PR #1272, opened 2017-04-12
Fix typo: connection -> connecting

justinj:
I notice that there's a little inconsistency in the use of `log.Fatal` vs. `log.Fatalf`, not a big deal but might be something worth standardizing in the docs on a larger scale at some point?

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1272)

### DIFF
--- a/_includes/app/basic-sample.go
+++ b/_includes/app/basic-sample.go
@@ -12,7 +12,7 @@ func main() {
 	// Connect to the "bank" database.
 	db, err := sql.Open("postgres", "postgresql://maxroach@localhost:26257/bank?sslmode=disable")
 	if err != nil {
-		log.Fatalf("error connection to the database: %s", err)
+		log.Fatalf("error connecting to the database: %s", err)
 	}
 
 	// Create the "accounts" table.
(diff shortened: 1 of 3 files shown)


sploiselle:
Glad to accept this change, @justinj ; can you give me a `TL;DR` on the differences between `log.Fatal` and `log.Fatalf`?

justinj:
No difference in this case! `log.Fatal` vs. `log.Fatalf` is the same as the distinction between `log.Print` and `log.Printf`, `log.Print` prints out each of its arguments as strings and `log.Printf` takes a format string and a bunch of values and inserts the values into the format string. So in this particular case, since the `%s` is at the end of the format string it's the same either way. You might use `log.Printf` if you wanted to insert a value into the middle of a string (`log.Printf("we got %s as the error", err)` is a contrived example, the `%s` gets replaced with the string representation of `err`).

Being consistent with them is certainly not a big deal at all and I only noticed that there was a difference because I saw these three lines all at once, I probably wouldn't have otherwise.

PR #1286, opened 2017-04-12
making log.Fatal references consistent

sploiselle:
Fixing issue mentioned in #1272

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1286)


justinj:
You want to get rid of the `%s` in the calls to `log.Fatal`, otherwise they'll show up in the output:
```
log.Fatal("error connecting to the database: %s", err)
```
will output as
```
error connecting to the database: %s <whatever the error was>
```
so you want to remove it to be just
```
log.Fatal("error connecting to the database: ", err)
```

sploiselle:
@justinj Fixed that. Thanks for explaining!

PR #1320, opened 2017-04-21
Prevent popup from opening on mobile

### LINKED ISSUE #1308: Stop displaying the "sign up for our newsletter" popup on phones
@jseldess & @sploiselle 

Having now tested the docs site on a phone multiple times in the last few days, I've noticed how annoying the newsletter modal is. The x to close out is pretty small, and Safari on iOS does not allow third party cookies, so the "do not show again" setting is moot. 

What do you think of setting it to not display on phones?

deronjohnson:
Fixes #1308

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1320)

### DIFF
--- a/_includes/newsletter-popout.html
+++ b/_includes/newsletter-popout.html
@@ -65,7 +65,7 @@
 
 		$(window).scroll(function() {
 			var cookie = getCookie('popout');
-			if(!cookie){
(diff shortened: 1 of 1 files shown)


jseldess:
Thanks, @deronjohnson. LGTM.

PR #1322, opened 2017-04-22
Document operator operation and precedence

mjibson:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1322)

### DIFF
--- a/functions-and-operators.md
+++ b/functions-and-operators.md
@@ -14,7 +14,7 @@ toc: true
 
 ## Operators
 
-The following table lists all CockroachDB operators from highest to lowest precedence, i.e., the order in which they will be evaluated within a statement.
+The following table lists all CockroachDB operators from highest to lowest precedence, i.e., the order in which they will be evaluated within a statement. Operators with the same precedence are left associative. This means that those operators are grouped together starting from the left and moving right.
 
 | Order of Precedence | Operator | Name | Operator Arity |
 | --- |
@@ -54,10 +54,6 @@ The following table lists all CockroachDB operators from highest to lowest prece
 | 15 | `NOT` | Logical NOT | unary |
 | 16 | `AND` | Logical AND | binary |
 | 17 | `OR` | Logical OR | binary |
-
-Operators with the same precedence are left associative. This means that those operators are grouped together starting from the left and moving right.
-
-Comparison operators (including `LIKE`, `~`, and `IS`) are non associative.
 
 ### Supported Operations
 
--- a/functions-and-operators.md
+++ b/functions-and-operators.md
@@ -14,6 +14,53 @@ toc: true
 
 ## Operators
 
+The following table lists all CockroachDB operators from highest to lowest precedence, i.e., the order in which they will be evaluated within a statement.
+
+| Order of Precedence | Operator | Name | Operator Arity |
+| --- |
+| 1 | `.` | Member field access operator | binary |
+| 2 | `::` | Type cast | binary |
+| 3 | `-` | Unary minus | unary |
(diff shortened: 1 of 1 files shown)


mjibson:
See: https://cloud.google.com/spanner/docs/functions-and-operators#operators

sploiselle:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[functions-and-operators.md, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1322#-KiXrUe05jeqscrKOIK0:-KiXrUe05jeqscrKOIK1:b-34hubr) ([raw file](https://github.com/cockroachdb/docs/blob/9ae0bb5e68648949c9ba9f9344917479dca09730/functions-and-operators.md#L58)):*
> ```Markdown
> | 17 | `OR` | Logical OR | binary |
> 
> Operators with the same precedence are left associative. This means that those operators are grouped together starting from the left and moving right.
> ```

Can we move this and the following line before the table?  Scrolling through it I was confused as to what operators with the same precedence meant and didn't know if the information was going to be in this document.

---

*[functions-and-operators.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1322#-KiXs7Ow4gr51nIvPab5:-KiXs7Ow4gr51nIvPab6:b-uwiwp5) ([raw file](https://github.com/cockroachdb/docs/blob/9ae0bb5e68648949c9ba9f9344917479dca09730/functions-and-operators.md#L60)):*
> ```Markdown
> Operators with the same precedence are left associative. This means that those operators are grouped together starting from the left and moving right.
> 
> Comparison operators (including `LIKE`, `~`, and `IS`) are non associative.
> ```

Just like you explain what "left associative" means, can I get a quick explanation of what "non associative" means? I don't quite understand its meaning in the Spanner docs, either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1322)*

mjibson:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[functions-and-operators.md, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1322#-KiXrUe05jeqscrKOIK0:-KiYL4A_N5YGZBI4UT6l:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/9ae0bb5e68648949c9ba9f9344917479dca09730/functions-and-operators.md#L58)):*

Can we move this and the following line before the table?  Scrolling through it I was confused as to what operators with the same precedence meant and didn't know if the information was going to be in this document.
</blockquote></details>

Done.

---

*[functions-and-operators.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1322#-KiXs7Ow4gr51nIvPab5:-KiYL4L2TcWVc6FR3Kqr:b-n322xt) ([raw file](https://github.com/cockroachdb/docs/blob/9ae0bb5e68648949c9ba9f9344917479dca09730/functions-and-operators.md#L60)):*

Just like you explain what "left associative" means, can I get a quick explanation of what "non associative" means? I don't quite understand its meaning in the Spanner docs, either.
</blockquote></details>

I've taken out the non associative stuff. It means that you can't do:

```
select true < true < true
```

even though you can do

```
select (true < true) < true
```

so it forces users to fully declare how the operators associate. I think we can get by without documenting it since no one writes "true < true < true" or similar things.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1322)*

mjibson:
Ping.

sploiselle:
LGTM

PR #1328, opened 2017-04-24
add porting postgres page

### LINKED ISSUE #1327: document postgres differences
Current known:

- float overflow (https://github.com/cockroachdb/cockroach/issues/15255)
- precedence of unary `~`
- int division (https://github.com/cockroachdb/cockroach/issues/14940)
- shift modulo (https://github.com/cockroachdb/cockroach/issues/14874)
- precedence of bitwise ops (https://github.com/cockroachdb/cockroach/pull/14944)

mjibson:
Fixes #1327

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1328)

### DIFF
--- a/_data/sidebar_doc.yml
+++ b/_data/sidebar_doc.yml
@@ -366,12 +366,12 @@ entries:
 
         - title: Information Schema
           url: /information-schema.html
-          thirdlevel:
 
+          thirdlevel:
           - title: Porting Applications
             thirdlevelitems:
 
-              - title: Postgres
+              - title: From PostgreSQL
                 url: /porting-postgres.html
 
     - title: Deploy
--- a/_data/sidebar_doc.yml
+++ b/_data/sidebar_doc.yml
@@ -366,6 +366,13 @@ entries:
 
         - title: Information Schema
           url: /information-schema.html
+          thirdlevel:
+
+          - title: Porting Applications
+            thirdlevelitems:
+
+              - title: Postgres
+                url: /porting-postgres.html
 
     - title: Deploy
       items:
--- a/porting-postgres.md
+++ b/porting-postgres.md
@@ -8,6 +8,8 @@ Although CockroachDB supports PostgreSQL syntax and drivers, it does not offer e
 
 Note that some of these differences only apply to rare inputs, and so no change will be needed, even if the listed feature is being used. In these cases, it is safe to ignore the porting instructions.
 
+{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting application that use [SQL features CockroachDB does not currently support](/sql-feature-support.html), such as arrays or the `ENUM` type.{{site.data.alerts.end}}
+
 <div id="toc"></div>
 
 ### Overflow of `float`
@@ -55,8 +57,6 @@ The unary `~` (bitwise not) operator in PostgreSQL has a low precedence. For exa
 SELECT ~1 + 2
 ```
 
-Is parsed as `~ (1 + 2)` because `~` has a lower precedence than `+`.
-
 In CockroachDB, unary `~` has the same (high) precedence as unary `-`, so the above expression will be parsed as `(~1) + 2`.
 
 **Porting instructions:** Manually add parentheses around expressions that depend on the PostgreSQL behavior.
@@ -69,7 +69,7 @@ The operators `|` (bitwise OR), `#` (bitwise XOR), and `&` (bitwise AND) in Post
 
 ### Integer division
 
-Division of integers in PostgreSQL results in an integer. For example, the following query returns `1`, since the `1 / 2` is truncated to `0` since it is performing integer division:
+Division of integers in PostgreSQL results in an integer. For example, the following query returns `1`, since the `1 / 2` is truncated to `0`:
 
 ```
 SELECT 1 + 1 / 2
--- a/porting-postgres.md
+++ b/porting-postgres.md
@@ -8,7 +8,7 @@ Although CockroachDB supports PostgreSQL syntax and drivers, it does not offer e
 
 Note that some of these differences only apply to rare inputs, and so no change will be needed, even if the listed feature is being used. In these cases, it is safe to ignore the porting instructions.
 
-{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting applications that use [SQL features CockroachDB does not currently support](/sql-feature-support.html), such as arrays or the `ENUM` type.{{site.data.alerts.end}}
+{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting applications that use <a href="/sql-feature-support.html">SQL features CockroachDB does not currently support</a>, such as arrays or the `ENUM` type.{{site.data.alerts.end}}
 
 <div id="toc"></div>
 
--- a/porting-postgres.md
+++ b/porting-postgres.md
@@ -8,7 +8,7 @@ Although CockroachDB supports PostgreSQL syntax and drivers, it does not offer e
 
 Note that some of these differences only apply to rare inputs, and so no change will be needed, even if the listed feature is being used. In these cases, it is safe to ignore the porting instructions.
 
-{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting application that use [SQL features CockroachDB does not currently support](/sql-feature-support.html), such as arrays or the `ENUM` type.{{site.data.alerts.end}}
+{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting applications that use [SQL features CockroachDB does not currently support](/sql-feature-support.html), such as arrays or the `ENUM` type.{{site.data.alerts.end}}
 
 <div id="toc"></div>
 
--- a/porting-postgres.md
+++ b/porting-postgres.md
@@ -4,7 +4,7 @@ summary: Porting an application from PostgreSQL
 toc: false
 ---
 
-Although CockroachDB supports Postgres syntax and drivers, it does not offer exact compatibility. This page documents the known list of differences between Postgres and CockroachDB for identical input. That is, a SQL statement of the type listed here will behave differently than in Postgres. Porting an existing application to CockroachDB will require changing these expressions.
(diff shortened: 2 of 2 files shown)


_data/sidebar_doc.yml
```diff
@@ -366,6 +366,13 @@ entries:
 
         - title: Information Schema
           url: /information-schema.html
+          thirdlevel:
```
jseldess: For consistency with the rest of the file, would you mind formatting this as follows?

```
thirdlevel:
  - title: Porting Applications
    thirdlevelitems:

      - title: From PostgreSQL
        url: /porting-postgres.html
```

_data/sidebar_doc.yml
```diff
@@ -366,6 +366,13 @@ entries:
...
+
+          - title: Porting Applications
+            thirdlevelitems:
+
+              - title: Postgres
```
jseldess: I think it'd be clearer to use the title `From PostgreSQL` here.

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Porting from PostgreSQL
+summary: Porting an application from PostgreSQL
+toc: false
+---
+
+Although CockroachDB supports Postgres syntax and drivers, it does not offer exact compatibility. This page documents the known list of differences between Postgres and CockroachDB for identical input. That is, a SQL statement of the type listed here will behave differently than in Postgres. Porting an existing application to CockroachDB will require changing these expressions.
```
jseldess: Let's use `PostgreSQL` throughout instead of `Postgres`.

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+The `float` type in Postgres will return an error if it overflows or an expression would return Infinity:
+
+```
+postgres=# select 1e300::float * 1e10::float;
+ERROR:  value out of range: overflow
+postgres=#  select pow(0::float, -1::float);
+ERROR:  zero raised to a negative power is undefined
+```
+
+In CockroachDB these expressions will instead return Infinity:
```
jseldess: Add comma after `In CockroachDB`

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+root@:26257/> select 1e300::float * 1e10::float;
```
jseldess: For the sql prompt, let's use `>`. Also, can you capitalize the keywords and break commands and responses into separate code blocks, where the commands have `sql` syntax highlighting and the responses have node?

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
++---------------------------+
+| +Inf                      |
++---------------------------+
+```
+
+### Precedence of unary `~`
+
+The unary `~` (bitwise not) operator in Postgres has a low precedence. For example:
+
+```
```
jseldess: Rather than break the sentence with the code block, I'd rewrite as:

> For example, the following query is parsed as `~ (1 + 2)` because `~` has a lower precedence than `+`:
> ```
> SELECT ~1 + 2`

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+In CockroachDB, unary `~` has the same (high) precedence as unary `-`, so the above expression will be parsed as `(~1) + 2`.
+
+**Porting instructions:** manually add parentheses around expressions that depend on the Postgres behavior.
+
+### Precedence of bitwise operators
+
+The operators `|` (bitwise OR), `#` (bitwise XOR), and `&` (bitwise AND) in Postgres all have the same precedence. In CockroachDB the precedence from highest to lowest is: `&`, `#`, `|`.
+
+**Porting instructions:** manually add parentheses around expressions that depend on the Postgres behavior.
```
jseldess: Capitalize `Manually`

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Integer division
+
+Division of integers in Postgres results in an integer. The query:
```
jseldess: Again, I'd rewrite this so the code block doesn't interrupt the sentence:

> For example, the following query returns `1`, since the `1 / 2` is truncated to `0` since it is performing integer division:
>```
>SELECT 1 + 1 / 2
>```
>CockroachDB instead provides the `//` operator to perform floor division.

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+SELECT ~1 + 2
+```
+
+Is parsed as `~ (1 + 2)` because `~` has a lower precedence than `+`.
+
+In CockroachDB, unary `~` has the same (high) precedence as unary `-`, so the above expression will be parsed as `(~1) + 2`.
+
+**Porting instructions:** manually add parentheses around expressions that depend on the Postgres behavior.
```
jseldess: Capitalize `Manually`

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+Division of integers in Postgres results in an integer. The query:
+
+```
+SELECT 1 + 1 / 2
+```
+
+Returns `1`, since the `1 / 2` is truncated to `0` since it is performing integer division. In CockroachDB integer division results in a `decimal`. CockroachDB instead provides the `//` operator to perform floor division.
+
+**Porting instructions:** change `/` to `//` in integer division where the result must be an integer.
```
jseldess: Capitalize `Change`

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Shift argument modulo
+
+The shift operators in Postgres (`<<`, `>>`) sometimes modulo their second argument to the bit size of the underlying type. For example:
```
jseldess: Again, revise so the sample doesn't interrupt the sentence:

>For example, the following query results in a `1` because the `int` type is 32 bits, and `32 % 32` is `0`, so this is the equivalent of `1 << 0`:
>```
>SELECT 1::int << 32
>```
>In CockroachDB, no such modulo is performed.

porting-postgres.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+SELECT 1::int << 32
+```
+
+Results in a `1` because the `int` type is 32 bits, and `32 % 32` is `0`, so this is the equivalent of `1 << 0`. In CockroachDB, no such modulo is performed.
+
+**Porting instructions:** manually add a modulo to the second argument. Also note that CockroachDB's int type is always 64 bits. For example:
```
jseldess: Capitalize `Manually`. Also, let's capitalize `INT` and link to our docs: 
```
[`INT`](int.html)
```

_data/sidebar_doc.yml
mjibson: done

_data/sidebar_doc.yml
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
mjibson: done

porting-postgres.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Precedence of bitwise operators
+
+The operators `|` (bitwise OR), `#` (bitwise XOR), and `&` (bitwise AND) in PostgreSQL all have the same precedence. In CockroachDB the precedence from highest to lowest is: `&`, `#`, `|`.
+
+**Porting instructions:** Manually add parentheses around expressions that depend on the PostgreSQL behavior.
+
+### Integer division
+
+Division of integers in PostgreSQL results in an integer. For example, the following query returns `1`, since the `1 / 2` is truncated to `0` since it is performing integer division:
```
jseldess: I'd actually remove `since it is performing integer division`. Feels redundant and makes the sentence a bit convoluted.

sploiselle:
I'd recommend mentioning that this guide primarily focuses on expressions and does not demonstrate how to port applications that use unsupported SQL features (e.g. arrays, enum).

jseldess:
Good point, @sploiselle. We may want to expand this page to cover a lot more, but for now, totally agree that it's best to be clear about the focus.

sploiselle:
Review status: 0 of 2 files reviewed at latest revision, 15 unresolved discussions.

---

*[porting-postgres.md, line 9 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1328#-Ki_yHiYY3Yn4JmCwrTZ:-Ki_yHiYY3Yn4JmCwrT_:b-rdd93d) ([raw file](https://github.com/cockroachdb/docs/blob/a420b84997412001498d203c2d394f57568b807d/porting-postgres.md#L9)):*
> ```Markdown
> Although CockroachDB supports PostgreSQL syntax and drivers, it does not offer exact compatibility. This page documents the known list of differences between PostgreSQL and CockroachDB for identical input. That is, a SQL statement of the type listed here will behave differently than in PostgreSQL. Porting an existing application to CockroachDB will require changing these expressions.
> 
> Note that some of these differences only apply to rare inputs, and so no change will be needed, even if the listed feature is being used. In these cases, it is safe to ignore the porting instructions.
> ```

After this line, recommend adding something like:

{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting application that use <a href="sql-feature-support.html">SQL features CockroachDB does not currently support</a>, such as arrays or the <code>ENUM</code> type.{{site.data.alerts.end}}

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1328)*

mjibson:
I've made those changes. RFAL.

porting-postgres.md
```diff
@@ -8,6 +8,8 @@ Although CockroachDB supports PostgreSQL syntax and drivers, it does not offer e
 
 Note that some of these differences only apply to rare inputs, and so no change will be needed, even if the listed feature is being used. In these cases, it is safe to ignore the porting instructions.
 
+{{site.data.alerts.callout_info}}This document currently only covers how to rewrite SQL expressions. It does not discuss strategies for porting application that use [SQL features CockroachDB does not currently support](/sql-feature-support.html), such as arrays or the `ENUM` type.{{site.data.alerts.end}}
```
jseldess: `porting applications` - needs to be plural

Issue #1319, opened 2017-04-21
security: certificate rotation


jseldess:
RFC: https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/certificate_rotation.md
PR: https://github.com/cockroachdb/cockroach/pull/14925

jseldess:
@mberhault, will need your help with this next week.

mberhault:
Please see PR #1331 for a draft doc about certificate rotation.

The actual reload of certificates is fairly easy and short.
However, I spend some time detailing the process to renew all certificates at about the same time which can cause issues if done improperly.

jseldess:
Closing in favor of https://github.com/cockroachdb/docs/issues/1499.

PR #1331, opened 2017-04-25
draft: explain certificate rotation

mberhault:
Part of #1319

This is maybe a bit more verbose than it needs to be.
I try to give some guidance on how to coordinate a full reload (CA cert
as well as other certs).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1331)

### DIFF
--- a/certificate-rotation.md
+++ b/certificate-rotation.md
@@ -1,0 +1,113 @@
+There are two parts to certificate rotation: generating the new certificates, and reloading them.
+
+Both have been improved, with the main difference being that we can now reload certificates without restarting a node.
+
+## Order of certificate regeneration
+
+Most of the complication in rotating certificates is proper handling of new CA certificates.
+It's important for all parts of the system (nodes and clients) to have the same CA certificates, or they may not be able to verify each other's client/node certificates.
+
+It is possible to have multiple CA certificates active at the same time, but not node or client certificates. Therefore, CA certificates should be generated and pushed first, followed by new node/client certificates using the new CA.
+
+Let's say we want to regenerate all certificates (CA, client, node), we should do the following:
+1. generate new CA certificate and combine it with the existing one
+1. push combined CA certificate to all nodes and reload certificates
+1. make combined CA available to all clients. We may need to wait some time (days/weeks) for all clients to make use of the new certificate
+1. generate and push new node/client certificates using the new CA certificate
+
+It is recommended to change CA certificates on a completely different schedule. eg: months before.
+**WARNING**: node certificates generated with the new CA will not be accepted by clients that do not yet have the new CA. Before all clients have the new CA cert, node certificates should be signed with the old CA (this is still a [TODO](https://github.com/cockroachdb/cockroach/blob/master/pkg/cli/cert.go#L91-L93))
+
+## Generating new CA certificates
+
+CA rotation includes generating a new CA certificate and key.
+
+If the original certificate and key are generated using:
+```
+$ cockroach cert create-ca --certs-dir=certs --ca-key=ca.key --logtostderr=true
+I170425 14:20:29.394473 1 security/certs.go:120  Generated CA key ca.key
+I170425 14:20:29.398081 1 security/certs.go:173  Wrote 1 certificates to certs/ca.crt
+```
+
+We can generate a new CA certificate and key with the following:
+```
+$ cockroach cert create-ca --certs-dir=certs --ca-key=ca.new.key --overwrite --logtostderr=true
+I170425 14:21:10.490341 1 security/certs.go:120  Generated CA key ca.new.key
+I170425 14:21:10.494007 1 security/certs.go:160  Found 1 certificates in certs/ca.crt
+I170425 14:21:10.494341 1 security/certs.go:173  Wrote 2 certificates to certs/ca.crt
+```
+
+Flags:
+* `--ca-key=ca.new.key`: path to the new key to generate. To re-use the existing key (not recommended), use `--ca-key=ca.key` and `--allow-key-reuse`.
+* `--overwrite`: add the new certificate to the existing `ca.crt` file.
+
+This results in a `ca.crt` containing the new certificate followed by the old certificate. This
+combined CA certificate should be used by all nodes and clients to verify old node/client certificates
+and well as new certificates.
+
+## Generating new node/client certificates
+
+Client and node certificates are simpler than CA certificates because we do not need to keep old
+certificates around.
+
+Generating a new certificate and key is the same as the original generation. If the files already exist, specify `--overwrite` to overwrite both certificate and key.
+
+If the original certificate and key for node `localhost` were created with:
+```
+$ cockroach cert create-node --certs-dir=certs --ca-key=ca.key --logtostderr=true localhost
+I170425 14:28:10.713448 1 security/certs.go:226  Generated node certificate: certs/node.crt
+I170425 14:28:10.713845 1 security/certs.go:232  Generated node key: certs/node.key
+```
+
+We can either delete the `node.crt` and `node.key` files and rerun the same command, or run:
+```
+roach cert create-node --certs-dir=certs --ca-key=ca.key --logtostderr=true --overwrite localhost
+I170425 14:28:45.839331 1 security/certs.go:226  Generated node certificate: certs/node.crt
+I170425 14:28:45.839556 1 security/certs.go:232  Generated node key: certs/node.key
+```
+
+## Deploying certificates
+
+Cockroach does not currently provide any tools to deploy certificates and keys to clients and nodes.
+The appropriate files should be copied to the hosts needing them. Specifically:
+
+* each node needs: `ca.crt`, `node.crt`, `node.key`, with the node certificate issued for its hostnames and IPs
+* a client with username `marc` needs: `ca.crt`, `client.marc.crt`, `client.marc.key`.
+
+When generated a new CA certificate and adding it to the existing file, you **must** push the newly 
+generated file containing both certificates.
+
+Node and client certificate and keys can be replaced entirely as they are re-generated.
+The key file must be the one generated at the same time as the corresponding certificate.
+
+## Reloading certificates on a node
+
+Certificates are loaded automatically when a node starts.
+It is also possible to tell a running node to re-scan the certificates directory and use the
+new certificates if found.
+
+This is done by issuing a `SIGHUP` signal to the cockroach process:
+```
+pkill -SIGHUP -x cockroach
+```
+
+Flags:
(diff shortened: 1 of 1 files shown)


certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+
+It is possible to have multiple CA certificates active at the same time, but not node or client certificates. Therefore, CA certificates should be generated and pushed first, followed by new node/client certificates using the new CA.
+
+Let's say we want to regenerate all certificates (CA, client, node), we should do the following:
+1. generate new CA certificate and combine it with the existing one
+1. push combined CA certificate to all nodes and reload certificates
+1. make combined CA available to all clients. We may need to wait some time (days/weeks) for all clients to make use of the new certificate
+1. generate and push new node/client certificates using the new CA certificate
+
+It is recommended to change CA certificates on a completely different schedule. eg: months before.
```
jseldess: Just to make sure I understand: You're saying it important to change CA certs a while before node/client certs?

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Deploying certificates
+
+Cockroach does not currently provide any tools to deploy certificates and keys to clients and nodes.
+The appropriate files should be copied to the hosts needing them. Specifically:
+
+* each node needs: `ca.crt`, `node.crt`, `node.key`, with the node certificate issued for its hostnames and IPs
+* a client with username `marc` needs: `ca.crt`, `client.marc.crt`, `client.marc.key`.
+
+When generated a new CA certificate and adding it to the existing file, you **must** push the newly
```
jseldess: This requirement is because the exiting node/client certs depend on the old CA cert, right? We should probably spell this out more.

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+It is possible to have multiple CA certificates active at the same time, but not node or client certificates. Therefore, CA certificates should be generated and pushed first, followed by new node/client certificates using the new CA.
+
+Let's say we want to regenerate all certificates (CA, client, node), we should do the following:
+1. generate new CA certificate and combine it with the existing one
+1. push combined CA certificate to all nodes and reload certificates
+1. make combined CA available to all clients. We may need to wait some time (days/weeks) for all clients to make use of the new certificate
+1. generate and push new node/client certificates using the new CA certificate
+
+It is recommended to change CA certificates on a completely different schedule. eg: months before.
+**WARNING**: node certificates generated with the new CA will not be accepted by clients that do not yet have the new CA. Before all clients have the new CA cert, node certificates should be signed with the old CA (this is still a [TODO](https://github.com/cockroachdb/cockroach/blob/master/pkg/cli/cert.go#L91-L93))
```
jseldess: Are we going to handle that TODO before 1.0? If not, we might want to delay these docs.

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Order of certificate regeneration
+
+Most of the complication in rotating certificates is proper handling of new CA certificates.
+It's important for all parts of the system (nodes and clients) to have the same CA certificates, or they may not be able to verify each other's client/node certificates.
+
+It is possible to have multiple CA certificates active at the same time, but not node or client certificates. Therefore, CA certificates should be generated and pushed first, followed by new node/client certificates using the new CA.
+
+Let's say we want to regenerate all certificates (CA, client, node), we should do the following:
+1. generate new CA certificate and combine it with the existing one
```
mberhault: this is detailed a bit more around L44:
```
This results in a `ca.crt` containing the new certificate followed by the old certificate. This
 combined CA certificate should be used by all nodes and clients to verify old node/client certificates
 and well as new certificates.
```

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+1. push combined CA certificate to all nodes and reload certificates
+1. make combined CA available to all clients. We may need to wait some time (days/weeks) for all clients to make use of the new certificate
```
mberhault: When we generate new node certificates (eg: adding a new node, or renewing the certificate on an old node), they will be signed by the new CA. This means that the CA needs to be on clients so that they can validate the new node cert.

certificate-rotation.md
mberhault: yup. Ideally, we want everyone to have the same CA certs. It's easy for nodes since we control that and can push the CA cert to all of them fairly quickly. We may not control how clients access the CA though. If it's all from a shared location, that may be easy, but even then the clients will need to reload the certificate (usually meaning a restart)

certificate-rotation.md
mberhault: probably not, but I may be able to get to it. Unfortunately, there's no great way of specifying this without making `cockroach cert` interfactive (bleh)

certificate-rotation.md
mberhault: that's correct.

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
+There are two parts to certificate rotation: generating the new certificates, and reloading them.
```
mberhault: there are generally three reasons to rotate certificates and keys:
* they're expiring soon so you're forced to (should not occur too often with out default of 10 years)
* you have a policy of frequent rotation for security reasons. if you have good PKI infrastructure, this may mean very frequent rotation
* some keys were compromised. rotation is probably not enough here, it might be necessary to just delete all old certs and generate completely new ones. but it could be just one client or node key, in which case rotating might be sufficient.

mberhault:
One thing to note: if we're not doing CA rotation, we can do node and client cert rotation whenever we like.
It's only when doing CA rotation that we have to be careful and make sure both new and old CAs are available to all nodes and clients.

certificate-rotation.md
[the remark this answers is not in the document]
Amruta-Ranade: @mberhault The new node certificates will be signed by the combined CA, right? So effectively, they will be signed by old as well as new CA. So clients should be able to validate new node cert, even if they have only old CA cert, right?

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+We can generate a new CA certificate and key with the following:
+```
+$ cockroach cert create-ca --certs-dir=certs --ca-key=ca.new.key --overwrite --logtostderr=true
+I170425 14:21:10.490341 1 security/certs.go:120  Generated CA key ca.new.key
+I170425 14:21:10.494007 1 security/certs.go:160  Found 1 certificates in certs/ca.crt
+I170425 14:21:10.494341 1 security/certs.go:173  Wrote 2 certificates to certs/ca.crt
+```
+
+Flags:
+* `--ca-key=ca.new.key`: path to the new key to generate. To re-use the existing key (not recommended), use `--ca-key=ca.key` and `--allow-key-reuse`.
```
Amruta-Ranade: @mberhault In which scenario would we need to reuse an existing key?

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+* `--overwrite`: add the new certificate to the existing `ca.crt` file.
```
Amruta-Ranade: @mberhault In the [Create Security Certificates](https://www.cockroachlabs.com/docs/stable/create-security-certificates.html#create-the-ca-certificate-and-key-pair) doc, we describe the --overwrite flag as "When running create-* subcommands, pass this flag to allow existing files in the certificate directory (--certs-dir) to be overwritten." 

Has the flag behavior been modified? Does the existing description need to be updated with this new description?

certificate-rotation.md
Amruta-Ranade: @mberhault Same question as before: If the combined CA has old CA as well, will this issue still occur?

Also, just to clarify, when we say "node certs should be signed with old CA", we mean use old CA key to generate new node cert and upload old CA cert to node, right?

certificate-rotation.md
```diff
@@ -0,0 +1,113 @@
...
+If the original certificate and key for node `localhost` were created with:
+```
+$ cockroach cert create-node --certs-dir=certs --ca-key=ca.key --logtostderr=true localhost
+I170425 14:28:10.713448 1 security/certs.go:226  Generated node certificate: certs/node.crt
+I170425 14:28:10.713845 1 security/certs.go:232  Generated node key: certs/node.key
+```
+
+We can either delete the `node.crt` and `node.key` files and rerun the same command, or run:
+```
+roach cert create-node --certs-dir=certs --ca-key=ca.key --logtostderr=true --overwrite localhost
```
Amruta-Ranade: @mberhault Shouldn't "ca.key" be replaced with "ca.new.key"?

certificate-rotation.md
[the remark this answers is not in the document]
mberhault: Certificates can only be signed by one CA. We include the old CA to ensure that old certificates can still be verified.
However, verifying newer certificates will require the new CA.
The reason we combine them is that we have a single file for the CA certificate. If we coordinate things properly (probably with a synchronized restart of servers and clients), we could drop the old CA entirely. However, it's operationally easier to accept both the old and the new certificates. Thus the combined CA certificate.

certificate-rotation.md
mberhault: Similar answer as before: if the client only has the old CA, it will not be able to verify certificates signed with the new CA.
Think of the combined CA certificate as two completely separate certificate/key pairs that are joined together. If we supported it, we could use two completely different files instead of one file containing both CAs.

certificate-rotation.md
mberhault: Reusing the old key is possible, just not recommended. For security reasons, it is considered good practice to rotate keys on a regular basis (that's the sensitive part). Someone could ignore the recommendation entirely for any reason they like (one less file to keep and distribute)

certificate-rotation.md
mberhault: It's similar, it just means that it can modify existing files. In the case of keys and node/client certificates, it means that it allows the cli to overwrite the files. In the case of the CA certificate, it allows modification as well (it's just not an overwrite, it adds the certificate instead)

certificate-rotation.md
mberhault: If we had renewed the CA certificate and key, then yes. However, CA certs and node/client certs can be regenerated on a different schedule (they have different expiration dates, with CA certs expiring much later).
For the purposes of the tutorial/docs, it may be better to change the CA key generation to:
1. rename the old CA key to `ca.old.key`
1. generate the new CA certificate and key: `ca.crt` and `ca.key`.

This way, the node/client certificate generation works regardless of whether we regenerated the CA.

jseldess:
Closing in favor of https://github.com/cockroachdb/docs/pull/1932. Thanks for all the help, @mberhault!

PR #1334, opened 2017-04-26
Recommend using multiple tiers of localities as a best practice

a-robinson:
This can really help out lease rebalancing in situations where
some datacenters are much closer together than others, as explained in
https://github.com/cockroachdb/cockroach/issues/15369

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1334)


sploiselle:
Thank you! This is a wonderful enhancement. I think Jesse has more insight into these docs, but I have some small suggestions.

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[configure-replication-zones.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1334#-Kig0tcJYtzXCawEkt1E:-Kig0tcJYtzXCawEkt1F:bjgo9nd) ([raw file](https://github.com/cockroachdb/docs/blob/10d315d20b47c02c0a34b1c76a316810962bbb79/configure-replication-zones.md#L61)):*
> ```Markdown
> Attribute Type | Description
> ---------------|------------
> **Node Locality** | Using the `--locality` flag, you can assign arbitrary key-value pairs that describe the locality of the node. Locality might include country, region, datacenter, rack, etc. The pairs should be ordered from most inclusive to lease inclusive, meaning that a country locality should be specified before datacenter, which should in turn be specified before rack.<br><br>CockroachDB attempts to spread replicas evenly across the cluster based on locality, with the order determining the priority. The keys themselves and the order of key-value pairs must be the same on all nodes, and it's typically better to include more pairs than fewer. For example:<br><br>`--locality=region=east,datacenter=us-east-1`<br>`--locality=region=east,datacenter=us-east-2`<br>`--locality=region=west,datacenter=us-west-1`
> ```

s/lease/least.

Also "least inclusive, meaning that a country locality..." might be a little more clear if it were split up into two sentences, like, "least inclusive. For example, a country locality...."

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1334#-Kig1USDpqKkvwRyuH2I:-Kig1USDpqKkvwRyuH2J:bgvk5hi) ([raw file](https://github.com/cockroachdb/docs/blob/10d315d20b47c02c0a34b1c76a316810962bbb79/recommended-production-settings.md#L22)):*
> ```Markdown
> - Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
> 
> - When replicating across datacenters, it's recommended to use datacenters on a single continent to ensure performance (inter-continent scenarios will improve in performance soon). Also, to ensure even replication across datacenters, it's recommended to specify which datacenter each node is in using the `--locality` flag (see this [example](configure-replication-zones.html#even-replication-across-datacenters) for more details). If some of your datacenters are much farther apart than others, [specifying multiple levels of locality is recommended](configure-replication-zones.html#descriptive-attributes-assigned-to-nodes).
> ```

"specifying multiple levels of locality (such as country and region)"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1334#-:-Kig1nkIddpmZmRPS-3_:b8cfw2u)*

a-robinson:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[configure-replication-zones.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1334#-Kig0tcJYtzXCawEkt1E:-Kig8-wWTr9bkRhlZY-R:b-po2n98) ([raw file](https://github.com/cockroachdb/docs/blob/10d315d20b47c02c0a34b1c76a316810962bbb79/configure-replication-zones.md#L61)):*

s/lease/least.

Also "least inclusive, meaning that a country locality..." might be a little more clear if it were split up into two sentences, like, "least inclusive. For example, a country locality...."
</blockquote></details>

Whoops, thanks. Done.

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1334#-Kig1USDpqKkvwRyuH2I:-Kig8EnCzdC6VB374QAj:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/10d315d20b47c02c0a34b1c76a316810962bbb79/recommended-production-settings.md#L22)):*

"specifying multiple levels of locality (such as country and region)"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1334)*

configure-replication-zones.md
```diff
@@ -58,7 +58,7 @@ When starting a node with the [`cockroach start`](start-a-node.html) command, yo
...
+**Node Locality** | Using the `--locality` flag, you can assign arbitrary key-value pairs that describe the locality of the node. Locality might include country, region, datacenter, rack, etc. The pairs should be ordered from most inclusive to least inclusive. For example, a country locality should be specified before datacenter, which should in turn be specified before rack.<br><br>CockroachDB attempts to spread replicas evenly across the cluster based on locality, with the order determining the priority. The keys themselves and the order of key-value pairs must be the same on all nodes, and it's typically better to include more pairs than fewer. For example:<br><br>`--locality=region=east,datacenter=us-east-1`<br>`--locality=region=east,datacenter=us-east-2`<br>`--locality=region=west,datacenter=us-west-1`
```
jseldess: I think this section might read better with some rearrangement. What do you think about:

> Using the `--locality` flag, you can assign arbitrary key-value pairs that describe the locality of the node. Locality might include country, region, datacenter, rack, etc. CockroachDB attempts to spread replicas evenly across the cluster based on locality.
>
>The key-value pairs should be ordered from most inclusive to least inclusive. For example, a country locality should be specified before datacenter, which should in turn be specified before rack. Also, the keys and the order of key-value pairs must be the same on all nodes, and it's typically better to include more pairs than fewer. For example...

a-robinson: Looks great! I've switched to that.

jseldess:
@a-robinson, what's your opinion: We don't need to do this immediately, but should we eventually add `--locality` as a best practice throughout our deployment docs, for example, [here](https://www.cockroachlabs.com/docs/manual-deployment.html#step-2-start-the-first-node)? Those guides already recommend reviewing the recommended production settings page, which has these details, but I wonder whether we should start pushing standard use of this feature.

a-robinson:
I'd lean toward leaving it off of pages like that. The system will work perfectly fine without specifying `--locality`, so I don't think it's worth cluttering up each individual deployment page with it. The recommended settings page seems like the right place.

jseldess:
OK, thanks, @a-robinson.

PR #1335, opened 2017-04-26
Release notes for 1.0-rc1

bdarnell:
As of cockroachdb/cockroach@18aaa37d38

This is still missing any mention of backup/restore and enterprise functionality, and awaiting resolution on whether distsql is making the cut.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1335)


1.0-rc.1.md
```diff
@@ -0,0 +1,101 @@
...
+  - `server.remote_debugging.mode`
+  - `sql.metrics.statement_details.dump_to_logs`
+  - `sql.metrics.statement_details.enabled`
+  - `sql.metrics.statement_details.threshold`
+  - `sql.trace.log_statement_execute`
+  - `sql.trace.session_eventlog.enabled`
+  - `sql.trace.txn.enable_threshold`
+- The `PREPARE` and `EXECUTE` statements can now be used to use prepared statements in a SQL shell. [#15256](https://github.com/cockroachdb/cockroach/pull/15256)
+- The `regexp_replace` function now handles newlines correctly, both with and without the `n` flag. [#15309](https://github.com/cockroachdb/cockroach/pull/15309)
+- Precedence for the binary `~` operator (regex match) has been changed. [#15264](https://github.com/cockroachdb/cockroach/pull/15264)
```
mjibson: May need to add: `This could cause DEFAULT or CHECK expressions to have different behavior if ~ was used in combination with other operators. These will need to be fixed manually.`

jseldess: Making this update, but @mjibson, doesn't this mean that we should list this under "Backwards-Incompatible Changes"?

mjibson: Yes, I think that's probably the most correct, although the chance that anyone will be affected by this change is near zero, so it's also fine to leave it here.

1.0-rc.1.md
```diff
@@ -0,0 +1,101 @@
...
+
+- CockroachDB now builds on Windows. Windows support is currently experimental and is recommended only for development, not production.
+- Building from source on macOS 10.11 works again. [#15168](https://github.com/cockroachdb/cockroach/pull/15168)
+- Building from source on older CPUs (without SSE4.2) works again. [#15247](https://github.com/cockroachdb/cockroach/pull/15247)
+
+### SQL Language Changes
+
+- New commands `SET CLUSTER SETTING` and `SHOW ALL CLUSTER SETTINGS` can now be used to change certain cluster-wide settings. The following cluster settings are available in this release:
+  - `diagnostics.reporting.enabled`
+  - `enterprise.enabled`
```
jseldess: Not sure if his change got in before the release candidate was picked, but @danhhz tells me that this setting is being removed. I'll remove it from the release notes, in any case.

bdarnell:
Added a new commit covering the latest updates to the RC, including a brief mention of distsql (which we may want to expand)

Issue #1182, opened 2017-03-13
Rolling upgrade (zero-downtime upgrades)


sploiselle:
- Conceptual doc for "upgrades" with notion of taking down one node at a time, replacing binary, and rejoining cluster.
- ASCII cast showing taking node down, upgrading binary, rejoining cluster
- Documentation for each environment/deployment because this is automatable.

jseldess:
Thanks for adding thisk @sploiselle . We definitely need it for 1.0.

jseldess:
First bullet is done. Tracking the second two in https://github.com/cockroachdb/docs/issues/1559.

PR #1338, opened 2017-04-26
manual rolling upgrade docs

sploiselle:
Work toward #1182 (not closing because this is only the manual process without any kind of ops-minded automation)

@mberhault or @bdarnell Do either of you have any guidance on improving this process or additional guidance we should include? e.g., if more than 5 minutes passes before rejoining the node to the cluster, wait for 5 more minutes to ensure you don't freeze the cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1338)

### DIFF
--- a/_data/sidebar_doc.yml
+++ b/_data/sidebar_doc.yml
@@ -462,6 +462,9 @@ entries:
         - title: View Version Details
           url: /view-version-details.html
 
+        - title: Upgrade a Cluster's Version
+          url: /upgrade-cockroach-version.html
+          
         - title: Back up Data
           url: /back-up-data.html
 
--- a/upgrade-cockroach-version.md
+++ b/upgrade-cockroach-version.md
@@ -1,0 +1,95 @@
+---
+title: Upgrade a Cluster's Version
+summary: Learn how to upgrade your CockroachDB cluster to a new version.
+toc: false
+toc_not_nested: true
+---
+
+Because of CockroachDB's multi-active availability design, you can perform a "rolling upgrade" of CockroachDB on your cluster. This means you can upgrade individual nodes in your cluster one at a time without any downtime for your cluster.
+
+However, there are a few things you should keep in mind:
+
+  - For rolling upgrades to work, your cluster must be behind a load balancer or your clients must be configured to talk to multiple nodes. If your application communicates with a single node, bringing it down to upgrade its CockroachDB binary will cause your application to fail when the node goes offline.
+
+  - We recommend creating scripts to upgrade CockroachDB and not attempting to do so by hand.
+
+  - Only bring down only one node at a time, and wait at least one minute after it rejoins the cluster to bring down the next node. This reduces the number of nodes you have offline at any point in time, minimizing the chance you'll lose a majority of the nodes in your cluster which can cause service outages.
+
+## Perform a Rolling Upgrade
+
+For each node in your cluster, complete the following steps.
+
+1. Connect to the node.
+
+2. Download and install the CockroachDB binary you want to use:
+
+   - **Mac**:
+
+     ~~~ shell
+     $ curl -O https://binaries.cockroachdb.com/cockroach-{{site.data.strings.version}}.darwin-10.9-amd64.tgz
+     $ tar xfz cockroach-{{site.data.strings.version}}.darwin-10.9-amd64.tgz
+
+     # Optional: Place cockroach in your $PATH
+     $ cp -i cockroach-{{site.data.strings.version}}.darwin-10.9-amd64/cockroach /usr/local/bin/cockroach
+     ~~~
+
+   - **Linux**:
+
+     ~~~ shell
+     $ wget https://binaries.cockroachdb.com/cockroach-{{site.data.strings.version}}.linux-amd64.tgz
+     $ tar xfz cockroach-{{site.data.strings.version}}.linux-amd64.tgz 
+
+     # Optional: Place cockroach in your $PATH
+     $ cp -i cockroach-{{site.data.strings.version}}.linux-amd64/cockroach /usr/local/bin/cockroach
+     ~~~
+
+3. If you're running with a process manager, have it stop `cockroach`.
+
+   Without a process manager, use this command:
+
+   ~~~ shell
+   $ pkill cockroach
+   ~~~
+
+4. If you use `cockroach` in your `$PATH`, rename the outdated `cockroach` binary, and then move the new one into its place:
+
+   - **Mac**:
+
+     ~~~ shell
+     $ i="$(which cockroach)"; mv "$i" "$i"_old
+     $ cp -i cockroach-{{site.data.strings.version}}.darwin-10.9-amd64/cockroach /usr/local/bin/cockroach
+     ~~~
+
+   - **Linux**:
+
+     ~~~ shell
+     $ i="$(which cockroach)"; mv "$i" "$i"_old
+     $ cp -i cockroach-{{site.data.strings.version}}.linux-amd64/cockroach /usr/local/bin/cockroach
+     ~~~
+
+   If you leave versioned binaries on your servers, you don't need to do anything.
+
+5. If you're running with a process manager, have the node rejoin the cluster by starting it.
+
+   Without a process manager, use this command:
+
+   ~~~ shell
+   $ cockroach start --join=[any other node's IP address] [other flags]
+   ~~~
+
+   `[other flags]` includes any flags you [use to a start node](start-a-node.html), such as its `--host`.
+
+6. Verify the node has rejoined the cluster through its output to `stdout` or through the [admin UI](explore-the-admin-ui.html).
+
+7. If you use `cockroach` in your `$PATH`, you can remove the old binary:
+
+   ~~~ shell
+   $ rm /usr/local/bin/cockroach_old
+   ~~~
+
+8. Wait at least one minute after the node has rejoined the cluster, and then repeat these steps for the next node.
+
+## See Also
+
+- [View Version Details](view-version-details.html)
+- [Release notes for our latest version]({{site.data.strings.version}}.html)


upgrade-cockroach.md
```diff
@@ -0,0 +1,69 @@
...
+## Manually Perform a Rolling Upgrade
+
+In advance of this procedure, you will need the IP addresses of all your cluster's nodes.
+
+1. Connect to the node.
+
+2. Stop `cockroach`:
+
+   ~~~ shell
+   $ cockroach quit [flags]
```
bdarnell: I'd recommend using `kill` instead of `cockroach quit` when you're running on the node itself (which you must be fore the next steps).

upgrade-cockroach.md
```diff
@@ -0,0 +1,69 @@
...
+     ~~~ shell
+     $ wget https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz
+     $ tar xfz cockroach-latest.linux-amd64.tgz
+     $ cp -i cockroach-latest.linux-amd64/cockroach "$i"
+     ~~~
+
+5. Rejoin the node to the cluster using any other node's IP address:
+
+   ~~~ shell
+   $ cockroach start --join=[any other node's IP address] [other flags]
```
bdarnell: Any serious deployment should be running under some sort of process manager - supervisord, systemd, etc. So in most cases this would be done via a command to that process manager.

upgrade-cockroach.md
```diff
@@ -0,0 +1,69 @@
...
+   ~~~
+
+   `[other flags]` includes any flags you [use to start nodes](start-a-node.html), such as its `--host`.
+
+6. Once you've confirmed that the node has rejoined the cluster, you can remove the old binary:
```
bdarnell: Mention that you should generally restart nodes no less than about a minute apart from each other.

upgrade-cockroach.md
```diff
@@ -0,0 +1,69 @@
...
+   ~~~ shell
+   $ cockroach quit [flags]
+   ~~~
+
+   `[flags]` includes the flags you [use to communicate with a node](stop-a-node.html), such as its `--host`.
+
+3. Rename the outdated `cockroach` binary:
+
+   ~~~ shell
+   $ i="$(which cockroach)"; mv "$i" "$i"_old
```
bdarnell: This looks kind of magical, and it may not work as written (because the binary may not be on `$PATH`, or it may not be in a writeable directory). I think it's better in a server context to use the versioned (not `latest` tarballs) and leave the binaries in their versioned directories instead of copying them onto `$PATH`. (but this is one of the hazards of providing this kind of manual instruction - everyone has their own opinions about the right way to do things).

upgrade-cockroach.md
```diff
@@ -0,0 +1,69 @@
...
+   ~~~
+
+4. Download and install the CockroachDB binary you want to use (this example uses the `latest` build for simplicity's sake):
```
bdarnell: I think we're planning to discontinue the `latest` tarballs; this should use an explicit version. (that was going to be less of a headache when we just had the one install page that would need to be updated)

upgrade-cockroach.md
sploiselle: Done.

upgrade-cockroach.md
sploiselle: Done.

upgrade-cockroach.md
sploiselle: Done.

upgrade-cockroach.md
sploiselle: Done-ish. Will follow up with @mberhault on this.

upgrade-cockroach.md
sploiselle: Made mention of this within the intro.

sploiselle:
@bdarnell or @mberhault Can I get a final review on this?

bdarnell:
LGTM

---

Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[upgrade-cockroach.md, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-KigeYC0bbfQNjsvmYfG-r1-32:-Kji5iGEl-LfSvYmMBtm:btaitt9) ([raw file](https://github.com/cockroachdb/docs/blob/4001bedcdfbad23b4b1637ae56447581383586bd/upgrade-cockroach.md#L32)):*

Done.
</blockquote></details>

The download part of this should come before stopping the old version. The only thing you want to do in between stopping the old process and starting the new one is copying the binary onto your path (if you're doing that) or pointing your process manager at the new binary.

---

*[upgrade-cockroach.md, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-KigeYC0bbfQNjsvmYfG-r1-58:-Kji6OkVIuI9rGYz6tFY:b3yotog) ([raw file](https://github.com/cockroachdb/docs/blob/4001bedcdfbad23b4b1637ae56447581383586bd/upgrade-cockroach.md#L58)):*

Made mention of this within the intro.
</blockquote></details>

The "one minute" comment above is not what I meant here. What I meant was that after restarting one node, wait one minute before stopping the next.

---

*[upgrade-cockroach.md, line 12 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5FochGB7GSt0Jdiw:-Kji5FochGB7GSt0Jdix:b-5v2t5z) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L12)):*
> ```Markdown
> However, there are a few things you should keep in mind:
> 
>   - For rolling upgrades to work, your cluster must be behind a load balancer. If your application communicates with a single node, bringing it down to upgrade its CockroachDB binary will cause your application to fail when the node goes offline.
> ```

Must be behind a load balancer *or* your clients must be configured to talk to multiple nodes. 

---

*[upgrade-cockroach.md, line 16 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5Lb_I8PAAMY9COoW:-Kji5Lb_I8PAAMY9COoX:b-yzsytj) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L16)):*
> ```Markdown
>   - We recommend creating scripts to upgrade CockroachDB and not attempting to do so by hand.
> 
>   - Only bring down only one node at a time, which should be restarted within one minute to prevent CockroachDB from entering a failover state.
> ```

I don't think there's anything significant about one minute (there's a threshold at five minutes)

---

*[upgrade-cockroach.md, line 43 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5n0DXnTSeppnnJQ7:-Kji5n0DXnTSeppnnJQ8:bgvg9zm) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L43)):*
> ```Markdown
> 
>      ~~~ shell
>      $ wget https://binaries.cockroachdb.com/cockroach-{{site.data.strings.version}}.darwin-10.9-amd64.tgz
> ```

`wget` is not installed by default on macs; this should use `curl` instead. (I don't know whether `wget` or `curl` is more likely to be present on linux)

---

*[upgrade-cockroach.md, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji619cZSCHohy-NloO:-Kji619cZSCHohy-NloP:b-bp3vas) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L47)):*
> ```Markdown
> 
>      # Optional: Place cockroach in your $PATH
>      $ cp -i cockroach-{{site.data.strings.version}}.darwin-10.9-amd64/cockroach "$i"
> ```

I think I'd use `/usr/local/bin/cockroach` here instead of `$i` (which is only going to be set if you ran step 3 in the same shell session)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-:-Kji6v9CjYsUqVQQEwJv:bb6ywmx)*

sploiselle:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[upgrade-cockroach.md, line 32 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-KigeYC0bbfQNjsvmYfG-r1-32:-Kjii9ElGXnBH_vOQjgb:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4001bedcdfbad23b4b1637ae56447581383586bd/upgrade-cockroach.md#L32)):*

The download part of this should come before stopping the old version. The only thing you want to do in between stopping the old process and starting the new one is copying the binary onto your path (if you're doing that) or pointing your process manager at the new binary.
</blockquote></details>

Done.

---

*[upgrade-cockroach.md, line 58 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-KigeYC0bbfQNjsvmYfG-r1-58:-KjiiizPMCZSnwuCL1zY:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4001bedcdfbad23b4b1637ae56447581383586bd/upgrade-cockroach.md#L58)):*

The "one minute" comment above is not what I meant here. What I meant was that after restarting one node, wait one minute before stopping the next.
</blockquote></details>

Done.

---

*[upgrade-cockroach.md, line 12 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5FochGB7GSt0Jdiw:-KjihvFtjjhfLxeQVY8T:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L12)):*

Must be behind a load balancer *or* your clients must be configured to talk to multiple nodes. 
</blockquote></details>

Done.

---

*[upgrade-cockroach.md, line 16 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5Lb_I8PAAMY9COoW:-KjiiepGjsZuOc5Y8izO:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L16)):*

I don't think there's anything significant about one minute (there's a threshold at five minutes)
</blockquote></details>

Done.

---

*[upgrade-cockroach.md, line 43 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji5n0DXnTSeppnnJQ7:-KjiiA37uQxvxhdUD0eT:b-5f6pvk) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L43)):*

`wget` is not installed by default on macs; this should use `curl` instead. (I don't know whether `wget` or `curl` is more likely to be present on linux)
</blockquote></details>

Done. (wget's a little more widely available by default)

---

*[upgrade-cockroach.md, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1338#-Kji619cZSCHohy-NloO:-KjiiLL2gMg2n1iQBZgr:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/55dc2a2244cf644ccdff90d8c177f6884ebf5086/upgrade-cockroach.md#L47)):*

I think I'd use `/usr/local/bin/cockroach` here instead of `$i` (which is only going to be set if you ran step 3 in the same shell session)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1338)*

sploiselle:
@jseldess Ready for copy edit.

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+
+Because of CockroachDB's multi-active availability design, you can perform a "rolling upgrade" of CockroachDB on your cluster. This means you can upgrade individual nodes in your cluster one at a time without any downtime for your cluster.
+
+However, there are a few things you should keep in mind:
+
+  - For rolling upgrades to work, your cluster must be behind a load balancer or your clients must be configured to talk to multiple nodes. If your application communicates with a single node, bringing it down to upgrade its CockroachDB binary will cause your application to fail when the node goes offline.
+
+  - We recommend creating scripts to upgrade CockroachDB and not attempting to do so by hand.
+
+  - Only bring down only one node at a time, and wait at least one minute after it rejoins the cluster to bring down the next node to prevent CockroachDB from entering a failover state.
```
jseldess: nit: Delete second `only`, and I'd suggest restructuring the sentence a bit to start with the goal:

> To prevent CockroachDB from entering a failover state, bring down only one node at a time, and wait at least one minute after it rejoins the cluster before bringing down the next node.

Also, I'm not exactly sure what we mean by "failover state"...

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+     # Optional: Place cockroach in your $PATH
+     $ cp -i cockroach-{{site.data.strings.version}}.linux-amd64/cockroach /usr/local/bin/cockroach
+     ~~~
+
+2. If you're running with a process manager, have it stop `cockroach`.
+
+   Without a process manager, use this command:
+
+   ~~~ shell
+   $ pkill cockroach
```
jseldess: Just curious: Why wouldn't we recommend `cockroach quit`? Cause we don't want cockroach to drain the node/transfer table leases?

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+
+3. If you use `cockroach` in your `$PATH`, rename the outdated `cockroach` binary:
+
+   ~~~ shell
+   $ i="$(which cockroach)"; mv "$i" "$i"_old
+   ~~~
+
+   If you leave versioned binaries on your servers, you don't need to do anything.
+
+
```
jseldess: nit: Extra line

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+5. If you're running with a process manager, have the node rejoin the cluster by starting it.
+
+   Without a process manager, use this command:
+
+   ~~~ shell
+   $ cockroach start --join=[any other node's IP address] [other flags]
+   ~~~
+
+   `[other flags]` includes any flags you [use to start nodes](start-a-node.html), such as its `--host`.
```
jseldess: Probably should remove `its` since `nodes` is plural.

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+
+6. If you use `cockroach` in your `$PATH`, once you've confirmed that the node has rejoined the cluster, you can remove the old binary:
```
jseldess: Should there be a step to confirm that the node has rejoined the cluster via the admin UI?

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
...
+
+   ~~~ shell
+   $ rm "$i"_old
+   ~~~
+
+Repeat these steps for each node in the cluster, waiting at least one minute after a node rejoins the cluste before bringing down the next one.
```
jseldess: `cluste` > `cluster`

Also, I think I'd make this a final step saying:

> Wait at least one minute after the node has rejoined the cluster, and then repeat these steps for the next node. 

And then use this as an intro to the steps:

> For each node in your cluster, complete the following steps.

_data/sidebar_doc.yml
```diff
@@ -462,6 +462,9 @@ entries:
         - title: View Version Details
           url: /view-version-details.html
 
+        - title: Upgrade Cluster Version
```
jseldess: nit: `Upgrade a Cluster` might be more straightforward?

upgrade-cockroach.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Upgrade Your Cluster's Version of CockroachDB
```
jseldess: Similar to feedback on the sidenav, I think `Upgrade a Cluster` might be simpler and sufficient. But I'm fine with this as well.

upgrade-cockroach.md
bdarnell: `kill` will do the draining too. The advantage of `kill` is that you don't need certificates, etc, and it's easier to integrate with process managers.

upgrade-cockroach.md
sploiselle: Simplified title but kept "version" keyword for SEO.

upgrade-cockroach.md
jseldess: Ah, thanks for the clarification, Ben.

upgrade-cockroach.md
sploiselle: Would rather lead with the direction rather than the benefit to improve scannability. Changing "failover state" to be more descriptive

upgrade-cockroach.md
sploiselle: Done.

upgrade-cockroach.md
sploiselle: Done.

upgrade-cockroach.md
sploiselle: Done.

_data/sidebar_doc.yml
sploiselle: "Version" is a commonly used term when dealing with OSS. Important to have this in here for SEO.

upgrade-cockroach-version.md
```diff
@@ -0,0 +1,95 @@
...
+
+Because of CockroachDB's multi-active availability design, you can perform a "rolling upgrade" of CockroachDB on your cluster. This means you can upgrade individual nodes in your cluster one at a time without any downtime for your cluster.
+
+However, there are a few things you should keep in mind:
+
+  - For rolling upgrades to work, your cluster must be behind a load balancer or your clients must be configured to talk to multiple nodes. If your application communicates with a single node, bringing it down to upgrade its CockroachDB binary will cause your application to fail when the node goes offline.
+
+  - We recommend creating scripts to upgrade CockroachDB and not attempting to do so by hand.
+
+  - Only bring down only one node at a time, and wait at least one minute after it rejoins the cluster to bring down the next node. This reduces the number of nodes you have offline at any point in time, minimizing the chance you'll lose a majority of the nodes in your cluster which can cause service outages.
```
jseldess: @sploiselle, very minor, but there's still an extra `only` in here.

sploiselle: Good catch! What I get for trying to finish something before a meeting. Fixed in #1388

PR #1341, opened 2017-04-27
Use single quotes throughout JS example

justinj:
There's no functional difference here, but it's a bit more consistent
with all the other uses to change these ones, especially because the
styling on the page highlights double-quoted strings a different colour
than single-quoted strings.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1341)

### DIFF
--- a/_includes/app/basic-sample.js
+++ b/_includes/app/basic-sample.js
@@ -25,11 +25,11 @@ pg.connect(config, function (err, client, done) {
   async.waterfall([
     function (next) {
       // Create the "accounts" table.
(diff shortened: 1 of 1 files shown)


sploiselle:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1341#-:-KilW4gmBfHkBAz9K2_-:bnfp4nl)*

PR #1343, opened 2017-04-27
Use single quotes throughout ruby example

justinj:
There isn't really a standard style for this in Ruby as far as I'm
aware, so I arbitrarily chose standardizing on single quotes because I
like the colour they highlight to on the docs page better. There's one
use of double quotes left because it uses string interpolation (which
single quotes disallow).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1343)

### DIFF
--- a/_includes/app/activerecord-basic-sample.rb
+++ b/_includes/app/activerecord-basic-sample.rb
@@ -5,11 +5,11 @@ require 'activerecord-cockroachdb-adapter'
 # Connect to CockroachDB through ActiveRecord.
 # In Rails, this configuration would go in config/database.yml as usual.
 ActiveRecord::Base.establish_connection(
-  adapter:  "cockroachdb",
-  username: "maxroach",
-  password: "",
-  database: "bank",
-  host:     "localhost",
+  adapter:  'cockroachdb',
+  username: 'maxroach',
+  password: '',
+  database: 'bank',
+  host:     'localhost',
   port:     26257,
 )
 
@@ -42,5 +42,5 @@ Account.create(id: 2, balance: 250)
 
 # Retrieve accounts and print out the balances
 Account.all.each do |acct|
-    puts "#{acct.id} #{acct.balance}"
+  puts "#{acct.id} #{acct.balance}"
 end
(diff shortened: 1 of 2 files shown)


sploiselle:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1343#-:-KilWTypGw7lfse0V__l:bnfp4nl)*

_includes/app/activerecord-basic-sample.rb
```diff
@@ -42,5 +42,5 @@ def change
 
 # Retrieve accounts and print out the balances
 Account.all.each do |acct|
-    puts "#{acct.id} #{acct.balance}"
+  puts "#{acct.id} #{acct.balance}"
```
jseldess: Did you mean to change to single quotes here as well?

justinj: In Ruby you can't use string interpolation within single quotes (they're just completely literal strings), so this one has to be double quoted. Afaik (and a Google seems to confirm) there isn't a standard style for which (single or double) to use by default, unfortunately.

jseldess: Ah, ok. Thanks for the explanation.

justinj:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[_includes/app/activerecord-basic-sample.rb, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1343#-KilFbLgBSyCaYNQAxoD-r1-45:-KipqUumRMJwxxyjb0MF:bn1wz0t) ([raw file](https://github.com/cockroachdb/docs/blob/0390ede7c4ba1e361d05caeb522a7c2513f34095/_includes/app/activerecord-basic-sample.rb#L45)):*

Ah, ok. Thanks for the explanation.
</blockquote></details>

:thumbsup:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1343)*

jseldess:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1343#-:-Kipr7upTYhRszimNrf8:bnfp4nl)*

PR #1357, opened 2017-05-03
Release notes for v1.0-rc.2

bdarnell:
As of cockroachdb/cockroach@198ba3aa75a0802b46aa2bea99540aa0d9004222

I'll be out tomorrow so here's an early version of the release notes so far. Update this with whatever else makes the cut.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1357)

PR #1371, opened 2017-05-04
add page to show all releases and their downloads

benesch:
This commit adds a page to display all releases and their downloads,
including binaries for Linux, Mac, and Windows (where available), and
source tarballs (where available). The sidebar is adjusted to display
only the latest version and a link to this page. The "old versions" page
is removed in favor of the new all-releases page.

Additionally adjusts the CSS to use the source archive icon consistently
across this page and the individual release notes pages.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1371)


benesch:
Here's a screenshot:

screenshot 2017-05-04 18 53 55

I expect we'll replace the "Latest" badge with "Recommended" or "Stable" once 1.0 lands.

benesch:
Also, @jseldess, I decided to stick with my icon because I think it's more evocative (and standard) than the generic down arrow. Let me know if you feel otherwise! This commit also adjusts the CSS for the button on the release page, though you can't see the effect because those buttons are added in the v1.0-rc.2 PR.

releases.md
```diff
@@ -0,0 +1,85 @@
...
+    vertical-align: middle;
+}
+</style>
+
+After downloading your desired release, learn how to [Install CockroachDB](install-cockroachdb.html).
+
+<table class="release-table">
+<thead>
+<tr>
+  <td>Version</td>
```
sploiselle: <td>Version &amp; Release Notes</td> just so it's clear what's at the other end of the link

benesch: Done!

benesch:
@sploiselle, want me to Title Case all the column headings? (Also, did you want "Version & Release Notes" instead of "Version and Release Notes"?)

sploiselle:
I think most everything is titled cased, but I can't find a table header example easily, so I'll defer to @jseldess.

I prefer ampersands because they take up less real estate on the screen.

benesch:
It's title-cased in [your backup documentation](https://www.cockroachlabs.com/docs/backup.html#object-dependencies), at least ;). Also in [Build an App with CockroachDB}(https://www.cockroachlabs.com/docs/build-an-app-with-cockroachdb.html).

Sending up a fix in a hot second.

jseldess:
Wohoo! This is very nice. Yes, let's title-case the column-headings, and I like `&` for the same reason as @sploiselle.

It'd be great if we could hold off on merging this until just after https://github.com/cockroachdb/docs/pull/1357, which will mean some minor rebasing here.

releases.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Releases
+summary: Release notes for older versions of CockroachDB.
+toc: false
+---
+
+<style>
```
jseldess: To separate the content from the styling, how about adding this as `/css/components/_release.scss` and then adding `@import "components/release";` to `/css/customstyles.scss`?

jseldess:
@benesch, free to merge now, once you've addressed any open issues.

benesch:
Ok, I think I've addressed everything! Merging as soon as this goes green.

jseldess:
@benesch, not sure what we missed, but the published version of the page has some styling issue: 

- The latest tag is not colored right.
- The columns aren't spaced well.

Ideas? 

screen shot 2017-05-05 at 12 25 37 pm

benesch:
Hmm, it seems okay on my machine. My hunch is you're picking up a cached version of `customstyles.css`. Try clearing your cache or incognito mode?

benesch:
Turned out to be caching, luckily!

jseldess:
Yep. Thanks, @benesch!

PR #1375, opened 2017-05-05
use versioned tarball in install instructions

### LINKED ISSUE #1368: Link to source tarball in release notes.
Porting over from https://github.com/cockroachdb/cockroach/issues/15684:
CC: 	@williamh

Hello, this is a question about building from source.

I am a maintainer on Gentoo Linux, and I want to package cockroach-1.0-rc1, but your documentation refers to a source tarball which is just named cockroach-latest.src.tgz with no version number in the name. Is this 1.0-rc1, or is it some snapshot later than that?

I have spent most of the day trying to figure out how to check out your git repositories and build from there without the build checking out anything from git but haven't had any luck.

If the tarball referred to in your documentation is the source for 1.0-rc1 do you mind versioning your tarball names in the future? If not, is there a simple way to download the release from http://github.com/cockroachdb/cockroach/releases, put the vendored code in the vendor directory, then build everything without the build process trying to pull from git?

Thanks,

William

benesch:
The `cockroach-latest*.tgz` tarballs have the unfortunate downside of not
including the version in either the filename or the extracted folder.
This was a convenient way to avoid updating the install instructions to
reflect the new filenames for every release. Now, though, we track the
latest release's name in _data/releases.yml, so we can use that to
automatically derive release filenames.

This is an important step in resolving cockroachdb/cockroach#14731.

Closes #1368.

/cc @mberhault @tamird @bdarnell if we land this change, will there be anything else that depends on the these tarballs?

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1375)

### DIFF
--- a/deploy-cockroachdb-on-aws-insecure.md
+++ b/deploy-cockroachdb-on-aws-insecure.md
@@ -110,11 +110,11 @@ AWS offers fully-managed load balancing to distribute traffic between instances.
 
 	~~~ shell
 	# Get the latest CockroachDB tarball.
-	$ wget https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz
+	$ wget https://binaries.cockroachdb.com/cockroach-{{ site.data.strings.version }}.linux-amd64.tgz
 
 	# Extract the binary.
-	$ tar -xf cockroach-latest.linux-amd64.tgz  \
-	--strip=1 cockroach-latest.linux-amd64/cockroach
+	$ tar -xf cockroach-{{ site.data.strings.version }}.linux-amd64.tgz  \
+	--strip=1 cockroach-{{ site.data.strings.version }}.linux-amd64/cockroach
 
 	# Move the binary.
 	$ sudo mv cockroach /usr/local/bin
@@ -141,11 +141,11 @@ At this point, your cluster is live and operational but contains only a single n
 
 	~~~ shell
 	# Get the latest CockroachDB tarball.
-	$ wget https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz
+	$ wget https://binaries.cockroachdb.com/cockroach-{{ site.data.strings.version }}.linux-amd64.tgz
 
 	# Extract the binary.
-	$ tar -xf cockroach-latest.linux-amd64.tgz  \
-	--strip=1 cockroach-latest.linux-amd64/cockroach
+	$ tar -xf cockroach-{{ site.data.strings.version }}.linux-amd64.tgz  \
+	--strip=1 cockroach-{{ site.data.strings.version }}.linux-amd64/cockroach
 
 	# Move the binary.
 	$ sudo mv cockroach /usr/local/bin
--- a/deploy-cockroachdb-on-aws.md
+++ b/deploy-cockroachdb-on-aws.md
@@ -213,11 +213,11 @@ Locally, you'll need to [create the following certificates and keys](create-secu
 
     ~~~ shell
     # Get the latest CockroachDB tarball.
-    $ wget https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz
+    $ wget https://binaries.cockroachdb.com/cockroach-{{ site.data.strings.version }}.linux-amd64.tgz
 
     # Extract the binary.
-    $ tar -xf cockroach-latest.linux-amd64.tgz  \
-    --strip=1 cockroach-latest.linux-amd64/cockroach
+    $ tar -xf cockroach-{{ site.data.strings.version }}.linux-amd64.tgz  \
+    --strip=1 cockroach-{{ site.data.strings.version }}.linux-amd64/cockroach
 
     # Move the binary.
     $ sudo mv cockroach /usr/local/bin
@@ -245,11 +245,11 @@ At this point, your cluster is live and operational but contains only a single n
 
     ~~~ shell
     # Get the latest CockroachDB tarball.
-    $ wget https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz
+    $ wget https://binaries.cockroachdb.com/cockroach-{{ site.data.strings.version }}.linux-amd64.tgz
 
     # Extract the binary.
-    $ tar -xf cockroach-latest.linux-amd64.tgz  \
-    --strip=1 cockroach-latest.linux-amd64/cockroach
+    $ tar -xf cockroach-{{ site.data.strings.version }}.linux-amd64.tgz  \
+    --strip=1 cockroach-{{ site.data.strings.version }}.linux-amd64/cockroach
 
     # Move the binary.
     $ sudo mv cockroach /usr/local/bin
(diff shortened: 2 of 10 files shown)


sploiselle:
@benesch We developed the `latest` tarballs so we didn't have to worry about people downloading old versions of Cockroach if the URL were copied and pasted elsewhere. For instance, on Digital Ocean, we don't have a reliable way of updating the URLs they display on their webpages. `latest` ensures that the person is getting a fresh version and not something from 6+ months ago (or, eventually 2 years ago). We decided to use it on our site, as well, because unsolicited authors often grab whatever URL you plunk on these kind of pages.

If there is a specific version of the `cockroach` binary someone wants, I suggest instead adding a note pointing them to the (real amazing) page you made that tracks the versions.

Alternatively, if you'd like to have the "unzipped" version of the `latest` tarball expose, the version, I'm sure that's a possibility.

benesch:
@sploiselle, sorry! I should have provided more context. There's been an ongoing discussion that's been unfortunately fragmented across cockroachdb/cockroach#14731, #engineering, and #release about what to do with these latest tarballs.

The core issue, as you noted, is that the latest tarball unzips to `cockroach-latest` instead of `cockroach-VERSION`. This rightly confused @williamh in #1368 and was the impetus for @bdarnell filing cockroachdb/cockroach#14731.

At the very least, as you mention, we want to adjust these latest links to be an HTTP redirect to the versioned tarball, rather than a completely separate tarball that unzips to `cockroach-latest`. 

You raise a good argument against nuking these tarballs completely. Let's see if we can centralize the discussion here and come to a consensus. I don't have strong feelings one way or another, but by the time 1.0 is released we should do one or the other: either scrub all references to the latest tarballs or make them HTTP redirects.

tamird:
:lgtm: fwiw

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1375#-:-KjOgxxygjWkhOby_50A:bb71oku)*

sploiselle:
@benesch If `latest` HTTP redirects to the latest version, that's fine with me. Or `latest-stable` or whatever y'all think is suitable.

I do strongly believe, though, that we need evergreen URLs to download the binary because it greatly improves many user's experience––the fewer hoops users jump through, the better. Once the package is on their machine, I care much less what's in it. If it's versioned at that point, it's totally fine.

williamh:
@sploiselle What do you mean by evergreen URLs?

jseldess:
These changes LGTM, but would you make the same changes to all the `deploy-cockroachdb-on-xxx` cloud tutorials, as well as to the `manual-deployment` tutorials? Those currently use `https://binaries.cockroachdb.com/cockroach-latest.linux-amd64.tgz` as well.

I'm also in favor of making the `latest` tarballs an HTTP redirect to the dated tarballs.

bdarnell:
@williamh An evergreen URL here is one like `-latest.tar.gz` that tracks new releases as they become available. 

I'm skeptical that an evergreen tarball URL is appropriate - if it becomes an HTTP redirect, then your instructions get messier since you have to say vague things like `after expanding the tarball, cd into the directory it created`. Most projects in my experience don't have evergreen urls like this. When they do, it's not the tarball itself that's served behind a redirect but something like an installer script (or another meta-thingy like an apt repository). 

"Latest" is also a slippery concept - if we discover a serious bug in 1.0  and release 1.0.3 after 1.1 has already been released, we don't want 1.0.3 to become "latest". But does that mean we need a separate "latest" url for each major release? 

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1375#-:-Kja2exkclb7jm_fHCD6:b9haeoj)*

mberhault:
I think `latest` made sense when we had weekly betas. However, with longer release cycles coming up, we will at the very least need something like `latest-beta`, `latest-stable`, etc... which may even be too much.

I think we may want to drop using `latest` entirely in the docs. Where we do want an actual link/filename (install docs, cloud scripts, etc..) we can use the latest docs version (same problem here, but this is manually updated, so easier to manage). Where we don't need it, we should just link to the list of release downloads.

jseldess:
As far as I can tell, if @benesch updates the other pages I mention above, we will no longer use `latest` in our docs. But as @sploiselle says, we will still use it on external sites that we don't have immediate control over, e.g., tutorials on Digital Ocean.

bdarnell:
If we can't update the tutorial, can we be sure that it will keep working with whatever we change the `latest` link to in the future? The proposed change (to replace the directory name `cockroach-latest.linux-amd64` with a versioned directory) would break the [digitalocean tutorial](https://www.digitalocean.com/community/tutorials/how-to-deploy-cockroachdb-on-a-three-node-cluster-on-ubuntu-16-04) anyway. I think that in general docs like this should at least limit themselves to a major version of cockroach - it would be OK if it started installing 1.0.1 as soon as that version is released, but having it pick up 1.1 automatically would be risky.

sploiselle:
I think it's important to note the different degrees of failure that come from the proposed changes. Not being able to download a file at all is a more complicated problem to solve than not being able to copy/paste an unzip command for a file that's already on your machine.

My goal with this is simply to make it as easy as possible for people to get up and running with Cockroach. I find it frustrating as a user when I install something outside of a package manager and doesn't have simple access to the latest version of the software. I hoped that `latest` would simply serve the same purpose as something like `apt-get cockroachdb`. However, this issue sounds like something that requires greater effort than is willing to be invested to make it successful, so I'll defer to Marc's and Ben's suggestions.

jseldess:
@benesch, not sure where we are with this. Can you update?

mberhault:
Another issue we're going to run into is the meaning of the `latest` links when we have a `1.0.X` alongside alpha/beta/RC releases for `1.1`. We definitely don't want `latest` to point to those, but we may not want it to remain on `1.0`. Instead, we may have to drop it entirely and link to the download page from external docs.

benesch:
Oof, good point. I'm going to finish addressing @jseldess's comments now. As you point out, the current implementation of the latest redirect will switch between 1.1-pre and 1.0.1, based on whichever was released most recently, and that's pretty unforgivable.

We'll have to work something out for the DO tutorial. @sploiselle, I do think you've raised a great point, and I think I might have a solution in mind.

sploiselle:
I can email Digital Ocean with an update to the text whenever you all decide what you want to do.

jseldess:
@benesch, @sploiselle, are we waiting on this until we've reach out to Digital Ocean? Or can we get this merged sooner than that?

sploiselle:
@benesch had mentioned he had a few things in progress; waiting on those to email DO with the requested edits.

sploiselle:
@benesch Are the changes in this PR good to clean up/merge?

benesch:
Sorry for dragging my feet on this! I've updated the additional pages you mentioned, @jseldess, and this is ready for re-review.

@sploiselle, nothing to do re Digital Ocean yet; we still don't have a better solution in place.

PR #1379, opened 2017-05-08
Minimum Windows version for Windows binary

### LINKED ISSUE #1377: Windows binary requires Windows 8+
I believe (still waiting on confirmation) that the Windows binaries to fail for this user on the forum because they're running a version of Windows pre-8: https://forum.cockroachlabs.com/t/where-are-the-windows-binaries/602/4

We should document the minimum required Windows version wherever we link to the binaries.

@tamird, can you confirm? Seems to be a RocksDB issue? https://github.com/facebook/rocksdb/issues/726.

jseldess:
Fixes #1377

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1379)

### DIFF
--- a/install-cockroachdb.md
+++ b/install-cockroachdb.md
@@ -461,7 +461,7 @@ $(document).ready(function(){
 <div id="download-the-binary-windows" class="install-option">
 <h2>Download the Binary</h2>
 
-{{site.data.alerts.callout_info}}Support for running CockroachDB on Windows is experimental and has not been extensively tested by Cockroach Labs. The Windows binary is provided as a convenience for local development and experimentation; production deployments of CockroachDB on Windows are strongly discouraged.{{site.data.alerts.end}}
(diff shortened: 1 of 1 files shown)


tamird:
Sorry for not seeing it before merging. I think this text is misleading in that it conflates "the Windows binary" with the source code which is what actually requires Windows 8+.

In other words, the current wording suggests that building from source might remove this requirement, but it would not.

jseldess:
@tamird, "build from source" isn't even presented as an option on Windows, as it is on Mac and Linux. We only present either "download the binary" or "use docker". 

If we want to provide instructions for building from source on Windows, I'll need help understanding how the process differs. Feels unnecessary, though, given that Windows support is only experimental...

benesch:
I suppose it could say "CockroachDB requires Windows 8 or later. Windows
support is experimental and has not been extensively tested by Cockroach
Labs. The binary is provided..."

The current wording doesn't bother me, though.

tamird:
@jseldess I think it is naive to assume that users won't try something just because it isn't presented in the docs (we have seen a variety of platforms and even had folks try ARMv8, after all).

I also appreciate that we don't necessarily want to consider those users when writing our docs, since they're off in uncharted territory, yet I find that the language here is imprecise and that making  it precise could be achieved without making it more pedantic or convoluted.

Would you accept a PR with such rewording? I'd be happy to take the task.

jseldess:
Of course, @tamird. Thanks for offering.

PR #1381, opened 2017-05-08
1.0 Release notes and known limitations

jseldess:
- Add v1.0 known limitations
- Add known limitations page to sidenav

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1381)


v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+In the [built-in SQL shell](use-the-built-in-sql-client.html), each previously executed command is stored in the SQL shell history. In cases where identical commands are executed consecutively, this can make it incoventient to return to a command prior to the identical batch of commands.
+
+## Using `\|` to perform a large input in the SQL shell
+
+In the [built-in SQL shell](use-the-built-in-sql-client.html), using the [`\|`](use-the-built-in-sql-client.html#sql-shell-commands) operator to perform a large number of queries from a file can cause the server to close the connection. This is because `\|` sends the entire file as a single query to the server, which can exceed the upper bound on the size of a packet the server can accept from any client (16MB).
+
+As a workaround, [execute the file from the command line](use-the-built-in-sql-client.html#execute-sql-statements-from-a-file) instead of from within the interactive shell.
+
+## New values generated by `DEFAULT` expressions during `ALTER TABLE ADD COLUMN`
```
jseldess: @knz, @andreimatei, please check these. If anything needs to be changed or added, I'd appreciate draft language.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+## `unique_rowid()` values in Javascript
+
+The `unique_rowid()` [function](functions-and-operators.html) returns values that cannot be represented exactly by numbers in Javascript. For example, a `235191684988928001` value returned by `unique_rowid()` would get represented as `235191684988928000` in Javascript. Notice that the last digit is different. In such cases, if the Javascript application temporarily stores the value `235191684988928001` and queries using it, it won't match the value that actually got stored.
+
+## Repeated commands in the SQL shell history
+
+In the [built-in SQL shell](use-the-built-in-sql-client.html), each previously executed command is stored in the SQL shell history. In cases where identical commands are executed consecutively, this can make it incoventient to return to a command prior to the identical batch of commands.
+
+## Using `\|` to perform a large input in the SQL shell
```
jseldess: @knz, the following is based on https://github.com/cockroachdb/cockroach/issues/15645. Please check and let me know if anything should be changed/expanded.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+When inserting/updating all columns of a table, and the table has no secondary indexes, it's recommended to use an `UPSERT` statement instead of the equivalent `INSERT ON CONFLICT` statement. Wherease the `INSERT ON CONFLICT` always has to read to determine the necessary writes, the `UPSERT` statement writes without first reading and so will be faster.
+
+This limitations may be particularly relevant if you are using a simple SQL table of two columns to [simulate direct KV access](frequently-asked-questions.html#can-i-use-cockroachdb-as-a-key-value-store). In that case, be sure to use the `UPSERT` statement.
+
+## `unique_rowid()` values in Javascript
+
+The `unique_rowid()` [function](functions-and-operators.html) returns values that cannot be represented exactly by numbers in Javascript. For example, a `235191684988928001` value returned by `unique_rowid()` would get represented as `235191684988928000` in Javascript. Notice that the last digit is different. In such cases, if the Javascript application temporarily stores the value `235191684988928001` and queries using it, it won't match the value that actually got stored.
+
+## Repeated commands in the SQL shell history
```
jseldess: @knz, the following is based on https://github.com/cockroachdb/cockroach/issues/14938. Please check and let me know if anything should be changed/expanded.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+<div id="toc"></div>
+
+## `INSERT ON CONFLICT` vs. `UPSERT`
+
+When inserting/updating all columns of a table, and the table has no secondary indexes, it's recommended to use an `UPSERT` statement instead of the equivalent `INSERT ON CONFLICT` statement. Wherease the `INSERT ON CONFLICT` always has to read to determine the necessary writes, the `UPSERT` statement writes without first reading and so will be faster.
+
+This limitations may be particularly relevant if you are using a simple SQL table of two columns to [simulate direct KV access](frequently-asked-questions.html#can-i-use-cockroachdb-as-a-key-value-store). In that case, be sure to use the `UPSERT` statement.
+
+## `unique_rowid()` values in Javascript
```
jseldess: @petermattis, the following is based on https://github.com/cockroachdb/cockroach/issues/15721. Please let me know if we need more details, and if we have a specific suggested workaround.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+title: v1.0 Known Limitations
+summary:
+toc: false
+---
+
+This page identifies limitations we've identified in the v1.0 release and plan to investigate and possibly resolve in a future release.
+
+<div id="toc"></div>
+
+## `INSERT ON CONFLICT` vs. `UPSERT`
```
jseldess: @danhhz, the following is based on https://github.com/cockroachdb/cockroach/issues/14482. We've documented this elsewhere already, but nonetheless, let me know if anything needs tweaking here.

jseldess:
I could use help adding more known limitations to this PR. If I've put your name next to one, I'd greatly appreciate it if you would add a comment to this PR with suggested language for explaining/describing the known limitation. If you don't have time to help, please suggest someone else. 

- [ ] [storage: Load-based lease rebalancing performs poorly if localities are under-specified in uneven latency deployments](https://github.com/cockroachdb/cockroach/issues/15369) - @a-robinson ?
- [ ] [sql: roundtrip to string does not respect precedence of ::: and -](https://github.com/cockroachdb/cockroach/issues/15617) - @eisenstatdavid ?
- [ ] [sql: requests against tables with FKs are slow](https://github.com/cockroachdb/cockroach/issues/15157) - @vivekmenezes ? 
- [ ] [sql: TestSteady index backfill seems to not get run properly](https://github.com/cockroachdb/cockroach/issues/15038) - @vivekmenezes ?
- [ ] [sql: improve overload resolution](https://github.com/cockroachdb/cockroach/issues/10679) - @eisenstatdavid ?
- [ ] [sql: support a UUID type which text-marshals to valid UUID syntax](https://github.com/cockroachdb/cockroach/issues/8719) - @bdarnell, @knz, @dt ?

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+When inserting/updating all columns of a table, and the table has no secondary indexes, it's recommended to use an `UPSERT` statement instead of the equivalent `INSERT ON CONFLICT` statement. Wherease the `INSERT ON CONFLICT` always has to read to determine the necessary writes, the `UPSERT` statement writes without first reading and so will be faster.
+
+This limitations may be particularly relevant if you are using a simple SQL table of two columns to [simulate direct KV access](frequently-asked-questions.html#can-i-use-cockroachdb-as-a-key-value-store). In that case, be sure to use the `UPSERT` statement.
+
+## `unique_rowid()` values in Javascript
+
+The `unique_rowid()` [function](functions-and-operators.html) returns values that cannot be represented exactly by numbers in Javascript. For example, a `235191684988928001` value returned by `unique_rowid()` would get represented as `235191684988928000` in Javascript. Notice that the last digit is different. In such cases, if the Javascript application temporarily stores the value `235191684988928001` and queries using it, it won't match the value that actually got stored.
```
knz: The workaround is to ensure that the JavaScript code that uses a column initialized with unique_rowid converts the id value to a string server-side before inspecting/using it.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Repeated commands in the SQL shell history
+
+In the [built-in SQL shell](use-the-built-in-sql-client.html), each previously executed command is stored in the SQL shell history. In cases where identical commands are executed consecutively, this can make it incoventient to return to a command prior to the identical batch of commands.
```
knz: incoventient -> inconvenient

Also, I have tried to reproduce #14938 and was unable to. It does not appear that we duplicate the history, as the issue implies. I would be more cautious with the phrasing, for example:

"In the [built-in SQL shell](use-the-built-in-sql-client.html), each previously executed command is stored in the SQL shell history. In some cases, commands are unexpectedly duplicated in the history."

v1.0-known-limitations.md
knz: :+1:

v1.0-known-limitations.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Using `\|` to perform a large input in the SQL shell
+
+In the [built-in SQL shell](use-the-built-in-sql-client.html), using the [`\|`](use-the-built-in-sql-client.html#sql-shell-commands) operator to perform a large number of queries from a file can cause the server to close the connection. This is because `\|` sends the entire file as a single query to the server, which can exceed the upper bound on the size of a packet the server can accept from any client (16MB).
+
+As a workaround, [execute the file from the command line](use-the-built-in-sql-client.html#execute-sql-statements-from-a-file) instead of from within the interactive shell.
```
knz: I would be more concrete here and say `[ execute ... ]( ... ) with "cat data.sql | cockroach sql", instead ...`

v1.0-known-limitations.md
knz: works for me

knz:
Some additional limitations:

- support for grouping DDL (schema change) statements inside a transaction is limited; and so is the combination of DDL and DML statements operating on the new schema inside the same transaction. For details see here: https://github.com/cockroachdb/cockroach/projects/4 (we may need a better high-level overview, @dt and @vivekmenezes can provide)

- by default, a single transaction can contain at most 100000 write operations (e.g. changes to individual row cells) and at most 64MiB of combined updates. These limits are configurable with the cluster settings `kv.raft.command.max_size` and `kv.transaction.max_intents` however non-default values are not supported yet. #15350

- the combined size of all the values in the columns of a single column family on a single row should not exceed the configured range size (see zone configs, 64MiB by default) for that table, or otherwise a write may fail or general cluster slowness may ensue. 

- the two previous points imply in particular that it is not yet possible to store BLOBs larger than 64MB (recommended: less than 32MiB) inside a single table cell in CockroachDB, without changing internal tuning parameters to non-supported values.

- the mechanisms implemented inside CockroachDB to limit the number of clients and queries based on how much free RAM is available are not complete (see blog post): on a loaded system, there exists a combination of high number of simultaneous connections and number of running queries beyond which the node accepting them may crash due to memory exhaustion. This can be prevented by monitoring CockroachDB memory usage and ensuring there is some margin between maximum CockroachDB memory usage and available system RAM.

- many SQL operators (ORDER BY, UNION, GROUP BY, window functions) accumulate intermediate results in RAM on the node where the operator is processed. If the operator attempts to process more rows than can fit into RAM, either a memory capacity error will be reported to the client that issued the query, or the node may crash. (for details see same blog post as previous point)

- using `COUNT(DISTINCT t.*)` to count distinct pairs of only one table when multiple tables are selected (e.g. via a join) will return incorrect results.  To work around, list the columns explicitly with e.g. `COUNT(DISTINCT (t.x, t.y, ...))`. #15750

knz:
The point about UUIDs is not a known limitation, it's rather a non-standard SQL feature which we have chosen not to implement yet. I'm not sure we want to make a list of every SQL feature under the sun that other RDBMs support but we don't.

eisenstatdavid:
sql: roundtrip to string -- "queries with constant expressions that evaluate to 2\*\*-63 may incorrectly rejected".

sql: improve overload resolution -- "many string operations are not properly overloaded for collated strings".

petermattis:
* `TRUNCATE` may fail with errors about commands being "too large". The implementation of `TRUNCATE` is currently quite inefficient. A workaround is to use `DROP TABLE` instead and to subsequently recreate the table.

---

Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-:-KjcM85iRprzzIh9Fhc6:b-718sru)*

v1.0-known-limitations.md
jseldess: Done.

v1.0-known-limitations.md
jseldess: Done.

v1.0-known-limitations.md
jseldess: Done.

jseldess:
@petermattis, please check whether the `Removing all rows from large tables` limitation is correct. I expanded your guidance a bit.

petermattis:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[v1.0-known-limitations.md, line 19 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-KjcZlDzt75tu97LcoC7:-KjcZlDzt75tu97LcoC8:bnfp4nl) ([raw file](https://github.com/cockroachdb/docs/blob/3de76ab62d03506952b137bef391d34aa57e28d2/v1.0-known-limitations.md#L19)):*
> ```Markdown
> 1. Use [`SHOW CREATE TABLE`](show-create-table.html) to get the table schema.
> 2. Use [`DROP TABLE`](drop-table.html) to remove the table.
> 3. Use [`CREATE TABLE`](create-table.html) with the output from step 1 to recreate the table.
> ```

:lgtm:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381)*

jseldess:
@eisenstatdavid, can you please provide a concrete SQL example for each? These single sentences don't feel like quite enough.

> sql: roundtrip to string -- "queries with constant expressions that evaluate to 2**-63 may be incorrectly rejected".
> sql: improve overload resolution -- "many string operations are not properly overloaded for collated strings".

jseldess:
@knz, thanks for adding those additional known limitations, but shouldn't there be cockroach issues with the `known-limitation` flag for each one of them to ensure that we investigate and possible resolve them in the future? I'd hate for this PR to be the only record.

eisenstatdavid:
roundtrip:
```
root@:26257/> create database d;
CREATE DATABASE
root@:26257/> create table d.t (i int primary key);
CREATE TABLE
root@:26257/> select * from d.t where (i, 2*i) < (42, -9223372036854775807 - 1);
pq: ($0, 2 * $0) < (42, - 9223372036854775808:::INT): tuples ($0, 2 * $0), (42, - 9223372036854775808:::INT) are not comparable at index 2: numeric constant out of int64 range
```

overloads:
```
root@:26257/> select 'string1' || 'string2';
+------------------------+
| 'string1' || 'string2' |
+------------------------+
| string1string2         |
+------------------------+
(1 row)
root@:26257/> select ('string1' collate en) || ('string2' collate en);
pq: unsupported binary operator: <collatedstring{en}> || <collatedstring{en}>
```

a-robinson:
[storage: Load-based lease rebalancing performs poorly if localities are under-specified in uneven latency deployments](https://github.com/cockroachdb/cockroach/issues/15369). This isn't likely to effect most people, and I'm not sure how to make it clear without being pretty long-winded but here's a somewhat condensed explanation:

If your cluster:
1. is running in datacenters that are very different distances away from each other (e.g. from datacenter A to datacenter B is 10ms and from A to C is 100ms)
1. has a single locality label assigned to each datacenter (e.g. `--locality datacenter=a`)
1. gets most of its requests sent to a single datacenter because that's where all your application traffic is

...then you might see increased latency caused by a continuously high rate of lease transfers between datacenters. You can tell whether this is happening by seeing whether the "Leases Transferred / second" line on the "Replication Queue" graph on the "Queues" dashboard of the admin UI is consistently larger than 0.

If you're seeing this problem, you can fix it (thus improving request latency) by adding a second tier to each node's locality specification. In the example above where A and B are close together and C is far apart, you might change A and B's localities to `--locality region=foo,datacenter=a` and `--locality region=foo,datacenter=b` while changing C's to `--locality region=bar,datacenter=c`. This will inform the replication logic about A and B's relative proximity.

knz:
@jseldess I have added the links to issues in my previous comment

jseldess:
Thanks, @a-robinson. I reworked your explanation a bit. Please take a look.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,68 @@
...
+
+When nodes are started with the [`--locality`](start-a-node.html) flag, CockroachDB attempts to place the replica lease holder (the replica that client requests are forwarded to) on the node that is closest to the source of the request. This means that as client requests move geographically, so does the replica lease holder.
+
+However, you might see increased latency caused by a continuously high rate of lease transfers between datacenters in the following case:
+
+- Your cluster is running in datacenters that are very different distances away from each other.
+- Each node was started with a single tier of `--locality`, e.g., `--locality=datacenter=a`.
+- Most client requests get sent to a single datacenter because that's where all your application traffic is.
+
+To detect that this is happening, open the [Admin UI](explore-the-admin-ui.html), select the **Queues** dashboard, hover over the **Replication Queue** graph, and check the **Leases Transferred / second** data point. If that data point is larger than 0, you should consider stopping and restarting each node with additional tiers of locality to improve request latency.
```
a-robinson: It's not necessarily bad if the data point is larger than 0 once in a while. The issue is if it's continuously/consistently greater than 0, indicating that leases are constantly being transferred. The occasional transfer is fine.

jseldess: Done.

jseldess:
@knz, that [UUID issue](https://github.com/cockroachdb/cockroach/issues/8719) is flagged as a known limitation. Should we remove the label?

petermattis:
@jseldess Yes, the lack of UUID support doesn't fall under my definition of known-limitation. We can't document all of our missing functionality. I think known-limitation should be constrained to functionality that we're providing but which has significant limitations, surprises or bugs.

jseldess:
Got it, @petermattis. I'll remove it from my list.

eisenstatdavid:
:lgtm: wrt 2**-63 and collated strings.

---

Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-:-KjhbbBHGtJ4ypSsHO_u:b-68mkrx)*

jseldess:
@bdarnell, @petermattis, @mberhault, @spencerkimball, @jess-edwards, @dianasaur323, here's a basic draft of the 1.0 release notes page: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/5d6625d82e86a3685919a40bb55cd147b95f46b1/v1.0.html

Let's review together during the docs team meeting today.

v1.0-known-limitations.md
```diff
@@ -0,0 +1,111 @@
...
+> SELECT (-9223372036854775808) ::: INT;
+
+> SELECT i FROM t WHERE (i, i) < (1, -9223372036854775808);
+~~~
+
+~~~
+pq: ($0, $0) < (1, - 9223372036854775808:::INT): tuples ($0, $0), (1, - 9223372036854775808:::INT) are not comparable at index 2: numeric constant out of int64 range
+~~~
+
+## Overload resolutation for collated strings
```
mberhault: "resolution"

_data/sidebar_doc.yml
```diff
@@ -495,8 +495,11 @@ entries:
     - title: Releases
       items:
 
-        - title: v1.0-rc.2
-          url: /v1.0-rc.2.html
+        - title: 1.0
+          url: /v1.0.html
+
+        - title: Known Limitations
```
mberhault: We're sticking the release notes and limitations into the "releases" sidebar? That seems a bit odd, no?

v1.0.md
```diff
@@ -0,0 +1,45 @@
...
+### Downloads
+
+<div id="os-tabs" class="clearfix">
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.0.darwin-10.9-amd64.tgz"><button id="mac" data-eventcategory="mac-binary-release-notes">Mac</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.0.linux-amd64.tgz"><button id="linux" data-eventcategory="linux-binary-release-notes">Linux</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.0.windows-6.2-amd64.zip"><button id="windows" data-eventcategory="windows-binary-release-notes">Windows</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.0.src.tgz"><button id="source" data-eventcategory="source-release-notes">Source</button></a>
+</div>
+
+### General Availability
```
mberhault: I think we should link to the limitations from here rather than the sidebar. Or both if you really want the sidebar.

v1.0-known-limitations.md
jseldess: Done

_data/sidebar_doc.yml
jseldess: Do you think it's enough to just link from the release notes? We don't need to show the known limitations page somewhere in the sidenav?

v1.0.md
jseldess: Added link to release notes. See my question above about the sidebar.

_data/sidebar_doc.yml
bdarnell: We should put it somewhere in the sidebar, but I'd put it under FAQ instead of Releases. And maybe it should be named `known-limitations.md` instead of `v1.0-known-limitations.md`. I think we want to be able to update this page in the future by by removing things as they are fixed (and documenting the known limitations of older versions should be handled by whatever generic doc-versioning scheme we adopt).

jseldess: OK, I've renamed the file, and I've moved it under **FAQ** in the sidenav.

vivekmenezes:
Known limitations of schema changes within transactions

- Certain CREATE TABLE statements that use Foreign-Keys or Interleaved tables cannot be followed by statements that reference the new table.
- DDL statements cannot follow DML statements in the same transaction. The work around is to rearrange the statements in the transaction.
- DROP table  doesn't  drop the name of the table immediately. Therefore a DROP TABLE followed by a CREATE TABLE using the same name will not work in the same transaction.

---

Review status: 0 of 4 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-:-KjiIsZ_Ug5DG6qr3ptK:b-hswetq)*

jseldess:
Thanks, @vivekmenezes. Please check my edits.

known-limitations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Creating views with array types
+
+It's not possible to [create a view](create-view.html) with an array type in the view's `SELECT` query. Attempting to do so causes the node that receives the request to crash.
+
+## Write and update limits for a single transaction
+
+A single transaction can contain at most 100,000 write operations (e.g., changes to individual columns) and at most 64MiB of combined updates. When a transaction exceeds these limits, it gets aborted.
+
+If you need to increase these limits, you can update the [cluster-wide settings](cluster-settings.html) `kv.transaction.max_intents` and `kv.raft.command.max_size`. However, we strongly encourage you to discuss your goals with CockroachDB before doing so.
```
knz: LGTM

known-limitations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Max size of a single column family
+
+When creating or updating a row, if the combined size of all the values in a single [column family](column-families.html) exceeds the max range size (64MiB by default) for the table, the operation may fail, or cluster performance may suffer.
+
+As a workaround, you can either [manually split a table's columns into multiple column families](column-families.html#manual-override), or you can [create a table-specific zone configuration](configure-replication-zones.html#create-a-replication-zone-for-a-table) with an increased max range size.
```
knz: LGTM

known-limitations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Simultaneous client connections and running queries on a single node
+
+When a node has both a high number of client connections and running queries, the node may crash due to memory exhaustion. This is due to CockroachDB not limiting the number of clients and queries based on how much free RAM is available on the node.
```
knz: "This is due to CockroachDB not accurately limiting ..."

knz: the rest lgtm

known-limitations.md
```diff
@@ -0,0 +1,181 @@
...
+
+To prevent memory exhaustion, monitor each node's memory usage and ensure there is some margin between maximum CockroachDB memory usage and available system RAM. For more details about memory usage in CockroachDB, see [this blog post](https://www.cockroachlabs.com/blog/memory-usage-cockroachdb/).
+
+## SQL subexpressions and memory usage
+
+Many SQL subexpressions (e.g., `ORDER BY`, `UNION`/`INTERSECT`/`EXCEPT`, `GROUP BY`, subqueries, window functions) accumulate intermediate results in RAM on the node processing the query. If the operator attempts to process more rows than can fit into RAM, either a memory capacity error will be reported to the client that issued the query, or the node may crash. For more details about memory usage in CockroachDB, see [this blog post](https://www.cockroachlabs.com/blog/memory-usage-cockroachdb/).
```
knz: LGTM

known-limitations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Counting distinct rows in a table
+
+When using `count(DISTINCT a.*)` to count distinct rows in a table, the results are almost always incorrect, for example:
```
knz: " ... to count distinct rows in a table based on a subset of the columns, as opposed to `count(*)`, the results ..."

the rest lgtm

known-limitations.md
jseldess: Done.

known-limitations.md
jseldess: Done.

jess-edwards:
@jseldess This is looking great.

jseldess:
@vivekmenezes, can you write up the following known limitations for me please? 
-https://github.com/cockroachdb/cockroach/issues/15038
- https://github.com/cockroachdb/cockroach/issues/15157

jseldess:
@tamird, @eisenstatdavid, @jordanlewis: can one of you write up this know limitation for me, with any workarounds? https://github.com/cockroachdb/cockroach/issues/15800

known-limitations.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Schema changes within transactions
+
+Within a single [transaction](transactions.html):
+
+- DDL statements cannot follow DML statements. As a workaround, arrange DML statements before DDL statements, or split the statements into separate transactions.
+- A [`CREATE TABLE`](create-table.html) statement containing [`FOREIGN KEY`](foreign-key.html) or [`INTERLEAVE`](interleave-in-parent.html) clauses cannot be followed by statements that reference the new table.
+- A table cannot be dropped and then recreated with the same name. This is not possible within a single transaction because `DROP TABLE` does not immediately drop the name of the table. As a workaround, split the [`DROP TABLE`](drop-table.html) and [`CREATE TABLE`](create-table.html) statements into separate transactions.
+
+## Nodes need access to the first key-value range on startup
```
jseldess: @spencerkimball, @a-robinson, is this sufficient for documenting https://github.com/cockroachdb/cockroach/issues/13027? If not, please help me revise/expand as necessary.

eisenstatdavid:
I think that the `select max(true)` panic was introduced subsequent to rc2, but I'll double check and get back to you if not.

eisenstatdavid:
Indeed, rc2 is fine.

jseldess:
Thanks, @eisenstatdavid.

known-limitations.md
```diff
@@ -0,0 +1,205 @@
...
+
+When a CockroachDB node is started, logs are written to a file by default, with a symlink referencing the log file. On Windows, symlink creation requires admin permissions. Therefore, when running CockroachDB on Windows as a non-admin user, you must pass the `--log-dir=` to the `cockroach start` command to have logs written to `stdout` instead of to a file:
+
+~~~ shell
+$ cockroach.exe start --insecure --log-dir=
+~~~
+
+## Query planning for `OR` expressions
+
+Given a query like `SELECT * FROM foo WHERE a > 1 OR b > 2`, even if there are appropriate indexes to satisfy both `a > 1` and `b > 2`, the query planner will perform a full table or index scan because it can't use both conditions at once.
```
knz: LGTM

a-robinson:
Review status: 0 of 6 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[known-limitations.md, line 29 at r11](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-KjiD7SFhjso-6k5cplm-r11-29:-Kjm_j7ATL-0Vp0AaoHv:b-30xye8) ([raw file](https://github.com/cockroachdb/docs/blob/f41f87c8c162f790f3ab18b7552a2c2f44e054c1/known-limitations.md#L29)):*

@spencerkimball, @a-robinson, is this sufficient for documenting https://github.com/cockroachdb/cockroach/issues/13027? If not, please help me revise/expand as necessary.
</blockquote></details>

First off, this only affects clusters that are being restarted with entirely new IP addresses. I'd frame the known limitation as only affecting situations in which you're restoring a backup onto a new set of machines that have different IP addresses.

Second, I'm fairly confident that telling the first node to join itself won't actually fix the problem. Even if I did think that it would work, we never actually tested it so I'd still be a little hesitant to rely on recommending that. Instead, I'd either suggest including all node addresses in the join flags or including node (n-1)'s address (or n for the first node) when restoring from backup onto a new set of machines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381)*

a-robinson:
Review status: 0 of 10 files reviewed at latest revision, 21 unresolved discussions.

---

*[known-limitations.md, line 31 at r11](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-KjmbgqM6bKW_wO_VoN-:-KjmbgqNLPD-VB31l-Jd:b-ewbzsp) ([raw file](https://github.com/cockroachdb/docs/blob/f41f87c8c162f790f3ab18b7552a2c2f44e054c1/known-limitations.md#L31)):*
> ```Markdown
> ## Nodes need access to the first key-value range on startup
> 
> As show in our [deployment tutorials](manual-deployment.html), when starting the first node of a cluster, the `--join` flag should be empty, but when starting all subsequent nodes, `--join` should be set to the address of node 1. This approach ensures that all nodes have access to a copy of the first key-value range, which is part of a meta-index identifying where all range replicas are stored, and which nodes requires to initialize themselves and start accepting incoming connections.
> ```

s/As show/As shown/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381)*

a-robinson:
:lgtm: other than my last two comments

---

Reviewed 1 of 4 files at r8, 1 of 4 files at r9, 2 of 3 files at r10, 2 of 2 files at r11, 3 of 6 files at r12.
Review status: 7 of 10 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1381#-:-KjmdU_ynev4c7vCQw7m:bszw2ji)*

PR #1392, opened 2017-05-10
Reword Windows symlink permissions note

tamird:
In particular, the previous wording of "the `--log-dir=`" was odd,
implying that the user must pass "the `--log-dir` flag", rather than
being clear about the need fdor passing the empty value.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1392)

### DIFF
--- a/known-limitations.md
+++ b/known-limitations.md
@@ -190,10 +190,10 @@ As a workaround, list the columns explicitly, for example:
 
 ## Running on Windows as a non-admin user
 
-When a CockroachDB node is started, logs are written to a file by default, with a symlink referencing the log file. On Windows, symlink creation requires admin permissions. Therefore, when running CockroachDB on Windows as a non-admin user, you must pass the `--log-dir=` to the `cockroach start` command to have logs written to `stdout` instead of to a file:
+By default, CockroachDB logs to a file which is periodically rotated; in addition to these log files, a symlink pointing to the current log file is maintained. On Windows, symlink creation requires admin permissions, which can cause CockroachDB to fail on start. As a workaround, when running CockroachDB on Windows as a non-admin user, pass `--log-dir=` (with the empty value) to the `cockroach start` command to have logs written to `stdout` instead of to a file, e.g.:
 
 ~~~ shell
-$ cockroach.exe start --insecure --log-dir=
+$ cockroach.exe start --log-dir= --insecure
 ~~~
 
 ## Query planning for `OR` expressions


sploiselle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-KjmvzGuUSehG1UN-DCv:b-rxo191) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*
Agreed this is much clearer. Do you think we could shorten it a little bit without losing clarity? Something like:

> When running CockroachDB as a non-admin user, you must pass `--log-dir=` (with the empty value) to the `cockroach start` command, which writes logs to `stdout` instead of to a file. You must do this because non-admin users cannot create symlinks on Windows, which CockroachDB requires to write logs to files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

tamird:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-KjmxAju-yR6X2z_rnVD:bxmr75r) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*

Agreed this is much clearer. Do you think we could shorten it a little bit without losing clarity? Something like:

> When running CockroachDB as a non-admin user, you must pass `--log-dir=` (with the empty value) to the `cockroach start` command, which writes logs to `stdout` instead of to a file. You must do this because non-admin users cannot create symlinks on Windows, which CockroachDB requires to write logs to files.
</blockquote></details>

I don't know the prevailing style in the docs, so I'll leave the final decision to you, but I have a slight preference for building the user's intuition by explaining the source of the limitation before providing the workaround.

In other words, I prefer that we first explain why a workaround is necessary, and then offer the particular workaround as a suggestion, rather than prescribing a "you must" directive and only then explaining the problem it is intended to address.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

sploiselle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-KjmxcSzogTsjwpuYNKd:b-gjmfz2) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*
I assume people read the beginning and not end of sentences, but if you're more comfortable with *reason>solution* that's totally cool:

> Because non-admin users cannot create symlinks on Windows, which CockroachDB requires to write logs to files, you must write logs to `stdout` instead of files. To do this,   pass `--log-dir=` (with the empty value) to the `cockroach start` command.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

jseldess:
*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-Kjn-H64m1DO8VAQr59U:b3sou5n) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*
Normally, I'd go with Sean's approach, but for this page, I think it's best to explain the limitation first and then provide the workaround. 

I'd only suggest, we lead into the code block:

> ... to `stdout` instead of to a file, for example:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

jseldess:
:lgtm: with a nit

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-:-Kjn-j8nMKCxR7xAmuss:b-d5ftqp)*

tamird:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-Kjn0aaplq3737OFM92D:b-gacteb) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*
@sploiselle 

> Because non-admin users cannot create symlinks on Windows, which CockroachDB requires to write logs to files, you must write logs to stdout instead of files.

I dislike this wording because it leaves me wondering "why does cockroachDB require symlinks to write to logs to files? that seems odd!". I think it's worthwhile to explain why we do this, in addition to stating that we do.

@jseldess done and flipped the two flags to be extra super explicit about the empty value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

sploiselle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-Kjn8OrZNuR5Z6jJUj0X:bscf16e) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

@sploiselle 

> Because non-admin users cannot create symlinks on Windows, which CockroachDB requires to write logs to files, you must write logs to stdout instead of files.

I dislike this wording because it leaves me wondering "why does cockroachDB require symlinks to write to logs to files? that seems odd!". I think it's worthwhile to explain why we do this, in addition to stating that we do.

@jseldess done and flipped the two flags to be extra super explicit about the empty value.
</blockquote></details>

Sorry I wasn't clear. I meant these edits to be illustrative and not directional. I'll leave a rewrite of it to you but here are the issues:

- Avoid using semicolons
- "which is periodically rotated" is passive voice. Convert to active.
- "is maintained" is passive voice. Convert to active.
- This issue sounds like it's not a "can cause" but instead just "causes."
- "As a workaround" sounds like it's a hack. Instead, provide clear direction of what you want the user to do without making it sound like we require hacky behavior to make out software work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392)*

tamird:
I think I'm just going to fix the underlying problem instead!

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[known-limitations.md, line 193 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-KjmvzGuUSehG1UN-DCu:-KjnXZvES6mDV4U4QXAf:b-xorepp) ([raw file](https://github.com/cockroachdb/docs/blob/79ad06e97b2341da883ce92366c6b6cd1ab43fdd/known-limitations.md#L193)):*

Sorry I wasn't clear. I meant these edits to be illustrative and not directional. I'll leave a rewrite of it to you but here are the issues:

- Avoid using semicolons
- "which is periodically rotated" is passive voice. Convert to active.
- "is maintained" is passive voice. Convert to active.
- This issue sounds like it's not a "can cause" but instead just "causes."
- "As a workaround" sounds like it's a hack. Instead, provide clear direction of what you want the user to do without making it sound like we require hacky behavior to make out software work.
</blockquote></details>

Thanks for the wonderfully detailed feedback! 

It looks like none of this advice is documented in CONTRIBUTING.md! Consider adding it, I think it'd be incredibly valuable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1392#-:-KjnXX4Br65AvTMZkzXD:b-smm9od)*

jseldess:
@tamird, amazing that you're just going to fix the problem, but that fix isn't in the current 1.0 release, so I think your improvements are still relevant. Consider reopening the issue, or if you'd prefer, I can take over your changes?

tamird:
Go ahead - I agree that those improvements are relevant, but given that
we'll release 1.0.1 in the next few weeks, it might be just as well to wait.

sploiselle:
@jseldess Opening up a PR with copy edits to this page. Will include Tamir's edits there.

jseldess:
Perfect. Thank you, @sploiselle!

PR #1396, opened 2017-05-10
Window Functions are SQL standard

### LINKED ISSUE #1395: Issue on docs for SQL Feature Support
The docs ( https://www.cockroachlabs.com/docs/sql-feature-support.html ) reference window functions as a common extension, when they have been part of the standard since SQL:2003.

justinpitts:
Fix #1395 by stating that window functions are a standard feature.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1396)


sploiselle:
Thanks so much, @justinpitts!

PR #1404, opened 2017-05-12
Draw more attention to Windows 8 requirement

jseldess:
In response to https://github.com/cockroachdb/cockroach/issues/15887#issuecomment-301103361.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1404)


jseldess:
@tamird, we don't have icons like that, but it's a good idea to design each of the three callout types (tips, infos, and warnings) with an appropriate icon. I'll open an issue for that.

PR #1416, opened 2017-05-15
build from source produces binary in source dir

couchand:
On my machine `make build` produces a binary in the source directory, there is no `./bin` to be found.  I'm on a Linux machine so I updated that section; I also updated the Mac section on the assumption that it behaves similarly, but feel free to tell me if I'm mistaken.  I can't seem to find any explicit reference in the Makefile that dictates where the binary ends up, so perhaps it's an artifact of my particular go/cmake/make versions/config.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1416)

### DIFF
--- a/install-cockroachdb.md
+++ b/install-cockroachdb.md
@@ -218,7 +218,7 @@ $(document).ready(function(){
   <li>
   <p>Move the binary into your <code>PATH</code> so it's easy to execute <a href="cockroach-commands.html">cockroach commands</a> from any shell:</p>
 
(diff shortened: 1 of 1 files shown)


jseldess:
Thanks for the change, @couchand. 

@benesch, can you take a look?

benesch:
Hey @couchand, thanks so much for reporting this! You're spot on: the binary no longer builds into `./bin/cockroach`, but rather `./src/github.com/cockroachdb/cockroach/cockroach`.

We actually added a `make install` command so that you wouldn't have to construct the right `cp -i ...` invocation. I've opened a PR (#1418) to that effect instead.

Thanks again!

PR #1419, opened 2017-05-16
fixed broken layout (ascii table)

tlossen:
one ascii table on this page wasn't escaped properly, causing the layout to break.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1419)


sploiselle:
Thanks @tlossen!

PR #1420, opened 2017-05-17
update homebrew install to use core formula instead of tap

benesch:
As of homebrew/homebrew-core#13431, CockroachDB is now in Homebrew
proper! Point the install instructions at the core cockroach formula
(`cockroach`) instead of our tap's cockroach formula
(`cockroachdb/cockroach/cockroach`).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1420)


benesch:
Note to self: update the release instructions when this lands.

cc @bdarnell any reason not to do this?

benesch:
TFTRs!

PR #1422, opened 2017-05-17
Unstable releases

bdarnell:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1422)


jseldess:
@benesch, since you set up the current release page, you should probably take a quick glance as well.

benesch:
No complaints here! Thanks, @bdarnell!

Issue #957, opened 2017-01-04
FAQ: Why is my empty database so big?


bdarnell:
New users who set up a cockroach cluster and return to it after a few days are often surprised by the amount of memory (rocksdb cache) and disk (timeseries storage) its using. We should add something about this to the FAQ page (in addition to making it less surprising in other ways, like attributing this data to its source in the admin UI, or even better just using less space for this stuff)

jseldess:
From Gitter:
```
DBuTbKa @DBuTbKa 03:18
Hello everyone. Can anybody tell me. I was up a cockroachdb cluster with 6 containers on Kubernetes, create couple empty databases. In one databse create 5 tables and put some data (about 200KB) and wait couple days. Now I was look to admin UI and see what my DB consume a lot of memory (up to 20GB) and db size is growing up (up to 2.5GB).  
database has about 500KB data

Ben Darnell @bdarnell 03:29
@DBuTbKa the memory usage is the rocksdb cache, which defaults to 25% of the memory available to the container (and it will eventually grow to use all that space regardless of how much real data there is). if you need to limit the cache usage independently of the container size (or if you’re not using containers), you can use the —cache-size flag
the data usage is the time-series data used by graphs on the admin UI. we keep this data around for two weeks so it will stop growing at that point (and we may make this less space-hungry in the future)
```

a-robinson:
This got hit again in https://github.com/cockroachdb/cockroach/issues/15993. It'd be really nice if we had a well-written docs entry to point people to. How quickly can we get something up? If help would speed things up, let me know where you think we should put something about it and I can write up an explanation.

sploiselle:
For the time being @a-robinson, if you want to throw an explanation in `frequently-asked-questions.md`, that would be great.

a-robinson:
Looking at it now, the `frequently-asked-questions.md` page only has questions that someone would ask when investigating whether to use cockroach. It isn't a useful page for people that are already using it and have questions about it.

The SQL FAQ has questions that current users would care about, but they're all obviously SQL related, which this really isn't. Ideally there'd be a different page for operational questions or we'd slightly re-theme the SQL FAQ page.

I'll send a PR adding it to the Known Limitations page for now, and a better place can be found later.

sploiselle:
Sorry, @a-robinson; that's a good point––I recommended throwing it there as a punt for the time being until we figure out a better knowledge base-style place for it, or expand an operator/DBA-focused FAQ.

[...]

PR #1431, opened 2017-05-18
Explain background disk and memory usage growth

a-robinson:
To address #957. I'll move these wherever you want them, but put them here for the sake of sending them out.

If they look good, I'll also throw them up on stack overflow.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1431)

### DIFF
--- a/known-limitations.md
+++ b/known-limitations.md
@@ -195,3 +195,13 @@ Given a query like `SELECT * FROM foo WHERE a > 1 OR b > 2`, even if there are a
 ## Privileges for `DELETE` and `UPDATE`
 
 Every [`DELETE`](delete.html) or [`UPDATE`](update.html) statement constructs a `SELECT` statement, even when no `WHERE` clause is involved. As a result, the user executing `DELETE` or `UPDATE` requires both the `DELETE` and `SELECT` or `DELETE` and `UPDATE` [privileges](privileges.html) on the table.
+
+## Background increase in amount of data
+
+The timeseries data used to power the graphs in the admin UI is stored within the cluster and accumulates for 30 days before it starts getting truncated. As a result, for the first 30 days or so of a cluster's life you will see a steady increase in disk usage and the number of ranges in the cluster even if you aren't writing data to it yourself.
+
+As of the 1.0 release, there is no way to change the number of days before timeseries data gets truncated. As a workaround, however, you can start each node with the `COCKROACH_METRICS_SAMPLE_INTERVAL` environment variable set higher than its default of `10s` to store fewer data points. For example, you could set it to `1m` to only collect data every 1 minute, which would result in storing 6x less timeseries data than the default setting.
+
+## Increasing memory usage despite lack of traffic
+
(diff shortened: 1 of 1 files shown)


known-limitations.md
```diff
@@ -195,3 +195,11 @@ Given a query like `SELECT * FROM foo WHERE a > 1 OR b > 2`, even if there are a
 ## Privileges for `DELETE` and `UPDATE`
 
 Every [`DELETE`](delete.html) or [`UPDATE`](update.html) statement constructs a `SELECT` statement, even when no `WHERE` clause is involved. As a result, the user executing `DELETE` or `UPDATE` requires both the `DELETE` and `SELECT` or `DELETE` and `UPDATE` [privileges](privileges.html) on the table.
+
+## Background increase in amount of data
+
+The timeseries data used to power the graphs in the admin UI is stored within the cluster and accumulates for 30 days before it starts getting truncated. As a result, for the first 30 days or so of a cluster's life you will see a steady increase in disk usage and the number of ranges in the cluster even if you aren't writing data to it yourself. As of the 1.0 release, there is not yet a way to delete the timeseries data sooner than after 30 days, but if you want to save data point less frequently you can start each node with the `COCKROACH_METRICS_SAMPLE_INTERVAL` environment variable set higher than its default of `10s`. For example, you could set it to `1m` to only collect data every 1 minute, which would result in storing 6x less timeseries data than the default setting.
```
jseldess: Looks good overall. Just a few nits:

- Start new paragraph at `As of the 1.0 release`, and split that sentence in two with a bit more punctuation:
  > As of the 1.0 release, there is no way to change the number of days before timeseries data gets truncated. As a workaround, however, you can start each node with the `COCKROACH_METRICS_SAMPLE_INTERVAL` environment variable set higher than its default of `10s`. For example, you could set it to `1m` to only collect data every 1 minute, which would result in storing 6x less timeseries data than the default setting.

jseldess: Btw, this feels like feels like it does belong in know limitations, if we plan to make it possible to change the truncation period. If so, perhaps we should title it something like `Truncation period for timeseries data`, since the inability to change that period is the limitation?

known-limitations.md
```diff
@@ -195,3 +195,11 @@ Given a query like `SELECT * FROM foo WHERE a > 1 OR b > 2`, even if there are a
...
+
+## Increasing memory usage despite lack of traffic
```
jseldess: This feels like it would belong more on an `Admin FAQ` page. @sploiselle, thoughts?

a-robinson: Yup, I'd say both of them belong on an Admin/Operational FAQ page.

known-limitations.md
a-robinson: Done.

As for your second comment -- the lack of configurability is a limitation, yes, but I expect that many more people will be confused by the increase in data usage than will be coming to this page looking for a way to store less timeseries data.

known-limitations.md
jseldess: Alex. Sorry this has been sitting for a while. I'll take over the PR and move this content to a new Admin/Operations FAQ page.

a-robinson: Thanks!

jseldess:
Continuing in https://github.com/cockroachdb/docs/pull/1450/files. Closing this PR.

PR #1441, opened 2017-05-20
Add info on setting zone configs for system ranges

### LINKED ISSUE #1280: storage: add the ability to set/adjust zone configs for system ranges
PRs: 
- https://github.com/cockroachdb/cockroach/pull/14740
- https://github.com/cockroachdb/cockroach/pull/12513

a-robinson:
Fixes #1280

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1441)

### DIFF
--- a/configure-replication-zones.md
+++ b/configure-replication-zones.md
@@ -25,6 +25,8 @@ There are three replication zone levels:
 
 When replicating a piece of data, CockroachDB uses the most granular zone available: If there's a replication zone for the table containing the data, CockroachDB uses it; otherwise, it uses the replication zone for the database containing the data. If there's no applicable table or database replication zone, CockroachDB uses the cluster-wide replication zone.
 
+In addition to the databases and tables that are visible via SQL, CockroachDB stores additional internal data in what are called system ranges. You can configure replication zones for parts of these internal data ranges if you'd like to override the cluster-wide settings. See [Create a Replication Zone for System Ranges](#create-a-replication-zone-for-system-ranges) for more details.
+
 ### Replication Zone Format
 
 A replication zone is specified in [YAML](https://en.wikipedia.org/wiki/YAML) format and looks like this:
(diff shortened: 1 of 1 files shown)


a-robinson:
I can add a scenario, but we may actually want to wait on publishing this due to the bug that I found while testing the docs (https://github.com/cockroachdb/cockroach/pull/16048). While the core logic inside the database works fine, the CLI for getting/setting them doesn't handle partial zone configs properly. I originally only tested with entire config files, which worked, but found the other day that just setting one field (as in `echo 'num_replicas: 7' | cockroach zone set .meta --insecure -f -`) doesn't work.

If we cherry-pick #16048 into 1.0.1 or 1.0.2, we can publish these changes afterwards.

a-robinson:
Added a scenario. Let me know what you think.

a-robinson:
Did you forget to add the comments/questions?

I don't think we need doc updates if we don't publish this until the bug fix has been released. The existence of the three special zone configs is entirely undocumented.

configure-replication-zones.md
```diff
@@ -381,7 +422,7 @@ There's no need to make zone configuration changes; by default, the cluster is c
 
 **Scenario:**
 
-- You have a 7 nodes, 5 with SSD drives and 2 with HDD drives.
+- You have 7 nodes, 5 with SSD drives and 2 with HDD drives.
```
jseldess: Can you make the same correction under **Even Replication Across Datacenters**?

configure-replication-zones.md
```diff
@@ -437,6 +478,86 @@ There's no need to make zone configuration changes; by default, the cluster is c
 
    Data in the table will be replicated 5 times, and the positive constraint will place data in the table on nodes with `ssd` drives whenever possible.
 
+### Tweaking the replication of system ranges
```
jseldess: nit: Let's use title case:

> Tweaking the Replication of System Ranges

configure-replication-zones.md
```diff
@@ -25,6 +25,8 @@ There are three replication zone levels:
 
 When replicating a piece of data, CockroachDB uses the most granular zone available: If there's a replication zone for the table containing the data, CockroachDB uses it; otherwise, it uses the replication zone for the database containing the data. If there's no applicable table or database replication zone, CockroachDB uses the cluster-wide replication zone.
 
+In addition to the databases and tables that are visible via SQL, CockroachDB stores some additional internal data in what are called system ranges. You can configure replication zones for parts of these internal data ranges if you'd like to override the cluster-wide settings. See [Create a Replication Zone for System Ranges](#create-a-replication-zone-for-system-ranges) for more details.
```
jseldess: nit: ...CockroachDB stores ~~some~~ additional internal data...

jseldess:
Sorry, @a-robinson. Not sure why those comments got lost. 

And I didn't realize the bug you mentioned relates only to zone configs for these special ranges. Makes sense to just hold off publishing this until the patch release then.

configure-replication-zones.md
a-robinson: Done.

configure-replication-zones.md
a-robinson: Done.

configure-replication-zones.md
a-robinson: Done.

jseldess:
Excellent. Thanks for handling this, @a-robinson!

jseldess:
Merging ahead of the 1.0.2 release notes so I can link to the new content.

PR #1465, opened 2017-05-25
Update commands to work as designed.

DBarney:
Missing backslash causes command to not background correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1465)

### DIFF
--- a/demo-data-replication.md
+++ b/demo-data-replication.md
@@ -123,7 +123,7 @@ $ cockroach start --insecure \
 --host=localhost \
 --port=26258 \
 --http-port=8081 \
---join=localhost:26257
+--join=localhost:26257 \
 --background
 
 # Add node 3:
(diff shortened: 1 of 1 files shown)


benesch:
Nice catch, @DBarney, and thanks for the fix! Much appreciated.

DBarney:
No problem, I was busy going through the docs trying things out when those two commands hung. Figured I should fix it if no one else already had.

jseldess:
Yes, thanks very much, @DBarney! Merging now. But I'm curious. Did you use the on-page **Contribute > Edit This Page** option to make this change?

DBarney:
I did actually, I was prepared to fork the repo and do it on my machine, but saw the option and decided to try it out. It was actually pretty slick!

jseldess:
Very nice to hear! Thanks, @DBarney.

PR #1473, opened 2017-05-26
Update activerecord sample to actually set the primary key

### LINKED ISSUE #1461: Build a Ruby App with CockroachDB Doc Update
Re: [Build a Ruby App with CockroachDB](https://cockroachlabs.com/docs//build-a-ruby-app-with-cockroachdb-activerecord.html)

## Issue Description

From the ActiveRecord docs:
```The name of the primary key, if one is to be added automatically. Defaults to id. If :id is false this option is ignored.

Note that Active Record models will automatically detect their primary key. This can be avoided by using self.primary_key= on the model to define the key explicitly.
```

However, in the docs, it shows `create_table :accounts, force: true, id: false, primary_key: 'id'`. That has caused me some confusion since it definitely does not create the `id` column as the primary key.

## Suggested Resolution 

Remove `primary_key` option and include a mention to using `self.primary_key=` on the model.

jseldess:
Fixes #1461

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1473)

### DIFF
--- a/_includes/app/activerecord-basic-sample.rb
+++ b/_includes/app/activerecord-basic-sample.rb
@@ -25,8 +25,7 @@ end
 # In Rails, this would go in db/migrate/ as usual.
 class Schema < ActiveRecord::Migration
   def change
-    create_table :accounts, force: true, id: false, primary_key: 'id' do |t|
-      t.integer :id
+    create_table :accounts, force: true do |t|
       t.integer :balance
     end
   end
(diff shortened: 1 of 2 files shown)


jseldess:
@jordanlewis, seems like this small change is all that's needed?

jordanlewis:
Actually, I think this is more confusing than helpful. Rails auto-creates an integer primary `id` column with a reasonable autoincrementing default if you don't specify anything else.

So the schema migration should look like this in our sample:

```
class Schema < ActiveRecord::Migration
  def change
    create_table :accounts, force: true do |t|
      t.integer :balance
    end
  end
end
```

jseldess:
Yep, that's a lot simpler, @jordanlewis. Please take another look.

jordanlewis:
LGTM

PR #1476, opened 2017-05-30
Update footer to match hugo website

deronjohnson:
@jseldess

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1476)


jseldess:
Thanks, @deronjohnson, looks good overall, but the line height is a bit different between **CONNECT WITH US** and all the other footer sections. And if you switch between the docs staging and the live homepage, there's also a discrepancy in terms of line hight/tightness.

deronjohnson:
Should be resolved now @jseldess.

jseldess:
Thanks, @deronjohnson. The line height issue looks good now, but the footer seems to have a slightly smaller bottom margin on the main site vs. the docs. Can you take a look?

deronjohnson:
@jseldess I believe that fixed it. There's a global line height set on the body that doesn't exist on the main site, so it needs to be overwritten.

jseldess:
Yep. LGTM. Thanks, @deronjohnson!

PR #1479, opened 2017-05-31
Clarify max-offset update instructions

bdarnell:
In gitter, @freeaqingme mentioned that the docs were ambiguous about whether or not a rolling restart is allowed.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1479)


sploiselle:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[start-a-node.md, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1479#-KlTv9pkbCeh61NSemG-:-KlTv9pkbCeh61NSemG0:b-gm54mt) ([raw file](https://github.com/cockroachdb/docs/blob/441af8174486188fa87295c0ae95a741426356a8/start-a-node.md#L48)):*
> ```Markdown
> `--listening-url-file` | The file to which the node's SQL connection URL will be written on successful startup, in addition to being printed to the [standard output](#standard-output).<br><br>This is particularly helpful in identifying the node's port when an unused port is assigned automatically (`--port=0`).
> `--locality` | Arbitrary key-value pairs that describe the locality of the node. Locality might include country, region, datacenter, rack, etc.<br><br>CockroachDB attempts to spread replicas evenly across the cluster based on locality, with the order determining the priority. The keys themselves and the order of key-value pairs must be the same on all nodes, for example:<br><br>`--locality=region=east,datacenter=us-east-1`<br>`--locality=region=west,datacenter=us-west-1`<br><br>These can be used to influence the location of data replicas. See [Configure Replication Zones](configure-replication-zones.html#replication-constraints) for full details.
> `--max-offset` | The maximum allowed clock offset for the cluster. If observed clock offsets exceed this limit, servers will crash to minimize the likelihood of reading inconsistent data. Increasing this value will increase the time to recovery of failures as well as the frequency of uncertainty-based read restarts.<br><br>Note that this value must be the same on all nodes in the cluster. In order to change it, every node in the cluster must be stopped at once and then all must be restarted with the new value.<br><br>**Default:** 500ms
> ```

Since this is the raison d'être for the change, I recommend adding "(i.e. you cannot perform a rolling restart)."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1479)*

jseldess:
:lgtm:, but with a few questions/comments.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1479#-:-KlU6TeONc9lYtVmZ9Ey:b7wkuqj)*

bdarnell:
@jseldess What are those questions/comments? They didn't seem to come through.

jseldess:
*[start-a-node.md, line 48 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1479#-KlTv9pkbCeh61NSemG-:-KlU6e8KEqL17NVssJIU:b-12s2kt) ([raw file](https://github.com/cockroachdb/docs/blob/441af8174486188fa87295c0ae95a741426356a8/start-a-node.md#L48)):*
What does "at once" really mean? They need to stop all node simultaneously? If they stop nodes sequentially, they'll have problems? 

If the point is that all nodes must be stopped, I think we could make the language more explicit, e.g.:
> In order to change it, first stop every node in the cluster. Then, once the entire cluster is offline, restart each node with the new value.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1479)*

bdarnell:
Incorporated both Sean and Jesse's suggestions.

bdarnell:
"Start" sounds better than "restart" to me here. In my mind "start" is used for the first time and "restart" for all subsequent times. "restart" may sometimes mean "stop and start" but not necessarily.

jseldess:
@bdarnell, you last comment seems to suggest that you want to change `restart` to `start`? I feel that `restart` is more accurate in the case that you're not starting a node for the first time, but I'm fine with whatever you think is clearest.

bdarnell:
Sorry, I got mixed up. I meant I prefer "restart".

jseldess:
Great. Thanks for this update, @bdarnell. And thanks for the prompt, @Freeaqingme!

PR #1480, opened 2017-05-31
Fix the other instance of sql.trace.txn.threshold.

petermattis:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1480)


Freeaqingme:
Thanks guys, was about to create an report for precisely this issue :+1:

PR #1486, opened 2017-06-01
Improve the doc for CREATE TABLE ... AS

### LINKED ISSUE #1467: CREATE TABLE AS Doc Update - vague language
Re: [CREATE TABLE AS](https://cockroachlabs.com/docs//create-table-as.html)

The text says:
> These tables are not designed for long-term use, as they do not support some common table features like Primary Keys and interleaving.

IMHO, that's not a very useful comment. "not designed for long-term use" doesn't mean anything to me :). Also, "they don't support ..." is not quite accurate. These are regular tables, they support whatever all other tables support.
I think we should concretely describe the schema that these tables will have and then say that if that's not the schema you want, go do something else. I believe this schema will be an autogenerated PK column, and the other cols will have the date type inferred by our type inference rules from the SELECT (e.g. if a col from another table is selected, we'll get its type), and they'll all be NULL-able. @knz would probably have better words (the best words?) for this.

knz:
Fixes #1467

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1486)

### DIFF
--- a/create-table-as.md
+++ b/create-table-as.md
@@ -1,18 +1,36 @@
 ---
 title: CREATE TABLE AS
-summary: The CREATE TABLE AS statement creates a new table from the results of a SELECT statement.
+summary: The CREATE TABLE AS statement persists the result of a query into the database for later reuse.
 toc: false
 ---
 
-The `CREATE TABLE AS` statement creates a new table from the results of a [`SELECT`](select.html) statement.
-
-These tables are not designed for long-term use, as they do not support some common table features like [Primary Keys](primary-key.html) and [interleaving](interleave-in-parent.html). For similar functionality with more robust feature support, [create a table](create-table.html) and then use [`INSERT INTO SELECT`](insert.html#insert-from-a-select-statement).
+The `CREATE TABLE ... AS` statement creates a new table from the results of a query.
 
 <div id="toc"></div>
 
+## Intended Use
+
+Tables created with `CREATE TABLE ... AS` are intended to persist the
+result of a query for later reuse.
+
+This can be more efficient than a [view](create-view.html) when the
+following two conditions are met:
+
+- The result of the query is used as-is multiple times.
+- The copy needs not be kept up-to-date with the original table over time.
+
+When the results of a query are reused multiple times within a larger
+query, a view is advisable instead. The query optimizer can "peek"
+into the view and optimize the surrounding query using the primary key
+and indices of the tables mentioned in the view query.
+
+A view is also advisable when the results must be up-to-date; a view
+always retrieves the current data from the tables that the view query
+mentions.
+
 ## Required Privileges
 
-The user must have the `CREATE` [privilege](privileges.html) on the parent database. 
+The user must have the `CREATE` [privilege](privileges.html) on the parent database.
 
 ## Synopsis
 
@@ -31,11 +49,78 @@ table td:first-child {
 | `IF NOT EXISTS` | Create a new table only if a table of the same name does not already exist in the database; if one does exist, do not return an error.<br><br>Note that `IF NOT EXISTS` checks the table name only; it does not check if an existing table has the same columns, indexes, constraints, etc., of the new table. |
 | `any_name` | The name of the table to create, which must be unique within its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). When the parent database is not set as the default, the name must be formatted as `database.name`.<br><br>The [`UPSERT`](upsert.html) and [`INSERT ON CONFLICT`](insert.html) statements use a temporary table called `excluded` to handle uniqueness conflicts during execution. It's therefore not recommended to use the name `excluded` for any of your tables. |
 | `name` | The name of the column you want to use instead of the name of the column from `select_stmt`. |
-| `select_stmt` | The [`SELECT`](select.html) statement whose results you want to use to create the table. |
+| `select_stmt` | The query whose results you want to use to create the table. This can use [`SELECT`](select.html), `TABLE` or `VALUES`. |
+
+## Limitations
+
+The [primary key](primary-key.html) of tables created with `CREATE
+TABLE ... AS` is not derived from the query results. Like for other
+tables, it is not possible to add or change the primary key after
+creation. Moreover, these tables are not
+[interleaved](interleave-in-parent.html) with other tables. The
+default rules for [column families](column-families.html) apply.
+
+For example:
+
+~~~ sql
+> CREATE TABLE logoff (
+    user_id INT PRIMARY KEY,
+    user_email STRING UNIQUE,
+    logoff_date DATE NOT NULL,
+);
+> CREATE TABLE logoff_copy AS TABLE logoff;
+> SHOW CREATE TABLE logoff_copy;
+~~~
+~~~
++-------------+-----------------------------------------------------------------+
+|    Table    |                           CreateTable                           |
++-------------+-----------------------------------------------------------------+
+| logoff_copy | CREATE TABLE logoff_copy (␤                                     |
+|             |     user_id INT NULL,␤                                          |
+|             |     user_email STRING NULL,␤                                    |
+|             |     logoff_date DATE NULL,␤                                     |
+|             |     FAMILY "primary" (user_id, user_email, logoff_date, rowid)␤ |
+|             | )                                                               |
++-------------+-----------------------------------------------------------------+
+(1 row)
+~~~
+
+The example illustrates that the primary key, unique and "not null"
+constraints are not propagated to the copy.
+
+It is however possible to
+[create a secondary index](create-index.html) after `CREATE TABLE
+... AS`.
+
+For example:
+
+~~~ sql
+> CREATE INDEX logoff_copy_id_idx ON logoff_copy(user_id);
+> SHOW CREATE TABLE logoff_copy;
+~~~
+~~~
++-------------+-----------------------------------------------------------------+
+|    Table    |                           CreateTable                           |
++-------------+-----------------------------------------------------------------+
+| logoff_copy | CREATE TABLE logoff_copy (␤                                     |
+|             |     user_id INT NULL,␤                                          |
+|             |     user_email STRING NULL,␤                                    |
+|             |     logoff_date DATE NULL,␤                                     |
+|             |     INDEX logoff_copy_id_idx (user_id ASC),␤                    |
+|             |     FAMILY "primary" (user_id, user_email, logoff_date, rowid)␤ |
+|             | )                                                               |
++-------------+-----------------------------------------------------------------+
+(1 row)
+~~~
+
+For maximum data storage optimization, consider using separately
+[`CREATE`](create-table.html) followed by
+[`INSERT INTO ...`](insert.html) to populate the table using the query
+results.
 
 ## Examples
 
-### Create a Table from a `SELECT` Statement
+### Create a Table from a `SELECT` Query
 
 ~~~ sql
 > SELECT * FROM customers WHERE state = 'NY';
@@ -64,6 +149,11 @@ table td:first-child {
 
 ### Change Column Names
 
+
+
+This statement creates a copy of an existing table but with changed column names.
+
+
 ~~~ sql
 > CREATE TABLE customers_ny (id, first_name) AS SELECT id, name FROM customers WHERE state = 'NY';
 
@@ -78,9 +168,49 @@ table td:first-child {
 +----+------------+
 ~~~
 
+### Create a Table from a `VALUES` Clause
+
+~~~ sql
+> CREATE TABLE tech_states AS VALUES ('CA'), ('NY'), ('WA');
+
+> SELECT * FROM tech_states;
+~~~
+~~~
++---------+
+| column1 |
++---------+
+| CA      |
+| NY      |
+| WA      |
++---------+
+(3 rows)
+~~~
+
+
+### Create a Copy of an Existing Table
+
+~~~ sql
+> CREATE TABLE customers_ny_copy AS TABLE customers_ny;
+
+> SELECT * FROM customers_ny_copy;
+~~~
+~~~
++----+------------+
+| id | first_name |
++----+------------+
+|  6 | Dorotea    |
+| 15 | Thales     |
++----+------------+
+~~~
+
+When a table copy is created this way, the copy is not associated to
+any primary key, secondary index or constraint that was present on the
+original table.
+
 ## See Also
 
 - [`CREATE TABLE`](create-table.html)
+- [`CREATE VIEW`](create-view.html)
 - [`SELECT`](select.html)
 - [`INSERT`](insert.html)
 - [`DROP TABLE`](drop-table.html)


jseldess:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[create-table-as.md, line 2 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZtbCyHVpjNoMHglvv:-KlZtbCyHVpjNoMHglvw:bq1gp9q) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L2)):*
> ```Markdown
> ---
> title: CREATE TABLE AS
> ```

Should we change the title to `CREATE TABLE ... AS`? Do you think we should change this in the sidenav and in the `sql-statements.md` list as well?

---

*[create-table-as.md, line 3 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZtSv9xNa_PUnwAbzf:-KlZtSv9xNa_PUnwAbzg:btr1y4x) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L3)):*
> ```Markdown
> ---
> title: CREATE TABLE AS
> summary: The CREATE TABLE AS statement creates a new table from the results of a SELECT statement.
> ```

We should update the `summary` field in the front-matter based on the final version of the intro sentence.

---

*[create-table-as.md, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZoteI82bXBOPJ7q4-:-KlZoteI82bXBOPJ7q40:b-aaqv3z) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L7)):*
I like that you're honing in the the ultimate purpose of using this statement, but it seems confusing to me that we're not directly stating that the result of this statement is a new table. And the paragraph following is a bit hard to parse.  What would you think about combining these versions, e.g.:

> The `CREATE TABLE ... AS` statement creates a new table from the result of a query.  

> Tables created with `CREATE TABLE ... AS` are generally used to persist the result of a query for later use. However, they are not optimized for data storage or retrieval since they are created without a primary key and cannot be interleaved into other tables. For maximum data optimization, use `CREATE TABLE` followed by `INSERT` statements.

I know I've left out the bit about secondary indexes being allowed. Might want to add mention of that back in, as well as the ability to add constraints after the fact.

---

*[create-table-as.md, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZxebIXnfMnuNdJ6mI:-KlZxebIXnfMnuNdJ6mJ:b8op0ez) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L74)):*
It's a great idea to add some explanation, but I think this could be clearer. What about:

> This statement creates a copy of an existing table but with changed column names. 

---

*[create-table-as.md, line 111 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZvZX8wAYt8cHVyvKz:-KlZvZX8wAYt8cHVyvL-:bc1xf2z) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L111)):*
> ```Markdown
> 
> 
> ### Create an Unstructured Copy of an Existing Table
> ```


I like this title, but "unstructured" seems confusing. I think we could get away with just `Create a Copy of an Existing Table`. Perhaps it would make most sense to use this is the first example, followed by the current first example renamed to `Create a Partial Copy of an Existing Table`. Following each example with an explanation that only the schema and data are copied, not the primary key, indexes, or constraints. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486)*

knz:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[create-table-as.md, line 2 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZtbCyHVpjNoMHglvv:-Kl_IWjyQLNEOnFUhaxr:bdydz4h) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L2)):*

Should we change the title to `CREATE TABLE ... AS`? Do you think we should change this in the sidenav and in the `sql-statements.md` list as well?
</blockquote></details>

I don't think it matters much. Perhaps the current form is better because it is shorter.

---

*[create-table-as.md, line 3 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZtSv9xNa_PUnwAbzf:-Kl_I_AdTnAUQdeKh6V4:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L3)):*

We should update the `summary` field in the front-matter based on the final version of the intro sentence.
</blockquote></details>

Done.

---

*[create-table-as.md, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZoteI82bXBOPJ7q4-:-Kl_LzDnxPoR3B7TFl-t:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L7)):*

I like that you're honing in the the ultimate purpose of using this statement, but it seems confusing to me that we're not directly stating that the result of this statement is a new table. And the paragraph following is a bit hard to parse.  What would you think about combining these versions, e.g.:

> The `CREATE TABLE ... AS` statement creates a new table from the result of a query.  

> Tables created with `CREATE TABLE ... AS` are generally used to persist the result of a query for later use. However, they are not optimized for data storage or retrieval since they are created without a primary key and cannot be interleaved into other tables. For maximum data optimization, use `CREATE TABLE` followed by `INSERT` statements.

I know I've left out the bit about secondary indexes being allowed. Might want to add mention of that back in, as well as the ability to add constraints after the fact.
</blockquote></details>

Done.

---

*[create-table-as.md, line 74 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZxebIXnfMnuNdJ6mI:-Kl_M2dF_Bc3WSF5VN3E:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L74)):*

It's a great idea to add some explanation, but I think this could be clearer. What about:

> This statement creates a copy of an existing table but with changed column names. 
</blockquote></details>

Done.

---

*[create-table-as.md, line 111 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZvZX8wAYt8cHVyvKz:-Kl_M7wwKC0-cYkLZJU7:b-i1aynl) ([raw file](https://github.com/cockroachdb/docs/blob/a6de2f9c56bf5670d3afc199c21244be50f90f06/create-table-as.md#L111)):*

I like this title, but "unstructured" seems confusing. I think we could get away with just `Create a Copy of an Existing Table`. Perhaps it would make most sense to use this is the first example, followed by the current first example renamed to `Create a Partial Copy of an Existing Table`. Following each example with an explanation that only the schema and data are copied, not the primary key, indexes, or constraints. 
</blockquote></details>

I have added new sections to make this point better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486)*

create-table-as.md
```diff
@@ -31,11 +47,75 @@ table td:first-child {
...
+It is however possible to
+[create a secondary index](create-index.html) after `CREATE TABLE
+... AS`.
+
+For example:
+
+~~~ sql
+> CREATE INDEX logoff_copy_id_idx ON logoff_copy(user_id);
+> SHOW CREATE TABLE logoff_copy;
++-------------+-----------------------------------------------------------------+
```
jseldess: Can you put the response in separate code block, as you've done above?

create-table-as.md
```diff
@@ -31,11 +47,75 @@ table td:first-child {
...
 | `any_name` | The name of the table to create, which must be unique within its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). When the parent database is not set as the default, the name must be formatted as `database.name`.<br><br>The [`UPSERT`](upsert.html) and [`INSERT ON CONFLICT`](insert.html) statements use a temporary table called `excluded` to handle uniqueness conflicts during execution. It's therefore not recommended to use the name `excluded` for any of your tables. |
 | `name` | The name of the column you want to use instead of the name of the column from `select_stmt`. |
-| `select_stmt` | The [`SELECT`](select.html) statement whose results you want to use to create the table. |
+| `select_stmt` | The query whose results you want to use to create the table. This can use [`SELECT`](select.html), `TABLE` or `VALUES`. |
+
+## Limitations
+
+The [primary key](primary-key.html) of tables created with `CREATE
```
jseldess: Nice!

create-table-as.md
```diff
@@ -1,18 +1,34 @@
...
+This can be more efficient than a [view](create-view.html) when the following two conditions are met:
+- the result of the query is used as-is multiple times;
+- the copy needs not be kept up-to-date with the original table over time.
+
+When the results of a query are reused multiple times within a larger
+query, a view is advisable instead: the query optimizer can "peek"
+into the view and optimize the surrounding query using the primary key
+and indices of the tables mentioned in the view query.
+
+A view is also advisable when the results must be up-to-date: q view
```
jseldess: Colon should be a semicolon. But feel free to use a period instead.

Also, `q` should be `A`.

create-table-as.md
```diff
@@ -1,18 +1,34 @@
...
+## Intended use
+
+Tables created with `CREATE TABLE ... AS` are intended to persist the
+result of a query for later reuse.
+
+This can be more efficient than a [view](create-view.html) when the following two conditions are met:
+- the result of the query is used as-is multiple times;
+- the copy needs not be kept up-to-date with the original table over time.
+
+When the results of a query are reused multiple times within a larger
```
jseldess: Colon should be a semicolon. But feel free to use a period instead.

create-table-as.md
```diff
@@ -1,18 +1,34 @@
...
 ---
 
-The `CREATE TABLE AS` statement creates a new table from the results of a [`SELECT`](select.html) statement.
-
-These tables are not designed for long-term use, as they do not support some common table features like [Primary Keys](primary-key.html) and [interleaving](interleave-in-parent.html). For similar functionality with more robust feature support, [create a table](create-table.html) and then use [`INSERT INTO SELECT`](insert.html#insert-from-a-select-statement).
+The `CREATE TABLE ... AS` statement creates a new table from the results of a query.
 
 <div id="toc"></div>
 
+## Intended use
```
jseldess: Nice! Just a few nits:

- Use title case for the header: `Intended Use`
- Make each bullet a sentence, that is, capitalize `The` and use a period instead of a semicolon.

knz:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[create-table-as.md, line 11 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-11:-Kl_RxOKSKuCJOxqdphG:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L11)):*

Nice! Just a few nits:

- Use title case for the header: `Intended Use`
- Make each bullet a sentence, that is, capitalize `The` and use a period instead of a semicolon.
</blockquote></details>

Done.

---

*[create-table-as.md, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-20:-Kl_S43dFH1buekaHD7v:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L20)):*

Colon should be a semicolon. But feel free to use a period instead.
</blockquote></details>

Done.

---

*[create-table-as.md, line 25 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-25:-Kl_S4pizUXLEo6bgH6d:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L25)):*

Colon should be a semicolon. But feel free to use a period instead.

Also, `q` should be `A`.
</blockquote></details>

Done.

---

*[create-table-as.md, line 97 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-97:-Kl_S8BOS6VOYb_gzaDH:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L97)):*

Can you put the response in separate code block, as you've done above?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486)*

andreimatei:
:lgtm:

Thanks!

The one observation I have is that, given the grammar diagram on this page, it was surprising to me that CREATE TABLE AS TABLE x is allowed. I think the diagram suggest that it isn't, as `select_stmt` needs to start with a `SELECT`. Perhaps we can improve the diagram?

---

Review status: 0 of 1 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[create-table-as.md, line 59 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-Kl_bXOK8Y5vBTAA8PGA:-Kl_bXOK8Y5vBTAA8PGB:b-jqmvc8) ([raw file](https://github.com/cockroachdb/docs/blob/221ef181ce1b41c1814ce57336b5071fdd0cf21e/create-table-as.md#L59)):*
> ```Markdown
> tables, it is not possible to add or change the primary key after
> creation. Moreover, these tables are not
> [interleaved](interleave-in-parent.html) with other tables.
> ```

consider saying something about the column family that will be used for the new table (or say that there's no control over col families and the default is used).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-:-Kl_bJ6cGerQY3vtwI60:b-cfu65o)*

jseldess:
:lgtm:, with a few final nits. Thanks, @knz!

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[create-table-as.md, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-20:-KlaedsvWbZg5ACL-y8h:b-k0rezg) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L20)):*

Done.
</blockquote></details>

Looks like this didn't change...

---

*[create-table-as.md, line 25 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-25:-Klael5eM-5_ks3xyir4:b-eidb50) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L25)):*

Done.
</blockquote></details>

Looks like the colon still needs to be changed to a semicolon or period.

---

*[create-table-as.md, line 16 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlaeaHyakAXLCmil7k8:-KlaeaHyakAXLCmil7k9:bauir3d) ([raw file](https://github.com/cockroachdb/docs/blob/221ef181ce1b41c1814ce57336b5071fdd0cf21e/create-table-as.md#L16)):*
> ```Markdown
> result of a query for later reuse.
> 
> This can be more efficient than a [view](create-view.html) when the following two conditions are met.
> ```

nit: Best to still introduce the list with a colon, as you had previously. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-:-Klaf0ZmeDQCA-0YWoJj:bamy23f)*

knz:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[create-table-as.md, line 20 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-20:-Kld3GPbbcL_KM_1vO3a:b8ihy9h) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L20)):*

Looks like this didn't change...
</blockquote></details>

My bad; fixed.

---

*[create-table-as.md, line 25 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlZQClBG3Fla2mzbiOI-r2-25:-Kld3KpB7mIfjtER5VW1:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/be99d05c534006509c5a3eb8a268f618b32c8957/create-table-as.md#L25)):*

Looks like the colon still needs to be changed to a semicolon or period.
</blockquote></details>

Done.

---

*[create-table-as.md, line 16 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-KlaeaHyakAXLCmil7k8:-Kld33CVISlTuTQ13nxQ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/221ef181ce1b41c1814ce57336b5071fdd0cf21e/create-table-as.md#L16)):*

nit: Best to still introduce the list with a colon, as you had previously. 
</blockquote></details>

Done.

---

*[create-table-as.md, line 59 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-Kl_bXOK8Y5vBTAA8PGA:-Kld3XhFKNeEPZtsAiYf:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/221ef181ce1b41c1814ce57336b5071fdd0cf21e/create-table-as.md#L59)):*

consider saying something about the column family that will be used for the new table (or say that there's no control over col families and the default is used).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486)*

knz:
There's some TC error which I do not understand. Can't merge this myself.

sploiselle:
Re-running TC tests for you

jseldess:
One note on the diagram: We could probably tweak it to show a simplified version of https://www.cockroachlabs.com/docs/sql-grammar.html#simple_select, perhaps with just `SELECT`, `VALUES`, and `TABLE` vertically stacked in place of `select_stmt`. Happy to play with that in a follow-up, if you like.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1486#-:-KldN8n4cyvhc25mIh7X:b-jyb8r7)*

knz:
@jseldess I tried to do that this morning. I have a patch ready, but I wasn't able to regenerate the diagram successfully. You can achieve the same by editing `generate/main.go` like this:

```
        {
          name:   "create_table_as_stmt",
          inline: []string{"opt_column_list", "name_list", "select_stmt", "select_no_parens", "simple_select"},
        },
```

knz:
Also perhaps this work could be a combined followup to also address #1488 :)

knz:
(The TC thing is still failing btw)

jseldess:
@knz, yes, let's handle that in a follow up, and take care of it across all the pages you mentioned in your issue. Thanks for adding clarity! The link test was failing on an external link. Rerunning often solves this. Looks like that's already restarted.

knz:
Here we go! TFYR

PR #1489, opened 2017-06-02
fix UPDATE railroad diagram generation

benesch:
Mark the `update_stmt` as `nosplit`; the update statement is now too complicated for the railroad diagram generator to condense the split form of the grammar.

This commit holds off on actually updating the update railroad digram, because I'm not sure whether it's the same in 1.0 and 1.1.

(CC'ing @knz, just in case he wasn't aware of this neat trick.)

### DIFF
--- a/generate/main.go
+++ b/generate/main.go
@@ -515,7 +515,11 @@ func main() {
 					replace: map[string]string{"stmt_list": "'CREATE' 'TABLE' table_name '(' ( column_def ( ',' column_def )* ) ( 'CONSTRAINT' name | ) 'UNIQUE' '(' ( column_name ( ',' column_name )* ) ')' ( table_constraints | ) ')'"},
 					unlink:  []string{"table_name", "check_expr", "table_constraints"},
 				},
-				{name: "update_stmt", inline: []string{"relation_expr_opt_alias", "set_clause_list", "set_clause", "single_set_clause", "multiple_set_clause", "ctext_row", "ctext_expr_list", "ctext_expr", "from_clause", "from_list", "where_clause", "returning_clause"}},
+				{
+					name:    "update_stmt",
+					inline:  []string{"relation_expr_opt_alias", "set_clause_list", "set_clause", "single_set_clause", "multiple_set_clause", "ctext_row", "ctext_expr_list", "ctext_expr", "from_clause", "from_list", "where_clause", "returning_clause"},
+					nosplit: true,
+				},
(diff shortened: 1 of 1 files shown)


knz:
I think the first sentence in the commit message is missing some words.

benesch:
Whoops, thanks! Fixed.

mjibson:
I have no recollection of what nosplit does, and apparently I didn't document it at all. Could you add a comment about nosplit so others can correctly use it in the future?

sploiselle:
Thanks @benesch (I think, though I'm not sure what this does)! Can I get you to throw in a quick description of `nosplit` in this [railroad diagram wiki](https://github.com/cockroachdb/docs/wiki/SQL-Grammar-Railroad-Diagram-Changes)?

benesch:
If I'm understanding the code correctly, if you have a rule like

```bnf
foo := bar "and" baz
bar := red_fish | blue_fish
baz := ...
```

and specify that `foo` should be inlined, you'll see something like this by default:

```bnf
foo := red_fish "and" baz | blue_fish "and" baz
```

With `nosplit`, you'll see something like this:

```bnf
foo := (red_fish | blue_fish) "and"
```

At least, I think. `nosplit` seems way better because it doesn't cause an exponential blowup for every inlined term, but I guess we needed split to properly implement `match` and `exclude`? As in, I don't think you can specify `match`, `exclude`, and `nosplit: true` all at once.

Does this jog your memory, @mjibson? I haven't yet sunk the time in to fully grok the code.

mjibson:
Yes, the explanation of what it does sounds correct. But I'm not sure about the reason we would need to do that. I don't think it's just due to the increase in number of terms. I remember something like it was because it caused some infinite recursion or something maybe.

knz:
The infinite recursion sounds familiar. I hit that yesterday. I didn't know about nosplit then though.

benesch:
Well, it wasn't quite infinite recursion, because it did eventually terminate:

screenshot 2017-06-02 11 36 40

(Unfortunately I don't have the full image.) Though I suppose it could be infinite recursion with a limit on the maximum number of calls.

knz:
ok I think you can add the explanation with a link to the screenshot to explain what nosplit does.

benesch:
Ok, added to the wiki! Is this good to merge?

jseldess:
Thanks, @benesch. LGTM!

PR #1491, opened 2017-06-05
Fix link in CHECK constraint

neoeinstein:
Correct to relative url for `ADD CONSTRAINT` link.


benesch:
Fixed TeamCity! (At least, it's running HTMLProofer now; should be finished in a few minutes.)

PR #1513, opened 2017-06-05
Add analytics attributes to links on install page

deronjohnson:
@jseldess @jess-edwards


deronjohnson:
It seems like that might be the case. Our analytics guy said they existed at some point and asked me to add them again.

jseldess:
Oh, now I remember. These links were added when we updated our build from source instructions. Previously, they were cloning the repo. Now they're downloading a source tarball. 

In any case, thanks for adding the events. I'll merge this now.

Issue #1492, opened 2017-06-05
style guide: paragraphs/code blocks in a list item 


jseldess:
After https://github.com/cockroachdb/docs/pull/1477 is merged, paragraphs or code blocks following a list item (bullet or number) need to be indented 4 times. We should add this guidance to our basic [style guide](https://github.com/cockroachdb/docs/blob/gh-pages/CONTRIBUTING.md#code-samples).

PR #1521, opened 2017-06-06
fixes spacing for redcartpet

sploiselle:
Implements the style mentioned in #1492


sploiselle:
@jseldess re:
> A nested list needs to be directly under a list item, with each nested item indented only 2 spaces?

List items need either 2 or 4 space. However, consistently doing 4 is obviously cleaner. I'm seeing how easily I can make all of these consistent based on the style guide.

PR #1526, opened 2017-06-07
Clarify docs for ORDER BY

### LINKED ISSUE #1490: Without ORDER BY, order is unspecified
One of the [examples](https://www.cockroachlabs.com/docs/select.html#order-retrieved-rows-by-one-column) states that by default, results are ordered by primary key. This is untrue: without an ORDER BY, there are no guarantees about the order of results, and in distsql the order is unpredictable (the order is based on which nodes respond the quickest). We should remove this claim, and add an explicit notice that order is unspecified unless an ORDER BY is specified (this sometimes confuses users; see cockroachdb/cockroach#16306)

jseldess:
Fixes #1490

### DIFF
--- a/select.md
+++ b/select.md
@@ -35,7 +35,7 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
 | `INTERSECT` | Only retrieve rows that exist in both the preceding and following `SELECT` statements. Returns distinct values. |
 | `EXCEPT` | Only retrieve rows that are in the preceding `SELECT` statement but not in the following `SELECT` statement.  Returns distinct values.|
 | `ALL` | Include duplicate rows in the returned values of `UNION`, `INTERSECT`, or `EXCEPT`. |
-| `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column. |
+| `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column.<br><br>When <code>ORDER BY</code> is not included in a query, rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.|
 | `LIMIT limit_val` | Only retrieve `limit_val` number of rows. |
 | `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` to "paginate" through retrieved rows. |
 
@@ -409,7 +409,6 @@ WHERE state_opened IN ('AZ', 'NY');
 +-----------------+
 ~~~
 
-
 #### Intersect: Retrieve Intersection of Two Queries
 
 `INTERSECT` finds only values that are present in both `SELECT` queries.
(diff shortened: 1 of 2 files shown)


sploiselle:
Awesome, @jseldess. I think this would also make a solid entry in the [SQL FAQ](https://www.cockroachlabs.com/docs/sql-faqs.htm).

---

---

*[select.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-Km-UTw9YMbIQCLpp8ht:-Km-UTw9YMbIQCLpp8hu:b-r5c1ww) ([raw file](https://github.com/cockroachdb/docs/blob/b2669069193d50a05fcbc3e607861c59b1c4f0ca/select.md#L38)):*
"Otherwise" is a little confusing here because it's a double-negative-like: "When this _isn't_ included..._Otherwise_..."

Maybe something like...

> When `ORDER BY` is not included in a query, rows are not sorted by any consistent criteria. Instead of ordering rows, CockroachDB returns them as the coordinating node receives them.

---

*[select.md, line 458 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-Km-WFLE6olUT3W8jm-4:-Km-WFLE6olUT3W8jm-5:b80be68) ([raw file](https://github.com/cockroachdb/docs/blob/b2669069193d50a05fcbc3e607861c59b1c4f0ca/select.md#L458)):*
> ```Markdown
> You can use an `ORDER BY` clause to sort retrieved rows by one or more columns.
> 
> {{site.data.alerts.callout_info}}When <code>ORDER BY</code> is not included in a query, CockroachDB retrieves rows as they become available from relevant nodes. Otherwise, the rows are not sorted by any consistent criteria.{{site.data.alerts.end}}
> ```

Same recommendation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-:-Km-UJo2yhkYCiU3rCS9:bq6jinp)*

jseldess:
TFTR, @sploisell!

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[select.md, line 38 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-Km-UTw9YMbIQCLpp8ht:-Km-iSom3FxOqKejwk8h:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b2669069193d50a05fcbc3e607861c59b1c4f0ca/select.md#L38)):*

"Otherwise" is a little confusing here because it's a double-negative-like: "When this _isn't_ included..._Otherwise_..."

Maybe something like...

> When `ORDER BY` is not included in a query, rows are not sorted by any consistent criteria. Instead of ordering rows, CockroachDB returns them as the coordinating node receives them.
</blockquote></details>

Done.

---

*[select.md, line 458 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-Km-WFLE6olUT3W8jm-4:-Km-iTvGR6WzPwFT2O8M:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b2669069193d50a05fcbc3e607861c59b1c4f0ca/select.md#L458)):*

Same recommendation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-:-Km-kdC6pK5AvfFXNf_4:b3zvhp9)*

sploiselle:
:lgtm: fwiw

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-:-Km-lIdSOU6JkH8LqXbH:bb71oku)*

bdarnell:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1526#-:-Km2BL1Tgv0r_ZRLLBhV:bnfp4nl)*

Issue #1132, opened 2017-02-28
Minimum and recommended hardware configuration per cloud provider


jseldess:
Minimum and recommended, per cloud provider:
- CPU
- RAM
- Disk types

Also related: #960 (configuring NTP)

jseldess:
@mberhault, can you help me here? If so, I'll set up time.

mberhault:
sure.
The basic hardware profile we're using right now is:
AWS: m3.large (2cpu/7.5Ram)
GCE: custom (8cpu/16Ram)
Azure: mostly D3V2 (4cpu/14Ram)

They all have local SSD as well, in varying amounts.

I still have it on my todo list to test various things (cockroach speed, disk/network speed) on various machines types on all clouds. I'm hoping to have those in this quarter.

jseldess:
Excellent, @mberhault. These initial details are great. Let me know once you get to that testing.

bdarnell:
On Azure, the A series VMs eat data and must be avoided (cockroachdb/cockroach#12875). The D and Dv2 series seem to be fine (I think those are the only ones we have enough experience to say something about).

jseldess:
MemSQL's installation requirements: http://docs.memsql.com/docs/system-requirements

jseldess:
From @spencerkimball in https://github.com/cockroachdb/docs/issues/965: Please include minimum recommended configuration as well.

jseldess:
@mberhault, it'd be great to have minimum and recommendation hardware configurations in place for 1.0, which leaves us ~3 weeks. Have you had more time to dig into this, and do you have updates?

I'm wondering if we could provide details similar to what [memsql](http://docs.memsql.com/docs/system-requirements) has...

sploiselle:
From @mberhault 

Minimum hardware requirements:
- Not less than 2 cores; adding cores will help throughput.
- Not less than 1 Gig of RAM; impact on caching at RocksDB level
- If performance is a concern, use SSD; but all SSDs not equal across clouds.
- Triplicate data by default: Storage = ⅓ raw disk space

Cluster size: 
- Thoroughly tested up to 32 nodes, with experiments up to 64 with no issues.
- Won’t have network blowup up to 100 nodes.

More nodes vs bigger nodes: 
- More nodes: If node dies, amount of data to recover is smaller; more aggregate bandwidth across cluster.
- Bigger nodes: Might have better power; on cloud, more likely to get your own node.

Cloud-specific:
 - **Azure**
[...]

PR #1527, opened 2017-06-07
Basic hardware recs

jseldess:
Starts to address #1132, though not yet per-cloud.

### DIFF
--- a/recommended-production-settings.md
+++ b/recommended-production-settings.md
@@ -5,16 +5,30 @@ toc: false
 toc_not_nested: true
 ---
 
-This page provides recommended settings for production deployments. 
+This page provides recommended settings for production deployments.
 
 <div id="toc"></div>
+
+## Hardware
+
+Use at least 3 nodes for a replicated cluster, with each node having at least 2 GB of RAM.
+
+For best performance:
+
+- Use SSDs over HDDs.
+- Use larger/more powerful nodes. Adding more CPU is usually more beneficial than adding more RAM.
+
+For best resiliency:
+
+- Use many smaller nodes instead of fewer larger ones. Recovery from a failed node is faster when data is spread across more nodes.
+- Use [zone configs](configure-replication-zones.html) to increase the replication factor from 3 (the default) to 5. You can do this for the [entire cluster](configure-replication-zones.html#edit-the-default-replication-zone) or for specific [databases](configure-replication-zones.html#create-a-replication-zone-for-a-database) or [tables](configure-replication-zones.html#create-a-replication-zone-for-a-table).
 
 ## Cluster Topology
 
 When running a cluster with more than one node, each replica will be on a different node and a majority of replicas must remain available for the cluster to make progress. Therefore:
 
-- Run at least three nodes to ensure that a majority of replicas (2/3) remains available if a node fails. 
- 
+- Use at least three nodes to ensure that a majority of replicas (2/3) remains available if a node fails.
+
 - Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
 
 - Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
@@ -43,8 +57,8 @@ CockroachDB can use a large number of open file descriptors, often more than is 
 
 For each CockroachDB node:
 
-- At a **minimum**, the file descriptors limit must be 1956 (1700 per store plus 256 for networking). If the limit is below this threshold, the node will not start. 
-- It is **recommended** to set the file descriptors limit to unlimited; otherwise, the recommended limit is at least 15000 (10000 per store plus 5000 for networking). This higher limit ensures performance and accommodates cluster growth. 
+- At a **minimum**, the file descriptors limit must be 1956 (1700 per store plus 256 for networking). If the limit is below this threshold, the node will not start.
+- It is **recommended** to set the file descriptors limit to unlimited; otherwise, the recommended limit is at least 15000 (10000 per store plus 5000 for networking). This higher limit ensures performance and accommodates cluster growth.
 - When the file descriptors limit is not high enough to allocate the recommended amounts, CockroachDB allocates 10000 per store and the rest for networking; if this would result in networking getting less than 256, CockroachDB instead allocates 256 for networking and evenly splits the rest across stores.
 
 ### Increase the File Descriptors Limit
@@ -130,13 +144,13 @@ $(document).ready(function(){
 
 To adjust the file descriptors limit for a single process in Mac OS X Yosemite and later, you must create a property list configuration file with the hard limit set to the recommendation mentioned [above](#file-descriptors-limit). Note that CockroachDB always uses the hard limit, so it's not technically necessary to adjust the soft limit, although we do so in the steps below.
 
-For example, for a node with 3 stores, we would set the hard limit to at least 35000 (10000 per store and 5000 for networking) as follows: 
+For example, for a node with 3 stores, we would set the hard limit to at least 35000 (10000 per store and 5000 for networking) as follows:
 
 1.  Check the current limits:
 
     ~~~ shell
     $ launchctl limit maxfiles
-    maxfiles    10240          10240      
+    maxfiles    10240          10240
     ~~~
 
     The last two columns are the soft and hard limits, respectively. If `unlimited` is listed as the hard limit, note that the hidden default limit for a single process is actually 10240.
@@ -174,7 +188,7 @@ For example, for a node with 3 stores, we would set the hard limit to at least 3
 
     ~~~ shell
     $ launchctl limit maxfiles
-    maxfiles    35000          35000      
+    maxfiles    35000          35000
     ~~~
 
 #### Older versions
@@ -187,7 +201,7 @@ For example, for a node with 3 stores, we would set the hard limit to at least 3
 
     ~~~ shell
     $ launchctl limit maxfiles
-    maxfiles    10240          10240      
+    maxfiles    10240          10240
     ~~~
 
     The last two columns are the soft and hard limits, respectively. If `unlimited` is listed as the hard limit, note that the hidden default limit for a single process is actually 10240.
@@ -198,13 +212,13 @@ For example, for a node with 3 stores, we would set the hard limit to at least 3
     limit maxfiles 35000 35000
     ~~~
 
-3.  Save the file, and restart the system for the new limits to take effect. 
+3.  Save the file, and restart the system for the new limits to take effect.
 
 4.  Verify the new limits:
 
     ~~~ shell
     $ launchctl limit maxfiles
-    maxfiles    35000          35000      
+    maxfiles    35000          35000
     ~~~
 
 </div>
@@ -247,7 +261,7 @@ For example, for a node with 3 stores, we would set the hard limit to at least 3
 
 #### With Systemd
 
-Alternately, if you're using [Systemd](https://en.wikipedia.org/wiki/Systemd): 
+Alternately, if you're using [Systemd](https://en.wikipedia.org/wiki/Systemd):
 
 1.  Edit the service definition to configure the maximum number of open files:
 


jseldess:
@bdarnell, @mberhault, is it worth publishing these basic hardware recommendations in advance of having cloud-specific guidance? If so, anything need to be tweaked or added? 

Question: What exactly should I say about disk space? Cassandra has some nice **[docs](https://docs.datastax.com/en/landing_page/doc/landing_page/planning/planningHardware.html#planningHardware__disk) as well as a procedure for [estimating usable disk space](https://docs.datastax.com/en/landing_page/doc/landing_page/planning/planPlanningDiskCapacity.html).

sploiselle:
@jseldess Don't know if it's worth trying to fix, but the failing Azure page on TC loads much more quickly with a trailing slash (i.e., `https://azure.microsoft.com/en-us/updates/create-empty-resource-groups/`)

jseldess:
@sploiselle, hmm, we're already using a trailing slash...

sploiselle:
@jseldess Ah. The TC error was showing a URL without the slash; I didn't check the source. My mistake!

bdarnell:
Yes, we should go ahead and publish this information even though we don't yet have it translated into the cloud provider's machine classes. 

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[recommended-production-settings.md, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CVK6EDI8R9lYMsCm:-Km2CVK6EDI8R9lYMsCn:b32pfvi) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L14)):*
> ```Markdown
> ## Hardware
> 
> Each node of a CockroachDB cluster requires at a mimimum:
> ```

This would be a good place to repeat the requirement that you need a minimum of three nodes for a replicated cluster.

---

*[recommended-production-settings.md, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2BgfZvCZmC-sX5k4G:-Km2Bgf_lOI8VszPI6oQ:b-pjkpgo) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L16)):*
> ```Markdown
> Each node of a CockroachDB cluster requires at a mimimum:
> 
> - 2 CPU cores
> ```

@mberhault Why make this a hard requirement? More is better, of course, but everything should work with a single core.  (i'd rather run with one core on an SSD than two cores on HDD, for example)

---

*[recommended-production-settings.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2DReKvRiAwgnFU1R9:-Km2DReKvRiAwgnFU1RA:bd6pq9f) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L21)):*
> ```Markdown
> If performance is a priority, also consider the following:
> 
> - Use SSDs over HHDs.
> ```

s/HHD/HDD/

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km2CLqdYQaGHNWFMlPL:b-tgd8dq) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*
> ```Markdown
> 
> - Use SSDs over HHDs.
> - Since CockroachDB balances data approximately evenly across nodes, use more nodes to increase the work that can be handled in parallel and to reduce the amount of data to recover in case of node failure.
> ```

It's not very clear to me what this line is trying to say. 

---

*[recommended-production-settings.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2D5SQIzIQtAzaOp--:-Km2D5SQIzIQtAzaOp-0:b-xox1s6) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L23)):*
> ```Markdown
> - Use SSDs over HHDs.
> - Since CockroachDB balances data approximately evenly across nodes, use more nodes to increase the work that can be handled in parallel and to reduce the amount of data to recover in case of node failure.
> - Use larger nodes to increase processing power and, on cloud platforms, to increase the chance of getting a dedicated machine.
> ```

We should also say that in most cases, adding more CPUs will be more beneficial than adding more RAM. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-:-Km2Dck6WVNdJalCnUAf:biys6jy)*

jseldess:
*[recommended-production-settings.md, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2BgfZvCZmC-sX5k4G:-Km2EuEBjYMVlR4k2CUH:b-dvpori) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L16)):*

@mberhault Why make this a hard requirement? More is better, of course, but everything should work with a single core.  (i'd rather run with one core on an SSD than two cores on HDD, for example)
</blockquote></details>

Oh, @mberhault's rec was 1 core. I saw a later comment from you in the issue and thought you were suggesting that 2 should be our minimum. I must've misunderstood.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527)*

jseldess:
*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km2F6DydngcSiuVJh78:bwcaqv1) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*

It's not very clear to me what this line is trying to say. 
</blockquote></details>

Hmm, the main point not being made clearly is that recovery from node failure will be quicker with more nodes? How would you explain that? 

I guess the other point I was trying to make - the the processing of queries is distributed just as the data is - isn't really relevant here...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527)*

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km2fSIsGhByLSNGTVun:b-gxfwts) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*

Hmm, the main point not being made clearly is that recovery from node failure will be quicker with more nodes? How would you explain that? 

I guess the other point I was trying to make - the the processing of queries is distributed just as the data is - isn't really relevant here...
</blockquote></details>

More nodes will improve recovery speed, but will generally reduce performance (it's true that some queries will be able to take advantage of more nodes, but *all* queries will be able to take advantage of faster nodes). This is why I was confused to see this in the "performance" section. 

How about making that tradeoff explicit with something like this:

For best performance:
- Use SSDs over HDDs
- Use larger/more powerful nodes. Adding more CPU is usually more beneficial than adding more RAM

For best resiliency:
- Use many smaller nodes instead of fewer larger ones. Recovery from a failed node is faster when data is spread across more nodes.
- Use zone configs to set a replication factor of 5 (instead of the default 3)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527)*

jseldess:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[recommended-production-settings.md, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CVK6EDI8R9lYMsCm:-Km2lq8sidHMZJdC7Ea8:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L14)):*

This would be a good place to repeat the requirement that you need a minimum of three nodes for a replicated cluster.
</blockquote></details>

Done.

---

*[recommended-production-settings.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2DReKvRiAwgnFU1R9:-Km2eePrt-8HKlh-mkgf:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L21)):*

s/HHD/HDD/
</blockquote></details>

Done.

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km2lroQkonjyt7whEV0:b-fld0l8) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*

More nodes will improve recovery speed, but will generally reduce performance (it's true that some queries will be able to take advantage of more nodes, but *all* queries will be able to take advantage of faster nodes). This is why I was confused to see this in the "performance" section. 

How about making that tradeoff explicit with something like this:

For best performance:
- Use SSDs over HDDs
- Use larger/more powerful nodes. Adding more CPU is usually more beneficial than adding more RAM

For best resiliency:
- Use many smaller nodes instead of fewer larger ones. Recovery from a failed node is faster when data is spread across more nodes.
- Use zone configs to set a replication factor of 5 (instead of the default 3)
</blockquote></details>

Done. 

For the zone config bullet, I've just linked to the start of the zone config page. Is the more specific recommendation here to change the default zone config to replicate 5 ways? If so, I could link directly to this example instead: https://www.cockroachlabs.com/docs/configure-replication-zones.html#edit-the-default-replication-zone

---

*[recommended-production-settings.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2D5SQIzIQtAzaOp--:-Km2lsXCfAFQRAbd3_Ni:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L23)):*

We should also say that in most cases, adding more CPUs will be more beneficial than adding more RAM. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527)*

bdarnell:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[recommended-production-settings.md, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2BgfZvCZmC-sX5k4G:-Km2yI4Xy7uO3TZZhAq9:b-mmpxd9) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L16)):*

Oh, @mberhault's rec was 1 core. I saw a later comment from you in the issue and thought you were suggesting that 2 should be our minimum. I must've misunderstood.
</blockquote></details>

I'm not sure if we need to list this if it's just 1 (unless it's to distinguish VMs that could otherwise get a fractional CPU)

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km2yG2jPp3oaFyvTOXA:b-mjh13h) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*

Done. 

For the zone config bullet, I've just linked to the start of the zone config page. Is the more specific recommendation here to change the default zone config to replicate 5 ways? If so, I could link directly to this example instead: https://www.cockroachlabs.com/docs/configure-replication-zones.html#edit-the-default-replication-zone
</blockquote></details>

Yeah, linking to that example would be good. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-:-Km2ykjbhLk9_DW5T3sC:bnfp4nl)*

jseldess:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[recommended-production-settings.md, line 16 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2BgfZvCZmC-sX5k4G:-Km3G588rI4yrI64lCsh:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L16)):*

I'm not sure if we need to list this if it's just 1 (unless it's to distinguish VMs that could otherwise get a fractional CPU)
</blockquote></details>

Done.

---

*[recommended-production-settings.md, line 22 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1527#-Km2CLqdYQaGHNWFMlPK:-Km3D6eNRG1q420H7fY-:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ad4eea541c719ba78bfe3fdf818ecb9284f03850/recommended-production-settings.md#L22)):*

Yeah, linking to that example would be good. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1527)*

jseldess:
@mberhault, any feedback before I merge these initial hardware recommendations?

mberhault:
LGTM, sorry about the delay.

PR #1540, opened 2017-06-08
Clean up DROP DATABASE documentation

### LINKED ISSUE #486: sql: Clean up DROP DATABASE
For the sake of consistency, add a **Parameters** section and convert **Usage** to **Examples**.

Amruta-Ranade:
Closes #486

### DIFF
--- a/drop-database.md
+++ b/drop-database.md
@@ -61,6 +61,9 @@ To avoid an error in case the database does not exist, you can include `IF EXIST
 +----------+
 ~~~
 
+{{site.data.alerts.callout_danger}}<code>DROP DATABASE</code> drops <em>all</em> tables within the database as well as objects dependent on the tables without listing the tables or the dependent objects. This can lead to inadvertent and difficult-to-recover losses. To avoid potential harm, we recommend dropping objects individually in most cases.{{site.data.alerts.end}}
+
+
 ## See Also
 
 - [`CREATE DATABASE`](create-database.html)
--- a/drop-database.md
+++ b/drop-database.md
@@ -4,7 +4,7 @@ summary: The DROP DATABASE statement removes a database and all its objects from
 toc: false
 ---
 
-The `DROP DATABASE` [statement](sql-statements.html) removes a database and all its objects from a CockroachDB cluster. 
+The `DROP DATABASE` [statement](sql-statements.html) removes a database and all its objects from a CockroachDB cluster.
 
 <div id="toc"></div>
 
--- a/drop-database.md
+++ b/drop-database.md
@@ -16,21 +16,15 @@ The `DROP DATABASE` [statement](sql-statements.html) removes a database and all 
 
 The user must have the `DROP` [privilege](privileges.html) on the database and on all tables in the database. 
 
-## Usage
+## Parameters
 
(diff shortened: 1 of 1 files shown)


jseldess:
@Amruta-Ranade, the restructuring looks good, but there's a little but of expanding to do as well, I just realized. As the intro says, `DROP DATABASE` drops a database *and all its objects*. That second part is missing from the diagram and from the documentation. Based on the [full grammar](https://www.cockroachlabs.com/docs/sql-grammar.html#drop_stmt), there should be an `opt_drop_behavior` at the end. When that's inlined, that turns into two options, `CASCADE` and `RESTRICT`. These options are available in the [`DROP TABLE`](https://www.cockroachlabs.com/docs/drop-table.html) docs as well, so it's probably good to go and read about them there first. 

Then you'll want to: 
1. Rest out how those options work in the SQL shell.
2. Rework the diagram to show those options.
3. Add them to the parameters list.
4. Cover them in the examples. 

I assume that deleting all dependent objects of a database means deleting all tables in the database, all indexes on those tables, and all views depending on any of those tables, but you'll need to test things out to confirm. Also, feel free to ask CockroachDB devs questions in Gitter. @dt might be a good first contact.

Amruta-Ranade:
@jseldess I checked the full grammar for drop statements, and it appears that opt_drop_behavior is only for dropping tables and indexes, not databases. I tested the drop database commands using the Cascade and Restrict options, but an error was displayed. The options worked perfectly for dropping tables though. So I think opt_drop_behavior does not apply to the drop database command. I will cross-check with the CockroachDB devs on Gitter.

jseldess:
I think you're right, @Amruta-Ranade. I just created a database, created a table inside it, and then created a view dependent on that table. Despite those objects, I was able to drop the datase without `CASCADE`:

```
root@:26257/> create table test.test (a int primary key);
CREATE TABLE
root@:26257/> create view test.test_view as select a from test.test;
CREATE VIEW
root@:26257/> drop database test;
DROP DATABASE
```

So I think you're right that those options aren't relevant. What this tells us, though, is that we need a clear warning, something like the one here https://www.cockroachlabs.com/docs/drop-table.html#remove-a-table-and-dependent-objects-with-cascade.

Amruta-Ranade:
Yes, I agree. Will work on that.

Amruta-Ranade:
@jseldess I checked the drop table warning again. The warning is for dropping dependent objects when using the cascade option. The dependent objects refer to other tables in the database who might be dependent on the table being dropped, hence the warning is required to make the user aware that dependent tables will be dropped as well. We are not recommending that the users drop the columns or other objects within the table being dropped. 

In case of the drop database, there seems to be no scenario where other databases are dependent on the dropped database. And I don't think we recommend the users drop the individual tables before dropping the database. Hence the warning might not be applicable for the drop database documentation. Please let me know what you think.

jseldess:
@Amruta-Ranade, sorry for the delay. You're right that the situation is a bit different than dropping a table. But I do think we should add a simple warning stating that dropping a database will drop all tables and views within the db. We don't want users doing that accidentally.

jseldess:
Thanks, @Amruta-Ranade. Looks great. I'll merge.

PR #1548, opened 2017-06-09
Update diagnostics reporting page

### LINKED ISSUE #1547: Document how to verify that opt out worked
Requested in https://github.com/cockroachdb/cockroach/issues/16389#issuecomment-307104726.

jseldess:
- Make two opt out options more explicit
- Explain how to verify opt out

Fixes #1547

### DIFF
--- a/diagnostics-reporting.md
+++ b/diagnostics-reporting.md
@@ -14,7 +14,7 @@ This page explains the details that get shared and how to opt out of sharing.
 
 ## What Gets Shared
 
-When diagnostics reporting is on, each node of a CockroachDB cluster shares anonymized storage details, SQL table structure details, and SQL query statistics with Cockroach Labs on an hourly basis. Please note that the details that get shared may change over time, but as that happens, we will update this page and announce the changes in release notes.
+When diagnostics reporting is on, each node of a CockroachDB cluster shares anonymized storage details, SQL table structure details, and SQL query statistics with Cockroach Labs on an hourly basis, as well as crash reports as they occur. Please note that the details that get shared may change over time, but as that happens, we will update this page and announce the changes in release notes.
 
 ### Storage Details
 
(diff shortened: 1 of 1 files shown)


jseldess:
@bdarnell, can you give this a quick look?

bdarnell:
:lgtm:

---

---

*[diagnostics-reporting.md, line 205 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1548#-KmRhs0EQM1rUGF2P_0y:-KmRhs0FcoWoGYT3d34q:b-igag2q) ([raw file](https://github.com/cockroachdb/docs/blob/2f407f87fd4f6269159bdb07684d892d14df4218/diagnostics-reporting.md#L205)):*
> ```Markdown
> ## Opt Out of Diagnostics Reporting
> 
> ### At Cluster Startup
> ```

Call this "Initialization" instead of "Startup" to make it more clear that this only takes effect the first time any node is started.

---

*[diagnostics-reporting.md, line 238 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1548#-KmRiC6Mbnsl90s_5fCW:-KmRiC6Mbnsl90s_5fCX:ba7guys) ([raw file](https://github.com/cockroachdb/docs/blob/2f407f87fd4f6269159bdb07684d892d14df4218/diagnostics-reporting.md#L238)):*
> ```Markdown
> If the setting is `false`, diagnostics reporting is off; if the setting is `true`, diagnostics reporting is on.
> 
> Also, if you opted out [at cluster startup](#at-cluster-startup), you can find the following message in the [logs](debug-and-error-logs.html) to confirm that diagnostics reporting is off:
> ```

This doesn't actually verify that reporting is off - it will show that the environment variable was recognized, but the env var will also show up here on subsequent restarts even if was missing the first time (which means it wasn't doing anything). We should probably leave this part out and just recommend `SHOW CLUSTER SETTING diagnostics.reporting.enabled`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1548#-:-KmRip3emci0FeGp03qW:bnfp4nl)*

jseldess:

---

*[diagnostics-reporting.md, line 205 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1548#-KmRhs0EQM1rUGF2P_0y:-KmRnxStBP_AfqSyd2x7:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2f407f87fd4f6269159bdb07684d892d14df4218/diagnostics-reporting.md#L205)):*

Call this "Initialization" instead of "Startup" to make it more clear that this only takes effect the first time any node is started.
</blockquote></details>

Done.

---

*[diagnostics-reporting.md, line 238 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1548#-KmRiC6Mbnsl90s_5fCW:-KmRoLdd5K3YoSPu5vsB:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2f407f87fd4f6269159bdb07684d892d14df4218/diagnostics-reporting.md#L238)):*

This doesn't actually verify that reporting is off - it will show that the environment variable was recognized, but the env var will also show up here on subsequent restarts even if was missing the first time (which means it wasn't doing anything). We should probably leave this part out and just recommend `SHOW CLUSTER SETTING diagnostics.reporting.enabled`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1548)*

PR #1549, opened 2017-06-09
Don't use unsupported non-const expr in AS OF SYSTEM TIME

tschottdorf:
https://github.com/cockroachdb/cockroach/issues/16424

### DIFF
--- a/backup.md
+++ b/backup.md
@@ -48,7 +48,7 @@ Full backups contain an unreplicated copy of your data and can always be used to
 
 #### Incremental Backups
 
-Incremental backups are smaller and faster to produce than full backups because they contain only the data that has changed since a base set of backups you specify (which must include one full backup, and can include many incremental backups). 
+Incremental backups are smaller and faster to produce than full backups because they contain only the data that has changed since a base set of backups you specify (which must include one full backup, and can include many incremental backups).
 
 You can only create incremental backups within the garbage collection period of the base backup's most recent timestamp. This is because incremental backups are created by finding which data has been created or modified since the most recent timestamp in the base backup––that timestamp data, though, is deleted by the garbage collection process.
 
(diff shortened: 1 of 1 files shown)


sploiselle:

---

*[backup.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1549#-KmXbjMz8tao-Lp5r9R0:-KmXbjMz8tao-Lp5r9R1:b-p4nog4) ([raw file](https://github.com/cockroachdb/docs/blob/d9e18644c1722d87ffce21896e1a528027f0f758/backup.md#L61)):*
Because this is no longer dynamically calculated, can I get you to switch this to:

> For best performance, we also recommend always starting backups with a specific [timestamp](timestamp.html) at least 10 seconds in the past. For example:

---

*[backup.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1549#-KmXcZYK7JiD2ETPCxYN:-KmXcZYK7JiD2ETPCxYO:bkjska0) ([raw file](https://github.com/cockroachdb/docs/blob/d9e18644c1722d87ffce21896e1a528027f0f758/backup.md#L152)):*
> ```Markdown
> ~~~
> 
> ### Create Backups as of a Specific System Time
> ```

Can you delete this section since all examples are now as of a specific time?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1549)*

jseldess:
:lgtm:, but @sp

---

---

*[backup.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1549#-KmXbjMz8tao-Lp5r9R0:-Kn01zwGUb4KIRJbuX-t:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d9e18644c1722d87ffce21896e1a528027f0f758/backup.md#L61)):*

Because this is no longer dynamically calculated, can I get you to switch this to:

> For best performance, we also recommend always starting backups with a specific [timestamp](timestamp.html) at least 10 seconds in the past. For example:
</blockquote></details>

Done.

---

*[backup.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1549#-KmXcZYK7JiD2ETPCxYN:-Kn025bk0eLzGJhGEzuW:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d9e18644c1722d87ffce21896e1a528027f0f758/backup.md#L152)):*

Can you delete this section since all examples are now as of a specific time?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1549#-:-KmGX7pkaUtG38S3fAZu:b-iy93c7)*

PR #1107, opened 2017-02-20
Refactor the SHOW/SET documentation.

knz:
Both SHOW and SET are really single statements which act on the
session (and they are also handled using a single code path in
CockroachDB!), with the exception of SET TRANSACTION which acts on the
current transaction. This patch acknowledges the commonality and
merges the various documentation pages into two more general pages.

Supports the changes in https://github.com/cockroachdb/cockroach/pull/13678.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1107)

PR #1588, opened 2017-06-13
Refactor the SHOW/SET documentation.

benesch:
Rebase of #1107. Feel free to close if you rebase #1107 on master, @knz!

----

Both SHOW and SET are really single statements which act on the
session (and they are also handled using a single code path in
CockroachDB!), with the exception of SET TRANSACTION which acts on the
current transaction. This patch acknowledges the commonality and
merges the various documentation pages into two more general pages.


benesch:
This failed, but it failed with a legitimate HTML proofer violation.

knz:
Thanks for this. I'll rebase #1107.

PR #1590, opened 2017-06-13
generate raw HTML tables on "Functions and Operators" page

### LINKED ISSUE #1536: A few rows on Functions and Operators are not rendering
screen shot 2017-06-08 at 7 58 35 am

benesch:
Due to vmg/redcarpet#477, it's not possible to use pipe characters (`|`)
in tables. This currently breaks several table rows on our "Functions
and Operators" page. Since these tables are auto-generated, fix this by
just directly outputting HTML tables, where pipe characters have no
special meaning.

Fixes #1536.

### DIFF
--- a/_includes/sql/operators.md
+++ b/_includes/sql/operators.md
@@ -1,230 +1,252 @@
-`#` | Return
---- | ---
-<a href="int.html">int</a> `#` <a href="int.html">int</a> | <a href="int.html">int</a>
-
-`%` | Return
---- | ---
-<a href="decimal.html">decimal</a> `%` <a href="decimal.html">decimal</a> | <a href="decimal.html">decimal</a>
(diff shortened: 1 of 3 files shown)


dt:
Hm. I know I made this change once while working on this page with @sean and then we decided not to make the switch it, but I don't remember why and thus if my reasoning then still applies. Any recollections of this @sploiselle?

sploiselle:
I don't think anything was broken at the time, so there was no impetus to convert from Kramdown to HTML––IIRC (heavy on the _if_).

Also, LGTM!

benesch:
TFTRs!

benesch:
Whoops, I took your comment as an LGTM, @dt. LMK if you have any post-review feedback.

dt:
Yep, it looks good to me!

PR #1592, opened 2017-06-14
Update docs on adding the Foreign Key constraint

### LINKED ISSUE #1519: sql: don't try to add index in ALTER...ADD FK
PRs: 
- https://github.com/cockroachdb/cockroach/pull/16333
- cockroachdb/cockroach#16175

jseldess:
- Document need to add index before adding FK constraint to an existing column,
- Update `ADD COLUMN` page:
  -  State that it's not possible to add a column with a FK constraint.
  - Provide workaround: add column without the constraint, index the column, and then add the constraint.

Fixes #1519

### DIFF
--- a/add-constraint.md
+++ b/add-constraint.md
@@ -21,7 +21,7 @@ The <a href="primary-key.html">Primary Key</a> and <a href="not-null.html">Not N
 
 ## Required Privileges
 
-The user must have the `CREATE` [privilege](privileges.html) on the table. 
+The user must have the `CREATE` [privilege](privileges.html) on the table.
 
 ## Parameters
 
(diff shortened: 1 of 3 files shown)


add-constraint.md
```diff
@@ -51,11 +51,66 @@ Adding the [Check constraint](check.html) requires that all of a column's values
 
 ### Add the Foreign Key Constraint
 
-Adding the [Foreign Key constraint](foreign-key.html) requires that all of a column's values are equal to an existing value in the column it references.
-
-``` sql
-> ALTER TABLE orders ADD CONSTRAINT customer_fk FOREIGN KEY (customer) REFERENCEs customers (id);
-```
+To add the [Foreign Key](foreign-key.html) constraint to an existing column or group of columns, you must first use [`CREATE INDEX`](create-index.html) to index the column or columns that you plan to constrain. Then you can use the `ADD CONSTRAINT` statement to add the Foreign Key constraint to the column or columns.
```
dt: to clarify, only need to create index if there isn't already an index that contains the column(s).

add-constraint.md
```diff
@@ -51,11 +51,66 @@ Adding the [Check constraint](check.html) requires that all of a column's values
...
+|        |     status STRING NOT NULL,␤                                                                                |
+|        |     CONSTRAINT "primary" PRIMARY KEY (id ASC),␤                                                             |
+|        |     FAMILY "primary" (id, customer_id, status),␤                                                            |
+|        |     CONSTRAINT check_status CHECK (status IN ('open':::STRING, 'complete':::STRING, 'cancelled':::STRING))␤ |
+|        | )                                                                                                           |
++--------+-------------------------------------------------------------------------------------------------------------+
+(1 row)
+~~~
+
+To ensure that each value in the `orders.customer_id` column matches a unique value in the `orders.id` column, you want to add the Foreign Key constraint to `orders.customer_id`. So you first create an index on `orders.customer_id`:
```
dt: We could explain somewhere that deleting a customer would require finding any orders for that customer -- if you _already had an index on `customer_id`_ -- e.g. because your app sometimes needs to find all the orders for the customer's account page-- then that index could also be used by the FK.

add-constraint.md
jseldess: Done. Also updated the foreign key constraint page.

add-constraint.md
jseldess: Hmm, can't find a smooth way to work that into the example. Can you help? Otherwise, is it ok to leave this bit out?

dt: yeah, probably doesn't need to be in the example, and might fit better on the FK page.

might be something like "an index on the columns in a foreign key is required because updates and deletes on the referenced table will need to search the referencing table for any matching records, to ensure those operations would not violate existing references. In practice, such index is likely also needed by the application using foreign keys, since finding all records which belong to some entity, for example all `orders` for a given `customer`, is very common."

jseldess: Done. Added to the foreign key constraint page.

PR #1594, opened 2017-06-14
Fix formatting on debug and error logs page

jseldess:


jseldess:
Thanks, @sploiselle. Yeah the only problem was indentation.

PR #1595, opened 2017-06-14
Revoke statement document: Added Parameters section, converted Usage to Examples, updated SQL diagram

### LINKED ISSUE #492: sql: Clean up REVOKE
For the sake of consistency, add a **Parameters** section and convert **Usage** to **Examples**.

Amruta-Ranade:
Closes #492

### DIFF
--- a/generate/main.go
+++ b/generate/main.go
@@ -430,7 +430,7 @@ func main() {
 						"table_pattern_list": "table_name ( ',' table_name )*",
 						"name_list": "database_name ( ',' database_name )*",
 						"'FROM' name ( ',' name )*": "'FROM' user_name ( ',' user_name )*"},
-					unlink: []string{"table_name","database_name"}, 
+					unlink: []string{"table_name","database_name","user_name"}, 
 				},
 				{
 					name:    "rollback_transaction",
--- a/generate/main.go
+++ b/generate/main.go
@@ -423,7 +423,15 @@ func main() {
 					},
 					unlink: []string{"destination", "timestamp", "full_backup_location", "incremental_backup_location"},
 				},
-				{name: "revoke_stmt", inline: []string{"privileges", "privilege_list", "privilege", "targets", "grantee_list"}},
+				{
+					name: "revoke_stmt", 
+					inline: []string{"privileges", "privilege_list", "privilege", "targets", "grantee_list"},
+					replace: map[string]string{
+						"table_pattern_list": "table_name ( ',' table_name )*",
+						"name_list": "database_name ( ',' database_name )*",
+						"'FROM' name ( ',' name )*": "'FROM' user_name ( ',' user_name )*"},
+					unlink: []string{"table_name","database_name"}, 
+				},
 				{
 					name:    "rollback_transaction",
 					stmt:    "transaction_stmt",
--- a/_includes/sql/diagrams/revoke.html
+++ b/_includes/sql/diagrams/revoke.html
@@ -60,11 +60,11 @@
          <rect x="497" y="509" width="58" height="32" rx="10"></rect>
          <rect x="495" y="507" width="58" height="32" class="terminal" rx="10"></rect>
          <text class="terminal" x="505" y="527">FROM</text>
-         <a xlink:href="sql-grammar.html#user_name" xlink:title="user_name">
+         
             <rect x="595" y="509" width="86" height="32"></rect>
             <rect x="593" y="507" width="86" height="32" class="nonterminal"></rect>
             <text class="nonterminal" x="603" y="527">user_name</text>
-         </a>
+         
          <rect x="595" y="465" width="24" height="32" rx="10"></rect>
          <rect x="593" y="463" width="24" height="32" class="terminal" rx="10"></rect>
          <text class="terminal" x="603" y="483">,</text>
--- a/_includes/sql/diagrams/revoke.html
+++ b/_includes/sql/diagrams/revoke.html
@@ -1,68 +1,74 @@
-<div><svg width="714" height="498">
+<div><svg width="728" height="542">
          
-         <polygon points="9 17 1 13 1 21"></polygon>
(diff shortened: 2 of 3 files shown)


jseldess:
@Amruta-Ranade, sorry for the delay. I'm reviewing now.

generate/main.go
```diff
@@ -423,7 +423,15 @@ func main() {
...
 				},
-				{name: "revoke_stmt", inline: []string{"privileges", "privilege_list", "privilege", "targets", "grantee_list"}},
+				{
+					name: "revoke_stmt", 
+					inline: []string{"privileges", "privilege_list", "privilege", "targets", "grantee_list"},
+					replace: map[string]string{
+						"table_pattern_list": "table_name ( ',' table_name )*",
+						"name_list": "database_name ( ',' database_name )*",
+						"'FROM' name ( ',' name )*": "'FROM' user_name ( ',' user_name )*"},
+					unlink: []string{"table_name","database_name"},
```
jseldess: We should also unlink `"user_name"`.

Amruta-Ranade:
@jseldess No problem! I fixed the issue. Please let me know if any changes are needed. Thanks!

sploiselle:
:lgtm: mod one pedantic nit

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[revoke.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1595#-Kn0fqyYjPdG6Ro9m1rY:-Kn0fqyYjPdG6Ro9m1rZ:bngirnh) ([raw file](https://github.com/cockroachdb/docs/blob/e2b020d86196b14d08f3c8ec9079c11ff186e7a5/revoke.md#L25)):*
> ```Markdown
> Parameter | Description
> ----------|------------
> `table_name` | A comma-separated list of table names. Alternatively, to revoke privileges for all tables, use `*`. 
> ```

According to the diagram, `table_name` isn't a list (that's what the comma track implies). I'd make this more direct and simply say something like "The name of the table..."

---

*[revoke.md, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1595#-Kn0g8XePfa-Q5fOLsPW:-Kn0g8XePfa-Q5fOLsPX:b33azd1) ([raw file](https://github.com/cockroachdb/docs/blob/e2b020d86196b14d08f3c8ec9079c11ff186e7a5/revoke.md#L26)):*
> ```Markdown
> ----------|------------
> `table_name` | A comma-separated list of table names. Alternatively, to revoke privileges for all tables, use `*`. 
> `database_name` | A comma-separated list of database names.<br><br>Privileges revoked for databases will be revoked for any new tables created in the databases.
> ```

Same as `table_name` feedback above

---

*[revoke.md, line 27 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1595#-Kn0gCmsPd0gzQu47Oth:-Kn0gCmsPd0gzQu47Oti:b33azd1) ([raw file](https://github.com/cockroachdb/docs/blob/e2b020d86196b14d08f3c8ec9079c11ff186e7a5/revoke.md#L27)):*
> ```Markdown
> `table_name` | A comma-separated list of table names. Alternatively, to revoke privileges for all tables, use `*`. 
> `database_name` | A comma-separated list of database names.<br><br>Privileges revoked for databases will be revoked for any new tables created in the databases.
> `user_name` | A comma-separated list of [users](create-and-manage-users.html).
> ```

Same as `table_name` feedback above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1595#-:-Kn0gLyZizBH11YVZk6x:b-mfisa6)*

Amruta-Ranade:
@sploiselle Doesn't the comma-track imply a list? I checked documents for other statements (for example, the grant statement), and the comma-track seems to imply a list. Am I misunderstanding something?

sploiselle:
@Amruta-Ranade This is probably confusing if we’re inconsistent in the documentation!

The idea behind the railroad diagrams is to show the “path” through the SQL syntax. If you follow that path and hit `table_name`, `database_name`, or `user_name`, what the SQL diagram expects is the name of an individual element (or `*` for all elements). To list multiple items, you follow the track through the optional comma track above it and re-enter an individual `table_name`, `database_name`, or `user_name`.

If you were to treat those elements as “comma-separated lists” themselves, under which circumstance would you ever divert from that element and proceed to the track with the optional comma? You never would, which means those values themselves represent individual names.

If you can point out where else you’ve seen this pattern in the docs, I’m glad to fix them so we’re consistent/accurate.

Amruta-Ranade:
@sploiselle Hmm..it is indeed confusing :) Thank you for the clarification! I see the comma-track pattern in several documents: Drop Table, Drop View, Create Table, Show Grants. But I also see the pattern as you explain it in Select. And then a third pattern where we do mention in the Parameters section that the elements can be a comma-separated list, but not in the SQL diagram (example: Insert, Upsert). 

Please advise which pattern should I follow for the revoke statement. I think the second pattern is most suitable.

jseldess:
@Amruta-Ranade, I think I'm guilty of the language you were emulating. Please follow @sploiselle's guidance, and we'll update the other docs in a separate PR.

Amruta-Ranade:
@jseldess @sploiselle Actually, I can see the logic behind all three variations.  I agree the parameter definition can start with "The name of..." but I think we also need to state that optionally, a comma-separated list can be used. I hadn't understood the comma-track usage just from the SQL diagram till I read the parameter description, so a new user might be confused as well. Permission to use a fourth variation of the description please :)

jseldess:
I think that sounds like a good, user-friendly idea, @Amruta-Ranade. @sploiselle, you good with that?

sploiselle:
Sounds good.

jseldess:
Looks good, @Amruta-Ranade. Thank you!

PR #1596, opened 2017-06-15
Clarify that hardware decisions depend on scenario and workload

jseldess:
@robert-s-lee, based on some feedback in the docs meeting, I've updated our hardware recs to clarify that decisions depend on scenario and workload. PTAL.


jseldess:
Thanks, @robert-s-lee. Updated this section. Please take another look. @bdarnell, would like your review, too.

PR #1597, opened 2017-06-15
clarify timestamp time zone details

### LINKED ISSUE #685: sql: TIMESTAMPTZ is incorrect
The docs for TIMESTAMPTZ say: "TIMESTAMPTZ stores a date and time pair with a time zone offset from UTC" however this isn't correct as both TIMESTAMP and TIMESTAMPTZ are stored in the same way: number of seconds and nanoseconds since the epoch. No timezone information is stored at all in either case.

In general the TZ type cares about the session offset and normal TIMESTAMP doesn't, but that's not always true. Ping me when you'd like to go over specifics of how these two types differ, as there are many edge cases here and pretty much only David knows how it all works.

sploiselle:
closes #685

Chose not to link to docs for setting session time zone because it is apparently not a best practice.

### DIFF
--- a/v1.0/timestamp.md
+++ b/v1.0/timestamp.md
@@ -4,13 +4,34 @@ summary: The TIMESTAMP data type stores a date and time pair in UTC, whereas TIM
 toc: false
 ---
 
-The `TIMESTAMP` [data type](data-types.html) stores a date and time pair in UTC, whereas `TIMESTAMPTZ` stores a date and time pair with a time zone offset from UTC. 
+The `TIMESTAMP` [data type](data-types.html) stores a date and time pair in UTC.
 
 <div id="toc"></div>
 
+## Time Zone Details
+
+`TIMESTAMP` has two variants:
+
+- `TIMESTAMP WITH TIME ZONE` converts `TIMESTAMP` values from UTC to the client's session time zone (unless another time zone is specified for the value). However, it is conceptually important to note that `TIMESTAMP WITH TIME ZONE` *does not* store any time zone data.
+
+   {{site.data.alerts.callout_info}}The default session time zone is UTC, which means that by default `TIMESTAMP WITH TIME ZONE` values display in UTC.{{site.data.alerts.end}}
+
+- `TIMESTAMP WITHOUT TIME ZONE` presents all `TIMESTAMP` values in UTC.
+
+The difference between these two types is that `TIMESTAMP WITH TIME ZONE` uses the client's session time zone, while the other simply does not. This behavior extends to functions like `now()` and `extract()` on `TIMESTAMP WITH TIME ZONE` values.
+
+### Best Practices
+
+We recommend always using the `...WITH TIME ZONE` variant because the `...WITHOUT TIME ZONE` variant can sometimes lead to unexpected behaviors when it ignores a session offset. 
+
+However, we also recommend you avoid setting a session time for your database. Instead, convert values from the default time zone (UTC) on the client side.
+
 ## Aliases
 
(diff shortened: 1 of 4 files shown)


timestamp.md
```diff
@@ -4,13 +4,27 @@ summary: The TIMESTAMP data type stores a date and time pair in UTC, whereas TIM
...
+## Time Zone Details
+
+`TIMESTAMP` has two variants:
+
+- `TIMESTAMP WITH TIME ZONE` converts `TIMESTAMP` values from UTC to the client's session time zone (unless another time zone is specified for the value). However, it is conceptually important to note that `TIMESTAMP WITH TIME ZONE` *does not* store any time zone data.
+- `TIMESTAMP WITHOUT TIME ZONE` presents all `TIMESTAMP` values in UTC.
+
+The difference between these two types is that `TIMESTAMP WITH TIME ZONE` uses the client's session time zone, while the other simply does not. This behavior extends to functions like `now()` and `extract()` on `TIMESTAMP WITH TIME ZONE` values.
+
+We recommend always using the `...WITH TIME ZONE` variant because the `...WITHOUT TIME ZONE` variant can sometimes lead to unexpected behaviors when it ignores a session offset. However, we also recommend you avoid setting a session time for your database.
```
mjibson: Can we note somewhere that the default session time zone is UTC?

jseldess:
@sploiselle, just bubbling this back up. Looks like there's just one open request from Matt.

sploiselle:
@mjibson I added some language around this to the `SET TIME ZONE` docs. Can I get you to take a look at those and see what you think?

v1.0/timestamp.md
```diff
@@ -4,13 +4,30 @@ summary: The TIMESTAMP data type stores a date and time pair in UTC, whereas TIM
...
 
+## Time Zone Details
+
+`TIMESTAMP` has two variants:
+
+- `TIMESTAMP WITH TIME ZONE` converts `TIMESTAMP` values from UTC to the client's session time zone (unless another time zone is specified for the value). However, it is conceptually important to note that `TIMESTAMP WITH TIME ZONE` *does not* store any time zone data.
+
+- `TIMESTAMP WITHOUT TIME ZONE` presents all `TIMESTAMP` values in UTC.
+
+The difference between these two types is that `TIMESTAMP WITH TIME ZONE` uses the client's session time zone, while the other simply does not. This behavior extends to functions like `now()` and `extract()` on `TIMESTAMP WITH TIME ZONE` values.
```
mjibson: If we aren't going to note here that the session time zone defaults to UTC, can we at least provide a link to the SET TIME ZONE section? Users need a fast way to discover the default.

sploiselle:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[v1.0/timestamp.md, line 19 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1597#-Koi0iT8-pVG5gYcBfgW-r2-19:-KoiFl-fKYGOvHCYoDV5:b-xghq49) ([raw file](https://github.com/cockroachdb/docs/blob/af78f2e3e608d33b8b2c181092bd6e71eb3e45d0/v1.0/timestamp.md#L19)):*
<details><summary><i>Previously, mjibson (Matt Jibson) wrote…</i></summary><blockquote>

If we aren't going to note here that the session time zone defaults to UTC, can we at least provide a link to the SET TIME ZONE section? Users need a fast way to discover the default.
</blockquote></details>

Misunderstood what you were suggesting! Tried to make that more clear in this doc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1597)*

PR #1601, opened 2017-06-16
untangle and speed up sidebar rendering

benesch:
Rendering the sidebar with YAML and Liquid is expensive. Liquid loops,
in particular, are quite slow, and the sidebar loops must be rerun for
every page to inject the "active" class where appropriate.

Since our sidebar already doesn't function without JavaScript, we can
avoid all this overhead by instead rendering the sidebar with JavaScript
on users' browsers. This gets our total build time down to under 3s. The
only downside is that the sidebar data is now JSON, which is a bit more
annoying to edit by hand.

---

JSON is a little less readable than YAML, but makes my life versioning the sidebar easier and speeds up the builds significantly to boot.


benesch:
Heh, whoops. Looks like that `env.BUILD_VCS_NUMBER` issue has actually been breaking the topnav links in TC test builds for quite a while. Fixed!

jseldess:
Thanks for the fix, @benesch. I think this is good to merge now. @sploiselle, mind taking a final look?

benesch:
TFTRs!

PR #1603, opened 2017-06-17
Propagate operation results through node.js txnWrapper example

twrobel3:
Modifies the callback behavior of the node.js transaction wrapper
code to propogate up all arguments to the callback in the user
defined operation.

Current implementation would only propogate the errors up, and not
any data that the user wishes to have returned.

### DIFF
--- a/_includes/app/txn-sample.js
+++ b/_includes/app/txn-sample.js
@@ -37,6 +37,7 @@ function txnWrapper(client, op, next) {
         if (err) {
           return handleError(err);
         }
+        var opResults = arguments;
 
         // If we reach this point, release and commit.
         client.query('RELEASE SAVEPOINT cockroach_restart', function (err) {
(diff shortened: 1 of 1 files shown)


twrobel3:
Brief explanation:  In the example code, the `results` variable in the transaction wrapper callback on [line 112](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.js#L112) is always (barring errors) the result of the `COMMIT` command on [line 60](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.js#L60).  Having the commit result be returned is a bit redundant, because the same information is encapsulated by the presence of the error.

This modification allows for the `op` provided by the user to have an arbitrary number of callback, and for those values to be propagated through the wrapper and back to the outer callback.  In this manner, an operation can include `SELECTS` or other operations returning values from the DB, and have those values be accessible from outside the wrapper.

twrobel3:
@jseldess The CONTRIBUTING doc says to assign you, but I don't seem to have privileges to assign, so hopefully tagging will suffice :)

jseldess:
@twrobel3, thanks for the suggestion! I'll loop in some CockroachDB devs familiar with node to do the actual review.

Issue #54, opened 2016-02-22
Docs versioning


jseldess:
Figure out how to provide docs for multiple versions of CockroachDB

benesch:
Since I didn't see an existing proposal, I figured I'd submit a strawman RFC for docs versioning here.

  * Create a branch for each released major version of CockroachDB: 1.0, 1.1, 1.2, 2.0, etc. For consistency with cockroachdb/cockroach, name these branches `release-x.x`, as in `release-1.0`. Right now, we'd only need `release-1.0`.

  * Auto-deploy each release branch to https://cockroachlabs.com/docs/x.x/. 1.0 documentation would live at https://cockroachlabs.com/docs/1.0, for example.

  * Auto-deploy the latest-stable documentation to https://cockroachlabs.com/docs/ so that existing URLs continue to work.

  * Auto-deploy the tip of gh-pages to https://cockroachlabs.com/docs/edge.

  * Don't handle pre-release/unstable versions specially; users on the unstable channel can use the edge docs.

  * When deploying, inject `<link rel="canonical" href="https://cockroachlabs.com/docs/FOO.html">` into non-default versions of the page with the same filename. So [https://cockroachlabs.com/docs/**1.0**/FOO.html](https://cockroachlabs.com/docs/1.0/FOO.html) and [https://cockroachlabs.com/docs/**edge**/FOO.html](https://cockroachlabs.com/docs/edge/FOO.html) would indicate their canonical URLs as https://cockroachlabs.com/docs/FOO.html. This should prevent Google from linking to non-default of pages, unless those old versions are legitimately far more relevant to the search query. (PostgreSQL documentation suffers tremendously from this problem, for example.) Pages that do not yet exist in the default version (e.g. https://cockroachlabs.com/docs/edge/NEW-FEATURE.html) would not list a canonical URL.

  * Backport commits on `gh-pages` to release branches as necessary.

  * Consult with design to build a version switcher.

The downside of this approach is that nearly every commit on `gh-pages` in the next few weeks will need to be backported to `release-1.0`. I don't see a great way around that, considering we've already started landing 1.1 features to master that will need documentation but won't make it into the 1.0.x series.

/cc @bdarnell @sploiselle

jseldess:
Your proposal looks very good to me, @benesch! Thanks for writing it up. Adding @mjibson since I've discussed basic approaches with him, too. 

[...]

PR #1605, opened 2017-06-19
Versions

benesch:
Version the docs as discussed in #54. Since the v1.0 and v1.1 docs are identical (because this PR doesn't include v1.1 updates), this PR includes a "demo" version, v2.0, to showcase how docs look once they diverge. v2.0 only includes pages for "Install Client Drivers" and "SQL Feature Support," so you'll only see it appear in the version switcher on those two pages. "SQL Feature Support" is particularly important, as it shows how pages that are renamed (in this case, to "SQL Support") can still be linked back to their old name.

Traveling tomorrow (Monday, 6/19), but can answer any questions on Tuesday!


benesch:
A few more thoughts:

* This PR ships a bare-bones version selector—just an unstyled `<select>` box. @kuanluo, what's the status of the implementation for the actual version selector?

* Build times are back up to ~7.5s on my machine, after the sidebar refactor brought them down to ~3s. (This PR essentially doubles the number of pages in the Jekyll repository, unfortunately.) This seems acceptable.

* Includes will need to be versioned as necessary, e.g. via

```
{% include some-instructions-{{ page.version }}.md %}
```

where `_includes/some-instructions-v1.0.md` and `_includes/some-instructions-v1.1.md` both exist. The only include that needed versioning so far was `sidebar-data.json`.

kuanluo:
The bare-bone version selector looks good, @benesch. I believe @jseldess's thought is to wait until your work is done to begin front-end. Just give me a heads-up when you're all good to go, and I will get Division Of involved.

jseldess:
@benesch, once again, thanks for the awesome work! I pulled a copy of your branch and am testing, and everything seems to work as you say. A few questions:

- The docs home (`index.md`) and the release notes pages don't show a version switcher. In the case of `index.md`, I think that's just because the file only exists in 1.0. Is that intentional? In the case of the release notes pages, is that just a side-effect of the pages not being versioned at all? If so, I wonder if there's a workaround, since it's weird for one page to show the switcher and another not show the switcher. 
- As you say, if an include needs to be different between versions, we'll need multiple instances of the include file. This is true of sql diagram images as well, but the sql diagram generator currently outputs a version-agnostic image. We'll need to change the generate to somehow factor the version into the file name, right, @mjibson? 
- When I serve up the site locally, I'm seeing some `not found` errors, but I can't figure out why (see below). Any ideas?

```
[2017-06-19 18:18:25] ERROR `/docs/js.hsforms.net/forms/v2.js' not found.
[2017-06-19 18:18:25] ERROR `/docs/https:/cdn.optimizely.com/js/7825500667.js' not found.
[2017-06-19 18:18:25] ERROR `/docs/js.hsforms.net/forms/v2.js' not found.
[2017-06-19 18:18:25] ERROR `/docs/https:/cdn.optimizely.com/js/7825500667.js' not found.
[2017-06-19 18:18:30] ERROR `/docs/js.hsforms.net/forms/v2.js' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/sidenav-arrow.svg' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/back_to_top.png' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/close_icon.png' not found.
[2017-06-19 18:18:30] ERROR `/docs/js.hsforms.net/forms/v2.js' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/sidenav-arrow.svg' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/back_to_top.png' not found.
[2017-06-19 18:18:30] ERROR `/docs/stable/images/close_icon.png' not found.
[2017-06-19 18:18:31] ERROR `/docs/stable/images/favicon.png' not found.
[2017-06-19 18:18:38] ERROR `/docs/js.hsforms.net/forms/v2.js' not found.
[2017-06-19 18:18:38] ERROR `/docs/edge/images/sidenav-arrow.svg' not found.
[2017-06-19 18:18:38] ERROR `/docs/edge/images/back_to_top.png' not found.
[2017-06-19 18:18:38] ERROR `/docs/edge/images/close_icon.png' not found.
```
- On the teamcity test site, I always see `stable`, `edge`, or `v2.0` in the URL. The `stable` content will also be available at `/doc` without any version indication as well, right?

jseldess:
@sploiselle, @mjibson, this definitely needs your review, too.

sploiselle:
If the page doesn't exist in a version, I would prefer to still have it show up in the selector with a behavior that tells the user what happened if they try to version switch to a non-existent page.

Rather than hiding versions that don't have a version of the currently loaded page, I think we can provide a more sensible experience by:
- Showing all versions on all pages
- If a user tries to 'version switch' to a page that DNE, redirect them to the home page with a notification that the page they were on doesn't exist in the version they tried to load.

My rationale behind this is that it's unclear from an uninitiated user's perspective why the UX is inconsistent. There's also a potential issue of "skipping versions" if we re-add a page with the same name; e.g. the user sees versions 1.0,1.1,2.1,3.0.

FWIW, I have not yet had time to map out all of these behaviors so we can come to an agreement on the desired behaviors buuuuuuut it's near the top of my priorities.

benesch:
Ok, I've updated this PR to always show all versions in the version selector, rather than hiding the versions where the current page doesn't exist. Versions where the current page doesn't exist are instead marked as "disabled," which means they render as greyed out and unselectable. I think this is even better better UX than what you proposed, @sploiselle, since users are shown which versions aren't available (perhaps we could even provide an explanatory tooltip?) before they've wasted time clicking on the link. If @sploiselle's on board, @kuanluo, could you give us a design for how these unselectable versions would look?

@jseldess, I've addressed everything you mentioned, I think! In order:

  * The docs home is now versioned, so it shows a functional version switcher. The release notes are still unversioned, but the version switcher now displays anyway, albeit with all versions marked as disabled. 

  * We'll definitely need to make the generator smarter, but the workaround you propose, I hope, is good enough for now? I want to avoid expanding the scope of this PR any more than necessary given the number of conflicts it's already generated. @dt and @mjibson are planning to move that generator into cockroachdb/cockroach soon anyway.

  * The not found errors were images and scripts I'd forgotten to update for the new URL scheme. Those should all be fixed now.

  * `/docs/FOO.html` now redirects to `/docs/stable/FOO.html`. I prefer the redirect to duplication here. We already have `/v1.0` and `/stable`, which together serve the two important purposes: a permalink for those who want to refer to the same content until the end of time, and a convenience link for those who want to always point at the latest released version. A copy at `/` doesn't serve a clear purpose (besides not breaking existing links, which the redirect handles equally well), and its URL doesn't communicate which version its serving, either.

Let me know what you guys think! If you're on board with this behavior, let's get Division Of working on implementing @kuanluo's mockups? Essentially every PR that lands on `master` causes a conflict with this one, so the sooner we can land this the better. (In that vein, it would also be great to defer any minor tweaks to behavior until after this lands. Major tweaks, of course, are worth sorting out now so we don't commit ourselves to e.g. URL schemes we'll regret.)

kuanluo:
@benesch Your solution sounds good to me. 

All, just give me a "go ahead" and I will assign the cosmetic part of the versioning to Division Of pronto.

sploiselle:
@benesch Even better than my proposal! Thanks so much. Looks great to me.

jseldess:
Update: @benesch, aka Superroach, is going to wrap up the styling work that we were planning to punt to Division Of.

benesch:
RFAL! I took a shot at the mobile styles, too. They're not perfect, but hopefully good enough to get this merged.

kuanluo:
Awesome, @benesch! A few nits: 

- The tooltip looks a bit funny with breaking lines. Can we use single line for the full sentence? Another idea would be to say something like  "v2.0 (unavailable)" to avoid the tooltip completely. 
image
- The horizontal line above is a tad too thick. Just 1px would be fine.
image

On mobile, moving the versioning into the main dropdown nav makes the most sense to me. It's unlikely people are switching between versions on pages, and we can also test this hypothesis for the first few weeks the versioning is live. Happy to punt this part of the work over to DivOf, if that's easier. 

cc: @jseldess

jseldess:
@benesch, I spoke to @petermattis, @nstewart, and @bdarnell about our url conventions. Let's go with `/stable` and `/dev`. 

Also, in the version switcher, I'd prefer that we do `Version: <version number>` and just list those numbers in the drop-down as well. So right now for stable, it'd be `Version: 1.0`. For dev, it'd be `Version: 1.1-alpha`. 

I don't think we need to repeat "stable" and "dev" in the switcher, but perhaps I'm being overly minimal. Perhaps we could just have a "stable" label-type thing in the drop-down but not in `Version: 1.0`? Thoughts? 

Of course, beyond renaming the url structure, if you want to punt any of this to Division Of, that's totally fine.

benesch:
Ok, I think this is mergeable. PTAL!

Updates:

* URLs use `stable` and `dev`.
* Version switcher displays "Version: vFOO" by default. Expanding it reveals "v1.0 (Stable)", "v1.1-alpha (Dev)", and "v2.0". Punting on making "(Stable)" and "(Dev)" proper, well-designed badges for now.
* v1.0 includes all changes on master and v1.1 includes all changes on release-1.1. Punting on #1609 for now.

benesch:
omg it's finally green! @jseldess let me know when this is good to merge!

kuanluo:
@benesch @jseldess 

Sorry my feedback came in a little late. The versioning control on mobile behavior is a bit funky:
image

And in one of the earlier versions, I saw the attempt to combine the versioning with the main nav. I like that idea a lot - are we still sticking with at?

benesch:
Yes, absolutely! I just wanted to punt any remaining design work to Division Of. Since this PR made such sweeping changes, every day it stayed open made it that much harder to merge, so we figured it was good enough to merge for now.

kuanluo:
Ah, got it. Thanks for the clarification, @benesch! I will assign it right away.

PR #1610, opened 2017-06-21
Cloud migration tutorial

jseldess:
A tutorial showing users how to use a local cluster to simulate cross-cloud deployment and then migration from one cloud to another. 

I'm planning to add a simple screencast (with audio or text) demoing this across real cloud platforms, and offering that video as something to watch before you begin. But I'm opening this to get initial feedback on the flow and basic messaging.


demo-cloud-migration.md
```diff
@@ -0,0 +1,199 @@
...
+Make sure you have already [installed CockroachDB](install-cockroachdb.html).
+
+## Step 1. Start a 3-node cluster on "cloud 1"
+
+If you've already [started a local cluster](start-a-local-cluster.html), these commands should be familiar to you. The new flag to note is [`--locality`](configure-replication-zones.html#descriptive-attributes-assigned-to-nodes), which accepts key-value pairs that describe the locality of a node. In this case, you're using the flag to specify that the first 3 nodes are running on "cloud platform 1".
+
+~~~ shell
+# In a new terminal, start node 1 on cloud 1:
+$ cockroach start --insecure \
+--locality=cloud=platform1 \
```
spencerkimball: Why "cloud=platform1" instead of just "cloud=1"?

demo-cloud-migration.md
```diff
@@ -0,0 +1,199 @@
...
+--host=localhost \
+--port=25259 \
+--http-port=8082 \
+--join=localhost:26257 \
+--cache=100MB
+~~~
+
+## Step 2. Set up HAProxy load balancing
+
+Each CockroachDB node is an equally suitable SQL gateway to your cluster, but to ensure client performance and reliability, it's important to use TCP load balancing. [HAProxy](http://www.haproxy.org/) is one of the most popular open-source TCP load balancers, and CockroachDB includes a built-in command for generating a configuration file that is preset to work with your running cluster, so you'll use that tool here.
```
spencerkimball: How about: "..., but to ensure an even balancing of client requests across available nodes, we can use a TCP load balancer."

demo-cloud-migration.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Step 6. Watch data balance across all 6 nodes
+
+Back in the Admin UI, hover over the **Replicas per Node** graph again. Because you set [`--locality`](configure-replication-zones.html#descriptive-attributes-assigned-to-nodes) settings to specify that nodes are running across 2 clouds, you'll see an approximately even replica count on each node, indicating that CockroachDB has automatically rebalanced replicas across the cloud platforms:
+
+<img src="images/admin_ui_replicas_migration2.png" alt="CockroachDB Admin UI" style="border:1px solid #eee;max-width:100%" />
+
+## Step 7. Migrate all data to "cloud 2"
+
+In a new terminal, edit the default replication zone, adding a [hard constraint](configure-replication-zones.html#replication-constraints) that all replicas must be on nodes with `--locality=cloud=platform2`:
```
spencerkimball: Why don't you provide some colorful additional narration to these steps that are demonstrating the value proposition. For example: "OK, we're now running three nodes one cloud 1. But what if we'd like to start experimenting with resources provided by another cloud vendor? Let's try that by adding three more nodes to a new cloud platform."

"We now have the cluster replicating across our two simulated cloud vendors. But after experimentation, we're happy with cloud vendor 2, and we decide that we'd like to move everything there. Can we do that without interruption to our live client traffic? Yes, and it's as simple as adding a constraint to our replication configuration that all replicas must be chosen from cloud vendor 2."

You get the idea...we need to give people a scenario so it's not just an abstraction that only folks who've been struggling specifically with these problems will understand.

demo-cloud-migration.md
jseldess: Good point. Done.

demo-cloud-migration.md
jseldess: Done.

demo-cloud-migration.md
jseldess: Great feedback. I'll work on this.

bdarnell:
LGTM

---

Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[demo-cloud-migration.md, line 7 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBkuBv8ybzkKyoJTfi:-KnBkuBv8ybzkKyoJTfj:bd9b2ui) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L7)):*
> ```Markdown
> ---
> 
> CockroachDB's flexible [replication controls](configure-replication-zones.html) make it trivially easy to run a single CockroachDB cluster across cloud platforms or to migrate a cluster from one cloud to another without any service interruption. This page walks you through a local simulation of the process.
> ```

Cross-cloud migration is just a special case of cross-DC migration. Since this is just a simulation, it seems weird to say it's a simulation of the special case instead of the general case. But I guess that's what gets the clicks...

---

*[demo-cloud-migration.md, line 55 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBlMR8K6QUuRSa4_9T:-KnBlMR8K6QUuRSa4_9U:b-in9hpr) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L55)):*
> ```Markdown
> 
> ~~~ shell
> $ brew install haproxy
> ```

This is mac-specific. I'm OK with not having instructions for other platforms at this point, but we should say something like "In a new terminal, install HAProxy. If you're on a Mac and using Homebrew, use `brew install haproxy`"

---

*[demo-cloud-migration.md, line 64 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBlotj0phmUGnk9I1P:-KnBlotj0phmUGnk9I1Q:b8x2bar) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L64)):*
> ```Markdown
> ~~~
> 
> This command generates an `haproxy.cfg` file automatically configured to work with the 3 nodes of your running cluster. In the file, change `bind :26257` to `bind :26270`. This changes the port on which HAProxy accepts requests to a port that is not already in use by a node and that won't be used by the nodes you'll add later.
> ```

These port numbers all start to run together. How about a round number like 26000 instead of 26270?

---

*[demo-cloud-migration.md, line 66 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBmm9_TEOW4yp1a_YK:-KnBmm9_TEOW4yp1a_YL:b-g7fz9g) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L66)):*
> ```Markdown
> This command generates an `haproxy.cfg` file automatically configured to work with the 3 nodes of your running cluster. In the file, change `bind :26257` to `bind :26270`. This changes the port on which HAProxy accepts requests to a port that is not already in use by a node and that won't be used by the nodes you'll add later.
> 
> ~~~ shell
> ```

`shell` mode doesn't highlight this correctly (it colors `bind` but not any other keywords). Unless there's a more precise option, better to just remove the `shell` tag completely.

---

*[demo-cloud-migration.md, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBn9TBThWFEPX3mXhi:-KnBn9TBThWFEPX3mXhj:b-2vuaqr) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L98)):*
> ```Markdown
> 
> ~~~ shell
> $ go get github.com/cockroachdb/loadgen/ycsb
> ```

There should be an introductory section that tells people they'll need to install Go (and we could put the haproxy installation instructions there too)

---

*[demo-cloud-migration.md, line 183 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBoaVwSS_d1s3pGOSb:-KnBoaVwSS_d1s3pGOSc:b-cqrap5) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L183)):*
> ```Markdown
> ## Step 9. Stop the cluster
> 
> Stop HAProxy and YCSB by switching into their terminals and pressing **CTRL + C**. Do the same for each CockroachDB node.
> ```

Stop YCSB first, or it will start spewing errors as soon as you stop haproxy. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-:-KnBp6kzhhd-2WoHwWoW:bb6ywmx)*

jseldess:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[demo-cloud-migration.md, line 64 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBlotj0phmUGnk9I1P:-KnGKhPlcDSnreaikA8p:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L64)):*

These port numbers all start to run together. How about a round number like 26000 instead of 26270?
</blockquote></details>

Done.

---

*[demo-cloud-migration.md, line 66 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBmm9_TEOW4yp1a_YK:-KnGKgr2saQYz66BTkFy:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L66)):*

`shell` mode doesn't highlight this correctly (it colors `bind` but not any other keywords). Unless there's a more precise option, better to just remove the `shell` tag completely.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1610)*

nstewart:
Minor feedback (we discussed in person too)
1. it would be useful to have people start in some sort of ./cloud-migration-test directory so they don't get a bunch of data files in their home
2. as written, the ycsb command doesn't work after the go-install. need to include the command to build
3. We shouldn't pre-split the ycsb ranges. I think this breaks (or at least doesn't reinforce) a beginners mental model for how CRDB works. Taking off the split shows new ranges emerging as new load comes in, vs just having it be flat the whole time.
4. In the final 'aha' graph where the load switches we ask users to hover over the graph to see the replica counts, but in my instance all but 1 were zero, which clearly wasn't correct. I think I would just say look at the graph, it's clear from the shape and color change that the load has moved to the other cloud.  For the record, I just realized my crdb local install is beta, but according to Jesse this is still a bug.

jseldess:
Review status: 0 of 10 files reviewed at latest revision, 9 unresolved discussions.

---

*[demo-cloud-migration.md, line 55 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBlMR8K6QUuRSa4_9T:-KnGeISlNiRsNA9VxBaA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L55)):*

This is mac-specific. I'm OK with not having instructions for other platforms at this point, but we should say something like "In a new terminal, install HAProxy. If you're on a Mac and using Homebrew, use `brew install haproxy`"
</blockquote></details>

Done.

---

*[demo-cloud-migration.md, line 98 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBn9TBThWFEPX3mXhi:-KnGeJqg2AksbbSfHvSg:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L98)):*

There should be an introductory section that tells people they'll need to install Go (and we could put the haproxy installation instructions there too)
</blockquote></details>

Done.

---

*[demo-cloud-migration.md, line 183 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1610#-KnBoaVwSS_d1s3pGOSb:-KnGLEFR45aTZpy8AT8p:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f8645eb6b95b234363c8ca97c85ab28b6092f424/demo-cloud-migration.md#L183)):*

Stop YCSB first, or it will start spewing errors as soon as you stop haproxy. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1610)*

jseldess:
@nstewart:

- Added a "before you begin" suggestion to create a new directory for running nodes.
- Moved all installations to "before you begin". And since `go get` does in fact build ycsb in `$HOME/go/bin` by default, I've added that path to the command to start ycsb.
- No longer pre-splitting ranges with ycsb. 
- Called out that the replica count on hover will be off for a few minutes but that the graph lines are accurate. Also pointed to the node list view for another accurate view of per-node replica counts.

jseldess:
Merging. Will open a follow-up PR for the screencast part.

Issue #1606, opened 2017-06-19
Build an App: Update txn code samples to return values as part of the transaction


jseldess:
In our code samples with transaction retry logic, the transaction doesn't return anything but errors or, sometimes, a success message. The code performs a select on the updated account balances *after* the transaction. 

Instead, we should adapt the transaction wrapper to allow values to be returned to the client, and then return the updated account balances as part of the transaction. 

- [ ] [Go](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.go)
- [ ] [Python](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.py)
- [ ] [Ruby](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.rb)
- [ ] [Java](https://github.com/cockroachdb/docs/blob/master/_includes/app/TxnSample.java)
- [x] [Node](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.js) (The wrapper is ready (see https://github.com/cockroachdb/docs/pull/1603), but the transaction itself needs to be updated and the `SELECT` following the transaction needs to be removed) 
- [ ] [C++](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.cpp)
- [ ] [Clojure](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.clj)
- [ ] [PHP](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.php)
- [ ] [Rust](https://github.com/cockroachdb/docs/blob/master/_includes/app/txn-sample.rs)

awoods187:
@vy-ton can PTAL at this?

vy-ton:
@jseldess I'm taking a look at this.

Are you suggesting perform the [select](https://docs.google.com/document/d/1uOsExAcVaGfT_Hh1VPBmFpg1x8R43g3OrKNKAHrCSAs/edit) as part of the transaction itself?

jseldess:
We have closed this issue because it is more than 3 years
old. If this issue is still relevant, please add a comment and reopen
the issue. Thank you for your contribution to CockroachDB docs!

PR #1615, opened 2017-06-22
Update node txn sample to pass results back from the transaction

twrobel3:
Adjust the node.js transaction sample code to perform a SELECT as
part of the transaction and return the results through the txn
wrapper.

Addresses node component of issue #1606

### DIFF
--- a/_includes/app/txn-sample.js
+++ b/_includes/app/txn-sample.js
@@ -84,15 +84,20 @@ function transferFunds(client, from, to, amount, next) {
     var acctBal = results.rows[0].balance;
     if (acctBal >= amount) {
       // Perform the transfer.
-      async.series([
+      async.waterfall([
         function (next) {
           // Subtract amount from account 1.
           client.query('UPDATE accounts SET balance = balance - $1 WHERE id = $2', [amount, from], next);
         },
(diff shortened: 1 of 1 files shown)


twrobel3:
@jseldess @justinj 

Following up on the changes made in #1603, this returns the updated balances within the transaction in the node sample code, addressing the node.js component of #1606

_includes/app/txn-sample.js
```diff
@@ -84,15 +84,20 @@ function transferFunds(client, from, to, amount, next) {
     var acctBal = results.rows[0].balance;
     if (acctBal >= amount) {
       // Perform the transfer.
-      async.series([
+      async.waterfall([
```
twrobel3: Switched from series to waterfall here, since with waterfall only the last async step has it's results sent to the final async callback.  Series returns the results of each function as separate callback parameters.

With series the outer function would need to have a signature along the lines of `function (err, firstUpdateResults, secondUpdateResults, selectResults) {` which I didn't think was as clear as this, where the transfer funds function returns just the updated funds.

jseldess:
Also, @justinj tested your change, and everything works great. Merging.

PR #1616, opened 2017-06-22
Fix minikube instructions.

mberhault:
The `minikube.sh` script no longer exists as of
https://github.com/cockroachdb/cockroach/pull/16536

This is because it's no longer needed.


jseldess:
@mberhault, changes look good, I think, but it looks like your branch branch is failing tc because it is missing a directory that got added to master sometime recently, `/ci`.

Can you getting an updated `master` and then rebasing against it?

PR #1618, opened 2017-06-22
Use --host to connect sql client

robert-s-lee:
Add --host when connecting to prevent the following when following the doc.

$ cockroach sql \
> --certs-dir=certs
Error: unable to connect or connection lost.

Please check the address and credentials such as certificates (if attempting to
communicate with a secure cluster).

dial tcp 127.0.0.1:26257: getsockopt: connection refused
Failed running "sql"

The change below with host works shown

$ cockroach sql --certs-dir=certs --host=10.0.3.98
# Welcome to the cockroach SQL interface.
# All statements must be terminated by a semicolon.
# To exit: CTRL + D.
root@10.0.3.98:26257/>

### DIFF
--- a/manual-deployment.md
+++ b/manual-deployment.md
@@ -201,7 +201,8 @@ CockroachDB replicates and distributes data for you behind-the-scenes and uses a
 
 	~~~ shell
 	$ cockroach sql \
-	--certs-dir=certs
+	--certs-dir=certs \
+	--host=<node1 address>
 	~~~
 
 	~~~ sql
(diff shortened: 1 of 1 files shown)


robert-s-lee:
made the same change in insecure.   https://github.com/cockroachdb/docs/pull/1620

jseldess:
Thanks, @robert-s-lee. In the future, no need to open a separate PR for a related change like that. You can just add it as a separate commit. Happy to show you how sometime if you like.

PR #1619, opened 2017-06-22
Tweaks to Node.js build an app tutorial

jseldess:
- Comment edit in node txn wrapper
- Minor updates to flow of node txn step

### DIFF
--- a/build-a-nodejs-app-with-cockroachdb.md
+++ b/build-a-nodejs-app-with-cockroachdb.md
@@ -43,7 +43,7 @@ Now that you have a database and a user, you'll run code to create a table and i
 
 First, use the following code to connect as the `maxroach` user and execute some basic SQL statements, creating a table, inserting rows, and reading and printing the rows.
 
-You can download the <a href="https://raw.githubusercontent.com/cockroachdb/docs/master/_includes/app/basic-sample.js" download><code>basic-sample.js</code></a> file or create the file yourself and copy the code into it.
+Download the <a href="https://raw.githubusercontent.com/cockroachdb/docs/master/_includes/app/basic-sample.js" download><code>basic-sample.js</code></a> file, or create the file yourself and copy the code into it.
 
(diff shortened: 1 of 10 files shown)


build-a-nodejs-app-with-cockroachdb.md
```diff
@@ -43,18 +43,32 @@ Now that you have a database and a user, you'll run code to create a table and i
...
 <a href="https://raw.githubusercontent.com/cockroachdb/docs/master/_includes/app/basic-sample.js" download>download it directly</a>.
 
 ~~~ js
 {% include app/basic-sample.js %}
 ~~~
 
+Then run the code:
+
+~~~ shell
+$ node basic-sample.js
```
justinj: Have we instructed them at this point to call the file this? Does that matter?

build-a-nodejs-app-with-cockroachdb.md
```diff
@@ -63,7 +77,21 @@ You can copy the code or
 {% include app/txn-sample.js %}
 ~~~
 
-After running the code, use the [built-in SQL client](use-the-built-in-sql-client.html) to verify that funds were transferred from one account to another:
+Then run the code:
+
+~~~ shell
+$ node txn-sample.js
```
justinj: same here? I don't actually feel strongly, just want to be sure this is intentional.

build-a-nodejs-app-with-cockroachdb.md
jseldess: Good point. I've tweaked the language a bit to be clearer about this. PTAL.

build-a-nodejs-app-with-cockroachdb.md
jseldess: Same as above. If you prefer this new approach, I'll also update the other build an app tutorials accordingly.

justinj:
LGTM!

Issue #1070, opened 2017-02-04
Add a "copy to clipboard" button


jseldess:
To make it easier/faster to run our sample commands and code, we should have an option to add a "copy to clipboard" feature to a code block. 

This needs to be an option (not universally applied to all code blocks) because some blocks don't warrant copy/past (command responses), and other blocks will need to be edited to work meaningfully with copy/paste (e.g., multiple sql commands in a single block will need to be split into separate blocks).

jseldess:
How other sites do it:

# Django

Django provides a copy to clipboard feature only for entire files:

![screen shot 2017-06-07 at 10 34 04 am](https://user-images.githubusercontent.com/15893490/26884135-0955b84c-4b6d-11e7-8022-f61c82d7c2b8.png)

Not for individual commands:

![screen shot 2017-06-07 at 10 35 54 am](https://user-images.githubusercontent.com/15893490/26884170-21673604-4b6d-11e7-9e4c-7f4a90305f9d.png)

# Google SDK

Google dev docs provide a copy to clipboard feature (shows up on hover) for individual commands:

![screen shot 2017-06-07 at 10 41 36 am](https://user-images.githubusercontent.com/15893490/26884549-05d643ac-4b6e-11e7-847f-0c6765cafc22.png)

Probably also for entire files, but I can't find any examples of that at the moment.

# Kubernetes

Kubernetes docs provide a copy to clipboard feature for entire files:

![screen shot 2017-06-07 at 10 47 13 am](https://user-images.githubusercontent.com/15893490/26884833-c00a129e-4b6e-11e7-9431-e3f9ba8e3f83.png)

But not for individual commands.

jseldess:
@kuanluo, examples of where we could use this feature in our docs:

## Single CLI command

[Start a Local Cluster](https://www.cockroachlabs.com/docs/start-a-local-cluster.html#step-1-start-the-first-node):

![screen shot 2017-06-07 at 1 26 35 pm](https://user-images.githubusercontent.com/15893490/26892062-11c08332-4b85-11e7-85eb-7957c0b30716.png)

## Single SQL command

[`CREATE VIEW`](https://www.cockroachlabs.com/docs/create-view.html#example):

![screen shot 2017-06-07 at 1 30 27 pm](https://user-images.githubusercontent.com/15893490/26892208-8d0928c8-4b85-11e7-929b-7a2682505d23.png)

## Entire file

[...]

PR #1624, opened 2017-06-23
Add copy to clipboard

deronjohnson:
Addresses #1070 and #1529.

@jseldess For this to work, we need to add some markup (_includes/copy-clipboard.html) to the code snippets, as well as the class "highlight--clipboard" to the code wrapper. However, I'm not sure how we want to implement this, because the code markup is generated by Redcarpet, which I could not find a way to alter. I noticed a few other places in the repository (ex. /install-cockroachdb) where the snippets were hardcoded, it seems to add analytics attributes to code blocks. We might have to take this same approach for snippets that can be copied, although I'm open to other ideas. For now, I added two plain html examples to /start-a-local-cluster so you could see the functionality. The second example strips backslashes followed by new lines, which you discussed in #1529.


jseldess:
@deronjohnson, thanks for this first attempt. The install page is an exception in the way it hardcodes the html for code samples, as you say, for tracking purposes. That's way too burdensome to do throughout the docs, though, so it seems we do need to find another way. 

@benesch and @mjibson, do either of you have ideas?

@deronjohnson, is the requirement that the button be inside the block making this difficult. Would it be easier if it were, say, just beyond the right margin?

deronjohnson:
@jseldess, I was actually able to solve this with CSS. I removed the two examples I hard coded and made the first two code snippets on /start-a-local-cluster copyable. All that is required now is placing {% include copy-clipboard.hrtml %} above a code snippet. You'll notice that the second example doesn't strip the new lines. I wasn't sure if you wanted to concatenate everything to one line only when there were backslashes, or always. Let me know and I can update.

benesch:
Looks awesome, @deronjohnson!

@jseldess, we want these on essentially every code sample, right? If so, perhaps we can just loop over every `<pre>` element on `document.ready` and inject the copy-to-clipboard HTML with JavaScript?

jseldess:
This is excellent, @deronjohnson! 

One request: It's great that you're stripping newlines when there are backslashes! But we seem to also be adding a final newline. When pasted into the terminal, this makes the command immediately execute. Instead, it'd be good for users to be able to review and replace values. So can you strip that final newline from copied code?

@benesch, we actually don't want this feature applied by default. Some code blocks show command responses/outputs. So for now, we'll need the ability to apply the feature only where relevant.

benesch:
Ah, gotcha. @deronjohnson's solution seems about as terse as it's going to get, in that case!

deronjohnson:
@jseldess, the first example should now work as you described, pasting without the final newline.

start-a-local-cluster.md
```diff
@@ -29,11 +29,12 @@ Also, feel free to watch this process in action before going through the steps y
 
 ## Step 1. Start the first node
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ cockroach start --insecure \
 --host=localhost
 ~~~
-
+{% include copy-clipboard.html %}
```
jseldess: Can you remove the include here, since the block shows a command response that users won't need to copy?

jseldess:
Thanks, @deronjohnson! Looks great. I'll merge.

jseldess:
@deronjohnson, just noticed a bug. The copy function seems to remove periods. For example,  copying this:

```
CREATE TABLE bank.accounts (id INT PRIMARY KEY, balance DECIMAL);
```

ends up as this:

```
CREATE TABLE bankaccounts (id INT PRIMARY KEY, balance DECIMAL);
```

Or copying this:

```
INSERT INTO bank.accounts VALUES (1, 1000.50);
```

ends up as this:

```
INSERT INTO bankaccounts VALUES (1, 100050);
```

Any ideas?

benesch:
@jseldess: https://github.com/cockroachdb/docs/pull/1631

PR #1627, opened 2017-06-26
Update restore.md

robert-s-lee:
Fix double quote to single quote


jseldess:
Nevermind, @robert-s-lee. I just made that final tweak myself.

PR #1631, opened 2017-06-27
preserve dots when copying to clipboard

benesch:
Looks like the copy-to-clipboard sanitizing regex was sourced from a StackOverflow question titled "Regex for removing leading and trailing carriage returns and line feeds from a sentence" (https://stackoverflow.com/a/2550845/1122351), but the question text sneakily also asked for the solution to combine this with "removing periods from within the string."

We want to keep these dots when copying to the clipboard, so remove the part of the regex that strips these dots.

### DIFF
--- a/js/customscripts.js
+++ b/js/customscripts.js
@@ -250,7 +250,7 @@ $(function() {
     },
     text: function(trigger) {
       var text = $(trigger).next().find('code').text();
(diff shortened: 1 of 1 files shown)


benesch:
(@jseldess: heads up, I haven't tested this yet.)

jseldess:
Thanks, @benesch. Will test in the morning. 

@deronjohnson, heads up that @benesch may have resolved this already.

jseldess:
Just tested. Works like a charm. Thanks again, @benesch!

PR #1636, opened 2017-06-29
FAQ: What happens when node clocks are not properly synchronized?

### LINKED ISSUE #1303: managing clock drift - commonly asked question
In response to the quote from Jepsen: Like Spanner, Cockroach's correctness depends on the strength of its clocks. If any node drifts beyond the clock-offset threshold (by default, 250 ms), all guarantees are basically out the window.

Default of clock-offset threshold is set at 500ms. Nodes will spontaneously abort when they detect a drift half of the configured value. The guarantee here is serialized consistency - if you drift beyond that, you can potentially get stale reads and write skews. However, with NTP set up, you should not be experiencing clock drifts beyond 500ms. We are constantly measuring clock offsets and have more than a generous default max offset. You would need to have a node's clock jump suddenly by more than 500ms, and even then, you would be facing 3s window of vulnerability. NTP keeps your clock offsets within tens of ms.

In the worst case if it actually happened, reads could see stale data. Writes are mostly unaffected by clocks, except to the extent that you may write data derived from stale reads.

jseldess:
- Add FAQ on clock synchronization
- Update `--max-offset` flag description
- Add links from general FAQ to more details

Fixes #1303


operational-faqs.md
```diff
@@ -43,6 +43,12 @@ As of the 1.0 release, there is no way to change the number of days before times
...
+
+The one rare case to note is when a node's clock suddenly jumps beyond the maximum offset before the node detects it. Although extremely unlikely, this could occur, for example, when running CockroachDB inside a VM and the VM hypervisor decides to migrate the VM to different hardware with a different time. In this case, there can be a small window of time between when the node's clock becomes unsynchronized and when the node spontaneously shuts down. During this window, it would be possible for a client to read stale data and write data derived from stale reads.
+
```
jseldess: Not sure if the last sentence is sufficient, or if we need to go into greater detail, with an example, e.g:

> Stale reads: Each range is comprised of multiple replicas of the same span of key-values. At any given time, one replica holds a "lease" indicating that it is the replica that receives and coordinates read and write requests for its range. If a node with a slow clock is the lease-holder, the lease could expire and a node with a fast clock could acquire it and start making writes while the slow node still thinks its lease is valid and continues to serve stale reads.

operational-faqs.md
```diff
@@ -43,6 +43,12 @@ As of the 1.0 release, there is no way to change the number of days before times
...
 
+## What happens when node clocks are not properly synchronized?
+
+By default, CockroachDB's maximum allowed clock offset is 500ms. When a node detects that its clock offset, relative to other nodes, is half of the maximum allowed (250ms), it spontaneously shuts down. This is well in advance of the maximum offset, beyond which [serializable consistency](https://en.wikipedia.org/wiki/Serializability) is not guaranteed and stale reads and write skew could occur. With NTP or other clock synchronization software running on each node, there's very little risk of ever exceeding the maximum offset and encountering such anomolies, and even on well-functioning hardware not running synchronization software, slow clock drift is most common, which CockroachDB handles safely.
```
knz: "is half or more of the maximum allowed (250ms)"

"and write skews"

"anomalies"

jseldess: Done.

bdarnell:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1636#-:-Knoa4x9Pk2lki5pJ2Ro:bnfp4nl)*

PR #1638, opened 2017-06-29
Add recommendation against using Azure A-series VMs

### LINKED ISSUE #1356: Microsoft Azure best practices
- Incidental VM migrations wipe disks; can reduce past quorum 
- Clock issues (cc @mberhault for more detail)

jseldess:
Fixes #1356


jseldess:
@bdarnell, @mberhault, this new warning only refers to corruption and performance degradation due to bad filesystems after A-series VM migrations. Should I also add details about clock issues on Azure? I may be able to piggyback on https://github.com/cockroachdb/docs/pull/1636...

mberhault:
We haven't had clock issues in Azure in a while. Maybe they finally rolled out their new version that was supposed to fix it?

mberhault:
LGTM

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[_includes/azure-warning.html, line 1 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1638#-KnoZVvilyVG3cNmX-h6:-KnoZVvilyVG3cNmX-h7:b-qwoqko) ([raw file](https://github.com/cockroachdb/docs/blob/570ce006b1957988583a82de6bf476c25c4c920d/_includes/azure-warning.html#L1)):*
> ```HTML
> {{site.data.alerts.callout_danger}}On <a href="deploy-cockroachdb-on-microsoft-azure.html">Microsoft Azure</a>, A-series VMs can come back from VM migrations with bad filesystems, which can in turn cause data corruption and performance degradation. We therefore recommend not using A-series VMs. Instead, use D-series VMs, which have not been seen to exhibit these problems. Before using any other Azure VM series in production, you should test them extensively.{{site.data.alerts.end}}
> ```

This is not about VM migrations (that issue is common to *all* azure VMs. It's not data corruption, though, just a performance problem). The issue is that the local disk of A-series VMs just eats data randomly. 

Actually, these azure instructions don't specify where the data lives, so they're using the root filesystem. This works (and appears to be immune to the A-series problems), but you have very limited space there. We should update the docs to discuss how to use the local disk (as well as the alternative of using premium storage. Microsoft heavily recommends this and is reluctant to provide support if you're not using it). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1638)*

jseldess:
Closing. Will open another PR once the approach is clear.

PR #1609, opened 2017-06-19
Refactor the SHOW/SET documentation; include USE and SET NAMES.

knz:
Both SHOW and SET are really single statements which act on the
session (and they are also handled using a single code path in
CockroachDB!), with the exception of SET TRANSACTION which acts on the
current transaction. This patch acknowledges the commonality and
merges the various documentation pages into two more general pages.

Extends #1107, off `release-1.1`.

Fixes #1502.
Fixes #1503.

Also covers the change implemented in cockroachdb/cockroach#16459.

Issue #1546, opened 2017-06-08
Move /generate to cockroach repo


jseldess:
If we move all the sql railroad generation files into cockroach, we'll be able to version the generation file alongside the sql grammer. This isn't a huge problem now but will probably become more of a problem. 

cc: @mjibson and @dt

jseldess:
@benesch, @dt, @mjibson, any updates on when we might be able to get to this?

benesch:
It's halfway done! https://github.com/cockroachdb/cockroach/pull/17024

Are the diagrams/functions/operations out of date?

dt:
@mjibson you think you'll move the diagram generator? I'm happy take a stab at it, but I'm less familiar with it (I don't have the jar, for example)

mjibson:
I won't be able to work on this until after the feature freeze. There's no need for the jar: it'll use the web service if the jar isn't present. I'm happy to do the work later. If it's important for it to be done sooner, it shouldn't be too hard. The only thing to note is that we have a copy of the yacc parser already in pkg/internal/yacc, but otherwise everything just gets copied, and the relative directory for the output needs fixing.

jseldess:
@benesch, the 1.0 sql diagrams and functions docs should be correct, but it's of course possible that somethings are behind.

jseldess:
Related to https://github.com/cockroachdb/cockroach/issues/18644

jseldess:
Closed by https://github.com/cockroachdb/docs/pull/2355.

PR #1648, opened 2017-06-30
version the sql diagrams

benesch:
#1609 is blocked on this, and it's an important step towards #1546.


benesch:
This should be the last massive PR like this!

jseldess:
@benesch, finally getting to this. So sorry for the delay. The change make sense in principle, but it doesn't seem to work with the diagram generator. For example, when I try to regenerate a single diagram:

```
$ ~/src/github.com/cockroachdb/docs$ cd generate/
~/src/github.com/cockroachdb/docs/generate$ go run main.go --filter add_column
2017/07/10 11:02:23 generate BNF: https://raw.githubusercontent.com/cockroachdb/cockroach/master/pkg/sql/parser/sql.y
2017/07/10 11:02:24 parse: alter_table_stmt, inline: [alter_table_cmds alter_table_cmd column_def], descend: false
2017/07/10 11:02:28 add_column: generated railroad diagram
2017/07/10 11:02:28 add_columnopen ../_includes/sql/diagrams/add_column.html: no such file or directory
exit status 1
```

The program fails due to this directory structure change. 

Runnig `make` in the root directory to regenerate everything, including function docs, also doesn't work:

```
~/src/github.com/cockroachdb/docs/generate$ cd ..
~/src/github.com/cockroachdb/docs$ make
make: Nothing to be done for `all'.
```

That, however, happens even without this PR's changes. Something we didn't catch with the versioning PR. 

Any ideas?

benesch:
Sorry for not being more clear! This is a necessary step towards making the diagram generator work with versioning. It's expected that it won't work until we move it to the cockroachdb/cockroach repo (#1546).

jseldess:
Sorry for misunderstanding this, @benesch. LGTM then. With this change in place, we won't be able to use the diagram generator until the rest of the work is done, though. Are you planning to move it over?

benesch:
Yup! Or talking @dt and @mjibson into it.

dt:
Moving the markdown generation is on my radar -- I hadn't looked at the diagram generator at all (though I imagine it probably is straightforward?)

PR #1650, opened 2017-06-30
Style guide updates

deronjohnson:
In progress: updating styling across the site to adhere to the style guide and match the hugo site's styling. Besides straightforward updates, I removed some outdated code and removed three of the callout styles that were currently unused, as the style guide prefers the three in use going forward. We haven't done a visual QA on our end yet, so there will be more changes coming.


deronjohnson:
@jseldess, I'm planning on merging the hugo style guide updates after you merge here. There's a dev project on the hugo repo that requires getting those changes integrated before I can start, so I was hoping we could get this merged by the end of the day.

jseldess:
@deronjohnson, the last commit doesn't seem to have actually change the toc bullet size. Did you mean to revert to 14px? Let me know what you intended and then I'll merge.

deronjohnson:
This was a change Andy asked me to make until we make a final decision on table of content type sizing, which was changing the size of the bullet itself. I had originally thought you wanted to reduce the size of the bullet, not the type, and Andy thought it should stay consistent. Removing the 14px rule makes the bullet the same size as the font (17px).

jseldess:
I see. OK. I'll just merge now. In my opinion, the type size and spacing is too large/spacious in comparison to the sidenav. But I don't want to block this.

PR #1477, opened 2017-05-30
Bring compilation down to ~4s

benesch:
See individual commits for details!

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1477)

PR #1652, opened 2017-06-30
fix mobile breadcrumbs

benesch:
I broke this in #1477. Guess it wasn't "dead code." 🤦‍♂️

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -55,6 +55,10 @@
 
         <ul id="mysidebar" class="nav {% if page.url == '/' or page.url == '/search.html' %}nav--home{% endif %}" style="display: none">
             <div class="sidenav-arrow"><img src="{{ 'images/sidenav-arrow.svg' | relative_url }}" alt="Sidenav arrow"></div>
+            <div class="collapsed-header">
+              <div id="ch-breadcrumbs" class="collapsed-header__pre"></div>
+              <div id="ch-title"></div>
+            </div>
             <li class="search-wrap">
                 <div class="search">
                     <span class="fa fa-search"></span>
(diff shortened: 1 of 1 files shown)


benesch:
(Here's the original implementation: https://github.com/benesch/docs/blob/9f9757a8b825c9d6575a84e8d9e95798ac38a28a/_includes/sidebar.html#L54-L77)

_includes/sidebar.html
```diff
@@ -89,32 +93,38 @@
...
                     if (!items || items.length == 0)
                         return $();
 
                     var lis = items.map(function (item) {
                       var urls = (item.urls || []).map(function (url) {
                           return baseUrl + url.replace("${VERSION}", pageVersion);
                       });
-                      var subitems = renderItems(item.items, tier + 1);
+                      var active = (urls.indexOf(location.pathname) !== -1);
+                      if (active) {
```
justinj: I'm not personally a huge fan of doing mutation within a `map`, maybe it would be clearer to change it to a `forEach` and append to a list at the end? Or even just extract this block into a function (if there's an easy way to do that)?

_includes/sidebar.html
```diff
@@ -89,32 +93,38 @@
...
+                        $("#ch-breadcrumbs")
```
justinj: I think I see what's happening here, you want to eventually overwrite with the value that occurs at the deepest point in the recursion? After figuring this out I think this logic is subtle enough that it could do with a comment explaining what's happening here, or maybe even pull this whole thing out into a separate function?

benesch:
All good points, but going to address in a follow-up PR to get this merged!

kuanluo:
Thank you, @benesch for the quick fix! 👍

benesch:
Heh, sorry for breaking it in the first place.

PR #1021, opened 2017-01-23
Proposal for a new sidebar organisation for sql docs.

knz:
This PR forks off #1008.

This is a proposal to initiate a discussion about the structure of the sidebar.

The motivation from this stems from the observation that the sidebar is hard to navigate in the "develop" section and fails to identify the hierarchical nature of how developers approach documentation about a new programming environment.

The proposal approaches this by defining the following top sections:

- Writing SQL queries (About the SQL language as supported by CockroachDB)
- Using transactions (About how the SQL queries are handled server-side)
- Data definition (About how to organize data)
- Misc: Security ("Privileges" in proposal) and Client Sessions (not yet written, but will need to be)

It further separates the language features that belong to each of these categories. The classification is "clean" in that it partitions the entire set of SQL statements along these lines without overlap (with one exception, "CREATE TABLE AS" which is really "CREATE TABLE" combined with "INSERT" and is a non-standard extension).

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/1021)

PR #1657, opened 2017-07-04
fix mobile breadcrumbs, round two

benesch:
On mobile, properly render "Docs Menu" in the collapsed header on pages without a sidebar entry, like the homepage and search bar. This addresses additional breakage from #1477 that was not addressed by #1652.

Also address review feedback from #1652 by adding some documenting comments.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -53,12 +53,9 @@
             });
         </script>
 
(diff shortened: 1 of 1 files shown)


benesch:
(FWIW I tried doing it in three loops, but it got a lot more complicated as
measured by lines of code, and we're probably throwing this sidebar away
eventually: #1021)

benesch:
TFTR!

PR #1665, opened 2017-07-05
Update GCE load balancing guidance

### LINKED ISSUE #1602: Deploy CockroachDB on Google Cloud Platform GCE Doc Bug
Re: [Deploy CockroachDB on Google Cloud Platform GCE](https://cockroachlabs.com/docs//deploy-cockroachdb-on-google-cloud-platform.html)

## Issue Description

This how-to, which includes instructions for running a secure cluster, instructs the reader to setup a Google Cloud TCP load balancer with the health checks for that LB pointing to port 8080 and the `/health` endpoint.

However, I don't believe this is possible to make work as the TCP (Network) load balancer that Google offers [only supports legacy HTTP health checks](https://cloud.google.com/compute/docs/load-balancing/health-checks) and if you make a request to the CRDB server with HTTP it redirects the client with a 308 Permanent Redirect to the HTTPS version of that /health check URL. This non-200 response causes the LB health check to fail and all CRDB hosts behind the load balancer are marked as unavailable.

Here is an example request on a CRDB running in secure mode, redirecting all requests to an `http` url to an `https` url.

```
$ curl -k -i http://0.0.0.0:8080/health
HTTP/1.1 308 Permanent Redirect
Location: https://0.0.0.0:8080/health
Date: Sat, 17 Jun 2017 04:32:44 GMT
Content-Length: 63
Content-Type: text/html; charset=utf-8

<a href="https://0.0.0.0:8080/health">Permanent Redirect</a>.

```
The outcome of this is that it appears to be impossible to use the Google TCP load balancer with CRDB, while also having the ability to take CRDB servers out of the rotation if they are no longer healthy.

## Suggested Resolution 

Stop CRDB from redirecting calls the `/health` to an `https` URL. Update the documentation to provide an alternative that works.

jseldess:
Fixes #1602

### DIFF
--- a/v1.0/deploy-cockroachdb-on-google-cloud-platform-insecure.md
+++ b/v1.0/deploy-cockroachdb-on-google-cloud-platform-insecure.md
@@ -10,7 +10,7 @@ toc_not_nested: true
   <button class="filter-button current"><strong>Insecure</strong></button></a>
 </div>
 
-This page shows you how to manually deploy an insecure multi-node CockroachDB cluster on Google Cloud Platform's Compute Engine (GCE), using Google's managed load balancing service to distribute client traffic.
+This page shows you how to manually deploy an insecure multi-node CockroachDB cluster on Google Cloud Platform's Compute Engine (GCE), using Google's TCP Proxy Load Balancing service to distribute client traffic.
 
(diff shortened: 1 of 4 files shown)


mberhault:
LGTM. Thanks for working through this, it should really help people.

PR #1670, opened 2017-07-06
update v1.1 build from source instructions for autotools build dep

benesch:
cockroachdb/cockroach#16859 added a build dependency on the autotools suite. Update the v1.1 build from source instructions to match CockroachDB's CONTRIBUTING.md.

### DIFF
--- a/v1.1/install-cockroachdb.md
+++ b/v1.1/install-cockroachdb.md
@@ -196,6 +196,10 @@ $(document).ready(function(){
         <td>CMake</td>
         <td>Versions 3.81+ are known to work.</td>
       </tr>
+      <tr>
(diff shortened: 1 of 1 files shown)


benesch:
Definitely helpful! Turns out we're about to land a change to remove the dependency on automake/libtool, though, so holding off until the dust clears on that PR.

benesch:
Ok, updated this PR to reflect that we only have a dependency on autoconf, not automake or libtool.

Holding off on adding full-on install instructions per platform, since we probably want both `apt-get` (Debian/Ubuntu/friends) and `yum` (CentOS/Fedora/friends) instructions.

jseldess:
Thanks for the update, @benesch.

PR #1680, opened 2017-07-06
Specify hostname on swarm services so that UI can work

### LINKED ISSUE #1664: Admin UI not reachable in a secure docker swarm cluster
As reported in https://forum.cockroachlabs.com/t/accessing-admin-ui-on-docker-swarm/770

Depending on how that thread gets resolved, we should either add in a `--hostname` flag to our docker swarm commands to make the UI work or note in the docs that it doesn't currently as a limitation.

a-robinson:
In the process, get rid of the advertise-host flag, since it's pointless
to set it to the same value as the container's hostname.

Also, for future reference, is there an easier way to update both the 1.0 and 1.1 docs for something than to just manually edit both versions?

Fixes #1664


sploiselle:
@a-robinson Unfortunately cross-version changes have to be done manually.

This LGTM, though.

PR #1691, opened 2017-07-10
Rework node cert workflow in cert and deployment docs

### LINKED ISSUE #1678: Create Security Certificates workflow adjustments
Re: [Create Security Certificates](https://cockroachlabs.com/docs//v1.0/create-security-certificates.html)

## Issue Description

overwriting a node certificate then uploading them to another nodes is somewhat confusing in terms of workflow. 

## Suggested Resolution 

in the CREATE THE CA CERTIFICATE AND KEY PAIR:
create and then copy the following to all of the nodes first
safe-directory/ca.key 
certs/client.root.crt
certs/client.root.key
certs/ca.crt

in CREATE THE CERTIFICATE AND KEY PAIRS FOR NODES:
replace 5 steps with the single step below

cockroach cert create-node \
node2.example.com \
node2.another-example.com \
--certs-dir=certs \
--ca-key=my-safe-directory/ca.key \

jseldess:
After generating and uploading certs to a node, remove
the node cert and key before generating those for the next
node. This is instead of using our custom `--overwrite` flag,
though I still mention the flag as an alternative.

Fixes #1678

### DIFF
--- a/v1.0/create-security-certificates.md
+++ b/v1.0/create-security-certificates.md
@@ -96,11 +96,14 @@ If you need to troubleshoot this command's behavior, you can change its [logging
 
 ### Create the CA certificate and key pair
 
-1. Create a safe directory for the CA key:
+1. Create two directories:
 
     ~~~ shell
+    $ mkdir certs
     $ mkdir my-safe-directory
     ~~~
+    - `certs`: You'll generate your CA certificate and all node and client certificates and keys in this directory and then upload some of the files to your nodes.
+    - `my-safe-directory`: You'll generate your CA key in this directory and then reference the key when generating node and client certificates. After that, you'll keep the key safe and secret; you will not upload it to your nodes.
 
 2. Generate the CA certificate and key:
 
@@ -113,12 +116,13 @@ If you need to troubleshoot this command's behavior, you can change its [logging
     ~~~
 
     ~~~
-    -rw-r--r--  1 maxroach  maxroach  1.1K Apr 19 09:35 ca.crt
+    total 8
+    -rw-r--r--  1 maxroach  maxroach  1.1K Jul 10 14:12 ca.crt
     ~~~
 
 ### Create the certificate and key pairs for nodes
 
-1. Create the certificate and key for the first node:
+1. Generate the certificate and key for the first node:
 
     ~~~ shell
     $ cockroach cert create-node \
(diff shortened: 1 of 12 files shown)


v1.0/create-security-certificates.md
```diff
@@ -131,54 +153,48 @@ If you need to troubleshoot this command's behavior, you can change its [logging
...
+    -rw-------  1 maxroach  maxroach  1.6K Jul 10 14:16 node.key
     ~~~
 
 2. Upload the files to the first node.
 
-3. Create the certificate and key for the second node, using the `--overwrite` flag to replace the files created for the first node:
+3. Delete the local copy of the first node's certificate and key:
+
+    ~~~ shell
+    $ rm -rf certs/node.crt certs/node.key
```
mberhault: you can drop the `-rf`. We're definitely not deleting directories (`-r`) and the permissions set by `cockroach cert` means we don't need to force the delete (`-f`).

v1.0/manual-deployment.md
```diff
@@ -86,19 +94,29 @@ Locally, you'll need to [create the following certificates and keys](create-secu
 	# Create the certs directory:
 	$ ssh <username>@<node1 address> "mkdir certs"
 
-	# Upload the CA certificate, client (root) certificate and key, and node certificate and key:
+	# Upload the CA certificate, client certificates and keys, and node certificate and key:
```
mberhault: We don't need to copy the client certificate and key to the node. I can understand `root` being useful for local debugging, but other users really don't need to be there.

v1.0/create-security-certificates.md
```diff
@@ -113,12 +116,31 @@ If you need to troubleshoot this command's behavior, you can change its [logging
     ~~~
 
     ~~~
-    -rw-r--r--  1 maxroach  maxroach  1.1K Apr 19 09:35 ca.crt
+    total 8
+    -rw-r--r--  1 maxroach  maxroach  1.1K Jul 10 14:12 ca.crt
     ~~~
 
+### Create the certificate and key pair for a client
```
mberhault: This was fine as the last step, keeping node and client certs separate makes things a bit clearer.

jseldess: Done.

v1.0/create-security-certificates.md
jseldess: Done.

v1.0/manual-deployment.md
jseldess: Done.

v1.0/create-security-certificates.md
```diff
@@ -96,106 +96,144 @@ If you need to troubleshoot this command's behavior, you can change its [logging
...
-    -rw-r--r--  1 maxroach  maxroach  1.1K Apr 19 09:35 ca.crt
-    -rw-r--r--  1 maxroach  maxroach  1.2K Apr 19 09:36 node.crt
-    -rw-------  1 maxroach  maxroach  1.6K Apr 19 09:36 node.key
+    total 24
+    -rw-r--r--  1 maxroach  maxroach  1.1K Jul 10 14:12 ca.crt
+    -rw-r--r--  1 maxroach  maxroach  1.2K Jul 10 14:16 node.crt
+    -rw-------  1 maxroach  maxroach  1.6K Jul 10 14:16 node.key
     ~~~
 
 2. Upload the files to the first node.
```
mberhault: uh. both `upload the files to the Nth node` steps are empty in this doc. Is it because this is an overview of how certs work? It seems odd though.

bdarnell:
Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v1.0/manual-deployment.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1691#-KogXyR48Dgyfhfahe-1-r1-97:-KoiV9nGuWesbUqdWlJH:bitj7u8) ([raw file](https://github.com/cockroachdb/docs/blob/6ba70db1592924854fe3c9c9fac99148faf582d7/v1.0/manual-deployment.md#L97)):*

Done.
</blockquote></details>

We may want to mention that copying the root cert is optional (and I'd consider it a bad practice, although it's not technically more of a security risk than the node certificates that are already there)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1691)*

jseldess:
*[v1.0/manual-deployment.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1691#-KogXyR48Dgyfhfahe-1-r1-97:-KolnY6dUOy--3avpvUk:b-o7rizt) ([raw file](https://github.com/cockroachdb/docs/blob/6ba70db1592924854fe3c9c9fac99148faf582d7/v1.0/manual-deployment.md#L97)):*

We may want to mention that copying the root cert is optional (and I'd consider it a bad practice, although it's not technically more of a security risk than the node certificates that are already there)
</blockquote></details>

Since this is considered bad practice, I updated all deployment tutorials to not copy the root user's cert/key to nodes and added a note about doing that if node-local debugging is needed. 

Also updated the "test a cluster" instructions to start the sql client locally instead of on nodes.

Hope these changes are acceptable, @bdarnell. Please let me know.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1691)*

jseldess:
*[v1.0/create-security-certificates.md, line 157 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1691#-KogXyR48Dgyfhfahdzx-r3-157:-Kolo-rj2mzjEBU-ZxzQ:b-13swmu) ([raw file](https://github.com/cockroachdb/docs/blob/81ff4815eb109985874bc12b1e02b3c5d4601556/v1.0/create-security-certificates.md#L157)):*

uh. both `upload the files to the Nth node` steps are empty in this doc. Is it because this is an overview of how certs work? It seems odd though.
</blockquote></details>

Added those details

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1691)*

bdarnell:
Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v1.0/manual-deployment.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1691#-KogXyR48Dgyfhfahe-1-r1-97:-KomGQMO5FsIeoHjQRD9:b49opeb) ([raw file](https://github.com/cockroachdb/docs/blob/6ba70db1592924854fe3c9c9fac99148faf582d7/v1.0/manual-deployment.md#L97)):*

Since this is considered bad practice, I updated all deployment tutorials to not copy the root user's cert/key to nodes and added a note about doing that if node-local debugging is needed. 

Also updated the "test a cluster" instructions to start the sql client locally instead of on nodes.

Hope these changes are acceptable, @bdarnell. Please let me know.
</blockquote></details>

LGTM (but does this break any of our other instructions? If we're relying on ssh'ing to nodes and running cockroach commands then we may want to leave it in for now)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1691)*

jseldess:
*[v1.0/manual-deployment.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1691#-KogXyR48Dgyfhfahe-1-r1-97:-KomHw8vBHGyigTITg1p:bvj7qxl) ([raw file](https://github.com/cockroachdb/docs/blob/6ba70db1592924854fe3c9c9fac99148faf582d7/v1.0/manual-deployment.md#L97)):*

LGTM (but does this break any of our other instructions? If we're relying on ssh'ing to nodes and running cockroach commands then we may want to leave it in for now)
</blockquote></details>

Ah, there is still one place in the manual deployment guide where the root cert is needed on the server: When running `cockroach gen haproxy` to generate the haproxy config. I'll revise the instructions to generate that file locally and copy it to the haproxy server instead. 

Otherwise, I'm pretty sure all cockroach client commands are run locally, not on nodes, in these guides.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1691)*

PR #1697, opened 2017-07-11
Ensure font fall-backs for code samples

knz:
Hi @deronjohnson please check this PR out. This change is only for the docs, but the blog is affected as well. Can you ensure that the `pre` and `code` classes have fall-backs beyond Monaco please? Right now all our width-sensitive code samples are broken on non-OSX systems (Monaco is a proprietary Apple font)

### DIFF
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -145,7 +145,7 @@ pre {
 
 pre, code {
   background-color: #F3F3F3;
-  font-family: 'Monaco';
+  font-family: Monaco, Menlo, Consolas, "Courier New", monospace;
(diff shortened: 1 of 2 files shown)


kuanluo:
@knz Thank you for noticing/fixing!

jseldess:
Yes, thanks very much for the catch, @knz! We need to have people not on Macs testing involved in testing in the future, clearly. 

@deronjohnson, if you're good with this change, I'll merge.

deronjohnson:
@jseldess, absolutely. Thanks for catching @knz!

PR #1702, opened 2017-07-12
Cross-cloud deployment & migration screencast

jseldess:
I recorded a live demo of deploying across Digital Ocean and GCE and then migrating from DO to GCE, as a complement to the tutorial on this topic. 

This PR embeds the recording (on Youtub) in the tutorial and renames the broadens the name of the tutorial from "Automatic Cloud Migration" to "Cross-Cloud Deployment & Migration".

### DIFF
--- a/_includes/sidebar-data-v1.0.json
+++ b/_includes/sidebar-data-v1.0.json
@@ -135,7 +135,7 @@
             ]
           },
           {
-            "title": "Automatic Cloud Migration",
+            "title": "Cross-Cloud Deployment & Migration",
             "urls": [
               "/${VERSION}/demo-automatic-cloud-migration.html"
             ]
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -135,7 +135,7 @@
             ]
           },
           {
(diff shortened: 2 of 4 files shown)


jseldess:
@nstewart, @spencerkimball, @robert-s-lee, @bdarnell: I'd love your feedback on the screencast. As you'll see, this isn't very polished or scripted. I kept it casual but feel it adds plenty of value. If you agree, we should do more of these, as they're relatively easy to make. 

To review, either click the link above and go to the "Watch a Live Demo" section. Or you can just go directly to the video on Youtube: https://www.youtube.com/watch?v=eoSfqobp3jc

spencerkimball:
@jseldess it's awesome! Definitely worth putting up and in fact we should do a short blog post which references it.

I think it would be very worthwhile to add slightly to the demo by changing the constraints so things balance all the way back to just digital ocean, so you get to see the replicas all rebalance in the other direction. You should also make the point in the video that while rebalancing is happening, the YCSB load continues to run; it does not have to weather an outage during the migration, nor are any active clients interrupted or otherwise made aware that the migration is happening.

jseldess:
Thanks for the suggestions, @spencerkimball. I'll do another take today.

PR #1708, opened 2017-07-13
Remove research banner from Docs home

jseldess:


deronjohnson:
@jseldess, not sure if you noticed or not, but if you want this banner to show up without the "note", you can wrap it in `<div class="bs-callout bs-callout--info"></div>` instead of `{{site.data.alerts.callout_info}}`. The callouts are automatically injecting "note", "tip", and "warning" now.

jseldess:
Thanks for that tip, @deronjohnson!

PR #1714, opened 2017-07-17
v1.1 SQL shell updates

### LINKED ISSUE #1633: cli/sql: report the cluster ID and server version upon new connections
PR: https://github.com/cockroachdb/cockroach/pull/16703

jseldess:
- Add SQL shell welcome details, calling out version and cluster ID details
- Add `node_id` session variable to `show-vars.md` and deployment tutorials

Fixes #1633

### DIFF
--- a/v1.1/deploy-cockroachdb-on-aws-insecure.md
+++ b/v1.1/deploy-cockroachdb-on-aws-insecure.md
@@ -250,11 +250,13 @@ To test this, install CockroachDB locally and use the [built-in SQL client](use-
 
 	As you can see, the load balancer redirected the query to one of the CockroachDB nodes.
 
-4. Check which node you were redirected to:
+4. Use the `node_id` [session variable](show-vars.html) to check which node you were redirected to:
 
-	~~~ sql
-	> SELECT node_id FROM crdb_internal.node_build_info LIMIT 1;
-	~~~
+    {% include copy-clipboard.html %}
+    ~~~ sql
+    > SHOW node_id;
+    ~~~
+
 	~~~
 	+---------+
 	| node_id |
--- a/v1.1/deploy-cockroachdb-on-aws.md
+++ b/v1.1/deploy-cockroachdb-on-aws.md
@@ -404,11 +404,11 @@ To test this, use the [built-in SQL client](use-the-built-in-sql-client.html) lo
 
     As you can see, the load balancer redirected the query to one of the CockroachDB nodes.
 
-3. Check which node you were redirected to:
+3. Use the `node_id` [session variable](show-vars.html) to check which node you were redirected to:
 
     {% include copy-clipboard.html %}
     ~~~ sql
-    > SELECT node_id FROM crdb_internal.node_build_info LIMIT 1;
+    > SHOW node_id;
     ~~~
 
     ~~~
--- a/v1.1/deploy-cockroachdb-on-digital-ocean-insecure.md
+++ b/v1.1/deploy-cockroachdb-on-digital-ocean-insecure.md
@@ -230,11 +230,13 @@ To test this, install CockroachDB locally and use the [built-in SQL client](use-
 
 	As you can see, the load balancer redirected the query to one of the CockroachDB nodes.
 
-4. Check which node you were redirected to:
+4. Use the `node_id` [session variable](show-vars.html) to check which node you were redirected to:
 
-	~~~ sql
-	> SELECT node_id FROM crdb_internal.node_build_info LIMIT 1;
-	~~~
+    {% include copy-clipboard.html %}
+    ~~~ sql
+    > SHOW node_id;
+    ~~~
+
 	~~~
 	+---------+
 	| node_id |
--- a/v1.1/deploy-cockroachdb-on-digital-ocean.md
+++ b/v1.1/deploy-cockroachdb-on-digital-ocean.md
@@ -379,11 +379,11 @@ To test this, use the [built-in SQL client](use-the-built-in-sql-client.html) lo
 
     As you can see, the load balancer redirected the query to one of the CockroachDB nodes.
 
-3. Check which node you were redirected to:
+3. Use the `node_id` [session variable](show-vars.html) to check which node you were redirected to:
 
     {% include copy-clipboard.html %}
     ~~~ sql
-    > SELECT node_id FROM crdb_internal.node_build_info LIMIT 1;
+    > SHOW node_id;
     ~~~
 
     ~~~
--- a/v1.1/set-vars.md
+++ b/v1.1/set-vars.md
@@ -32,17 +32,17 @@ The variable name is case insensitive. The value can be a list of one or more it
 
 ### Supported Variables
 
-| Variable name                   | Description  | Initial value | Can be viewed with [`SHOW`](show-vars.html)? |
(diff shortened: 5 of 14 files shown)


v1.1/deploy-cockroachdb-on-aws-insecure.md
```diff
@@ -264,6 +266,13 @@ To test this, install CockroachDB locally and use the [built-in SQL client](use-
 	(1 row)
 	~~~
 
+    The `node_id` [session variable](show-vars.html) used above is an alias for the following query, which you can use as well:
```
knz: This is not true any more. 

Also, as a general principle, I would not recommend / advertise `crdb_internal` in user-facing docs. It's called "internal" for a reason :)
If you find over time that some features available via `crdb_internal` tables happen to also be useful to users, let the eng team know so we can devise a properly advertisable feature outside of the "internal" namespace.

v1.1/deploy-cockroachdb-on-aws.md
```diff
@@ -420,6 +420,13 @@ To test this, use the [built-in SQL client](use-the-built-in-sql-client.html) lo
     (1 row)
     ~~~
 
+    The `node_id` [session variable](show-vars.html) used above is an alias for the following query, which you can use as well:
```
knz: ditto

v1.1/deploy-cockroachdb-on-digital-ocean-insecure.md
```diff
@@ -244,6 +246,13 @@ To test this, install CockroachDB locally and use the [built-in SQL client](use-
 	(1 row)
 	~~~
 
+    The `node_id` [session variable](show-vars.html) used above is an alias for the following query, which you can use as well:
```
knz: ditto

v1.1/deploy-cockroachdb-on-digital-ocean.md
```diff
@@ -395,6 +395,13 @@ To test this, use the [built-in SQL client](use-the-built-in-sql-client.html) lo
     (1 row)
     ~~~
 
+    The `node_id` [session variable](show-vars.html) used above is an alias for the following query, which you can use as well:
```
knz: ditto

v1.1/set-vars.md
```diff
@@ -32,17 +32,17 @@ The variable name is case insensitive. The value can be a list of one or more it
...
-| `standard_conforming_strings`   | Ignored; recognized for compatibility with PostgreSQL clients. | N/A                                                                                                                                                                                                            | No  |
-| `extra_float_digits`            | Ignored; recognized for compatibility with PostgreSQL clients. | N/A                                                                                                           | No  |
+| Variable name | Description  | Initial value | Can be viewed with [`SHOW`](show-vars.html)? |
+|---------------|--------------|---------------|----------------------------------------------|
+| `application_name` | The current application name for statistics collection. | Empty string | Yes |
+| `client_encoding` | Ignored; recognized for compatibility with PostgreSQL clients. Only possible value is "`UTF8`". | N/A | No |
```
knz: I thought it was better for UX to place the "ignored / compatibility" values at the bottom of the table. For users seeking "useful" values they are just noise.

v1.1/show-vars.md
```diff
@@ -28,22 +28,24 @@ It may be enclosed in double quotes; this is useful if the variable name itself
...
-| `server_version`                | The version of PostgreSQL that CockroachDB emulates. | Version-dependent.                                                                                                                                                         | No |
-| `client_min_messages`           | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
-| `client_encoding`               | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
-| `extra_float_digits`            | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
-| `max_index_keys`                | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
-| `standard_conforming_strings`   | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
+| Variable name | Description | Initial value |  Can be modified with [`SET`](set-vars.html)? |
+|---------------|-------------|---------------|-----------------------------------------------|
+| `application_name` | The current application name for statistics collection. | Empty string | Yes |
+| `client_encoding` | (Reserved; exposed only for ORM compatibility.) | (Reserved) | No |
```
knz: ditto

v1.1/deploy-cockroachdb-on-aws-insecure.md
```diff
@@ -252,9 +252,11 @@ To test this, install CockroachDB locally and use the [built-in SQL client](use-
 
 4. Check which node you were redirected to:
```
knz: I think you can write this entire section just once in e.g. `somesteps.md` then reuse it in each of the DC tutorials via `{% include %}`. If that is not possible, let's investigate together how to make it possible. This will save editing work in the future!

v1.1/set-vars.md
jseldess: I agree with you in principle, but in a straight-up reference table like this, I think it's also user-friendly to list the variables in the order in which they see them in the `SHOW ALL` output, which happens to be alphabetical. As variables are added, it's also easier to update the docs.  

An alternative would be to split this into an alphabetical list of "valuable" variables, and then a follow-on table for the compatibility variables. Let me know if you'd prefer that approach.

v1.1/deploy-cockroachdb-on-aws-insecure.md
jseldess: Done.

v1.1/deploy-cockroachdb-on-aws.md
jseldess: Done.

v1.1/deploy-cockroachdb-on-digital-ocean-insecure.md
jseldess: Done.

v1.1/deploy-cockroachdb-on-digital-ocean.md
jseldess: Done.

v1.1/deploy-cockroachdb-on-aws-insecure.md
jseldess: Totally agree that we need to use includes throughout these tutorials. I'd prefer to handle that in a separate PR, though. There are many minor differences between cloud deployment guides that need to be considered when breaking them up into includes.

v1.1/set-vars.md
knz: Yes the latter approach would appeal more to me as a reader.

jseldess:
@knz, @benesch, we already know how to re-use markdown, images, etc. across files: https://github.com/cockroachdb/docs/blob/master/CONTRIBUTING.md#auto-included-content. We do in already in many places. I opened a separate issue to refactor these tutorials with includes: https://github.com/cockroachdb/docs/issues/1718

v1.1/set-vars.md
jseldess: OK, in the end, I've decided to just go with your initial approach of putting the compatability variables at the bottom of the list. In addition, I've alphabetized the "valuable" variables above them. Thanks, @knz!

knz:
Thanks.

LGTM

PR #1723, opened 2017-07-17
[WIP] Multi-Active Availability Feature Description

sploiselle:
Adding content that describes what MAA is/does, as well as a light treatment of some of the technologies we use to enable the features for customers.

@bdarnell Can I get you to review from a technical standpoint? The intended audience is not as highly technical as the architecture docs, so I think we can speak mostly to larger strokes of the software's functionality (e.g. Raft).

### DIFF
--- a/v1.0/multi-active-availability.md
+++ b/v1.0/multi-active-availability.md
@@ -1,0 +1,67 @@
+---
+title: Multi-Active Availability
+summary: Learn about CockroachDB's high availability model, known as Multi-Active Availability.
+toc: false
+---
+
+CockroachDB's availability model is described as "Multi-Active Availability." In essence, multi-active availability provides benefits similar to traditional notions of high availability, but also lets you read and write from every node in your cluster without generating any conflicts.
+
+<div id="toc"></div>
+
+## What is High Availability?
+
+High availability lets an application continue running even if a system hosting one of its services fails. This is achieved by scaling the application's services horizontally, i.e., replicating the service across many machines or systems. If any one of them fails, the others can simply step in and perform the same service.
+
+Before diving into the details of CockroachDB's multi-active availability, we'll review the two most common high availability designs: [Active-Passive](#active-passive) and [Active-Active](#active-active) systems.
+
+### Active-Passive
+
+In active-passive systems, all traffic is routed to a single, "active" replica. Changes to the replica's state are then copied to a backup "passive" replica, in an attempt to always mirror the active replica as closely as possible.
+
+However, this design has downsides:
+
+- If you use asynchronous replication, you cannot guarantee that any data is ever successfully replicated to passive followers––meaning you can easily lose data. Depending on your industry, this could have pretty dire consequences.
+- If you use synchronous replication and any passive replicas fail, you have to either sacrifice availability for the entire application or risk inconsistencies.
+
+### Active-Active
+
+In active-active systems, multiple replicas run identical services, and traffic is routed to all of them. If any replica fails, the others simply handle the traffic that would've been routed to it.
+
+For databases, though, active-active replication is incredibly difficult to instrument for most workloads. For example, if you let multiple replicas handle writes for the same keys, how do you keep them consistent?
+
+#### Example: Conflicts with Active-Active Replication
+
+For this example, we have 2 replicas (**A**, **B**) in an active-active high availability cluster.
+
+1. **A** receives a write for key `xyz` of `'123'`, and then immediately fails.
+2. **B** receives a read of key `xyz`, and returns a `NULL` because it cannot find the key.
+3. **B** then receives a write for key `xyz` of `'456'`.
+4. **A** is restarted and attempts to rejoin **B**––but what do you about key `xyz`? There's an inconsistency in the system without a clear way to resolve it.
+
+{{site.data.alerts.callout_info}}In this example, the cluster remained active the entire time. But in terms of the <a href="https://en.wikipedia.org/wiki/CAP_theorem">CAP theorem</a>, this is an AP system; it favored being available instead of consistent when partitions occur.{{site.data.alerts.end}}
+
+## What is Multi-Active Availability?
+
+Multi-active availability is CockroachDB's version of high availability (keeping your application online in the face of partial failures), which we've designed to avoid the downsides of both active-passive and traditional active-active systems.
+
+Like active-active designs, all replicas can handle traffic, including both reads and writes. However, CockroachDB improves upon that design by also ensuring that data remains consistent across them, which we achieve by using "consensus replication." In this design, replication requests are sent to at least 3 replicas, and are only considered committed when a majority of replicas acknowledge that they've received it. This means that you can still have failures without compromising availability.
+
+To prevent conflicts and guarantee your data's consistency, clusters that lose a majority of replicas stop responding because they've lost the ability to reach a consensus on the state of your data. When a majority of replicas are restarted, your database resumes operation.
+
+### Consistency Example
+
+For this example, we have 3 CockroachDB nodes (**A**, **B**, **C**) in a multi-active availability cluster.
+
+1. **A** receives a write on `xyz` of `'123'`. It communicates this write to nodes **B** and **C**, who confirm that they've received the write, as well. Once **A** receives the first confirmation, the change is committed.
+2. **A** fails.
+3. **B** receives a read of key `xyz`, and returns the result `'123'`.
+4. **C** then receives an update for key `xyz` to the values `'456'`. It communicates this write to node **B**, who confirms that its received the write, as well. After receiving the confirmation, the change is committed.
+5. **A** is restarted and rejoins the cluster. It receives an update that the key `xyz` had its value changed to `'456'`.
+
+{{site.data.alerts.callout_info}}In this example, if nodes <strong>B</strong> or <strong>C</strong> failed at any time, the cluster would have stopped responding. In terms of the <a href="https://en.wikipedia.org/wiki/CAP_theorem">CAP theorem</a>, this is an CP system; it favored being consistent instead of available when partitions occur.{{site.data.alerts.end}}
+
+## What's next?
+
+To get a greater understanding of how CockroachDB is a survivable system that enforces strong consistency, check out our [architecture documentation](/architecture/overview.html).
+
+To see Multi-Active Availability in action, see this [availability demo](demo-fault-tolerance-and-recovery.html).
--- a/_includes/sidebar-data-v1.0.json
+++ b/_includes/sidebar-data-v1.0.json
@@ -117,6 +117,12 @@
         "title": "Explore Core Features",
         "items": [
           {
+            "title": "Multi-Active Availability",
+            "urls": [
+              "/${VERSION}/multi-active-availability.html"
+            ]
+          },
+          {
             "title": "Data Replication",
             "urls": [
               "/${VERSION}/demo-data-replication.html"
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -1297,6 +1297,12 @@
         "title": "CockroachDB Features",
         "items": [
           {
+            "title": "Multi-Active Availability",
+            "urls": [
+              "/${VERSION}/multi-active-availability.html"
+            ]
+          },
+          {
             "title": "Simplified Deployment",
             "urls": [
               "/${VERSION}/simplified-deployment.html"
--- a/_includes/sidebar-data-v2.0.json
+++ b/_includes/sidebar-data-v2.0.json
@@ -1338,6 +1338,12 @@
         "title": "CockroachDB Features",
         "items": [
           {
+            "title": "Multi-Active Availability",
+            "urls": [
+              "/${VERSION}/multi-active-availability.html"
+            ]
+          },
+          {
             "title": "Simplified Deployment",
             "urls": [
               "/${VERSION}/simplified-deployment.html"
--- a/releases/v1.0-rc.1.md
+++ b/releases/v1.0-rc.1.md
@@ -53,7 +53,7 @@ Get future release notes emailed to you:
 
 - Distributed backup and restore capabilities are now available to [enterprise-license customers](https://www.cockroachlabs.com/cockroachdb-community-license/). Using the new [`BACKUP`](../v1.0/backup.html) SQL statement, a cluster's schema and data can be backed up in full or incrementally as of a given timestamp, and can be stored on the platforms you're already using, including AWS S3, Google Cloud Storage, NFS, or HTTP storage. In the unlikely event of irreparable cluster failure, the new [`RESTORE`](../v1.0/restore.html) SQL statement can be used to restore a cluster from enterprise backups. Importantly, backup and restore operations are distributed across the nodes of a cluster, minimizing the impact on cluster performance.
 
-  For details about pricing, [contact Cockroach Labs](https://www.cockroachlabs.com/pricing/sales/).
+  For details about pricing, [contact Cockroach Labs](https://www.cockroachlabs.com/pricing/request-a-license/).
 
 ### SQL Language Changes
 
--- a/v1.1/enterprise-licensing.md
+++ b/v1.1/enterprise-licensing.md
@@ -21,7 +21,7 @@ Type | Description
 
 ## Request a Trial or an Enterprise License Key
 
-[Contact us](https://www.cockroachlabs.com/pricing/sales/) to request a trial or an enterprise license key.
+[Contact us](https://www.cockroachlabs.com/pricing/request-a-license/) to request a trial or an enterprise license key.
 
 ## Set the Trial or Enterprise License Key
 
--- a/v1.1/frequently-asked-questions.md
+++ b/v1.1/frequently-asked-questions.md
@@ -154,7 +154,7 @@ For more insight, see [CockroachDB in Comparison](cockroachdb-in-comparison.html
 
 Yes, although CockroachDB is unlikely to be a drop-in replacement at this time. Due to differences in available features and syntax, migrating data from these databases to CockroachDB involves some manual effort.
 
-As a first step, check our [SQL Feature Support](sql-feature-support.html) page against your application's high-level SQL requirements. If essential SQL features are missing, consider workarounds and/or reach out to us via [our forum](https://forum.cockroachlabs.com/) or [sales contact form](https://www.cockroachlabs.com/pricing/sales/).
+As a first step, check our [SQL Feature Support](sql-feature-support.html) page against your application's high-level SQL requirements. If essential SQL features are missing, consider workarounds and/or reach out to us via [our forum](https://forum.cockroachlabs.com/)  or <sales@cockroachlabs.com>.
 
 Once you're ready to migrate, we recommend [importing your data via CSV](import.html). The process may expose places where you need to make changes for compatability. When migrating from PostgreSQL, for example, be sure to check this list of [known differences for identical input](porting-postgres.html).
 
--- a/v1.1/multi-active-availability.md
+++ b/v1.1/multi-active-availability.md
@@ -1,0 +1,67 @@
+---
+title: Multi-Active Availability
+summary: Learn about CockroachDB's high availability model, known as Multi-Active Availability.
+toc: false
+---
+
+CockroachDB's availability model is described as "Multi-Active Availability." In essence, multi-active availability provides benefits similar to traditional notions of high availability, but also lets you read and write from every node in your cluster without generating any conflicts.
+
+<div id="toc"></div>
+
+## What is High Availability?
+
+High availability lets an application continue running even if a system hosting one of its services fails. This is achieved by scaling the application's services horizontally, i.e., replicating the service across many machines or systems. If any one of them fails, the others can simply step in and perform the same service.
+
+Before diving into the details of CockroachDB's multi-active availability, we'll review the two most common high availability designs: [Active-Passive](#active-passive) and [Active-Active](#active-active) systems.
+
+### Active-Passive
+
+In active-passive systems, all traffic is routed to a single, "active" replica. Changes to the replica's state are then copied to a backup "passive" replica, in an attempt to always mirror the active replica as closely as possible.
+
+However, this design has downsides:
+
+- If you use asynchronous replication, you cannot guarantee that any data is ever successfully replicated to passive followers––meaning you can easily lose data. Depending on your industry, this could have pretty dire consequences.
+- If you use synchronous replication and any passive replicas fail, you have to either sacrifice availability for the entire application or risk inconsistencies.
+
+### Active-Active
+
+In active-active systems, multiple replicas run identical services, and traffic is routed to all of them. If any replica fails, the others simply handle the traffic that would've been routed to it.
+
+For databases, though, active-active replication is incredibly difficult to instrument for most workloads. For example, if you let multiple replicas handle writes for the same keys, how do you keep them consistent?
+
+#### Example: Conflicts with Active-Active Replication
+
+For this example, we have 2 replicas (**A**, **B**) in an active-active high availability cluster.
+
+1. **A** receives a write for key `xyz` of `'123'`, and then immediately fails.
+2. **B** receives a read of key `xyz`, and returns a `NULL` because it cannot find the key.
+3. **B** then receives a write for key `xyz` of `'456'`.
+4. **A** is restarted and attempts to rejoin **B**––but what do you about key `xyz`? There's an inconsistency in the system without a clear way to resolve it.
+
+{{site.data.alerts.callout_info}}In this example, the cluster remained active the entire time. But in terms of the <a href="https://en.wikipedia.org/wiki/CAP_theorem">CAP theorem</a>, this is an AP system; it favored being available instead of consistent when partitions occur.{{site.data.alerts.end}}
+
+## What is Multi-Active Availability?
+
+Multi-active availability is CockroachDB's version of high availability (keeping your application online in the face of partial failures), which we've designed to avoid the downsides of both active-passive and traditional active-active systems.
+
+Like active-active designs, all replicas can handle traffic, including both reads and writes. However, CockroachDB improves upon that design by also ensuring that data remains consistent across them, which we achieve by using "consensus replication." In this design, replication requests are sent to at least 3 replicas, and are only considered committed when a majority of replicas acknowledge that they've received it. This means that you can still have failures without compromising availability.
+
+To prevent conflicts and guarantee your data's consistency, clusters that lose a majority of replicas stop responding because they've lost the ability to reach a consensus on the state of your data. When a majority of replicas are restarted, your database resumes operation.
+
+### Consistency Example
+
+For this example, we have 3 CockroachDB nodes (**A**, **B**, **C**) in a multi-active availability cluster.
+
+1. **A** receives a write on `xyz` of `'123'`. It communicates this write to nodes **B** and **C**, who confirm that they've received the write, as well. Once **A** receives the first confirmation, the change is committed.
+2. **A** fails.
+3. **B** receives a read of key `xyz`, and returns the result `'123'`.
+4. **C** then receives an update for key `xyz` to the values `'456'`. It communicates this write to node **B**, who confirms that its received the write, as well. After receiving the confirmation, the change is committed.
+5. **A** is restarted and rejoins the cluster. It receives an update that the key `xyz` had its value changed to `'456'`.
+
+{{site.data.alerts.callout_info}}In this example, if nodes <strong>B</strong> or <strong>C</strong> failed at any time, the cluster would have stopped responding. In terms of the <a href="https://en.wikipedia.org/wiki/CAP_theorem">CAP theorem</a>, this is an CP system; it favored being consistent instead of available when partitions occur.{{site.data.alerts.end}}
+
+## What's next?
+
+To get a greater understanding of how CockroachDB is a survivable system that enforces strong consistency, check out our [architecture documentation](/architecture/overview.html).
+
+To see Multi-Active Availability in action, see this [availability demo](demo-fault-tolerance-and-recovery.html).
--- a/v2.0/enterprise-licensing.md
+++ b/v2.0/enterprise-licensing.md
@@ -21,7 +21,7 @@ Type | Description
 
 ## Request a Trial or an Enterprise License Key
 
-[Contact us](https://www.cockroachlabs.com/pricing/sales/) to request a trial or an enterprise license key.
+[Contact us](https://www.cockroachlabs.com/pricing/request-a-license/) to request a trial or an enterprise license key.
 
 ## Set the Trial or Enterprise License Key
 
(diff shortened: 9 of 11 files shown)


bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v1.1/multi-active-availability.md, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL82f3xXvT4ksEno2o:-KpL82f3xXvT4ksEno2p:bu493fz) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L13)):*
> ```Markdown
> ## What is High Availability?
> 
> High availability is a common feature that lets an application continue running even if a machine running one of its services fails. This is achieved by scaling the application's services horizontally, i.e., running the same service on many machines. If any one of the machines fails, the other machines in your deployment can simply perform the same tasks that the downed machine was.
> ```

Use "replicating" instead of "scaling".

There are two components to high availability: the availability of spare capacity ready to take over for failing nodes, and an *automated* process for doing so. The latter part is worth a mention here.

---

*[v1.1/multi-active-availability.md, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL92UL0LN6QfltyPM-:-KpL92UL0LN6QfltyPM0:b-q43zsw) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L17)):*
> ```Markdown
> ### Active-Active Systems
> 
> Though there are many ways to instrument high availability, the most common strategy is known as "Active-Active." In this setup, multiple machines run identical services and, in the case one fails, the other machines can handle the traffic.
> ```

I think active-passive systems are more common, at least for SQL databases. What distinguishes active-active mode is that all nodes are processing traffic even when there are no failures. 

Active-passive setups are desirable because they exhibit strong consistency "most of the time", only risking inconsistency during failovers. 

---

*[v1.1/multi-active-availability.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL9nWDpD0K8_menzuP:-KpL9nWDpD0K8_menzuQ:bpzvdmb) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L21)):*
> ```Markdown
> However, creating consistency between machines poses a very real difficulty. For databases, this is a particularly pressing problem to solve because of its inherently stateful nature, which greatly benefits from consistency across every machine running the database. 
> 
> This difficulty's most commonly experienced when bringing a node back online from a failover state; often, there will be conflicts in the data that must be mediated, which are either difficult to reason about or not an efficient use of an operator's time.
> ```

"Failover" is an active-passive concept; it doesn't apply in active-active systems. In active-passive systems, I've never seen anyone actually try to resolve potential conflicts from the failed node. They just wipe it and set it up as a new replica.

In active-active systems, either the system provides strong consistency at all times, or conflicts are a potential issue at all times (not just during failures). Either way, operators don't really enter in to it; an automated rule applies to "resolve" the conflict (generally something simple like "last writer wins")

---

*[v1.1/multi-active-availability.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpLBGl85SPwuayGpgvw:-KpLBGl85SPwuayGpgvx:bo6rjrw) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L40)):*
> ```Markdown
> ## What is Multi-Active Availability?
> 
> Multi-Active Availability provides all of the benefits of high availability (remaining live despite individual machine failures), while overcoming some of the difficulties or downsides of Active-Active systems.
> ```

Multi-active is not something fundamentally different from active-active. It's just active-active for more than two nodes. That's significant because it *enables* consensus-based replication, but you could still have a multi-active inconsistent system.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723)*

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v1.1/multi-active-availability.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpLBGl85SPwuayGpgvw:-KpMOB3D19nx5YsCL9wJ:b-exvtac) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L40)):*

Multi-active is not something fundamentally different from active-active. It's just active-active for more than two nodes. That's significant because it *enables* consensus-based replication, but you could still have a multi-active inconsistent system.
</blockquote></details>

On further reflection, the leap from two to "multi" is a bigger deal than I was giving it credit for. We see a lot of people wanting to run cockroachdb in two datacenters, but that won't actually give them the ability to survive a failure of either datacenter. So I think we do need to drive this point home, that even though MAA is "just" active-active for 3+ replicas, that third replica changes the capabilities of the system.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723)*

jseldess:
@sploiselle, thanks for writing this! Will add a lot of value. I know this is in early review with Ben, so just let me know when you want me to take a close look. My only high-level suggestions:

- As Ben suggests (I think), it'd be great to cover both Active-Passive and Active-Active.
- Diagrams (or animated gifs) would be very effective here. Robert/Kuan could help, I'm sure.

sploiselle:
@bdarnell Could I get you to review again for technical accuracy?

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[v1.1/multi-active-availability.md, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL82f3xXvT4ksEno2o:-KtI5YGGan2UTHgH0Hdc:btvf0cf) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L13)):*

Use "replicating" instead of "scaling".

There are two components to high availability: the availability of spare capacity ready to take over for failing nodes, and an *automated* process for doing so. The latter part is worth a mention here.
</blockquote></details>

"scaling" still needs to be "replicating" here. 

---

*[v1.1/multi-active-availability.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL9nWDpD0K8_menzuP:-KtI7QagsxQ6bqzyRg93:bg6q92w) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L21)):*

"Failover" is an active-passive concept; it doesn't apply in active-active systems. In active-passive systems, I've never seen anyone actually try to resolve potential conflicts from the failed node. They just wipe it and set it up as a new replica.

In active-active systems, either the system provides strong consistency at all times, or conflicts are a potential issue at all times (not just during failures). Either way, operators don't really enter in to it; an automated rule applies to "resolve" the conflict (generally something simple like "last writer wins")
</blockquote></details>

Active-active systems must remain below 50% utilization in the absence of failure, so it's not really any better than active-passive for hardware costs (it just looks better to see 40% and 40% utilization instead of 80% and 0%). 

Active-active systems can be CAP-consistent, in which case they do solve the problem of losing data when machines fail (but then they aren't really HA)
CAP-Available active-active systems can often also experience conflicts or inconsistencies even when there is no failure. 

---

*[v1.1/multi-active-availability.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpLBGl85SPwuayGpgvw:-KtID46rl8LS-yi3dPoL:bvjezi5) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L40)):*

On further reflection, the leap from two to "multi" is a bigger deal than I was giving it credit for. We see a lot of people wanting to run cockroachdb in two datacenters, but that won't actually give them the ability to survive a failure of either datacenter. So I think we do need to drive this point home, that even though MAA is "just" active-active for 3+ replicas, that third replica changes the capabilities of the system.
</blockquote></details>

You talk about consensus down in the details below, but I think it deserves to be promoted to the top level. The benefits of multi-active described here are the benefits of consensus replication; the reason that active-active and multi-active are different is that consensus requires a replication factor of at least three. 

---

*[v1.1/multi-active-availability.md, line 23 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtI5lzwz0Uk3QS2kdbA:-KtI5lzwz0Uk3QS2kdbB:b1b8t8g) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L23)):*
> ```Markdown
> This design has some significant drawbacks, though:
> 
> - Because you must adequately prepare for failover, your passive machines must provide production-level capacity, but remain inert the majority of the time––so you pay for hardware that you know you won't use to its full capacity.
> ```

This is true to some extent for all HA solutions. The difference is one of degree: with active-passive or active-active configurations, the need for spare capacity doubles your overall hardware needs, while with multi-active you only need 50% more hardware (or less as you go to more than three datacenters). (With multi-active no machines are never idle, but if you let them go above 66% utilization while all three DCs are online, things will break when one goes down and the other two have to pick up the slack)

---

*[v1.1/multi-active-availability.md, line 61 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIEfTOJBou51_HBgBF:-KtIEfTOJBou51_HBgBG:bytg5dk) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L61)):*
> ```Markdown
> 3. B receives a read of key `abc`, and returns the result `'123'`.
> 4. C then receives an update for key `abc` to the values `'456'`. It communicates this write to node B, who confirms that its received the write, as well. After receiving the confirmation, the change is committed.
> 5. Node A is restarted and rejoins the cluster. It receives an update that the key `abc` had its value changed to `'456'`.
> ```

It's also important that a node that previously failed will not be allowed to serve any queries until it has caught back up. 

---

*[v1.1/multi-active-availability.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIF2TSQkfUrWNpBQOu:-KtIF2TSQkfUrWNpBQOv:b-n5hu1x) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L75)):*
> ```Markdown
> ### Raft Consensus Algorithm
> 
> To create consistency among replicated ranges (also known as replicas), each replica becomes a member of a "Raft group." These groups use the Raft consensus algorithm to ensure that a majority (or consensus) of nodes acknowledge that any changes have happened to the range.
> ```

Replicated ranges *have* replicas; the terms are not synonymous. You could say "To create consistency among the replicas of a range"

"majority (or quorum)", not "or consensus". 

---

*[v1.1/multi-active-availability.md, line 81 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIFTtwAD8-4mdre7EO:-KtIFTtwAD8-4mdre7EP:b6jbq6p) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L81)):*
> ```Markdown
> ### Symmetrical Nodes
> 
> Because of CockroachDB's strong consistency, each node in your deployment reliably sees the same data. This means that CockroachDB was able to be designed to allow reads and writes from every node in the cluster, even when other nodes in the cluster are down.
> ```

This paragraph may be misleading; it sounds like you can always read from your local replica, instead of having to go to the lease holder for the range. 

---

*[v1.1/multi-active-availability.md, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIFfKvg26ZrEhen96F:-KtIFfKvg26ZrEhen96G:b3g68az) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L85)):*
> ```Markdown
> To achieve this, every node can act as a gateway for client RPCs. The gateway node then find the nodes responsible for the ranges and lets them handle the changes to the data throguh Raft. After the changes have been committed (i.e., acknowledged by a majority of the replicas), the gateway node responds to the client.
> 
> ### Gossip
> ```

This is accurate, but I don't really see why it's relevant to this page. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723)*

sploiselle:
Review status: 0 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[v1.1/multi-active-availability.md, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL82f3xXvT4ksEno2o:-Krspi6zGMyZlKngdRB3:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L13)):*

Use "replicating" instead of "scaling".

There are two components to high availability: the availability of spare capacity ready to take over for failing nodes, and an *automated* process for doing so. The latter part is worth a mention here.
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL92UL0LN6QfltyPM-:-KuLz_VVCzlmfuNUzEIM:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L17)):*

I think active-passive systems are more common, at least for SQL databases. What distinguishes active-active mode is that all nodes are processing traffic even when there are no failures. 

Active-passive setups are desirable because they exhibit strong consistency "most of the time", only risking inconsistency during failovers. 
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL9nWDpD0K8_menzuP:-KuMkPFZbxAr62mFlAO_:b-uriil0) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L21)):*

Active-active systems must remain below 50% utilization in the absence of failure, so it's not really any better than active-passive for hardware costs (it just looks better to see 40% and 40% utilization instead of 80% and 0%). 

Active-active systems can be CAP-consistent, in which case they do solve the problem of losing data when machines fail (but then they aren't really HA)
CAP-Available active-active systems can often also experience conflicts or inconsistencies even when there is no failure. 
</blockquote></details>

Can/would you run Active-Active services on > 2 machines? I can understand the complexity 

---

*[v1.1/multi-active-availability.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpLBGl85SPwuayGpgvw:-KuM56Mw5owWPW7kR2PC:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L40)):*

You talk about consensus down in the details below, but I think it deserves to be promoted to the top level. The benefits of multi-active described here are the benefits of consensus replication; the reason that active-active and multi-active are different is that consensus requires a replication factor of at least three. 
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 23 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtI5lzwz0Uk3QS2kdbA:-KuLxy3k8KM0Kb5IbIjC:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L23)):*

This is true to some extent for all HA solutions. The difference is one of degree: with active-passive or active-active configurations, the need for spare capacity doubles your overall hardware needs, while with multi-active you only need 50% more hardware (or less as you go to more than three datacenters). (With multi-active no machines are never idle, but if you let them go above 66% utilization while all three DCs are online, things will break when one goes down and the other two have to pick up the slack)
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 61 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIEfTOJBou51_HBgBF:-KuM5HDt0gQmNz1ZhLn5:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L61)):*

It's also important that a node that previously failed will not be allowed to serve any queries until it has caught back up. 
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 75 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIF2TSQkfUrWNpBQOu:-KuM5SYAgPj8x3Lygt18:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L75)):*

Replicated ranges *have* replicas; the terms are not synonymous. You could say "To create consistency among the replicas of a range"

"majority (or quorum)", not "or consensus". 
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 81 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIFTtwAD8-4mdre7EO:-KuM67MPAUJ4OcJcoYJk:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L81)):*

This paragraph may be misleading; it sounds like you can always read from your local replica, instead of having to go to the lease holder for the range. 
</blockquote></details>

Done.

---

*[v1.1/multi-active-availability.md, line 85 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KtIFfKvg26ZrEhen96F:-KuM73GiaXvEZog-Cp5z:b-v6p8xk) ([raw file](https://github.com/cockroachdb/docs/blob/8e566d3cf40ee8297fa68cb69c47cd640728f9d9/v1.1/multi-active-availability.md#L85)):*

This is accurate, but I don't really see why it's relevant to this page. 
</blockquote></details>

Removing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723)*

sploiselle:
@bdarnell I added a large section with details about how MAA reduces the amount of hardware you need. Got everything else you mentioned, but that section is new and novel and could definitely use a look. I could really use any guidance you have about expressing "an arbitrary unit of CPU work" in a way that's easy to grok.

---

Review status: 0 of 4 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-:-KuQNh0nxc1WYub6h_NC:b-j579fi)*

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[v1.0/multi-active-availability.md, line 23 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KuUnr3ZJAz7JZJtnYqU:-KuUnr3ZJAz7JZJtnYqV:b7631d3) ([raw file](https://github.com/cockroachdb/docs/blob/b811a40af6f6e9f86475c38e303220c506432f45/v1.0/multi-active-availability.md#L23)):*
> ```Markdown
> ### Active-Passive
> 
> In Active-Passive systems, all traffic is routed to a single, "active" machine. Any kind of stateful changes made to the machine are then duplicated to a backup "passive" machine, in an attempt to always mirror the active machine as closely as possible.
> ```

I'd say "replica" or "system" instead of "machine" here - each replica could be composed of multiple machines in a sharded system. (in fact, I think it needs to be for the math below to work out)

---

*[v1.0/multi-active-availability.md, line 77 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KuUokTE9hH2yEIfbpOG:-KuUokTE9hH2yEIfbpOH:b-f0qz7o) ([raw file](https://github.com/cockroachdb/docs/blob/b811a40af6f6e9f86475c38e303220c506432f45/v1.0/multi-active-availability.md#L77)):*
> ```Markdown
> In addition to providing scalable consistency, Multi-Active Availability also reduces the amount of hardware your deployment requires.
> 
> While both Active-Passive and Active-Active require twice the hardware, CockroachDB only requires an additional 50% due to Multi-Active Availability. This is an emergent principle in the way that Multi-Active Availability maximizes the amount of your machine's computing power that's in use.
> ```

Twice the hardware of what? We never clearly say that we're comparing to the single-replica low-availability option. 

Multi-active is not always exactly 3x replication (and the other scenarios are not always 2x). 5x replication gives you the ability to survive more simultaneous failures, and costs more storage but has even less overhead in terms of idle CPU (40%)

---

*[v1.0/multi-active-availability.md, line 83 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KuUp2Gv7LqHFHn7ipSk:-KuUp2Gv7LqHFHn7ipSl:b-eo2idt) ([raw file](https://github.com/cockroachdb/docs/blob/b811a40af6f6e9f86475c38e303220c506432f45/v1.0/multi-active-availability.md#L83)):*
> ```Markdown
> This table shows you the total number of machines you must have to achieve relative throughput for each type of availability.
> 
> Here we use an arbitrary unit we're calling **Effective Machine**. The aim here is to create an arbitrary unit of work––the CPU required to process 1,000 or 5,000 queries per second––and then show multipliers of that value. So 2.0 Effective Machines is simply twice the processing power of 1.0 Effective Machines.
> ```

An "effective machine" is "the CPU, memory, and storage required", not just the CPU. Or we could perhaps simplify things by being more concrete and just talking about cores:

"Suppose your database needs 8 CPU cores to process your peak workload. In active/passive and active/active arrangements, you'd need two 8-core machines. in a 3x multi-active arrangement, you'd need three 4-core machines."

It's probably best to focus on cores here because multi-active requires more storage than the other scenarios; the savings come from the CPU side (and potentially read IOPS)

Up to a point, a machine that's twice as fast doesn't cost twice as much, so 3 "half power" machines are probably more expensive than 2 "full power" ones. For 3x MAA to be a cost savings, I think you have to be talking about whole machines instead of fractional ones. That means the active/x setups need to be clusters of two machines in each replica so the MAA can be three machines. 

Maybe the cost savings are a can of worms that we don't want to open here. I wouldn't consider the cost savings to be a significant reason to choose MAA over active/active. 

---

*[v1.1/multi-active-availability.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-KpL9nWDpD0K8_menzuP:-KuUvfAjaq8nYmPwj2ui:b2tbe6a) ([raw file](https://github.com/cockroachdb/docs/blob/1550b0e76d99805faeced22c9fd875c161204962/v1.1/multi-active-availability.md#L21)):*

Can/would you run Active-Active services on > 2 machines? I can understand the complexity 
</blockquote></details>

Yes, you can. Google Reader was active-active-active (inconsistent) replication across three bigtable clusters. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723)*

jseldess:
@sploiselle, is this finishing this still on your plan?

sploiselle:
@jseldess Yep! Don't have a firm timeline due to other priorities, but will definitely wrap on this after the training is done.

jseldess:
Sounds good. Thanks, @sploiselle.

sploiselle:
@bdarnell Can I get you to give this a quick read for me? All 3 files have the same content, so review of just one of them is great.

bdarnell:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1723#-:-L4HixvOl0-xEDfsD27H:bnfp4nl)*

sploiselle:
@jseldess Can I get a copy edit from the docs team on this? Could really use it by next Wednesday to ensure we have time to integrate the links with the upcoming PDF release

jseldess:
@sploiselle, @lhirata has agreed to give it a read through. Excited to see this almost live!

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Multi-Active Availability
+summary: Learn about CockroachDB's high availability model, known as Multi-Active Availability
```
lhirata: Add period

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+toc: false
+---
+
+CockroachDB's availability model is described as "Multi-Active Availability." In essence, multi-active availability provides benefits similar to traditional notions of high availability, but also lets you read and write from every node in your cluster without generating any conflicts.
+
+<div id="toc"></div>
+
+## What is High Availability?
+
+High availability lets an application continue running even if a system hosting one of its services fails. This is achieved by scaling the application's services horizontally, i.e., replicating the service across many machines or system. If any one of them fails, the others can simply step in and perform the same service.
```
lhirata: "[...] across many machines or system." -> "[...] across many machines or systems."

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Multi-Active Availability
+summary: Learn about CockroachDB's high availability model, known as Multi-Active Availability
+toc: false
+---
+
+CockroachDB's availability model is described as "Multi-Active Availability." In essence, multi-active availability provides benefits similar to traditional notions of high availability, but also lets you read and write from every node in your cluster without generating any conflicts.
```
lhirata: Might want to link to the "What is Multi-Active Availability" section here.

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+High availability lets an application continue running even if a system hosting one of its services fails. This is achieved by scaling the application's services horizontally, i.e., replicating the service across many machines or system. If any one of them fails, the others can simply step in and perform the same service.
+
+Before diving into the details of CockroachDB's multi-active availability, we'll review the two most common high availability designs: [Active-Passive](#active-passive) and [Active-Active](#active-active) systems.
+
+### Active-Passive
+
+In active-passive systems, all traffic is routed to a single, "active" replica. Changes to the replica's state are then copied to a backup "passive" replica, in an attempt to always mirror the active replica as closely as possible.
+
+However, this design has downsides:
+- If you use asynchronous replication, you cannot guarantee that any data is ever successfully replicated to passive followers––meaning you can easily lose data. Depending on your industry, this could have pretty dire consequences.
```
lhirata: Add a line break here - the HTML formatting is off.

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+#### Example: Conflicts with Active-Active Replication
+
+For this example, we have 2 replicas (**A**, **B**) in an active-active high availability cluster.
+
+1. **A** receives a write for key `xyz` of `'123'`, and then immediately fails.
+2. **B** receives a read of key `xyz`, and returns a `NULL` because it cannot find the key.
+3. **B** then receives a write for key `xyz` of `'456'`.
+4. **A** is restarted and attempts to rejoin **B**––but what do you about key `xyz`? There's an inconsistency in the system without a clear way to resolve it.
+
+{{site.data.alerts.callout_info}}In this example, the cluster remained active the entire time. But in terms of the CAP theorem, this is an AP system; it favored being available instead of consistent.{{site.data.alerts.end}}
```
lhirata: Can / should we link to an explanation of CAP theorem? This is the first mention of it and I'd like to know more.

Also, does "AP" stand for "active-passive"?

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+
+## What is Multi-Active Availability?
+
+Multi-active availability is CockroachDB's version of high availability (keeping your application online in the face of partial failures), which we've designed to avoid the downsides of both active-passive and traditional active-active systems.
+
+Like active-active designs, all replicas can handle traffic, including both reads and writes. However, CockroachDB improves upon that design by also ensuring that data remains consistent across them, which we achieve by using "consensus replication." In this design, replication requests are sent to at least 3 replicas, and are only considered committed when a majority of replicas acknowledge that they've received it. This means that you can still failures without compromising availability.
```
lhirata: "[...] can still failures" -> "[...] can still have failures"

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+
+For this example, we have 3 CockroachDB nodes (**A**, **B**, **C**) in a multi-active availability cluster.
+
+1. **A** receives a write on `xyz` of `'123'`. It communicates this write to nodes **B** and **C**, who confirm that they've received the write, as well. Once **A** receives the first confirmation, the change is committed.
+2. **A** fails.
+3. **B** receives a read of key `xyz`, and returns the result `'123'`.
+4. **C** then receives an update for key `xyz` to the values `'456'`. It communicates this write to node **B**, who confirms that its received the write, as well. After receiving the confirmation, the change is committed.
+5. **A** is restarted and rejoins the cluster. It receives an update that the key `xyz` had its value changed to `'456'`.
+
+{{site.data.alerts.callout_info}}In this example, if nodes <strong>B</strong> or <strong>C</strong> failed at any time, the cluster would have stopped responding. In terms of the CAP theorem, this is an CP system; it favored being consistent instead of available.{{site.data.alerts.end}}
```
lhirata: Same as the callout in the first example

v1.0/multi-active-availability.md
```diff
@@ -0,0 +1,66 @@
...
+
+## What's next?
+
+To get a greater understanding of how CockroachDB is a survivable system that enforces strong consistency, check out our [architecture documentation](/architecture).
```
lhirata: Change link to /architecture/overview.html

lhirata:
@sploiselle minor edits. otherwise, LGTM!

v1.0/multi-active-availability.md
sploiselle: AP in the CAP theorem is "available/partition" which means that the system stays available (instead of consistent) when partitions occur. The other paradigm (which CockroachDB is part of) favors consistency over availability when partitions occur.

v1.0/multi-active-availability.md
sploiselle: Given that the TOC is right beneath this and the information is sequential, will leave out link but can add it in later if we get more feedback that users expect it.

sploiselle:
@lhirata Thank you! Left comments where appropriate and implemented your edits.

v1.0/multi-active-availability.md
lhirata: Gotcha!

PR #1728, opened 2017-07-18
use full urls in sitemap by using jekyll-sitemap plugin

### LINKED ISSUE #1725: Use full URLs in sitemap.xml
The [docs sitemap](https://www.cockroachlabs.com/docs/sitemap.xml) is currently reporting errors in the webmaster console because all the links are relative instead of absolute. This may be contributing to google being slow to pick up changes as noted in #1724.

benesch:
The previous sitemap used relative URLs, which was breaking the Google webmaster console. This may explain why Google has been so slow to notice deleted pages. Defer plugin generation to a plugin, jekyll-sitemap, which gets this right.

Fixes #1725.


benesch:
@jseldess after this lands you'll need to run `bundle`.

jseldess:
@sploiselle and @Amruta-Ranade, see Nikhil's note above about running `bundle`.

bdarnell:
I've confirmed that this resolves the errors I saw in the google webmaster console.

PR #1729, opened 2017-07-18
[WIP] Deployment overview

sploiselle:
Based on feedback from users, trying to add a guide for deploying CockroachDB ahead of more detailed documentation coming in future quarters.

@mberhault Could I get you to take a quick look at this page to see if I'm covering the right topics/have provided correct guidance?

### DIFF
--- a/v1.0/deploy-cockroachdb-to-production.md
+++ b/v1.0/deploy-cockroachdb-to-production.md
@@ -1,0 +1,424 @@
+---
+title: How to Deploy CockroachDB to Production
+summary: Get an overview of how best to deploy CockroachDB as a consistent, scalable database for your application.
+toc: false
+---
+
+Though CockroachDB itself is easy to use, deploying a database to production requires a lot of preparation. This page attempts to cover common things you need to understand before deploying CockroachDB––or any service––to a production environment.
+
+{{site.data.alerts.callout_info}}For help training your team on CockroachDB's deployments or best practices, <a href="https://www.cockroachlabs.com/pricing/sales/">contact the Cockroach Labs team</a>.{{site.data.alerts.end}}
+
+<div id="toc"></div>
+
+## Deployment Architecture
+
+To deploy CockroachDB to production, you need the following infrastructure.
+
+#### Nodes
+
+3+ machines that can communicate with one another.
+
+- Use at least three nodes to ensure that a majority of replicas (2/3) remains available if a node fails. Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
+
+- Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
+
+- Ideally, these machines should be located in you would want these machines located in separate locations to maximize survivability.
+
+#### Load Balancer
+
+A highly available load balancer (such as those provided by Google Cloud or through an orchestration tool like Kubernetes). You'll use this to connect to multiple nodes in your cluster; you don't want to rely on a single node, because if it goes down, your application will go down despite having other available nodes.
+
+#### Service Discovery
+
+Service discovery (such as Zookeeper or through an orchestration tool like Kubernetes) so new nodes can be found.
+
+#### Monitoring
+
+It's crucial to know when machines fail and be alerted. While CockroachDB provides an Admin UI with this information, you will also want to integrate your cluster with a tool like Prometheus to alert your team of outages.
+
+## Platform
+
+To deploy CockroachDB, it's important to understand the platform on which you plan to deploy it. 
+
+### Networking
+
+While CockroachDB can easily be deployed on Google Cloud, AWS, Microsoft Azure, or Digital Ocean, each of them has idiosyncrasies, particularly around how they handle networking. Important elements of your network to understand include:
+
+- Hostnames
+- Firewalls
+
+#### Hostnames
+
+CockroachDB requires nodes to communicate with one another, as well as accept client requests from your application. This is often handled by hostnames, but it's important to understand what those are on your platform, or if they're accessible network names.
+
+Depending on how you plan to deploy Cockroach, you need to ensure that the following hostnames resolve:
+
+- **Client communication** to let your application or load balancer communicate with your cluster. This defaults to the node's `$HOSTNAME`, but can be controlled with `cockroach start`'s `--host` flag.
+- **Inter-node communication** so nodes can communicate with one another. This defaults to the same value as the client communication hostname (i.e., `$HOSTNAME` or the value of `--host`), but can be independently controlled with `cockroach start`'s `--advertise-host` flag.
+
+#### Firewalls
+
+Because CockroachDB benefits from widely distributed deployments, it's crucial to ensure that nodes can communicate with one another. There are two common ports that you should make sure nodes can use:
+
+- `26257` is the default port for the `cockroach` binary to accept client requests, as well as communicate with one another
+- `8080` is the default port for Admin UI of CockroachDB.
+
+#### Troubleshooting
+
+Because networking between CockroachDB nodes is by far the most confounding element of deploying it, if you run into issues, check out our [Networking Troubleshooting documentation](cluster-setup-troubleshooting.html#networking-troubleshooting).
+
+### Topology & Availability Zones
+
+Because CockroachDB doesn't have an intrinsic understanding of your deployment's topology, you must manually identify the node's location (also known as "availability zones") using the `--locality` flag when starting your nodes. CockroachDB then uses these flags as a way of replicating data in a way that ensures survivability.
+
+Your nodes' `--locality` flag should express a hierarchy of topologies. For example:
+
+1. Country
+2. Region
+3. Datacenter
+
+For a deployment in the United States, your `--locality` flags might look like: 
+
+- `--locality=country=us,region=east,datacenter=us-east-1`
+- `--locality=country=us,region=east,datacenter=us-east-2`
+- `--locality=country=us,region=central,datacenter=us-central-1`
+- `--locality=country=us,region=central,datacenter=us-central-2`
+- `--locality=country=us,region=west,datacenter=us-west-1`
+- `--locality=country=us,region=west,datacenter=us-west-2`
+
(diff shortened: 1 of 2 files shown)


v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
+---
+title: How to Deploy CockroachDB to Production
+summary: Get an overview of how best to deploy CockroachDB as a consistent, scalable database for your application.
+toc: false
+---
+
+Though CockroachDB itself is easy to use, deploying a database to production requires a lot of preparation. This page attempts to cover common things you need to understand before deploying CockroachDB––or any service––to a production environment.
```
mberhault: I would drop `or any service`. we're special :)

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+## Deployment Architecture
+
+To deploy CockroachDB to production, you need the following infrastructure.
+
+#### Nodes
+
+3+ machines that can communicate with one another.
+
+- Use at least three nodes to ensure that a majority of replicas (2/3) remains available if a node fails. Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
```
mberhault: Without changing the replication factor, adding more machines doesn't prevent data unavailability, it reduces the percentage of ranges that will be unavailable.
With enough ranges, losing two machines in a 100 node cluster will still result in some ranges becoming unavailable.

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+- Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
+
+- Ideally, these machines should be located in you would want these machines located in separate locations to maximize survivability.
```
mberhault: uh. I think you mean `these machines should be located in separate locations ...`

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Load Balancer
+
+A highly available load balancer (such as those provided by Google Cloud or through an orchestration tool like Kubernetes). You'll use this to connect to multiple nodes in your cluster; you don't want to rely on a single node, because if it goes down, your application will go down despite having other available nodes.
```
mberhault: All clouds offer this, we don't need to be too nice to Google here. Let's just say "Cloud-provided load balancer"

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+## Platform
+
+To deploy CockroachDB, it's important to understand the platform on which you plan to deploy it. 
+
+### Networking
+
+While CockroachDB can easily be deployed on Google Cloud, AWS, Microsoft Azure, or Digital Ocean, each of them has idiosyncrasies, particularly around how they handle networking. Important elements of your network to understand include:
+
+- Hostnames
+- Firewalls
```
mberhault: I would add private networks as well.

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Hostnames
+
+CockroachDB requires nodes to communicate with one another, as well as accept client requests from your application. This is often handled by hostnames, but it's important to understand what those are on your platform, or if they're accessible network names.
+
+Depending on how you plan to deploy Cockroach, you need to ensure that the following hostnames resolve:
+
+- **Client communication** to let your application or load balancer communicate with your cluster. This defaults to the node's `$HOSTNAME`, but can be controlled with `cockroach start`'s `--host` flag.
```
mberhault: you should mention why we may want to use one vs another. eg: hostnames valid within the cluster due to a private subnet, vs hostnames that can be used from outside the subnet.

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Monitoring
+
+While an orchestration tool can automatically bring up new resources if machines fail, if you choose to deploy CockroachDB manually, you'll need to integrate it with a monitoring tool to alert your team of node failures.
+
+You can monitor your cluster with basically any tool, but we have documentation on integrating CockroachDB with [Prometheus](monitor-cockroachdb-with-prometheus.html).
+
+## Security
+
+To deploy CockroachDB, you *must* use SSL encryption. To make this as simple as possible, the `cockroach` binary can work a certificate authority (CA), which provides everything you need to generate SSLs.
```
mberhault: I think you meant `the cockroach binary can work as a certificate authority`, which isn't really true, we just create certs and keys.
Instead, maybe something like: `the cockroach binary provides everything you need ...`

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+  Alternatively, CockroachDB does [support password authentication](create-and-manage-users.html#secure-cluster), though we typically recommend using client certificates instead.
+
+### Deploying Certificates
+
+#### Nodes
+
+Each node requires local copies of the following:
+
+- Your certificate authority (CA) certificate to validate the signature of all other certificates
+- A node key and certificate signed by your CA with the common name of `node`
```
mberhault: Should probably mention the ip/dns addresses that can be used to reach the nodes.

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+- Use many smaller nodes instead of fewer larger ones. Recovery from a failed node is faster when data is spread across more nodes.
+- Use [zone configs](configure-replication-zones.html) to increase the replication factor from 3 (the default) to 5. You can do this for the [entire cluster](configure-replication-zones.html#edit-the-default-replication-zone) or for specific [databases](configure-replication-zones.html#create-a-replication-zone-for-a-database) or [tables](configure-replication-zones.html#create-a-replication-zone-for-a-table).
+
+## Clock Synchronization
+
+{% include faq/clock-synchronization.html %}
+
+## Cache Size
+
+If you run multiple applications on the same machine as a CockroachDB node, you might consider manually setting the cache size instead of using the default 25% of available memory.
```
mberhault: it can also apply if running other memory-intensive processes. let's say a java app on the same machine.

v1.0/deploy-cockroachdb-to-production.md
```diff
@@ -0,0 +1,424 @@
...
+
+To manually set the limit of the cache size, start the node using the [`--cache` flag](start-a-node.html#flags). For example, the following command limits a node's cache to 5GB:
+
+```shell
+$ cockroach start --cache=5GB <other start flags>
+```
+
+## File Descriptors Limit
```
mberhault: Aren't all those details on the production recommendations page? You just put a link for clock sync, but the other extreme in terms of details for this bit.

bdarnell:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[v1.0/deploy-cockroachdb-to-production.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1729#-KpMSGazwOTeSPtmwdsE:-KpMSGazwOTeSPtmwdsF:bm7z4r7) ([raw file](https://github.com/cockroachdb/docs/blob/2d818b2f295291184c491763fe3791c5473c3c20/v1.0/deploy-cockroachdb-to-production.md#L21)):*
> ```Markdown
> 3+ machines that can communicate with one another.
> 
> - Use at least three nodes to ensure that a majority of replicas (2/3) remains available if a node fails. Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
> ```

Be explicit that the replication factor is distinct from the number of nodes. A four-node cluster with a replication factor of three is just fine for a cluster that has slightly outgrown its first three nodes. 

---

*[v1.0/deploy-cockroachdb-to-production.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1729#-KpMSePB9ltBaCnLxyd3:-KpMSePB9ltBaCnLxyd4:bciny7e) ([raw file](https://github.com/cockroachdb/docs/blob/2d818b2f295291184c491763fe3791c5473c3c20/v1.0/deploy-cockroachdb-to-production.md#L25)):*
> ```Markdown
> - Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
> 
> - Ideally, these machines should be located in you would want these machines located in separate locations to maximize survivability.
> ```

Delete "you would want these machines located in"

---

*[v1.0/deploy-cockroachdb-to-production.md, line 33 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1729#-KpMSnMMMJKP1Klcb8wL:-KpMSnMMMJKP1Klcb8wM:b-160tw1) ([raw file](https://github.com/cockroachdb/docs/blob/2d818b2f295291184c491763fe3791c5473c3c20/v1.0/deploy-cockroachdb-to-production.md#L33)):*
> ```Markdown
> #### Service Discovery
> 
> Service discovery (such as Zookeeper or through an orchestration tool like Kubernetes) so new nodes can be found.
> ```

How would you use zookeeper for this? I don't think service discovery tools will help much with your cockroachdb deployment. (Clients may want service discovery tools to find the load balancer, but in many cases the load balancer's hostname is sufficient)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1729)*

jseldess:
@sploiselle, seeing this quite late. Question: Based on previous discussions, you're planning to create issues for training topics that still need to be written, with high prio ones marked for 1.1. The end-goal is to be able to assemble/resculpt a full training curriculum.

Is the idea here to have a single page that covers everything in brief? I think it may be a valuable addition, but given the amount of work ahead of us, we need to make sure we're prioritizing carefully.

sploiselle:
@jseldess Good call! I'm planning on using this page as the "springboard" into all of the deployment-related docs, which would also double as a high-level overview. I will add an overview item to the curriculum to ensure it's accounted for.

sploiselle:
@jseldess @lhirata I think this is sufficiently dead, but wanted to ping both of you before closing in case there was something you wanted to potentially keep.

jseldess:
Thanks, Sean. I think it's ok to close this at this point since we're covering most of what's here in the production checklist.

PR #1743, opened 2017-07-20
Add 1.1 callouts

### LINKED ISSUE #1639: New/changed in 1.1 callout
Once https://github.com/cockroachdb/docs/pull/1605 lands, and we start changing and expanded 1.1 pages, we'll need a way to callout out these changes to users. 

First thought is to design a little intuitive icon that we can place next to a heading or paragraph that, on hover, says `New in 1.1` or `Changed in 1.1`. @kuanluo, thoughts?

jseldess:
- Add guidance on version tags to CONTRIBUTING.md
- Add version tag to node_id variable
- Add version tag to SQL Shell Welcome section

Fixes #1639

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,23 +106,41 @@ Field | Description | Default
 
 The CockroachDB Jekyll theme can auto-generate a page-level table of contents listing all h2 and h3 headers or just all h2 headers on the page. Related files: `js/toc.js` and `_includes/toc.html`.
 
--   To add a page TOC to the very top of the page, set `toc: true` in the page's front-matter.
+- To add a page TOC to the very top of the page, set `toc: true` in the page's front-matter.
 
--   To add a page TOC anywhere else on the page (for example, after an intro paragraph), set `toc: false` in the page's front-matter and add the following HTML where you want the toc to appear on the page:
+- To add a page TOC anywhere else on the page (for example, after an intro paragraph), set `toc: false` in the page's front-matter and add the following HTML where you want the toc to appear on the page:
 
     ``` html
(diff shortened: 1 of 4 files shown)


jseldess:
@sploiselle, @Amruta-Ranade, we now have a way to callout when a feature is new/changed in a new version. This PR adds a section to `CONTRIBUTING.md` and adds a few callouts to 1.1 docs. Would be great if you both could review and let me know if it all makes sense.

CONTRIBUTING.md
```diff
@@ -106,24 +106,42 @@ Field | Description | Default
...
+- To add a version tag to a paragraph, place `<span class="version-tag">New in vX.X:</span>` at the start of the paragraph, e.g:
+
+    ```
+    <span class="version-tag">New in v1.1:</span> The `user_privileges` view identifies global priveleges.
+    ```
+
+- To add a version tag to a heading, place `<div class="version-tag">New in vX.X</div>` to the right of the heading, e.g.:
+
+    ```
+    ## SQL Shell Welcome <div class="version-tag">New in v1.1</div>
```
Amruta-Ranade: I am wondering why we use <div> for headings and <span> for paragraphs. Given the issues I faced with <div>, I prefer <span>. Also, technically, <span> seems like the appropriate choice because it is an inline element.

jseldess: Good point. `<span>` works for headings, too. I've updated the guidance here as well as the implementation in `use-the-built-in-sql-client.md`.

@deronjohnson, just to be safe, I want to ask you: Is there any reason why a version tag would need to be in a `<div>` for headings as opposed to paragraphs?

PR #1754, opened 2017-07-26
Remove documentation of positive constraints.

a-robinson:
According to Bram, they were never really finished and are something
that we'd prefer to fully think through before committing to supporting
them. See https://github.com/cockroachdb/cockroach/issues/14163 for more detail.


jseldess:
LGTM

a-robinson:
Are you still on board with this, @BramGruneir?

BramGruneir:
!00%.  Thanks for doing this.

LGTM

PR #1756, opened 2017-07-27
Comparison chart update

deronjohnson:
Updates comparison chart to use dropdowns. Adds select2 for dropdown functionality and styling.

@jess-edwards @sploiselle @kuanluo


sploiselle:
@deronjohnson Can I get you to add an `alt` tag to `/images/icon_info.svg`? Something like "tooltip icon"?

deronjohnson:
@sploiselle done!

PR #1762, opened 2017-08-02
Update set-cluster-setting.md

vivekmenezes:


asubiotto:
The [v1.1 page](https://github.com/cockroachdb/docs/blob/483df71298821df9c0c1fc15f01dfb602328b08b/v1.1/set-cluster-setting.md) also needs to be updated

sploiselle:
Adding 1.1 file

sploiselle:
@vivekmenezes @asubiotto Was this a typo in the docs, or was it a change to the binary? If the latter, which subversion does it apply to going forward?

asubiotto:
It was a typo in the docs. [Here is the line in the commit](https://github.com/cockroachdb/cockroach/commit/4534d7c07570e62cbdbe1b0d3ed2d9c8c2804bea#diff-0714f3775a457dbf0c6ce7b5f702e5bdR120) adding the cluster setting.

sploiselle:
Thanks, folks!

PR #1763, opened 2017-08-02
Install binary through terminal

### LINKED ISSUE #1667: Improve experience getting binary onto machine
Mixing the browser download of the binary, followed by execution through the terminal causes a lot of confusion for unfamiliar devs.

There are many potential solutions to this problem, the simplest of which might be to have the install instructions rely entirely on the terminal instead of mixing browser and terminal use.

sploiselle:
Closes #1667

@jess-edwards Can you put me in touch with the right person from DivOf to ensure that the tracking for this looks like it would work?

Additional work item in the future is to include Copy to Clipboard button here. I couldn't get this to work right away because the HTML/MarkDown mixture is preventing any MarkDown from processing. Tried placing the Copy to Clipboard include on the page and it wasn't rendering properly.


jess-edwards:
@sploiselle Email to Igor sent.

jseldess:
@sploiselle, I think I know how to get copy-to-clipboard to work here.

1. Add `markdown="1"` to the div for each install method.
2. Place `{% include copy-clipboard.html %}` before the div for each code block where you want the feature.
3. In the opening `div` tag for the code block, make sure `class="highlight"` is present. In most cases, it looks like the opening div tags use the now-defunct `class="highlighter-rouge"`.

Example:

```
<div id="download-the-binary" class="install-option" markdown="1">
  <h2>Download the Binary</h2>
  <ol>
    <li>
      <p>Download the CockroachDB archive for OS X:</p>

      <div class="highlight"><pre class="highlight"><code data-eventcategory="mac-binary-step1"><span class="gp" data-eventcategory="mac-binary-step1">$ </span>curl -O https://binaries.cockroachdb.com/cockroach-{{page.release_info.version}}.darwin-10.9-amd64.tgz</code></pre></div>
    </li>
```

That results in:

screen shot 2017-08-03 at 7 39 58 am

sploiselle:
@jseldess ` markdown="1"`! I feel like I should have known (or _did_ know) that. Will add that in a bit.

v1.0/install-cockroachdb.md
```diff
@@ -316,7 +318,10 @@ $(document).ready(function(){
 
   <ol>
     <li>
-      <p>Download the <a href="https://binaries.cockroachdb.com/cockroach-{{ page.release_info.version }}.linux-amd64.tgz" data-eventcategory="linux-binary-step1">CockroachDB {{ page.release_info.version }} archive for Linux</a>.</p>
+      <p>Download the CockroachDB archive for Linux:</p>
+
+      <div class="highlighter-rouge"><pre class="highlight"><code data-eventcategory="linux-binary-step2"><span class="gp" data-eventcategory="linux-binary-step2">$ </span>wget https://binaries.cockroachdb.com/cockroach-{{page.release_info.version}}.linux-amd64.tgz</code></pre>
```
bdarnell: I think we still want a link somewhere. Either the URL itself could be a link (if that doesn't interfere with copy/paste), or we could make "CockroachDB archive" a link.

jseldess:
@sploiselle, if you don't have time to wrap this up before your vacation, I'm happy to wrap it up.

sploiselle:
@jseldess Will get this up today and will let you fix it up while I'm out!

v1.0/install-cockroachdb.md
sploiselle: Done.

sploiselle:
@jseldess Content here is what we were looking for:

1. Has customers download archive from terminal (added bonus, doesn't keep junky archive file)
2. Adds Copy-to-Click. In adding the feature to this page, I discovered some bugs in copy-to-click. I'll create a bug for tracking(#1807), but don't have a code sample ready off-hand until this merges.

So, this is good-to-go for a takeover/copy edits. Thanks!

jseldess:
LGTM, with a few minor updates forthcoming. But something seems to have broken the hubspot form:

screen shot 2017-08-10 at 11 15 25 am

jseldess:
@benesch, I can't figure out what change in this PR caused the hubspot form to break (see screenshot above). Do you have time to help troubleshoot? If not, I can loop in Division Of.

benesch:
If I had to guess, it's setting markdown=1 that causes the problem. What requires Markdown? (Sorry, on my phone.) Can we work around by avoiding Markdown instead?

sploiselle:
The Click to Copy button didn't work without enabling Markdown :(

benesch:
Huh, okay. OOO ATM, but let me see if I can fix this when I get in.

jseldess:
Actually, click to copy does work fine without `markdown="1"` for me, locally. I suspect the rendering issues @sploiselle was seeing earlier were due to incorrect class names.

In any case, even without `markdown="1"`, the hubspot forms aren't rendering...

jseldess:
Removed the commit for bypassing htmlproofer. Added that to https://github.com/cockroachdb/docs/pull/1815, which will land sooner.

benesch:
Ok, just pushed a commit that should fix it. See the commit message for details.

jseldess:
@jess-edwards, any word from Igor on additional work to be done to properly track installs after this change?

jess-edwards:
@jseldess - Igor asked us to flag Deron on this. Doing so below:

@deronjohnson Can you work with Igor to add the Google Analytics conversion tracking to this page? To be tracked are instructions for install on: (1) MacOS-Binary and (2) LinuxOS-Binary
1. "CockroachDB Archive": Should maintain left-click and right-click tracking - likely no action/change required.
2. The "copy" button on the code snippet in step one should get click tracking. This can be the same as left-click on "CockroachDB Archive". 

Other feedback: 
* The addition of the URL on "CockroachDB Archive" makes step 1 a little confusing. It looks like you're downloading the binary twice. I know @bdarnell requested the URL to be added, but the addition makes that step more confusing.

* Something about the "Note" feels a bit out of place or awkward. I think it's the phrasing that's causing the awkwardness. Perhaps something like: 
"Note: To download other versions of the binary, visit our Releases page." 

* The "copy" button is overlapping the code snippet in step 1. Can you check the padding on it?

jseldess:
Thanks, @jess-edwards. I'll wait to merge this until I get confirmation from @deronjohnson that event tracking has been updated.

In terms of the feedback:
- @bdarnell, can you clarify why we need to keep the archive link?
- I can work on the note.
- Let's handle the "copy" button styling separately. I'll open an issue.

bdarnell:
We don't *need* to keep the archive link, but if I'm writing my own download script instead of following the exact instructions here (or adapting the instructions for my preferences/quirks), I'd rather have a link I can copy instead of having to select the URL out of the curl command line. Personally I would always copy the URL and type the commands myself instead of copy/pasting the whole command, because I want to understand what I'm running.

jseldess:
@deronjohnson, any updates? I'd like to get this PR merged soon.

deronjohnson:
@jess-edwards @jseldess Sorry, for whatever reason I missed being pinged on this. I'll check in with @igork29 (now on GitHub) on this and get you an update asap.

jseldess:
Thanks, @deronjohnson.

igork29:
@jess-edwards @jseldess apologies for missing this. I am now on GotHub so you can tag me here. Tracking has been added to the button click action on step 1 for Mac and Linux binary. 

@jess-edwards should this be set as a goal in GA?

Also, I noticed that the data-EventCategory class was missing from the download binary link. We added it back, but not sure how it disappeared. We use data-EventCategory ="" tags to target elements for tracking purposes. They should not be removed when updates are made to pages.

deronjohnson:
@jseldess just wanted to make sure you saw my pull request for the tracking updates [here](https://github.com/cockroachdb/docs/pull/1862)

jseldess:
@deronjohnson, to make things easier, I just added the changes in https://github.com/cockroachdb/docs/pull/1862/files into this PR. Please take a look and let me know if I missed anything. Sorry that those data attributes went missing at some point.

Would you be willing to also add event tracking for all other uses of the copy clipboard button, if that's easy? I'd like to track that.

deronjohnson:
@jseldess, do you want to add tracking to the copy button site wide, or just on this page?

jseldess:
@deronjohnson, site-wide, please.

jseldess:
Everything's looking/working good now. Since adding site-wide tracking on copy-to-clipboard is independent of this, I'll merge.

jess-edwards:
@igork29 Yes, please, for the Install page only, this should be tracked as a conversion in GA. 

If you add the tracking to the copy button site-wide, please be sure that any instances of it outside of the install page are NOT tracked as conversions in GA.  

cc @deronjohnson

sploiselle:
Thank you, @jseldess and @jess-edwards for your fastidiousness in getting this done!

PR #1766, opened 2017-08-04
correct release notes

andreimatei:


sploiselle:
@andreimatei Is this change substantial enough that it ought to be documented, or was this a more subtle fix of an existing kind-of-unexpected behavior?

releases/v1.1-alpha.20170803.md
```diff
@@ -57,7 +57,7 @@ Get future release notes emailed to you:
 
 ### Bug Fixes
 
-- The first statement (or batch of statements) in a transaction can no longer return "retry transaction" errors; the server will retry the statement automatically. [#16874](https://github.com/cockroachdb/cockroach/pull/16874)
+- Retryable errors that were encountered by the first batch of statements in a transaction but were deferred and only returned to the client at COMMIT time are now handled transparently by no longer deferring them and automatically retrying the first batch. [#16874](https://github.com/cockroachdb/cockroach/pull/16874)
```
bdarnell: I think this is hard to follow because it introduces a lot of new concepts (deferred errors?). How about something like "Transactions that encounter a retryable error in their first batch of statements now retry automatically"? I don't think we need to be perfectly precise here; the PR link is there if anyone needs more details.

andreimatei:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[releases/v1.1-alpha.20170803.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1766#-KqjIE-aCTrNNk8H6YAi-r1-60:-KqykLi3dqTJFecAPtbc:bfuue58) ([raw file](https://github.com/cockroachdb/docs/blob/727aff5f167084f7341b30cc9f74de425f8abd84/releases/v1.1-alpha.20170803.md#L60)):*

I think this is hard to follow because it introduces a lot of new concepts (deferred errors?). How about something like "Transactions that encounter a retryable error in their first batch of statements now retry automatically"? I don't think we need to be perfectly precise here; the PR link is there if anyone needs more details. 
</blockquote></details>

Well what bothered me about the initial phrasing (same as your suggested one) is the implication that statements in the first batch _could_ have returned retryable errors. That's not true - and we tell people that it's not true.
It's not particularly important; we can also leave it alone.

But what about "A behavior causing statements in the first batch of a transaction to essentially defer their retryable errors until COMMIT time was removed; this partially defeated the effort to transparently retry first batches. As a result, fewer retryable errors should be encountered by COMMIT."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1766)*

bdarnell:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[releases/v1.1-alpha.20170803.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/1766#-KqjIE-aCTrNNk8H6YAi-r1-60:-KqylndEahiVXZDXJrqQ:bkae59i) ([raw file](https://github.com/cockroachdb/docs/blob/727aff5f167084f7341b30cc9f74de425f8abd84/releases/v1.1-alpha.20170803.md#L60)):*

Well what bothered me about the initial phrasing (same as your suggested one) is the implication that statements in the first batch _could_ have returned retryable errors. That's not true - and we tell people that it's not true.
It's not particularly important; we can also leave it alone.

But what about "A behavior causing statements in the first batch of a transaction to essentially defer their retryable errors until COMMIT time was removed; this partially defeated the effort to transparently retry first batches. As a result, fewer retryable errors should be encountered by COMMIT."
</blockquote></details>

That's fine with me. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1766#-:-KqymLNxC0cAfl_3oxWl:bnfp4nl)*

sploiselle:
> We removed a behavior causing statements in the first batch of a transaction to essentially defer their retryable errors until `COMMIT`; now, instead, we automatically retry statements in the first batch. As a result, fewer retryable errors should be encountered by `COMMIT`.

If this is technically inaccurate, sorry! Idea is just to tell people what we do now that we changed this behavior.

andreimatei:
As Sean said.

PR #1790, opened 2017-08-07
Parallel statement execution doc

### LINKED ISSUE #1290: sql: Support parallel statement execution using RETURNING NOTHING syntax
RFC: https://github.com/cockroachdb/cockroach/pull/13160
Implementation PRs:
- https://github.com/cockroachdb/cockroach/pull/14188
- https://github.com/cockroachdb/cockroach/pull/14553
- https://github.com/cockroachdb/cockroach/pull/14803
- etc.

The parallel statement execution feature can now be used and is demonstrating the expected speedups for target workloads (near-linear speedup for fully-independent statement sequences). I'm starting to write a blog post for this, but figured that we should have documentation for it as well before the post goes live.

Amruta-Ranade:
Closes #1290


jseldess:
Thanks for opening, @Amruta-Ranade. In addition to my review, you should assign a dev as reviewer as well. Nathan?

Amruta-Ranade:
I did assign Nathan as a reviewer. Is it not showing up?

jseldess:
Oh, yes, it is. Sorry about that.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,132 @@
+---
+title: Parallel Statement Execution
+summary: The parallel statement execution feature allows parallel execution of multiple independent SQL statements within a transaction.
+toc: false
+---
+
+CockroachDB supports parallel execution of independent INSERT, UPDATE, UPSERT, and DELETE statements within a single transaction. 
+
+To execute statements in parallel, use the `RETURNING NOTHING` clause for the independent SQL statements.
```
nvanbenschoten: Let's reword

> use the `RETURNING NOTHING` clause for the independent SQL statements

to 

> append the `RETURNING NOTHING` suffix on to each SQL statement.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,132 @@
...
+
+<div id="toc"></div>
+
+## Independent SQL statements
+
+CockroachDB considers SQL statements to be independent if they operate on independent tables and their execution can be safely reordered without affecting their results.
```
nvanbenschoten: This jumps into how we classify independent statements without really discussing why that's important. I think we should pull up a lot of the detail we have in the examples to here. For instance, we first need to discuss how `RETURNING NOTHING` statements will be run asynchronously and will return immediately to the client before returning any result. We then need to discuss how statements running asynchronously will be run at the same time as long as that won't impact correctness. Finally, we can discuss how statement independence allows us to infer whether running statements concurrently affects correctness. After that, we'll need to discuss statement synchronization and error handling (which you discuss below, but which I think get lost down there).

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,132 @@
...
+|  3 |  3000.0 | Pam      |
+|  4 |    3782 | Heer     |
+|  5 |    9887 | Julio    |
++----+---------+----------+
+(5 rows)
+~~~
+
+The execution timeline for the parallel execution of the independent statements is:
+
+BEGIN &nbsp; &nbsp; &nbsp; INSERT-----\INSERT &nbsp; &nbsp; &nbsp; COMMIT
```
nvanbenschoten: Could we somehow put this into a graphic?

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,132 @@
...
+<p>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; INSERT-----\INSERT </p>
+<p>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; UPDATE-----\UPDATE </p>
+
+
+### Handling independent and dependent SQL statements
```
nvanbenschoten: I don't think these sections should be nested under the examples header. @jseldess, what do you think?

v1.1/parallel-statement-execution.md
Amruta-Ranade: Yes, I had the same thought. Will look into it.

v1.1/parallel-statement-execution.md
Amruta-Ranade: I agree. I have been going back-and-forth about how to structure the document. I think the way we usually highlight things is through examples, and I also thought it was important to first clarify what we mean by independent statements so the users have a clear idea of when they can optimize the performance and when they shouldn't. In short, I prioritized the "when" and "why" above the "how". But I agree that in explaining the independent statements, the details about using the RETURNING NOTHING clause are getting buried. Will look into restructuring the document.

v1.1/parallel-statement-execution.md
Amruta-Ranade: We changed it from 'RETURNING NOTHING parameter' to 'RETURNING NOTHING clause' just yesterday :) Need to decide on the terminology. I would still prefer 'parameter' or 'clause' because that's what we use in other docs.

jseldess: I'm not sure there's a great answer here. Typically, we use `clause` when values need to be provided following an important keyword like `ORDER BY` or `WHERE`. In places where `RETURNING` is followed by values, therefore, it makes sense to refer to the `RETURNING` clause. But here, there's nothing but keywords. `suffix` is precise in that these keywords must always be at the end of the statement. I'd suggest we just go with that for now.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,132 @@
+---
+title: Parallel Statement Execution
+summary: The parallel statement execution feature allows parallel execution of multiple independent SQL statements within a transaction.
+toc: false
+---
+
+CockroachDB supports parallel execution of independent INSERT, UPDATE, UPSERT, and DELETE statements within a single transaction.
```
jseldess: Need to code-format these sql statements and link to their respective docs:

```
[`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html)
```

v1.1/parallel-statement-execution.md
jseldess: @Amruta-Ranade, this is a good start, but I agree with Nathan that it's not clear why this feature is important. You've covered the `how` well, but I don't come away with any sense of `why` you'd want to use this.

Instead of structuring this page the way we do for sql statements, perhaps take a look at some of the pages that cover broader sql features, for example, [Views](view.html) or [Interleaved Tables](interleave-in-parent.html). Even if not the best final structure, it might help you make progress to do what the views page does and divide the page into "Why Use Parallel Statement Execution?" and "How It Works". 

I'm happy to sync in person to sketch out some other ideas as well. @robert-s-lee will have good ideas, too. He should certainly review this at a later stage, since he's recommended this feature to prospects already.

v1.1/parallel-statement-execution.md
jseldess: Since the overall structure needs to be rethought, now sure where these will sit best. Let's see. If we end up with something like "Why Use Parallel Statement Execution?" and "How It Works", these would fit better under the latter.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Thank you so much for the links! I was structuring the document as per the SQL statements documents I have worked on so far. But this document is more closely aligned with the broader SQL topics. Will rework the structure accordingly.

robert-s-lee:
there is result of Apache jMeter demo showing 3x throughput improvement using this feature at https://docs.google.com/spreadsheets/d/1HHR4QxLAgMYluw0Zv1NWouLROddUzIXZP06e_3FG764/edit#gid=0  An inclusion of demo and example would be useful in the "how it works section"

Amruta-Ranade:
@nvanbenschoten I restructured the document as per the discussion. I am still copy-editing, but just wanted to check if I am on the right track. Let me know what you think :)

Amruta-Ranade:
@robert-s-lee @nvanbenschoten About including the demo and example, I think it is a better fit for the blog post Nathan's working on rather than the how-to document. What do you think?

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+
+On encountering the error, CockroachDB will stop the execution of all parallel statements and [abort and retry the transaction](transactions.html#error-handling). So even an error message mismatch occurs, the result will be the same as aborting the transaction.
+
+### `RETURNING NOTHING` clause appended to dependent statements
+
+If two consecutive statements are not independent, and yet a `RETURNING NOTHING` clause is added to the statements, CockroachDB will detect the dependence and serialize the execution. You can thus use the `RETURNING NOTHING` clause with SQL statements without worrying about their dependence.
+
+### Delayed execution of dependent statements
```
nvanbenschoten: I think what you mean here is synchronous (returning results) vs. asynchronous (running in the background without returning results), not dependent vs independent.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+toc: false
+---
+
+CockroachDB supports parallel execution of independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single transaction. Parallel execution of independent SQL statements reduces the overall execution time of multiple SQL statements, thereby improving query performance.
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+Applications usually issue multiple SQL statements within a single transaction. SQL engines normally execute each SQL statement sequentially. Because the queries are executed locally, the results can be delivered fast. But CockroachDB being a distributed database, execution of queries takes longer than normal SQL engines because of high round-trip latencies required for consensus. To counteract the issue and optimize query performance, CockroachDB supports parallel execution of independent SQL statements. Thus, despite the higher latency of single data-modifying statements, the overall latency of multiple such statements can be reduced.
```
nvanbenschoten: > execution of queries takes longer

let's change this to "execution of mutating statements can take longer"

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+
+## How Parallel Statement Execution Works
+
+SQL statements within a single transaction can be executed in parallel if the statements are independent. Normal execution of SQL statements returns a successful return value immediately from the server. After the return value is received, the next SQL statement is executed. Appending the `RETURNING NOTHING` clause with independent SQL statements prevents the server from sending return values and allows CockroachDB to execute the statements in parallel.
```
nvanbenschoten: To expand on what I meant above, in this section only the first use of "independent" is necessary. Statements don't need to be independent for the other two to be true.

v1.1/parallel-statement-execution.md
nvanbenschoten: my other point was "can take longer" instead of "take longer". We don't need to be overly pessimistic :)

nvanbenschoten: Also, I'm not sure I'd use the word "issue". Perhaps consider the word "challenge"

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,125 @@
...
+
+## How Parallel Statement Execution Works
+
+SQL statements within a single transaction can be executed in parallel if the statements are independent. CockroachDB considers SQL statements within a single transaction to be independent if they operate on independent tables and their execution can be safely reordered without affecting their results. 
+
+Normal execution of SQL statements returns a successful return value immediately from the server. After the return value is received, the next SQL statement is executed. Appending the `RETURNING NOTHING` clause with SQL statements prevents the server from sending return values and allows CockroachDB to execute the statements in parallel. Thus to execute statements in parallel, append the `RETURNING NOTHING` suffix to the SQL statements.
```
nvanbenschoten: It doesn't have to be "successful", it can also return an error.

For reference, the equivalent paragraph in my blog post was:
> The challenge with introducing any form of parallelism in SQL is that the language is traditionally exposed as a statement at-a-time conversational API. This means that a client sends a statement to a SQL database, the database processes the statement, and finally the statement’s results are sent back to the client. Only at this time can the next statement be issued. In order to gain any parallelism between these statements, we first needed to break this mold.

> We did this by introducing a new syntax called `RETURNING NOTHING`. Usually statements like INSERT and UPDATE return the number of rows affected. However, they can be augmented with a `RETURNING` clause to specify a set of columns from the affected rows to return. Much like other `RETURNING` clauses, our new `RETURNING NOTHING` clause is appended on to the end of statements like INSERT and UPDATE. Instead of returning a set of columns though, this new syntax does exactly what it says; it returns nothing. In doing so, it creates an opportunity for CockroachDB to return to the client with nothing but an acknowledgement and allow it to issue more statements before the first one has finished executing. When a second statement is issued, it can then be run in parallel with the first statement. Furthermore, if a chain of statements within a transaction are run with the `RETURNING NOTHING` syntax, they can all be run together and in parallel!

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Error message mismatch
+
+It is possible to see a mismatch in the error message being displayed and the statement being executed. Consider a scenario where a dependent statement is executed after a batch of independent statements, and one of the independent statements results in an error. Now because the `RETURNING NOTHING` clause is used with the independent statements, CockroachDB cannot return the error for the independent statement. The error message for the independent statement is then reported on the next dependent statement. Thus you might see a mismatch with the error message to the statement being executed.
```
nvanbenschoten: It sounds like dependent vs. independent is still used in places where we wanted sequential vs. parallel

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,125 @@
...
+
+If two consecutive statements are not independent, and yet a `RETURNING NOTHING` clause is added to the statements, CockroachDB will detect the dependence and serialize the execution. You can thus use the `RETURNING NOTHING` clause with SQL statements without worrying about their dependence.
+
+### Delayed execution of sequentially-executed statements 
+
+The SQL statements to be executed in parallel should have a `RETURNING NOTHING` clause, while sequentially-executed statements do not have the `RETURNING NOTHING` clause. Consider a scenario where a sequentially-executed SQL statement follows a batch of parallely-executed statements within a single transaction. The sequentially-executed statement will be executed after all the parallel statements are done executing. Thus it may seem as if the sequentially-executed statement is taking longer to execute, but it's in fact waiting on the parallel statements. Even then, the total time required for parallel execution of statements followed by the sequential execution of the SQL statement should be less than time required for the sequential execution of all statements. 
+
+### Error message mismatch
```
nvanbenschoten: I don't think mismatch is the right word here. This is part of the contract of parallel statements. If there is an error when executing them, it WILL be returned on the first sequential statement following the parallel batch (the barrier statement, as the blog post calls it).

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,125 @@
...
+> UPDATE a SET b = 2 WHERE y = 1;
+> UPDATE a SET b = 3 WHERE y = 1;
+~~~
+
+~~~ sql
+> UPDATE a SET y = true  WHERE y = false;
+> UPDATE a SET y = false WHERE y = true;
+~~~
+
+## Required Privileges
```
nvanbenschoten: Is this section needed?

v1.1/parallel-statement-execution.md
Amruta-Ranade: Agreed. Will delete "successful"
Also, rewording the why use parallel statement section. Does seem better for the doc:

## Why use parallel statement execution
To maintain strong consistency across a globally-distributed cluster, CockroachDB replicates writes using the [consensus protocol](https://www.cockroachlabs.com/blog/consensus-made-thrive/). The protocol requires majority of replicas to agree on a write before committing it. Thus writes to CockroachDB require coordination between globally-distributed nodes. Communication between these nodes takes time and contributes to communication latency. Executing SQL statements sequentially, as in traditional SQL engines, leads to higher cumulative communication latency.

Parallel execution of statements can help reduce the communication latency. With parallel statement execution, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements.

v1.1/parallel-statement-execution.md
Amruta-Ranade: But it is a mismatch in the sense you would not expect to see the error message after the sequential statement. Like in example in the blog post, you would not expect to see the error message for primary key constraint after the barrier statement. I don't think "error message handling" is correct, because the error isn't being handled, the message is just out of order.

v1.1/parallel-statement-execution.md
Amruta-Ranade: I think so.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Reworded:
It is possible to see a mismatch in the error message being displayed and the statement being executed. Consider a scenario where a sequentially-executed statement is executed after a batch of statements executed in parallel, and one of the parallel statements results in an error. Now because the `RETURNING NOTHING` clause is used with the parallel statements, CockroachDB cannot return the error for the parallel statement. The error message for the parallel statement is then reported on the next sequential statement. Thus you might see a mismatch with the error message to the statement being executed.

v1.1/parallel-statement-execution.md
nvanbenschoten: Why? What does a user learn from this at the moment? These are all privileges of the statements themselves and are not specific to RETURNING NOTHING. In that way, this is very similar to views, which are orthogonal to the specific statements they use.

v1.1/parallel-statement-execution.md
nvanbenschoten: This sounds so much better!

v1.1/parallel-statement-execution.md
Amruta-Ranade: My thinking was they need privileges to execute the statements, so by extension, they need the privileges for RETURNING NOTHING.

jseldess: @Amruta-Ranade, I agree with Nathan that it's unnecessary to put this here. Privileges are spelled out on each statement page.

nvanbenschoten: I guess I don't feel very strongly about this, but it just seems redundant. We wouldn't say that there are required privileges to use WHERE clauses, would we? 

Also, right now it sounds like users need all of those privileges when in reality they need privileges only for whichever statements they're adding RETURNING NOTHING on to. For instance, `INSERT ... RETURNING NOTHING` only needs `INSERT` privileges, not all 4.

Amruta-Ranade: Hmm..that makes sense. Will delete the section.

jseldess:
@Amruta-Ranade, are you ready for me to do another review?

Amruta-Ranade:
@jseldess Yes, I think the document is ready for review :)

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,122 @@
...
+toc: false
+---
+
+CockroachDB supports parallel execution of multiple independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single transaction. Parallel execution of SQL statements reduces the overall execution time of multiple SQL statements.
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+To maintain strong consistency across a globally-distributed cluster, CockroachDB replicates writes using the [consensus protocol](https://www.cockroachlabs.com/blog/consensus-made-thrive/). The protocol requires majority of replicas to agree on a write before committing it. Thus writes to CockroachDB require coordination between globally-distributed nodes. Communication between these nodes takes time and contributes to communication latency. Executing SQL statements sequentially, as in traditional SQL engines, leads to higher cumulative communication latency.
```
jseldess: Copy edit suggestion:

> To maintain strong consistency across a distributed cluster, CockroachDB replicates writes using the Raft consensus protocol, which requires that a majority of replicas agree before a write is committed. Replicas always live on separate nodes, and communication between these nodes takes time. Executing SQL statements sequentially, as in traditional SQL engines, leads to higher cumulative communication latency.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,122 @@
...
+
+Parallel execution of statements can help reduce the communication latency. With parallel statement execution, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements.
```
jseldess: This paragraph is nice but feels a little bit redundant in places. I'd suggest trimming it down to:

> With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,122 @@
+---
+title: Parallel Statement Execution
+summary: The parallel statement execution feature allows parallel execution of multiple independent SQL statements within a transaction.
+toc: false
+---
+
+CockroachDB supports parallel execution of multiple independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single transaction. Parallel execution of SQL statements reduces the overall execution time of multiple SQL statements.
```
jseldess: Also in the first sentence, let's make "transaction" a link: `[transaction](transactions.html)`

Amruta-Ranade:
@jseldess @nvanbenschoten I agree that the message is not as clear as I would like it to be. I was trying to strike a balance between including technical details, yet keeping it to-the-point. On reading through Nathan's blog post, I think I can leave out the technical details and point readers to the blog post for details, since Nathan covers them well in the post. With regards to your suggested paragraph, we also need to talk about communication latency and how parallel statement execution helps compensate for latency of executing multiple statements.

As for independent statements touching the same table, me and Nathan had decided to put it in for now, till it is fixed in code. Now that it's done, I will happily remove it :)

jseldess:
@Amruta-Ranade, when users come to this page, they should be able to understand and use the feature, without bouncing away, so I don't think we should depend on the blog post too much. We can get the message across while also conveying technical details. Let's sync on Monday and do another iteration, ok?

Amruta-Ranade:
@jseldess Sounds good :)

Amruta-Ranade:
@jseldess @nvanbenschoten As per our discussion, I have included diagrams and examples in the document. This is a preliminary draft to check if I am on the right track. If yes, I will edit the document further. Please let me know what you think of it.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+CockroachDB supports parallel execution of independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single [transaction](https://www.cockroachlabs.com/docs/dev/transactions.html). Executing statements in parallel helps reduce aggregate latency and improve performance. 
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+To understand why parallel statement execution is required, let's consider a sample scenario to demonstrate how SQL statements are executed. 
+
+Consider a scenario where a user's last name, favorite movie, and favorite song are to be updated on a social networking application. Suppose the database has three tables: username, movies, and songs, that need to be updated to update the user information. Also suppose that the user's ID is 1. Then the traditional transaction to update the user's information is as follows:
```
nvanbenschoten: ~Also suppose that the user's ID is 1.~ No need to say that :)

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+> BEGIN;
+> UPDATE username SET last_name = 'Smith' WHERE id = 1;
+> UPDATE movies SET favorite_movie = 'The Matrix' WHERE id = 1;
+> UPDATE songs SET favorite_song = 'All this time' WHERE id = 1;
+> COMMIT;
+~~~
+
+In this case, the statements are executed sequentially. Sequential execution of SQL statements returns a return value immediately from the server. After the return value is received, the next SQL statement is executed. The following conceptual diagram shows the sequential execution of the statements.
+
+<img src="{{ 'images/Sequential_Statement_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution Error Mismatch" style="border:1px solid #eee;max-width:100%" />
```
nvanbenschoten: I don't think we need the "transaction" bracket in the diagrams. It's just extra clutter and the Begin/Commit statements already make it very clear.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+~~~ sql
+> BEGIN;
+> UPDATE username SET last_name = 'Smith' WHERE id = 1;
+> UPDATE movies SET favorite_movie = 'The Matrix' WHERE id = 1;
+> UPDATE songs SET favorite_song = 'All this time' WHERE id = 1;
+> COMMIT;
+~~~
+
+In this case, the statements are executed sequentially. Sequential execution of SQL statements returns a return value immediately from the server. After the return value is received, the next SQL statement is executed. The following conceptual diagram shows the sequential execution of the statements.
```
nvanbenschoten: This is a super important paragraph! It's good now, but I think it could be clarified just a bit. Here's a suggestion that you can take or leave:

> In this case, the statements are executed sequentially, with each statement being run to completion before its results are returned to the client. After the return value is received by the client, the next SQL statement is delivered and executed. This is often described as a "conversational API" and the following conceptual diagram demonstrates why.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+<img src="{{ 'images/Sequential_Statement_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution Error Mismatch" style="border:1px solid #eee;max-width:100%" />
+
+Traditional SQL engines execute the statements sequentially without affecting performance. But in case of globally-distributed CockroachDB clusters, sequential execution of statements adds to aggregate latency. This is because to maintain strong consistency across a globally-distributed cluster, CockroachDB replicates writes using the Raft [consensus protocol](https://www.cockroachlabs.com/blog/consensus-made-thrive/). The protocol requires majority of replicas to agree before a write is committed. Replicas always live on separate nodes, and communication between these nodes takes time. Executing SQL statements sequentially leads to higher cumulative communication latency.
```
nvanbenschoten: > Traditional SQL engines execute the statements sequentially without affecting performance. But in case of globally-distributed CockroachDB clusters, sequential execution of statements adds to aggregate latency

This doesn't really make sense. Traditional SQL engines still take time to execute statements! It's just that they're usually a bit faster than us because of the reason you provide in the next few sentences. This means that they would also benefit from a feature like this, the speedup just wouldn't be as pronounced.

> Executing SQL statements sequentially leads to higher cumulative communication latency.

Executing SQL statements sequentially, especially when each statement requires long distance communication leads to higher cumulative latency.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+<img src="{{ 'images/Sequential_Statement_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution Error Mismatch" style="border:1px solid #eee;max-width:100%" />
+
+Traditional SQL engines execute the statements sequentially without affecting performance. But in case of globally-distributed CockroachDB clusters, sequential execution of statements adds to aggregate latency. This is because to maintain strong consistency across a globally-distributed cluster, CockroachDB replicates writes using the Raft [consensus protocol](https://www.cockroachlabs.com/blog/consensus-made-thrive/). The protocol requires majority of replicas to agree before a write is committed. Replicas always live on separate nodes, and communication between these nodes takes time. Executing SQL statements sequentially leads to higher cumulative communication latency. 
+
+With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements.
```
nvanbenschoten: Short and direct. Nice!

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements. 
+
+## How Parallel Statement Execution Works
+
+SQL statements within a single transaction can be executed in parallel if the statements are independent. CockroachDB considers SQL statements within a single transaction to be independent if their execution can be safely reordered without affecting their results. 
+
+As seen earlier, sequential execution of SQL statements returns a return value immediately from the server. The next statement is executed only after the return value is received. Appending the `RETURNING NOTHING` clause with SQL statements prevents the server from sending return values and allows CockroachDB to execute the statements in parallel. To execute statements in parallel, append the `RETURNING NOTHING` suffix to the SQL statements. The statements are executed in parallel till CockroachDB encounters a **barrier statement**. A barrier statement is a statement that needs to be executed sequentially.
```
nvanbenschoten: "prevents" might have the wrong connotation. Try "stops". I think it's also important to note here somewhere that we return to the client before executing the statement. This is actually the key that allows us to execute multiple statements in parallel.

nvanbenschoten: Also, s/till/until/

nvanbenschoten: > A barrier statement is a statement that needs to be executed sequentially.

This is going to be confusing when people read it first. I'd just say "A barrier statement is any other statement without the `RETURNING NOTHING` suffix"

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+In our example, the transaction would be as follows:
```
nvanbenschoten: s/In our example/In our previous example/

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+> UPDATE movies SET favorite_movie = 'The Matrix' WHERE id = 1 RETURNING NOTHING;
+> UPDATE songs SET favorite_song = 'All this time' WHERE id = 1 RETURNING NOTHING;
+> COMMIT;
+~~~
+
+In this case, because the statements within the transaction are independent of each other, they can be executed in parallel without affecting the results. The COMMIT statement is executed sequentially, and thus acts as a barrier statement. Barrier statements are executed only after all the parallel statements are executed.
+
+The following conceptual diagram shows how the transaction is executed sequentially and in parallel. The diagram also shows how executing statements in parallel reduces the aggregate latency.
+
+<img src="{{ 'images/Parallel_Statement_Normal_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution" style="border:1px solid #eee;max-width:100%" />
```
nvanbenschoten: Do we need the sequential timeline again? It's right above this diagram already.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+~~~ sql
+> BEGIN;
+> UPDATE username SET last_name = 'Smith' WHERE id = 1 RETURNING NOTHING;
+> UPDATE movies SET favorite_movie = 'The Matrix' WHERE id = 1 RETURNING NOTHING;
+> UPDATE songs SET favorite_song = 'All this time' WHERE id = 1 RETURNING NOTHING;
+> COMMIT;
+~~~
+
+In this case, because the statements within the transaction are independent of each other, they can be executed in parallel without affecting the results. The COMMIT statement is executed sequentially, and thus acts as a barrier statement. Barrier statements are executed only after all the parallel statements are executed.
```
nvanbenschoten: Here and below: s/are executed/are finished executing/

v1.1/parallel-statement-execution.md
nvanbenschoten: Side note, the diagrams are great!! Nice job 🎉

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+As stated earlier, barrier statements are executed only after all parallel statements are executed. Thus it may seem as if the barrier statement is taking longer to execute, but it's in fact waiting on the parallel statements. Even then, the total time required for parallel execution of statements followed by the sequential execution of the SQL statement should be less than time required for the sequential execution of all statements. 
+
+Referring to the previous diagram, all UPDATE statements are executed before executing COMMIT. Hence it might seem as if COMMIT is taking longer to execute, but it is, in fact, waiting on the UPDATE statements to be executed.
+
+### Error message mismatch
+
+In case of parallel statement execution, it is possible to see a mismatch in the error message being displayed and the statement being executed. 
+
+Because the `RETURNING NOTHING` clause is used with parallel statements, CockroachDB cannot return the error for the parallel statement. The error message for the parallel statement is then reported on the next barrier statement. Thus you might see a mismatch with the error message to the statement being executed.
```
nvanbenschoten: > Thus you might see a mismatch with the error message to the statement being executed. 

Thus, if any parallel statement encounters an error, you **will** see the error message as the result of the barrier statement.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+Referring to the previous diagram, all UPDATE statements are executed before executing COMMIT. Hence it might seem as if COMMIT is taking longer to execute, but it is, in fact, waiting on the UPDATE statements to be executed.
+
+### Error message mismatch
+
+In case of parallel statement execution, it is possible to see a mismatch in the error message being displayed and the statement being executed. 
+
+Because the `RETURNING NOTHING` clause is used with parallel statements, CockroachDB cannot return the error for the parallel statement. The error message for the parallel statement is then reported on the next barrier statement. Thus you might see a mismatch with the error message to the statement being executed. 
+
+On encountering the error, CockroachDB will stop the execution of all parallel statements and [abort and retry the transaction](transactions.html#error-handling). So even an error message mismatch occurs, the result will be the same as aborting the transaction.
```
nvanbenschoten: `s/So even an error/So even if an error/`

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+Revisiting our sample scenario, suppose an error occurs while executing the second UPDATE statement. In sequential execution, the transaction will be aborted and the error message will be sent immediately. But in case of parallel execution, the transaction will be aborted and error message will be sent after COMMIT. The following diagram illustrates the concept of error message mismatch:
+
+<img src="{{ 'images/Parallel_Statement_Execution_Error_Mismatch.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution Error Mismatch" style="border:1px solid #eee;max-width:100%" />
+
+### `RETURNING NOTHING` clause appended to dependent statements
+
+If two consecutive statements are not independent, and yet a `RETURNING NOTHING` clause is added to the statements, CockroachDB will detect the dependence and serialize the execution. You can thus use the `RETURNING NOTHING` clause with SQL statements without worrying about their dependence.
```
nvanbenschoten: 👍 nice!

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+
+Revising our sample scenario, suppose we want to create a new user on the social networking app. We need to create entries for the last name of the user, their favorite movie, and favorite song. We need to insert entries into three tables: username, movies, and songs. The transaction would be as follows:
+
+~~~ sql
+> BEGIN;
+> id = 2;
```
nvanbenschoten: no need for `> id = 2;`

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+> INSERT INTO username VALUES last_name = 'Pavlo' WHERE id = 2 RETURNING NOTHING;
+> INSERT INTO movies VALUES favorite_movie = 'Godfather' WHERE id = 2 RETURNING NOTHING;
+> INSERT INTO songs VALUES favorite_song = 'Remember' WHERE id = 2 RETURNING NOTHING;
+> COMMIT;
+~~~
+
+In this case, the second and third INSERT statements are dependent on the first INSERT statement. Thus even though we appended the RETURNING NOTHING clause to the first statement, CockroachDB will execute the first statement sequentially. After the first statement is executed, the second and third statements will be executed in parallel. The following conceptual diagram shows how the transaction will be executed in sequential and parallel modes:
```
nvanbenschoten: > are dependent on the first INSERT statement

Add to this "because the `movies` and `songs` tables both have a foreign key constraint on the `username` table".

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,128 @@
...
+## Why Use Parallel Statement Execution
+
+To understand why parallel statement execution is required, let's consider a sample scenario to demonstrate how SQL statements are executed. 
+
+Consider a scenario where a user's last name, favorite movie, and favorite song are to be updated on a social networking application. Suppose the database has three tables: username, movies, and songs, that need to be updated to update the user information. Also suppose that the user's ID is 1. Then the traditional transaction to update the user's information is as follows:
+
+~~~ sql
+> BEGIN;
+> UPDATE username SET last_name = 'Smith' WHERE id = 1;
+> UPDATE movies SET favorite_movie = 'The Matrix' WHERE id = 1;
```
nvanbenschoten: Here and below: for the movies and the songs table, we should use `WHERE user_id = 1`

v1.1/parallel-statement-execution.md
Amruta-Ranade: Will the hybrid case be evident without id=2?

v1.1/parallel-statement-execution.md
Amruta-Ranade: I like the side-by-side comparison diagram that shows the performance improvement. Hence the repetition :)

nvanbenschoten: Ah yeah, that's a very good point.

v1.1/parallel-statement-execution.md
nvanbenschoten: I think so.

v1.1/parallel-statement-execution.md
Amruta-Ranade: I think we can just say "Traditional SQL engines execute statements sequentially. But in case of globally-distributed CockroachDB cluster....". So basically, remove the "without affecting performance" part. Would that work?

v1.1/parallel-statement-execution.md
Amruta-Ranade: About noting that we return to client: I think it's a technical detail that is already evident from the diagram. Spelling it out might distract from the main point, which is we stop server from sending return values, thereby allowing parallel execution. Thoughts?

v1.1/parallel-statement-execution.md
Amruta-Ranade: Can we say "Thus, if any parallel statement encounters an error, you will see the error message after the barrier statement"?

v1.1/parallel-statement-execution.md
Amruta-Ranade: Hmm..I agree..let's check with @jseldess though.

jseldess: Yep, removing those brackets works for me, too!

v1.1/parallel-statement-execution.md
nvanbenschoten: "after" isn't the right word though. It needs to be super clear that the error message will always be returned **as the result of** the barrier statement.

Amruta-Ranade: "As the result of" is eerily similar to "as a result of". It could be misread as the error message is the result of the barrier statement, which isn't the case. "After" seems more appropriate in that context.

v1.1/parallel-statement-execution.md
nvanbenschoten: I think this is the main point though. The client can't send the next statement until it hears something from the server (that's why the conversation analogy is so good!). I would change `the client sends` to `the client can then send` here and below. The reason for this is subtle but kind of important. The reader of these docs is the client. We're describing the contract we (the server) are making with their application (the client).

> sequential execution of SQL statements returns a return value immediately from the server

It's not immediate, it's after executing the statement, right?

Amruta-Ranade: I reworded the section a bit:
As seen earlier, sequential execution of SQL statements returns a return value after executing the statement. The client can send the next statement to be executed only after it receives the return value from the server. In CockroachDB, appending the `RETURNING NOTHING` clause with SQL statements causes the server to send an acknowledgement immediately instead of completing the execution and then sending the return value to the client. The client can then send the next statement to be executed on receiving the acknowledgement. This allows CockroachDB to execute the statements in parallel. The statements are executed in parallel until CockroachDB encounters a **barrier statement**. A barrier statement is any statement without the `RETURNING NOTHING` suffix and is executed sequentially.

Does that seem clearer?

Amruta-Ranade:
@jseldess The document's ready for review :)

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+
+CockroachDB supports parallel execution of independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single [transaction](https://www.cockroachlabs.com/docs/dev/transactions.html). Executing statements in parallel helps reduce aggregate latency and improve performance. 
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+To understand why parallel statement execution is required, let's first understand sequential execution of SQL statements.
+
+Consider a scenario where we want to update a user's last name, favorite movie, and favorite song on a social networking application. Suppose the database has three tables: users, favorite_movies, and favorite_songs, that need to be updated to update the user information. Then the traditional transaction to update the user's information is as follows:
```
jseldess: @Amruta-Ranade, I think we need to be even more direct in this section and shouldn't lean so heavily on an example to explain "why". 

Perhaps we could repurpose/condense the content in this section to something like this:

> SQL engines traditionally execute the statements in a transaction sequentially: The server executes one statement to completion, sends the return value to the client, and only then starts the next statement, as demonstrated by the following conceptual diagram: 
>
> DIAGRAM
>
> In the case of a traditional single-node SQL database, execution of statements never leaves the single machine and so there isn't much communication latency. However, in the case of a multi-node replicated database like CockroachDB, execution of statements can span multiple machines, with communication and coordination between machines resulting in a higher cumulative latency. In such a case, executing SQL statement sequentially can add even more latency.
>
> THEN NEED TO END WITH WHY USE PARALLEL.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+
+With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements. 
+
+## How Parallel Statement Execution Works
+
+SQL statements within a single transaction can be executed in parallel if the statements are independent. CockroachDB considers SQL statements within a single transaction to be independent if their execution can be safely reordered without affecting their results.
```
jseldess: If we remove the scenario from "why" above, we'll need to lead into it in this section. Perhaps this is the section to the start with something like:

> To understand how parallel statement execution works, let's consider a scenario:
> - We want to update a user's last name, favorite movie, and favorite song on a social networking application.
> - The database has three tables that need to be updated: `users`, `favorite_movies`, and `favorite_songs`. 

And then intro the sequential/parallel diagrams and talk them through.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Error message mismatch
+
+In case of parallel statement execution, you might see a mismatch in the error message being displayed and the statement being executed. 
+
+Because the `RETURNING NOTHING` clause is used with parallel statements, CockroachDB cannot return the error for the parallel statement. The server reports the error message for the parallel statement after the next barrier statement. 
+
+On encountering the error, CockroachDB stops the execution of all parallel statements and [aborts and retries the transaction](transactions.html#error-handling). So even if an error message mismatch occurs, the result is the same as aborting the transaction.
+
+Revisiting our sample scenario, suppose an error occurs while executing the second `UPDATE` statement. In sequential execution, the transaction is aborted and the error message is sent immediately. However, in case of parallel execution, the transaction is aborted and the error message is sent after `COMMIT`. The following diagram illustrates the concept of error message mismatch:
```
jseldess: I think this paragraph actually has the key points. We might want to start with this and condense some of the rest:

> With sequential execution, as soon as an error happens, the transaction is aborted and an error message is sent to the client. However, with parallel execution, this happens not when the error is encountered but rather after the next barrier statement. This can result in the client receiving an error message that doesn't match the statement being executed. The following diagram illustrates this concept:
>
> DIAGRAM

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Why Use Parallel Statement Execution
+
+SQL engines traditionally execute the SQL statements sequentially. The server executes each statement to completion and sends the return value of each statement to the client. Only after the client receives the return value of a statement, it sends the next SQL statement to be executed. 
+
+In the case of a traditional single-node SQL database, statements are executed on the single machine, and so the execution does not result in any communication latency. However, in the case of a globally-distributed and replicated database like CockroachDB, execution of statements can span multiple nodes. The coordination between nodes results in communication latency. Executing SQL statements sequentially results in higher cumulative latency.
+
+With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements. 
+
+## When to use Parallel Statement Execution
```
jseldess: nit: capitalize "Use"

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+toc: false
+---
+
+CockroachDB supports parallel execution of independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single [transaction](https://www.cockroachlabs.com/docs/dev/transactions.html). Executing statements in parallel helps reduce aggregate latency and improve performance. 
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+SQL engines traditionally execute the SQL statements sequentially. The server executes each statement to completion and sends the return value of each statement to the client. Only after the client receives the return value of a statement, it sends the next SQL statement to be executed.
```
jseldess: In the first sentence, let's be more explicit:
> SQL engines traditionally execute the SQL statements in a transaction sequentially.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+CockroachDB supports parallel execution of independent [`INSERT`](insert.html), [`UPDATE`](update.html), [`UPSERT`](upsert.html), and [`DELETE`](delete.html) statements within a single [transaction](https://www.cockroachlabs.com/docs/dev/transactions.html). Executing statements in parallel helps reduce aggregate latency and improve performance. 
+
+<div id="toc"></div>
+
+## Why Use Parallel Statement Execution
+
+SQL engines traditionally execute the SQL statements sequentially. The server executes each statement to completion and sends the return value of each statement to the client. Only after the client receives the return value of a statement, it sends the next SQL statement to be executed. 
+
+In the case of a traditional single-node SQL database, statements are executed on the single machine, and so the execution does not result in any communication latency. However, in the case of a globally-distributed and replicated database like CockroachDB, execution of statements can span multiple nodes. The coordination between nodes results in communication latency. Executing SQL statements sequentially results in higher cumulative latency.
```
jseldess: I still think we should change "globally-distribute" to just "distributed". Most of our users aren't going to be running a globally distributed cluster.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+With parallel statement execution, however, multiple SQL statements within a transaction are executed at the same time, thereby reducing the aggregate latency over multiple SQL statements.
```
jseldess: Let's remove "over multiple SQL statements" at the end. We already say "multiple SQL statements" earlier in the sentence.

v1.1/parallel-statement-execution.md
jseldess: Also, I think this should be combined with the section on Independent Statements. See my comment in that section.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+> INSERT INTO favorite_movies VALUES movies = 'Godfather' WHERE user_id = 2 RETURNING NOTHING;
+> INSERT INTO facvorite_songs VALUES songs = 'Remember' WHERE user_id = 2 RETURNING NOTHING;
+> COMMIT;
+~~~
+
+In this case, the second and third `INSERT` statements are dependent on the first `INSERT` statement because the movies and songs tables both have a foreign key constraint on the users table. So even though we append the `RETURNING NOTHING` clause to the first statement, CockroachDB executes the first statement sequentially. After the first statement is executed to completion, the second and third `INSERT` statements are executed in parallel. The following conceptual diagram shows how the transaction is executed in sequential and parallel modes:
+
+<img src="{{ 'images/Parallel_Statement_Hybrid_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Hybrid Execution" style="border:1px solid #eee;max-width:100%" />
+
+## Independent SQL statements
```
jseldess: Let's make this section "When to Use Parallel Statement Execution" and start with:

> SQL statements within a single transaction can be executed in parallel if the statements are independent. CockroachDB considers SQL statements within a single transaction to be independent if their execution can be safely reordered without affecting their results. 
>
> For example, ...

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+<img src="{{ 'images/Sequential_Statement_Execution.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution" style="border:1px solid #eee;max-width:100%" />
+
+The SQL statements in our sample scenario can be executed in parallel since they are independent of each other. To execute statements in parallel, the client should be able to send the next statement to be executed without waiting for the return value of the earlier statement. In CockroachDB, on appending the `RETURNING NOTHING` clause with SQL statements,  the server sends an acknowledgment immediately, instead of waiting to complete the statement execution and sending the return value to the client. The client sends the next statement to be executed on receiving the acknowledgment. This allows CockroachDB to execute the statements in parallel. The statements are executed in parallel until CockroachDB encounters a **barrier statement**. A barrier statement is any statement without the `RETURNING NOTHING clause. The server executes a barrier statement sequentially.
```
jseldess: Need a closing backtick after the last `RETURNING NOTHING`.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Perceived delay in execution of barrier statements 
+
+As stated earlier, the server executes a barrier statement only after all the preceding parallel statements have finished executing. So it may seem as if the barrier statement is taking longer to execute, but it is waiting on the parallel statements. Even then, the total time required for parallel execution of statements followed by the sequential execution of the barrier statement should be less than the time required for the sequential execution of all statements. 
+
+Referring to the previous diagram, the server executes all `UPDATE` statements to completion before executing `COMMIT`. Hence it might seem as if `COMMIT` is taking longer to execute, but it is, in fact, waiting on the `UPDATE` statements to finish executing.
+
+### Error message mismatch
+
+In case of parallel statement execution, you might see a mismatch in the error message being displayed and the statement being executed.
```
jseldess: Let's remove this sentence. It's a repetition of what's said in the next paragraph.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+In case of parallel statement execution, you might see a mismatch in the error message being displayed and the statement being executed. 
+
+With sequential execution, as soon as an error happens, the transaction is aborted and an error message is sent to the client. However, with parallel execution, the message is sent not when the error is encountered but after the next barrier statement. This can result in the client receiving an error message that doesn't match the statement being executed. The following diagram illustrates this concept:
+
+<img src="{{ 'images/Parallel_Statement_Execution_Error_Mismatch.png' | relative_url }}" alt="CockroachDB Parallel Statement Execution Error Mismatch" style="border:1px solid #eee;max-width:100%" />
+
+### `RETURNING NOTHING` clause appended to dependent statements
+
+If two consecutive statements are not independent, and yet a `RETURNING NOTHING` clause is added to the statements, CockroachDB detects the dependence and executes the statements sequentially. You can use the `RETURNING NOTHING` clause with SQL statements without worrying about their dependence.
```
jseldess: Need a transition between these two sentence:

> This means that you can use the `RETURNING NOTHING`...

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+Revising our sample scenario, suppose we want to create a new user on the social networking app. We need to create entries for the last name of the user, their favorite movie, and favorite song. We need to insert entries into three tables: users, favorite_movies, and favorite_songs. The transaction would be as follows:
```
jseldess: Need to put backticks around the table names.

v1.1/parallel-statement-execution.md
```diff
@@ -0,0 +1,131 @@
...
+
+~~~ sql
+> UPDATE a SET y = true  WHERE y = false;
+> UPDATE a SET y = false WHERE y = true;
+~~~
+
+
+{{site.data.alerts.callout_info}}Parallel statement execution in CockroachDB is different than parallel query execution in PostgreSQL. For PostgreSQL, parallel query execution refers to “creating multiple query processes that divide the workload of a single SQL statement and executing them in parallel”. For CockroachDB’s parallel statement execution, an individual SQL statement is not divided into processes. Instead, multiple independent SQL statements within a single transaction are executed in parallel.{{site.data.alerts.end}}
+
+## See Also
```
jseldess: Let's also link to `[Transactions](transactions.html)`.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

v1.1/parallel-statement-execution.md
Amruta-Ranade: Fixed.

jseldess:
Thanks for all your efforts here, @Amruta-Ranade! Merging now.

nvanbenschoten:
One last question: do we want to go back at some point in the future and style these diagrams? It certainly doesn't need to be immediate, but I can see us wanting to have consistent style/branding across all diagrams in our docs.

jseldess:
@nvanbenschoten, yes, good point. I have another PR (on node decommissioning) with a bunch of diagrams. I've opened an issue for us to work with Design at some point to come up with a consistent style and approach: https://github.com/cockroachdb/docs/issues/1886

PR #1803, opened 2017-08-09
Remove limitation about on conflict and returning

### LINKED ISSUE #1693: sql: UPSERT and INSERT ON CONFLICT with RETURNING
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwdV1sM5KEK2L14K/recfh5aGbH5i9kwn9
PR: https://github.com/cockroachdb/cockroach/pull/17149

jseldess:
Fixes #1693


jseldess:
@couchand, this just involved removing text stating that on conflict with returning is not supported. I don't see the need for anything else. Do you?

@Amruta-Ranade, the language around `returning nothing` in the `upsert` and `insert` docs, and probably in other sql statement docs, needs to be revisited once your [Parallel Statement Execution PR](https://github.com/cockroachdb/docs/pull/1790) is further along.

couchand:
@jseldess LGTM

I'll also draw your attention to https://github.com/cockroachdb/cockroach/pull/17435, which I don't think needs anything in particular, since we hadn't documented the missing functionality anyway.  But it might be nice, for those unfamiliar with the details of PGSQL and uninclined to dig into the SQL syntax page, to point out that the `DO UPDATE` can have a `WHERE` filter after the `SET` clause.

jseldess:
TFTR, @couchand. I also added an example showing `DO UPDATE SET` with a `WHERE` filter and called it out as new in 1.1.

jseldess:
TC only failing on what should be valid links to old cockroach prs. Overriding TC and merging in the hope that the GitHub hiccup will pass.

Issue #119, opened 2016-03-15
certs: provide openssl commands to generate cockroach-compatible certs.


mberhault:
For each `cockroach cert` command, we need the  equivalent openssl command.
This will provide clarity into what the cert commands do, allow paranoid users to use a familiar tool instead, allow easier integration with existing certs-generating tools, and allow more complex scenarios (eg: real CA instead of self-signed).

jjzazuet:
Yes, please. Thanks!

jseldess:
@mberhault, are you still planning to draft up docs here? If so, any updates?

sploiselle:
@mberhault I know OpenSSL is maybe-the-least-fun thing in the world to work with (_"I'd rather write in binary!"_, I hear you say)––is there any chance I could get even a dump of a list of commands you might use that I could fashion into documentation, though?

mberhault:
I've started working on that. The PR listed above shows how to generate the CA cert and key with the same fields (almost). I'll add node and client certs to that as well.

mberhault:
The PR above has openssl equivalents for all `cockroach cert` commands. You'll notice that it needs three config files and 5 fairly verbose commands (8 if you include key creation, but in the example I just reuse the cockroach ones to make certificate comparison better).

I'm happy to add comments clarifying everything, and I definitely can't claim that this is the cleanest or shortest way to do it. @bdarnell: how's your openssl-fu?

mberhault:
Moved PR to #1804

bdarnell:
>  @bdarnell: how's your openssl-fu?

In this area, just cargo-culting. I think when I've done this in the past I've used command-line args instead of config files for the `openssl req` command, so the CA is the only thing that needs a config file.

mberhault:
Yeah, I tried going that way, but then you end up sticking the DN and altSubjectName in ugly-looking args.

jseldess:
@Amruta-Ranade, here's the issue to document the openssl commands for cockroach-compatible certs. See also Marc's draft PR linked above.

Amruta-Ranade:
Docs that might need to be updated:

- Create security certificates
- Rotate certificates
- Start a local cluster (secure)
- Create user
- Create and manage users
- Manual deployment
- Deploy CockroachDB on AWS, Azure, Digital Ocean, GCE

jjzazuet:
[...]

PR #1804, opened 2017-08-09
draft: generate cockroach certs using openssl

mberhault:
This PR contains the commands and config files necessary to generate
nearly-identical certificates using `cockroach cert` and `openssl`.

This is needed for #119

We generate all certs using cockroach:
* CA cert/key
* node cert/key
* client cert/key for user `maxroach`

We then generate the same certificates using openssl, but we reuse the
cockroach keys to make diffs smaller:
1. generate CA certificate
1. generate node CSR (certificate signing request)
1. sign node CSR resulting in node certificate
1. generate client CSR
1. sign client CSR resulting in client certificate

There are various config files used for each step:
* ca.conf for 1) 3) and 5)
* node.conf for 2)
* client.conf for 4)

### DIFF
--- a/openssl/ca.conf
+++ b/openssl/ca.conf
@@ -1,0 +1,38 @@
+[ ca ]
+default_ca = CA_default
+
+[ CA_default ]
+default_days = 3660
+database = index.txt
+serial = serial.txt 
+default_md = sha256 
+copy_extensions = copy
+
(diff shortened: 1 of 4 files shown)


jseldess:
Closing in favor of https://github.com/cockroachdb/docs/pull/1944.

PR #1808, opened 2017-08-10
Add some late additions to 20170803 release notes

bdarnell:
Covers
https://github.com/cockroachdb/cockroach/compare/64be211fa...1e88c32588fd3d7cad826437c18b349fd70e4762

Only adding the highlights since the release is already done.


jseldess:
Bypassing TC, since the broken links are addressed in https://github.com/cockroachdb/docs/pull/1763.

PR #1814, opened 2017-08-10
Add response field descriptions and remove --deps flag

### LINKED ISSUE #1081: cli: update cockroach version docs to cover Distribution, Build SHA-1, and Build Type
PR: https://github.com/cockroachdb/cockroach/pull/13102

### LINKED ISSUE #1812: Remove --deps flag from cockroach version doc
Re: [View Version Details](https://cockroachlabs.com/docs//v1.0/view-version-details.html)

## Issue Description
`cockroach version --deps`
```
Usage:
  cockroach version [flags]

Global Flags:
      --log-backtrace-at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log-dir string                   if non-empty, write log files in this directory
      --log-dir-max-size bytes           maximum combined size of all log files (default 100 MiB)
      --log-file-max-size bytes          maximum size of each log file (default 10 MiB)
      --log-file-verbosity Severity      minimum verbosity of messages written to the log file (default INFO)
      --logtostderr Severity[=DEFAULT]   logs at or above this threshold go to stderr (default NONE)
      --no-color                         disable standard error log colorization
      --verbosity level                  log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Error: unknown flag: --deps
Failed running "version"
```
`cockroach version` 

```
Build Tag:    v1.0.4
Build Time:   2017/07/27 17:54:36
Distribution: CCL
Platform:     linux amd64
Go Version:   go1.8.3
C Compiler:   gcc 6.3.0
Build SHA-1:  5b757262d33d814bda1deb2af20161a1f7749df3
Build Type:   release
```

jseldess:
Fixes #1812 
Fixes #1081

### DIFF
--- a/v1.0/view-version-details.md
+++ b/v1.0/view-version-details.md
@@ -6,13 +6,20 @@ toc: false
 
 To view version details for a specific `cockroach` binary, run the `cockroach version` [command](cockroach-commands.html):
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ cockroach version
+~~~
+
+~~~
 Build Tag:   {{page.release_info.version}}
 Build Time:  {{page.release_info.build_time}}
-Platform:    darwin amd64
-Go Version:  go1.7.1
-C Compiler:  4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)
+Distribution: CCL
+Platform:     darwin amd64
+Go Version:   go1.8.3
+C Compiler:   4.2.1 Compatible Clang 3.8.0 (tags/RELEASE_380/final)
+Build SHA-1:  5b757262d33d814bda1deb2af20161a1f7749df3
+Build Type:   release
 ~~~
 
 The `cockroach version` command outputs the following fields:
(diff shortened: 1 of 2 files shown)


jseldess:
@mberhault or @tamird , I can't figure out when/if the `Build Type` field will be [`release-gnu`](https://github.com/cockroachdb/cockroach/blob/8ac54961ab8edc641e41880f8e410032dc1843eb/Makefile#L114) or [`release-musl`](https://github.com/cockroachdb/cockroach/blob/8ac54961ab8edc641e41880f8e410032dc1843eb/Makefile#L116) in place of just `release`. Can you help clarify?

tamird:
@jseldess the `Build Type` field value will be `release` on any platform that isn't Linux. On Linux, it is `release-gnu` when the binary requires [glibc](https://en.wikipedia.org/wiki/GNU_C_Library) at runtime (glibc is present in almost all Linux distributions, with Alpine Linux being the notable exception), and `release-musl` when the binary is linked against [musl](https://en.wikipedia.org/wiki/Musl) instead of glibc. Since musl can be fully statically linked (glibc [can't](https://sourceware.org/bugzilla/show_bug.cgi?id=19341)), that is indeed how we link it, which means that those binaries can run on _any_ Linux distribution. glibc is much more widely used than musl, so we are conservative in promoting it, which is why our official Linux binaries are `release-gnu`.

As a side note, I discovered that we did have a request for official fully static (meaning musl) builds which is still open: https://github.com/cockroachdb/cockroach/issues/7261. Since those binaries are now being built, all that remains is to link to them from the docs.

mberhault:
LGTM. I don't think we need to go into too much detail on build types here, dropping the no-longer-existing `--deps` is the critical part.

jseldess:
Thanks, @tamird and @mberhault, I'll wrap this up soon.

jseldess:
Updated field definition. Opened new issue to link to static builds.

PR #1816, opened 2017-08-11
Change go txn sample to use updated crdb.ExecuteTx signature

twrobel3:
The cockroach-go `ExecuteTxn` signature was changed in a backward-incompatible way in https://github.com/cockroachdb/cockroach-go/pull/41, breaking the transaction example.

This change updates the sample transaction code to use the new signature.

### DIFF
--- a/_includes/app/txn-sample.go
+++ b/_includes/app/txn-sample.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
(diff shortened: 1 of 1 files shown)


twrobel3:
@jseldess FYI

cc @andreimatei and @tamird who reviewed the change in cockroach-go

jseldess:
LGTM, @twrobel3. Thanks for the update. And sorry for the delay.

PR #1837, opened 2017-08-17
Created document for Array data type, updated related files

### LINKED ISSUE #1561: sql: Enable storing non-PK ARRAY columns
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwdV1sM5KEK2L14K/recIfrzTv0P5w7Gf7
RFC: https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/array_encoding.md
PRs:
- cockroachdb/cockroach#16764
- https://github.com/cockroachdb/cockroach/pull/17318
- https://github.com/cockroachdb/cockroach/pull/17331
- https://github.com/cockroachdb/cockroach/pull/17489
- https://github.com/cockroachdb/cockroach/pull/17455 (sql: support <type> ARRAY syntax in CREATE

Amruta-Ranade:
Closes #1561 

- Created document for Array data type
- Updated sidebar navigation to include Array in Data Types category
- Updated Bytes document as per discussion on Slack about increasing size to 1 MB.
- Updated Functions and Operators document with array functions.

### DIFF
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -693,7 +693,7 @@
             ]
           },
           {
-            "title": "<code>ARRAYS</code>",
+            "title": "<code>ARRAY</code>",
             "urls": [
               "/${VERSION}/array.html"
             ]
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -691,6 +691,12 @@
             "urls": [
               "/${VERSION}/bytes.html"
             ]
+          },
+          {
+            "title": "<code>ARRAYS</code>",
+            "urls": [
+              "/${VERSION}/array.html"
+            ]
           }
         ]
       },
--- a/_includes/sql/v1.1/functions.md
+++ b/_includes/sql/v1.1/functions.md
@@ -260,17 +260,6 @@ Function &rarr; Returns | Description
 <code>translate(input: <a href="string.html">string</a>, find: <a href="string.html">string</a>, replace: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">In `input`, replaces the first character from `find` with the first character in `replace`; repeat for each character in `find`. <br/><br/>For example, `translate('doggie', 'dog', '123');` returns `1233ie`.</span>
 <code>upper(val: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">Converts all characters in `val`to their to their upper-case equivalents.</span>
 
-### Array Functions
-
-Function &rarr; Returns | Description
---- | ---
-<code>array_append(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Appends the specified element to the end of the array.
-<code>array_prepend(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Appends the specified element to the beginning of the array.
-<code>array_cat(input: anyarray[], anyarray[])&rarr;anyarray[]</code> | <span class="funcdesc"> Concatenates arrays.
-<code>array_remove(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Removes the specified element from the array.
-<code>array_replace(input: anyarray[], anyelement, anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Replaces the specified element with the new element.
-<code>array_position(input: anyarray[], anyelement)&rarr;int</code> | <span class="funcdesc"> Returns the position of the specified element in the array.
-
 ### System Info Functions
 
 Function &rarr; Returns | Description
--- a/_includes/sql/v1.1/functions.md
+++ b/_includes/sql/v1.1/functions.md
@@ -260,6 +260,17 @@ Function &rarr; Returns | Description
 <code>translate(input: <a href="string.html">string</a>, find: <a href="string.html">string</a>, replace: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">In `input`, replaces the first character from `find` with the first character in `replace`; repeat for each character in `find`. <br/><br/>For example, `translate('doggie', 'dog', '123');` returns `1233ie`.</span>
 <code>upper(val: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">Converts all characters in `val`to their to their upper-case equivalents.</span>
 
+### Array Functions
+
+Function &rarr; Returns | Description
+--- | ---
+<code>array_append(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Appends the specified element to the end of the array.
+<code>array_prepend(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Appends the specified element to the beginning of the array.
+<code>array_cat(input: anyarray[], anyarray[])&rarr;anyarray[]</code> | <span class="funcdesc"> Concatenates arrays.
+<code>array_remove(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Removes the specified element from the array.
+<code>array_replace(input: anyarray[], anyelement, anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Replaces the specified element with the new element.
+<code>array_position(input: anyarray[], anyelement)&rarr;int</code> | <span class="funcdesc"> Returns the position specified element in the array.
+
 ### System Info Functions
 
 Function &rarr; Returns | Description
--- a/_includes/sql/v1.1/functions.md
+++ b/_includes/sql/v1.1/functions.md
@@ -269,7 +269,7 @@ Function &rarr; Returns | Description
 <code>array_cat(input: anyarray[], anyarray[])&rarr;anyarray[]</code> | <span class="funcdesc"> Concatenates arrays.
 <code>array_remove(input: anyarray[], anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Removes the specified element from the array.
 <code>array_replace(input: anyarray[], anyelement, anyelement)&rarr;anyarray[]</code> | <span class="funcdesc"> Replaces the specified element with the new element.
-<code>array_position(input: anyarray[], anyelement)&rarr;int</code> | <span class="funcdesc"> Returns the position specified element in the array.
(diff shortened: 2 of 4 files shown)


v1.1/array.md
```diff
@@ -0,0 +1,64 @@
+---
+title: ARRAYS
+summary: The ARRAYS data type stores 
+toc: false
+---
+
+The `ARRAY` data type creates one-dimensional, 1-indexed, homogenous arrays of any non-array [data type](data-types.html).
```
justinj: I don't know how we word this usually, so feel free to ignore this, but it might be more correct to phrase this as "The `ARRAY` data type *is* a one-dimensional, ..."

v1.1/array.md
```diff
@@ -0,0 +1,64 @@
...
+
+<div id="toc"></div>
+
+{{site.data.alerts.callout_info}} CockroachDB does not support nested arrays, indexing on arrays, and ordering by arrays.{{site.data.alerts.end}}
```
justinj: This is fine, but just as a heads up it's caused some confusion during development that the word "indexing" here has two meanings - creating a database index (the meaning you're referring to here) and also indexing *into* an array (`SELECT (ARRAY[1,2,3])[2]`). I don't know if that's a concern for you but I vaguely suspect it could be a source of confusion, particularly among users new to databases.

_includes/sql/v1.1/functions.md
```diff
@@ -150,6 +150,17 @@ Function &rarr; Returns | Description
 <code>translate(input: <a href="string.html">string</a>, find: <a href="string.html">string</a>, replace: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">In `input`, replaces the first character from `find` with the first character in `replace`; repeat for each character in `find`. <br/><br/>For example, `translate('doggie', 'dog', '123');` returns `1233ie`.</span>
 <code>upper(val: <a href="string.html">string</a>) &rarr; <a href="string.html">string</a></code> | <span class="funcdesc">Converts all characters in `val`to their to their upper-case equivalents.</span>
 
+### Array Functions
```
justinj: This doesn't mention `||` as an append operator. All of its features can be used by using one of the functions below, but IMO it's the most pleasant way to work with arrays and we should probably highlight it somewhere. I guess it doesn't go here with the functions though, not sure of the right place for it.

v1.1/array.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Syntax
+A column of data type `ARRAY` can be created in the following ways:
```
knz: Indeed the same distinction between "columns" and "values" exist for arrays and should be worded as such.

justinj:
I actually thing "keeping the array data closer to the row itself" is probably not a huge selling point for arrays in Cockroach, given that interleaved tables also solve that problem (and all else equal, we would probably push people towards that option, rather than arrays). Avoiding joins is more compelling, as is compatibility (though I'm not sure compatibility is something to include on a docs page).

I can't reply to your specific comment for some reason, but "1-indexed" means the array indices start from 1, rather than 0, (so `(ARRAY[1,2,3])[1]` is `1`, not `0`), and homogenous does indeed mean that all values have the same type. These are both the technical terms for these concepts but I certainly wouldn't object to more accessible terminology!

_includes/sql/v1.1/functions.md
Amruta-Ranade: @justinj Yes, I am trying to figure out the right place for it as well. Let me think about it a bit more.

v1.1/array.md
Amruta-Ranade: @jseldess Yes, that's what we mean. I assume readers would know what it means, since we don't elaborate on it in the value expressions docs either. Thoughts, @justinj ?

Amruta-Ranade:
@jseldess @justinj About the comment: "we should find a way to communicate the value of this data type. Based on the roadmap description, users can "use array columns to improve performance by avoiding joins and keeping the array data closer to the row itself.""

The way I understand it, ARRAY data type in PostGreSQL does this already. So it is not a new value-add we are providing. As Justin said, we are making it compatible with PostGreSQL. So the doc need not point out the value of the data type. Let me know what you think.

knz:
To clarify, **arrays are actually a rather bad idea in SQL** -- it's would not be entirely incorrect to say that *we should strongly discourage users from using them unless they need them for whatever compatibility reasons*. 

The answer to "we should find a way to communicate the value of this data type" is "we should communicate as little as possible so that users do not notice them too much. This documentation page should be strictly minimal - explain just enough that we can say it's documented, but not anything more!"

v1.1/array.md
justinj: That wording sounds find to me! Copying my comment from below since I couldn't comment here the other day for some reason:

>"1-indexed" means the array indices start from 1, rather than 0, (so (ARRAY[1,2,3])[1] is 1, not 2), and homogenous does indeed mean that all values have the same type. These are both the technical terms for these concepts but I certainly wouldn't object to more accessible terminology!

v1.1/array.md
Amruta-Ranade: Fixed.

_includes/sidebar-data-v1.1.json
```diff
@@ -673,6 +673,12 @@
             "urls": [
               "/${VERSION}/bytes.html"
             ]
+          },
+          {
+            "title": "<code>ARRAYS</code>",
```
Amruta-Ranade: Fixed.

v1.1/array.md
```diff
@@ -0,0 +1,64 @@
...
+toc: false
+---
+
+The `ARRAY` data type creates one-dimensional, 1-indexed, homogenous arrays of any non-array [data type](data-types.html).
+
+<div id="toc"></div>
+
+{{site.data.alerts.callout_info}} CockroachDB does not support nested arrays, indexing on arrays, and ordering by arrays.{{site.data.alerts.end}}
+
+## Syntax
```
Amruta-Ranade: Fixed.

v1.1/array.md
```diff
@@ -0,0 +1,64 @@
...
+A column of data type `ARRAY` can be created in the following ways:
+
+
+- Appending square brackets ([]) to any non-array [data type](data-types.html) in column definition.
```
Amruta-Ranade: Fixed.

v1.1/array.md
Amruta-Ranade: Reworded as suggested. 
Also, I don't think the content in the [value expressions doc](https://www.cockroachlabs.com/docs/dev/sql-expressions.html#array-constructors) has been affected by this PR. @justinj Can you verify please?

v1.1/array.md
jseldess: I was just not familiar enough with these terms. A quick search of `1-indexed` matches the intended meaning here. And "homogenous" is also fine, as long as we think it's clear to users that a array can contain only values of the same non-array type.

For the "1-indexed" aspect, I think we should add another example showing how to get a value out of an array based on its index position. @justinj, perhaps you already have such an example?

Amruta-Ranade: @jseldess Found this in the test file, I think this example would work:
SELECT (ARRAY['a', 'b', 'c'])[2]
----
b

In the similar vein, can we add an example for the append (with ||) functionality? I know we normally don't include examples of functions in the data type docs, but can we make an exception?

v1.1/array.md
```diff
@@ -0,0 +1,131 @@
...
++---------------------------+
+| (ARRAY['a', 'b', 'c'])[2] |
++---------------------------+
+| b                         |
++---------------------------+
+(1 row)
+~~~
+
+### Appending an element to an array
+#### Using the array_append function
```
jseldess: Insert empty line between the h3 and h4 headings.

v1.1/array.md
```diff
@@ -0,0 +1,131 @@
...
+~~~
++------------+
+|     d      |
++------------+
+| {10,20,30} |
++------------+
+(1 row)
+~~~
+
+### Accessing an array element using array index
```
jseldess: Nice! My only comment here is that it might be even nicer to show selecting a specific array element from an array column in a table, say the table created in the previous example:

```
> SELECT d[2] FROM c;
```

```
+------+
| d[2] |
+------+
|   20 |
+------+
(1 row)
```

v1.1/array.md
```diff
@@ -0,0 +1,131 @@
...
+~~~
++---------------------------+
+| (ARRAY['a', 'b', 'c'])[2] |
++---------------------------+
+| b                         |
++---------------------------+
+(1 row)
+~~~
+
+### Appending an element to an array
```
jseldess: Just chatted with @justinj about this: Let's change these `select` examples  of actually updating on-disk arrays, which is something people might more often do. 

**For the `array_append()` function:**

```
> SELECT * FROM d;
```

```
+------------+
|     d      |
+------------+
| {10,20,30} |
+------------+
(1 row)
```

```
> UPDATE c SET d = array_append(d, 40) WHERE d[3] = 30;
```

```
> SELECT * FROM c;
```

```
+---------------+
|       d       |
+---------------+
| {10,20,30,40} |
+---------------+
(1 row)
```

**Then for the `||` operator:**

```
> SELECT * FROM d;
```

```
+---------------+
|       d       |
+---------------+
| {10,20,30,40} |
+---------------+
(1 row)
```

```
> UPDATE c SET d = d || 50 WHERE d[4] = 40;
```

```
> SELECT * FROM c;
```

```
+------------------+
|        d         |
+------------------+
| {10,20,30,40,50} |
+------------------+
(1 row)```

File: v1.1/array.md (hunk 8 above)
Comment (jseldess) [reply v1.1/array.md]: Also, let's code-format the function name: `array_append`

File: v1.1/array.md [hunk 11]
```diff
@@ -0,0 +1,131 @@
+---
+title: ARRAYS
+summary: The ARRAYS data type stores 
+toc: false
+---
+
+<span class="version-tag">New in v1.1:</span>The `ARRAY` data type stores one-dimensional, 1-indexed, homogenous arrays of any non-array [data type](data-types.html).
```
jseldess: In response to @knz note that arrays aren't great in sql, we could add a second paragraph like:

> The `ARRAY` type is useful for ensuring compatibility with ORMs and other tools. However, if such compatibility is not a concern, it's more flexible to design your schema with normalized tables.

@knz, does that sound ok?

v1.1/array.md
Amruta-Ranade: Fixed.

v1.1/array.md
Amruta-Ranade: Fixed.

v1.1/array.md
Amruta-Ranade: Fixed.

v1.1/array.md
Amruta-Ranade: Fixed.

PR #1862, opened 2017-08-24
Add event tracking data attributes

deronjohnson:
@jseldess 

Adds tracking attributes to mac and linux archive links and code snippet copy buttons. I had to inline the copy to clipboard partial so that we could add unique attributes.


jseldess:
Thanks, @deronjohnson. I just ported these changes over to https://github.com/cockroachdb/docs/pull/1763, to make things easier. So I'll close this.

PR #1875, opened 2017-09-01
Diagram generator improvements + new diagrams

### LINKED ISSUE #1806: SQL diagrams missing, too broad, or too narrow on 1.1 pages
So far I've seen this on the [`UPSERT`](https://cockroachlabs.com/docs/dev/show-cluster-setting.html) and pretty much all of the `SHOW` pages, i.e., [`SHOW CLUSTER SETTINGS`](https://cockroachlabs.com/docs/dev/show-cluster-setting.html)
 
screen shot 2017-08-09 at 6 17 22 pm

knz:
Needs https://github.com/cockroachdb/cockroach/pull/18144

Maybe fixes #1806?

@benesch plz double check 4th commit


sploiselle:
This looks great, @knz. As far as I can tell, this also closes @jseldess 's issue in #1806. Thanks for addressing this so quickly!

One thing is that some element names have changed. For example in the `BACKUP` doc, what used to be `non_reserved_word_or_sconst` is now `string_or_placeholder`. Is there any way we can reliably find these mismatches?

mjibson:
Regarding the first commit: the java throttling is done at https://github.com/cockroachdb/docs/blob/e5b76e3d140807c877699384a795739a144c7cdf/generate/extract/extract.go#L35

I think it makes sense to just decrease that number to 1 instead of adding yet another throttle.

benesch:
Fourth commit LGTM!

jseldess:
The generator changes LG(great)TM, @knz. Thank you! A number of the diagrams now are either too broad or too narrow (in terms of parameters, not in terms of appearance). I think there's probably no way for us to avoid case-by-case cleanup, though, which I'm willing to do closer to the 1.1 release date.

knz:
@jesse if you have specific examples of inelegant diagrams I am ready to have a look (possibly in a different PR)

@mjibson I actually removed the javaThrottle entirely, so that it is subsumed by -max-workers. We need -max-workers too in the case the jar file is not present and we have to use the HTTP API instead -- that thing also needs throttling!

@sploiselle: the grammar for BACKUP has changed, the new parameter is really of a different type. As to how to find these "mismatches"... I think the best solution is to look at the diffs.

---

Review status: 0 of 62 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1875#-:-KtXxiB3eUYVbllFvBmC:b339dz)*

knz:
feel free to merge when you're happy or tell me what needs further improvement!

jseldess:
This is just one example of a diagram that is now too broad: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/7157d613823c5705a0c315f9f2c17f53029a6761/dev/add-constraint.html

Compare that to the currently live version: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/7157d613823c5705a0c315f9f2c17f53029a6761/dev/add-constraint.html

But let's just merge this and I (we) can handle the diagrams separately.

knz:
It's interesting, the alter table railroad diagram is just plainly wrong. The live version is also wrong. I wonder how many other diagrams are also wrong in that way.

PR #1895, opened 2017-09-08
Fix `|` rendering problem by adding a single space

sesam:
See https://github.com/vmg/redcarpet/issues/546 for more on how redcarpet cares about spaces. There's a lax_spacing setting that might be part of the problem or solution, but as is right now just remembering to add some extra space now and again is a good enough quick fix.


jseldess:
@sesam, unfortunately, this doesn't render any differently. See http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/db78cbdbf7db91ae9c13b45562d5fd15158ebc0b/stable/use-the-built-in-sql-client.html#sql-shell-commands.

sesam:
Oh, strange. Maybe something "normalizes" the markdown source before
rendering? I noticed my added spaces get removed when I reopen the page
source. How about adding space &nbsp; space?

jseldess:
Think I figured this out. See https://github.com/cockroachdb/docs/pull/1896. Thanks for your efforts, @sesam!

PR #1903, opened 2017-09-11
Add CASCADE/RESTRICT to DROP DATABASE docs

### LINKED ISSUE #1902: sql: extend DROP DATABASE with an optional RESTRICT/CASCADE
PR: https://github.com/cockroachdb/cockroach/pull/17926

jseldess:
Fixes #1902

### DIFF
--- a/_includes/sql/v1.1/diagrams/drop_database.html
+++ b/_includes/sql/v1.1/diagrams/drop_database.html
@@ -1,4 +1,4 @@
-<div><svg width="642" height="68">
+<div><svg width="628" height="112">
          
          <polygon points="9 17 1 13 1 21"></polygon>
          <polygon points="17 17 9 13 9 21"></polygon>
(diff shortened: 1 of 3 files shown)


knz:
I reviewed this. The details of the change are OK.
However I'd like to take this opportunity to highlight a high-level
problem in the approach, which I have noticed may be also present in
several other pages: that of the handling of "default syntax".

So here the example is DROP DATABASE vs. DROP DATABASE RESTRICT.

In general, for every feature in CockroachDB where, say, we support a
statement "FROBNICATE FOO" that also exists as "FROBNICATE FOO UNDER X"
and "FROBNICATE FOO ABOVE Y", *and* it so happens that "FROBNICATE FOO"
is equivalent to specifying "UNDER X", we need to document those
variants and we need to document the "default", but there's a structure
to it.

Right now, the approach we have is like:

"FROBNICATE FOO is a statement to frobnicate foo. It has two parameters:
UNDER X and ABOVE Y. UNDER X is the default.

Here are some examples:

- FROBNICATE FOO UNDER X, for when you want to be on top.

- FROBNICATE FOO ABOVE Y, for when you want to be on the bottom.
"

The two problems with this approach are:

- "UNDER X" and "ABOVE Y" are not separate parameters! (more on this below)
- "UNDER X" is not a "default" - at least not in the sense people use it
technically, like we have defaults for command-line options.

If I were to explain this myself, I would say it this way:

"FROBNICATE FOO frobnicates foo. It has one parameter that determines
where the frobnication takes place. You can use it as follows:

FROBNICATE FOO [UNDER X]: for when you want to be on top. This is the
most common choice, so we also allow you to leave UNDER X out.

FROBNICATE FOO ABOVE Y: for when you want to be on the bottom.

Some examples:

- FROBNICATE FOO (reminder: "UNDER X" is optional)
- FROBNICATE FOO ABOVE Y
"

The root of the problem here is the confusion between when two things
are separate parameters, and when there is a single parameter with
multiple alternate values. This confusion is pervasive throughout our
docs, so I guess we can only fix this incrementally. Also, a lot of this
problem is caused by SQL not making a clear syntactic distinction
between the two (SQL is actually rather badly designed in that regard).
But here for DROP DATABASE we have a chance to make it clearer straight
away.

Some guidelines:

- IF EXISTS / IF NOT EXISTS / empty: 1 parameter. The 3 variants of that
single parameters have different meanings.
- CASCADE / RESTRICT / empty: 1 parameter. The variant "empty" is
equivalent to "RESTRICT".
- NULL / NOT NULL / empty: 1 parameter. The variant "empty" is
equivalent to NULL.
- DISTINCT / empty: 1 parameter. The two variants have different
meanings. This one is actually interesting: when DISTINCT is left out,
this is still a specification for that parameter (we could make it
clearer by also accepting NOT DISTINCT, but that's just not how SQL works).



-- 
Raphael 'kena' Poss

jseldess:
TFTR, @knz. I understand your points, but I'm not sure the changes you suggest impact user comprehension much. The Postgres docs, in fact, also list `CASCADE` and `RESTRICT` separate under **Parameters**: https://www.postgresql.org/docs/9.2/static/sql-altertype.html

The easy solution here is to not "inline" diagrams as often as we are currently. That would result in parent "parameters" being represented visually in the diagrams, in which case we could define just one parameter with its multiple options. 

However, we don't have the time to change this approach throughout the sql docs right now, given the imminent 1.1 release. So: If these changes are ok and will make sense to users, can you log your  suggestion as a separate issue?

jseldess:
@knz, in reviewing your feedback, I see that you're suggesting we take an iterative approach in making these improvements. Let me correct the page in question now then. More soon.

jseldess:
@knz, I looked at the Postgres docs some more and see that they provide a list of "Parameters" that mixes options with parameters. See their `DROP TABLE` docs for example: https://www.postgresql.org/docs/9.6/static/sql-droptable.html. They also represent certain possible values (`RESTRICT`) as the "default". 

I feel this this approach is intuitive and user-friendly, even if not completely semantically correct. My vote would be to continue mirroring how Postgres does this. But I of course understand and appreciate your perspective, so my last commit meets you half-way by pulling "options" into a separate table. Thoughts?


I'm still not convinced this is the right approach

knz:
:lgtm: as-is. I'll trust your experience on this.

---

Reviewed 3 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1903#-:-KtwgmejOhkoLSaa4kur:b-yce3wm)*

jseldess:
TFYR, @knz.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1903#-:-Ktwo1sofgsKQ31JoCIz:b-yevgdp)*

Issue #1291, opened 2017-04-13
sql: Document `system` database


sploiselle:
Briefly mentioned in #1160, but the tables of the `system` database should each be documented.

We don't see many questions about this currently (so I don't think it meets the 1.0 milestone requirement), but think it warrants documentation.

Currently the only reference to the `system` database is in [Learn CockroachDB SQL](https://www.cockroachlabs.com/docs/learn-cockroachdb-sql.html), which says:

> CockroachDB comes with a single default system database, which contains CockroachDB metadata and is read-only.

rmloveland:
I think this should be internal engineering/support documentation.  IIRC, users should not be doing anything with the `system` database other than knowing it exists and is used by the system.  What do you think @sploiselle ?  Can I convince you of this position?  🙂

rmloveland:
> users should not be doing anything with the system database

After reading some internal discussion about tables that live in `system.*` I realize this is not a correct opinion. :-)

rmloveland:
E.g., `DELETE FROM system.table_statistics WHERE name LIKE '%foo%';`

ericharmeling:
Closing in favor of https://github.com/cockroachdb/docs/issues/10340

PR #1905, opened 2017-09-11
Various updates to cockroach command docs

### LINKED ISSUE #1861: cli/sql: --safe-updates flag
PRs:
- https://github.com/cockroachdb/cockroach/pull/17889
- https://github.com/cockroachdb/cockroach/pull/17604 

Summary: New `cockroach sql` flag called `--safe-updates`. If `true`, the following types of statements are not allowed:

- `DELETE` without a `WHERE` clause
- `UPDATE` without a `WHERE` clause
- `DROP DATABASE` on a non-empty database
- `ALTER TABLE DROP COLUMN`

Defaults to `true`.

jseldess:
- Update `cockroach sql` docs with `--unsafe-updates` flag and example,
  `--echo-sql` flag and example, new SQL shell options, and
  corrected `--format` flag and examples.
- Update `cockroach dump` docs with `--echo-sql` flag.
- Update `cockroach user` docs with `--echo-sql` flag and example.
- Update `cockroach zone` docs with `--echo-sql` flag and example.


knz:
A) `--echo-sql` is a valid parameter for *all* `cockroach` sub-commands
that send SQL to the server, not just `cockroach sql`. For example
`cockroach node ls` also supports it. The examples should thus
illustrate multiple sub-commands.

B) I would not mention sql_safe_updates *at all* in the
DELETE/UPDATE/statement docs.

B.1) Most developers are going to use SQL non-interactively, from their
apps, and there sql_safe_updates is  *not* set by default so they will
never have to care about it.

B.2) in the CLI, the error message mentions already what to look at so
users can know *from the error message* what they should do. The docs
don't need to add a layer of this.

C) sql_safe_updates is a session variable, not a CLI shell variable, so
it should be documented with the other session variables. In the CLI SQL
shell page, a mere mention that the session variable receives a
different default in interactive shells is sufficient (with a "See Also"
link to the session variables at the bottom)

-- 
Raphael 'kena' Poss

jseldess:
TFTR, @knz!

> A) `--echo-sql` is a valid parameter for *all* `cockroach` sub-commands
that send SQL to the server, not just `cockroach sql`. For example
`cockroach node ls` also supports it. The examples should thus
illustrate multiple sub-commands.

Thanks for the clarification! I didn't realize it could be used with other commands. I corrected the flag description and example for `cockroach sql` and added the flag and examples for `cockroach dump`, `cockroach user`, and `cockroach zone`. `--echo-sql` is not available for any other commands like `cockroach node`; if that's an oversight, I'm happy to open an issue. 

> B) I would not mention sql_safe_updates *at all* in the
DELETE/UPDATE/statement docs.
>
> B.1) Most developers are going to use SQL non-interactively, from their
apps, and there sql_safe_updates is  *not* set by default so they will
never have to care about it.
>
> B.2) in the CLI, the error message mentions already what to look at so
users can know *from the error message* what they should do. The docs
don't need to add a layer of this.

Thanks for this clarification, too. I removed mention from those docs. 

> C) sql_safe_updates is a session variable, not a CLI shell variable, so
it should be documented with the other session variables. In the CLI SQL
shell page, a mere mention that the session variable receives a
different default in interactive shells is sufficient (with a "See Also"
link to the session variables at the bottom)

I clarified the language a bit. 

Please take another look.

knz:
I like this!
However now that you provide more examples I found two things that need fixing:

1) regardless of whether we show examples or not, we must hide any uses of `crdb_internal` from the documentation. Whatever `crdb_internal` provides *must not be documented anywhere*. This means that if we keep the `--echo-sql` examples for `cockroach sql`, `dump` and `user rm`, we must doctor the examples to remove the queries to `crdb_internal` tables. 

2) Both Andrei and I think it is interesting to reveal the use of `system` tables by `zone` and `user` and other sub-commands, so these examples are justified (although Andrei also recommends we document the `system` tables at some point). However, I think that the `dump` example output for `--echo-sql` is very verbose and does not add value because it doesn't issue anything a user wouldn't otherwise expect `dump` to do. I would suggest eliding the `dump --echo-sql` example altogether, while keeping the parameter listed in the table above.

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 4 of 4 files at r3.
Review status: 4 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-:-Ktw6qNCQ4H_pdl1hTFt:b-yerlbe)*

knz:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[v1.1/use-the-built-in-sql-client.md, line 99 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-Ktw84V_sEK4uc2IuFZH:-Ktw84VaOKBlCTUCacz9:b-umzvpm) ([raw file](https://github.com/cockroachdb/docs/blob/9556d36e17a562aab943d1089a4768b8f12a9b1b/v1.1/use-the-built-in-sql-client.md#L99)):*
> ```Markdown
> Client Options | Description
> ---------------|------------
> `ECHO` | <span class="version-tag">New in v1.1:</span> Reveal the SQL statements sent implicitly by the SQL shell.<br><br>This option is disabled by default. To enable it, run `\set ECHO`. For a demonstration, see the [example](#reveal-the-sql-statements-sent-implicitly-by-the-command-line-utility) below.
> ```

Another note: we're generally moving away from using capitals for anything that is not a SQL keyword. I would thus recommend lower casing the entire "Client options" column.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905)*

jseldess:
@knz, please take another look.  I've removed the `--echo-sql` example from the `dump` doc, and I've doctored the `sql` and `user` examples to not show the implicit query to `crdb_internal`. User WILL see that, however, so hiding from the docs seems overly slippery to me. 

As for documenting system tables, yes, we've wanted to do that for a long while, just hasn't been a priority. Here's the issue: https://github.com/cockroachdb/docs/issues/1291

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-:-KtwN1_48SN53weN104H:brv2fok)*

jseldess:
*[v1.1/use-the-built-in-sql-client.md, line 99 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-Ktw84V_sEK4uc2IuFZH:-KtwOxKwGpWv_7Ik1u8L:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/9556d36e17a562aab943d1089a4768b8f12a9b1b/v1.1/use-the-built-in-sql-client.md#L99)):*

Another note: we're generally moving away from using capitals for anything that is not a SQL keyword. I would thus recommend lower casing the entire "Client options" column.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905)*

knz:
:lgtm: with nit

---

Reviewed 3 of 3 files at r5.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v1.1/create-and-manage-users.md, line 189 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-Ktwfv_5ezmiD8trE0GP:-Ktwfv_5ezmiD8trE0GQ:bc1im4y) ([raw file](https://github.com/cockroachdb/docs/blob/05247738bc56ff252ce170909c6a6800d0d4e9ac/v1.1/create-and-manage-users.md#L189)):*
> ```Markdown
> ~~~
> # Server version: CockroachDB CCL f44a747ea (darwin amd64, built 2017/09/08 14:30:37, go1.8) (same version as client)
> # Cluster ID: 847a4ba5-c78a-465a-b1a0-59fae3aab520
> ```

You can remove the two lines starting with `#` as they do not pertain to this command's operation. (Also, the fact they are displayed at all here is a bug -- https://github.com/cockroachdb/cockroach/issues/18486 -- that I intend to fix.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-:-KtwgM8nLhfFzIy2J3Jr:b-xv64lu)*

jseldess:
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[v1.1/create-and-manage-users.md, line 189 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/1905#-Ktwfv_5ezmiD8trE0GP:-KtwlBvgT4FoCY8qefrI:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/05247738bc56ff252ce170909c6a6800d0d4e9ac/v1.1/create-and-manage-users.md#L189)):*

You can remove the two lines starting with `#` as they do not pertain to this command's operation. (Also, the fact they are displayed at all here is a bug -- https://github.com/cockroachdb/cockroach/issues/18486 -- that I intend to fix.)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1905)*

PR #1908, opened 2017-09-13
Document production limits

### LINKED ISSUE #1830: Document product limits
We need a recommendation against letting a single row get close to 64MB (the default size at which we split a range). In addition to the current row data, all historical versions of the row that have not been garbage collected count toward the overall size. 

Normally, a range contains many rows, in which case when the range gets to the limit, we split into 2 ranges. If a single row takes up all of a range, however, we won't split the range but rather let it get larger than the max limit. 

- As the row size increases further, memory usage may get spikier (eventually reaching a point where servers will start running out of memory and dying).
- At some point, we’ll stop rebalancing the range as well because it is too large.

knz:
Fixes https://github.com/cockroachdb/cockroach/issues/15771.

Fixes  #1830.

@jseldess  the way it's going to work is that I am going to iterate on this with Peter and Ben until we align on the technical content. Then I will hand this PR over to you so you can perform the necessary editorial fixes and cleanups.

### DIFF
--- a/v1.1/limits.md
+++ b/v1.1/limits.md
@@ -40,7 +40,7 @@ and unlimited disk space, we observe a practical limit of about 50.000
 ranges per node (3.2TB per node). Our experiments suggest the
 bottleneck is currently on CPU time. A higher capacity per node can
 thus be reached by increasing CPU core counts and RAM size according
-to needs.
+to needs. We are also working on reducing the CPU overhead per range.
 
 To estimate the maximum number of ranges across an entire cluster, one
 must consider the replication factor and the maximum number of nodes.
@@ -83,9 +83,12 @@ values of that type.
 
 Further limits apply:
 
-- the combined size of all primary key values in a single row must fit
-  within 16KB. This is the maximum key size in the underlying RocksDB
-  store on each node.
+- the combined size of all primary key values in a single row should
+  fit within a few kilobytes. Although CockroachDB might support keys
+  up to 8MB in size, which is [the maximum key size in the underlying
+  RocksDB store](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ)
+  on each node, keys larger than a few kilobytes will cause a
+  significant decrease in CockroachDB performance.
 - the combined sized of a row must fit within a fraction of the range
   size. We highly recommend this to be kept within 30% of the range
   size, although larger fractions may also work.
--- a/v1.1/limits.md
+++ b/v1.1/limits.md
@@ -21,10 +21,10 @@ In order to frame the various limits that typically matter to
 application developers and architects, one must consider the range
 limits first. The other limits are derived from that afterwards.
 
-### Range limits
+### Range limits per node
 
-By default, the maximum size of a range is 64MiB. To manage a range,
-CockroachDB needs:
+By default, the maximum size of a range is 64MiB. To manage a range on
+a single node, CockroachDB needs:
 
 - 64MiB on disk,
 - some amount of RAM, a few kilobytes per range at a minimum,
@@ -42,8 +42,18 @@ bottleneck is currently on CPU time. A higher capacity per node can
 thus be reached by increasing CPU core counts and RAM size according
 to needs. We are also working on reducing the CPU overhead per range.
 
+CockroachDB also supports changing the range size (e.g. to increase
+it) but this is currently not fully tested. Different zones can use
+different range sizes.
+
+### Range limits across a cluster
+
 To estimate the maximum number of ranges across an entire cluster, one
-must consider the replication factor and the maximum number of nodes.
+must consider:
+
+- the replication factor;
+- the maximum number of nodes;
+- the *meta* ranges that index all the other ranges.
 
 The replication factor is configurable per zone, and is set to 3 by
 default.
@@ -59,19 +69,50 @@ cluster (about 10TB). This limit can be increased further by
 increasing resources per node (as described above) and increasing the
 number of nodes.
 
-CockroachDB also supports changing the range size (e.g. to increase
-it) but this is currently not fully tested. Different zones can use
-different range sizes.
+Meanwhile, CockroachDB uses a tree structure to index all ranges in a
+cluster (so as to know on which node each range is present). The top
+of the tree is a single range called "meta1", containing descriptors
+to "meta2" ranges, containing descriptors to the data ranges.  A range
+descriptor is about 100-200 bytes, so theoretically with 64MB ranges
+meta1 can index ~550k meta2 ranges, and each meta2 range can index
+~550k data ranges. This gives at the default range size a theoretical
+maximum of 10PB capacity in a CockroachDB cluster.
+
+However, CockroachDB v1.1 has a [known
+limitation](https://github.com/cockroachdb/cockroach/issues/18032)
+which prevents the creation of multiple meta2 ranges, so **until
+CockroachDB v1.2 a cluster may become increasingly slower when using
+more than ~550k data ranges in total: clusters with a capacity of
+about 1TB or beyond will not operate at max performance.** This is
+because the single meta2 range will be forced to grow beyond the
+nominal maximum range size, which in turn has a performance penalty
+throughout the cluster.
 
 ### Derived SQL schema limits
 
-- Maximum number of databases = 2^63
-- Maximum number of tables = approx. maximum number of ranges or 2^63, whichever is smallest
-- Maximum number of indexes / FKs = approx. maximum number of ranges
-- Maximum number of indexes / FKs per table = maximum number of indexes or 2^63, whichever is smallest
-- Maximum number of rows in a single table = approx. maximum number of ranges, times the number of rows per range
-- Maximum number of rows per range = range size / minimum row size
-- Minimum number of rows per range = range size / maximum row size
+- Maximum number of databases = 2^63;
+- Maximum number of tables = approx. maximum number of ranges or 2^63, whichever is smallest (but also see note below);
+- Maximum number of indexes / FKs = approx. maximum number of ranges;
+- Maximum number of indexes / FKs per table = maximum number of indexes or 2^63, whichever is smallest;
+- Maximum number of rows in a single table = approx. maximum number of ranges, times the number of rows per range;
+- Maximum number of rows per range = range size / minimum row size;
+- Minimum number of rows per range = range size / maximum row size.
+
+The maximum number of tables/views is a combination of the following limits:
+
+- each table and database has a 63-bit numeric identifier.
+- each non-interleaved table (not view) uses at least one range,
+  because CockroachDB ensures that the data from two different
+  non-interleaved tables never share a range. This implies the maximum
+  number of tables is limited by the maximum number of ranges.
+- in CockroachDB v1.1, all the table and view descriptors (metadata)
+  are stored in a single range, which is duplicated in RAM across all
+  the cluster. Table/view descriptors can have a variable size,
+  depending on the schema definition and the number of schema changes
+  over time. Assuming a descriptor size of around 1K, the nominal
+  suggested maximum is thus about 64.000 tables/views (descriptors
+  that fit in a range), although that range will be allowed to grow
+  beyond the nominal maximum range size, albeit with a performance penalty.
 
 ### SQL Min/max row sizes
 
--- a/v1.1/limits.md
+++ b/v1.1/limits.md
@@ -1,0 +1,94 @@
+---
+title: Production limits
+summary: How to estimate the maximum capacity of a CockroachDB cluster
+toc: true
+---
+
+<div id="toc"></div>
+
+
+### Introduction
+
+The limits in CockroachDB are all based around the notion of *range*.
+
+For example, the maximum amount of data that can be stored in a
+CockroachDB cluster is directly related to the maximum number of
+ranges and the maximum size of a range, and the maximum size of a row
+in a table is limited by the size of a single range, as described
+below.
+
+In order to frame the various limits that typically matter to
+application developers and architects, one must consider the range
+limits first. The other limits are derived from that afterwards.
+
+### Range limits
+
+By default, the maximum size of a range is 64MiB. To manage a range,
+CockroachDB needs:
+
+- 64MiB on disk,
+- some amount of RAM, a few kilobytes per range at a minimum,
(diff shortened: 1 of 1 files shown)


jseldess:
Sounds good to me, @knz. Thanks.

bdarnell:
Review status: 0 of 1 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

*[v1.1/limits.md, line 10 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBs3sWD7kkXN1ERB60:-KuBs3sWD7kkXN1ERB61:bsz7txa) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L10)):*
> ```Markdown
> 
> 
> ### Introduction
> ```

I think it would be good to provide a summary of the major limits before going into how they are derived. 

---

*[v1.1/limits.md, line 29 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuAEg6Km62GKJs1sAU2:-KuAEg6Km62GKJs1sAU3:b-t0ty9d) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L29)):*
> ```Markdown
> a single node, CockroachDB needs:
> 
> - 64MiB on disk,
> ```

"Up to 64MiB on disk" (or maybe "Approximately"). It's rare for a range to be exactly full, and we store the data compressed (but split based on uncompressed size). On the other hand, rocksdb has some space amplification, so growing ranges will take up more space than their actual size until compaction occurs. 

---

*[v1.1/limits.md, line 39 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBp0dMwlQwg52s0IC1:-KuBp0dMwlQwg52s0IC2:b-g44h9g) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L39)):*
> ```Markdown
> 
> For example, in our test clusters with 8 cores per node, 16GB of RAM,
> and unlimited disk space, we observe a practical limit of about 50.000
> ```

Write numbers in US style (commas not periods for thousands separators).

---

*[v1.1/limits.md, line 46 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuAaiZzGF2sEljeuror:-KuAaiZzGF2sEljeuros:bjcstzr) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L46)):*
> ```Markdown
> 
> CockroachDB also supports changing the range size (e.g. to increase
> it) but this is currently not fully tested. Different zones can use
> ```

If we're cpu-limited we should definitely test this. My estimate is that increasing range sizes will reduce CPU usage but increase memory. 

---

*[v1.1/limits.md, line 76 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuA_DENjLziMXcgkRGX:-KuA_DENjLziMXcgkRGY:bxlg80z) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L76)):*
> ```Markdown
> of the tree is a single range called "meta1", containing descriptors
> to "meta2" ranges, containing descriptors to the data ranges.  A range
> descriptor is about 100-200 bytes, so theoretically with 64MB ranges
> ```

The size of a range descriptor is influenced by the application: indexing large string columns (for example) will result in larger range descriptors on average due to larger key sizes, and therefore lower overall capacity. 

---

*[v1.1/limits.md, line 78 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuAaSStAi8NYkDvb3Ob:-KuAaSStAi8NYkDvb3Oc:b-4n2y2n) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L78)):*
> ```Markdown
> descriptor is about 100-200 bytes, so theoretically with 64MB ranges
> meta1 can index ~550k meta2 ranges, and each meta2 range can index
> ~550k data ranges. This gives at the default range size a theoretical
> ```

Given the variability in range descriptor sizes, "theoretical" probably isn't the right word to use here. 

---

*[v1.1/limits.md, line 79 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuA_oj-ZX_NiqQ0NTXy:-KuA_oj-ZX_NiqQ0NTXz:bkjkrm) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L79)):*
> ```Markdown
> meta1 can index ~550k meta2 ranges, and each meta2 range can index
> ~550k data ranges. This gives at the default range size a theoretical
> maximum of 10PB capacity in a CockroachDB cluster.
> ```

The design doc does a similar computation with a result of 4 exabytes. We should compute this limit the same way in both places (updating the design doc if we now have more accurate information about range descriptor sizes).

---

*[v1.1/limits.md, line 84 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuAa_IdkBhUi8N6z4VG:-KuAa_IdkBhUi8N6z4VH:b-rx2sgz) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L84)):*
> ```Markdown
> limitation](https://github.com/cockroachdb/cockroach/issues/18032)
> which prevents the creation of multiple meta2 ranges, so **until
> CockroachDB v1.2 a cluster may become increasingly slower when using
> ```

Will it get slower? It should have similar consequences to increasing the max range size. 

---

*[v1.1/limits.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBpfA34I_a_xYGbJuo:-KuBpfA34I_a_xYGbJup:bmxblw7) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L86)):*
> ```Markdown
> CockroachDB v1.2 a cluster may become increasingly slower when using
> more than ~550k data ranges in total: clusters with a capacity of
> about 1TB or beyond will not operate at max performance.** This is
> ```

`550k * 64MiB` is 33TiB (and because 550k is a limit on the number of ranges, not the number of replicas, we don't need to divide by 3). The 1.1 limit therefore is 33TiB of user data stored on 100TiB of raw disk, which must be spread across at least 30 nodes. 

---

*[v1.1/limits.md, line 93 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBfirKaIyv0z0Ls4h0:-KuBfirKaIyv0z0Ls4h1:b-28cllt) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L93)):*
> ```Markdown
> ### Derived SQL schema limits
> 
> - Maximum number of databases = 2^63;
> ```

We can call 2^63 "unlimited".

However, IIRC databases also result in the creation of a range (which will always be empty), so they do run into the range limit. (Maybe we should stop auto-splitting ranges at schema id boundaries)

---

*[v1.1/limits.md, line 95 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBfx5-04nINip4RY5g:-KuBfx5-04nINip4RY5h:b-kk3ipn) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L95)):*
> ```Markdown
> - Maximum number of databases = 2^63;
> - Maximum number of tables = approx. maximum number of ranges or 2^63, whichever is smallest (but also see note below);
> - Maximum number of indexes / FKs = approx. maximum number of ranges;
> ```

Indexes/FKs don't create their own range; the only limit for them is that they increase the size of their table descriptors. 

---

*[v1.1/limits.md, line 97 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBgarLM5NBXDg4adKv:-KuBgarLM5NBXDg4adKw:blgo1gi) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L97)):*
> ```Markdown
> - Maximum number of indexes / FKs = approx. maximum number of ranges;
> - Maximum number of indexes / FKs per table = maximum number of indexes or 2^63, whichever is smallest;
> - Maximum number of rows in a single table = approx. maximum number of ranges, times the number of rows per range;
> ```

This is only true if essentially all of your data is in one table. I think that phrasing it this way confuses more than it helps; we should just say that there is no limit on the size of a table except for the total limit on the data size of the cluster. 

---

*[v1.1/limits.md, line 99 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBh712itwByr6brtNN:-KuBh712itwByr6brtNO:b-7bt5gm) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L99)):*
> ```Markdown
> - Maximum number of rows in a single table = approx. maximum number of ranges, times the number of rows per range;
> - Maximum number of rows per range = range size / minimum row size;
> - Minimum number of rows per range = range size / maximum row size.
> ```

These last two are true, but are they useful? 

---

*[v1.1/limits.md, line 108 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuAb8HckC3k6cmfpr7V:-KuAb8HckC3k6cmfpr7W:bxsel5n) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L108)):*
> ```Markdown
>   non-interleaved tables never share a range. This implies the maximum
>   number of tables is limited by the maximum number of ranges.
> - in CockroachDB v1.1, all the table and view descriptors (metadata)
> ```

Bring this comment up to the top of this section. It is the relevant limit in nearly all cases, and not the range count or 2^63. 

---

*[v1.1/limits.md, line 112 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBnuEcKXWRsVnSiHNb:-KuBnuEcKXWRsVnSiHNc:bjwaoob) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L112)):*
> ```Markdown
>   the cluster. Table/view descriptors can have a variable size,
>   depending on the schema definition and the number of schema changes
>   over time. Assuming a descriptor size of around 1K, the nominal
> ```

1KB seems high for a table descriptor, although some applications use very wide tables that will be even larger. Again, this is highly application-dependent. 

---

*[v1.1/limits.md, line 117 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBlzZooZA_8Hpsc1Wz:-KuBlzZooZA_8Hpsc1X-:bhqec46) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L117)):*
> ```Markdown
>   beyond the nominal maximum range size, albeit with a performance penalty.
> 
> ### SQL Min/max row sizes
> ```

A row has two sizes: the size of the row itself (roughly equal to `keyEncoding(pkColumns) * numColumnFamilies + valueEncoding(nonPKColumns)`), which is the size that must be kept under the max range size, and the total amount of data that it contributes to the table's size (which includes the row itself plus secondary indexes: `keyEncoding(idxColumns) + valueEncoding(pkColumns) + valueEncoding(storedColumns)` for each index).

---

*[v1.1/limits.md, line 128 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBhptYl9tJlBI8Ubew:-KuBhptYl9tJlBI8Ubex:b4hma0z) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L128)):*
> ```Markdown
> 
> - the combined size of all primary key values in a single row should
>   fit within a few kilobytes. Although CockroachDB might support keys
> ```

Kilobyte-sized primary keys would still be pretty bad. I'd recommend no more than 250 bytes for a primary key, with a strong preference for 8 or 16 bytes. (A few KB is a good guideline for fields used in secondary indexes)

---

*[v1.1/limits.md, line 133 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/1908#-KuBkvqkIoFJy8QDoOuS:-KuBkvqkIoFJy8QDoOuT:b2amxlp) ([raw file](https://github.com/cockroachdb/docs/blob/a728ef9976a547f5b04f2786a4cfd64024e8a021/v1.1/limits.md#L133)):*
> ```Markdown
>   on each node, keys larger than a few kilobytes will cause a
>   significant decrease in CockroachDB performance.
> - the combined sized of a row must fit within a fraction of the range
> ```

We should also discuss the retention of MVCC versions, and that the combined size of a row includes all of its historical versions within the GC TTL (plus some factor to account for the fact that GC is not instantaneous when the TTL is reached). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1908)*

jseldess:
@knz, any plans to continue working on this? No rush, but I think it'd be great to have for 1.2. In conjunction, I'll be reworking our production deployment guidance.

knz:
Yes I plan to work on this; however, the core team is currently making changes that heavily influence this document, and I'm waiting for the code to stabilize before I make the corresponding changes.

jseldess:
@robert-s-lee, @vivekmenezes, @cuongdo, this is the fine-grained production limits document I referenced in our meeting earlier. Based on @knz's last comment, it seems like we need to wait from some core work before proceeding.

jseldess:
@knz, would you mind adding some details about the core changes that will effect these limits? I discussed this briefly with @petermattis in the docs meeting today, and he wasn't sure what changes you might mean.

jseldess:
@knz, @bdarnell, is this something we want to resurrect?

jseldess:
We have closed this issue because it is more than 3 years
old. If this issue is still relevant, please add a comment and reopen
the issue. Thank you for your contribution to CockroachDB docs!

Issue #1288, opened 2017-04-12
core: Explicit init command


jseldess:
Background: 
  - https://airtable.com/tblD3oZPLJgGhCmch/viwdV1sM5KEK2L14K/recdaFNmgzO5oBwe7
  - RFC: https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/init_command.md

Owner: @bdarnell 

PRs: https://github.com/cockroachdb/cockroach/pull/16371

jseldess:
@bdarnell , this is the overall docs issue for the init command work you're doing for 1.1. Once it's ready for docs, please assign the `ready` label and add relevant PR links?

sploiselle:
Landed in cockroachdb/cockroach#16371

jseldess:
Thanks for tracking this, @sploiselle!

dianasaur323:
@jseldess do you mind letting me know when your docs for this land? It's the last feature on my list of features to acceptance test.

jseldess:
Will do, @dianasaur323. Working on upgrade docs now. `init` is next.

PR #1910, opened 2017-09-13
Update v1.1 Docker Swarm instructions to use the init command

a-robinson:
Also switch to using an `attachable` network which makes running one-off
containers easier and simplify/fix the docs as applicable.

We could also go a little further with the secure guide and remove the
client keys from each container now that we can connect to the cluster
with one-off containers, but it'll probably still come in handy in
debugging situations to have the keys in all the containers.

Fixes part of https://github.com/cockroachdb/docs/issues/1288. I'm not yet making the change suggested in https://github.com/cockroachdb/cockroach/issues/17133 because it would make our insecure and secure setups diverge quite a bit, and I'm not thrilled about that.

### DIFF
--- a/v1.0/orchestrate-cockroachdb-with-docker-swarm.md
+++ b/v1.0/orchestrate-cockroachdb-with-docker-swarm.md
@@ -295,7 +295,7 @@ A secure CockroachDB cluster uses TLS certificates for encrypted inter-node and 
     --ca-key=my-safe-directory/ca.key
     ~~~
 
-10. Create Docker secrets for the `root` user's certificate and key:
+12. Create Docker secrets for the `root` user's certificate and key:
 
     {% include copy-clipboard.html %}
     ~~~ shell
@@ -528,7 +528,7 @@ To increase the number of nodes in your CockroachDB cluster:
 4. Create security resources for the node (see [Step 5.7 and 5.8](#step-5-create-security-resources)).
 5. Create a new service to start another node and join it to the CockroachDB cluster (see [Step 6.2](#step-6-start-the-cockroachdb-cluster)).
 
-## Step 10. Stop the cluster
+## Step 11. Stop the cluster
 
 To stop the CockroachDB cluster, on the instance running your manager node, remove the services:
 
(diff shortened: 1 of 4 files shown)


a-robinson:
I tested this using the alexrobinson/cockroach:v1.1-beta.20170907 docker image, in case anyone wants to try it themselves.

v1.0/orchestrate-cockroachdb-with-docker-swarm.md
```diff
@@ -528,7 +528,7 @@ To increase the number of nodes in your CockroachDB cluster:
 4. Create security resources for the node (see [Step 5.7 and 5.8](#step-5-create-security-resources)).
 5. Create a new service to start another node and join it to the CockroachDB cluster (see [Step 6.2](#step-6-start-the-cockroachdb-cluster)).
 
-## Step 10. Stop the cluster
+## Step 11. Stop the cluster
```
jseldess: Thanks for catching these, @a-robinson!

v1.1/orchestrate-cockroachdb-with-docker-swarm-insecure.md
```diff
@@ -197,57 +201,31 @@ $ sudo docker network create --driver overlay cockroachdb
...
-    ~~~ shell
-    $ sudo docker ps | grep cockroachdb
-    ~~~
-
-    ~~~
-    9539871cc769        cockroachdb/cockroach:{{page.release_info.version}}   "/cockroach/cockroach"   2 minutes ago        Up 2 minutes         8080/tcp, 26257/tcp   cockroachdb-1.1.0wigdh8lx0ylhuzm4on9bbldq
-    ~~~
-
-2. Use the `sudo docker exec` command to open the built-in SQL shell in interactive mode inside the container:
+1. Use the `sudo docker run` command to run a new container attached to the CockroachDB network and connect to the cluster:
```
jseldess: I think it's helpful to call out that we're launching the built-in SQL shell, but I see that you removed that language. Do you feel that "connect to the cluster" conveys the same thing?

a-robinson: You're right, being more specific about how we're connecting is better. I've changed these back.

PR #1912, opened 2017-09-13
Soften the warning about running cross-continent clusters

a-robinson:
It's safe to do so, we just want to make sure users are aware of the
performance penalty involved.


a-robinson:
Done.

jseldess:
Thanks, @a-robinson!

PR #1928, opened 2017-09-26
Add FAQ on how ints are returned as strings in JavaScript

jseldess:
Addresses https://github.com/cockroachdb/cockroach/issues/15721

### DIFF
--- a/v1.0/sql-faqs.md
+++ b/v1.0/sql-faqs.md
@@ -86,6 +86,36 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column 
 
 When an [`ORDER BY`](select.html#sorting-retrieved-values) clause is not used in a `SELECT` query, retrieved rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.
 
+## Why are my `INT` columns returned as strings in JavaScript?
+
+In CockroachDB, all `INT`s are represented with 64 bits of precision, but JavaScript numbers only have 53 bits of precision. This means that large integers stored in CockroachDB are not exactly representable as JavaScript numbers. For example, JavaScript will round the integer `235191684988928001` to the nearest representable value, `235191684988928000`. Notice that the last digit is different. This is particularly problematic when using the `unique_rowid()` [function](functions-and-operators.html), since `unique_rowid()` nearly always returns integers that require more than 53 bits of precision to represent.
+
+To avoid this loss of precision, Node's [`pg` driver](https://github.com/brianc/node-postgres) will, by default, return all CockroachDB `INT`s as strings.
+
+~~~ javascript
+// Schema: CREATE TABLE users (id INT DEFAULT unique_rowid(), name STRING);
+pgClient.query("SELECT id FROM users WHERE name = 'Roach' LIMIT 1", function(err, res) {
+  var idString = res.rows[0].id;
+  // idString === '235191684988928001'
+  // typeof idString === 'string'
+});
+~~~
+
+To perform another query using the value of `idString`, you can simply use `idString` directly, even where an `INT` type is expected. The string will automatically be coerced into a CockroachDB `INT`.
+
+~~~ javascript
+pgClient.query("UPDATE users SET name = 'Ms. Roach' WHERE id = $1", [idString], function(err, res) {
+  // All should be well!
+});
+~~~
+
+If you instead need to perform arithmetic on `INT`s in JavaScript, you will need to use a big integer library like [Long.js](https://www.npmjs.com/package/long). Do _not_ use the built-in `parseInt` function.
+
+~~~ javascript
+parseInt(idString, 10) + 1; // WRONG: returns 235191684988928000
+require('long').fromString(idString).add(1).toString(); // GOOD: returns '235191684988928002'
+~~~
+
 ## See Also
 
 - [Product FAQs](frequently-asked-questions.html)
--- a/v1.1/sql-faqs.md
+++ b/v1.1/sql-faqs.md
@@ -86,6 +86,36 @@ Yes. For more details, see [`UUID`](uuid.html).
 
 When an [`ORDER BY`](select.html#sorting-retrieved-values) clause is not used in a `SELECT` query, retrieved rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.
 
+## Why are my `INT` columns returned as strings in JavaScript?
+
+In CockroachDB, all `INT`s are represented with 64 bits of precision, but JavaScript numbers only have 53 bits of precision. This means that large integers stored in CockroachDB are not exactly representable as JavaScript numbers. For example, JavaScript will round the integer `235191684988928001` to the nearest representable value, `235191684988928000`. Notice that the last digit is different. This is particularly problematic when using the `unique_rowid()` [function](functions-and-operators.html), since `unique_rowid()` nearly always returns integers that require more than 53 bits of precision to represent.
+
+To avoid this loss of precision, Node's [`pg` driver](https://github.com/brianc/node-postgres) will, by default, return all CockroachDB `INT`s as strings.
+
+~~~ javascript
+// Schema: CREATE TABLE users (id INT DEFAULT unique_rowid(), name STRING);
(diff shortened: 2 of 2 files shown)


v1.0/sql-faqs.md
```diff
@@ -86,6 +86,16 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column
...
+## How do I use `unique_rowid()` in JavaScript?
+
+The `unique_rowid()` [function](functions-and-operators.html) returns values that cannot be represented exactly by numbers in Javascript. For example, a `235191684988928001` value returned by `unique_rowid()` would get represented as `235191684988928000` in Javascript. Notice that the last digit is different. In such cases, if the Javascript application temporarily stores the value `235191684988928001` and queries using it, it won't match the value that actually got stored.
+
+As a workaround when using JavaScript or another language that doesn't support 64-bit integer types, cast the integer to a string server-side before inspecting or using it, for example:
+
+~~~ sql
+> SELECT CAST (id AS string) FROM my_table;
+~~~
+
```
benesch: Actually, there's no need to explicitly cast to a string. That happens automatically. Here's how I would explain it.

**Why are my `INT` columns returned as strings in JavaScript?**

In CockroachDB, all `INT`s are represented with 64 bits of precision, but JavaScript numbers only have 53 bits of precision. This means that large integers stored in CockroachDB are not exactly representable as JavaScript numbers. For example, JavaScript will round the integer `235191684988928001` to the nearest representable value, `235191684988928000`. Notice that the last digit is different. This is particularly problematic when using the `unique_rowid()` function, since `unique_rowid()` nearly always returns integers that require more than 53 bits of precision to represent.

To avoid this loss of precision, Node's [`pg` driver](https://github.com/brianc/node-postgres) will, by default, return all CockroachDB `INT`s as strings. 

```js
// Schema: CREATE TABLE users (id INT DEFAULT unique_rowid(), name STRING);
pgClient.query('SELECT id FROM users WHERE name = 'Roach' LIMIT 1', function(err, res) {
  var idString = res.rows[0].id; 
  // idString == '235191684988928001'
  // typeof idString == 'string'
});
```

To perform another query using the value of `idString`, you can simply use `idString` directly, even where an `INT` type is expected. Either the Postgres driver or CockroachDB will automatically convert the string to a CockroachDB `INT`.

```js
pgClient.query('UPDATE users SET name = 'Ms. Roach' WHERE id = $1', [idString], function(err, res) {
  // All should be well!
});
```

If you instead need to perform arithmetic on `INT`s in JavaScript, you will need to use a big integer library like [Long.js](https://www.npmjs.com/package/long). Do *not* use the built-in `parseInt` function.

```js
parseInt(idString, 10) + 1; // WRONG: returns 235191684988928000
require('long').fromString(idString).add(1).toString(); // GOOD: returns '235191684988928002'
```

jseldess: Done. Thanks very much for the revision, @benesch!

v1.0/sql-faqs.md
```diff
@@ -86,6 +86,36 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column
...
 
+## Why are my `INT` columns returned as strings in JavaScript?
+
+In CockroachDB, all `INT`s are represented with 64 bits of precision, but JavaScript numbers only have 53 bits of precision. This means that large integers stored in CockroachDB are not exactly representable as JavaScript numbers. For example, JavaScript will round the integer `235191684988928001` to the nearest representable value, `235191684988928000`. Notice that the last digit is different. This is particularly problematic when using the `unique_rowid()` [function](functions-and-operators.html), since `unique_rowid()` nearly always returns integers that require more than 53 bits of precision to represent.
+
+To avoid this loss of precision, Node's [`pg` driver](https://github.com/brianc/node-postgres) will, by default, return all CockroachDB `INT`s as strings.
+
+~~~ javascript
+// Schema: CREATE TABLE users (id INT DEFAULT unique_rowid(), name STRING);
+pgClient.query('SELECT id FROM users WHERE name = 'Roach' LIMIT 1', function(err, res) {
```
benesch: Shoot, this isn't valid JavaScript as written. Needs to use double quotes to allow single quotes within.

```sql
"SELECT id FROM users WHERE name = 'Roach' LIMIT 1"
```

v1.0/sql-faqs.md
```diff
@@ -86,6 +86,36 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column
...
+  var idString = res.rows[0].id;
+  // idString == '235191684988928001'
+  // typeof idString == 'string'
+});
+~~~
+
+To perform another query using the value of `idString`, you can simply use `idString` directly, even where an `INT` type is expected. Either the Postgres driver or CockroachDB will automatically convert the string to a CockroachDB `INT`.
+
+~~~ javascript
+pgClient.query('UPDATE users SET name = 'Ms. Roach' WHERE id = $1', [idString], function(err, res) {
```
benesch: Ditto here. 

```js
"UPDATE users SET name = 'Ms. Roach' WHERE id = $1"
```

jseldess: Done.

v1.0/sql-faqs.md
jseldess: Good catch. Done.

v1.0/sql-faqs.md
```diff
@@ -86,6 +86,36 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column
...
+
+In CockroachDB, all `INT`s are represented with 64 bits of precision, but JavaScript numbers only have 53 bits of precision. This means that large integers stored in CockroachDB are not exactly representable as JavaScript numbers. For example, JavaScript will round the integer `235191684988928001` to the nearest representable value, `235191684988928000`. Notice that the last digit is different. This is particularly problematic when using the `unique_rowid()` [function](functions-and-operators.html), since `unique_rowid()` nearly always returns integers that require more than 53 bits of precision to represent.
+
+To avoid this loss of precision, Node's [`pg` driver](https://github.com/brianc/node-postgres) will, by default, return all CockroachDB `INT`s as strings.
+
+~~~ javascript
+// Schema: CREATE TABLE users (id INT DEFAULT unique_rowid(), name STRING);
+pgClient.query("SELECT id FROM users WHERE name = 'Roach' LIMIT 1", function(err, res) {
+  var idString = res.rows[0].id;
+  // idString == '235191684988928001'
```
justinj: doesn't really matter since it's a comment, but my javascript head alarm goes off whenever it sees `==`, since `===` should generally be used instead.

v1.0/sql-faqs.md
```diff
@@ -86,6 +86,36 @@ Not at this time, but storing a 16-byte array in a [`BYTES`](bytes.html) column
...
+  // typeof idString == 'string'
+});
+~~~
+
+To perform another query using the value of `idString`, you can simply use `idString` directly, even where an `INT` type is expected. Either the Postgres driver or CockroachDB will automatically convert the string to a CockroachDB `INT`.
```
justinj: what is meant by "Either the Postgres driver or CockroachDB"? shouldn't it always be Cockroach doing this conversion? or does the driver somehow check if the string is within the int bounds?

benesch: Node's `pg` driver does its own type conversion provided you use prepared queries. So in a query like

```query("UPDATE foo SET int_col = $1", ["1234"])```

the `pg` driver sees an integer type hint on "$1" and will send the integer `1234` over the wire, instead of the string `"1234"`.

The reason we needed string -> number literal coercion in CockroachDB itself was to cope with Sequelize, which eschews prepared queries in favor of string mashing, e.g.:

```
query("UPDATE foo SET int_col = '" + escape("1234") + "'")
```

That's too detailed an explanation, of course, and I don't think we want a string mashing example in our docs. Too easy for people to think that we're encouraging string mashing.

Perhaps we can replace the sentence with "The string will automatically be coerced into a CockroachDB `INT`", and leave out mentioning exactly which layer is doing the conversion?

jseldess: Done.

v1.0/sql-faqs.md
jseldess: Done.

PR #1932, opened 2017-09-26
Created certificate rotation document

### LINKED ISSUE #674: CA cert expiration
From Gitter, 9/25/16:

Dolf Schimmel @Freeaqingme
Hi folks. Just noticed the CA cert generated by Cockroach is only valid for a year. Is the expiry of the certificates and CA checked anywhere within cockroach? If so, will my cluster collapse in a year?

kena @knz 10:46
Freeaqingme: yes it is checked. The cluster will not "collapse" you will just see connection refusals when you attempt to restart nodes. Hopefully, before then we'll introduce ways to transition during cert renewals

Dolf Schimmel @Freeaqingme 10:53
Okay, thanks. Just to be safe I guess I'll generate the CA by hand but then give it an expiry for 10 years.

Dolf Schimmel @Freeaqingme 14:01
Is it possible to associate a client certificate with a specific user?

Dolf Schimmel @Freeaqingme 14:38
oh lol, nvm. Looks like the CN is actually the username. Didn't get that from the docs

### LINKED ISSUE #1499: core: Cert rotation
Background: 
 - RFC: https://github.com/cockroachdb/cockroach/pull/14254
 - https://airtable.com/tblD3oZPLJgGhCmch/viwdV1sM5KEK2L14K/recSJKR91duNibBlb

PRs: 
 - docs draft: https://github.com/cockroachdb/docs/pull/1331
 - https://github.com/cockroachdb/cockroach/pull/14925
 - https://github.com/cockroachdb/cockroach/pull/16055
 - https://github.com/cockroachdb/cockroach/pull/16101

Amruta-Ranade:
Fixes #1499 
Fixes #674

### DIFF
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -1072,6 +1072,12 @@
           "/${VERSION}/create-a-file-server.html"
         ]
       },
+      {  
+        "title": "Rotate Security Certificates",
+        "urls": [
+          "/${VERSION}/rotate-certificates.html"
+        ]
+      },
       {
         "title": "Generate CockroachDB Resources",
         "urls": [
--- a/v1.1/create-security-certificates.md
+++ b/v1.1/create-security-certificates.md
@@ -4,11 +4,19 @@ summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and clie
 toc: false
 ---
 
-A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication and requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [command](cockroach-commands.html) with the appropriate subcommands and flags.
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communications, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [command](cockroach-commands.html) with the appropriate subcommands and flags.
 
-When using `cockroach cert` to create node and client certificates, you will need access to a local copy of the CA certificate and key. It is therefore recommended to create all certificates and keys in one place and then distribute node and client certificates and keys appropriately. For the CA key, be sure to store it somewhere safe and keep a backup; if you lose it, you will not be able to add new nodes or clients to your cluster. For a walkthrough of this process, see [Manual Deployment](manual-deployment.html).
+{{site.data.alerts.callout_success}}For details about when and how to change security certificates without restarting nodes, see <a href="rotate-certificates.html">Rotate Security Certificates</a>.{{site.data.alerts.end}}
 
 <div id="toc"></div>
+
+## How Security Certificates Work
+
+1. Using the `cockroach cert` command, you create a CA certificate and key and then node and client certificates that are signed by the CA certificate. Since you need access to a copy of the CA certificate and key to create node and client certs, it's best to create everything in one place.
+
+2. You then upload the appropriate node certificate and key and the CA certificate to each node, and you upload the appropriate client certificate and key and the CA certificate to each client.
+
+3. When nodes establish contact to each other, and when clients establish contact to nodes, they use the CA certificate to verify each other's identity.
 
 ## Subcommands
 
@@ -268,6 +276,8 @@ Certificate directory: certs
 
 ## See Also
 
-- [Manual Deployment](manual-deployment.html): Walkthrough starting a multi-node secure cluster and accessing it from a client.
-- [Start a Node](start-a-node.html): Learn more about the flags you pass when adding a node to a secure cluster.
+- [Manual Deployment](manual-deployment.html)
+- [Cloud Deployment](cloud-deployment.html)
+- [Local Test Deployment](secure-a-cluster.html)
+- [Rotate Security Certificates](rotate-certificates.html)
 - [Other Cockroach Commands](cockroach-commands.html)
--- a/v1.1/rotate-certificates.md
+++ b/v1.1/rotate-certificates.md
@@ -2,183 +2,152 @@
 title: Rotate Security Certificates
 summary: Rotate the security certificates of a secure CockroachDB cluster by creating and reloading new certificates.
 toc: false
+toc_not_nested: true
 ---
 
-<span class="version-tag">New in v1.1:</span> CockroachDB allows you to rotate security certificates without restarting the nodes. 
+<span class="version-tag">New in v1.1:</span> CockroachDB allows you to rotate security certificates without restarting nodes.
+
+{{site.data.alerts.callout_success}}For an introduction to how security certificates work in a secure CockroachDB cluster, see <a href="create-security-certificates.html">Create Security Certificates</a>.{{site.data.alerts.end}}
 
 <div id="toc"></div>
-
-## How CockroachDB Security Certificates Work
-
-A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communications, which requires CA, node, and client certificates and keys. 
-
-The node and client certificates are signed using the CA certificate. While establishing inter-node or client-node communication, the nodes and clients use the CA certificate to verify each other's identity. 
-
-For each node, the CA certificate, node certificate, and node keys are uploaded to certs directory on the node. For each client, the CA certificate, client certificate, and client keys are uploaded to the client. 
-
-For more information, see [Create security certificates](create-security-certificates.html).
 
 ## When to Rotate Certificates
 
 You may need to rotate the node, client, or CA certificates in the following scenarios:
 
-- The node, client, or CA certificates are expiring soon. 
-- Your organization's compliance policy may require periodical certificate rotation.
+- The node, client, or CA certificates are expiring soon.
+- Your organization's compliance policy requires periodical certificate rotation.
 - The key (for a node, client, or CA) is compromised.
-- You need to modify the contents of a certificate, for example, to add another DNS name/IP address of load balancer through which a node can be reached. In this case, you need to rotate only the node certificates.
+- You need to modify the contents of a certificate, for example, to add another DNS name or the IP address of a load balancer through which a node can be reached. In this case, you would need to rotate only the node certificates.
 
-## Rotating Client Certificates
+## Rotate Client Certificates
 
-Rotating a client certificate involves the following steps:
+1. Create a new client certificate and key:
 
-### Step 1. Create a new client key and certificate
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach cert create-client <username>
+    --certs-dir=certs \
+    --ca-key=my-safe-directory/ca.key
+    ~~~
 
-{% include copy-clipboard.html %}
-~~~ shell
-$ cockroach cert create-client <username>
---certs-dir=certs \
---ca-key=my-safe-directory/ca.key
-~~~
+2. Upload the new client certificate and key to the client using your preferred method.
 
-### Step 2. Upload the new client certificate and key to the client
+3. Have the client use the new client certificate.
 
-Upload the new client certificate and key to the client using your preferred method.
+    This step is application-specific and may require restarting the client.
 
-### Step 3. Have the client use the new client certificate
+## Rotate Node Certificates
 
-This step is application-specific and may require restarting the client.
+To rotate a node certificate, you create a new node certificate and key and reload them on the node.
 
-## Rotating Node Certificates 
+1. Create a new node certificate and key:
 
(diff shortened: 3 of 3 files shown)


v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
+---
+title: Rotate Security Certificates
```
mberhault: We'll want a link from `Create Security Certificates` along with a quick mention of expiration and the need for rotation.
Possibly from other places as well, but definitely that one.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+-------------|------------
+`ca.crt`     | CA certificate
+`node.crt`   | Server certificate
+`node.key`   | Key for server certificate
+`client.<user>.crt` | Client certificate for `<user>` (eg: `client.root.crt` for user `root`)
+`client.<user>.key` | Key for the client certificate
+
+The node and client certificates are signed by the CA certificate. 
+
+For each node, the CA certificate, node certificate, and node keys are uploaded to certs directory on the node. For each client, the CA certificate, client certificate, and client keys are uploaded to the client.
```
mberhault: I think mentioning the CA key is necessary. Probably best to keep it outside the `name | usage` table and emphasize that it is *NOT* copied to the nodes or clients.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+For more information, see [Create security certificates](create-security-certificates.html).
+
+## When to Rotate Certificates
+
+You may need to rotate the node, client, or CA certificates in the following scenarios:
+
+- The node, client, or CA certificates are expiring soon. 
+- Your organization's compliance policy may require periodical certificate rotation.
+- The key (for the node, client, or CA) is compromised.
+- You add a new load balancer and need to update the node certificates with the load balancer names and IP addresses. In this case, you need to rotate only the node certificates.
```
mberhault: A more general case might be better. eg: `You need to modify the contents of a certificate (eg: add another DNS name/ip address the node can be reached through, such as a load balancer)`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Rotating Client Certificates
+
+To rotate a client certificate, [generate a new client certificate](create-security-certificates.html#create-the-certificate-and-key-pair-for-a-client), upload it to the client using your preferred method, and restart the client.
```
mberhault: This is a bit short, especially compared to the node certificate details below.
Instead, we may want to spell out the exact method (`cockroach cert create-client <username>`).
Also, instead of saying `restart the client`, maybe say `Have the client use the new certificate. This is application specific and may involve a restart of the client application.`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Rotating Node Certificates 
+
+Rotating a node certificate requires generating a new node certificate and reloading the certificate on the node. Reloading the certificate involves uploading the node certificate to the node and rescanning the certs directory and using the new node certificate in the directory:
```
mberhault: I think you can skip the second sentence, it just summarizes the bullet points.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+ 
+### Why rotate CA certificates in advance 
+
+Going back to the CA certificate rotation process, the second step of the process involves rotating the node and client certificates. This involves signing the node certificates with new CA certificates. These new node certificates are uploaded to the nodes as soon as the certs directory on the node is rescanned. But the clients will get their new client certificates signed with the new CA certificates only when they are restarted. The clients won't even have the new CA certificates till they are restarted. Thus node certificates signed by the new CA certificate will not be accepted by clients that do not have the new CA certificate yet. To ensure all nodes and clients have the latest CA certificate, change CA certificates on a completely different schedule, for instance, months before changing the node and client certificates.
```
mberhault: Instead of `These new node certificates are uploaded to the nodes ...`, it should be `These new nodes certificates are used by the nodes ...`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,154 @@
...
+Going back to the CA certificate rotation process, the second step of the process involves rotating the node and client certificates. This involves signing the node certificates with new CA certificates. These new node certificates are uploaded to the nodes as soon as the certs directory on the node is rescanned. But the clients will get their new client certificates signed with the new CA certificates only when they are restarted. The clients won't even have the new CA certificates till they are restarted. Thus node certificates signed by the new CA certificate will not be accepted by clients that do not have the new CA certificate yet. To ensure all nodes and clients have the latest CA certificate, change CA certificates on a completely different schedule, for instance, months before changing the node and client certificates. 
+
+## Steps to rotate CA certificate
+
+### Step 1: Generate new CA certificate and append it to existing CA certificate:
+
+To rotate CA certificates, use the `--overwrite` flag to generate a new CA certificate and key. 
+
+~~~ shell
+$ cockroach cert create-ca --certs-dir=certs --ca-key=ca.new.key --overwrite
```
mberhault: I think we should change this to two steps:
* rename the old CA key: `mv ca.key ca.old.key`
* generate the new certificate: `cockroach cert create-ca --certs-dir=certs --ca-key=ca.key --overwrite`

This way, the instructions to use `--ca-key=ca.key` to generate node/client certificates are always accurate, even after rotation.

v1.1/certificate-rotation.md
```diff
@@ -96,26 +115,31 @@ To overcome the issue, we take advantage of the fact that multiple CA certificat
...
 
-### Step 1: Generate new CA certificate and append it to existing CA certificate:
+Suppose your existing CA and node certificates are created in the `certs` directory and your CA key is stored in a directory called `my-safe-directory`. 
 
-To rotate CA certificates, use the `--overwrite` flag to generate a new CA certificate and key. 
+Then to rotate the CA certificate, first rename the existing CA key:
 
 ~~~ shell
-$ cockroach cert create-ca --certs-dir=certs --ca-key=ca.new.key --overwrite 
+$ mv ca.key ca.old.key
```
mberhault: We should point to the same path as below: `my-safe-directory/ca.key` and `my-safe-directory/ca.old.key`

Amruta-Ranade: @mberhault Thank you! Although, I am not clear on why we need to do this. As in, why do we need to retain the old key when we do not yet support signing new node/client certificates with old  CA certificates?

mberhault: For one, because we will need the key once we allow signing with the old CA. But also because it's rarely a good idea to just delete a key completely.

mberhault:
Still LGTM. Do you want someone else's input on this? Might be good to have Jesse or Sean take a look too.

jseldess:
I'm hoping to review this soon.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+## How CockroachDB Security Certificates Work
+
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication and requires CA, node, and client certificates and keys:
+
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate
+`node.crt`   | Server certificate
+`node.key`   | Key for server certificate
+`client.<user>.crt` | Client certificate for `<user>` (eg: `client.root.crt` for user `root`)
```
jseldess: nit: `e.g.,` instead of `eg:`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+title: Rotate Security Certificates
+summary: Rotate the security certificates of a secure CockroachDB cluster by generating and reloading new certificates.
+toc: false
+---
+
+<span class="version-tag">New in v1.1:</span> CockroachDB allows you to rotate security certificates without restarting the nodes. 
+
+<div id="toc"></div>
+
+## How CockroachDB Security Certificates Work
```
jseldess: Most of this is a repetition of https://www.cockroachlabs.com/docs/dev/create-security-certificates.html#certificate-directory. And this isn't really "how cert rotation works". Let's remove from this page and expand the section I linked to. We can add a link to that section in this page's intro.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+
+For more information, see [Create security certificates](create-security-certificates.html).
+
+## When to Rotate Certificates
+
+You may need to rotate the node, client, or CA certificates in the following scenarios:
+
+- The node, client, or CA certificates are expiring soon. 
+- Your organization's compliance policy may require periodical certificate rotation.
+- The key (for the node, client, or CA) is compromised.
```
jseldess: nit: `for a node, a client, or the CA`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+- You need to modify the contents of a certificate (example: add another DNS name/IP address the node can be reached through, such as a load balancer). In this case, you need to rotate only the node certificates.
```
jseldess: Let's integrate the parenthetical into the sentence:

> You need to modify the contents of a certificate, for example, to add the DNS name/IP address of a load balancer through which a node can be reached. In this case, you need to rotate only the node certificates.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+
+To make a running node rescan the certificates directory and use the new certificates without restarting the node, issue a `SIGHUP` signal to the cockroach process:
+	
+~~~ shell
+pkill -SIGHUP -x cockroach
+~~~
+
+The `SIGHUP` signal must be sent by the same user running the process (for example: run with `sudo` if cockroach is running under user `root`).	
+
+## Rotating CA Certificate
```
jseldess: nit: `Rotating the CA Certificate`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+- The node, client, or CA certificates are expiring soon. 
+- Your organization's compliance policy may require periodical certificate rotation.
+- The key (for the node, client, or CA) is compromised.
+- You need to modify the contents of a certificate (example: add another DNS name/IP address the node can be reached through, such as a load balancer). In this case, you need to rotate only the node certificates.
+
+## Rotating Client Certificates
+
+Rotating a client certificate involves the following steps:
+
+#### Step 1: Generating a new client key and certificate
```
jseldess: Let's make these steps h3 headers (`###`). Also, let's use a period and active verbs for steps:

> Step 1. Generate a new client key and certificate

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+Upload the new certificate to the client using your preferred method.
+
+#### Step 3: Have the client use the new certificate
+
+This step is application-specific and may require restarting the client.
+
+
+## Rotating Node Certificates 
+
+Rotating a node certificate requires generating a new node certificate and reloading the certificate on the node. Reloading the certificate involves uploading the node certificate to the node and rescanning the certs directory and using the new node certificate in the directory:
```
jseldess: Let's condense this to just the first sentence. The steps handle the rest.

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Step 1: Generate new node certificate and key
```
jseldess: Missing the `a`: `Step 1. Generate a new node certificate and key`

v1.1/certificate-rotation.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Step 3: Reload the node certificate
+
+To make a running node rescan the certificates directory and use the new certificates without restarting the node, issue a `SIGHUP` signal to the cockroach process:
+	
+~~~ shell
+pkill -SIGHUP -x cockroach
+~~~
+
+The `SIGHUP` signal must be sent by the same user running the process (for example: run with `sudo` if cockroach is running under user `root`).
```
jseldess: @mberhault, is there a way to validate that the new certificate is in use? That might be a final step here.

v1.1/certificate-rotation.md
jseldess: I'll get back to reviewing this later or tomorrow, but my initial sense is that there's way too much prose before the actual procedure. Need to think about this.

v1.1/certificate-rotation.md
mberhault: Two ways:
1. look at the certificate report on the node (https://omega.crdb.io:8080/#/reports/certificates/1 for node 1)
1. look at monitoring (we should have a section on that)

Amruta-Ranade: I don't think we expose the certificate report to the users. 
Also, is this the monitoring section you are referring to: https://www.cockroachlabs.com/docs/stable/create-security-certificates.html#list-certificates-and-keys

mberhault: It's linked to from the `/debug/` page, assuming access is allowed (by a cluster setting).
By monitoring, I mean the actual metrics: `security_certificate_expiration_node` and `security_certificate_expiration_ca`

PR #1941, opened 2017-09-30
Update txn-sample.rs

tschottdorf:
Update the example to work against postgres-0.15.1.


nvanbenschoten:
LGTM, but could you point at the change that broke the previous version?

tschottdorf:
@nvanbenschoten done, is that what you had in mind?

nvanbenschoten:
Yep, thanks!

PR #1944, opened 2017-10-02
Created openssl document

### LINKED ISSUE #119: certs: provide openssl commands to generate cockroach-compatible certs.

Amruta-Ranade:
Closes #119

### DIFF
--- a/_includes/prod_deployment/secure-generate-certificates.md
+++ b/_includes/prod_deployment/secure-generate-certificates.md
@@ -1,4 +1,4 @@
-You can use either `cockroach cert` commands or [`openssl` commands](create-security-certificates-openssl.html) to generate the security certificates. This section discusses the procedure to generate certificates using the `cockroach cert` commands.
+You can use either `cockroach cert` commands or [`openssl` commands](create-security-certificates-openssl.html) to generate security certificates. This section features the `cockroach cert` commands.
 
 Locally, you'll need to [create the following certificates and keys](create-security-certificates.html):
 
--- a/_includes/prod_deployment/secure-generate-certificates.md
+++ b/_includes/prod_deployment/secure-generate-certificates.md
@@ -1,3 +1,5 @@
+You can use either `cockroach cert` commands or [`openssl` commands](create-security-certificates-openssl.html) to generate the security certificates. This section discusses the procedure to generate certificates using the `cockroach cert` commands.
+
 Locally, you'll need to [create the following certificates and keys](create-security-certificates.html):
 
 - A certificate authority (CA) key pair (`ca.crt` and `ca.key`).
--- a/v1.1/create-security-certificates-openssl.md
+++ b/v1.1/create-security-certificates-openssl.md
@@ -5,8 +5,8 @@ toc: false
 ---
 
 <div class="filters filters-big clearfix">
-  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use Cockroach Cert</button>
-  <button style="width:28%" class="filter-button current">Use <strong>OpenSSL</strong></button></a>
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use cockroach cert</button>
+  <button style="width:28%" class="filter-button current">Use <strong>openssl</strong></button></a>
 </div>
 
 A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
@@ -69,9 +69,9 @@ Note the following:
     - `certs`: Create your CA certificate and all node and client certificates and keys in this directory and then upload the relevant files to the nodes and clients.
     - `my-safe-directory`: Create your CA key in this directory and then reference the key when generating node and client certificates. After that, keep the key safe and secret; do not upload it to your nodes or clients.
 
-2. Create the `ca.cnf` file:
+2. Create the `ca.cnf` file and copy the following configuration into it.
 
-    Create the `ca.cnf` file and copy the following configuration into it. You can set the CA certificate expiration period using the `default_days` parameter. We recommend using the CockroachDB default value of the CA certificate expiration period, which is 3660 days.
+    You can set the CA certificate expiration period using the `default_days` parameter. We recommend using the CockroachDB default value of the CA certificate expiration period, which is 3660 days.
 
     {% include copy-clipboard.html %}
     ~~~ shell
@@ -254,7 +254,7 @@ In the following steps, replace the placeholder text in the code with the actual
 
 7. Repeat steps 1 - 6 for each additional node.
 
-8. Remove the .pem files in the `certs` directory. These files are unnecessary duplicates of the .crt files that CockroachDB requires.
+8. Remove the `.pem` files in the `certs` directory. These files are unnecessary duplicates of the `.crt` files that CockroachDB requires.
 
 ### Create the certificate and key pair for a client
 
@@ -321,7 +321,7 @@ In the following steps, replace the placeholder text in the code with the actual
 
 6. Repeat steps 1 - 5 for each additional client.
 
-7. Remove the .pem files in the `certs` directory. These files are unnecessary duplicates of the .crt files that CockroachDB requires.
+7. Remove the `.pem` files in the `certs` directory. These files are unnecessary duplicates of the `.crt` files that CockroachDB requires.
  
 ## See Also
 
--- a/v1.1/create-security-certificates-openssl.md
+++ b/v1.1/create-security-certificates-openssl.md
@@ -1,0 +1,329 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: false
+---
+
+<div class="filters filters-big clearfix">
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use Cockroach Cert</button>
+  <button style="width:28%" class="filter-button current">Use <strong>OpenSSL</strong></button></a>
+</div>
+
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
+
+<div id="toc"></div>
+
+## Subcommands
+
+Subcommand | Usage
+-----------|------
+[`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) | Create an RSA private key.
+[`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) | Create CA certificate and CSRs (certificate signing requests).
+[`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) | Create node and client certificates using the CSRs.
+
+## Configuration Files
+
+To use [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) and [`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) subcommands, you need the following configuration files:
+
+File name pattern | File usage
+-------------|------------
+`ca.cnf`     | CA configuration file
+`node.cnf`   | Server configuration file
+`client.cnf` | Client configuration file
+
+## Certificate Directory
+
+To create node and client certificates using the OpenSSL commands, you need access to a local copy of the CA certificate and key. We recommend creating all certificates (node, client, and CA certificates), and node and client keys in one place and then distributing them appropriately. Store the CA key somewhere safe and keep a backup; if you lose it, you will not be able to add new nodes or clients to your cluster. 
+
+Use the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) and [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) subcommands to create all certificates, and node and client keys in a single directory, with the files named as follows:
+
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate
+`node.crt`   | Server certificate
+`node.key`   | Key for server certificate
+`client.<user>.crt` | Client certificate for `<user>` (for example: `client.root.crt` for user `root`)
+`client.<user>.key` | Key for the client certificate
+
+Note the following:
+
+- The CA key should not be uploaded to the nodes and clients, so it should be created in a separate directory.
+
+- Keys (files ending in `.key`) must not have group or world permissions (maximum permissions are 0700, or `rwx------`). This check can be disabled by setting the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK=true`.
+
+## Examples
+
+### Create the CA key and certificate pair
+
+1. Create two directories:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir certs
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir my-safe-directory
+    ~~~
+    - `certs`: Create your CA certificate and all node and client certificates and keys in this directory and then upload the relevant files to the nodes and clients.
+    - `my-safe-directory`: Create your CA key in this directory and then reference the key when generating node and client certificates. After that, keep the key safe and secret; do not upload it to your nodes or clients.
+
+2. Create the `ca.cnf` file:
+
+    Create the `ca.cnf` file and copy the following configuration into it. You can set the CA certificate expiration period using the `default_days` parameter. We recommend using the CockroachDB default value of the CA certificate expiration period, which is 3660 days.
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # OpenSSL CA configuration file
+    [ ca ]
+    default_ca = CA_default
+
+    [ CA_default ]
+    default_days = 3660
+    database = index.txt
+    serial = serial.txt 
+    default_md = sha256 
+    copy_extensions = copy
+
+    # Used to create the CA certificate.
+    [ req ]
+    prompt=no
+    distinguished_name = distinguished_name
+    x509_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = Cockroach CA
+
+    [ extensions ]
+    keyUsage = critical,digitalSignature,nonRepudiation,keyEncipherment,keyCertSign
+    basicConstraints = critical,CA:true,pathlen:1
+
+    # Common policy for nodes and users.
+    [ signing_policy ]
+    organizationName = supplied
+    commonName = supplied
+
+    # Used to sign node certificates.
+    [ signing_node_req ]
+    keyUsage = critical,digitalSignature,keyEncipherment
+    extendedKeyUsage = serverAuth,clientAuth
+
+    # Used to sign client certificates.
+    [ signing_client_req ]
+    keyUsage = critical,digitalSignature,keyEncipherment
+    extendedKeyUsage = clientAuth
+    ~~~
+
+    {{site.data.alerts.callout_info}}The <code>keyUsage</code> and <code>extendedkeyUsage</code> parameters are vital for CockroachDB functions. You can modify or omit other parameters as per your preferred OpenSSL configuration, but do not omit the <code>keyUsage</code> and <code>extendedkeyUsage</code> parameters. {{site.data.alerts.end}}
+  
+3. Create the CA key using the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl genrsa -out my-safe-directory/ca.key 2048
+    ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell 
+    $ chmod 400 my-safe-directory/ca.key
+    ~~~
+
+4. Create the CA certificate using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl req \
+    -new \
+    -x509 \
+    -config ca.cnf \
+    -key my-safe-directory/ca.key \
+    -out certs/ca.crt \
+    -days 3660 \
+    -batch
+    ~~~
+
+5. Reset database and index files.
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ rm -f index.txt serial.txt
+    ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ touch index.txt
+    ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ echo '01' > serial.txt
+    ~~~
+
+### Create the certificate and key pairs for nodes
+
+In the following steps, replace the placeholder text in the code with the actual username and node address.
+
+1. Create the `node.cnf` file for the first node and copy the following configuration into it:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # OpenSSL node configuration file
+    [ req ]
+    prompt=no
+    distinguished_name = distinguished_name
+    req_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = node
+
+    [ extensions ]
+    subjectAltName = DNS:<node-hostname>,DNS:<node-domain>,IP:<IP Address>
+    ~~~
+
+    {{site.data.alerts.callout_info}}The <code>commonName</code> and <code>subjectAltName</code> parameters are vital for CockroachDB functions. It is also important that <code>commonName</code> be set to <code>node</code>. You can modify or omit other parameters as per your preferred OpenSSL configuration, but do not omit the <code>commonName</code> and <code>subjectAltName</code> parameters.  {{site.data.alerts.end}}
+  
+2. Create the key for the first node using the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl genrsa -out certs/node.key 2048 
+    ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ chmod 400 certs/node.key
+    ~~~
+
+3. Create the CSR for the first node using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create Node certificate signing request.
+    $ openssl req \
+    -new \
+    -config node.cnf \
+    -key certs/node.key \
+    -out node.csr \
+    -batch
+    ~~~
+
+4. Sign the node CSR to create the node certificate for the first node using the [`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) command. 
+
+    You can set the node certificate expiration period using the `days` flag. We recommend using the CockroachDB default value of the node certificate expiration period, which is 1830 days.
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Sign the CSR using the CA key.
+    $ openssl ca \
+    -config ca.cnf \
+    -keyfile my-safe-directory/ca.key \
+    -cert certs/ca.crt \
+    -policy signing_policy \
+    -extensions signing_node_req \
+    -out certs/node.crt \
+    -outdir certs/ \
+    -in node.csr \
+    -days 1830 \
+    -batch
+    ~~~
+
+5. Upload certificates to the first node:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create the certs directory:
+    $ ssh <username>@<node1 address> "mkdir certs"
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Upload the CA certificate and node certificate and key:
+    $ scp certs/ca.crt \
+    certs/node.crt \
+    certs/node.key \
+    <username>@<node1 address>:~/certs
+    ~~~
+
+6. Delete the local copy of the first node's certificate and key:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ rm certs/node.crt certs/node.key
+    ~~~
+
+    {{site.data.alerts.callout_info}}This is necessary because the certificates and keys for additional nodes will also be named <code>node.crt</code> and <code>node.key</code>.{{site.data.alerts.end}}
+
+7. Repeat steps 1 - 6 for each additional node.
+
+8. Remove the .pem files in the `certs` directory. These files are unnecessary duplicates of the .crt files that CockroachDB requires.
+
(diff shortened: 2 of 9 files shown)


v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+    prompt=no
+    distinguished_name = distinguished_name
+    x509_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = Cockroach CA
+
+    [ extensions ]
+    keyUsage = critical,digitalSignature,nonRepudiation,keyEncipherment,keyCertSign
```
mberhault: We should point out the important part of the file.
For `ca.conf`, that's `keyUsage` and `extendedKeyUsage` in multiple sections.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create CA certificate.
+    openssl req \
+      -new \
+      -x509 \
+      -config ca.conf \
+      -key ${CCERTS}/ca.key \
+      -out ${OSCERTS}/ca.crt \
+      -days 3660 \
```
mberhault: We should point out the reasoning behind some flags.
Here, just `-days 3660` to match the `cockroach cert` default values.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+    prompt=no
+    distinguished_name = distinguished_name
+    req_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = <node-hostname>
+
+    [ extensions ]
+    subjectAltName = DNS:<node-hostname>,DNS:<node-domain>,IP:<IP Address>
```
mberhault: point out `subjectAltName` and `commonName`

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    [ req ]
+    prompt=no
+    distinguished_name = distinguished_name
+    req_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = <node-hostname>
```
mberhault: this needs to be `node`, as in that actual string, not the hostname of the machine. When used as a client certificate, this is the username used by the various nodes.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+3. Generate the CSR for the first node using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create Node certificate signing request.
+    openssl req \
+      -new \
+      -config node.conf \
+      -key ${certs}/node.key \
+      -out ${certs}/node.csr \
```
mberhault: we want to store the csr somewhere else so it doesn't end up in the final `certs` directory. Could be `/tmp/` or even just the current directory (eg: `-out /tmp/node.csr` or `-out node.csr`).
Alternatively, we could delete it after generating the final `.crt`

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    [ req ]
+    prompt=no
+    distinguished_name = distinguished_name
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = <username>
```
mberhault: point out `commonName` here.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+When using OpenSSL commands to create node and client certificates, you will need access to a local copy of the CA certificate and key. It is therefore recommended to create all certificates and keys in one place and then distribute node and client certificates and keys appropriately. For the CA key, be sure to store it somewhere safe and keep a backup; if you lose it, you will not be able to add new nodes or clients to your cluster. For a walkthrough of this process, see [Manual Deployment](manual-deployment.html).
+
+<div id="toc"></div>
+
+## Subcommands
+
+Subcommand | Usage
+-----------|------
+[`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) | Generate an RSA private key
+[`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) | Create a private key and a CSR (certificate signing request)
```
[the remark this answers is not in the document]
Amruta-Ranade: @mberhault To be specific, this would be "self-signed CA certificate and CSR", right?

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,292 @@
...
+[`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) | Issue a root CA certificate based on the CSR.
```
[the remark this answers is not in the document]
Amruta-Ranade: @mberhault Not sure what "CA operations" means. Would "Sign the CSR using the CA key" make more sense?

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: @mberhault I was actually wondering about this. Because we are no longer trying to keep the openssl doc closely parallel to the cert commands doc, can we leave out this parameter? The users can set it to the value they want, right?

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: On reading about commonName on Google, I still think it should be the hostname of the node. Can you clarify please?

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Point out how?

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Can we leave out the directory altogether? I remember discussing that the working directory will be used in that case, and that will be okay.

v1.1/create-security-certificates-openssl.md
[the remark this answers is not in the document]
Amruta-Ranade: "Create CA certificate and CSR"

v1.1/create-security-certificates-openssl.md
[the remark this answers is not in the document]
Amruta-Ranade: "Create node and client certificates using CSR"

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: "Do not change this, else CRDB won't work".

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: "Recommended days for expiration of CA certificate"

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: "node"

Amruta-Ranade: Also, "do not change this"

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: "Do not omit this"

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: For CSR, leave out the directory

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+      -cert certs/ca.crt \
+      -policy signing_policy \
+      -extensions signing_node_req \
+      -out certs/node.crt \
+      -outdir certs/ \
+      -in node.csr \
+      -days 1830 \
+      -batch
+    ~~~
+
```
mberhault: I missed a step here. Turns out openssl also writes the plaintext certificate (as opposed to the encoded certificate `node.crt`) to `certs/<certificate id>.pem`.
We can either:
1. change outdir to `/tmp/`
2. run `rm certs/*.pem` to delete that file

The file is not needed since it's the same as `node.crt`, just a different encoding.
It doesn't hurt to leave it, but that's just extra stuff lying around.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+      -cert certs/ca.crt \
+      -policy signing_policy \
+      -extensions signing_client_req \
+      -out certs/client.maxroach.crt \
+      -outdir certs/ \
+      -in certs/client.maxroach.csr \
+      -days 1830 \
+      -batch
+    ~~~    
+
```
mberhault: same here regarding the `certs/<certificate id>.pem` file.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+3. Create the CSR for the first client using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create client certificate signing request
+    $ openssl req \
+      -new \
+      -config client.cnf \
+      -key certs/client.maxroach.key \
```
mberhault: we've been using `<username>` in the config file and commands above. We should do the same here, or use `maxroach` everywhere.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: So we either change outdir to tmp OR delete the file?

mberhault: Correct, we need one or the other. A third possibility is to just leave it and hope noone minds. Since we scp only the files we need, that might be fine.

Amruta-Ranade: Are we sure the plaintext file won't be needed for something else in the user's workflow? There might be a reason OpenSSL creates it in the first place.

Amruta-Ranade: I don't think we should change the outdir to tmp instead of certs - that will break the parallelism with the cockroach cert command doc. I prefer either removing the file or letting it be.

mberhault: The plaintext file is needed if we don't use `--out node.crt`. By default, openssl doesn't create that `node.crt` file, just the plaintext one. However, there's no option to have the encoded file without the plaintext file.
It's fine keeping it though.

Amruta-Ranade: As discussed, we are keeping the file :)

Amruta-Ranade:
To-Do: Figure out if/how the certificate rotation and other related docs need to be updated with openssl workflows.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates.md
```diff
@@ -1,15 +1,20 @@
...
+<div class="filters filters-big clearfix">
+  <button class="filter-button current"><strong>Cockroach Cert Commands</strong></button>
+  <a href="create-security-certificates-openssl.html"><button class="filter-button">OpenSSL Commands</button></a>
+</div>
 
 <div id="toc"></div>
 
+When using `cockroach cert` to create node and client certificates, you will need access to a local copy of the CA certificate and key. It is therefore recommended to create all certificates and keys in one place and then distribute node and client certificates and keys appropriately. For the CA key, be sure to store it somewhere safe and keep a backup; if you lose it, you will not be able to add new nodes or clients to your cluster. For a walkthrough of this process, see [Manual Deployment](manual-deployment.html).
```
jseldess: In the openssl version of the page, you put this info at the start of the `Certificate Directory` section. I like that placement. Can you do the same on this page?

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: false
+---
+
+A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [OpenSSL commands](https://wiki.openssl.org/index.php/).
```
jseldess: nit: since we say "cockroach cert commands", I'd suggest we say "openssl commands".

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+
+<div class="filters filters-big clearfix">
```
jseldess: Can we try putting these toggles at the top of the page, like we do [here](https://www.cockroachlabs.com/docs/stable/build-a-go-app-with-cockroachdb.html)?

Also, I'd suggest trying this text:
- Use Cockroach Commands
- Use OpenSSL Commands

We may have to force wider toggles. See https://github.com/cockroachdb/docs/blame/master/v1.1/build-a-python-app-with-cockroachdb.md#L10 for an example.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir my-safe-directory
+    ~~~
+    - `certs`: Create your CA certificate and all node and client certificates and keys in this directory and then upload the relevant files to the nodes and clients.
+    - `my-safe-directory`: Create your CA key in this directory and then reference the key when generating node and client certificates. After that, keep the key safe and secret; do not upload it to your nodes or clients.
+
+2. Create the ca.cnf file:
```
jseldess: Let's code-format `ca.cnf`.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+
+    Create the ca.cnf file and copy the following configuration into it. You can set the CA certificate expiration period using the `default_days` parameter. We recommend using the CockroachDB default value of the CA certificate expiration period, which is 3660 days.
```
jseldess: Same as above.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    keyUsage = critical,digitalSignature,keyEncipherment
+    extendedKeyUsage = clientAuth
+    ~~~
+
+    {{site.data.alerts.callout_info}}The <code>keyUsage</code> and <code>extendedkeyUsage</code> parameters are vital for CockroachDB functions. You can modify or omit other parameters as per your preferred OpenSSL configuration, but do not omit the <code>keyUsage</code> and <code>extendedkeyUsage</code> parameters. {{site.data.alerts.end}}
+  
+3. Create the CA key using the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
```
jseldess: Let's split these two commands into two code blocks, each with copy-clipboard.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    $ openssl genrsa -out my-safe-directory/ca.key 2048 
+    $ chmod 400 my-safe-directory/ca.key
+    ~~~
+
+4. Create the CA certificate using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create CA certificate.
```
jseldess: Let's remove this comment. Already stated in step.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    $ openssl req \
+      -new \
+      -x509 \
+      -config ca.cnf \
+      -key my-safe-directory/ca.key \
+      -out certs/ca.crt \
+      -days 3660 \
+      -batch
+
+    # Reset database and index files.
```
jseldess: Let's make this a step 5. Also, please put each command in a separate code block, each with copy-clipboard.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    $ rm -f index.txt serial.txt
+    $ touch index.txt
+    $ echo '01' > serial.txt
+    ~~~
+
+### Create the certificate and key pairs for nodes
+
+1. Create the node.cnf file for the first node and copy the following configuration into it:
```
jseldess: Need to code-format the filename.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    ~~~
+
+    {{site.data.alerts.callout_info}}The <code>commonName</code> and <code>subjectAltName</code> parameters are vital for CockroachDB functions. It is also important that <code>commonName</code> be set to <code>node</code>. You can modify or omit other parameters as per your preferred OpenSSL configuration, but do not omit the <code>commonName</code> and <code>subjectAltName</code> parameters.  {{site.data.alerts.end}}
+  
+
+2. Create the key for the first node using the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl genrsa -out certs/node.key 2048
```
jseldess: Split these into separate code blocks.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    prompt=no
+    distinguished_name = distinguished_name
+    req_extensions = extensions
+
+    [ distinguished_name ]
+    organizationName = Cockroach
+    commonName = node
+
+    [ extensions ]
+    subjectAltName = DNS:<node-hostname>,DNS:<node-domain>,IP:<IP Address>
```
jseldess: We need to call out that these placeholders need to be replace with real values.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: @jseldess The reason I put the toggles after the first paragraph is that the paragraph is common to both. So if we move the toggles to the top, I will have to repeat the para on both pages. Not that I mind doing that. Thoughts?

v1.1/create-security-certificates.md
Amruta-Ranade: Umm..I think in both docs I place the info right after the toc. In any case, I think we need that info in the beginning, because the `Subcommands` section needs it.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    $ rm certs/node.crt certs/node.key
+    ~~~
+
+    {{site.data.alerts.callout_info}}This is necessary because the certificates and keys for additional nodes will also be named <code>node.crt</code> and <code>node.key</code>.{{site.data.alerts.end}}
+
+7. Repeat steps 1 - 6 for each additional node.
+
+### Create the certificate and key pair for a client
+
+1. Create the client.cnf file for the first client and copy the following configuration into it:
```
jseldess: Need to code-format `client.cnf`.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,306 @@
...
+    ~~~
+
+    {{site.data.alerts.callout_info}}The <code>commonName</code> parameter is vital for CockroachDB functions. You can modify or omit other parameters as per your preferred OpenSSL configuration, but do not omit the <code>commonName</code> parameter.  {{site.data.alerts.end}}
+  
+
+2. Create the key for the first client using the [`openssl genrsa`](https://wiki.openssl.org/index.php/Manual:Genrsa(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl genrsa -out certs/client.<username>.key 2048
```
jseldess: Split these commands into two code blocks.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Reworded toggle text, increased width.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates-openssl.md
Amruta-Ranade: Done.

v1.1/create-security-certificates.md
jseldess: Hmm. I must've been looking at an outdated version or something. In any case, I'm glad the content is consistently place, but it feels a bit strange orphaned below the TOC. 

- Can you perhaps add a h2 header `Considerations`?
- Can you make sure the formulations are identify on both pages, aside from necessary differences?

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: false
+---
+
+A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
+
+<div class="filters filters-big clearfix">
```
jseldess: Looks like one of my comments might have gotten lost. In addition to renaming these, can you move them to the top of the page, like we do here: https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb.html

Amruta-Ranade: My reply got lost too :( I replied saying that the first paragraph is common to both the docs, so I would prefer to keep them it at the top, then segue into the links for each doc. But that's just my preference, I am open to changing it too.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use Cockroach Cert Commands</button>
+  <button style="width:28%" class="filter-button current"><strong>Use OpenSSL Commands</strong></button></a>
```
jseldess: Can you start the `<strong>` emphasis at `OpenSSL`? Same on non-openssl page.

The button text now feels too verbose. Could we try:
- Use **cockroach cert**
- Use **openssl**

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+    ~~~ shell 
+    $ chmod 400 my-safe-directory/ca.key
+    ~~~
+
+4. Create the CA certificate using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl req \
+      -new \
```
jseldess: Can you de-indent all lines but the first. That way, when pasted into the terminal, there aren't extra spaces between flags.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+    $ chmod 400 certs/node.key
+    ~~~
+
+3. Create the CSR for the first node using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create Node certificate signing request.
+    $ openssl req \
+      -new \
```
jseldess: Same as above: Please de-indent everything but the first line.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+
+4. Sign the node CSR to create the node certificate for the first node using the [`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) command. 
+
+    You can set the node certificate expiration period using the `days` flag. We recommend using the CockroachDB default value of the node certificate expiration period, which is 1830 days.
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Sign the CSR using the CA key.
+    $ openssl ca \
+      -config ca.cnf \
```
jseldess: Please de-indent.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+    $ chmod 400 certs/client.<username>.key
+    ~~~
+
+3. Create the CSR for the first client using the [`openssl req`](https://wiki.openssl.org/index.php/Manual:Req(1)) command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    # Create client certificate signing request
+    $ openssl req \
+      -new \
```
jseldess: Please de-indent.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+      -out client.<username>.csr \
+      -batch
+    ~~~
+
+4. Sign the client CSR to create the client certificate for the first client using the [`openssl ca`](https://wiki.openssl.org/index.php/Manual:Ca(1)) command. You can set the client certificate expiration period using the `days` flag. We recommend using the CockroachDB default value of the client certificate expiration period, which is 1830 days.
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ openssl ca \
+      -config ca.cnf \
```
jseldess: Please de-indent.

v1.1/create-security-certificates-openssl.md
jseldess: If you're willing, I'd prefer them at the top, like we've done on most other pages, for example:
- https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb.html
- https://www.cockroachlabs.com/docs/stable/start-a-local-cluster.html
- https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-on-aws.html

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+6. Delete the local copy of the first node's certificate and key:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ rm certs/node.crt certs/node.key
+    ~~~
+
+    {{site.data.alerts.callout_info}}This is necessary because the certificates and keys for additional nodes will also be named <code>node.crt</code> and <code>node.key</code>.{{site.data.alerts.end}}
+
+7. Repeat steps 1 - 6 for each additional node.
```
jseldess: Marc and I realized that openssl creates `.pem` files that are exact duplicates of the `.crt` files. So let's add a step `8`:

> 8. Remove the `.pem` files in the `certs`. These files are unnecessary duplicates of the `.crt` files that CockroachDB requires.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,326 @@
...
+      -out certs/client.<username>.crt \
+      -outdir certs/ \
+      -in client.<username>.csr \
+      -days 1830 \
+      -batch
+    ~~~    
+
+5. Upload certificates to the first client using your preferred method. 
+
+6. Repeat steps 1 - 5 for each additional client.
```
jseldess: Same as above. Let's add a step `7`:

> 7. Remove the `.pem` files in the `certs`. These files are unnecessary duplicates of the `.crt` files that CockroachDB requires.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,329 @@
...
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ rm certs/node.crt certs/node.key
+    ~~~
+
+    {{site.data.alerts.callout_info}}This is necessary because the certificates and keys for additional nodes will also be named <code>node.crt</code> and <code>node.key</code>.{{site.data.alerts.end}}
+
+7. Repeat steps 1 - 6 for each additional node.
+
+8. Remove the .pem files in the `certs` directory. These files are unnecessary duplicates of the .crt files that CockroachDB requires.
```
jseldess: Need to code-format `.pem` and `.crt`.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,329 @@
...
+    -in client.<username>.csr \
+    -days 1830 \
+    -batch
+    ~~~    
+
+5. Upload certificates to the first client using your preferred method. 
+
+6. Repeat steps 1 - 5 for each additional client.
+
+7. Remove the .pem files in the `certs` directory. These files are unnecessary duplicates of the .crt files that CockroachDB requires.
```
jseldess: Need to code-format `.pem` and `.crt`

_includes/prod_deployment/secure-generate-certificates.md
```diff
@@ -1,3 +1,5 @@
+You can use either `cockroach cert` commands or [`openssl` commands](create-security-certificates-openssl.html) to generate the security certificates. This section discusses the procedure to generate certificates using the `cockroach cert` commands.
```
jseldess: Can tighten this up a bit:

In the first sentence: `generate the security certificates` > `generate security certificates`

The second sentence can be cut to: `This section features the `cockroach cert` commands.`

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,329 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: false
+---
+
+<div class="filters filters-big clearfix">
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use Cockroach Cert</button>
```
jseldess: Since we're referring to commands, I think we should use the lowercase for `cockroach cert` and `openssl`. I don't think we can code-format inside a button, but the lowercase and bold (when selected) should be enough.

v1.1/create-security-certificates-openssl.md
```diff
@@ -0,0 +1,329 @@
...
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir my-safe-directory
+    ~~~
+    - `certs`: Create your CA certificate and all node and client certificates and keys in this directory and then upload the relevant files to the nodes and clients.
+    - `my-safe-directory`: Create your CA key in this directory and then reference the key when generating node and client certificates. After that, keep the key safe and secret; do not upload it to your nodes or clients.
+
+2. Create the `ca.cnf` file:
```
jseldess: nit: I think we can simplify this to

> 2. Create the `ca.cnf` file and copy the following configuration into it.

with that sentence removed from the paragraph following. So that paragraph would just start with `You can set the...`.

PR #1956, opened 2017-10-04
remove code-based doc generation

dt:
the generation tools now live along side the code from which they generate in the main repo.
the generated content is also checked in to the main repo with CI jobs verifying it stays up-to-date.
the docs repo now just copies those generated files from the main repo to update them, rather than running its own generation.


benesch:
LGTM! You documented the new process for copying files over somewhere, right?

jseldess:
@dt, please see my comment [here](https://github.com/cockroachdb/cockroach/pull/19043#issuecomment-334328297). Let's please wait until after the 1.1 release before making this change official since there's still lots of sql docs work in flight, and a change to the workflow now will slow us down.

jseldess:
Also, in the meantime, would you be willing to update this wiki page with new instructions from generating sql diagrams and functions? Or point me to where this guidance lives (again)?

dt:
@jseldess yep, we can sit on this until we're ready to kill the generator on the docs side. we'll be able to diff the files in main repo to ensure that we caught all the changes before we remove this one too.

mjibson:
LGTM

PR #1970, opened 2017-10-07
Remove `cockroach version` step from install docs

### LINKED ISSUE #1939: Remove `cockroach version` step from install instructions
Re: [Install CockroachDB](https://cockroachlabs.com/docs//v1.0/install-cockroachdb.html)

## Issue Description
Do we really need step 3?

We currently have a third step to the install: "Make sure the CockroachDB executable works" -- how often does this not work? and if it didnt work from here we don't have any other next steps to fix it. I'm thinking we can cut this step and make the third step the add to the email list one. At least AB test having 3 steps instead of 4.

jseldess:
Fixes #1939 

Exception: Didn't remove from the Windows binary installation instructions because the step is unfortunately the only example of running a cockroach command in a Windows environment (powershell).

### DIFF
--- a/v1.0/install-cockroachdb.html
+++ b/v1.0/install-cockroachdb.html
@@ -112,11 +112,6 @@ $(document).ready(function(){
       <p>If you get a permissions error, prefix the command with <code>sudo</code>.</p>
     </li>
     <li>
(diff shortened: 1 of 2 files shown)


nstewart:
LGTM

PR #1974, opened 2017-10-08
Fix 1.1 upgrade doc

### LINKED ISSUE #1964: Upgrade a Cluster's Version Doc Update
Running through these docs:

Step 1 Num2: decommission flag is not available in 1.0, so this doesn't work
Step 1 Num3: same thing, ranges flag is not available in 1.0, so this doesn't work
Step 2 Num2: you can't cp the path in while the cockroach node is still running

jseldess:
Fixes #1964 

- Use callout at top to make the upgrade path clearer.
- Fix prepare step: Encouraging users to check that all nodes are on the same version before upgrading. Also add note about using `--ranges` flag when upgrading within 1.1 series.
- Fix the node upgrade sequence: Previously, we were telling users to download the new binary
and move it onto the path before stoping and renaming the old binary. Now, we tell users to stop the process, download the new binary, and then rename the old binary and move the new binary onto the `PATH`.
- Improve usability: Use max/linux toggles.

### DIFF
--- a/v1.0/upgrade-cockroach-version.md
+++ b/v1.0/upgrade-cockroach-version.md
@@ -38,47 +38,7 @@ For each node in your cluster, complete the following steps.
 
 1. Connect to the node.
 
-2. Download and install the CockroachDB binary you want to use:
-    - **Mac**:
-
-        {% include copy-clipboard.html %}
-        ~~~ shell
-        # Get the CockroachDB tarball:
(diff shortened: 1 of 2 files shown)


dianasaur323:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/1974)*

PR #1977, opened 2017-10-08
Clarify that root user must user cert-based authentication

### LINKED ISSUE #1700: cli: return an error if using root without valid certificates
PR: https://github.com/cockroachdb/cockroach/pull/16953

jseldess:
Fixes #1700


mberhault:
LGTM

PR #1978, opened 2017-10-08
Update identifier rules

### LINKED ISSUE #1737: sql: make database, table, view and columns lookups properly case-sensitive
PR: https://github.com/cockroachdb/cockroach/pull/16884

jseldess:
Fixes #1737 

Hint that you can use double-quotes to preserve case-sensitivity
in database, table, view, and column names.

@knz, is this sufficient, or should we also add notes to `CREATE DATABASE`, `CREATE TABLE`, etc.?


knz:
this seems sufficient to me.
LGTM

PR #1982, opened 2017-10-10
Fix the SQL function docs.

### LINKED ISSUE #1889: sql: add current_user, session_user, user builtins
PR: https://github.com/cockroachdb/cockroach/pull/17903

knz:
Fixes #1889.


knz:
Ok I re-ran the update against `master` and it produced the same diff as in this PR, so I figured I would just reuse it.

@jseldess I hand this over to you, as you see there are new data types that have appeared. If you can't find the bandwidth ping me again and I can probably stub them for you.

knz:
Also note regarding the `user` functions that only the `current_user` function actually exists as a function; the words `user` and `session_user` in the grammar are automatically replaced by `current_user`.

knz:
ok, fixed

jseldess:
Wohoo, thank you, @knz. Merging.

PR #1987, opened 2017-10-10
Fixing nav for architecture-related content

sploiselle:
General nav cleanup given where the architecture docs landed.

- Until we decide to backport the v1.1 architecture docs to v1.0, leaving stub in place with link to 1.1. Hardcoding 1.1 link because it will always be the closest approximation.
- Adding redirects to `architecture/overview.html`––first attempt is experimental. @benesch, would love some help here to ensure I get this right.
- Removing vestigial `cockroachdb-architecture.html` page

### DIFF
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -1239,12 +1239,6 @@
         ]
       },
       {
-        "title": "CockroachDB Architecture",
-        "urls": [
-          "/${VERSION}/cockroachdb-architecture.html"
-        ]
-      },
-      {
         "title": "CockroachDB Features",
         "items": [
           {
--- a/_plugins/versions.rb
+++ b/_plugins/versions.rb
@@ -10,11 +10,13 @@
 # additional aliased versions. (At the time of writing, we named a `stable` and
 # `edge` version.) These are implemented as directory symlinks. For example, if
 # `stable: v1.0` is specified, a symlink `stable -> v1.0` will be written to the
-# build directory. If a page is renamed, set `key: original-file-name.md` in its
-# front matter so the version switcher can find its old versions. The `stable`
-# mapping is special and must exist, as `docs/FOO.html` will be automatically
-# redirected to `docs/stable/FOO.html` for every FOO.md in the stable directory.
-# Other name mappings are optional and can be added and removed at will.
+# build directory. The `stable` mapping is special and must exist, as
+# `docs/FOO.html` will be automatically redirected to `docs/stable/FOO.html` for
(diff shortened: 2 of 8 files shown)


benesch:
@sploiselle I took the liberty of making the tweaks directly to the PR—hope you don't mind! I amended your first commit to use `key: cockroachdb-architecture.html` instead of `redirect-from: cockroachdb-architecture`. Specifying `key:` allows the version switcher to link directly to the new/old name, avoiding a client-side redirect. I also adjusted the index redirect to ask for a redirect of `index.html` instead of `architecture/index.html`, as redirect specifications are relative to the directory of the file. (Your original redirect specification was creating a redirect at `v1.1/architecture/architecture/index.html`.)

Also took the opportunity to fix some bugs in the versioning code, which wasn't designed to handle pages in versioned subdirectories, like the `v1.1/architecture` directory. Previously, we'd install redirects like `/overview.html -> /stable/architecture/overview.html` instead of `/architecture/overview.html -> /stable/architecture/overview.html`.

sploiselle:
Thanks as always, @benesch. Two questions:

- The `key` path is from the "root?" In this case, `/docs/version/`
- `redirect_from` is relative to the file's path? in this case, `/docs/version/architecture`

benesch:
Yes on both counts!

To be pedantic, the key isn't so much a path but a stable identifier for a
page across versions. If you set `key: sean` in a page in both v1.0 and
v1.1, that would also tie those two pages together in the version switcher.
Here I'm relying on the fact that if you leave key unset, it defaults to
the path to the page minus the /docs/version prefix. This is safer than
making up keys on the fly, because different pages must have different
keys, and path-to-page is guaranteed to be unique within a version.

If that doesn't make any sense, just ignore me and carry on.

sploiselle:
@benesch So with `key: cockroachdb-architecture.html` in place, `.../v1.0/cockroachdb-architecture.html` will direct to `.../v1.1/architecture/overview.html`?

If I'm still missing the point, would love an example.

benesch:
With `key: cockroachdb-architecture.html` in place, `.../v1.0/cockroachdb-architecture.html` looks just like before, except that the version switcher on that page knows that v1.1 should link to `.../v1.1/architecture/overview.html`. Similarly, the version switcher on `.../v1.1/architecture/overview.html` will link v1.0 to `.../v1.0/cockroachdb-architecture.html`.

sploiselle:
@benesch 💡 –– just the version switcher. Got it.

benesch:
Exactly!

jseldess:
LGTM

PR #1990, opened 2017-10-11
Document v1.1 known limitations

### LINKED ISSUE #1835: Document 1.1 known limitations
**Note: we should link the issues to this one (in additional comments or using GH linking) instead of editing the PR description directly, so as to prevent edit conflicts.** Also add the label "known-limitation" to the linked issue.

- Because range splits are no longer allowed in the meta2 range, clusters that grow beyond 64TB will have problems. Note from @petermattis: 64TB sounds right with the default range size. It is really a limit on the number of ranges. O(1m) is reasonable with a single meta2 range.

jseldess:
Fixes #1835

### DIFF
--- a/v1.0/known-limitations.md
+++ b/v1.0/known-limitations.md
@@ -1,6 +1,6 @@
 ---
 title: Known Limitations in CockroachDB v1.0
-summary:
+summary: Known limitations is CockroachDB v1.0.
 toc: false
 ---
 
@@ -10,7 +10,7 @@ This page describes limitations we identified in the [CockroachDB v1.0](../relea
 
 ## Removing all rows from large tables
 
-{{site.data.alerts.callout_info}}Resolved as of version 1.1. See <a href="https://github.com/cockroachdb/cockroach/pull/17016">#17016</a>.{{site.data.alerts.end}}
+{{site.data.alerts.callout_info}}Resolved as of v1.1. See <a href="https://github.com/cockroachdb/cockroach/pull/17016">#17016</a>.{{site.data.alerts.end}}
 
 When removing all rows from a table via a [`TRUNCATE`](truncate.html) statement or a [`DELETE`](delete.html#delete-all-rows) statement without a `WHERE` clause, CockroachDB batches the entire operation as a single [transaction](transactions.html). For large tables, this can cause the nodes containing the table data to either crash or exhibit poor performance due to elevated memory and CPU usage.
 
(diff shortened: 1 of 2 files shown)


v1.0/known-limitations.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Known Limitations in CockroachDB v1.0
-summary:
+summary: Known limitations is CockroachDV v1.0.
```
bdarnell: "DB", not "DV"

v1.0/known-limitations.md
```diff
@@ -28,6 +28,36 @@ Within a single [transaction](transactions.html):
 - A [`CREATE TABLE`](create-table.html) statement containing [`FOREIGN KEY`](foreign-key.html) or [`INTERLEAVE`](interleave-in-parent.html) clauses cannot be followed by statements that reference the new table.
 - A table cannot be dropped and then recreated with the same name. This is not possible within a single transaction because `DROP TABLE` does not immediately drop the name of the table. As a workaround, split the [`DROP TABLE`](drop-table.html) and [`CREATE TABLE`](create-table.html) statements into separate transactions.
 
+## Schema changes between executions of prepared statements
+
+When the schema of a table targeted by a prepared statement changes before the prepared statement is executed, CockroachDB should return an error. Instead, CockroachDB incorrectly allows the prepared statement to return results based on the changed table schema, for example:
```
bdarnell: Should this really be an error? The trace below looks reasonable to me. We should add a best-practice warning against using `SELECT *` for queries that will be repeated (whether via prepared statements or not), but I don't think we need to document this as a limitation. (We might want to say that this is *deprecated* if we plan to change this behavior in the future as we do more caching of query plans or output schemas). 

The insert/update cases below are appropriate to document as known limitations.

v1.1/known-limitations.md
```diff
@@ -1,46 +1,94 @@
...
+
+- If the number of columns has increased, the prepared statement returns an error but nonetheless writes the data.
+- If the number of columns remains the same but the types have changed, the prepared statement writes the data and does not return an error.
+
+### Join flags when restoring a backup onto new machines
+
+All nodes in a cluster must have access to a copy of the first key-value range, which is part of a meta-index identifying where all range replicas are stored, and which nodes require to initialize themselves and start accepting incoming connections.
 
 Ensuring that all nodes have access to a copy of the first key-value range is more difficult when restoring from a whole-cluster backup onto machines with different IP addresses than the original cluster. In this case, the `--join` flags must form a fully-connected directed graph. The easiest way to do this is to put all of the new nodes' addresses into each node's `--join` flag, which ensures all nodes can join a node with a copy of the first key-value range.
```
bdarnell: I'd recommend putting the first 3-5 nodes as join targets, not all the nodes. 

This is recommended in all cases, not just when restoring from a backup (we're even talking about making it mandatory in the future).

v1.1/known-limitations.md
```diff
@@ -1,46 +1,94 @@
...
 
-## `INSERT ON CONFLICT` vs. `UPSERT`
+### `INSERT ON CONFLICT` vs. `UPSERT`
 
 When inserting/updating all columns of a table, and the table has no secondary indexes, we recommend using an [`UPSERT`](upsert.html) statement instead of the equivalent [`INSERT ON CONFLICT`](insert.html) statement. Whereas `INSERT ON CONFLICT` always performs a read to determine the necessary writes, the `UPSERT` statement writes without reading, making it faster.
 
 This issue is particularly relevant when using a simple SQL table of two columns to [simulate direct KV access](frequently-asked-questions.html#can-i-use-cockroachdb-as-a-key-value-store). In this case, be sure to use the `UPSERT` statement.
 
-## Repeated or combined commands in the SQL shell history
+### Repeated or combined commands in the SQL shell history
```
bdarnell: This has gotten better in 1.1, although it won't be completely fixed until 1.2.

v1.1/known-limitations.md
```diff
@@ -63,7 +111,7 @@ To detect if this is happening, open the [Admin UI](admin-ui-access-and-navigate
 
 For example, let's say that latency is 10ms from nodes in datacenter A to nodes in datacenter B but is 100ms from nodes in datacenter A to nodes in datacenter C. To ensure A's and B's relative proximity is factored into lease holder rebalancing, you could restart the nodes in datacenter A and B with a common region, `--locality=region=foo,datacenter=a` and `--locality=region=foo,datacenter=b`, while restarting nodes in datacenter C with a different region, `--locality=region=bar,datacenter=c`.
 
-## Roundtrip to `STRING` does not respect precedences of `:::` and `-`
+### Roundtrip to `STRING` does not respect precedences of `:::` and `-`
```
bdarnell: I think this was fixed: cockroachdb/cockroach#15617

v1.1/known-limitations.md
jseldess: Since this case is now not different than any other, I'd say this isn't known limitation anymore. I'm removing, but I can put it back post-release, if you can help me understand how to phrase it so it's a limitation.

v1.0/known-limitations.md
jseldess: Thanks, @bdarnell. I've left this in, for now, and reworded it to simply explain that this can happen and propose that recommendation. Post-release, I can work on moving it somewhere else.

v1.0/known-limitations.md
jseldess: Done.

v1.1/known-limitations.md
jseldess: I guess I'll hold off on saying anything until it's fixed.

v1.1/known-limitations.md
jseldess: Great. Updated in the 1.0 limitations and removed from the 1.1 limitations.

PR #1991, opened 2017-10-11
[WIP] Add secure csharp code samples

BramGruneir:


jseldess:
@BramGruneir, insecure code samples and tutorials landed with https://github.com/cockroachdb/docs/pull/2236. Do you want to refocus this PR on secure examples?

leomkkwan:
Thanks a lot to make C# as a support language.

For my use case, I am running a secure cluster with username/password for client authentication. So in my connection string, I’ve aded “SSL Mode=Require;Trust Server Certificate=True;Use SSL Stream=True”.

BramGruneir:
@leomkkwan, you might need to convert the certs to something that dotnet can read.  I used openssl to do this. The details should be in the files, but let me know if you have any issues.  

What platform are you using?
This works on linux and windows, but won't work on osx due to a limitation involving password-less certs.  And I'm actively looking for a workaround.

leomkkwan:
@BramGruneir In my use case, I did not generate a ssl cert for each client, instead, the client is authenticated by username/password only. However, the cluster is in secure mode, and each node have it own cert. So, when I am trying to connect from my C# apps, I need to enforce ssl is required and trusted server cert. In this case, I believe the traffic is over SSL and I don't need to have cleint cert.

BramGruneir:
Great.  Let me know if you have any issues.

jseldess:
https://github.com/cockroachdb/docs/pull/4213 landed, so this can be closed.

PR #1994, opened 2017-10-12
Un-break links to the dev docs

a-robinson:
This can be switched back to pointing to v1.2 once we have a v1.2
release to point at.

Temporary band-aid on top of f3eb4a30ca59e0be2b56495627300405175df79b to fix the main repo's build, which expects the dev docs to be functional. Other URLs out there on the web may also refer to the dev docs. I'm not sure whether this will work, but teamcity should make it apparent soon.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,6 +19,7 @@ redcarpet:
 
 versions:
   stable: v1.1
+  dev: v1.1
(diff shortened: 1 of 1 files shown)


a-robinson:
It appears to work on the temporary aws site 🎉

PR #1996, opened 2017-10-12
Clarify DROP DATABASE CASCADE/RESTRICT

jseldess:
Clarify that `CASCADE` is the default `DROP DATABASE` behavior for all clients except the built-in SQL client. And add note that, in an upcoming patch release, `CASCADE` will be made the default behavior for the SQL client as well.


knz:
I fear that's not quite correct.
In 1.1.0 erroneously RESTRICT is the default for all clients. It was CASCADE in 1.0.
This is a breaking change for some clients, and the change will be  reverted in 1.1.1 back to the 1.0 behavior.

jseldess:
Ugh, thanks for the explanation, @knz. How does it look now?

PR #1998, opened 2017-10-13
Fix link to Replication Layer page

jseldess:
It was incorrectly pointing to the Transaction Layer page.


sploiselle:
Thanks, @jseldess! LGTM

PR #2002, opened 2017-10-16
RFC: Training Content

sploiselle:
RFC for training content

### DIFF
--- a/rfc/training-content-delivery.md
+++ b/rfc/training-content-delivery.md
@@ -1,0 +1,89 @@
+- Feature Name: Training Delivery
+- Status: draft
+- Start Date: 2017-08-12
+- Authors: sploiselle
+- RFC PR: cockcroachdb/docs/#2002
+- Cockroach Doc Issue: n/a
+
+# Summary
+
+I propose Cockroach Labs delivers training content through a Jekyll-generated HTML site, containing links to other hypertext deliverables (e.g. Google Sheets, embedded YouTube videos).
+
+This approach leverages our existing toolchains (so it's familiar to create and iterate on content), as well as ensures that the content leverages our current branding, so it looks professional and cohesive. A secondary benefit is that it familiarizes users with Cockroach Labs' design language, making it easier to navigate through our other properties, as well.
+
+Because of this approach's integration with our existing workflow, we can easily achieve it by placing this content in a subdirectory off of `cockroachlabs.com`, using a similar approach to both `/docs` and `/blog`.
+
+# Motivation
+
+We've begun developing training curriculum but have not yet determined where it lives, or how to best deliver it to attendees. Because this approach could drastically alter the authoring process, it's best to determine how this is done before we make substantial headway into producing these deliverables.
+
+By the end of this RFC, there will be a plan for the method we plan to use for delivering training content.
+
+# Guide-level explanation
+
+### Elements
+
+CockroachDB training is comprised of 3 elements:
+
+- **Presentations** give the instructor a tool to make sure the class follows the outlined curriculum, as well as reinforce topics discussed for students with a quick summary to help them mentally model the content, as well as a canvas for visual aids.
+- **Labs** provide attendees a chance to work with CockroachDB, using the information they just learned in the presentation. This should include a mixture of reinforcement of content from the presentation and step-by-step instructions to accomplish a task directly related to the presentation.
+- **Demos** give the instructor the option to demonstrate something that might be difficult for attendees to quickly reproduce themselves, e.g. anomalous reads from something like Jepsen. These can be "staged" labs to help students understand that something is true or real without having to relate the conditions for it.
+
+### Delivery
+
+To deliver the material, each element has its own needs:
+
+- Presentations require there being only a single bit of text presented at a time. To achieve this, I recommend using Google Slides.
+- Labs require much more text, as well as other hypertext tools that we've already developed on `/docs`. To achieve this, I recommend using Markdown rendered into HTML through Jekyll.
+- Demos each have different needs, but are often broadcast in the classroom from the instructor's computer. I recommend either recording demos that can be shown on YouTube, or solving demos on a case-by-case basis.
+
+### Structure
+
+As a wrapper around all of this content, the training material is hosted on `cockroachlabs.com/training`. We currently do not have password-restricted access on this content, but may in the future.
+
+This site is written in Markdown and converted to HTML through Jekyll.
+
+#### Organization
+
+The training site is organized in this way
+
+`home/<day>/<module>/`
+
+That is:
+- 1 home page for all training material (if this training system developers further, it's possible to insert levels above this)
+- 1 page for each day, which contain links to its modules
+- 1 page page for each module, which contains contains a brief summary, as well as links to both the presentations and the labs.
+
+Off of the `../<module>/` subdirectory, there would also be a page for any labs from that day. We would not have a page for presentations because these could just be links to Google sheets.
+
+### UX
+
+Users go to the training site and the instructor guides them through each of the presentations. After the presentation, they return to the module page and follow through any labs related to the module.
+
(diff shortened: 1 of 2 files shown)


rfc/training-content-delivery.md
```diff
@@ -0,0 +1,75 @@
...
+
+To deliver the material, each element has its own needs:
+
+- Presentations require there being only a single bit of text presented at a time. To achieve this, I recommend using Google Slides.
+- Labs require much more text, as well as other hypertext tools that we've already developed on `/docs`. To achieve this, I recommend using Markdown rendered into HTML through Jekyll.
+- Demos each have different needs, but are often broadcast in the classroom from the instructor's computer. I recommend either recording demos that can be shown on YouTube, or solving demos on a case-by-case basis.
+
+### Structure
+
+As a wrapper around all of this content, the training material is hosted on `cockroachlabs.com/training`. We currently do not have password-restricted access on this content, but may in the future.
```
jseldess: As discussed offline, it's fine for this content to be public; when used in classrooms, the students get added value from guided exercises and the instructors' expertise.

rfc/training-content-delivery.md
```diff
@@ -0,0 +1,75 @@
...
+
+This site is written in Markdown and converted to HTML through Jekyll and has the following structure:
+
+`home/<day>/<module>/`
+
+That is:
+- 1 home page for all training material (if this training system developers further, it's possible to insert levels above this)
+- 1 page for each day, which contain links to its modules
```
jseldess: For now, sounds fine to group by day. But if we decide, down the road, to expose this content for self-service use, we may then want to reassess the "day"-based approach.

jseldess: An alternative would be grouping by categories like:
- Introduction
- Ops Basics
- Dev Basics
- Monitoring/Troubleshooting
- Maintenance

rfc/training-content-delivery.md
```diff
@@ -0,0 +1,75 @@
...
+
+### UX
+
+Users go to the training site and the instructor guides them through each of the presentations. After the presentation, they return to the module page and follow through any labs related to the module.
+
+At the end of each module page and each lab, there are links to the subsequent modules.
+
+## Drawbacks
+
+This implementation _does not_ provide password protection, and therefor relies on obscurity to "protect" the content's IP. Because the training content is potentially monetizable, this could jeopardize that strategy.
```
jseldess: As mentioned above, I think this is fine. That's the impression I got from discussions with @spencerkimball, too.

rfc/training-content-delivery.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Rationale and Alternatives
+
+This design creates a straighforward user experience that's consistent with our docs site. It leverages our existing toolchain and requires little-to-no ramp up effort.
+
+However, to overcome the drawbacks (i.e., as a means of providing password authentication), training content could be delivered solely through Google Drive, swapping out HTML-rendered Markdown for Google Docs, and using Google Sheets as navigation elements. However, I think that the UX of this is very poor and places a burden onto the instructor and the attendees because of our hopes of monetizing the content.
+
+## Unresolved questions
+
+- Which repo should this content belong to?
```
jseldess: This is a good question. Since Docs will be managing this content for now, operationally it makes sense to have the content in the docs repo. But I agree that the content should be parallel, in terms of url structure. I'm not sure if that means it has to be in the hugo repo?

rfc/training-content-delivery.md
```diff
@@ -0,0 +1,75 @@
...
+- Should this content be versioned?
```
jseldess: It wouldn't hurt to version this content, given that some guidance may change from version to version. That may be one reason to put it in the docs repo, where we already have versioning logic in place.

dianasaur323:
*[rfc/training-content-delivery.md, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-Kwl9z8AeojzDySO_Rq1:-Kwl9z8AeojzDySO_Rq2:b704qt3) ([raw file](https://github.com/cockroachdb/docs/blob/ede5c5cfa07012b0449cdf231051d60f64271f19/rfc/training-content-delivery.md#L26)):*
> ```Markdown
> ### Elements
> 
> CockroachDB training is comprised of 3 elements:
> ```

I understand that we want to make everything as open as possible, however, is there some content that we would prefer to keep private?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2002)*

dianasaur323:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[rfc/training-content-delivery.md, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-Kwl9z8AeojzDySO_Rq1:-KwlAUIHGZyPkVDeIbGq:bf7occx) ([raw file](https://github.com/cockroachdb/docs/blob/ede5c5cfa07012b0449cdf231051d60f64271f19/rfc/training-content-delivery.md#L26)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I understand that we want to make everything as open as possible, however, is there some content that we would prefer to keep private?
</blockquote></details>

Nvm, you mention it later on.

---

*[rfc/training-content-delivery.md, line 50 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-Kw_rW57Gif3COqSEwA--r1-50:-KwlAl3_p0QGuvhzDJb2:bizw804) ([raw file](https://github.com/cockroachdb/docs/blob/ede5c5cfa07012b0449cdf231051d60f64271f19/rfc/training-content-delivery.md#L50)):*

An alternative would be grouping by categories like:
- Introduction
- Ops Basics
- Dev Basics
- Monitoring/Troubleshooting
- Maintenance
</blockquote></details>

Yep, I agree that we might want to group by categories eventually when we have more comment

---

*[rfc/training-content-delivery.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-Kw_rW57Gif3COqSEwA--r1-73:-KwlBEErjxOe2ItNUu0W:b-n36t5d) ([raw file](https://github.com/cockroachdb/docs/blob/ede5c5cfa07012b0449cdf231051d60f64271f19/rfc/training-content-delivery.md#L73)):*

This is a good question. Since Docs will be managing this content for now, operationally it makes sense to have the content in the docs repo. But I agree that the content should be parallel, in terms of url structure. I'm not sure if that means it has to be in the hugo repo?
</blockquote></details>

I think we might want to have a different cockroachlabs.com/___ though, since this isn't really docs anymore

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2002)*

dianasaur323:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-:-KwlBJiIZOZmCSwQMgS6:bnfp4nl)*

jseldess:
*[rfc/training-content-delivery.md, line 73 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2002#-Kw_rW57Gif3COqSEwA--r1-73:-KwlFSmZ1LJbpYWMohhO:bbohpyl) ([raw file](https://github.com/cockroachdb/docs/blob/ede5c5cfa07012b0449cdf231051d60f64271f19/rfc/training-content-delivery.md#L73)):*
<details><summary><i>Previously, dianasaur323 (Diana Hsieh) wrote…</i></summary><blockquote>

I think we might want to have a different cockroachlabs.com/___ though, since this isn't really docs anymore
</blockquote></details>

Yep, we're all agreeing that this content should reflect a parallel url structure to `/docs`, `/blog`, etc. Just not sure which repo the files should live in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2002)*

PR #2048, opened 2017-10-18
./cockroach no longer needed with docker

amq:
Only insecure docker swarm doc is affected.


jseldess:
Thanks, @amq! @a-robinson, can you take a quick look at this change?

PR #2072, opened 2017-10-18
Release notes for v1.1.1

bdarnell:
The `--advertise-port` flag is (deliberately) undocumented, so maybe we should drop it from the release notes. I also don't know if we've documented the range debug page enough that it needs a mention here.


jseldess:
Thanks, @bdarnell. LGTM overall. I agree that we should remove `--advertise-port` as well as the debug page reference, since we haven't documented those anywhere. We'll start that work for 1.2.

I can make these changes.

releases/v1.1.1.md
```diff
@@ -0,0 +1,51 @@
...
+- Some issues that could cause non-restorable backups are now detected during the backup process. [#19286](https://github.com/cockroachdb/cockroach/pull/19286)
+- Fixed a deadlock in the crash reporting code, making more information available about certain fatal errors. [#19287](https://github.com/cockroachdb/cockroach/pull/19287)
+- The `--advertise-port` flag now works correctly. [#18974](https://github.com/cockroachdb/cockroach/pull/18974)
+- Very large arguments to the SQL `repeat()` function now return an error instead of panicking. [#19066](https://github.com/cockroachdb/cockroach/pull/19066)
+- It is now possible to drop a foreign key constraint on a column that is part of the primary key. [#19092](https://github.com/cockroachdb/cockroach/pull/19092)
+- Attempting to use `SHOW TRACE FOR` while session tracing is enabled no longer turns off session tracing. [#19099](https://github.com/cockroachdb/cockroach/pull/19099)
+- `SHOW TRACE FOR` no longer includes detailed KV-level information (these messages are included in the separate `SHOW KV TRACE FOR` command). [#19143](https://github.com/cockroachdb/cockroach/pull/19143)
```
knz: No mention of #19209?

v1.1/known-limitations.md
```diff
@@ -12,6 +12,8 @@ This page describes newly identified limitations in the CockroachDB v1.1 release
 
 ### Enterprise restores after creating, dropping, or truncating tables
 
+{{site.data.alerts.callout_info}}Resolved as of v1.1.1. It is no longer possible to create an incremental backup if a table has been been created, dropped, or truncated after the full backup. See <a href="https://github.com/cockroachdb/cockroach/pull/19286">#19286</a>.{{site.data.alerts.end}}
```
bdarnell: I would say "changed" instead of "resolved". You still can't make a backup after dropping a table. The only change is that now it will tell you the backup has failed instead of leaving you with a broken backup and a false sense of security.

jseldess: Rephrased.

PR #2093, opened 2017-10-24
add redirect for renamed admin UI overview

benesch:

PR #2097, opened 2017-10-24
Redirected Explore the Admin UI page to Overview page

### LINKED ISSUE #2096: 404 on explore-the-admin-ui.html
Going to `https://www.cockroachlabs.com/docs/stable/explore-the-admin-ui.html` 404s while `https://www.cockroachlabs.com/docs/v1.0/explore-the-admin-ui.html` succeeds.

@Amruta-Ranade To resolve this, can you add a `redirect_from: explore-the-admin-ui.html` to whatever the "key" page is for the Admin UI docs?

Amruta-Ranade:
Closes #2096

### DIFF
--- a/v1.1/admin-ui-overview.md
+++ b/v1.1/admin-ui-overview.md
@@ -2,8 +2,8 @@
 title: Admin UI Overview
 summary: Use the Admin UI to monitor and optimize cluster performance.
 toc: false
-redirect_from:
-- explore-the-admin-ui.html
+redirect_from: explore-the-admin-ui.html
+key: explore-the-admin-ui.html
 ---
 
 The CockroachDB Admin UI provides details about your cluster and database configuration, and helps you optimize cluster performance by monitoring the following areas:
--- a/v1.1/admin-ui-overview.md
+++ b/v1.1/admin-ui-overview.md
@@ -2,6 +2,8 @@
 title: Admin UI Overview
 summary: Use the Admin UI to monitor and optimize cluster performance.
 toc: false
+redirect_from:
+- explore-the-admin-ui.html
 ---
 
 The CockroachDB Admin UI provides details about your cluster and database configuration, and helps you optimize cluster performance by monitoring the following areas:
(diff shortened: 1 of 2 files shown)


jseldess:
@Amruta-Ranade, it seems @benesch opened a PR to fix the same problem but implemented some additional front-matter fields. Can you chat with him about https://github.com/cockroachdb/docs/pull/2093 and decide how to proceed (with your or his PR)?

benesch:
The reason I prefer `redirect_to` is because `redirect_from` has some weird interaction with the `/stable` alias. The former will redirect /explore-the-admin-ui.html to /stable (which I think is correct) and the latter will redirect /explore-teh-admin-ui.html to v1.1 (which is odd).

http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/d315de0f5c6335f59ab20f45777f4eebc8a7d6da/explore-the-admin-ui.html
http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/980b30c2a234d2647f14c0981672943fc623681d/explore-the-admin-ui.html

Specifying `key:` means the version switcher can link directly to the renamed page instead of causing a client-side redirect

benesch:
Sorry we didn't get a chance to talk in person this week, @Amruta-Ranade!

jseldess:
Thanks for the added details, @benesch. It actually doesn't bother me much that `redirect_from` takes you to `v1.1` rather than stable, and I like that we don't need to keep the `explore-the-admin-ui.md` file in v1.1 just to put `redirect_to` in place. So I'll just go with this PR, add the `key` field, and see if anyone complains. Hope that's ok.

jseldess:
Chatted more with @benesch. Since we use `redirect_from` already in several places, we'll go with that for now. But `redirect_to` is safer in that someone won't in the future create a file that conflicts with the redirect, whereas with `redirect_from`, the logic is kind of hidden and that someone could, in principle, create such a conflict. We can revisit this problem down the road.

Issue #1067, opened 2017-02-03
SQL pagination


jseldess:
From @bdarnell, overheard in a thread with a user:

limit/offset is a really inefficient way of doing pagination. unless you really need to be able to jump to arbitrary pages, it's better to have each key include the index of the last entry returned (or the first entry not returned) so you can pick up the query from there on the next request

if you do need to use limit/offset, then it's often better to split it into two queries: one query that uses limit/offset and just reads the task_id (so it can use the index alone without joining with the primary table), and a second query that uses `where task_id in (...)` to get the data (assuming task_id is the primary key)

Also:

limit/offset pagination gets a lot more efficient if your index includes enough columns to satisfy the group by/order by clauses without joining against the primary data

And then from @RaduBerinde: 

yes, if the index is sorted like the order by clause, we won't need to sort, which means we will just read (limit + offset) rows. This applies even if we are doing an index-join because other columns are needed by the select.

sploiselle:
Potential reference point: https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/

jseldess:
Comment from @danhhz, from https://github.com/cockroachdb/docs/issues/657:

The usual way to do this is one SELECT query per page that uses LIMIT/OFFSET, but since it doesn't make sense to run this in a sql transaction, any insertions/deletions could make this an inconsistent view of the table.

CockroachDB has AS OF SYSTEM TIME implemented, which means we can let the user iterate over a consistent version of their database without a transaction. This would work by selecting `cluster_logical_timestamp()` in the first SELECT, passing it through the page (like limit and offset are) and using it in subsequent queries. @mjibson has already implemented paging exactly like this for the `./cockroach dump` command.

See the discussion in https://github.com/cockroachdb/cockroach/issues/9227#issuecomment-246157219

wzrdtales:
@jseldess That might sound like a good idea, but apart from dumping scenarios it mostly is not. `AS OF SYSTEM TIME` will also deliver outdated results, so for a pagintion you really only want to avoid strange pagination behaviors, but you do not want to deliver outdated results.

wzrdtales:
The most practical thing would be to allow instead of an offset telling it to show everything after a specified id within the current sorting.

wzrdtales:
Since there is nothing but time as a serial factor here, the tables affected should not allow inserting anything with the same timestamp or anything int that window at a later time.

wzrdtales:
[...]

PR #2099, opened 2017-10-25
Added content for FETCH FIRST and FETCH NEXT

### LINKED ISSUE #1497: sql: support FETCH FIRST syntax in SELECTs
The standard SQL syntax FETCH FIRST N ROWS ONLY is now supported as an alias for LIMIT.

PR: https://github.com/cockroachdb/cockroach/pull/16083

Amruta-Ranade:
Closes #1497 

As discussed, added information about FETCH FIRST and FETCH NEXT in same row as LIMIT, so as not to call attention, yet inform that we do support the parameters now. Also added examples.

### DIFF
--- a/v1.1/select.md
+++ b/v1.1/select.md
@@ -36,7 +36,7 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
 | `EXCEPT` | Only retrieve rows that are in the preceding `SELECT` statement but not in the following `SELECT` statement.  Returns distinct values.|
 | `ALL` | Include duplicate rows in the returned values of `UNION`, `INTERSECT`, or `EXCEPT`. |
 | `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column.<br><br>When <code>ORDER BY</code> is not included in a query, rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.|
-| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. <br><br><span class="version-tag">New in v1.1:</span> CockroachDB also supports `FETCH FIRST limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases for `LIMIT`. If `limit_val` is omitted, then one row is fetched. `FETCH NEXT` is often used in conjunction with `OFFSET`.|
+| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. <br><br><span class="version-tag">New in v1.1:</span> CockroachDB also supports `FETCH FIRST limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases for `LIMIT`. If `limit_val` is omitted, then one row is fetched.|
 | `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` to "paginate" through retrieved rows.|
 
 ## Examples
--- a/v1.1/select.md
+++ b/v1.1/select.md
@@ -36,8 +36,8 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
 | `EXCEPT` | Only retrieve rows that are in the preceding `SELECT` statement but not in the following `SELECT` statement.  Returns distinct values.|
 | `ALL` | Include duplicate rows in the returned values of `UNION`, `INTERSECT`, or `EXCEPT`. |
 | `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column.<br><br>When <code>ORDER BY</code> is not included in a query, rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.|
-| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. |
-| `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` to "paginate" through retrieved rows. |
+| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. <br><br><span class="version-tag">New in v1.1:</span> CockroachDB also supports `FETCH FIRST limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases for `LIMIT`. If `limit_val` is omitted, then one row is fetched. `FETCH NEXT` is often used in conjunction with `OFFSET`. |
+| `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` or `FETCH NEXT` to "paginate" through retrieved rows.|
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


v1.1/select.md
```diff
@@ -528,6 +528,25 @@ LIMIT 5;
 +----+------------------+
 ~~~
 
+Alternatively, you can use `FETCH FIRST`.
+
+~~~ sql
+> SELECT id, name 
+FROM accounts 
+FETCH FIRST 5 ROWS ONLY;
```
justinj: I think we should put a warning alongside this if we're going to include an example of limit/fetch first that doesn't also include an `ORDER BY`, since the results are not going to be stable across invocations.

v1.1/select.md
```diff
@@ -528,6 +528,25 @@ LIMIT 5;
...
++----+------------------+
+|  1 | Bjorn Fairclough |
+|  2 | Bjorn Fairclough |
+|  3 | Arturo Nevin     |
+|  4 | Arturo Nevin     |
+|  5 | Naseem Joossens  |
++----+------------------+
+~~~
+
 #### Paginate Through Limited Results
```
justinj: I know this is from before this PR, but I'm not sure we should be advocating this, `OFFSET` is rarely what you want - it's not particularly efficient and it's sort of error prone.

sploiselle: Would be glad to improve this––what do you recommend instead?

justinj: It depends on the use case, really, but the inefficiency is that the database has to process all the offset-ed rows as well - `SELECT ... LIMIT 1 OFFSET 1000000000` will have to process `1000000001` rows. One better way is to store the sort key of the last row processed and then `SELECT ... WHERE key > key ORDER BY key ... LIMIT n`

I think there are some other things that can go wrong too, but they're slipping my mind at the moment...

Amruta-Ranade:
@justinj The rationale behind documenting this functionality is that if we are providing support for it, we need to let users know about it, but not necessarily call attention to it. Also, FETCH FIRST is part of the SQL Standard, while LIMIT is not. However, LIMIT is the more popular usage. Considering these factors, I vote that we document it. 

@jseldess Thoughts?

v1.1/select.md
tschottdorf: https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/

sploiselle: Derp. This is already tracked in #1067. Sure we'll get to it eventually.

Amruta-Ranade:
@justinj So does the following query replace the OFFSET one (as described in the article):

> SELECT id, name 
+FROM accounts 
+WHERE N>5
+ORDER BY n ASC
+FETCH NEXT 5 ROWS ONLY;

Amruta-Ranade:
@justinj Is this a better way to paginate instead of OFFSET:
> SELECT id, "name" FROM accounts WHERE id > 5 ORDER BY id ASC LIMIT 5;

justinj:
Yeah, where the `5` is actually whatever the last observed row was before. This way the database can make use of its indexes. Makes sense I guess to document it, so someone coming across a query using it will understand!

v1.1/select.md
jseldess: I think it's sufficient to call out this compatibility-based syntax in the parameter list above. Let's remove this alternative example.

v1.1/select.md
jseldess: Ugh, that's an old issue. But I'm glad it's tracked, and we will eventually get to it.

v1.1/select.md
```diff
@@ -550,6 +569,26 @@ OFFSET 5;
 +----+------------------+
 ~~~
 
+Alternatively, you can use `FETCH NEXT`.
```
jseldess: Same as above. Let's remove this alternate example and just go with the callout in the parameters table.

v1.1/select.md
```diff
@@ -36,8 +36,8 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
...
 | `ALL` | Include duplicate rows in the returned values of `UNION`, `INTERSECT`, or `EXCEPT`. |
 | `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column.<br><br>When <code>ORDER BY</code> is not included in a query, rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.|
-| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. |
-| `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` to "paginate" through retrieved rows. |
+| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. <br><br><span class="version-tag">New in v1.1:</span> CockroachDB also supports `FETCH FIRST limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases for `LIMIT`. If `limit_val` is omitted, then one row is fetched. `FETCH NEXT` is often used in conjunction with `OFFSET`. |
```
jseldess: Let's extend the first part to:

> Only retrieve `limit_val` number of rows.  This is often used in conjunction with `OFFSET` to "paginate" through retrieved rows.

And then I think we can remove:

> `FETCH NEXT` is often used in conjunction with `OFFSET`.

v1.1/select.md
```diff
@@ -36,8 +36,8 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
...
+| `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` or `FETCH NEXT` to "paginate" through retrieved rows.|
```
jseldess: Let's just remove:

> or `FETCH NEXT`

v1.1/select.md
Amruta-Ranade: @jseldess And retain description for `OFFSET` or combine that with `LIMIT` as well?

v1.2/select.md
```diff
@@ -36,9 +36,8 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
 | `EXCEPT` | Only retrieve rows that are in the preceding `SELECT` statement but not in the following `SELECT` statement.  Returns distinct values.|
 | `ALL` | Include duplicate rows in the returned values of `UNION`, `INTERSECT`, or `EXCEPT`. |
 | `ORDER BY sortby_list` | Sort retrieved rows in the order of comma-separated column names you include in `sortby_list`. You can optionally specify `ASC` or `DESC` order for each column.<br><br>When <code>ORDER BY</code> is not included in a query, rows are not sorted by any consistent criteria. Instead, CockroachDB returns them as the coordinating node receives them.|
-| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. |
-| `OFFSET offset_val` | Do not include the first `offset_value` number of rows.<br/><br/>`OFFSET` is often used in conjunction with `LIMIT` to "paginate" through retrieved rows. |
-
+| `LIMIT limit_val` | Only retrieve `limit_val` number of rows. <br><br>CockroachDB also supports `FETCH FIRST limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases for `LIMIT`. If `limit_val` is omitted, then one row is fetched. `FETCH NEXT` is often used in conjunction with `OFFSET`.|
```
jseldess: Only one more change: Instead of this at the end of the FETCH NEXT note:

> `FETCH NEXT` is often used in conjunction with `OFFSET`.

Let's expand the first sentence to:

> Only retrieve `limit_val` number of rows. `LIMIT` is often used in conjunction with `OFFSET` to "paginate" through retrieved rows.

PR #2100, opened 2017-10-26
Update replication-layer.md

louishust:
fix typo


sploiselle:
@louishust Thank you for catching and fixing this!

PR #2123, opened 2017-11-01
Release notes for CockroachDB v1.1.2

jseldess:


jseldess:
@tschottdorf, I need help understanding and summarizing the following commits. Can you please: 
- Identify the commits warranting callout in the v1.1.2 release notes.
- For each of those, provide a sentence of explanation and the corresponding PR. Also, I'm assuming they're all bug fixes; if not, please let me know what other category they fit under (performance improvements, cli change, etc.).

```
commit 52f16b32e486274f67941a7328f9b66b12d48472
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 23 21:42:46 2017 +0200

    log: special-case Wrap(safe, "foo")
    
    Extend the previous special case to allow reporting a wrapped
    error verbatim if it is the only argument.
    
    This should allow the changes in #19447 to *actually* report the wrapped
    error including its type (before the string would be reported, but losing
    the wrapped type).

commit de18698919fdcba5c1f94a072888a41e1c35a75b
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 23 17:37:19 2017 +0200

    storage: return error from split/merge lock acquisition
    
    Also report the error to sentry because we suspect it to have caused
    the bug referenced below in versions of CockroachDB not running with
    this commit.
    
    See #19172.

commit b6d4eae8a8a2262e432c0a89c7628b2ea51650b7
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 23 18:00:51 2017 +0200

    log: export SendCrashReport

commit 3f1fdd8b4e123aa18f395b57cef1988645d1f117
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 23 17:54:39 2017 +0200

    storage: report failed raft command application
    
    Previously, if `applyRaftCommand` returned an error, it would mark the replica
    as corrupt but then go on and execute the side effects and potentially the
    assertions in `assertState()`. These were then likely to fail and return a
    misleading error, as likely seen in #16004.
    
    Instead, cause a fatal error right when observing the error, and potentially
    capture the root cause on sentry.io. The (perhaps too optimisic) expectation
    is that after accounting these disk corruption/space errors, there will be
    much fewer (possibly no) reports triggered by `assertState()`.
    
    Touches #16004.

commit 6a8129b44fc0c500ce77f84f1b5b2dc0de981159
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Wed Oct 18 17:37:03 2017 +0200

    storage: order liveness update criteria
    
    Though Expiration should never move backwards in practice, it is a
    theoretical possibility (at least when using clockless reads).
    Use an ordering that prioritizes Epoch to address that.

commit 5d1c746f79d65594b5b2bf96bcb18bef8040ad72
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 16 17:24:18 2017 +0200

    storage: remove `self` from liveness
    
    Storing entries for the own node in `self` but also having the `nodes` map
    was confusing, and the original motivation, if any, was hard to come by.
    It seems more straightforward to treat the own node like any other, and to
    handle out-of-order updates by comparing livenesses. Admittedly this has
    some shortcomings, but especially for the own entry, these appear to be
    acceptable.

commit bed0bbf4d30be4fa14039aab24d1c9eb7ab0ffeb
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 16 16:33:34 2017 +0200

    storage: check liveness in setSelf

commit ed363dc41c6beb4d887cc071f25f0b39660ffb96
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 16 16:21:20 2017 +0200

    storage: less liveness clobbering in IncrementEpoch
    
    See #18219.

commit a97b409b44c7b8ee37b6e62c453c8e71d881339d
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Mon Oct 16 16:09:53 2017 +0200

    storage: avoid some liveness clobbering
    
    Sharpen the conditions under which Gossip updates overwrite the existing
    liveness entry. I believe this clobbering doesn't happen in practice,
    but this commit also prepares the code to be reusable in other locations
    that are known to clobber.
    
    See #18219.

commit ecb89ea380d1f01dfa8f0786f559f046a15a8763
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Thu Oct 5 17:18:24 2017 -0400

    storage: prevent accidentally unpoisoning aborted txn
    
    Discovered by @bdarnell in https://github.com/cockroachdb/cockroach/pull/18635#issuecomment-333393480.
    Making poisoning happen less often (to reduce contention) is planned but
    requires more care.

commit b32d56a7e064f2bb34490b4c3eee7d4234a079f1
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Fri Oct 6 13:09:46 2017 -0400

    storage: declare AbortCache only when used
    
    This ensures that we won't declare the AbortSpan key for a txn
    when committing or moving intents.
```

jseldess:
@bdarnell, I excluded the following test-related commits. Please let me know if any should in fact be included.

Also, it'd be great if you could check my language in the release notes.

```
commit e401b5bbb1d656e86a2059118b4371c3ef3583a4
Author: Radu Berinde <radu@cockroachlabs.com>
Date:   Wed Oct 25 14:06:41 2017 -0400

    sql: fix anonymous alias set up for natural join
    
    Fixes #19536.

commit d9c4fa0898756b93d4725cd0f141289f4f7433bc
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Thu Oct 12 15:28:26 2017 +0200

    storage: de-flake TestNodeLiveness
    
    Heartbeat() is not guaranteed to return success.
    
    Fixes #18340.

commit 72ef8f97e2815a439aacb2807bd408d43f6e53e7
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Fri Oct 20 10:08:44 2017 +0200

    storage: deflake TestNodeHeartbeatCallback
    
    The node could become live by observing the Gossip update (triggered by its
    heartbeat write) before getting to the point in `heartbeatInternal` at which
    it would invoke the heartbeat callback. As a result, one location in this
    test needs to be prepared to wait for a split second.
    
    Flaky in the hundreds before, now not flaky with 2k+ iterations.
    
    Fixes #19362.

commit 20175ae45e6969417582cb39b38e663e52c55f1a
Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
Date:   Wed Oct 18 19:58:10 2017 +0200

    storage: de-flake TestNodeIsLiveCallback
    
    This test exercised that liveness callbacks fired after a node's liveness
    changed from false to true. However, the trigger was tied to Gossip updates
    only, but the liveness also updated "eagerly" after the heartbeat.
    
    Recent refactorings made it much more likely that the eager update took
    precedence, in which case the Gossip handler would avoid updating a
    second time, forgetting the callbacks as a consequence.
    
    This was addressed by making callback invocation the duty of the code
    that ingests livenesses.
    
    The test was extremely flaky before, and passed several thousand iterations as
    of this commit.
    
    Fixes #19344.

commit e8537cffa5501187b9bedc334079b8e2aa67aec0
Author: Matt Jibson <matt.jibson@gmail.com>
Date:   Tue Oct 17 14:17:58 2017 -0400

    cli: test dump with non-default PK name
    
    Although this test is already passing, it is being added to master
    so we don't regress and so it can be backported to 1.1, where it
    is failing.
    
    See #18500

commit 2a7a423cb37307f391e3c842d5a35380402c2bad
Author: Nikhil Benesch <nikhil.benesch@gmail.com>
Date:   Sun Oct 15 18:00:41 2017 -0400

    server: allow more than one lease in TestRange[s]Response
    
    If the server is slow to boot, leases might be renewed and generate
    additional history entries. Loosen the assertion of the number of
    history entries in TestRangeResponse and TestRangesResponse from
    "exactly one history entry" to "at least one history entry."

commit 9e3d69f3f8fe03aa9e2bc2675bfe587283478f58
Author: Nikhil Benesch <nikhil.benesch@gmail.com>
Date:   Sun Oct 15 22:12:46 2017 -0400

    storage: avoid miscounting scan requests in TestStoreScanIntents
    
    Previously, in TestScoreScanIntents, if the race detector slowed things
    down sufficiently, a periodic gossip would be triggered before the test
    completed. This gossip would generate scan requests that were included
    in the test's bookkeeping, causing spurious failures.
    
    Filter out scan requests that aren't generated by the test by using a
    unique prefix for the test's keys. Enabling the DisablePeriodicGossip
    testing knob would have the same effect, but this approach is more
    robust against future features that might generate background scan
    requests.

commit 897270209934b6c70956279dc5d6163237152b74
Author: Alex Robinson <alexdwanerobinson@gmail.com>
Date:   Thu Oct 5 12:51:49 2017 -0400

    build: Increase testrace timeout to allow for logic tests taking longer
    
    Fixes #19058
```

tschottdorf:
Copy-pasted your list, removing everything that didn't seem noteworthy.

#19448: improved handling of errors during range splits.
#19447: prevent correctness problems when nodes run out of disk.
#19279: prevent decommissioning from hanging even though target node is decommissioned.
#19279: prevent aborted transactions from not seeing their own writes.
#19279: improve performance for some transaction conflicts

RaduBerinde:
Description of 19419 LGTM

tschottdorf:
Language LGTM. Definitely better than my wordings above anyway 😄

bdarnell:
@RaduBerinde 's natural join change is a bug fix that should be mentioned. "Fixed natural join when one side of the join is a subquery" or something like that.

jseldess:
Added that natural join change, @bdarnell. Thanks for catching that.

@tschottdorf, added those changes. Please check my language.

tschottdorf:
Did you mean to link to somewhere in `NATURAL JOIN`?

LGTM

jseldess:
TFTR, @tschottdorf. We don't have any specific docs on "natural joins." We need much better join docs overall, soon.

jseldess:
@knz, please check the latest commit, which updates 1.1 and 1.2 `DROP DATABASE` docs.

knz:
Reviewed 1 of 1 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 6 of 6 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[v1.1/set-vars.md, line 41 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgkC1LDw2nuzB0eQV:-KxxgkC2tuRWiOGGBy6K:b-9zo8jd) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.1/set-vars.md#L41)):*
> ```Markdown
> | `database` | The default database for the current session. | Database in connection string, or empty if not specified | Yes |
> | `default_transaction_isolation` | The default transaction isolation level for the current session. See [Transaction parameters](transactions.html#transaction-parameters) and [`SET TRANSACTION`](set-transaction.html) for more details. | Settings in connection string, or "`SERIALIZABLE`" if not specified  | Yes |
> | `sql_safe_updates` | If `false`, allow potentially unsafe SQL statements, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for sessions from the [built-in SQL client](use-the-built-in-sql-client.html),<br>`false` for sessions from other clients | Yes |
> ```

`true` for interactive sessions from the built-in SQL client.

If the session is not interactive, `sql_safe_updates` is not set either.

---

*[v1.1/show-vars.md, line 44 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgrAlrLxPaKU6wk2l:-KxxgrAlrLxPaKU6wk2m:b-7ropln) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.1/show-vars.md#L44)):*
> ```Markdown
> | `server_version` | The version of PostgreSQL that CockroachDB emulates. | Version-dependent | No |
> | `session_user` | The user connected for the current session. | User in connection string | No |
> | `sql_safe_updates` | If `false`, potentially unsafe SQL statements are allowed, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for sessions from the [built-in SQL client](use-the-built-in-sql-client.html),<br>`false` for sessions from other clients | Yes |
> ```

ditto

---

*[v1.2/set-vars.md, line 41 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgvGU67n7uiXRKtrk:-KxxgvGVd8t-O4UPDCEM:b-7ropln) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.2/set-vars.md#L41)):*
> ```Markdown
> | `database` | The default database for the current session. | Database in connection string, or empty if not specified | Yes |
> | `default_transaction_isolation` | The default transaction isolation level for the current session. See [Transaction parameters](transactions.html#transaction-parameters) and [`SET TRANSACTION`](set-transaction.html) for more details. | Settings in connection string, or "`SERIALIZABLE`" if not specified  | Yes |
> | `sql_safe_updates` | If `false`, allow potentially unsafe SQL statements, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for sessions from the [built-in SQL client](use-the-built-in-sql-client.html),<br>`false` for sessions from other clients | Yes |
> ```

ditto

---

*[v1.2/show-vars.md, line 44 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgwgIsSIMA36fPMWv:-KxxgwgIsSIMA36fPMWw:b-7ropln) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.2/show-vars.md#L44)):*
> ```Markdown
> | `server_version` | The version of PostgreSQL that CockroachDB emulates. | Version-dependent | No |
> | `session_user` | The user connected for the current session. | User in connection string | No |
> | `sql_safe_updates` | If `false`, potentially unsafe SQL statements are allowed, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for sessions from the [built-in SQL client](use-the-built-in-sql-client.html),<br>`false` for sessions from other clients | Yes |
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2123)*

jseldess:
Review status: 7 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[v1.1/set-vars.md, line 41 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgkC1LDw2nuzB0eQV:-Kxxk7HuQnpYmJao_Pd8:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.1/set-vars.md#L41)):*

`true` for interactive sessions from the built-in SQL client.

If the session is not interactive, `sql_safe_updates` is not set either.
</blockquote></details>

Done.

---

*[v1.1/show-vars.md, line 44 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgrAlrLxPaKU6wk2l:-Kxxk6heoMmrEb-C0RIu:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.1/show-vars.md#L44)):*

ditto
</blockquote></details>

Done.

---

*[v1.2/set-vars.md, line 41 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgvGU67n7uiXRKtrk:-Kxxk5pMZV7ufXOj3ipU:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.2/set-vars.md#L41)):*

ditto
</blockquote></details>

Done.

---

*[v1.2/show-vars.md, line 44 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2123#-KxxgwgIsSIMA36fPMWv:-Kxxk5CaTTq5EEXbIsMc:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/3daa4c7083ffb20ed96ceac7c2446f7e32283c3a/v1.2/show-vars.md#L44)):*

ditto
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2123)*

PR #2126, opened 2017-11-02
Test App: Add explicit release savepoint

### LINKED ISSUE #2121: java code sample missing release
The "Transaction (with retry logic)" code sample for java seems wrong:
https://www.cockroachlabs.com/docs/stable/build-a-java-app-with-cockroachdb.html#transaction-with-retry-logic

As someone noticed on gitter, it's missing a `releaseSavepoint()` call. Besides that making the sample not handle retriable errors on COMMIT correctly, this user seems to claim that the jdbc driver doesn't even let you run the `db.commit()` method.

@arjunravinarayan, assigning to you since I think you wrote this sample originally.

```
i13026 @i13026 09:13
db.setAutoCommit(false)

val savePoint = db.setSavepoint("COCKROACH_RESTART")

while (true) {
  try {
    // statement 1
// statement 2
// statement 3

    db.releaseSavepoint(savePoint)
    db.commit

    return
  } catch {
    case e: SQLException =>
      if (e.getErrorCode == 40001 || e.getSQLState == "40001") {
        db.rollback(savePoint)
      } else {
        throw e
      }
  }
}
Is this right? In https://www.cockroachlabs.com/docs/stable/build-a-java-app-with-cockroachdb.html#transaction-with-retry-logic the comment says // If we reach this point, commit the transaction, which implicitly releases the savepoint.
conn.commit();
but it without release it does not work

Tobias Schottdorf @tschottdorf 09:15
what error do you get?

i13026 @i13026 11:08
without the releaseSavepoint: org.postgresql.util.PSQLException: ERROR: SAVEPOINT COCKROACH_RESTART

Andrei Matei @andreimatei 11:34
@i13026 our java code sample seems wrong to me. The release statement is indeed needed. I’ll see about fixing it. The idea is that the release call can encounter a retriable error, and at that point we can still restart the txn. If the release is missing, then the commit would encounter that retriable error, except it’s too late to restart after a commit (more precisely, the ROLLBACK TO SAVEPOINT would fail after a COMMIT).
The comment about commit implicitly releasing the savepoint is, in itself, not wrong. If the database sees a COMMIT, then the savepoint is automatically released. I believe this is Postgres behavior which we’re emulating (I have to check that too, though, now that we’re talking).
However, it would appear that jdbc or the jdbc Postgres driver doesn’t let you do that - I interpret your error as the driver insisting that all savepoints are released. Will investigate a bit.
```

arjunravinarayan:
Fixes #2121. While the previous version was correct, some JDBC drivers
do not implicitly release the savepoint on transaction commit as the
comment indicated, so this is a better example.

### DIFF
--- a/_includes/app/TxnSample.java
+++ b/_includes/app/TxnSample.java
@@ -6,6 +6,12 @@ import java.sql.*;
   You can download the postgres JDBC driver jar from https://jdbc.postgresql.org.
 */
 
+// Ambiguous whether the transaction committed or not.
+class AmbiguousCommitException extends SQLException{
+    public AmbiguousCommitException(Throwable cause) {
+        super(cause);
+    }
+}
 class InsufficientBalanceException extends Exception {}
 class AccountNotFoundException extends Exception {
     public int account;
@@ -17,14 +23,14 @@ class AccountNotFoundException extends Exception {
 // A simple interface that provides a retryable lambda expression.
 interface RetryableTransaction {
     public void run(Connection conn)
-        throws SQLException, InsufficientBalanceException, AccountNotFoundException;
+        throws SQLException, InsufficientBalanceException, AccountNotFoundException, AmbiguousCommitException;
 }
 
 public class TxnSample {
     public static RetryableTransaction transferFunds(int from, int to, int amount) {
         return new RetryableTransaction() {
             public void run(Connection conn)
-                throws SQLException, InsufficientBalanceException, AccountNotFoundException {
+                throws SQLException, InsufficientBalanceException, AccountNotFoundException, AmbiguousCommitException {
                 // Check the current balance.
                 ResultSet res = conn.createStatement().executeQuery("SELECT balance FROM accounts WHERE id = " + from);
                 if(!res.next()) {
@@ -42,28 +48,31 @@ public class TxnSample {
     }
 
     public static void retryTransaction(Connection conn, RetryableTransaction tx)
-        throws SQLException, InsufficientBalanceException, AccountNotFoundException {
+        throws SQLException, InsufficientBalanceException, AccountNotFoundException, AmbiguousCommitException {
         Savepoint sp = conn.setSavepoint("cockroach_restart");
         while(true) {
+            boolean releaseAttempted = false;
             try {
-                // Attempt the transaction.
                 tx.run(conn);
-
-                // If we reach this point, commit the transaction,
-                // which implicitly releases the savepoint.
-                conn.commit();
-                break;
-            } catch(SQLException e) {
+                releaseAttempted = true;
+                conn.releaseSavepoint(sp);
+            }
+            catch(SQLException e) {
+                String sqlState = e.getSQLState();
                 // Check if the error code indicates a SERIALIZATION_FAILURE.
-                if(e.getErrorCode() == 40001) {
+                if(sqlState.equals("40001")) {
                     // Signal the database that we will attempt a retry.
                     conn.rollback(sp);
+                    continue;
+                } else if(releaseAttempted) {
+                    throw new AmbiguousCommitException(e);
                 } else {
-                    // This is a not a serialization failure, pass it up the chain.
                     throw e;
                 }
             }
+            break;
         }
+        conn.commit();
     }
 
     public static void main(String[] args) throws ClassNotFoundException, SQLException {
@@ -81,7 +90,9 @@ public class TxnSample {
                 // tutorial.
                 RetryableTransaction transfer = transferFunds(1, 2, 100);
                 retryTransaction(db, transfer);
+
                 // Check balances after transfer.
+                db.setAutoCommit(true);
                 ResultSet res = db.createStatement().executeQuery("SELECT id, balance FROM accounts");
                 while (res.next()) {
                     System.out.printf("\taccount %s: %s\n", res.getInt("id"), res.getInt("balance"));
@@ -90,6 +101,8 @@ public class TxnSample {
                 System.out.println("Insufficient balance");
             } catch(AccountNotFoundException e) {
                 System.out.println("No users in the table with id " + e.account);
+            } catch(AmbiguousCommitException e) {
+                System.out.println("Ambiguous result encountered: " + e);
             } catch(SQLException e) {
                 System.out.println("SQLException encountered:" + e);
             } finally {
--- a/v1.0/transactions.md
+++ b/v1.0/transactions.md
@@ -53,7 +53,7 @@ To handle errors in transactions, you should check for the following types of se
 
 - **Retryable Errors**: Errors with the code `40001` or string `retry transaction`, which indicate the transaction failed because another concurrent or recent transaction accessed the same values. To handle these errors, you should [retry the transaction](#client-side-transaction-retries).
 
-- **Ambiguous Errors**: Errors with the code `XX000` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
+- **Ambiguous Errors**: Errors with the code `40003` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
 
   For example, you might want to read values from the database to see if the transaction successfully wrote values before attempting to write the values again or, alternatively, you might write the data again without seeing if the first write attempt succeeded.
 
--- a/v1.1/transactions.md
+++ b/v1.1/transactions.md
@@ -53,7 +53,7 @@ To handle errors in transactions, you should check for the following types of se
 
 - **Retryable Errors**: Errors with the code `40001` or string `retry transaction`, which indicate the transaction failed because another concurrent or recent transaction accessed the same values. To handle these errors, you should [retry the transaction](#client-side-transaction-retries).
 
-- **Ambiguous Errors**: Errors with the code `XX000` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
+- **Ambiguous Errors**: Errors with the code `40003` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
 
   For example, you might want to read values from the database to see if the transaction successfully wrote values before attempting to write the values again or, alternatively, you might write the data again without seeing if the first write attempt succeeded.
 
--- a/v1.2/transactions.md
+++ b/v1.2/transactions.md
@@ -53,7 +53,7 @@ To handle errors in transactions, you should check for the following types of se
 
 - **Retryable Errors**: Errors with the code `40001` or string `retry transaction`, which indicate the transaction failed because another concurrent or recent transaction accessed the same values. To handle these errors, you should [retry the transaction](#client-side-transaction-retries).
 
-- **Ambiguous Errors**: Errors with the code `XX000` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
+- **Ambiguous Errors**: Errors with the code `40003` that are returned in response to `RELEASE SAVEPOINT` (or `COMMIT` when not using `SAVEPOINT`), which indicate that the state of the transaction is ambiguous, i.e., you cannot assume it either committed or failed. How you handle these errors depends on how you want to resolve the ambiguity.
 
   For example, you might want to read values from the database to see if the transaction successfully wrote values before attempting to write the values again or, alternatively, you might write the data again without seeing if the first write attempt succeeded.
 


andreimatei:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyTrWH6_6Eow024US0:-KxyTrWH6_6Eow024US1:b33q9yk) ([raw file](https://github.com/cockroachdb/docs/blob/326bb4ef9d29a00679c90d7a718d5647d35eded3/_includes/app/TxnSample.java#L54)):*
> ```Java
>                 // If we reach this point, release the savepoint and commit the transaction.
>                 conn.releaseSavepoint(sp);
>                 conn.commit();
> ```

I'd move this `conn.commit();` away, outsides of the loop, for clarity. I don't know if the code is currently correct or not (i.e. if the commit can get a retriable error or not given that we've done the release); in case it does, the rollback call below would be illegal.

If you move the commit away, consider giving some hints about the ambiguity of errors from commit (unless jdbc already makes that clear somehow). In the go client lib we have a special type of error for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-:-KxyUaqJ-rygjxREsGcp:bnfp4nl)*

arjunravinarayan:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyTrWH6_6Eow024US0:-KxyXnN47HvepVO0deU4:b-z2gep6) ([raw file](https://github.com/cockroachdb/docs/blob/326bb4ef9d29a00679c90d7a718d5647d35eded3/_includes/app/TxnSample.java#L54)):*

I'd move this `conn.commit();` away, outsides of the loop, for clarity. I don't know if the code is currently correct or not (i.e. if the commit can get a retriable error or not given that we've done the release); in case it does, the rollback call below would be illegal.

If you move the commit away, consider giving some hints about the ambiguity of errors from commit (unless jdbc already makes that clear somehow). In the go client lib we have a special type of error for that.

</blockquote></details>

I took a stab at it, not sure if I got it right. Would appreciate another round of review.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

andreimatei:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyTrWH6_6Eow024US0:-Kxya3Dw3tfjBjiVBDnF:b4k9cja) ([raw file](https://github.com/cockroachdb/docs/blob/326bb4ef9d29a00679c90d7a718d5647d35eded3/_includes/app/TxnSample.java#L54)):*
<details><summary><i>Previously, arjunravinarayan (Arjun Narayan) wrote…</i></summary><blockquote>

I took a stab at it, not sure if I got it right. Would appreciate another round of review.
</blockquote></details>

I made a mistake: it's not an error from the `conn.commit()` call that's ambiguous, but (some) error from `conn.releaseSavepoint()` - that's the place where we actually commit the transaction on the server.

So, if the release gets an error, we need to discriminate between retriable and other errors (and even within those others, we ideally should identify network errors - those are the ambiguous ones).
Exceptions from the `conn.commit()` call are not ambiguous, they should be left alone.

---

*[_includes/app/TxnSample.java, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyZ7y_oDLUt1nmj38u:-KxyZ7y_oDLUt1nmj38v:b-374loz) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L9)):*
> ```Java
> */
> 
> class AmbiguousResultException extends Exception{}
> ```

I'd say `AmbiguousCommitException`. And I don't really know the java mindset, but wouldn't it be better if this was a subclass of SQLException? Particularly given that it's a checked exception - we don't necessarily want everybody handling it (at least not more so than any database/driver should already want that :) ).

---

*[_includes/app/TxnSample.java, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_P5tDvenzRfa93jJ:-Kxy_P5tDvenzRfa93jK:b31f21h) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L9)):*
> ```Java
> */
> 
> class AmbiguousResultException extends Exception{}
> ```

ideally this would wrap another exception.
Btw, given that this is a code sample and not a library, in my opinion a comment at the commit site would also be sufficient (replacing this exception), and then this is moot.

---

*[_includes/app/TxnSample.java, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyZE-ApjCcM17siXPz:-KxyZE-ApjCcM17siXQ-:bq9rc2q) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L53)):*
> ```Java
>                 tx.run(conn);
> 
>                 // If we reach this point, release the savepoint.
> ```

nit: this comment now says nothing new in addition to the of code line below

---

*[_includes/app/TxnSample.java, line 71 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_-FtX4ZW36zIZuIz:-Kxy_-FtX4ZW36zIZuJ-:bi37m2g) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L71)):*
> ```Java
>             conn.commit();
>         } catch(SQLException e) {
>             if(e.getErrorCode() % 1000 == 0) {
> ```

wip? We 

---

*[_includes/app/TxnSample.java, line 73 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_2-9cu25KBlVkT0h:-Kxy_2-9cu25KBlVkT0i:b-vmuymu) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L73)):*
> ```Java
>             if(e.getErrorCode() % 1000 == 0) {
>                 // It's ambiguous whether the commit succeeded or not.
>                 // TODO(andrei): taking recommendations for what this comment should say.
> ```

I think the comment is all right. Perhaps "ambiguous whether the transaction committed or not". And I'd say the comment should be on the exception declaration, not here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

arjunravinarayan:
PTAL

---

Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[_includes/app/TxnSample.java, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyTrWH6_6Eow024US0:-Ky2pD0Y25PJZPlbk_dM:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/326bb4ef9d29a00679c90d7a718d5647d35eded3/_includes/app/TxnSample.java#L54)):*

I made a mistake: it's not an error from the `conn.commit()` call that's ambiguous, but (some) error from `conn.releaseSavepoint()` - that's the place where we actually commit the transaction on the server.

So, if the release gets an error, we need to discriminate between retriable and other errors (and even within those others, we ideally should identify network errors - those are the ambiguous ones).
Exceptions from the `conn.commit()` call are not ambiguous, they should be left alone.
</blockquote></details>

Done.

---

*[_includes/app/TxnSample.java, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyZ7y_oDLUt1nmj38u:-Ky2-oixtkUklwR8ILtP:b-7utpkv) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L9)):*

I'd say `AmbiguousCommitException`. And I don't really know the java mindset, but wouldn't it be better if this was a subclass of SQLException? Particularly given that it's a checked exception - we don't necessarily want everybody handling it (at least not more so than any database/driver should already want that :) ).
</blockquote></details>

I don't really know the Java(TM) MindSet(TM) either, but I thought the whole point of AmbiguousCommit was that it *needs* the user to handle it based on client logic? It can't be automatically handled - it depends very much on whether the user is doing something idempotent (in which case repeat it), or something where you can do some extra work to figure out what state you are in (a read to check whether it did succeed).

---

*[_includes/app/TxnSample.java, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_P5tDvenzRfa93jJ:-Ky208cqO41aO2XXm8GD:b-w0fg89) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L9)):*

ideally this would wrap another exception.
Btw, given that this is a code sample and not a library, in my opinion a comment at the commit site would also be sufficient (replacing this exception), and then this is moot.
</blockquote></details>

I'd still like to give good guidance here since people are raising questions (see [this Forum post](https://forum.cockroachlabs.com/t/cockroach-transaction-throwing-exceptions/1071/8) that references not just the example code, but also this PR).

---

*[_includes/app/TxnSample.java, line 53 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KxyZE-ApjCcM17siXPz:-Ky20H_S6hfUKC66of1v:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L53)):*

nit: this comment now says nothing new in addition to the of code line below
</blockquote></details>

Done.

---

*[_includes/app/TxnSample.java, line 71 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_-FtX4ZW36zIZuIz:-Ky2p9kWQJWRiYfqgR6d:b-i290sr) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L71)):*

wip? We 
</blockquote></details>

Was following docs instructions, which I assume are wrong? Take a look at the other files I added to this PR .

---

*[_includes/app/TxnSample.java, line 73 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-Kxy_2-9cu25KBlVkT0h:-Ky2p9D5VH11mR_OF3Na:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2cc92f9895da93b24afbb0f4e8b1eee67db1a604/_includes/app/TxnSample.java#L73)):*

I think the comment is all right. Perhaps "ambiguous whether the transaction committed or not". And I'd say the comment should be on the exception declaration, not here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-:-Ky2po8YvYySZLKcnede:bb71pro)*

andreimatei:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

*[_includes/app/TxnSample.java, line 10 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH5bldZsyp7yX2HWiD:-KyH5bldZsyp7yX2HWiE:b-vfbwyb) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L10)):*
> ```Java
> 
> // Ambiguous whether the transaction committed or not.
> class AmbiguousCommitException extends SQLException{}
> ```

you've renamed the exception here, but haven't updated all the uses

---

*[_includes/app/TxnSample.java, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH6WjDwTD5p0Cc1jnU:-KyH6WjDwTD5p0Cc1jnV:bgzs2v8) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L61)):*
> ```Java
>                     // Signal the database that we will attempt a retry.
>                     conn.rollback(sp);
>                 } else if(e.getErrorCode() == 40007) {
> ```

what's this about? Where did you get the code 40007 from? You probably want 40003, but that's probably not the only case where we want to throw `AmbiguousCommitError` - errors in communication with the server are also ambiguous. 
In the go lib we return the respective error whenever `release` fails with a non-retriable error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

arjunravinarayan:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 10 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH5bldZsyp7yX2HWiD:-KyM_rEPi7CMxGd3KwtN:benrr71) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L10)):*

you've renamed the exception here, but haven't updated all the uses
</blockquote></details>

Fixed. Didn't notice because it still compiled because I had the .class files lying around.

---

*[_includes/app/TxnSample.java, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH6WjDwTD5p0Cc1jnU:-KyMaAN5kDVVYmI7IgRc:b-ec3dgi) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L61)):*

what's this about? Where did you get the code 40007 from? You probably want 40003, but that's probably not the only case where we want to throw `AmbiguousCommitError` - errors in communication with the server are also ambiguous. 
In the go lib we return the respective error whenever `release` fails with a non-retriable error.
</blockquote></details>

can I make this just the `else` clause?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

andreimatei:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[_includes/app/TxnSample.java, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH6WjDwTD5p0Cc1jnU:-KyMvqh71j9rz5KvRAWf:b-vvzx0p) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L61)):*
<details><summary><i>Previously, arjunravinarayan (Arjun Narayan) wrote…</i></summary><blockquote>

can I make this just the `else` clause?
</blockquote></details>

So, I'd refrain for adding any code about `40003` here, at least for the moment. There's problems with how we use that error code - we return it for things other than commits. What we should be doing is convert it to a retriable code unless the ambiguity was actually about a commit.

What I would do is mimic the go lib - throw `AmbiguousCommitException` whenever `releaseSavepoint` has run (so you need to keep track about whether that ran or not).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

andreimatei:
I believe this sample also has another problem - we're never calling `conn.rollback()`. I think we have to call it in catch/finally whenever any exception is propagating. The JDBC  doc is not clear on whether this needs to be called or not if `conn.commit()` throws, but it's probably best to always call it. Any exceptions from rollback should be swallowed.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[_includes/app/TxnSample.java, line 94 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyRSqLT3cYsJcfOQe0R:-KyRSqLT3cYsJcfOQe0S:b-2h13vj) ([raw file](https://github.com/cockroachdb/docs/blob/55973be720e12e09782926dbad355e3a6a97b576/_includes/app/TxnSample.java#L94)):*
> ```Java
>                 retryTransaction(db, transfer);
>                 // Check balances after transfer.
>                 ResultSet res = db.createStatement().executeQuery("SELECT id, balance FROM accounts");
> ```

I think you want to `db.setAutoCommit(true);` before doing this, otherwise you're starting a txn that you're never committing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-:-KyRPFPeKfYIKMYeq8yX:b-2yhbnl)*

jseldess:
@arjunravinarayan, @andreimatei, thank you for making these improvements! Are we getting close to merging the changes?

arjunravinarayan:
@jseldess sorry for the delay. We had to sync up in person to discuss some of the subtleties.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[_includes/app/TxnSample.java, line 61 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyH6WjDwTD5p0Cc1jnU:-KzQWXJZsmcTWrGatCQ2:bulbvcf) ([raw file](https://github.com/cockroachdb/docs/blob/a0056b662f6ff239c29d33a6426c386c57510216/_includes/app/TxnSample.java#L61)):*

So, I'd refrain for adding any code about `40003` here, at least for the moment. There's problems with how we use that error code - we return it for things other than commits. What we should be doing is convert it to a retriable code unless the ambiguity was actually about a commit.

What I would do is mimic the go lib - throw `AmbiguousCommitException` whenever `releaseSavepoint` has run (so you need to keep track about whether that ran or not).
</blockquote></details>

Resolved offline with @andreimatei.

---

*[_includes/app/TxnSample.java, line 94 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KyRSqLT3cYsJcfOQe0R:-KzQNhSMoC2re7riebSi:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/55973be720e12e09782926dbad355e3a6a97b576/_includes/app/TxnSample.java#L94)):*

I think you want to `db.setAutoCommit(true);` before doing this, otherwise you're starting a txn that you're never committing.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-:-KzQQA-ShdxzIWndZG2O:b-jkxkjx)*

andreimatei:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 105 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KzQcPBGJxMAxWOG5JF9:-KzQcPBGJxMAxWOG5JFA:b-z5cgtv) ([raw file](https://github.com/cockroachdb/docs/blob/0e2716975f339833b2075813b9ca6b255f397631/_includes/app/TxnSample.java#L105)):*
> ```Java
>                 System.out.println("No users in the table with id " + e.account);
>             } catch(AmbiguousCommitException e) {
>                 System.out.println("Ambiguous result encountered: " +e);
> ```

nit: space in +e.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-:-KzQcoTZLAR5YDc83PtX:bnfp4nl)*

arjunravinarayan:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[_includes/app/TxnSample.java, line 105 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2126#-KzQcPBGJxMAxWOG5JF9:-KzR-IBuWX1K5jcCydBY:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f87037e09883daab9c8c0029f813d7dd4ff00696/_includes/app/TxnSample.java#L105)):*

nit: space in +e.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2126)*

PR #2127, opened 2017-11-03
Remove recent blog links from docs home

jseldess:
Also remove mechanism for auto-generating the links.

Decided to do this with @kuanluo, since the links are below-the-fold.


benesch:
Aww, I kind of liked these. Is the problem that they're always out of date?

jseldess:
I don't mind them either, @benesch, but they don't get many clicks, and @kuanluo suggested they're not great from a design perspective since they're entirely below-the-fold. We'll soon be revising the docs home to feature a video, in which case there'll be even less room.

kuanluo:
@benesch The blog entry was a nice idea to drive traffic to blogs, but because of its position, it doesn't really do that well. And given we're going to do bigger changes on Docs Home, I suggested removing to keep docs focused.

jseldess:
Thanks you very much though, @benesch, for creating the mechanism to refresh these easily! The feature may come in handy again in the future, so would you recommend I leave the rake and xml files in tact with a comment, or is the git history enough?

PR #2138, opened 2017-11-07
Add Pricing option to Product dropdown in header

j-low:
__[description]:__
* Add 'Pricing' to 'Product' dropdown in header.


jseldess:
LGTM, but you need to add it to the mobile version of the topnav as well.

j-low:
@jseldess ah yes I didn't see that.  Update forthcoming.

PR #2141, opened 2017-11-08
typos in the first note "data.alerts.callout"

jess-edwards:
- add "s" on "detail" -- "This guide details how..."
- change "do" to "does" -- "but does not explain..."


sploiselle:
Thank you, @jess-edwards!

PR #2158, opened 2017-11-14
Update install-cockroachdb.html

knz:
Reported by @richardwu

### DIFF
--- a/v1.2/install-cockroachdb.html
+++ b/v1.2/install-cockroachdb.html
@@ -374,6 +374,13 @@ $(document).ready(function(){
         <td>Autoconf</td>
         <td>Version 2.68 or higher is required.</td>
       </tr>
+      <tr>
+        <td>Terminfo / NCurses development libraries</td>
+        <td>This may be
+            part of a ncurses development package (e.g. <code>libtinfo-dev</code> on
+            Debian/Ubuntu, but <code>ncurses-devel</code> on CentOS).
+        </td>
+      </tr>
     </table>
     <p>A 64-bit system is strongly recommended. Building or running CockroachDB on 32-bit systems has not been tested. You'll also need at least 2GB of RAM. If you plan to run our test suite, you'll need closer to 4GB of RAM.</p>
   </li>


v1.1/install-cockroachdb.html
```diff
@@ -374,6 +374,13 @@ <h2 id="whats-next">What's Next?</h2>
         <td>Autoconf</td>
         <td>Version 2.68 or higher is required.</td>
       </tr>
+      <tr>
+        <td>Terminfo / NCurses development libraries</td>
+        <td>This may be
+            part of a ncurses development package (e.g. `libtinfo-dev` on
```
richardwu: I think you'll need to wrap `libtinfo-dev` in `<code>libtinfo-dev</code>` (also `ncurses-devel`).

v1.1/install-cockroachdb.html
```diff
@@ -374,6 +374,13 @@ <h2 id="whats-next">What's Next?</h2>
         <td>Autoconf</td>
         <td>Version 2.68 or higher is required.</td>
       </tr>
+      <tr>
+        <td>Terminfo / NCurses development libraries</td>
```
richardwu: This doesn't wrap too nicely, maybe just `Terminfo/NCurses`?

knz: That would be inaccurate. Folk usually already have libncurses installed, but it's not sufficient.

v1.1/install-cockroachdb.html
knz: Done.

jseldess:
Thanks, @knz. Mind if I take this over and apply the update to the 1.2 docs as well?

knz:
Please do

knz:
Oh wait, actually I made a mistake: this change should **only** be applied to the 1.2 docs. It was an error to change this for 1.1 -- v1.1 is not affected.

jseldess:
OK, @knz. I'll move the change just to the 1.2 version of the page.

PR #2192, opened 2017-11-17
[WIP] Info arch test env

j-low:
__[desc]:__
* Prototype environment for user testing changes proposed changes to docs informational architecture & organization.


j-low:
@jseldess

jseldess:
@j-low, please check out my last commit. After playing around for a while, I have the feel that using a single sidenav per version, like we currently do, is going to be more manageable. The approach now is to include `section: guides`, `section: reference`, etc., in each page's front-matter, and then use conditional logic in the sidenav file to display the relevant content in the sidenav. 

For the 1.1 docs, I've set everything up but the SQL Reference pages.

nstewart:
I think the IA of the sidebar being tied to the new top subnav is really confusing. To me it's not obvious that the two are linked -- I have to look at too many places on the page to get my bearings. Here's an example of how I've seen this solved in the past: https://cdn.dribbble.com/users/87516/screenshots/2646554/ps-nav-01_1x.png

jseldess:
Thanks, @nstewart. @j-low and I are meeting to discuss tomorrow at 2:30. If you're free, would be great to have you in the mix. 

Another option, as mentioned during the docs meeting, is to not retain the same topnav as the rest of the site. I'd very much prefer this option, if we can accept it as a company. Here's a example of where the docs topnav is different than the main site: https://www.twilio.com/docs/

If we're going to let docs diverge from the rest of the site more broadly, we should probably use the full width of the screen, putting the sidnav into the left margin and perhaps putting the page-specific toc in the right margin, like we see in many docs, for example, https://angular.io/tutorial/toh-pt2, https://grpc.io/docs/quickstart/java.html

j-low:
@nstewart while I agree that the implicitly connecting the topnav and sidenav requires some additional attentional investment on the part of the user, it *is* a pattern I've seen on a many docs sites in the past and one that I think is familiar enough to most developers that it wouldn't cause much confusion.

That said (and as I've discussed with @jseldess),  I think there are some things happening in the topnav itself that could be improved to help strengthen the implicit connection between it and the sidenav (e.g. stronger background color; reorganization of nav links, search bar, and version switcher; increasing nav link font size).

Lastly, I think there is something to be said for the case of letting the docs layout diverge more greatly from the main site and may provide some opportunities for additionally improving our information organization (e.g. moving the TOC out of the main content area and widening the main content area both seem worth exploring).

In any case, I think there are compelling arguments for various approaches and I'd be interested to hash them out in a bit more detail.

lhirata:
Have we thought about adding a hover state / tooltip to the secondary top nav to give a preview / overview of what's included in that section?

Currently, the side nav lets you toggle the different sections to see what's included. In the new structure, a reader may need to use trial and error to find relevant info. This may very well be a non-issue and user testing would definitely help determine if we need to provide some kind of section preview.

j-low:
@lhirata are you suggesting something like a hover state dropdown menu that would be populated with clickable section subheadings?  (Basically similar interface as e.g. hovering 'Customers' in the main site topnav)

lhirata:
@j-low yeah, possibly-- a hover dropdown that would, for example, say Reference > "A list of SQL commands we support" (or something better). I'm not sure if we need the clickable subheadings since that might get confusing with the sidebar.

j-low:
@lhirata I think the dropdown clickable subheadings would be redundant in the event that you are already viewing content contained in that parent section.  There may be value in being able to navigate directly to another section's content directly without having to first navigate to the section's main page, if that makes sense.

jseldess:
Supplanted by https://github.com/cockroachdb/docs/pull/3133.

PR #2194, opened 2017-11-18
Tutorial on "follow the workload"

### LINKED ISSUE #1843: Leaseholder rebalancing and leaseholder locality (follow-the-sun)
RFCs:
- [Leaseholder rebalancing](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/leaseholder_rebalancing.md)
- [Leaseholder locality](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/leaseholder_locality.md)
 
Additional explanations of load-based leaseholder rebalancing (new in 1.1): 
[explanations.pdf](https://github.com/cockroachdb/docs/files/1234759/explanations.pdf)

jseldess:
Closes #1843


jseldess:
@a-robinson, @awoods187, PTAL. The overview section is still missing some images to clarify the behavior (ideally animated gifs), but the rest should be pretty solid. I haven't seen the blog post on this topic yet, so I hope this serves as a good companion.

@spencerkimball and @Amruta-Ranade, would appreciate your eyes on this, too.

jseldess:
Added images, so this is now a complete draft.

sploiselle:
This is great, Jesse. I gave it a quick glance-over and wanted some quick detail about how a user would be able to implement this feature on their own cluster, and didn't find a "scannable" solution. Maybe adding a section with the head "Implement Follow the Workload on Your Cluster" would make it more obvious how to achieve it myself?

jseldess:
@sploiselle, all the user needs to do is start each node with the `--locality` flag. It's important for me to know that that doesn't come across on "scan". I'll work on that.

v1.1/demo-follow-the-workload.md
```diff
@@ -0,0 +1,298 @@
...
+
+However, if the nodes were started with the `--locality` flag, after a short while, the cluster would move range 3's lease to node 1, which is closer to the origin of the workload, thus reducing the network roundtrips and increasing the speed of reads.
+
+<img src="{{ 'images/follow-workload-2.png' | relative_url }}" alt="Follow the workload example" style="max-width:100%" />
+
+## Tutorial
+
+### Step 1. Install prerequisites
+
+In this tutorial, you'll use CockroachDB, a Go network tool to simulate network latency on your local workstation, and a Go load generator to simulate client workloads. Before you begin, make sure these applications are installed:
```
Amruta-Ranade: Nit: Can we phrase this as "In this tutorial, you'll use CockroachDB, comcast (a Go network tool....)..." 

The way it reads now seems odd - as if CockroachDB is the Go network tool :)

v1.1/demo-follow-the-workload.md
```diff
@@ -0,0 +1,298 @@
...
+
+This increases the speed of reads, but it doesn't guarantee that the range lease will be anywhere close to the origin of requests. If requests are coming from the US West, for example, and the relevant range lease is on a node in the US East, the requests would likely enter a gateway node in the US West and then get routed to the node with the range lease in the US East.
+
+However, you can cause the cluster to actively move range leases for even better read performance by starting each node with the `--locality` flag. With this flag specified, the cluster knows about the location of each node, so when there's high latency between nodes, the cluster will move active range leases to a node closer to the origin of the majority of the workload. This is especially helpful for applications with workloads that move around throughout the day (e.g., most of the traffic is in the US East in the morning and in the US West in the evening).
```
Amruta-Ranade: Nit: "--locality flag" can be hyperlinked to the locality flag documentation? Also, the locality flag documentation can be updated to reference this document.

awoods187:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzP6n7iRbKA7ehvXYuS:-KzP6n7iRbKA7ehvXYuT:b-tu0qjk) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L7)):*
> ```Markdown
> ---
> 
> "Follow the workload" refers to CockroachDB's ability to dynamically optimize read latency for the location from which most of the workload is originating. This page describes how this works and walks you through a simple demonstration using a local cluster.
> ```

Nit but I think it should be "follow-the-workload"

---

*[v1.1/demo-follow-the-workload.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzP73bun6Shk5yhStbw:-KzP73bun6Shk5yhStbx:bxehkuh) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L25)):*
> ```Markdown
> ### How It Works
> 
> "Follow the workload" is based on the way **range leases** handle read requests. Read requests bypass the Raft consensus protocol, accessesing the range replica with which the range lease is co-located (the leaseholder) and sending the results to the client without needing to coordinate with any of the other range replicas. Bypassing Raft, and the network roundtrips involved, is possible because the leaseholder is guaranteed to be up-to-date already due to the way write request are handled by Raft.
> ```

nit --"with which the range" is a bit hard to read

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

awoods187:
really like this tutorial explanation--i'm going to run through it later this week and will circle back with any additional feedback

---

Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-:-KzP7dBh6whiaeXmIIlC:blt4mmh)*

a-robinson:
I also really like this, nice work!

---

Reviewed 2 of 2 files at r1, 4 of 4 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzP73bun6Shk5yhStbw:-KzPCOWjvqUBNBQjDfN7:b8sgxq1) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L25)):*
> the range replica with which the range lease is co-located

Yeah, I think "the range replica which holds the range lease" would be much clearer.

Also, s/accessesing/accessing

---

*[v1.1/demo-follow-the-workload.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPCbbE7plJR9Yo_dAy:-KzPCbbE7plJR9Yo_dAz:b-5l4ojp) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L25)):*
> due to the way write request are handled by Raft.

It's more due to the fact that all write requests also go to the leaseholder, so it's guaranteed to be up-to-date.

---

*[v1.1/demo-follow-the-workload.md, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDE6qEL41bdX0Jmvi:-KzPDE6qEL41bdX0Jmvj:b-vl4y4w) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L65)):*
> ```Markdown
> ~~~
> 
> This command causes a 100 millisecond delay for all requests on your local workstation.
> ```

If we want to be more specific, it introduces delay for all requests on the loopback interface of your local workstation, meaning it will only effect connections from the machine to itself, not to/from the Internet. We don't necessarily need to be more specific, though.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPEKtVrrHNSjO05XSY:-KzPEKtVrrHNSjO05XSZ:b2fop1y) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L133)):*
> ```Markdown
>     ~~~ shell
>     $ kv -duration 1m -max-rate 100 'postgresql://root@localhost:26259?sslmode=disable'
> ```

Note that this means there will also be considerable latency between the `kv` process and the cockroachdb processes. That doesn't effect the demonstration of the feature at all, but it does make the latency/throughput numbers below look pretty terrible.

There are a couple options:
1. Add more concurrent clients using kv's `-concurrency` flag (e.g. `-concurrency 1000`). This won't help the latency numbers, but will make the throughput look less terrible.
2. Start the cockroach processes with `--advertise-host=localhost` instead of `--host=localhost`, then connect to them via kv with a connection string more like `"postgresql://root@$(hostname):26259?sslmode=disable". This way all the node-to-node traffic would be affected by comcast's rules, but the kv-to-cockroach traffic wouldn't. I'm not 100% sure this'll work since the kv traffic might end up on the loopback interface anyway. I'll check.

The two could also be done in tandem. I still need to test option 2. I tried it locally, but realized that not even your original instructions were working to initialize the cluster. It's possible that my laptop's network interface is messed up -- I'll try to figure out what's going on and whether there's anything wrong with the instructions.

---

*[v1.1/demo-follow-the-workload.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDdrPIvYufyWVD74U:-KzPDdrPIvYufyWVD74V:b-j06s43) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L152)):*
> ```Markdown
>     ~~~
>     _elapsed___errors____________ops___ops/sec(cum)__p95(ms)__p99(ms)_pMax(ms)_____seq(begin/end)
>       120.0s        0           2330           19.4    419.4    637.5   1610.6           0 / 2338
> ```

Nit, but this clearly ran for 2 minutes, which is out of sync with the recommendation to run for 1 minute.

---

*[v1.1/demo-follow-the-workload.md, line 154 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDtGxoM6TZzgEeMlX:-KzPDtGxoM6TZzgEeMlY:b-hkb71u) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L154)):*
> ```Markdown
>       120.0s        0           2330           19.4    419.4    637.5   1610.6           0 / 2338
> 
>     BenchmarkBlocks     2330      51502608.4 ns/op
> ```

Is this garbage output from something else? I don't think it's from kv.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-:-KzPJHf3PkpNGeZArBvk:bockkq8)*

a-robinson:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzPKfpSnNHxvpWDSZtH:b-6ugt2l) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*
> ```Markdown
> ~~~ shell
> $ comcast --device lo0 --latency 100
> ```

This device name is going to be different on different machines. On a GCE VM, it was just `lo`, which is the standard name for Linux machines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPMM9_ajfVMfOEXnz_:-KzPMM9_ajfVMfOEXnza:b-tk76ok) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L133)):*
> ```Markdown
>     ~~~ shell
>     $ kv -duration 1m -max-rate 100 'postgresql://root@localhost:26259?sslmode=disable'
> ```

Also, this is going to 50% reads, 50% writes, but this demo is primarily about read latency, so I'd recommend setting `-read-percent 100`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPEKtVrrHNSjO05XSY:-KzPN3vUXiy6mW7MGpAy:beykl0d) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L133)):*

Note that this means there will also be considerable latency between the `kv` process and the cockroachdb processes. That doesn't effect the demonstration of the feature at all, but it does make the latency/throughput numbers below look pretty terrible.

There are a couple options:
1. Add more concurrent clients using kv's `-concurrency` flag (e.g. `-concurrency 1000`). This won't help the latency numbers, but will make the throughput look less terrible.
2. Start the cockroach processes with `--advertise-host=localhost` instead of `--host=localhost`, then connect to them via kv with a connection string more like `"postgresql://root@$(hostname):26259?sslmode=disable". This way all the node-to-node traffic would be affected by comcast's rules, but the kv-to-cockroach traffic wouldn't. I'm not 100% sure this'll work since the kv traffic might end up on the loopback interface anyway. I'll check.

The two could also be done in tandem. I still need to test option 2. I tried it locally, but realized that not even your original instructions were working to initialize the cluster. It's possible that my laptop's network interface is messed up -- I'll try to figure out what's going on and whether there's anything wrong with the instructions.
</blockquote></details>

Alright, the second option doesn't fix it even if I manually put in in the host's non-loopback IP address, since the kernel is too smart. We should just bump the concurrency a bit to `-concurrency 32`, set `-read-percent 100`, and add a note here explaining why the latency visible in the kv output is 200ms instead of just a couple milliseconds (it's because the kv process's packets get the same 100ms-each-way delay as the packets between the cockroach processes -- this shows that the cockroach process receiving the request is the leaseholder since it doesn't have to consult the others).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPEKtVrrHNSjO05XSY:-KzPOkEuxT8QaPuscRWX:b-jhhojm) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L133)):*

Alright, the second option doesn't fix it even if I manually put in in the host's non-loopback IP address, since the kernel is too smart. We should just bump the concurrency a bit to `-concurrency 32`, set `-read-percent 100`, and add a note here explaining why the latency visible in the kv output is 200ms instead of just a couple milliseconds (it's because the kv process's packets get the same 100ms-each-way delay as the packets between the cockroach processes -- this shows that the cockroach process receiving the request is the leaseholder since it doesn't have to consult the others).
</blockquote></details>

(I should add that there are more advanced ways we could get around this, but they'd just distract from the main point of the demo)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPEKtVrrHNSjO05XSY:-KzPPSrQJPutulsxa1Lu:b7r60cu) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L133)):*

(I should add that there are more advanced ways we could get around this, but they'd just distract from the main point of the demo)
</blockquote></details>

And I figured out what was going on on my laptop, I was accidentally using an old binary that didn't support the init command.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

jseldess:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 7 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzP6n7iRbKA7ehvXYuS:-KzPpFhjPzgI-6BuNwMb:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L7)):*

Nit but I think it should be "follow-the-workload"
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzP73bun6Shk5yhStbw:-KzPpgfYSeo1hSxX-cIe:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L25)):*

> the range replica with which the range lease is co-located

Yeah, I think "the range replica which holds the range lease" would be much clearer.

Also, s/accessesing/accessing
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 25 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPCbbE7plJR9Yo_dAy:-KzPqCNTZn0S9wO1zw2Q:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L25)):*

> due to the way write request are handled by Raft.

It's more due to the fact that all write requests also go to the leaseholder, so it's guaranteed to be up-to-date.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 65 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDE6qEL41bdX0Jmvi:-KzQ0ZH2H0JLGauSjjXQ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L65)):*

If we want to be more specific, it introduces delay for all requests on the loopback interface of your local workstation, meaning it will only effect connections from the machine to itself, not to/from the Internet. We don't necessarily need to be more specific, though.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 133 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPEKtVrrHNSjO05XSY:-KzQ5AStwtSmTF1CENdX:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L133)):*

And I figured out what was going on on my laptop, I was accidentally using an old binary that didn't support the init command.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 152 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDdrPIvYufyWVD74U:-KzQ5I0nazJJoNg7xI6q:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L152)):*

Nit, but this clearly ran for 2 minutes, which is out of sync with the recommendation to run for 1 minute.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 154 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDtGxoM6TZzgEeMlX:-KzQ5IKi89Sun2RpnvUE:boolaly) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L154)):*

Is this garbage output from something else? I don't think it's from kv.
</blockquote></details>

I ran it again and still see that final line. But I don't think this summary is necessary anyway, so I've removed.

---

*[v1.1/demo-follow-the-workload.md, line 29 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzCg_JDvPFbxHjYkuOH-r3-29:-KzPw5xfgCnTnm50qoDK:bamdrv5) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L29)):*

Nit: "--locality flag" can be hyperlinked to the locality flag documentation? Also, the locality flag documentation can be updated to reference this document.
</blockquote></details>

Good call. I updated the `--locality` flag description and cross-referenced.

---

*[v1.1/demo-follow-the-workload.md, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzCg_JDvPFbxHjYkuOH-r3-45:-KzQ0A3ZwjXMgRBmaRpU:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L45)):*

Nit: Can we phrase this as "In this tutorial, you'll use CockroachDB, comcast (a Go network tool....)..." 

The way it reads now seems odd - as if CockroachDB is the Go network tool :)
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzQ0C0E2OgMqi2RUSYn:bmft0bh) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*

This device name is going to be different on different machines. On a GCE VM, it was just `lo`, which is the standard name for Linux machines.
</blockquote></details>

Ah, right. Is there any easy command to find this device name on any machine?

---

*[v1.1/demo-follow-the-workload.md, line 133 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPMM9_ajfVMfOEXnz_:-KzQ5R1DGTyRefclL979:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L133)):*

Also, this is going to 50% reads, 50% writes, but this demo is primarily about read latency, so I'd recommend setting `-read-percent 100`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
:lgtm:

---

Reviewed 1 of 5 files at r4, 4 of 4 files at r5.
Review status: 8 of 9 files reviewed at latest revision, 9 unresolved discussions.

---

*[v1.1/demo-follow-the-workload.md, line 154 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPDtGxoM6TZzgEeMlX:-KzQAoQw-pJ0VU4iSPJp:b-9lpp5h) ([raw file](https://github.com/cockroachdb/docs/blob/1e8dda32c80e2470fa4f04c5c83e560ac8260ce3/v1.1/demo-follow-the-workload.md#L154)):*

I ran it again and still see that final line. But I don't think this summary is necessary anyway, so I've removed.
</blockquote></details>

Ah, I see. You're just running an older version of kv. The current version prints `BenchmarkLoadgenKV`. Removing it LGTM.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzQ87TIoblGElh-RwJu:b-d58r0c) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*

Ah, right. Is there any easy command to find this device name on any machine?
</blockquote></details>

I think we can just say to use `lo0` if you're on Mac, `lo` if you're on Linux, and to check the output of `ifconfig` for the interface responsible for `127.0.0.1` if neither works. There might be an easy command out there somewhere, but I don't know it and a couple different google searches didn't turn up anything that does more than just listing the available interfaces.

---

*[v1.1/demo-follow-the-workload.md, line 65 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ93re1Cquf45xSIXs:-KzQ93re1Cquf45xSIXt:b-9pmyum) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L65)):*
> ```Markdown
> ~~~
> 
> This command causes a 100 millisecond delay for all requests on the loopback interface of your local workstation. It will only effect connections from the machine to itself, not to/from the Internet.
> ```

s/effect/affect

---

*[v1.1/demo-follow-the-workload.md, line 136 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ9DNcrJEATw6cV12m:-KzQ9DNcrJEATw6cV12n:b-pbi6r9) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L136)):*
> ```Markdown
>     ~~~
> 
>     This command initiates 32 concurrent read-only workloads for 1 minute but limits each worker to 100 operations per second (since you're running everything on a single machine). While `kv` is running, it will print some stats to the terminal:
> ```

It doesn't limit each "worker" to 100 operations per second, it limits the entire `kv` process to 100 operations per second.

---

*[v1.1/demo-follow-the-workload.md, line 148 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ9M7pHcSxsAml_-3x:-KzQ9M7pHcSxsAml_-3y:b-ddvsma) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L148)):*
> ```Markdown
>     ~~~
> 
>     {{site.data.alerts.callout_info}}The latency numbers printed reflect the 100 millisecond delay caused by the <code>comcast</code> tool. Normally, these numbers would be much lower.{{site.data.alerts.end}}
> ```

Maybe it's just me, but this doesn't reflect the point I'd like to get across about how the delay is coming in between the `kv` process and `cockroach`, not within cockroach, and thus in a real deployment the latency would be drastically lower. This is a bit wordy, but for example:

"The latency numbers printed reflect are over 200 milliseconds because the 100 millisecond delay in each direction (200ms round-trip) caused by the `comcast` tool also applies to the traffic going from the `kv` process to the `cockroach` process. If you were to set up more advanced rules that excluded the `kv` process's traffic or to run this on a real network with real network delay, you'd see these numbers would be down in the single-digit milliseconds."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-:-KzQB-daDe76BswceRRy:bnfp4nl)*

jseldess:
Review status: 8 of 9 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzQF0HkicC4GltvrYJJ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*

I think we can just say to use `lo0` if you're on Mac, `lo` if you're on Linux, and to check the output of `ifconfig` for the interface responsible for `127.0.0.1` if neither works. There might be an easy command out there somewhere, but I don't know it and a couple different google searches didn't turn up anything that does more than just listing the available interfaces.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 65 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ93re1Cquf45xSIXs:-KzQF-jBZ0ZpMQTqPvEY:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L65)):*

s/effect/affect
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 136 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ9DNcrJEATw6cV12m:-KzQEmeBB1rZdfOE5g6C:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L136)):*

It doesn't limit each "worker" to 100 operations per second, it limits the entire `kv` process to 100 operations per second.
</blockquote></details>

Done.

---

*[v1.1/demo-follow-the-workload.md, line 148 at r6](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzQ9M7pHcSxsAml_-3x:-KzQFoCoBFOYW8Y_sKqR:btz2dkq) ([raw file](https://github.com/cockroachdb/docs/blob/23cb680b6dd2fdb5ce8490cfe5514aca3ad4d7a9/v1.1/demo-follow-the-workload.md#L148)):*

Maybe it's just me, but this doesn't reflect the point I'd like to get across about how the delay is coming in between the `kv` process and `cockroach`, not within cockroach, and thus in a real deployment the latency would be drastically lower. This is a bit wordy, but for example:

"The latency numbers printed reflect are over 200 milliseconds because the 100 millisecond delay in each direction (200ms round-trip) caused by the `comcast` tool also applies to the traffic going from the `kv` process to the `cockroach` process. If you were to set up more advanced rules that excluded the `kv` process's traffic or to run this on a real network with real network delay, you'd see these numbers would be down in the single-digit milliseconds."
</blockquote></details>

Done. Thanks, Alex.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

a-robinson:
:thumbsup: 

---

Reviewed 1 of 6 files at r7.
Review status: 3 of 9 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzQJ0_ZlyiMBz6TbGKi:b-wqwhq4) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*

Done.
</blockquote></details>

Super minor, but in case someone doesn't know what `ifconfig` is, it might be more understandable to say "check the output of running the `ifconfig` command to find the interface ..."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-:-KzQJUGnJF0Otfh0crKl:b4c9lyr)*

jseldess:
Review status: 3 of 9 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[v1.1/demo-follow-the-workload.md, line 62 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2194#-KzPKfpSnNHxvpWDSZtG:-KzQNV0eqEyhMIJT7fLI:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/22b6d5003f7f86d4db202eb537c325347090ef83/v1.1/demo-follow-the-workload.md#L62)):*

Super minor, but in case someone doesn't know what `ifconfig` is, it might be more understandable to say "check the output of running the `ifconfig` command to find the interface ..."
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2194)*

jseldess:
Added a callout that the only use action required to enable this feature is using `--locality` on `cockroach start`.

PR #1934, opened 2017-09-27
New docs on select clauses

knz:
Fixes #1488.

Also includes using statements between '[...]'

PR #2199, opened 2017-11-20
Updated docs for multi-row DML content

Amruta-Ranade:


petermattis:
:lgtm: modulo figuring out what we want to say about `UPDATE`.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[v1.1/insert.md, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvN1-bRyhKUs8UpNk:-KzVvN1-bRyhKUs8UpNl:bzgqr3h) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L14)):*
> ```Markdown
> 
> - [Multi-row `INSERT`](insert.html#insert-multiple-rows) is faster than multiple single-row `INSERT` statements. Whenever possible, use multi-row `INSERT` instead of multiple single-row `INSERT` statements. 
> - Generating and retrieving unique IDs with traditional databases involves using `INSERT` with `SELECT`. But with CockroachDB, use `RETURNING` clause with `INSERT` instead. See [Insert and Return Values](insert.html#insert-and-return-values) for more details.
> ```

I think that should read: "use a `RETURNING` clause".

---

*[v1.1/update.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvf5F-T1rJ687943w:-KzVvf5F-T1rJ687943x:b-5uvuvm) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/update.md#L108)):*
> ```Markdown
> ### Update Multiple Rows
> 
> {{site.data.alerts.callout_success}}Multi-row <code>`UPDATE`</code> is faster than multiple single-row <code>`UPDATE`</code> statements. Whenever possible, use multi-row <code>`UPDATE`</code> instead of multiple single-row <code>`UPDATE`</code> statements.{{site.data.alerts.end}}
> ```

I'm less confident of this statement as we haven't done as much testing with it and the SQL is quite ugly. While it is possible to do multiple updates in a single statement, doing many might be problematic.

Cc @knz, @bdarnell 

---

*[v1.1/upsert.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvyypFK7CUXUUIijz:-KzVvyypFK7CUXUUIik-:b-2lbnq2) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/upsert.md#L102)):*
> ```Markdown
> ~~~
> ~~~ sql
> > UPSERT INTO accounts (id, balance) VALUES (4, 1970.4),(5, 2532.9),(6, 4473.0);
> ```

For consistency: add a space after each comma separating the tuples in the `VALUES` list.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-:-KzW3KQoAEtI6ZTLsQuH:b-cls9oy)*

knz:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[v1.1/insert.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzYGSG31reJ0wGfZsAe:-KzYGSG31reJ0wGfZsAf:bzvh4c) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L39)):*
> ```Markdown
> `AS name` | An alias for the table name. When an alias is provided, it completely hides the actual table name.
> `qualified_name_list` | A comma-separated list of column names, in parentheses.
> `select_stmt` | A comma-separated list of column values or [value expressions](sql-expressions.html) for a single row, in parentheses. To insert values into multiple rows, use a comma-separated list of parentheses. Alternately, you can use [`SELECT`](select.html) statements to retrieve values from other tables and insert them as new rows. See the [Insert a Single Row](#insert-a-single-row), [Insert Multiple Rows](#insert-multiple-rows), [Insert from a `SELECT` Statement](#insert-from-a-select-statement) examples below.<br><br>Each value must match the [data type](data-types.html) of its column. Also, if column names are listed (`qualified_name_list`), values must be in corresponding order; otherwise, they must follow the declared order of the columns in the table.
> ```

It's unfortunate that the PR #1934 has not landed yet. This entry in the table right now is rather incomplete.

---

*[v1.1/insert.md, line 91 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzYGg_g6vLHKjawGAcs:-KzYGg_h93Kd4yEIBFhO:bfsnn8t) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L91)):*
> ```Markdown
> 
> ~~~ sql
> > INSERT INTO accounts (id, balance) VALUES (3, 8100.73), (4, 9400.10);
> ```

This section is misleading in that a VALUES clause is not the only way to insert multiple rows.
See the changes in #1934 - this section should call out that any select clause can be used here as long as it produces multiple rows.

---

*[v1.1/update.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvf5F-T1rJ687943w:-KzYGudAQHjNaAn9koEu:b-wgbbu0) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/update.md#L108)):*

I'm less confident of this statement as we haven't done as much testing with it and the SQL is quite ugly. While it is possible to do multiple updates in a single statement, doing many might be problematic.

Cc @knz, @bdarnell 
</blockquote></details>

Three things here:

1) the term "multi-row UPDATE" is not defined anywhere. UPDATE and DELETE are very unlike INSERT/UPSERT because they do not accept a select clause. The idea that UPDATE/DELETE can "work on multiple rows" exists but it works differently. I think a dedicated section should explain how.

2) as peter says, we indeed have not yet evaluated whether an UPDATE that touches multiple rows is indeed better for performance. I'd be wary to suggest that here until we have measurements.

3) The example given below is not very good. For this particular use case two separate UPDATE statements would have been better than just one -- there is no clear benefit from updating 2 rows instead of just one. If there are any benefits from using UPDATE with many rows instead of just one, you'd need to update dozens of rows at a time. In that case, a CASE .. END expression would make the query extremely difficult to express. If you are looking for an example of an UPDATE that works on many rows at a time, then you'd want to show this kind of example:

```
UPDATE accounts SET balance = balance * 1.05 WHERE id in (2, 3) -- increase the balance by 5% for these customers
```

---

*[v1.1/upsert.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzYINMsgrF7YuqStWmQ:-KzYINMsgrF7YuqStWmR:b95lkbv) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/upsert.md#L87)):*
> ```Markdown
> ### Upsert that Inserts Multiple Rows 
> 
> In this example, the `UPSERT` statement inserts multiple rows into the table.
> ```

Same remark as for INSERT: as explained in #1934 an UPSERT statement takes any select clause as argument, not just VALUES. There are many ways to make it operate on multiple rows at a time, not just VALUES.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2199)*

arjunravinarayan:
Amruta-Ranade are you sure you wanted to request a review from me on this PR? I don't really have much to add re: `UPDATE`, and think what you have here is fine (so LGTM!). I do want to chime in on the other performance best practices you have on your list (especially planning merge JOINs, guidelines for adding/removing secondary indexes, storing secondary indexes), so please include me on any subsequent WIPs and reviews.

Amruta-Ranade:
@arjunravinarayan Yes, just wanted to make sure I didn't miss out on anything in this PR. Will keep you in the loop for the upcoming ones.

arjunravinarayan:
Sounds good. LGTM!

jseldess:
:lgtm:, with some copyedits. Also, you'll need to duplicate these changes in the v1.2 docs.

---

Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[v1.1/insert.md, line 14 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvN1-bRyhKUs8UpNk:-L-DaaNxQoRUFlhdONv-:b-pih3lp) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L14)):*
Yep, I'd also suggest a slight rewording:

> In traditional SQL databases, generating and retrieving unique IDs involves using `INSERT` followed by `SELECT`. In CockroachDB, use a `RETURNING` clause instead.... 

---

*[v1.1/insert.md, line 39 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzYGSG31reJ0wGfZsAe:-L-DbB3_Hm7pvI38nlAg:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L39)):*

It's unfortunate that the PR #1934 has not landed yet. This entry in the table right now is rather incomplete.
</blockquote></details>

Done.

---

*[v1.1/insert.md, line 91 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzYGg_g6vLHKjawGAcs:-L-DbK4NbOMdiEdn8KO5:bsvmf51) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/insert.md#L91)):*

This section is misleading in that a VALUES clause is not the only way to insert multiple rows.
See the changes in #1934 - this section should call out that any select clause can be used here as long as it produces multiple rows.
</blockquote></details>

@knz, would you recommend we expand this example, or perhaps just link to the new Selection Clauses doc?

---

*[v1.1/insert.md, line 13 at r8](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-L-DcKx8mDelViw61fsm:-L-DcKx8mDelViw61fsn:bjg7ob0) ([raw file](https://github.com/cockroachdb/docs/blob/2c4f84b252c3d3e49feb70b7c7469c5d6bcd3964/v1.1/insert.md#L13)):*
This first bullet would read better if it started:

> A single [multi-row `INSERT`](#insert-multiple-rows) statement is faster than multiple single-row `INSERT` statements. Whenever possible, use a multi-row `INSERT` instead of multiple single-row `INSERT` statements.

Also not that, for on-page anchors, you only need the hash, i.e., `@insert-multiple-rows` instead of `insert.html@insert-multiple-rows).
 

---

*[v1.1/insert.md, line 88 at r8](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-L-DbyRf1cHhx5l7lbaP:-L-DbyRf1cHhx5l7lbaQ:bw2pr1p) ([raw file](https://github.com/cockroachdb/docs/blob/2c4f84b252c3d3e49feb70b7c7469c5d6bcd3964/v1.1/insert.md#L88)):*
> ```Markdown
> ### Insert Multiple Rows
> 
> {{site.data.alerts.callout_success}} Multi-row inserts are faster than multiple single-row <code>`INSERT`</code> statements. As a performance best practice, we recommend using multi-row <code>`INSERT`</code> instead of multiple single-row <code>`INSERT`</code> statements whenever possible. {{site.data.alerts.end}}
> ```

You need to remove the backticks from inside the `<code></code>` segments.

---

*[v1.1/update.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvf5F-T1rJ687943w:-L-DdSWkpD8MnI1LABcL:b-qz06ig) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/update.md#L108)):*

Three things here:

1) the term "multi-row UPDATE" is not defined anywhere. UPDATE and DELETE are very unlike INSERT/UPSERT because they do not accept a select clause. The idea that UPDATE/DELETE can "work on multiple rows" exists but it works differently. I think a dedicated section should explain how.

2) as peter says, we indeed have not yet evaluated whether an UPDATE that touches multiple rows is indeed better for performance. I'd be wary to suggest that here until we have measurements.

3) The example given below is not very good. For this particular use case two separate UPDATE statements would have been better than just one -- there is no clear benefit from updating 2 rows instead of just one. If there are any benefits from using UPDATE with many rows instead of just one, you'd need to update dozens of rows at a time. In that case, a CASE .. END expression would make the query extremely difficult to express. If you are looking for an example of an UPDATE that works on many rows at a time, then you'd want to show this kind of example:

```
UPDATE accounts SET balance = balance * 1.05 WHERE id in (2, 3) -- increase the balance by 5% for these customers
```



</blockquote></details>

@amruta, we ended up not making any changes to this statement, right?

---

*[v1.1/upsert.md, line 17 at r8](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-L-Ddtyb-pEf9wKuM1G0:-L-Ddtyb-pEf9wKuM1G1:b-x8sz38) ([raw file](https://github.com/cockroachdb/docs/blob/2c4f84b252c3d3e49feb70b7c7469c5d6bcd3964/v1.1/upsert.md#L17)):*
Same edit as for `INSERT`:

> A single [multi-row `UPSERT`](#upsert-that-inserts-multiple-rows) statement is faster than multiple single-row `UPSERT` statements. Whenever possible, use a multi-row `UPSERT` instead of multiple single-row `UPSERT` statements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-:-L-DeTsNoc-DffmxOuTC:b-9320ik)*

Amruta-Ranade:
@knz Yes, we did not make any changes to UPDATE. I just removed extra space at the end of the doc.

bdarnell:
Review status: 0 of 5 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[v1.1/update.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2199#-KzVvf5F-T1rJ687943w:-L-bgcQqBD9CK2jghkVp:b-o0yzls) ([raw file](https://github.com/cockroachdb/docs/blob/2432f24f72f28a822503522a87d16aa9e8c0994f/v1.1/update.md#L108)):*

@amruta, we ended up not making any changes to this statement, right?
</blockquote></details>

Yeah, we shouldn't talk about this more complex sql until we're more confident that it's a significant win. There are reasons to think so (parallelism, maybe even 1PC eligibility), but I'm not at all confident that we're able to take advantage of these things and actually make it faster.

I think the example using CASE to update two rows with different values is worth showing as a special case (assuming it's faster); we don't need to generalize it to very large numbers of updated rows. I'd rather show that example instead of `UPDATE WHERE id IN` (this type of multi-row update to set the same value everywhere is fairly rare in my experience and is a straightforward application of the WHERE clause. The CASE example highlights something that users may not know about on their own).

This single-statement transformation can be useful to enable server-side automatic retries, even if it doesn't improve performance. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2199)*

knz:
Reviewed 1 of 1 files at r8, 4 of 4 files at r9.
Review status: all files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2199)*

PR #2207, opened 2017-11-23
Fix typos in documentation

rvbaz:
- Fix repeated words
- Fix broken links


rvbaz:
@jseldess No: I was just reading the docs, spotted the typos, then used  `git grep` to get all the similar typos.

PR #2208, opened 2017-11-23
Fix typos in documentation

rvbaz:
Fix repeated words


rvbaz:
@jseldess Now you've tricked me into whipping up a quick bash one-liner to find other duplicated words in the doc ;)

sploiselle:
@rvbaz What a wonderful bunch of catches. Thank you so much for taking the time to do this!

rvbaz:
@jseldess It's very rough (catches a few false positives like repeated integers & co, etc.) but here it is: `find . -type f -name "*.md" -print0 | xargs -0 -n 1 grep -o -E '\b((\w+)\s+)(\2|\1)+\b' | sort | uniq`

jseldess:
Thanks for sharing this, @rvbaz!

PR #2210, opened 2017-11-27
Document how to truncate timeseries data sooner

### LINKED ISSUE #2180: ts: Make timeseries truncation threshold configurable
PR: https://github.com/cockroachdb/cockroach/pull/18632

From release notes:
> New cluster setting `timeseries.resolution_10s.storage_duration` can be used to reduce the storage used by built-in monitoring.

jseldess:
Fixes #2180

### DIFF
--- a/v1.2/admin-ui-overview-dashboard.md
+++ b/v1.2/admin-ui-overview-dashboard.md
@@ -40,6 +40,8 @@ Ranges are subsets of your data, which are replicated to ensure survivability. R
 
 For details about how to control the number and location of replicas, see [Configure Replication Zones](configure-replication-zones.html).
 
+{{site.data.alerts.callout_info}}The timeseries data used to power the graphs in the admin UI is stored within the cluster and accumulates for 30 days before it starts getting truncated. As a result, for the first 30 days or so of a cluster's life, you will see a steady increase in disk usage and the number of ranges even if you aren't writing data to the cluster yourself. For more details, see this <a href="operational-faqs.html#why-is-disk-usage-increasing-despite-lack-of-writes">FAQ</a>.{{site.data.alerts.end}}
+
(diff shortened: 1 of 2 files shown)


a-robinson:
:lgtm: but I'd wait for @mrtracy to chime in as well. In particular, I think we'll want to explain how the two different controls (sampling frequency and storage duration) interact with each other, and I'm guessing it's not obvious judging by the name of the storage duration cluster setting.

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2210#-:-Kzy8D39r35EoAejR0S9:b26qqgv)*

mrtracy:
I would *not* document the second method at all; changing that setting is not tested, and I personally do not think it is plumbed through the system correctly.

mrtracy:
Opened cockroachdb/cockroach/issues/20310 to follow up on my concerns with this setting.

jseldess:
@mrtracy, I removed mention of that second method from the 1.2 docs. Should we also remove from the 1.1 and 1.0 docs? The cluster setting doesn't exist in those versions, so the env variable is the only way to affect timeseries data storage.

mrtracy:
:lgtm:, thanks Jesse

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2210#-:-L-4K2pPoM5QwGNrQGjJ:bq9wyew)*

jseldess:
Merging despite the still broken libpqxx c++ driver site.

PR #2230, opened 2017-11-30
Clarify that truncate is equivalent to dropping and recreating a table

### LINKED ISSUE #2065: kv/storage: re-enable meta2 splits (again), replace RangeLookupRequest
PR: https://github.com/cockroachdb/cockroach/pull/18970

@nvanbenschoten, is it ok to resolve [this limitation](https://www.cockroachlabs.com/docs/stable/known-limitations.html#maximum-cluster-size) in the 1.2 docs now?

### LINKED ISSUE #2229: Update known limitation
Based on thread in Gitter:

The third bullet in this [known limitation](https://www.cockroachlabs.com/docs/stable/known-limitations.html#schema-changes-within-transactions) also applies to `truncate`. We should say so. 

We might also want to update the `truncate` docs to explicitly say that, behind-the-scenes, the command drops and recreates the table.

jseldess:
Fixes #2065 
Fixes #2229

### DIFF
--- a/v1.0/known-limitations.md
+++ b/v1.0/known-limitations.md
@@ -25,7 +25,7 @@ As a workaround, when you need to remove all rows from a large table:
 Within a single [transaction](transactions.html):
 
 - DDL statements cannot follow DML statements. As a workaround, arrange DML statements before DDL statements, or split the statements into separate transactions.
-- A [`CREATE TABLE`](create-table.html) statement containing [`FOREIGN KEY`](foreign-key.html) or [`INTERLEAVE`](interleave-in-parent.html) clauses cannot be followed by statements that reference the new table.
+- A [`CREATE TABLE`](create-table.html) statement containing [`FOREIGN KEY`](foreign-key.html) or [`INTERLEAVE`](interleave-in-parent.html) clauses cannot be followed by statements that reference the new table. This also applies to running [`TRUNCATE`](truncate.html) on such a table because `TRUNCATE` implicitly drops and recreates the table.
 - A table cannot be dropped and then recreated with the same name. This is not possible within a single transaction because `DROP TABLE` does not immediately drop the name of the table. As a workaround, split the [`DROP TABLE`](drop-table.html) and [`CREATE TABLE`](create-table.html) statements into separate transactions.
 
 ## Schema changes between executions of prepared statements
--- a/v1.0/truncate.md
+++ b/v1.0/truncate.md
@@ -6,6 +6,8 @@ toc: false
 
 The `TRUNCATE` [statement](sql-statements.html) deletes all rows from specified tables.
 
+{{site.data.alerts.callout_info}}The <code>TRUNCATE</code> removes all rows from a table by dropping the table and recreating a new table with the same name. For large tables, this is much more performant than deleting each of the rows. However, for smaller tables, it's more performant to use a <a href="delete.html#delete-all-rows"><code>DELETE</code> statement without a <code>WHERE</code> clause<a>.{{site.data.alerts.end}}
+
 <div id="toc"></div>
 
 ## Synopsis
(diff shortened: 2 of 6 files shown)


jseldess:
@Amruta-Ranade, I had to update a known limitation about truncate, so I went ahead and added a note to the truncate docs about using with large vs. small tables. Feel free to tweak along with your performance best practices, if you see fit.

v1.0/known-limitations.md
```diff
@@ -28,6 +28,8 @@ Within a single [transaction](transactions.html):
 - A [`CREATE TABLE`](create-table.html) statement containing [`FOREIGN KEY`](foreign-key.html) or [`INTERLEAVE`](interleave-in-parent.html) clauses cannot be followed by statements that reference the new table.
 - A table cannot be dropped and then recreated with the same name. This is not possible within a single transaction because `DROP TABLE` does not immediately drop the name of the table. As a workaround, split the [`DROP TABLE`](drop-table.html) and [`CREATE TABLE`](create-table.html) statements into separate transactions.
 
+    Note that this limitation also applies to [`TRUNCATE`](../v1.0/truncate.html), which implicitly drops and recreates a table.
+
```
vivekmenezes: Actually this limitation needs to be placed after the CREATE TABLE limitation. TRUNCATE drops and then creates a table, and an INSERT will fail only because the CREATE table has an FK

jseldess: Hmm, I guess I'm not being clear here. Since `truncate` is equivalent to dropping and recreating a table with the same name, I'm saying that the last bullet also means `truncate` can't be used in a transaction. 

Can you clarify how this connects to create table? I have a feeling your referencing another limitation that we haven't yet documented...

vivekmenezes: TRUNCATE can be used in a transaction, but since it creates a table it suffers from the limitations of CREATE. All limitations of CREATE apply to it.

jseldess:
@vivekmenezes, ah, I see. The third bullet states that you can't drop and recreate a table with the same name in a single transaction. Based on our conversation, my understanding is the truncate does just that (drops and recreates a table), so I thought it also shouldn't be allowed within a single transaction. But based on some simple testing, that doesn't actually seem to be a limit for truncate:

```
root@:26257/> begin;
           -> truncate test.t1;
           -> commit;
COMMIT

Time: 25.857ms

root@:26257/> begin;
           -> drop table test.t1;
           -> create table test.t1 (a serial primary key, b string);
           -> commit;
pq: relation "t1" already exists
``` 

I'll rework the limitation to apply to the `create table` bullet instead.

v1.0/known-limitations.md
jseldess: Updated. PTAL.

v1.0/truncate.md
```diff
@@ -6,6 +6,8 @@ toc: false
 
 The `TRUNCATE` [statement](sql-statements.html) deletes all rows from specified tables.
 
+{{site.data.alerts.callout_info}}The <code>TRUNCATE</code> removes all rows from a table by droppioing the table and recreating a new table with the same name. For large tables, this is much more performant than deleting each of the rows. However, for smaller tables, it's more performant to use a <a href="delete.html#delete-all-rows"><code>DELETE</code> statement without a <code>WHERE</code> clause<a>.{{site.data.alerts.end}}
```
vivekmenezes: s/droppioing/dropping

v1.0/known-limitations.md
vivekmenezes: you need to update this.

v1.0/truncate.md
vivekmenezes: and two other locations

jseldess: Sorry. Done.

v1.0/known-limitations.md
jseldess: Done.

vivekmenezes:
LGTM

PR #2231, opened 2017-11-30
Sync styles between docs and main site

j-low:
__[desc]:__
* increase font size on header links
* update font family on header button


j-low:
@kuanluo PTAL

kuanluo:
LGTM! Thank you for the quick fix, @j-low. 👍

PR #2234, opened 2017-12-01
Add video to docs home

jseldess:
@kuanluo, @j-low, @sploiselle, here's a prototype of how we might tweak the docs homepage to feature the new intro video. Basically, I've just removed the bottom three roach tiles and added the video (using tips from https://stackoverflow.com/questions/38270661/displaying-a-youtube-video-with-iframe-full-width-of-page to make it fill the width but stay responsive). 

Not at all tied to this approach. Just want to throw something out there to get things started. PTAL. @j-low, feel free to take over.


jseldess:
@kuanluo, I know you'd like to remove the roach images entirely, so perhaps we could make that change as part of this PR as well. Not sure what to use in place of the images, though.

kuanluo:
@jseldess We can potentially use icons as you suggested to be more inline with the Admin UI nav icon style. Let me mock up something quickly - will attach here shortly.

kuanluo:
@jseldess @sploiselle @j-low @josueeee Two rough mocks below:

**Option 1:**
- Removed the critter illustrations and simplified the tile modules
- Used the card styles elsewhere on the site for the 3 tiles

image

**Option 2:**
The video could benefit from an intro, and if it's given more prominence, the tiles can even be further simplified as text links underneath the video:

image

sploiselle:
This looks great. A few thoughts:

- I would love to have one of the primary actions to be "Install CockroachDB." I try to find that page from the docs site all the time, and only just now realized I could actually get there from the upper right-hand corner. This is also a common design pattern on lots of folks' docs site; let me know if you'd like to dredge up examples.
- Larger buttons for the primary actions make more sense to me. The green text links don't feel like they emphasize the actions we want you to take from that page (and instead really assert a primacy of the video).

jseldess:
I like both designs and don't mind the primacy of the video in the second. But I'm happy to go with @sploiselle's suggestion of using the larger tiles, adding a fourth for install. 

Like the icons, @kuanluo! Can you find one for "Install" and update the design?

j-low:
I agree with @sploiselle re: "Install CockroachDB".  Perhaps that could replace "Quick Start" in the second mockup?  I think "Install" and "Quick Start" content would likely contain a lot of overlap.

@kuanluo I like the layout of the first page.  Since it's a docs page (vs. a marketing page) I think having the action tiles at the top is desirable. As a developer who theoretically visits our doc pages somewhat frequently, I think I might find a top-positioned video obstructive after the first visit when simply I want to get to the necessary info as quickly as possible.

kuanluo:
Thanks all for the feedback! Let's go with Option 1. Instead of having 4 buttons/options on top of the video, what do you think of presenting the following 3 options?

- Install CockroachDB 
- Start a Cluster
- Build an app

Not only would 4 be a bit awkward space-wise, the more buttons we add on top, the more friction we add to the video, which summarizes the benefits nicely with the title slide (global, scalable, resilient).

jseldess:
The video doesn't hit on all of the benefits we have tutorials for, but I see you're point, @kuanluo. I'm fine keeping it to 3: Install, Start, Build. Can you create an updated mock with appropriate icons?

kuanluo:
Here it is: 
image

sploiselle:
@kuanluo Should we do something to introduce the video, like some kind of text that says "Get an Overview of CockroachDB"? I'm realizing there isn't a good title card in the video :S

j-low:
@jseldess @kuanluo PTAL.  Current Teamcity build reflects latest mock.

@sploiselle I see your point re: text to introduce the video, however I personally find that giant "Global / Scalable / Resilient" really grabs my attention already.  That said, that represents my initial reaction and I'm open to suggestion thereupon.

jseldess:
Looks great, @j-low! 

I also see @sploiselle's point. He and I briefly discussed the possibility of uploading a title slide to youtube, if that's an option. If Sean and Kuan can come up with a good title slide and communicates the focus of the video, I think we should test it.

j-low:
@jseldess @sploiselle: yes, I'm definitely for adding a title card to the video itself.

kuanluo:
Agree with adding a title slide for the video. Also, I'm wondering if the play button on the video can catch the style of the homepage video.

Also, @j-low small nits for the white cards: 

- The style (no border, shadow, etc) should match the other white cards on the website. See the homepage, blog article page. 
- On hover, icon + text should be green.

sploiselle:
Added title card to video itself (hat tip to @kuanluo)

jseldess:
Looks great! Thanks, @sploiselle and @kuanluo. 

Once @j-low makes Kuan's final style updates, are we good to merge?

Oh, @j-low, we actually need to update the v1.0 and v1.2 `index.html` files as well. Those pages aren't styled correctly: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/4a277a4e3cec66cb1fd1ce00dce6569b62a93ec4/dev/. So either we should make them match the 1.1 version or correct their styling.

j-low:
@jseldess v1.0 & v1.2 pages are updated (sorry for overlooking those).

Also, @sploiselle I like the change to add the title card!

kuanluo:
LGTM! Thank you, @j-low for accommodating all the nits.

jseldess:
Wohoo. Looks great. I'm merging.

We should follow-up by making sure we can track usage of the video.

PR #2235, opened 2017-12-01
Clarify that featured drivers and ORMs are not yet fully supported

### LINKED ISSUE #2241: Clarify that featured drivers and ORMs have beta-level support
We want the community to understand that, though we have done basic testing of drivers and ORMs recommended on our site, more testing needs to happen before we can claim full support.

jseldess:
@awoods187, as discussed, this PR updates our docs on drivers and ORMs to clarify that we've tested them enough to claim beta-level, but not full, support. PTAL. @cuongdo, would like your eyes, too.

Fixes #2241

### DIFF
--- a/hackathon.md
+++ b/hackathon.md
@@ -3,7 +3,7 @@ title: CockroachDB Hackathon Guide
 summary: Get Up & Running with CockroachDB at a Hackathon
 toc: false
 toc_not_nested: true
-twitter: true
+twitter: false
 no_sidebar: true
 ---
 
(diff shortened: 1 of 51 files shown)


jseldess:
BTW, @awoods187, @benesch mentioned we should eventually document what different levels of "support" mean and suggested looking here for inspiration: https://appdb.winehq.org/objectManager.php?sClass=application&iId=10

awoods187:
LGTM--I agree we should define this publicly in the future. Perhaps once we feel "confident" that we are supporting something to a high standard we can also publish definitions. Intuitively, I think people understand Beta + file a support ticket.

PR #2237, opened 2017-12-01
Clarify treatment of NULLs in SELECTs with ORDER BY clause

### LINKED ISSUE #2233: Highlight treatment of NULLs in SELECTs with ORDER BY clause
Copied from https://github.com/cockroachdb/cockroach/issues/20356:

Postgres sorts `NULL`s last and first with `ASC` and `DESC` directions, respectively.

In Cockroach, `NULL`s sort first and last with `ASC` and `DESC`, respectively. Note: this deviates from postgres behavior.

This is briefly noted in the ["NULL handling" section of the docs](https://www.cockroachlabs.com/docs/stable/null-handling.html#nulls-and-sorting) for `ASC`, but not for `DESC`. We should also crosslink this section in the [`ORDER BY` section of the docs](https://www.cockroachlabs.com/docs/stable/select.html#sorting-retrieved-values).

jseldess:
Fixes #2233

### DIFF
--- a/v1.1/null-handling.md
+++ b/v1.1/null-handling.md
@@ -4,7 +4,7 @@ summary: Learn how NULL values are handled in CockroachDB SQL.
 toc: false
 ---
 
-This page summarizes how `NULL` values are handled in CockroachDB SQL. Each topic is demonstrated via the [built-in SQL client](use-the-built-in-sql-client.html), using the table data in the first section, [NULLs and Logic](#nulls-and-logic). 
+This page summarizes how `NULL` values are handled in CockroachDB SQL. Each topic is demonstrated via the [built-in SQL client](use-the-built-in-sql-client.html), using the table data in the first section, [NULLs and Logic](#nulls-and-logic).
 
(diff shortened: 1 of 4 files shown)


v1.1/null-handling.md
```diff
@@ -262,7 +262,7 @@ For example, let's say you want to calculate the average value of column `b` as
 
 ## NULLs and Sorting
 
-When sorting a column containing `NULL` values, CockroachDB orders `NULL`s lower than the first non-`NULL` value. This differs from PostgreSQL, which orders `NULL`s higher than the last non-`NULL` value.
+When [sorting a column](select.html#sorting-retrieved-values) containing `NULL` values, CockroachDB orders `NULL`s lower than the first non-`NULL` value. This differs from PostgreSQL, which orders `NULL`s higher than the last non-`NULL` value.
```
richardwu: Perhaps add the `NULL`s first with `ASC`, and `NULL`s last  with `DESC` here, too?

I guess the example below shows that pretty clearly but `NULL`s lower than the first non-`NULL` isn't the case with `DESC`.

jseldess: Done. PTAL.

PR #2246, opened 2017-12-05
Update secure-generate-certificates.md

guotie:
I think here should be node2?


jseldess:
Btw, @guotie, if you had any other difficulties with the deployment docs, or any recommendations for improvement, please let us know!

PR #2266, opened 2017-12-08
Release notes v1.2 alpha.20171211

jseldess:
For https://github.com/cockroachdb/cockroach/issues/20457.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -33,9 +33,9 @@ release_info:
     start_time: 2017-11-27 15:10:52.34274101 +0000 UTC
   v1.2:
     name: v1.2-alpha
-    version: v1.2-alpha.20171204
-    build_time: 2017/12/04 14:48:26 (go1.9)
-    start_time: 2017-12-04 15:10:52.34274101 +0000 UTC
+    version: v1.2-alpha.20171211
+    build_time: 2017/12/11 14:48:26 (go1.9)
+    start_time: 2017-12-11 15:10:52.34274101 +0000 UTC
 
 training:
   ccl_license: crl-0-ELzpyNMFGAIiF0NvY2tyb2FjaCBMYWJzIFRyYWluaW5n
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -25,6 +25,8 @@
       version: v1.0
 - title: Testing Releases
   releases:
+    - date: Dec 11, 2017
+      version: v1.2-alpha.20171211
     - date: Dec 4, 2017
       version: v1.2-alpha.20171204
     - date: Nov 13, 2017
--- a/releases/v1.2-alpha.20171211.md
+++ b/releases/v1.2-alpha.20171211.md
@@ -32,27 +32,54 @@ Get future release notes emailed to you:
 
 ### General Changes
 
-
-### Build Changes
-
+- Alpha and beta releases are now published as [Docker images under the name cockroachdb/cockroach-unstable](https://hub.docker.com/r/cockroachdb/cockroach-unstable/). [#20331](https://github.com/cockroachdb/cockroach/pull/20331)
 
 ### SQL Language Changes
 
+- The protocol statement tag for [`CREATE TABLE ... AS ...`](../v1.2/create-table-as.html) is now [`SELECT`](../v1.2/select.html), like in PostgreSQL. [#20268](https://github.com/cockroachdb/cockroach/pull/20268)
+- OIDs can now be compared with inequality operators. [#20367](https://github.com/cockroachdb/cockroach/pull/20367)
+- The [`CANCEL JOB`](../v1.2/cancel-job.html) statement now supports canceling [`IMPORT`](../v1.2/import.html) jobs. [#20343](https://github.com/cockroachdb/cockroach/pull/20343)
+- [`PAUSE JOB`](../v1.2/pause-job.html)/[`RESUME JOB`](../v1.2/resume-job.html)/[`CANCEL JOB`](../v1.2/cancel-job.html) statements can now be used within SQL [transactions](../v1.2/transactions.html). [#20185](https://github.com/cockroachdb/cockroach/pull/20185)
+- Added a cache for the internal `system.table_statistics` table. [#20212](https://github.com/cockroachdb/cockroach/pull/20212)
+- The `intervalstyle` [session variable](../v1.2/set-vars.html) is now supported for PostgreSQL compatibility. [#20274](https://github.com/cockroachdb/cockroach/pull/20274)
+- The [`SHOW [KV] TRACE`](../v1.2/show-trace.html) statement now properly extracts file/line number information when analyzing traces produced in debug mode. Also, the new `SHOW COMPACT [KV] TRACE` statement provides a more compact view on the same data. [#20093](https://github.com/cockroachdb/cockroach/pull/20093)
+- Some queries using `IS NOT NULL` conditions are now better optimized. [#20436](https://github.com/cockroachdb/cockroach/pull/20436)
+- [Views](../v1.2/views.html) now support `LIMIT` and `ORDER BY`. [#20246](https://github.com/cockroachdb/cockroach/pull/20246)
 
 ### Command-Line Interface Changes
 
+- Reduced temporary disk space usage for the `cockroach debug compact` command. [#20460](https://github.com/cockroachdb/cockroach/pull/20460)
+- The [`cockroach node status`](../v1.2/view-node-details.html) and [`cockroach node ls`](../v1.2/view-node-details.html) commands now support a timeout. [#20308](https://github.com/cockroachdb/cockroach/pull/20308)
 
 ### Admin UI Changes
 
+- The Admin UI now sets the `Last-Modified` header when serving assets to permit browser caching. This improves page load times, especially on slow connections [#20429](https://github.com/cockroachdb/cockroach/pull/20429).
 
 ### Bug Fixes
 
+- Removed the possibility for OOM errors during distributed [`IMPORT`](../v1.2/import.html) from csv. [#20506](https://github.com/cockroachdb/cockroach/pull/20506)
+- Fixed a crash triggered by some corner-case queries containing [`ORDER BY`](../v1.2/select.html#sorting-retrieved-values). [#20489](https://github.com/cockroachdb/cockroach/pull/20489)
+- Added missing Distributed SQL flows to the exported `sql.distsql.flows.active` and `sql.distsql.flows.total` metrics and the "Active Flows for Distributed SQL Queries" admin UI graph. [#20503](https://github.com/cockroachdb/cockroach/pull/20503)
+- Fixed an issue with stale buffer data when using the binary format for [`ARRAY`](../v1.2/array.html) values. [#20461](https://github.com/cockroachdb/cockroach/pull/20461)
+- The [`cockroach sql`](../v1.2/use-the-built-in-sql-client.html) shell now better reports the number of rows inserted by a [`CREATE TABLE ... AS ...`](../v1.2/create-table-as.html) statement. Note,  however, that the result are still formatted incorrectly if the `CREATE TABLE ... AS ...` statement creates zero rows in the new table. [#20268](https://github.com/cockroachdb/cockroach/pull/20268)
+- Self-referencing tables can now reference a non-primary index without manually adding an index on the referencing column. [#20325](https://github.com/cockroachdb/cockroach/pull/20325)
+- Fixed an issue where spans for descending indexes were displaying incorrectly and updated `NOT NULL` tokens from `#` to `!NULL`. [#20318](https://github.com/cockroachdb/cockroach/pull/20318)
+- Fixed `BACKUP` jobs to correctly resume in all conditions. [#20185](https://github.com/cockroachdb/cockroach/pull/20185)
+- Fix various race conditions with jobs. [#20185](https://github.com/cockroachdb/cockroach/pull/20185)
+- It is no longer possible to use conflicting `AS OF SYSTEM TIME` clauses in different parts of a query. [#20267](https://github.com/cockroachdb/cockroach/pull/20267)
+- Fixed a panic caused by dependency cycles with `cockroach dump`. [#20255](https://github.com/cockroachdb/cockroach/pull/20255)
+- Prevented context cancellation during lease acquisition from leaking to coalesced requests. [#20424](https://github.com/cockroachdb/cockroach/pull/20424)
 
-### Enterprise Edition Changes
+### Performance Improvements
 
+- Improved handling of `IS NULL` conditions. [#20366](https://github.com/cockroachdb/cockroach/pull/20366)
+- Improved p99 latencies for garbage collection of previous versions of a key, when there are many versions. [#20373](https://github.com/cockroachdb/cockroach/pull/20373)
+- Smoothed out disk usage under very write heavy workloads by syncing to disk more frequently. [#20352](https://github.com/cockroachdb/cockroach/pull/20352)
+- Improved garbage collection of very large [transactions](../v1.2/transactions.html) and large volumes of abandoned write intents. [#20396](https://github.com/cockroachdb/cockroach/pull/20396)
+- Improved table scans and seeks on interleaved parent tables by skipping interleaved children rows at the end of a scan. [#20235](https://github.com/cockroachdb/cockroach/pull/20235)
+- Replaced the interval tree structure in `CommandQueue` with arena-backed concurrent skiplist. This reduces global locking and garbage collection pressure, improving average and tail latencies. [#20300](https://github.com/cockroachdb/cockroach/pull/20300)
 
 ### Doc Updates
 
-
-### New RFCs
-
+- Added an [introduction to CockroachDB video](../v1.2/). [#2234](https://github.com/cockroachdb/docs/pull/2234)
+- Clarified that we have tested the PostgreSQL-compatible drivers and ORMs featured in our documentation enough to claim **beta-level** support. This means that applications using advanced or obscure features of a driver or ORM may encounter incompatibilities. If you encounter problems, please [open an issue](https://github.com/cockroachdb/cockroach/issues/new) with details to help us make progress toward full support. [#2235](https://github.com/cockroachdb/docs/pull/2235)
--- a/releases/v1.2-alpha.20171211.md
+++ b/releases/v1.2-alpha.20171211.md
@@ -1,0 +1,58 @@
+---
+title: What&#39;s New in v1.2-alpha.20171211
+toc: false
+summary: Additions and changes in CockroachDB version v1.2-alpha.20171211
+---
+
+## Dec 11, 2017
+
+Get future release notes emailed to you:
+
+<div class="hubspot-install-form install-form-1 clearfix">
+    <script>
+        hbspt.forms.create({
+            css: '',
+            cssClass: 'install-form',
+            portalId: '1753393',
+            formId: '39686297-81d2-45e7-a73f-55a596a8d5ff',
+            formInstanceId: 1,
+            target: '.install-form-1'
+        });
+    </script>
+</div>
+
+### Downloads
+
+<div id="os-tabs" class="clearfix">
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.2-alpha.20171211.darwin-10.9-amd64.tgz"><button id="mac" data-eventcategory="mac-binary-release-notes">Mac</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.2-alpha.20171211.linux-amd64.tgz"><button id="linux" data-eventcategory="linux-binary-release-notes">Linux</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.2-alpha.20171211.windows-6.2-amd64.zip"><button id="windows" data-eventcategory="windows-binary-release-notes">Windows</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.2-alpha.20171211.src.tgz"><button id="source" data-eventcategory="source-release-notes">Source</button></a>
+</div>
+
+### General Changes
+
+
+### Build Changes
+
+
+### SQL Language Changes
+
+
+### Command-Line Interface Changes
+
+
+### Admin UI Changes
+
+
+### Bug Fixes
+
+
+### Enterprise Edition Changes
+
+
+### Doc Updates
+
+
+### New RFCs
+
--- a/v1.2/install-cockroachdb.html
+++ b/v1.2/install-cockroachdb.html
@@ -85,7 +85,7 @@ $(document).ready(function(){
 <div id="mac-installs" class="clearfix install-wrap">
 <a href="#use-homebrew" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-homebrew">Use <div class="c2a">Homebrew</div></a>
 <a href="#download-the-binary" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-binary">Download the <div class="c2a">Binary</div></a>
-<!-- <a href="#build-from-source" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-source">Build from <div class="c2a">Source</div></a> -->
+<a href="#build-from-source" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-source">Build from <div class="c2a">Source</div></a>
 <a href="#use-docker" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-docker">Use <div class="c2a">Docker</div></a>
 </div>
 
@@ -168,7 +168,7 @@ $(document).ready(function(){
 
 </div>
 
-<!-- <div id="build-from-source" class="install-option" style="display: none;">
+<div id="build-from-source" class="install-option" style="display: none;">
 <ol>
   <li>
     <p>Install the following prerequisites, as necessary:</p>
@@ -246,7 +246,7 @@ $(document).ready(function(){
 
 {% include diagnostics-callout.html %}
 
-</div> -->
+</div>
 
 <div id="use-docker" class="install-option" style="display: none;">
 
@@ -303,7 +303,7 @@ $(document).ready(function(){
 
 <div id="linux-installs" class="clearfix install-wrap">
 <a href="#download-the-binary-linux" class="install-button linux-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="linux-binary" data-eventlabel="">Download the <div class="c2a">Binary</div></a>
-<!-- <a href="#build-from-source-linux" class="install-button linux-button" data-eventcategory="buttonClick-doc-install" data-eventaction="linux-source">Build from <div class="c2a">Source</div></a> -->
+<a href="#build-from-source-linux" class="install-button linux-button" data-eventcategory="buttonClick-doc-install" data-eventaction="linux-source">Build from <div class="c2a">Source</div></a>
 <a href="#use-docker-linux" class="install-button linux-button" data-eventcategory="buttonClick-doc-install" data-eventaction="linux-docker">Use <div class="c2a">Docker</div></a>
 </div>
 
@@ -348,7 +348,7 @@ $(document).ready(function(){
 
 </div>
 
-<!-- <div id="build-from-source-linux" class="install-option" style="display: none;">
+<div id="build-from-source-linux" class="install-option" style="display: none;">
 <ol>
   <li>
     <p>Install the following prerequisites, as necessary:</p>
@@ -434,7 +434,7 @@ $(document).ready(function(){
 
 {% include diagnostics-callout.html %}
 
-</div> -->
+</div>
 
 <div id="use-docker-linux" class="install-option" style="display: none;">
 


jseldess:
Need some help from the following devs. I assume most of these don't require a note, but please check your name off after either adding a comment with release notes or confirming that none are needed:

- [ ] @benesch 
- [ ] @tschottdorf 
- [ ] @nvanbenschoten 
- [ ] @knz 
- [ ] @justinj 
- [ ] @m-schneider 
- [ ] @mberhault 
- [ ] @a-robinson 
- [ ] @lego 
- [ ] @spencerkimball 
- [ ] @RaduBerinde 
- [ ] @jordanlewis 

```
* | |   commit ff49a9d83d20e201f3546bef70ced64a74b2a714
|\ \ \  Merge: d0472f0d2 749885cc7
| | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | Date:   Wed Dec 6 11:18:51 2017 -0500
| | | | 
| | | |     Merge pull request #20429 from benesch/ui-cache
| | | |     
| | | |     ui: allow browsers to cache assets
| | | | 
| * | | commit 749885cc76435b6de737909515bfdb67876ed8a8
| | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | Date:   Sun Dec 3 12:40:01 2017 -0500
| | | | 
| | | |     ui: allow browsers to cache assets
| | | |     
| | | |     Include the the UI asset files' mtimes in bindata.go so that Go's HTTP
| | | |     server can properly compute a Last-Modified header. This allows
| | | |     browsers to cache the UI assets, which can improve performance
| | | |     significantly on slow connections.
| | | |     
| | | |     (The metadata is not sensitive; it was previously stripped because the
| | | |     compiled UI was checked into version control and so needed to be
| | | |     reproducible.)
| | | |     
| | | |     Partially fixes #20425.
| | | |   
* | | |   commit d0472f0d233598de089d62ef718c22f444dba8ea
|\ \ \ \  Merge: 459184b05 34c65badc
| | | | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | | | Date:   Wed Dec 6 11:16:40 2017 -0500
| | | | | 
| | | | |     Merge pull request #16838 from tschottdorf/rfc_change_feeds
| | | | |     
| | | | |     RFC: Storage-Level Primitive for Change Feeds
| | | | | 
| * | | | commit 34c65badc7f07e5f6d215bf650c17770fa55f5b9
| | |_|/  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| |/| |   Date:   Wed Dec 6 10:38:41 2017 -0500
| | | |   
| | | |       RFC: Storage-Level Primitive for Change Feeds
| | | |       
| | | |       This RFC is incomplete. Is is merged in `draft` status.
| | | |       
| | | |       This RFC proposes a mechanism for subscribing to changes to a set of key ranges
| | | |       starting at an initial timestamp.
| | | |       
| | | |       It is for use by Cockroach-internal higher-level systems such as distributed
| | | |       SQL, change data capture, or a Kafka producer endpoint. A "sister RFC" detailing
| | | |       these higher-level systems is in #17535. See also #19222 for a related
| | | |       follower-reads RFC.
| | | |   
* | | |   commit 459184b0544ff0eeb86ea5afc548e75d5441d079
|\ \ \ \  Merge: ccc98368d aa9833fe8
| | | | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | | | Date:   Wed Dec 6 10:42:06 2017 -0500
| | | | | 
| | | | |     Merge pull request #20184 from tschottdorf/cli-debug-intents
| | | | |     
| | | | |     cli: pretty-print intents in `debug keys`
| | | | | 
| * | | | commit aa9833fe8a820c8d3d5772c8c9150ad182c0d8f4
| | | | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | | | Date:   Mon Nov 20 17:19:58 2017 -0500
| | | | | 
| | | | |     cli: pretty-print intents in `debug keys`
| | | | |     
| | | | |     Can improve this by pretty-printing the inlined value, but may not
| | | | |     be necessary for a while.
| | | | | 
| * | | | commit 29e072159ffc33400615e77c0255767eb4a921d2
|  / / /  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | |   Date:   Mon Nov 20 17:19:24 2017 -0500
| | | |   
| | | |       cli: print timestamps first in `debug keys`
| | | |       
| | | |       It is more pleasant to read and if undesired, the can easily be stripped off using `cut`.
| | | |   
* | | | | | |   commit b2e7cd6a81c5a1381fff62e7b7a5edd74d236cac
|\ \ \ \ \ \ \  Merge: 26fff804f 7f7ef4b84
| | | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | | Date:   Tue Dec 5 14:43:26 2017 -0500
| | | | | | | | 
| | | | | | | |     Merge pull request #20424 from nvanbenschoten/nvanbenschoten/leaseCtx
| | | | | | | |     
| | | | | | | |     storage: use background context for async lease requests
| | | | | | | | 
| * | | | | | | commit 7f7ef4b8482126bfdb1f316817570822a4e7bdff
| | | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | | Date:   Sat Dec 2 19:35:20 2017 -0500
| | | | | | | | 
| | | | | | | |     storage: use background context for async lease requests
| | | | | | | |     
| | | | | | | |     Fixes #20195
| | | | | | | |     Closes #20340
| | | | | | | |     
| | | | | | | |     Lease requests are special in that multiple goroutines can join to a
| | | | | | | |     single request. The actual lease request previously inherited the
| | | | | | | |     context of the originating request. If that context was canceled the
| | | | | | | |     cancellation error would be propagated to all of the waiters, which
| | | | | | | |     caused flakiness in tests and was generally bad.
| | | | | | | |     
| | | | | | | |     This change implements the approach discussed in:
| | | | | | | |     https://github.com/cockroachdb/cockroach/pull/20340#issuecomment-348245665
| | | | | | | |     
| | | | | | | |     The async lease request is now given the background context instead of
| | | | | | | |     the originating request's context. Each goroutine then takes out a
| | | | | | | |     handle on the lease request which allows it to unregister from the
| | | | | | | |     request on its own context cancellations. The async task performing the
| | | | | | | |     lease request is then only canceled if the reference count of handles
| | | | | | | |     waiting on it drops to zero.
| | | | | | | | 
| * | | | | | | commit f1f555e46505e49308a4417ccb1abfd5c3b0f648
| | |_|_|/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| |/| | | | |   Date:   Sat Dec 2 16:57:54 2017 -0500
| | | | | | |   
| | | | | | |       storage: hold reference to Replica in pendingLeaseRequest
| | | | | | |       
| | | | | | |       We previously passed a `Replica` to each of its functions. This was
| | | | | | |       messy and created an opportunity for the wrong replica to be provided.
| | | | | | |   
* | | | | | |   commit 26fff804f9abe591a47a8a36f8203243f462ce7d
|\ \ \ \ \ \ \  Merge: fd906d460 9b7d32306
| |_|_|_|/ / /  Author: kena <knz@users.noreply.github.com>
|/| | | | | |   Date:   Tue Dec 5 14:33:46 2017 -0500
| | | | | | |   
| | | | | | |       Merge pull request #20436 from sydnever/20237
| | | | | | |       
| | | | | | |       sql: any "not-null" filters should be trimmed according to non-null…
| | | | | | | 
| * | | | | | commit 9b7d323068506afcdb00de896946cca39d8adb00
| | | | | | | Author: Syd <324760805@qq.com>
| | | | | | | Date:   Mon Dec 4 16:59:41 2017 +0800
| | | | | | | 
| | | | | | |     sql: removes any unnecessary IS NOT NULL filters on non-nullable columns.
| | | | | | |   
* | | | | |   commit 6377b6d310fb69f44ec4b18f484de76689870829
|\ \ \ \ \ \  Merge: 46415a816 0ba44d256
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Tue Dec 5 11:35:12 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20414 from nvanbenschoten/nvanbenschoten/reduceOOM
| | | | | | |     
| | | | | | |     storage: reduce tscache page size in tests
| | | | | | | 
| * | | | | | commit 0ba44d2565baff669148e9d9b7edc7f6cff5cc13
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Fri Dec 1 18:35:30 2017 -0500
| | | | | | | 
| | | | | | |     storage: reduce tscache page size in tests
| | | | | | |     
| | | | | | |     Fixes  #20401.
| | | | | | |     Closes #20405.
| | | | | | |     
| | | | | | |     We were already reducing the page size of the `sklImpl` timestamp cache
| | | | | | |     during race testing. We now do it during all testing that uses test
| | | | | | |     configurations.
| | | | | | |   
* | | | | | |   commit 46415a81642e8d9a74f70b2de4173c4d982d175f
|\ \ \ \ \ \ \  Merge: aa04a575d 6bcdbf7a5
| | | | | | | | Author: Justin Jaffray <justin.jaffray@gmail.com>
| | | | | | | | Date:   Tue Dec 5 10:44:33 2017 -0500
| | | | | | | | 
| | | | | | | |     Merge pull request #20214 from taku-k/support-json_object_keys
| | | | | | | |     
| | | | | | | |     sql: support json_object_keys of JSON builtins
| | | | | | | | 
| * | | | | | | commit 6bcdbf7a5dfcc4f7f5eb6b32ed4521bf0bff6a3f
|  / / / / / /  Author: Takuya Kuwahara <taakuu19@gmail.com>
| | | | | | |   Date:   Wed Nov 22 07:25:19 2017 +0900
| | | | | | |   
| | | | | | |       sql: support json{b}_object_keys of JSON builtins
| | | | | | |       
| | | | | | |       Add json builtin functions as follows:
| | | | | | |       json_object_keys
| | | | | | |       jsonb_object_keys
| | | | | | |   
* | | | | | |   commit 5bc744dc01ce9febb19b27f75e8292674ff830e4
|\ \ \ \ \ \ \  Merge: e571b68f1 a8e1a1253
| | | | | | | | Author: kena <knz@users.noreply.github.com>
| | | | | | | | Date:   Mon Dec 4 18:12:49 2017 -0500
| | | | | | | | 
| | | | | | | |     Merge pull request #20416 from knz/20171202-destruct-start
| | | | | | | |     
| | | | | | | |     sql: do not recurse the initialization of local execution via Start()
| | | | | | | | 
| * | | | | | | commit a8e1a1253c846019fe4d57253b8ae459440c56be
| | | | | | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | | | | | Date:   Sat Dec 2 02:05:20 2017 +0100
| | | | | | | | 
| | | | | | | |     sql: simplify startPlan()
| | | | | | | |     
| | | | | | | |     .. and also simplify the SCRUB implementation to use the `runParams`
| | | | | | | |     struct more.
| | | | | | | | 
| * | | | | | | commit c4d6bc11cc95368efaa2b13b94987ab73201a526
| | | | | | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | | | | | Date:   Sat Dec 2 01:42:59 2017 +0100
| | | | | | | | 
| | | | | | | |     sql: make startExec() start sub-queries itself
| | | | | | | | 
| * | | | | | | commit 408a987977993bd7d4b9ddddee3cc606879fb569
| | | | | | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | | | | | Date:   Sat Dec 2 01:30:50 2017 +0100
| | | | | | | | 
| | | | | | | |     sql: remove the planNode Start() method
| | | | | | | |     
| | | | | | | |     The Start() method on planNode had two roles:
| | | | | | | |     - initialize local execution
| | | | | | | |     - recurse into sub-plans.
| | | | | | | |     
| | | | | | | |     Eventually we want the code for local execution to become independent
| | | | | | | |     from the logical planning, and in particular it should not directly
| | | | | | | |     know/assume how to access the "children nodes" of a given relational
| | | | | | | |     operator. In order to make the code agnostic, we can delegate the task
| | | | | | | |     of the recursion to a suitable dispatch method.
| | | | | | | |     
| | | | | | | |     It so happens that we already have such a dispatch function -
| | | | | | | |     walkPlan() already used for other things.
| | | | | | | |     
| | | | | | | |     This patch thus isolates the code that initializes local execution in
| | | | | | | |     a new method startExec() which doesn't need to recurse, and changes
| | | | | | | |     startPlan() to call this method via walkPlan().
| | | | | | | | 
| * | | | | | | commit 12495018b46e6ed35cc4db1ca95ef708a619367f
| | | | | | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | | | | | Date:   Sat Dec 2 01:26:14 2017 +0100
| | | | | | | | 
| | | | | | | |     sql: make the planObserver interface smarter
| | | | | | | |   
* | | | | | | |   commit e571b68f10d0a68cbcb21299ce3678b6174aae20
|\ \ \ \ \ \ \ \  Merge: 500123d16 b07e426c5
| | | | | | | | | Author: Masha Schneider <masha@cockroachlabs.com>
| | | | | | | | | Date:   Mon Dec 4 17:49:18 2017 -0500
| | | | | | | | | 
| | | | | | | | |     Merge pull request #20172 from m-schneider/invindexsql
| | | | | | | | |     
| | | | | | | | |     sql: support INVERTED when creating indexes
| | | | | | | | | 
| * | | | | | | | commit b07e426c5abc2676b940be0296afa73e73c859af
| | |_|_|_|_|_|/  Author: Masha Schneider <masha@cockroachlabs.com>
| |/| | | | | |   Date:   Mon Nov 20 13:09:12 2017 -0500
| | | | | | | |   
| | | | | | | |       sql: support INVERTED when creating indexes
| | | | | | | |       
| | | | | | | |       Adding parsing support for CREATE INVERTED INDEX...
| | | | | | | |   
* | | | | |   commit c51247857e1838cdef9170a5ed3674aa0ef944c6
|\ \ \ \ \ \  Merge: c58a3fa07 f147a8fa6
| |_|/ / / /  Author: marc <marc.berhault@gmail.com>
|/| | | | |   Date:   Mon Dec 4 15:22:50 2017 -0500
| | | | | |   
| | | | | |       Merge pull request #20440 from mberhault/marc/fix_return_nil
| | | | | |       
| | | | | |       misc: fix "if err != nil { return nil }"
| | | | | | 
| * | | | | commit f147a8fa6ec9d5aa306248c06edc8583aaa7f4bb
| | |_|_|/  Author: marc <marc@cockroachlabs.com>
| |/| | |   Date:   Mon Dec 4 09:47:51 2017 -0500
| | | | |   
| | | | |       misc: fix "if err != nil { return nil }"
| | | | |       
| | | | |       I encountered a random one somewhere and poked around for more.
| | | | |       
| | | | |       I **think** this is right. There are no comments in any of them about
| | | | |       not returning an error on errors.
| | | | |   
* | | | | |   commit 8dc1d751b1177aa5d263b4403bea308102d2ca7b
|\ \ \ \ \ \  Merge: 333060610 5891f5b46
| | | | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | | | Date:   Mon Dec 4 11:17:09 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20442 from a-robinson/rocks-cp
| | | | | | |     
| | | | | | |     c-deps: Pull in fixes for rocksdb disk contention issues
| | | | | | | 
| * | | | | | commit 5891f5b465812256d22fef44ae8f6deb142bee66
|/ / / / / /  Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | |   Date:   Mon Dec 4 10:40:20 2017 -0500
| | | | | |   
| | | | | |       c-deps: Pull in fixes for rocksdb disk contention issues
| | | | | |       
| | | | | |       https://github.com/facebook/rocksdb/pull/3211 fixes the CMake build
| | | | | |       flags to enable the `sync_file_range` syscall which is required by the
| | | | | |       `bytes_per_sync` feature.
| | | | | |       
| | | | | |       https://github.com/facebook/rocksdb/pull/3208 modernizes how background
| | | | | |       compactions/flushes are configured, which helped prevent stalls in some
| | | | | |       of my testing.
| | | | | |       
| | | | | |       cc #19699
| | | | | |   
* | | | | |   commit 333060610f00b66cfe5ce5b0d50d896053646eec
|\ \ \ \ \ \  Merge: 4fbc79df1 e71508004
| | | | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | | | Date:   Mon Dec 4 10:22:50 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20432 from a-robinson/cqueuetime
| | | | | | |     
| | | | | | |     storage: Include command timestamps in command queue debug string
| | | | | | | 
| * | | | | | commit e71508004628a767b927574a3cbf4f62d81a15ba
| |/ / / / /  Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | |   Date:   Sun Dec 3 13:08:46 2017 -0500
| | | | | |   
| | | | | |       storage: Include command timestamps in command queue debug string
| | | | | |       
| | | | | |       Would have been nice when looking into #20114
| | | | | |   
* | | | | |   commit 4fbc79df133e843de90c54efe2e4f5208375f42e
|\ \ \ \ \ \  Merge: edf092fd1 1f83330b6
| | | | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | | | Date:   Mon Dec 4 10:22:07 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20431 from a-robinson/distsenderslow
| | | | | | |     
| | | | | | |     kv: Print which replica we're waiting on if an RPC is slow
| | | | | | | 
| * | | | | | commit 1f83330b6f4b0a7338b86f4212c8a85a2d986f4c
| | | | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | | | Date:   Sun Dec 3 14:27:53 2017 -0500
| | | | | | | 
| | | | | | |     kv: Print which replica we're waiting on if an RPC is slow
| | | | | | |     
| | | | | | |     This has been bugging me for quite a while -- normally you can figure
| | | | | | |     out which replica is being waited on pretty easily when a cluster is
| | | | | | |     still running, but not if all you have are logs provided by a user after
| | | | | | |     the fact.
| | | | | | | 
| * | | | | | commit 119e41a75855b268db086ccf937169e4f0691d5c
| |/ / / / /  Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | |   Date:   Sun Dec 3 13:19:48 2017 -0500
| | | | | |   
| | | | | |       kv: Indicate need for lock to be held in method name
| | | | | |   
* | | | | |   commit edf092fd1b8c7bfb3313844e20c9a54780fc6e66
|\ \ \ \ \ \  Merge: f437f7d9b 410361e4f
| |_|/ / / /  Author: Joey Pereira <legojoey17@gmail.com>
|/| | | | |   Date:   Mon Dec 4 10:05:34 2017 -0500
| | | | | |   
| | | | | |       Merge pull request #19960 from lego/scrub-constraints-checks
| | | | | |       
| | | | | |       sql,parser: Add SCRUB support for checking SQL CHECK and FOREIGN KEY
| | | | | | 
| * | | | | commit 410361e4f2a744d33929363034cc37807024785c
| | | | | | Author: Joey Pereira <joey@pereira.io>
| | | | | | Date:   Wed Nov 8 15:44:56 2017 -0500
| | | | | | 
| | | | | |     sql: Implement scrub constraint check for SQL FOREIGN KEY
| | | | | |     
| | | | | |     Added support for validating SQL FOREIGN KEY constraints with the
| | | | | |     `CONSTRAINT ALL` scrub option.
| | | | | | 
| * | | | | commit 376ea3c6e60dbc80c02930a5ceb48a185f6860df
| | | | | | Author: Joey Pereira <joey@pereira.io>
| | | | | | Date:   Wed Nov 8 14:50:00 2017 -0500
| | | | | | 
| | | | | |     sql: Implement scrub constraint check for SQL CHECK
| | | | | |     
| | | | | |     Added support for validating SQL CHECK constraints with the
| | | | | |     `CONSTRAINT ALL` scrub option.
| | | | | | 
| * | | | | commit a871690aea5ffc14d3cf43e5b6c10d7f2658ee46
|/ / / / /  Author: Joey Pereira <joey@pereira.io>
| | | | |   Date:   Wed Nov 8 13:11:13 2017 -0500
| | | | |   
| | | | |       sql,parser: Add SCRUB TABLE ... WITH OPTIONS CONSTRAINT syntax
| | | | |       
| | | | |       This adds the stubbed scrub command to check constraints
| | | | |       
| | | | |           SCRUB TABLE <table> WITH OPTIONS CONSTRAINT ALL
| | | | |           SCRUB TABLE <table> WITH OPTIONS CONSTRAINT (<constraint_name>...)
| | | | |   
* | | | | |   commit 68a40e4a7d14bc5503e13de67c1ce88dc5b451e1
|\ \ \ \ \ \  Merge: c98e4ea2d 4e5222968
| | | | | | | Author: Spencer Kimball <spencer.kimball@gmail.com>
| | | | | | | Date:   Fri Dec 1 10:16:00 2017 -0800
| | | | | | | 
| | | | | | |     Merge pull request #20398 from spencerkimball/cleanup-all-txn-intents
| | | | | | |     
| | | | | | |     storage: fix a bug which failed to resolve all intents after pushing txn
| | | | | | | 
| * | | | | | commit 4e52229688710a03092a84d99422c461f0865773
| | | | | | | Author: Spencer Kimball <spencer.kimball@gmail.com>
| | | | | | | Date:   Fri Dec 1 09:45:00 2017 -0800
| | | | | | | 
| | | | | | |     storage: fix a bug which failed to resolve all intents after pushing txn
| | | | | | |     
| | | | | | |     This has been in the code a long time. We would ignore all but the first
| | | | | | |     intent for a transaction being pushed. Note that this wouldn't happen
| | | | | | |     on a consistent read, only an inconsistent read, so it would be
| | | | | | |     infrequent in practice.
| | | | | | |   
* | | | | |   commit 7668b831ec355a0161e1b20fe8586922358f18af
|\ \ \ \ \ \  Merge: 911228dd7 45d88619e
| |/ / / / /  Author: Nikhil Benesch <nikhil.benesch@gmail.com>
|/| | | | |   Date:   Thu Nov 30 17:06:00 2017 -0500
| | | | | |   
| | | | | |       Merge pull request #20361 from benesch/bump-gogoproto
| | | | | |       
| | | | | |       Bump gogoproto to fix protobuf compilation warning
| | | | | | 
| * | | | | commit 45d88619e49c0d59b60425700a594285d29b9952
| | | | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | | | Date:   Thu Nov 30 15:37:36 2017 -0500
| | | | | | 
| | | | | |     enginepb: remove unused proto import
| | | | | |     
| | | | | |     This import is strictly unnecessary, but was provided as a workaround
| | | | | |     for gogo/protobuf#207, which is now fixed.
| | | | | | 
| * | | | | commit 08ba590aa74c554093dd146f18c4b82e4c900b18
| | | | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | | | Date:   Thu Nov 30 15:34:25 2017 -0500
| | | | | | 
| | | | | |     deps: bump gogoproto to fix protobuf compilation warning
| | | | | |     
| | | | | |     Pick up a fix for gogo/protobuf#207, which was causing spurious warnings
| | | | | |     when compiling protobufs.
| | | | | |   
* | | | | |   commit 911228dd7f97b7e997e66d02f2668fcf6f072f85
|\ \ \ \ \ \  Merge: 46f46522f 33f9ceca6
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Thu Nov 30 16:17:57 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20309 from nvanbenschoten/nvanbenschoten/tscacheMetrics
| | | | | | |     
| | | | | | |     tscache: add intervalSkl metrics
| | | | | | | 
| * | | | | | commit 33f9ceca68e49c204e0de6606e0853734ef6b62b
| |/ / / / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | |   Date:   Tue Nov 28 14:40:25 2017 -0500
| | | | | |   
| | | | | |       tscache: add intervalSkl metrics
| | | | | |       
| | | | | |       These metrics will help us track page count and page rotation rate so
| | | | | |       that we can monitor that the tscache remains healthy.
| | | | | |   
* | | | | |   commit 46f46522f2b157a213b074c0860e172441de630e
|\ \ \ \ \ \  Merge: 4824320c3 a2cd1e147
| | | | | | | Author: marc <marc.berhault@gmail.com>
| | | | | | | Date:   Thu Nov 30 16:13:29 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20358 from mberhault/marc/cliflags
| | | | | | |     
| | | | | | |     cli: use FlagInfo in CCL code.
| | | | | | | 
| * | | | | | commit a2cd1e1476b32d944b038dec5ad1fcb3083e7bc4
| | | | | | | Author: marc <marc@cockroachlabs.com>
| | | | | | | Date:   Thu Nov 30 14:27:45 2017 -0500
| | | | | | | 
| | | | | | |     cli: use FlagInfo in CCL code.
| | | | | | |     
| | | | | | |     * move some of the FlagInfo helpers to `cliflags`
| | | | | | |     * use a similar flag definition layout in cliccl
| | | | | | |     * add `StringSliceFlag`
| | | | | | |     * convert all ccl flags to `FlagInfo`
| | | | | | |     * use the `FlagInfo` name for the encryption flag
| | | | | | |     
| | | | | | |     I'm leaving the `cli.<type>Flag` functions in `cli` to avoid having to
| | | | | | |     prepend `cliflags.` to all of them.
| | | | | | |   
* | | | | | |   commit 4824320c37aa282ef41883944a995e34d0656256
|\ \ \ \ \ \ \  Merge: 2329e1d14 710a0f6fd
| |_|_|_|/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
|/| | | | | |   Date:   Thu Nov 30 16:10:59 2017 -0500
| | | | | | |   
| | | | | | |       Merge pull request #20288 from nvanbenschoten/nvanbenschoten/cleanupQuota
| | | | | | |       
| | | | | | |       storage: don't use quota pool throttling for NodeLiveness range
| | | | | | | 
| * | | | | | commit 710a0f6fd827f3369510c42878b896e32df96c5e
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Thu Nov 30 15:12:25 2017 -0500
| | | | | | | 
| | | | | | |     storage: don't use quota pool throttling for NodeLiveness range
| | | | | | |     
| | | | | | |     We don't want to throttle updates to the `NodeLiveness` range even
| | | | | | |     if a follower is falling behind because this could result in
| | | | | | |     cascading failures.
| | | | | | | 
| * | | | | | commit 88478eb4ef006f5aa41ccea651ad2e67c08afb35
| | |_|/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| |/| | | |   Date:   Mon Nov 27 14:28:53 2017 -0500
| | | | | |   
| | | | | |       storage: clean up quota pool logic for active replicas
| | | | | |       
| | | | | |       Follow up to #19819.
| | | | | |   
* | | | | |   commit dfc4c66a47b3459f8c083831fba1d4afa2c75f5d
|\ \ \ \ \ \  Merge: da27e442d ac0f8f33e
| |_|_|/ / /  Author: Nikhil Benesch <nikhil.benesch@gmail.com>
|/| | | | |   Date:   Thu Nov 30 15:44:52 2017 -0500
| | | | | |   
| | | | | |       Merge pull request #20246 from benesch/view-limit-order
| | | | | |       
| | | | | |       sql: support LIMIT and ORDER BY in views
| | | | | | 
| * | | | | commit ac0f8f33e380ba75961d6b9c8fb2bf3218bb58d6
| | | | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | | | Date:   Wed Nov 29 16:51:24 2017 -0500
| | | | | | 
| | | | | |     logictest: upcase some view definitions
| | | | | |     
| | | | | |     For consistency with other logic tests that use uppercase view
| | | | | |     definitions.
| | | | | | 
| * | | | | commit 568f810b2f664c38a663adc6973ef220b962aee0
|  / / / /  Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | |   Date:   Thu Nov 23 16:19:56 2017 -0500
| | | | |   
| | | | |       sql: support LIMIT and ORDER BY in views
| | | | |       
| | | | |       Previously, LIMIT and ORDER BY clauses in view definitions were silently
| | | | |       ignored. For example, given the following view
| | | | |       
| | | | |           CREATE VIEW v AS SELECT generate_series FROM generate_series(1, 100) ORDER BY 1 DESC LIMIT 10
| | | | |       
| | | | |       the following query would return 100 results in ascending order, instead
| | | | |       of the expected ten results in descending order:
| | | | |       
| | | | |           SELECT * FROM v
| | | | |       
| | | | |       This commit ensures the above query instead observes the ORDER BY and
| | | | |       LIMIT specified in the view definition, as Postgres does. As in
| | | | |       Postgres, additional ORDER BY and LIMIT clauses can be specified at query time
| | | | |       
| | | | |           SELECT * FROM v ORDER BY 1 ASC LIMIT 10
| | | | |       
| | | | |       in which case the effect is the same as if the view definition had been
| | | | |       specified inline as a subquery:
| | | | |       
| | | | |           SELECT *
| | | | |             FROM (SELECT generate_series FROM generate_series(1, 100) ORDER BY 1 DESC LIMIT 10)
| | | | |             ORDER BY 1 ASC LIMIT 10
| | | | |   
* | | | | | |   commit 4b42b3c14d73fef8733fc234b50a1eaf8b91c9f3
|\ \ \ \ \ \ \  Merge: e678539f3 0cf73ec2b
| |_|_|/ / / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
|/| | | | | |   Date:   Thu Nov 30 13:43:03 2017 -0500
| | | | | | |   
| | | | | | |       Merge pull request #20232 from nvanbenschoten/nvanbenschoten/sklImplCleanup
| | | | | | |       
| | | | | | |       tscache: switch to skiplist implementation by default, remove Request interface
| | | | | | | 
| * | | | | | commit 0cf73ec2b1af47a2c3f7bbd71e7c254ceb8129da
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Wed Nov 22 18:51:44 2017 -0500
| | | | | | | 
| | | | | | |     tscache: remove Request type and {Add,Expand}Request methods
| | | | | | |     
| | | | | | |     Now that all `tscache.Cache` implementations are thread safe and do nothing in
| | | | | | |     their `ExpandRequests` methods, we can remove the idea of a `tscache.Request`
| | | | | | |     entirely. We do this by removing the associated methods and type and instead
| | | | | | |     exporting the `Add` method to be used directly. This avoids all associated
| | | | | | |     overhead of the `Request` struct, which is a big improvement for `sklImpl`.
| | | | | | | 
| * | | | | | commit bd1d6adc02242f387f4725379e081c0f66f802ac
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Wed Nov 22 16:06:01 2017 -0500
| | | | | | | 
| | | | | | |     tscache: remove requests btree from treeImpl
| | | | | | |     
| | | | | | |     This change removes the `requests` btree from treeImpl. Since `treeImpl`
| | | | | | |     is no longer used by default, this optimization is no longer important.
| | | | | | |     
| | | | | | |     This will allow us to remove `tscache.Request` completely.
| | | | | | | 
| * | | | | | commit 5c98a8f32e0dabf22a257840f5089d091eabf368
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Wed Nov 22 13:20:48 2017 -0500
| | | | | | | 
| | | | | | |     tscache: remove Cache.ThreadSafe method entirely
| | | | | | |     
| | | | | | |     Because all `Cache` implementation must now be thread-safe, this change
| | | | | | |     removes the `ThreadSafe` method entirely. It then removes the `tsCacheMu`
| | | | | | |     lock on `Store`.
| | | | | | | 
| * | | | | | commit 213c44953dd981005603137da78433597e012419
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Wed Nov 22 12:53:29 2017 -0500
| | | | | | | 
| | | | | | |     tscache: make treeImpl thread-safe
| | | | | | |     
| | | | | | |     This change makes `treeImpl` thread-safe, as is reflected by its
| | | | | | |     `ThreadSafe` method. This makes the structure slightly less efficient
| | | | | | |     because it will result in locking at a finer-granularity. Now that it's
| | | | | | |     not used by default, that's no a huge deal.
| | | | | | | 
| * | | | | | commit bcbde0291fa1df8ace01241ed081429db04cda6a
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Wed Nov 22 12:33:47 2017 -0500
| | | | | | | 
| | | | | | |     tscache: use sklImpl as default tscache.Cache implementation
| | | | | | | 
| * | | | | | commit e29588c75df4f3fd3518407a4fb33a84e35c64e7
| | | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | | Date:   Thu Nov 23 23:34:04 2017 -0500
| | | | | | | 
| | | | | | |     util: only add preemption points in specific race tests
| | | | | | |     
| | | | | | |     The new `sklImpl` tscache structure includes a large number of
| | | | | | |     preemption points when race tested. This change adds a new
| | | | | | |     `util.EnableRacePreemptionPoints` function so that these preemption
| | | | | | |     points are only used in specific tests.
| | | | | | |   
* | | | | |   commit b6f4a57059201bf7186f8968805db30e92349932
|\ \ \ \ \ \  Merge: 1e4b34e28 054806984
| | | | | | | Author: RaduBerinde <radu@cockroachlabs.com>
| | | | | | | Date:   Thu Nov 30 12:44:48 2017 -0500
| | | | | | | 
| | | | | | |     Merge pull request #20350 from RaduBerinde/stat-name
| | | | | | |     
| | | | | | |     stats: plumb stat name
| | | | | | | 
| * | | | | | commit 05480698448e078b0322e44a90ef6cd37fe99e52
| | | | | | | Author: Radu Berinde <radu@cockroachlabs.com>
| | | | | | | Date:   Wed Nov 29 17:12:26 2017 -0500
| | | | | | | 
| | | | | | |     stats: plumb stat name
| | | | | | |     
| | | | | | |     Plumbing the name of the stat through the processors and planning code.
| | | | | | |   
* | | | | | |   commit 1e4b34e280385b3f2d64a6692be2804337764294
|\ \ \ \ \ \ \  Merge: 56685a6b6 98419a535
| |_|/ / / / /  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
|/| | | | | |   Date:   Thu Nov 30 12:30:32 2017 -0500
| | | | | | |   
| | | | | | |       Merge pull request #20329 from tschottdorf/refactor/cmd
| | | | | | |       
| | | | | | |       batcheval: absorb command registration
| | | | | | | 
| * | | | | | commit 98419a53554b96519545635d934054fd55b3570e
| | |_|_|/ /  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| |/| | | |   Date:   Thu Nov 30 11:27:54 2017 -0500
| | | | | |   
| | | | | |       batcheval: absorb command registration
| | | | | |       
| | | | | |       You'll notice that `EndTransaction` is still lingering in `replica_command.go`,
| | | | | |       but that will be done in a follow-up.
| | | | | |       
| | | | | |       Touches #18779.
| | | | | |   
* | | | |   commit 25ad9d4f650cf9bca2cd95f03d6a2e64c165ce6a
|\ \ \ \ \  Merge: f680e2623 4f6967288
| |_|/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
|/| | | |   Date:   Thu Nov 30 10:00:48 2017 -0500
| | | | |   
| | | | |       Merge pull request #20342 from nvanbenschoten/nvanbenschoten/canceled
| | | | |       
| | | | |       lint: 🇺🇸🇺🇸🇺🇸 use US locale in TestMisspell 🇺🇸🇺🇸🇺🇸
| | | | | 
| * | | | commit 4f6967288978bde3c11968301ae6cd7d2cc4a8df
| |/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | |   Date:   Wed Nov 29 21:59:56 2017 -0500
| | | |   
| | | |       lint: 🇺🇸🇺🇸🇺🇸 use US locale in TestMisspell 🇺🇸🇺🇸🇺🇸
| | | |       
| | | |       This change adds a US locality preference to `TestMisspell`. This causes
| | | |       `misspell` to add rules from its `DictAmerican` dictionary which catch
| | | |       british vs. american spelling differences like "cancelled" vs. "canceled"
| | | |       and "marshalled" vs. "marshaled".
| | | |       
| | | |       The bulk of this change was automated by running the command:
| | | |       ```
| | | |       git ls-files pkg/ | xargs ./bin/misspell -locale US -i "licence,analyse" -w
| | | |       ```
| | | |       
| | | |       After this change:
| | | |       ```
| | | |       Nathans-MBP:cockroach$ git grep -i cancelled | wc -l
| | | |              0
| | | |       ```
| | | |   
* | | |   commit f680e2623e691656185f1ef4f5f7fae71c3ad3f6
|\ \ \ \  Merge: e09462604 82debfb14
| | | | | Author: RaduBerinde <radu@cockroachlabs.com>
| | | | | Date:   Thu Nov 30 09:38:18 2017 -0500
| | | | | 
| | | | |     Merge pull request #20326 from RaduBerinde/stats-planning
| | | | |     
| | | | |     sql: CREATE STATISTICS statement and distsql planning for stats
| | | | | 
| * | | | commit 82debfb14e5c8f1c63ca6ebc8ef633f61a0ca8bc
| | | | | Author: Radu Berinde <radu@cockroachlabs.com>
| | | | | Date:   Wed Nov 29 13:31:18 2017 -0500
| | | | | 
| | | | |     stats: use unique_rowid for statisticsID
| | | | | 
| * | | | commit d8dc2931128e373750d492f5b6e4df03162b95b9
| | | | | Author: Radu Berinde <radu@cockroachlabs.com>
| | | | | Date:   Wed Nov 29 11:35:00 2017 -0500
| | | | | 
| | | | |     update hyperloglog dependency
| | | | |     
| | | | |     Updating to include a fix for spurious race detector errors.
| | | | | 
| * | | | commit 71b82dbe853b42a9dafe5ae93f80a3589e87959b
| | | | | Author: Radu Berinde <radu@cockroachlabs.com>
| | | | | Date:   Mon Nov 27 13:49:15 2017 -0500
| | | | | 
| | | | |     distsql: planning for statistics
| | | | |     
| | | | |     Sample plan:
| | | | |     https://cockroachdb.github.io/distsqlplan/decode.html?eJzElLGOozAQhvt7CvRXd5ILG0guR5Ur01xOyXUnCgePEFqCkW2k3Y149xVQZIMSs6soS8mYz98_GmtOqLSiP_JIFsl_CDCEYIjAEINhgZShNjoja7XpfhmAjXpGwhmKqm5cV04ZMm0IyQmucCUhwT95KGlHUpEBgyIni7KX1KY4SvOyVtJJMGwblwRrgbRl0I07X2idzAmJaNnHpXt5rMuxcCjui1dKAsE552DYO-nXhje1Z1tTaaPIkLrwpe3NYL_z3FAunf5EwuB7yHlwaLIncvbHzbzRRV4xx2wmpI-aTThHrxPSR_UazdHrhPRRvcZz9Doh_Yr9ckW7I1vrytJoz1y_mXf7h1ROw7KyujEZ_TU66zXD57bn-oIi64ZTMXxsquGoC_geFl44vIDFGA795gl15KVjPxzfk3vhhZd-8_Ie808vvPKbV_eYf_lnxSeeif-Rjd1p--0tAAD__0dOyfM=
| | | | | 
| * | | | commit 962fc7f75b581e9cf7a36ebec3e6a43174a49526
| | | | | Author: Radu Berinde <radu@cockroachlabs.com>
| | | | | Date:   Mon Nov 27 16:40:11 2017 -0500
| | | | | 
| | | | |     sql: add CREATE STATISTICS statement
| | | | |   
* | | | |   commit e09462604583b8fd0c6890a12433a64f617fa3c1
|\ \ \ \ \  Merge: 49ec6e5e9 59bd094db
| |_|/ / /  Author: Jordan Lewis <jordanthelewis@gmail.com>
|/| | | |   Date:   Wed Nov 29 23:06:29 2017 -0500
| | | | |   
| | | | |       Merge pull request #20335 from jordanlewis/dry-cmp-ops
| | | | |       
| | | | |       sql: DRY the declaration of comparison operators
| | | | | 
| * | | | commit 59bd094dbced2374466dd96b334a8d135e07294e
| |/ / /  Author: Jordan Lewis <jordanthelewis@gmail.com>
| | | |   Date:   Wed Nov 29 16:38:23 2017 -0500
| | | |   
| | | |       sql: DRY the declaration of comparison operators
| | | |       
| | | |       Also, sort them alphabetically within their sections.
| | | |   
* | | |   commit 1ea9c20ae771fb736fa38f698895fbdd2d2e9791
|\ \ \ \  Merge: 566aaf5a8 cede579ad
| |_|/ /  Author: Joey Pereira <legojoey17@gmail.com>
|/| | |   Date:   Wed Nov 29 16:46:48 2017 -0500
| | | |   
| | | |       Merge pull request #20177 from lego/scrub-mergejoin
| | | |       
| | | |       sql: Improve scrub index check performance
| | | | 
| * | | commit cede579adad83ca6deca226e46b1b65543ca58d8
| | | | Author: Joey Pereira <joey@pereira.io>
| | | | Date:   Wed Nov 29 12:05:16 2017 -0500
| | | | 
| | | |     sql: Fix scrub error for tables with implicit primary key
| | | |     
| | | |     After changing the scrub index check to use a sorted merge join, it
| | | |     failed to handle tables with an implicit primary key (i.e. a hidden
| | | |     column `rowid`).
| | | |     
| | | |     This fixes the bug and adds a test to ensure scrub checks in general
| | | |     will not fail when executing on such a table.
| | | | 
| * | | commit 6336d9ae56f06ec62c0ff901c986bc1a6ea75377
| | | | Author: Joey Pereira <joey@pereira.io>
| | | | Date:   Thu Nov 16 13:56:07 2017 -0500
| | | | 
| | | |     sql: Change scrub index check to use distSQL merge join
| | | |     
| | | |     This significantly improves the performance of the check after #19889.
| | | |     After adding the condition to filter false positives, the check was
| | | |     running an expensive and slow hash join.
| | | |     
| | | |     After this patch, the scan and check uses a sort and merge join and is
| | | |     much faster. In order to do this, we needed to leverage outer joins in
| | | |     the distSQL merge joiner and the capability to have `NULL = NULL` be
| | | |     truth-y.
| | | |     
| | | |     Roughly estimating, prior to the patch a check of 5000 rows took >10s.
| | | |     After the patch this is well under 1s.
| | | | 
| * | | commit 1f2b5c3dfec4582f695800e2d825c3a951f52b50
| | | | Author: Joey Pereira <joey@pereira.io>
| | | | Date:   Thu Nov 16 12:20:04 2017 -0500
| | | | 
| | | |     distsqlrun: Plumb NullEquality into MergeJoiner
| | | |     
| | | |     This allows a toggle to change the behaviour of the MergeJoiner to
| | | |     make NULL=NULL to be truth-y.
| | | | 
| * | | commit c77e53e06f4f3a1af8debb60a1585822e141645e
|/ / /  Author: Joey Pereira <joey@pereira.io>
| | |   Date:   Thu Nov 16 14:46:14 2017 -0500
| | |   
| | |       sql: Plan MergeJoiner for OUTER joins.
| | |       
| | |       Currently, everything was already completed for executing OUTER joins
| | |       using the distSQL sorted merge joiner. This change now plans them.
| | |   
* | | |   commit 9d071f1d2b196536d636b452b42b5ba161dc4fc5
|\ \ \ \  Merge: 1333b16b9 cf3270475
| |/ / /  Author: Alex Robinson <alexdwanerobinson@gmail.com>
|/| | |   Date:   Wed Nov 29 15:59:20 2017 -0500
| | | |   
| | | |       Merge pull request #20331 from a-robinson/docker-unstable
| | | |       
| | | |       build: Publish a separate docker image for unstable releases
| | | | 
| * | | commit cf3270475c0db1f37680c26d89a894051c79332a
| | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | Date:   Wed Nov 29 15:03:46 2017 -0500
| | | | 
| | | |     build: Publish a separate docker image for unstable releases
| | | |     
| | | |     Tested locally to push out an image for our latest alpha release. All
| | | |     unstable images get published to
| | | |     https://hub.docker.com/r/cockroachdb/cockroach-unstable/
| | | |   
* | |   commit f292bed21360f1c90fa3710111d1bf1773ba23d3
|\ \ \  Merge: d7de7099c e635ac7dd
| |_|/  Author: marc <marc.berhault@gmail.com>
|/| |   Date:   Wed Nov 29 08:50:32 2017 -0500
| | |   
| | |       Merge pull request #20252 from mberhault/marc/cleanup_privileges
| | |       
| | |       sql: cleanup privileges a bit
| | | 
| * | commit e635ac7dd88f3639dc0c5917c2bd1d57f79a61fc
|/ /  Author: marc <marc@cockroachlabs.com>
| |   Date:   Fri Nov 24 08:18:09 2017 -0500
| |   
| |       sql: cleanup privilege interfaces
| |       
| |       * use ID names, privilege helper for system tables.
| |       * make privilege checks methods on planner
| |       * use privilege check methods instead of manual `RootUser` checks
| |   
* | | |   commit 4b5e546152850e5275b4427a897fdc84cb5152c7
|\ \ \ \  Merge: 604ab8f65 832c1e47d
| | | | | Author: kena <knz@users.noreply.github.com>
| | | | | Date:   Tue Nov 28 19:25:24 2017 +0100
| | | | | 
| | | | |     Merge pull request #20229 from knz/20171122-split-run
| | | | |     
| | | | |     sql: extract the run-time state from planNodes
| | | | | 
| * | | | commit 832c1e47d4b8aa8beea99754a8865bf3ec1c78b5
| | | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | | Date:   Wed Nov 22 21:42:43 2017 +0100
| | | | | 
| | | | |     sql: extract the run-time state from planNodes
| | | | |     
| | | | |     The new "Run" structs contain the fields from each
| | | | |     planNodes that are either:
| | | | |     
| | | | |     - written to during local execution, and thus need to be reset
| | | | |       if local execution takes place multiple times (e.g.
| | | | |       if we reuse the planNode).
| | | | |     - only read from during local execution, i.e. they are
| | | | |       initialized once during logical planning and never read again
| | | | |       until local execution takes place.
| | | | |     
| | | | |     They expose the following invariants:
| | | | |     
| | | | |     - every field *not* in a Run struct should be considered immutable
| | | | |       after planning has completed,
| | | | |     
| | | | |     - every field in a Run struct needs not be initialized if local
| | | | |       execution is known not to take place.
| | | | |     
| | | | |     This patch places some fields in the Run struct that are immutable
| | | | |     across executions (after planning has completed). It is not the case
| | | | |     that *all* Run fields must be re-initialized on every execution.
| | | | |     
| | | | |     The fields that are allowed in the Run struct and do not remain in the
| | | | |     main planNode are those that are known to be produced during planning
| | | | |     but not *used* during planning.
| | | | |     
| | | | |     This patch *may* have kept some immutable fields in the main planNode
| | | | |     that are not used during planning (and only used during execution).
| | | | |     Subsequent patches can strive to migrate more fields from the planNode
| | | | |     to the Run struct if possible.
| | | | |   
* | | | | |   commit 7d9974d34da7821bb71994761f32120653ae86ae
|\ \ \ \ \ \  Merge: ba09b7933 d48bbe5af
| |_|_|_|/ /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
|/| | | | |   Date:   Tue Nov 28 11:00:02 2017 -0500
| | | | | |   
| | | | | |       Merge pull request #20300 from nvanbenschoten/nvanbenschoten/reuseSkl
| | | | | |       
| | | | | |       tscache: re-use memory arena when cycling intervalSkl pages
| | | | | | 
| * | | | | commit d48bbe5afdbe5ffc72b051ab4795e78b5879d71c
| | | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | | Date:   Tue Nov 28 10:03:45 2017 -0500
| | | | | | 
| | | | | |     tscache: remove unnecessary atomic operations
| | | | | |     
| | | | | |     These atomics are already protected by a write lock.
| | | | | | 
| * | | | | commit d30aa98e6a209cc68906061b950bfd594cb5108d
|  / / / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | |   Date:   Tue Nov 28 08:12:47 2017 -0500
| | | | |   
| | | | |       tscache: re-use memory arena when cycling intervalSkl pages
| | | | |       
| | | | |       The `intervalSkl` structure used by `sklImpl` uses 32 MB page sizes.
| | | | |       These are large enough to warrant special care to prevent from
| | | | |       re-allocating them continuously. This change re-uses arenas during
| | | | |       page rotation whenever possible (which is most of the time).
| | | | |   
* | | |   commit ac51bb85589b5baeaf6307cb7df8c71a6e40aba0
|\ \ \ \  Merge: d6756190e ebe5f376d
| |_|/ /  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
|/| | |   Date:   Mon Nov 27 22:49:28 2017 -0500
| | | |   
| | | |       Merge pull request #19634 from dsaveliev/19523-startup-sequence-interceptors
| | | |       
| | | |       server: implement the startup sequence interceptors
| | | | 
| * | | commit ebe5f376d1622029256ebdd33eab81e63b3be576
|/ / /  Author: Dmitry Saveliev <d.e.saveliev@gmail.com>
| | |   Date:   Mon Oct 30 13:48:50 2017 +0400
| | |   
| | |       server: simplify startup listener handling
| | |       
| | |       Rework and simplify the way we bootstrap our listeners by introducing a
| | |       `serveMode` which is either "initializing" or "operational". In the latter,
| | |       everything is allowed -- the server is fully functional. In the former, the
| | |       server is either waiting for an Init command or busy bringing its moving parts
| | |       into motion: only a select subset of "safe" grpc endpoints are available
| | |       (gossip, init, ping) and connections opened against the Postgres listener are
| | |       eagerly closed (a TODO in the code hints at a straightforward UX improvement
| | |       here).
| | |       
| | |       Fortified the init server to prevent overlapping init requests, and made its
| | |       resilient to double-initialization (which in turn simplified the listener
| | |       handling). As a side effect, patched up small holes during which services would
| | |       not be available.
| | |   
* | |   commit d6756190e40df9fbc30c124022689baf341ec7f4
|\ \ \  Merge: f5e5af60a 0921e3df0
| | | | Author: kena <knz@users.noreply.github.com>
| | | | Date:   Mon Nov 27 23:28:06 2017 +0100
| | | | 
| | | |     Merge pull request #20093 from knz/20171116-fix-show-trace
| | | |     
| | | |     sql: SHOW TRACE: misc improvements
| | | | 
| * | | commit 0921e3df048c64c3f805185ec687305b8ce7ce87
| | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | Date:   Mon Nov 27 17:17:45 2017 +0100
| | | | 
| | | |     sql: new variant SHOW COMPACT TRACE for common cases
| | | |     
| | | |     Requested by @andreimatei: in most cases SHOW TRACE is used to
| | | |     investigate the duration of each operation and the message. Also the
| | | |     file/lineno location as prefix of the message is desired in this case.
| | | |     
| | | |     To respond to this common case, this patch introduces the variant
| | | |     `SHOW COMPACT [KV] TRACE` which only renders the age, message, tag and
| | | |     operation columns. The file/lineno information is prepended to the
| | | |     message.
| | | |     
| | | |     For example:
| | | |     
| | | |     ```sql
| | | |     > SHOW COMPACT TRACE FOR SESSION;
| | | |     +----------------+-----------------------------------------------------------------------------------------+---------------------------------------+------------------+
| | | |     |      age       |                                                 message                                 |                         tag           |     operation    |
| | | |     +----------------+-----------------------------------------------------------------------------------------+---------------------------------------+------------------+
| | | |     | 0s             | === SPAN START: sql txn implicit ===                                                    |                                       | sql txn implicit |
| | | |     | 174µs525ns     | === SPAN START: sql txn ===                                                             |                                       | sql txn          |
| | | |     | 190µs114ns     | sql/executor.go:1184 executing 1/6: BEGIN TRANSACTION                                   | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| | | |     | 225µs320ns     | sql/executor.go:1184 executing 2/6: SELECT 1                                            | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| | | |     | 269µs362ns     | sql/executor.go:2099 query not supported for distSQL: unsupported node *sql.unaryNode   | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| | | |     | 298µs987ns     | sql/executor.go:1184 executing 3/6: COMMIT TRANSACTION                                  | [client=127.0.0.1:46367,user=root,n1] | sql txn          |
| | | |     ...
| | | |     ```
| | | | 
| * | | commit d6b73683161f7bdcf993fc903ad1d28da28b4a11
|  / /  Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | |   Date:   Thu Nov 16 13:39:06 2017 +0100
| | |   
| | |       sql: split the tag and filename/lineno information in tracing output
| | |       
| | |       Prior to this patch, the filename/lineno information in tracing output
| | |       wasn't properly recognized, in turn causing the tag not to be split
| | |       off properly. Also, the extraction was performed using the SQL
| | |       function `regexp_replace`, causing some RE lookup overhead.
| | |       
| | |       This patch fixes the issues by extracing the filename/lineno in a
| | |       separate column.  This adds new columns `tag` and `loc` in
| | |       `crdb_internal.session_trace`.  The extraction is done in Go, with a
| | |       well-identified, pre-compiled RE.
| | |       
| | |       It also exposes the `loc` column in the output of SHOW TRACE.  The
| | |       `context` column is renamed to `tag`, since this name more closely
| | |       mirrors the terminology used in the source code, and users looking at
| | |       the output of SHOW TRACE will usually also look at the source code
| | |       too.
| | |   
* |   commit 4ebabadc7db413cc3164cc878f8efc0414b0e3ff
|\ \  Merge: 4c0c5997e 60fe5ad05
| | | Author: marc <marc.berhault@gmail.com>
| | | Date:   Mon Nov 27 15:00:44 2017 -0500
| | | 
| | |     Merge pull request #20255 from mberhault/marc/fix_dump
| | |     
| | |     cli: dump: fix infinite loop in when resolving dependencies.
| | | 
| * | commit 60fe5ad05b86e09be403ebd2dbe8087c134556fa
| | | Author: marc <marc@cockroachlabs.com>
| | | Date:   Fri Nov 24 16:03:05 2017 -0500
| | | 
| | |     cli: dump: fix infinite loop in when resolving dependencies.
| | |     
| | |     While this will avoid the loop and keep dependency order, there may
| | |     still be problems with dependency cycles.
| | |     
| | |     This came up in #20254 when encountering a foreign key referring to the
| | |     same table.
| | |     
| | |     Before the fix, the test fails with:
| | |     ```
| | |     runtime: goroutine stack exceeds 1000000000-byte limit
| | |     fatal error: stack overflow
| | |     ```
| | |   
```

mberhault:
Just one from me:
cli: dump: fix crash on dependency cycles [#20255]

All others are cleanups/refactors/work for future features.

tschottdorf:
Nothing from me (checked my box).

a-robinson:
Sorry, @jseldess! Hopefully this'll get easier now that there's a lint warning in place to help us remember to include these.

The additional only release note that may be needed from my commits is:

(General Change) Alpha and beta releases are [now published as Docker images under the name `cockroachdb/cockroach-unstable`](https://hub.docker.com/r/cockroachdb/cockroach-unstable/)  #20331

nvanbenschoten:
(Performance improvement) replace interval tree structure in CommandQueue with arena-backed concurrent skiplist. Reduces global locking and GC pressure, improving average and tail latencies.

(Bug fix) prevent context cancellation during lease acquisition from leaking to coalesced requests.

petermattis:
@nvanbenschoten Did you mean `s/CommandQueue/TimestampCache/g`?

nvanbenschoten:
... wow yeah, but one can dream!

RaduBerinde:
None required for any of my commits.

knz:
Release note (sql): the SHOW [KV] TRACE statement now properly extracts file/line number information when analyzing traces produced in debug mode. A new SHOW COMPACT [KV] TRACE statement is also provided to provide a more compact view on the same data. #20093 

Release note (sql): some queries using IS NOT NULL conditions are now better optimized. #20436

benesch:
Release note (sql change): Views now support LIMIT and ORDER BY (#20246).
Release note (admin ui change): The Admin UI now sets the Last-Modified header when serving assets to permit browser caching. This improves page load times, especially on slow connections (#20429).

jseldess:
@justinj, checking you off, since the commit concerns JSON, which we're not ready to document.

jseldess:
@m-schneider, checking you off, too, since your commit also concerns JSON, I believe.

sploiselle:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[releases/v1.2-alpha.20171211.md, line 41 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2266#-L06JC2om6VJMSluB974:-L06JC2om6VJMSluB975:b-ri4jti) ([raw file](https://github.com/cockroachdb/docs/blob/276402ffd180254a7aa1007dc21560254d155b40/releases/v1.2-alpha.20171211.md#L41)):*
> ```Markdown
> - The protocol statement tag for [`CREATE TABLE ... AS ...`](../v1.2/create-table-as.html) is now [`SELECT`](../v1.2/select.html), like in PostgreSQL. [#20268](https://github.com/cockroachdb/cockroach/pull/20268)
> - OIDs can now be compared with inequality operators. [#20367](https://github.com/cockroachdb/cockroach/pull/20367)
> - The [`CANCEL JOB`](../v1.2/cancel-job.html) statement now supports cancelling [`IMPORT`](../v1.2/import.html) jobs. [#20343](https://github.com/cockroachdb/cockroach/pull/20343)
> ```

s/cancelling/canceling

---

*[releases/v1.2-alpha.20171211.md, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2266#-L06Kfufgv67bhg41Jqi:-L06Kfufgv67bhg41Jqj:bkqyr4a) ([raw file](https://github.com/cockroachdb/docs/blob/276402ffd180254a7aa1007dc21560254d155b40/releases/v1.2-alpha.20171211.md#L64)):*
> ```Markdown
> - Added missing Distributed SQL flows to the exported `sql.distsql.flows.active` and `sql.distsql.flows.total` metrics and the "Active Flows for Distributed SQL Queries" admin UI graph. [#20503](https://github.com/cockroachdb/cockroach/pull/20503)
> - Fixed an issue with stale buffer data when using the binary format for [`ARRAY`](../v1.2/array.html) values. [#20461](https://github.com/cockroachdb/cockroach/pull/20461)
> - The [`cockroach sql`](../v1.2/use-the-built-in-sql-client.html) shell now better reports the number of rows inserted by a [`CREATE TABLE ... AS ...`](../v1.2/create-table-as.html) statement. Note,  however, that the result are still formatted incorrectly if the `CREATE TABLE ... AS ...` statement create zero rows in the new table. [#20268](https://github.com/cockroachdb/cockroach/pull/20268)
> ```

s/create/creates/

---

*[releases/v1.2-alpha.20171211.md, line 78 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2266#-L06LQExUMItchdXf4v8:-L06LQExUMItchdXf4v9:b9dm6s2) ([raw file](https://github.com/cockroachdb/docs/blob/276402ffd180254a7aa1007dc21560254d155b40/releases/v1.2-alpha.20171211.md#L78)):*
> ```Markdown
> - Improved p99 latencies for garbage collection of previous versions of a key, when there are many versions. [#20373](https://github.com/cockroachdb/cockroach/pull/20373)
> - Smoothed out disk usage under very write heavy workloads by syncing to disk more frequently. [#20352](https://github.com/cockroachdb/cockroach/pull/20352)
> - Improved garbage collection of very large [transactions](../v1.2/transactions.html) and large volumes of abandoned writes (intents). [#20396](https://github.com/cockroachdb/cockroach/pull/20396)
> ```

s/writes (intents)/write intents/

or...something like that "writes intents" is odd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2266)*

PR #2271, opened 2017-12-11
Update secure Kubernetes instructions with RBAC support

a-robinson:
This enables them to work on more clusters (including non-GKE clusters)
by default.

Documents the changes in https://github.com/cockroachdb/cockroach/pull/20116. I'll coordinate their merging to avoid any confusion caused by the docs and config files not matching up.

### DIFF
--- a/_includes/orchestration/start-cluster.md
+++ b/_includes/orchestration/start-cluster.md
@@ -8,6 +8,11 @@ $ kubectl create -f https://raw.githubusercontent.com/cockroachdb/cockroach/mast
 ~~~
 
 ~~~
+serviceaccount "cockroachdb" created
+role "cockroachdb" created
+clusterrole "cockroachdb" created
+rolebinding "cockroachdb" created
+clusterrolebinding "cockroachdb" created
 service "cockroachdb-public" created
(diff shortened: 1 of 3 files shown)


a-robinson:
I saw this before refreshing the page on the latest teamcity build -- I assume it's just an effect of how these temporary builds work, but let me know if it isn't. Things looked fine after the page refresh.

screen shot 2017-12-11 at 2 27 38 pm

a-robinson:
More importantly, I'm stumped on why this isn't formatting properly. Any ideas?

screen shot 2017-12-11 at 2 39 33 pm

a-robinson:
@jseldess ping

jseldess:
Sorry for the delay, @a-robinson. Looking at this now.

jseldess:
@a-robinson, the `{% include copy-clipboard.html %}`s are causing that rendering issue. For some reason, that liquid tag doesn't behave when nested within `<div>` tags, which we're using to filter based on environment. In this case, just remove the include snippets.

jseldess:
Superseded by https://github.com/cockroachdb/docs/pull/2281.

PR #2274, opened 2017-12-11
releases: Correct reference to timestamp cache

bdarnell:
cc @nvanbenschoten


nvanbenschoten:
LGTM, thanks for catching this. I didn't realize how efficient @jseldess was and thought I had been quick enough to get my edit in in time.

PR #2281, opened 2017-12-13
Update secure Kubernetes instructions with RBAC support

jseldess:
This enables them to work on more clusters (including non-GKE clusters)
by default.

This PR replaces https://github.com/cockroachdb/docs/pull/2271.

### DIFF
--- a/_includes/orchestration/start-kubernetes.md
+++ b/_includes/orchestration/start-kubernetes.md
@@ -26,6 +26,31 @@ Choose whether you want to orchestrate CockroachDB with Kubernetes using the hos
 
     The process can take a few minutes, so don't move on to the next step until you see a `Creating cluster cockroachdb...done` message and details about your cluster.
 
+{% if page.secure == true %}
+
+3. Get the email address associated with your Google Cloud account:
+
+    ~~~ shell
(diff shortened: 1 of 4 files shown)


_includes/orchestration/start-kubernetes.md
```diff
@@ -26,6 +26,31 @@ Choose whether you want to orchestrate CockroachDB with Kubernetes using the hos
 
     The process can take a few minutes, so don't move on to the next step until you see a `Creating cluster cockroachdb...done` message and details about your cluster.
 
+{% if page.secure == true %}
```
jseldess: @a-robinson, I added this filter to make sure these RBAC steps don't get pulled into the insecure tutorials.

jseldess: Also made sense to move this content into this `include` file. But it's only showing up for `Hosted GKE`. Is RBAC relevant for the other methods as well?

a-robinson: RBAC in general is relevant for other methods. These two steps below are a workaround for GKE limitations and aren't relevant for any other environments.

Issue #2014, opened 2017-10-17
SQL sequences


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwINAgXA5on9pr2T/recpKacCHw2xQwqzn

vilterp:
Hi Jesse, getting started on implementing this (it will take a few PRs): here's a tracking issue so you can see the progress https://github.com/cockroachdb/cockroach/issues/19723 (it also links to the RFC). Let me know how I can help documenting this once it gets a little further on.

jseldess:
Thanks for the heads up, @vilterp.

jseldess:
@vilterp, `SEQUENCES` is going to be Lauren's starter docs project! She's going to be reaching out sometime next week for an intro.

vilterp:
Great! There are definitely plenty of little details to document — most of which are similar to Postgres at least. Their docs are here as a reference:

- `CREATE SEQUENCE`: https://www.postgresql.org/docs/current/static/sql-createsequence.html
- sequence functions: https://www.postgresql.org/docs/10/static/functions-sequence.html

lhirata:
closing this issue - new issues will be opened for remaining sequence docs.

PR #2292, opened 2017-12-14
docs for sequence SQL statements

lhirata:


v1.2/alter-sequence.md
```diff
@@ -4,13 +4,15 @@ summary:
...
 ---
 
-<span class="version-tag">New in v1.2:</span> The `ALTER SEQUENCE` [statement](sql-statements.html)...
+<span class="version-tag">New in v2.0:</span> The `ALTER SEQUENCE` [statement](sql-statements.html) applies a schema change to a sequence.
+
+{{site.data.alerts.callout_info}}To understand how CockroachDB changes schema elements without requiring table locking or other user-visible downtime, see <a href="https://www.cockroachlabs.com/blog/how-online-schema-changes-are-possible-in-cockroachdb/">Online Schema Changes in CockroachDB</a>.{{site.data.alerts.end}}
 
 <div id="toc"></div>
 
 ## Required Privileges
```
lhirata: How do you check for privileges?

v1.2/alter-sequence.md
```diff
@@ -34,11 +36,50 @@ table td:first-child {
 
  Parameter | Description
 -----------|------------
+`IF EXISTS` | Modify the sequence only if it exists; if it does not exist, do not return an error.
+`name` | The name of the sequence you want to modify.
+`INCREMENT` | The new value by which the sequence is incremented. A negative number creates a descending sequence. A positive number creates an ascending sequence.
+`MINVALUE` | The new minimum value of the sequence.
+`MAXVALUE` | The new maximum value of the sequence.
+`START` | The new starting value of the sequence.
```
lhirata: Doesn't work yet

v1.2/create-sequence.md
```diff
@@ -33,21 +35,92 @@ table td:first-child {
...
-`MINVALUE` |
-`MAXVALUE` |
-`START` |
-`CYCLE` |
+`seqname` | The name of the sequence to create, which must be unique within its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). When the parent database is not set as the default, the name must be formatted as `database.name`.
+`INCREMENT` | The value by which the sequence is incremented. A negative number creates a descending sequence. A positive number creates an ascending sequence.<br><br>**Default:** `1`
+`MINVALUE` | The minimum value of the sequence. Default values apply if not specified or if you enter `NO MINVALUE`.<br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `MININT`
+`MAXVALUE` | The maximum value of the sequence. Default values apply if not specified or if you enter `NO MAXVALUE`.<br><br>**Default for ascending:** `MAXINT` <br><br>**Default for descending:** `-1`
+`START` | The first value of the sequence. <br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `-1`
+`CYCLE` | The sequence will wrap around when the sequence value hits the maximum or minimum value. If `NO CYCLE` is set, the sequence will not wrap.
```
lhirata: What's the default for CYCLE?

v1.2/create-sequence.md
```diff
@@ -33,21 +35,92 @@ table td:first-child {
...
+  customer string,
+  address string
+  );
+~~~
+
+Insert a few records to see the sequence.
+
+{% include copy-clipboard.html %}
+~~~ sql
+> INSERT INTO customer_list (customer, address) VALUES ('Lauren', '123 Main Street'), ('Jesse', '456 Broad Ave'), ('Amruta', '9876 Green Parkway');
```
lhirata: Should this code block be formatted so there's no horizontal scrolling? If so, where should the line breaks be?

v1.2/drop-sequence.md
```diff
@@ -4,13 +4,13 @@ summary:
 toc: false
 ---
 
-<span class="version-tag">New in v1.2:</span> The `DROP SEQUENCE` [statement](sql-statements.html)...
+<span class="version-tag">New in v2.0:</span> The `DROP SEQUENCE` [statement](sql-statements.html) removes a sequence from a database.
```
lhirata: Currently, not working correctly.

v1.2/drop-sequence.md
```diff
@@ -28,11 +28,39 @@ table td:first-child {
 
  Parameter | Description
 -----------|------------
+`IF EXISTS` |  Drop the sequence only if it exists; if it does not exist, do not return an error.
+`sequenceName` | The name of the sequence you want to drop. Find the sequence name with `SHOW CREATE TABLE` on the table that uses the sequence.
+`CASCADE` | Drop all objects (such as [constraints](constraints.html) and tables) that depend on the sequence.<br><br>`CASCADE` does not list objects it drops, so should be used cautiously.
```
lhirata: Will CASCADE drop the table(s) that use the sequence?

v1.2/drop-sequence.md
```diff
@@ -28,11 +28,39 @@ table td:first-child {
...
+`RESTRICT` | _(Default)_ Do not drop the sequence if any objects (such as [constraints](constraints.html) and tables) depend on it.
 
 ## Examples
```
lhirata: Is there a way to view all of the sequences in a database?

v1.2/drop-sequence.md
```diff
@@ -4,13 +4,13 @@ summary:
...
+<span class="version-tag">New in v2.0:</span> The `DROP SEQUENCE` [statement](sql-statements.html) removes a sequence from a database.
 
 <div id="toc"></div>
 
 ## Required Privileges
 
-The user must have the...
+The user must have the `CREATE` [privilege](privileges.html) on the parent database.
 
 ## Synopsis
```
lhirata: Parameter names are inconsistent (e.g., seqname vs. sequenceName vs. name). Do we care / should we standardize across the sequence docs?

jseldess:
Congrats on opening this PR, @lhirata! I'll hold off on reviewing until you'd done a round with @vilterp. Quick tip, though: If you are asking a question of someone in particular, you should probably `@` them in the comment so they get a notification.

v1.2/create-sequence.md
vilterp: false.

v1.2/create-sequence.md
vilterp: Maybe

```
INSERT INTO customer_list (customer, address)
VALUES
  ('Lauren', '123 Main Street'),
  ('Jesse', '456 Broad Ave'),
  ('Amruta', '9876 Green Parkway');
```

v1.2/drop-sequence.md
vilterp: `SELECT * FROM pg_catalog.pg_sequence` or `SELECT * FROM information_schema.sequences`

v1.2/alter-sequence.md
vilterp: This doesn't set the sequence value, but it sets:

- the value that it goes to if you do `ALTER SEQUENCE foo RESTART` (not implemented yet)
- the value that it will wrap around to if you hit the `MAXVALUE` and `CYCLE` is on. This should work. E.g.:

```
CREATE SEQUENCE foo MAXVALUE 5 CYCLE
=> CREATE SEQUENCE
... increment up to 5 ...
ALTER SEQUENCE foo START -1000
=> ALTER SEQUENCE
SELECT nextval('foo')
=> 5
```

If we hadn't done the alter sequence, it would have cycled back to 1 (the default start)

v1.2/alter-sequence.md
vilterp: Not quite sure… You can do `SELECT * FROM information_schema.table_privileges`, and see the schema there alongside with other tables.

I guess things should be tested here like "you can't call nextval() on a sequence if you don't have the UPDATE privilege on it"… Which probably fails now. That's something I should fix and put in the automated tests though… Sigh

v1.2/drop-sequence.md
vilterp: They seem to be standardized in [functions.md](https://github.com/cockroachdb/cockroach/blob/0832bc5ed19e0654d131400e7a7fe9c82f1d5221/docs/generated/sql/functions.md#sequence-functions) as `sequence_name`; not sure where they're inconsistent

v1.2/create-sequence.md
```diff
@@ -33,21 +35,92 @@ table td:first-child {
...
+
+In this example, we create a table using the sequence we created in the first example as the table's primary key.
+
+{% include copy-clipboard.html %}
+~~~ sql
+> CREATE TABLE customer_list (
+  id INT PRIMARY KEY DEFAULT nextval('customer_seq'),
+  customer string,
+  address string
+  );
```
vilterp: nit: indent the columns:

```sql
CREATE TABLE customer_list (
  id INT PRIMARY KEY DEFAULT nextval('customer_seq'),
  customer string,
  address string
);
```

v1.2/create-sequence.md
```diff
@@ -33,21 +35,92 @@ table td:first-child {
...
+|  2 | Jesse    | 456 Broad Ave      |
+|  3 | Amruta   | 9876 Green Parkway |
++----+----------+--------------------+
+~~~
+
+### View the Current Value of a Sequence
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT currval('customer_seq');
```
vilterp: Not-so-fun fact: this doesn't just straightforwardly return the current value, as one would expect. It returns the last value that _you got for this sequence in the current session_. This means that:

- If you fire up a `cockroach sql` and run `currval` before running any `nextval`s, it errors out: `currval of sequence "seq" is not yet defined in this session`
- if you call `currval` and see 5, then other users of the DB increment it to 10, and then you call `currval` again, you'll still see 5

Why doesn't it just get the current value? Because we have to mimic PG. Why does PG do this? I don't know. How do you actually get the current value? `SELECT * FROM my_sequence`, which isn't implemented yet.

v1.2/create-sequence.md
```diff
@@ -4,13 +4,15 @@ summary:
 toc: false
 ---
 
-<span class="version-tag">New in v1.2:</span> The `CREATE SEQUENCE` [statement](sql-statements.html)...
+<span class="version-tag">New in v2.0:</span> The `CREATE SEQUENCE` [statement](sql-statements.html) creates a new sequence in a database. Use a sequence to auto-increment integers in a table.
+
+{{site.data.alerts.callout_info}}Using sequences may slow down your application. Use the serial data type unless a sequence is preferred or required.{{site.data.alerts.end}}
```
vilterp: nit: maybe add "may slow down your application _because it does a write to persistent storage_" or similar, just to be clear.

vilterp: another important gotcha that should be called out: sequences can leave gaps. Users might expect that when they use a sequence they'll never have gaps (i.e. they'll see `1, 2, 3, …` for that column in their table) but there actually can be gaps if you roll back a transaction that uses a sequence. E.g.:

```
INSERT INTO my_sequence_using_table VALUES ('foo', 'bar');
=> gets id 1
BEGIN;
INSERT INTO my_sequence_using_table VALUES ('bloop', 'baz');
=> gets id 2
ROLLBACK;
-- at this point even though the row isn't inserted, id 2 is still considered taken.
-- the `ROLLBACK` doesn't "put it back" in the sequence
INSERT INTO my_sequence_using_table VALUES ('beep', 'blorp');
=> gets id 3
-- now we have a gap: 1 => 3
```

This is done for performance reasons, and matches PG. PG docs here: https://www.postgresql.org/docs/10/static/functions-sequence.html, specifically "To avoid blocking concurrent transactions that obtain numbers from the same sequence, a nextval operation is never rolled back; that is, once a value has been fetched it is considered used and will not be returned again. [...]"

v1.2/drop-sequence.md
vilterp: Apparently yes; that's what PG does. (Of course, I haven't implemented this yet)

vilterp:
Looks good overall; really excited to see this!

v1.2/create-sequence.md
lhirata: added.

v1.2/create-sequence.md
lhirata: added.

lhirata: added. @jseldess - i've included two notes here back to back. kosher?

v1.2/drop-sequence.md
lhirata: added to example.

v1.2/create-sequence.md
lhirata: added.

v1.2/drop-sequence.md
lhirata: updated.

v1.2/create-sequence.md
lhirata: updated

v1.2/create-sequence.md
lhirata: updated

v1.2/create-sequence.md
lhirata: updated

vilterp:
Strangely, the backtick formatting didn't work in the note:
screen shot 2017-12-19 at 12 04 05 pm

vilterp:
FYI, moved remaining issues to a GitHub Project: https://github.com/cockroachdb/cockroach/projects/13

Hopefully this is easier to track.

lhirata:
Thanks @vilterp, this is super helpful. Making some final edits / getting a final review on this PR and then will open another PR when your remaining issues are closed.

lhirata:
@vilterp - do you mind reviewing the new doc called RENAME SEQUENCE? also made minor edits to hide info that isn't implemented yet.

@Amruta-Ranade - do you mind doing a review for style / grammar / clarity? @jseldess took a quick look yesterday but i want one more set of eyeballs before i merge. thanks!!

v2.0/create-sequence.md
```diff
@@ -4,13 +4,18 @@ summary:
...
 ---
 
-<span class="version-tag">New in v2.0:</span> The `CREATE SEQUENCE` [statement](sql-statements.html)...
+<span class="version-tag">New in v2.0:</span> The `CREATE SEQUENCE` [statement](sql-statements.html) creates a new sequence in a database. Use a sequence to auto-increment integers in a table.
 
 <div id="toc"></div>
 
+## Considerations
+
+- Using a sequence is slower than using the `SERIAL` data type– incrementing a sequence requires a write to persistent storage, whereas generating a new `SERIAL`value does not. Use the `SERIAL` data type unless a sequence is preferred or required.
```
Amruta-Ranade: nit: "type - increment" Missing space between type and -

Amruta-Ranade: When will a sequence be required? Can we provide example use-cases?

v2.0/create-sequence.md
```diff
@@ -4,13 +4,18 @@ summary:
...
+- A column that uses a sequence can have a gap if a transaction increments the sequence and is then rolled back. Sequence updates are committed immediately and aren't rolled back along with their containing transaction. This is done to avoid blocking concurrent transactions that use the same sequence.
```
Amruta-Ranade: Not sure what this means. So is there a missing value in the column, or is there an erroneous value for the sequence?

v2.0/create-sequence.md
```diff
@@ -33,21 +38,90 @@ table td:first-child {
 
  Parameter | Description
 -----------|------------
-`seqname` |
-`INCREMENT` |
-`MINVALUE` |
-`MAXVALUE` |
-`START` |
-`CYCLE` |
+`seqname` | The name of the sequence to create, which must be unique within its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). When the parent database is not set as the default, the name must be formatted as `database.name`.
```
Amruta-Ranade: nits: "be created", "these" > "the",

lhirata:
Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[v1.2/alter-sequence.md, line 13 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2292#-L0L2bv5Vkw74te8IS1Y-r1-13:-L0uXHFMNr0gx0YPT8-_:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2668a1fc1a38766080bca2c7e525b7dc621da1c5/v1.2/alter-sequence.md#L13)):*

Not quite sure… You can do `SELECT * FROM information_schema.table_privileges`, and see the schema there alongside with other tables.

I guess things should be tested here like "you can't call nextval() on a sequence if you don't have the UPDATE privilege on it"… Which probably fails now. That's something I should fix and put in the automated tests though… Added to my [tracking issue](https://github.com/cockroachdb/cockroach/issues/19723).
</blockquote></details>

Done.

---

*[v1.2/alter-sequence.md, line 44 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2292#-L0L2bv5Vkw74te8IS1Y-r1-44:-L0uWvpfnU6dBSPJR7Wn:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/2668a1fc1a38766080bca2c7e525b7dc621da1c5/v1.2/alter-sequence.md#L44)):*

This doesn't set the sequence value, but it sets:

- the value that it goes to if you do `ALTER SEQUENCE foo RESTART` (not implemented yet)
- the value that it will wrap around to if you hit the `MAXVALUE` and `CYCLE` is on. This should work. E.g.:

```
CREATE SEQUENCE foo MAXVALUE 5 CYCLE
=> CREATE SEQUENCE
... increment up to 5 ...
ALTER SEQUENCE foo START -1000
=> ALTER SEQUENCE
SELECT nextval('foo')
=> 5
```

If we hadn't done the alter sequence, it would have cycled back to 1 (the default start)
</blockquote></details>

Done

---

*[v1.2/drop-sequence.md, line 33 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2292#-L0L2bv5Vkw74te8IS1_-r1-33:-L0uX6E-HnL2LGDWaQ20:b6gja7a) ([raw file](https://github.com/cockroachdb/docs/blob/2668a1fc1a38766080bca2c7e525b7dc621da1c5/v1.2/drop-sequence.md#L33)):*

Apparently yes; that's what PG does. (Of course, I haven't implemented this yet)
</blockquote></details>

commented out for now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2292)*

v2.0/alter-sequence.md
```diff
@@ -4,24 +4,25 @@ summary:
 toc: false
 ---
 
-<span class="version-tag">New in v2.0:</span> The `ALTER SEQUENCE` [statement](sql-statements.html)...
+<span class="version-tag">New in v2.0:</span> The `ALTER SEQUENCE` [statement](sql-statements.html) applies a schema change to a sequence.
```
Amruta-Ranade: This sounds awkward. Can we say "The `ALTER SEQUENCE` changes the name and increment values of a sequence"? The "applies a schema change" wording sounds right for a table or database, not sure it is applicable to sequences.

v2.0/alter-sequence.md
```diff
@@ -34,11 +35,50 @@ table td:first-child {
 
  Parameter | Description
 -----------|------------
+`IF EXISTS` | Modify the sequence only if it exists; if it does not exist, do not return an error.
+`sequence_name` | The name of the sequence you want to modify.
+`INCREMENT` | The new value by which the sequence is incremented. A negative number creates a descending sequence. A positive number creates an ascending sequence.
+`MINVALUE` | The new minimum value of the sequence.
```
Amruta-Ranade: You might want to specify the default values for `MINVALUE` and `MAXVALUE`

v2.0/alter-sequence.md
```diff
@@ -34,11 +35,50 @@ table td:first-child {
...
+`START` | The value the sequence starts at if you `RESTART` or if the sequence hits the `MAXVALUE` and `CYCLE` is set. <br><br>`RESTART` and `CYCLE` are not implemented yet.
+`CYCLE` | The sequence will wrap around when the sequence value hits the maximum or minimum value. If `NO CYCLE` is set, the sequence will not wrap.
+`newname` | The new name of the sequence, which must be unique to its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). <br><br>Name changes do not propagate to the  table(s) using the sequence.
 
 ## Examples
 
-### Example 1
+### Change the Increment of a Sequence
+
+In this example, we're going to change the increment of a sequence from its current state (i.e., `1`) to `2`.
```
Amruta-Ranade: Should it be "increment value"? "Change the increment" sounds odd.

Amruta-Ranade:
Looks good, @lhirata!

v2.0/alter-sequence.md
lhirata: edited

v2.0/alter-sequence.md
lhirata: edited

v2.0/alter-sequence.md
lhirata: added

v2.0/create-sequence.md
lhirata: 1. i think, technically, the em dash doesn't need a space on either side / it depends on the style guide we're adhering to. that said, i added in the space :)
2. sometimes, an app might require using a sequence instead of a randomized int. i'm not sure of a specific example - @vilterp can you provide one?

v2.0/create-sequence.md
lhirata: this is referring to a gap in increment values. for example, if your sequence increments by 1, it would number the rows as 1, 2, 3, 4, 5... but if the update is rolled back, you could end up with 1, 2, 3, 5 (a gap between 3 and 5). do you think this needs to be clarified?

Amruta-Ranade: Ohh! Yes, I think it needs to be clarified. The way it reads now, it seems like there would be a gap in the column, like a missing value in the column -not a gap in sequences.

v2.0/create-sequence.md
lhirata: edited

v2.0/create-sequence.md
lhirata: ok, edited to hopefully be clearer

lhirata:
closes #2014. will open new issues for remaining sequence work.

PR #2295, opened 2017-12-15
More detailed docs on clock sync

### LINKED ISSUE #960: Configuring NTP
From Gitter:

Yan Long @yznming
guys, I setup a 9 nodes cluster using the beta-20161215 version, but some nodes self-terminate several days later because of the clock offset. And I have setup a crontab task to sync the local time with the NTP server every hour. Are there any I’m missing?

@yznming you shouldn’t be syncing ntp with cron, you should be running an ntp client process that keeps the clocks in sync continuously
(and we have some issues with leap seconds currently so lots of CRDB servers crashed on 12/31. we’ll get those fixed before the next one)

Yan Long @yznming 09:41
@bdarnell Thank you for answers. BTW, I think the deploy doc can give more detail about configuring the NTP service.

Thanks, @dianasaur323, for calling this out to me.

### LINKED ISSUE #1690: Add clock sync to troubleshooting guide
From Gitter:

Hi, I am trying to setup cockroachdb cluster using these instructions - https://www.cockroachlabs.com/docs/stable/cluster-setup-troubleshooting.html I stared node1 and then let node2 join. This works fine but when I try to add node3 I am getting a crash

```
 I170708 00:01:58.437337 23 gossip/client.go:131  [n?] started gossip client to 10.160.138.187:26257
F170708 00:01:58.440256 26 server/server.go:162  [n?] fewer than half the known nodes are within the maximum offset of 500ms (0 of 1)
goroutine 26 [running]:
github.com/cockroachdb/cockroach/pkg/util/log.getStacks(0x0, 0x1a3dc698, 0x2bf5ac0, 0x10)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/clog.go:858 +0xa7
github.com/cockroachdb/cockroach/pkg/util/log.(*loggingT).outputLogEntry(0x2bf72a0, 0xc400000004, 0x2485ed9, 0x10, 0xa2, 0xc4206e17a0, 0x54)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/clog.go:724 +0x591
github.com/cockroachdb/cockroach/pkg/util/log.addStructured(0x7f38ddda8000, 0xc420191bc0, 0x4, 0x2, 0x0, 0x0, 0xc420577cf8, 0x1, 0x1)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/structured.go:158 +0x4b3
github.com/cockroachdb/cockroach/pkg/util/log.logDepth(0x7f38ddda8000, 0xc420191bc0, 0x1, 0x4, 0x0, 0x0, 0xc420577cf8, 0x1, 0x1)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/log.go:68 +0x8c
github.com/cockroachdb/cockroach/pkg/util/log.Fatal(0x7f38ddda8000, 0xc420191bc0, 0xc420577cf8, 0x1, 0x1)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/log/log.go:171 +0x76
github.com/cockroachdb/cockroach/pkg/server.NewServer.func1()
        /go/src/github.com/cockroachdb/cockroach/pkg/server/server.go:162 +0xba
github.com/cockroachdb/cockroach/pkg/rpc.(*Context).runHeartbeat(0xc4202afb00, 0xc420308680, 0xc420208500, 0x14, 0x0, 0x0)
        /go/src/github.com/cockroachdb/cockroach/pkg/rpc/context.go:415 +0x55a
github.com/cockroachdb/cockroach/pkg/rpc.(*Context).GRPCDial.func1.2.1(0x7f38ddda8000, 0xc42047db60)
        /go/src/github.com/cockroachdb/cockroach/pkg/rpc/context.go:297 +0x6b
github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).RunWorker.func1(0xc4201ab1e0, 0xc42041a8c0, 0xc42047daa0)
        /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:215 +0xf7
created by github.com/cockroachdb/cockroach/pkg/util/stop.(*Stopper).RunWorker
        /go/src/github.com/cockroachdb/cockroach/pkg/util/stop/stopper.go:216 +0xad
```

Would really appreciate if someone can help with getting the cluster set up. I had to do some firewall setting changes to allow communication between node1 and node2.

@mberhault:

This error means that the clocks on the various hosts are more than 500ms apart. We recommend running ntp (or similar) for best results (see https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#clock-synchronization).

### LINKED ISSUE #1799: Error: `fewer than half the known nodes are within the maximum offset of...`
If debug logs contain the following error:

> `fewer than half the known nodes are within the maximum offset of...`

It's possible that the clock's offset is too large. The solution for this is most likely to configure NTP or synchronize the clocks in another way.

### LINKED ISSUE #1867: Document GCE's non-standard NTP servers
Clusters that span both GCE and other hosting providers can have problems on leap second days, because Google's NTP servers handle leap seconds differently from the standard NTP pool. If you mix GCE and non-GCE nodes in the same cluster, you must either point the GCE nodes at a non-google NTP server or point all the non-GCE nodes at google's (https://developers.google.com/time/).

### LINKED ISSUE #2129: Deploy CockroachDB on AWS EC2 (Insecure) Doc Update
Re: [Deploy CockroachDB on AWS EC2 (Insecure)](https://cockroachlabs.com/docs//v1.1/deploy-cockroachdb-on-aws-insecure.html)

## Issue Description
When following these instructions I occasionally get 
>clock synchronization error: this node is more than 500ms away from at least half of the known nodes (0 of 1 are within the offset)

Even though my AWS instance has NTP running
`$ ntpstat \n
synchronised to NTP server (173.71.68.101) at stratum 2 
   time correct to within 122 ms
   polling server every 64 s`
## Suggested Resolution 

NTP slowly adjusts the time (in teeny tiny changes), so it may take a while to catch up. To force it (before cockroach is running), people tend to do something like `ntpdate -b time.nist.gov` which forces it but you need to stop ntp before running this and restart it after.

`sudo service ntp stop`
`sudo ntpdate -b ntp.ubuntu.com`
` sudo service ntp start`

jseldess:
Fixes #960
Fixes #1867
Fixes #2129
Fixes #1799 
Fixes #1690 

@bdarnell, @mberhault, I've added clock sync instructions for digital ocean, gce, aws, azure, and for manual deployment. However, I haven't yet touched on centOS or Redhat, both of which @robert-s-lee has suggested would be helpful. I'm hesitant to document those distros until we've done more testing on them. Please let me know what you think.

### DIFF
--- a/_includes/prod_deployment/synchronize-clocks.md
+++ b/_includes/prod_deployment/synchronize-clocks.md
@@ -1,0 +1,173 @@
+CockroachDB requires moderate levels of [clock synchronization](recommended-production-settings.html#clock-synchronization) to preserve data consistency. For this reason, when a node detects that its clock is out of synch with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default), it spontaneously shuts down. This avoids the risk of consistency anomalies, but it's best to prevent clocks from drifting too far in the first place by running clock synchronization software on each node.
+
+{% if page.title contains "Digital Ocean" or page.title contains "Manual" %}
+
+[`ntpd`](http://doc.ntp.org/) should keep offsets in the single-digit milliseconds, so that software is featured here, but other methods of clock synchronization are suitable as well.
+
+1. SSH to the first machine.
+
+2. Disable `timesyncd`, which tends to be active by default on some Linux distributions:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo timedatectl set-ntp no
+    ~~~
+
+    Verify that `timesyncd` is off:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ timedatectl
+    ~~~
+
+    Look for `Network time on: no` or `NTP enabled: no` in the output.
+
+3. Install the `ntp` package:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo apt-get install ntp
+    ~~~
+
+4. Stop the NTP daemon:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp stop
+    ~~~
+
+5. Synch the machine's clock with Google's NTP service:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpd -b time.google.com
+    ~~~
+
+    To make this change permanent, in the `/etc/ntp.conf` file, remove or comment out any lines starting with `server` or `pool` and add the following lines:
+
+    {% include copy-clipboard.html %}
+    ~~~
+    server time1.google.com iburst
+    server time2.google.com iburst
+    server time3.google.com iburst
+    server time4.google.com iburst
+    ~~~
+
+    Restart the NTP daemon:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp start
+    ~~~
+
+    {{site.data.alerts.callout_info}}We recommend Google's external NTP service because they handle <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>. If you use a different NTP service that doesn't smear the leap second, you must configure client-side smearing manually and do so in the same way on each machine.{{site.data.alerts.end}}
+
+6. Verify that the machine is using a Google NTP server:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpq -p
+    ~~~
+
+    The active NTP server will be marked with an asterisk.
+
+7. Repeat these steps for each machine where a CockroachDB node will run.
+
+{% elsif page.title contains "Google" %}
+
+Compute Engine instances are preconfigured to use [NTP](http://www.ntp.org/), which should keep offsets in the single-digit milliseconds. However, Google can’t predict how external NTP services, such as `pool.ntp.org`, will handle the leap second. Therefore, you should:
+
+- [Configure each GCE instances to use Google's internal NTP service](https://cloud.google.com/compute/docs/instances/managing-instances#configure_ntp_for_your_instances).
+- If you plan to run a hybrid cluster across GCE and other cloud providers or environments, [configure the non-GCE machines to use Google's external NTP service](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks).
+
+{% elsif page.title contains "AWS" %}
+
+Amazon provides the [Amazon Time Sync Service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html), which uses a fleet of satellite-connected and atomic reference clocks in each AWS Region to deliver accurate current time readings. The service also smears the leap second.
+
+- If you plan to run your entire cluster on AWS, [configure each AWS instance to use the internal Amazon Time Sync Service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html#configure-amazon-time-service).
+- However, if you plan to run a hybrid cluster across AWS and other cloud providers or environments, [configure all machines to use Google's external NTP service](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks), which is comparably accurate and also handles <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>.
+
+{% elsif page.title contains "Azure" %}
+
+[`ntpd`](http://doc.ntp.org/) should keep offsets in the single-digit milliseconds, so that software is featured here. However, to run `ntpd` properly on Azure VMs, it's necessary to first unbind the Time Synchronization device used by the Hyper-V technology running Azure VMs; this device aims to synchronize time between the VM and its host operating system but has been known to cause problems.
+
+1. SSH to the first machine.
+
+2. Find the the ID of the Hyper-V Time Synchronization device:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ curl -O https://raw.githubusercontent.com/torvalds/linux/master/tools/hv/lsvmbus
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ python lsvmbus -vv | grep -w "Time Synchronization" -A 3
+    ~~~
+
+    ~~~
+    VMBUS ID 12: Class_ID = {9527e630-d0ae-497b-adce-e80ab0175caf} - [Time Synchronization]
+        Device_ID = {2dd1ce17-079e-403c-b352-a1921ee207ee}
+        Sysfs path: /sys/bus/vmbus/devices/2dd1ce17-079e-403c-b352-a1921ee207ee
+        Rel_ID=12, target_cpu=0
+    ~~~
+
+3. Unbind the device, using the `Device_ID` from the previous command's output:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ echo <DEVICE_ID> | sudo tee /sys/bus/vmbus/drivers/hv_util/unbind
+    ~~~
+
+4. Install the `ntp` package:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo apt-get install ntp
+    ~~~
+
+5. Stop the NTP daemon:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp stop
+    ~~~
+
+6. Synch the machine's clock with Google's NTP service:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpd -b time.google.com
+    ~~~
+
+    To make this change permanent, in the `/etc/ntp.conf` file, remove or comment out any lines starting with `server` or `pool` and add the following lines:
+
+    {% include copy-clipboard.html %}
+    ~~~
+    server time1.google.com iburst
+    server time2.google.com iburst
+    server time3.google.com iburst
+    server time4.google.com iburst
+    ~~~
+
+    Restart the NTP daemon:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp start
+    ~~~
+
+    {{site.data.alerts.callout_info}}We recommend Google's NTP service because they handle <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>. If you use a different NTP service that doesn't smear the leap second, be sure to configure client-side smearing in the same way on each machine.{{site.data.alerts.end}}
+
+7. Verify that the machine is using a Google NTP server:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpq -p
+    ~~~
+
+    The active NTP server will be marked with an asterisk.
+
+8. Repeat these steps for each machine where a CockroachDB node will run.
+
+{% endif %}
--- a/_includes/faq/clock-synchronization.html
+++ b/_includes/faq/clock-synchronization.html
@@ -1,5 +1,15 @@
-CockroachDB needs moderately accurate time to preserve data consistency, so it's important to run [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
-
-By default, CockroachDB's maximum allowed clock offset is 500ms. When a node detects that its clock offset, relative to other nodes, is half or more of the maximum allowed, it spontaneously shuts down. This is well in advance of the maximum offset (500ms by default), beyond which [serializable consistency](https://en.wikipedia.org/wiki/Serializability) is not guaranteed and stale reads and write skews could occur. With NTP or other clock synchronization software running on each node, there's very little risk of ever exceeding the maximum offset and encountering such anomalies, and even on well-functioning hardware not running synchronization software, slow clock drift is most common, which CockroachDB handles safely.
+CockroachDB requires moderate levels of clock synchronization to preserve data consistency. For this reason, when a node detects that its clock is out of synch with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default), it spontaneously shuts down. This avoids the risk of violating [serializable consistency](https://en.wikipedia.org/wiki/Serializability) and causing stale reads and write skews, but it's important to prevent clocks from drifting too far in the first place by running [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
 
 The one rare case to note is when a node's clock suddenly jumps beyond the maximum offset before the node detects it. Although extremely unlikely, this could occur, for example, when running CockroachDB inside a VM and the VM hypervisor decides to migrate the VM to different hardware with a different time. In this case, there can be a small window of time between when the node's clock becomes unsynchronized and when the node spontaneously shuts down. During this window, it would be possible for a client to read stale data and write data derived from stale reads.
+
+For guidance on synchronizing clocks, see the tutorial for your deployment environment:
+
+Environment | Recommended Approach
+------------|---------------------
+[Manual](manual-deployment.html#step-1-synchronize-clocks) | Use NTP with Google's external NTP service.
+[AWS](deploy-cockroachdb-on-aws.html#step-3-synchronize-clocks) | Use the Amazon Time Sync Service.
+[Azure](deploy-cockroachdb-on-microsoft-azure.html#step-3-synchronize-clocks) | Disable Hyper-V time synchronization and use NTP with Google's external NTP service.
+[Digital Ocean](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks) | Use NTP with Google's external NTP service.
+[GCE](deploy-cockroachdb-on-google-cloud-platform.html#step-3-synchronize-clocks) | Use NTP with Google's internal NTP service.
+
+{{site.data.alerts.callout_info}}In most cases, we recommend Google's external NTP service because they handle <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>. If you use a different NTP service that doesn't smear the leap second, you must configure client-side smearing manually and do so in the same way on each machine.{{site.data.alerts.end}}
--- a/v1.0/operational-faqs.md
+++ b/v1.0/operational-faqs.md
@@ -45,7 +45,11 @@ Collecting information about CockroachDB's real world usage helps us prioritize 
 
 ## What happens when node clocks are not properly synchronized?
 
-{% include faq/clock-synchronization.html %}
+CockroachDB needs moderately accurate time to preserve data consistency, so it's important to run [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
+
+By default, CockroachDB's maximum allowed clock offset is 500ms. When a node detects that its clock offset, relative to other nodes, is half or more of the maximum allowed, it spontaneously shuts down. This is well in advance of the maximum offset (500ms by default), beyond which [serializable consistency](https://en.wikipedia.org/wiki/Serializability) is not guaranteed and stale reads and write skews could occur. With NTP or other clock synchronization software running on each node, there's very little risk of ever exceeding the maximum offset and encountering such anomalies, and even on well-functioning hardware not running synchronization software, slow clock drift is most common, which CockroachDB handles safely.
+
+The one rare case to note is when a node's clock suddenly jumps beyond the maximum offset before the node detects it. Although extremely unlikely, this could occur, for example, when running CockroachDB inside a VM and the VM hypervisor decides to migrate the VM to different hardware with a different time. In this case, there can be a small window of time between when the node's clock becomes unsynchronized and when the node spontaneously shuts down. During this window, it would be possible for a client to read stale data and write data derived from stale reads.
 
 ## See Also
 
--- a/v1.0/recommended-production-settings.md
+++ b/v1.0/recommended-production-settings.md
@@ -44,7 +44,11 @@ For details about controlling the number and location of replicas, see [Configur
 
 ## Clock Synchronization
 
-{% include faq/clock-synchronization.html %}
+CockroachDB needs moderately accurate time to preserve data consistency, so it's important to run [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
+
+By default, CockroachDB's maximum allowed clock offset is 500ms. When a node detects that its clock offset, relative to other nodes, is half or more of the maximum allowed, it spontaneously shuts down. This is well in advance of the maximum offset (500ms by default), beyond which [serializable consistency](https://en.wikipedia.org/wiki/Serializability) is not guaranteed and stale reads and write skews could occur. With NTP or other clock synchronization software running on each node, there's very little risk of ever exceeding the maximum offset and encountering such anomalies, and even on well-functioning hardware not running synchronization software, slow clock drift is most common, which CockroachDB handles safely.
+
+The one rare case to note is when a node's clock suddenly jumps beyond the maximum offset before the node detects it. Although extremely unlikely, this could occur, for example, when running CockroachDB inside a VM and the VM hypervisor decides to migrate the VM to different hardware with a different time. In this case, there can be a small window of time between when the node's clock becomes unsynchronized and when the node spontaneously shuts down. During this window, it would be possible for a client to read stale data and write data derived from stale reads.
 
 ## Cache Size
 
--- a/v1.1/common-errors.md
+++ b/v1.1/common-errors.md
@@ -137,9 +137,17 @@ When running a multi-node CockroachDB cluster, if you see an error like the one 
 
 ## clock synchronization error: this node is more than 500ms away from at least half of the known nodes
 
-This message indicates that a node has spontaneously shut down because it detected that its clock is out of sync with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default). CockroachDB requires moderate levels of [clock synchronization](recommended-production-settings.html#clock-synchronization) to preserve data consistency, so the node shutting down in this way avoids the risk of consistency anomalies.
+This error indicates that a node has spontaneously shut down because it detected that its clock is out of synch with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default). CockroachDB requires moderate levels of [clock synchronization](recommended-production-settings.html#clock-synchronization) to preserve data consistency, so the node shutting down in this way avoids the risk of consistency anomalies.
 
-To prevent this from happening, you should run [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
+To prevent this from happening, you should run clock synchronization software on each node. For guidance on synchronizing clocks, see the tutorial for your deployment environment:
+
+Environment | Recommended Approach
+------------|---------------------
+[Manual](manual-deployment.html#step-1-synchronize-clocks) | Use NTP with Google's external NTP service.
+[AWS](deploy-cockroachdb-on-aws.html#step-3-synchronize-clocks) | Use the Amazon Time Sync Service.
+[Azure](deploy-cockroachdb-on-microsoft-azure.html#step-3-synchronize-clocks) | Disable Hyper-V time synchronization and use NTP with Google's external NTP service.
+[Digital Ocean](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks) | Use NTP with Google's external NTP service.
+[GCE](deploy-cockroachdb-on-google-cloud-platform.html#step-3-synchronize-clocks) | Use NTP with Google's internal NTP service.
 
 ## context deadline exceeded
 
--- a/v1.1/deploy-cockroachdb-on-aws-insecure.md
+++ b/v1.1/deploy-cockroachdb-on-aws-insecure.md
@@ -70,7 +70,11 @@ You can create these rules using [Security Groups' Inbound Rules](http://docs.aw
 - Running at least 3 CockroachDB nodes to ensure survivability.
 - Selecting the same continent for all of your instances for best performance.
 
-## Step 3. Set up load balancing
+## Step 3. Synchronize clocks
+
+{% include prod_deployment/synchronize-clocks.md %}
+
+## Step 4. Set up load balancing
 
 Each CockroachDB node is an equally suitable SQL gateway to your cluster, but to ensure client performance and reliability, it's important to use TCP load balancing:
 
@@ -87,23 +91,23 @@ AWS offers fully-managed load balancing to distribute traffic between instances.
 
 {{site.data.alerts.callout_info}}If you would prefer to use HAProxy instead of AWS's managed load balancing, see <a href="manual-deployment-insecure.html">Manual Deployment</a> for guidance.{{site.data.alerts.end}}
 
-## Step 4. Start nodes
+## Step 5. Start nodes
 
 {% include prod_deployment/insecure-start-nodes.md %}
 
-## Step 5. Initialize the cluster
+## Step 6. Initialize the cluster
 
 {% include prod_deployment/insecure-initialize-cluster.md %}
 
-## Step 6. Test the cluster
+## Step 7. Test the cluster
 
 {% include prod_deployment/insecure-test-cluster.md %}
 
-## Step 7. Test load balancing
+## Step 8. Test load balancing
 
 {% include prod_deployment/insecure-test-load-balancing.md %}
 
-## Step 8. Use the cluster
+## Step 9. Use the cluster
 
 Now that your deployment is working, you can:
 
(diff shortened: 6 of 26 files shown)


jseldess:
Added instructions for GCE as well.

jseldess:
Added AWS. Remaining: Azure and manual deployment.

jseldess:
Also added clock offset to the troubleshooting docs.

_includes/prod_deployment/synchronize-clocks.md
```diff
@@ -0,0 +1,145 @@
...
+4. Force the machine's clock to synch with Google's NTP service:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp stop
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpd -b time.google.com
```
mberhault: `ntpdate`, not `ntpd`. We should also mention why we're doing that.
`ntpdate` also need to be installed, it's not included in the `ntp` package. So the apt-get line becomes: `sudo apt-get install ntpdate ntp`.

If we're using `time.google.com`, then we need to modify `/etc/ntp.conf` as well.

_includes/prod_deployment/synchronize-clocks.md
```diff
@@ -0,0 +1,145 @@
...
+- If you plan to run your cluster across cloud providers, [configure the non-GCE machines to use Google's external NTP service](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks).
+
+{% elsif page.title contains "AWS" %}
+
+Amazon provides the [Amazon Time Sync Service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html), which uses a fleet of satellite-connected and atomic reference clocks in each AWS Region to deliver accurate current time readings. The service also smears the leap second.
+
+- [Configure each AWS instance to use the internal Amazon Time Sync Service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html#configure-amazon-time-service).
+- If you plan to run your cluster across cloud providers, [configure the non-AWS machines to use Google's external NTP service](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks), which is comparably accurate and also handles <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>.
```
mberhault: Does amazon not offer an externally-available time server that we can recommended for non-AWS instances?

_includes/prod_deployment/synchronize-clocks.md
```diff
@@ -0,0 +1,145 @@
...
+5. Force the machine's clock to synch with Google's NTP service:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo service ntp stop
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo ntpd -b time.google.com
```
mberhault: ditto: ntpdate.

bdarnell:
Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0Rswr6IgUAW_0zhmMU:b-oreaky) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

`ntpdate`, not `ntpd`. We should also mention why we're doing that.
`ntpdate` also need to be installed, it's not included in the `ntp` package. So the apt-get line becomes: `sudo apt-get install ntpdate ntp`.

If we're using `time.google.com`, then we need to modify `/etc/ntp.conf` as well.
</blockquote></details>

Any discussion of time.google.com needs to be accompanied by a warning that if any of your nodes use google's time server, they all must. Reconfiguring servers to use google's time service is more involved than simply enabling ntp with the default pool, so I'm not sure if we want to recommend this by default (OTOH, unless you do something about smearing the cluster will crash on the leap second)

---

*[_includes/prod_deployment/synchronize-clocks.md, line 74 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-74:-L0RuFWnMFTH2_6NKrto:b-7yvhlx) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L74)):*

Does amazon not offer an externally-available time server that we can recommended for non-AWS instances?
</blockquote></details>

I don't think they do. 

---

*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L0RtjjfzbcD8YIoTSEU:b-sjp2tl) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*
> ```Markdown
>     ~~~ shell
>     $ curl -O https://raw.githubusercontent.com/torvalds/linux/master/tools/hv/lsvmbus
> ```

Suggesting that users copy these commands one at a time is awkward. We should be providing a script that can be run in a single shot. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0UjW_7tRHFHg1Tvorg:b-mez1nk) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*
Thanks, @mberhault and @bdarnell. I'll make changes soon. But @mberhault, why is it important to use `ntpdate` as opposed to `ntpd`? The [docs](http://doc.ntp.org/4.1.1/ntpdate.htm) say:

> Disclaimer: The functionality of this program is now available in the ntpd program. See the -q command line option in the ntpd - Network Time Protocol (NTP) daemon page. After a suitable period of mourning, the ntpdate program is to be retired from this distribution

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0UkPkFultKtPjysvKj:b-k8xjsl) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

Thanks, @mberhault and @bdarnell. I'll make changes soon. But @mberhault, why is it important to use `ntpdate` as opposed to `ntpd`? The [docs](http://doc.ntp.org/4.1.1/ntpdate.htm) say:

> Disclaimer: The functionality of this program is now available in the ntpd program. See the -q command line option in the ntpd - Network Time Protocol (NTP) daemon page. After a suitable period of mourning, the ntpdate program is to be retired from this distribution
</blockquote></details>

Also, @mberhault, if we say to modify `/etc/ntp.conf`, when would they do that? After or before the steps here? And would we just tell them to do [this](https://developers.google.com/time/guides#linux_ntpd)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0UkvvJUAoyWZu3k6aQ:b9xse2b) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

Also, @mberhault, if we say to modify `/etc/ntp.conf`, when would they do that? After or before the steps here? And would we just tell them to do [this](https://developers.google.com/time/guides#linux_ntpd)?
</blockquote></details>

@bdarnell, are you ok with our recommending google's time server? Seems easiest way to account for leap second smearing. If so, you mention that the process is more involved. What's missing here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

mberhault:
Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0Ul0bsw_3dc6-LXD6-:b-bpf9ay) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

@bdarnell, are you ok with our recommending google's time server? Seems easiest way to account for leap second smearing. If so, you mention that the process is more involved. What's missing here?
</blockquote></details>

oh, yeah, it's `ntpd -q` (just set the time and quit). Definitely no `-b` though, that's a `ntpdate` flag.
If we're using custom servers to `ntpd -q`, then we need to make sure those same servers are in the config file. It needs to be done after `ntpd` is stopped and before it is started up again.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

bdarnell:
Review status: 0 of 19 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 9 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0WGAI7mUP6oj25VUod:-L0WGAI7mUP6oj25VUoe:b-nij5i0) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L9)):*
> ```Markdown
> 1. SSH to the first machine.
> 
> 2. Disable `timesyncd`, which tends to be active by default on some Linux distributions:
> ```

`timesyncd` is the default on ubuntu. Could/should we recommend that ubuntu users point `timesyncd` at `time.google.com` instead of switching to `ntpd`? Is the preference for `ntpd` just because that's the one we're most familiar with or is there some reason to favor it over other NTP clients?

---

*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L0WDvcOZqK7sgevTjsz:b773l3o) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

oh, yeah, it's `ntpd -q` (just set the time and quit). Definitely no `-b` though, that's a `ntpdate` flag.
If we're using custom servers to `ntpd -q`, then we need to make sure those same servers are in the config file. It needs to be done after `ntpd` is stopped and before it is started up again.
</blockquote></details>

The "more involved" part is the fact that you need to edit `/etc/ntp.conf` and *make sure* that you do this for every current and future node. I agree that this is probably still the simplest way to account for the leap second, though. (since to do client-side slewing it looks like you need to use chronyd instead of ntpd)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 9 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0WGAI7mUP6oj25VUod:-L0XN3S1WyWMCWyUKqYx:b-swq3wq) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L9)):*

`timesyncd` is the default on ubuntu. Could/should we recommend that ubuntu users point `timesyncd` at `time.google.com` instead of switching to `ntpd`? Is the preference for `ntpd` just because that's the one we're most familiar with or is there some reason to favor it over other NTP clients?
</blockquote></details>

Good question. I've read here and here that `timesyncd` should be sufficient. I could have the user check for `timesyncd` and, if it's present, configure it, if it's not present, install and configure `ntp`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 9 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0WGAI7mUP6oj25VUod:-L0XSjOuzYDZGVItSPl2:b-yd0znh) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L9)):*

Good question. I've read here and here that `timesyncd` should be sufficient. I could have the user check for `timesyncd` and, if it's present, configure it, if it's not present, install and configure `ntp`?
</blockquote></details>

I've also read that [`chrony` performs better](https://chrony.tuxfamily.org/comparison.html). And [configuration seems just as easy]( https://www.server-world.info/en/note?os=Ubuntu_16.04&p=ntp&f=2).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
@mberhault, @bdarnell, any updates/thoughts on this work? I believe Mark was going to try to use ntp with the external google time service on one of our production clusters before we publish these recommendations...

_includes/prod_deployment/synchronize-clocks.md
```diff
@@ -0,0 +1,145 @@
...
+
+[NTP](http://www.ntp.org/) should keep offsets in the single-digit milliseconds, so that software is featured here, but other methods of clock synchronization are suitable as well.
+
+1. SSH to the first machine.
+
+2. Disable `timesyncd`, which tends to be active by default on some Linux distributions:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ sudo timedatectl set-ntp no
```
mberhault: I can confirm that this disables `systemd-timesyncd.service` permanently (as opposed to just stopping it). This is mentioned in the [timedatectl docs](https://www.freedesktop.org/software/systemd/man/timedatectl.html):
```
This command is hence mostly equivalent to: systemctl enable --now systemd-timesyncd.service
and systemctl disable --now systemd-timesyncd.service, but is protected by a different access policy.
```

_includes/prod_deployment/synchronize-clocks.md
mberhault: Making this change permanent involves changing `/etc/ntp.conf` and restarting the service.
Google [has docs](https://developers.google.com/time/guides#linux_ntpd) on this for various systems.
For ubuntu, the new file looks like:
```
# Specify one or more NTP servers.

# Use servers from the NTP Pool Project. Approved by Ubuntu Technical Board
# on 2011-02-08 (LP: #104525). See http://www.pool.ntp.org/join.html for
# more information.
#pool 0.ubuntu.pool.ntp.org iburst
#pool 1.ubuntu.pool.ntp.org iburst
#pool 2.ubuntu.pool.ntp.org iburst
#pool 3.ubuntu.pool.ntp.org iburst

# Use Ubuntu's ntp server as a fallback.
#pool ntp.ubuntu.com

server time1.google.com iburst
server time2.google.com iburst
server time3.google.com iburst
server time4.google.com iburst
```

Where I removed the `ntp.ubuntu.com` and switched to using the servers specified in the Google docs.

jseldess:
Review status: 0 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 41 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-41:-L2ewvNnj7xt3VsWWe7S:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L41)):*

Making this change permanent involves changing `/etc/ntp.conf` and restarting the service.
Google [has docs](https://developers.google.com/time/guides#linux_ntpd) on this for various systems.
For ubuntu, the new file looks like:
```
# Specify one or more NTP servers.

# Use servers from the NTP Pool Project. Approved by Ubuntu Technical Board
# on 2011-02-08 (LP: #104525). See http://www.pool.ntp.org/join.html for
# more information.
#pool 0.ubuntu.pool.ntp.org iburst
#pool 1.ubuntu.pool.ntp.org iburst
#pool 2.ubuntu.pool.ntp.org iburst
#pool 3.ubuntu.pool.ntp.org iburst

# Use Ubuntu's ntp server as a fallback.
#pool ntp.ubuntu.com

server time1.google.com iburst
server time2.google.com iburst
server time3.google.com iburst
server time4.google.com iburst
```

Where I removed the `ntp.ubuntu.com` and switched to using the servers specified in the Google docs.
</blockquote></details>

Done.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L0UjzX4PCt8V9XcKM5O:bm7eq7a) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*

Suggesting that users copy these commands one at a time is awkward. We should be providing a script that can be run in a single shot. 
</blockquote></details>

Marc, Ben, I think the instructions are now up to date with our latest understanding. Do you think it's still meaningful to write script? They have to edit the `conf` file, which might need to be done, manually, right?

---

*[_includes/prod_deployment/synchronize-clocks.md, line 124 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0NkYiXba0uM-fx3Teu-r3-124:-L2exLSYHSmS-yclFBt0:b-oif3on) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L124)):*

ditto: ntpdate.
</blockquote></details>

I can't find where, but I'm pretty sure we agreed to go with `ntpd` instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

_includes/prod_deployment/synchronize-clocks.md
jseldess: Great, thanks.

jseldess:
@mberhault, @bdarnell, the docs should be up-to-date with our latest discussions. Please take another look before I apply the changes to the 2.0 docs as well. Thanks for all the help here!

bdarnell:
Reviewed 2 of 6 files at r4.
Review status: 2 of 13 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[_includes/faq/clock-synchronization.html, line 1 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2fefEIX1nKMwmnlo4u:-L2fefEIX1nKMwmnlo4v:b-fgdmrk) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/faq/clock-synchronization.html#L1)):*
> ```HTML
> CockroachDB requires moderate levels of clock synchronization to preserve data consistency. For this reason, when a node detects that its clock is out of synch with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default), it spontaneously shuts down. This avoids the risk of violating [serializable consistency](https://en.wikipedia.org/wiki/Serializability) and causing stale reads and write skews, but it's best to prevent clocks from drifting too far in the first place by running [NTP](http://www.ntp.org/) or other clock synchronization software on each node.
> ```

I think "it's best" is too weak here - I'd say "it's important"

---

*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L2ffwXC0I8MiITZBDyV:bfcz6r7) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*

Marc, Ben, I think the instructions are now up to date with our latest understanding. Do you think it's still meaningful to write script? They have to edit the `conf` file, which might need to be done, manually, right?
</blockquote></details>

Yes, I still think we should write a script for this.

The conf file would not be edited manually in practice - you'd have a master copy in some config repository and copy it over. 

---

*[_includes/prod_deployment/synchronize-clocks.md, line 5 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2fevbFByTFSiB8kyrO:-L2fevbFByTFSiB8kyrP:b90qovm) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L5)):*
> ```Markdown
> {% if page.title contains "Digital Ocean" or page.title contains "Manual" %}
> 
> [NTP](http://www.ntp.org/) should keep offsets in the single-digit milliseconds, so that software is featured here, but other methods of clock synchronization are suitable as well.
> ```

The software we're featuring here is `ntpd`, not NTP. NTP is the protocol, which is used by all three software packages (ntpd, timesyncd, chrony). 

---

*[_includes/prod_deployment/synchronize-clocks.md, line 46 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffIA5MmpkklQzhDGA:-L2ffIA5MmpkklQzhDGB:b-3e0qk5) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L46)):*
> ```Markdown
>     ~~~
> 
>     To make this change permanent, in the `/etc/ntp.conf` file, remove or comment out any lines starting with `server` and add the following lines:
> ```

From Marc's previous comment, it looks like you also have to remove any `pool` lines as well. 

Is there anything in the file that we would want to keep, or can we just say to replace the entire file with these four `server` lines?

---

*[_includes/prod_deployment/synchronize-clocks.md, line 78 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffU8DP9DUWQ4Egzji:-L2ffU8DP9DUWQ4Egzjj:b-pmgds) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L78)):*
> ```Markdown
> {% elsif page.title contains "Google" %}
> 
> Compute Engine instances are preconfigured to use [NTP](http://www.ntp.org/), which should keep offsets in the single-digit milliseconds. However, Google can’t predict how external NTP services, such as `pool.ntp.org`, will handle the leap second. Therefore, you should:
> ```

Whether they are pre-configured with NTP depends on which base image you're using; I don't think this is guaranteed. 

---

*[_includes/prod_deployment/synchronize-clocks.md, line 88 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffhQVzs2CvnBT-1Ex:-L2ffhQVzs2CvnBT-1Ey:bmqhwuc) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L88)):*
> ```Markdown
> 
> - [Configure each AWS instance to use the internal Amazon Time Sync Service](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html#configure-amazon-time-service).
> - If you plan to run your cluster across cloud providers, [configure the non-AWS machines to use Google's external NTP service](deploy-cockroachdb-on-digital-ocean.html#step-2-sychronize-clocks), which is comparably accurate and also handles <a href="https://developers.google.com/time/smear">"smearing" the leap second</a>.
> ```

I don't think it's safe to mix amazon's internal time servers with google's external ones. I think if you envision a cross-cloud deployment you should use google's external time servers even on AWS. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
Note: Mention somewhere that you can get the current cluster offset via the `clock_offset_* variables` at `/_status/vars`.

mberhault:
Review status: 2 of 13 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 46 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffIA5MmpkklQzhDGA:-L34Sklv2yZrsVbh4RYG:b4ayjl5) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L46)):*

From Marc's previous comment, it looks like you also have to remove any `pool` lines as well. 

Is there anything in the file that we would want to keep, or can we just say to replace the entire file with these four `server` lines?
</blockquote></details>

Yup, the pool lines all need to go away. I kept the rest of the file as the GCE machines had them, but I'm not sure which parts are useful.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
Review status: 2 of 13 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[_includes/faq/clock-synchronization.html, line 1 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2fefEIX1nKMwmnlo4u:-L34LwqyO_eHUoeZXPZM:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/faq/clock-synchronization.html#L1)):*

I think "it's best" is too weak here - I'd say "it's important"
</blockquote></details>

Done.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L34Tk6EEhS8VG1me31F:b-mlqn2r) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*

Yes, I still think we should write a script for this.

The conf file would not be edited manually in practice - you'd have a master copy in some config repository and copy it over. 
</blockquote></details>

I'd like to get this merged, so I'd prefer to convert these steps to a script in a follow-up PR. 

---

*[_includes/prod_deployment/synchronize-clocks.md, line 5 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2fevbFByTFSiB8kyrO:-L34Pz9epqqRESGKGClA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L5)):*

The software we're featuring here is `ntpd`, not NTP. NTP is the protocol, which is used by all three software packages (ntpd, timesyncd, chrony). 
</blockquote></details>

Done.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 78 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffU8DP9DUWQ4Egzji:-L34Qagp8Qw4eoUHAHmP:b-5cbul4) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L78)):*
The [GCE docs](https://cloud.google.com/compute/docs/instances/managing-instances#configure-ntp) claim:

> To help keep system clocks in sync, Compute Engine instances are preconfigured to use network time protocol (NTP).

So I'd prefer to trust them and keep this simple until someone tells us it's wrong. That ok?

---

*[_includes/prod_deployment/synchronize-clocks.md, line 88 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffhQVzs2CvnBT-1Ex:-L34S84PpISmpS51pjJh:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L88)):*

I don't think it's safe to mix amazon's internal time servers with google's external ones. I think if you envision a cross-cloud deployment you should use google's external time servers even on AWS. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 46 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffIA5MmpkklQzhDGA:-L34UOvYdpXjwhHzeoKq:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L46)):*

Yup, the pool lines all need to go away. I kept the rest of the file as the GCE machines had them, but I'm not sure which parts are useful.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

mberhault:
Review status: 0 of 24 files reviewed at latest revision, 10 unresolved discussions, some commit checks pending.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L34Z8AUAI9qRrN5Ni0p:b-2r03bl) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*

I'd like to get this merged, so I'd prefer to convert these steps to a script in a follow-up PR. 
</blockquote></details>

I think a single copyable text block would do rather than individual blocks (for single-shot copy/paste). Description of what it does can be moved above. That should be just for the Azure bits, the ntp switcheroo can remain individual blocks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

bdarnell:
Reviewed 1 of 14 files at r5.
Review status: 1 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

*[_includes/prod_deployment/synchronize-clocks.md, line 78 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L2ffU8DP9DUWQ4Egzji:-L34bfTZDndm0bJt94_q:b32w0x) ([raw file](https://github.com/cockroachdb/docs/blob/30261bf9fa486c14d3fd8397fec9f53921af7933/_includes/prod_deployment/synchronize-clocks.md#L78)):*

The [GCE docs](https://cloud.google.com/compute/docs/instances/managing-instances#configure-ntp) claim:

> To help keep system clocks in sync, Compute Engine instances are preconfigured to use network time protocol (NTP).

So I'd prefer to trust them and keep this simple until someone tells us it's wrong. That ok?
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

mberhault:
LGTM. Let's leave scripting for later if we need it. We would have to account for various systems before providing a full copy/paste solution.

---

Review status: 1 of 24 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-:-L34eFnQDzQ-KZjE0fPb:b-l7vfmd)*

jseldess:
*[_includes/prod_deployment/synchronize-clocks.md, line 86 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2295#-L0RtjjfzbcD8YIoTSET:-L34if9frRTIfQmZrcPe:bq51y3y) ([raw file](https://github.com/cockroachdb/docs/blob/b5a09f443c69e5babf2d7c24610eeda1da188f86/_includes/prod_deployment/synchronize-clocks.md#L86)):*

I think a single copyable text block would do rather than individual blocks (for single-shot copy/paste). Description of what it does can be moved above. That should be just for the Azure bits, the ntp switcheroo can remain individual blocks.
</blockquote></details>

We discussed offline. Given that unbinding the device relies on retrieving the DEVICE_ID from the output of the second command, and it's not guaranteed to be identical on all system, we agreed to keep the instructions explicit for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2295)*

PR #2298, opened 2017-12-18
Added note about escaping percentage sign

### LINKED ISSUE #2082: Document that % in cockroach flag values may need to be escaped
See https://github.com/cockroachdb/cockroach/issues/19393 for reference.

Relevant `cockroach start` flags:
- `--cache=25%`
- `--max-sql-memory=25%`
- `--max-disk-temp-storage=10%`
`--store=path=/mnt/ssd01,size=20%`

While we're at it, a bunch of the examples in the `store` > `size` field description are `--store-path=...`, whereas they should be `--store=path=...`

Amruta-Ranade:
Closes #2082 

@jseldess I took a shot at what the docs requirement was. Let me know if the note I added is sufficient. 

As for the comment, "While we're at it, a bunch of the examples in the store > size field description are --store-path=..., whereas they should be --store=path=...", I can't see any instances in the store>size description. Am I missing something?


bdarnell:
This is kind of vague and not actionable. When and how might I need to escape the percent sign? (systemd configs are one known case according to cockroachdb/cockroach#19393, where the percent sign needs to be doubled). It's better to be specific here. 

This change also needs to be copied to the 2.0 version of the docs.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2298#-:-L0f0SYWomfkKKIOce9q:b71cpzx)*

Amruta-Ranade:
@bdarnell What are the specific instances where the sign needs to be escaped?

bdarnell:
The only case I know about is in systemd service files.

Amruta-Ranade:
Okay..so will it suffice to add "for instance, while configuring CockroachDB through systemd service files" ? Or is there a way I can find specific instances?

bdarnell:
Yes, that would be fine.

Amruta-Ranade:
Okay..a GitHub issue..Should I restore this branch and make the changes, or make a new branch?

jseldess:
Totally fine to just open a follow-up PR, @Amruta-Ranade.

Issue #631, opened 2016-09-07
Add secure option for "Build an App"


dianasaur323:
Current "Build a Test App" tutorial assumes insecure clusters. Add in an option for secure clusters.

`conn = psycopg2.connect(database='bank',user='maxroach',host='localhost',port=26257, sslmode='require', sslrootcert='certs/ca.cert', sslcert='certs/maxroach.cert', sslkey='certs/maxroach.key')`

mberhault:
There are a few steps here, for each example:
* determine how the listed library accepts certs/keys. parameter names, file formats etc..
* when file formats differ, document how to convert our certs/keys to the expected format
* consider having the `cockroach cert` tools generate other formats.

knz:
A user shared a working node.js example on the forum: https://forum.cockroachlabs.com/t/basic-example-in-secure-mode/1339/6?u=knz

jseldess:
From https://github.com/cockroachdb/docs/issues/2089:

Here is a quick secure example for PHP if needed.

Current insecure example:
```
$dbh = new PDO('pgsql:host=localhost;port=26257;dbname=bank;sslmode=disable',
    'maxroach', null, array(
      PDO::ATTR_ERRMODE          => PDO::ERRMODE_EXCEPTION,
      PDO::ATTR_EMULATE_PREPARES => true,
  ));
```

Secure example:
```
$dbh = new PDO('pgsql:localhost=host;port=26257;dbname=bank;sslmode=require;sslcert=[path]/client.maxroach.crt;sslkey=[path]/client.maxroach.key;sslrootcert=[path]/ca.crt;',
    'maxroach', null, array(
      PDO::ATTR_ERRMODE          => PDO::ERRMODE_EXCEPTION,
      PDO::ATTR_EMULATE_PREPARES => true,
  ));
```

jseldess:
See Java post on forum: See https://forum.cockroachlabs.com/t/connecting-to-an-ssl-secure-server-using-jdbc-java-and-client-certificate-authentication/400

mberhault:
Ping on this: java is particularly tricky and could use an example. See https://github.com/cockroachdb/cockroach/issues/24487

jseldess:
Thanks, @mberhault. @awoods187, let's talk about prioritizing this as a docs/eng project for 2.1, perhaps focusing just on jdbc and hibernate.

jseldess:
For 2.1, part of https://github.com/cockroachdb/docs/issues/2943.

[...]

PR #2311, opened 2017-12-19
Troubleshooting common errors

jseldess:
This PR revises the troubleshooting overview to prescribe
intial steps, converts the general troubleshooting page into a
list of common errors with solutions, and cleans up some page
names.


jseldess:
Some other scenarios discussed with @tschottdorf and @petermattis that I think we might want to cover:
- kill a majority of nodes
- node won’t join cluster
- permissions on data directory don’t let node start (need to chmod this back) 
- use the wrong advertise-host (other networking problems)
- decommission a majority

jseldess:
@knz, @a-robinson, @andreimatei, PTAL at least at the errors you know about.

andreimatei:
:lgtm:

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[v1.1/common-errors.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kFA9YDJUssdAlCnNh:-L0kFA9YDJUssdAlCnNi:b-greigt) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L60)):*
> ```Markdown
> ## retry transaction
> 
> Messages with the error code `40001` or the string `retry transaction` indicate that a transaction failed because another concurrent or recent transaction accessed the same values.
> ```

s/or the string/and the string

Messages with the error code `40001` and the string `retry transaction` indicate that a transaction failed because it conflicted with another concurrent or recent transaction accessing the same data. The transaction needs to be retried by the the client. See [client-side transaction retries](transactions.html#client-side-transaction-retries).

---

*[v1.1/common-errors.md, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kFlu64CXihnGrILe9:-L0kFlu64CXihnGrILeA:bi8m69i) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L144)):*
> ```Markdown
> ## context deadline exceeded
> 
> This message occurs when a component of CockroachDB gives up because it was relying on another component that has not behaved as expected. To investigate further, look in the node's logs for the root cause where the primary failure is occurring.
> ```

e.g. another CockroachDB node dropped a network connection for some reason (perhaps it crashed).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-:-L0kGD7Y0CxqcajV7_np:bnfp4nl)*

bdarnell:
:lgtm:

---

Reviewed 7 of 7 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[v1.1/common-errors.md, line 28 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJk_4jYT9MVXIt-SW:-L0kJk_506ZvfUTnM71m:b-c6mobh) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L28)):*
> ```Markdown
> 
> - If the node hasn't yet been started, [start the node](start-a-node.html).
> - If you specified a `--host` flag when starting the node, you must include it with all other [`cockroach` commands](cockroach-commands.html).
> ```

The `COCKROACH_HOST` environment variable can be used for this (both COCKROACH_HOST and COCKROACH_PORT should work for all commands except `start`)

---

*[v1.1/common-errors.md, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJyOej7SQThkqZFEs:-L0kJyOej7SQThkqZFEt:b-hz0lk2) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L47)):*
> ```Markdown
> This message indicates that cluster is using TLS encryption to protect network communication, and the client is trying to open a connection without using the required TLS certificates.
> 
> To resolve this issue, use the [`cockroach cert client-create`](create-security-certificates.html) command to generate a client certificate and key for the user trying to connect. For a secure deployment walkthrough, including generating security certificates and connecting clients, see [Manual Deployment](manual-deployment.html).
> ```

We should also link to the per-language tutorials to show how to configure your clients to use these certs (especially important for java which has the PEM-vs-DER issue). 

---

*[v1.1/common-errors.md, line 56 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kK5YWttUwAqc5kt27:-L0kK5YWttUwAqc5kt28:binqxcg) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L56)):*
> ```Markdown
> 
> - [Upgrade the cluster to CockroachDB v1.1](upgrade-cockroach-version.html).
> - Use a [CockroachDB v1.0 binary](../releases/v1.0.6.html) to start the built-in SQL client.
> ```

Doesn't need to be solved in this PR, but it would be nice if we had somewhere to point to for "latest release in the 1.0 branch" instead of linking directly to 1.0.6. 

---

*[v1.1/common-errors.md, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kKbgaeZlSXwB9osEp:-L0kKbgaeZlSXwB9osEq:bp09y8v) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L64)):*
> ```Markdown
> To handle these errors, implement [client-side transaction retries](transactions.html#client-side-transaction-retries).
> 
> ## node belongs to cluster \<cluster ID> but is attempting to connect to a gossip network for cluster \<another cluster ID>
> ```

This can also be caused if you restart a node without a `--join` flag. The solution in this case is to update your scripts to *always* use `--join` and to use the `init` command instead of relying on implicit initialization when the `--join` flag is omitted.

---

*[v1.1/transactions.md, line 55 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJCFoXbNVN1e2wZuC:-L0kJCFoXbNVN1e2wZuD:b1c775i) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/transactions.md#L55)):*
> ```Markdown
> 
> 
> **Retryable Errors** | Errors with the code `40001` or string `retry transaction`, which indicate the transaction failed because another concurrent or recent transaction accessed the same values. To handle these errors, you should [retry the transaction](#client-side-transaction-retries).
> ```

In Java, you access error codes with the `getSQLState()` method instead of `getErrorCode()`. I'm not sure where we should point this out, but it's tripped people up before. I don't know if other languages have similar quirks. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-:-L0kKmy_RaJEL5ds5pEU:bnfp4nl)*

a-robinson:
:lgtm:

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[v1.1/common-errors.md, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kOXPfFhIIIVwd9Lx8:-L0kOXPgJCVZe3JYBpQ6:b-c4b9ow) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L45)):*
> ```Markdown
> ## node is running secure mode, SSL connection required
> 
> This message indicates that cluster is using TLS encryption to protect network communication, and the client is trying to open a connection without using the required TLS certificates.
> ```

s/that cluster/that the cluster/

---

*[v1.1/common-errors.md, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kR0w0qcNSYL3z8Pyt:-L0kR0w0qcNSYL3z8Pyu:bfozqj5) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L138)):*
> ```Markdown
> ## clock synchronization error: this node is more than 500ms away from at least half of the known nodes
> 
> This message indicates that a node has spontaneously shut down because it detected that its clock is out of synch with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default). CockroachDB requires moderate levels of [clock synchronization](recommended-production-settings.html#clock-synchronization) to preserve data consistency, so the node shutting down in this way avoids the risk of consistency anomalies.
> ```

s/synch/sync/

---

*[v1.1/common-errors.md, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJyOej7SQThkqZFEs:-L0kOcv-V4UHWATCZUpH:bpjyzym) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L47)):*

We should also link to the per-language tutorials to show how to configure your clients to use these certs (especially important for java which has the PEM-vs-DER issue). 
</blockquote></details>

+1, this stuff is really confusing to newcomers

---

*[v1.1/common-errors.md, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kKbgaeZlSXwB9osEp:-L0kPIvr3CW0-BAZ2NgW:b-ld8kt5) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L64)):*

This can also be caused if you restart a node without a `--join` flag. The solution in this case is to update your scripts to *always* use `--join` and to use the `init` command instead of relying on implicit initialization when the `--join` flag is omitted.
</blockquote></details>

Can you explain in more detail the situation you're thinking of? I'd expect that case to result in a different error message.

To be specific, in that case I'd expect `tocommit(123) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?` in v1.1, or `client cluster ID foo doesn't match server cluster ID bar` in v1.2, because the issue is that the other nodes in the cluster will try to send requests to the node that was restarted without its data (and thus created a new cluster ID for itself).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-:-L0kR977OuSU8g4llrFR:bnfp4nl)*

bdarnell:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[v1.1/common-errors.md, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kKbgaeZlSXwB9osEp:-L0kZXweH6ZdGs8f5hW8:bxvvosi) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L64)):*

Can you explain in more detail the situation you're thinking of? I'd expect that case to result in a different error message.

To be specific, in that case I'd expect `tocommit(123) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?` in v1.1, or `client cluster ID foo doesn't match server cluster ID bar` in v1.2, because the issue is that the other nodes in the cluster will try to send requests to the node that was restarted without its data (and thus created a new cluster ID for itself).
</blockquote></details>

I think in 1.1 both the `tocommit out of range` and `gossip network for other cluster` errors are both possible for the same underlying issues. I'm not positive, though. In any case, all of these become the rpc "cluster id doesn't match" error in 2.0.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311)*

knz:
:lgtm:

---

Reviewed 7 of 7 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-:-L0kxCbDKqqqqIQ3kinn:bnfp4nl)*

jseldess:
Review status: 6 of 9 files reviewed at latest revision, 9 unresolved discussions.

---

*[v1.1/common-errors.md, line 45 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kOXPfFhIIIVwd9Lx8:-L0k_QykCdAGcjqpISf6:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L45)):*

s/that cluster/that the cluster/
</blockquote></details>

Done.

---

*[v1.1/common-errors.md, line 60 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kFA9YDJUssdAlCnNh:-L0k_T5-T6XWuX8_zNZI:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L60)):*

s/or the string/and the string

Messages with the error code `40001` and the string `retry transaction` indicate that a transaction failed because it conflicted with another concurrent or recent transaction accessing the same data. The transaction needs to be retried by the the client. See [client-side transaction retries](transactions.html#client-side-transaction-retries).
</blockquote></details>

Done.

---

*[v1.1/common-errors.md, line 138 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kR0w0qcNSYL3z8Pyt:-L0kzXi7FbFFBldIbJks:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L138)):*

s/synch/sync/
</blockquote></details>

Done.

---

*[v1.1/common-errors.md, line 144 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kFlu64CXihnGrILe9:-L0kzpTD_iDjMCVZiV8R:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1ce8cfe29d53f59c55775f53beed7582fb6bcff0/v1.1/common-errors.md#L144)):*

e.g. another CockroachDB node dropped a network connection for some reason (perhaps it crashed).
</blockquote></details>

Done.

---

*[v1.1/common-errors.md, line 28 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJk_4jYT9MVXIt-SW:-L0kZ2Oh5-xhWHVISFim:b2vtk5p) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L28)):*

The `COCKROACH_HOST` environment variable can be used for this (both COCKROACH_HOST and COCKROACH_PORT should work for all commands except `start`)
</blockquote></details>

Done, I think. PTAL.

---

*[v1.1/common-errors.md, line 47 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJyOej7SQThkqZFEs:-L0kZC_5tdWiyl0vXZEd:bnu4ani) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L47)):*

+1, this stuff is really confusing to newcomers
</blockquote></details>

Our per-language tutorials are still insecure only, unfortunately. This is the main issue: https://github.com/cockroachdb/docs/issues/631.

@awoods187, any thoughts/updates on whether we can do this work for 2.0? I think we may have decided to just focus on one or two languages/clients?

---

*[v1.1/common-errors.md, line 56 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kK5YWttUwAqc5kt27:-L0kZxlsuwnQ3215HYMb:blnpz2x) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L56)):*

Doesn't need to be solved in this PR, but it would be nice if we had somewhere to point to for "latest release in the 1.0 branch" instead of linking directly to 1.0.6. 
</blockquote></details>

I agree, though I'm not sure how to do that. Will think about it.

---

*[v1.1/common-errors.md, line 64 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kKbgaeZlSXwB9osEp:-L0l3XCMMUnpIRUvjaPp:b2vtk5p) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/common-errors.md#L64)):*

I think in 1.1 both the `tocommit out of range` and `gossip network for other cluster` errors are both possible for the same underlying issues. I'm not positive, though. In any case, all of these become the rpc "cluster id doesn't match" error in 2.0.
</blockquote></details>

Done, I think. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311)*

jseldess:
Review status: 4 of 21 files reviewed at latest revision, 9 unresolved discussions.

---

*[v1.1/transactions.md, line 55 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2311#-L0kJCFoXbNVN1e2wZuC:-L0mJhcJUupk2mu_Lmkw:b68jon9) ([raw file](https://github.com/cockroachdb/docs/blob/5101209e325561e3d65bff317a18f1806e6ffb86/v1.1/transactions.md#L55)):*

In Java, you access error codes with the `getSQLState()` method instead of `getErrorCode()`. I'm not sure where we should point this out, but it's tripped people up before. I don't know if other languages have similar quirks. 
</blockquote></details>

The Java "build a test app" tutorial demonstrates using `getSQLState()`. I added a note there, too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2311)*

PR #2312, opened 2017-12-19
Revise basic hardware recs and add cloud-specific recs

### LINKED ISSUE #1132: Minimum and recommended hardware configuration per cloud provider

### LINKED ISSUE #1356: Microsoft Azure best practices

### LINKED ISSUE #2181: Increase bare minimum hardware requirements
We currently say `1 CPU and 2 GB of RAM per node` is the bare minimum. 

Since some potential customers have started off with 1 CPU and been disappointed, @robert-s-lee suggests we increase to 2 CPUs. 

@bdarnell, @nstewart, @mberhault, thoughts?

jseldess:
Fixes #2181
Fixes #1132
Fixes #1356

### DIFF
--- a/v1.1/deploy-cockroachdb-on-aws-insecure.md
+++ b/v1.1/deploy-cockroachdb-on-aws-insecure.md
@@ -65,10 +65,15 @@ You can create these rules using [Security Groups' Inbound Rules](http://docs.aw
 
 ## Step 2. Create instances
 
-[Create an instance](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/launching-instance.html) for each node you plan to have in your cluster. We [recommend](recommended-production-settings.html#cluster-topology):
+[Create an instance](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/launching-instance.html) for each node you plan to have in your cluster.
 
-- Running at least 3 CockroachDB nodes to ensure survivability.
-- Selecting the same continent for all of your instances for best performance.
+- Run at least 3 nodes to [ensure survivability](recommended-production-settings.html#cluster-topology).
+
+- Use `m` (general purpose), `c` (compute-optimized), or `i` (storage-optimized) [instances](https://aws.amazon.com/ec2/instance-types/), with SSD-backed [EBS volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html) or [Instance Store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html). For example, Cockroach Labs has used `m3.large` instances (2 vCPUs and 7.5 GiB of RAM per instance) for internal testing.
+
+- **Do not** use ["burstable" `t2` instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-instances.html), which limit the load on a single core.
+
+For more details, see [Hardware Recommendations](recommended-production-settings.html#hardware) and [Cluster Topology](recommended-production-settings.html#cluster-topology).
 
 ## Step 3. Synchronize clocks
 
--- a/v1.1/deploy-cockroachdb-on-aws.md
+++ b/v1.1/deploy-cockroachdb-on-aws.md
@@ -65,10 +65,15 @@ You can create these rules using [Security Groups' Inbound Rules](http://docs.aw
 
 ## Step 2. Create instances
 
-[Create an instance](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/launching-instance.html) for each node you plan to have in your cluster. We [recommend](recommended-production-settings.html#cluster-topology):
+[Create an instance](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/launching-instance.html) for each node you plan to have in your cluster.
 
-- Running at least 3 nodes to ensure survivability.
-- Selecting the same continent for all of your instances for best performance.
+- Run at least 3 nodes to [ensure survivability](recommended-production-settings.html#cluster-topology).
+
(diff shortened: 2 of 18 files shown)


bdarnell:
I'm having second thoughts about this because the AWS `t2.medium` instance advertises 2 vCPUs, but throttled to 20% utilization, which would have the same problems that I think we're trying to avoid with this recommendation. Let's keep discussing in #2181.

jseldess:
@bdarnell, @mberhault, please review the revised basic and new cloud-specific hardware recommendations. Before adding them to the cloud tutorials (and replicated to v2.0 docs), I'd like to know what's wrong, what needs expansion, etc. Thanks very much for the help thus far.

bdarnell:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[v1.1/recommended-production-settings.md, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3naCon-7vOxwkIytwm:-L3naCon-7vOxwkIytwn:b-juia8j) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L20)):*
> ```Markdown
> - Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
> 
> - Configurations with odd numbers of replicas are more robust than those with even numbers. Clusters of three and four nodes can each tolerate one node failure and still reach a majority (2/3 and 3/4 respectively), so the fourth replica doesn't add any extra fault-tolerance. To survive two simultaneous failures, you must have five replicas.
> ```

We should clarify that the *number of nodes* and the *replication factor* are two different things. The replication factor (set via zone configs) should be an odd number, while you can use any number of nodes (as long as you have at least as many nodes as your replication factor). 

---

*[v1.1/recommended-production-settings.md, line 35 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3nao159BJF0ePE6xkU:-L3nao16dDJzchBX8zpr:b-6fbnw3) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L35)):*
> ```Markdown
> 
> - At a bare minimum, each node should have **2 GB of RAM and one entire core**. More data, complex workloads, higher concurrency, and faster performance require additional resources.
>     {{site.data.alerts.callout_danger}}Avoid "burstable" or "shared-core" machines that limit the load on a single core.{{site.data.alerts.end}}
> ```

Add "virtual" before "machines" here to clarify that this applies to virtualized cloud scenarios and not bare metal. 

---

*[v1.1/recommended-production-settings.md, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3ncCAvzwN6L4fWYQf7:-L3ncCAvzwN6L4fWYQf8:bwhgphs) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L62)):*
> ```Markdown
>     For example, Cockroach Labs has used `Standard_L4s` VMs (4 vCPUs and 32 GiB of RAM per VM) for internal testing.
> - If you choose local SSD storage, on reboot, the VM can come back with the `ntfs` filesystem. Be sure your automation monitors for this and reformats the disk to the Linux filesystem you chose initially.
> - **Do not** use ["burstable" B-series](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/b-series-burstable) VMs, which limit the load on a single core.
> ```

If we're going to list machine types to stay away from on azure we also need to list A-series (HDD, data corruption) and D-series (irregular disk performance). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-:-L3ncjZDfQ6-HqKXl3Cg:bnfp4nl)*

jseldess:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[v1.1/recommended-production-settings.md, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3naCon-7vOxwkIytwm:-L40q2Q0uAyJN0PH34_r:b-eievsi) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L20)):*

We should clarify that the *number of nodes* and the *replication factor* are two different things. The replication factor (set via zone configs) should be an odd number, while you can use any number of nodes (as long as you have at least as many nodes as your replication factor). 
</blockquote></details>

If it's ok with you, I'm going to revise/expand this section in a follow-up PR. Already have plans to provide sample cluster patterns and more details about fault tolerance, so I assume this section will change at least a bit. 

---

*[v1.1/recommended-production-settings.md, line 35 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3nao159BJF0ePE6xkU:-L40kzjhLJ4fc5Aoe492:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L35)):*

Add "virtual" before "machines" here to clarify that this applies to virtualized cloud scenarios and not bare metal. 
</blockquote></details>

Done.

---

*[v1.1/recommended-production-settings.md, line 62 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2312#-L3ncCAvzwN6L4fWYQf7:-L40k5kc72YmqZJS93DH:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f9dc979cd19ad573d64d793c32d7ef48bbd489be/v1.1/recommended-production-settings.md#L62)):*

If we're going to list machine types to stay away from on azure we also need to list A-series (HDD, data corruption) and D-series (irregular disk performance). 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2312)*

PR #2319, opened 2017-12-21
Be more explicit about feature parity of DC/OS deployment

a-robinson:
This was specifically requested before we get our package updated (https://github.com/mesosphere/universe/pull/1580). Updating the section on scaling a node was also requested, but I'm not yet clear on what actually needs to be done there.

cc @ryadav88

PR #2321, opened 2017-12-21
Update Mesosphere docs to reflect support for newer cockroach versions

a-robinson:
I don't know whether to include the doc in v2.0 or not - you can't
actually run unstable releases using the DC/OS package we provide, but
is it weird if we don't have the doc for it? Will we forget to add it
later? Let me know what you'd like.

The first commit is just https://github.com/cockroachdb/docs/pull/2319.

### DIFF
--- a/_includes/sidebar-data-v1.1.json
+++ b/_includes/sidebar-data-v1.1.json
@@ -928,6 +928,12 @@
               "/${VERSION}/orchestrate-cockroachdb-with-docker-swarm.html",
               "/${VERSION}/orchestrate-cockroachdb-with-docker-swarm-insecure.html"
             ]
+          },
+          {
+            "title": "Mesosphere DC/OS",
+            "urls": [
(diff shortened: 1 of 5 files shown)


a-robinson:
You can hold off on reviewing this for now. My update got merged but I still need to verify I can install the new version from the real package manager and figure out how to specify that you want the old version for the v1.0 doc.

a-robinson:
This is now ready for review, but not ready for merge because I managed to mess up the S3 URLs in the package definition (https://github.com/mesosphere/universe/pull/1587)

a-robinson:
This is ready for both review and merge now that https://github.com/mesosphere/universe/pull/1587 is in.

a-robinson:
The `docker run` command wouldn't work because it'd try to pull `cockroachdb/cockroach:v2.0-alpha.20171218`, which doesn't exist. That's a good catch -- I've changed all versions of the doc to use the `docker_image` variable you added recently.

Issue #2114, opened 2017-10-30
Code blocks: Improving syntax highlighting


kuanluo:
@jseldess originally posted in this [airtable task](https://airtable.com/tblFbdnCTXgR9xAjG/viwHgKuTrS4RlVx87/recDGIBtotLaw22Pe): 
 
1. Rethink the overall contrast of code blocks. Our pages have a white background. Should the blocks be light grey, as they are now? Should they rather be dark (more contrast), for example, https://developer.github.com/v3/guides/getting-started/ and https://devcenter.heroku.com/articles/getting-started-with-python#deploy-the-app? Even better, should we give users the option to choose light or dark, for example, https://developers.google.com/google-apps/tasks/quickstart/python#step_2_install_the_google_client_library (hover over block and change the color theme). 

2. Are we highlighting syntax in a helpful way for SQL, shell, and the various languages we have in examples?

3. Should we include line numbering in cases where we're showing the contents of a file?

4. Should code blocks have a "header" identifying the type of sample or the name of the file, e.g., https://stripe.com/docs/libraries#java-library?

Some explorations below:
- Light/dark background with line numbering
![image](https://user-images.githubusercontent.com/1486483/32181085-f085900c-bd69-11e7-8d7b-bba5feb6cf2d.png)

- Light/dark background with file types on top
![image](https://user-images.githubusercontent.com/1486483/32181116-02e4ba48-bd6a-11e7-824c-d6e739aea3d3.png)

- Option to allow users to pick color modes
![image](https://user-images.githubusercontent.com/1486483/32181181-21419952-bd6a-11e7-884d-30713a637fc2.png)

The full iteration can be view on InVision here: https://invis.io/QBE7BYFDK Commenting directly on InVision welcome!

cc: @Amruta-Ranade @sploiselle @jseldess

sploiselle:
OMG 😍 

My one comment is that I think the dark background could use a brighter blue for strings and a brighter purple for comments (i.e. more contrast).

I also _love_ the idea of letting users choose the mode. I know that some people with different vision abilities bemoan the fact that they're forced to read code with a specific color scheme.

Such great work, @kuanluo!

bdarnell:
[...]

PR #2323, opened 2017-12-27
Syntax highlighting colors

j-low:
__[description]:__
* Re: Issue #2114
* Update syntax highlighting font to use `SourceCodePro-Regular`
* Update syntax highlighting colors


j-low:
@kuanluo @jseldess PTAL

kuanluo:
Hi @j-low, I'm ready to review this, but the link above no longer works. :/ Sorry for taking so long - would you mind resending? Thank you!

jseldess:
@kuanluo, I just reran TeamCity, which will create a new test build for us. Link should be posted shortly. I'll review then as well.

jseldess:
LGTM overall. Thanks, @j-low! 

The font is a clear improvement. The colors look softer now, which I think is fine if that's what you and @kuanluo prefer. The only other change I can detect is that comments are now grey instead of blueish. Again, looks fine to me, if that's what we want. If there's otherwise anything specific to check, please let me know.

kuanluo:
Looks awesome, @j-low! Thank you. 

One question for @jseldess and you - We use the navy blue as one of the main colors for the syntax and it's frequently paired with black. Example: 
image

I wonder if using a more distinguishable color, such as the light blue, would be better? Example:
image

jseldess:
Yes, I agree with @kuanluo that a bit more contrast in such cases would be great.

kuanluo:
Awesome, thanks @jseldess to validate my assumption. @j-low, can we use the light blue instead in the case?

jseldess:
@j-low, that color change looks good, but I just realized that `code` formatted text that's also a link no longer gets the green link coloring. See `INSERT` here for an example: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/30746693c48db86bc872e37f589e7be3bcc29ff3/stable/learn-cockroachdb-sql.html#insert-rows-into-a-table

jseldess:
@j-low, the link color for code-formatted text is different than the link color for regular text. That feels confusing to me. Was that intentional?

kuanluo:
LGTM, @j-low!

PR #2324, opened 2017-12-27
Fixed incremental backup example

### LINKED ISSUE #1894: Update AS OF SYSTEM TIME syntax example
Re: [BACKUP](https://cockroachlabs.com/docs//v1.0/backup.html)

## Issue Description

Some docs updates needed. In incremental backup, AS OF SYSTEM TIME doesn't work if it's placed after the entire command. It needs to be placed after just the BACKUP command. We also need to be explicit that incremental backups can only be created within the 24 hour GC window.

Amruta-Ranade:
Closes #1894

### DIFF
--- a/v1.0/backup.md
+++ b/v1.0/backup.md
@@ -143,10 +143,10 @@ AS OF SYSTEM TIME '2017-06-09 16:13:55.571516+00:00';
 Incremental backups must be based off of full backups you've already created.
 
 ~~~ sql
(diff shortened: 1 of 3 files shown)


dianasaur323:
:lgtm: thank you!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2324#-:-L1SNOE1-pZoDd24FbNa:bro3093)*

PR #2328, opened 2018-01-03
Architecture documentation polish

sploiselle:
@bdarnell Is there anyone from the Core team you'd nominate to review these updates to the architecture docs for technical accuracy?

@jseldess These are largely stylistic updates, but would really like an engineering vouch before moving into final review.


bdarnell:
LGTM. I only looked at the 1.1 versions on the assumption that the 2.0 versions are identical.

---

Reviewed 6 of 12 files at r1.
Review status: 6 of 12 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

*[v1.1/architecture/distribution-layer.md, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yyYIyw1hupf0DrFCH:-L1yyYIyw1hupf0DrFCI:bffox4g) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L20)):*
> ```Markdown
> 
>   - **Efficient scans**: By ordering data, scanning over ranges of data is both simple and fast.
>   - **Lookups with less overhead**: Rather than maintaining the location of each individual key, CockroachDB tracks the address of entire ranges, resulting in much simpler operations.
> ```

Both sorted and hashed maps support low-overhead lookups. The difference between the two is really just the efficient ordered scans. 

---

*[v1.1/architecture/distribution-layer.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yz3zWBo9JBi__O-e-:-L1yz3zWBo9JBi__O-e0:b-xbfl8y) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L40)):*
> ```Markdown
> For this two-level indexing to work, though, every node keeps information on where to locate the `meta1` range.
> 
> Nodes also cache values of the `meta2` ranges it has accessed before, which optimizes access of that data in the future. Whenever a node discovers that its `meta2` cache is invalid for a specific key, the cache is updated by performing a regular read on the `meta2` ranges.
> ```

This cache is used the same way for both meta1 and meta2. 

---

*[v1.1/architecture/distribution-layer.md, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yzcXouayT4iwfVqXI:-L1yzcXouayT4iwfVqXJ:b-rgu60d) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L49)):*
> ```Markdown
> 
> - **Keys** are the table's indexed columns (e.g. your table's `PRIMARY KEY`)
> - **Values** are the non-indexed/stored columns (e.g. the rest of your table's columns)
> ```

This might be a little misleading, because the same data may be stored as a key in one place and a value in another. A table with a secondary index is stored in two spans: The primary span has the PK columns as keys and all other columns as values. The secondary span has the secondary index's columns as keys and the PK columns as values. 

---

*[v1.1/architecture/distribution-layer.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-2HXcF4pVW2FMZbP:-L1z-2HYC8XasYiMbjiL:b-gxpo5) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L61)):*
> ```Markdown
> #### Example
> 
> For a client to read the key `12345`, it would go through this process:
> ```

Should we make this `/Table/51/12345` since all client access in practice goes through a table? 

---

*[v1.1/architecture/distribution-layer.md, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-bZXPubDqfT3tUVa:-L1z-bZXPubDqfT3tUVb:beolmmx) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L66)):*
> ```Markdown
> 2. `meta1` finds the addresses of the `meta2` replicas for keys in your cluster between––for example––`10000` and `19999`, and returns these addresses.
> 3. The Gateway sends the request to the first node in the list, which looks up in its `meta2` range the address of the replicas for the keyspace between `12000` and `12999`.
> 4. Lastly, the Gateway sends the request to the first node in this list, which will process the request because it contains the data you're looking for.
> ```

But the first node may not have the lease for this range, so you may get redirected again to the leaseholder node (the leaseholder information is also cached for future use). 

When the cluster is configured with localities, we use the nearest node in the list instead of the first one. 

---

*[v1.1/architecture/distribution-layer.md, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-vT5nmlJX_opS0vU:-L1z-vT5nmlJX_opS0vV:b8ugnlc) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L68)):*
> ```Markdown
> 4. Lastly, the Gateway sends the request to the first node in this list, which will process the request because it contains the data you're looking for.
> 
> Often times, this process is be optimized if the Gateway had looked up a nearby key before because it would have a cache of the `meta2` address and wouldn't need to get information from `meta1`.
> ```

Maybe we should add the cache lookups as explicit steps above? The cache is used the vast majority of the time so we don't often have to go to the meta1 and meta2 ranges. 

---

*[v1.1/architecture/distribution-layer.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z0kGSGVvqsxEN13u7:-L1z0kGSGVvqsxEN13u8:b-xe3s2n) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L87)):*
> ```Markdown
> ### BatchRequest
> 
> To send KV operations between nodes, the `TxnCoordSender` bundles them into a protobuf known as a `BatchRequest`. Each `BatchRequest` also includes its destination, as well as a pointer to the request's Transaction Record. (On the other side, when a node is replying to a `BatchRequest`, it uses a protobuf––`BatchResponse`.)
> ```

This is `internal/client.Txn`, not `TxnCoordSender`. 

---

*[v1.1/architecture/overview.md, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1CZloPVaGjjjFb-q:-L1z1CZloPVaGjjjFb-r:bpx22va) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/overview.md#L54)):*
> ```Markdown
> Term | Definition
> -----|-----------
> **Consistency** | While "consistency" is a loaded term in the world of database technology, CockroachDB uses the term to simply mean that your data is anomaly-free.
> ```

I think we should preserve the old text here (at least the parts that explicitly acknowledge both the ACID and CAP usages of the word "consistency"). 

---

*[v1.1/architecture/replication-layer.md, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1fhD1d5v3g5CrrSK:-L1z1fhD1d5v3g5CrrSL:b-uij1sn) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L17)):*
> ```Markdown
> High availability requires that your database can tolerate nodes going offline without interrupting service to your application. This means replicating data between nodes to ensure the data remains accessible.
> 
> Ensuring consistency with nodes offline, though, is a challenge many databases fail. To solve this problem, CockroachDB uses a consensus algorithm, which requires a quorum of replicas to agree on any changes to a range before those changes are committed. Because 3 is the smallest number that can achieve quorum (i.e., 2 out of 3), CockroachDB requires 3 nodes to become highly available.
> ```

"at least 3"?

---

*[v1.1/architecture/replication-layer.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1lGEhkB42fgf6h_p:-L1z1lGEhkB42fgf6h_q:btvoz78) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L21)):*
> ```Markdown
> The number of failures that can be tolerated is equal to *(Replication factor - 1)/2*. For example, with 3x replication, one failure can be tolerated; with 5x replication, two failures, and so on. You can control the replication factor at the cluster, database, and table level using [Replication Zones](../configure-replication-zones.html).
> 
> When failures happen, though, CockroachDB waits for a small number of minutes and then begins redistributing data to continue maximizing survivability. This process also works the other way around: when new nodes join your cluster, data automatically rebalances onto it, ensuring your load is evenly distributed.
> ```

There are two levels of failure detection. After a few *seconds*, surviving nodes will pick up any leases held by nodes that have failed and begin serving traffic for those ranges. After a few minutes, we decide that the failed notes are not coming back quickly, so we begin the more expensive process of copying data from the surviving replicas to new nodes to bring the range back up to the desired number of replicas. 

---

*[v1.1/architecture/replication-layer.md, line 64 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z2TTNnMB3pH8A0QzS:-L1z2TTNnMB3pH8A0QzT:br9wr5a) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L64)):*
> ```Markdown
> A single node in the Raft group acts as the Leaseholder, which is the only node that can serve reads or propose writes to the Raft group leader (both actions are received as `BatchRequests` from [`DistSender`](distribution-layer.html#distsender)).
> 
> When serving reads, Leaseholders bypass Raft; all of the Leaseholder's writes must have already achieved quorum, so a second consensus on the same data is unnecessary. This skips the networking round trips required by Raft and greatly increases the speed of reads without sacrificing consistency.
> ```

This isn't really right: writes have "already achieved quorum" whether you're on the leaseholder or not. The real reason for leases has to do with the timestamp cache and read/write conflicts (as discussed in the transaction page). In order to keep the timestamp cache up to date, we must either send all reads through consensus so that every node agrees, or send all reads *and* writes through a single node (the leaseholder) so that node is responsible for the timestamp cache. We do the latter to avoid the network latency associated with extra consensus round trips.

---

*[v1.1/architecture/replication-layer.md, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z6_PSRLNzL39YmdYf:-L1z6_PSRLNzL39YmdYg:bymg47t) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L68)):*
> ```Markdown
> If there is no Leaseholder, any node receiving a request will attempt to become the Leaseholder for the range. To prevent two nodes from acquiring the lease, the requester includes a copy of the last valid lease it had; if another node became the Leaseholder, its request is ignored.
> 
> Once it becomes the Leaseholder for the range, the [Range Descriptor](distribution-layer.html#range-descriptors) is updated and propogated to `meta2`. 
> ```

This is not true; the identity of the leaseholder is only recorded on the range itself (in all its replicas), not in meta2. 

---

*[v1.1/architecture/replication-layer.md, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z6newMDlzK_5JQb6F:-L1z6newMDlzK_5JQb6G:bmx9cke) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L110)):*
> ```Markdown
> ### Replication & Storage Layers
> 
> Committed Raft commands are written to the Raft log and as KV data, both of which are stored in RocksDB on each node.
> ```

I'd refer to the storage layer instead of rocksdb. In theory we could have multiple storage layers with different implementations. 

---

*[v1.1/architecture/transaction-layer.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zABNjJk1CUtgnX3bi:-L1zABNjJk1CUtgnX3bj:b-givvju) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L23)):*
> ```Markdown
> ### Concurrency Control
> 
> To enable transactions that can span your entire cluster (including cross-range and cross-table transactions), CockroachDB optimizes correctness through a somewhat-pessimistic concurrency control system.
> ```

I've learned we should be careful about using the words "optimistic" or "pessimistic". Our system includes elements of both. 

---

*[v1.1/architecture/transaction-layer.md, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zAlxBcCBgo-bVqekZ:-L1zAlxBcCBgo-bVqek_:b8pgn3z) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L34)):*
> ```Markdown
> - A list of all writes related to the transaction (known as Write Intents).
> 
> It's also important to realize that this Transaction Record is also a KV entry, so it benefits from [replication](replication-layer.html), ultimately letting a transaction continue even if a node containing it fails.
> ```

On the other hand, a transaction cannot survive the failure of its gateway node, so this bit of survivability is of limited use.

---

*[v1.1/architecture/transaction-layer.md, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zB8josFoYX6FtjG2U:-L1zB8josFoYX6FtjG2V:bxlxir4) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L88)):*
> ```Markdown
> - Responding to the client
> 
> Though CockroachDB is a distributed, highly available database, if a the gateway node dies during a transaction, the operation will time out and the client will have to re-send the transaction to the cluster.
> ```

Depending on how exactly the gateway node dies, you're likely to get a connection error instead of a timeout. 

---

*[v1.1/architecture/transaction-layer.md, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zBUub4uEyGbSsU4gF:-L1zBUub4uEyGbSsU4gG:bkm7awc) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L98)):*
> ```Markdown
> This timestamp is included for all operations, including those sent to other nodes. When nodes receive requests, they inform their local HLC of the request's timestamp, which enforces distributed consistency by guaranteeing that all data that is read or overwritten is at a timestamp less than the next HLC time (i.e. is "in the past").
> 
> CockroachDB leverages HLC timestamp ordering to optimize read performance, as well. Whenever the read's timestamp is greater than the MVCC value it's reading (i.e., the read always happens "after" the write), we don't need to achieve consensus––the write already did––and can let a single node's replica serve the read operation.
> ```

This single node must also be the leaseholder so it can update the timestamp cache and ensure that future conflicting writes with lower timestamps are not allowed. 

---

*[v1.1/architecture/transaction-layer.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zBc7Eb8GlkmCPrVfo:-L1zBc7Eb8GlkmCPrVfp:bejykav) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L102)):*
> ```Markdown
> #### Max Clock Offset Enforcement
> 
> To ensure correctness among distributed nodes, you can identify a Maximum Clock Offset. Because CockroachDB relies on clock synchronization, nodes run a version of [Marzullo’s algorithm](http://infolab.stanford.edu/pub/cstr/reports/csl/tr/83/247/CSL-TR-83-247.pdf) amongst themselves to measure maximum clock offset within the cluster. If a node's timestamp exceeds the maximum offset, it commits suicide to prevent any potential inconsistencies within the cluster.
> ```

"If the difference between a node's timestamp and its peers exceeds the maximum offset, it commits suicide..."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-:-L1zC5J0fMsmmtNhIVEL:bze9014)*

sploiselle:
Review status: 2 of 12 files reviewed at latest revision, 18 unresolved discussions.

---

*[v1.1/architecture/distribution-layer.md, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yyYIyw1hupf0DrFCH:-L20qrfWMEFduI-6soBk:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L20)):*

Both sorted and hashed maps support low-overhead lookups. The difference between the two is really just the efficient ordered scans. 
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yz3zWBo9JBi__O-e-:-L20r5nOfjO5FgMVDAMp:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L40)):*

This cache is used the same way for both meta1 and meta2. 
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 49 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1yzcXouayT4iwfVqXI:-L20rq6sFDpD4YChLGdJ:b-eowr3y) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L49)):*

This might be a little misleading, because the same data may be stored as a key in one place and a value in another. A table with a secondary index is stored in two spans: The primary span has the PK columns as keys and all other columns as values. The secondary span has the secondary index's columns as keys and the PK columns as values. 
</blockquote></details>

Clarified.

---

*[v1.1/architecture/distribution-layer.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-2HXcF4pVW2FMZbP:-L20sMm7s6dvd_jTYdY-:b-7xk4w3) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L61)):*

Should we make this `/Table/51/12345` since all client access in practice goes through a table? 
</blockquote></details>

Ended up removing this example and generalizing the steps above it.

---

*[v1.1/architecture/distribution-layer.md, line 66 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-bZXPubDqfT3tUVa:-L21e3ovKknKuKwaQjw4:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L66)):*

But the first node may not have the lease for this range, so you may get redirected again to the leaseholder node (the leaseholder information is also cached for future use). 

When the cluster is configured with localities, we use the nearest node in the list instead of the first one. 
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z-vT5nmlJX_opS0vU:-L21e4YD9lOg8RuJBULJ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L68)):*

Maybe we should add the cache lookups as explicit steps above? The cache is used the vast majority of the time so we don't often have to go to the meta1 and meta2 ranges. 
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z0kGSGVvqsxEN13u7:-L20yjqeJo1GLWkjzrnZ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/distribution-layer.md#L87)):*

This is `internal/client.Txn`, not `TxnCoordSender`. 
</blockquote></details>

Done.

---

*[v1.1/architecture/overview.md, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1CZloPVaGjjjFb-q:-L20z10XX75dtLYLzz5x:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/overview.md#L54)):*

I think we should preserve the old text here (at least the parts that explicitly acknowledge both the ACID and CAP usages of the word "consistency"). 
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 17 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1fhD1d5v3g5CrrSK:-L20z7DKqKy6jR8YeoRA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L17)):*

"at least 3"?
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 21 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z1lGEhkB42fgf6h_p:-L210ZzVQb7T6Ij37bfD:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L21)):*

There are two levels of failure detection. After a few *seconds*, surviving nodes will pick up any leases held by nodes that have failed and begin serving traffic for those ranges. After a few minutes, we decide that the failed notes are not coming back quickly, so we begin the more expensive process of copying data from the surviving replicas to new nodes to bring the range back up to the desired number of replicas. 
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 64 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z2TTNnMB3pH8A0QzS:-L21a4At88poY61LqE5W:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L64)):*

This isn't really right: writes have "already achieved quorum" whether you're on the leaseholder or not. The real reason for leases has to do with the timestamp cache and read/write conflicts (as discussed in the transaction page). In order to keep the timestamp cache up to date, we must either send all reads through consensus so that every node agrees, or send all reads *and* writes through a single node (the leaseholder) so that node is responsible for the timestamp cache. We do the latter to avoid the network latency associated with extra consensus round trips.
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 68 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z6_PSRLNzL39YmdYf:-L21a4ec2a9UJQq3om-0:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L68)):*

This is not true; the identity of the leaseholder is only recorded on the range itself (in all its replicas), not in meta2. 
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1z6newMDlzK_5JQb6F:-L21aG7lYfRHwe371brp:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/replication-layer.md#L110)):*

I'd refer to the storage layer instead of rocksdb. In theory we could have multiple storage layers with different implementations. 
</blockquote></details>

Done.

---

*[v1.1/architecture/transaction-layer.md, line 23 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zABNjJk1CUtgnX3bi:-L21admeBPkjaObbpntj:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L23)):*

I've learned we should be careful about using the words "optimistic" or "pessimistic". Our system includes elements of both. 
</blockquote></details>

Done.

---

*[v1.1/architecture/transaction-layer.md, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zAlxBcCBgo-bVqekZ:-L21afBvlKHU9vvU8V2l:bvodsyl) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L34)):*

On the other hand, a transaction cannot survive the failure of its gateway node, so this bit of survivability is of limited use.
</blockquote></details>

I thought the same thing, but Murat Demirbas felt that it was worth mentioning. Clarifying a bit.

---

*[v1.1/architecture/transaction-layer.md, line 88 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zB8josFoYX6FtjG2U:-L21b220gjdC1Q-DQnUi:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L88)):*

Depending on how exactly the gateway node dies, you're likely to get a connection error instead of a timeout. 
</blockquote></details>

Done.

---

*[v1.1/architecture/transaction-layer.md, line 98 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zBUub4uEyGbSsU4gF:-L21dwH1wvI35ylqB3vz:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L98)):*

This single node must also be the leaseholder so it can update the timestamp cache and ensure that future conflicting writes with lower timestamps are not allowed. 
</blockquote></details>

Done.

---

*[v1.1/architecture/transaction-layer.md, line 102 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2328#-L1zBc7Eb8GlkmCPrVfo:-L21dwajyKn_HW6gd4wP:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b45e05877cddd53805a04009748b2e4b51f701b7/v1.1/architecture/transaction-layer.md#L102)):*

"If the difference between a node's timestamp and its peers exceeds the maximum offset, it commits suicide..."
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2328)*

sploiselle:
@bdarnell @a-robinson Can either of you take a quick peek at this section I added, [Acquiring and Maintaining a Lease](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/a72f3d3f126faa0af412d2c6d562888f37b58f86/stable/architecture/replication-layer.html#acquiring-and-maintaining-a-lease)?

a-robinson:
The new section LGTM.

sploiselle:
@jseldess Ready for copy edits whenever you get a chance

v1.1/architecture/replication-layer.md
```diff
@@ -53,17 +62,25 @@ After loading the snapshot, the node gets up to date by replaying all actions fr
...
+
+{{site.data.alerts.callout_info}}CockroachDB uses a single node's timestamp cache for efficiency's sake. This mechanism could be alternatively designed to use a consensus-based timestamp cache and serve reads from any replica, but each read would incur more network round trips.{{site.data.alerts.end}}
+
+#### Acquiring and Maintaining a Lease
+
+Any replica that can communicate with a majority of the range's replicas can become the Leaseholder. A replica establishes itself as the Leaseholder by committing a special lease acquisition log entry through Raft (which is a write and therefore requires consensus). The Leaseholder maintains the lease by continually updating the lease's expiration (which is another write and requires consensus). If the node were to be partitioned away from a majority of the range's replicas, it would no longer be able to extend its lease's expiration.
```
bdarnell: We only update the lease's expiration for a few special ranges (the meta and liveness ranges). For other ranges, the lease includes a reference to the node's liveness record, and we periodically update the expiration of the liveness record (one per node) instead of the leases (one per range). The lease is then considered to be expired when the corresponding node liveness record expires. Atomic test-and-set operations are used to ensure that leases and liveness records stay consistent.

v1.1/architecture/replication-layer.md
```diff
@@ -53,17 +62,25 @@ After loading the snapshot, the node gets up to date by replaying all actions fr
...
 
-When serving reads, Leaseholders bypass Raft; for the Leaseholder's writes to have been committed in the first place, they must have already achieved consensus, so a second consensus on the same data is unnecessary. This has the benefit of not incurring networking round trips required by Raft and greatly increases the speed of reads (without sacrificing consistency).
+If there is no active Leaseholder, any node receiving a request will attempt to become the Leaseholder for the range. To prevent two nodes from acquiring the lease, the requester includes a copy of the last valid lease it had; if another node became the Leaseholder, its request is ignored.
 
-CockroachDB attempts to elect a Leaseholder who is also the Raft group leader, which can also optimize the speed of writes.
+Once it becomes the Leaseholder for the range, the [Range Descriptor](distribution-layer.html#range-descriptors) is updated with the new Leaseholder's address.
```
bdarnell: Range descriptors do not contain information about the leaseholder. This information is maintained in memory on each replica of the range.

v1.1/architecture/sql-layer.md
```diff
@@ -22,72 +22,66 @@ When developers send requests to the cluster, they arrive as SQL statements, but
...
 
-CockroachDB implements a large portion of the ANSI SQL standard to manifest its relational structure. You can view [all of the SQL features CockroachDB supports here](../sql-feature-support.html).
+CockroachDB's relational structure is accessed by our implementation of large portions of the ANSI SQL standard. You can view [all of the SQL features CockroachDB supports here](../sql-feature-support.html).
 
 Importantly, through the SQL API, we also let developers use ACID-semantic transactions just like they would through any SQL database (`BEGIN`, `END`, `ISOLATION LEVELS`, etc.)
 
 ### PostgreSQL Wire Protocol
 
-SQL queries reach your cluster through the PostgreSQL wire protocol. This makes connecting your application to the cluster simple by supporting most PostgreSQL-compatible drivers, as well as many PostgreSQL ORMs, such as GORM (Go) and Hibernate (Java).
+Clients connect to your cluster through the PostgreSQL wire protocol, which means you can simply use most PostgreSQL-compatible drivers, as well as many PostgreSQL ORMs, such as [GORM (Go)](../build-a-go-app-with-cockroachdb-gorm.html) and [Hibernate (Java)](../build-a-java-app-with-cockroachdb-hibernate.html).
```
jseldess: nit: I'd cut `simply`

jseldess:
@sploiselle, any chance you can wrap this one up once the 2.0 dust settles?

sploiselle:
Definitely. Will try to get everything finalized by April 13.

sploiselle:
@bdarnell Could I get you to give a quick look at these two sections I changed?

- [Node Liveness](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/3d1e3fa7fe8158701e8a52f8596d7a6b2808dcc1/stable/architecture/distribution-layer.html#node-liveness)
- [Acquiring and Maintaining a Lease](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/3d1e3fa7fe8158701e8a52f8596d7a6b2808dcc1/stable/architecture/replication-layer.html#acquiring-and-maintaining-a-lease)

bdarnell:
New sections look good to me.

---

Review status: 0 of 14 files reviewed at latest revision, 28 unresolved discussions, all commit checks successful.

---

*[v1.1/architecture/replication-layer.md, line 77 at r4](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L9pl5ZH8PMUMFQC0jVL:-L9pl5ZH8PMUMFQC0jVM:b-2ztfdq) ([raw file](https://github.com/cockroachdb/docs/blob/344ac76b1540d5f7e93c38b617876f9fc49db97e/v1.1/architecture/replication-layer.md#L77)):*
> ```Markdown
> As long as the node can continue to update its liveness record it maintains the lease. Each other replica in the Raft group tracks the lease in memory and cross-references it with the node liveness table to ensure the Leaseholder is still valid.
> 
> If a node detects that a lease for one of its ranges is invalid (i.e. because the Leaseholder could no longer update its node liveness record), this node attempts to become the Leaseholder for the range. To prevent two nodes from acquiring the lease, the requester includes a copy of the last valid lease it was aware of; if another node became the Leaseholder in the interrim, its lease is out of date, and its request to become the Leaseholder is ignored.
> ```

Change the first sentence to start with "If a node wants to do something that requires a lease and it detects that a lease for one of its ranges is invalid". We do not immediately attempt to acquire leases as soon as we discover that a lease has expired; we only do so on demand (so it is normal for a range that no one is touching to go for a while without a lease)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-:-L9plTaMBkOSBPtYHBCO:bji20kg)*

sploiselle:
Review status: 0 of 14 files reviewed at latest revision, 29 unresolved discussions, all commit checks successful.

---

*[v1.1/architecture/distribution-layer.md, line 15 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-15:-L9q6Hzj0Ob9HZISC7Hf:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L15)):*

nits:
- `survivability, but still needs provide` > `survivability but still needs to provide` 
- `to track all of a range's replicas physical locations (i.e., nodes)` > `to track the physical locations (i.e., nodes) of all of a range's replicas`
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 37 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-37:-L9q6JJL6ed59zGn-0bK:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L37)):*

nit: `Nodes also cache values...it has accessed before` > `Nodes also cache values...they have access before`
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 60 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-60:-L9q6JxI--kIAf6coY9H:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L60)):*

nit: We use both `Leaseholder Cache` and `Leaseholder cache`. We should pick one.
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 64 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-64:-L9q6NbE-waQaH2NZlPh:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L64)):*

`and then requesting node` > `and the requesting node`
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 95 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-95:-L9q6OP-A8qU3jSFm-RU:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L95)):*

nit: I'd remove the parentheses and just start with `This retry happens...`.
</blockquote></details>

Done.

---

*[v1.1/architecture/distribution-layer.md, line 167 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBY-r3-167:-L9q6PFVE9HWS5IifDNj:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/distribution-layer.md#L167)):*

Should `range's key's values` be `range's keys' values`?
</blockquote></details>

Done.

---

*[v1.1/architecture/overview.md, line 28 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPyrvJ0LNxdlSUBZ-r3-28:-L9q6Q9r2XXBDYM-HTCv:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/overview.md#L28)):*

Replace `/` with period at end of sentence.
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 73 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPysm_HOAZlD5AVj-r3-73:-L9q6SLLEFajMswZjDkX:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/replication-layer.md#L73)):*

We only update the lease's expiration for a few special ranges (the meta and liveness ranges). For other ranges, the lease includes a reference to the node's liveness record, and we periodically update the expiration of the liveness record (one per node) instead of the leases (one per range). The lease is then considered to be expired when the corresponding node liveness record expires. Atomic test-and-set operations are used to ensure that leases and liveness records stay consistent. 
  
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 77 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPysm_HOAZlD5AVj-r3-77:-L9q6Svo5ALZmsZyrvgF:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/replication-layer.md#L77)):*

Range descriptors do not contain information about the leaseholder. This information is maintained in memory on each replica of the range. 
</blockquote></details>

Done.

---

*[v1.1/architecture/replication-layer.md, line 77 at r4](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L9pl5ZH8PMUMFQC0jVL:-L9q6UAS6x2gS6r5TdRA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/344ac76b1540d5f7e93c38b617876f9fc49db97e/v1.1/architecture/replication-layer.md#L77)):*

Change the first sentence to start with "If a node wants to do something that requires a lease and it detects that a lease for one of its ranges is invalid". We do not immediately attempt to acquire leases as soon as we discover that a lease has expired; we only do so on demand (so it is normal for a range that no one is touching to go for a while without a lease)
</blockquote></details>

Done.

---

*[v1.1/architecture/sql-layer.md, line 43 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2328#-L1xdPysm_HOAZlD5AVk-r3-43:-L9q6VEsC41jN4vQkTRE:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/75ca7a7eaa657b97939d885644da4e83bc3620d9/v1.1/architecture/sql-layer.md#L43)):*

nit: I'd cut `simply`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2328)*

sploiselle:
@jseldess Two sections that could use at least a cursory edit:

- [Node Liveness](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/3ed52e90d16d8f1e09cb9108e7330fd8770cba4c/stable/architecture/distribution-layer.html#node-liveness)
- [Acquiring and Maintaining a Lease](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/3ed52e90d16d8f1e09cb9108e7330fd8770cba4c/stable/architecture/replication-layer.html#acquiring-and-maintaining-a-lease)

jseldess:
@rmloveland, this is an old PR with various improvements to our architecture docs. It never got merged, unfortunately. Can you please take a look and see if it's worth reviving?

rmloveland:
@jseldess yep, added to TODO list - thanks for the ping

sploiselle:
Gonna close this just so it's no longer on my personal list of open PRs. Won't delete the branch so everything is still salvageable if there's any interest in resurrecting it.

PR #2329, opened 2018-01-04
Fix spelling of datacenter and round trip on various pages

a-robinson:
Let me know if you'd prefer "round-trip" to "round trip", but the
misspellings should be fixed either way.


jseldess:
Thanks, @a-robinson! I'd prefer `round-trip`. LGTM once that's used everywhere.

a-robinson:
I'm getting extra grammatical now - I used "round trip" for all nouns and "round-trip" for adjectives.

jseldess:
`round-trip` is definitely correct as an adjective, but there aren't fixed rules about hyphenating compound nouns, so both `round trip` and `round-trip` are correct as a noun. `round trip` is fine for now. The docs writers want to create a more extensive style guide soonish, and this type of convention should be part of the process (cc @Amruta-Ranade and @lhirata).

jseldess:
Sorry for letting this sit so long, @a-robinson. LGTM. Will merge once it passes TeamCity.

a-robinson:
No worries, I forgot about it too :)

PR #2331, opened 2018-01-05
Document difference between `~` and `~*` operators

bdarnell:


bdarnell:
Done. The "edit this page" link on every page is kind of useless when we have to update multiple versions of a page at once. :(  Maybe we should drop that and just have the "report an issue" option.

jseldess:
Thanks. Yeah, I agree that versioning the docs has made the button less useful. But for community edits, I thinks it fine to let users edit one version and then take on the work ourselves to apply the update to other versions.

PR #2333, opened 2018-01-07
Release notes for v1.1.4

jseldess:
Release notes for v1.1.4 release. See https://github.com/cockroachdb/cockroach/issues/21144.

### DIFF
--- a/releases/v1.1.4.md
+++ b/releases/v1.1.4.md
@@ -1,0 +1,39 @@
+---
+title: What&#39;s New in v1.1.4
+toc: false
+summary: Additions and changes in CockroachDB version v1.1.4
+---
+
+## Jan 8, 2018
+
+Get future release notes emailed to you:
+
+<div class="hubspot-install-form install-form-1 clearfix">
+    <script>
+        hbspt.forms.create({
+            css: '',
+            cssClass: 'install-form',
+            portalId: '1753393',
+            formId: '39686297-81d2-45e7-a73f-55a596a8d5ff',
+            formInstanceId: 1,
+            target: '.install-form-1'
+        });
+    </script>
+</div>
+
+### Downloads
+
+<div id="os-tabs" class="clearfix">
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.1.4.darwin-10.9-amd64.tgz"><button id="mac" data-eventcategory="mac-binary-release-notes">Mac</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.1.4.linux-amd64.tgz"><button id="linux" data-eventcategory="linux-binary-release-notes">Linux</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.1.4.windows-6.2-amd64.zip"><button id="windows" data-eventcategory="windows-binary-release-notes">Windows</button></a>
+    <a href="https://binaries.cockroachdb.com/cockroach-v1.1.4.src.tgz"><button id="source" data-eventcategory="source-release-notes">Source</button></a>
+</div>
+
+### SQL Language changes
+
+
+### Bug Fixes
+
+
+### Doc Updates
--- a/releases/v1.1.4.md
+++ b/releases/v1.1.4.md
@@ -30,10 +30,51 @@ Get future release notes emailed to you:
     <a href="https://binaries.cockroachdb.com/cockroach-v1.1.4.src.tgz"><button id="source" data-eventcategory="source-release-notes">Source</button></a>
 </div>
 
-### SQL Language changes
+### SQL Language Changes
 
+- For compatibility with PostgreSQL, the `transaction isolation level`, `transaction priority`, and `transaction status` [session variables](../v1.1/set-vars.html) have been renamed `transaction_isolation`, `transaction_priority`, and `transaction_status`. [#20301](https://github.com/cockroachdb/cockroach/pull/20301)
+- The [`IMPORT`](../v1.1/import.html) statement now supports `201` and `204` return codes from HTTP storage. [#20516](https://github.com/cockroachdb/cockroach/pull/20516)
+
+### Command-Line Changes
+
+- Client [`cockroach` commands](../v1.1/cockroach-commands.html) that use SQL (`cockroach sql`, `cockroach node ls`, etc.) now print a warning if the server is running an older version of CockroachDB than the client. Also, this and other warning messages are now clearly indicated with the "warning:" prefix. [#20945](https://github.com/cockroachdb/cockroach/pull/20945)
+- The [`cockroach sql`](../v1.1/use-the-built-in-sql-client.html) command no longer fails when the server is running a version of CockroachDB that does not support the `sql_safe_updates` [session variable](../v1.1/set-vars.html#supported-variables). [#20945](https://github.com/cockroachdb/cockroach/pull/20945)
+- The [`cockroach sql`](../v1.1/use-the-built-in-sql-client.html) command now performs client-side syntax checking automatically only if the client and server are running the same version of CockroachDB. This restriction will be lifted in CockroachDB v2.0. [#21135](https://github.com/cockroachdb/cockroach/pull/21135)
+- Reduced temporary disk space usage for the `debug compact` command. [#20502](https://github.com/cockroachdb/cockroach/pull/20502)
 
 ### Bug Fixes
 
+- Improved rebalancing to make thrashing back and forth between nodes much less likely, including when [localities](../v1.1/start-a-node.html#locality) have very different numbers of nodes. [#20934](https://github.com/cockroachdb/cockroach/pull/20934)
+- Fixed a scenario where a range that is too big to snapshot can lose availability even with a majority of nodes alive. [#20906](https://github.com/cockroachdb/cockroach/pull/20906)
+- Fixed a replica corruption that could occur if a process crashed in the middle of a range split. [#20939](https://github.com/cockroachdb/cockroach/pull/20939)
+- Fixed an issue where seemingly irrelevant error messages were being returned for certain [`INSERT`](../v1.1/insert.html) statements.[#20855](https://github.com/cockroachdb/cockroach/pull/20855)
+- Crash details are now properly copied to the log file even when a node was started with `--logtostderr` as well as in other circumstances when crash details could be lost previously. [#20842](https://github.com/cockroachdb/cockroach/pull/20842)
+- Fixed an issue where wildcards (`_`, `%`) and `\` escaped characters in `LIKE` patterns were not handled properly. [#20654](https://github.com/cockroachdb/cockroach/pull/20654)
+- Fixed the return type signature of the `array_positions()` [built-in function](../v1.1/functions-and-operators.html). This caused a crash if the function was used in query execution by the distributed execution engine. [#20652](https://github.com/cockroachdb/cockroach/pull/20652)
+- Fixed a crash caused by null [collated strings](../v1.1/collate.html). [#20648](https://github.com/cockroachdb/cockroach/pull/20648)
+- Fixed a crash caused by `NATURAL JOINS` and `USING` in conjunction with a filter. [#20582](https://github.com/cockroachdb/cockroach/pull/20582)
+- Fixed build when new versions of `libgtest` are installed. [#20598](https://github.com/cockroachdb/cockroach/pull/20598)
+- Fixed a crash caused by an edge-case when queries are run against a table undergoing a schema change. [#20541](https://github.com/cockroachdb/cockroach/pull/20541)
+- Fixed a race condition that would result in some queries hanging after [cancellation](../v1.1/cancel-query.html). [#20534](https://github.com/cockroachdb/cockroach/pull/20534)
+- Fixed a crash caused by some corner-case queries containing `ORDER BY`. [#20512](https://github.com/cockroachdb/cockroach/pull/20512)
+- Prevented generating empty sstables at the [storage layer](../v1.1/architecture/storage-layer.html), which could cause corruption. [#20284](https://github.com/cockroachdb/cockroach/pull/20284)
+- Fixed a crash caused by [`cockroach dump`](../v1.1/sql-dump.html) in the presence of reference cycles. [#20283](https://github.com/cockroachdb/cockroach/pull/20283)
+- It is no longer possible to use conflicting `AS OF SYSTEM TIME` clauses in different parts of a query. [#20286](https://github.com/cockroachdb/cockroach/pull/20286)
+- Improved S3 resiliency during [`BACKUP`](../v1.1/backup.html)/[`RESTORE`](../v1.1/restore.html)/[`IMPORT`](../v1.1/import.html). [#21003](https://github.com/cockroachdb/cockroach/pull/21003)
+
+### Performance Improvements
+
+- Improved garbage collection of very large [transactions](../v1.1/transactions.html) and large volumes of abandoned write intents. [#20446](https://github.com/cockroachdb/cockroach/pull/20446)
+
+### Enterprise Edition Changes
+
+- [`BACKUP`](../v1.1/backup.html)/[`RESTORE`](../v1.1/restore.html) jobs now support `201` and `204` return codes from HTTP storage. [#20516](https://github.com/cockroachdb/cockroach/pull/20516)
 
 ### Doc Updates
+
+- Added a tutorial on [orchestrating CockroachDB v1.1 with Mesosphere DC/OS](../v1.1/orchestrate-cockroachdb-with-mesosphere-insecure.html). [#2321](https://github.com/cockroachdb/docs/pull/2321)
+- Improved the [initial steps for troubleshooting a problem](../v1.1/troubleshooting-overview.html), and added a list of [common errors with recommended solutions](../v1.1/common-errors.html). [#2311](https://github.com/cockroachdb/docs/pull/2311)
+- Updated docs styling to increase information density. [#2304](https://github.com/cockroachdb/docs/pull/2304)
+- Documented how to [request and set trial and enterprise license keys for CockroachDB](../v1.1/enterprise-licensing.html). [#2278](https://github.com/cockroachdb/docs/pull/2278)
+- Fixed the placement of `AS OF SYSTEM TIME` in enterprise [`BACKUP` examples](../v1.1/backup.html#examples). [#2324](https://github.com/cockroachdb/docs/pull/2324)
+- Updated the [secure Kubernetes](../v1.1/orchestrate-cockroachdb-with-kubernetes.html) tutorial with RBAC support. [#2281](https://github.com/cockroachdb/docs/pull/2281)
--- a/_config.yml
+++ b/_config.yml
@@ -29,11 +29,11 @@ release_info:
     build_time: 2017/09/14 15:13:51 (go1.8.3)
     start_time: 2017-09-14 13:33:22.046027038 -0400 EDT
   v1.1:
-    name: v1.1.3
-    version: v1.1.3
+    name: v1.1.4
+    version: v1.1.4
     docker_image: cockroachdb/cockroach
-    build_time: 2017/11/27 14:48:26 (go1.8.3)
-    start_time: 2017-11-27 15:10:52.34274101 +0000 UTC
+    build_time: 2018/01/08 14:48:26 (go1.8.3)
+    start_time: 2018-01-08 15:10:52.34274101 +0000 UTC
   v2.0:
     name: v2.0-alpha
     version: v2.0-alpha.20171218
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -1,8 +1,10 @@
 - title: Production Releases
   releases:
+    - date: Jan 8, 2018
+      version: v1.1.4
+      latest: true
     - date: Nov 27, 2017
       version: v1.1.3
-      latest: true
     - date: Nov 2, 2017
       version: v1.1.2
     - date: Oct 19, 2017
--- a/_includes/sidebar-releases.json
+++ b/_includes/sidebar-releases.json
@@ -2,6 +2,12 @@
   "title": "Releases",
   "items": [
     {
+      "title": "v1.1.4",
+      "urls": [
+        "/releases/v1.1.4.html"
+      ]
+    },
+    {
       "title": "v1.1.3",
       "urls": [
         "/releases/v1.1.3.html"
--- a/v1.1/common-errors.md
+++ b/v1.1/common-errors.md
@@ -11,7 +11,6 @@ Topic | Message
 ------|--------
 Client connection | [`connection refused`](#connection-refused)
 Client connection | [`node is running secure mode, SSL connection required`](#node-is-running-secure-mode-ssl-connection-required)
-Client connection | [`unknown variable: "sql_safe_updates"`](#unknown-variable-sql_safe_updates)
 Transactions | [`retry transaction`](#retry-transaction)
 Node startup | [`node belongs to cluster <cluster ID> but is attempting to connect to a gossip network for cluster <another cluster ID>`](#node-belongs-to-cluster-cluster-id-but-is-attempting-to-connect-to-a-gossip-network-for-cluster-another-cluster-id)
 Node configuration | [`clock synchronization error: this node is more than 500ms away from at least half of the known nodes`](#clock-synchronization-error-this-node-is-more-than-500ms-away-from-at-least-half-of-the-known-nodes)
@@ -46,15 +45,6 @@ $ cockroach start [flags]
 This message indicates that the cluster is using TLS encryption to protect network communication, and the client is trying to open a connection without using the required TLS certificates.
 
 To resolve this issue, use the [`cockroach cert client-create`](create-security-certificates.html) command to generate a client certificate and key for the user trying to connect. For a secure deployment walkthrough, including generating security certificates and connecting clients, see [Manual Deployment](manual-deployment.html).
-
-## unknown variable: "sql_safe_updates"
-
-This message indicates that the [built-in SQL client](use-the-built-in-sql-client.html) (`cockroach sql`) is using CockroachDB v1.1.x or later to connect to a node running CockroachDB v1.0.x.
-
-To resolve this issue, do one of the following:
-
-- [Upgrade the cluster to CockroachDB v1.1](upgrade-cockroach-version.html).
-- Use a [CockroachDB v1.0 binary](../releases/v1.0.6.html) to start the built-in SQL client.
 
 ## retry transaction
 
--- a/v2.0/common-errors.md
+++ b/v2.0/common-errors.md
@@ -11,7 +11,6 @@ Topic | Message
 ------|--------
 Client connection | [`connection refused`](#connection-refused)
 Client connection | [`node is running secure mode, SSL connection required`](#node-is-running-secure-mode-ssl-connection-required)
-Client connection | [`unknown variable: "sql_safe_updates"`](#unknown-variable-sql_safe_updates)
 Transactions | [`retry transaction`](#retry-transaction)
 Node startup | [`node belongs to cluster <cluster ID> but is attempting to connect to a gossip network for cluster <another cluster ID>`](#node-belongs-to-cluster-cluster-id-but-is-attempting-to-connect-to-a-gossip-network-for-cluster-another-cluster-id)
 Node configuration | [`clock synchronization error: this node is more than 500ms away from at least half of the known nodes`](#clock-synchronization-error-this-node-is-more-than-500ms-away-from-at-least-half-of-the-known-nodes)
@@ -46,15 +45,6 @@ $ cockroach start [flags]
 This message indicates that the cluster is using TLS encryption to protect network communication, and the client is trying to open a connection without using the required TLS certificates.
 
 To resolve this issue, use the [`cockroach cert client-create`](create-security-certificates.html) command to generate a client certificate and key for the user trying to connect. For a secure deployment walkthrough, including generating security certificates and connecting clients, see [Manual Deployment](manual-deployment.html).
-
-## unknown variable: "sql_safe_updates"
-
-This message indicates that the [built-in SQL client](use-the-built-in-sql-client.html) (`cockroach sql`) is using CockroachDB v1.1.x or later to connect to a node running CockroachDB v1.0.x.
-
-To resolve this issue, do one of the following:
-
-- [Upgrade the cluster to CockroachDB v1.1](upgrade-cockroach-version.html).
-- Use a [CockroachDB v1.0 binary](../releases/v1.0.6.html) to start the built-in SQL client.
 
 ## retry transaction
 


jseldess:
```
git log v1.1.3...b794b52cbfffa2340cdaabf1c33be716ebde1db4

*   commit b794b52cbfffa2340cdaabf1c33be716ebde1db4
|\  Merge: 268fe806d d808f067b
| | Author: kena <knz@users.noreply.github.com>
| | Date:   Mon Jan 1 03:05:44 2018 +0100
| | 
| |     Merge pull request #21135 from knz/20171230-check-syntax
| |     
| |     cherry-pick 1.1: cli/sql: avoid check_syntax on client/server version mismatch
| | 
| * commit d808f067ba6a6e378cdd682af43b71b7432d30a4
|/  Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
|   Date:   Sat Dec 30 23:42:10 2017 +0100
|   
|       cli/sql: avoid check_syntax on client/server version mismatch
|       
|       This is a toned down cherry-pick of #21119, as advised by #19445.
|       
|       Release note (cli change): client-side syntax checking is now
|       only automatically enabled if the client and server are running
|       the same version of CockroachDB. This restriction will be
|       lifted in CockroachDB 2.0.
|   
*   commit 268fe806db4b43fee5f00d621db095687a76ddd5
|\  Merge: 666df6cce 41ab818e9
| | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | Date:   Sun Dec 24 09:18:30 2017 -0500
| | 
| |     Merge pull request #21034 from tschottdorf/cp-20989
| |     
| |     cherrypick-1.1: gossip: better reporting for a tripped assertion
| | 
| * commit 41ab818e9a613c0125df642cc2005fbc808756e6
| | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | Date:   Thu Dec 21 10:31:57 2017 -0500
| | 
| |     gossip: better reporting for a tripped assertion
| |     
| |     Touches #20986.
| |     
| |     Release note: None
| |   
* |   commit 666df6cce121964d101b90e3e0d11705e5c6d9c9
|\ \  Merge: 6ac21a04a dbaa5489b
| | | Author: Matt Jibson <matt.jibson@gmail.com>
| | | Date:   Thu Dec 21 16:04:06 2017 -0500
| | | 
| | |     Merge pull request #21003 from mjibson/cherrypick-19556
| | |     
| | |     cherrypick-1.1: ccl/storageccl: retry s3 region fetch
| | | 
| * | commit dbaa5489b0f61245247fb4ce01f5c3fc3a925cf8
| | | Author: Matt Jibson <matt.jibson@gmail.com>
| | | Date:   Thu Oct 26 03:24:09 2017 -0400
| | | 
| | |     ccl/storageccl: retry s3 region fetch
| | |     
| | |     Retry an S3 operation that has been causing transient errors. If a
| | |     503 is returned, we may need to slow down our request rate. Although
| | |     the retry package does wait a bit, it is only some 10s of ms. We
| | |     instead wait 5s on these errors to attempt to slightly decrease the
| | |     request rate. This 5s is purely a guess, but it is also a low enough
| | |     time that it won't matter too much if we are waiting longer than
| | |     minimally needed.
| | |     
| | |     For now put this only in the place we have seen the failure. We can
| | |     optionally retry the other locations if we see problems there. This
| | |     is the same approach we used for Azure and GCE with good results.
| | |     
| | |     Fixes #19502
| | |   
* | |   commit 6ac21a04afffccd9ac85f0a0f451d13299fce9ee
|\ \ \  Merge: 2cded8b6a 02788d830
| | | | Author: kena <knz@users.noreply.github.com>
| | | | Date:   Thu Dec 21 14:52:15 2017 +0100
| | | | 
| | | |     Merge pull request #20945 from knz/20171220-cherrypick-20935
| | | |     
| | | |     cherry-pick 1.1: cli, cli/sql: version checks
| | | | 
| * | | commit 02788d830bd7b555f731ea5cbf4e969bb9a7384f
| | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | Date:   Thu Dec 21 14:42:36 2017 +0100
| | | | 
| | | |     cli/interactive_tests: fix typo in test controller
| | | |     
| | | |     Release note: None
| | | | 
| * | | commit 959f629180c866980b4bcc9d9f228050ca3a5078
| | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | Date:   Wed Dec 20 21:19:26 2017 +0100
| | | | 
| | | |     cli: add a check that client version <= server version
| | | |     
| | | |     This patch introduces two features:
| | | |     
| | | |     - it ensures that a v1.1+ `cockroach` client connecting to a v1.0
| | | |       server properly recognizes that the server is running v1.0 (prior to
| | | |       this patch, the version strings from a v1.0 server were not
| | | |       recognized).
| | | |     
| | | |     - it causes a warning message to be printed if the client is newer
| | | |       than the server.
| | | |     
| | | |     For example:
| | | |     
| | | |     ```
| | | |     kena@kenax ~/cockroach % ./cockroach sql --insecure
| | | |     
| | | |     warning: server version older than client! proceed with caution; some features may not be available.
| | | |     
| | | |     root@:26257/>
| | | |     ```
| | | |     
| | | |     Release note (cli change): client commands that use SQL (including
| | | |     `cockroach sql`, `cockroach node ls`, etc) now print a warning if the
| | | |     server version is older than the client.
| | | | 
| * | | commit c8d5f02783c4cf05e626156b5e4bee849283fd5b
| | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | Date:   Wed Dec 20 19:25:54 2017 +0100
| | | | 
| | | |     cli: de-emphasize the seriousness of errors in ancillary SQL client services
| | | |     
| | | |     The code that implements the SQL connection object utilized by various
| | | |     CLI utilities (including, but not limited to, `cockroach sql`) provide
| | | |     ancillary services for the user's convenience, like checking whether
| | | |     the client and the server version are the same, etc.
| | | |     
| | | |     When these checks fail, the failure should not indicate malfunction --
| | | |     they can simply indicate that the utility was run against a previous
| | | |     CockroachDB version that does not yet support some service, or that
| | | |     `cockroach sql` is being run against a PostgreSQL server.
| | | |     
| | | |     Prior to this patch, the errors due to ancillary failures were
| | | |     reported with serious-sounding messages, often starting with the word
| | | |     "unable" or "error".
| | | |     
| | | |     This patch de-emphasizes them by prefixing them with the string
| | | |     "warning:".
| | | |     
| | | |     Release note (cli change): some warning messages are now more clearly
| | | |     indicated with the "warning:" prefix.
| | | | 
| * | | commit 1d22b22757ca4232cee3721934a92b70fc534dbc
| | | | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | | | Date:   Wed Dec 20 19:22:12 2017 +0100
| | | | 
| | | |     cli/sql: do not fail if sql_safe_updates cannot be set
| | | |     
| | | |     Prior to this patch, an interactive SQL shell (`cockroach sql`) would
| | | |     error out if connecting to a server which doesn't support the session
| | | |     variable `sql_safe_updates` -- e.g. CockroachDB 1.0 or PostgreSQL
| | | |     itself.
| | | |     
| | | |     This patch ensures that a failure is only reported as a warning.
| | | |     
| | | |     Release note (cli change): `cockroach sql` does not fail any more when
| | | |     the server does not support the `sql_safe_updates` session variable.
| | | |   
* | | |   commit 2cded8b6a855397f158bae29a50ae6a11e7a72c3
|\ \ \ \  Merge: 22440455c f01eb24f5
| | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | Date:   Thu Dec 21 00:59:47 2017 -0500
| | | | | 
| | | | |     Merge pull request #20934 from a-robinson/cherrypick_20241
| | | | |     
| | | | |     cherrypick-1.1: storage: avoid replica thrashing when localities are different sizes
| | | | | 
| * | | | commit f01eb24f5a4862d74b7d335230ce1c3b2b25d7d9
| | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | Date:   Fri Dec 15 13:18:57 2017 -0500
| | | | | 
| | | | |     storage: Avoid replica thrashing when localities are different sizes
| | | | |     
| | | | |     Fixes #20241
| | | | |     
| | | | |     Release note (bug fix): avoid rebalance thrashing when localities have
| | | | |     very different numbers of nodes
| | | | | 
| * | | | commit 06f7d7664796558f4e8df1faab63e0be5f8a5fde
| | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | Date:   Fri Dec 15 12:51:45 2017 -0500
| | | | | 
| | | | |     storage: Always simulate RemoveTarget when rebalancing
| | | | |     
| | | | |     Skipping the simulation when raftStatus.Progress is nil can make
| | | | |     for undesirable thrashing of replicas, as seen when testing #20241.
| | | | |     It's better to run the simulation without properly filtering replicas
| | | | |     than to not run it at all.
| | | | |     
| | | | |     Release note: None
| | | | | 
| * | | | commit fd46e3f3949efcf0cafe215fa1ef51a4f8bbb6eb
| | | | | Author: Alex Robinson <alexdwanerobinson@gmail.com>
| | | | | Date:   Wed Dec 13 16:50:51 2017 -0500
| | | | | 
| | | | |     storage: Fix simulation of rebalance removals to actually remove targets
| | | | |     
| | | | |     If the first target attempted was rejected due to the simulation
| | | | |     claiming that it would be immediately removed, we would reuse the
| | | | |     modified `rangeInfo.Desc.Replicas` that had the target added to it,
| | | | |     messing with future iterations of the loop.
| | | | |     
| | | | |     Also, we weren't properly modifying the `candidates` slice, meaning that
| | | | |     we could end up trying the same replica multiple times.
| | | | |     
| | | | |     Release note (bug fix): Improve data rebalancing to make thrashing
| | | | |     back and forth between nodes much less likely.
| | | | | 
| * | | | commit 5dbe05eab5a45d0d969c0bad3472ea5f877d7c2f
| | | | | Author: Tamir Duberstein <tamird@gmail.com>
| | | | | Date:   Wed Oct 25 20:47:26 2017 -0400
| | | | | 
| | | | |     storage: allocate the slice just once
| | | | | 
| * | | | commit 21d4cd9a90f6a35820315396941b058e169e7903
| | | | | Author: a6802739 <songhao9021@gmail.com>
| | | | | Date:   Thu Oct 12 16:33:24 2017 +0800
| | | | | 
| | | | |     storage : fix bug for the testrace of TestSystemZoneConfigs
| | | | |     
| | | | |     Fixes #19180.
| | | | |     Fixes #19207.
| | | | | 
| * | | | commit 5dded67dd86919be3555e8a7f938509df4ccaa50
| | | | | Author: a6802739 <songhao9021@gmail.com>
| | | | | Date:   Fri Sep 8 16:33:50 2017 +0800
| | | | | 
| | | | |     storage: Consider which replica will be removed when adding a replica to improve balance
| | | | | 
| * | | | commit 489326825997fd76903202f4708cc2095ae0200c
| | | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | | Date:   Sun Oct 8 16:52:22 2017 -0400
| | | | | 
| | | | |     storage: updating local Store before gossip should not crash
| | | | |     
| | | | |     Updating a target store write stats immediately after rebalancing was
| | | | |     recently addressed in #18425. With that change, if `updateLocalStoreAfterRebalance`
| | | | |     is called before the `StorePool` had seen the `StoreDescriptor` in gossip,
| | | | |     it will trigger a NPE. This change fixes this by making the update a
| | | | |     no-op if the descriptor has yet to be seen in gossip.
| | | | | 
| * | | | commit 6ef56ae8cac5853f956b55ccfbf57ee48c6524f3
|  / / /  Author: a6802739 <songhao9021@gmail.com>
| | | |   Date:   Mon Sep 11 16:42:48 2017 +0800
| | | |   
| | | |       storage: immediately update target store write stats after rebalance
| | | |   
* | | |   commit 22440455ce7e423025f07f2a683de5578f3d47b5
|\ \ \ \  Merge: 02bdfdfae 4f44c54b5
| |/ / /  Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
|/| | |   Date:   Wed Dec 20 14:48:49 2017 -0500
| | | |   
| | | |       Merge pull request #20906 from nvanbenschoten/nvanbenschoten/cherrypick-20589
| | | |       
| | | |       cherry-pick-1.1: storage: add permitLargeSnapshots flag to replica
| | | | 
| * | | commit 4f44c54b51a48e15eddabc996eb1297f9dbd9750
| | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | Date:   Sat Dec 9 04:09:46 2017 -0500
| | | | 
| | | |     storage: add permitLargeSnapshots flag to replica
| | | |     
| | | |     In a privately reported user issue, we've seen that [our attempts](https://github.com/cockroachdb/cockroach/pull/7788)
| | | |     at [preventing large snapshots](https://github.com/cockroachdb/cockroach/issues/7581)
| | | |     can result in replica unavailability. Our current approach to limiting
| | | |     large snapshots assumes is that its ok to block snapshots indefinitely
| | | |     while waiting for a range to first split. Unfortunately, this can create
| | | |     a dependency cycle where a range requires a snapshot to split (because it
| | | |     can't achieve an up-to-date quorum without it) but isn't allowed to perform
| | | |     a snapshot until its size is reduced below the threshold. This can result
| | | |     in unavailability even when a majority of replicas remain live.
| | | |     
| | | |     Currently, we still need this snapshot size limit because unbounded snapshots
| | | |     can result in OOM errors that crash entire nodes. However, once snapshots
| | | |     are streamed from disk to disk, never needing to buffer in-memory on the
| | | |     sending or receiving side, we should be able to remove any snapshot size
| | | |     limit (see #16954).
| | | |     
| | | |     As a holdover, this change introduces a `permitLargeSnapshots` flag on a
| | | |     replica which is set when the replica is too large to snapshot but observes
| | | |     splits failing. When set, the flag allows snapshots to ignore the size
| | | |     limit until the snapshot goes through and splits are able to succeed
| | | |     again.
| | | |     
| | | |     Release note (bug fix): Fixed a scenario where a range that is too big
| | | |     to snapshot can lose availability even with a majority of nodes alive.
| | | | 
| * | | commit 6adf047196e5fd6f253e550b8ddf1bcc92fa1872
| | | | Author: Nathan VanBenschoten <nvanbenschoten@gmail.com>
| | | | Date:   Sat Dec 9 01:38:35 2017 -0500
| | | | 
| | | |     storage: respect context deadline in idAllocator.Allocate
| | | |     
| | | |     Release note: None
| | | |   
* | | |   commit 02bdfdfae41166ba2f30288380f1e0d7e132a625
|\ \ \ \  Merge: 5fe44557e d2e0bec99
| |/ / /  Author: Ben Darnell <ben@bendarnell.com>
|/| | |   Date:   Wed Dec 20 13:59:42 2017 -0500
| | | |   
| | | |       Merge pull request #20939 from bdarnell/cherrypick-split-atomic
| | | |       
| | | |       cherrypick-1.1: storage: Write HardState atomically with committing splits
| | | | 
| * | | commit d2e0bec9909f9ffd5e8c7d4389d712a7757e12a9
|/ / /  Author: Ben Darnell <ben@cockroachlabs.com>
| | |   Date:   Wed Dec 13 17:24:39 2017 -0500
| | |   
| | |       storage: Write HardState atomically with committing splits
| | |       
| | |       Prior to this change, an ill-timed crash (between applying the raft
| | |       command and calling splitPostApply) would leave the replica in a
| | |       persistently broken state (no HardState).
| | |       
| | |       Found via jepsen.
| | |       
| | |       Fixes #20629
| | |       Fixes #20494
| | |       
| | |       Release note (bugfix): Fixed a replica corruption that could occur if
| | |       a process crashed in the middle of a range split.
| | |   
* | |   commit 5fe44557e8812ed372b323726e7bcdb5e0d7324d
|\ \ \  Merge: c03511030 f362da8ac
| | | | Author: Richard Wu <richardwu1997@gmail.com>
| | | | Date:   Tue Dec 19 13:45:08 2017 -0500
| | | | 
| | | |     Merge pull request #20855 from richardwu/fix-err-indexes-1.1
| | | |     
| | | |     1.1 fix/cherry-pick: fix error index alignment in multiple partial distSender batches
| | | | 
| * | | commit f362da8acd7ea5539b701f067f16dfc65efa303d
|/ / /  Author: Richard Wu <richardwu1997@gmail.com>
| | |   Date:   Mon Dec 18 18:26:19 2017 -0500
| | |   
| | |       1.1 fix/cherry-pick: fix error index alignment in multiple partial
| | |       distSender batches
| | |       
| | |       Release notes: (bug fix): fixed an issue where seemingly irrelevant
| | |       error messages were being returned for certain insert statements.
| | |   
* | |   commit c03511030c9eee85a4588e8f738de376f8a4e8e2
|\ \ \  Merge: 59721409f f529c4206
| | | | Author: kena <knz@users.noreply.github.com>
| | | | Date:   Tue Dec 19 16:18:03 2017 +0100
| | | | 
| | | |     Merge pull request #20842 from knz/20171218-cherrypick-20839
| | | |     
| | | |     cherry-pick 1.1: log: ensure panic details are logged to file even when stderr is not redirected
| | | | 
| * | | commit f529c42062a69e5abb6be868ed330c209f24b949
|/ / /  Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | |   Date:   Mon Dec 18 23:33:38 2017 +0100
| | |   
| | |       log: ensure panic details are logged to file even when stderr is not redirected
| | |       
| | |       Prior to this patch, panic details would not be captured to logs
| | |       unless the log file was forcing redirection of stderr. This was
| | |       notoriously not the case when passing `--logtostderr` or
| | |       `--logtostderr=INFO`.  (In contrast, `--logtostderr=WARNING` would
| | |       redirect stderr and cause the panic details to be properly captured.)
| | |       
| | |       This patch fixes the issue by capturing the panic details every time
| | |       stderr is not redirected.
| | |       
| | |       Release note (bug fix): the crash details are now properly copied to
| | |       the log file when starting a node with `--logtostderr` (and in some
| | |       other circumstances where they could be lost previously).
| | |   
* | |   commit 59721409fa340497db440d899855362b684dedea
|\ \ \  Merge: 999ecbda7 f65062730
| | | | Author: Richard Wu <richardwu1997@gmail.com>
| | | | Date:   Tue Dec 12 15:28:37 2017 -0500
| | | | 
| | | |     Merge pull request #20654 from richardwu/fix-like-match-v1.1
| | | |     
| | | |     1.1 fix: sql: fix LIKE with '_...%', '%..._' and '\' escaped character patterns
| | | | 
| * | | commit f6506273039ae8a791498a80e7f8f780129ea4bb
| | | | Author: Richard Wu <richardwu1997@gmail.com>
| | | | Date:   Tue Dec 12 11:20:09 2017 -0500
| | | | 
| | | |     1.1 fix: sql: fix LIKE with '_...%', '%..._' and '\' escaped character patterns
| | | |     
| | | |     This cherry-pick (with modifications) fixes LIKE comparison operators
| | | |     where we did not handle single wildcards ('_') properly in the optimized
| | | |     like pathway whenever we had a wildcard ('%') at the beginning or end.
| | | |     
| | | |     It also fixes how we handle '\' escaped characters with LIKE patterns.
| | | |     This fix should align Cockroach's LIKE pattern matching with Postgres.
| | | |     
| | | |     The original PR that fixed this in 2.0 is #20600.
| | | |     
| | | |     Release note: fixed an issue where wildcards ('_', '%') and '\' escaped
| | | |     characters in LIKE patterns were not handled properly.
| | | |   
* | | |   commit 999ecbda714cf62993e3f5f27d48f34a8a0d6110
|\ \ \ \  Merge: dc3bb421d 8031d1af7
| |/ / /  Author: Joey Pereira <legojoey17@gmail.com>
|/| | |   Date:   Tue Dec 12 13:56:56 2017 -0500
| | | |   
| | | |       Merge pull request #20652 from lego/1_1-20171212-fix-array-positions-returntype
| | | |       
| | | |       1.1: sql: Fix return type signature of array_positions built-in
| | | | 
| * | | commit 8031d1af7b78764dd94d60a37a0baf3bd1dd4063
|/ / /  Author: Joey Pereira <joey@pereira.io>
| | |   Date:   Tue Dec 12 13:00:47 2017 -0500
| | |   
| | |       sql: Fix return type signature of array_positions built-in
| | |       
| | |       This error was fixed in #20524, and just the fix is
| | |       being backported to 1.1.
| | |       
| | |       The following type signature was changed:
| | |       
| | |           array_positions(<TYPE>[], <TYPE>) now returns a INT[] (previously
| | |               <TYPE>[])
| | |       
| | |       Due to the incorrect type signature returned, a query that contains this
| | |       function would cause a panic in distSQL, as follows:
| | |       
| | |           panic: invalid datum type given: int[], expected string[]
| | |       
| | |           goroutine 336 [running]:
| | |           github.com/cockroachdb/cockroach/pkg/sql/sqlbase.DatumToEncDatum(0xf, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc4202ff0e8, 0x6673720, 0xc420428450, ...)
| | |             /Users/joey/.go/src/github.com/cockroachdb/cockroach/pkg/sql/sqlbase/encoded_datum.go:122 +0x292
| | |           github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*ProcOutputHelper).EmitRow(0xc420c04000, 0x6665820, 0xc42012e120, 0xc42106a9a0, 0x3, 0x3, 0x3, 0x0, 0x0)
| | |             /Users/joey/.go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/processors.go:286 +0xa79
| | |           github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*tableReader).Run(0xc420c04000, 0x6665820, 0xc42012e120, 0xc420669398)
| | |             /Users/joey/.go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/tablereader.go:214 +0x412
| | |           created by github.com/cockroachdb/cockroach/pkg/sql/distsqlrun.(*Flow).Start
| | |             /Users/joey/.go/src/github.com/cockroachdb/cockroach/pkg/sql/distsqlrun/flow.go:394 +0x3d7
| | |       
| | |       Release note (bug fix): Fixed the return type signature of the
| | |       array_positions built-in funtion. This caused a panic if the
| | |       function was used in query executing in the distributed execution
| | |       engine.
| | |   
* | |   commit dc3bb421d8c00ba2375fc2b07ef7ae408b60065e
|\ \ \  Merge: 26243fc8b e48524429
| | | | Author: Jordan Lewis <jordanthelewis@gmail.com>
| | | | Date:   Tue Dec 12 11:07:04 2017 -0500
| | | | 
| | | |     Merge pull request #20648 from jordanlewis/cp-col-null
| | | |     
| | | |     cherrypick-1.1: sql: bugfix to evalling null as collated string
| | | | 
| * | | commit e48524429b22c761df42c69e5aaebd06cd1275ee
| | |/  Author: Jordan Lewis <jordanthelewis@gmail.com>
| |/|   Date:   Mon Dec 11 17:30:15 2017 -0500
| | |   
| | |       sql: bugfix to evalling null as collated string
| | |       
| | |       Previously, evaluating a prepared statement where a placeholder was
| | |       directly collated with the `COLLATE` operator could panic if the input
| | |       was null.
| | |       
| | |       Release note (bug fix): Fix a panic with null collated strings.
| | |   
* | |   commit 26243fc8b8271638fa3e944acf55ed186e863650
|\ \ \  Merge: 235cc62c6 e4bcc41bb
| |/ /  Author: Matt Jibson <matt.jibson@gmail.com>
|/| |   Date:   Tue Dec 12 04:40:40 2017 -0500
| | |   
| | |       Merge pull request #20516 from mjibson/cherrypick-20027
| | |       
| | |       cherrypick-1.1: storageccl: support 201 and 204 return codes
| | | 
| * | commit e4bcc41bbf12ff2aaddee8920e13f4686454b6da
|/ /  Author: Matt Jibson <matt.jibson@gmail.com>
| |   Date:   Mon Nov 13 16:57:12 2017 -0500
| |   
| |       storageccl: support 201 and 204 return codes
| |       
| |       Also correctly close response bodies when unused.
| |       
| |       Fixes #20017
| |   
* |   commit 235cc62c6cac65fe47b20e8933ed99d63cd64f56
|\ \  Merge: 3361b4d00 487f2ba83
| | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | Date:   Mon Dec 11 13:59:06 2017 -0500
| | | 
| | |     Merge pull request #20577 from tschottdorf/cp11-rocksdb-emptydir
| | |     
| | |     cherrypick-1.1: cli: check existing instance in RocksDB debug
| | | 
| * | commit 487f2ba83e53afb39a732b70bbe1d1d7ed076f7e
| | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | Date:   Wed Dec 6 09:43:04 2017 -0500
| | | 
| | |     cli: check existing instance in RocksDB debug
| | |     
| | |     Running, for example, `./cockroach debug keys no-such-dir` would result
| | |     in a new RocksDB instance to be initialized at `no-such-dir`. This is
| | |     confusing and annyoing; instead, return an error.
| | |     
| | |     Release note: None
| | |   
* | |   commit 3361b4d00a24e2499edb95067c5514d69a69ca12
|\ \ \  Merge: f782e38f4 d1814df92
| | | | Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | | | Date:   Mon Dec 11 12:48:29 2017 -0500
| | | | 
| | | |     Merge pull request #20576 from tschottdorf/cp11-trace-schema-changer
| | | |     
| | | |     cherrypick-1.1: sql: trace schema changes
| | | | 
| * | | commit d1814df92c63be9df3b5f06340ed43384b337e95
| |/ /  Author: Tobias Schottdorf <tobias.schottdorf@gmail.com>
| | |   Date:   Fri Dec 8 12:29:37 2017 -0500
| | |   
| | |       sql: trace schema changes
| | |       
| | |       Motivated by the fact that previously, `DROP TABLE` did not allow any
| | |       kind of introspection. Now, at least there's something you can look
| | |       at in `/debug/requests` to see activity.
| | |       
| | |       Touches #19004
| | |       
| | |       Release note: None
| | |   
* | |   commit f782e38f4a5666cb9a14200a1df0bad56c454154
|\ \ \  Merge: 1ee122c05 a47716a1c
| | | | Author: RaduBerinde <radu@cockroachlabs.com>
| | | | Date:   Sun Dec 10 15:56:50 2017 -0500
| | | | 
| | | |     Merge pull request #20582 from RaduBerinde/join-plan-panic-1.1
| | | |     
| | | |     1.1 fix: distsql: fix panic during join planning
| | | | 
| * | | commit a47716a1c14492b86b668fbdf1a482d69d0faa82
| |/ /  Author: Radu Berinde <radu@cockroachlabs.com>
| | |   Date:   Fri Dec 8 16:25:33 2017 -0500
| | |   
| | |       distsql: fix panic during join planning
| | |       
| | |       This change fixes a panic that was based on a faulty assumption: that we either
| | |       have merged columns (NATURAL JOIN, USING) or we have no merged columns and an ON
| | |       condition. This seems true based on the syntax, but for inner joins filters
| | |       (WHERE) get pushed down and become ON conditions.
| | |       
| | |       In 1.2 this was fixed as part of a some bigger changes that removed merged
| | |       columns altogether (#19640). This fix is specific to 1.1 (this is not a
| | |       cherry-pick).
| | |       
| | |       Release note: fixed a crash caused by NATURAL JOINS and USING in conjunction
| | |       with a filter.
| | |       
| | |       Fixes #20569.
| | |   
* | |   commit 1ee122c053f440fe777fcdb2769a261167a8fbf7
|\ \ \  Merge: 9134f8f47 146e05930
| | | | Author: marc <marc.berhault@gmail.com>
| | | | Date:   Sat Dec 9 18:48:18 2017 -0500
| | | | 
| | | |     Merge pull request #20598 from mberhault/marc/specify_snappy_target
| | | |     
| | | |     build: specify snappy target.
| | | | 
| * | | commit 146e05930a3c11536360787076e1855ba924d6e2
|/ / /  Author: marc <marc@cockroachlabs.com>
| | |   Date:   Sat Dec 9 17:56:26 2017 -0500
| | |   
| | |       build: specify snappy target.
| | |       
| | |       Release Note: fix build when new versions of libgtest are installed.
| | |       
| | |       Fixes #20596
| | |       
| | |       When `libgtest` is installed, the snappy unittests fail to build.
| | |       Since we never run the snappy tests, let's stop building them entirely.
| | |   
* | |   commit 9134f8f4705de95be3723336bc34381162ad3e6b
|\ \ \  Merge: a5cbd0396 25f400f0c
| |/ /  Author: Richard Wu <richardwu1997@gmail.com>
|/| |   Date:   Fri Dec 8 18:01:54 2017 -0500
| | |   
| | |       Merge pull request #20541 from richardwu/cherry-pick-index
| | |       
| | |       cherrypick-1.1: fix panic when indexing selecting with additional schema change columns.
| | | 
| * | commit 25f400f0ce5d23b3e67ae12cafb8775cdf74394b
| | | Author: Richard Wu <richardwu1997@gmail.com>
| | | Date:   Wed Dec 6 17:04:23 2017 -0500
| | | 
| | |     cherrypick-1.1: fix panic when indexing selecting with additional schema
| | |     change columns.
| | |     
| | |     Release note (bug fix): fixed a panic for an edge-case when queries are
| | |     run against a table undergoing a schema change.
| | |   
* | |   commit a5cbd0396d0055c6b3b41a9ca0a4c377f8a84482
|\ \ \  Merge: acd2334ef 664784488
| | | | Author: Alfonso Subiotto Marqués <alfonso@cockroachlabs.com>
| | | | Date:   Wed Dec 6 19:36:45 2017 -0500
| | | | 
| | | |     Merge pull request #20534 from asubiotto/asubiotto/cherrypick_20088
| | | |     
| | | |     cherrypick-1.1: distsqlrun: close orderedSynchronizer sources that were neglected
| | | | 
| * | | commit 664784488deb826b6b0e616fe287759afe492054
| |/ /  Author: Alfonso Subiotto Marqués <alfonso@cockroachlabs.com>
| | |   Date:   Tue Nov 14 16:03:03 2017 -0500
| | |   
| | |       distsqlrun: close orderedSynchronizer sources that were neglected
| | |       
| | |       These sources were neglected because they were not added to the heap
| | |       when a heap initialization error happened. The ordered synchronizer
| | |       would then go into a draining phase, but it would only close sources
| | |       that were added to the heap.
| | |       
| | |       Fixes #19951
| | |       
| | |       This commit also improves the testing of distsql query cancellation. The
| | |       issue that this commit fixes is caught by stress testing the new test.
| | |       
| | |       Release note (bug fix): Fixes a race condition that would result in some
| | |       queries hanging after cancellation.
| | |   
* | |   commit acd2334ef4d93856d0ccef7ade57047bcf1357a4
|\ \ \  Merge: c6e58a8e4 dc6c5c17a
| | | | Author: Nikhil Benesch <nikhil.benesch@gmail.com>
| | | | Date:   Wed Dec 6 17:53:04 2017 -0500
| | | | 
| | | |     Merge pull request #20542 from benesch/1.1-logicrace-split
| | | |     
| | | |     cherrypick-1.1: build: separate out race logic tests
| | | | 
| * | | commit dc6c5c17a800d90840cad91ed8767763deef623b
|  / /  Author: David Taylor <tinystatemachine@gmail.com>
| | |   Date:   Fri Oct 6 00:10:50 2017 +0000
| | |   
| | |       build: separate out race logic tests
| | |       
| | |       the testrace build is going over 20min lately, so pulling logic tests (~15min) into their own build should help a bit.
| | |       a follow up step might be to split default from the rest of the logic tests if testlogicrace is still the slowest build.
| | |   
* | |   commit c6e58a8e4b3220a4c58a80d0e6011d6abf700835
|\ \ \  Merge: 9ad86b861 777d1817c
| |/ /  Author: Spencer Kimball <spencer.kimball@gmail.com>
|/| |   Date:   Wed Dec 6 17:39:37 2017 -0500
| | |   
| | |       Merge pull request #20446 from cockroachdb/cp11-gcqueue
| | |       
| | |       cherrypick-1.1: storage: move txn intent cleanup out of GC critical path, & GC keys early
| | | 
| * | commit 777d1817c78bbeeb02020c79b28bc0bd9813998d
| |/  Author: Spencer Kimball <spencer.kimball@gmail.com>
| |   Date:   Fri Dec 1 09:04:15 2017 -0800
| |   
| |       storage: move txn intent cleanup out of GC critical path, & GC keys early
| |       
| |       This change first moves the cleanup of potentially "fat" transactions out
| |       of the GC critical path by sending them to the intent resolver's asynchronous
| |       cleanup mechanism. This prevents a txn laden with significant numbers of
| |       unresolved intents from gumming up the GC queue and experiencing context
| |       timeouts.
| |       
| |       Moved GC of keys scanned during the GC process to execute before intents
| |       encountered during the scan are processed. This makes progress in the GC
| |       queue more likely and greatly lessens the chance of a stubborn range
| |       entering into an infinite GC loop.
| |       
| |       Removed unused code for the `ResolveOptions.Wait` boolean and changed
| |       `intentResolver.resolveIntents` to send batches of 100 serially with a new
| |       timeout for each batch. In concert with this, removed the timeout previously
| |       set in `intentResolver.processIntents`.
| |       
| |       Release note: improve garbage collection of very large transactions and
| |       large volumes of abandoned writes (intents).
| |   
* |   commit 9ad86b861cd8eb7b8d9ebca02ada738a77aae811
|\ \  Merge: 59fda2ec5 cf6b27e89
| | | Author: RaduBerinde <radu@cockroachlabs.com>
| | | Date:   Wed Dec 6 12:20:44 2017 -0500
| | | 
| | |     Merge pull request #20512 from RaduBerinde/crash-plan-sort-1.1
| | |     
| | |     cherrypick-1.1: distsql: fix crash in addSorters
| | | 
| * | commit cf6b27e89e47e64ffa535500097a58156a8d4fe9
| |/  Author: Radu Berinde <radu@cockroachlabs.com>
| |   Date:   Tue Dec 5 13:14:40 2017 -0500
| |   
| |       distsql: fix crash in addSorters
| |       
| |       Fixing a crash triggered by a case where we don't actually need a sortNode
| |       column.
| |       
| |       Fixes #20481.
| |       
| |       Release note (bug fix): fixed a crash triggered by some corner-case queries
| |       containing ORDER BY.
| |   
* |   commit 59fda2ec5213bf18f537bedeec174cafb410056d
|\ \  Merge: 29dba0b97 71fd8360e
| |/  Author: Peter Mattis <peter@cockroachlabs.com>
|/|   Date:   Tue Dec 5 19:16:37 2017 -0500
| |   
| |       Merge pull request #20502 from petermattis/pmattis/cherrypick-debug-compact
| |       
| |       cherry-pick-1.1: libroach: avoid significant temp disk space usage in DBCompact
| | 
| * commit 71fd8360ed18ded7c9651c7e9c510554cdc73b9c
|/  Author: Peter Mattis <petermattis@gmail.com>
|   Date:   Mon Dec 4 17:18:19 2017 -0500
|   
|       libroach: avoid significant temp disk space usage in DBCompact
|       
|       Change DBCompact to individually compact the ranges delineated by the
|       sstables in the bottom-most level. This reduces temporary disk space
|       usage for compact range significantly. Before this change, running
|       `debug compact` on a 2.5GB storage directory temporarily grew the size
|       of the storage directory to 3.5GB. After this change, the storage
|       directory remains at 2.5GB (or below).
|       
|       Release note (cli): reduce temporary disk space usage for the `debug
|       compact` command.
|       
|       Fixes #20338
|   
*   commit 29dba0b97e4dc0aced12274265120631ee935b89
|\  Merge: ca0beda9b 18d5c11fe
| | Author: marc <marc.berhault@gmail.com>
| | Date:   Thu Nov 30 14:34:38 2017 -0500
| | 
| |     Merge pull request #20284 from mberhault/marc/update_rocksdb
| |     
| |     cherrypick-1.1: skip generating empty sst
| | 
| * commit 18d5c11fe5a186b19e8842b65d89eabda492379b
|   Author: marc <marc@cockroachlabs.com>
|   Date:   Mon Nov 27 15:29:37 2017 -0500
|   
|       cherrypick-1.1: skip generating empty sst
|       
|       Release Node (rocksdb): don't write empty sstables (causes corruption).
|       
|       Fixes #19891.
|   
*   commit ca0beda9bc92432a34887198b6845db0aa3bddae
|\  Merge: 9f38d19bf fb9e4a1df
| | Author: kena <knz@users.noreply.github.com>
| | Date:   Tue Nov 28 18:08:40 2017 +0100
| | 
| |     Merge pull request #20301 from knz/20171128-cherrypick-20264
| |     
| |     cherry-pick 1.1: sql: name the transaction settings like PG does
| | 
| * commit fb9e4a1df60b323cff57b2aa10190c9a8df25efb
| | Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
| | Date:   Sun Nov 26 20:32:37 2017 +0100
| | 
| |     sql: name the transaction settings like PG does
| |     
| |     In PostgreSQL, the transaction isolation level is reported by the
| |     session variable `transaction_isolation` (two words separated by an
| |     underscore), not `transaction isolation level` (three words separated
| |     by spaces). Also the transaction level is reported in lowercase.
| |     
| |     Clients actually care about this stuff.
| |     
| |     This patch makes CockroachDB behave more like PG.
| |     
| |     This also aligns `transaction_priority` and `transaction_status`
| |     (CockroachDB extension) with `transaction_isolation` (PG standard
| |     setting), for more consistency in UX.
| |     
| |     Patch series: ["Just because you are unique does not mean you are
| |     useful."](https://duckduckgo.com/?q=just+because+you+are+unique+does+not+mean+you+are+useful&iar=images)
| |     
| |     ----
| |     
| |     Release note (sql): the session settings `transaction isolation
| |     level`, `transaction priority` and `transaction status` are now called
| |     `transaction_isolation`, `transaction_priority` and
| |     `transaction_status` for better compatibility with PostgreSQL.
| | 
| * commit 3078ba97938d261cef37e66eaab7c9d4ce88ad51
|/  Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
|   Date:   Sun Nov 26 20:12:40 2017 +0100
|   
|       sql: comment the session variables
|   
*   commit 9f38d19bfdd17ffda1e9995ce84d7bad9190eec5
|\  Merge: ca5cb21b0 1c2adc934
| | Author: marc <marc.berhault@gmail.com>
| | Date:   Tue Nov 28 10:01:24 2017 -0500
| | 
| |     Merge pull request #20283 from mberhault/marc/cherrypick_20255
| |     
| |     cherrypick-1.1: cli: dump: fix infinite loop in when resolving depend…
| | 
| * commit 1c2adc934b75e8a19be53b2562727594361b3d2d
|   Author: marc <marc@cockroachlabs.com>
|   Date:   Fri Nov 24 16:03:05 2017 -0500
|   
|       cherrypick-1.1: cli: dump: fix infinite loop in when resolving dependencies.
|       
|       Release Note (cli): fix panic on `cockroach dump` in the presence of
|       reference cycles.
|       
|       While this will avoid the loop and keep dependency order, there may
|       still be problems with dependency cycles.
|       
|       This came up in #20254 when encountering a foreign key referring to the
|       same table.
|       
|       Before the fix, the test fails with:
|       ```
|       runtime: goroutine stack exceeds 1000000000-byte limit
|       fatal error: stack overflow
|       ```
|   
*   commit ca5cb21b061f734bd54e4e8d324d29e007a0bd7e
|\  Merge: de82822e1 c9078243a
| | Author: Andrei Matei <andreimatei1@gmail.com>
| | Date:   Mon Nov 27 17:59:09 2017 -0500
| | 
| |     Merge pull request #20285 from andreimatei/extra-10-compat
| |     
| |     1.1 only: distsqlrun: introduce --extra-1.0-compatibility
| | 
| * commit c9078243ab7fc42e9104f4e57e79ac52bee72842
|   Author: Andrei Matei <andrei@cockroachlabs.com>
|   Date:   Mon Nov 27 15:29:39 2017 -0500
|   
|       1.1 only: distsqlrun: introduce --extra-1.0-compatibility
|       
|       Introduce a new flag forcing 1.1 nodes to be compatible with DistSQL
|       flows produced by 1.0 gateways. This is motivated by some clients not
|       finding it easy to upgrade from 1.0 to 1.1 because of DistSQL on 1.1
|       refusing flows produced by 1.0, and so a mixed-version cluster running
|       for an extended period has a problem.
|       
|       --extra-1.0-compatibility does the following:
|       - makes consumers not send handshake messages to producers upon stream
|       connection. These messages were new in 1.1 and would confuse 1.0 code.
|       These handshakes are not actually used at the moment, albeit they may be
|       in a future version.
|       - an old proto format for AggregationSpec's is supported.
|       - the DistSQL MinAcceptedVersion is reset to 3, which is 1.0.x's
|       version.
|       
|       These changes are done only on the 1.1 branch. 1.2 will not support this
|       flag.
| 
* commit de82822e1b9d96bbf8ad9b557d79b04a9af269c1
| Merge: 0629bcd52 693d9dcc1
| Author: kena <knz@users.noreply.github.com>
| Date:   Mon Nov 27 23:17:05 2017 +0100
| 
|     Merge pull request #20286 from knz/20171127-cherrypick-20267
|     
|     cherry-pick 1.1: sql: ensure that AS OF SYSTEM TIME is handled consistently
| 
* commit 693d9dcc1eb27bc884b9ee9931fb88a8fd692afa
  Author: Raphael 'kena' Poss <knz@cockroachlabs.com>
  Date:   Mon Nov 27 00:06:14 2017 +0100
  
      sql: ensure that AS OF SYSTEM TIME is handled consistently
      
      Prior to this patch, two inconsistencies were present in the code:
      
      - a user-visible inconsistency: `AS OF SYSTEM TIME` (henceforth
        referred to as "AOST") was accepted anywhere in a query as soon as
        it was present at the top level SELECT statement, including with
        conflicting timestamps. Accepting AOST in multiple places doesn't
        feel undesirable, but allowing conflicting timestamps to be
        specified is unsound with the current underlying mechanism (a shared
        timestamp for the entire transaction).
      
      - an internal inconsistency, visible to CockroachDB developers: the
        presence of an AOST clause during planning was erroneously conflated
        with the flag that disables caching of table descriptors
        (`planner.avoidCachedDescriptors`). This is erroneous because,
        although AOST *implies* `avoidCachedDescriptors == true` (we can't
        use the cache while time travelling), the converse is not true: when
        processing view descriptors (and perhaps, in the future, for other
        reasons), we also disable descriptor caching although AOST is not
        involved.
      
      This patch rectifies this situation as follows:
      
      - a new planner flag `asOfSystemTime` is introduced to indicate
        that an AOST clause was properly recognized at the top level.
      
      - the logic that allows or refuses AOST clauses in FROM clauses in
        arbitrarily nested SELECT clauses (including, potentially, those
        expanded from views), is modified to use this new flag.
      
      - the timestamps of AOST clauses, if multiple are specified, are
        checked to be equal to the one set at the top level. This
        restriction might be lifted in the future if we ever support
        different AOST clauses per data source.
      
      In addition, the error message when an AOST clause is not given in the
      proper syntactic position is improved to hint where it should be
      placed instead.
      
      ----
      
      Release note (sql, bug fix): it is not possible any more to indicate
      conflicting `AS OF SYSTEM TIME` clauses in different part of a query.
```

jseldess:
@andreimatei and @mjibson, do the following changes need to be mention in the 1.1.4 release notes? If so, please draft it up for me and let me know which category to put it under.

```
* |   commit 666df6cce121964d101b90e3e0d11705e5c6d9c9
|\ \  Merge: 6ac21a04a dbaa5489b
| | | Author: Matt Jibson <matt.jibson@gmail.com>
| | | Date:   Thu Dec 21 16:04:06 2017 -0500
| | | 
| | |     Merge pull request #21003 from mjibson/cherrypick-19556
| | |     
| | |     cherrypick-1.1: ccl/storageccl: retry s3 region fetch
| | | 
| * | commit dbaa5489b0f61245247fb4ce01f5c3fc3a925cf8
| | | Author: Matt Jibson <matt.jibson@gmail.com>
| | | Date:   Thu Oct 26 03:24:09 2017 -0400
| | | 
| | |     ccl/storageccl: retry s3 region fetch
| | |     
| | |     Retry an S3 operation that has been causing transient errors. If a
| | |     503 is returned, we may need to slow down our request rate. Although
| | |     the retry package does wait a bit, it is only some 10s of ms. We
| | |     instead wait 5s on these errors to attempt to slightly decrease the
| | |     request rate. This 5s is purely a guess, but it is also a low enough
| | |     time that it won't matter too much if we are waiting longer than
| | |     minimally needed.
| | |     
| | |     For now put this only in the place we have seen the failure. We can
| | |     optionally retry the other locations if we see problems there. This
| | |     is the same approach we used for Azure and GCE with good results.
| | |     
| | |     Fixes #19502
| | |   
*   commit ca5cb21b061f734bd54e4e8d324d29e007a0bd7e
|\  Merge: de82822e1 c9078243a
| | Author: Andrei Matei <andreimatei1@gmail.com>
| | Date:   Mon Nov 27 17:59:09 2017 -0500
| | 
| |     Merge pull request #20285 from andreimatei/extra-10-compat
| |     
| |     1.1 only: distsqlrun: introduce --extra-1.0-compatibility
| | 
| * commit c9078243ab7fc42e9104f4e57e79ac52bee72842
|   Author: Andrei Matei <andrei@cockroachlabs.com>
|   Date:   Mon Nov 27 15:29:39 2017 -0500
|   
|       1.1 only: distsqlrun: introduce --extra-1.0-compatibility
|       
|       Introduce a new flag forcing 1.1 nodes to be compatible with DistSQL
|       flows produced by 1.0 gateways. This is motivated by some clients not
|       finding it easy to upgrade from 1.0 to 1.1 because of DistSQL on 1.1
|       refusing flows produced by 1.0, and so a mixed-version cluster running
|       for an extended period has a problem.
|       
|       --extra-1.0-compatibility does the following:
|       - makes consumers not send handshake messages to producers upon stream
|       connection. These messages were new in 1.1 and would confuse 1.0 code.
|       These handshakes are not actually used at the moment, albeit they may be
|       in a future version.
|       - an old proto format for AggregationSpec's is supported.
|       - the DistSQL MinAcceptedVersion is reset to 3, which is 1.0.x's
|       version.
|       
|       These changes are done only on the 1.1 branch. 1.2 will not support this
|       flag.
| 
```

andreimatei:
Not for mine.

mjibson:
Yes: `bug fix: improve S3 resiliency during BACKUP/RESTORE/IMPORT`

bdarnell:
LGTM

releases/v1.1.4.md
```diff
@@ -0,0 +1,80 @@
...
+- Fixed an issue where seemingly irrelevant error messages were being returned for certain [`INSERT`](../v1.1/insert.html) statements.[#20855](https://github.com/cockroachdb/cockroach/pull/20855)
+- Crash details are now properly copied to the log file even when a node was started with `--logtostderr` as well as in other circumstances when crash details could be lost previously. [#20842](https://github.com/cockroachdb/cockroach/pull/20842)
+- Fixed an issue where wildcards (`_`, `%`) and `\` escaped characters in `LIKE` patterns were not handled properly. [#20654](https://github.com/cockroachdb/cockroach/pull/20654)
+- Fixed the return type signature of the `array_positions()` [built-in function](../v1.1/functions-and-operators.html). This caused a crash if the function was used in query execution by the distributed execution engine. [#20652](https://github.com/cockroachdb/cockroach/pull/20652)
+- Fix a crash caused by null [collated strings](../v1.1/collate.html). [#20648](https://github.com/cockroachdb/cockroach/pull/20648)
```
Amruta-Ranade: nit: 'Fix' to 'Fixed' to maintain consistency

jseldess: Thanks, @Amruta-Ranade. I'm going to let TC finish so I can fix any broken links with this update.

jseldess: Done.

PR #2337, opened 2018-01-08
Added SERIAL aliases

### LINKED ISSUE #2320: Add aliases for SERIAL types
PRs: 
- https://github.com/cockroachdb/cockroach/issues/20774
- https://github.com/cockroachdb/cockroach/pull/20776#issuecomment-352910761

Additional context from @vilterp:

I don’t thinking you have to update sequence docs about `serial` though maybe should add a note to the `serial` docs that `serial2` & `serial4` are actually 8 bytes, and you need all 8 bytes to hold the response of `unique_rowid` cuz if you cut off some of the bytes you’ll lose data…

lhirata:
fixes #2320


v2.0/serial.md
```diff
@@ -16,8 +16,9 @@ The `SERIAL` type is equivalent to [`INT DEFAULT unique_rowid()`](int.html).
 
 In CockroachDB, the following are aliases for `SERIAL`:
 
-- `SMALLSERIAL`
-- `BIGSERIAL`
+Name | Allowed Width | Aliases
```
jseldess: Since `SERIAL` always generates a 64-bit integer, and you never manually do this yourself, instead of listing `Allowed Width` here, I'd suggest we update the intro sentence to:

> The `SERIAL` data type is a column data type that, on insert, generates a 64-bit signed integer from the timestamp and ID of the node executing the insert.

We already state that on the [data type overview page](https://www.cockroachlabs.com/docs/stable/data-types.html). 

Then I think we can go back to a simple bulleted list of aliases. Thoughts, @lhirata?

jseldess:
Separately, I think we need to update the [`unique_rowid()` function description](https://www.cockroachlabs.com/docs/stable/functions-and-operators.html#id-generation-functions) to explicitly state that it generates a 64-bit value and, thus, when inserting into a column, the column type needs to be able to accommodate that width. Does that sound right, @vilterp? If so, I can create a separate issue for that.

v2.0/serial.md
lhirata: Yeah, that sounds good to me!

vilterp:
Yeah, what we want to avoid here is people thinking that because `SERIAL2` and `SERIAL4` exist, they can fit a value from `unique_rowid` into 2 or 4 bytes. If they use a 2- or 4-byte type in their application code, part of the value will be truncated when it's read into that application code, causing bugs. So yes, a warning about this would be good.

If they really need to use a 2- or 4-byte type in their application code (e.g. because they're trying to run legacy software that they can't/don't want to change on top of CR), they can use sequences (which will be slower of course).

jseldess:
I see. So @lhirata, it seems like we need to do 2 things:

1. Following the aliases list, I think we need to add a warning based on what @vilterp says above, 
that the `serail2` and `serial4` alisaes aren't different than `serial`. They store 8-byte (64-bit) values, not 2- or 4-byte values that the names might suggest.
2. We should create a docs issue to update the `unique_rowid()` function description. This change will have to be made in the cockroach repo, but you can create a docs issue. I'll help you work it out.

lhirata:
New issue for `unique_rowid()`: #2351

lhirata:
@vilterp can you review? thx!

vilterp:
:lgtm: with a small nit

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[v2.0/serial.md, line 7 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2337#-L2X5hlWm7QaXibmZ9l7:-L2X5hlWm7QaXibmZ9l8:b7wgms9) ([raw file](https://github.com/cockroachdb/docs/blob/0c90d7a1eede47cb80af7eadbd4f5c05f8ad9909/v2.0/serial.md#L7)):*
> ```Markdown
> ---
> 
> The `SERIAL` [data type](data-types.html) is a column data type that, on insert, generates a 64-bit signed integer from the timestamp and ID of the node executing the insert. This combination is likely to be globally unique except in extreme cases (see this [example](create-table.html#create-a-table-with-auto-generated-unique-row-ids) for more details). Also, because value generation does not require talking to other nodes, it is much faster than sequentially auto-incrementing a value, which requires distributed coordination.
> ```

I don't think you have to say `signed` here, just `64-bit integer`.  It's a bunch of stuff concatenated together into an integer, so I don't think it meaningfully has a sign.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2337#-:-L2YRUf3EU2NssIYfT0n:bm5fgra)*

PR #2355, opened 2018-01-10
remove generate scripts from docs

### LINKED ISSUE #2353: Remove docs generators from repo
This should probably replace https://github.com/cockroachdb/docs/pull/1956.

mjibson:
Fixes #2353


mjibson:
Thanks for that. I was about to start tracking that failure down.

PR #2361, opened 2018-01-12
Document how to reduce or disable timeseries data

### LINKED ISSUE #2040: Provide more cluster settings to configure storing historical operation
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwINAgXA5on9pr2T/rec4M9yFaIlgVDPWR

PRs:
- https://github.com/cockroachdb/cockroach/pull/18632

jseldess:
Fixes #2040

### DIFF
--- a/v2.0/operational-faqs.md
+++ b/v2.0/operational-faqs.md
@@ -2,7 +2,7 @@
 title: Operational FAQs
 summary: Get answers to frequently asked questions about operating CockroachDB.
 toc: false
-toc_not_nester: true
+toc_not_nested: true
 ---
 
 <div id="toc"></div>
@@ -38,7 +38,15 @@ Like most databases, CockroachDB caches the most recently accessed data in memor
 
 The timeseries data used to power the graphs in the admin UI is stored within the cluster and accumulates for 30 days before it starts getting truncated. As a result, for the first 30 days or so of a cluster's life, you will see a steady increase in disk usage and the number of ranges even if you aren't writing data to the cluster yourself.
 
-To truncate timeseries data sooner, you can change the `timeseries.resolution_10s.storage_duration` cluster setting to an [`INTERVAL`](interval.html) value less than `720h0m0s` (30 days). For example, to truncate timeseries data after 15 days, you would execute the following [`SET CLUSTER SETTING`](set-cluster-setting.html) command:
+## Can I reduce or disable the storage of timeseries data? <span class="version-tag">New in v2.0</span>
+
+Yes. By default, CockroachDB stores timeseries data for the last 30 days for display in the Admin UI, but you can [reduce the interval for timeseries storage](#reduce-the-interval-for-timeseries-storage) or [disable timeseries storage entirely](#disable-timeseries-storage-entirely).
+
+{{site.data.alerts.callout_info}}After reducing or disabling timeseries storage, it can take up to 24 hours for timeseries data to be deleted and for the change to be reflected in Admin UI metrics.{{site.data.alerts.end}}
+
+### Reduce the interval for timeseries storage
+
+To reduce the interval for storage of timeseries data, change the `timeseries.resolution_10s.storage_duration` cluster setting to an [`INTERVAL`](interval.html) value less than `720h0m0s` (30 days). For example, to store timeseries data for the last 15 days, run the following [`SET CLUSTER SETTING`](set-cluster-setting.html) command:
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -57,6 +65,38 @@ To truncate timeseries data sooner, you can change the `timeseries.resolution_10
 | 360h                                       |
 +--------------------------------------------+
 (1 row)
+~~~
+
+### Disable timeseries storage entirely
+
+{{site.data.alerts.callout_info}}Disabling timeseries storage is recommended only if you exclusively use a third-party tool such as <a href="monitor-cockroachdb-with-prometheus.html">Prometheus</a> for timeseries monitoring. Prometheus and other such tools do not rely on CockroachDB-stored timeseries data; instead, they ingest metrics exported by CockroachDB from memory and then store the data themselves.{{site.data.alerts.end}}
+
+To disable the storage of timeseries data entirely, run the following command:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING timeseries.storage.enabled = false;
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SHOW CLUSTER SETTING timeseries.storage.enabled;
+~~~
+
+~~~
++----------------------------+
+| timeseries.storage.enabled |
++----------------------------+
+| false                      |
++----------------------------+
+(1 row)
+~~~
+
+If you want all existing timeseries data to be deleted, change the `timeseries.resolution_10s.storage_duration` cluster setting as well:     
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING timeseries.resolution_10s.storage_duration = '0s';
 ~~~
 
 ## Why does CockroachDB collect anonymized cluster usage details by default?
--- a/v2.0/admin-ui-access-and-navigate.md
+++ b/v2.0/admin-ui-access-and-navigate.md
@@ -32,6 +32,9 @@ The Admin UI displays time series graphs of key metrics. Time series graphs are 
 
 <img src="{{ 'images/admin_ui_hovering.gif' | relative_url }}" alt="CockroachDB Admin UI" style="border:1px solid #eee;max-width:100%" />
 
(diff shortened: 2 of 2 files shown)


v2.0/operational-faqs.md
```diff
@@ -38,26 +38,53 @@ Like most databases, CockroachDB caches the most recently accessed data in memor
...
+    ~~~
+    +----------------------------+
+    | timeseries.storage.enabled |
+    +----------------------------+
+    | false                      |
+    +----------------------------+
+    (1 row)
+    ~~~
+
+{{site.data.alerts.callout_info}}After adjusting either of these cluster settings, it can take up to 24 hours for timeseries data to be deleted and for the change to be reflected in Admin UI metrics.{{site.data.alerts.end}}
```
jseldess: @dianasaur323, @mrtracy, @vilterp, this sound ok?

vilterp: Based on seeing this in the code: https://github.com/cockroachdb/cockroach/blob/3dbe1dae8e734a27df53591f882efcd1543de9cd/pkg/storage/ts_maintenance_queue.go#L31-L33

Now that I look again though, I see that that's a _per-replica_ timer, and replica timers could be staggered. So maybe a little more accurate to say that it could take up to 24 hours for it to be deleted _across all replicas_.

I'm not sure if that adds a lot though; what you have now seems good. Thoughts @mrtracy?

vilterp:
High-level question on this issue: does turning off our internal TS DB break Prometheus? (Since Prometheus pulls data from us). Or do the metrics on the `/_status/vars` endpoint come from memory, in which case turning off our TS DB would be fine?

mrtracy:
@vilterp No, disabling TS storage does not disable prometheus.  Prometheus scrapes the current values of metrics (from memory), it does not read anything from the TS system.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v2.0/operational-faqs.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2fp5ynIOMTURuz6yoU-r1-87:-L2g5qam3n5yy4xSbhom:b-lbiquo) ([raw file](https://github.com/cockroachdb/docs/blob/2f541daead6c081b29045cb6049b63df73d65ef5/v2.0/operational-faqs.md#L87)):*

Based on seeing this in the code: https://github.com/cockroachdb/cockroach/blob/3dbe1dae8e734a27df53591f882efcd1543de9cd/pkg/storage/ts_maintenance_queue.go#L31-L33

Now that I look again though, I see that that's a _per-replica_ timer, and replica timers could be staggered. So maybe a little more accurate to say that it could take up to 24 hours for it to be deleted _across all time series replicas_.

I'm not sure if that adds a lot though; what you have now seems good. Thoughts @mrtracy? 
</blockquote></details>

@jseldess This seems to imply that setting the `timeseries.storage.enable` to false will cause existing time series data to be deleted; that is not true. It will disable any additional time series data from being written, but existing data will still be retained (and eventually deleted according to the storage_duration setting).  In order to disable *and* delete existing data immediately, you would need to adjust both settings.

@pvilter The replica scanner is used to *discover* time series that need to be pruned (as an alternative to keeping a central list of time series somewhere else).  Once a time series is discovered, the operation "prunes" it, deleting all data for that series older than the retention threshold across the cluster.  Therefore, in practice the deletion should happen faster than 24 hours, because the multiple replicas give multiple chances for each time series to be discovered and pruned.  24 hours is an upper limit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-:-L2g5X1okbeLDhxiTT-T:b5s4cos)*

jseldess:
*[v2.0/operational-faqs.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2fp5ynIOMTURuz6yoU-r1-87:-L2jXlMVtihVyFcFcgfx:b-sky2gw) ([raw file](https://github.com/cockroachdb/docs/blob/2f541daead6c081b29045cb6049b63df73d65ef5/v2.0/operational-faqs.md#L87)):*
<details><summary><i>Previously, mrtracy (Matt Tracy) wrote…</i></summary><blockquote>

@jseldess This seems to imply that setting the `timeseries.storage.enable` to false will cause existing time series data to be deleted; that is not true. It will disable any additional time series data from being written, but existing data will still be retained (and eventually deleted according to the storage_duration setting).  In order to disable *and* delete existing data immediately, you would need to adjust both settings.

@pvilter The replica scanner is used to *discover* time series that need to be pruned (as an alternative to keeping a central list of time series somewhere else).  Once a time series is discovered, the operation "prunes" it, deleting all data for that series older than the retention threshold across the cluster.  Therefore, in practice the deletion should happen faster than 24 hours, because the multiple replicas give multiple chances for each time series to be discovered and pruned.  24 hours is an upper limit.
</blockquote></details>

PTAL. I've tried to get more explicit by breaking the FAQ into two parts. I've also linked to this FAQ from a spot in the admin ui docs where we explicitly call out time series graphs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361)*

mrtracy:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[v2.0/operational-faqs.md, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2zmerGj5xfjt_FDG-X:-L2zmerGj5xfjt_FDG-Y:b17trpo) ([raw file](https://github.com/cockroachdb/docs/blob/9f5f1efdb81cd3961c18a9540339f4e396592e10/v2.0/operational-faqs.md#L45)):*
> ```Markdown
> Yes. By default, CockroachDB stores timeseries data for the last 30 days for display in the Admin UI, but you can [reduce the interval for timeseries storage](#reduce-the-interval-for-timeseries-storage) or [disable timeseries storage entirely](#disable-timeseries-storage-entirely).
> 
> {{site.data.alerts.callout_info}}After reducing or disabling timeseries storage, it can take up to 24 hours for timeseries data to be deleted and for the change to be reflected in Admin UI metrics.{{site.data.alerts.end}}
> ```

I'm not sure how exact we need to be, but this is still a little off by my reading; after reducing time series storage, the pruning may take up to 24hrs; however, disabling time series storage has an immediate effect, time series collection will cease right away.

That said, I'm not sure if this is one of those situations where being more specific will end up being more confusing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361)*

mrtracy:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[v2.0/operational-faqs.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2fp5ynIOMTURuz6yoU-r1-87:-L2znz4C5KShqUt4TY2l:b40ex5k) ([raw file](https://github.com/cockroachdb/docs/blob/2f541daead6c081b29045cb6049b63df73d65ef5/v2.0/operational-faqs.md#L87)):*

PTAL. I've tried to get more explicit by breaking the FAQ into two parts. I've also linked to this FAQ from a spot in the admin ui docs where we explicitly call out time series graphs.
</blockquote></details>

This part looks good to me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361)*

mrtracy:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[v2.0/operational-faqs.md, line 45 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2zmerGj5xfjt_FDG-X:-L2zodbRtpfBFlDhwUIV:b-8t0j20) ([raw file](https://github.com/cockroachdb/docs/blob/9f5f1efdb81cd3961c18a9540339f4e396592e10/v2.0/operational-faqs.md#L45)):*
<details><summary><i>Previously, mrtracy (Matt Tracy) wrote…</i></summary><blockquote>

I'm not sure how exact we need to be, but this is still a little off by my reading; after reducing time series storage, the pruning may take up to 24hrs; however, disabling time series storage has an immediate effect, time series collection will cease right away.

That said, I'm not sure if this is one of those situations where being more specific will end up being more confusing.
</blockquote></details>

After trying to rewrite this, I think this is likely fine.  Getting more detailed seems to make it sound more complicated than it is, and the linked sections cover the caveat.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-:-L2zrf4lzU6btlhlFnRq:bnfp4nl)*

dianasaur323:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[v2.0/operational-faqs.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-L2fp5ynIOMTURuz6yoU-r1-87:-L3-99jQqEb0chan4iYP:b84nhm5) ([raw file](https://github.com/cockroachdb/docs/blob/2f541daead6c081b29045cb6049b63df73d65ef5/v2.0/operational-faqs.md#L87)):*
<details><summary><i>Previously, mrtracy (Matt Tracy) wrote…</i></summary><blockquote>

This part looks good to me.
</blockquote></details>

lgtm. thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-:-L3-9CwYT33sgZFfO8i-:bnfp4nl)*

jseldess:
Thanks for the reviews!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2361#-:-L3-DWFHUPP7RLrq0V_L:b-hlc8bm)*

PR #2374, opened 2018-01-19
Solicit more specific "No" feedback

### LINKED ISSUE #1747: More specific "No" feedback on docs
When a user selects **Was this page helpful?** > **No**, we should offer options for providing more details. We used to do this via a popup, but we found that to discourage submissions and to condition users to not use the feedback mechanism. So instead of a popup, we should offer the options on the page.

Spotify can serve as an example:

screen shot 2017-07-21 at 8 56 12 am
screen shot 2017-07-21 at 8 56 21 am

jseldess:
Fixes #1747


jseldess:
@j-low, as mentioned in the related issue, I need help applying our styling to the form. Also, we have to figure out the confirmation message after "Submit". If you provide guidance, I'm happy to continue trying. Otherwise, feel free to take over the PR.

j-low:
@jseldess PTAL but should be good to go.

PR #2375, opened 2018-01-20
Remove homebrew install option for non-stable releases

### LINKED ISSUE #2153: Homebrew instructions are not version aware
The Homebrew instructions on [dev version of Install CockroachDB](https://www.cockroachlabs.com/docs/dev/install-cockroachdb.html) instead the latest stable version.

@benesch Do you have suggestions for a fix for this that work globally, incl. for versions that are beta-only?

jseldess:
Our homebrew recipe is currently only for the latest stable release.

Fixes #2153

### DIFF
--- a/v2.0/install-cockroachdb.html
+++ b/v2.0/install-cockroachdb.html
@@ -83,13 +83,14 @@ $(document).ready(function(){
 <p>See <a href="../releases/{{page.release_info.version}}.html" data-eventcategory="mac-releasenotes-download">Release Notes</a> for what's new in the latest release, {{ page.release_info.version }}. </p>
 
 <div id="mac-installs" class="clearfix install-wrap">
-<a href="#use-homebrew" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-homebrew">Use <div class="c2a">Homebrew</div></a>
-<a href="#download-the-binary" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-binary">Download the <div class="c2a">Binary</div></a>
+<!-- <a href="#use-homebrew" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-homebrew">Use <div class="c2a">Homebrew</div></a> -->
+<a href="#download-the-binary" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-binary">Download the <div class="c2a">Binary</div></a>
 <a href="#build-from-source" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-source">Build from <div class="c2a">Source</div></a>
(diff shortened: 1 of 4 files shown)


v2.0/install-cockroachdb.html
```diff
@@ -83,13 +83,14 @@
 <p>See <a href="../releases/{{page.release_info.version}}.html" data-eventcategory="mac-releasenotes-download">Release Notes</a> for what's new in the latest release, {{ page.release_info.version }}. </p>
 
 <div id="mac-installs" class="clearfix install-wrap">
-<a href="#use-homebrew" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-homebrew">Use <div class="c2a">Homebrew</div></a>
-<a href="#download-the-binary" class="install-button mac-button" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-binary">Download the <div class="c2a">Binary</div></a>
+<!-- <a href="#use-homebrew" class="install-button mac-button current" data-eventcategory="buttonClick-doc-install" data-eventaction="mac-homebrew">Use <div class="c2a">Homebrew</div></a> -->
```
benesch: I believe you can also use a conditional to avoid needing to update this whenever a version is promoted to stable:

```
{% if page.version.stable? %}
```

jseldess:
Good point, @benesch, but I can't quite figure out the right variable to use, though. `{% if page.version.stable? %}` doesn't seem to work. Nor does `{% if page.version == stable %}` or `{% if page.version.name == stable %}`. Any ideas?

Also, even if I can get this bit of logic to work, I still need to add the `active` class to the next button in the list, and remove the `display: none` styling from the next content section as well. Do you see any easy way to accomplish that?

benesch:
Ah, sorry, it required just a touch of Ruby: https://github.com/cockroachdb/docs/pull/2375/commits/bd80bf6ec15be26d20b6efafe2c2c848526b73fe#diff-dd7d09ea0294ee5e0584776264833760R41

I also took the opportunity to rewrite the tabbing JavaScript. It's no longer necessary to explicitly mark things as `current`. Whatever tab is first will be automatically marked `current` when the page loads.

v2.0/install-cockroachdb.html
```diff
@@ -8,67 +8,32 @@
...
-        $("#macinstall").hide();
-        $("#windowsinstall").hide();
-    }
-
-    function toggleWindows(){
-        $(".windows-button:first").trigger('click');
-        $("#windowsinstall").show();
-        $("#macinstall").hide();
-        $("#linuxinstall").hide();
+    // Detect OS and display corresponding tab by default.
```
jseldess: @m-schneider, who's on linux, said this defaults to linux in chrome but to mac in ff. Any ideas why?

Also, don't you want to make these changes to the 1.1 version of the page as well?

benesch: Hmm, as a result of this change, or always?

jseldess:
Thanks for the help, @benesch!

benesch:
Most definitely!

PR #2380, opened 2018-01-23
Fix typo in demo-fault-tolerance-and-recovery.md

puilp0502:


jseldess:
@puilp0502, thanks very much for catching and fixing this!

Issue #2372, opened 2018-01-19
Code blocks must not be hyphenated automatically


knz:
See for example:

![schermafdruk_2018-01-19_18-26-43](https://user-images.githubusercontent.com/642886/35163242-5e7afd42-fd46-11e7-9cc0-03622e84eea1.png)

The SE-LECT is hy-phe-na-ted. No good, it makes copy-pasting impossible.

cc @jseldess @j-low

j-low:
Addressed in #2404.  Closing issue.

PR #2404, opened 2018-01-23
Remove hyphens from code syntax

j-low:
__[description]__
* re: #2372


jseldess:
LGTM, but I think @knz should check, too.

css/customstyles.scss
```diff
@@ -103,11 +109,6 @@ pre, code {
   color: $cl_blue;
```
knz: I think hyphenation should be disabled for both `pre` and `code`. I can't see of any valid case for automatically hyphenating content that the author explicitly wanted to render preformatted.

Also, incidentally, I don't clearly see why we use a different border for `code` and `pre`.

j-low: Agree re: de-hyphenating the `pre` blocks as well.

Re: border color: `pre` currently has no border.  Are you suggesting to give it a border similar to the `code` blocks?

knz:
Yes I suggest <pre> gets a border.

Also and I think I've filed a separate issue for that already, i'd also suggest <code> *loses* both its background and border *while in the context of a title* and gets a font size with the same height as the surrounding text. 
-- 

j-low:
@knz yes, I saw that issue over in the Hugo repo.  I think you were erroneously directed to file it there-- anything that lives under the 'docs' section can be filed here; bugs from any other section can be filed in the Hugo repo.  Sorry for the confusion.  

Would you be able to re-register that bug here in docs?  It will get addressed but I will have to do so in the next round of bug fixes as I've got some higher priority stuff on my plate that I need to address.

j-low:
@knz, okay so I have to punt on the issue of borders for `pre` elements as well and treat it as a separate issue.  The reason for this is that in cases where a `pre` element has a 'copy to clipboard' option, there are actually two elements side-by-side: a `pre` element and a `highlight` element.  

Thus in a 'copy-to-clipboard' context would require a top/bottom/left border on `pre` elements and top/right/bottom border on `highlight` elements; however in the context with no 'c-2-c', the `pre` element would receive a top/right/bottom/left border.  I can't think of a trivial fix for this at the moment.  So for now, I am going to restrict the scope of this PR to strictly the hyphenation issue which I believe has been addressed by the current work.

knz:
Oh I see, thanks for explaining.
Let's keep `pre` on the side for now. Carry on!

j-low:
@knz: sure thing!  I didn't realize there was that particular complication until diving in so it's good to become aware of that issue.

PR #2442, opened 2018-01-30
Detailed SQL Standard conformance

Amruta-Ranade:
@jseldess @awoods187 
This is the first draft of the SQL Standard compatibility matrix - just making sure I am on the right track.

I was considering breaking up the table into three tables: supported, unsupported, and partially-supported features. Thoughts?


awoods187:
Great first step! Can you remove the "??" ? Let's go ahead and label those as "No." Also, is it possible to filter this list? Could someone just look at what we do or don't support?

awoods187:
Let's go ahead and mark it for 1.1

Makes sense re filter follow up

bdarnell:
We shouldn't just copy the same file from 2.0 to 1.1 - some of those checkboxes won't be the same. Unless someone wants to go through and make sure that all the differences are noted I think it would be better to start this page in 2.0 and update it going forward.

v2.0/detailed-sql-support.md
```diff
@@ -0,0 +1,692 @@
+---
+title: Detailed SQL Standard Comparison
+summary: Details of CockroachDB's conformance to the SQL standard and which common extensions it supports.
+---
+
+<div id="toc"></div>
+
+## Overview
+
+This page lists which SQL standard features are supported, partially-supported, and unsupported by CockroachDB.
```
jseldess: I'd suggest simplifying this page by removing the toc and **Overview** headingh and starting the page with this:

> This page outlines to what extent CockroachDB conforms to the current SQL standard. 
> 
> - **Feature ID** ...
> - **Description** ...
> - **CockroachDB Support** ...

jseldess: Applies also the 1.1 version of the page.

jseldess:
Makes sense to me, @bdarnell. @Amruta-Ranade, let's just remove this content from 1.1 docs. Sorry to cause confusion here.

PR #2443, opened 2018-01-30
Exclude training pages from search results

### LINKED ISSUE #2434: Exclude training from search results
Our training pages are incomplete and in flux, but they are sometimes high up in google search results. For now, we should exclude the training pages from search results.

jseldess:
Fixes #2434.


jseldess:
@benesch, @bdarnell, anything else I could/should do to get these pages excluded?

bdarnell:
LGTM

This will hide them from all search engines, which is appropriate if they're still in development. If we want to exclude them from our docs search but still let them be found on google, we'll have to do something different with the search restrictions (maybe easier with @j-low's rewrite of our search)

jseldess:
TFTR, @bdarnell. We do want to hide them from all search results for now.

PR #2456, opened 2018-02-05
Updates to training

jseldess:
- Change sidenav
- Revise the basic ops labs and slides
- Add troubleshooting labs
- Start on maintenance labs


jseldess:
Merging. Will get reviews in different format.

PR #2459, opened 2018-02-06
Update answers to When is CockroachDB a good / not good choice

jess-edwards:
Update answers to: 

When is CockroachDB a good choice? 
Adds: `Multi-region deployments, Cloud migrations`

Question: When is CockroachDB not a good choice? 
Removes: `Complex SQL Joins`

cc @awoods187 @petermattis @nstewart


jess-edwards:
@jseldess Good point. Adding a PTAL for @nstewart, as well. 

And yes, please take over and apply it to the 2.0 FAQ. Much appreciated!

Issue #1856, opened 2017-08-23
Install page should change URL for each option


nvanbenschoten:
It's often the case that I want to link to an OS/Install Method combination on https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html. Can we add these options to the URL so we can properly link into a specific option?

jseldess:
@benesch, I can't figure out why the `allowed-hashes` don't work on the install page. They work fine on https://www.cockroachlabs.com/docs/stable/start-a-local-cluster-in-docker.html. Do you have any ideas?

benesch:
IIRC allowed-hashes was a hack to avoid HTML proofer errors when another page linked to a different version. (#linux doesn't actually exist on the page, so HTML proofer would complain.)

I think the install page simply doesn't have the JavaScript necessary to support linking to a particular OS directly.

PR #2462, opened 2018-02-07
clarify install docker directions.

arjunravinarayan:
The "Make Sure You Use Docker" instructions could be a little more
user-error proofed, as by default, the link takes you to
the "install via homebrew" instructions.

If there was a way to link directly to the "use docker" link, that
would be better, but I could not figure it out.

I have not wordsmithed it, happy to take suggestions (and then I can add it to the other versions of the doc page).

### DIFF
--- a/v2.0/start-a-local-cluster-in-docker.md
+++ b/v2.0/start-a-local-cluster-in-docker.md
@@ -33,7 +33,7 @@ Once you've [installed the official CockroachDB Docker image](install-cockroachd
 <div class="filter-content" markdown="1" data-scope="os-windows">
 ## Before You Begin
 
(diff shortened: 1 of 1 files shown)


jseldess:
@arjunravinarayan, so sorry for the delay. I tried to push some minor edits to your branch but wasn't able, so I pulled this change into a separate PR: https://github.com/cockroachdb/docs/pull/2732. 

Unfortunately, we don't currently have a way to link directly to an installation method on the install page. This is a known issue that we really should resolve soon: https://github.com/cockroachdb/docs/issues/1856

In the meantime, this adds some clarification. When the user links to the install page, we have some JavaScript that exposes the options for their OS, so it really should just be a matter of looking at **Use Docker**.

PR #2468, opened 2018-02-09
Reorder footer columns to match Hugo site; update copyright year

j-low:
__[description]__
* Reorder footer columns to match Hugo site 
* update copyright year to 2018


jseldess:
LGTM overall, but there's some very minor differences you may want to check and deal with now rather than later:

- Font/font-size of column headers, e.g., Product, Resources, etc.
- Padding between columns.

Issue #2470, opened 2018-02-09
Code-formatted links aren't getting proper link styling


j-low:
<img width="289" alt="screen shot 2018-01-25 at 9 45 34 pm" src="https://user-images.githubusercontent.com/6320333/36047772-0a00abe8-0dab-11e8-81e7-cdaa0bed040d.png">

Link text is not green which conceals that it is, in fact, a link.  Status as a link is especially concealed when not hovered and no underline is visible.

All code-formatted links should be styled with the same link color as is used site-wide.

cc: @jseldess

j-low:
Addressed in #2471.  Closing issue.

PR #2471, opened 2018-02-10
Fix color of code-formatted links

j-low:
__[description]__
* re: #2470
* replace variable instance in lavish-boostrap with hex value


jseldess:
Thanks very much, @j-low. LGTM. 

The TeamCity failures should go away after you rebase against master.

PR #2475, opened 2018-02-11
 Update locality module to run nodes in background

jseldess:


petermattis:
It sure would be nice to be able to use `roachdemo` to start up these multi-node local clusters.

jseldess:
I've edited again to mirror your process, @a-robinson. @petermattis, I agree that it'd be nice to use a tool like `roachdemo` for trainings like this, but I think we may have to tackle that later?

petermattis:
> I agree that it'd be nice to use a tool like roachdemo for trainings like this, but I think we may have to tackle that later?

Yes, let's not muddle with this right now. I've been going through the modules and they are a bit manual, but I think that is probably better than hiding everything behind roachdemo. Gives a real sense for the separate processes and what not.

PR #2476, opened 2018-02-11
Add node decommissioning lab

jseldess:


a-robinson:
> We could also include a cockroach node status command in addition to / instead of waiting for the UI to list the node as decommissioned

I take it back, apparently `cockroach node status` doesn't include information about whether a node is decommissioned in v1.1.

PR #2482, opened 2018-02-12
Added partitioning doc and updated sidebar nav

Amruta-Ranade:
I am dividing the partitioning documentation task in two phases:

1. Add the stand-alone partitioning doc.
2. Update the related individual docs.

This PR is for phase 1. 

@danhhz @benesch Please review the doc for technical accuracy and completeness
@awoods187 Please review the doc for positioning and terminology
@jseldess @lhirata  Please review the doc for structure and completeness

I will copy-edit the doc extensively after the first round of reviews.

### DIFF
--- a/_includes/sidebar-data-v2.0.json
+++ b/_includes/sidebar-data-v2.0.json
@@ -1088,7 +1088,7 @@
         ]
       },
       {
-        "title": "Define Table Partitions",
+        "title": "Define Table Partitions (Enterprise)",
         "urls": [
           "/${VERSION}/partitioning.html"
         ]
--- a/_includes/sidebar-data-v2.0.json
+++ b/_includes/sidebar-data-v2.0.json
@@ -1088,6 +1088,12 @@
         ]
       },
       {
+        "title": "Partition Tables",
+        "urls": [
+          "/${VERSION}/partitioning.html"
+        ]
+      },
+      {
         "title": "Enterprise Licensing",
         "urls": [
           "/${VERSION}/enterprise-licensing.html"
--- a/_includes/sidebar-data-v2.0.json
+++ b/_includes/sidebar-data-v2.0.json
@@ -1088,7 +1088,7 @@
         ]
       },
       {
-        "title": "Partition Tables",
+        "title": "Define Table Partitions",
         "urls": [
           "/${VERSION}/partitioning.html"
         ]
--- a/v2.0/partitioning.md
+++ b/v2.0/partitioning.md
@@ -19,23 +19,51 @@ Table partitioning helps you reduce latency and cost:
 
 ## How to Use Table Partitioning
 
-1. [Start the nodes with the right constraints](#start-the-nodes-with-the-right-constraints).
-2. [Set the enterprise license key](#set-the-enterprise-license-key). 
-3. [Create table with partitions and define the primary key](#create-the-table-with-partitions-and-define-the-primary-key).
-4. [Set zone configurations](#set-zone-configurations).
-5. [Verify table partitions](#verify-table-partitions).
+1. [Identify the Partitioning method](#identify-the-partitioning-method).
+2. [Start the nodes with the right constraints](#start-the-nodes-with-the-right-constraints).
+3. [Set the enterprise license key](#set-the-enterprise-license-key). 
+4. [Define the table partitions using a primary key or secondary index](#define-table-partitions-using-primary-key-or-secondary-index).
+5. [Set zone configurations](#set-zone-configurations).
+6. [Verify table partitions](#verify-table-partitions).
 
-## Start the Nodes with the Right Constraints
+### Identify the Partitioning Method
+
+You can define partitions and subpartitions over one or more columns of a table. You can declare which values belong to each partition in one of two ways: 
+
+- **List partitioning**: Enumerate all possible values for each partition. List partitioning is a good choice when the number of possible values is small. List partitioning is well-suited for geo-partitioning.
+- **Range partitioning**: Specify a contiguous range of values for each partition by specifying lower and upper bounds. Range partitioning is a good choice when the number of possible values is too large to explicitly list out. Range partitioning is well-suited for archival-partitioning.
+
+#### Partition by List
+
+`PARTITION BY LIST` lets you map one or more tuples to a partition.
+
+To partition a table by list, use the `PARTITION BY LIST` syntax while creating the table. While defining a list partition, you can also set the `DEFAULT` partition that acts as a catch-all if none of the rows match the requirements for the defined partitions. 
+
+See [Partition by List](#partition-by-list) example below for more details.
+
+#### Partition by Range
+
+`PARTITION BY RANGE` lets you map ranges of tuples to a partition.
+
+To define a table partition by range, use the `PARTITION BY RANGE` syntax while creating the table.  While defining a range partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
+
+{{site.data.alerts.callout_info}}The lower bound of a range partition is inclusive, while the upper bound is exclusive. For range partitions, <code>NULL</code> is considered less than any other data, which is consistent with our key encoding ordering and <code>ORDER BY</code> behavior.{{site.data.alerts.end}}
+
+Partition values can be any SQL expression, but it’s only evaluated once. If you create a partition with value `< (now() - '1d')` on 2017-01-30, it would be contain all values less than 2017-01-29. It would not update the next day, it would continue to contain values less than 2017-01-29.
+
+See [Partition by Range](#partition-by-range) example below for more details.
+
+### Start the Nodes with the Right Constraints
 
 Start the nodes with the right constraits (localities, storage devices, and other constraints).
 
 See [Configure replication zones](configure-replication-zones.html) for more information.
 
-## Set the Enterprise License Key
+### Set the Enterprise License Key
 
 You require an enterprise license to partition or repartition a table, and to add or edit the zone configurations of a partition or an index. To enable partitioning, [set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key).
 
-### Expired License
+#### Expired License
 
 The following features don't work with an expired license:
 
@@ -51,17 +79,15 @@ However, the following features continue to work even with an expired enterprise
 - Unpartitioning a partitioned table
 - Making non-partitioning changes to a partitioned table (for example, adding a column/index/foreign key/check constraint) 
 
-## Create the Table with Partitions and Define the Primary Key
-
-### Define the Primary Key
+### Define Table Partitions using Primary Key or Secondary Index
 
 A table can be partitioned based on either its primary key or secondary index. 
 
+#### Define Table Partitions using Primary Key
+
 The primary key required for partitioning is different from the conventional primary key. To define the primary key for partitioning, prefix the unique identifier(s) in the primary key with all columns you want to partition and subpartition the table on, in the order in which you want to nest your subpartitions. 
 
-#### Example
-
-Consider the database of a global online learning portal that has a table for students of all the courses across the world. If you want to geo-partition the table based on the countries of the students, then the primary key needs to be defined as:
+For instance, consider the database of a global online learning portal that has a table for students of all the courses across the world. If you want to geo-partition the table based on the countries of the students, then the primary key needs to be defined as:
 
 {% include copy-clipboard.html %} 
 ~~~ sql
@@ -74,12 +100,12 @@ Consider the database of a global online learning portal that has a table for st
     PRIMARY KEY (country, id));
 ~~~
 
-### Partitioning Key Considerations
+**Primary Key Considerations**
 
 - For v2.0, you cannot change the primary key after you create the table. Provision for all future subpartitions by including those columns in the primary key. In the example of the online learning portal, if you think you might want to subpartition based on `expected_graduation_date` in the future, define the primary key as `(country, expected_graduation_date, id)`. v2.1 will allow you to change the primary key.
 - The order in which the columns are defined in the primary key is important. The partitions and subpartitions need to follow that order. In the example of the online learning portal, if you define the primary key as `(country, expected_graduation_date, id)`, the primary partition is by `country`, and then subpartition is by `expected_graduation_date`. You can’t skip `country` and partition by `expected_graduation_date`.
 
-### Partitioning and Index Columns
+#### Define Table Partitions using Secondary Index
 
 The primary key discussed above has two drawbacks: 
 
@@ -103,44 +129,17 @@ CREATE INDEX foo_b_idx ON foo (b) PARTITION BY LIST (b) (
 
 Consider using a naming scheme that uses the index name to avoid conflicts. For example, the partitions above could be named `primary_idx_bar`, `primary_idx_default`, `b_idx_baz`, `b_idx_default`.
 
-### Define the Table Partitions
-
-You can define partitions and subpartitions over one or more columns of a table. You can declare which values belong to which partition in one of two ways: 
-
-- **List partitioning**: Enumerate all possible values for each partition. List partitioning is a good choice when the number of possible values is small.
-- **Range partitioning**: Specify a contiguous range of values for each partition by specifying lower and upper bounds. Range partitioning is a good choice when the number of possible values is too large to explicitly list out.
-
-#### Partition by List
-
-`PARTITION BY LIST` lets you map one or more tuples to a partition.
-
-To partition a table by list, use the `PARTITION BY LIST` syntax while creating the table. While defining a list partition, you can also set the `DEFAULT` partition that acts as a catch-all if none of the rows match the requirements for the defined partitions. 
-
-See [Partition by List](#partition-by-list) example below for more details.
-
-#### Partition by Range
-
-`PARTITION BY RANGE` lets you map ranges of tuples to a partition.
-
-To define a table partition by range, use the `PARTITION BY RANGE` syntax while creating the table.  While defining a range partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
-
-{{site.data.alerts.callout_info}}The lower bound of a range partition is inclusive, while the upper bound is exclusive. For range partitions, <code>NULL</code> is considered less than any other data, which is consistent with our key encoding ordering and <code>ORDER BY</code> behavior.{{site.data.alerts.end}}
-
-Partition values can be any SQL expression, but it’s only evaluated once. If you create a partition with value `< (now() - '1d')` on 2017-01-30, it would be contain all values less than 2017-01-29. It would not update the next day, it would continue to contain values less than 2017-01-29.
-
-See [Partition by Range](#partition-by-range) example below for more details.
-
-### Define Partitions on Interleaved Tables
+#### Define Partitions on Interleaved Tables
 
 For [interleaved tables](interleave-in-parent.html), partitions can be defined only on the root table of the interleave hierarchy, while children are interleaved the same as their parents.
 
-## Set Zone Configurations
+### Set Zone Configurations
 
 On their own, partitions are inert and simply apply a label to the rows of the table that satisfy the criteria of the defined partitions. Applying functionality to a partition requires creating and applying the [zone configurations](configure-replication-zones.html) to the corresponding partitions.
 
 CockroachDB uses the most granular zone config available. Zone configs that target a partition are considered more granular than those that target a table or index, which in turn are considered more granular than those that target a database. 
 
-## Verify Table Partitions
+### Verify Table Partitions
 
 To verify the table partitions, use the [`SHOW CREATE TABLE`](show-create-table.html) statement.
 
@@ -153,9 +152,11 @@ Consider a global online learning portal, RoachLearn, that has a database contai
 - We want to keep the data of the students located in the United States and Canada in the United States datacenter.
 - We want to keep the data of students located in Australia and New Zealand in the Australian datacenter. 
 
-We can achieve this by partitioning on `country` and using the `PARTITION BY LIST` syntax. 
+#### Step 1. Identify the partitioning method
 
-#### Step 1. Start each node with its datacenter location specified in the `--locality` flag:
+We want to geo-partition the table to keep the students' data closer to their locations. We can achieve this by partitioning on `country` and using the `PARTITION BY LIST` syntax.
+
+#### Step 2. Start each node with its datacenter location specified in the `--locality` flag
 
 {% include copy-clipboard.html %}
 ~~~ shell
@@ -181,11 +182,11 @@ $ cockroach start --insecure \
 --join=<node1 hostname>:26257,<node2 hostname>:26258
 ~~~
 
-#### Step 2. Set the enterprise license
+#### Step 3. Set the enterprise license
 
 To set the enterprise license, see [Set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key)
 
-#### Step 3. Create a table with the appropriate partitions:
+#### Step 4. Create a table with the appropriate partitions
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -202,7 +203,7 @@ To set the enterprise license, see [Set the Trial or Enterprise License Key](ent
       PARTITION DEFAULT VALUES IN (default));
 ~~~
 
-#### Step 4. Create and apply corresponding zone configurations:
+#### Step 5. Create and apply corresponding zone configurations
 
 Create appropriate zone configurations:
 
@@ -230,7 +231,7 @@ $ cockroach zone set roachlearn.students_by_list.north_america --insecure  -f no
 $ cockroach zone set roachlearn.students_by_list.australia --insecure  -f australia.zone.yml
 ~~~
 
-#### Step 5. Verify table partitions
+#### Step 6. Verify table partitions
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -260,13 +261,17 @@ Time: 7.209032ms
 
 ### Define Table Partitions by Range
 
-Suppose we want to store the data of current students on fast and expensive storage devices (e.g., SSD) and store the data of the graduated students on slower, cheaper storage devices (e.g., HDD). We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax. 
+Suppose we want to store the data of current students on fast and expensive storage devices (e.g., SSD) and store the data of the graduated students on slower, cheaper storage devices (e.g., HDD). 
 
-#### Step 1. Set the enterprise license
+#### Step 1. Identify the partitioning method
+
+We want to archival-partition the table to keep newer data on faster devices and older data on slower devices. We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax. 
+
+#### Step 2. Set the enterprise license
 
 To set the enterprise license, see [Set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key)
 
-#### Step 2. Start each node with the appropriate storage device specified in the `--store` flag
+#### Step 3. Start each node with the appropriate storage device specified in the `--store` flag
 
 {% include copy-clipboard.html %}
 ~~~ shell
@@ -288,7 +293,7 @@ $ cockroach start --insecure \
 --join=<node1 hostname>:26257,<node2 hostname>:26258
 ~~~
 
-#### Step 3. Create a table with the appropriate partitions
+#### Step 4. Create a table with the appropriate partitions
 
 {% include copy-clipboard.html %} 
 ~~~ sql
@@ -304,7 +309,7 @@ $ cockroach start --insecure \
       PARTITION current VALUES FROM ('2017-08-15') TO (MAXVALUE));
 ~~~
 
-#### Step 4. Create and apply corresponding zone configurations
+#### Step 5. Create and apply corresponding zone configurations
 
 Create appropriate zone configurations:
 
@@ -332,7 +337,7 @@ $ cockroach zone set roachlearn.students_by_range.current --insecure  -f current
 $ cockroach zone set roachlearn.students_by_range.graduated --insecure  -f graduated.zone.yml
 ~~~
 
-#### Step 5. Verify table partitions
+#### Step 6. Verify table partitions
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -365,9 +370,11 @@ Going back to RoachLearn's scenario, suppose we want to do all of the following:
 - Store the current students' data on faster storage devices.
 - Store the graduated students' data on slower, cheaper storage devices (example: HDD).
 
-We can achieve this by partitioning the table first by location and then by date.
+#### Step 1. Identify the Partitioning method
 
-#### Step 1. Start each node with the appropriate storage device specified in the `--store` flag
+We want to geo-partition as well as archival-partition the table. We can achieve this by partitioning the table first by location and then by date.
+
+#### Step 2. Start each node with the appropriate storage device specified in the `--store` flag
 
 Start a node in the US datacenter: 
 
@@ -399,11 +406,11 @@ Initialize the cluster:
 $ cockroach init --insecure --host=<node1 hostname>
 ~~~
 
-#### Step 2. Set the enterprise license
+#### Step 3. Set the enterprise license
 
 To set the enterprise license, see [Set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key)
 
-#### Step 3. Create a table with the appropriate partitions
+#### Step 4. Create a table with the appropriate partitions
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -422,7 +429,7 @@ To set the enterprise license, see [Set the Trial or Enterprise License Key](ent
 
 Subpartition names must be unique within a table. In our example, even though `graduated` and `current` are sub-partitions of distinct partitions, they still need to be uniquely named. Hence the names `graduated_au`, `graduated_us`, and `current_au` and `current_us`.
 
-#### Step 4. Create and apply corresponding zone configurations
+#### Step 5. Create and apply corresponding zone configurations
 
 Create appropriate zone configurations:
 
@@ -472,7 +479,7 @@ $ cockroach zone set roachlearn.students.current_au --insecure -f current_au.zon
 $ cockroach zone set roachlearn.students.graduated_au --insecure -f graduated_au.zone.yml
 ~~~
 
-#### Step 5. Verify table partitions
+#### Step 6. Verify table partitions
 
 {% include copy-clipboard.html %}
 ~~~ sql
--- a/v2.0/partitioning.md
+++ b/v2.0/partitioning.md
@@ -1,5 +1,5 @@
 ---
-title: Define Table Partitions
+title: Define Table Partitions (Enterprise)
 summary: Partitioning is an enterprise feature that gives you row-level control of how and where your data is stored.
 toc: false
 ---
@@ -10,19 +10,48 @@ CockroachDB allows you to define table partitions, thus giving you row-level con
 
 <div id="toc"></div>
 
-## Overview
+## Why Use Table Partitioning
 
 Table partitioning helps you reduce latency and cost:
 
-- **Reduce latency**: Geo-partitioning allows you to keep user data close to the user, which reduces the distance that the data needs to travel, thereby reducing latency. To geo-partition a table, define location-based partitions while creating a table, create location-specific zone configurations, and apply the zone configurations to the corresponding partitions. 
-- **Reduce costs**: Archival-partitioning allows you to store infrequently-accessed data on slower and cheaper storage. To archival-partition a table, define frequency-based partitions while creating a table, create frequency-specific zone configurations with appropriate storage devices constraints, and apply the zone configurations to the corresponding partitions.
+- **Geo-partitioning** allows you to keep user data close to the user, which reduces the distance that the data needs to travel, thereby **reducing latency**. To geo-partition a table, define location-based partitions while creating a table, create location-specific zone configurations, and apply the zone configurations to the corresponding partitions. 
+- **Archival-partitioning** allows you to store infrequently-accessed data on slower and cheaper storage, thereby **reducing costs**. To archival-partition a table, define frequency-based partitions while creating a table, create frequency-specific zone configurations with appropriate storage devices constraints, and apply the zone configurations to the corresponding partitions.
 
-## Define Table Partitions
+## How to Use Table Partitioning
 
-1. [Set the enterprise license key.](enterprise-licensing.html#set-the-trial-or-enterprise-license-key) 
-2. [Start the nodes with the right constraints](configure-replication-zones.html) (localities, storage devices, and other constraints).
-3. [Create table with partitions (and subpartitions)](partitioning.html#create-a-table-with-partitions) and [define the primary key.](partitioning.html#define-the-primary-key)
-4. [Create appropriate location-specific or storage-specific zone configurations and apply zone configurations to corresponding partitions](partitioning.html#set-zone-configurations)
+1. [Start the nodes with the right constraints](#start-the-nodes-with-the-right-constraints).
+2. [Set the enterprise license key](#set-the-enterprise-license-key). 
+3. [Create table with partitions and define the primary key](#create-the-table-with-partitions-and-define-the-primary-key).
+4. [Set zone configurations](#set-zone-configurations).
+5. [Verify table partitions](#verify-table-partitions).
+
+## Start the Nodes with the Right Constraints
+
+Start the nodes with the right constraits (localities, storage devices, and other constraints).
+
+See [Configure replication zones](configure-replication-zones.html) for more information.
+
+## Set the Enterprise License Key
+
+You require an enterprise license to partition or repartition a table, and to add or edit the zone configurations of a partition or an index. To enable partitioning, [set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key).
+
+### Expired License
+
+The following features don't work with an expired license:
+
+- Creating new table partitions or adding new zone configurations for partitions
+- Changing the partitioning scheme on any table or index
+- Changing the zone config for a partition
+
+However, the following features continue to work even with an expired enterprise license:
+
+- Querying a partitioned table (for example, `SELECT foo PARTITION`)
+- Inserting or updating data in a partitioned table
+- Dropping a partitioned table
+- Unpartitioning a partitioned table
+- Making non-partitioning changes to a partitioned table (for example, adding a column/index/foreign key/check constraint) 
+
+## Create the Table with Partitions and Define the Primary Key
 
 ### Define the Primary Key
 
@@ -45,12 +74,12 @@ Consider the database of a global online learning portal that has a table for st
     PRIMARY KEY (country, id));
 ~~~
 
-#### Partitioning Key Considerations
+### Partitioning Key Considerations
 
-- Once you set the primary key, you can’t change it later. Provision for all future subpartitions by including those columns in the primary key. In the example of the online learning portal, if you think you might want to subpartition based on `expected_graduation_date` in the future, define the primary key as `(country, expected_graduation_date, id)`.
+- For v2.0, you cannot change the primary key after you create the table. Provision for all future subpartitions by including those columns in the primary key. In the example of the online learning portal, if you think you might want to subpartition based on `expected_graduation_date` in the future, define the primary key as `(country, expected_graduation_date, id)`. v2.1 will allow you to change the primary key.
 - The order in which the columns are defined in the primary key is important. The partitions and subpartitions need to follow that order. In the example of the online learning portal, if you define the primary key as `(country, expected_graduation_date, id)`, the primary partition is by `country`, and then subpartition is by `expected_graduation_date`. You can’t skip `country` and partition by `expected_graduation_date`.
 
-#### Partitioning and Index Columns
+### Partitioning and Index Columns
 
 The primary key discussed above has two drawbacks: 
 
@@ -74,66 +103,44 @@ CREATE INDEX foo_b_idx ON foo (b) PARTITION BY LIST (b) (
 
 Consider using a naming scheme that uses the index name to avoid conflicts. For example, the partitions above could be named `primary_idx_bar`, `primary_idx_default`, `b_idx_baz`, `b_idx_default`.
 
-### Create a Table with Partitions
+### Define the Table Partitions
 
 You can define partitions and subpartitions over one or more columns of a table. You can declare which values belong to which partition in one of two ways: 
 
-- List partitioning: Enumerate all possible values for each partition. List partitioning is a good choice when the number of possible values is small.
-- Range partitioning: Specify a contiguous range of values for each partition by specifying lower and upper bounds. Range partitioning is a good choice when the number of possible values is too large to explicitly list out.
+- **List partitioning**: Enumerate all possible values for each partition. List partitioning is a good choice when the number of possible values is small.
+- **Range partitioning**: Specify a contiguous range of values for each partition by specifying lower and upper bounds. Range partitioning is a good choice when the number of possible values is too large to explicitly list out.
 
-#### Define Table Partitions by List
+#### Partition by List
 
-To partition a table by list, use the `PARTITION BY LIST` syntax while creating the table. `PARTITION BY LIST` lets you map one or more tuples to a partition.
+`PARTITION BY LIST` lets you map one or more tuples to a partition.
 
-While defining a list partition, you can also set the `DEFAULT` partition that acts as a catch-all if none of the rows match the requirements for the defined partitions. 
+To partition a table by list, use the `PARTITION BY LIST` syntax while creating the table. While defining a list partition, you can also set the `DEFAULT` partition that acts as a catch-all if none of the rows match the requirements for the defined partitions. 
 
-See [Partition by List example](#partition-by-list).
+See [Partition by List](#partition-by-list) example below for more details.
 
-#### Define Table Partitions by Range
+#### Partition by Range
 
-To define a table partition by range, use the `PARTITION BY RANGE` syntax while creating the table. `PARTITION BY LIST` lets you map ranges of tuples to a partition.
+`PARTITION BY RANGE` lets you map ranges of tuples to a partition.
 
-While defining a range partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
+To define a table partition by range, use the `PARTITION BY RANGE` syntax while creating the table.  While defining a range partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
 
-{{site.data.alerts.callout_info}}The lower bound of a range partition is inclusive, while the upper bound is exclusive. For range partitions, `NULL` is considered less than any other data, which is consistent with our key encoding ordering and `ORDER BY` behavior.{{site.data.alerts.end}}
-
-See [Partition by Range example](#partition-by-range).
+{{site.data.alerts.callout_info}}The lower bound of a range partition is inclusive, while the upper bound is exclusive. For range partitions, <code>NULL</code> is considered less than any other data, which is consistent with our key encoding ordering and <code>ORDER BY</code> behavior.{{site.data.alerts.end}}
 
 Partition values can be any SQL expression, but it’s only evaluated once. If you create a partition with value `< (now() - '1d')` on 2017-01-30, it would be contain all values less than 2017-01-29. It would not update the next day, it would continue to contain values less than 2017-01-29.
 
-#### Define Partitions on Interleaved Tables
+See [Partition by Range](#partition-by-range) example below for more details.
+
+### Define Partitions on Interleaved Tables
 
 For [interleaved tables](interleave-in-parent.html), partitions can be defined only on the root table of the interleave hierarchy, while children are interleaved the same as their parents.
 
-### Set Zone Configurations
+## Set Zone Configurations
 
 On their own, partitions are inert and simply apply a label to the rows of the table that satisfy the criteria of the defined partitions. Applying functionality to a partition requires creating and applying the [zone configurations](configure-replication-zones.html) to the corresponding partitions.
 
 CockroachDB uses the most granular zone config available. Zone configs that target a partition are considered more granular than those that target a table or index, which in turn are considered more granular than those that target a database. 
 
-### Synopsis
-
-~~~ shell
-# Set enterprise license
-$ SET CLUSTER SETTING cluster.organization = 'organization_name';
-$ SET CLUSTER SETTING enterprise.license = 'xxxxxxxxxxxx'; 
-
-# Partition a table by list (geo-partitioning)
-$ CREATE TABLE <table-name> ( <elements...>, PRIMARY KEY (<primary_keys>))
-  PARTITION BY LIST(primary_key_component) (PARTITION <partition-name> VALUES IN ( <list-expr>... ));
-
-# Partition a table by range (archival-partitioning)
-$ CREATE TABLE <table-name> ( <elements...>, PRIMARY KEY (<primary_keys>)) 
-  PARTITION BY RANGE(primary_key_component) (PARTITION <partition-name> VALUES FROM <range-expr> TO <range-expr>);
-
-# Create/edit the replication zone for a partition:
-$ cockroach zone set <database.table.partition> --file=<zone-content.yaml> <flags>
-
-# Remove the replication zone for a partition:
-$ cockroach zone rm <database.table.partition> <flags>
-~~~
-
-### Verify Table Partitions
+## Verify Table Partitions
 
 To verify the table partitions, use the [`SHOW CREATE TABLE`](show-create-table.html) statement.
 
@@ -148,22 +155,39 @@ Consider a global online learning portal, RoachLearn, that has a database contai
 
 We can achieve this by partitioning on `country` and using the `PARTITION BY LIST` syntax. 
 
-#### Step 1: Start each node with its datacenter location specified in the `--locality` flag:
+#### Step 1. Start each node with its datacenter location specified in the `--locality` flag:
 
+{% include copy-clipboard.html %}
 ~~~ shell
 # Start the node in the US datacenter:
-$ cockroach start --insecure --locality=region=us-1  --store=node1 --host=<node1 hostname> --port=26257 --http-port=8080 --join=<node1 hostname>:26257,<node2 hostname>:26258
-
-# Start the node in the AUS datacenter:
-$ cockroach start --insecure --locality=region=aus-1 --store=node2 --host=<node2 hostname> --port=26258 --http-port=8081 --join=<node1 hostname>:26257,<node2 hostname>:26258
+$ cockroach start --insecure \
+--locality=region=us-1  \
+--store=node1 \
+--host=<node1 hostname> \
+--port=26257 \
+--http-port=8080 \
+--join=<node1 hostname>:26257,<node2 hostname>:26258
 ~~~
 
-#### Step 2: Set the enterprise license
+{% include copy-clipboard.html %}
+~~~ shell
+# Start the node in the AUS datacenter:
+$ cockroach start --insecure \
+--locality=region=aus-1 \
+--store=node2 \
+--host=<node2 hostname> \
+--port=26258 \
+--http-port=8081 \
+--join=<node1 hostname>:26257,<node2 hostname>:26258
+~~~
+
+#### Step 2. Set the enterprise license
 
 To set the enterprise license, see [Set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key)
 
-#### Step 3: Create a table with the appropriate partitions:
+#### Step 3. Create a table with the appropriate partitions:
 
+{% include copy-clipboard.html %}
 ~~~ sql
 > CREATE TABLE students_by_list (
     id SERIAL,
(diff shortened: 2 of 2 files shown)


lhirata:
Review status: 0 of 2 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[v2.0/partitioning.md, line 20 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5A_8GnQshcZxAKQZeG:-L5A_8GnQshcZxAKQZeH:b-908ltw) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L20)):*
> ```Markdown
> - **Reduce costs**: Archival-partitioning allows you to store infrequently-accessed data on slower and cheaper storage. To archival-partition a table, define frequency-based partitions while creating a table, create frequency-specific zone configurations with appropriate storage devices constraints, and apply the zone configurations to the corresponding partitions.
> 
> ## Partitioning a Table
> ```

Change these headings to be active: Partition a Table, Define the Primary Key, Create a Table.., etc.

---

*[v2.0/partitioning.md, line 24 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AYswq8mCQpCDN6kML:-L5AYswq8mCQpCDN6kMM:b-bis1fg) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L24)):*
> ```Markdown
> You can define partitions and subpartitions over one or more columns of a table. You can repartition and unpartition existing partitions. You can also use partitioning with indexes and interleaved tables. 
> 
> To partition a table:
> ```

These steps don't display as a list. I would recommend using a numbered list instead of "Step 1". (This is something we should note for our inevitable style guide)

---

*[v2.0/partitioning.md, line 26 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AZMyNJcs4noaKKd40:-L5AZMyNJcs4noaKKd41:btt6a4q) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L26)):*
> ```Markdown
> To partition a table:
> 
> Step 1: Set the enterprise license key. 
> ```

Should these steps be linked to the subheadings? I was expecting the subheadings to follow the list of prescribed steps but they don't, which is a little confusing.

---

*[v2.0/partitioning.md, line 40 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ab4xsTRCgO2MktSLG:-L5Ab4xsTRCgO2MktSLH:b-ej3vi9) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L40)):*
> ```Markdown
> #### Example
> 
> Consider the database of a global online learning portal that has a table for students of all the courses across the world. Suppose you want to geo-partition the table based on the countries of the students. Then the composite primary key needs to be defined as:
> ```

"Suppose you want to [...]" > "If you want to [...] , then the composite primary key needs to be defined as:"

---

*[v2.0/partitioning.md, line 42 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AbKsLypC0WInIcRGP:-L5AbKsLypC0WInIcRGQ:b-h4j73n) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L42)):*
> ```Markdown
> Consider the database of a global online learning portal that has a table for students of all the courses across the world. Suppose you want to geo-partition the table based on the countries of the students. Then the composite primary key needs to be defined as:
> 
> ~~~ sql
> ```

Should this have a {% include copy-clipboard.html %} ?

---

*[v2.0/partitioning.md, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AbhSkXdXtA-SZvunY:-L5AbhSl4NMwe3gxSYU1:b-w9no0q) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L54)):*
> ```Markdown
> #### Partitioning Key Considerations
> 
> - Once you set the composite primary key, you can’t change it later. So provision for all future subpartitions by including those columns in the composite primary key. In the example of the online learning portal, if you think you might want to subpartition based on `last_seen` in the future, define the composite primary key as `(country, last_seen, id)`.
> ```

"So provision for all future [...]" > "Provision for all [...]"

---

*[v2.0/partitioning.md, line 59 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AcAEzyaG0BpWU-7W4:-L5AcAEzyaG0BpWU-7W5:b-w5wmmm) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L59)):*
> ```Markdown
> #### Partitioning and Index Columns
> 
> The composite primary key discussed above has two drawbacks: It does not enforce that the identifier column is globally unique, and it does not provide fast lookups on the identifier. To ensure uniqueness or fast lookups, create a unique, unpartitioned secondary index on the identifier.
> ```

You might want to format this differently. Suggestion:

"The composite primary key discussed above has two drawbacks:

1. It does not...
2. It does not...

To ensure uniqueness..."

---

*[v2.0/partitioning.md, line 63 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ac_VTaUgx9DhTaaLZ:-L5Ac_VTaUgx9DhTaaL_:b5ghwau) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L63)):*
> ```Markdown
> Indexes can also be partitioned, but are not required to be. Each partition is required to have a name that is unique among all partitions on that table or index.
> 
> See [Partitioning Indexes example](#partitioned-secondary-indexes).
> ```

"See [Partitioning...]" > "See the [Partitioning...]"

---

*[v2.0/partitioning.md, line 69 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ad7gqHLSTRvwryBZv:-L5Ad7gqHLSTRvwryBZw:bagv5rm) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L69)):*
> ```Markdown
> You can partition a table either by list or by range:
> 
> #### Partition a table by list
> ```

Change to title case

---

*[v2.0/partitioning.md, line 75 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ad9m0vgM0DjVDQ4E8:-L5Ad9m0vgM0DjVDQ4E9:biuyu32) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L75)):*
> ```Markdown
> While defining a list partition, you can also set the `DEFAULT` partition that acts as a catch-all if none of the rows match the requirements for the defined partitions. 
> 
> See [Partition by List example](#partition-by-list).
> ```

"See [Partition...]" > "See the [Partition...]"

---

*[v2.0/partitioning.md, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdHJx4yHOWttkulOJ:-L5AdHJx4yHOWttkulOK:bagv5rm) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L77)):*
> ```Markdown
> See [Partition by List example](#partition-by-list).
> 
> #### Partition a table by range
> ```

Change to title case

---

*[v2.0/partitioning.md, line 79 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdWzq2nLt7Qr5LynE:-L5AdWzq2nLt7Qr5LynF:bze0lsr) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L79)):*
> ```Markdown
> #### Partition a table by range
> 
> To partition a table by range, use `PARTITION BY RANGE` syntax while creating the table. `PARTITION BY LIST` lets you map ranges of tuples to a partition.
> ```

Change to "use the `PARTITION BY RANGE` syntax [...]"

---

*[v2.0/partitioning.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdfoDh-iXL5ZiKKCb:-L5AdfoDh-iXL5ZiKKCc:bl0p8x6) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L83)):*
> ```Markdown
> While defining a list partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
> 
> {{site.data.alerts.callout_info}}Note that the lower bound of a range partition is inclusive, while the upper bound is exclusive. Note also that a `NULL` value in a range-partitioned column sorts into the first range, which is consistent with our key encoding ordering and `ORDER BY` behavior.{{site.data.alerts.end}}
> ```

Redundant - change "Note that the lower [...]" > "The lower bound of a range [...]". Also, change the back ticks to <code>.

---

*[v2.0/partitioning.md, line 85 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdzNE8nAmpecM8gEC:-L5AdzNE8nAmpecM8gED:biuyu32) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L85)):*
> ```Markdown
> {{site.data.alerts.callout_info}}Note that the lower bound of a range partition is inclusive, while the upper bound is exclusive. Note also that a `NULL` value in a range-partitioned column sorts into the first range, which is consistent with our key encoding ordering and `ORDER BY` behavior.{{site.data.alerts.end}}
> 
> See [Partition by Range example](#partition-by-range).
> ```

"See [Partition...]" > "See the [Partition...]"

---

*[v2.0/partitioning.md, line 87 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AeTe2GkYFbb_i5Weu:-L5AeTe2GkYFbb_i5Wev:b-3iw4it) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L87)):*
> ```Markdown
> See [Partition by Range example](#partition-by-range).
> 
> Partition values can be any SQL expression, but it’s only evaluated once. so if you create a partition with value `< (now() - '1d')` on 2017-01-30, it would be contain all values less than 2017-01-29. It would not update the next day, it would continue to contain values less than 2017-01-29.
> ```

"so if you create a partition with value `< (now() - '1d')` on 2017-01-30, it would be contain all values less than 2017-01-29." > "If you create [...], it would contain all values less than [...]"

---

*[v2.0/partitioning.md, line 91 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Af-Z1J-VtHm_cFduQ:-L5Af-Z1J-VtHm_cFduR:bxgi5gy) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L91)):*
> ```Markdown
> #### Partitioning Interleaved Tables
> 
> Partitioning [interleaved tables](interleave-in-parent.html) partitions only the parent table, while children will naturally be interleaved the same as their parents.
> ```

This is a little confusing to me. Maybe it can be clarified by saying something like "Partitioning interleaved tables affects only the parent table, [...]" 

---

*[v2.0/partitioning.md, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Afy-1x1tXZCVlpcvb:-L5Afy-1x1tXZCVlpcvc:b-mq80xf) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L95)):*
> ```Markdown
> ### Setting Zone Configurations
> 
> On their own, partitions are inert. Applying functionality to a partition requires [zone configs](configure-replication-zones.html).
> ```

Is "config" considered jargon / do we have any guidelines around when you should use "config" vs. "configuration"?

---

*[v2.0/partitioning.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ago-6SVsJ1m4Bjs7T:-L5Ago-6SVsJ1m4Bjs7U:btume9d) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L108)):*
> ```Markdown
> # Partition a table by list (geo-partitioning)
> $ CREATE TABLE <table-name> ( <elements...>, PRIMARY KEY (<primary_keys>))
>   PARTITION BY LIST(primary_key_component)<partition-name> VALUES IN ( <list-expr>... )
> ```

Does this need ; at the end?

---

*[v2.0/partitioning.md, line 112 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AgkJBcUFYM_1hVX_n:-L5AgkJBcUFYM_1hVX_o:btume9d) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L112)):*
> ```Markdown
> # Partition a table by range (archival partitioning)
> $ CREATE TABLE <table-name> ( <elements...>, PRIMARY KEY (<primary_keys>)) 
>   PARTITION BY RANGE(primary_key_component) <partition-name> VALUES FROM <range-expr> TO <range-expr>
> ```

Does this need ; at the end?

---

*[v2.0/partitioning.md, line 129 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AhM8Y8NpF5wAkFQLA:-L5AhM8Y8NpF5wAkFQLB:b-8uqec0) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L129)):*
> ```Markdown
> ### Partition by List
> 
> Consider a table of the users of a global e-commerce website, RoachMart. Suppose we have two datacenters: one in the United States and another in Australia. To reduce latency, we want to keep the users' data closer to their locations: We want to keep data of the users located in United States and Canada in the United States datacenter, and the data of users located in Australia and New Zealand in the Australian datacenter. We can achieve this by partitioning on country and use the `PARTITION BY LIST` syntax. 
> ```

Don't capitalize after the colon.  Suggestion: change "To reduce latency, we want to keep the users' data closer to their locations: We want to keep data of the users [...]" to "To reduce latency, we want to keep the users' data closer to their locations, i.e., we want to keep data of the users [...]"

---

*[v2.0/partitioning.md, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AiAu_nwZDm2FOpFJq:-L5AiAu_nwZDm2FOpFJr:b-xu7odr) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L189)):*
> ```Markdown
> ### Partition by Range
> 
> Consider the table of users of the same global e-commerce website, RoachMart. Suppose we want to store the recent users on fast and expensive storage devices (example: ssd) and store the data of the users who have not been to the website recently on slower, cheaper storage devices(example: hdd). We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax. 
> ```

Missing space in "cheaper storage devices(example: hdd)."

---

*[v2.0/partitioning.md, line 235 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AibXo2KZxco96xsV2:-L5AibXo2KZxco96xsV3:b-1jp24z) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L235)):*
> ```Markdown
> ### Sub-Partition a Table
> 
> The `PARTITION BY` syntax is recursive so that partitions can be themselves partitioned any number of times, using either list or range partitioning. 
> ```

This sentence is a little confusing. Suggestion: "The PARTITION BY syntax is recursive; partitions can be used any number of times, using either list or range partitioning."

---

*[v2.0/partitioning.md, line 237 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Aj74W2st-Vys-PImF:-L5Aj74W2st-Vys-PImG:b-eol1k3) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L237)):*
> ```Markdown
> The `PARTITION BY` syntax is recursive so that partitions can be themselves partitioned any number of times, using either list or range partitioning. 
> 
> Going back to RoachMart's scenario, suppose we want to keep the users' data close to their location, as well as store the recent users' data on faster storage devices and store the data of the users who have not been to the website recently on slower, cheaper storage devices(example: hdd). We can achieve this by partitioning the table first by location and then by date. 
> ```

This is a really long sentence that I think you can format differently so it's easier to read. Suggestion:

"Going back to RoachMart's scenario, suppose we want to do all of the following:

- Keep the users' data close to their location
- Store the recent users' data on faster storage devices
- Store the data of the users who have not been to the website recently on slower, cheaper storage devices (example: hdd)

We can achieve this by partitioning the table first by location and then by date."

---

*[v2.0/partitioning.md, line 270 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AjencdZuZkQ4Eu19M:-L5AjencdZuZkQ4Eu19N:bguhzh8) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L270)):*
> ```Markdown
> ~~~
> 
> {{site.data.alerts.callout_info}} Subpartition names must be unique within a table. In our example, even though `archived` and `recent` are sub-partitions of distinct partitions, they still need to be uniquely named. Hence the names `archived_au`, `archived_us`, and `recent_au` and `recent_us`).{{site.data.alerts.end}}
> ```

Change back ticks to <code></code>

---

*[v2.0/partitioning.md, line 322 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ak3LvK28YmAbDxHDP:-L5Ak3LvK28YmAbDxHDQ:b-p4d32j) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L322)):*
> ```Markdown
> ## Licensing Requirements
> 
> You require an enterprise license to partition or repartition a table, and to add or edit the zone configurations of a partition or an index. To enable partitioning, [set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key).
> ```

Change to "An enterprise license is required to partition [...]"

---

*[v2.0/partitioning.md, line 326 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AkHkfFczfV7HwL9T_:-L5AkHkfFczfV7HwL9Ta:bd2u11p) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L326)):*
> ```Markdown
> ### Expired License
> 
> The following features will no longer work with an expired license:
> ```

Change "will no longer work" to "will not work" (unless you were previously able to do use these features with an expired license?)

---

*[v2.0/partitioning.md, line 333 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AkTyEiV-Umah2TMrd:-L5AkTyEiV-Umah2TMre:bkyvkqw) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L333)):*
> ```Markdown
> However, the following features will continue to work even with an expired enterprise license:
> 
> - Querying on a partitioned table (e.g., SELECT foo PARTITION)
> ```

Add back ticks around `SELECT foo PARTITION`

---

*[v2.0/partitioning.md, line 337 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AkcDHVl1nVF7nt4Qb:-L5AkcDHVl1nVF7nt4Qc:b-fkydw1) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L337)):*
> ```Markdown
> - Dropping a partitioned table
> - Unpartitioning a partitioned table
> - Making non-partitioning changes to a partitioned table (e.g., adding a column/index/fk/check constraint) 
> ```

"fk" > "foreign key"

---

*[v2.0/partitioning.md, line 341 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Akn2EubciGtDOdFsq:-L5Akn2EubciGtDOdFsr:bkogoia) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L341)):*
> ```Markdown
> ## Locality–Resilience Tradeoff
> 
> There exists a tradeoff between making reads/writes fast and surviving failures. Consider a partition with three replicas of roachmart.users for Australian users. If only one replica is pinned to an Australian datacenter, then reads may be fast (via leases follow the sun) but writes will be slow. If two replicas are pinned to an Australian datacenter, than reads and writes will be fast (as long as the cross-ocean link has enough bandwidth that the third replica doesn’t fall behind and hit the quota pool). If those two replicas are in the same datacenter, then loss of one datacenter can lead to data unavailability, so some deployments may want two separate Austrialian datacenters. If all three replicas are in Australian datacenters, then three Australian datacenters are needed to be resilient to a datacenter loss.
> ```

"There exists a tradeoff [...]" > "There is a tradeoff"

Put back ticks around `roachmart.users`

"(via leases follow the sun)" > "following the sun" (But don't we call this "following the workload?" Also, maybe we should link to relevant docs here) 

---

*[v2.0/partitioning.md, line 349 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AlWR5NBfap-MYH5eM:-L5AlWR6vFBhy0zsBCfR:b4m6x32) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L349)):*
> ```Markdown
> - MySQL and Oracle specify partitions inline in `CREATE TABLE`, similar to CockroachDB.
> - Microsoft SQL Server requires four steps: allocating physical storage called “filegroups” for each partition with `ALTER DATABASE… ADD FILE`, followed by `CREATE PARTITION FUNCTION` to define the partition split points, followed by `CREATE PARTITION SCHEME` to tie the partition function output to the created file groups, followed by `CREATE TABLE... ON partition_scheme` to tie the table to the partitioning scheme.
> - PostgreSQL 10 takes a hybrid approach: the partition columns and scheme (i.e, RANGE or LIST) are specified in the `CREATE TABLE` statement, but the partition split points are specified by running `CREATE TABLE… PARTITION OF… FOR VALUES`once for each partition.
> ```

Add space after `CREATE TABLE… PARTITION OF… FOR VALUES`

---

*[v2.0/partitioning.md, line 351 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AlcmQi676_8XD8gZI:-L5AlcmQi676_8XD8gZJ:b-ypmkfo) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L351)):*
> ```Markdown
> - PostgreSQL 10 takes a hybrid approach: the partition columns and scheme (i.e, RANGE or LIST) are specified in the `CREATE TABLE` statement, but the partition split points are specified by running `CREATE TABLE… PARTITION OF… FOR VALUES`once for each partition.
> 
> Other databases use the partitioning for three additional use cases: secondary indexes, sharding, and bulk loading/deleting. CockroachDB addresses these use-cases not by using partitioning, but in the following ways:
> ```

"Other databases use the partitioning" > "Other databases use partitioning"

---

*[v2.0/partitioning.md, line 353 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Alur775yZ-u4nukqy:-L5Alur775yZ-u4nukqz:b-hlf0kw) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L353)):*
> ```Markdown
> Other databases use the partitioning for three additional use cases: secondary indexes, sharding, and bulk loading/deleting. CockroachDB addresses these use-cases not by using partitioning, but in the following ways:
> 
> - Changes to secondary indexes: CockroachDB solves these changes through online schema changes. Online schema changes are a superior feature to partitioning because they require zero-downtime and eliminate the potential for consistency problems. 
> ```

Bold the beginning of each bullet point

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

lhirata:
Nice draft!

---

Review status: 0 of 2 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-:-L5AmG2veCs60p377a-k:b-6dk4n6)*

danhhz:
:raised_hands: here's some initial comments. @benesch, i'd love for your feedback as well

---

Review status: 0 of 2 files reviewed at latest revision, 54 unresolved discussions, all commit checks successful.

---

*[v2.0/partitioning.md, line 2 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KdvUgnVPvDLofZuJL:-L5KdvUgnVPvDLofZuJM:b-jbcp0) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L2)):*
> ```Markdown
> ---
> title: Partition Tables 
> ```

this is more @awoods187's domain, but I'm fairly sure we decided to only use "Geo Partitioning" and "Archival Partitioning" to attempt to avoid confusion, so we may want to make the page title reflect that

but if I'm wrong about that, I prefer "Partitioned Tables". "Partition Tables" sounds to me like something else

---

*[v2.0/partitioning.md, line 34 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Keuhdl0vtECUiuNsj:-L5Keuhdl0vtECUiuNsk:b-mat6w2) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L34)):*
> ```Markdown
> ### Defining the Primary Key
> 
> A table can be partitioned based on either its composite primary key or secondary index. 
> ```

the repeated use of "composite primary key" is odd to me (and confusing). i'd remove all of the references to "composite" except perhaps once when introducing how the primary key in a partitioned table might be different than what would be used in the unpartitioned equivalent of the same table

---

*[v2.0/partitioning.md, line 55 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KgXC2364wMfC9KbUs:-L5KgXC2364wMfC9KbUt:br0l6fy) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L55)):*
> ```Markdown
> 
> - Once you set the composite primary key, you can’t change it later. So provision for all future subpartitions by including those columns in the composite primary key. In the example of the online learning portal, if you think you might want to subpartition based on `last_seen` in the future, define the composite primary key as `(country, last_seen, id)`.
> - The order in which the columns are defined in the composite primary key is important. The partitions and subpartitions need to follow that order. In the example of the online learning portal, if you define the composite primary key as `(country, last_seen, id,)`, the primary partition is by `country`, and then subpartition is by `last_seen`. You can’t skip `country` and partition by `last_seen`.
> ```

nit: extra comma after `id`

---

*[v2.0/partitioning.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KgcssqTDvotG5eVuU:-L5KgcssqTDvotG5eVuV:bf85zrf) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L61)):*
> ```Markdown
> The composite primary key discussed above has two drawbacks: It does not enforce that the identifier column is globally unique, and it does not provide fast lookups on the identifier. To ensure uniqueness or fast lookups, create a unique, unpartitioned secondary index on the identifier.
> 
> Indexes can also be partitioned, but are not required to be. Each partition is required to have a name that is unique among all partitions on that table or index.
> ```

unique among all partitions on that table and its indexes

---

*[v2.0/partitioning.md, line 81 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdUC7I4OgtVfeDxrL:-L5AdUC7I4OgtVfeDxrM:b-tfi6dx) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L81)):*
> ```Markdown
> To partition a table by range, use `PARTITION BY RANGE` syntax while creating the table. `PARTITION BY LIST` lets you map ranges of tuples to a partition.
> 
> While defining a list partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
> ```

list partition -> range partition

---

*[v2.0/partitioning.md, line 83 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AdZsIiRNyOlCrJMBN:-L5AdZsIiRNyOlCrJMBO:b-9ncfkq) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L83)):*
> ```Markdown
> While defining a list partition, you can use CockroachDB-defined `MINVALUE` and `MAXVALUE` parameters to define the lower and upper bounds of the ranges respectively. 
> 
> {{site.data.alerts.callout_info}}Note that the lower bound of a range partition is inclusive, while the upper bound is exclusive. Note also that a `NULL` value in a range-partitioned column sorts into the first range, which is consistent with our key encoding ordering and `ORDER BY` behavior.{{site.data.alerts.end}}
> ```

"first" is no longer quite right since you can now define range partitions in any order. maybe say something like "For the purpose of range partitions, NULL is considered less than any other data, which is consistent with our ..."

for your understanding (i don't think we need to call this out in the docs) it's possible to define a range partition `FROM (NULL) TO (1)`

---

*[v2.0/partitioning.md, line 91 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AePMkzEvj_h5jy0oa:-L5AePMkzEvj_h5jy0ob:b-qnl4dh) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L91)):*
> ```Markdown
> #### Partitioning Interleaved Tables
> 
> Partitioning [interleaved tables](interleave-in-parent.html) partitions only the parent table, while children will naturally be interleaved the same as their parents.
> ```

is only allowed on the parent table. also, it's technically only on the "root table" of the "interleave hierarchy" both of which are jargon introduced in the interleaved tables docs

---

*[v2.0/partitioning.md, line 95 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KhZkppnj3xVdwISQT:-L5KhZkppnj3xVdwISQU:bsjyg5a) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L95)):*
> ```Markdown
> ### Setting Zone Configurations
> 
> On their own, partitions are inert. Applying functionality to a partition requires [zone configs](configure-replication-zones.html).
> ```

partitions are inert and simply give a name to a subset of a table (or rows in a table or something like that)

---

*[v2.0/partitioning.md, line 97 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KhoMHs10jdXQXVfCz:-L5KhoMHs10jdXQXVfD-:bygog9o) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L97)):*
> ```Markdown
> On their own, partitions are inert. Applying functionality to a partition requires [zone configs](configure-replication-zones.html).
> 
> CockroachDB uses the most granular zone config available. Zone configs that target a partition are considered more granular than those that target a table, which in turn are considered more granular than those that target a database. 
> ```

we talk about partitioned indexes above, but not here. (index would be between table and partition)

---

*[v2.0/partitioning.md, line 108 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AekIo9hzrvb71AgY4:-L5AekIo9hzrvb71AgY5:b-6mkpaf) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L108)):*
> ```Markdown
> # Partition a table by list (geo-partitioning)
> $ CREATE TABLE <table-name> ( <elements...>, PRIMARY KEY (<primary_keys>))
>   PARTITION BY LIST(primary_key_component)<partition-name> VALUES IN ( <list-expr>... )
> ```

This is missing parens around `<partition-name> VALUES IN ( <list-expr>... )` and also `<partition-name> VALUES IN ( <list-expr>... )` can be repeated. Same with below

---

*[v2.0/partitioning.md, line 110 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KiD16vaxLXdFBbe_y:-L5KiD16vaxLXdFBbe_z:b87kwum) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L110)):*
> ```Markdown
>   PARTITION BY LIST(primary_key_component)<partition-name> VALUES IN ( <list-expr>... )
> 
> # Partition a table by range (archival partitioning)
> ```

nit: archival-partitioning is hyphenated in other places

---

*[v2.0/partitioning.md, line 115 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AesQCEcc_1V18ljj2:-L5AesQCEcc_1V18ljj3:b-a06qa6) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L115)):*
> ```Markdown
> 
> # Create/edit the replication zone for a partition:
> $ cockroach zone set <database.table.partition> --file=<zone-conent.yaml> <flags>
> ```

nit: zone-content

---

*[v2.0/partitioning.md, line 129 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Ki_gDDajb6ckWAuRt:-L5Ki_gDDajb6ckWAuRu:bbbkqmq) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L129)):*
> ```Markdown
> ### Partition by List
> 
> Consider a table of the users of a global e-commerce website, RoachMart. Suppose we have two datacenters: one in the United States and another in Australia. To reduce latency, we want to keep the users' data closer to their locations: We want to keep data of the users located in United States and Canada in the United States datacenter, and the data of users located in Australia and New Zealand in the Australian datacenter. We can achieve this by partitioning on country and use the `PARTITION BY LIST` syntax. 
> ```

i think there's an opportunity to consolidate on fewer examples. maybe pick one of `students` or`roachmart` for list partitioning

---

*[v2.0/partitioning.md, line 149 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KjAA7qR3Y2qcRYZBC:-L5KjAA7qR3Y2qcRYZBD:b-msi7nx) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L149)):*
> ```Markdown
> 
> ~~~ sql
> > CREATE TABLE users_by_list (
> ```

why the `_by_list` suffix?

---

*[v2.0/partitioning.md, line 156 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KjFJBLpCUuBPXkgio:-L5KjFJBLpCUuBPXkgip:b-i9ro7m) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L156)):*
> ```Markdown
>     last_seen DATE,   
>     PRIMARY KEY (country, last_seen, id))
>     PARTITION BY LIST (country)(PARTITION north_america VALUES IN ('CA','US'), PARTITION australia VALUES IN ('AU','NZ'), PARTITION DEFAULT VALUES IN (default));
> ```

put each PARTITION ... VALUES IN on its own line, indented in from PARTITION BY LIST

---

*[v2.0/partitioning.md, line 181 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KjOeEKbuf-ft2ufZY:-L5KjOeEKbuf-ft2ufZZ:b8k6ugo) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L181)):*
> ```Markdown
> 
> ~~~ sql
> > CREATE INDEX part_idx ON roachmart.orders (user_country, part_id)
> ```

you haven't introduced `orders` anywhere

---

*[v2.0/partitioning.md, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Kjw7NikGNd64Kn0jk:-L5Kjw7NikGNd64Kn0jl:b-ew9u2y) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L189)):*
> ```Markdown
> ### Partition by Range
> 
> Consider the table of users of the same global e-commerce website, RoachMart. Suppose we want to store the recent users on fast and expensive storage devices (example: ssd) and store the data of the users who have not been to the website recently on slower, cheaper storage devices(example: hdd). We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax. 
> ```

this example is a bit contrived, since updating a user's pk on every website access is going to be really expensive. I think you want something where the date you're using for partitioning is write-once

---

*[v2.0/partitioning.md, line 265 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Kjosy3W9vTachQk2B:-L5Kjosy3W9vTachQk2C:b-373pqq) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L265)):*
> ```Markdown
>     PRIMARY KEY (country, last_seen, id))
>     PARTITION BY LIST (country)(
>         PARTITION australia VALUES IN ('AU','NZ') PARTITION BY RANGE (last_seen)(PARTITION archived_au VALUES FROM (MINVALUE) TO ('2017-06-04'), PARTITION recent_au VALUES FROM ('2017-06-04') TO (MAXVALUE)),
> ```

ditto. way more newlines here

---

*[v2.0/partitioning.md, line 276 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KkY9j708W-cscpOmq:-L5KkY9j708W-cscpOmr:b76sn0d) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L276)):*
> ```Markdown
> ~~~ shell
> $ cat recent_us.zone.yml
> constraints: [+ssd]
> ```

does this work? i'd expect it to be `constraints: [+ssd,+datacenter=us1]`

---

*[v2.0/partitioning.md, line 341 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KkvdYuHSdaxqVi9oO:-L5KkvdZGDUCSoYvkWpR:b-s7q3ki) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L341)):*
> ```Markdown
> ## Locality–Resilience Tradeoff
> 
> There exists a tradeoff between making reads/writes fast and surviving failures. Consider a partition with three replicas of roachmart.users for Australian users. If only one replica is pinned to an Australian datacenter, then reads may be fast (via leases follow the sun) but writes will be slow. If two replicas are pinned to an Australian datacenter, than reads and writes will be fast (as long as the cross-ocean link has enough bandwidth that the third replica doesn’t fall behind and hit the quota pool). If those two replicas are in the same datacenter, then loss of one datacenter can lead to data unavailability, so some deployments may want two separate Austrialian datacenters. If all three replicas are in Australian datacenters, then three Australian datacenters are needed to be resilient to a datacenter loss.
> ```

remove "and hit the quota pool"

---

*[v2.0/partitioning.md, line 347 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Kl7GqiNNNfKK1Q3__:-L5Kl7GqiNNNfKK1Q3_a:bmm2u90) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L347)):*
> ```Markdown
> There is unfortunately no SQL standard for partitioning. As a result, separate partitioning syntaxes have emerged:
> 
> - MySQL and Oracle specify partitions inline in `CREATE TABLE`, similar to CockroachDB.
> ```

remove all references to the syntax of partitioning in other databases, it's irrelevant

---

*[v2.0/partitioning.md, line 355 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KlH9l00_t83X3KFqS:-L5KlH9l00_t83X3KFqT:bsobxbk) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L355)):*
> ```Markdown
> - Changes to secondary indexes: CockroachDB solves these changes through online schema changes. Online schema changes are a superior feature to partitioning because they require zero-downtime and eliminate the potential for consistency problems. 
> - Sharding: CockroachDB automatically shards data as a part of its distributed database architecture. 
> - Bulk Loading & Deleting: CockroachDB does not have a feature that supports this use case as of now. 
> ```

bulk loading is import csv. bulk deleting is on the roadmap. talk to andy about something you can link to for that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-:-L5KlRin6nwNJtoN17uZ:btjxrkn)*

Amruta-Ranade:
@danhhz @lhirata Thank you for the reviews!
@benesch If you'd prefer, you can review the doc once I incorporate Dan and Lauren's comments. That way, you can review the updated draft. If you want to review the doc as it stands now, that's okay as well :)

awoods187:
Review status: 0 of 2 files reviewed at latest revision, 55 unresolved discussions, all commit checks successful.

---

*[v2.0/partitioning.md, line 2 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KdvUgnVPvDLofZuJL:-L5LB501_YQNlCZaQqzk:b-u56l5l) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L2)):*

this is more @awoods187's domain, but I'm fairly sure we decided to only use "Geo Partitioning" and "Archival Partitioning" to attempt to avoid confusion, so we may want to make the page title reflect that

but if I'm wrong about that, I prefer "Partitioned Tables". "Partition Tables" sounds to me like something else
</blockquote></details>

Yeah i think we should title this geo-partitioning and archival partitioning.

---

*[v2.0/partitioning.md, line 54 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5LBUiQaLLRomLcs2gr:-L5LBUiQaLLRomLcs2gs:b-i3givx) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L54)):*
> ```Markdown
> #### Partitioning Key Considerations
> 
> - Once you set the composite primary key, you can’t change it later. So provision for all future subpartitions by including those columns in the composite primary key. In the example of the online learning portal, if you think you might want to subpartition based on `last_seen` in the future, define the composite primary key as `(country, last_seen, id)`.
> ```

Should we reference the (messy) workaround of reimporting your data?

---

*[v2.0/partitioning.md, line 355 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KlH9l00_t83X3KFqS:-L5LC594wOB9Om1Wpmkp:bqi071r) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L355)):*

bulk loading is import csv. bulk deleting is on the roadmap. talk to andy about something you can link to for that
</blockquote></details>

Bulk deleting is on the roadmap for 2.1 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

benesch:
Made a few comments!

---

Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 60 unresolved discussions, all commit checks successful.

---

*[v2.0/partitioning.md, line 61 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5KgcssqTDvotG5eVuU:-L5Q9jJIfkLrjogGvkYF:b-yx6jxv) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L61)):*

unique among all partitions on that table and its indexes
</blockquote></details>

This is surprising enough that it probably warrants another paragraph. Something like this, perhaps?

The uniqueness requirements can be surprising. For example, the following `CREATE INDEX` will fail because it reuses the name of a partition of the primary key:

```
CREATE TABLE foo (a STRING PRIMARY KEY, b STRING) PARTITION BY LIST (a) (
    bar VALUES IN ('bar'),
    default VALUES IN (DEFAULT)
)
CREATE INDEX foo_b_idx ON foo (b) PARTITION BY LIST (b) (
    baz VALUES IN ('baz'),
    default VALUES IN (DEFAULT)
)
```

Consider using a naming scheme that uses the index name to avoid conflicts. For example, the partitions above could be named `primary_idx_bar`, `primary_idx_default`, `b_idx_baz`, `b_idx_default`.

---

*[v2.0/partitioning.md, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QAdFhcz3QbQsW-Flh:-L5QAdFhcz3QbQsW-Fli:bhdm7pi) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L67)):*
I might explain a bit more here. Something like this?

> To declare the mapping between partitions and values. 

When declaring a table's partitions, you can declare what values belong to each partition by enumerating all possible values for each partition (`PARTITION BY LIST`)

---

*[v2.0/partitioning.md, line 235 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AibXo2KZxco96xsV2:-L5QDTamVGmJr6TG8jfp:b-dwt938) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L235)):*
Range partitions actually can't be partitioned. List partitions can be subpartitioned ad infinitum. 

> partitions can be used any number of times

I'm worried this isn't quite clear enough. What about something like this?

> A list partition can itself be partitioned, forming a subpartition. There is no limit on the number of levels of subpartitioning; that is, list partitions can be infinitely nested.
> Note that range partitions cannot be subpartitioned.

---

*[v2.0/partitioning.md, line 299 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QD8yy4TGa1aMsZIxy:-L5QD8yy4TGa1aMsZIxz:b-y4x1s6) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L299)):*
> ```Markdown
> 
> $ cockroach zone set blog.users.recent_au --insecure  -f recent_au.zone.yml
> $ cockroach zone set blog.users.archived_au --insecure  -f archived_au.zone.yml
> ```

Extra spaces after insecure.

---

*[v2.0/partitioning.md, line 304 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QD11DzBmJIsdjVjB8:-L5QD11DzBmJIsdjVjB9:bs9y4h9) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L304)):*
> ```Markdown
> ### Repartition a Table
> 
> Consider the partitioned table of users of RoachMart. Suppose the table has been partitioned on range to store the recent users on fast and expensive storage devices (example: ssd) and store the data of the users who have not been to the website recently on slower, cheaper storage devices(example: hdd). Now suppose we want to change the date after which the users will be considered recent to `2018-03-01`. We can achieve this by using the `ALTER TABLE` command. 
> ```

"SSD", "HDD"

---

*[v2.0/partitioning.md, line 318 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QCuG4BBXZHG03ceDD:-L5QCuG4BBXZHG03ceDE:b-is8unz) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L318)):*
> ```Markdown
> ~~~ sql
> > ALTER TABLE users_by_range PARTITION BY NOTHING;
> ~~~
> ```

Similarly, you can remove the partitions of an index using `ALTER INDEX ... PARTITION BY NOTHING`:

--example here---

---

*[v2.0/partitioning.md, line 326 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AkHkfFczfV7HwL9T_:-L5QCMxDk61WMLw9S7Z7:bqei7pi) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L326)):*
> unless you were previously able to do use these features with an expired license?

Nope. "Will not work" is better.

---

*[v2.0/partitioning.md, line 330 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QCZrApqbLJJYBXve9:-L5QCZrApqbLJJYBXveA:b3ev2uc) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L330)):*
> ```Markdown
> - Changing the partitioning scheme on any table or index
> - Changing the zone config for a partition
> ```

Also creating new partitioned tables or adding new zone configs for a partition won't work.

---

*[v2.0/partitioning.md, line 333 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5AkTyEiV-Umah2TMrd:-L5QCTKhW4Efsdw6hQX-:b-v8y8qj) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L333)):*

Add back ticks around `SELECT foo PARTITION`
</blockquote></details>

"Querying a partitioned table" (drop the on)

---

*[v2.0/partitioning.md, line 341 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Akn2EubciGtDOdFsq:-L5QC3CLGpjIllwzw_Ib:b1w21tr) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L341)):*

"There exists a tradeoff [...]" > "There is a tradeoff"

Put back ticks around `roachmart.users`

"(via leases follow the sun)" > "following the sun" (But don't we call this "following the workload?" Also, maybe we should link to relevant docs here) 
</blockquote></details>

Yeah, the terminology has been updated to "follow the workload" since the RFC was written.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-:-L5QFY-aINKCd-BJfy44:b-3uimdc)*

Amruta-Ranade:
*[v2.0/partitioning.md, line 189 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Kjw7NikGNd64Kn0jk:-L5oUHMyIaeFwXxvxPZZ:b81cey) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L189)):*

this example is a bit contrived, since updating a user's pk on every website access is going to be really expensive. I think you want something where the date you're using for partitioning is write-once
</blockquote></details>

Hmm..how about articles on a blog?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

Amruta-Ranade:
*[v2.0/partitioning.md, line 347 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5Kl7GqiNNNfKK1Q3__:-L5oWwFhmSw51LL6qAEk:b-v01pet) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L347)):*

remove all references to the syntax of partitioning in other databases, it's irrelevant
</blockquote></details>

Yeah, I wasn't sure about this section. Can I just delete the bullet list and directly talk about how other databases use partitioning for the three use cases?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

Amruta-Ranade:
Review status: 1 of 2 files reviewed at latest revision, 60 unresolved discussions, all commit checks successful.

---

*[v2.0/partitioning.md, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QAdFhcz3QbQsW-Flh:-L5oQbSJEwsBxJT7wUw7:bkqhp5w) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L67)):*
> When declaring a table's partitions, you can declare what values belong to each partition by enumerating all possible values for each partition (PARTITION BY LIST)

What about `PARTITION BY RANGE`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

benesch:
Review status: 1 of 2 files reviewed at latest revision, 60 unresolved discussions, all commit checks successful.

---

*[v2.0/partitioning.md, line 67 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2482#-L5QAdFhcz3QbQsW-Flh:-L5oa1XZcgOkkAen77FR:b91yqjo) ([raw file](https://github.com/cockroachdb/docs/blob/79330c132dab3685c195ec6921942f995dafe2cc/v2.0/partitioning.md#L67)):*

> When declaring a table's partitions, you can declare what values belong to each partition by enumerating all possible values for each partition (PARTITION BY LIST)

What about `PARTITION BY RANGE`?
</blockquote></details>

Oops, unfinished thought. Something like this?

You can declare which values belong to which partition in one of two ways. In list partitioning, (`PARTITION BY LIST`) you enumerate all possible values for each partition. In range partitioning (`PARTITION BY RANGE`), you specify a contiguous "range" of values for each partition by specifying lower and upper bounds. List partitioning is a good choice when the number of possible values is small; conversely, range partitioning is a good choice when the number of possible values is too large to explicitly list out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2482)*

Amruta-Ranade:

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+    PARTITION BY LIST (country)
+      (PARTITION north_america VALUES IN ('CA','US'), 
+      PARTITION australia VALUES IN ('AU','NZ'), 
+      PARTITION DEFAULT VALUES IN (default));
+~~~
+
+#### Step 4: Create corresponding zone configurations:
+
+~~~ shell
+$ cat > north_america.zone.yml
```
jseldess: Are these commands supposed to just show the contents of the files? If so, let's split the command from the response. 

Also, I don't think you need the `>`. You can just do `cat north_america.zone.yml`. 

Same applies below.

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+#### Step 5: Apply zone configurations to corresponding partitions:
+ 
+~~~ shell
+$ cockroach zone set roachlearn.students_by_list.north_america --insecure  -f north_america.zone.yml
+$ cockroach zone set roachlearn.students_by_list.australia --insecure  -f australia.zone.yml
+~~~
+
+### Define Table Partitions by Range
+
+Consider the table of students of the global online learning portal, RoachLearn. Suppose we want to store the data of current students on fast and expensive storage devices (example: SSD) and store the data of the graduated students on slower, cheaper storage devices(example: HDD). We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax.
```
jseldess: I don't think you need the first sentence.

nits: 
- `(example: SSD)` > `(e.g., SSD)`
- `(example: HDD)` > `(e.g., HDD)`

Also need a space before the HDD parentheses.

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Define Table Partitions by Range
+
+Consider the table of students of the global online learning portal, RoachLearn. Suppose we want to store the data of current students on fast and expensive storage devices (example: SSD) and store the data of the graduated students on slower, cheaper storage devices(example: HDD). We can achieve this by partitioning the table by date and using the `PARTITION BY RANGE` syntax. 
+
+#### Step 1: Set the enterprise license
+
+To set the enterprise license, see [Set the Trial or Enterprise License Key](enterprise-licensing.html#set-the-trial-or-enterprise-license-key)
+
+#### Step 2: Start each node with the appropriate storage device specified in the `--store` flag:
```
jseldess: In the example, you're not using the attribute field of the store flag. You're using the node-level `attrs` flag. If the examples are right, you need to update the header language. Or vice versa if the header language is right.

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+
+Going back to RoachLearn's scenario, suppose we want to do all of the following:
+
+- Keep the students' data close to their location.
+- Store the current students' data on faster storage devices.
+- Store the graduated students' data on slower, cheaper storage devices (example: HDD).
+
+We can achieve this by partitioning the table first by location and then by date.
+
+#### Step 1: Start each node with the appropriate storage device specified in the `--store` flag:
```
jseldess: Ah, now I see you using store-level attributes. Perhaps this is what you wanted to do in the example above as well?

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+
+Other databases use partitioning for three additional use cases: secondary indexes, sharding, and bulk loading/deleting. CockroachDB addresses these use-cases not by using partitioning, but in the following ways:
+
+- **Changes to secondary indexes:** CockroachDB solves these changes through online schema changes. Online schema changes are a superior feature to partitioning because they require zero-downtime and eliminate the potential for consistency problems. 
+- **Sharding:** CockroachDB automatically shards data as a part of its distributed database architecture. 
+- **Bulk Loading & Deleting:** CockroachDB does not have a feature that supports this use case as of now. 
+
+## See Also
+
+[Other Cockroach Commands](cockroach-commands.html)
```
jseldess: This feels strange, given that this page is not focused only on a cockroach command. Can you think of other relevant docs tied to this page?

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Unpartition a Table
+
+You can remove the partitions on a table by using the `PARTITION BY NOTHING` syntax with the `ALTER TABLE` command:
+ 
+~~~ sql
+> ALTER TABLE students PARTITION BY NOTHING;
+~~~
+
+## Licensing Requirements
```
jseldess: Perhaps you should move this under **How Table Partitioning Works**. I mentioned above that it feels like a gap there in any case.

v2.0/partitioning.md
```diff
@@ -0,0 +1,455 @@
...
+    name STRING,
+    email STRING,
+    country STRING,
+    expected_graduation_date DATE,   
+    PRIMARY KEY (country, id));
+~~~
+
+#### Partitioning Key Considerations
+
+- Once you set the primary key, you can’t change it later. Provision for all future subpartitions by including those columns in the primary key. In the example of the online learning portal, if you think you might want to subpartition based on `expected_graduation_date` in the future, define the primary key as `(country, expected_graduation_date, id)`.
```
nstewart: I think from a product perspective we need to note that this is on the roadmap. Asking someone to plan (potentially years in advance) the ways in which they may want to change the table is off brand. 

cc @awoods187 @dianasaur323

awoods187: +1 (but it's coming in 2.1)

nstewart: Yup -- let's just say that!

v2.0/partitioning.md
```diff
@@ -0,0 +1,371 @@
...
+);
+CREATE INDEX foo_b_idx ON foo (b) PARTITION BY LIST (b) (
+    baz VALUES IN ('baz'),
+    default VALUES IN (DEFAULT)
+);
+~~~
+
+Consider using a naming scheme that uses the index name to avoid conflicts. For example, the partitions above could be named `primary_idx_bar`, `primary_idx_default`, `b_idx_baz`, `b_idx_default`.
+
+### Create a Table with Partitions
```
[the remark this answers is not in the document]
Amruta-Ranade: Renamed the section.

Amruta-Ranade:
@jseldess Reorganized the content as we discussed yesterday. Let me know if you think it makes things clearer (I think it does).

_includes/sidebar-data-v2.0.json
```diff
@@ -1033,6 +1033,12 @@
           "/${VERSION}/configure-replication-zones.html"
         ]
       },
+      {
+        "title": "Define Table Partitions",
```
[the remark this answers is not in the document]
jseldess: This isn't actually done, @Amruta-Ranade. You still need to change `"title": "Define Table Partitions"` to `"title": "Define Table Partitions (Enterprise)".

v2.0/partitioning.md
```diff
@@ -0,0 +1,543 @@
...
+### Unpartition a Table
+
+You can remove the partitions on a table by using the `PARTITION BY NOTHING` syntax with the `ALTER TABLE` command:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> ALTER TABLE students PARTITION BY NOTHING;
+~~~
+
+## Locality–Resilience Tradeoff
```
jseldess: This section is hard to follow and feels a bit strange all the way at the end. But for now, perhaps we can just improve via bullets?

> - If only one replica is pinned to an Australian datacenter, then reads may be fast (via leases follow the workload) but writes will be slow. 
> - If two replicas are pinned to an Australian datacenter, then reads and writes will be fast (as long as the cross-ocean link has enough bandwidth that the third replica doesn’t fall behind). If those two replicas are in the same datacenter, then the loss of one datacenter can lead to data unavailability, so some deployments may want two separate Australian datacenters. 
> - If all three replicas are in Australian datacenters, then three Australian datacenters are needed to be resilient to a datacenter loss.

PR #1, opened 2016-01-13
Completed draft of start a local cluster

jseldess:
I have a first draft of "Start a Local Cluster". Please review http://cockroach-draft-docs.s3-website-us-east-1.amazonaws.com/start-a-local-cluster.html and send thoughts and feedback. 

@spencerkimball, I've assigned this to Marc, but please send along your suggestions, too.

PR #2485, opened 2018-02-12
Fix data unavailability tutorial

### LINKED ISSUE #3: install cockroachdb
- Install instructions for Mac, Linux, and Windows.
- Tab-type thing on page to show instructions only for relevant OS.

a-robinson:
It was broken for a few main reasons:

1. There was no guarantee that important system data wouldn't end up on
the two nodes that we're bringing down, since we were only using two
localities.
2. Even if we used 3 localities, the pattern of starting up a 3-node
cluster first and then adding more nodes means that the system ranges
may not be properly spread across the diversities, since in v1.1 we
don't proactively move data around to improve diversity.
3. A read-only query isn't guaranteed to hang even if a range is
unavailable. If we only kill the 2 non-leaseholders, the leaseholder
will still be able to keep extending its lease (via node liveness)
and serve reads.

To fix #1, I've modified this to spin up a 9 node cluster across 3
localities.
To fix #2, I've spun up all the nodes before running cockroach init.
We can go back to the old way of doing this once the labs use v2.0.
To fix #3, I've switched from demo-ing a SELECT to using an INSERT.


a-robinson:
Looks like I need to regenerate a couple screenshots. Will have those up in a few minutes.

a-robinson:
Screenshots updated.

petermattis:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[training/data-unavailability-troubleshooting.md, line 194 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5Bj1nfVQKbISQUVQmF:-L5Bj1nfVQKbISQUVQmG:b-4r0osq) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L194)):*
> ```Markdown
>     | Start Key | End Key | Replicas | Lease Holder |
>     +-----------+---------+----------+--------------+
>     | NULL      | NULL    | {3,6,9}  |            9 |
> ```

Shouldn't this be `7,8,9`.

---

*[training/data-unavailability-troubleshooting.md, line 214 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5BjEDznQFeWM7io8bp:-L5BjEDznQFeWM7io8bq:bb9qma6) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L214)):*
> ```Markdown
>     --insecure \
>     --port=26264
>     ~~~
> ```

In other modules, we've told the trainees to ctrl-c the appropriate nodes. Any reason to do this differently here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-:-L5BjZLMJwV3XLSolFGk:bnfp4nl)*

a-robinson:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[training/data-unavailability-troubleshooting.md, line 194 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5Bj1nfVQKbISQUVQmF:-L5BlSQ6SqVwkXuaK7hw:b-oq70qs) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L194)):*

Shouldn't this be `7,8,9`.
</blockquote></details>

That might be clearer, but it wouldn't reflect reality given that we're starting all the nodes before running `init`. And we have to do that due to point #2 in the commit message.

---

*[training/data-unavailability-troubleshooting.md, line 214 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5BjEDznQFeWM7io8bp:-L5BldrPuOzkkw02UcpC:bv8nd92) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L214)):*

In other modules, we've told the trainees to ctrl-c the appropriate nodes. Any reason to do this differently here?
</blockquote></details>

Because the instructions in step one start up all processes in the background using `&` to avoid needing 10 terminals for this lab (9 for the nodes, 1 for other commands). If you think using 10 terminals would be less confusing, I can change it back. Using `--background` isn't an option because it doesn't work until after `init` has been run.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2485)*

petermattis:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[training/data-unavailability-troubleshooting.md, line 194 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5Bj1nfVQKbISQUVQmF:-L5BoP0m8ta2xGm_v0k0:b-j1c4w7) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L194)):*

That might be clearer, but it wouldn't reflect reality given that we're starting all the nodes before running `init`. And we have to do that due to point #2 in the commit message.
</blockquote></details>

Ah, the nodes here won't be deterministic. You'll have to call that out in the presentation. I was wondering why you recommend visiting the `nodes` report. Now it makes sense.

---

*[training/data-unavailability-troubleshooting.md, line 214 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2485#-L5BjEDznQFeWM7io8bp:-L5BmU5_GrAl6Od8bqK1:bs6xziz) ([raw file](https://github.com/cockroachdb/docs/blob/16ad166a7ff525c5857742216d74fee508aa9be8/training/data-unavailability-troubleshooting.md#L214)):*

Because the instructions in step one start up all processes in the background using `&` to avoid needing 10 terminals for this lab (9 for the nodes, 1 for other commands). If you think using 10 terminals would be less confusing, I can change it back. Using `--background` isn't an option because it doesn't work until after `init` has been run.
</blockquote></details>

Ah, I missed the `&`. Carry on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2485)*

training/data-unavailability-troubleshooting.md
```diff
@@ -100,29 +191,45 @@ In preparation, add three more nodes with a distinct `--locality`, add a table,
...
     | Start Key | End Key | Replicas | Lease Holder |
     +-----------+---------+----------+--------------+
-    | NULL      | NULL    | {4,5,6}  |            6 |
+    | NULL      | NULL    | {3,6,9}  |            9 |
     +-----------+---------+----------+--------------+
     (1 row)
     ~~~
 
-## Step 2. Simulate the problem
+4. Note that the node IDs above may not match the order in which we started the nodes, because node IDs only get allocated after `cockroach init` is run. We can verify that the nodes listed by `SHOW TESTING_RANGES`are all in the `datacenter=us-east-3` locality by opening the Node Diagnostics debug page at <a href="http://localhost:8080/#/reports/nodes" data-proofer-ignore>http://localhost:8080/#/reports/nodes</a> and checking the locality for each of the 3 node IDs.
```
jseldess: nit: need a space after `SHOW TESTING RANGES`. I'll fix in follow-up PR.

PR #2487, opened 2018-02-12
Fix ycsb download/untar in Fault Tolerance and Replication tutorial

a-robinson:
I somehow forgot to remove the filenames in
a196f8ce0a6280f614d79ceddea58bcfc3cf8ef1.


petermattis:
Jesse, should that be `tar xz`?

a-robinson:
@jseldess that's because the file doesn't appear to be hosted on the review site. If you omit the `| tar xJ` the command just hangs forever.

Does the real hosted file work for you? i.e. `curl https://www.cockroachlabs.com/docs/training/resources/crdb-ycsb-mac.tar.gz | tar xJ`

a-robinson:
> Jesse, should that be `tar xz`?

I'd think so as well, but our ["Install CockroachDB" page](https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html) has recommended using `tar -xJ` for as long as I can remember, so I didn't want to mess with that at the last minute here.

I did just switch this from `tar xJ` to `tar -xJ` for consistency with that page, but I doubt that matters.

jseldess:
Oh, yes, the real hosted file works for me, @a-robinson. Thanks for the clarification.

jseldess:
Also, @a-robinson, feel free to change to `xz` if that's better.

a-robinson:
> Also, @a-robinson, feel free to change to xz if that's better.

Sticking with what's been working for binary downloads for a long time is better than changing it at the last minute.

PR #2490, opened 2018-02-13
Add How Cockroach Labs Debugs page

petermattis:


petermattis:
Cargo-culting here so I might have done something wrong.

petermattis:
Fixed the side-bar issue with the original commit. Seems to work now.

PR #2496, opened 2018-02-14
training: Fix node cleanup text, remove final What's Next section

a-robinson:
There's also the fact that the next module doesn't actually have the steps to create a cluster from scratch, but I don't want to tweak any of @petermattis's content without his confirmation at this point.

### DIFF
--- a/v1.1/training/cluster-upgrade.md
+++ b/v1.1/training/cluster-upgrade.md
@@ -286,7 +286,3 @@ This is the last module of the training, so feel free to stop you cluster and cl
     ~~~ shell
     $ rm -rf node1 node2 node3
     ~~~
-
-## What's Next?
-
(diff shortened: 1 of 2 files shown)


a-robinson:
Sounds good! I was also wondering why the Backup/Restore section ended up before the Cluster Upgrade section, which was originally planned to come first and seems to make more sense coming beforehand. It's no big deal either way, though.

jseldess:
@a-robinson, if we put cluster upgrade at the end, I think we'd have to use the 2.0 binary for backup/restore. Figured it's better to stick with 1.1 for that. If you disagree, happy to switch them.

a-robinson:
Not now that we name the v2.0 binary differently (`cockroach-v2.0`).

PR #2497, opened 2018-02-14
Update backup and restore training

jseldess:
@petermattis and @a-robinson, I have to get the license to work in the morning, but this should be good to go otherwise. To test it yourself, you can get an internal license from the #backup channel.  

Sorry for the late addition. If it's too late to include in the training, no problem. I can hide the content/slides.


jseldess:
I'm just going to go ahead and merge this, since it's an improvement on what was there in any case. I'll open a follow-up if there's any feedback, and once I have an updated enterprise settings.

PR #2536, opened 2018-02-16
training: Use fixed ycsb binary that rate limits better

a-robinson:
Pulls in https://github.com/cockroachdb/loadgen/pull/120

Just fixes a minor annoyance that made it look like Cockroach couldn't even handle 300qps locally, due to the ycsb binary's rate limiting not working well.


a-robinson:
Thanks for the reminder. It also reminded me to fix a few commands that I never added a Copy button to on the network partitioning lab.

PR #2539, opened 2018-02-19
Various updates to sql feature support page

jseldess:
- Mark Sequences as supported in 2.0
- Add INET for 2.0
- Change UUID for sql standard to postgres extension for 1.1 and 2.0


knz:
Thanks!

we need to check, it may be that array data types are also not in the
sql standard.

jseldess:
The Postgres docs for `ARRAY` include the following statement that suggests that `ARRAY` is in fact in the sql standard:

> An alternative syntax, which conforms to the SQL standard by using the keyword `ARRAY`, can be used for one-dimensional arrays...

jseldess:
@knz, just found array in the SQL:1999 specification, so I think we're ok to categorize it as standard.

PR #2547, opened 2018-02-21
Fix Google Tag Manager tags

sploiselle:
We had some inconsistent `dava-eventcategory` tags on the install page, which caused conflicts/tracking failures in Google Tag manager. This PR aligns them.

Bundled in here is also a formatting fix for the `docker version` commands, which weren't formatted in HTML and didn't prevent the user from selecting the terminal marker.


jseldess:
LGTM. Thanks, @sploiselle.

PR #2548, opened 2018-02-22
Withdraw the v2.0-alpha.20180122 release

a-robinson:
Fixes https://github.com/cockroachdb/cockroach/issues/22947

I chose not to comment out the release notes as was done in #2366 because I didn't want to have to roll them all forward into the next alpha's release notes, but let me know if you think we should do that instead.

Also, is this all that we do to withdraw a release?

### DIFF
--- a/_data/releases.yml
+++ b/_data/releases.yml
@@ -38,6 +38,7 @@
       no_windows: true
     - date: Jan 22, 2018
       version: v2.0-alpha.20180122
+      withdrawn: true
     - date: Jan 16, 2018
       version: v2.0-alpha.20180116
       withdrawn: true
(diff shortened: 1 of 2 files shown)


petermattis:
:lgtm: though I'm not sure if there is anything else we do. Ben and Jesse would probably know and yet both are out this week.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2548#-:-L5yvt_C-BXMM9LHNXy4:bob9og5)*

a-robinson:
Well judging by the 20180116 release, we don't delete the Github tag or release, we don't remove the unstable docker image, and we don't remove the binaries from binaries.cockroachdb.com, so I'm inclined to think this is all we do.

jseldess:
Thanks, @a-robinson. I think you took care of everything.

PR #2552, opened 2018-02-23
Clarify that CANCEL QUERY accepts a single ID

### LINKED ISSUE #2545: Clarify use of CANCEL QUERY
In [this subsection](https://www.cockroachlabs.com/docs/v1.1/cancel-query.html#cancel-a-query-via-a-nested-select-
statement) we explain a useful trick for cancelling a specific query. A user reading this might be concerned that they could accidentally cancel another query alongside the one they intended to. It might be useful to specify that `CANCEL QUERY` only takes a single value, so if the subquery returns multiple values the `CANCEL QUERY` statement will fail.

jseldess:
Fixes #2545

### DIFF
--- a/v1.1/cancel-query.md
+++ b/v1.1/cancel-query.md
@@ -25,7 +25,7 @@ The `root` user can cancel any currently active queries, whereas non-`root` user
 
 Parameter | Description
 ----------|------------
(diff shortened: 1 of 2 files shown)


justinj:
LGTM!

PR #2583, opened 2018-03-01
Recommend against running Kubernetes instructions from a cloud shell

a-robinson:
This can be confusing for users because the port-forward method of
viewing the admin UI only works if you're port-forwarding from your
local machine.


a-robinson:
The note in the Monitor Cluster section might be overkill. I'm very interested in your opinions on that.

kannanlakshmi:
LGTM - thanks Alex!

jseldess:
:lgtm:, too!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2583#-:-L6aK7AMmApiGqB4fRwV:b-3h41f2)*

Issue #2146, opened 2017-11-10
Copy-to-clipboard styled incorrectly when in <div> tags


jseldess:
Example:

```
2. Stop Kubernetes:

    <div class="filter-content" markdown="1" data-scope="gce-hosted">

    {% include copy-clipboard.html %}
    ~~~ shell
    $ gcloud container clusters delete cockroachdb
    ~~~

    </div>
```

<img width="854" alt="screen shot 2017-11-09 at 11 30 01 pm" src="https://user-images.githubusercontent.com/15893490/32642977-f2a4d070-c5a5-11e7-9a44-0839ad1ae7fd.png">

It should look like:

<img width="830" alt="screen shot 2017-11-09 at 11 30 39 pm" src="https://user-images.githubusercontent.com/15893490/32642984-07cf8f44-c5a6-11e7-894f-f810fa465934.png">

jseldess:
@j-low, when you have some time, would love you help with this.

jseldess:
@j-low, ping on this, whenever you have time.

j-low:
@jseldess looking into this issue this week

jseldess:
@j-low, this problem has reappeared in more severe form in https://github.com/cockroachdb/docs/pull/2585. In that case, using an include snippet inside a div causes the html to not render properly. 

While looking at that, I found a workaround: Whenever we use an include snippet within a div, changing the `<div></div>` tags to `<section></section>` resolves the problem. However, it seems important to **not** make this change when there's no include within the div. I don't understand what's going on, but I suspect the html generated by the include snippet is somehow invalid and therefore causing chaos when nested in another div.

I think this workaround is sufficient for now, but it'd still be great if you could figure out why this combination creates chaos.

luciaaaaaaa:
does the problem persist if we remove the icon? and have the button just be 'COPY'

luciaaaaaaa:
Link to designs in Figma: https://www.figma.com/file/6XcxaUdad2RkjpEqnVHKAG/Education-docs-redesign?node-id=2228%3A12946

PR #2585, opened 2018-03-01
Add missing COPY buttons to Kubernetes docs

a-robinson:
Also add a couple of missing `kubectl get pods` commands and clean up
the formatting of the commands to be entered into the SQL shell.

### DIFF
--- a/_includes/orchestration/start-kubernetes.md
+++ b/_includes/orchestration/start-kubernetes.md
@@ -10,7 +10,7 @@ Choose whether you want to orchestrate CockroachDB with Kubernetes using the hos
 
 ## Step 2. Start Kubernetes
 
-<div class="filter-content" markdown="1" data-scope="gke-hosted">
+<section class="filter-content" markdown="1" data-scope="gke-hosted">
 
 1. Complete the **Before You Begin** steps described in the [Google Kubernetes Engine Quickstart](https://cloud.google.com/kubernetes-engine/docs/quickstart) documentation. The documentation offers the choice of using Google's Cloud Shell product or using a local shell on your machine. Choose to use a local shell if you want to be able to view the CockroachDB Admin UI using the steps in this guide.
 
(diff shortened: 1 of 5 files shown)


a-robinson:
Any idea why the `gcloud container clusters create` command isn't rendering properly?

screen shot 2018-03-01 at 5 56 32 pm

jseldess:
Ugh, unfortunately, it looks like, when we're wrapping markdown in `<div class="filter-content" markdown="1" data-scope="BLAH">` tags for filtering purposes, using the `{% include copy-clipboard.html %}` within the `<div>` breaks the page. I don't see an immediate solution right now because even if you remove `{% include copy-clipboard.html %}` from the first code block in the div, the second one with the include snippet breaks. 

For now, I'd suggest reverting all of the includes you added within divs and opening an issue for @j-low to help us fix the include snippet. We already have an issue for incorrect styling of the copy button in cases like this, so perhaps we should just expand that issue: https://github.com/cockroachdb/docs/issues/2146.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2585#-:-L6aHdUHsRs5k0absyjd:b-oq181x)*

jseldess:
We could also get around this by splitting the hosted gke, manual gce, and manual aws into separate pages. But for now, I think accepting a slightly annoying UX of not having copy to clipboard on code snippets inside those divs is ok.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2585#-:-L6aKLZ_8VKijVjORUDG:b-xgbsqq)*

jseldess:
OK, I found a workaround. Whenever you're using an `{% include copy-clipboard.html %}` within a `<div> </div>`, change those div tags to `<section> </section>`. I tested this on your branch, and it works fine. However, it seems important to **not** make this change when there's no include within the div. I don't understand what's going on, but I suspect the html div generated by the include snippet is somehow invalid and therefore causing choas when nested in another div. 

Would you prefer I make this change and push to your branch?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2585#-:-L6b14yA26jQQd63MIG2:bxj8rpt)*

a-robinson:
That is certainly unfortunate. Feel free to take over, or I can just remove the buggy COPY buttons if you'd prefer.

a-robinson:
Your changes LGTM. Thanks, @jseldess.

PR #2588, opened 2018-03-02
Release Notes Draft

Amruta-Ranade:


Amruta-Ranade:
@jseldess Ready for your review + need Doc updates for the release.

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+- CockroachDB now builds with go 1.9.4 and higher. [#22608][#22608]
+
+### General Changes
+
+- Diagnostics reports now include information about changed settings and statistics on types of errors encountered during SQL execution. [#22705][#22705], [#22693][#22693], [#22948][#22948]
+
+### Enterprise Edition Changes
+
+- Bulk IO rate limiting is now done per-store rather than per-node. [#22729][#22729]
+- Revision history backup and restore no longer considered experimental. [#22679][#22679]
```
jseldess: `is no longer`

Also, does `revision history` = `point-in-time`? If so, is `revision history` the term we're going with?

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+- Revision history backup/restore now handles schema changes. [#21717][#21717]
+- CockroachDB now checks that backup actually contains requested restore time. [#22659][#22659]
```
jseldess: some missing articles: `that a backup actually contains the requested restore time`.

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+- Improved `BACKUP`'s handling of tables after `TRUNCATE` [#21895][#21895]
```
jseldess: Missing period.

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+- Removed `into_db` option in `IMPORT`. The database is now specified as part of the table name. [#21813][#21813]
```
jseldess: `Removed the...`

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+
+- The type determined for constant `NULL` expressions is renamed to `unknown` for better compatibility with PostgreSQL. [#23150][#23150]
+- Deleting multiple rows at once now consumes less memory. [#23013][#23013]
+- Attempts to modify virtual schemas with DDL statements now fail with a clearer error message. [#23041][#23041]
+- The new statement `SHOW SCHEMAS` reveals which are the valid virtual schemas next to the physical schema `public`. [#23041][#23041]
+- CockroachDB now recognizes the special syntax `SET SCHEMA <name>` as an alias for `SET search_path = <name>` for better compatibility with PostgreSQL. [#23041][#23041]
+- `current_role` is recognized as an alias for `current_user` for better compatibility with PostgreSQL. [#23041][#23041]
+- The special identifier `current_catalog` is supported as alias for `current_database()` for better compatibility with PostgreSQL. [#23041][#23041]
+- CockroachDB now returns the correct error code for division by zero. [#22948][#22948]
+- The GC of table data after a `DROP TABLE` now respects changes to the GC TTL interval specified in the zone config [#22903][#22903]
```
jseldess: Either "after `DROP TABLE`" or "after a `DROP TABLE` statement".

releases/v2.0-beta.20180305.md
```diff
@@ -0,0 +1,406 @@
...
+- The full names of tables/view/sequences are now properly logged in the system event log. [#22848][#22848]
```
jseldess: Are we announcing the "system even log" here, or just announcing a fix of some bug? If the latter, let's move under bug fixes. If the former, let's say "are logged in the new system event log".

jseldess: We mention "auditing" and "SQL execution log" in a few bullets below. I suspect we're talking about the same thing using different terms. If so, we should consolidate or, at the least, use the same terminology. cc @knz for his insights.

knz: No the audit log and the execution log are two distinct, unrelated features.

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[releases/v2.0-beta.20180305.md, line 53 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2588#-L6qtYlPCazDpzonwun6:-L6qtYlPCazDpzonwun7:ba9yk4s) ([raw file](https://github.com/cockroachdb/docs/blob/1f021909a0e037d62b0a9a73b3c3cd014ca5db69/releases/v2.0-beta.20180305.md#L53)):*
> ```Markdown
> - [Sequences](../v2.0/create-sequence.html) are now supported in enterprise [`BACKUP`](../v2.0/backup.html)/[`RESTORE`](../v2.0/restore.html) jobs.
> 
>     This changes how sequences are stored in the key-value storage layer, so existing sequences must be dropped and recreated. Since a sequence cannot be dropped while it is being used in a column's `DEFAULT` expression, those expressions must be dropped before the sequence is dropped, and recreated after the sequence is recreated. The `setval` function can be used to set the value of a sequence to what it was previously. [#21684][#21684]
> ```

This is a backwards-incompatible change that needs to be called out separately from the enterprise section. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2588)*

knz:
> Thanks, @knz. Are the "system event log" and "SQL execution logs" different?

Yes, these are different too.

Here's the overview:

- system event log: logs high-level events like schema changes (tables created/deleted/renamed); job completions, node joins / decommissioned, etc.

- SQL execution log: logs all statements executed if cluster setting `sql.trace.log_statement_execute` is set

- audit log: logs all SQL statements that access specific tables when auditing has been enabled for those specific tables

PR #2595, opened 2018-03-03
Add script that checks for unused images

jseldess:
And remove unused images.


jseldess:
@knz, PTAL when you have time.

knz:
Your commit has two identical python scripts. Did you mean to just keep one?

jseldess:
Yes, I’ll fix that, @knz. But I’m still also trying  I figure out how to get the script to recognize all unused images on one run.

scripts/find-unused-images.py
```diff
@@ -0,0 +1,38 @@
...
+
+docs = list()
+for d in dirs:
+    for path, subdirs, files in os.walk(d):
+        for name in files:
+            docs.append(os.path.join(path, name))
+
+unused_images = images
+unread_files = list()
+for doc in docs:
```
mberhault: are you missing a commit? The change we added to skip the `not_checked_for_images.txt` file isn't there. Ditto the `onError` argument for `os.walk`

scripts/find-unused-images.py
```diff
@@ -0,0 +1,38 @@
...
+    for f in files:
+        images.append(f)
+print("Total # of image files:", len(images))
+
+dirs = list()
+for path, subdirs, files in os.walk("../"):
+    dirs.append(path)
+
+docs = list()
+for d in dirs:
```
mberhault: You can just call `for ... in os.walk("../"):` here, os.walk is recursive.

scripts/find-unused-images.py
jseldess: Sorry. Added `onerror` back in. I'm no longer writing to `not_checked_for_images.txt`, so skipping isn't necessary.

scripts/find-unused-images.py
jseldess: So when I do that, the script incorrectly identifies a specific image file as unused. Here's the code:

```
import os

"""
This script identifies images that are not used in any doc pages
so they can be removed. Run it from within the scripts directory.
"""

def _error(exc):
    raise exc

images = list()
for root, dirs, files in os.walk("../images", onerror=_error):
    for f in files:
        images.append(f)
print("Total # of image files:", len(images))

# dirs = list()
# for path, subdirs, files in os.walk("../", onerror=_error):
#     dirs.append(path)

docs = list()
# for d in dirs:
for path, dirs, files in os.walk("../", onerror=_error):
    for name in files:
        docs.append(os.path.join(path, name))

unused_images = images
# unread_files = list()
for doc in docs:
    try:
        with open(doc) as file:
            data = file.read()
            for i in images:
                if i in data and i in unused_images:
                    unused_images.remove(i)
    except Exception as e:
        """
        Uncomment the next line to see the files that cannot be read.
        Mostly, these will be binary files.
        """
        # print("exception:", e, "\n", "doc:", doc)
        continue

print("# of unused image files:", len(unused_images))
print("Unused image files:", unused_images)

# open("not_checked_for_images.txt", "a").write(str(unread_files))
```

And here's the output:

```
~/src/github.com/cockroachdb/docs/scripts$ python find-unused-images.py
Total # of image files: 110
# of unused image files: 1
Unused image files: ['training-5.png']
```

Contrast that to the output of the current script:

```
~/src/github.com/cockroachdb/docs/scripts$ python find-unused-images.py
Total # of image files: 110
# of unused image files: 0
Unused image files: []
```

I have no idea what's happening. Any ideas?

PR #2600, opened 2018-03-07
Doc updates for planned maintenance events

### LINKED ISSUE #2038: Improved behavior during maintenance events
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwINAgXA5on9pr2T/recWjo7BZnP2A0Zt4

PRs:

jseldess:
- Add FAQ on handling planned node maintenance
- Add guidance on using server.time_until_store_dead for rolling upgrades
- Minor improvements to cockroach quit doc
- Retitle Remove Nodes to Decommission Nodes

Fixes #2038

### DIFF
--- a/_includes/faq/planned-maintenance.md
+++ b/_includes/faq/planned-maintenance.md
@@ -1,0 +1,13 @@
+By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. Before temporarily stopping nodes for planned maintenance (e.g., upgrading system software), if you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window.
+
+For example, let's say you want to maintain a group of servers, and the nodes running on the servers may be offline for up to 15 minutes as a result. Before shutting down the nodes, you would change the `server.time_until_store_dead` cluster setting as follows:
+
+~~~ sql
+> SET CLUSTER SETTING server.time_until_store_dead = '15m0s';
+~~~
+
+After completing the maintenance work and [restarting the nodes](start-a-node.html), you would then change the setting back to its default:
+
+~~~ sql
+> SET CLUSTER SETTING server.time_until_store_dead = '5m0s';
+~~~
--- a/v1.1/remove-nodes.md
+++ b/v1.1/remove-nodes.md
@@ -1,5 +1,5 @@
 ---
-title: Remove Nodes
+title: Decommission Nodes
 summary: Permanently remove one or more nodes from a cluster.
 toc: false
 toc_not_nested: true
--- a/v2.0/operational-faqs.md
+++ b/v2.0/operational-faqs.md
@@ -107,6 +107,10 @@ Collecting information about CockroachDB's real world usage helps us prioritize 
 
 {% include faq/clock-synchronization_v2.0.html %}
 
+## How do I prepare for planned node maintenance?
+
+{% include faq/planned-maintenance.md %}
+
 ## See Also
 
 - [Product FAQs](frequently-asked-questions.html)
--- a/v2.0/remove-nodes.md
+++ b/v2.0/remove-nodes.md
@@ -1,5 +1,5 @@
 ---
-title: Remove Nodes
+title: Decommission Nodes
 summary: Permanently remove one or more nodes from a cluster.
 toc: false
 toc_not_nested: true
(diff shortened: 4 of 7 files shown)


jseldess:
TFTR, @dianasaur323!

bdarnell:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[v2.0/upgrade-cockroach-version.md, line 34 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2600#-L716cAij3uVBZN99NIv:-L716cAij3uVBZN99NIw:b-ie6dmu) ([raw file](https://github.com/cockroachdb/docs/blob/870d0e95a9d8a7b37fc5740baf750d6f009c6d9d/v2.0/upgrade-cockroach-version.md#L34)):*
> ```Markdown
> 5. By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. If you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window.
> 
>     For example, let's say you expect nodes to be offline for up to 10 minutes during the rolling upgrade. Before starting, you would change the `server.time_until_store_dead` cluster setting as follows:
> ```

Why would you expect this much downtime during a rolling upgrade? You should easily be able to complete your cockroach upgrade within the default 5m window (as long as you're automating things instead of typing them by hand during the upgrade). 

In my experience the only thing that might take long enough to warrant an increase in server.time_until_store_dead is a kernel upgrade. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2600)*

jseldess:
*[v2.0/upgrade-cockroach-version.md, line 34 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2600#-L716cAij3uVBZN99NIv:-L719vas2ejP8k6B7NVA:bjuaadt) ([raw file](https://github.com/cockroachdb/docs/blob/870d0e95a9d8a7b37fc5740baf750d6f009c6d9d/v2.0/upgrade-cockroach-version.md#L34)):*

Why would you expect this much downtime during a rolling upgrade? You should easily be able to complete your cockroach upgrade within the default 5m window (as long as you're automating things instead of typing them by hand during the upgrade). 

In my experience the only thing that might take long enough to warrant an increase in server.time_until_store_dead is a kernel upgrade. 
</blockquote></details>

Downgraded this to a note that states 5 min should be more than enough. Calls out the setting just in case. Is that ok, @bdarnell?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2600)*

bdarnell:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v2.0/upgrade-cockroach-version.md, line 34 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2600#-L716cAij3uVBZN99NIv:-L71MGUf1N3NYK8wgegu:b-ksl8mw) ([raw file](https://github.com/cockroachdb/docs/blob/870d0e95a9d8a7b37fc5740baf750d6f009c6d9d/v2.0/upgrade-cockroach-version.md#L34)):*

Downgraded this to a note that states 5 min should be more than enough. Calls out the setting just in case. Is that ok, @bdarnell?
</blockquote></details>

I'd probably leave it out completely, but I guess since we're writing these instructions as if you're running commands by hand instead of automating it it may be worth leaving in a mention. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2600)*

bdarnell:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v2.0/upgrade-cockroach-version.md, line 34 at r5](https://reviewable.io:443/reviews/cockroachdb/docs/2600#-L716cAij3uVBZN99NIv:-L76A2-L3UFb3IOmyQLt:b-ft4fwt) ([raw file](https://github.com/cockroachdb/docs/blob/870d0e95a9d8a7b37fc5740baf750d6f009c6d9d/v2.0/upgrade-cockroach-version.md#L34)):*

OK, removed. 

At the beginning of the "Perform the rolling upgrade" section, we have a tip:

> We recommend creating scripts to perform these steps instead of performing them by hand.

But I'd like to get more information from you about how this page could be more focused on automation. Seems to me that manual steps are necessary given that users will use differing tools. But I'd love to create and link to an example script or config for a specific system as well. 
</blockquote></details>

My view is that we have to walk through the manual process here so that people will understand what they have to write in their scripts. Over time we can develop a library of scripts/configs (like we have for kubernetes, etc), and those automated versions will displace these manual steps. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2600)*

PR #2618, opened 2018-03-07
Update clock offset shutdown language

rmloveland:
Specifically:

- Use less charged language to describe node shutdown during max clock
  offset situations

- Add a link to 'Remove Nodes' to provide more detail about what happens
  during the node shutdown process

### DIFF
--- a/v1.1/architecture/transaction-layer.md
+++ b/v1.1/architecture/transaction-layer.md
@@ -74,7 +74,7 @@ This then lets the node primarily responsible for the range (i.e., the Leasehold
 
 #### Max Clock Offset Enforcement
 
-To ensure correctness among distributed nodes, you can identify a Maximum Clock Offset. Because CockroachDB relies on clock synchronization, nodes run a version of [Marzullo’s algorithm](http://infolab.stanford.edu/pub/cstr/reports/csl/tr/83/247/CSL-TR-83-247.pdf) amongst themselves to measure maximum clock offset within the cluster. If the configured maximum offset is exceeded by any node, it commits suicide, preventing it from potentially creating consistency issues within the cluster.
+To ensure correctness among distributed nodes, you can identify a Maximum Clock Offset. Because CockroachDB relies on clock synchronization, nodes run a version of [Marzullo’s algorithm](http://infolab.stanford.edu/pub/cstr/reports/csl/tr/83/247/CSL-TR-83-247.pdf) amongst themselves to measure maximum clock offset within the cluster. If the configured maximum offset is exceeded by any node, it shuts down as soon as possible, preventing it from potentially creating consistency issues within the cluster.
 
 For more detail about the risks that large clock offsets can cause, see [What happens when node clocks are not properly synchronized?](../operational-faqs.html#what-happens-when-node-clocks-are-not-properly-synchronized)
 
--- a/v1.1/architecture/transaction-layer.md
+++ b/v1.1/architecture/transaction-layer.md
@@ -74,7 +74,9 @@ This then lets the node primarily responsible for the range (i.e., the Leasehold
 
 #### Max Clock Offset Enforcement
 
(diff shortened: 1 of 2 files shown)


rmloveland:
@jseldess I'm less confident about the link to 'Remove Nodes' but it seemed relevant.  If it's not (or somewhat misleading in this case) I'm happy to remove.

bdarnell:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v1.1/architecture/transaction-layer.md, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2618#-L71dFsJFEj9eXObWXmF-r1-77:-L767wQVTB2IdNUpchMd:bwvx4kw) ([raw file](https://github.com/cockroachdb/docs/blob/7c4740ed6bc504d5cf23296b2442c490eea198cf/v1.1/architecture/transaction-layer.md#L77)):*

I like this change, but I think it might make sense to replace this entire paragraph with the first paragraph in this include: https://github.com/cockroachdb/docs/blob/master/_includes/faq/clock-synchronization_v2.0.html. This include language is newer and, I think, a bit more accurate. 

It's a shame not to use the entire include here as well, but I don't think anything but the first paragraph is necessary.

As for node shutdown, I don't know if it's a "graceful" shutdown when due to clock skew. @bdarnell, can you clarify? 
</blockquote></details>

No, it is deliberately not a graceful shutdown (I'm actually in the process of making it even less graceful). We want to shut everything down as abruptly as possible because anything in progress could be serving inconsistent results. "Crashes" or "halts" would be better than "shuts down" here, and we don't go though any of the process described in the "Remove nodes" page. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2618)*

rmloveland:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v1.1/architecture/transaction-layer.md, line 77 at r1](https://reviewable.io:443/reviews/cockroachdb/docs/2618#-L71dFsJFEj9eXObWXmF-r1-77:-L7a3xr25bR_pOObsNJo:b-x64i7v) ([raw file](https://github.com/cockroachdb/docs/blob/7c4740ed6bc504d5cf23296b2442c490eea198cf/v1.1/architecture/transaction-layer.md#L77)):*

No, it is deliberately not a graceful shutdown (I'm actually in the process of making it even less graceful). We want to shut everything down as abruptly as possible because anything in progress could be serving inconsistent results. "Crashes" or "halts" would be better than "shuts down" here, and we don't go though any of the process described in the "Remove nodes" page. 
</blockquote></details>

@jseldess, I replaced the whole paragraph with the one you suggested in 499a2b94, except that I updated the language as @bdarnell suggested to say it **crashes immediately** (in bold so it's hard to miss :-) )

I also removed the links to 'Remove Nodes' since those are about a different type of shutdown (I don't have the commit SHA handy because I did something weird when fixing conflicts).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2618)*

bdarnell:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2618#-:-L7a50ozCWUOHd2ZXX91:bnfp4nl)*

jseldess:
:lgtm:, Rich.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2618#-:-L7aDojx-tl6Etd1Dt2T:bxsxuq5)*

PR #2623, opened 2018-03-08
Fix explanation of unavailable ranges in admin ui docs

### LINKED ISSUE #2602: Definition of "unavailable ranges" is wrong
https://www.cockroachlabs.com/docs/dev/admin-ui-replication-dashboard.html#review-of-cockroachdb-terminology

A range is unavailable when a majority of its replicas are unavailable. That means multiple, simultaneous node outages.

a-robinson:
Fixes #2602


jseldess:
LGTM

PR #2631, opened 2018-03-08
start a node: add clarification on specifying flags every time

### LINKED ISSUE #1813: Clarify --join flag
From @bdarnell:  

`--join` is only *required* when there is no data directory

as long as the data directory is preserved, the previous join targets can be recovered from there

however, it is highly *recommended* to always set `--join`, to avoid mistakes if the data directory has been wiped or if node addresses have changed

cc: @robert-s-lee

### LINKED ISSUE #1855: Start a Node Doc Update
Re: [Start a Node](https://cockroachlabs.com/docs//v1.0/start-a-node.html)

## Issue Description

I had an issue recently where I restarted a node without advertise host, and it was able to join the cluster, but none of the other nodes could talk to it an queries could not run.

## Suggested Resolution 

For restarting a node in a cluster in the cloud, we should make it clear that users need to advertise host.

arjunravinarayan:
Fixes #1855, fixes #1813.

### DIFF
--- a/v1.0/start-a-node.md
+++ b/v1.0/start-a-node.md
@@ -25,7 +25,13 @@ $ cockroach start --help
 
 ## Flags
 
-The `start` command supports the following [general-use](#general) and [logging](#logging) flags.
+The `start` command supports the following [general-use](#general) and
+[logging](#logging) flags. All flags must be specified each time the
+node is started, as they will not be remembered, with the exception of
+the `--join` flag. Nevertheless, we recommend specifying
+_all_ flags every time, including the `--join` flag, as that will
+allow restarted nodes to join the cluster even if their data directory
+was destroyed.
 
 {{site.data.alerts.callout_success}}When adding a node to an existing cluster, include the <code>--join</code> flag.{{site.data.alerts.end}}
 
--- a/v1.1/start-a-node.md
+++ b/v1.1/start-a-node.md
@@ -29,7 +29,13 @@ $ cockroach start --help
 
 ## Flags <span class="version-tag">Changed in v1.1</span>
 
-The `start` command supports the following [general-use](#general) and [logging](#logging) flags.
+The `start` command supports the following [general-use](#general) and
+[logging](#logging) flags. All flags must be specified each time the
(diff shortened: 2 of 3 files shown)


v1.0/start-a-node.md
```diff
@@ -25,7 +25,13 @@ $ cockroach start --help
 
 ## Flags
 
-The `start` command supports the following [general-use](#general) and [logging](#logging) flags.
+The `start` command supports the following [general-use](#general) and
+[logging](#logging) flags. All flags must be specified each time the
+node is started, as they will not be remembered, with the exception of
+the <code>--join</code> flag. Nevertheless, we recommend specifying
```
jseldess: nit: can you use backticks instead of `<code></code>`?

v1.1/start-a-node.md
```diff
@@ -29,7 +29,13 @@ $ cockroach start --help
 
 ## Flags <span class="version-tag">Changed in v1.1</span>
 
-The `start` command supports the following [general-use](#general) and [logging](#logging) flags.
+The `start` command supports the following [general-use](#general) and
+[logging](#logging) flags. All flags must be specified each time the
+node is started, as they will not be remembered, with the exception of
+the <code>--join</code> flag. Nevertheless, we recommend specifying
```
jseldess: Same as above.

v2.0/start-a-node.md
```diff
@@ -29,7 +29,13 @@ $ cockroach start --help
 
 ## Flags <span class="version-tag">Changed in v1.1</span>
 
-The `start` command supports the following [general-use](#general) and [logging](#logging) flags.
+The `start` command supports the following [general-use](#general) and
+[logging](#logging) flags. All flags must be specified each time the
+node is started, as they will not be remembered, with the exception of
+the <code>--join</code> flag. Nevertheless, we recommend specifying
```
jseldess: Same as above.

arjunravinarayan:
Done.

jseldess:
@arjunravinarayan, you need to update your copy of master, rebase this branch against it, and push again to get past the TC failure. Then we can merge!

PR #2635, opened 2018-03-08
delete: note that deletes dont free disk space immediately

### LINKED ISSUE #1622: FAQ about deletion tombstones
Several users have recently been surprised by scan on seemingly small tables taking a long time. This happened because their tables had a large number of rows that have been deleted. For such rows, crdb leaves behind tombstones; these tombstones are skipped during table scans, but still add cost. The point of not simply wiping the data is the support for historical queries.
This scenario is common when users use a table as a work queue - tasks are enqueued and dequeued by inserting and deleting in a table. At any one time there are few "live" rows, but many tombstones.

This behavior may also be surprising for an operator that observes that their running low on disk space, deletes some data and then is surprised to see that nothing has changed (in fact, disk space might have increased due to the deletion :) ).
I think we should explain this, perhaps in an FAQ.

One solution is to lower the TTL on the table's zone
```
./cockroach zone set .default -f - --insecure <<EOF
gc:
  ttlseconds: 600
EOF
```

On the product side, we should consider whether our current TTL default of 24h makes sense; but that'll be a separate issue.

arjunravinarayan:
Fixes #1622.

### DIFF
--- a/v1.0/delete.md
+++ b/v1.0/delete.md
@@ -43,6 +43,28 @@ Successful `DELETE` statements return one of the following:
 |`DELETE` _`int`_ | _int_ rows were deleted.<br><br>`DELETE` statements that don't delete any rows respond with `DELETE 0`. When `RETURNING NOTHING` is used, this information is not included in the response. |
 |Retrieved table | Including the `RETURNING` clause retrieves the deleted rows, using the columns identified by the clause's parameters.<br><br>[See an example.](#return-deleted-rows)|
 
+## Disk Space Usage After Deletes
+
+Deleting a row does not immediately free up the disk space. This is
+due to the fact that CockroachDB retains [the ability to query tables
+historically](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).
+
+If disk usage is a concern, there are two potential solutions. The
+first is to [reduce the time-to-live](configure-replication-zones.html)
+(TTL) for the zone, which will cause garbage collection to clean up
+deleted rows more frequently. Second, unlike `DELETE`,
+[truncate](truncate.html) immediately deletes the entire table, so
+consider if you can use `TRUNCATE` instead.
+
+## Select Performance on Deleted Rows
+
+Queries that scan across tables that have lots of deleted rows will
+have to scan over deletions that have not yet been garbage
+collected. Certain database usage patterns that frequently scan over
+and delete lots of rows will want to reduce the
+[time-to-live](configure-replication-zones.html) values to clean up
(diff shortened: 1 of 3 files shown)


v1.0/delete.md
```diff
@@ -43,6 +43,27 @@ Successful `DELETE` statements return one of the following:
...
 |Retrieved table | Including the `RETURNING` clause retrieves the deleted rows, using the columns identified by the clause's parameters.<br><br>[See an example.](#return-deleted-rows)|
 
+## Disk Space Usage After Deletes
+
+Deleting a row does not immediately free up the disk space. This is
+due to the fact that CockroachDB retains [the ability to query tables
+historically](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).
+
+If disk usage is a concern, there are two potential solutions. The
+first is to [reduce the Time-to-live](configure-replication-zones.html)
```
jseldess: nit: I'd use the lowercase `time-to-live` here, since you use it below.

v1.0/delete.md
```diff
@@ -43,6 +43,27 @@ Successful `DELETE` statements return one of the following:
...
+[truncate](truncate.html) immediately deletes the entire table, so
+consider if you can use `TRUNCATE` instead.
+
+## Select Performance on Deleted Rows
+
+Queries that scan across tables that have lots of deleted rows will
+have to scan over deletions that have not yet been garbage
+collected. Certain database usage patterns that frequently scan over
+and delete lots of rows will want to reduce the
+[time-to-live](configure-replication-zones.html) values.
```
jseldess: nit: `values to clean up deleted rows more frequently.` Like you say above.

v1.1/delete.md
```diff
@@ -43,6 +43,27 @@ Successful `DELETE` statements return one of the following:
 |`DELETE` _`int`_ | _int_ rows were deleted.<br><br>`DELETE` statements that don't delete any rows respond with `DELETE 0`. When `RETURNING NOTHING` is used, this information is not included in the response. |
 |Retrieved table | Including the `RETURNING` clause retrieves the deleted rows, using the columns identified by the clause's parameters.<br><br>[See an example.](#return-deleted-rows)|
 
+## Disk Space Usage After Deletes
```
jseldess: Same changes as above.

jseldess:
I'll just take over the final edits.

v1.0/delete.md
jseldess: Done.

v1.1/delete.md
jseldess: Done.

PR #2639, opened 2018-03-08
 Attempt to clarify the node status command's updated_at field

### LINKED ISSUE #2249: Correct description of updated_at and is_live for `node status`
The other question is why the updated_at column seems to disagree with the is_live column. [Our docs](https://www.cockroachlabs.com/docs/stable/view-node-details.html#node-status) on this say that updated_at gives The date and time when the node last reported its status and that is_live states whether the node is live. Under the hood, these two mean different things. The former is referring to the last update time of the persisted node status summaries. The latter is referring to whether the node's liveness status is up-to-date in the NodeLiveness range.

So these two columns are referring to different things. While it looks strange when they diverge, it's not particularly unexpected when a cluster is unhealthy.

a-robinson:
Fixes #2249 

Also add the appropriate descriptions/field names for the v1.0 docs while I'm here.

I'm not sure this is really the best explanation, but I'm also skeptical that it's worth going into great detail here since it could easily confuse people in the common case that there's nothing weird about the values. It'd probably help if we had a good definition of what it means to be "live" that we could link to from here. Do we?

### DIFF
--- a/v1.0/view-node-details.md
+++ b/v1.0/view-node-details.md
@@ -56,19 +56,21 @@ For the `node ls` command, only the `id` field is returned for each node. For th
 
 Field | Description
 ----------|------------
-`id` |
-`address` |
-`build` |
-`updated_at` |
-`started_at` |
-`live_bytes` |
-`key_bytes` |
-`value_bytes` |
-`intent_bytes` |
-`system_bytes` |
-`leader_ranges` |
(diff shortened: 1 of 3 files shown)


jseldess:
@a-robinson, you just need to update your copy of master, rebase, and push again to pass TC.

PR #2643, opened 2018-03-08
CONTRIBUTING: suggest checking Allow edits from maintainers

dt:


jseldess:
You need to update your copy of master, rebase, and push one more time before we can merge.

PR #2653, opened 2018-03-08
Update guide for Azure Network Security Rules

### LINKED ISSUE #2418: Azure NSG docs out of date
Azure has apparently expanded their options for configuring Network Security Groups, and our "Azure" documentation is no longer up to date.

Specifically, the IP configuration has now been split into "source" and "destination".  The existing "Port" field is now "Destination Port", and two new fields have been added; "Destination" (allowing you to restrict the rule to a subset of the internal network) and "Source Port", allowing you to restrict the rule to specific ports from the source.

I believe the recommended defaults should be:

Source Port: "*"
Destination: "Any"
Destination Port (formerly just "port"): 8080.

Screenshot from a user on gitter who ran into this confusion:
azure ports

https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-on-microsoft-azure.html

mrtracy:
Azure has slightly modified the interface for its network security group
rules, and our cloud deployment docs no longer matched.

+ Explicitly specify that the required rules are "Inbound" rules.
+ Add Destination and Destination port fields.
+ Sort fields to match current Azure interface.

Fixes #2418 

Current Azure Interface for reference:
screen shot 2018-03-08 at 1 24 47 pm

### DIFF
--- a/v1.0/deploy-cockroachdb-on-microsoft-azure-insecure.md
+++ b/v1.0/deploy-cockroachdb-on-microsoft-azure-insecure.md
@@ -47,33 +47,35 @@ To enable this in Azure, you must create a Resource Group, Virtual Network, and 
 
 2. [Create a Virtual Network](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-create-vnet-arm-pportal) that uses your **Resource Group**.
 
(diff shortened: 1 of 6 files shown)


mrtracy:
@jseldess I found it while helping a customer run through the guide; they asked for help in Gitter when the guide didn't match the actual interface.

PR #2655, opened 2018-03-08
add descriptions of as of system time's benefits/drawbacks

### LINKED ISSUE #2612: Use AS OF SYSTEM TIME for reads that can be stale
In our [`BACKUP`](https://www.cockroachlabs.com/docs/stable/backup.html#performance) docs, we suggest:

> For best performance, we also recommend always starting backups with a specific timestamp at least 10 seconds in the past... This improves performance by decreasing the likelihood that the BACKUP will be retried because it contends with other statements/transactions.

We should add this recommendation for reads that can be a little stale. Using `AS OF SYSTEM TIME` in such cases makes it less likely that the reads will content with concurrent statements/transactions. 

See also https://forum.cockroachlabs.com/t/table-hints-for-isolation-mode-without-a-transaction/1412/6

sploiselle:
Closes #2612

### DIFF
--- a/v2.0/select.md
+++ b/v2.0/select.md
@@ -27,7 +27,7 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
 | `AS col_label` | In the retrieved table, change the column label to `col_label`. |
 | `table_ref` | The [table expression](table-expressions.html) you want to retrieve data from.|
 | `index_name` | The name of the index you want to use, also known as "[index hints](#force-index-selection-index-hints)." Find index names using [`SHOW INDEX`](show-index.html). <br/><br/>Forced index selection overrides [CockroachDB's index selection](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/). |
-| `AS OF SYSTEM TIME timestamp` | Retrieve data as it existed as of [`timestamp`](timestamp.html).<br/><br/>For more information, see [this example](#select-historical-data-time-travel) or our blog post [Time-Travel Queries](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/). |
+| `AS OF SYSTEM TIME timestamp` | Retrieve data as it existed as of [`timestamp`](timestamp.html).<br/><br/>For more information, see [this example](#select-historical-data-time-travel) or our blog post [Time-Travel Queries](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).<br/><br/>Because `AS OF SYSTEM TIME` returns historical data, your reads might be stale. |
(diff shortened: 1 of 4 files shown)


v2.0/select.md
```diff
@@ -27,7 +27,7 @@ The user must have the `SELECT` [privilege](privileges.html) on the table.
...
 | `table_ref` | The [table expression](table-expressions.html) you want to retrieve data from.|
 | `index_name` | The name of the index you want to use, also known as "[index hints](#force-index-selection-index-hints)." Find index names using [`SHOW INDEX`](show-index.html). <br/><br/>Forced index selection overrides [CockroachDB's index selection](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/). |
-| `AS OF SYSTEM TIME timestamp` | Retrieve data as it existed as of [`timestamp`](timestamp.html).<br/><br/>For more information, see [this example](#select-historical-data-time-travel) or our blog post [Time-Travel Queries](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/). |
+| `AS OF SYSTEM TIME timestamp` | Retrieve data as it existed as of [`timestamp`](timestamp.html).<br/><br/>For more information, see [this example](#select-historical-data-time-travel) or our blog post [Time-Travel Queries](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).<br/><br/>Because `AS OF SYSTEM TIME` returns historical data, your reads might be stale. |
```
jseldess: I think it'd be helpful to cross-reference the new section in `performance-best-practices-overview.md`.

v2.0/transactions.md
```diff
@@ -61,7 +61,7 @@ Type | Description
...
 
-Transactions may require retries if they experience deadlock or read/write contention with other concurrent transactions which cannot be resolved without allowing potential [serializable anomalies](https://en.wikipedia.org/wiki/Serializability).
+Transactions may require retries if they experience deadlock or read/write contention with other concurrent transactions which cannot be resolved without allowing potential [serializable anomalies](https://en.wikipedia.org/wiki/Serializability). (However, it's possible to mitigate read-write conflicts by performing reads using [`AS OF SYSTEM TIME`](select.html#select-historical-data-time-travel).)
```
jseldess: Instead of linking to the `SELECT` example, it might be more informative to link to the new section in `performance-best-practices-overview.md`.

v2.0/select.md
jseldess: Also, can you mention this use case in the example itself and link to `performance-best-practices-overview.md` from there, too?

v2.0/transactions.md
jseldess: Done.

v2.0/select.md
jseldess: Done.

PR #2658, opened 2018-03-08
Document ORDER BY and row ordering

### LINKED ISSUE #2518: Enhance description of ORDER BY in SELECT / UPDATE / DELETE
Specifically, the `ORDER BY sortby_list` definitions. This might be a good issue for Docs BugFixIt Day!

knz:
Fixes  #2518.

@RaduBerinde can you have a look at the new page. Can you in particular remind me where ordering is significant so I can add the cases to the list.


knz:
@RaduBerinde  in particular what is the rule again for order-sensitive aggregations like `array_agg` or `concat_agg`?

RaduBerinde:
Very nice! :lgtm_strong: 

Regarding `array_agg` and `concat_agg` - strictly speaking (SQL standard), I don't think they need to respect the datasource ordering. But in practice, I'm guessing people expect it. Postgres doc says "it will usually work" but notes that it's not portable; the recommended way is to use an ORDER BY clause inside the aggregate function call: `SELECT array_agg(a ORDER BY b DESC) FROM table;` (which we don't support AFAIK).

I believe that implementation-wise CRDB does respect the datasource ordering currently.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[v2.0/query-order.md, line 57 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76Vyoa9bFrPLK1WI8U:-L76Vyoa9bFrPLK1WI8V:b-14e807) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L57)):*
> ```Markdown
> significant:
> 
> - the ordering of the data source in a `SELECT` clause that uses
> ```

Is this true if the ORDER BY is inside the datasource, or is this only true for the top-level ORDER BY?

I'm pretty sure that currently we do preserve the ordering in both cases, just not sure of what semantics we need to promise.

---

*[v2.0/query-order.md, line 64 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76WdFu8y3MOQ2sF4Vv:-L76WdFu8y3MOQ2sF4Vw:b4rh42r) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L64)):*
> ```Markdown
>   statement that also uses `LIMIT`, is preserved.
> 
> (FIXME: Radu, what are the other cases?)
> ```

I can't think of any other cases to mention here (other than concat_agg/array_agg/json_agg).

---

*[v2.0/query-order.md, line 182 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76X3dFASJszWFqHQde:-L76X3dFASJszWFqHQdf:b-oes8r9) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L182)):*
> ```Markdown
> in the order of a given index.
> 
> The particular advantage is that for most queries, this guarantees the
> ```

for queries using that index

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-:-L76X93lB3qgZhNeqB2T:bxy5h9a)*

knz:
Yes you and I know what cockroachdb currently happens to do for local execution, but I'd say with the current code we're throwing ordering out of the door as soon as a query is distributed. So I think I won't tell anything about what crdb currently happens to preserve and stick to what's guaranteed.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[v2.0/query-order.md, line 57 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76Vyoa9bFrPLK1WI8U:-L76lzC2ECxrzlGAYx7u:bym5ebm) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L57)):*
<details><summary><i>Previously, RaduBerinde wrote…</i></summary><blockquote>

Is this true if the ORDER BY is inside the datasource, or is this only true for the top-level ORDER BY?

I'm pretty sure that currently we do preserve the ordering in both cases, just not sure of what semantics we need to promise.
</blockquote></details>

This point was bogus, I checked: DISTINCT ON respects the ORDER BY in the same SELECT.

---

*[v2.0/query-order.md, line 64 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76WdFu8y3MOQ2sF4Vv:-L76m3sxBXxcXRdxt1dm:b93vamr) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L64)):*
<details><summary><i>Previously, RaduBerinde wrote…</i></summary><blockquote>

I can't think of any other cases to mention here (other than concat_agg/array_agg/json_agg).
</blockquote></details>

There were two more: WITH ORDINALITY and scalar subqueries.

---

*[v2.0/query-order.md, line 182 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-L76X3dFASJszWFqHQde:-L76mIhm5gD8nlNM5t-G:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/3b0d4b35d0fd5a1359c9e3c73ccb90e3f3b69e68/v2.0/query-order.md#L182)):*
<details><summary><i>Previously, RaduBerinde wrote…</i></summary><blockquote>

for queries using that index
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-:-L76mMYnC1_fQnlb3ulU:b-yv9o3a)*

knz:
I added a paragraph about aggregation ordering and the known limitation.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2658#-:-L76nhQHBhx3d_RwJzyD:b-3qcbn3)*

knz:
@jseldess please take this over, including generating the missing diagram.
I might make a final pass to check the copy-edits

RaduBerinde:
LGTM

> Yes you and I know what cockroachdb currently happens to do for local execution, but I'd say with the current code we're throwing ordering out of the door as soon as a query is distributed.

I don't think that's the case. We've been careful to guarantee orderings required by sort nodes anywhere in the plan, and we specifically don't do local aggregation for ARRAY_AGG. E.g. check out these testcases: https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/logictest/testdata/logic_test/distsql_agg#L618

knz:
>  We've been careful to guarantee orderings required by sort nodes anywhere in the plan, and we specifically don't do local aggregation for ARRAY_AGG.

I see. Well maybe we could drop some of them. I'm ok with being closer to the sql standard and pg if it makes some queries run fast(er).

knz:
Do you agree that we should not document the current behavior to not provide these guarantees?

RaduBerinde:
Yes, definitely. We might indeed relax these in the optimizer.

jseldess:
Thanks very much, @knz. Hope to take this over sometime today.

jseldess:
@knz, sorry for the delay. Please take a look at my edits. In the meantime, I'll generate and add the sql diagram to the `query-order.md` file.

knz:
:lgtm: thank you!

---

Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[_includes/sidebar-data-v2.0.json, line 660 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2658#-L7VAgGEYUj3eeinL_iA:-L7VAgGEYUj3eeinL_iB:b5l05cu) ([raw file](https://github.com/cockroachdb/docs/blob/1be0966f34abc0ac3721f5d2880929bc51214083/_includes/sidebar-data-v2.0.json#L660)):*
> ```JSON
>               "/${VERSION}/query-order.html"
>             ]
>           },          
> ```

nit: stray spaces

---

*[v2.0/query-order.md, line 8 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2658#-L7VAoluh8PTi4gUkRQS:-L7VAoluh8PTi4gUkRQT:b870rrr) ([raw file](https://github.com/cockroachdb/docs/blob/1be0966f34abc0ac3721f5d2880929bc51214083/v2.0/query-order.md#L8)):*
> ```Markdown
> 
> The `ORDER BY` clause controls the order in which rows are returned or processed.
> It can be used in [`SELECT` clauses](select.html) and [`DELETE`](delete.html) and [`UPDATE`](update.html) statements.
> ```

Also INSERT and UPSERT. I forgot.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2658#-:-L7VBX3swZzPpkOl0RkH:bro3093)*

jseldess:
Review status: 4 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[_includes/sidebar-data-v2.0.json, line 660 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2658#-L7VAgGEYUj3eeinL_iA:-L7VE9a5QxuU-gc49cdq:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1be0966f34abc0ac3721f5d2880929bc51214083/_includes/sidebar-data-v2.0.json#L660)):*

nit: stray spaces
</blockquote></details>

Done.

---

*[v2.0/query-order.md, line 8 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2658#-L7VAoluh8PTi4gUkRQS:-L7VEFP4J6RTHCPadnY2:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1be0966f34abc0ac3721f5d2880929bc51214083/v2.0/query-order.md#L8)):*

Also INSERT and UPSERT. I forgot.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2658)*

jseldess:
Can't generate the `sort_clause` diagram until this is fixed: https://github.com/cockroachdb/docs/issues/2712

jseldess:
Updated the full grammar page for 2.0, which exposed a bunch of broken links.

jseldess:
@knz, I'm finally ready to merge this. But while regenerating a bunch of sql diagrams to match the full 2.0 grammar, I noticed that `INSERT` and `UPSERT` do not have any `sort_clause`:
 - [`INSERT`](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/7c3e503e8b7dad509d119c6c77a7659763046a73/dev/sql-grammar.html#insert_stmt)
- [`UPSERT`](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/7c3e503e8b7dad509d119c6c77a7659763046a73/dev/sql-grammar.html#upsert_stmt)

We call out those two statements in the **Ordering Query Results** page, though, so I'm a little confused. Any idea why the grammar wouldn't indicate support for `ORDERY BY` for those statements?

knz:
So insert and upsert accept "order by" because their operand is an entire `select_stmt` which itself can contain an order by clause. We'd need to massage the diagram parameters for them to reveal that properly.

knz:
Oh it's actually worse than that.

We say that "the SELECT statement and the SELECT clause are more or less the same thing" but this is very wrong actually. A SELECT *statement* can do several things more than a SELECT clause.

INSERT / UPSERT take an entire SELECT statement, which has more ORDER BY abilities than a SELECT clause. 

Also a SELECT statement can also be _any selection clause_, not just SELECT clauses.

I think if we spend a little time later to discuss (#2719) we should also discuss a strategy on how to reveal all this properly.

jseldess:
Thanks, @knz. I've recorded the open issues here: https://github.com/cockroachdb/docs/issues/2719. I'll merge this PR, since it already adds a lot of value, and schedule time for us to finish up.

Issue #2659, opened 2018-03-08
Jekyll not picking up include on pages with "allowed_hashes"


rmloveland:
On pages like those linked below (which have the `allowed_hashes` key in the YAML header -- maybe the clue) adding an include to the page only works if you click the tabs into the "right-most" option.

<https://github.com/cockroachdb/docs/blob/master/v2.0/install-cockroachdb.html> is a good example, because in order to see the included content (which you can replicate using the patch at the bottom of this comment), you need to click the *Windows* tab, and then the *Use Docker* tab.

This issue also affects <https://github.com/cockroachdb/docs/tree/master/v2.0/start-a-local-cluster-in-docker.md>, which prompted me to file the issue.  In that case I was trying to add the note, but it would only appear when I clicked the *Windows* tab.

```
diff --git a/v2.0/install-cockroachdb.html b/v2.0/install-cockroachdb.html
index 1f449f7c..c0a94bed 100644
--- a/v2.0/install-cockroachdb.html
+++ b/v2.0/install-cockroachdb.html
@@ -563,6 +563,10 @@ $(document).ready(function(){
 <h2 id="whats-next">What's Next?</h2>
 <p><a href="start-a-local-cluster-in-docker.html#os-windows">Quick start</a> a multi-node cluster across multiple Docker containers on a single host, using Docker volumes to persist node data, or explore running a physically distributed cluster in containers using <a href="orchestration.html">orchestration</a> tools.</p>
 
+{{site.data.alerts.callout_info}}
+{% include available-capacity-metric.md %}
+{{site.data.alerts.end}}
+
 {% include diagnostics-callout.html %}
 
 </div>
```

jseldess:
@rmloveland, I don't think this is an issue anymore, but please open if you disagree.

PR #2660, opened 2018-03-08
Add disk usage info to '{Start,Secure} Cluster'

### LINKED ISSUE #2569: 'Start a Local Cluster' is missing info re: "incorrect" disk usage
We don't note on either of the below pages that the disk usage will be "incorrect" when running multiple nodes on the same machine:

https://www.cockroachlabs.com/docs/stable/secure-a-cluster.html
https://www.cockroachlabs.com/docs/stable/start-a-local-cluster.html

The reasons why are explained in https://github.com/cockroachdb/cockroach/issues/5454

This was partially addressed in the actual Admin UI docs via https://github.com/cockroachdb/docs/pull/1849, but I ran into it as a new user when doing some onboarding work and I ended up bugging people with a question so thought I'd flag.  Since the "start a local cluster" docs also show the Admin UI (including the "weird" totals) it would be nice to note there as well IMO.

rmloveland:
Fixes #2569

Specifically:

1. This is listed as a known limitation of v1.1 and in v2.0, so we break
   that content out into an include

2. From there, we use that include in the various places that make note
   of that limitation (basically, the existing Admin UI docs)

3. Then, we can easily *add* the information to the various 'Start/Secure
   a Cluster' pages by snarfing the include

Note: I tried, but was unable to add the included note to the "Start a
Cluster in Docker" pages due to
https://github.com/cockroachdb/docs/issues/2659


_includes/available-capacity-metric.md
```diff
@@ -0,0 +1 @@
+If you are running multiple nodes on a single machine (not recommended in production) and didn't specify the maximum allocated storage capacity for each node using the [`--store`](start-a-node.html#store) flag, the available capacity shown in the [**Capacity**](admin-ui-storage-dashboard.html#capacity) graph in the Admin UI is incorrect. This is because when multiple nodes are running on a single machine, the machine's hard disk is treated as an available store for each node, while in reality, only one hard disk is available for all nodes. The total available capacity is then calculated as the hard disk size multiplied by the number of nodes on the machine.
```
jseldess: We should probably expand this to also reference **Capacity Used** in the **Summary** panel:

```
... the **Capacity Used** metric in the [**Summary**](admin-ui-access-and-navigate.html#summary-panel) panel and the available capacity shown in the [**Capacity**](admin-ui-storage-dashboard.html#capacity) graph...
```

v1.1/secure-a-cluster.md
```diff
@@ -212,6 +212,10 @@ As mentioned earlier, CockroachDB automatically replicates your data behind-the-
 
 The replica count on each node is identical, indicating that all data in the cluster was replicated 3 times (the default).
 
+{{site.data.alerts.callout_info}}
```
jseldess: To keep these getting started pages lean, perhaps instead of embedding the include, we could link to the known limitation, for example:

```
{{site.data.alerts.callout_info}}Capacity metrics can be incorrect when running multiple nodes on a single machine. For more details, see this [limitation](known-limitations.html#available-capacity-metric-in-the-admin-ui).{{site.data.alerts.end}}
```

v1.1/start-a-local-cluster.md
```diff
@@ -181,6 +181,10 @@ As mentioned earlier, CockroachDB automatically replicates your data behind-the-
 
 The replica count on each node is identical, indicating that all data in the cluster was replicated 3 times (the default).
 
+{{site.data.alerts.callout_info}}
```
jseldess: Same as comment above:

To keep these getting started pages lean, perhaps instead of embedding the include, we could link to the known limitation, for example:

```
{{site.data.alerts.callout_info}}Capacity metrics can be incorrect when running multiple nodes on a single machine. For more details, see this [limitation](known-limitations.html#available-capacity-metric-in-the-admin-ui).{{site.data.alerts.end}}
```

v2.0/admin-ui-access-and-navigate.md
```diff
@@ -56,7 +56,7 @@ Metric | Description
 --------|----
 Total Nodes | The total number of nodes in the cluster. <a href='admin-ui-access-and-navigate.html#decommissioned-nodes'>Decommissioned nodes</a> are not included in the Total Nodes count. <br><br>You can further drill down into the nodes details by clicking on [**View nodes list**](#nodes-list).
 Dead Nodes | The number of [dead nodes](admin-ui-access-and-navigate.html#dead-nodes) in the cluster.
-Capacity Used | The storage capacity used as a percentage of total storage capacity allocated across all nodes.<br><br>**Note:** If you are running multiple nodes on a single machine (not recommended), this value may be incorrect. This is because when multiple nodes are running on a single machine, the machine's hard disk is treated as separate stores for each node, while in reality, only one hard disk is used for all nodes. The Capacity Used is then displayed as the hard disk capacity used multiplied by the number of nodes on the machine.
+Capacity Used | The storage capacity used as a percentage of total storage capacity allocated across all nodes.<br /><br />**Note**: {% include available-capacity-metric.md %}
```
jseldess: It looks like using an include in the table does, in fact, break things: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/ae5878f35d2552936f680b7306a56385dee2ebbf/stable/admin-ui-access-and-navigate.html#summary-panel

Can you play around with it and, if necessary, paste the text here directly?

v2.0/secure-a-cluster.md
```diff
@@ -212,6 +212,10 @@ As mentioned earlier, CockroachDB automatically replicates your data behind-the-
 
 The replica count on each node is identical, indicating that all data in the cluster was replicated 3 times (the default).
 
+{{site.data.alerts.callout_info}}
```
jseldess: Same as for v1.1 version:

To keep these getting started pages lean, perhaps instead of embedding the include, we could link to the known limitation, for example:

```
{{site.data.alerts.callout_info}}Capacity metrics can be incorrect when running multiple nodes on a single machine. For more details, see this [limitation](known-limitations.html#available-capacity-metric-in-the-admin-ui).{{site.data.alerts.end}}
```

v2.0/start-a-local-cluster.md
```diff
@@ -181,6 +181,10 @@ As mentioned earlier, CockroachDB automatically replicates your data behind-the-
 
 The replica count on each node is identical, indicating that all data in the cluster was replicated 3 times (the default).
 
+{{site.data.alerts.callout_info}}
```
jseldess: Same as for v1.1 version:

To keep these getting started pages lean, perhaps instead of embedding the include, we could link to the known limitation, for example:

```
{{site.data.alerts.callout_info}}Capacity metrics can be incorrect when running multiple nodes on a single machine. For more details, see this [limitation](known-limitations.html#available-capacity-metric-in-the-admin-ui).{{site.data.alerts.end}}
```

v2.0/admin-ui-access-and-navigate.md
jseldess: You could also just do the include after the table, as you've done in other places.

jseldess:
Also, you need to update your copy of master, rebase this branch against it, and then push again to pass TC:

```
git checkout master
git pull master
git checkout fixitday-local-cluster-disk-usage
git rebase master
git push --force origin fixitday-local-cluster-disk-usage
```

_includes/available-capacity-metric.md
rmloveland: Sweet, thanks!  Updated as suggested in bac307f9

v1.1/secure-a-cluster.md
rmloveland: Great suggestion - updated in 3e80ace3

v1.1/start-a-local-cluster.md
rmloveland: Updated in 3e80ace3

v2.0/secure-a-cluster.md
rmloveland: Thanks, addressed in eacc8ede

v2.0/start-a-local-cluster.md
rmloveland: Addressed in eacc8ede

v2.0/admin-ui-access-and-navigate.md
rmloveland: Thanks for the catch!  I updated this page and the v1.1 version to do the include after the table in 7076917f.  Tables appear to be building correctly now "on my machine (tm)" :-)

rmloveland:
@jseldess I have a couple of questions:

- Do you have further feedback?  I think I've addressed the issues you pointed out, but I'm happy to fix any other things you noticed

- Should I ask for another review or is that not necessary since we are exposing information that already existed elsewhere in the docs, just in more places?

jseldess:
LGTM, @rmloveland! 

Please add `Fixes #<issue number>` to your PR description so the related issue auto-resolves. And then merge away!

Issue #2365, opened 2018-01-17
How to get the current cluster offset


jseldess:
Document how get the current cluster offset via the `clock_offset_*` variables at `/_status/vars`.

jseldess:
We can add a new question to the Operational FAQs page: https://www.cockroachlabs.com/docs/dev/operational-faqs.html#what-happens-when-node-clocks-are-not-properly-synchronized

PR #2663, opened 2018-03-08
 Add information on how to monitor clock offsets to the Operational FAQ

a-robinson:
Also consolidate FAQ answers about clock synchronization while I'm here.

Fixes https://github.com/cockroachdb/docs/issues/2365


jseldess:
@a-robinson, TC is reporting that Jekyll can't built the html:

```
[20:35:28]  Liquid Exception: Could not locate the included file 'faq/clock-synchronization_v2.0.html' in any of ["/docs/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in v2.0/recommended-production-settings.md
```

Since you renamed the include file, you have to update any references to it.

a-robinson:
Done, sorry for missing those other couples references.

PR #2670, opened 2018-03-08
Specify restrictions on the target of `init`

### LINKED ISSUE #2589: Correct cockroach init docs
Docs need to make explicit that the command must be run against a node that appears the `--join` flag of other nodes. The CLI help text already says this.

pbardea:
Explained that the target of the `cockroach init` command must be one of
the nodes which appear in the `--join` flag of the `cockroach start`
command.

Fixes #2589

### DIFF
--- a/v1.1/initialize-a-cluster.md
+++ b/v1.1/initialize-a-cluster.md
@@ -29,7 +29,7 @@ The `cockroach init` command supports the following [general-use](#general) and 
 Flag | Description
 -----|-----------
 `--certs-dir` | The path to the [certificate directory](create-security-certificates.html). If the cluster is secure, this directory must contain a valid CA certificate and a client certificate and key for the `root` user. Client certificates for other users are not supported.<br><br>**Env Variable:** `COCKROACH_CERTS_DIR`<br>**Default:** `${HOME}/.cockroach-certs/`
-`--host` | The server host to connect to. This can be the address of any node. <br><br>**Env Variable:** `COCKROACH_HOST`<br>**Default:**`localhost`
+`--host` | The server host to connect to. This must appear in the `--join` flag of the other nodes. <br><br>**Env Variable:** `COCKROACH_HOST`<br>**Default:**`localhost`
 `--insecure` | Run in insecure mode. If this flag is not set, the `--certs-dir` flag must point to valid certificates.<br><br>**Env Variable:** `COCKROACH_INSECURE`<br>**Default:** `false`
 `--port` | The server port to connect to. <br><br>**Env Variable:** `COCKROACH_PORT`<br>**Default:** `26257`
 
--- a/v2.0/initialize-a-cluster.md
+++ b/v2.0/initialize-a-cluster.md
@@ -29,7 +29,7 @@ The `cockroach init` command supports the following [general-use](#general) and 
 Flag | Description
 -----|-----------
 `--certs-dir` | The path to the [certificate directory](create-security-certificates.html). If the cluster is secure, this directory must contain a valid CA certificate and a client certificate and key for the `root` user. Client certificates for other users are not supported.<br><br>**Env Variable:** `COCKROACH_CERTS_DIR`<br>**Default:** `${HOME}/.cockroach-certs/`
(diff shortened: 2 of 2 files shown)


v2.0/initialize-a-cluster.md
```diff
@@ -4,7 +4,7 @@ summary: Perform a one-time-only initialization of a CockroachDB cluster.
 toc: false
 ---
 
-<span class="version-tag">New in v1.1:</span> This page explains the `cockroach init` [command](cockroach-commands.html), which you use to perform a one-time initialization of a new multi-node cluster. For a full walk-through of the cluster startup and initialization process, see one of the [Manual Deployment](manual-deployment.html) tutorials.
+<span class="version-tag">New in v1.1:</span> This page explains the `cockroach init` [command](cockroach-commands.html), which you use to perform a one-time initialization of a new multi-node cluster. The target of the `cockroach init` command must appear in the `--join` flag of the [`cockroach start`](start-a-cluster.html) command which have been run on the other nodes in the cluster. For a full walk-through of the cluster startup and initialization process, see one of the [Manual Deployment](manual-deployment.html) tutorials.
```
jseldess: See comment for the v1.1 version of the page.

pbardea: Which comment are you referring to on the v1.1 version of the page? I'm not seeing any Github comments on that page.

Also I've gone ahead and fixed TC if that is what it was referring to.

v1.1/initialize-a-cluster.md
```diff
@@ -4,7 +4,7 @@ summary: Perform a one-time-only initialization of a CockroachDB cluster.
 toc: false
 ---
 
-<span class="version-tag">New in v1.1:</span> This page explains the `cockroach init` [command](cockroach-commands.html), which you use to perform a one-time initialization of a new multi-node cluster. For a full walk-through of the cluster startup and initialization process, see one of the [Manual Deployment](manual-deployment.html) tutorials.
+<span class="version-tag">New in v1.1:</span> This page explains the `cockroach init` [command](cockroach-commands.html), which you use to perform a one-time initialization of a new multi-node cluster. The target of the `cockroach init` command must appear in the `--join` flag of the [`cockroach start`](start-a-node.html) command which have been run on the other nodes in the cluster. For a full walk-through of the cluster startup and initialization process, see one of the [Manual Deployment](manual-deployment.html) tutorials.
```
jseldess: Ugh, I seem to have not submitted this comment before. Sorry.

Instead of mentioning this here, I'd suggest updating the `--host` flag description as follows:

> The server host to connect to. This must appear in the `--join` flag of other nodes. 

@bdarnell, does the target host address need to appear in the join flag of all or any of the other nodes?

bdarnell: It's `--advertise-host`, not `--host` that must appear in `--join` and be used from client commands.

We recommend using the same `--join` arguments on all nodes, and while they may diverge over time they should at least be the same at `init` time. 

Putting this message in the first paragraph of the command seems strange to me; I would put it in the docs for the `--host` flag. I don't think we need to make this warning too prominent since at the time that you're initializing the cluster you shouldn't have too many nodes or a complex configuration.

jseldess: I'm missing something, @bdarnell. `--advertise-host` isn't even an option on client commands like `init`...  Or are you just saying that, when starting a cluster, if you use `--advertise-host` on nodes, that's what needs to be used in `--join` as well?

bdarnell: Sorry, I got mixed up. I thought you were suggesting that the server-side `--host` flag should say that this is what should appear in the `--join` flag.

In retrospect, instead of `--advertise-host` and `--host`, we should have had `--host` and `--bind-host`. It's the advertised host that matters everywhere but on the node itself.

pbardea:
Moved the message from the top of the page to the description `--host` flag, as suggested.

jseldess:
Thanks, @pbardea. @bdarnell, I've opened a separate issue to update the description of `cockroach start --join` based on your comments: https://github.com/cockroachdb/docs/issues/2703

PR #2677, opened 2018-03-08
Document changes to the built-in replication zones

### LINKED ISSUE #2387: Clusters are now initialized with default .meta and .liveness replication zones with lower GC TTL configurations.
This doc, and maybe other docs, need to be updated: https://www.cockroachlabs.com/docs/v2.0/configure-replication-zones.html#create-a-replication-zone-for-system-ranges

We also need to mention the `.liveness` replication zone.

PR: https://github.com/cockroachdb/cockroach/pull/17628

a-robinson:
Fixes #2387

### DIFF
--- a/v2.0/configure-replication-zones.md
+++ b/v2.0/configure-replication-zones.md
@@ -5,7 +5,7 @@ keywords: ttl, time to live, availability zone
 toc: false
 ---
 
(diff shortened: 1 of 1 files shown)


v2.0/configure-replication-zones.md
```diff
@@ -156,6 +156,23 @@ If you need to troubleshoot this command's behavior, you can change its [logging
 
 These examples focus on the basic approach and syntax for working with zone configuration. For examples demonstrating how to use constraints, see [Scenario-based Examples](#scenario-based-examples).
 
+###  View the Pre-Configured Replication Zones
```
jseldess: nit: I'd title this **List the Pre-Configured Replication Zones**, since we're not really viewing their configs.

a-robinson: Done.

jseldess:
LGTM!

PR #2678, opened 2018-03-08
Add copy-clipboard to commands on the Configure Replication Zones page

a-robinson:
Ignore the first commit, it's just #2677

### DIFF
--- a/v1.0/configure-replication-zones.md
+++ b/v1.0/configure-replication-zones.md
@@ -161,6 +161,7 @@ The cluster-wide replication zone (`.default`) is initially set to replicate dat
 
 To view the default replication zone, use the `cockroach zone get .default` command with appropriate flags:
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ cockroach zone get .default --insecure
 ~~~
@@ -179,6 +180,7 @@ constraints: []
 
 To edit the default replication zone, create a YAML file defining only the values you want to change (other values will not be affected), and use the `cockroach zone set .default -f <file.yaml>` command with appropriate flags:
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ cat default_update.yaml
 ~~~
@@ -187,6 +189,7 @@ $ cat default_update.yaml
 num_replicas: 5
 ~~~
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ cockroach zone set .default --insecure -f default_update.yaml
 ~~~
@@ -202,6 +205,7 @@ constraints: []
 
 Alternately, you can pass the YAML content via the standard input:
 
+{% include copy-clipboard.html %}
 ~~~ shell
 $ echo 'num_replicas: 5' | cockroach zone set .default --insecure -f -
 ~~~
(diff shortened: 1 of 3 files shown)


v1.0/configure-replication-zones.md
```diff
@@ -495,6 +517,7 @@ There's no need to make zone configuration changes; by default, the cluster is c
 
 1. Start each node with a different locality attribute:
 
+    {% include copy-clipboard.html %}
```
jseldess: Does the copy work when there are multiple commands? I'm not sure...

a-robinson: Yes. It copies the lines prefixed with `#` as well, but that's fine because shells will ignore those lines and execute the rest.

jseldess: Right, the comments aren't a problem but rather the multiple commands. All but the last command will attempt to execute and will fail (in this case, due to placeholder text). For example, here's what I get when I copy and paste into a terminal:

```
~/src/github.com/cockroachdb/cockroach$ # Start the two nodes in datacenter 1:
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node1 hostname> --locality=datacenter=us-1
-bash: node1: No such file or directory
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node2 hostname> --locality=datacenter=us-1 --join=<node1 hostname>:27257
-bash: node2: No such file or directory
~/src/github.com/cockroachdb/cockroach$
~/src/github.com/cockroachdb/cockroach$ # Start the two nodes in datacenter 2:
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node3 hostname> --locality=datacenter=us-2 --join=<node1 hostname>:27257
-bash: node3: No such file or directory
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node4 hostname> --locality=datacenter=us-2 --join=<node1 hostname>:27257
-bash: node4: No such file or directory
~/src/github.com/cockroachdb/cockroach$
~/src/github.com/cockroachdb/cockroach$ # Start the two nodes in datacenter 3:
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node5 hostname> --locality=datacenter=us-3 --join=<node1 hostname>:27257
-bash: node5: No such file or directory
~/src/github.com/cockroachdb/cockroach$ cockroach start --insecure --host=<node6 hostname> --locality=datacenter=us-3 --join=<node1 hostname>:27257
```

For that reason, we usually try to avoid using copy-to-clipboard on code blocks including multiple commands. One approach is to split the commands into separate code blocks, each with a copy-to-clipboard include. The other approach is to just not use copy-to-clipboard on these examples. What are you thoughts?

jseldess: I should really update https://github.com/cockroachdb/docs/blob/master/CONTRIBUTING.md#code-samples to be more explicit. Opened an issue: https://github.com/cockroachdb/docs/issues/2688.

jseldess:
@a-robinson, now that https://github.com/cockroachdb/docs/pull/2677 has merged, you'll need to resolve some conflicts here.

PR #2689, opened 2018-03-10
[Test 1] Try to speed up htmlproofer

jseldess:
- Disable external link checking.
- Disable image and script checking.
- Use 4 cpus.


jseldess:
TC completed in 11m:52s.

PR #2690, opened 2018-03-10
[Test 2] Try to speed up htmlproofer

jseldess:
Same as https://github.com/cockroachdb/docs/pull/2689
but without increasing cpu usage.


jseldess:
TC finished in 27m:42s. This makes it seem like disabling external links doesn't result in any benefit, which is confusing.

PR #2691, opened 2018-03-10
[Test 3] Try to speed up htmlproofer

jseldess:
Same as https://github.com/cockroachdb/docs/pull/2689
but increasing threads instead of cpu.


jseldess:
TC finished in 30m:44s. No apparent benefit to increasing threads.

PR #2694, opened 2018-03-10
[Test 6] Try to speed up htmlproofer

jseldess:
- Use 4 cpus.
- Do not disable external link checking.
- Do not disable image or script checking.


jseldess:
TC finished in 14m:30s.

PR #2698, opened 2018-03-11
Use single quotes rather than backticks on field names

nstewart:


lhirata:
LGTM, thanks for catching this!

PR #2700, opened 2018-03-12
 Use htmltest instead of HTML proofer for faster CI

### LINKED ISSUE #2503: Change CI for Docs
Currently, `html-proofer` runs against the entire site (all versions of docs) on each PR. This takes around 30 min, which is a long time to wait for link checking. It'll get worse every time we add another version subdirectory.

Proposed change:
- On each PR, run `hmtl-proofer` against new/changed pages only.
- Run a nightly test against the entire site, auto-creating an issue with the failure details. 

@benesch, let me know if you have other ideas.

benesch:
htmltest is a Go reimplemention of HTML proofer that is much faster.
Pull request CI now takes seconds instead of minutes.

Fix #2503.


benesch:
Check it out! A docs build in 50s: https://teamcity.cockroachdb.com/viewLog.html?buildId=559453&buildTypeId=Internal_Docs_GitHubCi

htmltest also turned up some broken links that HTML proofer wasn't finding. I fixed as many as I could, but I don't have the stamina to fix the broken grammar links at the moment.

benesch:
P.S. Checking external links still takes on the order of minutes. This PR adjusts things so that external links aren't checked on every pull request, but only once every night.

v1.0/multi-active-availability.md
```diff
@@ -62,6 +62,6 @@ For this example, we have 3 CockroachDB nodes (**A**, **B**, **C**) in a multi-a
 
 ## What's next?
 
-To get a greater understanding of how CockroachDB is a survivable system that enforces strong consistency, check out our [architecture documentation](/architecture/overview.html).
+To get a greater understanding of how CockroachDB is a survivable system that enforces strong consistency, check out our [architecture documentation]({{ '/architecture/overview.html' | relative_url }}).
```
jseldess: Thanks for catching this one `htmltest` and Nikhil! So whenever we link to a page in the same version but in another directory, we need to do this?

jseldess:
Btw, do you need anyone else on eng to review this?

benesch:
> I supposed we'll just have to make TC not block merging until I or someone on my team can find time to clean up the grammar diagrams, which might take a while.

Sorry, forgot to mention that I hacked around the issue for now. See the dummy targets I added to each versions `sql-grammar.md` files. That'll allow us to clean up the diagrams bit by bit.

v1.0/multi-active-availability.md
benesch: Yep! I'm surprised HTML proofer didn't catch this, actually.

benesch:
Discussed on Slack and decided this is ready to go!

PR #2707, opened 2018-03-12
Fix version switcher URLs

benesch:
Version switcher URLs should have a trailing slash iff they are a
directory. Jekyll already handles this logic for us, so we just need to
teach the version injector not to strip the trailing slash if it exists.


benesch:
You bet! Sorry for the chaos.

PR #2710, opened 2018-03-13
Remove withdrawn releases from full release list

jseldess:
But leave the release note pages in place so as
not to break incoming links.


a-robinson:
@jseldess I'm sorry, but I never rolled the release notes for the 20180122 release into the subsequent one (https://github.com/cockroachdb/docs/pull/2548), so there may still be a little more work to do.

PR #2725, opened 2018-03-15
Add more robots tags to training page headers

### LINKED ISSUE #2724: Tell Google to forget about old versions of our training pages
All pages already tell Google no to index them. But there are still lots of cached versions of the pages from prior to adding "noindex".

Using some admin console, @mberhault, already asked Google to temporarily remove all of those old versions from search results and their cache. In conjunction, we should add `nofollow`, `noarchive`, and `nocache` to all of the training pages so they disappear permanently.

jseldess:
- noindex
- nofollow
- noarchive
- nocache

Fixes #2724


mberhault:
LGTM

PR #2740, opened 2018-03-16
[WIP] Algolia site search

j-low:
__[description]__
* implement site search with Algolia
* implement versioned search based on site cookie


j-low:
cc: @jseldess @kuanluo @lhirata @Amruta-Ranade @rmloveland 

This PR is an MVP of site search with Algolia.  Algolia provides a lot of nice functionality that I would like to explore adding for this initial version.  I'll reach out to the cc'd above for your suggestions early next week.

jseldess:
@j-low, I'm running your branch locally. The search is fast, and it seems to return only content for the currently selected version, which is great. 

I'm seeing one main issue right now: When searching a term, the same page is represented multiple times in the search result. Here's an extreme example:

screen shot 2018-03-19 at 12 04 21 pm

Also, a few question:
- What is our plan for the quick results (that display before hitting Enter)?
- Are we returning all results or only one page worth? If the latter, what's the 1-page limit?

j-low:
@jseldess thanks for noting that.  What's happening there is that the index needs to be configured by to ignore the anchored links.  I tried a few options on the front end to fix that but it seems that since the index was compiled by Algolia some other tweak needs to be made.  I've been in contact with their dev as of last Thursday about resolving this particular issue.

jseldess:
Looking good, @j-low. I'm still seeing See Also texts in the results. I know you're working on excluding them. But now direct page hits have a better summary.

A nice-to-have nit: Can we bold the search term(s) in the results summaries? 

Post-2.0, it might be nice to replicate what Google does, where the link is to the page but the summary starts with a "Jump to XXX", if there's a section to jump to. 

screen shot 2018-03-21 at 11 10 41 am

j-low:
@jseldess 

re: "See Also" text: that config has been PR'd but not yet merged so this is expected.

re: highlight search terms: it looks like they provide a nice way to handle that so, yes, this can be worked in for this iteration.

re: "Jump to ... ": this would definitely be nice.  I don't have a clear idea how to do this right off the top of my head but definitely worth exploring.

jseldess:
Seeing something a bit strange now. Not sure if it was happening before. Some result summaries are only the search keyword, even though on the page the term shows up in the context of a sentence/paragraph.

screen shot 2018-03-21 at 1 22 57 pm

For example, if you click the first link, `UPSERT`, you're taking to an example: https://www.cockroachlabs.com/docs/stable/upsert.html?__hstc=181257784.62e02c0a517f58508c195cf17fdd48cc.1489812728556.1521644926081.1521652640628.522&__hssc=181257784.11.1521652640628&__hsfp=1496384207#upsert-that-fails-conflict-on-non-primay-key

The term doesn't actually appear anywhere in that section, but it does appear in the See Also just below. It also appears in a flag description further up the page. So I'm not sure what's happening here.

j-low:
@jseldess it looks like two separate issues.  I'm debugging them currently and should have some more info by later today.

For reference, here is the PR thread where I am discussing the issue with Algolia: https://github.com/algolia/docsearch-configs/pull/349#issuecomment-374977212

jseldess:
LGTM, @j-low. I think this is good to merge, given that we'll continue making improvements. Thanks for all your work!

PR #2766, opened 2018-03-22
Added pages for enabling node map

Amruta-Ranade:


v1.1/enable-node-map.md
```diff
@@ -0,0 +1,3 @@
+---
+redirect_to: ../v2.0/enable-node-map.md
```
jseldess: The redirect is failing because this needs to have the `.html` file ending, i.e., `redirect_to: ../v2.0/enable-node-map.html`.

Amruta-Ranade: Ah! Thanks!

PR #2772, opened 2018-03-23
Add python to json tutorial

jseldess:
- Add python code.
- Update tutorial to let user choose go or python.

### DIFF
--- a/_includes/json/json-sample.py
+++ b/_includes/json/json-sample.py
@@ -1,0 +1,44 @@
+import json
+import psycopg2
+import requests
+import time
+
+conn = psycopg2.connect(database="jsonb_test", user="maxroach", host="localhost", port=26257)
+conn.set_session(autocommit=True)
+cur = conn.cursor()
+
+# The Reddit API wants us to tell it where to start from. The first request
+# we just say "null" to say "from the start"; subsequent requests will use
+# the value received from the last call.
+url = "https://www.reddit.com/r/programming.json"
+after = {"after": "null"}
+
+for n in range(300):
+    # First, make a request to reddit using the appropriate "after" string.
+    req = requests.get(url, params=after, headers={"User-Agent": "Python"})
+
+    # Decode the JSON and set "after" for the next request.
+    resp = req.json()
+    after = {"after": str(resp['data']['after'])}
+
+    # Convert the JSON to a string to send to the database.
+    data = json.dumps(resp)
+
+    # The JSON reddit returns looks like this:
+    # {
+    #   "data": {
+    #     "children": [ ... ]
+    #   },
+    #   "after": ...
+    # }
+    # We structure our query so that we extract the `children` field, and then
+    # expand that and insert each individual element into the database as a
+    # separate row.
+    cur.execute("""INSERT INTO jsonb_test.programming (posts)
+            SELECT json_array_elements(%s->'data'->'children')""", (data,))
+
+    # Reddit limits to 30 requests per minute, so don't do any more than that.
+    time.sleep(2)
+
+cur.close()
+conn.close()
--- a/v2.0/demo-json-support.md
+++ b/v2.0/demo-json-support.md
@@ -10,8 +10,21 @@ toc: false
 
 ## Step 1. Install prerequisites
 
+<div class="filters filters-big clearfix">
+    <button class="filter-button" data-scope="go">Go</button>
+    <button class="filter-button" data-scope="python">Python</button>
+</div>
+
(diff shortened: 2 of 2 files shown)


justinj:
LGTM besides the `resp` thing

---

---

*[_includes/json/json-sample.py, line 36 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2772#-L8XWjNJCeNU3ekJdaWi:-L8XWjNJCeNU3ekJdaWj:b-9npx16) ([raw file](https://github.com/cockroachdb/docs/blob/9b92ec8c9a5dddb11d4e212f3a97e3e987701c75/_includes/json/json-sample.py#L36)):*
> ```Python
>     cur.execute("""INSERT INTO jsonb_test.programming (posts)
>             SELECT json_array_elements(%s->'data'->'children')
>             RETURNING %s->'data'->'after'""", (resp,resp))
> ```

doing the two `resp`s here is unfortunate, I'm starting to think we just shouldn't be fancy and just parse it on the client side as well :/ it also hadn't occurred to me that the `RETURNING` would happen for each row being inserted, which is sort of ugly and inefficient, I can PR a fix for that for the Go example (we can either parse it on the client side or use a CTE, I think).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2772#-:-L8XYj5k62vrSaR-Q4u9:b-dyxn5x)*

lhirata:
:lgtm: from a docs perspective. Thanks for adding!

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2772#-:-L8XbXcwHqXz0BP2swBK:bdcuhgh)*

jseldess:
*[_includes/json/json-sample.py, line 36 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2772#-L8XWjNJCeNU3ekJdaWi:-L8XezxoDh4ioidDdfXk:bi1lfxe) ([raw file](https://github.com/cockroachdb/docs/blob/9b92ec8c9a5dddb11d4e212f3a97e3e987701c75/_includes/json/json-sample.py#L36)):*
<details><summary><i>Previously, justinj (Justin Jaffray) wrote…</i></summary><blockquote>

doing the two `resp`s here is unfortunate, I'm starting to think we just shouldn't be fancy and just parse it on the client side as well :/ it also hadn't occurred to me that the `RETURNING` would happen for each row being inserted, which is sort of ugly and inefficient, I can PR a fix for that for the Go example (we can either parse it on the client side or use a CTE, I think).
</blockquote></details>

But doing the two `resp`s is just a python requirement. For each placeholder, you need a corresponding item. Is that in itself really a problem? 

I also agree that parsing client-side would probably be easier, but then we'd lose the demonstration of our support for functions like `json_array_elements()`. 

That said, totally your call on what we do here. If you like, I can put this on hold if you want to open a separate PR with changes to the Go sample.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2772)*

justinj:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[_includes/json/json-sample.py, line 36 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2772#-L8XWjNJCeNU3ekJdaWi:-L8c-ahC1KyytNIzZ1RV:bj4klcc) ([raw file](https://github.com/cockroachdb/docs/blob/9b92ec8c9a5dddb11d4e212f3a97e3e987701c75/_includes/json/json-sample.py#L36)):*

But doing the two `resp`s is just a python requirement. For each placeholder, you need a corresponding item. Is that in itself really a problem? 

I also agree that parsing client-side would probably be easier, but then we'd lose the demonstration of our support for functions like `json_array_elements()`. 

That said, totally your call on what we do here. If you like, I can put this on hold if you want to open a separate PR with changes to the Go sample.
</blockquote></details>

We don't necessarily have to lose the demonstration of `json_array_elements()`, it's just that if we can't send the data in a single time, we're definitely going to have to parse it again, whereas the benefit of getting back `after` from the query was that we only have to parse it once. If we'd rather leave it in just to demonstrate it's possible that's cool.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2772#-:-L8c1118ChTdN3ZNl3vO:bnfp4nl)*

jseldess:
I updated the code to get `after` on the client side. Thanks for the help, @justinj.

Issue #2522, opened 2018-02-14
cli: add sequences to `cockroach dump` output


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/21774

From release notes:
> The `cockroach dump` command now outputs `CREATE SEQUENCE` statements before the `CREATE TABLE` statements that use them.

Issue #2516, opened 2018-02-14
keys, sql: change sequence key format so backup/restore works


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/21684

From release notes:
> Sequences are now supported in enterprise `BACKUP`/`RESTORE` jobs.
> 
> This changes how sequences are stored in the key-value storage layer, so existing sequences must be dropped and recreated. Since a sequence cannot be dropped while it is being used in a column's `DEFAULT` expression, those expressions must be dropped before the sequence is dropped, and recreated after the sequence is recreated. The `setval` function can be used to set the value of a sequence to what it was previously.

lhirata:
@jseldess - i added sequences to BACKUP / RESTORE docs. do i need add a note to CREATE SEQUENCE regarding the backwards incompatibility?

jseldess:
I'm not sure, @lhirata. @vilterp, @mjibson, do you have opinions?

vilterp:
The note would have to be something like "sequences created in 2.0-alpha builds before <X alpha build>" will have to be dropped and recreated …". Seems like that would affect a very small number of users. I lean toward saying the release note above would be enough for them.

FWIW, looks like the backwards-incompatible change was merged Jan 31. Not sure how to get a sense of if anyone was using sequences before then… Maybe the reg cluster could tell us. But anyway, I'd say just leave the release note.

Issue #2510, opened 2018-02-14
sql: prevent dropping a sequence if it's being used


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/21364

From release notes:
> `DROP SEQUENCE` cannot drop a `SEQUENCE` that is currently in use.

PR #2779, opened 2018-03-26
Sequence Updates

### LINKED ISSUE #2718: sequences: correct description of what happens when sequence is renamed
As @mberhault points out in https://github.com/cockroachdb/cockroach/issues/21868, the docs don't quite match what happens when you rename a sequence that a table is using in a default expression:

- They say that the rename will work but won't be propagated to the DEFAULT expression.
- What actually happens is that you get an error telling you you can't rename the sequence.

I think that the actual behavior is better than what the docs say happens, so I'm not going to change the behavior. We should update the docs. Apologies for being wrong about this behavior when this was originally documented!

lhirata:
Closes #2718, #2510, #2516, #2522.

### DIFF
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -27,7 +27,7 @@ Dependent objects should be backed up at the same time as the objects they depen
 Object | Depends On
 -------|-----------
 Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys)).
-Table with a [sequence](create-sequence.html) | The sequence.
+Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses (however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences-new-in-v2-0).
 [Views](views.html) | The tables used in the view's `SELECT` statement.
 [Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy).
 
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -26,9 +26,10 @@ Dependent objects should be backed up at the same time as the objects they depen
 
 Object | Depends On
 -------|-----------
-Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys))
-[Views](views.html) | The tables used in the view's `SELECT` statement
-[Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy)
+Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys)).
+Table with a [sequence](create-sequence.html) | The sequence.
+[Views](views.html) | The tables used in the view's `SELECT` statement.
+[Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy).
 
 ### Users and Privileges
 
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -26,8 +26,8 @@ Dependent objects should be backed up at the same time as the objects they depen
 
 Object | Depends On
 -------|-----------
-Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys)).
-Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses (however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences-new-in-v2-0).
+Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES`; however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys).
+Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses; however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences-new-in-v2-0).
 [Views](views.html) | The tables used in the view's `SELECT` statement.
 [Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy).
 
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -27,7 +27,7 @@ Dependent objects should be backed up at the same time as the objects they depen
 Object | Depends On
 -------|-----------
 Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES`; however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys).
-Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses; however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences-new-in-v2-0).
+Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses; however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences).
 [Views](views.html) | The tables used in the view's `SELECT` statement.
 [Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy).
 
--- a/v2.0/create-sequence.md
+++ b/v2.0/create-sequence.md
@@ -43,7 +43,7 @@ table td:first-child {
 We support the following [SQL sequence functions](functions-and-operators.html):
 
 - `nextval('seq_name')`
-    {{site.data.alerts.callout_info}}If <code>nextval()</code> is used in conjunction with <code>RETURNING NOTHING</code> statements, the sequence increments can be reordered. For more information, see <a href="https://www.cockroachlabs.com/blog/how-online-schema-changes-are-possible-in-cockroachdb/">Parallel Statement Execution</a>.{{site.data.alerts.end}}
+    {{site.data.alerts.callout_info}}If <code>nextval()</code> is used in conjunction with <code>RETURNING NOTHING</code> statements, the sequence increments can be reordered. For more information, see <a href="parallel-statement-execution.html">Parallel Statement Execution</a>.{{site.data.alerts.end}}
 - `currval('seq_name')`
 - `lastval()`
 - `setval('seq_name', value, is_called)`
--- a/v2.0/create-sequence.md
+++ b/v2.0/create-sequence.md
@@ -43,6 +43,7 @@ table td:first-child {
 We support the following [SQL sequence functions](functions-and-operators.html):
 
 - `nextval('seq_name')`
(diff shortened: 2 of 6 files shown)


v2.0/rename-sequence.md
```diff
@@ -4,9 +4,9 @@ summary: The RENAME SEQUENCE statement changes the name of a sequence.
 toc: false
 ---
 
-<span class="version-tag">New in v2.0:</span> The `RENAME SEQUENCE` [statement](sql-statements.html) changes the name of a sequence.
+<span class="version-tag">New in v2.0:</span> The `RENAME TO` [statement](sql-statements.html) is part of [`ALTER SEQUENCE`](alter-sequence.html), and changes the name of a sequence.
 
-{{site.data.alerts.callout_danger}}If you rename a sequence that's being used in a table, the change will not propagate to the <code>DEFAULT</code> expressions that reference the sequence. {{site.data.alerts.end}}
+{{site.data.alerts.callout_danger}}You cannot rename a sequence that's being used in a table. To rename the sequence, drop the <code>DEFAULT</code> expressions that reference the sequence, rename the sequence, and add the <code>DEFAULT</code> expressions back.{{site.data.alerts.end}}
```
vilterp: 👍

vilterp: May want to mention or link to the syntax for doing this: `ALTER TABLE _ ALTER COLUMN _ DROP DEFAULT`

v2.0/sql-dump.md
```diff
@@ -14,8 +14,8 @@ The `cockroach dump` [command](cockroach-commands.html) outputs the SQL statemen
 
 When `cockroach dump` is executed:
 
-- Table and view schemas and table data are dumped as they appeared at the time that the command is started. Any changes after the command starts will not be included in the dump.
-- **New in v1.1:** Table and view schemas are dumped in the order in which they can successfully be recreated.
+- Table, sequence, and view schemas and table data are dumped as they appeared at the time that the command is started. Any changes after the command starts will not be included in the dump.
+- **New in v1.1:** Table, sequence, and view schemas are dumped in the order in which they can successfully be recreated.
```
vilterp: Sadly not true if you have cyclical foreign keys: https://github.com/cockroachdb/cockroach/issues/20254

Worth noting as a known limitation if it's not already.

v2.0/create-sequence.md
```diff
@@ -43,6 +43,7 @@ table td:first-child {
 We support the following [SQL sequence functions](functions-and-operators.html):
 
 - `nextval('seq_name')`
+    {{site.data.alerts.callout_info}}If <code>nextval()</code> is used in conjunction with <code>RETURNING NOTHING</code> statements, the sequence increments can be reordered. For more information, see <a href="parallel-statement-execution.html">Parallel Statement Execution</a>.{{site.data.alerts.end}}
```
vilterp: 👍

v2.0/backup.md
```diff
@@ -26,9 +26,10 @@ Dependent objects should be backed up at the same time as the objects they depen
 
 Object | Depends On
 -------|-----------
-Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys))
-[Views](views.html) | The tables used in the view's `SELECT` statement
-[Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy)
+Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys)).
+Table with a [sequence](create-sequence.html) | The sequence.
```
vilterp: There is actually an option `skip_missing_sequences` which works the same way as `skip_missing_foreign_keys`. Sorry for not calling it out better in release notes or otherwise! I think we need to add something to `restore.md` and link to it here, same as for `skip_missing_foreign_keys`.

Mimicking the section for foreign keys, the new section in `restore.md` would be something like:

> #### `skip_missing_sequences`
>
> - **Description**: If you want to restore a table that depends on a sequence but don't want to restore the sequence it references, you can drop the sequence dependency from a table (its sequence-using `DEFAULT` expression) and then have it restored.
> - **Key**: `skip_missing_sequences`
> - **Value**: *No value*
> - **Example**: `WITH skip_missing_sequences`

(Note: it's a bit odd that the `skip_missing_foreign_keys` section of that doc seems to link to its own heading; removed that)

lhirata:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v2.0/backup.md, line 30 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigr9-r3-30:-L8baJ2Kfm4GljlLY3de:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a5908f05c454295e9aac2733c44e7d2ed26c883f/v2.0/backup.md#L30)):*

There is actually an option `skip_missing_sequences` which works the same way as `skip_missing_foreign_keys`. Sorry for not calling it out better in release notes or otherwise! I think we need to add something to `restore.md` and link to it here, same as for `skip_missing_foreign_keys`.

Mimicking the section for foreign keys, the new section in `restore.md` would be something like:

> #### `skip_missing_sequences`
>
> - **Description**: If you want to restore a table that depends on a sequence but don't want to restore the sequence it references, you can drop the sequence dependency from a table (its sequence-using `DEFAULT` expression) and then have it restored.
> - **Key**: `skip_missing_sequences`
> - **Value**: *No value*
> - **Example**: `WITH skip_missing_sequences`

(Note: it's a bit odd that the `skip_missing_foreign_keys` section of that doc seems to link to its own heading; removed that)
</blockquote></details>

Done.

---

*[v2.0/rename-sequence.md, line 9 at r2](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigrB-r2-9:-L8bb2BfnAns3Rerk5WX:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7b9ebd7ed0812c104759848d816d2897c727d96a/v2.0/rename-sequence.md#L9)):*

May want to mention or link to the syntax for doing this: `ALTER TABLE _ ALTER COLUMN _ DROP DEFAULT`
</blockquote></details>

Done.

---

*[v2.0/sql-dump.md, line 18 at r3](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigrD-r3-18:-L8bcSnW2zYQ7AH9gUGd:b8yxl47) ([raw file](https://github.com/cockroachdb/docs/blob/a5908f05c454295e9aac2733c44e7d2ed26c883f/v2.0/sql-dump.md#L18)):*

Sadly not true if you have cyclical foreign keys: https://github.com/cockroachdb/cockroach/issues/20254

Worth noting as a known limitation if it's not already.
</blockquote></details>

Added clarification. Also tagged the bug ticket with "known limitation" so we can document it in the aggregated list.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2779)*

v2.0/restore.md
```diff
@@ -130,7 +131,14 @@ You can include the following options as key-value pairs in the `kv_option_list`
...
-- **Description**: If you want to restore a table with a foreign key but don't want to restore the table it references, you can [drop the Foreign Key constraint from the table](#skip_missing_foreign_keys) and then have it restored.
+- **Description**: If you want to restore a table with a foreign key but don't want to restore the table it references, you can drop the Foreign Key constraint from the table and then have it restored.
+- **Key**: `skip_missing_foreign_keys`
+- **Value**: *No value*
+- **Example**: `WITH skip_missing_foreign_keys`
+
+#### `skip_missing_sequences` <span class="version-tag">New in v2.0</span>
+
+- **Description**: If you want to restore a table that depends on a sequence but don't want to restore the sequence it references, you can drop the sequence dependency from a table (i.e., the `DEFAULT` expression that uses the sequence) and then have it restored.
 - **Key**: `skip_missing_foreign_keys`
```
jseldess: Should the key/value/example be different for sequences?

v2.0/backup.md
```diff
@@ -26,9 +26,10 @@ Dependent objects should be backed up at the same time as the objects they depen
 
 Object | Depends On
 -------|-----------
-Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys))
-[Views](views.html) | The tables used in the view's `SELECT` statement
-[Interleaved tables](interleave-in-parent.html) | The parent table in the [interleaved hierarchy](interleave-in-parent.html#interleaved-hierarchy)
+Table with [foreign key](foreign-key.html) constraints | The table it `REFERENCES` (however, this dependency can be [removed during the restore](restore.html#skip_missing_foreign_keys)).
+Table with a [sequence](create-sequence.html) | <span class="version-tag">New in v2.0:</span> The sequence it uses (however, this dependency can be [removed during the restore](restore.html#skip_missing_sequences-new-in-v2-0).
```
jseldess: nit: In this and the preceding row, could you change the parenthetical to a stand-alone sentence (or separate with a semi-colon)?

v2.0/restore.md
```diff
@@ -130,7 +131,14 @@ You can include the following options as key-value pairs in the `kv_option_list`
 
 #### `skip_missing_foreign_keys`
 
-- **Description**: If you want to restore a table with a foreign key but don't want to restore the table it references, you can [drop the Foreign Key constraint from the table](#skip_missing_foreign_keys) and then have it restored.
+- **Description**: If you want to restore a table with a foreign key but don't want to restore the table it references, you can drop the Foreign Key constraint from the table and then have it restored.
+- **Key**: `skip_missing_foreign_keys`
+- **Value**: *No value*
+- **Example**: `WITH skip_missing_foreign_keys`
+
+#### `skip_missing_sequences` <span class="version-tag">New in v2.0</span>
```
jseldess: Unfortunately, the callout doesn't render correctly for h4 and up. Would you mind filing this bug with @j-low and, for the time being, just moving this to the line below the heading?

lhirata:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[v2.0/backup.md, line 30 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigr9-r4-30:-L8cKrDIs9Ex4aSbrOHo:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e7c41ad9305e8631444f84e69c629108169eb59b/v2.0/backup.md#L30)):*

nit: In this and the preceding row, could you change the parenthetical to a stand-alone sentence (or separate with a semi-colon)?
</blockquote></details>

Done.

---

*[v2.0/restore.md, line 139 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigrC-r4-139:-L8cLNjdA-tTaThqrZZ3:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e7c41ad9305e8631444f84e69c629108169eb59b/v2.0/restore.md#L139)):*

Unfortunately, the callout doesn't render correctly for h4 and up. Would you mind filing this bug with @j-low and, for the time being, just moving this to the line below the heading? 
</blockquote></details>

Done.

---

*[v2.0/restore.md, line 142 at r4](https://reviewable.io:443/reviews/cockroachdb/docs/2779#-L8XhmSBX4VCHFRKigrC-r4-142:-L8cLRfaDf62jtPhe52m:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e7c41ad9305e8631444f84e69c629108169eb59b/v2.0/restore.md#L142)):*

Should the key/value/example be different for sequences?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cockroachdb/docs/2779)*

lhirata:
TFTRs! 🎉

PR #2795, opened 2018-03-28
Correct display message

aermolaev:


jseldess:
Thank you, @aermolaev! I'll validate this change soon.

jseldess:
LGTM. Thanks for this correction, and apologies for the delay. I'll apply this change to subsequent versions of the docs as well.

PR #2821, opened 2018-03-29
Update docs on `sql_safe_updates` var and flag

### LINKED ISSUE #2396: Changes to sql_safe_updates variable and --unsafe-updates flag
From release notes:

> - The sql_safe_updates session variable now defaults to false unless the shell is truly interactive (using cockroach sql, -e not specified, standard input not redirected) and --unsafe-updates is not specified. Previously, sql_safe_updates would always default to true unless --unsafe-updates was specified. PR: https://github.com/cockroachdb/cockroach/pull/20805
> - The --unsafe-updates of cockroach sql was renamed to --safe-updates. The default behavior is unchanged: The previous flag defaulted to false; the new flag defaults to true. PR: https://github.com/cockroachdb/cockroach/pull/20935
> - The cockroach sql command no longer fails when the server is running a version of CockroachDB that does not support the sql_safe_updates session variable. PR: https://github.com/cockroachdb/cockroach/pull/20935

rmloveland:
Fixes #2396 

Related to:

- https://github.com/cockroachdb/cockroach/pull/20805
- https://github.com/cockroachdb/cockroach/pull/20935

Specific docs changes:

- Note in several places that `--unsafe-updates` is now `--safe-updates`, with its default flipped to `true`

- Add a slight clarification on the `sql_safe_updates` session var that it will be `false` if the flag is passed on client startup


v2.0/set-vars.md
```diff
@@ -38,7 +38,7 @@ The variable name is case insensitive. The value can be a list of one or more it
...
-| `sql_safe_updates` | If `false`, allow potentially unsafe SQL statements, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for interactive sessions from the [built-in SQL client](use-the-built-in-sql-client.html),<br>`false` for sessions from other clients | Yes |
+| `sql_safe_updates` | If `false`, allow potentially unsafe SQL statements, including `DROP` of a non-empty database and all dependent objects, `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE .. DROP COLUMN`. See [Allow Potentially Unsafe SQL Statements](use-the-built-in-sql-client.html#allow-potentially-unsafe-sql-statements) for more details. | `true` for interactive sessions from the [built-in SQL client](use-the-built-in-sql-client.html) unless `--safe-updates=false` is specified,<br>`false` for sessions from other clients | Yes |
```
knz: I think the first sentence of the description should also be updated to match the description in the other places, that is, "If `true`, disallow potentially unsafe SQL statements, including `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE ... DROP COLUMN`."

rmloveland: Thanks @knz, updated with that language.

knz:
LGTM

PR #2825, opened 2018-03-30
Copy v1.0 known join flag limitation over to v1.1 known limitations

a-robinson:
I'd guess this got dropped from the known limitations because the --join
flags in our examples changed to include more than just the first node,
but the underlying issue still exists in clusters large enough that it
isn't guaranteed that a copy of the first range will be on nodes
included in the join flags.

It's been fixed for 2.0, but we should still include this in our 1.1
docs. I've tweaked the wording a bit to make more sense than the v1.0
wording verbatim would have.

### DIFF
--- a/v1.1/known-limitations.md
+++ b/v1.1/known-limitations.md
@@ -74,6 +74,12 @@ Also, a prepared [`INSERT`](insert.html), [`UPSERT`](upsert.html), or [`DELETE`]
 - If the number of columns has increased, the prepared statement returns an error but nonetheless writes the data.
 - If the number of columns remains the same but the types have changed, the prepared statement writes the data and does not return an error.
 
+## Join flags when restarting a cluster with different addresses
+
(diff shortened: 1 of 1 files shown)


v1.1/known-limitations.md
```diff
@@ -74,6 +74,12 @@ Also, a prepared [`INSERT`](insert.html), [`UPSERT`](upsert.html), or [`DELETE`]
...
+
+However, if the nodes in a cluster are restarted with different addresses for some reason, then it's not guaranteed that a copy of the first range will be on the nodes in the join flags. In such cases, the `--join` flags must form a fully-connected directed graph. The easiest way to do this is to put all of the new nodes' addresses into each node's `--join` flag, which ensures all nodes can join a node with a copy of the first key-value range.
```
bdarnell: A fully-connected graph is not strictly required (connecting each node to nodes 1..N-1 leaves the last node with no inbound edges). I'd just drop that sentence and leave the recommendation to put all the addresses in all the join flags.

a-robinson: If you've reduced the replication factor of the first range to 1 and its only copy is on the last node, then that configuration won't work. I do think it has to be fully-connected to guarantee success in cases like that. I can remove the wording, but it's also how we documented it in the 1.0 docs. Should I remove it there too?

bdarnell: I'm fine leaving it as-is.

jseldess:
Thanks, @a-robinson!

PR #2846, opened 2018-04-02
Add CACHE / NO CYCLE to CREATE SEQUENCE

### LINKED ISSUE #2793: NO CYCLE and CACHE 1 are now supported options during sequence creation
PR: https://github.com/cockroachdb/cockroach/pull/23518

lhirata:
Closes #2793.

@nvanbenschoten, I think `CACHE <seq_value>` need to be added to the `create_sequence` SQL diagram.

### DIFF
--- a/v2.0/create-sequence.md
+++ b/v2.0/create-sequence.md
@@ -36,9 +36,10 @@ table td:first-child {
 `MINVALUE` | The minimum value of the sequence. Default values apply if not specified or if you enter `NO MINVALUE`.<br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `MININT`
 `MAXVALUE` | The maximum value of the sequence. Default values apply if not specified or if you enter `NO MAXVALUE`.<br><br>**Default for ascending:** `MAXINT` <br><br>**Default for descending:** `-1`
 `START` | The first value of the sequence. <br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `-1`
-`CACHE <seq_value>` | The number of sequence values to allocate in memory for faster access. Currently, you can only `CACHE 1` or leave unspecified.
 `NO CYCLE` | Currently, all sequences are set to `NO CYCLE` and the sequence will not wrap.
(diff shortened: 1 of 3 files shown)


v2.0/create-sequence.md
```diff
@@ -36,7 +36,9 @@ table td:first-child {
 `MINVALUE` | The minimum value of the sequence. Default values apply if not specified or if you enter `NO MINVALUE`.<br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `MININT`
 `MAXVALUE` | The maximum value of the sequence. Default values apply if not specified or if you enter `NO MAXVALUE`.<br><br>**Default for ascending:** `MAXINT` <br><br>**Default for descending:** `-1`
 `START` | The first value of the sequence. <br><br>**Default for ascending:** `1` <br><br>**Default for descending:** `-1`
-`CYCLE` / `NO CYCLE` | Not yet implemented. The sequence will wrap around when the sequence value hits the maximum or minimum value. Currently, all sequences are set to `NO CYCLE` and the sequence will not wrap.
+`CACHE <seq_value>` | The number of sequence values to allocate in memory for faster access.
```
nvanbenschoten: We currently will throw an error if this is specified and it is not 1.

lhirata: Updated to clarify

lhirata:
TFTR @nvanbenschoten! I noticed that `CACHE` doesn't show up in the [full sql grammar](https://www.cockroachlabs.com/docs/dev/sql-grammar.html#sequence_option_elem). Is that something I can add or do you need to?

nvanbenschoten:
I think that was on purpose and now that you bring that up, I don't actually think we want to document the `CACHE` option at all. We're exposing only the bare minimum support for it so that we don't reject ORMs that already use the option, but I don't think we need to tell users that it's available.

lhirata:
Ok, thanks for the info. @jseldess, should I just close this PR?

nvanbenschoten:
The `NO CYCLE` option is still valuable to document.

lhirata:
Ah yeah, that's already documented. This PR just cleaned it up a little. I'll just remove the `CACHE` entry.

PR #2866, opened 2018-04-04
Edit 'Supporting legacy code' example/description

### LINKED ISSUE #2555: docs: simplify-supporting-legacy-code example doesn't work
https://www.cockroachlabs.com/docs/stable/views.html#simplify-supporting-legacy-code

I suppose you can get rid of the transaction

rmloveland:
Specifically:

- Rewrite the code sample to drop the transaction
- Rewrite the surrounding text to note that there may be a brief window after the rename but before the view creation where your application may need to do some error handling (which in theory it is already doing when accessing the DB)
- Link to the relevant section of the "known limitations" page

Fixes #2555

### DIFF
--- a/v1.0/views.md
+++ b/v1.0/views.md
@@ -163,22 +163,31 @@ pq: user bob does not have SELECT privilege on table accounts
 
 ### Simplify supporting legacy code
 
-When you need to alter a table in a way that will break your application, and you can't update your code right away, you can create a view with a name and schema identical to the original table. During and after changes to the original table, your application will continue uninterrupted, and you can update the legacy code at your leisure.
+When you need to alter a table in a way that will break your application, and you can't update your code right away, you can create a view with a name and schema identical to the original table. Once the new view is created, your application can continue uninterrupted, and you can update the legacy code at your leisure.
 
 #### Example
 
-Let's say you have a table called `user_accounts` that you want to rename to `client_accounts`, but your application has many queries to `user_accounts`. To ensure that your application continues uninterrupted, you could execute a [transaction](transactions.html) that renames the underlying table to `client_accounts` and creates a view with the old table name, `user_accounts`:
+Let's say you have a table called `user_accounts` that you want to rename to `client_accounts`, but your application has many queries to `user_accounts`.
+
+You can accomplish the change with a 2-step process:
+
+1. [Rename the underlying table](rename-table.html) to `client_accounts`
+2. [Create a view](create-view.html) with the old table name, `user_accounts`
+
+For example:
 
 ~~~ sql
-BEGIN;
 ALTER TABLE bank.user_accounts RENAME TO bank.client_accounts;
 CREATE VIEW bank.user_accounts
   AS SELECT type, email
   FROM bank.client_accounts;
-COMMIT;
 ~~~
 
-Your application would then continue referencing `user_accounts` without issue, and at a later time, you could update your application code to reference the new `client_accounts` table.
+Note that there will be a brief window of time after the table rename and before the view creation when there will be no relation (table or view) called `user_accounts`. During this window, if your application tries to access `user_accounts`, it will need to handle any error that occurs. In most ORMs, this will cause an exception to be thrown, followed by a retry. Once the view is created, the retry should succeed, and further accesses of `user_accounts` will work as expected.
+
+{{site.data.alerts.callout_info}}
+Unfortunately the 2-step operation described above can't be wrapped in a [transaction](transactions.html) due to a [known limitation](known-limitations.html#schema-changes-within-transactions) that a table cannot be renamed inside a transaction.
+{{site.data.alerts.end}}
 
 ## How Views Work
 
--- a/v1.1/views.md
+++ b/v1.1/views.md
@@ -163,22 +163,31 @@ pq: user bob does not have SELECT privilege on table accounts
 
 ### Simplify supporting legacy code
 
-When you need to alter a table in a way that will break your application, and you can't update your code right away, you can create a view with a name and schema identical to the original table. During and after changes to the original table, your application will continue uninterrupted, and you can update the legacy code at your leisure.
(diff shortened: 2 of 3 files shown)


rmloveland:
@andreimatei would appreciate your review, esp. around the "your ORM should retry after an exception" part, which I think is how the world *should* work but ... IDK :-)

Also related to the above, do you think it is reasonable to provide more information about how long one should expect the gap to be between the table rename and the subsequent view creation?  Right now it says there "will be a brief window of time", which may be too vague.

andreimatei:
> Also related to the above, do you think it is reasonable to provide more information about how long one should expect the gap to be between the table rename and the subsequent view creation?

The gap should generally be on the order of the latency of the slowest network link between two nodes in the cluster, but can be minutes if a node had crashed.
I think the only thing we can say is that, once the `CREATE VIEW` succeeds, further queries should work.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[v2.0/views.md, line 164 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LRR83CfqiEYbazqlp:-L9LRR83CfqiEYbazqlq:b3pg7b7) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L164)):*
> ```Markdown
> ~~~
> 
> ### Simplify supporting legacy code
> ```

My preference would be to get rid of this section altogether. I don't think it's particularly valuable.
For anything other than a rename of the original table, queries through the view will likely be too slow to claim that the app just works anyway. 

---

*[v2.0/views.md, line 186 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LSu203SL-KOF_qigH:-L9LSu203SL-KOF_qigI:b-j1k4op) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L186)):*
> ```Markdown
> ~~~
> 
> Note that there will be a brief window of time after the table rename and before the view creation when there will be no relation (table or view) called `user_accounts`. During this window, if your application tries to access `user_accounts`, it will need to handle any error that occurs. In most ORMs, this will cause an exception to be thrown, followed by a retry. Once the view is created, the retry should succeed, and further accesses of `user_accounts` will work as expected.
> ```

do we use the word "relation" anywhere else in our docs?

---

*[v2.0/views.md, line 186 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LT4Zf2IW39ua-7b1g:-L9LT4Zf2IW39ua-7b1h:b7ph9ii) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L186)):*
> ```Markdown
> ~~~
> 
> Note that there will be a brief window of time after the table rename and before the view creation when there will be no relation (table or view) called `user_accounts`. During this window, if your application tries to access `user_accounts`, it will need to handle any error that occurs. In most ORMs, this will cause an exception to be thrown, followed by a retry. Once the view is created, the retry should succeed, and further accesses of `user_accounts` will work as expected.
> ```

I think "followed by a retry" is unclear. No ORM will just retry on such an error. The client needs to retry... Which is unlikely - it's weird to suggest to do retries on schema errors.
I wouldn't say anything about retries.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-:-L9LTaPr7Dfxpuh28mQI:brn7l7r)*

jseldess:
*[v2.0/views.md, line 164 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LRR83CfqiEYbazqlp:-L9LV68h9pBuVwGxjyMI:b-b67a0s) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L164)):*

My preference would be to get rid of this section altogether. I don't think it's particularly valuable.
For anything other than a rename of the original table, queries through the view will likely be too slow to claim that the app just works anyway. 
</blockquote></details>

This is good input, @andreimatei. I stole this use case from other dbs' docs on views and assumed it was valuable, given that the alternative is interrupting the app for the duration of the schema change. But I defer to you on this. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2866)*

andreimatei:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[v2.0/views.md, line 164 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LRR83CfqiEYbazqlp:-L9LVgbjDA_rOcKH8KK2:b2772li) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L164)):*

This is good input, @andreimatei. I stole this use case from other dbs' docs on views and assumed it was valuable, given that the alternative is interrupting the app for the duration of the schema change. But I defer to you on this. 
</blockquote></details>

if we had support for atomic renames + create view, I'd be cool to showcase it. But since we don't, I think this section is adding little.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2866)*

jseldess:
*[v2.0/views.md, line 164 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2866#-L9LRR83CfqiEYbazqlp:-L9L_uUt7a_5W4xoYrNF:b5y0igx) ([raw file](https://github.com/cockroachdb/docs/blob/eb5e9b0a9f2ba972651782d383d21dfcc74dcd57/v2.0/views.md#L164)):*

if we had support for atomic renames + create view, I'd be cool to showcase it. But since we don't, I think this section is adding little.
</blockquote></details>

Cool. Let's just remove then, @rmloveland. Sorry for causing you unnecessary effort here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2866)*

PR #2871, opened 2018-04-05
Add a new yellow "important" callout

rmloveland:
This is intended to slot in as follows:

| Type      | Color  | Meaning                                      |
|-----------+--------+----------------------------------------------|
| Tip       | Green  | "hey check out this cool thing"              |
| Info      | Blue   | "FYI there is also this other consideration" |
| Important | Yellow | "This is approaching the danger zone, maybe" |
| Warning   | Red    | "DANGER ZONE"                                |

### DIFF
--- a/_data/alerts.yml
+++ b/_data/alerts.yml
@@ -6,6 +6,7 @@ end: '</div>'
 
 callout_danger: '<div class="bs-callout bs-callout--danger"><div class="bs-callout__label">Warning:</div>'
 callout_success: '<div class="bs-callout bs-callout--success"><div class="bs-callout__label">Tip:</div>'
+callout_important: '<div class="bs-callout bs-callout--warning"><div class="bs-callout__label">Note:</div>'
 callout_info: '<div class="bs-callout bs-callout--info"><div class="bs-callout__label">Note:</div>'
 
 hr_faded: '<hr class="faded"/>'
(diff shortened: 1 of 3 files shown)


rmloveland:
Did this as part of working on #2555 but we ended up deleting stuff in that instance, not adding.

Sorry if bikeshed-ish, figured this might come in handy for other work.

There are a few details to hammer out, e.g., do we want a different word than "Note" in the box?  Etc.

jseldess:
@rmloveland, thanks for taking the initiative here. I'm personally ambivalent about adding another callout type. We had the option initially, but the difference between tip / note / warning is much clearer than the difference between note / important. Though I do think we can make a case for it. 

Let's open this up to @lhirata and @Amruta-Ranade as well. 

Btw, if we add this, we should also update [this part of `CONTRIBUTING.md`](https://github.com/cockroachdb/docs/blob/master/CONTRIBUTING.md#notes-warnings--tips).

rmloveland:
Makes sense, I'll refrain from arguing more about this particular bikeshed paint color until until others on the team get a chance to comment. :-}

BTW here's what it would look like (save someone spinning up Jekyll on a branch):

screenshot 2018-04-05 13 25 58

lhirata:
My initial reaction is that it seems redundant to have "Information" and "Important" callouts. I would assume that "Information" could be extrapolated to mean "Important Information". I would think that if we use a callout rather than just include it as normal text, we're deeming it important / a gotcha / hey-you-should-consider-this. Similarly, the difference between yellow zone and red zone seems subjective. Do you have an example of what would warrant red vs yellow vs blue?

My other thought is that too many colors can be distracting and could diminish the effectiveness.

jseldess:
@rmloveland, for now, can we hold off on adding a fourth callout type and just use the info callout for these cases? During @lhirata style guide/conventions review, perhaps she can dig in a bit more and see if a fourth level is warranted.

rmloveland:
> for now, can we hold off on adding a fourth callout type

No problem!

PR #2872, opened 2018-04-05
Clarify provided geo-coordinates are AWS-only, make example match them

a-robinson:
We should follow-up with coordinates for the GCE and Azure regions, but
these are not correct for anything other than AWS - for example, AWS's
us-east-1 is in Northern Virginia, while GCE's us-east1 is in South Carolina.

### DIFF
--- a/_includes/azure-locations.md
+++ b/_includes/azure-locations.md
@@ -1,0 +1,30 @@
+|  Location | SQL Statement |
+|  -------- | ------------- |
+|  eastasia (East Asia) | `INSERT into system.locations VALUES ('region', 'eastasia', 22.267, 114.188)` |
+|  southeastasia (Southeast Asia) | `INSERT into system.locations VALUES ('region', 'southeastasia', 1.283, 103.833)` |
+|  centralus (Central US) | `INSERT into system.locations VALUES ('region', 'centralus', 41.5908, -93.6208)` |
+|  eastus (East US) | `INSERT into system.locations VALUES ('region', 'eastus', 37.3719, -79.8164)` |
+|  eastus2 (East US 2) | `INSERT into system.locations VALUES ('region', 'eastus2', 36.6681, -78.3889)` |
+|  westus (West US) | `INSERT into system.locations VALUES ('region', 'westus', 37.783, -122.417)` |
+|  northcentralus (North Central US) | `INSERT into system.locations VALUES ('region', 'northcentralus', 41.8819, -87.6278)` |
+|  southcentralus (South Central US) | `INSERT into system.locations VALUES ('region', 'southcentralus', 29.4167, -98.5)` |
+|  northeurope (North Europe) | `INSERT into system.locations VALUES ('region', 'northeurope', 53.3478, -6.2597)` |
+|  westeurope (West Europe) | `INSERT into system.locations VALUES ('region', 'westeurope', 52.3667, 4.9)` |
+|  japanwest (Japan West) | `INSERT into system.locations VALUES ('region', 'japanwest', 34.6939, 135.5022)` |
+|  japaneast (Japan East) | `INSERT into system.locations VALUES ('region', 'japaneast', 35.68, 139.77)` |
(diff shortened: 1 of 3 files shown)


a-robinson:
Rebased and fixed merge conflict.

jseldess:
@a-robinson, did you see @Amruta-Ranade's [comment](https://github.com/cockroachdb/docs/issues/2873#issuecomment-379025587)? Seems we need to resolve that before moving forward here.

v2.0/enable-node-map.md
```diff
@@ -195,11 +195,12 @@ The **Node Map** is displayed only for the locality levels that have latitude/lo
 
 ## Location Coordinates for Reference
```
jseldess: To be explicit, perhaps we should update this to `Location Coordinates for AWS and Azure`.

v2.0/enable-node-map.md
```diff
@@ -214,3 +215,35 @@ The **Node Map** is displayed only for the locality levels that have latitude/lo
 |  Asia Pacific (Sydney) | `INSERT into system.locations VALUES ('region', 'ap-southeast-2', -33.86882, 151.209296)` |
 |  Asia Pacific (Mumbai) | `INSERT into system.locations VALUES ('region', 'ap-south-1', 19.075984, 72.877656)` |
 |  South America (São Paulo) | `INSERT into system.locations VALUES ('region', 'sa-east-1', -23.55052, -46.633309)` |
+
+### Azure locations
```
jseldess: @a-robinson, this is awesome, but to make it easy to update this, can you please:

1. Make this azure table an include.
2. Add the python script to the `scripts/` directory. 

It'd be even better for the script to edit the relevant include instead of printing the table, but that's just a nice to have. Eventually, we could just wrap this into our TC builds.

jseldess:
To support my request that we add GCE coordinates, we got this docs feedback submission from Tierion just last last night:

Submitted April 5, 2018 6:45 PM (0 days ago)

Improvements
Something's broken (links, etc.)
What's broken?
The text has a link for "the latitudes and longitudes of common AWS/Azure/GC regions, see Locations Coordinates for Reference."
But the table at the bottom of the page has only locations for AWS. Values for Google Cloud would be helpful.
Email
xxx@tierion.com

cc @Amruta-Ranade

v2.0/enable-node-map.md
jseldess: @a-robinson, if you would prefer that I create the script separately, this PR LGTM.

v2.0/enable-node-map.md
a-robinson: Probably, although it sounds like @Amruta-Ranade would be changing it back shortly, so I'll leave it as is to avoid changing the URL location back and forth.

v2.0/enable-node-map.md
a-robinson: Done, in the sense that it's all runnable as a single command. I'm not sure whether more would be needed to fit it into a TC build.

jseldess: Thank you!

v2.0/enable-node-map.md
jseldess: SGTM.

_includes/azure-locations.md
```diff
@@ -0,0 +1,29 @@
+|  Location | SQL Statement |
+|  eastasia (East Asia) | `INSERT into system.locations VALUES ('region', 'eastasia', 22.267, 114.188)` |
```
jseldess: The table is broken without a separation between the header row and the first normal row:

```
|  Location | SQL Statement |
|----------|-----------------|
|  eastasia (East Asia) | `INSERT into system.locations VALUES ('region', 'eastasia', 22.267, 114.188)` |
```

Can you put that in place please?

a-robinson: Great catch! I'm sorry for missing that. Done.

jseldess: Thanks, Alex! This PR LGTM now. @Amruta-Ranade, once you're back, please add the GCE coordinates as well. I've assign this issue to you: https://github.com/cockroachdb/docs/issues/2885.

PR #2896, opened 2018-04-10
Performance tuning guide for running on Kubernetes

### LINKED ISSUE #2265: Document Kubernetes performance gotchas
I did a performance comparison last night between our default `StatefulSet` Kubernetes config file and an alternative [`DaemonSet` config](https://gist.github.com/a-robinson/ec2b86783ccbf053c83ba83170673d63) on a 3-node v1.7.8 GKE cluster of n1-standard-1 VMs. The performance difference was massive:

```
When running against daemonset with `hostNetwork` and `hostDir` (using the VMs' 100GB pd-standard boot disks):
_elapsed___errors____________ops___ops/sec(cum)__p95(ms)__p99(ms)_pMax(ms)_____seq(begin/end)
   78.8s        0         146093         1853.3    130.0    176.2    486.5          0 / 73356

When running against statefulset (using a 1GB pd-standard, which our config defaults to):
_elapsed___errors____________ops___ops/sec(cum)__p95(ms)__p99(ms)_pMax(ms)_____seq(begin/end)
   52.0s        0          18023          346.5    520.1    704.6   1208.0           0 / 9071

When running against statefulset (using a 100GB pd-standard to be more fair, since GCE disk size affects IOPS):
_elapsed___errors____________ops___ops/sec(cum)__p95(ms)__p99(ms)_pMax(ms)_____seq(begin/end)
   84.4s        0          30822          365.1    570.4    738.2   2147.5          0 / 15426
```

I suspect most of the difference is network-related -- we already know that our RPC layer can be a bottleneck, and going through the Docker bridge is not cheap. I would have expected it to be more like a 2x difference, though, not a 5x difference. It's also possible that the way the `StatefulSet`'s disk gets mounted is a little different from the machine's boot disk, but they were all using the same kind of disk (persistent HDDs).

If any Kubernetes customers are particularly worried about performance, I can try to confirm the hypothesis. Just let me know and I'll do so, otherwise I'll probably let this sit for a while. cc @robert-s-lee

a-robinson:
I'm sure it could be better written or easier to understand, so feedback or suggestions would be great.

Other questions:

* Should we provide more numbers about how much each change helps on a basic workload like `kv`?
* Should we be more prescriptive about which set of changes are most recommended?
* Should we provide any more material, such as a preconstructed YAML file with easy-to-replace template parameters showing where a user's customized numbers should go?

Fixes #2265

### DIFF
--- a/_includes/orchestration/kubernetes-performance.md
+++ b/_includes/orchestration/kubernetes-performance.md
@@ -1,0 +1,546 @@
+Kubernetes provides many useful abstractions for deploying and operating distributed systems, but some of the abstractions come with a performance overhead and an increase in underlying system complexity. This page explains potential bottlenecks to be aware of when [running CockroachDB in Kubernetes](orchestrate-cockroachdb-with-kubernetes.html) and shows you how to optimize your deployment for better performance.
+
+<div id="toc"></div>
+
+## Prerequisites
+
+Before you focus on optimizing a Kubernetes-orchestrated CockroachDB cluster:
+
+1. Go through the documentation for [running a CockroachDB cluster on Kubernetes](orchestrate-cockroachdb-with-kubernetes.html) to familiarize yourself with the necessary Kubernetes terminology and deployment abstractions.
+2. Verify that CockroachDB performs up to your requirements for your workload on identical hardware without Kubernetes. You may find that you need to [modify your workload](performance-best-practices-overview.html) or use [different machine specs](recommended-production-settings.html#hardware) to achieve the performance you need, and it's better to determine that up front than after spending a bunch of time trying to optimize your Kubernetes deployment.
+
+## Performance Factors
+
+There are a number of independent factors that affect the performance you observe when running CockroachDB on Kubernetes. Some are more significant than others or easier to fix than others, so feel free to pick and choose the improvements that best fit your situation. Note that most of these changes are easiest to make before you create your CockroachDB cluster. If you already have a running CockroachDB cluster in Kubernetes that you need to modify while keeping it running, extra work may be needed and extra care and testing is strongly recommended.
+
+In a number of the sections below, we have shown how to modify excerpts from our provided Kubernetes configuration YAML files. You can find the most up-to-date versions of these files on Github, [one for running CockroachDB in secure mode](https://github.com/cockroachdb/cockroach/blob/master/cloud/kubernetes/cockroachdb-statefulset-secure.yaml) and one for [running CockroachDB in insecure mode](https://github.com/cockroachdb/cockroach/blob/master/cloud/kubernetes/cockroachdb-statefulset.yaml).
+
+### Version of CockroachDB
+
+Because CockroachDB is under very active development, there are typically substantial performance gains in each release. If you aren't running the latest release and aren't getting the performance you desire, you should try the latest and see how much it helps.
+
+### Client workload
+
+Your workload is the single most important factor in database performance. Read through our [SQL performance best practices](performance-best-practices-overview.html) to determine whether there are any easy changes that you can make to speed up your application.
+
+### Machine Size
+
+The size of the machines you're using isn't a Kubernetes-specific concern, but it's always a good place to start if you want more performance. See our [hardware recommendations](recommended-production-settings.html#hardware) for specific suggestions, but using machines with more CPU will almost always allow for greater throughput. Be aware that because Kubernetes runs a set of processes on every machine in a cluster, you typically will get more bang for your buck by using fewer large machines than more small machines.
+
+### Disk Type
+
+CockroachDB makes heavy use of the disks you provide it, so using faster disks is an easy way to improve your cluster's performance. Our provided configuration does not specify what type of disks it wants, so in most environments Kubernetes will auto-provision disks of the default type. In the common cloud environments (AWS, GCP, Azure) this means you'll get slow disks that aren't optimized for database workloads (e.g. HDDs on GCE, SSDs without provisioned IOPS on AWS). However, we [strongly recommend using SSDs](recommended-production-settings.html#hardware) for the best performance, and Kubernetes makes it relatively easy to use them.
+
+#### Creating a different disk type
+
+Kubernetes exposes the disk types used by its volume provisioner via its [`StorageClass` API object](https://kubernetes.io/docs/concepts/storage/storage-classes/). Each cloud environment has its own default `StorageClass`, but you can easily change the default or create a new named class which you can then ask for when asking for volumes. To do this, pick the type of volume provisioner you want to use from the list in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/storage-classes/), take the example YAML file they provide, modify it to have the disk type you want, then run `kubectl create -f <your-storage-class-file.yaml>`. For example, in order to use the `pd-ssd` disk type on Google Compute Engine or Google Kubernetes Engine, you can use a `StorageClass` file like this:
+
+~~~ yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: <your-ssd-class-name>
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-ssd
+~~~
+
+You can then use this new disk type either by configuring the CockroachDB YAML file to request it or by making it the default. You may also want to set additional parameters as documented in the list of Kubernetes storage classes, such as configuring the `iopsPerGB` if you're creating a `StorageClass` for AWS's `io1` Provisioned IOPS volume type.
+
+#### Configuring the disk type used by CockroachDB
+
+To use a new `StorageClass` without making it the default in your cluster, you have to modify your application's YAML file to ask for it. In the CockroachDB `StatefulSet` configuration, that means adding a line to its `VolumeClaimTemplates` section. For example, that would mean taking these lines of the CockroachDB config file:
+
+~~~ yaml
+  volumeClaimTemplates:
+  - metadata:
+      name: datadir
+    spec:
+      accessModes:
+        - "ReadWriteOnce"
+      resources:
+        requests:
+          storage: 1Gi
+~~~
+
+And adding a `storageClassName` field to the `spec`, changing them to:
+
+~~~ yaml
+  volumeClaimTemplates:
+  - metadata:
+      name: datadir
+    spec:
+      accessModes:
+        - "ReadWriteOnce"
+      storageClassName: <your-ssd-class-name>
+      resources:
+        requests:
+          storage: 1Gi
+~~~
+
+If you make this change then run `kubectl create -f` on your YAML file, Kubernetes should create volumes for you using your new `StorageClass`.
+
+#### Changing the default disk type
+
+If you want your new `StorageClass` to be the default for all volumes in your cluster, you have to run a couple of commands to inform Kubernetes of what you want. First, get the names of your `StorageClass`es. Then remove the current default and add yours as the new default.
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl get storageclasses
+~~~
+
+~~~
+NAME                 PROVISIONER
+ssd                  kubernetes.io/gce-pd
+standard (default)   kubernetes.io/gce-pd
+~~~
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl patch storageclass standard -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
+~~~
+
+~~~
+storageclass "standard" patched
+~~~
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl patch storageclass ssd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
+~~~
+
+~~~
+storageclass "ssd" patched
+~~~
+
+Note that if you are running an older version of Kubernetes, you may need to use a beta version of the annotation instead of the form used above. In particular, on v1.8 of Kubernetes you need to use `storageclass.beta.kubernetes.io/is-default-class`. To determine for sure which to use, run `kubectl describe storageclass` and copy the annotation used by the current default.
+
+### Disk Size
+
+On some cloud providers (notably including all GCP disks and the AWS io1 disk type), the number of IOPS available to a disk is directly correlated to the size of the disk. In such cases, increasing the size of your disks can make for significantly better CockroachDB performance, as well as less risk of filling them up. Doing so is easy -- before you create your CockroachDB cluster, modify the `VolumeClaimTemplate` in the CockroachDB YAML file to ask for more space. For example, to give each CockroachDB instance 1TB of disk space, you'd change:
+
+~~~ yaml
+  volumeClaimTemplates:
+  - metadata:
+      name: datadir
+    spec:
+      accessModes:
+        - "ReadWriteOnce"
+      resources:
+        requests:
+          storage: 1Gi
+~~~
+
+To instead be:
+
+~~~ yaml
+  volumeClaimTemplates:
+  - metadata:
+      name: datadir
+    spec:
+      accessModes:
+        - "ReadWriteOnce"
+      resources:
+        requests:
+          storage: 1024Gi
+~~~
+
+Since [GCE disk IOPS scale linearly with disk size](https://cloud.google.com/compute/docs/disks/performance#type_comparison), a 1TiB disk gives 1024 times as many IOPS as a 1GiB disk, which can make a very large difference for write-heavy workloads.
+
+### Local Disks
+
+Up to this point, we have been assuming that you will be running CockroachDB in a `StatefulSet`, using auto-provisioned remotely attached disks. However, using local disks typically provides better performance than remotely attached disks, such as SSD Instance Store Volumes instead of EBS Volumes on AWS or Local SSDs instead of Persistent Disks on GCE. `StatefulSet`s have historically not supported using local disks, but [beta support for using "local" `PersistentVolume`s was added in Kubernetes v1.10](https://kubernetes.io/docs/concepts/storage/volumes/#local). We don't recommend using this for production data until the feature is more mature, but it's a promising development.
+
+There is also the option of using local disks if you don't run CockroachDB in a `StatefulSet`, but instead use a `DaemonSet`. For more details on what this entails, see the section on [Running in a DaemonSet](#running-in-a-daemonset).
+
+Note that when running with local disks, there is a greater chance of experiencing a disk failure than when using the cloud providers' network-attached disks that are often replicated underneath the covers. Consequently, you may want to [configure replication zones](configure-replication-zones.html) to increase the replication factor of your data to 5 from its default of 3 when using local disks.
+
+### Resource Requests and Limits
+
+When you ask Kubernetes to run a pod, either directly or indirectly through another resource type such as a `StatefulSet`, you can tell it to reserve certain amounts of CPU and/or memory for each container in the pod or to limit the CPU and/or memory of each container. Doing one or both of these can have different implications depending on how utilized your Kubernetes cluster is. For the authoritative information on this topic, see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/).
+
+#### Resource Requests
+
+Resource requests allow you to reserve a certain amount of CPU or memory for your container. If you add resource requests to your CockroachDB YAML file, Kubernetes will schedule each CockroachDB pod onto a node with sufficient unreserved resources and will ensure the pods are guaranteed the reserved resources using the applicable Linux container primitives. If you are running other workloads in your Kubernetes cluster, setting resource requests is very strongly recommended to ensure good performance, because if you don't set them then CockroachDB could be starved of CPU cycles or OOM killed before less important processes.
+
+To determine how many resources are usable on your Kubernetes nodes, you can run:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl describe nodes
+~~~
+
+~~~
+Name:               gke-perf-default-pool-aafee20c-k4t8
+[...]
+Capacity:
+ cpu:     4
+ memory:  15393536Ki
+ pods:    110
+Allocatable:
+ cpu:     3920m
+ memory:  12694272Ki
+ pods:    110
+[...]
+Non-terminated Pods:         (2 in total)
+  Namespace                  Name                                              CPU Requests  CPU Limits  Memory Requests  Memory Limits
+  ---------                  ----                                              ------------  ----------  ---------------  -------------
+  kube-system                kube-dns-778977457c-kqtlr                         260m (6%)     0 (0%)      110Mi (0%)       170Mi (1%)
+  kube-system                kube-proxy-gke-perf-default-pool-aafee20c-k4t8    100m (2%)     0 (0%)      0 (0%)           0 (0%)
+Allocated resources:
+  (Total limits may be over 100 percent, i.e., overcommitted.)
+  CPU Requests  CPU Limits  Memory Requests  Memory Limits
+  ------------  ----------  ---------------  -------------
+  360m (9%)     0 (0%)      110Mi (0%)       170Mi (1%)
+~~~
+
+This will output a lot of information for each of the nodes in your cluster, but if you focus in on the right parts you'll see how many "allocatable" resources are available on each node and how many resources are already being used by other pods. The "allocatable" resources are how much CPU and memory Kubernetes is willing to provide to pods running on the machine. The difference between the node's "capacity" and its "allocatable" resources is taken up by the operating system and Kubernetes' management processes. The "m" in "3920m" stands for "milli-CPUs", meaning "thousandths of a CPU".
+
+You'll also see a number of pods running here that you may not have realized were in your cluster. Kubernetes runs a handful of pods in the `kube-system` namespace that are part of the cluster infrastructure. These may make it tough to attempt to reserve all the allocatable space on your nodes for CockroachDB, since some of them are essential for the Kubernetes cluster's health. If you want to run CockroachDB on every node in your cluster, you'll have to leave room for these processes. If you are only running CockroachDB on a subset of the nodes in your cluster, you can choose to take up all the "allocatable" space other than what is being used by the `kube-system` pods that are on all the nodes in the cluster, such as `kube-proxy` or the `fluentd` logging agent.
+
+Note that it will be difficult to truly use up all of the allocatable space in the current versions of Kubernetes (v1.10 or older) because you'd have to manually preempt the `kube-system` pods that are already on the nodes you want CockroachDB to run on (by deleting them). This should become easier in future versions of Kubernetes when its [Pod Priority](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) feature gets promoted from alpha to beta. Once that feature is more widely available, you could set the CockroachDB pods to a higher priority, causing the Kubernetes scheduler to preempt and reschedule the `kube-system` pods onto other machines.
+
+Once you've picked out an amount of CPU and memory to reserve for Cockroach, you'll have to configure the resource request in your CockroachDB YAML file. They should go underneath the `containers` heading. For example, to use most of the available resources on the machines described above, you'd change these lines of your YAML config file:
+
+~~~ yaml
+      containers:
+      - name: cockroachdb
+        image: {{page.release_info.docker_image}}:{{page.release_info.version}}
+        imagePullPolicy: IfNotPresent
+        ports:
+        - containerPort: 26257
+          name: grpc
+        - containerPort: 8080
+          name: http
+~~~
+
+To be:
+
+~~~ yaml
+      containers:
+      - name: cockroachdb
+        image: {{page.release_info.docker_image}}:{{page.release_info.version}}
+        imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            cpu: "3500m"
+            memory: "12300Mi"
+        ports:
+        - containerPort: 26257
+          name: grpc
+        - containerPort: 8080
+          name: http
+~~~
+
+When you create the `StatefulSet`, you'll want to check to make sure that all the CockroachDB pods are scheduled successfully. If you see any get stuck in the pending state, run `kubectl describe pod <podname>` and check the `Events` for information about why they're still pending. You may need to manually preempt pods on one or more nodes by running `kubectl delete pod` on them to make room for the CockroachDB pods. As long as the pods you delete were created by a higher-level Kubernetes object such as a `Deployment` or a `StatefulSet`, they'll be safely recreated on another node.
+
+#### Resource Limits
+
+Resource limits are conceptually similar to resource requests, but serve a different purpose. They let you cap the resources used by a pod to no more than the provided limit, which can have a couple of different uses. For one, it makes for more predictable performance because your pods will not be allowed to use any excess capacity on their machines, meaning that they won't have more resources available to them at some times (during lulls in traffic) than others (busy periods where the other pods on a machine are also fully utilizing their reserved resources). Secondly, it also increases the ["Quality of Service" guaranteed by the Kubernetes runtime](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/resource-qos.md). In short, when a machine is oversubscribed, Kubernetes will first preempt pods that don't have resource requests or limits set ("best-effort" pods), then pods that have requests set but not limits ("burstable" pods), then finally pods that have both requests and limits set to the same values ("guaranteed" pods).
+
+Setting resource limits works about the same as setting resource requests. If you wanted to set resource limits in addition to requests on the config from the [Resource Requests](#resource-requests) section above, you'd change the config to:
+
+~~~ yaml
+      containers:
+      - name: cockroachdb
+        image: {{page.release_info.docker_image}}:{{page.release_info.version}}
+        imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            cpu: "3500m"
+            memory: "12300Mi"
+          limits:
+            cpu: "3500m"
+            memory: "12300Mi"
+        ports:
+        - containerPort: 26257
+          name: grpc
+        - containerPort: 8080
+          name: http
+~~~
+
+The pods would then be restricted to only use the resource they have reserved and guaranteed to not be preempted except in very exceptional circumstances. This typically won't give you better performance on an under-utilized Kubernetes cluster, but will give you more predictable performance as other workloads are run.
+
+{{site.data.alerts.callout_danger}}While setting memory limits is strongly recommended, <a href="https://github.com/kubernetes/kubernetes/issues/51135">setting CPU limits can hurt tail latencies as currently implemented by Kubernetes</a>. We recommend not setting CPU limits on Kubernetes versions 1.9 and below, and only doing so on Kubernetes 1.10 if you use an integer (non-fractional) number of CPUs and set your CPU and memory limits exactly equal to your requests in order for the <a href="https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy">Static CPU Management Policy</a> to kick in.{{site.data.alerts.end}}
+
+#### Default Resource Requests and Limits
+
+Note that even if you don't manually set resource requests yourself, you're likely unknowingly using them anyways. In many installations of Kubernetes, a [`LimitRange`](https://kubernetes.io/docs/tasks/administer-cluster/cpu-default-namespace/) is preconfigured for the `default` namespace that applies a default CPU request of `100m`, or one-tenth of a CPU. You can see this configuration by running
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl describe limitranges
+~~~
+
+~~~
+Name:       limits
+Namespace:  default
+Type        Resource  Min  Max  Default Request  Default Limit  Max Limit/Request Ratio
+----        --------  ---  ---  ---------------  -------------  -----------------------
+Container   cpu       -    -    100m             -              -
+~~~
+
+Experimentally, this does not appear to have a noticeable effect on CockroachDB's performance when a Kubernetes cluster isn't heavily utilized, but don't be surprised if you see CPU requests on your pods that you didn't set.
+
+### Other Pods on the Same Machines as CockroachDB
+
+As discovered in the above section on [Resource Requests and Limits](#resource-requests-and-limits), there will always be pods other than just CockroachDB running in your Kubernetes cluster, even if you don't create any other pods of your own. You can see them at any time by running:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl get pods --all-namespaces
+~~~
+
+~~~
+NAMESPACE     NAME                                             READY     STATUS    RESTARTS   AGE
+kube-system   event-exporter-v0.1.7-5c4d9556cf-6v7lf           2/2       Running   0          2m
+kube-system   fluentd-gcp-v2.0.9-6rvmk                         2/2       Running   0          2m
+kube-system   fluentd-gcp-v2.0.9-m2xgp                         2/2       Running   0          2m
+kube-system   fluentd-gcp-v2.0.9-sfgps                         2/2       Running   0          2m
+kube-system   fluentd-gcp-v2.0.9-szwwn                         2/2       Running   0          2m
+kube-system   heapster-v1.4.3-968544ffd-5tsb8                  3/3       Running   0          1m
+kube-system   kube-dns-778977457c-4s7vv                        3/3       Running   0          1m
+kube-system   kube-dns-778977457c-ls6fq                        3/3       Running   0          2m
+kube-system   kube-dns-autoscaler-7db47cb9b7-x2cc4             1/1       Running   0          2m
+kube-system   kube-proxy-gke-test-default-pool-828d39a7-dbn0   1/1       Running   0          2m
+kube-system   kube-proxy-gke-test-default-pool-828d39a7-nr06   1/1       Running   0          2m
+kube-system   kube-proxy-gke-test-default-pool-828d39a7-rc4m   1/1       Running   0          2m
+kube-system   kube-proxy-gke-test-default-pool-828d39a7-trd1   1/1       Running   0          2m
+kube-system   kubernetes-dashboard-768854d6dc-v7ng8            1/1       Running   0          2m
+kube-system   l7-default-backend-6497bcdb4d-2kbh4              1/1       Running   0          2m
+~~~
+
+These ["cluster add-ons"](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons) provide a variety of basic services like managing DNS entries for services within the cluster, powering the Kubernetes dashboard UI, or collecting logs or metrics from all the pods running in the cluster. If you don't like having them take up space in your cluster, you can prevent some of them from running by configuring your Kubernetes cluster appropriately. For example, on GKE, you can create a cluster with the minimal set of addons by running:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ gcloud container clusters create <your-cluster-name> --no-enable-cloud-logging --no-enable-cloud-monitoring --addons=""
+~~~
+
+However, essentials like `kube-proxy` and `kube-dns` are effectively required to have a compliant Kubernetes cluster. This means that you'll always have some pods that aren't yours running in your cluster, so it's important to understand and account for the possible effects of CockroachDB having to share a machine with other processes. The more processes there are on the same machine as a CockroachDB pod, the worse and less predictable its performance will likely be. To protect against this, it's strongly recommended to run with [Resource Requests](#resource-requests) on your CockroachDB pods to provide some level of CPU and memory isolation.
+
+Setting resource requests isn't a panacea, though. There can still be contention for shared resources like network I/O or, in [exceptional](https://sysdig.com/blog/container-isolation-gone-wrong/) [cases](https://hackernoon.com/another-reason-why-your-docker-containers-may-be-slow-d37207dec27f), internal kernel data structures. For these reasons and because of the Kubernetes infrastructure processes running on each machine, CockroachDB running on Kubernetes simply can't reach quite the same levels of performance as running directly on dedicated machines. Thankfully, it can at least get quite close if you use Kubernetes wisely.
+
+If for some reason setting appropriate resource requests still isn't getting you the performance you expect, you might want to consider going all the way to [dedicated nodes](#dedicated-nodes).
+
+#### Client Applications on the Same Machines as CockroachDB
+
+Running client applications such as benchmarking applications on the same machines as CockroachDB can be even worse than just having Kubernetes system pods on the same machines. They are very likely to end up competing for resources, because when the applications get more loaded than usual, so will the CockroachDB processes. The best way to avoid this is to [set resource requests and limits](#resource-requests-and-limits), but if you are unwilling or unable to do that for some reason, you can also set [anti-affinity scheduling policies](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) on your client applications. Anti-affinity policies are placed in the pod spec, so if you wanted to change our provided example load generator app, you'd change [these lines](https://github.com/cockroachdb/cockroach/blob/98c506c48f3517d1ac1aadb6a09e1b23ad672c37/cloud/kubernetes/example-app.yaml#L11-L12):
+
+~~~ yaml
+    spec:
+      containers:
+~~~
+
+To be:
+
+~~~ yaml
+    spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - loadgen
+              topologyKey: kubernetes.io/hostname
+          - weight: 99
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - cockroachdb
+              topologyKey: kubernetes.io/hostname
+      containers:
+~~~
+
+This configuration will first prefer to put the `loadgen` pods on different nodes from each other, which is important for the fault tolerance of the `loadgen` pods themselves. As a secondary priority, it will attempt to put the pods on nodes that don't already have a running `CockroachDB` pod. This will ensure the best possible balance of fault tolerance and performance for the load generator and CockroachDB cluster.
+
+### Networking
+
+[Kubernetes asks a lot of the network that it runs on](https://kubernetes.io/docs/concepts/cluster-administration/networking/) in order to provide a routable IP address and an isolated Linux network namespace to each pod in the cluster, among its other requirements. While this document isn't nearly large enough to properly explain the details, and those details themselves can depend heavily on specifically how you have set up the network for your cluster, it suffices to say that Docker and Kubernetes' networking abstractions often come with a performance penalty for high-throughput distributed applications such as CockroachDB.
+
+If you really want to eke more performance out of your cluster, networking is a good target to at least experiment with. You can either replace your cluster's networking solution with a more performant one or bypass most of the networking overhead by using the host machines' networks directly.
+
+#### Networking Solutions
+
+If you aren't using a hosted Kubernetes service, you'll typically have to choose how to set up the network when you're creating a Kubernetes cluster. There are [a lot of solutions out there](https://kubernetes.io/docs/concepts/cluster-administration/networking/#how-to-achieve-this), and they can have significantly different performance characteristics and functionality. We don't endorse any networking software or configurations in particular, but want to call out that your choice can have a meaningful affect on performance compared to running CockroachDB outside of Kubernetes.
+
+#### Using the Host's Network
+
+If you are already content with your cluster's networking setup or don't want to have to mess with it, Kubernetes does offer an escape hatch for exceptional cases that lets you avoid network performance overhead -- the `hostNetwork` setting, which allows you to run pods using their host machine's network directly and bypass the layers of abstraction. This comes with a number of downsides, of course. For example, two pods using `hostNetwork` on the same machine can't use the same ports, and it also can have serious security implications if your machines are reachable on the public Internet. If you want to give it a try, though, to see what effects it has for your workload, you just have to add two lines to the CockroachDB YAML configuration file and to any client applications that desparately need better performance, changing:
+
+~~~ yaml
+    spec:
+      affinity:
+~~~
+
+To be:
+
+~~~ yaml
+    spec:
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      affinity:
+~~~
+
+`hostNetwork: true` tells Kubernetes to put the pods in the host machine's network namespace, using its IP address, hostname, and entire networking stack. The `dnsPolicy: ClusterFirstWithHostNet` line tells Kubernetes to configure the pods to still be able to use the cluster's DNS infrastructure for service discovery.
+
+This won't work miracles, so use it with caution. In our testing, it pretty reliably gives about a 6% improvement in database throughput when running [our `kv` load generator](https://hub.docker.com/r/cockroachdb/loadgen-kv/) against a 3-node cluster on GKE.
+
+### Running in a DaemonSet
+
+In all of the examples so far, we've been using the standard CockroachDB `StatefulSet` configuration file and tweaking it slightly. An alternative that comes with a different set of tradeoffs is to completely switch from using a `StatefulSet` for orchestration to using a `DaemonSet`. A [`DaemonSet`](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) is a Kubernetes type that runs a pod on all nodes matching some selection criteria.
+
+This comes with a few main benefits -- it's a more natural abstraction for cordoning off onto [dedicated nodes](#dedicated-nodes), it naturally pairs with [using the host's network](#using-the-hosts-network) since you're already coupling CockroachDB processes one-to-one with nodes, and it allows you to use [local disks](#local-disks) without relying on the beta support for using local disks with `StatefulSets`. The biggest tradeoff is that you're limiting Kubernetes' ability to help your cluster recover from failures. It can't create new pods to replace pods on nodes that fail because it's already running a CockroachDB pod on all the matching nodes. This matches the behavior of running CockroachDB directly on a set of physical machines that are only manually replaced by human operators.
+
+<!-- TODO(a-robinson): link to the checked-in version of this file -->
+To set up a CockroachDB `DaemonSet`, a little more work is needed than for a `StatefulSet`. We will use [this example configuration file](https://gist.github.com/a-robinson/ec2b86783ccbf053c83ba83170673d63) as our base.
+
+First of all, unless you want CockroachDB running on every machine in your Kubernetes cluster, you should pick out which nodes you want to run CockroachDB on using either [node labels](#node-labels) or [node taints](#node-taints). Once you have chosen or created the nodes,  configure them and the `DaemonSet` YAML file appropriately as described in the relevant [Dedicated Nodes](#dedicated-nodes) section.
+
+Then, you must set the addresses in the CockroachDB `--join` flag in the YAML file. The file defaults to [using the host's network](#using-the-hosts-network), so we need to use the host machines' IP addresses or hostnames as join addresses. Pick out two or three of them to include and replace the list (`10.128.0.4,10.128.0.5,10.128.0.3`) in the provided file. Be aware that if the machines you choose are removed from the Kubernetes cluster, you will need to update your `--join` flag values or else new CockroachDB instances will not be able to join the cluster.
+
+Then, pick out the directory from the host that you would like to store CockroachDB's data in and replace the `path: /tmp/cockroach-data` line in the config file with your desired directory. If you're using local SSD, this should be wherever the SSDs are mounted on the machines.
+
+After taking those steps and making any other desired modifications, you should be all set to create the `DaemonSet`:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl create -f cockroachdb-daemonset.yaml
+~~~
+
+~~~
+daemonset "cockroachdb" created
+~~~
+
+To initialize the cluster pick one of the pod names and run:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl exec -it <pod-name> -- ./cockroach init --insecure
+~~~
+
+~~~
+Cluster successfully initialized
+~~~
+
+### Dedicated Nodes
+
+If your Kubernetes cluster is made up of heterogeneous hardware, it's very possible that you'd like to make sure CockroachDB only runs on certain machines. If you want to get as much performance as possible out of a set of machines, you might also want to make sure that nothing other than CockroachDB is run on them.
+
+#### Node Labels
+
+Node labels and node selectors are a way to tell Kubernetes which nodes you want a pod to be allowed on. To label a node, you can just use the `kubectl label node` command as such, substituting in your node's name and your preferred key-value pair for the label:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl label node <node-name> <key>=<value>
+~~~
+
+Some Kubernetes installation tools allow you to automatically apply labels to certain nodes. For example, when creating a new [GKE Node Pool](https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools), you can use the `--node-labels` flag to the `gcloud container node-pools create` command.
+
+Once you do set up labels for all the nodes you want, you can then [use a `NodeSelector`](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodeselector) to control where your pods are allowed to be scheduled. For example, in the `DaemonSet` file from the above example, you would change the lines:
+
+~~~ yaml
+    spec:
+      hostNetwork: true
+      containers:
+~~~
+
+To be:
+
+~~~ yaml
+    spec:
+      nodeSelector:
+        <key>: <value>
+      hostNetwork: true
+      containers:
+~~~
+
+#### Node Taints
+
+Alternatively, if you want to make sure that CockroachDB is the only thing running on a set of machines, you're better off using a pair of complementary features called [`Taints` and `Tolerations`](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) to instruct Kubernetes not to schedule anything else on them. You can set them up in a very similar fashion to how you can set up node labels and node selectors:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ kubectl taint node <node-name> <key>=<value>:NoSchedule
+~~~
+
+Just like for [node labels](#node-labels), some Kubernetes installation tools allow you to automatically apply taints to certain nodes. For example, when creating a new [GKE Node Pool](https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools), you can use the `--node-taints` flag to the `gcloud container node-pools create` command.
+
+Once you have applied the appropriate `Taint`s to each machine you want to only run CockroachDB, add the corresponding `Toleration`s to your CockroachDB config file. For example, in the `DaemonSet` file from the above example, you would change the lines:
+
+~~~ yaml
+    spec:
+      hostNetwork: true
+      containers:
+~~~
+
+To be:
+
+~~~ yaml
+    spec:
+      tolerations:
+      - key: <key>
+        operator: "Equal"
+        value: <value>
+        effect: "NoSchedule"
+      hostNetwork: true
+      containers:
+~~~
+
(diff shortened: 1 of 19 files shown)


bobvawter:
lgtm

Do we need to say anything about how to update the pods that result from a StatefulSet or a DaemonSet if changes to the configuration are desired?

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 112 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kGPZ67pwZ1P8YnA7a:-L9kGPZ67pwZ1P8YnA7b:bfslihg) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L112)):*
> ```Markdown
> ~~~ shell
> $ kubectl patch storageclass ssd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
> ```

Should that be `is-default-class:true` ?

---

*[_includes/orchestration/kubernetes-performance.md, line 398 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kJbGS5gbZnmghpqXR:-L9kJbGS5gbZnmghpqXS:b-5bgjqs) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L398)):*
> ```Markdown
> First of all, unless you want CockroachDB running on every machine in your Kubernetes cluster, you should pick out which nodes you want to run CockroachDB on using either [node labels](#node-labels) or [node taints](#node-taints). Once you have chosen or created the nodes,  configure them and the `DaemonSet` YAML file appropriately as described in the relevant [Dedicated Nodes](#dedicated-nodes) section.
> 
> Then, you must set the addresses in the CockroachDB `--join` flag in the YAML file. The file defaults to [using the host's network](#using-the-hosts-network), so we need to use the host machines' IP addresses or hostnames as join addresses. Pick out a few of them to include and replace the list (`10.128.0.4,10.128.0.5,10.128.0.3`) in the provided file.
> ```

"a few" --> "two or three"

"You only need to choose a handful of IP addresses since a cockroach instance will discover the other members of the cluster from these seed servers.  Be aware that if the seed machines are removed from the Kubernetes cluster, you will need to update your `-join` flag values, or new cockroach instances will not be able to join the cluster."

---

*[_includes/orchestration/kubernetes-performance.md, line 452 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kLJgZ6d9rQ80uvFDZ:-L9kLJgZ6d9rQ80uvFD_:b-cabnwx) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L452)):*
> ```Markdown
>     spec:
>       nodeSelector:
>         key: value
> ```

Saying `key: value` here seems a little vague.  Maybe something like

nodeSelector:
  # A key : value pair
  cockroach : here

---

*[_includes/orchestration/kubernetes-performance.md, line 483 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kLWyh32LfEfu8TNAw:-L9kLWyh32LfEfu8TNAx:bbz4ejb) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L483)):*
> ```Markdown
>       - key: "key"
>         operator: "Equal"
>         value: "value"
> ```

Use actual key:value pairs with an explanatory comment?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-L9kP6mz23258VF85jsY:b-y5k1aq)*

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 45 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQrFR0y36ZhLuOApy:-L9kQrFR0y36ZhLuOApz:bj2apbh) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L45)):*
> ```Markdown
> kind: StorageClass
> metadata:
>   name: ssd
> ```

It's not clear in these examples which names have to match up with what. For example, if I'm interpreting this correctly, `pd-ssd` must be exactly that for GCE, but the `ssd` on this line can be whatever you want as long as it matches up with the `storageClassName` in the next section. It might be nice to name the user-chosen names something like `my-ssd` to make it clear where this is the case. 

---

*[_includes/orchestration/kubernetes-performance.md, line 53 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQHwkAUYRbrMpZKnb:-L9kQHwkAUYRbrMpZKnc:baegpir) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L53)):*
> ```Markdown
> You can then use this new disk type either by configuring the CockroachDB YAML file to request it or by making it the default. You may also want to set additional parameters as documented in the list of Kubernetes storage classes, such as configuring the `iopsPerGB` if you're creating a `StorageClass` for AWS's `io1` Provisioned IOPS volume type.
> 
> ##### Configuring the disk type used by CockroachDB
> ```

Our stylesheet has h5 headers that are actually smaller than the body type. We should either use less deeply-nested headers or revise the stylesheet to have more usable h5/h6 headers. 

---

*[_includes/orchestration/kubernetes-performance.md, line 55 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kRL472TNkQM2Xc2GG:-L9kRL472TNkQM2Xc2GH:b-526r3r) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L55)):*
> ```Markdown
> ##### Configuring the disk type used by CockroachDB
> 
> To use a new `StorageClass` without making it the default in your cluster, you have to modify your application's YAML file to ask for it. In the CockroachDB `StatefulSet` configuration, that means adding a line to its `VolumeClaimTemplates` section. For example, that would mean taking [these lines](https://github.com/cockroachdb/cockroach/blob/b4c33af61e040243de046d7de7f2846bd56842d7/cloud/kubernetes/cockroachdb-statefulset.yaml#L123-L131) of the config file:
> ```

We're going to need to keep bumping this link as the file is changed. It might be better to just refer to the "`volumeClaimTemplates` section" and link to the master version of the file. We have already copied the specific lines below. 

---

*[_includes/orchestration/kubernetes-performance.md, line 155 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kWuV_4lQKMrfIDfhq:-L9kWuV_4lQKMrfIDfhr:b9ecdjn) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L155)):*
> ```Markdown
> ### Local Disks
> 
> Up to this point, we have been assuming that you will be running CockroachDB in a `StatefulSet`, using auto-provisioned remotely attached disks. However, using local disks typically provides better performance than remotely attached disks, such as SSD Instance Store Volumes instead of EBS Volumes on AWS or Local SSDs instead of Persistent Disks on GCE. `StatefulSet`s have historically not supported using local disks, but [beta support for using "local" `PersistentVolume`s was added in Kubernetes v1.10](https://kubernetes.io/docs/concepts/storage/volumes/#local). We don't recommend using this for production data until the feature is more mature, but it's a promising development.
> ```

We may want to suggest running with a replication factor of 5 instead of 3 when using local disks. 

---

*[_includes/orchestration/kubernetes-performance.md, line 227 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kT1I80f10Mjjz2jQy:-L9kT1I80f10Mjjz2jQz:b7t3w9l) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L227)):*
> ```Markdown
>         resources:
>           requests:
>             cpu: "3500m"
> ```

Does the use of `m` as `milli` need explanation in this context?

---

*[_includes/orchestration/kubernetes-performance.md, line 240 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kU2FY7w6xD79QDPIj:-L9kU2FY7w6xD79QDPIk:bqe00g5) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L240)):*
> ```Markdown
> #### Resource Limits
> 
> Resource limits are conceptually to resource requests, but serve a different purpose. They let you cap the resources used by a pod to no more than the provided limit, which can have a couple of different uses. For one, it makes for more predictable performance because your pods will not be allowed to use any excess capacity on their machines, meaning that they won't have more resources available to them at some times (during lulls in traffic) than others (busy periods where the other pods on a machine are also fully utilizing their reserved resources). Secondly, it also increases the ["Quality of Service" guaranteed by the Kubernetes runtime](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/resource-qos.md). In short, when a machine is oversubscribed, Kubernetes will first preempt pods that don't have resource requests or limits set ("best-effort" pods), then pods that have requests set but not limits ("burstable" pods), then finally pods that have both requests and limits set to the same values ("guaranteed" pods).
> ```

Missing word after "conceptually" in the first sentence. 

---

*[_includes/orchestration/kubernetes-performance.md, line 254 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kUKM5ATOko7PPI3OG:-L9kUKM5ATOko7PPI3OH:bb4v1lu) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L254)):*
> ```Markdown
>             memory: "12300Mi"
>           limits:
>             cpu: "3500m"
> ```

How does k8s enforce fractional CPU limits? I haven't tried this in k8s but I've had bad experience with fractional reservations before (because they're often implemented by hitting the process with SIGSTOP and SIGCONT to keep it below the target utilization) and I think it's best to stick with integer CPU limits.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

jseldess:
This is great, @a-robinson. I'll review closely tonight. But I'm also wondering if we should provide a YAML with this. We could then have two versions of the kubernetes tutorial and leverage these explanations in the context of the performance-focused tutorial. Perhaps the yaml would be our "prescriptive" recommendations?

---

Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 434 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9lKUqu2wIwnQaa5xui:-L9lKUqu2wIwnQaa5xuj:bm0qfjn) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L434)):*
> ```Markdown
> ~~~ shell
> $ kubectl label node <node-name> key=value
> ```

Should we make it `<key>=<value>`?

---

*[_includes/orchestration/kubernetes-performance.md, line 452 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kLJgZ6d9rQ80uvFDZ:-L9lK6S4C3zeVYmNzAu1:b-ssqvn7) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L452)):*
<details><summary><i>Previously, bobvawter (Bob Vawter) wrote…</i></summary><blockquote>

Saying `key: value` here seems a little vague.  Maybe something like

nodeSelector:
  # A key : value pair
  cockroach : here

</blockquote></details>

Or maybe just `<key>:<value>`?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-L9lKl-gEaYxdT22sSKH:b-i9laq0)*

jseldess:
cc @kannanlakshmi as well.

jseldess:
*[_includes/orchestration/kubernetes-performance.md, line 39 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9onFh01SNhw2ufYRO6:-L9onFh01SNhw2ufYRO7:bw8ttnw) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L39)):*
> ```Markdown
> #### Creating a different disk type
> 
> Kubernetes exposes the disk types used by its volume provisioner via its [`StorageClass` API object](https://kubernetes.io/docs/concepts/storage/storage-classes/). Each cloud environment has its own default `StorageClass`, but you can easily change the default or create a new named class which you can then ask for when asking for volumes. To do this, pick the type of volue provisioner you want to use from the list in the [Kubernetes documentation](https://kubernetes.io/docs/concepts/storage/storage-classes/), take the example YAML file they provide, modify it to have the disk type you want, then run `kubectl create -f <your-storage-class-file.yaml>`. For example, in order to use the `pd-ssd` disk type on Google Compute Enginer or Google Kubernetes Engine, you can use a `StorageClass` file like this:
> ```

`the type of volue` > `the type of volume`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

jseldess:
*[_includes/orchestration/kubernetes-performance.md, line 388 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9oovuEBUtuDF5AWvfk:-L9oovuEBUtuDF5AWvfl:b-b3uyt4) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L388)):*
> ```Markdown
> This won't work miracles, so use it with caution. In our testing, it pretty reliably gives about a 6% improvement in database throughput when running [our `kv` load generator](https://hub.docker.com/r/cockroachdb/loadgen-kv/) against a 3-node cluster on GKE.
> 
> ### Running in a DaemonSet
> ```

Given that using `DaemonSet` doesn't automate replacing pods, is this really something we want to advocate?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

jseldess:
This is awesome work, @a-robinson! If we go with a document like this, I'm happy to do some editing, though little is needed, really. But before going down that route, I'd like to consider some structural questions related to questions you asked in your initial comment:

1. Should our default k8 tutorial be production-ready? If so, I think we should explore changing the configs to follow the most performance-essential recommendations here. 
    - This would allows us to address a number of these concerns up-front as users learn how to use k8 with cockroach. Instead of providing a non-production-ready config and asking users to customize it for production, we could include tips about scaling back some configurations if performance isn't a focus. 
    - Given that disk and other factors are cloud-specific, however, this approach would mean either providing more configs or asking users to comment out/uncomment certain parts of the config depending on platform. 
2. With performance guidance threaded into the tutorial, I think it may make sense to add an "Orchestration/Kubernetes" section to "Production Checklist" with a reiteration of the essential concepts and recommendations. I'm not yet clear if that would be enough (tutorial plus orchestration section in production checklist) or whether we'd need extra page or pages.

Would be great to hear thoughts from everyone on this thread.

---







*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-L9opKpA6Jwe9NUocoAq:b-wvey)*

robert-s-lee:
> Should we provide more numbers about how much each change helps on a basic workload like kv?

This will be very valuable.  enumerating performance gain and reasons for those fains is very valuable.  Specifying the reason(s) for the observed benefits can help guide whether the should be used for customer's workloads.

bdarnell:
> Should our default k8 tutorial be production-ready?

As much as we can without adding complexity, but I think it's more important to keep the default instructions simple than to make them production-ready (I don't want people to think "kubernetes is too complicated/just for production, so i'll do it manually because it looks simpler"). So I think that means we should add resource reservations/limits to the defaults, but not things like new storage classes.

kannanlakshmi:
This is fantastic @a-robinson thank you! 

-Should we be more prescriptive about which set of changes are most recommended?
I learn towards being a bit more prescriptive. Right now, the various options are fantastic for a somewhat sophisticated user. Do we have a preference between say, daemon set v stateful sets for example, or are we suggesting the user figure out those trade offs themselves? My gut says at this stage users might prefer our recommendation rather than having to do this calculation themselves. 

-Should we provide any more material, such as a preconstructed YAML file with easy-to-replace template parameters showing where a user's customized numbers should go?
Yes, I think we should provide preconstructed YAMLs and take as much out of the devs hands as possible. Its helpful to cater to the lowest common denominator of knowledgeable user (at least for now). Users can always ignore/ make changes as they see fit but if we can do most of the heavy lifting that's best. To @jseldess point, we might need to do a few different configs for each cloud provider with customization/ best practices suggestions. 

@jseldess open to your thoughts on where this fits, but at minimum we should include this both under "orchestration" as well as under "production checklist."

a-robinson:
Thanks for the reviews!

@jseldess, where should we put this in the site hierarchy and link to it from? And should the name of the file (and thus the URL path) just be `kubernetes-performance` or something like that rather than the current verbose option of `orchestrate-cockroachdb-with-kubernetes-performance`?

---

Review status: 0 of 4 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 39 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9onFh01SNhw2ufYRO6:-L9pZ4S11huTlC8GAfb0:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L39)):*

`the type of volue` > `the type of volume`
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 45 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQrFR0y36ZhLuOApy:-L9pZ9xJ0xkMFu3zbT4u:b-58gp4x) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L45)):*

It's not clear in these examples which names have to match up with what. For example, if I'm interpreting this correctly, `pd-ssd` must be exactly that for GCE, but the `ssd` on this line can be whatever you want as long as it matches up with the `storageClassName` in the next section. It might be nice to name the user-chosen names something like `my-ssd` to make it clear where this is the case. 
</blockquote></details>

You're right. I've switched to putting customizable names in `<>` tags as suggested by @jseldess in a couple of places below.

---

*[_includes/orchestration/kubernetes-performance.md, line 53 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQHwkAUYRbrMpZKnb:-L9pc5ob14v4c-PAij2A:b-r82r05) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L53)):*

Our stylesheet has h5 headers that are actually smaller than the body type. We should either use less deeply-nested headers or revise the stylesheet to have more usable h5/h6 headers. 
</blockquote></details>

These h5 sections still make sense at h4, so I've switched them over to that. I'll leave the broader h5 docs question for @jseldess.

---

*[_includes/orchestration/kubernetes-performance.md, line 55 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kRL472TNkQM2Xc2GG:-L9pcS00ACJjPUTYA2rx:b-nabfnk) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L55)):*

We're going to need to keep bumping this link as the file is changed. It might be better to just refer to the "`volumeClaimTemplates` section" and link to the master version of the file. We have already copied the specific lines below. 
</blockquote></details>

I was planning on never updating it to keep up with line numbers, hence the link to the specific SHA. Linking to an old version could be confusing if the version at HEAD changes drastically, but if that happens then the YAML snippets below could also get confusing and need to be updated anyway.

Maybe I should just remove these links and link to the file once near the top here? What do you think, @jseldess?

---

*[_includes/orchestration/kubernetes-performance.md, line 112 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kGPZ67pwZ1P8YnA7a:-L9pZH5EE8D21nHN9q0k:b-iz9uyk) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L112)):*
<details><summary><i>Previously, bobvawter (Bob Vawter) wrote…</i></summary><blockquote>

Should that be `is-default-class:true` ?
</blockquote></details>

Yes! Thanks.

---

*[_includes/orchestration/kubernetes-performance.md, line 155 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kWuV_4lQKMrfIDfhq:-L9pfCO22h--pxNb0-Op:b-i8wb5g) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L155)):*

We may want to suggest running with a replication factor of 5 instead of 3 when using local disks. 
</blockquote></details>

Added a note.

@jseldess - let me know if this idea that you may want to run with a higher replication factor on local disks is referenced anywhere else in the docs. I couldn't find anything.

---

*[_includes/orchestration/kubernetes-performance.md, line 227 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kT1I80f10Mjjz2jQy:-L9pgi9-BtBvYiVTG4zp:b70cvtd) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L227)):*

Does the use of `m` as `milli` need explanation in this context?
</blockquote></details>

Maybe? I added a sentence about it above when it first appears.

---

*[_includes/orchestration/kubernetes-performance.md, line 240 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kU2FY7w6xD79QDPIj:-L9pgut925zR_ks0zkSn:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L240)):*

Missing word after "conceptually" in the first sentence. 
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 254 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kUKM5ATOko7PPI3OG:-L9pw4z10xb_P37vwUXC:bhr49b1) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L254)):*

How does k8s enforce fractional CPU limits? I haven't tried this in k8s but I've had bad experience with fractional reservations before (because they're often implemented by hitting the process with SIGSTOP and SIGCONT to keep it below the target utilization) and I think it's best to stick with integer CPU limits.
</blockquote></details>

CFS quota, typically (https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/). So I wouldn't think fractional reservations would be any different from non-fractional reservations.

Although digging up a source for that answer lead me to find https://github.com/kubernetes/kubernetes/issues/51135, which suggests that we should be discouraging CPU limits entirely for now. I didn't observe any issues in my testing, but I also was allocating almost the entire machine to cockroach.

I'll update this accordingly in my next update.

---

*[_includes/orchestration/kubernetes-performance.md, line 388 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9oovuEBUtuDF5AWvfk:-L9q48oOBUsb7B9Vl2l8:b-yatn2w) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L388)):*
It will always put a pod on any machine in the cluster with the appropriate Kubernetes labels. It will restart pods if they fail, it just won't move CockroachDB onto a machine without the appropriate labels in the case of a machine failure. If some external process (or human administrator) does add a new machine to the cluster to replace an old failed machine, Kubernetes will run cockroach on that new machine if it's also been given the appropriate Kubernetes labels.

> Given that using `DaemonSet` doesn't automate replacing pods, is this really something we want to advocate?

An analogous question would be: "Given that the machines in my data center aren't automatically replaced when one fails, is running on bare metal machines really something we want to advocate?" And of course the answer that question is yes. The answer to this one is the same, for the same reasons.

---

*[_includes/orchestration/kubernetes-performance.md, line 398 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kJbGS5gbZnmghpqXR:-L9pxd0jBn5Ih7VSaUw3:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L398)):*
<details><summary><i>Previously, bobvawter (Bob Vawter) wrote…</i></summary><blockquote>

"a few" --> "two or three"

"You only need to choose a handful of IP addresses since a cockroach instance will discover the other members of the cluster from these seed servers.  Be aware that if the seed machines are removed from the Kubernetes cluster, you will need to update your `-join` flag values, or new cockroach instances will not be able to join the cluster."
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 434 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9lKUqu2wIwnQaa5xui:-L9pgw_z1ftUVpvhqIkZ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L434)):*

Should we make it `<key>=<value>`?
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 452 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kLJgZ6d9rQ80uvFDZ:-L9pb5eEAAdo1Y7K5ssC:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L452)):*

Or maybe just `<key>:<value>`?
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 483 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kLWyh32LfEfu8TNAw:-L9pb6L46roCLikC_OSv:b-3pwjz1) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L483)):*
<details><summary><i>Previously, bobvawter (Bob Vawter) wrote…</i></summary><blockquote>

Use actual key:value pairs with an explanatory comment?
</blockquote></details>

Switched to `<key>` and `<value>` as suggested by @jseldess. It was indeed very unclear before.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-L9pW045-rxavuoBkebC:b-i1xubx)*

bdarnell:
Review status: 0 of 4 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 254 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kUKM5ATOko7PPI3OG:-L9qQ1Y-DigjveUysB-F:b-wymt7w) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L254)):*

CFS quota, typically (https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/). So I wouldn't think fractional reservations would be any different from non-fractional reservations.

Although digging up a source for that answer lead me to find https://github.com/kubernetes/kubernetes/issues/51135, which suggests that we should be discouraging CPU limits entirely for now. I didn't observe any issues in my testing, but I also was allocating almost the entire machine to cockroach.

I'll update this accordingly in my next update.
</blockquote></details>

OK. Looks like CFS quota resets every 100ms by default (https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt) which should be low enough that it doesn't cause availability problems for us, but I'd still advise integer CPU reservations until we've done more testing. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

_includes/orchestration/kubernetes-performance.md
```diff
@@ -0,0 +1,507 @@
...
+
+The size of the machines you're using isn't a Kubernetes-specific concern, but it's always a good place to start if you want more performance. See our [hardware recommendations](recommended-production-settings.html#hardware) for specific suggestions, but using machines with more CPU will almost always allow for greater throughput. Be aware that because Kubernetes runs a set of processes on every machine in a cluster, you typically will get more bang for your buck by using fewer large machines than more small machines.
+
+### Disk type
+
+CockroachDB makes heavy use of the disks you provide it, so using faster disks is an easy way to improve your cluster's performance. Our provided configuration does not specify what type of disks it wants, so in most environments Kubernetes will auto-provision disks of the default type. In the common cloud environments (AWS, GCP, Azure) this means you'll get a slow disks that aren't optimized for database workloads (e.g. HDDs on GCE, SSDs without provisioned IOPS on AWS). However, we [strongly recommend using SSDs](recommended-production-settings.html#hardware) for the best performance, and Kubernetes makes it relatively easy to use them.
+
```
kannanlakshmi: "you'll get **a** slow disk" > "you'll get slow disks"

_includes/orchestration/kubernetes-performance.md
```diff
@@ -0,0 +1,507 @@
...
+            cpu: "3500m"
+            memory: "12300Mi"
+        ports:
+        - containerPort: 26257
+          name: grpc
+        - containerPort: 8080
+          name: http
+~~~
+
+The pods would then be restricted to only use the resuorce they have reserved and guaranteed to not be preempted except in very exceptional circumstances. This typically won't give you better performance on an under-utilized Kubernetes cluster, but will give you more predictable performance as other workloads are run.
```
kannanlakshmi: "resuorce" > misspelt.

a-robinson:
> Do we need to say anything about how to update the pods that result from a StatefulSet or a DaemonSet if changes to the configuration are desired?

Probably. Added very basic info at the bottom of the page.

> This is great, @a-robinson. I'll review closely tonight. But I'm also wondering if we should provide a YAML with this. We could then have two versions of the kubernetes tutorial and leverage these explanations in the context of the performance-focused tutorial. Perhaps the yaml would be our "prescriptive" recommendations?

Yeah, that'd be nice. I'm just not looking forward to a future where we have 2^n configs, one for all possible combinations of (secure/insecure, default/optimized, statefulset/daemonset, etc). If we do this, which one file do we choose? Just an optimized secure statefulset I'd suppose, with insecure/daemonset configs left as exercises for the reader.

I'll be sending a PR with these to the main repo shortly.

> Should our default k8 tutorial be production-ready?

I agree with @bdarnell here. The default should Just Work no matter where you run it. We should optimize it as much as possible within that constraint, but making it work everywhere with no extra effort should take priority over improving its production-readiness. That can be handled by a separate "production ready" template file.

> With performance guidance threaded into the tutorial, I think it may make sense to add an "Orchestration/Kubernetes" section to "Production Checklist" with a reiteration of the essential concepts and recommendations.

Good point, I'll add that in next.

> So I think that means we should add resource reservations/limits to the defaults, but not things like new storage classes.

If we set a memory limit small enough to work any Kubernetes cluster, we'd have to go below our minimum production recommendations. And the requests would similarly have to be tiny to make the pods schedulable on dev environments like a GKE cluster of f1-micros/g1-smalls or on a local Minikube cluster running on your laptop. I can put them in and comment them out, but I'm not otherwise not convinced there's a good number to set them to.

---

Review status: 0 of 4 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 254 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kUKM5ATOko7PPI3OG:-LAEnqwv7wQD6jmp0R-m:b-ut9s9) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L254)):*

OK. Looks like CFS quota resets every 100ms by default (https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt) which should be low enough that it doesn't cause availability problems for us, but I'd still advise integer CPU reservations until we've done more testing. 
</blockquote></details>

As as I understand it, Kubernetes uses CFS quota for *all* CPU reservations in all versions older than 1.10 (which was just released a couple weeks ago), not just fractional reservations.

In 1.10 [it looks like a pod can get exclusive access to CPUs if it's in the `Guaranteed` pod class](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy) (i.e. if it sets `cpu-request == cpu-limit` and `memory-request == memory-limit`). So while I haven't seen performance problems caused by using cpu requests/limits, if you've had those issues before we should probably give opposite advice for the different k8s versions.

I've updated the warning accordingly. Look OK?

---

*[_includes/orchestration/kubernetes-performance.md, line 36 at r2](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9jvL3rlksG2kOXK5cH-r2-36:-LAEnoTcDLwGKM1XwW1l:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ce5969f69e28d87552553f96086017e25b1f395b/_includes/orchestration/kubernetes-performance.md#L36)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

"you'll get **a** slow disk" > "you'll get slow disks"
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 265 at r2](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9jvL3rlksG2kOXK5cH-r2-265:-LAF3tKbEv1wmjqcMf5H:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/ce5969f69e28d87552553f96086017e25b1f395b/_includes/orchestration/kubernetes-performance.md#L265)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

"resuorce" > misspelt. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-L9q7u9C5vomdFLQI0MK:bu0yvog)*

a-robinson:
Sent out https://github.com/cockroachdb/cockroach/pull/24869 with PoC "high-performance" configurations.

---

Review status: 0 of 9 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-LAJ49U23dL4ayzQLWCi:b-kezi78)*

bdarnell:
:lgtm:

---

Reviewed 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 45 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQrFR0y36ZhLuOApy:-LAJXyMU-s_usBceGKXZ:bh2jb1m) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L45)):*

You're right. I've switched to putting customizable names in `<>` tags as suggested by @jseldess in a couple of places below.
</blockquote></details>

Angle brackets can be confusing (are they significant in yaml syntax?). I prefer to use syntactically-valid files with names that suggest customization (`my-whatever`) instead of marking placeholders with special syntax. But I defer to consistency with the rest of the docs. 

---

*[_includes/orchestration/kubernetes-performance.md, line 55 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kRL472TNkQM2Xc2GG:-LAJYJoZ-zEk4KU42_Wi:b-sjvzeh) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L55)):*

I was planning on never updating it to keep up with line numbers, hence the link to the specific SHA. Linking to an old version could be confusing if the version at HEAD changes drastically, but if that happens then the YAML snippets below could also get confusing and need to be updated anyway.

Maybe I should just remove these links and link to the file once near the top here? What do you think, @jseldess?
</blockquote></details>

Yeah, I think just one link at the top is the way to go. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-LAJYyTM-SX_2jnmEauG:bnfp4nl)*

jseldess:
This is looking great. Some minor comments and structure suggestions below. In terms of naming and sidenav placement:

- Let's just call the file(s) `kubernetes-performance.md`. If we choose to rename, we can do a client-side redirect. The current page title seems fine.
- It's a little messy/verbose, but let's change the sidenav to:

   ```
   Orchestrated Deployment
      Overview
      Kubernetes Deployment
      Kubernetes Performance Optimization
      Docker Swarm Deployment
      Mesosphere DC/OS Deployment
   ``` 

Also, we should probably mention this new page on `orchestration.md` and add a note to the top of the Kubernetes tutorials referencing this performance guidance. Can you add that as well?

---

Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 45 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQrFR0y36ZhLuOApy:-LATw-RP4x9cEFgfkdLm:blxyveq) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L45)):*

Angle brackets can be confusing (are they significant in yaml syntax?). I prefer to use syntactically-valid files with names that suggest customization (`my-whatever`) instead of marking placeholders with special syntax. But I defer to consistency with the rest of the docs. 
</blockquote></details>

For now, I'd say we go with angle brackets, for the sake of consistency with the rest of the docs. But this is something we could reconsider globally.

---

*[_includes/orchestration/kubernetes-performance.md, line 53 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kQHwkAUYRbrMpZKnb:-LATw7q-BCWTUGohzOuP:bn5www3) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L53)):*

These h5 sections still make sense at h4, so I've switched them over to that. I'll leave the broader h5 docs question for @jseldess.
</blockquote></details>

Thanks, Alex. I don't think we ever use h5 in our docs. That's a degree of nesting that probably indicates the need to split the page into multiple pages. But also, the current styling for h5 is bad, as Ben mentions.

---

*[_includes/orchestration/kubernetes-performance.md, line 55 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kRL472TNkQM2Xc2GG:-LATwVjQ9ft-yjI-7kvm:b-3pef4a) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L55)):*

Yeah, I think just one link at the top is the way to go. 
</blockquote></details>

I agree that we should probably just link at the top. 

---

*[_includes/orchestration/kubernetes-performance.md, line 155 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kWuV_4lQKMrfIDfhq:-LATwrZD6mVEzhf2UzPU:bwfly1l) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L155)):*

Added a note.

@jseldess - let me know if this idea that you may want to run with a higher replication factor on local disks is referenced anywhere else in the docs. I couldn't find anything.
</blockquote></details>

No, I don't think we mention that anywhere. Perhaps worth calling out [here](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#basic-recommendations) or somewhere else on the Production Checklist?

---

*[_includes/orchestration/kubernetes-performance.md, line 9 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUBoeCDmIUn4TxCPGs:-LAUBoeCDmIUn4TxCPGt:bo9tu5p) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L9)):*
I think we can get rid of this section and merge the essential details with the top-of-page introduction, e.g.:

> Kubernetes provides many useful abstractions for deploying and operating distributed systems, but some of the abstractions come with a performance overhead and an increase in underlying system complexity. This page explains the potential CockroachDB-specific bottelnecks to be aware of and shows you how to optimize your deployment for better performance. 

Thoughts?

---

*[_includes/orchestration/kubernetes-performance.md, line 13 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAU8DOR22TK_DJgwpI5:-LAU8DOR22TK_DJgwpI6:biwq4rd) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L13)):*
> ```Markdown
> Kubernetes provides many useful abstractions for deploying and operating distributed systems, but some of those abstractions come with a performance overhead and an increase in underlying system complexity that can make it hard to diagnose performance problems. If you aren't getting the performance you need out of your CockroachDB cluster that's running in Kubernetes, it's worth being aware of potential bottlenecks and knowing how to optimize your deployment for better performance.
> 
> ## Benchmarking your workload
> ```

I think we may be able to rewrite "Prerequisites" to cover this, e,g,:

```
## Prerequisites

Before you focus on optimizing a Kubernetes-orchestrated CockroachDB cluster:

1. Go through the documentation for [running a CockroachDB cluster on Kubernetes](orchestrate-cockroachdb-with-kubernetes.html) to familiarize yourself with the necessary Kubernetes terminology and deployment abstractions.
2. Verify that CockroachDB performs up to your requirements for your workload on identical hardware without Kubernetes. You may find that you need to [modify your workload](performance-best-practices-overview.html) or use [different machine specs](recommended-production-settings.html#hardware) to achieve the performance you need, and it's better to determine that up front than after spending a bunch of time trying to optimize your Kubernetes deployment.
```
 

---

*[_includes/orchestration/kubernetes-performance.md, line 17 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUEwkhBnZlAESruDNQ:-LAUEwkhBnZlAESruDNR:b-ofinl6) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L17)):*
> ```Markdown
> Before investing time in targeting Kubernetes-specific performance improvements, you should verify that CockroachDB performs up to your requirements for your workload on identical hardware without Kubernetes. You may find that you need to [modify your workload](performance-best-practices-overview.html) or use [different machine specs](recommended-production-settings.html#hardware) to achieve the performance you need, and it's better to determine that up front than after spending a bunch of time trying to optimize your Kubernetes deployment.
> 
> ## Factors that can affect performance
> ```

I'd change this to `Performance Factors`.

---

*[_includes/orchestration/kubernetes-performance.md, line 33 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAU7xAm1bPCRuQbqiRb:-LAU7xAm1bPCRuQbqiRc:bi1gubu) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L33)):*
> ```Markdown
> The size of the machines you're using isn't a Kubernetes-specific concern, but it's always a good place to start if you want more performance. See our [hardware recommendations](recommended-production-settings.html#hardware) for specific suggestions, but using machines with more CPU will almost always allow for greater throughput. Be aware that because Kubernetes runs a set of processes on every machine in a cluster, you typically will get more bang for your buck by using fewer large machines than more small machines.
> 
> ### Disk type
> ```

nit: `Disk type `> `Disk Type`

---

*[_includes/orchestration/kubernetes-performance.md, line 161 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUDej01jxD53TgEDpq:-LAUDej01jxD53TgEDpr:bdrsym3) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L161)):*
> ```Markdown
> Note that when running with local disks, there is a greater chance of experiencing a disk failure than when using the cloud providers' network-attached disks that are often replicated underneath the covers. Consequently, you may want to [configure replication zones](#configure-replication-zones.html) to increase the replication factor of your data to 5 from its default of 3 when using local disks.
> 
> ### Setting Resource Requests and Limits
> ```

Since we having introduced the topics above with a verb, I'd say we should stay consistent here and use `Resource Requests and Limits`.

---

*[_includes/orchestration/kubernetes-performance.md, line 288 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUEUBC1qivuLrv46ky:-LAUEUBC1qivuLrv46kz:b-czb4yh) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L288)):*
> ```Markdown
> Experimentally, this does not appear to have a noticeable effect on CockroachDB's performance when a Kubernetes cluster isn't heavily utilized, but don't be surprised if you see CPU requests on your pods that you didn't set.
> 
> ### Other Pods on the Same Machines as CockroachDB
> ```

To be consistent with other parallel headings, and more concise, we could just use `Cluster Add-ons`. But the current heading is probably more helpfully descriptive.

---

*[_includes/orchestration/kubernetes-performance.md, line 368 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUDwgd8y-jDel-DbVA:-LAUDwgd8y-jDel-DbVB:bfglhzi) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L368)):*
> ```Markdown
> This configuration will first prefer to put the `loadgen` pods on different nodes from each other, which is important for the fault tolerance of the `loadgen` pods themselves. As a secondary priority, it will attempt to put the pods on nodes that don't already have a running `CockroachDB` pod. This will ensure the best possible balance of fault tolerance and performance for the load generator and CockroachDB cluster.
> 
> ### Using the Host's Network
> ```

Same as above: To stay consistent, perhaps we should just use `Networking` as the heading.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-LAU76KQ8YYu6VlElYzf:b-lkbvzm)*

a-robinson:
> add a note to the top of the Kubernetes tutorials referencing this performance guidance

How would suggest working it in? As a callout?

> Oh, also, I didn't see any updates to `recommended-production-settings.md`. Do you intend to do that as part of this PR?

Done, albeit very briefly.

---

Review status: 0 of 17 files reviewed at latest revision, 17 unresolved discussions, some commit checks pending.

---

*[_includes/orchestration/kubernetes-performance.md, line 55 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kRL472TNkQM2Xc2GG:-LAddbNf30HhW8D-jjV5:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L55)):*

I agree that we should probably just link at the top. 
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 155 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-L9kWuV_4lQKMrfIDfhq:-LAddlRlArxB_9j6BuSq:b-90rw21) ([raw file](https://github.com/cockroachdb/docs/blob/4f2273345bc306e50f75296fa00543799e9cbf65/_includes/orchestration/kubernetes-performance.md#L155)):*

No, I don't think we mention that anywhere. Perhaps worth calling out [here](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#basic-recommendations) or somewhere else on the Production Checklist?
</blockquote></details>

Sent https://github.com/cockroachdb/docs/pull/3001

---

*[_includes/orchestration/kubernetes-performance.md, line 9 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUBoeCDmIUn4TxCPGs:-LAdNdlB44IH1MKtVuCX:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L9)):*

I think we can get rid of this section and merge the essential details with the top-of-page introduction, e.g.:

> Kubernetes provides many useful abstractions for deploying and operating distributed systems, but some of the abstractions come with a performance overhead and an increase in underlying system complexity. This page explains the potential CockroachDB-specific bottelnecks to be aware of and shows you how to optimize your deployment for better performance. 

Thoughts?

</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 13 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAU8DOR22TK_DJgwpI5:-LAdO7ze6b3-XoEQZTwv:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L13)):*

I think we may be able to rewrite "Prerequisites" to cover this, e,g,:

```
## Prerequisites

Before you focus on optimizing a Kubernetes-orchestrated CockroachDB cluster:

1. Go through the documentation for [running a CockroachDB cluster on Kubernetes](orchestrate-cockroachdb-with-kubernetes.html) to familiarize yourself with the necessary Kubernetes terminology and deployment abstractions.
2. Verify that CockroachDB performs up to your requirements for your workload on identical hardware without Kubernetes. You may find that you need to [modify your workload](performance-best-practices-overview.html) or use [different machine specs](recommended-production-settings.html#hardware) to achieve the performance you need, and it's better to determine that up front than after spending a bunch of time trying to optimize your Kubernetes deployment.
```
 
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 17 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUEwkhBnZlAESruDNQ:-LAdOA2z4-WV9DtL3rUE:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L17)):*

I'd change this to `Performance Factors`.
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 33 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAU7xAm1bPCRuQbqiRb:-LAdOByFAmDN6F8AdK27:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L33)):*

nit: `Disk type `> `Disk Type`
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 161 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUDej01jxD53TgEDpq:-LAdRCMuAh7Qt52-iW6j:bfbq99i) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L161)):*

Since we having introduced the topics above with a verb, I'd say we should stay consistent here and use `Resource Requests and Limits`.
</blockquote></details>

Good point, done.

---

*[_includes/orchestration/kubernetes-performance.md, line 288 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUEUBC1qivuLrv46ky:-LAdRZkr98BwBImpdb6I:btv1lef) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L288)):*

To be consistent with other parallel headings, and more concise, we could just use `Cluster Add-ons`. But the current heading is probably more helpfully descriptive.
</blockquote></details>

I may not have ordered the information ideally since I do spend the first few paragraphs focused on the cluster add-ons, but the section is really meant to be about other pods on the same machines in general, not just about add-on pods. I can reorder the info if you think it would make things more clear, but titling the section `Cluster Add-ons` wouldn't encompass what this section is really supposed to be conveying.

---

*[_includes/orchestration/kubernetes-performance.md, line 368 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUDwgd8y-jDel-DbVA:-LAdT5z620eTPjvXwV1l:b7v8sr3) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L368)):*

Same as above: To stay consistent, perhaps we should just use `Networking` as the heading.
</blockquote></details>

I think ideally this should be split up into two sections underneath a `Networking` header -- one for using a more performant networking solution in your cluster, and one for falling back to using `hostNetwork`. How does this look?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-LAdf5UlC2eHIIlD7dlt:b-vv1bjn)*

jseldess:
:lgtm: with some final nits. Also, the PR needs to be rebased against master and applied to the new v2.1 content.

For referencing the performance page from the main tutorial, yeah, a callout might work best just about the page toc. Perhaps something like this?

```
{{site.data.alerts.callout_success}}For details about potential performance bottlenecks to be aware of when running CockroachDB in Kubernetes and guidance on how to optimize your deployment for better performance, see <a href="kubernetes-performance.html">CockroachDB Performance on Kubernetes</a>.{{site.data.alerts.end}}
```

---

Review status: 0 of 17 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 288 at r3](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAUEUBC1qivuLrv46ky:-LAh4tej6zj6jkg4Jo0U:b-15se3k) ([raw file](https://github.com/cockroachdb/docs/blob/7ac2fbb8f67bed4900d748dd242488c9959e4c6c/_includes/orchestration/kubernetes-performance.md#L288)):*

I may not have ordered the information ideally since I do spend the first few paragraphs focused on the cluster add-ons, but the section is really meant to be about other pods on the same machines in general, not just about add-on pods. I can reorder the info if you think it would make things more clear, but titling the section `Cluster Add-ons` wouldn't encompass what this section is really supposed to be conveying.
</blockquote></details>

Thanks for the clarification. This LGTM now.

---

*[_includes/orchestration/kubernetes-performance.md, line 264 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAh3HdMDYY-WdxUCoAb:-LAh3HdMDYY-WdxUCoAc:bw54ypc) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L264)):*
> ```Markdown
> The pods would then be restricted to only use the resource they have reserved and guaranteed to not be preempted except in very exceptional circumstances. This typically won't give you better performance on an under-utilized Kubernetes cluster, but will give you more predictable performance as other workloads are run.
> 
> {{site.data.alerts.callout_danger}}While setting memory limits is strongly recommended, [setting CPU limits can hurt tail latencies as currently implemented by Kubernetes](https://github.com/kubernetes/kubernetes/issues/51135). We recommend not setting CPU limits on Kubernetes versions 1.9 and below, and only doing so on Kubernetes 1.10 if you use an integer (non-fractional) number of CPUs and set your CPU and memory limits exactly equal to your requests in order for the [Static CPU Management Policy](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy) to kick in.{{site.data.alerts.end}}
> ```

Unfortunately, within a callout like this, you need to use html for the links, e.g., `<a href="https://github.com/kubernetes/kubernetes/issues/51135">setting CPU limits can hurt tail latencies as currently implemented by Kubernetes</a>`.

---

*[_includes/orchestration/kubernetes-performance.md, line 287 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAh3fv_0zRu_9z2p1WU:-LAh3fv_0zRu_9z2p1WV:bjl87yg) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L287)):*
> ```Markdown
> ### Other Pods on the Same Machines as CockroachDB
> 
> As discovered in the above section on [Setting Resource Requests and Limits](#setting-resource-requests-and-limits), there will always be pods other than just CockroachDB running in your Kubernetes cluster, even if you don't create any other pods of your own. You can see them at any time by running:
> ```

`[Setting Resource Requests and Limits](#setting-resource-requests-and-limits)` > `[Resource Requests and Limits](#resource-requests-and-limits)`

---

*[_includes/orchestration/kubernetes-performance.md, line 328 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAh4YhTERiezV9Wzeja:-LAh4YhTERiezV9Wzejb:bsntakg) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L328)):*
> ```Markdown
> #### Client Applications on the Same Machines as CockroachDB
> 
> Running client applications such as benchmarking applications on the same machines as CockroachDB can be even worse than just having Kubernetes system pods on the same machines. They are very likely to end up competing for resources, because when the applications get more loaded than usual, so will the CockroachDB processes. The best way to avoid this is to [set resource requests and limits](#setting-resource-requests-and-limits), but if you are unwilling or unable to do that for some reason, you can also set [anti-affinity scheduling policies](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) on your client applications. Anti-affinity policies are placed in the pod spec, so if you wanted to change our provided example load generator app, you'd change [these lines](https://github.com/cockroachdb/cockroach/blob/98c506c48f3517d1ac1aadb6a09e1b23ad672c37/cloud/kubernetes/example-app.yaml#L11-L12):
> ```

`(#setting-resource-requests-and-limits)` > `(#resource-requests-and-limits)`

---

*[_includes/orchestration/kubernetes-performance.md, line 377 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAh5PiEChmx61xPsB1d:-LAh5PiEChmx61xPsB1e:b-d7horj) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L377)):*
> ```Markdown
> #### Using the Host's Network
> 
> If you are already content with your cluster's networking setup or don't want to have to mess with it, Kubernetes does offer an escape hatch for exceptional cases that lets you avoid network performance overhead -- the `hostNetwork` setting, which allows you to run pods using their host machine's network directly and bypass the layers of abstraction. This comes with a number of downsides, of course. For example, two pods using `hostNetwork` on the same machine can't use the same ports, and it also can have serious security implications if your machines are reachable on the public Internet. If you want to give it a try, though, to see what effects it has for your workload, you can just have to add two lines to the CockroachDB YAML configuration file and to any client applications that desparately need better performance, changing:
> ```

`you can just have to add two lines` > `you just have to add two lines`.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-:-LAh6_Dr30cWnJtYjkHz:b-1lowte)*

jseldess:
*[_includes/orchestration/kubernetes-performance.md, line 156 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAmYtf41Rl0lyZBoOat:-LAmYtf5EicqK3dM_CMd:b-7at75h) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L156)):*
> ```Markdown
> There is also the option of using local disks if you don't run CockroachDB in a `StatefulSet`, but instead use a `DaemonSet`. For more details on what this entails, see the section on [Running in a DaemonSet](#running-in-a-daemonset).
> 
> Note that when running with local disks, there is a greater chance of experiencing a disk failure than when using the cloud providers' network-attached disks that are often replicated underneath the covers. Consequently, you may want to [configure replication zones](#configure-replication-zones.html) to increase the replication factor of your data to 5 from its default of 3 when using local disks.
> ```

broken link: `#configure-replication-zones.html` > `configure-replication-zones.html`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

jseldess:
*[_includes/orchestration/kubernetes-performance.md, line 10 at r5](https://beta.reviewable.io/reviews/cockroachdb/docs/2896#-LAmZIS87ejWfeip0Fa6:-LAmZIS87ejWfeip0Fa7:b-9sgzhs) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L10)):*
> ```Markdown
> 
> 1. Go through the documentation for [running a CockroachDB cluster on Kubernetes](orchestrate-cockroachdb-with-kubernetes.html) to familiarize yourself with the necessary Kubernetes terminology and deployment abstractions.
> 2. Verify that CockroachDB performs up to your requirements for your workload on identical hardware without Kubernetes. You may find that you need to [modify your workload](performance-best-practices-overview.html) or use [different machine specs](recommended-production-settings.html#hardware) to achieve the performance you need, and it's better to determine that up front than after spending a bunch of time trying to optimize your Kubernetes deployment.
> ```

This is annoying, but the `performance-best-practices-overview.html` page isn't available in 1.0, which is making TC fail. There are various things we can do here, but I think the best option is to just not apply any of these changes to the 1.0 docs. Thoughts?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2896)*

a-robinson:
Added the callout and the v2.1 content.

Thanks so much for the reviews!

---

Review status: 0 of 17 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[_includes/orchestration/kubernetes-performance.md, line 10 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAmZIS87ejWfeip0Fa6:-LApQF-hC0Qa3GeBeU3_:b-2dkt5) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L10)):*

This is annoying, but the `performance-best-practices-overview.html` page isn't available in 1.0, which is making TC fail. There are various things we can do here, but I think the best option is to just not apply any of these changes to the 1.0 docs. Thoughts?
</blockquote></details>

Yeah, it's not worth the effort to do much else. Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 156 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAmYtf41Rl0lyZBoOat:-LApQPf9EtX1NVyD7FiK:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L156)):*

broken link: `#configure-replication-zones.html` > `configure-replication-zones.html`
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 264 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAh3HdMDYY-WdxUCoAb:-LApQVKH9vndsDswyrwz:bo3vl6y) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L264)):*

Unfortunately, within a callout like this, you need to use html for the links, e.g., `<a href="https://github.com/kubernetes/kubernetes/issues/51135">setting CPU limits can hurt tail latencies as currently implemented by Kubernetes</a>`.
</blockquote></details>

Thanks, done.

---

*[_includes/orchestration/kubernetes-performance.md, line 287 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAh3fv_0zRu_9z2p1WU:-LApRbS2BriGPoq6yesa:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L287)):*

`[Setting Resource Requests and Limits](#setting-resource-requests-and-limits)` > `[Resource Requests and Limits](#resource-requests-and-limits)`
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 328 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAh4YhTERiezV9Wzeja:-LApRezf2nrZHBpMvPCr:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L328)):*

`(#setting-resource-requests-and-limits)` > `(#resource-requests-and-limits)`
</blockquote></details>

Done.

---

*[_includes/orchestration/kubernetes-performance.md, line 377 at r5](https://reviewable.io/reviews/cockroachdb/docs/2896#-LAh5PiEChmx61xPsB1d:-LApRjvYBcS5md_6hVg-:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/a89ba2f0d49b6e17098a383c1ec76907217511fb/_includes/orchestration/kubernetes-performance.md#L377)):*

`you can just have to add two lines` > `you just have to add two lines`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/2896#-:-LApTJU-EHudLtd6Rpqe:bcg29w8)*

kannanlakshmi:
Minor nit: under 'Running in a Daemon set' there is a "TODO" parenthesis that probably wasn't meant to be in the final draft? See below? 

_To set up a CockroachDB DaemonSet, a little more work is needed than for a StatefulSet. We will use this example configuration file as our base. (TODO: Replace with a checked-in version or two that also include the cockroachdb-public service)_

jseldess:
:lgtm:, once Lakshmi's catch has been addressed. Thanks for this, @a-robinson!

---







*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/2896#-:-LArm5Zu0-X2HrsgaqUL:bw38p2i)*

embano1:
@a-robinson 
Just read your documentation and have some questions/suggestions:

> We recommend not setting CPU limits on Kubernetes versions 1.9 and below, and only doing so on Kubernetes 1.10 if you use an integer (non-fractional) number of CPUs and set your CPU and memory limits exactly equal to your requests in order for the Static CPU Management Policy to kick in.

CPU Manager (beta in v1.10) is a flag that has to be activated on kubelet, may be add this for clarification (defaults to `none`) -> https://kubernetes.io/docs/reference/generated/kubelet/

> Secondly, it also increases the "Quality of Service" guaranteed by the Kubernetes runtime. In short, when a machine is oversubscribed, Kubernetes will first preempt pods that don't have resource requests or limits set ("best-effort" pods), then pods that have requests set but not limits ("burstable" pods), then finally pods that have both requests and limits set to the same values ("guaranteed" pods).

Strictly speaking, this is the behaviour of Kubernetes before v1.9. This has changed slightly in v1.9, where usage above `requests` is now included for eviction logic. So from v1.9 it can happen that a `burstable` pod with exhaustive memory usage could be evicted before a `best effort` pod only consuming little memory. Here's the differences:

v1.10: https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/kubelet/eviction/helpers.go#L568

v1.8: https://github.com/kubernetes/kubernetes/blob/release-1.8/pkg/kubelet/eviction/helpers.go#L577

a-robinson:
Thanks a ton for the corrections, @embano1! I really appreciate it. I've opened https://github.com/cockroachdb/docs/pull/3025 to fix the problems in the original text.

embano1:
You're more than welcome Alex. Keep up the good work!

PR #2902, opened 2018-04-11
Add replication zone details to diagnostics reporting

### LINKED ISSUE #2800: Zone config usage is now included in diagnostic reports
PR: https://github.com/cockroachdb/cockroach/pull/23750

jseldess:
Fixes #2800

### DIFF
--- a/v2.0/diagnostics-reporting.md
+++ b/v2.0/diagnostics-reporting.md
@@ -282,6 +282,67 @@ This JSON example shows what anonymized Admin UI page views information looks li
 }
 ~~~
 
+### Replication Zone Details <span class="version-tag">New in v2.0</span>
+
+CockroachDB collects anonymized data about [replication zone](configure-replication-zones.html#replication-zone-format) usage on an hourly basis.
+
+#### Example
+
+This JSON example shows an excerpt of what replication zone details look like when sent to Cockroach Labs. Note that all names and other strings have been scrubbed from the queries and replaced with `"unknown"`.
+
+~~~ json
+{
+  "zone_configs": {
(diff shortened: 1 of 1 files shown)


jseldess:
@dt, I assumed zone config details are shared hourly like other stats. Please correct if that's wrong. Also, in the related issue, you mentioned a few other new fields. Can you please provide more details on those?

dt:
Yep, there are new fields:  off the top of my head there's a localities field on nodes now, there's a list of altered_settings and and a map of error codes with counts. Additionally, any changes in the embedded messages might change the existing examples -- e.g. i'm sure table_descriptor has some new fields in it.

This has me thinking that this hand-maintained example report is prone to rot / might require disproportionate effort to maintain -- minor changes to table descriptors for example may hardly seem relevant to the docs at all, but would introduce inaccuracy in this example. 

I think we might want to explore an alternative approach. One option might be to automatically re-generate an example report during builds and embed that in the docs, like we do for functions & operators, settings, grammar, etc. Or maybe it makes sense to just have the docs contain a high-level summary of the report, and then pointing users to how they can get a real example report directly from their own cluster, e.g. cockroachdb/coackroach#24813.

jseldess:
Sorry for the long delay, @dt. I like both of your proposed alternatives. Now that there's an endpoint we can point users to, I particularly like the idea of providing a high-level summary in the docs and pointing them to the debug page for real examples. You ok with that option?

dt:
yep

dt:
though that endpoint isn't in 2.0 (unless we backport it in 2.0.2)

PR #2903, opened 2018-04-11
Document server.shutdown.drain_wait setting

### LINKED ISSUE #2791: Document server.shutdown.drain_wait cluster setting
PRs: 
- https://github.com/cockroachdb/cockroach/pull/23319
- https://github.com/cockroachdb/cockroach/pull/23629

From release notes:
> A node now waits in an unready state for the length of time specified by the server.shutdown.drain_wait cluster setting before draining. This helps ensure that load balancers won't send client traffic to a node about to be drained.
> Changed the server.drain_max_wait cluster setting to server.shutdown.query_wait.

jseldess:
Fixes #2791

### DIFF
--- a/_includes/faq/planned-maintenance.md
+++ b/_includes/faq/planned-maintenance.md
@@ -11,3 +11,9 @@ After completing the maintenance work and [restarting the nodes](start-a-node.ht
 ~~~ sql
 > SET CLUSTER SETTING server.time_until_store_dead = '5m0s';
 ~~~
+
+It's also important to ensure that load balancers don't send client traffic to a node about to be shut down, even if it will only be down for a few seconds. If you find that your load balancer's health check is not always recognizing a node as unready before the node shuts down, you can increase the `server.shutdown.drain_wait` setting, which tells the node to wait in an unready state for the specified duration. For example:
+
+ ~~~ sql
+ > SET CLUSTER SETTING server.shutdown.drain_wait = '10s';
+ ~~~
--- a/v2.0/monitoring-and-alerting.md
+++ b/v2.0/monitoring-and-alerting.md
@@ -91,7 +91,8 @@ Otherwise, it returns an HTTP `200 OK` status response code with details about t
 
 <span class="version-tag">New in v2.0:</span> The `http://<node-host>:<http-port>/health?ready=1` endpoint returns an HTTP `503 Service Unavailable` status response code with an error in the following scenarios:
 
-- The node is being [decommissioned](remove-nodes.html) or in the process of [shutting down](stop-a-node.html) and is therefore not able to accept SQL connections and execute queries. This is especially useful for making sure load balancers don't direct traffic to nodes that are live but not "ready", which is a necessary check during [rolling upgrades](upgrade-cockroach-version.html).    
+- The node is being [decommissioned](remove-nodes.html) or in the process of [shutting down](stop-a-node.html) and is therefore not able to accept SQL connections and execute queries. This is especially useful for making sure load balancers don't direct traffic to nodes that are live but not "ready", which is a necessary check during [rolling upgrades](upgrade-cockroach-version.html).
+    {{site.data.alerts.callout_success}}If you find that your load balancer's health check is not always recognizing a node as unready before the node shuts down, you can increase the <code>server.shutdown.drain_wait</code> <a href="cluster-settings.html">cluster setting</a> to cause a node to return <code>503 Service Unavailable</code> even before it has started shutting down.{{site.data.alerts.end}}
 - The node is unable to communicate with a majority of the other nodes in the cluster, likely because the cluster is unavailable due to too many nodes being down.
 
 {% include copy-clipboard.html %}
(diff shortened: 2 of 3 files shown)


_includes/faq/planned-maintenance.md
```diff
@@ -1,13 +1,21 @@
-By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. Before temporarily stopping nodes for planned maintenance (e.g., upgrading system software), if you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window.
+By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. Before temporarily stopping nodes for planned maintenance (e.g., upgrading system software), if you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window. In addition, to ensure that load balancers won't send client traffic to a node about to be shut down, you can increase the `server.shutdown.drain_wait` setting.
```
a-robinson: It's weird to me to mix the `drain_wait` setting in so closely with the `time_until_store_dead` setting, since they're not very related. `drain_wait` is important any time you're stopping a cockroach process, even if it's only going to be a few seconds before you restart it, while `time_until_store_dead` is only relevant if a node is going to be offline for more than 5 minutes.

We could give this a separate subsection, maybe? Or just fully explain one setting before mentioning the other rather than blending them?

Also, this is less clear cut, but you could argue that `drain_wait` is relevant for more than just "planned maintenance" situations since our docs up to this point seem to have defined "planned maintenance" to mean events longer than 5 minutes.

_includes/faq/planned-maintenance.md
```diff
@@ -1,13 +1,21 @@
...
+~~~
+
+After completing the maintenance work and [restarting the nodes](start-a-node.html), you would then change the settings back to their defaults:
 
 ~~~ sql
 > SET CLUSTER SETTING server.time_until_store_dead = '5m0s';
 ~~~
+
+~~~ sql
+> SET CLUSTER SETTING server.shutdown.drain_wait = '0s';
```
a-robinson: This is the kind of setting that you're very unlikely to want to reset back to zero. It's the sort of setting for which you want to figure out a value that works well for your load balancers (based on their health check frequency to make sure they have enough time to mark a node as unready) and then leave at that value forever.

jseldess: Removed.

_includes/faq/planned-maintenance.md
jseldess: Thanks, @a-robinson. I made some changes. PTAL.

_includes/faq/planned-maintenance.md
```diff
@@ -1,6 +1,6 @@
 By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. Before temporarily stopping nodes for planned maintenance (e.g., upgrading system software), if you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window.
 
-For example, let's say you want to maintain a group of servers, and the nodes running on the servers may be offline for up to 15 minutes as a result. Before shutting down the nodes, you would change the `server.time_until_store_dead` cluster setting as follows:
+For example, let's say you want to maintain a group of servers, and the nodes running on the servers may be offline for up to 15 minutes as a result. Before shutting down the nodes, you might change the `server.time_until_store_dead` cluster setting as follows:
```
a-robinson: For what it's worth, I thought the word "would" works well here. It doesn't really matter, though.

_includes/faq/planned-maintenance.md
```diff
@@ -11,3 +11,9 @@ After completing the maintenance work and [restarting the nodes](start-a-node.ht
 ~~~ sql
 > SET CLUSTER SETTING server.time_until_store_dead = '5m0s';
 ~~~
+
+It's also important to ensure that load balancers don't send client traffic to a node about to be shut down for a longer period or just for a few seconds. If you find that your load balancer's health check needs more time to mark a node as unready, you can increase the `server.shutdown.drain_wait` setting, for example:
```
a-robinson: If you don't mind a suggestion, I'd rephrase this paragraph as:

> It's also important to ensure that load balancers don't send client traffic to a node about to be shut down, even if it will only be down for a few seconds. If you find that your load balancer's health check is not always recognizing a node as unready before the node shuts down, you can increase the `server.shutdown.drain_wait` setting, which tells the node to wait in an unready state for the specified duration. For example:

_includes/faq/planned-maintenance.md
jseldess: I agree. Changed back.

_includes/faq/planned-maintenance.md
jseldess: Done. Also updated the tip in the health endpoint doc.

jseldess:
TFTR, @a-robinson!

PR #2924, opened 2018-04-15
Fix an order presentation example.

### LINKED ISSUE #2922: Description of ORDER BY preservation has invalid query
In this section of the docs: https://www.cockroachlabs.com/docs/stable/query-order.html#order-preservation

The following query is used as an example of when ORDER BY would be preserved even though it's part of a subquery:
`SELECT (1, 2, 3) = (SELECT a.x FROM a ORDER BY a.x);`

However, if I try to run this query in the terminal, I get the error:
`unsupported comparison operator: <tuple{int, int, int}> = <int>`

I think this query should either be removed or replaced with a different one. There is already another example with `ARRAY()` which may be sufficient...

knz:
Noticed by @rytaft.

Fixes #2922.

@rytaft if you can think of a way where a scalar subquery can provide a tuple (I couldn't find one in the end, although I was initially confident that was possible), we could add it here.

### DIFF
--- a/v2.0/query-order.md
+++ b/v2.0/query-order.md
@@ -95,7 +95,7 @@ For example, using a sub-query in scalar context:
 ~~~ sql
 > SELECT ARRAY(SELECT a.x FROM a ORDER BY a.x);
   -- ensures that the array is constructed using the values of a.x in sorted order.
-> SELECT (1, 2, 3) = (SELECT a.x FROM a ORDER BY a.x);
+> SELECT ARRAY[1, 2, 3] = ARRAY(SELECT a.x FROM a ORDER BY a.x);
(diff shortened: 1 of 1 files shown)


rytaft:
:lgtm: Thanks!

I also haven't been able to think of a way for a scalar subquery to provide a tuple. I had one idea which doesn't work either:
```
root@:26257/t> select (select a from abc) = (select a from abc);
pq: more than one row returned by a subquery used as an expression
```

That works in some databases but not in CockroachDB.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2924#-:-LAEbl-d1L7u3c6QotS2:b-yw4z1k)*

knz:
ok this means we can't do better for now. Thanks for checking though.

PR #2949, opened 2018-04-18
READ ONLY transactions

### LINKED ISSUE #2314: sql: support read-only transactions via PostgreSQL-compatible syntax
PR: https://github.com/cockroachdb/cockroach/pull/20547

Adds support for:
- `SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY/READ WRITE;`
- `SET TRANSACTION READ ONLY/READ WRITE;`
- `SET default_transaction_read_only;`
- `SET transaction_read_only;`

lhirata:
Closes #2314.


v2.0/set-vars.md
```diff
@@ -37,14 +37,16 @@ The variable name is case insensitive. The value can be a list of one or more it
...
 | `application_name` | The current application name for statistics collection. | Empty string | Yes |
 | `database` | The [current database](sql-name-resolution.html#current-database). | Database in connection string, or empty if not specified | Yes |
-| `default_transaction_isolation` | The default transaction isolation level for the current session. See [Transaction parameters](transactions.html#transaction-parameters) and [`SET TRANSACTION`](set-transaction.html) for more details. | Settings in connection string, or "`SERIALIZABLE`" if not specified  | Yes |
+| `default_transaction_isolation` | <span class="version-tag">New in v2.0:</span> The default transaction isolation level for the current session. See [Transaction parameters](transactions.html#transaction-parameters) and [`SET TRANSACTION`](set-transaction.html) for more details. | Settings in connection string, or "`SERIALIZABLE`" if not specified  | Yes |
+| `default_transaction_read_only` | The default transaction access mode for the current session. See [`SET TRANSACTION`](set-transaction.html) for more details. | `off` | Yes |
```
jseldess: We might want to add a bit more of an explicit description here, e.g., 

> If set to `on`, only read operations are allowed in transactions; if set to `off`, both read and write operations are allowed in transactions.

Thoughts? If we're more explicit here, we should probably be more explicit in the other places, too.

lhirata:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[v2.0/set-vars.md, line 41 at r1](https://beta.reviewable.io/reviews/cockroachdb/docs/2949#-LAP0uizkKruhxU_KrZj-r1-41:-LAnDNKx2UxHY6mnt_Us:b-lnogku) ([raw file](https://github.com/cockroachdb/docs/blob/611a392429d281490c41ce2a81239e6f0672eaae/v2.0/set-vars.md#L41)):*

We might want to add a bit more of an explicit description here, e.g., 

> If set to `on`, only read operations are allowed in transactions in the current session; if set to `off`, both read and write operations are allowed.

Thoughts? If we're more explicit here, we should probably be more explicit in the other places, too.
</blockquote></details>

Updated. Doesn't hurt to be more explicit.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/2949)*

PR #2950, opened 2018-04-18
Show how to split on a composite primary key

### LINKED ISSUE #2769: SPLIT AT Doc Update
Re: [SPLIT AT](https://cockroachlabs.com/docs//v2.0/split-at.html)

The docs say that SPLIT AT takes a `select statement` but that's not enough info. For example, what would be required to split at a table that has a composite primary key?

According to @jordanlewis we use the syntax `alter table .. split at values ((string, uuid), (string2, uuid2))`

We should make this clear -- I think this may come up when we release partitioning since it uses compound keys.

rmloveland:
Fixes #2769

### DIFF
--- a/v2.0/split-at.md
+++ b/v2.0/split-at.md
@@ -103,6 +103,87 @@ the ranges that store tables or indexes:
 (4 rows)
 ~~~
 
+### Split a Table with a Composite Primary Key
+
+You may want to split a table with a composite primary key (e.g., when working with [partitions](partitioning.html#partition-using-primary-key)).
+
+Given the table
+
+{% include copy-clipboard.html %}
+~~~ sql
+CREATE TABLE t (k1 INT, k2 INT, v INT, w INT, PRIMARY KEY (k1, k2));
+~~~
+
+we can split it at its primary key like so:
+
+{% include copy-clipboard.html %}
+~~~ sql
+ALTER TABLE t SPLIT AT VALUES (5,1), (5,2), (5,3);
+~~~
+
+~~~
++------------+-----------------+
+|    key     |     pretty      |
++------------+-----------------+
+| \xbc898d89 | /Table/52/1/5/1 |
+| \xbc898d8a | /Table/52/1/5/2 |
+| \xbc898d8b | /Table/52/1/5/3 |
++------------+-----------------+
+(3 rows)
+~~~
+
+To see more information about the range splits, run:
+
+{% include copy-clipboard.html %}
+~~~ sql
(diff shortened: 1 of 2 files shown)


nstewart:
Thanks LGTM

RaduBerinde:
Maybe worth mentioning that you can also split at prefix of the PK columns. In this case, you could split at /3 (which means that there is a split before all keys that start with `3`)

rmloveland:
> Maybe worth mentioning that you can also split at prefix of the PK columns. In this case, you could split at /3 (which means that there is a split before all keys that start with 3)

Fixed in latest push - thanks!

v2.0/split-at.md
```diff
@@ -103,6 +103,87 @@ the ranges that store tables or indexes:
...
++-----------+---------+----------+----------+--------------+
+| NULL      | /5/1    |      151 | {2,3,5}  |            5 |
+| /5/1      | /5/2    |      152 | {2,3,5}  |            5 |
+| /5/2      | /5/3    |      153 | {2,3,5}  |            5 |
+| /5/3      | NULL    |      154 | {2,3,5}  |            5 |
++-----------+---------+----------+----------+--------------+
+(4 rows)
+~~~
+
+Alternatively, you could instead split at a prefix of the primary key columns.  For example, to add a split before all keys that start with `3`, run:
```
jseldess: nits: Don't need both `alternatively` and `instead`. I'd remove `instead`. Also, use just one space following the period.

v2.1/split-at.md
```diff
@@ -103,6 +103,87 @@ the ranges that store tables or indexes:
...
++-----------+---------+----------+----------+--------------+
+| NULL      | /5/1    |      151 | {2,3,5}  |            5 |
+| /5/1      | /5/2    |      152 | {2,3,5}  |            5 |
+| /5/2      | /5/3    |      153 | {2,3,5}  |            5 |
+| /5/3      | NULL    |      154 | {2,3,5}  |            5 |
++-----------+---------+----------+----------+--------------+
+(4 rows)
+~~~
+
+Alternatively, you could instead split at a prefix of the primary key columns.  For example, to add a split before all keys that start with `3`, run:
```
jseldess: Same as above.

v2.0/split-at.md
rmloveland: Thanks, fixed.

v2.1/split-at.md
rmloveland: Fixed.

PR #3013, opened 2018-04-23
Remove a known limitation.

knz:
Requested by @andreimatei in https://github.com/cockroachdb/cockroach/pull/23373#issuecomment-383617998.

### DIFF
--- a/v2.1/known-limitations.md
+++ b/v2.1/known-limitations.md
@@ -20,15 +20,6 @@ In the upgrade process, after upgrading all binaries to v2.1, it's recommended t
 
 You can track this known limitation [here](https://github.com/cockroachdb/cockroach/issues/24490).
 
-### Write and update limits for a single statement
-
-A single statement can perform at most at most 64MiB of combined updates. When a statement exceeds these limits, its transaction gets aborted. `INSERT INTO ... SELECT FROM` queries commonly encounter these limits.
-
-If you need to increase these limits, you can update the [cluster-wide setting](cluster-settings.html) `kv.raft.command.max_size`, but note that increasing this setting can affect the memory utilization of nodes in the cluster. For `INSERT INTO .. SELECT FROM` queries in particular, another workaround is to manually page through the data you want to insert using separate transactions.
-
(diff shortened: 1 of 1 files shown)


jseldess:
Thanks, @knz, but it's misleading to remove the limitation before the fix has made it into an official release, don't you think? Instead of deleting the limitation from the page, we tend to add a note that it's been "resolved as of xxx release", for example, https://www.cockroachlabs.com/docs/v1.1/known-limitations.html#maximum-cluster-size. Should we just open a docs issue to do this work in conjunction with the next 2.1 alpha release?

knz:
We could leave this PR open until the relevant 2.1 alpha is released. I'm happy if you take this over. I created this docs PR just because @andreimatei didn't want me (us) to forget about documenting that this limitation is being lifted.

andreimatei:
I thought there was a place where we list the "resolved limitations" since the previous major release, but I can't find it now... @jseldess is there / should there be?

My opinion is that this PR can be merged. I feel that it's not worth being strict about when documentation for unreleased versions is updated.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/3013#-:-LAnef50722plF9HR-kg:bn7u48d)*

andreimatei:
:lgtm:

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/cockroachdb/docs/3013#-:-LAnf1KI-n_I_JyXAeWi:bnfp4nl)*

jseldess:
@andreimatei, the way we've done it up till now is we've added a "Resolved in xxx release" and then, in the known limitations for the following major release, we've remove the limitations entirely. 

I agree that we shouldn't be strict about merging changes for dev releases. It's a discussion topic in tomorrow's docs meeting, actually, so I'll return to this once the team is on the same page here.

jseldess:
The docs team now has a more formalized approach to [documenting across versions](https://github.com/cockroachdb/docs/wiki/Updating-Docs-Across-Versions), which includes updating dev (2.1) docs before releases, if need be. So merging this. Thanks, @knz and @andreimatei.

PR #3018, opened 2018-04-24
document supported formats for AOST timestamps

### LINKED ISSUE #2928: Time interval now supported in BACKUP ... AS OF SYSTEM TIME
This was added in https://github.com/cockroachdb/cockroach/pull/24768. Our docs (https://www.cockroachlabs.com/docs/stable/backup.html#performance) mention that users in production should be running BACKUPs with a timestamp from 10 seconds ago to avoid affecting running transactions but this was previously manual. It can now be something like `BACKUP .. AS OF SYSTEM TIME '-10s'`.

### LINKED ISSUE #3014: AS OF SYSTEM TIME syntax is not fully described
Re: [Simple SELECT Clause](https://cockroachlabs.com/docs//v2.0/select-clause.html)
^ this page does not describe the syntax accepted by AOST. In fact, the syntax is not described at all; examples show a string being used as the timestamp expression. In fact, we accept strings, ints (nanoseconds since unix epoch), decimals (nanoseconds since epoch `.` logical timestamp), and intervals (which are added to the statement timestamp). These are all useful.

mjibson:
Fixes #2928
Fixes #3014


mjibson:
Not sure if there's a doc reviewer who owns this, so please reassign the review if needed.

v1.0/backup.md
```diff
@@ -89,7 +89,7 @@ Only the `root` user can run `BACKUP`.
 | `table_pattern` | The table or [view](views.html) you want to back up. |
 | `name` | The name of the database you want to back up (i.e., create backups of all tables and views in the database).|
 | `destination` | The URL where you want to store the backup.<br/><br/>For information about this URL structure, see [Backup File URLs](#backup-file-urls). |
-| `AS OF SYSTEM TIME timestamp` | Back up data as it existed as of [`timestamp`](timestamp.html). However, the `timestamp` must be more recent than your cluster's last garbage collection (which defaults to occur every 24 hours, but is [configurable per table](configure-replication-zones.html#replication-zone-format)). |
+| `AS OF SYSTEM TIME timestamp` | Back up data as it existed as of `timestamp`. {% include aost-v1.0.md %} However, the `timestamp` must be more recent than your cluster's last garbage collection (which defaults to occur every 24 hours, but is [configurable per table](configure-replication-zones.html#replication-zone-format)). |
```
jseldess: Not sure exactly why, but includes generally (or always?) don't work in basic markdown table cells:

screen shot 2018-04-24 at 10 01 19 am

Unless you can figure out how to get that to work, we have a few other options:

1. Just duplicate the content wherever it needs to be.
2. Create a separate section for `AS OF SYSTEM TIME`, link from the cell to that section, and then use that include in the new section. We do that for the `--locality` flag on `start-a-node.md`, for example. Not sure if there's enough content here to warrant this approach. Your call.

_includes/aost-v1.0.md
```diff
@@ -0,0 +1 @@
+`timestamp` can be an [`int`](int.html) (nanoseconds since the Unix epoch), or string ([`timestamp`](timestamp.html) or [`int`](int.html) of nanoseconds).
```
knz: we usually capitalize type names.

_includes/aost-v2.1.md
```diff
@@ -0,0 +1 @@
+`timestamp` can be an [`int`](int.html) (nanoseconds since the Unix epoch), negative [`interval`](interval.html) that is added to `statement_timestamp()`, or string ([`timestamp`](timestamp.html), [`int`](int.html) of nanoseconds, or negative [`interval`](interval.html)).
```
knz: ditto

jseldess:
@mjibson, would you like me to take over the final edits?

mjibson:
Yes, that'd be helpful.

jseldess:
Made some final tweaks, @mjibson. Sorry for the delay.

PR #3025, opened 2018-04-25
Correct kubernetes resource limit explanations

a-robinson:
To address the corrections provided by https://github.com/cockroachdb/docs/pull/2896#issuecomment-384255765


embano1:
@a-robinson 

> Secondly, it also increases the "Quality of Service" guaranteed by the Kubernetes runtime on Kubernetes versions **1.9** and below

Should be "1.8.x and below" because the QoS eviction order change has been implemented in [v1.9](https://github.com/kubernetes/kubernetes/blob/release-1.9/pkg/kubelet/eviction/helpers.go). Sorry for any confusion.

Michael

a-robinson:
> Should be "1.8.x and below" because the QoS eviction order change has been implemented in v1.9. Sorry for any confusion.

Thanks, fixed 👍

PR #3046, opened 2018-05-01
document IMPORT CSV and BYTES format

### LINKED ISSUE #3044: import: document new bytes format

mjibson:
Fixes #3044

### DIFF
--- a/v2.0/import.md
+++ b/v2.0/import.md
@@ -38,6 +38,19 @@ Your [`IMPORT`](import.html) statement must include a `CREATE TABLE` statement (
 
 We also recommend [all secondary indexes you want to use in the `CREATE TABLE` statement](create-table.html#create-a-table-with-secondary-and-inverted-indexes-new-in-v2-0). It is possible to add secondary indexes later, but it is significantly faster to specify them during import.
 
+### CSV Data
+
+The tabular data to import must be valid [CSV files](https://tools.ietf.org/html/rfc4180), with the caveat that the comma [delimiter](#delimiter) can be set to another single character. In particular:
+
+- Files must be UTF-8 encoded.
+- If the delimiter (`,` by default), a double quote (`"`), newline (`\n`), or carriage return (`\r`) appears in a field, the field must be enclosed by double quotes.
+- If double quotes are used to enclose fields, then a double quote appearing inside a field must be escaped by preceding it with another double quote.  For example:
+  `"aaa","b""bb","ccc"`
+
+CockroachDB-specific requirements:
+
+- If a column is of type [`BYTES`](bytes.html), it can either be a valid UTF-8 string or a [hex-encoded byte literal](sql-constants.html#hexadecimal-encoded-byte-array-literals) beginning with `\x`. For example, a field whose value should be the bytes `1`, `2` would be written as `\x0102`.
+
 ### Object Dependencies
 
 When importing tables, you must be mindful of the following rules because [`IMPORT`](import.html) only creates single tables which must not already exist:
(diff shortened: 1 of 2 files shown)


jseldess:
:lgtm:, with formatting suggestions.

---

---

*[v2.0/import.md, line 41 at r1](https://reviewable.io/reviews/cockroachdb/docs/3046#-LBW8yQkEfyASE-AmDLy:-LBW8yQkEfyASE-AmDLz:b85hlq7) ([raw file](https://github.com/cockroachdb/docs/blob/9b2f1a9aa98e8b6d49dd9edd8f4591233984ccb6/v2.0/import.md#L41)):*
This is great, @mjibdson. Thank you. I'd only suggest presenting this info in bulleted form to make it easier to separate the various points:

> The tabular data to import must be valid [CSV files](https://tools.ietf.org/html/rfc4180), with the caveat that the comma [delimiter](#delimiter) can be set to another single character. Also note the following requirements:
> 
> - Files must be UTF-8 encoded. 
> - If the delimiter, a double quote (`"`), newline (`\n`), or carriage return (`\r`) appears in a field, the field must be surrounded by double quotes. 
> - Double quotes appearing in a field must be escaped with a backslash (`\"`).
> - If a column is of type [`BYTES`](bytes.html), it can either be a valid UTF-8 string, or a [hex-encoded byte literal](sql-constants.html#hexadecimal-encoded-byte-array-literals) beginning with `\x`. For example, a field whose value should be the bytes `1`, `2` would be written as `\x0102`.

I think this is fine as a distinct section, but we could also consider integrating it into the "Requirements" section further up. 

---

*[v2.1/import.md, line 41 at r1](https://reviewable.io/reviews/cockroachdb/docs/3046#-LBWAa0L7n6FpNDMnSgq:-LBWAa0L7n6FpNDMnSgr:b-by9q0f) ([raw file](https://github.com/cockroachdb/docs/blob/9b2f1a9aa98e8b6d49dd9edd8f4591233984ccb6/v2.1/import.md#L41)):*
> ```Markdown
> We also recommend [all secondary indexes you want to use in the `CREATE TABLE` statement](create-table.html#create-a-table-with-secondary-and-inverted-indexes-new-in-v2-0). It is possible to add secondary indexes later, but it is significantly faster to specify them during import.
> 
> ### CSV Data
> ```

See other comment.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3046#-:-LBWAc9A9D8VKHzG0_-g:bxdoo4l)*

v2.0/import.md
```diff
@@ -38,6 +38,19 @@ Your [`IMPORT`](import.html) statement must include a `CREATE TABLE` statement (
...
+The tabular data to import must be valid [CSV files](https://tools.ietf.org/html/rfc4180), with the caveat that the comma [delimiter](#delimiter) can be set to another single character. In particular:
+
+- Files must be UTF-8 encoded.
+- If the delimiter (`,` by default), a double quote (`"`), newline (`\n`), or carriage return (`\r`) appears in a field, the field must be enclosed by double quotes.
+- If double quotes are used to enclose fields, then a double quote appearing inside a field must be escaped by preceding it with another double quote.  For example:
+  `"aaa","b""bb","ccc"`
+
+CockroachDB-specific requirements:
+
+- If a column is of type [`BYTES`](bytes.html), it can either be a valid UTF-8 string, or a [hex-encoded byte literal](sql-constants.html#hexadecimal-encoded-byte-array-literals) beginning with `\x`. For example, a field whose value should be the bytes `1`, `2` would be written as `\x0102`.
```
jseldess: nit: remove comma before `or`.

PR #3052, opened 2018-05-02
Script to auto-generate docs issues from release notes file

jseldess:
@knz and/or @benesch, this is ready for a review.

### DIFF
--- a/scripts/issues-from-release-notes.py
+++ b/scripts/issues-from-release-notes.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+
+"""
+Create docs issues from a release notes file.
+
+Before running, add <!--doc--> to the end of each bullet in the release notes file
+that needs a docs issue.
+
+Help: ./issues-from-release-notes.py --help
+Example use: ./issues-from-release-notes.py --release_notes=v2.0.1.md --milestone=2.0.x
+
+Requirements:
+  - Requests: http://docs.python-requests.org/en/master/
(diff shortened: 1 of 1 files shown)


scripts/issues-from-release-notes.py
```diff
@@ -0,0 +1,65 @@
+import argparse
```
benesch: If you add a line like

```
#!/usr/bin/env python3
```

(a "shebang"), then run

```
chmod +x scripts/issues-from-release-notes.py
```

then users can directly run scripts/issues-from-release-notes.py without needing to think about which Python version to use.

scripts/issues-from-release-notes.py
```diff
@@ -0,0 +1,65 @@
...
+args = parser.parse_args()
+
+release_notes = args.release_notes
+milestone = args.milestone
+
+# Map milestones to their internal IDs.
+if milestone == "2.1":
+    milestone = 8
+if milestone == "2.O.x":
+    milestone = 10
```
benesch: One thing to consider is asking GitHub for milestone title -> id mapping: https://developer.github.com/v3/issues/milestones/

scripts/issues-from-release-notes.py
jseldess: Done. Thanks!

scripts/issues-from-release-notes.py
jseldess: This is a great idea, but I'll think I'll take it up in a follow-up PR.

jseldess:
TFTR, @benesch!

PR #3104, opened 2018-05-05
Create FAQs for numbering problems.

knz:
As discussed on https://github.com/cockroachdb/cockroach/issues/9227 and other related issues.

Jesse please take this over.

### DIFF
--- a/_includes/faq/sequential-numbers.md
+++ b/_includes/faq/sequential-numbers.md
@@ -1,0 +1,5 @@
+As of v2.0, CockroachDB supports [SQL sequences](create-sequence.html).
+
+{{site.data.alerts.callout_info}}Sequences produce <emph>unique</emph> values, however not all values are guaranteed to be produced (e.g., when a transaction is canceled after it consumes a value) and the values may be slightly reordered (e.g., when a transaction that consumes a lower sequence number commits after a transaction that consumes a higher number).{{site.data.alerts.end}}
+
+{{site.data.alerts.callout_info}}For maximum performance, avoid using sequences to generate row IDs or indexed columns. This is because sequence values are logically close to each other and can cause contention on few data ranges during inserts. Instead, prefer <code>UUID</code>  identifiers or integer identifiers generated with <code>unique_rowid()</code>.{{site.data.alerts.end}}
--- a/_includes/faq/sequential-transactions.md
+++ b/_includes/faq/sequential-transactions.md
@@ -1,0 +1,28 @@
+We find that most use cases that appear to ask for total write
+ordering can be solved with other, more distribution-friendly
+solutions instead. For example, CockroachDB's [time travel queries
+(`AS OF SYSTEM
+TIME`)](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/)
+support the following:
+
+- Paginate through all the changes to a table or dataset.
+- Determine the order of changes to data over time.
+- Determine the state of data at some point in the past.
+- Determine the changes to data between two points of time.
+
+Should your application absolutely require it, it is still possible to
+create a strictly monotonic counter in CockroachDB, as follows:
+
+- initially: `CREATE TABLE cnt(val INT PRIMARY KEY); INSERT INTO cnt(val) VALUES(1);`
+- in each transaction: `INSERT INTO cnt(val) SELECT max(val)+1 FROM cnt RETURNING val;`
+
+This will cause all your INSERT transactions to conflict with each
+other and effectively force the transactions to commit one at a time
+throughout the cluster, which in turn guarantees the values generated
+in this way are strictly increasing over time without gaps. The caveat
+is that performance is severely limited as a result.
+
+Please contact us and describe your situation if you find yourself
+interested in this problem. We would be glad to help you find
+alternative solutions and possibly extend CockroachDB to better match
+your needs.
--- a/v2.0/sql-faqs.md
+++ b/v2.0/sql-faqs.md
@@ -14,6 +14,14 @@ Currently, you can bulk insert data with batches of [`INSERT`](insert.html) stat
 
 {% include faq/auto-generate-unique-ids_v1.1.html %}
 
+## How do I generate unique, slowly increasing sequential numbers in CockroachDB?
+
+{% include faq/sequential-numbers.md %}
+
+## How do I totally order writes to a table over time in CockroachDB?
+
+{% include faq/sequential-transactions.md %}
+
 ## How do I get the last ID/SERIAL value inserted into a table?
 
 There’s no function in CockroachDB for returning last inserted values, but you can use the [`RETURNING` clause](insert.html#insert-and-return-values) of the `INSERT` statement.
--- a/v2.1/sql-faqs.md
+++ b/v2.1/sql-faqs.md
@@ -14,6 +14,14 @@ Currently, you can bulk insert data with batches of [`INSERT`](insert.html) stat
 
 {% include faq/auto-generate-unique-ids_v1.1.html %}
 
+## How do I generate unique, slowly increasing sequential numbers in CockroachDB?
+
+{% include faq/sequential-numbers.md %}
+
+## How do I totally order writes to a table over time in CockroachDB?
+
+{% include faq/sequential-transactions.md %}
+
 ## How do I get the last ID/SERIAL value inserted into a table?
 
 There’s no function in CockroachDB for returning last inserted values, but you can use the [`RETURNING` clause](insert.html#insert-and-return-values) of the `INSERT` statement.


bdarnell:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[_includes/faq/sequential-numbers.md, line 5 at r1](https://reviewable.io/reviews/cockroachdb/docs/3104#-LBrREig6_TK8wpcWymM:-LBrREig6_TK8wpcWymN:bb7o9jq) ([raw file](https://github.com/cockroachdb/docs/blob/15ab607cae1dd2f623e6f21b49ccd09b0d52214a/_includes/faq/sequential-numbers.md#L5)):*
> ```Markdown
> {{site.data.alerts.callout_info}}Sequences produce <emph>unique</emph> values, however not all values are guaranteed to be produced (e.g., when a transaction is canceled after it consumes a value) and the values may be slightly reordered (e.g., when a transaction that consumes a lower sequence number commits after a transaction that consumes a higher number).{{site.data.alerts.end}}
> 
> {{site.data.alerts.callout_info}}For maximum performance, avoid using sequences to generate row IDs or indexed columns. This is because sequence values are logically close to each other and can cause contention on few data ranges during inserts. Instead, prefer <code>UUID</code>  identifiers or integer identifiers generated with <code>unique_rowid()</code>.{{site.data.alerts.end}}
> ```

`unique_rowid()` is still mostly sequential - it will perform more like sequences than like uuid with respect to contention. The main difference between `unique_rowid()` and sequences is that sequences produce smaller numbers (but generating them is much slower).  

So we should recommend that if you need a roughly-ordered id, use `unique_rowid()` unless you need the numbers to be small. And unless you specifically need a roughly-ordered id, you're probably better off using a UUID (it's important to emphasize this point to wean people off the habits of sequential IDs from other databases). 

---

*[v2.0/sql-faqs.md, line 21 at r1](https://reviewable.io/reviews/cockroachdb/docs/3104#-LBrVI_TDUwH2YniVa--:-LBrVI_TDUwH2YniVa-0:b-yevjs8) ([raw file](https://github.com/cockroachdb/docs/blob/15ab607cae1dd2f623e6f21b49ccd09b0d52214a/v2.0/sql-faqs.md#L21)):*
> ```Markdown
> {% include faq/sequential-numbers.md %}
> 
> ## How do I totally order writes to a table over time in CockroachDB?
> ```

I don't think "totally order" is the right term to use here. All rows in CRDB are totally ordered in the [mathematical sense](https://en.wikipedia.org/wiki/Total_order) . The question here is how to make that total order correspond with insertion order. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104)*

knz:
Amended based on Ben's suggestion. Also added a comparison table. RFAL.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[_includes/faq/sequential-numbers.md, line 5 at r1](https://reviewable.io/reviews/cockroachdb/docs/3104#-LBrREig6_TK8wpcWymM:-LC0QVcK7L9YLgnHMKJX:by6n0vq) ([raw file](https://github.com/cockroachdb/docs/blob/15ab607cae1dd2f623e6f21b49ccd09b0d52214a/_includes/faq/sequential-numbers.md#L5)):*

`unique_rowid()` is still mostly sequential - it will perform more like sequences than like uuid with respect to contention. The main difference between `unique_rowid()` and sequences is that sequences produce smaller numbers (but generating them is much slower).  

So we should recommend that if you need a roughly-ordered id, use `unique_rowid()` unless you need the numbers to be small. And unless you specifically need a roughly-ordered id, you're probably better off using a UUID (it's important to emphasize this point to wean people off the habits of sequential IDs from other databases). 
</blockquote></details>

Thanks for reminding me. Amended.

---

*[v2.0/sql-faqs.md, line 21 at r1](https://reviewable.io/reviews/cockroachdb/docs/3104#-LBrVI_TDUwH2YniVa--:-LC0Qv7GBgjvwIwoEWUt:b-us0tsb) ([raw file](https://github.com/cockroachdb/docs/blob/15ab607cae1dd2f623e6f21b49ccd09b0d52214a/v2.0/sql-faqs.md#L21)):*

I don't think "totally order" is the right term to use here. All rows in CRDB are totally ordered in the [mathematical sense](https://en.wikipedia.org/wiki/Total_order) . The question here is how to make that total order correspond with insertion order. 
</blockquote></details>

Yes, you're right. Rephrased.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104#-:-LC0Qwt64LxnV1P-z-Pf:b-rw1shi)*

bdarnell:
:lgtm:

---

Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[_includes/faq/differences-between-numberings.md, line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19CnFDFQE2I_GcJty:-LC19CnFDFQE2I_GcJtz:b9pzuev) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/_includes/faq/differences-between-numberings.md#L7)):*
> ```Markdown
> | Value distribution             | Uniformly distributed (128 bits)        | Contains time and space (node ID) components | Dense, small values            |
> | Performance cost at generation | Small, scalable                         | Small, scalable                              | Variable, can cause contention |
> | Locality                       | Maximally distributed, least contention | Somewhat local, may cause INSERT contention  | Highly local, most INSERT contention |
> ```

The locality issue is not about insert contention: unique_rowid values are extremely unlikely to directly contend with each other. The difference between unique_rowid and UUID is *parallelism*: looking up values by a UUID can make use of many nodes because the values will be spread across many ranges. Typical queries by a more time-ordered id will create more hotspots. 

---

*[v2.0/sql-faqs.md, line 37 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19wvm9YjrbbZ87bzg:-LC19wvm9YjrbbZ87bzh:b-3co6bi) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/v2.0/sql-faqs.md#L37)):*
> ```Markdown
> - [On the Way to Better SQL Joins](https://www.cockroachlabs.com/blog/better-sql-joins-in-cockroachdb/)
> 
> ## How do I get the last ID/SERIAL value inserted into a table?
> ```

Add this to the 2.1 docs too. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104#-:-LC1A-fb4Z7giOA0hm6W:bnfp4nl)*

bdarnell:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[_includes/faq/differences-between-numberings.md, line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19CnFDFQE2I_GcJty:-LC4delf6AaTcnaaGUJI:b6h15sn) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/_includes/faq/differences-between-numberings.md#L7)):*

I reworded, can you have another look?
</blockquote></details>

For "data locality" and "read performance", sequences and unique_rowid are equivalent. We don't need to make a distinction between them except for insert performance. 

---

*[_includes/faq/differences-between-numberings.md, line 9 at r3](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC4drU_3Ek6S8Djml9m:-LC4drU_3Ek6S8Djml9n:b-bhejb7) ([raw file](https://github.com/cockroachdb/docs/blob/66983e1cbba45eaaa3e79516882f53c787694270/_includes/faq/differences-between-numberings.md#L9)):*
> ```Markdown
> | Value distribution                  | Uniformly distributed (128 bits)        | Contains time and space (node ID) components  | Dense, small values            |
> | Data locality                       | Maximally distributed                   | Values generated close in time are co-located | Highly local                   |
> | INSERT performance when used as key | Highest                                 | Lower for values generated close in time      | Slowest                        |
> ```

Break "insert performance" into latency and throughput buckets. For latency, both uuid and unique_rowid have good latency while sequences are slower. For throughput, UUID has best throughput while unique_rowid and sequences are limited. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104)*

bdarnell:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[_includes/faq/differences-between-numberings.md, line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19CnFDFQE2I_GcJty:-LC4nKd37tcou7dA0cPG:bnnalng) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/_includes/faq/differences-between-numberings.md#L7)):*

Well I don't agree with that. Two rowids generated a day apart will have values far apart. Sequences are guaranteed to be close to each other. Or are you saying that the "value distance" due to time distance for rowids does not matter?
</blockquote></details>

The "value distance" only matters if there are other keys between them. Each key will (usually) be adjacent to the one generated before it, whether the difference between the keys is 1 or 1000. 

---

*[_includes/faq/differences-between-numberings.md, line 9 at r3](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC4drU_3Ek6S8Djml9m:-LC4nbYiAh4OyLW4Tx6h:bhbue45) ([raw file](https://github.com/cockroachdb/docs/blob/66983e1cbba45eaaa3e79516882f53c787694270/_includes/faq/differences-between-numberings.md#L9)):*

The latency increases under contention, doesn't it?
</blockquote></details>

Yes. Under low traffic, UUID and unique_rowid will have the same latency. As traffic increases, unique_rowid latency will degrade while UUID insertion latency will stay the same. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104)*

knz:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[_includes/faq/differences-between-numberings.md, line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19CnFDFQE2I_GcJty:-LC4pmLt0tCkLXcIPnnh:b51i0jx) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/_includes/faq/differences-between-numberings.md#L7)):*

The "value distance" only matters if there are other keys between them. Each key will (usually) be adjacent to the one generated before it, whether the difference between the keys is 1 or 1000. 
</blockquote></details>

Check. Now I get it (I think).

---

*[_includes/faq/differences-between-numberings.md, line 9 at r3](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC4drU_3Ek6S8Djml9m:-LC4qep4BpEENXx-Y5sF:b-r9ye9a) ([raw file](https://github.com/cockroachdb/docs/blob/66983e1cbba45eaaa3e79516882f53c787694270/_includes/faq/differences-between-numberings.md#L9)):*

Yes. Under low traffic, UUID and unique_rowid will have the same latency. As traffic increases, unique_rowid latency will degrade while UUID insertion latency will stay the same. 
</blockquote></details>

Ok updated.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104)*

bdarnell:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[_includes/faq/differences-between-numberings.md, line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/3104#-LC19CnFDFQE2I_GcJty:-LC4sgts37v8xXDBOqgB:b-su1hfs) ([raw file](https://github.com/cockroachdb/docs/blob/3d26424a01d61b027601fc8df447a7bbda973c0e/_includes/faq/differences-between-numberings.md#L7)):*

Check. Now I get it (I think).
</blockquote></details>

For the data locality line, I think both unique_rowid and sequences are "highly local". There's not a meaningful difference between the two. 

Similarly, what difference is "somewhat time-ordered" vs "highly time-ordered" supposed to indicate? I'd say that both are equally time-ordered. I think the amount of concurrency required to have out-of-order insertions is similar for both. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104)*

bdarnell:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3104#-:-LCA8OzR9YBLeaRN9yyO:bnfp4nl)*

v2.0/sql-faqs.md
```diff
@@ -14,6 +14,21 @@ Currently, you can bulk insert data with batches of [`INSERT`](insert.html) stat
 
 {% include faq/auto-generate-unique-ids_v1.1.html %}
 
+## How do I generate unique, slowly increasing sequential numbers in CockroachDB?
+
+{% include faq/sequential-numbers.md %}
+
+~~~
```
jseldess: Remove this extra empty code block.

v2.1/sql-faqs.md
```diff
@@ -14,6 +14,21 @@ Currently, you can bulk insert data with batches of [`INSERT`](insert.html) stat
 
 {% include faq/auto-generate-unique-ids_v1.1.html %}
 
+## How do I generate unique, slowly increasing sequential numbers in CockroachDB?
+
+{% include faq/sequential-numbers.md %}
+
+~~~
```
jseldess: Remove this extra empty code block.

_includes/faq/sequential-numbers.md
```diff
@@ -0,0 +1,22 @@
+Sequential numbers can be generated in CockroachDB using the built-in
+function `unique_rowid()` or using [SQL sequences](create-sequence.html).
+
+{{site.data.alerts.callout_info}}Unless you need roughly-ordered
```
jseldess: Need to use `<code></code>` instead of backticks within a callout.

_includes/faq/sequential-transactions.md
```diff
@@ -0,0 +1,32 @@
+We find that most use cases that appear to ask for a strong time-based
+write ordering can be solved with other, more distribution-friendly
+solutions instead. For example, CockroachDB's [time travel queries
+(`AS OF SYSTEM
+TIME`)](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/)
```
jseldess: Once https://github.com/cockroachdb/docs/pull/3018/files lands, we'll have a specific doc page to link to here, which I think it preferable to the blog post.

_includes/faq/sequential-transactions.md
```diff
@@ -0,0 +1,32 @@
...
+in the previous FAQ entries, also provide an approximate time ordering.
+
+Should your application absolutely require it, it is still possible to
+create a strictly monotonic counter in CockroachDB that increases over
+time, as follows:
+
+- initially: `CREATE TABLE cnt(val INT PRIMARY KEY); INSERT INTO cnt(val) VALUES(1);`
+- in each transaction: `INSERT INTO cnt(val) SELECT max(val)+1 FROM cnt RETURNING val;`
+
+This will cause all your INSERT transactions to conflict with each
```
jseldess: We should code-format `INSERT` here.

PR #3133, opened 2018-05-10
2.1 sidenav overhaul

jseldess:


jseldess:
@j-low, I'd love your help making some adjustments to the css:

1.  The new sidenav styling should only apply when `is_top_level` is set to `true`. Currently, although I've only configured the 2.1 sidenav with that setting, the styling seems to be getting applied to all versions of the docs.
2. We still need to style the deepest level of the sidenav, for example, http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/4beedcf77a27276e22ddc31a0cbef80d8c2a7794/dev/start-a-local-cluster.html and http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/4beedcf77a27276e22ddc31a0cbef80d8c2a7794/dev/deploy-cockroachdb-on-premises.html.

jseldess:
@lhirata, @Amruta-Ranade, @rmloveland, this is ready for your review. Direct link: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/d2529e45823acf9660ee3e91d780500afb6e07bc/dev/.

I've only applied the sidenav reorg to the 2.1 docs. Once we're comfortable, we can publish and then do some automated surveying of users over the next few months. Based on feedback, we can make further changes.

Amruta-Ranade:
LGTM 🎉 
A random thought: Should we rename "Maintain" to "Manage"?

jseldess:
Thanks, all. @Amruta-Ranade, we actually switched it from manage to maintain at some point in the past. Not sure which is better. We can think about that some more, too.

PR #3152, opened 2018-05-16
Style Guide Updates

lhirata:
Updates based on Rich's feedback

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -35,6 +35,8 @@ Included in this guide:
 
 CockroachDB docs should be helpful, humble, positive, and friendly. To achieve this, all docs should be factual and free from hyperbolic language.
 
+To expand upon the idea of "free from hyperbolic language", avoid the use of the word "simple" (along with "just", "easily", "actually", etc.) since it's not really possible to tell what might be easy or hard for the user. Something you think is simple may be challenging for them.
+
 Other general guidance about language and tone:
 
 - For [reference and general task-based docs](#reference-and-task-based-docs), use the second-person imperative present tense, also known as "[imperative mood](https://en.wikipedia.org/wiki/Imperative_mood)." These docs should be straightforward and conventional.
@@ -198,7 +200,7 @@ Our docs use three classes of highlighted text:
 - [Notes](#notes)
 - [Warnings](#warnings)
 
-The text of notes, warnings, and tips must be formatted in HTML instead of Markdown/Kramdown.
+The text of notes, warnings, and tips must be formatted in HTML instead of Markdown.
 
 #### Tips
 
@@ -216,7 +218,7 @@ To insert a tip, use the following code:
 
 Use notes to call attention to a piece of clarifying information; this information should not be crucial to accomplishing the task in the document.
 
-For example, you might use a note to let users know that the `DELETE` statement only deletes rows and that to delete columns you must use `ALTER TABLE`. This helps clarify `DELETE`'s purpose and point misguided users to the right place.
+For example, you might use a note to let users know that the `DELETE` statement only deletes rows and that to delete columns you must use `ALTER TABLE`. This helps clarify `DELETE`'s purpose and point users to the right place.
 
 To insert a note, use the following code:
 
@@ -265,7 +267,7 @@ SQL code samples are broken into two sections: commands and responses.
 
 **Copy to Clipboard**
 
-Many of our code blocks are written so users can copy and paste them directly into a terminal. To make that easier, add the **Copy to Clipboard** button by placing `{include copy-clipboard.html}` on the line directly preceding the code block, for example:
+Many of our code blocks are written so users can copy and paste them directly into a terminal. To make that easier, add the **Copy to Clipboard** button by placing `{% include copy-clipboard.html %}` on the line directly preceding the code block, for example:
 
 ```
 {% include copy-clipboard.html %}
--- a/STYLE.md
+++ b/STYLE.md
@@ -35,8 +35,6 @@ Included in this guide:
 
 CockroachDB docs should be helpful, humble, positive, and friendly. To achieve this, all docs should be factual and free from hyperbolic language.
 
-To expand upon the idea of "free from hyperbolic language", avoid the use of the word "simple" (along with "just", "easily", "actually", etc.) since it's not really possible to tell what might be easy or hard for the user. Something you think is simple may be challenging for them.
-
 Other general guidance about language and tone:
 
 - For [reference and general task-based docs](#reference-and-task-based-docs), use the second-person imperative present tense, also known as "[imperative mood](https://en.wikipedia.org/wiki/Imperative_mood)." These docs should be straightforward and conventional.
@@ -54,11 +52,13 @@ Other general guidance about language and tone:
 
     **Example:** `table name`: The name of the table you want to create audit logs for.
 
-- Use contractions to simplify language.
+- To expand upon the idea of "free from hyperbolic language", avoid the use of the word "simple" (along with "just", "easily", "actually", etc.) since it's not really possible to tell what might be easy or hard for the user. Something you think is simple may be challenging for them.
 
-    **Example:** A primary key can't be changed using `ALTER TABLE`.
+- Use contractions to simplify language, but not in cases where a clear directive or prohibition is being given (e.g., `do not` / `cannot` / `should not` instead of `don't` / `can't` / `shouldn't`).
 
-    **Example:** If you leave versioned binaries on your servers, you don't need to do anything.
+    **Example:** A primary key cannot be changed using `ALTER TABLE`.
+
+    **Example:** If you leave versioned binaries on your servers, you do not need to do anything.
 
 ## Capitalization and Punctuation
 
@@ -172,6 +172,8 @@ Do not use underlined text in CockroachDB docs. If it seems beneficial to emphas
 ### Links
 
 Whenever a CockroachDB feature is referenced, a link to the relevant documentation should be provided. Links to external resources can also be provided, but only if the resource is vetted and no CockroachDB documentation covers the topic.
+
+Use Markdown's reference-style links when several parts of the same page refer to the same target URLs (e.g., [Release Notes](https://raw.githubusercontent.com/cockroachdb/docs/master/releases/v2.1.0-alpha.20180507.md)).
 
 Link capitalization can be either title- or sentence-case:
 
@@ -316,7 +318,11 @@ Examples:
 
 ### Tables
 
-Use tables to display structured information in an easy-to-read format.
+Use tables to display structured information in an easy-to-read format. There are two types of tables we use: [Markdown](#markdown) and [HTML](#html).
+
+#### Markdown
+
+If table formatting can be kept simple (e.g., basic text formatting and using `<br>` tags for paragraph breaks), create a table using Markdown. This is the preferred table format.
 
 To create a table, use pipes (`|`) between columns and at least 3 dashes (`-`) separating the header cells from the body cells. A return denotes the start of the next row. The text within each column does not need to align in order to be rendered correctly, and you can inline Markdown or HTML.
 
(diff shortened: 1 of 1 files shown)


lhirata:

---

*[STYLE.md, line 38 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeUjQO6GB7qk1NG7Bc:-LCeUjQO6GB7qk1NG7Bd:b3vqkj4) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L38)):*
> ```Markdown
> CockroachDB docs should be helpful, humble, positive, and friendly. To achieve this, all docs should be factual and free from hyperbolic language.
> 
> To expand upon the idea of "free from hyperbolic language", avoid the use of the word "simple" (along with "just", "easily", "actually", etc.) since it's not really possible to tell what might be easy or hard for the user. Something you think is simple may be challenging for them.
> ```

@rmloveland commented: "This is more of an overall style thing and does not relate to examples in particular, but I think we should avoid the use of the word "simple" (along with "just", "easily", "actually", and friends). It's not really possible to tell what might be easy or hard for the reader. Something we think is simple may be pretty challenging for them."

Added Rich's guidance to the Language and Tone section. I think this is a really good observation, and falls in line with being "humble" and "free from hyperbolic language.

---

*[STYLE.md, line 59 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVxCN7WUWDoU4R6g2:-LCeVxCN7WUWDoU4R6g3:bch918) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L59)):*
> ```Markdown
> - Use contractions to simplify language.
> 
>     **Example:** A primary key can't be changed using `ALTER TABLE`.
> ```

@rmloveland: "I think in cases like this we should go the other way and not use contractions.

By "cases like this" I mean that I think in the case of "can't" it is clearer to say "a primary key can not be changed using ALTER TABLE".

This is to aid comprehension for readers who land on a page via search results. They are in a hurry and may skim-read, so it's easy for them to miss the difference between "can" and "can't" and other contractions because they're only one or so letters apart. There may also be some benefit for people for whom English is not their primary language."

--- 

Instead of saying "use contractions to simplify", do you think we should modify our stance and say, "do not use contractions"? I can see an argument for both:

1. "Use contractions to simplify language, but in important steps/scenarios, do not use contractions." Leaving it up to the discretion of the writer could make this guidance confusing.
2. "Do not use contractions." This is an easier piece of guidance to remember, but do we lose the "friendliness" by not allowing contractions?

---

*[STYLE.md, line 172 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVTtx9K2hdh6Ep0JS:-LCeVTtx9K2hdh6Ep0JT:bw83a18) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L172)):*
> ```Markdown
> Do not use underlined text in CockroachDB docs. If it seems beneficial to emphasize a word or phrase, use [bold](#bold).
> 
> ### Links
> ```

@rmloveland: "Is it going to be legal to use Markdown's reference-style links? They can be really handy when several different parts of a page refer to the same target URLs. I'm pretty OK with not allowing them too, if we prefer to value consistency."

Ah, interesting! I've never used reference-style links before. These are pretty cool. I don't think we have any currently. @Amruta-Ranade / @jseldess, what do you think about reference links? My initial thoughts are that we should keep links consistent. 

---

*[STYLE.md, line 270 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeUy9O7New5f-iKKY8:-LCeUy9O7New5f-iKKY9:b-w5zzma) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L270)):*
> ```Markdown
> **Copy to Clipboard**
> 
> Many of our code blocks are written so users can copy and paste them directly into a terminal. To make that easier, add the **Copy to Clipboard** button by placing `{% include copy-clipboard.html %}` on the line directly preceding the code block, for example:
> ```

@rmloveland: "Nit: in Copy to Clipboard, it looks like the inline include is not formatting correctly. I think it can be removed since the example just below shows how to use it."

Fixed the inline - I think it was just missing the %'s, unless you're seeing something else in Firefox?

---

*[STYLE.md, line 317 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeW7rq82KLDSZTs8wN:-LCeW7rr3U1_-c5CXI62:b-efpwtt) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L317)):*
> ```Markdown
> - [`CREATE TABLE`](https://www.cockroachlabs.com/docs/stable/create-table.html)
> 
> ### Tables
> ```

@rmloveland: "Do we have criteria for when we drop out of Markdown and just write an HTML table? For example, if there are strict formatting requirements or too much complexity with inline BR tags creating extremely long lines (among other issues), I sometimes prefer to write tables in HTML."

I originally had "move to HTML if the table starts getting hairy"-type of guidance, but we decided to remove it for now because we haven't been using HTML tables often. This could also not be an issue once we move to Hugo (right?), so not sure if it's worth extrapolating right now. Do you feel strongly about it? If so, we can work on this section. 

---

*[STYLE.md, line 419 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeWFcf7kamXg2ILEps:-LCeWFcf7kamXg2ILEpt:b-bds3xj) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L419)):*
> ```Markdown
> ```
> 
> ### Images
> ```

@rmloveland: "I think it would be good if we could add that CSS to our docs site stylesheet rather than adding it inline (maybe it's already there and I missed it? If so, sorry.)"

This is a good idea, but I'm not sure how to do this correctly. Can you add or should we ask Joe?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3152)*

jseldess:
*[STYLE.md, line 419 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeWFcf7kamXg2ILEps:-LCjgitO99xZ36enSqsD:b-etsps0) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L419)):*

@rmloveland: "I think it would be good if we could add that CSS to our docs site stylesheet rather than adding it inline (maybe it's already there and I missed it? If so, sorry.)"

This is a good idea, but I'm not sure how to do this correctly. Can you add or should we ask Joe?
</blockquote></details>

This could be a simple css rule, but we would probably need to add an id or class to all images to which we want this rule applied. Or we'd need to leave the inline styling for images that we do not want styled in a uniform way. 

@rmloveland, would you mind creating a docs issue for this and assigning it to Joe?

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3152)*

rmloveland:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[STYLE.md, line 59 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVxCN7WUWDoU4R6g2:-LCjhPsw15RKgBE1TNwM:bi85bnb) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L59)):*

After chatting with Rich in another channel, I think we could allow contractions to simplify language but not in cases where a clear directive or prohibition is being given. So maybe `do not`/`cannot`/`should not` instead of `don't`/`can't`/`shouldn't`.
</blockquote></details>

I like Jesse's formulation of it

---

*[STYLE.md, line 172 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVTtx9K2hdh6Ep0JS:-LCjh9Pl56DA5pENn_1-:b-zac2rb) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L172)):*
> By reference-style links, do we mean
yep!

not super opinionated on it though!

---

*[STYLE.md, line 317 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeW7rq82KLDSZTs8wN:-LCjhfrF6vBE--k09bBC:b45d1b8) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L317)):*

My feeling is we should allow html tables in specific cases, and we should probably define those cases pretty well. Initial thoughts:

- If table formatting can be kept simple, we should prefer that (e.g., basic text formatting and using `<br>` tags for paragraph breaks).
- If it's necessary to include more complex formatting (bullets, code blocks), we could accept writing the whole table in html. In these cases, though, it may be even harder for other writers to parse and maintain. 

We should probably also state what types of complex formatting are ok. I think bulleted lists are the most common. Maybe code blocks (though I don't know what they'd do to the rendering of the table). Anything else?
</blockquote></details>

Another option could be: Markdown tables with excessively long lines due to using multiple `br` tags to simulate paragraphs

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3152)*

jseldess:
*[STYLE.md, line 317 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeW7rq82KLDSZTs8wN:-LCjzZ3R6YgYksPxQ1P1:b-7cue5a) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L317)):*

Another option could be: Markdown tables with excessively long lines due to using multiple `br` tags to simulate paragraphs
</blockquote></details>

Definitely

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3152)*

lhirata:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

*[STYLE.md, line 38 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeUjQO6GB7qk1NG7Bc:-LD1gh4H1vs1lhWcZfV8:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L38)):*

I like this addition, but can you make it another bullet instead of an intro paragraph? I think this might be the seed for "terminology" guidance that we can add over time.
</blockquote></details>

Done.

---

*[STYLE.md, line 59 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVxCN7WUWDoU4R6g2:-LD1hBOW9XaYKCw5CQnG:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L59)):*

I like Jesse's formulation of it
</blockquote></details>

Done.

---

*[STYLE.md, line 172 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeVTtx9K2hdh6Ep0JS:-LD1i0Ej7cNo4NAofKsy:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L172)):*

> By reference-style links, do we mean
yep!

not super opinionated on it though!
</blockquote></details>

Done.

---

*[STYLE.md, line 317 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeW7rq82KLDSZTs8wN:-LD2wOfJ5ahyuPVIPjVJ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L317)):*

Definitely
</blockquote></details>

Done.

---

*[STYLE.md, line 419 at r1](https://reviewable.io/reviews/cockroachdb/docs/3152#-LCeWFcf7kamXg2ILEps:-LD2x177BlpKqYDG4mLl:bi52yrt) ([raw file](https://github.com/cockroachdb/docs/blob/110fb71b52ebbb6a30ac491f43d5ee90414897c3/STYLE.md#L419)):*

This could be a simple css rule, but we would probably need to add an id or class to all images to which we want this rule applied. Or we'd need to leave the inline styling for images that we do not want styled in a uniform way. 

@rmloveland, would you mind creating a docs issue for this and assigning it to Joe?
</blockquote></details>

Added issue: https://github.com/cockroachdb/docs/issues/3175

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3152)*

Amruta-Ranade:
LGTM 🎉

jseldess:
LGTM!

PR #3172, opened 2018-05-18
mention the SEQUENCE feature with the SERIAL type

dchenk:
The docs here explain that "Auto-Incrementing Is Not Always Sequential." Now that the SEQUENCE feature has been added to CRDB, I think it should be mentioned in this section as it achieves precisely what AUTO_INCREMENT does in MySQL and SERIAL in PostgreSQL.


jseldess:
So sorry for the delay, @dchenk. This LGTM, but I've waited too long and now can't get it merge. So I've moved this work into a [separate PR](https://github.com/cockroachdb/docs/pull/3310). Thanks for this improvement!

Issue #3164, opened 2018-05-17
hibernate example fails at `gradle run`


rmloveland:
Gradle version is 4.7 installed via `brew install gradle` on OS X

Java version is 

    java version "10.0.1" 2018-04-17
    Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
    Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)

Following the instructions at [Build a Java app with Hibernate](https://www.cockroachlabs.com/docs/dev/build-a-java-app-with-cockroachdb-hibernate.html) does not work on my machine (tm):

    $ gradle run --stacktrace
    > Task :compileJava UP-TO-DATE
    > Task :processResources UP-TO-DATE
    > Task :classes UP-TO-DATE

    > Task :run FAILED
    Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
        at org.hibernate.boot.spi.XmlMappingBinderAccess.<init>(XmlMappingBinderAccess.java:43)
        at org.hibernate.boot.MetadataSources.<init>(MetadataSources.java:87)
        at org.hibernate.cfg.Configuration.<init>(Configuration.java:123)
        at org.hibernate.cfg.Configuration.<init>(Configuration.java:118)
        at com.cockroachlabs.Sample.<clinit>(Sample.java:16)
    Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
        ... 5 more

    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':run'.
    > Process 'command '/Library/Java/JavaVirtualMachines/jdk-10.0.1.jdk/Contents/Home/bin/java'' finished with non-zero exit value 1

    * Try:
    Run with --info or --debug option to get more log output. Run with --scan to get full insights.

    * Exception is:
    org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':run'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:103)
[...]

PR #3176, opened 2018-05-21
Note that Java version must be <= 9

rmloveland:
Examples don't work with Java 10 due to ... reasons.

Related:
https://github.com/cockroachdb/docs/issues/3164

tl;dr from above: Suggestion was to downgrade to 9 as 10 is breaking lots of code due to the introduction of "modules" in v10.

I think we should reassess updating the examples once Java 10/11 are more widespread (?).  Hopefully adding these notes will help at least some users avoid a frustrating experience until then.


jordanlewis:
Thanks. I think the examples mostly work with Java 10. There's just something wrong with our dependency setup. If people use the examples as a code reference, things should still work - but I agree we should revisit this as Java 10 gets more popular.

rmloveland:
@jordanlewis sorry, I just want to make sure I understand: do you think we should re-word this note to say something about Java 10 dependency mgmt?  Or keep it roughly as is, for now?

jordanlewis:
Happy to see it kept as-is - that was just a note for you.

PR #3179, opened 2018-05-22
Do not mention negative AOST intervals for v1.1 and v2.0

knz:
Fixing a small mistake I made when copying the page over in the previous versions.


jseldess:
Thanks, @knz, but @mjibson already made this change in https://github.com/cockroachdb/docs/pull/3182.

PR #3189, opened 2018-05-24
Add drift script to docs site

jseldess:


jseldess:
I update all drift playbook targeting to exclude docs pages, except the 1 playbook for surveying users about our 2.1 sidenav. If I didn't do something right and drift shows up on docs where we don't want it, I'll revert this change.

PR #3195, opened 2018-05-28
Initial draft for Admin UI user auth changes

### LINKED ISSUE #1582: User authentication in the Admin UI
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwINAgXA5on9pr2T/recXCWr6wysvhsGN7
PM: @piyush-singh 
Eng Partner: 

RFC: https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/web_session_login.md
PRs: 
- https://github.com/cockroachdb/cockroach/pull/17343

Amruta-Ranade:
Closes #1582 

Initial draft for the Admin UI user auth changes. I have not tested the procedure yet, but opening the PR for structural review. @jseldess

### DIFF
--- a/_includes/sidebar-data-v2.1.json
+++ b/_includes/sidebar-data-v2.1.json
@@ -1545,7 +1545,7 @@
             ]
           },
           {
-            "title": "Enable User Authentication",
+            "title": "Secure the Admin UI",
             "urls": [
               "/${VERSION}/admin-ui-user-authentication.html"
             ]
--- a/_includes/sidebar-data-v2.1.json
+++ b/_includes/sidebar-data-v2.1.json
@@ -1545,6 +1545,12 @@
             ]
           },
           {
+            "title": "Enable User Authentication",
+            "urls": [
+              "/${VERSION}/admin-ui-user-authentication.html"
+            ]
+          },
+          {
             "title": "Overview Dashboard",
             "urls": [
               "/${VERSION}/admin-ui-overview-dashboard.html"
--- a/v2.1/admin-ui-access-and-navigate.md
+++ b/v2.1/admin-ui-access-and-navigate.md
@@ -8,6 +8,112 @@ toc: false
 
 ## Access the Admin UI
 
+By default, CockroachDB allows all users to access and view the Admin UI. For added security, you can choose to enable user authentication so that only authorized users can access and view the Admin UI.
+
+<div class="filters filters-big clearfix">
+  <button class="filter-button" data-scope="with-user-auth"><strong>Enable user authentication</strong></button>
+  <button class="filter-button" data-scope="without-user-auth"><strong>Ignore user authentication</strong></button>
+</div>
+
+<section class="filter-content" markdown="1" data-scope="with-user-auth">
+
+You can enable user authentication only for secure clusters. To enable user authentication:
+
+### Step 1. Create security certificates
+
+You can use either `cockroach cert` commands or [`openssl` commands](create-security-certificates-openssl.html) to generate security certificates. This section features the `cockroach cert` commands.
+
+~~~ shell
+# Create a certs directory and safe directory for the CA key.
+# If using the default certificate directory (`${HOME}/.cockroach-certs`), make sure it is empty.
+$ mkdir certs
+$ mkdir my-safe-directory
+
+# Create the CA key pair:
+$ cockroach cert create-ca \
+--certs-dir=certs \
+--ca-key=my-safe-directory/ca.key
+
+# Create a client key pair for the root user:
+$ cockroach cert create-client \
+root \
+--certs-dir=certs \
+--ca-key=my-safe-directory/ca.key
+
+# Create a key pair for the nodes:
+$ cockroach cert create-node \
+localhost \
+$(hostname) \
+--certs-dir=certs \
+--ca-key=my-safe-directory/ca.key
+~~~
+
+- The first command makes a new directory for the certificates.
+- The second command creates the Certificate Authority (CA) certificate and key: `ca.crt` and `ca.key`.
+- The third command creates the client certificate and key, in this case for the `root` user: `client.root.crt` and `client.root.key`. These files will be used to secure communication between the built-in SQL shell and the cluster (see step 4).
+- The fourth command creates the node certificate and key: `node.crt` and `node.key`. These files will be used to secure communication between nodes. Typically, you would generate these separately for each node since each node has unique addresses; in this case, however, since all nodes will be running locally, you need to generate only one node certificate and key.
+
+### Step 2. Set the environment variable for user authentication and start the first node
+
+~~~ shell
+$ COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE \
+  ./cockroach start --host=<node1 hostname> --certs-dir=certs
+~~~
+
+### Step 3. Add nodes to the cluster
+
+In a new terminal, add the second node:
+
+~~~ shell
+$ cockroach start \
+--certs-dir=certs \
+--store=node2 \
+--host=<node2 hostname \
+--join=<node1 hostname>:26257
+~~~
+
+In a new terminal, add the third node:
+
+~~~ shell
+$ cockroach start \
+--certs-dir=certs \
+--store=node3 \
+--host=<node3 hostname \
+--join=<node1 hostname>:26257
+~~~
+
+### Step 4. Initialize the cluster
+
+Run the [`cockroach init`](initialize-a-cluster.html) command with the `--certs-dir` flag set to the directory containing the `ca.crt` file and the files for the root user, and with the `--host` flag set to the address of any node:
+
+~~~ shell
+$ cockroach init --certs-dir=certs --host=<address of any node>
+~~~
+
+### Step 5. Create a user with a password
+
+Usernames are case-insensitive; must start with either a letter or underscore; must contain only letters, numbers, or underscores; and must be between 1 and 63 characters.
+
+~~~ sql
+> CREATE USER <username> WITH PASSWORD '<password>';
+~~~
+
+For secure clusters, you must also [create their client certificates](create-security-certificates.html).
+
+
+### Step 6. Access the Admin UI using the user credentials
+You can access the Admin UI from any node in the cluster.
+
+By default, you can access it via HTTP on port `8080` of the hostname or IP address you configured using the `--host` flag while [starting the node](https://www.cockroachlabs.com/docs/stable/start-a-node.html#general). For example, `https://<any node host>:8080`.
+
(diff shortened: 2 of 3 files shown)


v2.1/admin-ui-access-and-navigate.md
```diff
@@ -8,6 +8,53 @@ toc: false
...
+
+<section class="filter-content" markdown="1" data-scope="with-user-auth">
+
+User authentication can be enabled only for secure clusters. To enable user authentication:
+
+### Step 1. Start a secure cluster with the user authentication environment variable set for each node:
+
+To start a secure cluster, first [create security certificates](secure-a-cluster.html#step-1-create-security-certificates).
+
+Then set the environment variable, `COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE`, while [starting each node](secure-a-cluster.html#step-2-start-the-first-node) in the cluster:
```
vilterp: I thought we were not going to document this, since we'll be turning it off before 2.1 is released, and just tell our alpha testers about it in the meantime.

jseldess:
@Amruta-Ranade, thanks for soliciting early feedback. I think instead of expanding the "Access and Navigate" page, we should add a separate page called "Secure the Admin UI" and put the guidance there. On the "Access and Navigate" page, we could mention that a password is required to access if use auth is on. Thoughts?

v2.1/admin-ui-access-and-navigate.md
Amruta-Ranade: @vilterp Another argument was to document the state-of-the-product at the time of alpha and editing it as and when we remove the variable. That made more sense from the user's perspective, so went with that option. Are there any downsides to documenting the variable for now?

Amruta-Ranade:
@jseldess As I understand it, the user auth will be mandatory for secure clusters for 2.1. In that case, pointing to yet another document for securing the Admin UI might be cumbersome. Also, the user auth UX might undergo changes/get additional features. IMO, it might be better to wait for those changes to make it into the product before restructuring the doc set.

jseldess:
@Amruta-Ranade, you make a good point. But if user auth is in fact mandatory for the 2.1 release, the guidance will need to be built directly into our secure deployment tutorials. In the meantime, we need to put this info somewhere, and where it is currently feels strange. If I'm a user going to a page called "Access and Navigate the Admin UI", I wouldn't expect to get guidance on setting up user auth. Those are distinct objectives. 

Given that this content is temporary, I do think it's clearest to just create a distinct page. It won't be a problem to remove the page when the time comes, putting in place redirects, if necessary.

Let's chat about this in person more.

Amruta-Ranade:
@jseldess The way I understand it, the Admin UI user auth feature does not secure the Admin UI as much as it allows users to access the Admin UI for secure clusters. But I see your point, and agree that from a content reusability perspective, it's easier to document it as a new page. Also, because it is not a mandatory feature for this alpha, it makes sense to decouple it from the standard "access the Admin UI" content. Will work on the separate page today.

Amruta-Ranade:
Update: Tested the procedure on the CRDB alpha. It works just fine :)

Amruta-Ranade:
@jseldess Created an new page for the user auth content.

v2.1/admin-ui-access-and-navigate.md
vilterp: Only downside is that you'll have to do some editing, i.e. removing all references to this environment variable and just saying that "if your cluster is secure, the admin ui will be behind login; otherwise it will be open"

Amruta-Ranade: That's okay. I don't mind doing that :)

Amruta-Ranade:
@couchand I agree. This is the first feature being documented with our new Agile process, so I am aware that I might have to iterate on the content as required. Let's see how that works out.

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Enable User Authentication
+summary: Learn how to enable user authentication for secure clusters for the Admin UI.
+toc: false
+---
+
+By default, CockroachDB allows all users to access and view the Admin UI. For added security, you can choose to enable user authentication so that only authorized users can access and view the Admin UI.
```
jseldess: I'd suggest we call out that this is for secure cluster directly in the intro and remove the info callout:

> By default, CockroachDB allows all users to access the Admin UI. However, for secure clusters, you can enable user authentication so that only authorized users can access and view the Admin UI.

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+{{site.data.alerts.callout_info}}User authentication can be enabled only for secure clusters. {{site.data.alerts.end}}
+
+<div id="toc"></div>
+
+### Step 1. Start a secure cluster with the user authentication environment variable set for each node:
```
jseldess: Remove the colon at end of this heading

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+To start a secure cluster, first [create security certificates](secure-a-cluster.html#step-1-create-security-certificates).
```
jseldess: I'd suggest keeping this a lot simpler:

> Start a secure cluster as described in our [deployment tutorials](manual-deployment.html). However, when starting each node, be sure to set the `COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE` environment variable, for example:

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+Then set the environment variable, `COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE`, while [starting each node](secure-a-cluster.html#step-2-start-the-first-node) in the cluster:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE \
+  ./cockroach start --host=<node1 hostname> --certs-dir=certs
+~~~
+
+Finally, [initialize the cluster](initialize-a-cluster.html).
```
jseldess: If you revise as mentioned above, this can be removed.

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Step 2. Create a user with a password
```
jseldess: It's strange to keep this singular. I think it should be plural.

> Step 2. Create users with passwords

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+[Use the built-in SQL client](use-the-built-in-sql-client.html) to [create a user with a password](create-user.html). Usernames are case-insensitive; must start with either a letter or underscore; must contain only letters, numbers, or underscores; and must be between 1 and 63 characters.
```
jseldess: Based on my previous comment:

> For each user who should have access to the Admin UI, use the built-int SQL client...

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+{% include copy-clipboard.html %}
+~~~ sql
+> CREATE USER <username> WITH PASSWORD '<password>';
+~~~
+
+For secure clusters, you must also [create their client certificates](create-security-certificates.html).
```
jseldess: This is a bit confusing. We're only focusing on secure cluster, so why do we say "For secure clusters" here?

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Step 3. Access the Admin UI using the user credentials
```
jseldess: This step feels out of place to me. The instructions are for enabling user auth, not for a user to access the UI. 

Looking over everything again, I almost think headings are overkill for this task. What do you think about:

> 1. Start a secure cluster as described in our [deployment tutorials](manual-deployment.html). 
>
>     However, when starting each node, be sure to set the `COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE` environment variable as follows:
>    ```
>    $ COCKROACH_EXPERIMENTAL_REQUIRE_WEB_LOGIN=TRUE \
>    cockroach start --host=<node1 hostname> --certs-dir=certs
>    ```
> 2. For each user who should have access to the Admin UI, [create a user with a password](create-user.html).
>
>    Usernames are case-insensitive; must start with either a letter or underscore; must contain only letters, numbers, or underscores; and must be between 1 and 63 characters.
>
> On accessing the Admin UI, these users will see a Login screen, where they will need to enter their usernames and passwords.

_includes/sidebar-data-v2.1.json
```diff
@@ -1538,6 +1538,12 @@
               "/${VERSION}/admin-ui-access-and-navigate.html"
             ]
           },
+          {
+            "title": "Enable User Authentication",
```
jseldess: I defer to you, but I still like "Secure the Admin UI" better than "Enable User Authentication". User auth is much broader than the admin ui.

Amruta-Ranade:
@jseldess Ready for review round 2 :)

v2.1/admin-ui-user-authentication.md
```diff
@@ -0,0 +1,23 @@
...
+title: Secure the Admin UI
+summary: Learn how to enable user authentication for secure clusters for the Admin UI.
+toc: false
+---
+
+By default, CockroachDB allows all users to access and view the Admin UI. However, for secure clusters, you can enable user authentication so that only authorized users can access and view the Admin UI.
+
+<div id="toc"></div>
+
+1. Start a secure cluster as described in our [deployment tutorials](secure-a-cluster.html#step-1-create-security-certificates).
```
jseldess: This isn't the right link. It should be `manual-deployment.html`. From there, whatever deployment guide they choose will default to the secure version.

v2.1/admin-ui-access-and-navigate.md
```diff
@@ -10,6 +10,8 @@ toc: false
 
 You can access the Admin UI from any node in the cluster.
 
+{{site.data.alerts.callout_info}}By default, CockroachDB allows all users to access and view the Admin UI. For added security, you can choose to <a href="admin-ui-user-authentication.html">enable user authentication</a> for secure clusters so that only authorized users can access and view the Admin UI. {{site.data.alerts.end}}
```
jseldess: We should probably update this language a bit as well:

> For secure clusters, you can choose to <a href="admin-ui-user-authentication.html">enable user authentication</a> to restrict access to the Admin UI to authorized users.

Issue #2195, opened 2017-11-18
docs design has low information density


benesch:
As visually appealing as our docs are, I fear their information density is too low. There are many places where we barely pack two or three sentences onto the screen at a time:

![screenshot 2017-11-18 12 12 48](https://user-images.githubusercontent.com/882976/32982917-dcd1b6a4-cc59-11e7-9071-c61ec9722de4.png)

![screenshot 2017-11-18 12 13 02](https://user-images.githubusercontent.com/882976/32982922-e9275422-cc59-11e7-8f63-91fa9c4f590b.png)

Compare that to the PostgreSQL docs:

![screenshot 2017-11-18 12 12 07](https://user-images.githubusercontent.com/882976/32982923-ec38f58a-cc59-11e7-9036-d87e3edb2a5e.png)

To be quite honest this makes reading our documentation a very frustrating experience for me personally. Others may feel differently, of course.

@kuanluo, this is obviously your wheelhouse, so assigning to you for analysis/triage. @jseldess @Amruta-Ranade @sploiselle also curious to hear your thoughts on the matter.

/cc @knz who I'm sure has thoughts too.

knz:
Yes I agree with you fully. I would also like a much more compact documentation site overall. I've always wanted that, but somehow I got a signal early on that our stylesheet was not negotiable, so I didn't push further.

benesch:
Here's a rough prototype of how I think this should look: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/156f9e7b452641f9946360fb05a691e6d4b63166/stable/

Before:

![screenshot 2017-11-18 14 15 44](https://user-images.githubusercontent.com/882976/32983950-ff8a1766-cc6a-11e7-88d6-56f7345c5a27.png)

After:

![screenshot 2017-11-18 14 15 42](https://user-images.githubusercontent.com/882976/32983949-fbf0bcf4-cc6a-11e7-8eb5-69a3bec93357.png)

Before:

![screenshot 2017-11-18 14 17 15](https://user-images.githubusercontent.com/882976/32983970-3081637e-cc6b-11e7-80de-d1eed9375d49.png)

After:

![screenshot 2017-11-18 14 17 14](https://user-images.githubusercontent.com/882976/32983972-34ba8d1c-cc6b-11e7-9762-1b8a81daa111.png)

jseldess:
[...]

PR #3205, opened 2018-05-30
[prototype/WIP] widened side grid

j-low:
__[description]:__
* rough draft for what a widened site grid would appear as on docs


j-low:
cc: @jseldess

jseldess:
This is cool, @j-low. One quick comment: Instead of letting the main body expand to the right indefinitely, I think we want a max width. While we're doing this, we should move the page-level TOC to a right margin. @kuanluo  mocked this up a bit here: https://github.com/cockroachdb/docs/issues/2195#issuecomment-361985920

Some decent examples of using the right margin for page TOC:

- https://developers.google.com/identity/protocols/OAuth2
- https://valelint.github.io/docs/styles/#creating-a-style

j-low:
@jseldess sure thing...there will be progressive updates to this PR.  I'll be working on that third column for the TOC next.

jseldess:
For clarity, I'm closing this in favor of the more recent https://github.com/cockroachdb/docs/pull/3284.

PR #3225, opened 2018-05-30
 Fix documentation of GCS AUTH for import/export docs

a-robinson:
Also  fix documentation of EXPORT's nullas option.

As called out in https://github.com/cockroachdb/cockroach/issues/25223#issuecomment-392212300

### DIFF
--- a/_includes/external-urls-v2.0.md
+++ b/_includes/external-urls-v2.0.md
@@ -13,7 +13,7 @@
 
 #### Considerations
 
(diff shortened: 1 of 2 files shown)


a-robinson:
The GCS AUTH change is to accurately reflect the code at https://github.com/cockroachdb/cockroach/blob/98b1ceb6975874c5a8670b56e6540307ab791728/pkg/ccl/storageccl/export_storage.go#L626-L651

Amruta-Ranade:
LGTM! Thanks Alex :)

Issue #3140, opened 2018-05-14
Migration from Postgres or MySQL to CockroachDB


jseldess:
For 2.1, we want to add a new `Guides > Migrate` section after `Deploy` and before `Maintain` that should have something like 3 distinct pages:

```
Migrate
- Migrate from PostgreSQL
- Migrate from MySQL
- Migrate from Oracle
```

The first two tie into many `IMPORT` improvements in the 2.1 product. The last one can be based on [this example](https://github.com/cockroachdb/docs/issues/3539) and leveraging [this draft doc](https://docs.google.com/document/d/1D8F-MgNVPK4B8AoyN-9sNA21uXbDVPIOO4ipHb-r3wc/edit?ts=5b75a1f4) from @drewdeally.

jseldess:
Oracle migration guide is tracked in a separate issue: https://github.com/cockroachdb/docs/issues/2888.

Closing this issue, now that the Postgres and MySQL guides are live, with follow-up work tracked in https://github.com/cockroachdb/docs/issues/3819.

Issue #2936, opened 2018-04-17
Frictionless migration from MySQL to CockroachDB


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/rec341jbt9b5PVY22
PM: @nstewart 
Eng partner: @dt 

Part of https://github.com/cockroachdb/docs/issues/3140.

jseldess:
- Things that are different and always will be different.
- Things that will improve over time (e.g., we're slow at foreign keys)

rmloveland:
Fixed by #3766

PR #3229, opened 2018-05-31
Add 'Import from MySQL Dump' section to 'Import Data'

### LINKED ISSUE #3139: Mysqlout-to-csv utility
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recmpRqp5Y1B4uBdp
PM: @nstewart 

This is an intermediate approach for migrating from MySQL to cockroachdb. Basic instructions are here: https://docs.google.com/document/d/16yWcR4CiOGhYaODtsyA1X28oFTgtCvr5AUWvMBUOSic/edit

This will be superseded by production work in https://github.com/cockroachdb/docs/issues/2936.

rmloveland:
- Fixes #3139

- Starts down the path to #2936

- A small piece of #3140


v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+
+~~~
+mysqldump: Got error: 1290: The MySQL server is running with the --secure-file-priv option so it cannot execute this statement when executing 'SELECT INTO OUTFILE'
+~~~
+
+### Step 2. `IMPORT` the dump files
+
+To import the MySQL dump file for a table, issue an `IMPORT` statement like the one shown below. You must use the same settings for field enclosure and escape characters that you provided to `mysqldump --tab` in the previous step.
+
+You will need to look at the dump's `foo.sql` file that corresponds to the `foo.txt` data file and translate the MySQL `CREATE TABLE` statement there into something CockroachDB understands.
```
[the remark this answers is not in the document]
jseldess: Do we have any guidance on how to translate the MySQL CREATE TABLE statement into "something CockroachDB understands? I think this might actually be step 2 with import being step 3?

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+  birth_date DATE NOT NULL,
+  first_name STRING NOT NULL,
+  last_name STRING NOT NULL,
+  gender STRING NOT NULL,
+  hire_date DATE NOT NULL
+)
+MYSQLOUTFILE DATA ('s3://nate-external-storage/employees.txt?AWS_ACCESS_KEY_ID=ACCESSKEY&AWS_SECRET_ACCESS_KEY=SECRET')
+~~~
+
+Success will look something like the following:
```
jseldess: Can we tighten this up? 
> Success will look like:

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+
+To import the MySQL dump file for a table, issue an `IMPORT` statement like the one shown below. You must use the same settings for field enclosure and escape characters that you provided to `mysqldump --tab` in the previous step.
+
+You will need to look at the dump's `foo.sql` file that corresponds to the `foo.txt` data file and translate the MySQL `CREATE TABLE` statement there into something CockroachDB understands.
+
+This example uses S3. For a complete list of the types of cloud storage `IMPORT` can pull from, see [Import File URLs](import.html#import-file-urls).
+
+{% include copy-clipboard.html %}
+~~~ sql
+IMPORT TABLE employees (
```
jseldess: For SQL commands, we start the first line with `>`:

```
> IMPORT TABLE employees 
```

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+
+~~~
++--------------------+-----------+--------------------+------+---------------+----------------+-------+
+|       job_id       |  status   | fraction_completed | rows | index_entries | system_records | bytes |
++--------------------+-----------+--------------------+------+---------------+----------------+-------+
+| 352938237301293057 | succeeded |                  1 |    0 |             0 |              0 |     0 |
++--------------------+-----------+--------------------+------+---------------+----------------+-------+
+(1 row)
+
+Time: 2.129808512s
```
jseldess: I'd remove the Time. This will probably differ from case to case, and we don't wan't this example setting performance expectations.

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+~~~
+
+To load the data from local files, you have 2 options:
+
+1. Start up a [local HTTP server](create-a-file-server.html#using-caddy-as-a-file-server) to serve the files, and point `MYSQLOUTFILE DATA` at the server's URL. This is the recommended option.
```
jseldess: I'd make this a bulleted list to avoid people misunderstanding this as steps.

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+### Step 1. Dump the MySQL database
+
+Issue a command like the following (modified as needed for your environment):
+
+{% include copy-clipboard.html %}
+~~~ sh
+$ mysqldump -uroot --tab=/tmp employees --fields-enclosed-by='"' --fields-escaped-by='\'
+~~~
+
+Always specify an escape character when running `mysqldump --tab`. If your data happens to include any of the special characters, the dumped data cannot be decoded correctly if they are not escaped. We also recommend specifying an enclose character.
```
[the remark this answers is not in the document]
rmloveland: In https://docs.google.com/document/d/16yWcR4CiOGhYaODtsyA1X28oFTgtCvr5AUWvMBUOSic/edit we say it's "probably a good idea", so I went with it since this is "alpha" :-)  I know @dt wrote  in https://github.com/cockroachdb/cockroach/pull/25287 that 

<blockquote>
MySQL's OUTFILE (as used in mysqldump --tab) can be configured to almost  
CSV, but its escaping and field enclosing behavior does not quite match  
the CSV spec.
</blockquote>

I suspect it's related to that?  Not sure though, perhaps David has more detail.

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+
+Your MySQL user will need permission to execute [`SELECT INTO OUTFILE`](https://dev.mysql.com/doc/refman/8.0/en/select.html), as required by [`mysqldump --tab`](https://dev.mysql.com/doc/refman/8.0/en/mysqldump-delimited-text.html).
```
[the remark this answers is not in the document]
rmloveland: Fixed.

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
...
+
+If you get the following error, you will need to change the value of MySQL's [`secure_file_priv` system variable](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_secure_file_priv):
```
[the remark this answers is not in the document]
rmloveland: Fixed. (or 🦊ed? 😋)

v2.1/import-data.md
```diff
@@ -77,10 +77,94 @@ For reference, CockroachDB uses these defaults:
 - `[port]`: **26257**
 - `[user]`: **root**
 
+
+## Import from MySQL Dump
+
+{% include experimental-warning.md %}
+
+<span class="version-tag">New in v2.1:</span> This section has instructions for getting data from MySQL dump files into CockroachDB using [`IMPORT`](import.html). It uses the [employees data set](https://github.com/datacharmer/test_db) that is also used in the [MySQL docs](https://dev.mysql.com/doc/employee/en/).
```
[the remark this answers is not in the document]
rmloveland: That data set is what I used to write and test all of the instructions.  The `employees` bit in the `mysqldump` command is the table being dumped (but we also say to modify as needed for the user's environment if they want a different table).  Hope that helps!

v2.1/import-data.md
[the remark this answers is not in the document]
rmloveland: 🦊ed!

v2.1/import-data.md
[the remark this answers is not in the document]
rmloveland: > I'd suggest just using .sql and .txt.
🦊ed, thanks.

> Do we have any guidance on how to translate the MySQL CREATE TABLE statement into "something CockroachDB understands?

Not that I could find, a [search for MySQL](https://www.cockroachlabs.com/docs/search.html?q=mysql) only turned up two results.  However it was reasonably straightforward even for a SQL novice like me (at least in this case).

> I think this might actually be step 2 with import being step 3?

I went with making it part of the "import" step because crafting the `IMPORT` statement basically is the whole step since that one statement both creates the table and loads the data file.  However maybe there's a better way to do it.

v2.1/import-data.md
rmloveland: 🦊ed, thanks!

v2.1/import-data.md
rmloveland: Fixed thanks!

v2.1/import-data.md
rmloveland: Fixed!

v2.1/import-data.md
rmloveland: nice catch, thanks!

rmloveland:
> Are you planning to update the IMPORT doc as part of this, or as a follow-up? If as a follow-up, I'd say Address #3139 instead of Fixes #3139.

Thanks for the review @jseldess , you are the best!

I updated the language to 'Addresses'.  I will update IMPORT as a follow-up and link related issues/PRs to this one

dt:
We have two ways to import from mysqldump -- one described here is what is generated by `mysqldump --tab`, which is mysql's `OUTFILE` format. 
So, `MYSQLOUTFILE` is one of two `mysqldump` formats we'll support in 2.1.

There are some gotchas with the OUTFILE format.

First, you might not be able to run it, depending on your configuration: writing it (via `SELECT INTO OUTFILE` or with `mysqldump --tab`) requires a special mysql server configuration flag, that enabled writing to files, since the _server_ writes the file (usually `mysqldump` gets results back from the server over the network, and you write its output to a file yourself, without worrying about the server configuration).

Second, the format itself is a bit tricky: you have to specify field enclosing and/or escape characters -- if your data includes your field or row separators and you don't specify an escape char, it will dump ~useless data since you can't tell fields apart anymore. So if using outfile, it is very important to specify an escape character (`mysqldump`'s default is to not have one).

When importing from OUTFILE data, it is important to use the same options as were used to generate it, e.g. if you use `--fields-enclosed-by='"'` when running `mysqldump`, you need to provide the same value to `IMPORT ... MYSQLOUTFILE DATA ... WITH fields_enclosed_by =`, etc.

Fortunately, OUTFILE is only one of the `mysqldump` formats: the _default_ (i.e. when not passing `--tab), is just SQL statements, i.e. `CREATE TABLE`s and lots of `INSERT`s. Given outfile's problems, we're also supporting parsing the native SQL format.

Currently this is used by specifying `IMPORT ... MYSQLDUMP DATA (example.sql)`. However, we're actively working on getting this to `IMPORT FROM MYSQLDUMP (example.sql)` with getting the table names and structures from the dumpfile directly.

So, to recap, to load from mysql, we'll have (in order of ease of use):
- `IMPORT FROM MYSQLDUMP bar` that creates and loads all the tables in `mysqldump` file bar
- `IMPORT TABLE foo FROM MYSQLDUMP bar` that created and loads foo from `mysqldump` file bar
- `IMPORT TABLE foo (i int primary key...) MYSQLDUMP DATA (bar)` that creates foo using the specified schema, loading the _data_ from the `mysqldump` file bar.
- `IMPORT TABLE foo (i int primary key...) MYSQLOUTFILE DATA (bar) WITH <opts>` that creates foo using the specified schema, loading the `mysqldump --tab <opts>` file bar with matching opts.

dt:
Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[v2.1/import-data.md, line 99 at r1](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDqrP5qKR-ivBd_g276-r1-99:-LDrppbvARzRs_rcENGh:bth3jtv) ([raw file](https://github.com/cockroachdb/docs/blob/e51af0b3dd0ab16a56a95dfd9c09f9cc7d6cf0c1/v2.1/import-data.md#L99)):*

🦊ed!
</blockquote></details>

We should recommend always using an enclose, and even more importantly, always using an escape... though, even better, not using OUTFILE -- as you probably discovered, it is just not a very good format (or at least they picked poor defaults), and we should probably just avoid it and recommend native SQL mysqldump instead.

---

*[v2.1/import-data.md, line 140 at r2](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDrp5g72MpfyHY1lxvW:-LDrp5g72MpfyHY1lxvX:bxztmiw) ([raw file](https://github.com/cockroachdb/docs/blob/ce03636afa3ed6828fcb2569c4c5fba21b632644/v2.1/import-data.md#L140)):*
> ```Markdown
> ~~~
> 
> To load the data from local files, you have 2 options:
> ```

This is also true of CSV data, and any other format, so we should have a canned answer for "how do i get cockroach to interact with this file i have right here, on my local machine" rather than having a different answer in each usage.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3229)*

rmloveland:
Review status: 0 of 1 files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---

*[v2.1/import-data.md, line 99 at r1](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDqrP5qKR-ivBd_g276-r1-99:-LDrvz787QCcin8c6Sud:bm7olgo) ([raw file](https://github.com/cockroachdb/docs/blob/e51af0b3dd0ab16a56a95dfd9c09f9cc7d6cf0c1/v2.1/import-data.md#L99)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

We should recommend always using an enclose, and even more importantly, always using an escape... though, even better, not using OUTFILE -- as you probably discovered, it is just not a very good format (or at least they picked poor defaults), and we should probably just avoid it and recommend native SQL mysqldump instead.
</blockquote></details>

Darn, did not realize we would have MYSQLDUMP in this alpha release as well?  Building a new binary from the SHA in https://github.com/cockroachdb/cockroach/issues/25351 to test, will update doc accordingly.

---

*[v2.1/import-data.md, line 140 at r2](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDrp5g72MpfyHY1lxvW:-LDruvDf2_ajUuPGAaHc:b-khlz0p) ([raw file](https://github.com/cockroachdb/docs/blob/ce03636afa3ed6828fcb2569c4c5fba21b632644/v2.1/import-data.md#L140)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

This is also true of CSV data, and any other format, so we should have a canned answer for "how do i get cockroach to interact with this file i have right here, on my local machine" rather than having a different answer in each usage.
</blockquote></details>

By "we should have a canned answer" do you mean that we should not describe the workflow using `extern` and just point people at running a local server?

Although perhaps you mean that we should have something documented in one place that answers this across all of the various IMPORT etc. statements that we can just point to?  (In which case I agree but I'm not sure I can commit to doing that as part of this PR to get something out by Monday - can capture in an issue for future work tho)

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3229)*

rmloveland:
Thanks for sharing that knowledge, David!  I have pushed some updates to emphasize using the `MYSQLDUMP` variant which reads from SQL, not the more complex OUTFILE format.  I think we should also document that before the 2.1 release, but it seems better to focus on importing the SQL as the easier/better option in terms of docs we will publish Monday.  Do you agree?

---

Review status: 0 of 1 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[v2.1/import-data.md, line 99 at r1](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDqrP5qKR-ivBd_g276-r1-99:-LDwc_D-428zaV-IJmCb:b9hvmx5) ([raw file](https://github.com/cockroachdb/docs/blob/e51af0b3dd0ab16a56a95dfd9c09f9cc7d6cf0c1/v2.1/import-data.md#L99)):*

Darn, did not realize we would have MYSQLDUMP in this alpha release as well?  Building a new binary from the SHA in https://github.com/cockroachdb/cockroach/issues/25351 to test, will update doc accordingly.
</blockquote></details>

OK, just pushed an update switching to using the SQL dumped by `mysqldump` and reading it in with `IMPORT ... MYSQLDUMP DATA (example.sql)`. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3229#-:-LDwbw_U6JG3xm_o0L5O:b-d85qdw)*

v2.1/import-data.md
```diff
@@ -77,10 +77,89 @@ For reference, CockroachDB uses these defaults:
...
+|       job_id       |  status   | fraction_completed | rows | index_entries | system_records | bytes |
++--------------------+-----------+--------------------+------+---------------+----------------+-------+
+| 352938237301293057 | succeeded |                  1 |    0 |             0 |              0 |     0 |
++--------------------+-----------+--------------------+------+---------------+----------------+-------+
+(1 row)
+~~~
+
+To load the data from local files, you have 2 options:
+
+- Start up a [local HTTP server](create-a-file-server.html#using-caddy-as-a-file-server) to serve the files, and point `MYSQLOUTFILE DATA` at the server's URL. This is the recommended option.
```
dt: `MYSQLOUTFILE`

v2.1/import-data.md
```diff
@@ -77,10 +77,89 @@ For reference, CockroachDB uses these defaults:
...
+{% include experimental-warning.md %}
+
+<span class="version-tag">New in v2.1:</span> This section has instructions for getting data from MySQL dump files into CockroachDB using [`IMPORT`](import.html). It uses the [employees data set](https://github.com/datacharmer/test_db) that is also used in the [MySQL docs](https://dev.mysql.com/doc/employee/en/).
+
+- [Step 1. Dump the MySQL database](#step-1-dump-the-mysql-database)
+- [Step 2. `IMPORT` the dump files](#step-2-import-the-dump-files)
+
+### Step 1. Dump the MySQL database
+
+Make sure your MySQL user has permission to execute [`SELECT INTO OUTFILE`](https://dev.mysql.com/doc/refman/8.0/en/select.html), as required by [`mysqldump`](https://dev.mysql.com/doc/refman/8.0/en/mysqldump-sql-format.html). Then issue a command like the following (modified as needed for your environment):
```
dt: I believe this only applies to the `--tab` output mode of `mysqldump`, so I think since we're only showing the SQL dump, this note and the notes below about error 1290, don't apply any more.

I think we'll want a separate section for outfile, where we can put these warnings, along with recommendation for escape chars, etc but I think that is pretty low-priority: the SQL format is the default dump format and is better supported by cockroach, so I think it makes sense to focus on it initially.

rmloveland: > I believe this only applies to the --tab output mode of mysqldump, so I think since we're only showing the SQL dump, this note and the notes below about error 1290, don't apply any more.

Fixed, thanks

> I think we'll want a separate section for outfile, where we can put these warnings, along with recommendation for escape chars, etc but I think that is pretty low-priority: the SQL format is the default dump format and is better supported by cockroach, so I think it makes sense to focus on it initially.

Great!

v2.1/import-data.md
rmloveland: Thanks - went one step further and changed it to `MYSQLDUMP` since that is what we are showing in the example now.

jseldess:
:lgtm:, with a suggestion.

---

Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[v2.1/import-data.md, line 113 at r1](https://reviewable.io/reviews/cockroachdb/docs/3229#-LDqrP5qKR-ivBd_g276-r1-113:-LEA443UEjYFWtgBJ5cB:b-vzyc1f) ([raw file](https://github.com/cockroachdb/docs/blob/e51af0b3dd0ab16a56a95dfd9c09f9cc7d6cf0c1/v2.1/import-data.md#L113)):*

> I'd suggest just using .sql and .txt.
🦊ed, thanks.

> Do we have any guidance on how to translate the MySQL CREATE TABLE statement into "something CockroachDB understands?

Not that I could find, a [search for MySQL](https://www.cockroachlabs.com/docs/search.html?q=mysql) only turned up two results.  However it was reasonably straightforward even for a SQL novice like me (at least in this case).

> I think this might actually be step 2 with import being step 3?

I went with making it part of the "import" step because crafting the `IMPORT` statement basically is the whole step since that one statement both creates the table and loads the data file.  However maybe there's a better way to do it.
</blockquote></details>

Can you make `CREATE TABLE` a link to our docs for that statement? That would probably be enough.

---


*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3229#-:-LEA4K3-7I93ZiLYRrqI:bhkeom)*

PR #3230, opened 2018-05-31
Scroll monospaced output instead of wrapping

rmloveland:
We currently wrap monospaced tabular output which makes it unreadable. This patch turns on scrolling in such cases.

Our highlighted code samples appear to already have the scrolling turned on, so this change brings the non-highlighted output in line.

### DIFF
--- a/css/theme-blue.css
+++ b/css/theme-blue.css
@@ -148,3 +148,9 @@ body.print h1 {color: #015CAE !important; font-size:28px;}
 body.print h2 {color: #595959 !important; font-size:24px;}
 body.print h3 {color: #E50E51 !important; font-size:14px;}
 body.print h4 {color: #679DCE !important; font-size:14px; font-style: italic;}
+
+pre {
+    overflow-x: scroll;
+    overflow-y: hidden;
+    white-space: nowrap;
+}


rmloveland:
Sure!  See the section **Add a Non-Null Column with a Default Value** on [Add Column](https://www.cockroachlabs.com/docs/dev/add-column.html).  Screenshot is taken from Firefox 60.0.1 with the browser using the full screen width.

screenshot 2018-06-01 13 31 23

With this patch it looks like this:

screenshot 2018-06-01 13 35 15

rmloveland:
@jseldess I found some even more egregious examples on the [`EXPLAIN`](https://www.cockroachlabs.com/docs/dev/explain.html##metadata-option) page.  How do you feel about merging this? :-)

jseldess:
Sorry for the delay, @rmloveland. This is weird. I see it on FF, too, but not on Chrome. I'm fine with this change, but I'd like @j-low to be the reviewer, since he's inherited our messy css. He may want this rule somewhere else.

rmloveland:
w00t!  thanks @j-low and @jseldess !!!

PR #3239, opened 2018-06-01
Updated related docs for cancel schema change

Amruta-Ranade:


Amruta-Ranade:
@jseldess Can't see the suggestion.

v2.1/manage-long-running-queries.md
```diff
@@ -6,8 +6,6 @@ toc: false
 
 This page shows you how to identify and, if necessary, cancel SQL queries that are taking longer than expected to process.
 
-{{site.data.alerts.callout_info}}Schema changes (statements beginning with <code>ALTER</code>) cannot currently be cancelled. However, to monitor the progress of schema changes, you can use <a href="show-jobs.html"><code>SHOW JOBS</code></a>.{{site.data.alerts.end}}
```
jseldess: Instead of removing this, we should either:

1. Make this a tip that `SHOW JOBS` and `CANCEL JOBS` can be used to identify and cancel long-running schema changes. This is probably fine for now.
2. Revise this page to cover both standard queries and jobs, including schema changes. This is probably the right approach, but would require more time and thought.

jseldess:
@Amruta-Ranade, sorry about that. Not sure how my comment got lost. Recreated it for you.

PR #3241, opened 2018-06-03
Temporarily revert encryption docs

jseldess:
As requested by @mberhault and @nstewart, due to encryption bug that won't be resolved until after the v2.1.0-alpha.20180604 release.


mberhault:
LGTM. Thanks

PR #3253, opened 2018-06-05
Fix release notes

### LINKED ISSUE #3252: Incomplete/incorrect release notes
https://forum.cockroachlabs.com/t/release-notes-for-cockroachdb-v2-1-alpha-20180604/1711

- The latest published release note have an item that reads "It is now possible to use ALTER TABLE ... " but there is no issue or PR reference and it doesn't really say what it does.
- "order of precendence" should read "order of precedence"
- "Documented the TIMETZ data type. #3102 " - this is incorrect, we have pulled the TIMETZ doc for now because the implementation is invalid/incorrect
- the contributor "yuzefovich" is Yahor and his name should be spelled in full.

jseldess:
Fixes #3252

### DIFF
--- a/releases/v2.1.0-alpha.20180604.md
+++ b/releases/v2.1.0-alpha.20180604.md
@@ -115,7 +115,7 @@ Get future release notes emailed to you:
 
 ### Doc Updates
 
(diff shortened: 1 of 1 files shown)


lhirata:
It seems like the binary file names are wrong - I'm getting a 404. Correct URLs:

- MacOS: https://binaries.cockroachdb.com/cockroach-v2.1.0-alpha.20180604.darwin-10.9-amd64.tgz
- Linux: https://binaries.cockroachdb.com/cockroach-v2.1.0-alpha.20180604.linux-amd64.tgz
- Windows:  https://binaries.cockroachdb.com/cockroach-v2.1.0-alpha.20180604.windows-6.2-amd64.zip
- Source: https://binaries.cockroachdb.com/cockroach-v2.1.0-alpha.20180604.src.tgz

PR #3255, opened 2018-06-05
TIMETZ is not recommended for use in current state

### LINKED ISSUE #3254: urgent: de-announce TIMETZ and discourage users from using it
As found in https://github.com/cockroachdb/cockroach/issues/25224 the implementation of TIMETZ is currently incomplete/incorrect.

The feature will be either retracted or corrected, but the way forward will necessarily be backward-incompatible. 

A user who now creates a table with this type will see their table broken when they use the future corrected implementation.

The docs must de-emphasize the type, discourage users from using it and warn that any use will be broken by a future release.

rmloveland:
Fixes #3254.


jseldess:
LGTM. Thanks, @rmloveland!

PR #3270, opened 2018-06-12
Update 2.1 sidenav

jseldess:
- Expand first level by default.
- Remove the "accordian" behavior.

Conditional logic ensures that these changes affect only 2.1 docs.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -10,7 +10,7 @@
       var navgocoOptions = {
         caretHtml: '',
         {% if page.version.version == 'v2.1' %}accordion: false,{% else %}accordion: true,{% endif %}
-        openClass: 'active', // open
+        openClass: 'active--expand', // open
         save: false, // leave false or nav highlighting doesn't work right
         cookie: {
           name: 'navgoco',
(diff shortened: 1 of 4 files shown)


jseldess:
@lhirata, @rmloveland, @Amruta-Ranade, Joe helped find a way to make the first-level of the 2.1 sidenav expand by default. We've been getting positive feedback on the new sidenav, but I'd like to test out if people also like the 2.1 sidenav with more exposed. I think it might be a better middle-ground (less clicks to intuit the structure). 

I'd like your help testing this new nav. Please comment here with any UX issues or bugs you find.

jseldess:
Issues:

- First level expansion in mobile nav is overwhelming. I think we should probably revert to collapsing everything and using accordian behavior in the mobile nav.
- Top levels are highlighted green by default. When you click a top-level (Guides, Reference, etc.), all of the top-levels lose their highlighting. I think we might actually want to style the top levels slightly differently so they stand out. Perhaps using something other than the green "link" color?

Amruta-Ranade:
First impressions: I really, really like the expanded structure. It took me a long time to learn where to look for certain docs. With the expanded structure, I think we are taking out the guesswork from the process of locating the docs.

As for the top-level items losing their formatting, yes, that does seem weird. Wonder why that's happening though.

j-low:
@Amruta-Ranade the styling issue happens because the same CSS class that tells the top-level items to be expanded is also the same class that tells them to be green.  Fixing this issue will be straightforward and will just involve using separate classes to indicate `active` (i.e. green highlight) and `expanded`.

jseldess:
Oh, we also need to maker the sidenav independently scrollable.

lhirata:
This looks great! A lot easier to figure out where to navigate. I agree with the issues that've been brought up already (independent scrolling, highlighted headings) and would also echo Amruta's comment about the hamburger symbols. I'm not sure we need them anymore (at least for browser), and I'm not sure we need the ability to collapse sections.

j-low:
PR updated with the following changes:
- sidenav expands without top level nav items highlighted
- sidenav is _not_ expanded by default on mobile and tablet views

Still outstanding is the independent scrolling functionality.  As it turns out the layout is a _little_ weird with regards to how the containers are laid out and positioned.  Next week I'm going to be doing some refactoring of both the hugo site and docs to implement a wider site grid.  During that work I'm going to try refactoring the layout to make it conducive to using an independently scrollable sidenav.

jseldess:
Thanks, @j-low! There's one thing we might want to address before merging this: When you select a page in the sidenav, the sidenav collapses to the page's parent. So the selected page isn't visible. 

Example:

screen shot 2018-06-14 at 9 19 02 pm

It should look like:

screen shot 2018-06-14 at 9 19 11 pm

jseldess:
Merging now. We'll open a follow-up PR to address the remaining issues.

PR #3282, opened 2018-06-18
SQL Optimizer

### LINKED ISSUE #2975: Query Optimizer v1
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recZHaMd0J2cuxavB
PM: @awoods187 
Eng Partner: @andy-kimball

lhirata:
Closes #2975.


lhirata:
Open questions:

- Where in TOC should this be?
- Do we need more of an overview?

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,72 @@
+---
+title: SQL Optimizer
+summary: Learn about the SQL Optimizer
+toc: false
+---
+
+<span class="version-tag">New in v2.1:</span> The SQL Optimizer seeks the lowest cost for a query, usually related to time. **CockroachDB's cost-based Optimizer is enabled by default**, and the heuristic planner will only be invoked on functionality that is not supported by the Optimizer (i.e., on auto-fallback), or if you explicitly [set a cluster or session variable that disables the cost-based Optimizer](#how-to-turn-the-optimizer-off).
```
andy-kimball: Is this document intended for the July 2nd alpha, the July 25th alpha, or for the October production release? We won't enable by default until the July 25th alpha.

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,72 @@
...
+...
+|                                                                          |
++--------------------------------------------------------------------------+
+~~~
+
+The query above will be run with the Optimizer and `EXPLAIN (OPT)` returns the query plan tree.
+
+## How to turn the Optimizer off
+
+With the Optimizer turned on, your workloads may perform worse than expected (e.g., lower throughput, higher latency).
```
andy-kimball: Up to you, but I'd reword this a bit, because it sounds like we're expecting it to be worse all around. For example:
```
With the Optimizer turned on, the performance of some workloads may change. If your workload performs worse than expected (e.g. lower throughput, higher latency), you can turn off the cost-based Optimizer and use the heuristic Optimizer.
```

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,72 @@
...
+~~~ sql
+SET experimental_opt = 'off';
+~~~
+
+To turn the Optimizer off by cluster:
+
+{% include copy-clipboard.html %}
+~~~ sql
+SET CLUSTER SETTING sql.defaults.optimizer = 'off';
+~~~
```
andy-kimball: Maybe add a note that the cluster setting doesn't actually turn the Optimizer off, but instead changes the *default* session setting. This means that you need to restart your session to see any difference.

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Known limitations
+
+- The optimizer will not support automated use of statistics during this time period.
+- Many optimizations present in 2.0 may not be supported; however, the optimizer will fall back to the v2.0 code path for this functionality. If performance in the new alpha is worse than v2.0, you will need to manually force it to fallback to v2.0 (see [How to turn the Optimizer off](#how-to-turn-the-optimizer-off)).
```
andy-kimball: We don't fall back in order to do optimizations, but instead to enable functionality that we don't otherwise support.

So should be: "Some features present in 2.0 are not supported by the optimizer; however, the optimizer will fall back ..."

v2.1/sql-optimizer.md
awoods187: +1 to this reword.

v2.1/sql-optimizer.md
awoods187: i think this is actually fairly confusing and will need to be explained in depth

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,72 @@
...
+To turn the Optimizer off by cluster:
+
+{% include copy-clipboard.html %}
+~~~ sql
+SET CLUSTER SETTING sql.defaults.optimizer = 'off';
+~~~
+
+## Known limitations
+
+- The optimizer will not support automated use of statistics during this time period.
```
awoods187: @andy-kimball I think the answer to this is no, but do we want to encourage people to use create statistics?

andy-kimball: I don't see a big advantage to encouraging them to use them at this point.

v2.1/sql-optimizer.md
```diff
@@ -0,0 +1,76 @@
...
+
+{{site.data.alerts.callout_info}}
+Changing the cluster setting does not immediately turn the Optimizer off; instead, it changes the default session setting to `off`. To see the change, restart your session.
+{{site.data.alerts.end}}
+
+## Known limitations
+
+- The optimizer will not support automated use of statistics during this time period.
+- Some features present in v2.0 are not supported by the Optimizer; however, the Optimizer will fall back to the v2.0 code path for this functionality. If performance in the new alpha is worse than v2.0, you will need to manually force it to fallback to v2.0 (see [How to turn the Optimizer off](#how-to-turn-the-optimizer-off)).
+
```
awoods187: should we explicitly mention correlated subqueries?

lhirata:
@andy-kimball - Do we need to mention "Some correlated subqueries are not supported yet" in the Known limitations section?

lhirata:
@jseldess when you have time to review, can you prioritize this one? I want to merge this branch before I start on something else because both of them touch the EXPLAIN page. Thanks!

PR #3284, opened 2018-06-19
Widen site grid

j-low:
__[description]:__
* widen overall site grid
* move TOC to righthand gutter
* implement fixed and independently-scrollable sidenav


j-low:
cc: @jseldess @lhirata @rmloveland @Amruta-Ranade 

This first build implements the independently-scrollable sidenav which is QA-able at this point.  I have not merged or duplicated the other sidenav changes from the work performed last week.

lhirata:
For the independently scrolling side nav, is there a way to lock the minimum size of the sidebar container? What I'm noticing is that it gets squished when you scroll down to the bottom of the main content / footer. 

For example:

screen shot 2018-06-14 at 3 13 55 pm

Let me know if you need clarification!

rmloveland:
Hey @j-low thanks for working on this!  Really looking forward to a wider layout on the site.

Unfortunately I have one issue: The new layout causes the content text not to wrap to fit inside the browser window when the browser is using half the screen in a "2-up" configuration with another app (screenshot attached).

screenshot 2018-06-20 10 59 40

In other words, it seems like this layout really wants the browser to be full-screened or at least prefer to be a horizontal rectangle rather than a vertical rectangle.  I suspect this might break for some developers who read docs on a vertical monitor (something I've seen around), or for people like me who work in side-by-side windows.

j-low:
@rmloveland thanks for the QA!  That's actually a known issue that's being worked on.  It just didn't make it into the last commit.

j-low:
Ignore that last commit message; that was meant for a Hugo commit. There is another update coming with the _actual_ QA feedback implemented.

j-low:
The most recent build represents the first version ready for QA.  There are a few judgment calls I made we should discuss, e.g.:
* I added a `check for a no_toc` variable in the front matter to hide/show the righthand TOC column (there might be a more ideal way of doing this)
* Below 1072px the right hand TOC goes away and is instead rendered in its old position; this is because the screen real estate begins to get very narrow and those extra gutter pixels are needed to  maintain a functional layout.

There are likely other things that will probably come up in the QA process but perhaps we can have a brief stakeholder check-in early next week to run through it all.

rmloveland:
@j-low I **love** the new layout!  The highest compliment I can pay is this: once this is out I can stop using my custom page-widening CSS for the site (which was sorta broken anyway as I'm not a developer).  Thank you!

jseldess:
This is looking awesome, @j-low! 

A few observations and requests:

- I think we should experiment with removing the bullets in the toc, as seen in docs like **[this]https://developers.google.com/identity/protocols/OAuth2UserAgent). I think the toc line space could be tightened somewhat, too.
- In cases where the toc is longer than the screen, you can't get to the bottom of the toc unless you scroll to the bottom of the page. Example: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/ffe40a53f16e043aa0ec834a8beeb5cf517f8fc1/stable/recommended-production-settings.html. I think in this case (toc is longer than screen), we should use a scrollbar. Example: https://developers.google.com/identity/protocols/googlescopes
- I think while we're doing this, we might as well reduce the whitespace between the sidenav and the main content, and between the topnav and the rest of the page. Especially now that we're using so much horizontal space, the padding feels excessive. I don't mean to focus on them at the exclusion of all others, but the google docs referenced above are a lot tighter than ours.
- With the toc in the right gutter, the green background highlighting also feels strange. What do you think about using a left border like we currently do for nested selected items in the sidenav. Also an example: https://firebase.google.com/docs/android/setup
-

j-low:
@jseldess great stuff, thanks for the feedback.  I like all the suggestions.  The one about removing the green background @kuanluo had also suggested too but I had forgotten to remove it in this PR.  Bullets can definitely go.  I will experiment with making the TOC itself scrollable as well as removing all the extra whitespace.  I will notify the docs channel when the PR is updated.

@rmloveland glad this all seems to be an improvment!

kuanluo:
@j-low @jseldess 

I echo all of Jesse's feedback. 

> Re: the green background: do we want it to be removed when the screen is resized and the TOC is rendered in the original area?

When the screen is resized and TOC is rendered in the original area, we can also get rid of the green background, depend on the way it's redesigned. In Firebase's case, for example, the right blue sidebar works well in smaller screens.

kuanluo:
@j-low @jseldess Looks great! 

- Can we get rid of the gray bar on the right (screenshot below) before any scroll is needed? It shows up on TOC as well.  
image
- Is it possible to highlight the line of TOC as people scroll to the content? Ideally, if I clicked on a TOC line, the page jumps to that header, and the TOC line should be highlighted.

j-low:
@kuanluo

* re: browser scroll bar: now removed in most recent PR update
* re: highlighting the line: it's definitely something we can do but it's going to take me the rest of the day to implement and test it.  If it's worth delaying the PR I can go ahead with working on it.
* re: Drift widget/help button: this is something we should meet about as there aren't any simple solutions
* re: green banner: are you incognito and/or cookie-disabled?  If so the banner will appear every time, as the behavior is cookie-based.

jseldess:
This is looking better and better! Thanks, @j-low. I just messed with some things in-browser to get it to look like this:

screen shot 2018-06-27 at 9 48 25 am

The main changes were:
- For section headings in left nav (Get Started, Develop) and for the toc heading, I used 14px font size.
- For everything else in left nav and toc, I used 13px font size.
- For toc heading, I changed it from "On This Page" to just "Contents", removed the special font, and bolded. So it looks exactly like section headings in the left nav. 
- For both left nav and toc, I reduce line height to 16px. Also for nested layers of the left nav. 
- I removed the green left border for nested layers in left nav.
- I added the green left border to the entire toc to mark the separation a bit more (`border-left: 4px solid rgb(105, 204, 0)`). Perhaps this won't be necessary once other changes are made, not sure.
- I also think we should change the "Contribute" option to just be an edit pencil or something, to remove more verbal noise. But that's something we can do later. 

Thoughts?

kuanluo:
@jseldess @j-low agree it's looking better and better! Agree with all of your edits, with some comments below.

> For section headings in left nav (Get Started, Develop) and for the toc heading, I used 14px font size.

Would like to see 13px font size + bold style for the section headings. 

> I added the green left border to the entire toc to mark the separation a bit more (border-left: 4px solid rgb(105, 204, 0)). Perhaps this won't be necessary once other changes are made, not sure.

The border-left feels a little thick. We can try 2px and gray? 

> I also think we should change the "Contribute" option to just be an edit pencil or something, to remove more verbal noise. But that's something we can do later.

We can also try change the color of the Contribute font size to gray instead.

j-low:
@kuanluo @jseldess: style updates in new build.  Forthcoming (i.e. not included in latest build):
* max-width container on content
* fix for weird nav highlighting issue

kuanluo:
This is looking great, @j-low !

j-low:
cc @jseldess: check out the latest build.  I _think_ I got the sidebar issues figured out.  The one drawback is that the page now remembers what sidebar nav items you had open however-- since we're leaning towards defaulting them to being open-- I don't think that's going to get in the way all that much.

jseldess:
Hmm, I like collapsing the nested levels of the sidenav when moving to a different section, but this isn't a blocker if this is tricky to do right now. Generally, this looks better. 

A few remaining issues:

- I polled in docs, and it looks like we should make the sidenav sections all caps again. Sorry for that churn. 
- Do we need to adjust the size of the version selector as well?
- On pages that have `toc: false` and don't have any `<div id="toc">` on page, can we not render the toc in the right gutter? In fact, it'd be good to know what conditions we're looking for now for the toc.
- I think we need to look at the font sizing in the mobile sidenav:

screen shot 2018-06-27 at 5 38 02 pm

rmloveland:
As of latest build, AFAICT the area taken up by the actual content is not that much wider than in the version of the site it replaces (screenshots below).  I don't know if this work is explicitly trying to address #2195 but the content is ~10% wider and not much more vertically dense (with respect to the "fitting more content on the screen" sense described in the linked issue).

screenshot 2018-06-28 10 54 48

screenshot 2018-06-28 10 54 52

However I admittedly have a wide-page bias for technical docs, as you can see from this screenshot of the current site with my custom CSS applied.

screenshot 2018-06-28 11 04 36

Having said all of that, here is a web page created by an actual typography expert which uses extremely wide margins to limit the line width to , so "it's complicated" 🤷‍♂️ .

https://practicaltypography.com/line-length.html

In closing, I do think there is something to the idea that the large amount of whitespace makes it hard for a technical audience to read effectively since (assumption) they need to load more info into RAM than is being presented on a single screen (which I think is the core of what @benesch  was getting at in that issue).

j-low:
@rmloveland: what you raise reflects an open question.  I tend to concur with your analysis that these changes don't do much to enhance information density.  The current max-width is set at 1500px to conform with the changes being made to the Hugo site.  I'm not sure we can achieve both perfect site-wide conformity while also meaningfully impacting information density.  

I have tried experimenting with a max-width of 1800px which I think is a better width from the info density standpoint.  As a developer, I lean towards making docs as useful and as quickly-navigable as possible.  The amount of times that I navigate between a tech product's docs and their marketing site is near-zero and I suspect that is similar for most devs.  I am inclined to prioritize information density (and thereby usability) over site-wide conformity.  Whether we go in that direction is something we need to establish collectively.

j-low:
@jseldess  re: `toc: false` and not rendering `#toc-right`: we can definitely not render the TOC if the front matter param is set to `false`.  What I'm noticing is that every page has `toc: false` in the front matter yet clearly most pages have a TOC which seems to mean that the `toc` param is mostly unused.  This is confusing to make sense of.  Is it possible we can take this opportunity to establish a pattern in the codebase that makes clearer use of the `toc` param?  Specifically I think that setting `toc: false` should be the sole determinant of whether or not the TOC renders in the righthand column or in the main content area if possible.

rmloveland:
>  I am inclined to prioritize information density (and thereby usability) over site-wide conformity. Whether we go in that direction is something we need to establish collectively.

Totes hear you on this @j-low !  I def can't claim to have answers, only questions. Thank you for your hard work to make our docs site better while also taking into consideration a lot of internal and external users :-)

jseldess:
@j-low, @rmloveland, I totally agree that we haven't completely solved the info density question. But these changes are certainly a step in the right direction. I'd prefer to follow-up separately on the question of having docs diverge more from the marketing site, in terms of max-width and otherwise. We discussed this last year and decided it was too early. We should revisit.

@j-low, the toc situation is not intuitive right now. Prior to this change, here's how it worked:

- If you want a toc at the very top of the page, set `toc: true` and that's it. 
- If you want a toc somewhere else on the page, set `toc: false` and put `<div id="toc"></div>` where you want the toc to show up. 
- If you don't want a toc at all, just set `toc: false`. 

There's another front-matter flag to determine if you want just h2 and h3 to show up in the toc or just h2. If the former, do nothing. If the latter, add `toc_not_nested: true` to the front matter. 

With your changes, I think we could simplify and just say: 

- If `toc: true` is in front matter, show toc in right gutter. 
- Otherwise, don't. 

This would require us to refactor the pages that are putting toc in non-default places. I think it wouldn't be too hard to identify and fix those with regex or a script though. Want me to take that on before we merge this? Or after?

jseldess:
Just noticed a weirdness with the right toc: No matter which link you click, the first one is always highlighted. For example, in the follow case, I clicked "SQL Shell Welcome", but "Synopsis" is highlighted:

screen shot 2018-06-28 at 5 44 31 pm

j-low:
@jseldess that behavior you saw is b/c I started working on scroll highlighting in the `#toc-right` (which wasn't completed as of last Friday).  I just finished that implementation up which should fix that issue as well as achieve highlighting based on page scroll position.

jseldess:
h3s aren't getting highlighted in right toc: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/502ea07aeac303e780f6f0c977a5cb99d4446af6/dev/build-a-python-app-with-cockroachdb.html

j-low:
@jseldess PR updated with the changes we discussed earlier (e.g. TOC sub-item highlighting, merge conflicts resolved; first tier nav items default to open on all versions).

lhirata:
This looks great!

One minor thing I noticed is when you use search, the version toggle disappears. Since the search is limited to the selected version, it would be helpful to have an indicator of what version you're looking at.

j-low:
@lhirata I believe that ~feature~ bug has been in place for some time.  I think this is something we could tackle in the near-ish future.  @jseldess and I spoke about it a few months ago in the context of a broader discussion about better visibility around versioning-- search results page included.

jseldess:
@j-low, all pages now either have `toc: true` or `toc: false`. You can now put in log to show the toc in the right column only if `toc: true`.

PR #3291, opened 2018-06-21
Update Kubernetes docs to reflect that pods aren't "Ready" before init

a-robinson:
This is true as of the updates that we made to the Kubernetes config for
v2.0.


a-robinson:
I've gone and added a note to the initialization step that the pods should now be ready. Still look ok?

jseldess:
Still LGTM. Thanks, @a-robinson!

PR #3296, opened 2018-06-25
Update alter-view.md

Princesso:
Added copy to clipboard function to alter-view.md

### DIFF
--- a/v2.0/alter-view.md
+++ b/v2.0/alter-view.md
@@ -28,6 +28,7 @@ Parameter | Description
 
 ## Example
 
+{% include copy-clipboard.html %}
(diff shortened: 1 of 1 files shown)


rmloveland:
Thank you @Princesso !

jseldess:
@rmloveland, although these changes look perfect, they mess up the html rendering: https://www.cockroachlabs.com/docs/stable/alter-view.html

It's a pain, but we always need to check the rendered version before merging. On this page, we need to  put `<div>` tags around the sql diagram include to fix the problem. Can you please do that real when you have a sec?

rmloveland:
@jseldess on it

Issue #3301, opened 2018-06-27
Make htmltest validate HTML markup (not just links)


rmloveland:
Discovered the hard way by merging #3296 without visually inspecting the build output that unfortunately [htmltest](https://github.com/cockroachdb/htmltest) does not actually validate HTML output and will happily show green even though the site is generating invalid/broken HTML with unprocessed markdown inside to be served to users, e.g., 

```
         <polygon points="707 17 699 13 699 21"></polygon>
      </svg></div>

## Parameters

Parameter | Description
----------|------------
`IF EXISTS` | Rename the view only if a view of `view_name` exists; if one does not exist, do not return an error. 
`view_name` | The name of the view to rename. To find view names, use:<br><br>`SELECT * FROM information_schema.tables WHERE table_type = 'VIEW';` 
`name` | The new [`name`](sql-grammar.html#name) for the view, which must be unique to its database and follow these [identifier rules](keywords-and-identifiers.html#identifiers). 

## Example

<div class="copy-clipboard">
  <div class="copy-clipboard__text">copy</div>
  <svg id="copy-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 12"><style>.st0{fill:#A2A2A2;}</style><title>icon/buttons/copy</title><g id="Mask"><path id="path-1_1_" class="st0" d="M4.9 4.9v6h6v-6h-6zM3.8 3.8H12V12H3.8V3.8zM2.7 7.1v1.1H.1S0 5.5 0 0h8.2v2.7H7.1V1.1h-6v6h1.6z"/></g></svg>
  <svg id="copy-check" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 10"><style>.st1{fill:#54B30E;}</style><path id="path-1_2_" class="st1" d="M3.8 9.1c-.3 0-.5-.1-.6-.2L.3 6C0 5.7-.1 5.2.2 4.8c.3-.4.9-.4 1.3-.1L3.8 7 10.6.2c.3-.3.9-.4 1.2 0 .3.3.3.9 0 1.2L4.4 8.9c-.2.1-.4.2-.6.2z"/></svg>
</div>
```

In this case I'm not even sure why Jekyll is generating invalid markup - as shown in the code snippet above, the SVG railroad diagram has been wrapped in `div` tags (which is admittedly a cargo culted solution; it's unclear why it works or doesn't work).

jseldess:
We have closed this issue because it is more than 3 years
[...]

PR #3302, opened 2018-06-27
Finish wrapping all 2.0 SQL diagrams in div tags

rmloveland:
Related to need for HTML output validation as noted in #3301.

Left the diagrams which were already wrapped alone, including those with
`section` tags, to keep this commit smaller/easier to read.


Amruta-Ranade:
LGTM 🎉

PR #3307, opened 2018-06-28
Update/specify the eval order in a way compatible with pg

knz:
Relevant pg docs:

https://www.postgresql.org/docs/10/static/sql-expressions.html#SYNTAX-EXPRESS-EVAL

https://www.postgresql.org/docs/10/static/functions-conditional.html#FUNCTIONS-CASE

https://www.postgresql.org/docs/10/static/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL

### DIFF
--- a/v1.0/sql-expressions.md
+++ b/v1.0/sql-expressions.md
@@ -336,7 +336,7 @@ The subscripted expression must have an array type; the index expression
 must have type `INT`.  The result has the element type of the
 subscripted expression.
 
-## Conditional Expressions and Boolean Short-Circuit Operations
+## Conditional Expressions
 
 Expressions can test a conditional expression and, depending on whether
 or which condition is satisfied, evaluate to one or more additional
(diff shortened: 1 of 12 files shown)


andy-kimball:
:lgtm:

+@knz, did you verify that Postgres guarantees short-circuit behavior for the operators other than `CASE`? I don't want to go beyond what Postgres does...

---

Review status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale)

---



*Comments from [Reviewable](https://reviewable.io/reviews/cockroachdb/docs/3307#-:-LG5vPoG6RPWc10TGCL5:b-cc9639)*

knz:
> did you verify that Postgres guarantees short-circuit behavior for the operators other than CASE? I don't want to go beyond what Postgres does...

Have you checked the linked documentation? I put the links in the commit message exactly because I wanted you to satisfy yourself this is indeed the case.

(Regarding the IF syntax. This syntax is not provided by PostgreSQL but as we discussed before it should be desugared to CASE, so the CASE semantics apply.)

knz:
I fixed the broken links too.

knz:
I'll merge this for now, we can fix further in separate PRs if needed.

PR #3163, opened 2018-05-17
Initial draft for systemd process

Amruta-Ranade:
Closes #1772 

@jseldess @mberhault This is an initial draft of the systemd process. For now, I have updated only the Digital Ocean deployment guide. Once the content is finalized, I will port the changes to the other deployment guides.

To-Do:
- Figure out where to place the sample configuration templates
- Edit the "Scale the cluster" sections
- Port content to other deployment guides

(Update: All To-Dos completed)

Future to-dos:
- Figure out how backup/restore and upgrades work with systemd

PR #3352, opened 2018-07-03
Update --host and --advertise-host guidance in deployment tutorials

### LINKED ISSUE #1174: cli: clarify --host and --advertise-host usage in cockroach start
We should make sure the `--host` and `--advertise-host` flag descriptions are correct and helpful, and we should add better examples for specific scenarios.

@mberhault, in a thread with a customer:

```
`--host` controls which interface the server listens on. if you want it to tell the other nodes how to reach it (eg: hostname means nothing to the other nodes) you should use `--advertise-host`

[12:15]  
so let's say you have the following VM: hostname: `my-host`, ip: `192.168.1.1` and `my-host` does not resolve on the other hosts

[12:16]  
if you specify `--host=192.168.1.1`, it will only listen on eth1 (or whatever interface) and (I think) advertise `192.168.1.1` to the other nodes

[12:16]  
if you specify `--advertise-host=192.168.1.1`, it will listen on all interfaces but tell other nodes to use `192.168.1.1`
```

### LINKED ISSUE #1770: Networking guidance/scenarios
Customers continually experience problems deploying CockroachDB because of issues with networking. While an operator's network lies beyond the purview of CRL, it's crucial the documentation provide best practices and guidance for various scenarios.

Need to cover:

- Basics/best practices 
    - Hostnames/firewalls (as started in https://github.com/cockroachdb/docs/pull/1729)
    - `--host` vs. `--advertise-host` usage (#2883, #1174)
- Scenarios
    - Floating IPs (https://forum.cockroachlabs.com/t/floating-ip-addresses/1518)
    - Changing IPs based on maintenance (happens when racks get moved around, etc.)
    - Private IPs not routable across nodes/datacenters (#2538, https://forum.cockroachlabs.com/t/bind-cannot-assign-requested-address/1152/9)

### LINKED ISSUE #2883: host vs advertise-host clarification
See https://github.com/cockroachdb/cockroach/issues/23461

@bdarnell says: 

> I think you want --host (which controls which network interface gets bound), not --advertise-host in this case.
> 
> In a single-node cluster, there's no one to advertise to, so --advertise-host has no effect (but in a multi-node cluster, --advertise-host is more likely to be useful than --host). This is subtle and we should at least document it better.

@MarkHerHold says:

> For anyone coming across this issue in the future, this is what to use for local development:
> `cockroach start --insecure --host localhost`
> IMO, the brew start instructions should be updated.

jseldess:
This is a general follow-up to https://github.com/cockroachdb/docs/pull/3163.

Fixes #1174.
Fixes #2883.
Fixes #1770.


jseldess:
Still need a broader doc solution for networking scenarios/guidance: https://github.com/cockroachdb/docs/issues/1770.

knz:
The sentence "this has the node listen on all of its interfaces and communicate with other nodes on `os.Hostname()`" is nonsensical (from a tech perspective). Some general concerns:

- `os.Hostname()` is a go-specific notion. The audience will have no idea what this refers to. What you need to use is instead the phrase "the server's canonical host name" and the reader will infer the rest.

- the verb "communicate" does not conjugate as "the node will communicate with with other nodes on a hostname". You can't "communicate on a hostname".

- there's a mix up of interfaces, advertised addresses and hosts in this story.

Here's a suggestion:

> These flags control how the node sets up network connections and with other nodes and clients. `--host` determines which local interface it listens to. This mainly pertains to the local routing and firewall rules applicable on the server where the node is running. This flag can be set to either a IP address or a name that resolves to one of the host's network interfaces. Separately, `--advertise-host` determines which IP address or host name is advertised to other nodes, which they must be able to use to connect back to this node.

> The effect of the two flags can be summarized as follows:

|   | `--host` not specified | `--host` specified |
|---|---|----|
| `--advertise-host` not specified | Listen on all interfaces. Advertise the canonical host name to other nodes. | Listen on the interface specified by `--host`. Advertise the value of `--host` to other nodes. |
| `--advertise-host` specified | Listen on all interfaces. Advertise the value of `--advertise-host` to other nodes. | Listen on the interface specified by `--host`. Advertise the value of `--advertise-host` to other nodes  |

> When running nodes on a LAN or VLAN where the local IP address or name are visible anywhere on the network, both flags can be set to the same value.  If cluster nodes are running over a WAN, for example in different data centers, there is a different between the private IP address or host name that a node knows about itself (this you can use with `--host`) and the public IP address or host name that can be used to reach it from other DCs (this you must specify with `--advertise-host`).

> Be careful about the information advertised to other nodes (either via `--advertise-host`, or when using `--host` when `--advertise-host` is not specified): if using a host name, it must be resolvable from all nodes, and when using an IP address, it must be routable from all nodes.


You should include this information in a separate section (so we can have callouts and tables) and not merely as a few sentences in the table for command-line flags.

jseldess:
This is very helpful. Thank you, @knz.

jseldess:
@knz, @bdarnell, @mberhault, PTAL at the latest commit. I leveraged Raphael's feedback but may still have gotten some terms or facts wrong.

jseldess:
___
*[_includes/v1.1/prod-deployment/secure-start-nodes.md, line 61 at r2](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHi7583-4ydhBLb1Ko-:-LHiLJ4g09ZroMftew_6:bwtndmz) ([raw file](https://github.com/cockroachdb/docs/blob/3504ff6b865c7e2eff8e5da6e272e842e5caaaa2/_includes/v1.1/prod-deployment/secure-start-nodes.md#L61)):*

This text for --host/--advertise-host is still confused. It should not refer to `os.Hostname()`. 
</blockquote></details>

Sorry. Never meant to edit 1.1 docs. Reverted the change.

jseldess:
___
*[_includes/v2.0/prod-deployment/insecure-start-nodes.md, line 52 at r2](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHi7YK3AFzRYA08dg1A:-LHilyCVCQIwdGUxgKHa:b-x4cncv) ([raw file](https://github.com/cockroachdb/docs/blob/3504ff6b865c7e2eff8e5da6e272e842e5caaaa2/_includes/v2.0/prod-deployment/insecure-start-nodes.md#L52)):*

I see you are repeating this text over and over in many places. This make future maintenance of this text unnecessarily complex. May I suggest that you just place here a reference to a separate section and detail the argument in that section. The section would be written a file that's included from the individual pages.
</blockquote></details>

Unfortunately, I think repeating in these places is the least burdensome for users. In each version, this content appears 4 times: twice in the insecure start nodes include (one for manual, once for systemd) and twice on the secure start nodes include. I think I'd prefer duplication over too many includes within includes. Also, it doesn't seem possible to include just a few table rows into a markdown table, based on my testing.

jseldess:
___
*[v2.0/recommended-production-settings.md, line 137 at r5](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHoHnMMD4NG-Jv1_xss:-LHv9gyf7u3DvyNHdbZw:bqqqax0) ([raw file](https://github.com/cockroachdb/docs/blob/90f814aec74d9ba529aab21bfde9b70e0bb14a30/v2.0/recommended-production-settings.md#L137)):*

Whoops. I still have to deal with this.
</blockquote></details>

Done now, I think.

jseldess:
OK, thanks again, @knz. I'm OOO through next Wed, but I'll do another round of revisions once I'm back. It'll be nice to finally get this as useful and clear as possible for users!

jseldess:
___
*[v2.0/recommended-production-settings.md, line 115 at r7](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHwX-Si8WdceGIbM1fk:-LIB9XnZ8wwR6Hb8Q_-4:bh715vf) ([raw file](https://github.com/cockroachdb/docs/blob/f3feac2b2e54b7106ee08edd6e29cfc6ddca3529/v2.0/recommended-production-settings.md#L115)):*
@knz, this conflicts somewhat with the exchange above, where you and @bdarnell said we should talk about addresses not interfaces for `--host`:

> knz++. This should be "address", not "interface". (concrete example: on some platforms you can bind to 127.0.0.2, which is a different address on the loopback interface).

I also see that in this [related PR](https://github.com/cockroachdb/cockroach/pull/27800/files#diff-b7942517c0b2e2181a53a1ee4d6c9042R328), you're using the language, "address or hostname to listen on. I'm not going to focus on the language in that PR too much right now cause I want to get it right here first. 

So before doing another round of revisions, I'd like to settle on some language. How about we say this for the basic `--host`/`--advertise-host` descriptions above the table:

> `--host` determines which IP address the node listens on for connections from other nodes of the cluster and from clients. If left unspecified (recommended), the node listens on all IP addresses. If set to a hostname, the value must resolve to one of the node's IP addresses.
> `--advertise-host` determines the IP address or hostname to tell other nodes to use. If using an IP address, be routable from all nodes; if using a hostname, it must be resolvable from all nodes.This value must route to an IP address the node is listening on.

I'm tempted to leave this even simpler, and have the table spell out the details:

> `--host` determines which IP address(es) the node listens on for connections from other nodes and from clients. 
> `--advertise-host` determines which IP address or hostname to tell other nodes to use.

Thoughts?

jseldess:
___
*[v2.0/recommended-production-settings.md, line 115 at r7](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHwX-Si8WdceGIbM1fk:-LIBJpR32IncTk7tmsfG:b-qet58x) ([raw file](https://github.com/cockroachdb/docs/blob/f3feac2b2e54b7106ee08edd6e29cfc6ddca3529/v2.0/recommended-production-settings.md#L115)):*

@knz, this conflicts somewhat with the exchange above, where you and @bdarnell said we should talk about addresses not interfaces for `--host`:

> knz++. This should be "address", not "interface". (concrete example: on some platforms you can bind to 127.0.0.2, which is a different address on the loopback interface).

I also see that in this [related PR](https://github.com/cockroachdb/cockroach/pull/27800/files#diff-b7942517c0b2e2181a53a1ee4d6c9042R328), you're using the language, "address or hostname to listen on. I'm not going to focus on the language in that PR too much right now cause I want to get it right here first. 

So before doing another round of revisions, I'd like to settle on some language. How about we say this for the basic `--host`/`--advertise-host` descriptions above the table:

> `--host` determines which IP address the node listens on for connections from other nodes of the cluster and from clients. If left unspecified (recommended), the node listens on all IP addresses. If set to a hostname, the value must resolve to one of the node's IP addresses.
> `--advertise-host` determines the IP address or hostname to tell other nodes to use. If using an IP address, be routable from all nodes; if using a hostname, it must be resolvable from all nodes.This value must route to an IP address the node is listening on.

I'm tempted to leave this even simpler, and have the table spell out the details:

> `--host` determines which IP address(es) the node listens on for connections from other nodes and from clients. 
> `--advertise-host` determines which IP address or hostname to tell other nodes to use.

Thoughts?
 
</blockquote></details>

Another thought (and I personally prefer this) is that we don't stress the "listen on hostname" thing too much ([cassandra](https://docs.datastax.com/en/cassandra/2.1/cassandra/configuration/configCassandra_yaml_r.html#configCassandra_yaml_r__listen_address) uses this language, too) and have a prominent note about hostnames needing to resolve and route the right way. We have the start of such a note below.

jseldess:
___
*[v2.0/recommended-production-settings.md, line 127 at r7](https://reviewable.io/reviews/cockroachdb/docs/3352#-LHwYKKU-FsVWvbjRtsV:-LIGLt1l0VMziLbmX348:b9ar8it) ([raw file](https://github.com/cockroachdb/docs/blob/f3feac2b2e54b7106ee08edd6e29cfc6ddca3529/v2.0/recommended-production-settings.md#L127)):*

This first row is still somewhat hard to read and probably needs improvement. I'll break it up:

> Start each node with `--host` set to its private address and do not specify `--advertise-host`.

OK, that's clear. The result of this is that the private address will be advertised to other nodes and clients in the private network can connect to it.

> If load balancers/clients are outside the network, start each node with `--advertise-host` set to the private address and do not specify `--host`.

That's where it breaks down. If load balancers/clients are outside the network **and the network is private** (what you're assuming on this row), then necessarily (by the definition of "private") I'd expect that **no** external load balancers or clients can connect. Instead I'd expect that if indeed they could connect, there would be a security problem.

> Also configure firewalls to allow external traffic to reach the cluster. 

By doing so, a user would effectively open their network so that it is not private any more. I doubt this is acceptable.

</blockquote></details>

Removed the confused paragraph. Thanks, @knz.

jseldess:
@mberhault gave me the verbal LGTM as well, so I'll merge. If there are other improvements to make, I can handle them in a follow-up.

PR #3374, opened 2018-07-11
Add release notes for v2.0.4

jseldess:


benesch:
Binaries are live! In we go. Thanks, @jseldess.

PR #3378, opened 2018-07-13
Performance tuning tutorial

### LINKED ISSUE #3160: Performance tuning tutorial
As discussed with @robert-s-lee, we need a step-by-step walkthrough of how to optimize read and write performance on a schema. The basic expectations is that reads should return in 2ms and writes should return in 4ms. 

We'll aim to create a tutorial that evolves from one datacenter to multiple regions. It will:

- Explain topology to use, with specific machine types.
- Explain schema.
- Explain performance expectations. 
- Guide user to start cluster, import data, and test and tune read and write performance.

jseldess:
This PR adds a 2.0 performance tuning tutorial. It starts with a single-region deployment, focusing on common SQL techniques for getting faster reads and writes. It then expands into a multi-region deployment, focusing on table partitioning.

The PR also includes a simple Python client for read/write testing. It hopefully makes it easier to run a given statement multiple times and look at the average and/or cumulative latency.

For internal testers, I've added the corresponding `roachprod` commands as comments in the `performance-tuning.md` file and in a separate comment below. 

Fixes #3160.


v2.0/performance-tuning.md
```diff
@@ -0,0 +1,1275 @@
...
+| NULL      | NULL    |       27 | {1,2,3}  |            1 |
++-----------+---------+----------+----------+--------------+
+(1 row)
+~~~
+
+The results above tell us that the `rides` table is split across 7 ranges, with 1 range's leaseholder on node 3 and the leaseholders for the rest of the ranges on node 2, and the leaseholder of the single range for `users` is on node 1.
+
+With this information, we can visualize what's happening, assuming the request is sent to node 1 and ignoring non-involved ranges:
+
+<img src="{{ 'images/v2.0/perf_tuning_join1.png' | relative_url }}" alt="Perf tuning concepts" style="max-width:100%" />
```
knz: The previous diagrams were OK to follow along; this one in contrast has me thoroughly confused: I do not understand the meaning of the arrows that jump from one "lease" box to another on node 2.

If you mean to explain that node 1 is forwarding the full table scan to every range in the "rides" table, that must be done by using:

- 6 arrows from node 1 to node 2 (you may group them together because they actually flow along a single network link)
- 1 arrow from node 1 to node 3 
- 6 arrows back from node 2 to the outside of node 1 (for the results)
- 1 arrow back from node 3 to the outside of node 1

Then for the join there is no extra arrow needed because node 1 is already leaseholder.


There should not be any arrow from node 2 to itself, and neither from node 2 to the "lease" box of node 1.

v2.0/performance-tuning.md
```diff
@@ -0,0 +1,1275 @@
...
+|                      │             | group by  | @1-@1                                                                    |
+|                      └── render    |           |                                                                          |
+|                           └── scan |           |                                                                          |
+|                                    | table     | rides@primary                                                            |
+|                                    | spans     | ALL                                                                      |
++------------------------------------+-----------+--------------------------------------------------------------------------+
+(21 rows)
+~~~
+
+This is a complex query plan, but the important thing to note is the full table scan of `rides@primary` above the `subquery`. This shows you that, after the subquery returns the IDs of the top 5 vehicles, CockroachDB scans the entire primary index to find the rows with `max(end_time)` for each `vehicle_id`, although you might expect CockroachDB to more efficiently use the secondary index on `vehicle_id`.
```
knz: Mention that we're working to lift this limitation in a future version.

v2.0/performance-tuning.md
jseldess: Thanks, @knz. Do you think this section would be sufficient without the diagram? I think they were important in the intro concepts, but I'm unsure whether they're as useful embedded in the tutorial.

knz: I think the diagram is not necessary here indeed.

knz:
Discussed offline:

- because this guide uses a fixed schema there is no discussion of how schema design impacts performance. I do not know exactly how much ground we *should* cover here, but at least we *must* mention that new apps should be careful about which index keys they define for tables (either for primary keys or secondary indexes) because of the impact on contention. Perhaps link to the FAQ entries on this topic.

- there is a section about "dropping unused secondary indexes" which is already doing a good job. Two useful additions perhaps:
  - add a quick note there that adding a foreign key constraint also adds secondary indexes implicitly, but these are not automatically removed when FK relationship is dropped. So a user should be careful to remove both if they find that they do not need a FK constraint any more.
  - the performance relationship between the number of indexes and the latency of mutations (insert/update/etc) should *also* be captured in a concise entry of the FAQ page, separately from the performance tuning guide. I think this is important because it comes up often during discussions on gitter/forum etc and the FAQ entry would be more searchable/discoverable than a section in the middle of the perf tuning guide.

v2.0/performance-tuning.md
```diff
@@ -0,0 +1,1275 @@
...
+
+1. [Create 3 instances](https://cloud.google.com/compute/docs/instances/create-start-instance) for your CockroachDB nodes. While creating each instance:  
+    - Select the **us-east1-b** [zone](https://cloud.google.com/compute/docs/regions-zones/).
+    - Use the `n1-standard-4` machine type (4 vCPUs, 15 GB memory).
+    - [Create and mount a local SSD](https://cloud.google.com/compute/docs/disks/local-ssd#create_local_ssd).
+    - To apply the Admin UI firewall rule you created earlier, click **Management, disk, networking, SSH keys**, select the **Networking** tab, and then enter `cockroachdb` in the **Network tags** field.
+
+2. Note the internal IP address of each `n1-standard-4` instance. You'll need these addresses when starting the CockroachDB nodes.
+
+3. SSH to each instance and [optimize the local SSD for write performance](https://cloud.google.com/compute/docs/disks/performance#optimize_local_ssd) (see the **Disable write cache flushing** section).
```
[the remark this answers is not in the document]
jseldess: Hmm, I assumed this was safe since it's a technique that we're featuring in our tpc-c benchmarking. @nvanbenschoten, @bdarnell, can either of you confirm that this is in fact safe and, if so, help me come up with language that clarifies that fact?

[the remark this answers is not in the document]
jseldess: Also, just to be clear, @knz, the reason I'm featuring this is, based on my testing, it significantly improves performance. @robert-s-lee is also impressed with the speedup it gives you.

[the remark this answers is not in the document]
nvanbenschoten: cc. @petermattis and @arjunravinarayan. I think they'll be able to give the most accurate answers to the following questions: "is it safe?" and "should we recommend it?".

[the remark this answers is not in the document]
arjunravinarayan: I believe it is not safe to do on cloud instances. It is what one would do to replicate the performance characteristics of an on-prem capacitor-backed SSDs would give you, and this is why we have these instructions in our performance whitepaper. But I do not believe it is suitable guidance for a performance tuning guide in general.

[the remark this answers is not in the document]
jseldess: OK, thanks, @arjunravinarayan. If @petermattis agrees that this is not safe to do on cloud instances, I'll remove this and rerun all the numbers.

v2.0/performance-tuning.md
```diff
@@ -0,0 +1,1275 @@
...
+
+    {% include copy-clipboard.html %}
+    ~~~ sql
+    > ALTER TABLE rides
+    ADD CONSTRAINT fk_vehicle_city_ref_vehicles
+    FOREIGN KEY (vehicle_city, vehicle_id)
+    REFERENCES vehicles (city, id);
+    ~~~
+
+    **Add a note about why we need vehicle_city and link to ticket. Basically, we can't have 2 foreign keys constraints on the same column (city). So we duplicate city as vehicle_city and add check constraint to ensure that they are identical. https://github.com/cockroachdb/cockroach/issues/23580**
```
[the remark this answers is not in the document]
jseldess: I added a note in the `### Schema section`. PTAL.

v2.0/performance-tuning.md
jseldess: Removed it.

v2.0/performance-tuning.md
```diff
@@ -0,0 +1,1275 @@
...
+| york"/"\xcev\xeb\x9ah\x80C<\x86\xb6\x83\x00\xc4\x1f\xc6\x04"                          |                                                                                       |          |          |              |
++---------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------+----------+----------+--------------+
+(3 rows)
+~~~
+
+This tells us the the index is stored in 3 ranges, with the leaseholders for all of them on node 1. We already know that the leaseholder for the `users` table is also on node 1.
+
+With this information, we can visualize what's happening now, still assuming the request is sent to node 1 and ignoring non-involved ranges:
+
+<img src="{{ 'images/v2.0/perf_tuning_join2.png' | relative_url }}" alt="Perf tuning concepts" style="max-width:100%" />
```
jseldess: Also removed this image.

v2.0/performance-tuning.md
jseldess: Done.

jseldess:
___
*[v2.0/performance-tuning.md, line 141 at r1](https://reviewable.io/reviews/cockroachdb/docs/3378#-LHHFRRIcrYRpM-Y5ATP-r1-141:-LIpw4o2CEqOh4r1D3Zv:b-sq4q99) ([raw file](https://github.com/cockroachdb/docs/blob/10d9ee346e10e104e49c2907588d51a798e900f2/v2.0/performance-tuning.md#L141)):*

Ditto what @arjunravinarayan said. Note that Google Persistent SSD does not flush to the storage media (effectively acting like nobarrier is set), while Local SSD does. If running a high performance database on-prem you'd get hardware that had capacitor-backed or battery-backed SSDs. Future versions of cloud datacenters may have reliable enough batteries to allow recommending this as the default (and I've heard murmurings that some current Google datacenters have reliable batteries), but right now the recommendation should be that you have to know your hardware in order to specify `nobarrier` and the cloud providers do not provide enough transparency to do this safely.
</blockquote></details>

OK. Thanks, all. I'll remove this.

jseldess:
Roachprod commands for single-region portion:

```
<!-- roachprod instructions for single-region deployment
1. Reserve 12 instances across 3 GCE zone: roachprod create <yourname>-tuning --geo --gce-zones us-east1-b,us-west1-a,us-west2-a --local-ssd -n 12
2. Put cockroach` on all instances: `roachprod run <yourname>-tuning "curl https://binaries.cockroachdb.com/cockroach-v2.0.4.linux-amd64.tgz | tar -xvz; mv cockroach-v2.0.4.linux-amd64/cockroach cockroach"
3. Start the cluster in us-east1-b: roachprod start <yourname>-tuning:1-3
4. You'll need the addresses of all instances later, so list and record them somewhere: roachprod list -d <yourname>-tuning
5. Import the Movr dataset:
   - SSH onto instance 4: roachprod run <yourname>-tuning:4
   - Run the SQL commands in Step 4 below.
8. Install the Python client:
   - Still on instance 4, run commands in Step 5 below.
9. Test/tune read performance:
   - Still on instance 4, run commands in Step 6.
10. Test/tune write performance:
   - Still on instance 4, run commands in Step 7.
-->
```

Roachprod commands for multi-region portion:

```
<!-- roachprod instructions for multi-region deployment
You created all instanced up front, so no need to add more now.
1. Install the Python client on instance 8:
   - SSH to instance 8: roachprod run <yourname>-tuning:8
   - Run commands in Step 5 above.
2. Install the Python client on instance 12:
   - SSH onto instance 12: roachprod run <yourname>-tuning:12
   - Run commands in Step 5 above.
3. Check rebalancing:
   - SSH to instance 4, 8, or 12.
   - Run `SHOW EXPERIMENTAL_RANGES` from Step 11 below.
4. Test performance:
   - Run the SQL commands in Step 12 below. You'll need to SSH to instance 8 or 12 as suggested.
5. Partition the data:
   - SSH to any node and run the SQL in Step 13 below.
6. Check rebalancing after partitioning:
   - SSH to instance 4, 8, or 12.
   - Run `SHOW EXPERIMENTAL_RANGES` from Step 14 below.
7. Test performance after partitioning:
   - Run the SQL commands in Step 15 below. You'll need to SSH to instance 8 or 12 as suggested.
-->
```

robert-s-lee:
- `A few notes about the schema` uses the `compound key`.  `composite key` is used at https://www.cockroachlabs.com/docs/v2.0/split-at.html#split-a-table-with-a-composite-primary-key to refer to the same concept.  There is a nuance to  `composite key` vs `compound key`.   Is this truly `compound key`?

- `The rides table contains both city and the seemingly redundant vehicle_city. This redundancy is necessary` leaving myself a note to double check
 
- `Leaseholder` mention this is CockroachDB performance optimization of Raft Protocol 

- on the diagram, having the same colors for leaseholder and replicas took time to explain.  https://docs.google.com/presentation/d/1Tiq3lNmU-kOtJuppSqhYh0sAps2Rcs2rl7uH7eZ-z0c/edit#slide=id.g37076dbdf1_0_608 has @kuan thoughts on having the leaseholder and replica be the same color but hallow to make it easier to identify and highlight the importance of the leaseholder.   

- the sequence diagram mention `write`, but does not mention CockroachDB waits for the write to the disk.  This is an important distinction as some other databases don't necessarily wait thus not assuming multiple nodes holding the data in RAM is good enough for durability.   CockroachDB offers strong durability and maybe a quick description.  This will also aid in why having a fast disk is important to CockroachDB as well as the fast network.

- assuming network and io as the major latency bottleneck, 1ms network hops x 2 for read result in `Retrieving a single row based on the primary key will usually return in 2ms or less:` why response time of 2ms is expected and how the change in latency of network and IO will have big impact in performance.

- the same for the write.

jseldess:
Thanks, @robert-s-lee. I've changed compound to composite (based on this [wikipedia entry](https://en.wikipedia.org/wiki/Compound_key), I was just using the wrong term), and I've included more details about how the leaseholder mechanism bypasses Raft. I'm not sure how to proceed on the other points. We can talk about them in person.

jseldess:
@robert-s-lee, I expanded "Important concepts" to cover more about the Raft log and how it plays into writes, and I added a brief emphasis of network and disk i/o as performance bottlenecks. I don't think we can go into greater details in this tutorial without making the material much harder to parse. But I opened a [docs issue](https://github.com/cockroachdb/docs/issues/3534) to document the read and write paths in more detail separately. PTAL.

tim-o:
@jseldess  - this is really incredible. I didn't run into any edits or nits worth picking. I second Sean's thoughts on the epic scale. One thought while I was reading through it: maybe a series of blog posts? In any case, agreed that publishing now and modularizing later is a good approach.

PR #3405, opened 2018-07-18
Add image to training home page

### LINKED ISSUE #3393: Improvements to design of /training home page
@jseldess 

Now that  we're driving some traffic to this page, it's becoming apparent that its role as a landing page (home page for Training content) needs more investment. The current experience is a bit confusing when you land on it. 

For example: Because the page is so short, the "was this helpful: yes/no?" is taking up a prominent spot for visitors, and it appears as if it is the desired action you want a visitor to take. 

Some interim recommendations: 
1. Replace the "What's first?" section with a big CTA that says "Get started with training". This would click through to the first module. Could reuse the CTA style used on the download page to select OS. My recommendation would be to have it background-green by default, and bigger than the OS-selection CTA.

2. Hide (remove?) the "Was this page helpful?" module. Again, because the page is so short, it is jumping up the page to look like primary content. Because this is just a landing page that people should navigate away, we shouldn't be missing much valuable feedback on the yes/no click.

jseldess:
Fixes #3393.

@jess-edwards, with the image on the page, do you still think we need to make the "What's first" thing more visual?


jseldess:
TFTR, @jess-edwards!

PR #3423, opened 2018-07-26
openssl docs: allow multiple certificates for the same subject.

mberhault:
We create node certificates with subject `/O=Cockroach/CN=node`. By default,
openssl doesn't allow multiple certificates with the same subject to be
signed (is knows this by checking the local database files).

There are multiple ways of getting around this (delete/truncate the
database files, add more fields to the subject that vary for each node),
but this one is by far the easiest.


mberhault:
Also made the banner about setting `commonName` to `node` red and added a comment not to change it in the config. This keeps popping up.

Issue #2944, opened 2018-04-17
Under- or un-documented features


jseldess:
For 2.1, we need to expand our docs on:

- [x] Window functions (#710)
- [x] Online schema changes (#3492)
   - Only have this [blog post](https://www.cockroachlabs.com/blog/how-online-schema-changes-are-possible-in-cockroachdb/)
   - Priority 1: Also mention how we're working on improvements.

jseldess:
@bdarnell, @petermattis, what is our current opinion about documenting `crdb_internal`?

jseldess:
Comment from @bdarnell on `crdb_internal`:

> In general my thinking is that crdb_internal should remain internal-only and undocumented. we should take the stuff that's useful enough to document and put it behind SHOW commands or maybe make a separate `crdb_catalog` virtual schema with more stability/documentation guarantees

jseldess:
These features tracked separately, as potential docs projects for 2.2.:

- [ ] Joins (#534) 
- [ ] Placeholder support (#2168) 
- [ ] Server-side Prepared Statements (#1333)
- [ ] Interleaved secondary indexes (#761)

jseldess:
Window functions and online schema changes are live, so closing this.

PR #3426, opened 2018-07-26
Add docs for window functions

rmloveland:
Addresses the window functions portion of #2944

Includes:

- Introduction to and overview of window functions and related terms.

- A (possibly helpful) diagram showing how window functions work.

- Various examples of window functions using the soon-to-be-open-sourced
  'movr' database.

- Added a note to the existing `PARTITION BY` statement docs that links
  to the window functions page just in case the user is searching for
  the window functions version.


rmloveland:
Direct link to the new window functions page:
http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/4760e0b63348696e50e7ca38367c71f41593d996/dev/window-functions.html

rmloveland:
> There are some missing things: we already support specifying frames using ROWS (full support) and RANGE (only UNBOUNDED PRECEDING/CURRENT ROW/UNBOUNDED FOLLOWING support as of right now)

I added the following sentence so that we at least make some mention of frame clauses.

> If you are not seeing results you expect from a window function, this behavior may explain why.  You may need to specify the frame boundaries explicitly using a *frame clause* such as `ROWS BETWEEN <start> <end> [exclusion]` (fully supported) or `RANGE BETWEEN <start> <end> [exclusion]` (only `UNBOUNDED PRECEDING`/`CURRENT ROW`/`UNBOUNDED FOLLOWING` supported).

Would be good to add more info and examples in a future PR (Unfortunately I have to move on to some other work at the moment!).

Amruta-Ranade:
Incredible doc, @rmloveland! Was fun to try out and learned a lot! 🎉

Issue #3431, opened 2018-07-30
Pages not highlighted in sidenav when url does not include .html


jseldess:
https://www.cockroachlabs.com/docs/v2.1/sql-optimizer

lhirata:
@jseldess I added [SQL Optimizer](https://www.cockroachlabs.com/docs/v2.1/sql-optimizer.html) in the sidenav under Reference > SQL > Performance Optimization. It shows up when you add the .html at the end of the URL.

jseldess:
OK, thanks, @lhirata. Weird that this doesn't work without the `.html`. @j-low, do you have any insights?

lhirata:
Seems like it happens to other pages too if you delete the .html. You can still see the page but it doesn't show up in the nav.

jseldess:
Thanks, @lhirata. I'm converting this issue accordingly and assigning to @j-low.

j-low:
Hmmm, I don't have any immediate ideas. It seems like all the sidebar data links use `.html`...is there a reason we would avoid appending `.html`?

jseldess:
This is fixed.

PR #3434, opened 2018-07-30
Highlight sidebar links for URLs without explicitly-appended `.html` 

j-low:
__[description]:__
* re: #3431 
* handle active sidebar highlighting for URLs without .html explicitly appended


jseldess:
LGTM.

PR #3488, opened 2018-08-01
Multiple --store on a 3 node cluster limitations documented.

vaLski:
Added documentation about well known limitations of cockroach 2.0.x in a 3-node cluster configuration with multiple --stores per node.

Also added suggestions how one can work-around this but increase failure domain from disk "level" to machine "level"


jseldess:
Thanks for this improvement, @vaLski. This information is clearly important, but I might want to implement it differently and in a few different places. I'll review more closely soon and get back to you.

knz:
this advice (or some reworded version of it) should be considered for inclusion in our docs

vaLski:
> this advice (or some reworded version of it) should be considered for inclusion in our docs

Thanks for getting back to me.  If you find it useful free to use it as it is. If there are certain issues that should be fixed or reworded I assume that it would be easier to simply edit the MR instead  of going back and forth what I should say. However if you prefer the later way I am ok to adjust the commits according to your specifications.

jseldess:
I've taken over these changes in a new PR: https://github.com/cockroachdb/docs/pull/4320. Thanks again for submitting this change, @vaLski, and my apologies for the long delay.

PR #3498, opened 2018-08-06
2.1 Training: Intro + Ops Basics

lhirata:
Updates to 2.1 Training - Intro + Ops Basics section.

Not included:

- New lab content for Production Deployments / Debug Logs. _Is this needed?_


jseldess:
> New lab content for Production Deployments / Debug Logs. Is this needed?

@lhirata, no. After chatting with @rolandcrosby, I think we're going to consolidate these slide-only modules into more of a see also list. I'll handle that at some point.

PR #3506, opened 2018-08-06
Add docs to describe new default database(s)

### LINKED ISSUE #3206: sql: sanitize the "no current db" story
PR: https://github.com/cockroachdb/cockroach/pull/24735

From release notes:
> New clusters and existing clusters [upgraded](../v2.1/upgrade-cluster-version.html) to this version of CockroachDB will include two new empty databases, `defaultdb` and `postgres`. The `defaultdb` database is automatically used for clients that connect without a current database set (e.g., without a database component in the connection URL). The `postgres` database is provided for compatibility with PostgreSQL client frameworks that require it to exist when the database server has ben freshly installed. Both new databases behave like any other regular database and, if deemed unnecessary, can be [manually deleted](../v2.1/drop-database.html). [#24735][#24735] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #3206.

Summary of changes:

- Update [Client connection parameters][1] to note that `database` param
  defaults to `defaultdb` as of 2.1

- Update [Learn SQL][2] and [SHOW DATABASES][3] to describe `defaultdb`
  and `postgres`

[1]: https://www.cockroachlabs.com/docs/dev/connection-parameters.html
[2]: https://www.cockroachlabs.com/docs/dev/learn-cockroachdb-sql
[3]: https://www.cockroachlabs.com/docs/dev/show-databases

### DIFF
--- a/v2.1/learn-cockroachdb-sql.md
+++ b/v2.1/learn-cockroachdb-sql.md
@@ -14,75 +14,6 @@ Use an interactive SQL shell to try out these statements. If you have a cluster 
 CockroachDB aims to provide standard SQL with extensions, but some standard SQL functionality is not yet available. See our [SQL Feature Support](sql-feature-support.html) page for more details.
 {{site.data.alerts.end}}
 
-## Create a Database
-
-CockroachDB comes with a single default `system` database, which contains CockroachDB metadata and is read-only. To create a new database, use [`CREATE DATABASE`](create-database.html) followed by a database name:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> CREATE DATABASE bank;
-~~~
-
-Database names must follow [these identifier rules](keywords-and-identifiers.html#identifiers). To avoid an error in case the database already exists, you can include `IF NOT EXISTS`:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> CREATE DATABASE IF NOT EXISTS bank;
-~~~
-
-When you no longer need a database, use [`DROP DATABASE`](drop-database.html) followed by the database name to remove the database and all its objects:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> DROP DATABASE bank;
-~~~
-
-## Show databases
-
-To see all databases, use the [`SHOW DATABASES`](show-databases.html) statement:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> SHOW DATABASES;
-~~~
-
-~~~
-+---------------+
-| database_name |
-+---------------+
-| bank          |
-| defaultdb     |
-| postgres      |
-| system        |
-+---------------+
-(3 rows)
-~~~
-
-## Set the default database
-
-To set the default database, use the [`SET`](set-vars.html#examples) statement:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> SET DATABASE = bank;
-~~~
-
-When working with the default database, you do not need to reference it explicitly in statements. To see which database is currently the default, use the `SHOW DATABASE` statement (note the singular form):
-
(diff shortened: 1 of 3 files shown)


knz:
@rmloveland tip: typically we rebase and we don't merge master into the secondary branch

rmloveland:
@knz Oof, sorry - I made the mistake of using the Github web UI to fix a conflict.  Won't happen again.

v2.1/learn-cockroachdb-sql.md
```diff
@@ -10,76 +10,6 @@ This page walks you through some of the most essential CockroachDB SQL statement
 Use an interactive SQL shell to try out these statements. If you have a cluster already running, use the [`cockroach sql`](use-the-built-in-sql-client.html) command. Otherwise, use the [`cockroach demo`](cockroach-demo.html) command to open a shell to a temporary, in-memory cluster.
 {{site.data.alerts.end}}
 
-{{site.data.alerts.callout_info}}
-CockroachDB aims to provide standard SQL with extensions, but some standard SQL functionality is not yet available. See our [SQL Feature Support](sql-feature-support.html) page for more details.
```
jseldess: Why are we removing this callout?

v2.1/learn-cockroachdb-sql.md
```diff
@@ -10,76 +10,6 @@ This page walks you through some of the most essential CockroachDB SQL statement
...
-{{site.data.alerts.end}}
-
-## Create a Database
```
jseldess: Why are we removing create database and show database? Even though the defaults have changed, these still seem like relevant actions to highlight here...

v2.1/learn-cockroachdb-sql.md
```diff
@@ -143,30 +73,8 @@ To see all tables in the active database, use the [`SHOW TABLES`](show-tables.ht
 |  Table   |
 +----------+
 | accounts |
-| users    |
 +----------+
-(2 rows)
-~~~
-
-To view tables in a database that's not active, use `SHOW TABLES FROM` followed by the name of the database:
```
jseldess: I don't understand why we're removing this either. Can you clarify?

v2.1/show-databases.md
```diff
@@ -26,15 +25,18 @@ No [privileges](privileges.html) are required to list the databases in the Cockr
...
+| database_name |
++---------------+
+| defaultdb     |
+| postgres      |
+| system        |
++---------------+
+(3 rows)
 ~~~
 
+{% include {{page.version.version}}/default-databases.md %}
```
jseldess: Instead of embedding this as a callout following the example, I'd recommend we add an additional `Default Databases` section above the example. It'd make it much easier to cross-reference, if that's a need at any point. Thoughts?

knz:
@jseldess as I have explained to Rich earlier, we do not need any more to explain the "database concept" to a beginner user in the tutorial any more, because they do not need it (any more) to issue interesting SQL statements successfully.

The path to a successful SQL experience should be as short as possible and talking about databases that early would feel like a shamanic rain dance.

knz:
I do not feel strongly but I do want to point out that most tutorials for postgres and mysql skip talking about creating and using custom database names and simply assume the user has a default database already.

CockroachDB's current tutorial feels complex in comparison. There is no need for this complexity. I would prefer if we didn't need more step to onboard new users than they would need  with, say, postgres.

jseldess:
Hmm, when I do a quick "posgres getting started" search, I land on https://www.postgresql.org/docs/10/static/tutorial-start.html, which does cover creating a database. I don't want to belabor this. I'm fine removing it. But we do create databases in several other places in our getting started docs as well (all of the build an app tutorials, for example).

Issue #3337, opened 2018-07-02
importccl: support decompressing input files


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/26796

From release notes:
> Added support for de-compressing [`IMPORT`](../v2.1/import.html) files with gzip or bzip. [#26796][#26796] {% comment %}doc{% endcomment %}

Amruta-Ranade:
PR: https://github.com/cockroachdb/docs/pull/3510

Amruta-Ranade:
Addressed by https://github.com/cockroachdb/docs/pull/3510

PR #3510, opened 2018-08-07
Import doc updates: Skip n lines + support decompressing input files

### LINKED ISSUE #3093: cli: add "skip" option to IMPORT CSV DATA
PR: https://github.com/cockroachdb/cockroach/pull/23466

From release notes:
> Added the `skip` option to the `IMPORT` command. [#23466][#23466] {% comment %}doc{% endcomment %}

Amruta-Ranade:
@bobvawter Commit 1 addresses #3093 
@dt Commit 2 addresses #3337

### DIFF
--- a/v2.1/import.md
+++ b/v2.1/import.md
@@ -217,6 +217,31 @@ Skip the first *n* lines of a CSV file.
 	</tbody>
 </table>
 
+#### `decompress`
+
+Import compressed files.
+
+<table>
+	<tbody>
+		<tr>
+			<td><strong>Required?</strong></td>
+			<td>No</td>
+		</tr>
+		<tr>
+			<td><strong>Key</strong></td>
+			<td><code>decompress</code></td>
+		</tr>
+		<tr>
+			<td><strong>Value</strong></td>
+			<td>The decompression codec to be used: <code>gzip</code>, <code>bzip</code>, <code>auto</code>, or <code>none</code>. <br><br>The default option is <code>auto</code> which attempts to guess the codec based on the filename, matching the common extensions '.gz' or '.bz2' and .bz. <br><br> <code>none</code> can be used to disable any attempts at decompression.</td>
+		</tr>
+		<tr>
+			<td><strong>Example</strong></td>
+			<td>To import a file compressed using the <code>bzip</code> codec: <code>WITH decompress = 'bzip'</code></td>
+		</tr>
+	</tbody>
+</table>
+
 
 
 ## Examples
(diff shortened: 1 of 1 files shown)


v2.1/import.md
```diff
@@ -192,6 +192,58 @@ Convert values to SQL *NULL* if they match the specified string.
...
+			<td><strong>Key</strong></td>
+			<td><code>skip</code></td>
+		</tr>
+		<tr>
+			<td><strong>Value</strong></td>
+			<td>The number of rows to be skipped while importing a file</td>
+		</tr>
+		<tr>
+			<td><strong>Example</strong></td>
+			<td>To import CSV files with column headers <em>NULL</em>: <code>WITH skip = ''</code></td>
```
dt: What's `NULL` ? also should this read `skip = '1'` ?

v2.1/import.md
```diff
@@ -192,6 +192,58 @@ Convert values to SQL *NULL* if they match the specified string.
...
+			<td><strong>Required?</strong></td>
+			<td>No</td>
+		</tr>
+		<tr>
+			<td><strong>Key</strong></td>
+			<td><code>decompress</code></td>
+		</tr>
+		<tr>
+			<td><strong>Value</strong></td>
+			<td>The decompression codec to be used: <code>gzip</code>, <code>bzip</code>, <code>auto</code>, or <code>none</code>. <br><br>The default option is <code>auto</code> which attempts to guess the codec based on the filename, matching the common extensions '.gz' or '.bz2' and .bz. <br><br> <code>none</code> can be used to disable any attempts at decompression.</td>
```
dt: should `.bz` be in single-quotes as well?

v2.1/import.md
```diff
@@ -256,12 +323,27 @@ WITH
...
+### Import a compressed CSV file
+
+{% include copy-clipboard.html %}
+~~~ sql
+> IMPORT TABLE customers (
+		id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+		name TEXT,
+		INDEX name_idx (name)
+)
+CSV DATA ('azure://acme-co/customer-import-data.gz?AZURE_ACCOUNT_KEY=hash&AZURE_ACCOUNT_NAME=acme-co')
```
dt: it would probably be named `customer-import-data.csv.gz`
That said, if the file ends in `.gz` you don't actually need the decompress option at all since we could figure it out form the filename, so I'm not sure what makes sense as an example.

Amruta-Ranade: But we do have options `bzip` and `gzip` when they want to explicitly specify the codec, right?

dt: sure, you can pass `with decompress = 'gzip'` but you don't _need_ to if the filename is `.gz` since it would pick that anyway, so as an example, i'm not sure how much it communicates.

Amruta-Ranade: Hmmm..so what would be a better example?

dt: I'm not sure... maybe no extension?

dt: we could also call the example "force selection of gzip decompression codec" or something to make it clearer why you'd pass the option

Amruta-Ranade: Now I am wondering why we added the options in the first place 🤔

dt: the option is there so that if your filename _doesn't_ end in the common extension, you can force it to use a specific codec, but the default is to try to just figure it out.

Amruta-Ranade: Hmm..so the default way we want users to import decompressed files is by not specifying any options. I am going to reframe the example to reflect that.

PR #3517, opened 2018-08-08
RENAME / DROP DATABASE update

### LINKED ISSUE #3090: sql: reduce UX surprise with DROP DATABASE and ALTER DATABASE RENAME
PR: https://github.com/cockroachdb/cockroach/pull/24246

From release notes:
> `DROP DATABASE` and `ALTER DATABASE ... RENAME` now prevent the removal of a database name if that database is set as the current database (`SET database` / `USE`) and the session setting `sql_safe_updates` is also set. [#24246][#24246] {% comment %}doc{% endcomment %}

lhirata:
You cannot rename / drop a database if it is set as the current database.

Closes #3090.


lhirata:
TFTRs @rmloveland!

PR #3524, opened 2018-08-08
Describe runtime key rotation in encryption page.

mberhault:
Also mentioned key information written to log files.
Both changes occured recently.


mberhault:
Oh, also mentioned information written to the logs (also new). There will be a few more changes once another PR or two lands, but nothing major.

I threw this page together a while back. It could probably use a thorough examination and some serious polish.

PR #3527, opened 2018-08-09
Updated backup/restore/import job creation

### LINKED ISSUE #3522: Be more precise about initiation of long-running jobs
We have a few misleading statements in the docs for `BACKUP`, `RESTORE`, and `IMPORT` about exactly when an entry is actually created in the jobs table, which led to confusion in cockroachdb/cockroach#25129. Specifically, we currently make these statements (which appear to be in the docs for 1.1, 2.0, and 2.1):

[BACKUP](https://www.cockroachlabs.com/docs/dev/backup.html#viewing-and-controlling-backups-jobs):
> ## Viewing and controlling backups jobs
> Whenever you initiate a backup, CockroachDB registers it as a job, which you can view with SHOW JOBS.
> 
> After the backup has been initiated, you can control it with PAUSE JOB, RESUME JOB, and CANCEL JOB.

[RESTORE](https://www.cockroachlabs.com/docs/dev/restore.html#viewing-and-controlling-restore-jobs):
> ## Viewing and controlling restore jobs
> Whenever you initiate a restore, CockroachDB registers it as a job, which you can view with SHOW JOBS.
> 
> After the restore has been initiated, you can control it with PAUSE JOB, RESUME JOB, and CANCEL JOB.

[IMPORT](https://www.cockroachlabs.com/docs/dev/import.html#viewing-and-controlling-import-jobs):
> ## Viewing and controlling import jobs
> Whenever you initiate an import, CockroachDB registers it as a job, which you can view with SHOW JOBS.
> 
> After the import has been initiated, you can control it with PAUSE JOB, RESUME JOB, and CANCEL JOB.

These statements are misleading (or at least unclear), as there are circumstances in which a user may execute a BACKUP, RESTORE, or IMPORT query that fails before CockroachDB registers the operation as a job. Specifically, this may occur when a user specifies an inaccessible file path (which is what happened to Nate in the linked issue) or when the query is malformed in the first place.

I'm not sure the exact language to use to clarify this, but maybe something along the lines of "when CockroachDB starts to execute a backup/restore/import, it registers the backup/restore/import as a job" would be sufficient. If necessary I can also get more details about what errors the user should watch out for; just let me know.

Amruta-Ranade:
Closes #3522

### DIFF
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -92,7 +92,7 @@ Once the backup is complete, your client will receive a `BACKUP` response.
 
 ## Viewing and Controlling Backups Jobs
 
-Whenever CockroachDB starts executing a backup, it registers the backup as a job, which you can view with [`SHOW JOBS`](show-jobs.html).
+After CockroachDB successfully initiates a backup, it registers the backup as a job, which you can view with [`SHOW JOBS`](show-jobs.html).
 
 After the backup has been initiated, you can control it with [`PAUSE JOB`](pause-job.html), [`RESUME JOB`](resume-job.html), and [`CANCEL JOB`](cancel-job.html).
 
--- a/v2.0/backup.md
+++ b/v2.0/backup.md
@@ -92,7 +92,7 @@ Once the backup is complete, your client will receive a `BACKUP` response.
 
 ## Viewing and Controlling Backups Jobs
 
-Whenever you initiate a backup, CockroachDB registers it as a job, which you can view with [`SHOW JOBS`](show-jobs.html).
+Whenever CockroachDB starts executing a backup, it registers the backup as a job, which you can view with [`SHOW JOBS`](show-jobs.html).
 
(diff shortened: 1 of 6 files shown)


Amruta-Ranade:
IMO, the statement "Whenever CockroachDB starts executing a backup, it registers the backup as a job, which you can view with SHOW JOBS" implies that the job is registered only if the backup execution starts successfully, and not if the operation fails while preparing to run. But I might be wrong. 
@jseldess Thoughts?

rolandcrosby:
Yeah, I think you're right that the current wording implies that the job isn't necessarily going to be created, I just wasn't sure how explicit we need to be here. If @jseldess thinks this is clear enough I'm fine merging as is.

Amruta-Ranade:
@rolandcrosby How about "After CockroachDB successfully initiates a backup, it registers the backup as a job, which you can view with `SHOW JOBS`" ?

rolandcrosby:
I think that change would be fine, since it more strongly implies that the job only gets registered if it doesn't fail while starting.

rolandcrosby:
LGTM!

PR #3528, opened 2018-08-09
Fix typo: use 'decrease' instead of 'increase'

rmloveland:


jseldess:
@rmloveland, is this PR tied to a docs issue?

v2.0/performance-best-practices-overview.md
```diff
@@ -288,7 +288,7 @@ To avoid contention, multiple strategies can be applied:
   [normalization](https://en.wikipedia.org/wiki/Database_normalization)
   of the data to place parts of the same records that are modified by
   different transactions in different tables. Note however that this
-  is a double-edged sword, because denormalization can also increase
+  is a double-edged sword, because denormalization can also decrease
```
knz: The correct word was "increase" so this change is not warranted.

More details:

If a user decides to increase normalization (and thus decrease denormalization) because they want to decrease contention, they are reducing the number of copies.

Meanwhile more copies can increase performance because it parallelizes access to the data (by distributing the KV operations to them across more nodes/ranges).

So it's an unclear tradeoff, hence the word "double-edged".

Feel free to reformulate the paragraph to more clearly explain this trade-off.

rmloveland: Thanks for the explanation and additional information @knz - I will probably press pause on this for now, and return to it later.

rmloveland:
> @rmloveland, is this PR tied to a docs issue?

@jseldess No it just came up when I was doing some reading related to txn pipelining.  Will leave it be for now since there is much else to do. :-)

rmloveland:
Closing this since the current language is correct and there is other work to be done at this time.  Sorry for the noise folks.

PR #3530, opened 2018-08-09
Document pipelining of transactional writes

### LINKED ISSUE #3485: Transaction pipelining
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwqpUolS6pxnGsHf/rec5JVK6sW9YiZVBb
PM: @awoods187 
Eng partner: @nvanbenschoten 
PR: https://github.com/cockroachdb/cockroach/pull/26599

From release notes:
> Transactional writes are now pipelined when being replicated and when being written to disk, dramatically reducing the latency of transactions that perform multiple writes. This can be disabled using the new `kv.transaction.write_pipelining_enabled` [cluster setting](../v2.1/cluster-settings.html). [#26599][#26599] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #3485.

Summary of changes:

- Update *Writes* section of the 'Transaction Layer' doc:

  - note that as of v2.1 transactional writes are pipelined

  - Explain the effects / how it works at a high level

- Update 'Cluster Settings' page with the new flags (via the
  autogenerated docs from the CockroachDB build):

  - kv.transaction.write_pipelining_enabled
  - kv.transaction.write_pipelining_max_batch_size

### DIFF
--- a/v2.1/architecture/transaction-layer.md
+++ b/v2.1/architecture/transaction-layer.md
@@ -190,6 +190,38 @@ Whenever a write occurs, its timestamp is checked against the timestamp cache. I
 Whenever a transaction's timestamp has been pushed, additional checks are required before allowing serializable transactions to commit at the pushed timestamp: any values which the transaction previously read must be checked to verify that no writes have subsequently occurred between the original transaction timestamp and the pushed transaction timestamp. This check prevents serializability violation. The check is done by keeping track of all the reads using a dedicated `RefreshRequest`. If this succeeds, the transaction is allowed to commit (transactions perform this check at commit time if they've been pushed by a different transaction or by the timestamp cache, or they perform the check whenever they encounter a `ReadWithinUncertaintyIntervalError` immediately, before continuing).
 If the refreshing is unsuccessful, then the transaction must be retried at the pushed timestamp.
 
+### Transaction pipelining
+
+<span class="version-tag">New in v2.1:</span> Transactional writes are pipelined when being replicated and when being written to disk, dramatically reducing the latency of transactions that perform multiple writes. For example, consider the following transaction:
+
+{% include copy-clipboard.html %}
+~~~ sql
+-- CREATE TABLE kv (id UUID PRIMARY KEY DEFAULT gen_random_uuid(), key VARCHAR, value VARCHAR);
+> BEGIN;
+  SAVEPOINT cockroach_restart;
+  INSERT into kv (key, value) VALUES ('apple', 'red');
+  INSERT into kv (key, value) VALUES ('banana', 'yellow');
+  INSERT into kv (key, value) VALUES ('orange', 'orange');
+  RELEASE SAVEPOINT cockroach_restart;
+  COMMIT;
+~~~
+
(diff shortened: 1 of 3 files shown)


nvanbenschoten:
> Does txn pipelining change any of the information in the paragraphs just prior to this change in the Writing section of the Transaction layer page about txn records, write intents, etc.?

No, transaction records and write intents are not effected.

> Does txn pipelining affect any of the information in understanding and avoiding transaction contention?

Nope, there shouldn't be any changes in that regard.

> (Apart from the previous questions) Are there additional places in the documentation you think should be updated to mention this? AFAICT it is an optimization but does not necessarily affect things from a user/docs POV except by making things faster (which to be clear is awesome). Therefore, it's not clear what to document other than to note its operation at the txn layer, but I may be missing something.

The only thing I can think of is that transactional pipelining makes parallel statement execution less important and we should probably remove it as a top-level performance suggestion. It can still be used in certain cases to improve performance, but those cases are now a lot more nuanced.

We should talk a bit about this next time you're in though, because it looks like the description here is a little outdated. The idea of "dry run" requests were in the initial design of transactional pipelining, but we don't actually use them anymore. Instead, the approach that will be in 2.1 is more closely in-line with https://github.com/cockroachdb/cockroach/issues/16026#issuecomment-375904032.

rmloveland:
Hey @nvanbenschoten I've made some updates based on our conversation.  Mind taking a look again when you get a chance?

Direct links for convenience:

- [Parallel statement execution](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/a97732f5c24d726ea6870becb56af438c2085a2d/dev/parallel-statement-execution.html) - updated with a note stating that it is superseded by txn pipelining for most use cases.  Let me know if the language is too specific about the remaining valid use case, or if it's making too strong a claim.

- New section in [Transaction Layer](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/a97732f5c24d726ea6870becb56af438c2085a2d/dev/architecture/transaction-layer.html#txn-pipelining) describing txn pipelining.  It's reasonably high level (I think?) and hopefully correct.  Happy to update if the information is off base or if there are better ways of describing what is happening.

- The [cluster settings page](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/a97732f5c24d726ea6870becb56af438c2085a2d/dev/cluster-settings.html) has the new flags - this may not need a review since it's generated from the DB code (AFAICT?) but just giving a heads up that it's there.

rmloveland:
Oh, and as mentioned in the commit message I've removed parallel statement execution as a suggestion from the [SQL performance best practices page](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/a97732f5c24d726ea6870becb56af438c2085a2d/dev/performance-best-practices-overview.html) based on your earlier comment that

> The only thing I can think of is that transactional pipelining makes parallel statement execution less important and we should probably remove it as a top-level performance suggestion. It can still be used in certain cases to improve performance, but those cases are now a lot more nuanced.

rmloveland:
Hey @jseldess and @sploiselle , do you have any additional comments I should address?  I'd like to merge this if there are no blockers.

PR #3532, opened 2018-08-09
Update AS OF SYSTEM TIME

### LINKED ISSUE #3455: sql: don't use the grammar to enforce expression const-ness in AOST
PR: https://github.com/cockroachdb/cockroach/pull/27206

From release notes:
> `AS OF SYSTEM TIME` can now use some more complex expressions to compute the desired timestamp. [#27206][#27206] {% comment %}doc{% endcomment %}

lhirata:
AS OF SYSTEM TIME can now use some more complex expressions to compute the desired timestamp

Closes #3455.

### DIFF
--- a/v2.1/as-of-system-time.md
+++ b/v2.1/as-of-system-time.md
@@ -1,7 +1,7 @@
 ---
 title: AS OF SYSTEM TIME
 summary: The AS OF SYSTEM TIME clause executes a statement as of a specified time.
-toc: false
+toc: true
 ---
 
 The `AS OF SYSTEM TIME timestamp` clause causes statements to execute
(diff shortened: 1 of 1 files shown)


danhhz:
one example: we highly recommend that you run backups 10 seconds in the past to avoid interfering with production traffic and this is perfect for that

mjibson:
I don't think you can have expressions like `now()` or any other time function in the AOST clause. I think that we should just close this PR and issue because this change doesn't need to be documented beyond the release note. I can't come up with a good example of why a user would care about this. We made the change for internal code reasons, not any user-facing need.

lhirata:
Ok, thanks for info. Will close the issue / PR!

Issue #3542, opened 2018-08-14
Change "download" to "get cockroachdb"


jseldess:
Follow-up to https://github.com/cockroachlabs/hugo-website/pull/575.

j-low:
@jseldess thanks for the heads up.  I'll have a PR up sometime this afternoon.

PR #3543, opened 2018-08-14
Update download button text

j-low:
__[description]__
* re: #3542 
* update download button text to read "Get CockroachDB" to conform with marketing site


j-low:
@jseldess good catch.  Updates made.  Headers should be identical now.

PR #3548, opened 2018-08-15
Mention debug compact command to force re-encryption

mberhault:


mberhault:
@jseldess: I don't think we document the `cockroach debug` commands anywhere so the only help people get is from the command description (which is rather sparse). However, it is the only way to see detailed status and force re-encryption.

jseldess:
Currently, we only have docs on `cockroach debug zip`: https://www.cockroachlabs.com/docs/dev/debug-zip.html.

jseldess:
Perhaps we should just document `cockroach debug` and all of the user-facing subcommands at this point?

PR #3556, opened 2018-08-16
(WIP) 2.1 Training: SQL Basics

lhirata:
Editing what Nikhil had initially wrote for SQL Basics training

(I hope that I used git correctly... 🤞)


lhirata:
@benesch, can you help me with running a transaction?

I run this in one SQL shell:

```
root@:26257/defaultdb> BEGIN;
                    -> SELECT balance >= 250 FROM bank.accounts WHERE type = 'checking' AND customer_id = 2;
                    -> UPDATE bank.accounts SET balance = balance - 250 WHERE type = 'savings' AND customer_id = 2;
                    -> COMMIT;
+----------+
| ?column? |
+----------+
|   true   |
+----------+
(1 row)

Time: 21.871851ms
```

And the same in the other, but I'm not getting the expected error:

```
root@:26257/defaultdb> BEGIN;
                    -> SELECT balance >= 250 FROM bank.accounts WHERE type = 'checking' AND customer_id = 2;
                    -> UPDATE bank.accounts SET balance = balance - 250 WHERE type = 'savings' AND customer_id = 2;
                    -> commit;
+----------+
| ?column? |
+----------+
|   true   |
+----------+
(1 row)
```

I ran `SELECT *` and noticed that the account I was withdrawing from is now overdrawn:

```
root@:26257/defaultdb> select * from bank.accounts;
+----------+----------+-------------+
|   type   | balance  | customer_id |
+----------+----------+-------------+
| checking |   250.00 |           2 |
| savings  |  -685.85 |           2 |
| savings  | 42000.00 |           4 |
+----------+----------+-------------+
(3 rows)

Time: 1.641927ms
```

Am I doing something incorrectly?

lhirata:
@jseldess - I wasn't sure how much of the original descriptive content should be kept. Can you let me know if I should cut more?

I'm also having some issues with replicating the transaction steps (as mentioned in the previous comment), but the rest is ready for a cursory review. Thanks!

benesch:
Oh I see. This is confusing. Press enter twice between each statement and
see if that works.

lhirata:
thanks @benesch, that helped!

PR #3558, opened 2018-08-17
Document multi-region Kubernetes deployment

### LINKED ISSUE #2933: Global / Multi-DC Kubernetes configuration
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recauVlJw13HDn1Sf
PM: @kannanlakshmi 
Eng Partner: @a-robinson

jseldess:
Fixes #2933.

Follow-up work: https://github.com/cockroachdb/docs/issues/3591.

### DIFF
--- a/_includes/sidebar-data-v2.1.json
+++ b/_includes/sidebar-data-v2.1.json
@@ -232,10 +232,16 @@
                 ]
               },
               {
-                "title": "Kubernetes Deployment",
+                "title": "Kubernetes Single-Cluster Deployment",
                 "urls": [
                   "/${VERSION}/orchestrate-cockroachdb-with-kubernetes.html",
                   "/${VERSION}/orchestrate-cockroachdb-with-kubernetes-insecure.html"
+                ]
+              },
+              {
+                "title": "Kubernetes Multi-Cluster Deployment",
+                "urls": [
+                  "/${VERSION}/orchestrate-cockroachdb-with-kubernetes-multi-region.html"
                 ]
               },
               {
--- a/_includes/v2.1/orchestration/kubernetes-limitations.md
+++ b/_includes/v2.1/orchestration/kubernetes-limitations.md
@@ -1,5 +1,3 @@
-### Limitations
-
 #### Kubernetes version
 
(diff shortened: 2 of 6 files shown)


a-robinson:
I also still owe you an explanation of the firewall problem you hit when you tried this out in the shared GCP project.

jseldess:
___
*[v2.1/orchestrate-cockroachdb-with-kubernetes-multi-region.md, line 335 at r1](https://reviewable.io/reviews/cockroachdb/docs/3558#-LKwHc-71gSAzHimFKg8:-LKx4yx89yLZz2V0cYKq:bhc3f3) ([raw file](https://github.com/cockroachdb/docs/blob/33d96566e613cc45da463ce86f9f81d598bbd2a7/v2.1/orchestrate-cockroachdb-with-kubernetes-multi-region.md#L335)):*

We don't need to do this now, but instead of simulating node failure I think it would be much more interesting to simulate entire datacenter failure by either scaling the number of VM instances in one of the GKE clusters down to 0 (`gcloud container clusters resize <cluster-name> --zone=<zone> --size=0`) or all the pods in one of the clusters down to 0 (`kubectl scale statefulset cockroachdb --replicas=0 --namespace=<cluster-namespace> --context=<cluster-context>`).
</blockquote></details>

Great idea. I created an issue for follow-up work: https://github.com/cockroachdb/docs/issues/3590.

jseldess:
___
*[v2.1/orchestrate-cockroachdb-with-kubernetes.md, line 13 at r1](https://reviewable.io/reviews/cockroachdb/docs/3558#-LKSK2jT0nMEY0Gh2y-A:-LKx5CsmD0KDlJ_MXcUD:b-5v03g7) ([raw file](https://github.com/cockroachdb/docs/blob/33d96566e613cc45da463ce86f9f81d598bbd2a7/v2.1/orchestrate-cockroachdb-with-kubernetes.md#L13)):*

I'm not sure we really want to emphasize the single-region nature of this. This guide will still be the default for 90+% of people that want to run us on Kubernetes, and emphasizing points of difference/choice just makes things more confusing when getting started. This is the way to run Cockroach on Kubernetes. If you have special needs (like multi-region/cluster), then take a look at the other guide.

Maybe I'm just rambling, and the distinction doesn't really matter. That was my first impression reading this, though.
</blockquote></details>

I think it's confusing to have one tutorial that explicitly calls out its deployment context and another that does not. However, I hear you that the existing tutorial should be the default for most people. I'm hoping the 90% still go to the single-region tutorial. Is it ok to wait and see if there's confusion?

PR #3559, opened 2018-08-20
sqlfmt

mjibson:
Add a `make sqlfmt` target that runs sqlfmt on the sql of the 2.1 docs. I've attempted to exclude files and expressions that aren't formatted well or lose some context information. For example, statements where we demonstrate different kinds of string escaping are ignored since the formatter changes all of those strings into a canonical form.


mjibson:
@jseldess Redone with a new `?nofmt` modifier that doesn't appear to affect the code highlighting output, and is ignorable by sqlfmt. This should be ready for a real review now.

jseldess:
Hoping to review sometime today or tomorrow, @mjibson. Sorry for the delay.

jseldess:
Ugh. I'm sorry this never happened, @mjibson.

@yingzhuchin, when you have a chance, would you mind putting this on your backlog. This is a way to auto-format all SQL code examples in our docs, which would be amazing.

jseldess:
Oh, we actually have an issue already, @yingzhuchin: https://github.com/cockroachdb/docs/issues/7059

PR #3489, opened 2018-08-01
Backup runbook

Amruta-Ranade:
Closes #2447 

**Note to reviewers** 
Opening a WIP unedited PR for initial (alignment + structure) feedback:
- @dt and @robert-s-lee : Need your feedback about whether the doc meets our requirements of a one-stop, copy-pastable backup/restore runbook.
- @jseldess: Need feedback about the structural changes (consolidating the backup/restore docs, organizing information)

Summary of changes:
- Consolidated the backup and restore docs in the Maintain guides section
- Added content for performing core and enterprise (manual and automated) backups and restores
- Updated the sidebar nav

To-Do:
- The formatting is breaking for some reason - looking into that (TOC, sample script include)
- Might have some broken links that need to be fixed.
- Flesh out parameters descriptions (if required, not sure if I should repeat them in this doc, since we already have extensive docs about them)

PR #3567, opened 2018-08-22
Fix "Back and Restore" typo

solongordon:
One of the training pages had "Back and Restore" as a title rather than
"Backup and Restore."


jseldess:
Thanks, @solongordon. I think this is fixed in this other PR, which will be merged soon: https://github.com/cockroachdb/docs/pull/3489. Can you confirm and, if so, close this PR?

solongordon:
Ah, yup! Might want to expand the other PR to fix it in pre-2.1 versions as well.

PR #3571, opened 2018-08-22
Fix --warehouses argument in TPCC instructions

solongordon:
The --warehouses argument for loading TPCC fixtures was off by a factor
of 10.


lhirata:
@solongordon @jseldess - the first example is running TPC-C 1K: `1,000 warehouses (for a total dataset size of 200GB) on 3 nodes` so 1000 is actually correct. I'm going to revert the change, but will add a little more description to point out the difference.

lhirata:
@solongordon @jseldess - sorry! misread the 0s. Closing the revert, thanks for the fix :)

solongordon:
Np, thanks for verifying!

Issue #3388, opened 2018-07-17
backport-2.0: cli: Deprecate --wait=live in `node decommission`


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/27158

From release notes:
> CockroachDB now computes the correct number of replicas on down nodes. Therefore, when [decommissioning nodes](../v2.0/remove-nodes.html) via the [`cockroach node decommission`](../v2.0/view-node-details.html) command, the `--wait=all` option no longer hangs indefinitely when there are down nodes. As a result, the `--wait=live` option is no longer necessary and has been deprecated. [#27158][#27158] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closed by https://github.com/cockroachdb/docs/pull/3573

PR #3573, opened 2018-08-22
Decommissioning nodes doc updates

### LINKED ISSUE #3472: cli: Deprecate --wait=live in `node decommission`
PR: https://github.com/cockroachdb/cockroach/pull/27027

From release notes:
> CockroachDB now computes the correct number of replicas on down nodes. Therefore, when [decommissioning nodes](../v2.1/remove-nodes.html) via the [`cockroach node decommission`](../v2.0/view-node-details.html) command, the `--wait=all` option no longer hangs indefinitely when there are down nodes. As a result, the `--wait=live` option is no longer necessary and has been deprecated. The `--wait=all` option is now the default. [#27027][#27027] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closes #3472 and #3388 

Summary of changes:
- Remove `--wait=live` instances
- Add note about decommissioned nodes never disappearing from the **Decommissioned Nodes** list on the UI.

JFYI: I also spent considerable time testing the decommissioning workflow and reworking the document to prevent users from decommissioning multiple nodes simultaneously, but today Tobi fixed the problem in the product. So those doc changes are no longer valid - hence reverted them.


Amruta-Ranade:
@lhirata Fixed the nits :)

PR #3585, opened 2018-08-26
Update install-client-drivers.md with appropriate Ruby on Rails adapter specific instructions.

skunkworker:
On the CockroachDB blog https://www.cockroachlabs.com/blog/cockroachdb-hearts-activerecord-ruby-on-rails/ it is recommended that Ruby on Rails users use the `activerecord-cockroachdb-adapter` if using Ruby on Rails. This is not clearly communicated on the `Install Client Drivers` page where it should be.


jseldess:
Suitably covered in a separate PR: https://github.com/cockroachdb/docs/pull/4923.

Issue #3655, opened 2018-08-28
ui/cdc: Highwater timestamp on Jobs page


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/28535

From release notes:
> The [**Jobs** page](../v2.1/admin-ui-jobs-page.html) now supports indefinitely-running job types that have a "highwater timestamp", instead of the "fraction completed" used by jobs with a finite task. [#28535][#28535] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Discussion with @danhhz and @lhirata Need to document `highwater timestamp` for CDC and then update the Jobs page and link to the CDC doc.

Issue #3606, opened 2018-08-28
cdc: fix a couple small issues


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/28733

From release notes:
> The `CHANGEFEED` `timestamp` option has been split into `updated` and `resolved`. [#28733][#28733] {% comment %}doc{% endcomment %}

Issue #3605, opened 2018-08-28
changefeedccl: add production monitoring metrics


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/28162

From release notes:
> `CHANGEFEED`s now export metrics for production monitoring. [#28162][#28162] {% comment %}doc{% endcomment %}

Issue #3603, opened 2018-08-28
jobspb,changefeedccl: finalize ChangefeedDetails job proto for 2.1


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/27996

From release notes:
> `CHANGEFEED`s can now only target lists of physical tables. [#27996][#27996] {% comment %}doc{% endcomment %}

PR #3669, opened 2018-08-28
CDC updates

### LINKED ISSUE #3601: changefeedccl: various tests
PR: https://github.com/cockroachdb/cockroach/pull/27962

From release notes:
> `CREATE CHANGEFEED` is now restricted to superusers. [#27962][#27962] {% comment %}doc{% endcomment %}
>
> `CHANGEFEED`s now produce an error when column families are added (instead of returning incorrect results) and when targeting `system` tables (instead of operating with undefined behavior). [#27962][#27962] {% comment %}doc{% endcomment %}
>
> `CREATE CHANGEFEED` now requires an enterprise license when used with Kafka. [#27962][#27962] {% comment %}doc{% endcomment %}

lhirata:
Closes #3601, #3602, #3603, #3605, #3606, #3316, #3752, #3655

### DIFF
--- a/v2.1/change-data-capture.md
+++ b/v2.1/change-data-capture.md
@@ -7,29 +7,61 @@ toc: true
 <span class="version-tag">New in v2.1:</span> Change data capture (CDC) provides efficient, distributed, row-level change feeds into Apache Kafka for downstream processing such as reporting, caching, or full-text indexing.
 
 {{site.data.alerts.callout_danger}}
-**This feature is under active development** and only works for a targeted use case. Please [file a Github issue](file-an-issue.html) if you have feedback on the interface.
-
-In v2.1, CDC will be an enterprise feature and will have a core version.
+**This feature is under active development** and only works for a targeted use case. Please [file a Github issue](file-an-issue.html) if you have feedback on the roadmap.
 {{site.data.alerts.end}}
 
+{{site.data.alerts.callout_info}}
+CDC is an enterprise feature. There will be a core version in a future release.
+{{site.data.alerts.end}}
 
 ## What is change data capture?
 
 While CockroachDB is an excellent system of record, it also needs to coexist with other systems. For example, you might want to keep your data mirrored in full-text indexes, analytics engines, or big data pipelines.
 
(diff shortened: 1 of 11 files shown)


rolandcrosby:
We need to document the time format for WITH CURSOR - it’s an internal cluster timestamp and I don’t think we explicitly talk about the format anywhere. Dan or I can help if you need more details.

v2.1/change-data-capture.md
```diff
@@ -7,29 +7,61 @@ toc: true
...
 
 {{site.data.alerts.callout_danger}}
-**This feature is under active development** and only works for a targeted use case. Please [file a Github issue](file-an-issue.html) if you have feedback on the interface.
-
-In v2.1, CDC will be an enterprise feature and will have a core version.
+**This feature is under active development** and only works for a targeted use case. Please [file a Github issue](file-an-issue.html) if you have feedback on the roadmap.
 {{site.data.alerts.end}}
 
+{{site.data.alerts.callout_info}}
+CDC is an enterprise feature. There will be a core version in v2.2.
```
rolandcrosby: Do we generally commit to when future versions are coming, or do we want to say "a core version is planned for a future release"?

lhirata:
@danhhz @jseldess - added info about high-water timestamps and made edits based on the last round of feedback. can you review plz?

@Amruta-Ranade - you can take a quick look too (if you want). i think i added enough info to cover #3655.

jseldess:
___
*[v2.1/change-data-capture.md, line 14 at r6](https://reviewable.io/reviews/cockroachdb/docs/3669#-LL1XoYYIT89CORaQN0m-r6-14:-LMrl0Lw3ewvbv9OZ8Tp:b-9dfmm5) ([raw file](https://github.com/cockroachdb/docs/blob/761607ef83b9a3022a8f78bbf5e777bda1c75554/v2.1/change-data-capture.md#L14)):*

There's been enough interest in a core version of changefeeds that I think it's worth keeping a mention of our intention to have one. This is the first time that the core alternative to an enterprise feature is not available when the enterprise feature was launched, so I think it's breaking new ground a bit
</blockquote></details>

SGTM. In that case, Lauren, please just add a link to the enterprise licensing page.

PR #3675, opened 2018-08-30
Build docs for CockroachDB and Managed CockroachDB

jseldess:
This PR introduces the mechanics for building docs for CockroachDB and for Managed CockroachDB.

Summary of this approach:
- Use separate configs for each "product": `_config.yml` for the standard product (SP), and `_config_managed.yml` for the managed service (MS).
  - In `_config_managed.yml`:
    - Build the MS docs to the `/docs/managed` subdirectory.
    - Use an exclude list to exclude pages. This list is populated during the build (see bullet below on Makefile).
    - Add a `managed: true` variable to all pages included in the MS docs. On pages that need to differ for MS vs. SP, we can use conditional logic to include the right content. Example: v2.1/index.md.
    - Build the MS docs with a distinct sidenav. It's just a placeholder right now.
- Customize the Makefile so that:
    - `make managed` executes `scripts/exclude_from_managed_docs.py`, which scrapes all 2.1 pages and adds any without `build_for: managed` and `build_for: both` in the front-matter to the `exclude` list in `_config_managed.yml`, and then builds the MS docs.
    - `make standard` executes `scripts/exclude_from_standard_docs.py`, which scrapes all 2.1 pages and adds any with `build_for: managed` in the front-matter to the `exclude` list in `_config.yml`, and then builds the SP docs.
- Add logic to `_include/head.html` so MS pages:
   - Are ignored by search engines.
   - Point to the SP version as canonical, if a page exists in both contexts. This avoids negative SEO implications of identical content. Eventually, we'll need to not do this for pages only in the MS docs.  
- Update `ci/pull-request` to build both SP and MS docs for PRs.


jseldess:
@j-low, let's look at this together in our meeting tomorrow. @benesch, would be great to get your eyes on this, too.

j-low:
@jseldess per our conversation, we currently have a means to exclude standard pages from the cloud docs but no mechanism for excluding 'cloud-only' pages from the standard docs.  My suggestion here would be to place 'cloud-only' docs into a special subdirectory and add that subdirectory to the `excludes` of the `config.yml` for standard docs.  This approach will avoid the necessity of adding an extra bit of front matter state to _all_ pages (i.e. less maintenance overhead).

jseldess:
Thanks, @j-low. As discussed in our meeting, we're now using `cloud_only: true` and `core_only: true` to populate the exclude lists for the two builds.

benesch:
LGTM

benesch:
It's due to this commit which checks the sidebar links. That code will need to be adapted for the new managed sidebar. https://github.com/cockroachdb/docs/commit/d39c299b6f14768732167eb488dc091c3463feef#diff-cd23c6a5d63a98a34c5720b313be55e9

Let me know when this branch is otherwise ready to go and I can push a commit that fixes the sidebar link checker.

jseldess:
Thanks, @benesch. This PR is otherwise ready, so please go ahead and push that change, when you have time.

PR #3702, opened 2018-09-05
System range replication

### LINKED ISSUE #3552: Add known limitation about changing the replication factor
See https://github.com/cockroachdb/cockroach/issues/14990 (link was edited)

If a user decides to run with, say, 5x replication and changes the `.default` zone config accordingly, they're still not done. They also need to change the `.liveness` zone config as well as the zone config for the system database, which will have "copied over" the 3x at cluster creation time and won't automatically pick up the 5x. (And losing either of these is fatal).

We should also generally recommend that operators bump the replication factor for those parts of the keyspace to 5x, provided they are running at least a five node cluster.

### LINKED ISSUE #3699: Update `data replication` guide for latest stable version.
The guide expects all ranges across the entire cluster to end up with a replication factor of 5. However, 2.0.5 has three predefined zone configs: `.liveness`, `.meta`, and `system.jobs`.

Changing the `.default` config will not modify the ranges using those custom configs and some ranges will still only be replicated three times. This is confusing as the guide currently shows all ranges replicated 5 times.

jseldess:
Update tutorials, training, and production checklist to emphasize the importance of manually increasing the replication factor for system ranges when:
- Running a cluster of 5 or more nodes.
- Increasing the default replication factor for user data.

Also add a known limitation.

Fixes #3552.
Fixes #3699.


jseldess:
@tschottdorf, should I also add a new section to the [Known Limitations](https://www.cockroachlabs.com/docs/stable/known-limitations.html) page?

I definitely want to write to a deeper page on fault tolerance. We need to help people understand table data ranges vs. system ranges and data unavailability vs. system unavailability. Just scratched the surface in these [trainings](https://www.cockroachlabs.com/docs/stable/training/cluster-unavailability-troubleshooting.html).

PR #3704, opened 2018-09-05
Update v2.0 example code to use secure clusters

### LINKED ISSUE #3685: Include secure examples in 2.0 docs.
2.0 docs are still the default making those examples hard to find for people (see https://github.com/cockroachdb/cockroach/issues/29463).
Since the secure examples are not specific to a CockroachDB version, we should include them in the 2.0 docs.

rmloveland:
Fixes #3685.

Summary of changes:

- Update 2.0 docs to use the same pattern as 2.1

- Each updated example page now has 'secure' / 'insecure' tabs

- Updated pages include:

  - Go (pq & GORM)
  - Java (JDBC & Hibernate)
  - JavaScript (pg & sequelize)
  - PHP
  - Python (psycopg2 & SQLAlchemy)
  - Ruby (pg & ActiveRecord)


rmloveland:
> Something is wrong here. In the insecure version, step 2 is completely missing. This is the case in the live 2.1 version as well, we just didn't catch it earlier.

Thanks @jseldess !  Fixed in latest push.

jseldess:
LGTM!

Issue #2992, opened 2018-04-20
Support ALTER TABLE ... SET TYPE (change column type)


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recogtNUGtTUPczhc
PM: @nstewart 
Eng Partners: @mjibson, @dt, @bobvawter

jseldess:
Initial work: https://github.com/cockroachdb/cockroach/pull/25777

jseldess:
@Amruta-Ranade, you can write docs on this. There's a workaround to document in some cases (see the Airtable comments for details).

lhirata:
I'm updating the SQL diagram for ALTER COLUMN and documenting DROP STORED (computed columns, see #3220). Additional updates will need to be made for the other added parameters.

Amruta-Ranade:
Discussion with @mjibson: 
- On-disk representation/format shouldn't change (eg: can't go from STRING to an INT, even if the string is just a number)
- Any validation has to be successful (you can go from STRING[10] to STRING[20], but not to STRING [5])

Eg: TIMESTAMP <> TIMESTAPMTZ because it meets both criteria.

Tests for examples: https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/schemachange/alter_column_type_test.go
Look for things marked "ColumnConversionTrivial"

Amruta-Ranade:
Syntax:

`ALTER TABLE <table_name> ALTER <column_name> TYPE <new_type>;`

Amruta-Ranade:
@andreimatei gave an excellent example: For our TPCC schema, we have a column with decimal (10,2). He changed it to (15,2) when he had more records. But he couldn't have changed it to less than 10 (say, (8,2)) because that invalidates the earlier representation.

Amruta-Ranade:
Doc impact:
- Need a new doc (ALTER TYPE)
- Update the `ALTER TABLE` doc with reference to the new doc.

Amruta-Ranade:
@mjibson Will running `SHOW JOBS` show the change in data type?

mjibson:
No, this doesn't create a job.

Amruta-Ranade:
Closed via https://github.com/cockroachdb/docs/pull/3707

PR #3707, opened 2018-09-06
Document `Alter type` subcommand

Amruta-Ranade:
Addresses #2992


Amruta-Ranade:

PR #3712, opened 2018-09-06
Fix SQL Basics training

### LINKED ISSUE #3711: sql basics: confusion in transactions example
In https://www.cockroachlabs.com/docs/dev/training/sql-basics.html, in the Transactions section, we have:

`
SELECT balance >= 250 FROM bank.accounts WHERE type = 'checking' AND customer_id = 2;
UPDATE bank.accounts SET balance = balance - 250 WHERE type = 'savings' AND customer_id = 2;
`

This seems to not work as expected since we are checking that the "checking" account has over 250, but we're withdrawing from the "savings" account, instead of withdrawing from the account that is being checked.

lhirata:
Closes #3711.

### DIFF
--- a/v2.0/training/sql-basics.md
+++ b/v2.0/training/sql-basics.md
@@ -329,7 +329,7 @@ Here's the above example in a transaction. Again, don't run this example yet.
 
 -- If false, issue a ROLLBACK statement. Otherwise, continue.
 
-OPEN> UPDATE bank.accounts SET balance = balance - 250 WHERE type = 'savings' AND customer_id = 2;
+OPEN> UPDATE bank.accounts SET balance = balance - 250 WHERE type = 'checking' AND customer_id = 2;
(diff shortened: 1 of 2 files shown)


changangela:
LGTM, thanks! random question: why is there an `OPEN>` prompt?

lhirata:
Thanks for the catch @changangela! `OPEN>` is included to denote the open transaction. If you press enter twice after you start a transaction, you'll see the prompt. For example:

~~~
root@:26257/> begin;
           ->
BEGIN

Time: 741.873µs

root@:26257/  OPEN>
~~~

lhirata:
@changangela ah! I see that the `OPEN>` prompt in the training gets copied when you click the copy button. I'll fix that in another PR!

changangela:
@lhirata haha thanks for the clarification

Issue #2364, opened 2018-01-16
distsql session variable is poorly documented


solongordon:
I noticed we don't have good documentation for the `distsql` session variable. It's not mentioned at all in [`set-vars.md`](https://github.com/cockroachdb/docs/blob/master/v2.0/set-vars.md) and is listed without a description in [`show-vars.html`](https://github.com/cockroachdb/docs/blob/master/v2.0/show-vars.md).

lhirata:
Related to #3717

PR #3718, opened 2018-09-10
Add distsql to SET/SHOW vars

### LINKED ISSUE #3717: Edit SET distsql settings
Set distsql=2.0 - do we need a flag that reverts our behavior to 2.0?

Conclusion: yes - we can’t let customers have un-revertable regressions.

What are the settings now?
- off: run everything via local mode
- auto: default: distribute “should distribute” queries, but everything runs through distsql
- on: distribute all queries that are possible to distribute
- always: distribute every query even those that can’t be distributed (returns error if not)
Add a new one:
-2.0: distribute “should distribute” queries, everything else runs through local sql (same as “auto” in 2.0)

lhirata:
Add `distsql` to `SET vars` and `SHOW vars`. Will need to add the `2.0` setting later (#3720).

Note: Conceptual docs about the DistSQL engine / distributed query heuristic will be written later.

Closes #3717 and #2364.


v2.1/set-vars.md
```diff
@@ -40,6 +40,7 @@ The variable name is case insensitive. The value can be a list of one or more it
...
  `default_transaction_read_only` | The default transaction access mode for the current session. If set to `on`, only read operations are allowed in transactions in the current session; if set to `off`, both read and write operations are allowed. See [`SET TRANSACTION`](set-transaction.html) for more details. | `off` | Yes
+ `distsql` | <span class="version-tag">New in v2.1:</span> The query distribution mode for the current session.<br><br>If `auto`, all queries that should be distributed (determined by a heuristic) are distributed, and all other queries are run through the gateway node. **`auto` is recommended.** <br><br>If `on`, any queries that can be distributed are distributed, and all other queries are run through the gateway node.<br><br>If `off`, all queries are run through the local SQL engine. <br><br>If `always`, all queries are distributed, even those that can't be distributed (returns an error). This setting is used for internal testing and is not recommended. | `auto` | Yes
```
jordanlewis: I would flesh out the language `should` by a bit. Maybe say something like "if `auto`, all queries that CockroachDB believes will take less time to execute if distributed to multiple nodes will be executed". Probably there's a better way to word that, but the important point is that the heuristic is based on whether or not we think it'll be faster to run the query if we distribute it.

PR #3731, opened 2018-09-11
Fix typo

awreece:
> thought of as AN always close to

### DIFF
--- a/v2.1/architecture/transaction-layer.md
+++ b/v2.1/architecture/transaction-layer.md
@@ -65,7 +65,7 @@ In relationship to other layers in CockroachDB, the transaction layer:
 
 ### Time and hybrid logical clocks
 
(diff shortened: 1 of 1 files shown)


jseldess:
Thanks, @awreece.

@sploiselle, for the correct meaning, shouldn't we just remove `an/and`?

sploiselle:
@jseldess I meant it like [(thought-of-as) and (always-close-to) local wall time]. Could be rephrased to something like `...which are composed of a physical component (always close to local wall time)...`

@awreece Thanks for letting us know you stumbled on this. Just an FYI, if you wanted to implement your edit, you'd need to hyphenated it, i.e. `thought of as an-always-close-to-local wall time`, though that's pretty strange wording.

awreece:
Yeah, I think there is actually a bit of work we can do to simplify the grammar here :)

I just didn't want to rock the boat on my first PR! Would you be interested in me doing a bit more work on this section, or is perhaps not a great time for that?

jseldess:
Thanks for the clarification, @sploiselle. And for your offer to edit the section, @awreece. Please do propose simplifications, if you have them. If we stay close to the current formulation, I'd suggest something like: `...physical component (always close to local wall time)...`

jseldess:
I went ahead and opened a separate PR with this change and applied it to other versions of the page as well. Thanks for triggering this improvement, @awreece.

Issue #3647, opened 2018-08-28
ui: show sys+user combined cpu, normalized by # cores


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/28596

From release notes:
> The [**Hardware** dashboard](../v2.1/admin-ui-hardware-dashboard.html) now shows system and user CPU summed instead of separately, and normalized by number of CPUs. [#28596][#28596] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closed via https://github.com/cockroachdb/docs/pull/3744

Issue #3646, opened 2018-08-28
ui: add disk ops per second graph to hardware dashboard


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/28594

From release notes:
> Added disk read and write time charts to the [**Hardware** dashboard](../v2.1/admin-ui-hardware-dashboard.html). [#27977][#27977] [#28594][#28594] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closed via https://github.com/cockroachdb/docs/pull/3744

PR #3744, opened 2018-09-13
Doc Updates for Hardware Dashboard (content + screenshots)

### LINKED ISSUE #3694: ui: Hardware dashboard doc updates
- Document the new graphs added to the hardware dashboard
- Provide doc links in tooltips (Reference issue: https://github.com/cockroachdb/cockroach/issues/29525)

Amruta-Ranade:
Closes #3694, #3647, #3646


piyush-singh:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMh_82o7PDEOBKT7Umw:-LMh_82pD7b_gzJbVlJk:bb7ddje) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L17)):*
> ```Markdown
> - In the node view, the graph shows the amount of CPU resources used by the CockroachDB process as well as the CockroachDB system-level operations for the selected node.
> 
> - In the cluster view, the graph shows the amount of CPU resources used by the CockroachDB process as well as the CockroachDB system-level operations across all nodes.
> ```

We should include a note here about how these percentages are normalized. We are showing CPU percentages in a different way than we do elsewhere in the UI and in the past. Instead of showing CPU seconds per second, this is now dividing that value by the # of cores to give a true utilization percentage out of 100%. @vilterp is it worth putting a note about containerized environments here?

piyush-singh:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMhaMod7ai9JkThH7TZ:-LMhaMod7ai9JkThH7T_:b-6diukb) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L47)):*
> ```Markdown
> CockroachDB Admin UI Disk Read Time graph
> 
> - In the node view, the graph shows the 10-second average of the disk read time per second for all processes, including CockroachDB, for the selected node.
> ```

Out of curiosity, is this always a 10 second average? Don't we average over a longer time horizon when looking at larger time frames (1 month instead of 1 hour)? Not sure if we have a consistent way to reference that in docs. 

I will also point out that like CPU/s per second, Disk read/write time per second can get a little confusing. It might be worth including a sort explanation of what exactly that means or a short example. "For example, if during a 10 second interval, the disk is read from for .5 seconds at the beginning and 1.5 seconds at the end, the total read time will show 2s for that 10 second interval" or something to that effect might help clarify. I'm not sure of the best way to represent that here though, since it might be redundant to include this multiple times for each metric. I don't think it is critical to include this, but considering how much general confusion there was over CPU s/s, it might be worthwhile.

piyush-singh:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 91 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMhdM317_NuwcrYh50q:-LMhdM317_NuwcrYh50r:b-6eobds) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L91)):*
> ```Markdown
> CockroachDB Admin UI Disk Capacity graph
> 
> - In the node view, the graph shows the available storage capacity for the selected node.
> ```

Possibly worth mentioning that "Available" capacity factors in other non-cockroach data on the disk? So this gives an accurate picture of how much space is available to cockroach, not just non-cockroach disk size.

couchand:
> general confusion there was over CPU s/s

I would argue that the issue was not confusion, and particularly not regarding s/s.  The *inconsistency* is in how tooling reports CPU *percentages*, which is why it would be preferable to use an absolute unit that doesn't suffer from that problem.

Amruta-Ranade:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMh_82o7PDEOBKT7Umw:-LMn1WtM9g5i4Po4N_zX:b-gygb41) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L17)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

We should include a note here about how these percentages are normalized. We are showing CPU percentages in a different way than we do elsewhere in the UI and in the past. Instead of showing CPU seconds per second, this is now dividing that value by the # of cores to give a true utilization percentage out of 100%. @vilterp is it worth putting a note about containerized environments here?
</blockquote></details>

I am planning to tackle this holistically in the doc project #3753

v2.1/admin-ui-hardware-dashboard.md
```diff
@@ -8,29 +8,13 @@ The **Hardware** dashboard lets you monitor CPU usage, disk throughput, network
...
 The **Hardware** dashboard displays the following time series graphs:
 
-## User CPU Percent
+## CPU Percent
 
-<img src="{{ 'images/v2.1/admin_ui_user_cpu.png' | relative_url }}" alt="CockroachDB Admin UI User CPU Percent graph" style="border:1px solid #eee;max-width:100%" />
+<img src="{{ 'images/v2.1/admin_ui_cpu_percent.png' | relative_url }}" alt="CockroachDB Admin UI CPU Percent graph" style="border:1px solid #eee;max-width:100%" />
 
-- In the node view, the graph shows the amount of CPU resources used by the CockroachDB process for the selected node.
+- In the node view, the graph shows the amount of CPU resources used by the CockroachDB process as well as the CockroachDB system-level operations for the selected node.
```
vilterp: "for the selected node" seems a bit weird, since we are showing all nodes; there isn't really a way to "select" one. How about "for each node"? (This comment applies to your docs for every graph)

vilterp: Also "amount of CPU resources" is a bit weird, since we're only showing one resource: time. How about "…shows the CPU utilization of the CockroachDB process, where 100% means that CockroachDB is fully using all CPUs"

Especially important to explicitly state what 100% means, since people might think it means fully using only one core.

vilterp:
Also is this blurb gone?
image

I see it in the TeamCity link above but not in the source. I don't think it should be there; only really makes sense when the numbers are aggregated, not per-node as they are here.

v2.1/admin-ui-hardware-dashboard.md
vilterp: Oh, I guess you mean selecting "cluster" vs selecting one node in the selector on the top left. I guess that makes sense, although it seems a little heavyweight to repeat both forms for every single graph, as opposed to a little blurb at the top about how this works.

(Then again, "how this works" will change when we land a toggle)

v2.1/admin-ui-hardware-dashboard.md
```diff
@@ -56,6 +40,38 @@ For multi-core systems, the system CPU percent can be greater than 100%. Full ut
...
+
+<img src="{{ 'images/v2.1/admin_ui_disk_read_time.png' | relative_url }}" alt="CockroachDB Admin UI Disk Read Time graph" style="border:1px solid #eee;max-width:100%" />
+
+- In the node view, the graph shows the 10-second average of the disk read time per second for all processes, including CockroachDB, for the selected node.
+
+- In the cluster view, the graph shows the 10-second average of the disk read time per second for all processes, including CockroachDB, across all nodes.
+
+## Disk Write Time
+
+<img src="{{ 'images/v2.1/admin_ui_disk_write_time.png' | relative_url }}" alt="CockroachDB Admin UI Disk Write Time graph" style="border:1px solid #eee;max-width:100%" />
```
vilterp: removing: https://github.com/cockroachdb/cockroach/pull/30488

Amruta-Ranade:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMJd0UExt7_ek-e-AJ3-r1-15:-LNBZ6nBBC6UBXBcbkr7:b-g25wgk) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L15)):*

Oh, I guess you mean selecting "cluster" vs selecting one node in the selector on the top left. I guess that makes sense, although it seems a little heavyweight to repeat both forms for every single graph, as opposed to a little blurb at the top about how this works.

(Then again, "how this works" will change when we land a toggle)
</blockquote></details>

- "CPU utilization of the CockroachDB process, where 100% means that CockroachDB is fully using all CPUs" > should include "CockroachDB process as well as the CockroachDB system-level operations", right?
- "you mean selecting "cluster" vs selecting one node in the selector on the top left. I guess that makes sense, although it seems a little heavyweight to repeat both forms for every single graph, as opposed to a little blurb at the top about how this works.

(Then again, "how this works" will change when we land a toggle)"

Yup. Leaving it the way it is for now. I have it on my to-watch list to make relevant doc changes when the toggle lands.

Amruta-Ranade:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 53 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMJd0UExt7_ek-e-AJ3-r1-53:-LNBZspC6eiAJ69df2d3:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L53)):*

removing: https://github.com/cockroachdb/cockroach/pull/30488
</blockquote></details>

Done.

Amruta-Ranade:
As discussed, the blurb stays :)

Amruta-Ranade:
___
*[v2.1/admin-ui-hardware-dashboard.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/3744#-LMJd0UExt7_ek-e-AJ3-r1-15:-LNBiONMAHCtgfCvl08U:b-hkl8sy) ([raw file](https://github.com/cockroachdb/docs/blob/1c65022d604b52e01085a12c913c674f0cdd72b9/v2.1/admin-ui-hardware-dashboard.md#L15)):*

- "CPU utilization of the CockroachDB process, where 100% means that CockroachDB is fully using all CPUs" > should include "CockroachDB process as well as the CockroachDB system-level operations", right?
- "you mean selecting "cluster" vs selecting one node in the selector on the top left. I guess that makes sense, although it seems a little heavyweight to repeat both forms for every single graph, as opposed to a little blurb at the top about how this works.

(Then again, "how this works" will change when we land a toggle)"

Yup. Leaving it the way it is for now. I have it on my to-watch list to make relevant doc changes when the toggle lands.
</blockquote></details>

Follow-up with Pete: The distinction between CockroachDB process and system-level operations is not required in this graph. (We do need it in the CPU graph elsewhere).

v2.1/admin-ui-hardware-dashboard.md
```diff
@@ -12,9 +12,13 @@ The **Hardware** dashboard displays the following time series graphs:
...
 
-- In the node view, the graph shows the CPU utilization (across all cores) by the CockroachDB process for the selected node.
+- In the node view, the graph shows the percentage of CPU in use by the CockroachDB process for the selected node.
 
-- In the cluster view, the graph shows the CPU utilization (across all cores) by the CockroachDB process across all nodes.
+- In the cluster view, the graph shows the percentage of CPU in use by the CockroachDB process across all nodes.
+
+{{site.data.alerts.callout_info}}
+For multi-core systems, the percentage of CPU usage is calculated by normalizing the CPU usage across all cores, whereby 100% utilization indicates that all cores are fully utilized. 
+{{site.data.alerts.end}}
```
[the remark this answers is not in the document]
Amruta-Ranade: I tried to keep it consistent with the note for the other CPU percent graph, as well as AWS and GCP's terminology.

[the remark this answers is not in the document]
vilterp: Okey doke

v2.1/admin-ui-hardware-dashboard.md
```diff
@@ -4,32 +4,20 @@ summary: The Hardware dashboard lets you monitor CPU usage, disk throughput, net
...
+## CPU Percent
 
-<img src="{{ 'images/v2.1/admin_ui_user_cpu.png' | relative_url }}" alt="CockroachDB Admin UI User CPU Percent graph" style="border:1px solid #eee;max-width:100%" />
+<img src="{{ 'images/v2.1/admin_ui_cpu_percent.png' | relative_url }}" alt="CockroachDB Admin UI CPU Percent graph" style="border:1px solid #eee;max-width:100%" />
 
-- In the node view, the graph shows the amount of CPU resources used by the CockroachDB process for the selected node.
+- In the node view, the graph shows the percentage of CPU in use by the CockroachDB process for the selected node.
 
-- In the cluster view, the graph shows the amount of CPU resources used by the CockroachDB process across all nodes.
+- In the cluster view, the graph shows the percentage of CPU in use by the CockroachDB process across all nodes.
```
vilterp: nit: I would prefer "for each node" to "across all nodes"

Amruta-Ranade: Just on this one or for each graph (wherever applicable)?

PR #3746, opened 2018-09-13
New networking flags

### LINKED ISSUE #3634: New --http-addr, --advertise-addr, and --listen-addr flags
PR: https://github.com/cockroachdb/cockroach/pull/28502

From release notes:
> The new `--http-addr` flag recognizes both a hostname/address and port and replaces the `--http-host` flag, which is now deprecated. The port portion of `--http-addr` can be either a service name or numeric value; when specified as `0`, a port number is automatically allocated. [#28373][#28373] [#28502][#28502] {% comment %}doc{% endcomment %}
>
> The new `--advertise-addr` flag recognizes both a hostname/address and port and replaces the `--advertise-host` and `--advertise-port` flags, which are now deprecated.  The port portion of `--advertise-addr` can be either a service name or numeric value; when specified as `0`, a port number is automatically allocated. [#27800][#27800] [#28373][#28373] [#28502][#28502] {% comment %}doc{% endcomment %}
>
> The new `--listen-addr` flag recognizes both a hostname/address and port and replaces the `--host` and `--port` flags, which are now deprecated for `cockroach start` but remain valid for other *client* commands. The port portion of `--listen-addr` can be either a service name or numeric value; when specified as `0`, a port number is automatically allocated. [#27800][#27800] [#28373][#28373] [#28502][#28502] {% comment %}doc{% endcomment %}

jseldess:
- Update 2.1 docs with new `--advertiser-add`, `--listen-addr`, and `--http-addr` flags, including `cockroach start` docs, Production Checklist, and all tutorials, examples, and trainings. 
- Add a Tutorials section to the 2.1 sidenav.

Fixes #3634.


jseldess:
___
*[_includes/v2.1/prod-deployment/insecure-initialize-cluster.md, line 9 at r2](https://reviewable.io/reviews/cockroachdb/docs/3746#-LMNFy-e4vaQFqPqLeCJ:-LMNY_JjB92udMc6YPEq:b-cy66tm) ([raw file](https://github.com/cockroachdb/docs/blob/8248341891741033aadf56323e620d67b40345cf/_includes/v2.1/prod-deployment/insecure-initialize-cluster.md#L9)):*

Thanks, @knz. A few responses:

- Given that we're asking users to re-learn networking flags on `cockroach start`, one aspect of which is combining host and port in single flags, I think it's logical to ask user to extend that to networking flags on client commands. Consistency is a key aspect of usability. So I disagree that my process is very different than the process users will go through. 
- It's not best practice to explain _all_ syntax variations in instructions. Rather, we should focus on one form that we feel is common/intuitive. For all variations, they can refer to the reference docs, which would be `cockroach start`.
- I have, in fact, understood that host/address is optional. In some cases, I just think it's more helpful to users to explicitly show the common form, even if it means not leveraging defaults.
- All the above said, since you feel strongly, however, I can drop port from these examples. I'll want to keep them in local cluster tutorials though.
 
</blockquote></details>

Need one clarification on `when the host/address part is left unspecified, it defaults to localhost`:

My understanding from earlier work on networking, with you and @bdarnell, was different, that a node listens on all IP address and advertises the node's canonical hostname by default (if neither flag is set).

Is this no longer true?

PR #3764, opened 2018-09-18
Replace mentions of root with admin role

### LINKED ISSUE #3518: Replace mentions of root with superuser.
Since RBAC, SQL permission checks that required the `root` user changed to require a superuser. This [is defined](https://www.cockroachlabs.com/docs/stable/roles.html#terminology) as "a member of the `admin` role" (which includes the `root` user).

For example, `BACKUP` requires a superuser but [currently says](https://www.cockroachlabs.com/docs/stable/backup.html#required-privileges) `root`. This can be misleading for anyone using RBAC.

jseldess:
Fixes #3518.

### DIFF
--- a/v2.1/backup.md
+++ b/v2.1/backup.md
@@ -111,7 +111,7 @@ The `BACKUP` statement cannot be used within a [transaction](transactions.html).
 
 ## Required privileges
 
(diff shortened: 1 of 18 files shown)


mberhault:
LGTM. Thanks!

PR #3781, opened 2018-09-25
Delete hackathon.md

mjibson:
I can't see it anywhere so I think it's leftover and can be removed. Anyone aware of this being used?


jseldess:
@sploiselle, do you think we'll need this file for future hackathons?

sploiselle:
We aren't currently using it and it could be trivially re-added in the future. Think this is fine to delete. LGTM.

PR #3796, opened 2018-10-01
`cockroach debuz zip` doc updates

### LINKED ISSUE #3642: cli: include problem ranges in `debug zip` output
PR: https://github.com/cockroachdb/cockroach/pull/28253

From release notes:
> The file generated by running `cockroach debug zip` now contains the contents of the `system.rangelog` table, which is a record of range splits and rebalances in the cluster. The problem ranges report is now included as well. [#28396][#28396] [#28253][#28253] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closes #3642 

Summary of changes:
The original issue (#3642 ) was for adding Problem Ranges to the list of things in the debug zip. In addition to that, I also added other things that appear in the debug zip (like cluster settings, etc.) that were missing from the doc.

Also deleted one instance of a warning that appeared twice.

### DIFF
--- a/v2.1/debug-zip.md
+++ b/v2.1/debug-zip.md
@@ -17,8 +17,8 @@ The `debug zip` [command](cockroach-commands.html) connects to your cluster and 
 - [Cluster Settings](cluster-settings.html)
 - [Metrics](admin-ui-custom-chart-debug-page.html#available-metrics)
 - Alerts
-- Heap profiles (**new in v2.0**)
-- Problem ranges (**new in v2.1**)
+- **New in v2.0:** Heap profiles 
+- **New in v2.1:** Problem ranges
 
 {{site.data.alerts.callout_danger}}The file produced by <code>cockroach debug zip</code> can contain highly sensitive, unanonymized information, such as usernames, passwords, and possibly your table's data. You should share this data only with Cockroach Labs developers and only after determining the most secure method of delivery.{{site.data.alerts.end}}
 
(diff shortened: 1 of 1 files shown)


Amruta-Ranade:
Yeah..I kept it consistent with the flag for `heap profile`, but should change both to the new flag. Thanks for pointing it out!

Amruta-Ranade:
Also, do we keep the new in 2.0 flag in 2.1 docs? 🤔

v2.1/debug-zip.md
```diff
@@ -7,13 +7,18 @@ toc: true
...
 - Schema change events
 - Node liveness
 - Gossip data
 - Stack traces
 - Range lists
 - A list of databases and tables
-- Heap profiles (**new in v2.0**)
+- [Cluster Settings](cluster-settings.html)
+- [Metrics](admin-ui-custom-chart-debug-page.html#available-metrics)
+- Alerts
```
jseldess: How do "Alerts" and "Cluster events" differ?

Amruta-Ranade: As I understand it, Alerts = the configured alerts (https://www.cockroachlabs.com/docs/dev/monitoring-and-alerting.html#events-to-alert-on), while Cluster Events = events  like node joining the cluster, decommissioning, etc.

@bdarnell Please correct me if I am wrong.

bdarnell: Yeah, that's the right idea, although the alerts collected here are not exactly the same as those on the monitoring and alerting page. Alerts are things like "3 under-replicated ranges" or "raft group has no leader" that probably represent a problem with the cluster (the kind of thing we'd show in a red box in the web UI). Right now there's not much here but we plan to add to it. 

Cluster events, on the other hand, are "normal" things like node starts and stops, table creations, etc.

jseldess: OK, in that case, let's remove the cluster events link to the monitoring docs. That's misleading. Fine to just leave it a bullet.

Amruta-Ranade: Fixed.

PR #3812, opened 2018-10-03
Clarify that per-replica constraints don't need to add up to total replicas

### LINKED ISSUE #3097: *: let per-replica constraints specify less than all a zone's replicas
PR: https://github.com/cockroachdb/cockroach/pull/23057

From release notes:
> Per-replica constraints in replication zones no longer have to add up to the total number of replicas in a range. If all replicas aren't specified, then the remaining replicas will be allowed on any store. [#23057][#23057] {% comment %}doc{% endcomment %}

jseldess:
Fixes #3097.


jseldess:
Thanks, @a-robinson. I applied the change to the 2.0 version as well.

PR #3818, opened 2018-10-03
Document replication zones for secondary indexes

### LINKED ISSUE #3146: Setting replication zones for secondary indexes
This is missing from our [docs](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html). 

See https://github.com/cockroachdb/cockroach/issues/25035 for details.

jseldess:
Fixes #3146.


jseldess:
@a-robinson, just updated the note about what happens when you create a zone for an index, and the index data is stored in a range with table data. @bdarnell mentioned that `we always split on zone config boundaries, so if you give the table and index different zone configs they will get split`.

PR #3826, opened 2018-10-04
Sql comparison

### LINKED ISSUE #3801: Detailed SQL Standard Comparison Doc Update
Re: [Detailed SQL Standard Comparison](https://cockroachlabs.com/docs//v2.1/detailed-sql-support.html)

## Issue Description

Certain rows are highlighted in bold, but this does not provide any additional insight, instead it distracts from the reading.

## Suggested Resolution 

Remove the bold formatting from table rows.

lhirata:
- Removed bolding
- Replaced < and > so they can render correctly

Closes #3801.

### DIFF
--- a/v2.1/detailed-sql-support.md
+++ b/v2.1/detailed-sql-support.md
@@ -13,7 +13,7 @@ To understand the extent to which we support the standard SQL features, use the 
 
 ## Features
 
-|  **Feature ID** | **Description** | **CockroachDB Support** |
+|  Feature ID | Description | CockroachDB Support |
 |  ------ | ------ | ------ |
 |  B011 | Embedded Ada | No |
 |  B012 | Embedded C | No |
@@ -25,7 +25,7 @@ To understand the extent to which we support the standard SQL features, use the 
 |  B021 | Direct SQL | No |
 |  B031 | Basic dynamic SQL | No |
 |  B032 | Extended dynamic SQL | No |
-|  B032-01 | <describe input statement> | No |
+|  B032-01 | &lt;describe input statement&gt; | No |
 |  B033 | Untyped SQL-invoked function arguments | No |
 |  B034 | Dynamic specification of cursor attributes | No |
 |  B035 | Non-extended descriptor names | No |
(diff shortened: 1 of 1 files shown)


v2.1/detailed-sql-support.md
```diff
@@ -7,682 +7,682 @@ This page lists which SQL standard features are supported, partially-supported,
 
 To understand the extent to which we support the standard SQL features, use the table below.
 
-- **Feature ID** is the SQL Standard feature identification number.
-- **Description** provides details about the feature.
-- **CockroachDB Support** indicates whether the feature is supported, unsupported, or partially-supported by CockroachDB.
+- Feature ID is the SQL Standard feature identification number.
+- Description provides details about the feature.
+- CockroachDB Support indicates whether the feature is supported, unsupported, or partially-supported by CockroachDB.
```
rmloveland: I think the bold in this part is necessary to indicate that we are referring to column headers in the table below.  The original issue only mentions "remove bold formatting from table rows".

v2.1/detailed-sql-support.md
```diff
@@ -7,682 +7,682 @@ This page lists which SQL standard features are supported, partially-supported,
...
+X284 | XMLValidate: NAMESPACE without ELEMENT clause | No
+X286 | XMLValidate: NO NAMESPACE with ELEMENT clause | No
+X300 | XMLTable | No
+X301 | XMLTable: derived column list option | No
+X302 | XMLTable: ordinality column option | No
+X303 | XMLTable: column default option | No
+X304 | XMLTable: passing a context item | No
+X305 | XMLTable: initializing an XQuery variable | No
+X400 | Name and identifier mapping | No
+X410 | Alter column data type: XML type | No
```
rmloveland: Can this change be limited to only removing the bold formatting from the table rows as requested in the linked issue?  I totes understand the philosophy of "let's fix this while we're in here" but I think a formatting-only change of this size might be better considered in a separate PR.  This diff is huge and while it's probably fine I can't honestly say I'm able to give it a thorough review, and I wouldn't want to miss something obvious.

lhirata:
@rmloveland changed it back to just be removing the bolded rows. For some reason, Reviewable isn't showing the changes very well, so it might be easier to check the Files Changed tab instead.

jseldess:
I agree with @knz that the bolding seems to serve the purpose of calling out related sections of features, but I also agree that it feels inconsistent. This entire page is modeled on [postgres' equivalent page](https://www.postgresql.org/docs/9.5/static/features-sql-standard.html). They don't bold anything there. So I'd leave it up to @lhirata. Either we remove the bold formatting or add bold to any feature that tis the "superordinate" of other features.

knz:
FWIW I am OK with not bolding anything.

lhirata:
Ok, thanks for the feedback! I'll stick with no bolding. @rmloveland, can you review / unblock?

PR #3832, opened 2018-10-04
Add better examples / language around monotonicity

### LINKED ISSUE #612: Expand SERIAL documentation
Excerpts from Gitter:

Adam @apinckard 09:39
Hi all. Quick question is the SERIAL data type guaranteed to be monotonically increasing as well as globally unique? (understand that is not always sequential)

... So we are intending to use it for very high frequency. If two transactions occur concurrently can we guarantee monotonically increasing e.g. first to commit < second to commit?

kena @knz 09:50
right now the SERIAL values are based off the current time up to microsecond precision. You always get different values on different nodes, but if you have 2 concurrent transactions on the same node asking for SERIAL values at the same time, one could get a value and another the same value or just below (or above) it with no guarantee on ordering.
no the ordering is not related to commit order.
...
**the nearest we have to what you want/need is another function called cluster_logical_timestamp() which returns a DECIMAL value. It's heavier than a simple int**

Adam @apinckard 09:56
...
One follow up: within a single transaction is the output of serial guaranteed to be monotonically increasing?

kena @knz 10:43
within a single transaction, the output of serial is currently guaranteed to be monotonically increasing as long as your system clock does not jump backwards
if it does (which seems to happen in cloud servers, where VMs are migrated live), all bets are off
feel free to summarize the current situation and what you'd like to see instead in our github issue tracker.

### LINKED ISSUE #3491: SERIAL Doc Update
Re: [SERIAL](https://cockroachlabs.com/docs//v2.0/serial.html)

## Issue Description

Discussed on https://forum.cockroachlabs.com/t/unique-rowid-monotonicity-question/1888

The user was confused by the paragraph:

> It's a common misconception that the auto-incrementing types in PostgreSQL and MySQL generate strictly sequential values. In fact, each insert increases the sequence by one, even when the insert is not commited. This means that auto-incrementing types may leave gaps in a sequence.

The fact that there are gaps is one thing, but the text still suggests vaguely (using the word "increases") that the values increase monotonically.

However SERIAL currently uses `unique_rowid()` and is thus not monotonic.

Even if it did use sequences, like postgres does, it still would not be guaranteed to produce monotonic values. (Incidentally, postgres does not guarantee monotonicity either).

## Suggested Resolution 

The documentation for SERIAL should highlight that the data type does not provide any monotonicity guarantee.

Perhaps rephrase like this:

```
It's a common misconception that the auto-incrementing types in PostgreSQL and MySQL generate strictly sequential values. There can be gaps and the order is not completely guaranteed.

- each insert increases the sequence by one, even when the insert is not commited. This means that auto-incrementing types may leave gaps in a sequence.
- two concurrent transactions can commit in a different order than their use of sequences, and thus "observe" the values to decrease relative to each other. This effect is amplified by automatic transaction retries.

These are fundamental properties of a transactional systems with non-transactional sequences. Neither PostgreSQL, MySQL nor CockroachDB increase sequences transactionally with other SQL statements, so these effects can happen in any case.
```

lhirata:
Closes #3491
Closes #612

### DIFF
--- a/_includes/v2.1/faq/auto-generate-unique-ids.html
+++ b/_includes/v2.1/faq/auto-generate-unique-ids.html
@@ -84,4 +84,4 @@ If it's important for generated IDs to be stored in the same key-value range, yo
 (3 rows)
 ~~~
 
-On insert, the `unique_rowid()` function generates a default value from the timestamp and ID of the node executing the insert. Such time-ordered values are likely to be globally unique except in cases where a very large number of IDs (100,000+) are generated per node per second.
+On insert, the `unique_rowid()` function generates a default value from the timestamp and ID of the node executing the insert. Such time-ordered values are likely to be globally unique except in cases where a very large number of IDs (100,000+) are generated per node per second. Also, there can be gaps and the order is not completely guaranteed.
--- a/_includes/v2.1/sql/functions.md
+++ b/_includes/v2.1/sql/functions.md
@@ -375,7 +375,7 @@ has no relationship with the commit order of concurrent transactions.</p>
 </span></td></tr>
 <tr><td><code>gen_random_uuid() &rarr; <a href="uuid.html">uuid</a></code></td><td><span class="funcdesc"><p>Generates a random UUID and returns it as a value of UUID type.</p>
 </span></td></tr>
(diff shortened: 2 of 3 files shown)


lhirata:
@rmloveland or @jseldess when you have time, can one of you review this?

PR #3842, opened 2018-10-09
add class to search form for query tracking

sploiselle:
To understand users' queries, I need to add a class to track submissions to this form.


jseldess:
LGTM. This is for Heap, @sploiselle?

sploiselle:
@jseldess Yep

Issue #3591, opened 2018-08-27
Follow-up improvements to multi-region Kubernetes guide


a-robinson:
Unresolved comments from https://github.com/cockroachdb/docs/pull/3558:

- [x]  Port from v2.1 back to v2.0 (for @jseldess)
- [x] Clarify the namespace situation (for @a-robinson)
- [x] Document the firewall requirements (@a-robinson)
- [x] Simulate dc failure instead of node failure - see comment below (for @a-robinson)

a-robinson:
3. Document the firewall requirements (@a-robinson)

jseldess:
Also:

In the multi-region k8s tutorial, instead of simulating node failure I think it would be much more interesting to simulate entire datacenter failure by either scaling the number of VM instances in one of the GKE clusters down to 0 (`gcloud container clusters resize <cluster-name> --zone=<zone> --size=0`) or all the pods in one of the clusters down to 0 (`kubectl scale statefulset cockroachdb --replicas=0 --namespace=<cluster-namespace> --context=<cluster-context>`).

jseldess:
Hey @a-robinson, do you think you'll have time to handle the two bullets assigned to you? If not, not a problem to push to 2.2.

a-robinson:
I'm having trouble reproducing the firewall issue or even understanding how you ran into it in the first place. All of our GCE projects have a "default-allow-internal" firewall rule that allows traffic between all VMs in the network on all ports. Also, all GKE VMs (at least the ones in my personal project) are given a  "gke-<cluster-name>-<cluster-hash>-vms" firewall rule which allows traffic on all ports from VMs inside the project's network to all the VMs in the GKE cluster. And another firewall rule is added for each GKE cluster that allows communication from the pod IP ranges (if/when it falls outside the normal network CIDR) to all VMs in the project.

So I'm really not sure why you ran into a connectivity problem, or why adding a firewall rule that allows traffic to all VMs from all IPs on port 26257 did anything that the existing rules didn't. I'll see if I can repro it in that project, though, since trying in mine hasn't worked.

a-robinson:
Well I'm still not sure why it works in my project without the firewall rule to allow all traffic to port 26257, but I now see why it doesn't work in the shared project that you used (when I remove that rule).

[...]

PR #3845, opened 2018-10-10
Explain the unconventional namespace usage in multi-cluster k8s setup

a-robinson:
Fixes part of #3591 

Relies on https://github.com/cockroachdb/cockroach/pull/31196 getting in first, since it's what automates the creation of the cockroachdb-public service in each cluster's default namespace.


a-robinson:
I've added another commit that simulates entire datacenter failure instead of just single node failure.

PR #3855, opened 2018-10-12
[Do not merge] Update the IMPORT syntax

knz:
Goes with https://github.com/cockroachdb/cockroach/pull/31263

### DIFF
--- a/v2.1/import.md
+++ b/v2.1/import.md
@@ -257,7 +257,7 @@ WITH
 
 {% include copy-clipboard.html %}
 ~~~ sql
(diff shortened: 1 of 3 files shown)


knz:
Note that both forms (with and without parentheses) are already accepted by CockroachDB. The parentheses are now promoted for consistency with the syntax form using DATA.

knz:
let's not merge this just yet. David found a compelling argument why this PR may not be needed/useful after all.

jseldess:
OK. Just post an update here when you have one, @knz.

jseldess:
@knz, any updates here?

knz:
The change in CockroachDB was cancelled. There is nothng left to do here.

PR #3874, opened 2018-10-16
Managed docs UI

j-low:
__[description]__
* User interface for rollout of managed service

cc: @jseldess

### DIFF
--- a/v2.1/managed-sign-up-for-a-cluster.md
+++ b/v2.1/managed-sign-up-for-a-cluster.md
@@ -16,9 +16,9 @@ The Managed CockroachDB offering is currently in Limited Availability and accept
 
 To be considered for participation:
 
-1. [Sign up for an Enterprise Trial license](https://www.cockroachlabs.com/pricing/start-trial/).
+1. [Sign up for an Enterprise Trial license](https://www.cockroachlabs.com/get-cockroachdb/).
 
-2. During sign up, indicate that you're interested in the Managed offering. You'll be asked to provide details about your company, use case, and deployment requirements.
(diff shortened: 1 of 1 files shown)


jseldess:
@j-low, @benesch, I can't get the `CockroachDB` and `Managed CockroachDB` links in the secondary topnav to work. Also, @benesch, even after rebasing against master, I'm seeing the link checker problems again. Can you help?

One other problem to solve after the above: Getting "CockroachDB" to be highlighted in the secondary top nav when on a page in the standard docs, and getting "Managed CockroachDB" to be highlighted in the secondary top nav when on a page in the managed docs.

jseldess:
@benesch, I figured out the htmltest errors. Just needed to add alt attributes to the new CockroachDB /Managed CockroachDB nav images. 

Remaining things I'd love your help on:
1. I hard-coded the CockroachDB / Managed CockroachDB links. That's not ideal at all, of course. Can you see a way around that?
2. We need to find a way to highlight CockroachDB when the user is on those pages, and Managed CockroachDB when the user is on those pages.

benesch:
Ok, I think I've taken care of both issues. PTAL.

jseldess:
Nice! Thanks very much @benesch! There's a redirect happening whenever you click `CockroachDB` or `Managed CockroachDB`, but that's probably ok.

benesch:
Yeah, not seeing a great way around that redirect. Now that we use Netlify it should be possible to avoid the client side "redirecting you now" flicker. Perhaps something @mjibson could look into.

benesch:
Actually, maybe there is. One sec.

jseldess:
@j-low, one more minor problem:

When you click an anchor link, the heading ends up hidden behind the new nav:

screen shot 2018-10-26 at 3 14 46 pm

PR #3879, opened 2018-10-17
Update Encryption at Rest experimental warning

### LINKED ISSUE #3877: Make Encryption at Rest warning scarier
The Encryption at Rest feature is going to be "experimental" in 2.1. Right now we have a generic warning at the top of [the page](https://www.cockroachlabs.com/docs/v2.1/encryption.html) that says:
> Warning:
> This is an experimental feature. The interface and output of this feature are subject to change.

We need to make this warning stronger, as it doesn't really convey the danger of using encryption at rest in its current state. Specifically, in the case of a bug or user error, all the data on an encrypted node could be rendered unreadable.

rmloveland:
In particular, make it more explicit and scarier.

Fixes #3877.

### DIFF
--- a/v2.1/encryption.md
+++ b/v2.1/encryption.md
@@ -7,8 +7,12 @@ toc: true
 <span class="version-tag">New in v2.1:</span>
 Encryption at Rest provides transparent encryption of a node's data on the local disk.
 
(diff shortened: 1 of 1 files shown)


mberhault:
LGTM

PR #3887, opened 2018-10-18
Add page listing experimental features

### LINKED ISSUE #2914: Create docs section for "Experimental" features
From a chat with @awoods187: 

> it might be useful to have an `experimental` section in the docs for people to easily find the whole list of things subject to change to try out (at least when we've decided to document an experimental feature)

rmloveland:
Fixes #2914.

Summary of changes:

- Add a new page **SQL Syntax > Experimental Features** which lists the experimental features available in CockroachDB 2.1.

- For each experimental feature, either link to documentation (if available), or write a brief description and example using comments and statements adapted from source code in the cockroachdb/cockroach repo.

### DIFF
--- a/_includes/sidebar-data-v2.1.json
+++ b/_includes/sidebar-data-v2.1.json
@@ -1437,6 +1437,12 @@
                 ]
               }
             ]
+          },
+          {
+            "title": "Experimental Features",
+            "urls": [
+              "/${VERSION}/experimental-features.html"
+            ]
           }
         ]
       },
--- a/v2.1/experimental-features.md
+++ b/v2.1/experimental-features.md
@@ -1,0 +1,150 @@
+---
+title: Experimental Features
+summary: Learn about the experimental features available in CockroachDB v2.1
+toc: true
+---
+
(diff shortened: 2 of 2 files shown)


rmloveland:
Direct link for easier reference:

[Experimental Features](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/639d6567a9bdbfa2b3255fa4c62412724d619559/dev/experimental-features.html)

rmloveland:
@a-robinson re:

> An example of EXPERIMENTAL_RELOCATE might be nice, but it's not necessary. Your call. I'm not sure what the general approaches to documenting sql statements are.

In general we are committed to providing good examples (AIUI).  I would like to meet that standard here, but there are a couple of reasons why I didn't in this case:

1. We don't have the same commitment to documenting experimental features as "standard" ones.
2. I'm taking advantage of that a little to punt here, since creating a good example would require more effort than I can expend at this time (but I would def be open to improving these examples in the future).
3. This page still provides value, since before we had no docs for some of these things.

PS Not really telling you anything you don't know :-) - mostly leaving this here for context in case someone else on this issue is wondering / wants to discuss.

PR #3904, opened 2018-10-23
Inverted Index Comparisons

tim-o:
Documenting Ben's knowledge bomb from [the forum](https://forum.cockroachlabs.com/t/jsonb-filtering-on-numbers/2094/2).

This isn't super intuitive and it's probably going to come up again.

### DIFF
--- a/v2.1/inverted-indexes.md
+++ b/v2.1/inverted-indexes.md
@@ -82,7 +82,11 @@ Tables are not locked during index creation thanks to CockroachDB's [schema chan
 
 Indexes create a trade-off: they greatly improve the speed of queries, but slightly slow down writes (because new values have to be copied and sorted). The first index you create has the largest impact, but additional indexes only introduce marginal overhead.
 
-
+### Comparisons
+Currently, inverted indexes only support equality comparisons using the `=` operator. If you require comparisons using `>`, `<=`, et al., you can create an index on a computed column using your JSON payload, and then create a regular index on that. So if you wanted to write a query where the value of "foo" is greater than three, you would:
+1. Create your computed column: `create table test(id int, data jsonb, foo int as ((data->>'foo')::int) stored);`
+2. Create an index on your stored column: `create index test_idx on test (foo);`
+3. Execute your query: `select * from test where foo > 3;`
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


v2.1/inverted-indexes.md
```diff
@@ -156,6 +158,14 @@ Now, run a query that filters on the `JSONB` column:
 +--------------------------------------+----------------------------------+--------------------------------------------------------------------------+
 (2 rows)
 ~~~
+### Comparisons
+Currently, inverted indexes only support equality. If you require comparisons like `>`, `<=`, etc., you can create an index on a computed column using your json payload, and then create a regular index on that. So if you wanted to write a query where the value of "foo" is greater than three, you would:
```
rmloveland: 1. Suggest expanding the first part of the first sentence slightly to add context - e.g., something like:

> Currently, inverted indexes only support equality comparisons using the `=` operator.  If you require comparisons using `>`, ...

2. 'json' should be capitalized to 'JSON'.

3. This `### Comparisons` section doesn't really fit here (as a subsection of *Example*).  Suggest moving it up the page so it comes directly after the *Performance* section (keeping the existing `###` level)

v2.1/inverted-indexes.md
```diff
@@ -156,6 +158,14 @@ Now, run a query that filters on the `JSONB` column:
...
+1) Create your computed column:
```
rmloveland: The list of steps is rendering a little weird on the built page at http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/953ab54c7314a112c6b0be7ec78ee6dd8743b3c7/dev/inverted-indexes.html

1. Suggest editing slightly to:

```
... you would:

1. Create your computed column: `CREATE TABLE test(id INT, data JSONB, foo INT AS ((data->>'foo')::int) STORED);`
2. Create an index on your stored column: `CREATE INDEX test_idx ON test (foo);`
3. Execute your query: `SELECT * FROM test3 WHERE foo > 3;`
```

2. I don't see where `test3` is created before step 3; do you mean `SELECT * FROM test_idx` since `test_idx` was created in step 2?

tim-o: Test 3 was an artifact from Ben's example. Good catch. Using the test table should use test_idx implicitly.

v2.1/inverted-indexes.md
tim-o: 👍 Done.

tim-o:
Hrm, something still wonky with the formatting. I'll fix before merging.

tim-o:
Added spaces and clipboard... fingers crossed!

tim-o:
Looks good. Merged.

PR #3916, opened 2018-10-25
Re-use glossary of essential terms

jseldess:
- In architecture overview
- In performance tuning tutorial


sploiselle:
Nice! Good call, @jseldess. We'll need to change this relative link:

```
[Raft consensus protocol](architecture/replication-layer.html#raft)
```

to...something else (these pages are in different directories, so they have to use something other than a relative link with this structure). The sanest thing I can think to do is link from the site's root, which necessitates including a template string to identify the currently loaded version––is that something we have handy?

jseldess:
Ah, good catch. I don't know how to accomplish what you're suggesting, though it sounds like an efficient approach. For now, I'll just use an absolute url. That ok?

sploiselle:
LGTM

PR #3917, opened 2018-10-25
Fix Netlify managed redirects

benesch:
The _redirects file needs to be built for both the standard and managed
docs. I think.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,5 +1,6 @@
 ---
 layout: null
+build_for: [standard, managed]
 ---
 
 {% for version in site.versions %}


jseldess:
LGTM, at least in the TeamCity build on s3, though I don't understand this change. 

@mjibson, can you chime in? The problem is that, on the live site, the docs at `/docs/managed/` are only available when you use the explicit `v2.1` in the url (/docs/managed/v2.1/). The `stable` alias doesn't work (/docs/managed/stable). Here are examples:

https://www.cockroachlabs.com/docs/managed/stable/
https://www.cockroachlabs.com/docs/managed/v2.1/

benesch:
Ok, so here's the explanation. Version aliases are kinda tricky, because you want to totally duplicate the content under a different name. Jekyll is slow enough that generating every page twice is prohibitively slow. So when I initially set up versioning, I used symlinks. "stable" is a symlink for v2.0, "dev" is a symlink for v2.1, and so on. No copies necessary; Jekyll literally runs e.g. `ln -s v2.1 dev` when it generates the site.

This worked out really nicely, because `jekyll serve` does the right thing with symlinks, as does GitHub pages. Netlify, sadly, doesn't handle symlinks. Luckily, it supports a `_redirects` file, which can accomplish the same thing. In addition to writing the symlink, to support local servers, we teach Netlfiy about the symlink by adding a line to the _redirects file that looks like this:

```
/docs/stable /docs/v2.1 200
```

The standard build needs to output a `_redirects` file like the above, while the managed build needs to output a redirects file with a /managed prefix:

```
/docs/managed/stable /docs/managed/v2.1 200
```

Naturally, when I converted the Python scripts to Jekyll plugins, I overlooked this, and since the `_redirects` file didn't specify `build_for` it wasn't getting built for the managed docs, and so Netlify didn't know about the redirects.

jseldess:
Thanks for the write-up, @benesch.

mjibson:
LGTM

Issue #2964, opened 2018-04-20
Merge ranges


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recVwbKWUfu87YRE4
PM: @awoods187 
Eng Partner: @benesch

jseldess:
@benesch, I know this work is done, so when you feel like it's stable and ready for documentation, please work with Rich to understand the needed docs updates. Mostly, I think it'll be architecture doc updates.

benesch:
Sounds good. We still haven't committed to turning merges on for v2.1, so
no doc updates yet. I'd say it's 50–50 whether we turn them on. Maybe it
would be easier to prepare the docs ahead of time just in case? I don't
know how swamped you guys are.

jseldess:
Thanks, @benesch. Probably best to wait until we know if the feature is in or out. There's plenty left to do.

rmloveland:
Hey @benesch just checking in: is range merging going to be turned on by default for 2.1?

awoods187:
The plan is on by default unless we discover a bad bug between now and then

jseldess:
Off by default in 2.1.

awoods187:
It should be p-0. Range merges is in 2.1 (but off by default). We will want to recommend some users to try it in 2.1.2

sploiselle:
@sploiselle Merge into 2.2 docs

PR #3920, opened 2018-10-25
Remove range merge cluster setting from 2.1.0 docs

rmloveland:
Relates to #2964.

For v2.1.0, range merges will be disabled, and the cluster setting will not be available to users.


rmloveland:
@benesch this change comes from grabbing `docs/generated/settings.html` from the `release-2.1` branch.  Hopefully that was the right version to grab.

benesch:
Lovely! LGTM.

By the way, I feel compelled to share that cluster setting is still there secretly, so if you know the exact name of the setting you can still turn it on. But we don't want anyone stumbling across it and thinking "this looks fun!" and flipping it to on, because that will definitely crash their cluster until we fix a few more merge bugs.

rmloveland:
> that cluster setting is still there secretly, so if you know the exact name of the setting you can still turn it on

⚠️ DOUBLE SECRET CLUSTER SETTING ⚠️ 

gotcha thanks :-)

PR #3922, opened 2018-10-25
Ensure int/date/timestamp conversions are clearly marked as experimental

knz:


knz:
cc @sploiselle since this touches one of the known limitations you pinged me about.

knz:
TFYR

PR #3923, opened 2018-10-25
More correctly document SERIAL.

### LINKED ISSUE #3630: sql: introduce virtual sequences and use with SERIAL
PR: https://github.com/cockroachdb/cockroach/pull/28575

From release notes:
> CockroachDB now supports two experimental compatibility modes with how PostgreSQL handles [`SERIAL`](../v2.1/serial.html) and [sequences](../v2.1/create-sequence.html), to ease reuse of 3rd party frameworks or apps developed for PostgreSQL. These modes can be enabled with the `experimental_serial_normalization` session variable (per client) and `sql.defaults.serial_normalization` cluster setting (cluster-wide). The first mode, `virtual_sequence`, enables compatibility with many applications using `SERIAL` with maximum performance and scalability. The second mode, `sql_sequence`, enables maximum PostgreSQL compatibility but uses regular SQL sequences and is thus subject to performance constraints. [#28575][#28575] {% comment %}doc{% endcomment %}

knz:
SERIAL has always been a pseudo-type. It never possessed the
attributes that are required of real types (For example, it does not
have values of its own. It cannot be stored. It cannot be produced by
scalar expressions. It does not have comparison rules. It is not
independent, like other types, from the presence or absence of a
DEFAULT clause in the same column definition CREATE TABLE.)

The notion that it could be documented as a data type was erroneous, a
past misunderstanding.

Moreover, due to compatibility problems with PostgreSQL, is is not
impossible that the translation of SERIAL to `DEFAULT unique_rowid()`
be changed in the future to use SQL sequences instead; if/when that
occurs a new keyword may be added to preserve the `rowid()` syntax.

Therefore, it is desirable in the current doc to spell out `DEFAULT
unique_rowid()` throughout all the tutorials and examples to ensures
the table definitions are future-proof.

Fixes #3630.


knz:
@jseldess I agree the documentation was already *proper* in that it was very skillfully reflecting the understanding of the doc team when it was produced.

Nevertheless it was not *correct*. I will reword the title.

knz:
> can you explain the need to recategorize SERIAL as a pseudo-type and to tell users to not use it for new applications?

SERIAL has always been a pseudo-type. It never possessed the attributes that are required of real types (For example, it does not have values of its own. It cannot be stored. It cannot be produced by scalar expressions. It does not have comparison rules. It is not independent, like other types, from the presence or absence of a DEFAULT clause in the same column definition CREATE TABLE.) 

The notion that it could be documented as a data type was erroneous, a past misunderstanding.

jseldess:
Thanks for the added details, @knz!

knz:
The v2.0 docs are updated as well.

knz:
RFAL

jseldess:
LGTM!

PR #3949, opened 2018-10-30
Render search bar on Safari

benesch:
Safari inexplicably clips `position: fixed` elements to the boundaries
of their parent. This seems to violate the CSS spec--the only containing
box of a `position: fixed` element should be the viewport--but the
solution is to add some padding on the parent element so that the search
bar is not outside of its bounding box.

Also remove `height: fit-content`, which caused Safari to add about 50px
of whitespace beneath the search bar that was obscuring the navigation.
Not sure what the `height: fit-content` was for; things seem to render
fine without it.


j-low:
@benesch unfortunately this breaks in another way.  If you remove some of those `overflow-y` properties you end up with a sidebar that overruns the bottom boundary when scrolling down with a long sidebar ([example](/v2.0/deploy-cockroachdb-on-premises.html))

benesch:
Ooh, tricky! I knew I must've been overlooking something. PTAL at the new solution.

j-low:
@benesch looks great.  I'm going to push a few changes along with it that address some style issues but your won't affect your fix (e.g. only show the sidenav scroll bar on hover).  I'll merge once those are in place.

jseldess:
Thanks, @benesch and @j-low!

PR #3951, opened 2018-10-30
Make it easier to get docker image for specific versions

### LINKED ISSUE #3946: Document cockroach-unstable docker images
We publish docker images of our alpha and beta releases, but we don't document this clearly and users are [often unable to find them](https://forum.cockroachlabs.com/t/docker-images-of-testing-versions-beta-rc/2117). The only place this is mentioned is the `install-cockroachdb` page, which will use the correct image name, but only if the branch selector is pointing to a dev branch (today we don't have a dev branch so I can't link to it, but once we start the dev branch for 2.2 it'll show up again). Getting to the right version of the installation page is not obvious. I think we should mention the docker image on the release notes page alongside the download links.

Also, the "Docker Hub" link on both the 2.0 and 2.1 install pages points to the cockroach-unstable image. We should either fix this link or remove it.

jseldess:
Fixes #3946.

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -36,9 +36,15 @@ release_info:
   v2.1:
     name: v2.1.0
     version: v2.1.0
-    docker_image: cockroachdb/cockroach-unstable
+    docker_image: cockroachdb/cockroach
     build_time: 2018/10/30 14:48:26 (go1.10.3)
     start_time: 2018-10-30 15:10:52.34274101 +0000 UTC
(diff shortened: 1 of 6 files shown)


bdarnell:
LGTM. I wouldn't worry about backfilling older release notes; just do it going forward.

jseldess:
TFTR, @bdarnell. I'll see if I can get this into the release notes script, so we're sure to not miss it in future notes.

Issue #3902, opened 2018-10-23
"Contents" bar on right side of pages moves a lot when hovering cursor over it


a-robinson:
This is quite minor, but it was annoying me when looking through [the known limitations page](https://www.cockroachlabs.com/docs/dev/known-limitations.html).

When you aren't hovering over the "Contents" column on the right, all the text renders as if all the space over there is usable:

<img width="283" alt="screen shot 2018-10-23 at 10 50 42 am" src="https://user-images.githubusercontent.com/7085343/47373660-e38ca180-d6b1-11e8-8459-d6d65e18e328.png">

If you then hover over it to try to click on one of the headings, they all move by a line or two because the scroll bar gets rendered when you hover over the box, taking away some of the space previously used by the text and thus causing different line wrappings:

<img width="278" alt="screen shot 2018-10-23 at 10 50 53 am" src="https://user-images.githubusercontent.com/7085343/47373726-fef7ac80-d6b1-11e8-8b87-b0f187713479.png">

That makes it tougher than expected to click on one of the entries, since they all move when you put your mouse over them.

Also, the lack of bullet points or anything to differentiate the start of one entry from the end of another makes that whole column on the known limitations page a bit tough to quickly parse. All the entries run together visually. That's a separate concern, though, and one that probably affects this page more than any other on our site.

@j-low

jseldess:
Thanks for filing this, @a-robinson. These issues have bothered me at different times as well.

benesch:
Here's a supposed solution to the problem: https://stackoverflow.com/a/33264668/1122351

jseldess:
Fixed in the meantime.

PR #3953, opened 2018-10-30
Tweak some spacing

benesch:
Tweak the TOC spacing to make it more obvious where one entry ends and the next begins:

Before:

image

After:

image

Also add a tiny bit of spacing to the tiles on the docs homepage.

Before:

image

After:

image


j-low:
@benesch Definitely good with the changes.  This is the nittiest of nits but I added one extra pixel of bottom padding to the TOC sections.  Since the line height is getting pretty tight I think adding that in there helps to give a little more visual distinction between sections.  LGTM.

benesch:
Works for me, @j-low! @jseldess afraid that it doesn't actually address the issue in #3902. I added a comment there about one technique for fixing that, though.

PR #3956, opened 2018-10-31
Correct managed home page title

jseldess:
This fix causes the page title to be unique in browser tabs
and also helps us differentiate between standard docs home
and managed docs home in analytics data.


jseldess:
Merging to correct the analytics we're getting following the managed annoucement. @benesch, if you see optimizations, please comment and I'll follow up.

PR #3958, opened 2018-10-31
Remove SNAPSHOT support

### LINKED ISSUE #1585: sql: allow BEGIN READ WRITE
PR: https://github.com/cockroachdb/cockroach/pull/16348

sploiselle:
Mostly Rich's work, but includes BNF modification

Closes #1585


sploiselle:
@jseldess @rmloveland Found a bunch of mentions of "isolation level" that I hadn't thought to look up in the first pass. PTAL. One of the big questions I have is the treatment of `set-vars.md` given that there's a kind of impedance mismatch between the interface and the operation––you have any thoughts/suggestions there?

sploiselle:
@jseldess PTAL. I think that identifying these setting as _NOP_ is the most accurate.

After a brief discussion in cockroachdb/cockroach#32115, the setting is potentially important for ORM compatibility. My apprehension with removing it is that it could appear like the setting isn't available for the next generation of ORM developers that want to consider integrating CockroachDB support.

Issue #3415, opened 2018-07-23
More flexible certificates setup.


mberhault:
Ongoing work to be finished by 2.1: https://github.com/cockroachdb/cockroach/issues/26630

This will allow at least:
- separate CAs for node vs clients vs admin UI
- allow use of system CA pool for each category

And probably:
- required fields in certificates to tie certificates to cockroach clusters

And maybe:
- custom fields/values for the `node` client certificate

Assigning to myself for now to write the draft of this. Not all of will be possible with the `cockroach cert` command, this is meant more for integration with other certificate tools.

jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/27916

From release notes:
> CockroachDB now supports a separate CA (`ca-ui.crt`) and certificate (`ui.crt`) for the Web UI. [#27916][#27916] {% comment %}doc{% endcomment %}

@mberhault, is there anything left to do for this roadmap item?

mberhault:
I need to write the draft docs for this. 2.1 feature work is done.

mberhault:
Some very brief explanation:

Everything is still in the --certs-dir. If you want to use a separate CA for client certificates, you'll need the following on each node:
* ca-ui.crt: the CA certificate used to sign the client certs
* client.node.crt: the client certificate for user node
* client.node.key: the private key to use along with client.node.crt

This will do two things:
* nodes will use client.node.crt (and .key) in their client TLS config when talking to other nodes.
* nodes will use ca-ui.crt to verify client certificates. ca.crt will not be used to verify client certificates. So you need to make sure all client certificates are signed by it

As a side-effect, node.crt no longer needs Client Authentication and does not need CN=node (but it's fine to keep those)

Another addition (also optional) is ui.crt and ui.key which are used for the Admin UI. This allows things like real public certificates (eg: Let's Encrypt) in the Admin UI.

mberhault:
Another things to document: switching from the single `ca.crt` to split `ca.crt` and `ca-client.crt` is tricky, especially if the desired is for a rolling restart, or to continue supporting clients with certificates signed by `ca.crt`. Will need to document this as well.

mberhault:
Here is a description of the uses of certificates and their requirements, along with multiple examples for when we would want to use various certificates.
I don't think we need to have `openssl` instructions for all the new certificates, at least not yet.

Certificates are used to verify identities in three different ways:

## Verifying CockroachDB node identity on port 26257 (sql and internal):

[...]

PR #3959, opened 2018-10-31
Initial content updates

Amruta-Ranade:
Partially addresses #3415 

Summary of changes:
- New include for UI cert content
- Updated the `Create Security Certificates` doc and `Use the Admin UI` doc with the new include.

To-Do:
- Add `include` to deployment and start-a-secure-cluster guides
- Test the procedure: Figure out how to configure the DNS names on one of the cloud platforms, use the DNS name to generate a certificate from Let's Encrypt, and then run through our documented steps.

### DIFF
--- a/v2.1/create-security-certificates.md
+++ b/v2.1/create-security-certificates.md
@@ -9,7 +9,7 @@ toc: true
   <a href="create-security-certificates-openssl.html"><button style="width:28%" class="filter-button">Use openssl</button></a>
 </div>
 
-A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires Certificate Authority (CA), node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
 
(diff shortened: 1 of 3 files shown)


v2.1/create-security-certificates.md
```diff
@@ -9,7 +9,7 @@ toc: true
   <a href="create-security-certificates-openssl.html"><button style="width:28%" class="filter-button">Use openssl</button></a>
 </div>
 
-A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires CA, node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication, which requires Certificate Authority (CA), node, and client certificates and keys. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, or use [`openssl` commands](https://wiki.openssl.org/index.php/).
```
mberhault: This makes it sound like we need the CA key. We should clarify to make it clear that we only need the CA certificate, but we need the node and client certificates and keys.

v2.1/create-security-certificates.md
```diff
@@ -45,6 +49,8 @@ File name pattern | File usage
 `client.<user>.crt` | Client certificate for `<user>` (e.g., `client.root.crt` for user `root`)
 `client.<user>.key` | Key for the client certificate
 
+If you have a certificate issued by a public CA to securely access the Admin UI, you need to place the certificate in the directory specified by the `--certs-dir` flag.
```
mberhault: Can we mention `ui.crt` and `ui.key` here, and make it clear they're optional?

mberhault:
LGTM.

Amruta-Ranade:
@jseldess Added the include to the manual deployment guides.
I need help figuring out the updates to the orchestrated deployment guides (and also figure out if the UI cert applies to them).
And as for the `Start a local cluster` guide, I don't think the UI certs should be recommended for the local clusters (as I understand it, the users cannot get UI certs for localhost clusters).

Amruta-Ranade:
@jseldess I have made the changes that we discussed in Wednesday's 1:1
- Removed the _include file
- Added the UI cert and key content as an example in the `cockroach cert` doc
- Referenced the example from other docs (Admin UI navigation doc and all deployment guides)

Let me know if it looks okay now.

PR #4020, opened 2018-11-06
Updated Azure Backup URL

tim-o:
Currently the doc specifies that the Host should be the container name - this isn't correct. This URL:
`'azure://foo.blob.core.windows.net/employees.sql?AZURE_ACCOUNT_KEY=12345&AZURE_ACCOUNT_NAME=foo';` will throw an error from Azure since we attempt to conect to `foo.blob.core.windows.net/foo.blob.core.windows.net`. There should not be a host at all in the `azure://` string.


jseldess:
@lhirata, since you're on the ops and tools writer, would you mind verifying this change?

PR #4032, opened 2018-11-13
Split install into OS-specific pages

### LINKED ISSUE #1856: Install page should change URL for each option

### LINKED ISSUE #4031: OS-specific install pages
Historically, our core [Install CockroachDB](https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html) page gets a high number of "Was this page helpful? Yes" responses, but it also gets too many "Was this page helpful? No" responses. For example, since 1/18, the page got 227 yes votes and 72 no votes. 

Because the current page covers all OSs, it's hard to glean where users are having problems.

jseldess:
Historically, our core Install CockroachDB page gets a high number of
"Was this page helpful? Yes" responses, but it also gets too many
"Was this page helpful? No" responses. For example, since 1/18, the
 page got 227 Yes votes and 72 No votes.

Because the current page covers all OSs, it's hard to glean where users
are having problems. Therefore, this PR:

- Splits the install instructions into 3 OS-specific pages. This makes
it possible to track No votes per OS. It also make it easier to link
users to instructions for specific OS install methods.
- Changes the JS on the original install-cockroachdb.html to detect
the user's OS and redirect to the appropriate OS-specific install
page. All internal links to the install page still point here so as
to save the user the effort of selecting the right OS.

The OS-specific pages still include the html for tracking installs
for marketing.

Fixes #4031.
Fixes #1856.


v2.1/install-cockroachdb-mac.html
```diff
@@ -0,0 +1,219 @@
+---
+title: Install CockroachDB
+summary: Install CockroachDB on Mac, Linux, or Windows. Sign up for product release notes.
+tags: download, binary, homebrew
+toc: true
+key: install-cockroachdb-mac.html
+redirect_from: install-cockroachdb-mac.html
```
benesch: I think you meant to say `install-cockroachdb.html` here. This creates an infinite redirect. :)

jseldess: Thanks. Fixed.

jseldess:
Thanks, @benesch. I took your suggested approach and added `install-cockroachdb.html` with JS to auto-detect the OS and redirect to the corresponding page. PTAL. 

I agree that, even with this split, we still won't know which install method is the problem. That's a problem we could look into addressing in the future as well.

v2.1/install-cockroachdb.html
```diff
@@ -1,573 +1,18 @@
...
+    // Detect OS and redirect to corresponding install page by default.
     if (navigator.userAgent.indexOf("Linux") !== -1) {
-        $("#linux").click();
+        location = 'https://www.cockroachlabs.com/docs/{{ page.version.version }}/install-cockroachdb-linux.html';
     } else if (navigator.userAgent.indexOf("Win") !== -1) {
-        $("#windows").click();
+        location = 'https://www.cockroachlabs.com/docs/{{ page.version.version }}/install-cockroachdb-windows.html';
     } else {
-        $("#mac").click();
+        location = 'https://www.cockroachlabs.com/docs/{{ page.version.version }}/install-cockroachdb-mac.html';
```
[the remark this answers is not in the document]
jseldess: Done.

jseldess:
@rolandcrosby, PTAL at the second commit, which tries to improve the "What's next?" links from the install pages, as we discussed offline.

rolandcrosby:
Very nice improvements, thanks Jesse!

jseldess:
TFTR, @benesch and @rolandcrosby.

PR #4033, opened 2018-11-13
Improve certs documentation

Amruta-Ranade:
Addresses #3415

### DIFF
--- a/v2.1/create-security-certificates-custom-ca.md
+++ b/v2.1/create-security-certificates-custom-ca.md
@@ -1,0 +1,137 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: true
+---
+
+To secure your CockroachDB cluster's inter-node and client-node communication, you need to provide a Certificate Authority (CA) certificate that has been used to sign keys and certificates (SSLs) for:
+
+- Nodes
+- Clients
+- Admin UI (optional)
+
+To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, use [`openssl` commands](https://wiki.openssl.org/index.php/), or use a [custom CA](create-security-certificates-custom-ca.html) (for example, a public CA or your organizational CA).
+
+<div class="filters filters-big clearfix">
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use cockroach cert</button>
+  <a href="create-security-certificates-openssl.html"><button style="width:28%" class="filter-button">Use Openssl</button></a>
+  <button style="width:28%" class="filter-button current">Use <strong>custom CA</strong></button></a>
+</div>
+
+This document discusses the following advanced use cases for using security certificates with CockroachDB:
+
+Approach | Use case description
+-------------|------------
+[UI certificate and key](#accessing-the-admin-ui-for-a-secure-cluster) | When you want to access the Admin UI for a secure cluster and avoid clicking through a warning message to get to the UI.
+[Split-node certificate](#split-node-certificates) | When your organizational CA requires you to have separate certificates for the node's incoming connections (from SQL and Admin UI clients, and from other CockroachDB nodes) and for outgoing connections to other CockroachDB nodes.
+[Split-CA certificates](#split-ca-certificates) | When you have multiple CockroachDB clusters and need to restrict access to clients from accessing the other cluster.
+
+## Accessing the Admin UI for a secure cluster
+
+On [accessing the Admin UI](admin-ui-access-and-navigate.html#access-the-admin-ui) for a secure cluster, your web browser will consider the CockroachDB-issued certificate invalid, because the browser hasn't been configured to trust the CA that issued the certificate.
+
+For secure clusters, you can avoid getting the warning message by using a certificate issued by a public CA whose certificates are trusted by browsers, in addition to the CockroachDB-created certificates:
+
+1. Request a certificate from a public CA (for example, [Let's Encrypt](https://letsencrypt.org/)). The certificate must have the IP addresses and DNS names used to reach the Admin UI listed in the `Subject Alternative Name` field.
+2. Rename the certificate and key as `ui.crt` and `ui.key`.
+3. Add the `ui.crt` and `ui.key` to the [certificate directory](create-security-certificates.html#certificate-directory). `ui.key` must not have group or world permissions (maximum permissions are 0700, or rwx------). You can disable this check by setting the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK=true`.
+4. For nodes that are already running, load the `ui.crt` certificate without restarting the node by issuing a `SIGHUP` signal to the cockroach process:
+   {% include copy-clipboard.html %}
+   ~~~ shell
+   pkill -SIGHUP -x cockroach
+   ~~~
+   The `SIGHUP` signal must be sent by the same user running the process (e.g., run with sudo if the cockroach process is running under user root).
+
+### Node key and certificates
+
+A node must have the following files with file names as specified in the table:
+
+File name | File usage
+-------------|------------
+`ca.crt`     | CA certificate created using the `cockroach cert` command.
+`node.crt`   | Server certificate created using the `cockroach cert` command. <br><br> `node.crt` must have `CN=node` and the list of IP addresses and DNS names listed in `Subject Alternative Name` field. <br><br>Must be signed by `ca.crt`.
+`node.key`   | Server key created using the `cockroach cert` command.
+`ui.crt` | UI certificate signed by the public CA. `ui.crt` must have the IP addresses and DNS names used to reach the Admin UI listed in `Subject Alternative Name`.
+`ui.key` | UI key corresponding to `ui.crt`.
+
+### Client key and certificates
+
+A client must have the following files with file names as specified in the table:
+
+File name | File usage
+-------------|------------
+`ca.crt`     | CA certificate created using the `cockroach cert` command.
+`client.<user>.crt` | Client certificate for `<user>` (e.g., `client.root.crt` for user `root`). <br><br>Each `client.<user>.crt` must have `CN=<user>`  (for example, `CN=marc` for `client.marc.crt`) <br><br> Must be signed by `ca.crt`.
+`client.<user>.key` | Client key created using the `cockroach cert` command.
+
+## Split node certificates
+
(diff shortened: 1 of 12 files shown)


v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+Split-node certificate | When your organizational CA requires you to have separate certificates for when the node acts as a server and as a client.
+Split-CA certificates | When you have multiple CockroachDB clusters and need to restrict access to clients from accessing the other cluster
+
+## Accessing the Admin UI for a secure cluster
+
+On [accessing the Admin UI](admin-ui-access-and-navigate.html#access-the-admin-ui), your browser will consider the CockroachDB-created certificate invalid, so you’ll need to click through a warning message to get to the UI.
+
+For secure clusters, you can avoid getting the warning message by using a certificate issued by a public CA:
+
+1. Request a certificate from a public CA (for example, [Let's Encrypt](https://letsencrypt.org/)). The certificate must have the IP addresses and DNS names used to reach the Admin UI listed in the `Subject Alternative Names` field.
```
mberhault: This should be: "listed in the Subject's `Common Name` field, or `Subject Alternative Name` field" (note the singular on "subject alternative name", applicable here and the rest of the doc).

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+We do not recommend you use split CA certificates unless your organizational security practices mandate you to do so.
+{{site.data.alerts.end}}
+
+If you need to use separate CAs to sign node certificates and client certificates, then we need two CAs and their respective certificates and keys: `ca.crt` and `ca-client.crt`.
+
+### Node key and certificates
+
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate to sign node certificates
```
mberhault: let's use "to verify" instead of "to sign". The CA does the signing, we're only verifying things.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+`ca-client.crt` | CA certificate to sign client certificates
+`node.crt`   | Node certificate for when node acts as server. <br><br>Must have the list of IP addresses and DNS names listed in `Subject Alternative Names`
```
mberhault: Add "this certificate must be signed using `ca.crt`, were mentioning this in the other certs for this scenario.

mberhault:
@bdarnell: would you like to take a look? I'm a bit close to this to see if we missed anything important.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. Request a certificate from a public CA (for example, [Let's Encrypt](https://letsencrypt.org/)). The certificate must have the IP addresses and DNS names used to reach the Admin UI listed in the `Subject Alternative Names` field.
+2. Rename the certificate and key as `ui.crt` and `ui.key`.
+3. Add the `ui.crt` and `ui.key` to the [certificate directory](create-security-certificates.html#certificate-directory). `ui.key` must not have group or world permissions (maximum permissions are 0700, or rwx------). This check can be disabled by setting the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK=true`.
+4. For nodes that are already running, load the `ui.crt` certificate without restarting the node by issuing a `SIGHUP` signal to the cockroach process:
+   {% include copy-clipboard.html %}
+   ~~~ shell
+   pkill -SIGHUP -x cockroach
+   ~~~
+   The `SIGHUP` signal must be sent by the same user running the process (e.g., run with sudo if the cockroach process is running under user root).
```
bdarnell: Change this sentence to:
The `SIGHUP` signal must be sent by either `root` or the same user as the `cockroach` process (so it may be necessary to send the signal with `sudo`).

It's a little weird to discuss this under the UI cert section but not in other places (even though it applies for all cert changes). It might be nice to provide a little more context here, that letsencrypt uses relatively short-lived certificates that are updated more frequently.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+
+A secure CockroachDB cluster uses [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) for encrypted inter-node and client-node communication and requires a Certificate Authority (CA) certificate as well as keys and certificates for nodes, clients, and (optionally) the Admin UI. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, use [`openssl` commands](https://wiki.openssl.org/index.php/), or use a custom CA (for example, a public CA or your organizational CA).
+
+<div class="filters filters-big clearfix">
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use cockroach cert</button>
+  <a href="create-security-certificates-openssl.html"><button style="width:28%" class="filter-button">Use Openssl</button></a>
+  <button style="width:28%" class="filter-button current">Use <strong>custom CA</strong></button></a>
+</div>
+
+This document discusses the following advanced use cases for using security certificates with CockroachDB:
```
bdarnell: Should there be a "simple" example for custom CAs, discussing the fields that must be set for the minimum configuration?  We should also mention that the three advanced use cases can be combined.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+   The `SIGHUP` signal must be sent by the same user running the process (e.g., run with sudo if the cockroach process is running under user root).
+
+### Node key and certificates
+
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate
+`node.crt`   | Server certificate. <br><br> `node.crt` must have `CN=node` and the list of IP addresses and DNS names listed in `Subject Alternative Names` field. <br><br>Must be signed by `ca.crt`
+`node.key`   | Key for server certificate
+`ui.crt` | UI certificate. `ui.crt` must have the IP addresses and DNS names used to reach the Admin UI listed in `Subject Alternative Names`
```
bdarnell: Consider bolding the ui.crt and ui.key lines since they are the unique part of this section (and doing the same for `client.node.*` and ca-client.crt below).

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate
+`client.<user>.crt` | Client certificate for `<user>` (e.g., `client.root.crt` for user `root`) and must be signed by `ca.crt`. Each `client.<user>.crt` must have `CN=<user>`  (for example, `CN=marc` for `client.marc.crt`)
+`client.<user>.key` | Key for the client certificate
+
+## Split node certificates
+
+The node certificate discussed in the `cockroach cert` command documentation is multifunctional, which means the same certificate is presented when the node acts as a server as well as a client. To make the certificate multi-functional, the `node.crt` created using the `cockroach cert` command has `CN=node` and the list of IP addresses and DNS names listed in `Subject Alternative Names` field. This works if you are also using the CockroachDB CA created using the `cockroach cert` command. However, if you need to use an external public CA or your own organizational CA, the CA policy might not allow it to sign a node certificate containing a CN that is not an IP address or domain name.
```
bdarnell: I think that last sentence should say "CA policy might not allow it to sign a *server* certificate containing a CN that is not an IP address". In this configuration we still require the node *client* certificate to have CN=node.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,117 @@
...
+File name pattern | File usage
+-------------|------------
+`ca.crt`     | CA certificate
+`client.<user>.crt` | Client certificate for `<user>` (e.g., `client.root.crt` for user `root`). <br><br>Each `client.<user>.crt` must have `CN=<user>`  (for example, `CN=marc` for `client.marc.crt`) <br><br>Must be signed by `ca.crt`
+`client.<user>.key` | Key for the client certificate
+
+## Split CA certificates
+
+{{site.data.alerts.callout_danger}}
+We do not recommend you use split CA certificates unless your organizational security practices mandate you to do so.
```
bdarnell: Why not? It's unnecessary, but why make this a "danger" callout?

jseldess:
___
*[v2.1/create-security-certificates.md, line 62 at r3](https://reviewable.io/reviews/cockroachdb/docs/4033#-LRh5-7DAMm36_MZDW2c:-LRh9TWs49i4bOVphlyr:bau3rid) ([raw file](https://github.com/cockroachdb/docs/blob/7a8898e22868fb7ecaff53d3336d78a0b265390e/v2.1/create-security-certificates.md#L62)):*
That's better, but what do we really mean by "we"? 

> This is needed to make sure that, when the node is acting as a client, it is indeed connected to another CockroachDB node, not a man-in-the-middle.

??

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,120 @@
...
+
+A secure CockroachDB cluster uses TLS 1.2 for encrypted inter-node and client-node communication and requires a Certificate Authority (CA) certificate as well as keys and certificates for nodes, clients, and (optionally) the Admin UI. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, use [`openssl` commands](https://wiki.openssl.org/index.php/), or use a custom CA (for example, a public CA or your organizational CA).
+
+<div class="filters filters-big clearfix">
+  <a href="create-security-certificates.html"><button style="width:28%" class="filter-button">Use cockroach cert</button>
+  <a href="create-security-certificates-openssl.html"><button style="width:28%" class="filter-button">Use Openssl</button></a>
+  <button style="width:28%" class="filter-button current">Use <strong>custom CA</strong></button></a>
+</div>
+
+This document discusses the following advanced use cases for using security certificates with CockroachDB:
```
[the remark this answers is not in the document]
sploiselle: After reading through this, there's something I'm missing that I think could be added to the doc. What, explicitly, is the information we're providing here that the user doesn't already have? Is this saying that we need/expect these specific files to have these specific names to implement these scenarios? That's just a guess.

v2.1/create-security-certificates-custom-ca.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Create Security Certificates
+summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
+toc: true
+---
+
+A secure CockroachDB cluster uses TLS 1.2 for encrypted inter-node and client-node communication and requires a Certificate Authority (CA) certificate as well as keys and certificates for nodes, clients, and (optionally) the Admin UI. To create these certificates and keys, use the `cockroach cert` [commands](cockroach-commands.html) with the appropriate subcommands and flags, use [`openssl` commands](https://wiki.openssl.org/index.php/), or use a custom CA (for example, a public CA or your organizational CA).
```
sploiselle: It looks like this is a slight rewrite of the original intro, so I think it's worth improving it a bit (feel free to disregard). I think that the doc might be better served by talking about the user's experience/needs, rather than what the product does/needs. As a straw man:

> To secure your CockroachDB cluster's inter-node and client-node communication, you need to provide a Certificate Authority (CA) certificate that has been used to sign keys and certificates (SSLs) for:
> - Nodes
> - Clients
> - Admin UI _(optional)_

This is a subtle change, but I think it helps users more clearly understand who needs to do what.

Amruta-Ranade:
@rolandcrosby @jseldess Ready for your final reviews :)

PR #4044, opened 2018-11-14
Improve docs around automatic retries

### LINKED ISSUE #3748: Update docs on automatic retries
Re: [Transactions](https://cockroachlabs.com/docs//v2.1/transactions.html)

> CockroachDB automatically retries individual statements and transactions sent from the client as a single batch.

This is incomplete (which makes it not true). Regardless of how things things are sent, we only ever do automatic retries is the results fit in a 16KiB buffer. Once that overflows, we start streaming results to the client and we forfeit the option of doing automatic retries. Which is really bad is someone relies on this statement being always true, as a client did.

andreimatei:
... explaining that not all implicit transactions are automatically
retried.

Fixes #3748

### DIFF
--- a/v2.0/transactions.md
+++ b/v2.0/transactions.md
@@ -72,11 +72,26 @@ There are two cases for handling transaction retries:
 
 ### Automatic Retries
 
-CockroachDB automatically retries individual statements and transactions sent from the client as a single batch.
+CockroachDB automatically retries individual statements (implicit transactions)
+and transactions sent from the client as a single batch, as long as the size of
+the results being produced for the client (including protocol overhead) is less
+than 16KiB. Once that buffer overflows, CockroachDB starts streaming results back to
+the client, at which point automatic retries cannot be performed any more. As
+long as the results of a single statement or batch of statements are known to
+stay clear of this limit, the client does not need to worry about transaction
+retries.
+
+In future versions of CockroachDB, we plan on providing stronger guarantees for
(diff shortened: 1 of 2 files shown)


andreimatei:
If this looks good, I'll add similar changes for 2.0 docs to the PR.

Separately, I'm making several code improvements that should result in more doc updates.

jseldess:
___
*[v2.1/transactions.md, line 82 at r1](https://reviewable.io/reviews/cockroachdb/docs/4044#-LRNLQ431A7A-ni99bnv:-LRNNEst81WkcoIZb9be:b-4sph99) ([raw file](https://github.com/cockroachdb/docs/blob/1686e4c4a81827dd8efa8827ddb44662cfe04504/v2.1/transactions.md#L82)):*

I'd like to keep the terminology simple and consistent here, so let's change `a query (or a batch's) results` `the results of a single statement or batch of statements` 
</blockquote></details>

I meant: Let's change TO `the results of a single statement or batch of statements`

jseldess:
Sorry about the, Andrei. 

Create the v2.2 today for the first 2.2.0-alpha release. It'll be in place as soon as https://github.com/cockroachdb/docs/pull/4051 lands.

jseldess:
@andreimatei, I'll merge this and apply your changes to the 2.2 version of the page in a follow-up PR. Thanks for your help!

PR #4082, opened 2018-11-20
Update 2.2 cluster settings doc

### LINKED ISSUE #4061: server: add a configuration to enable GC of system.rangelog
PR: https://github.com/cockroachdb/cockroach/pull/30913

From release notes:
> The range log and system events logs now automatically purge records older than 30 and 90 days, respectively. This can be adjusted via the `server.rangelog.ttl` and `server.eventlog.ttl` [cluster settings](../v2.2/cluster-settings.html). [#30913][#30913] {% comment %}doc{% endcomment %}

jseldess:
Fixes #4061.


jseldess:
This is really just porting over auto-generated content from the `cockroach` repo. Nothing much to check, really, so merging.

PR #4094, opened 2018-11-21
Clarify when to use DELETE vs. TRUNCATE

### LINKED ISSUE #4088: Fix English language infinite loop between TRUNCATE and DELETE pages
On the [TRUNCATE][truncate] page we say (roughly) that

> It's faster to delete smaller tables using a DELETE without a WHERE clause than to use TRUNCATE

But we don't say what smaller is!

Then, on the [DELETE][delete] page we say that 

> The TRUNCATE statement removes all rows from a table by dropping the table and recreating a new table ... this is much more performant than deleting rows ...

Scope of work here is roughly:

1. Clarify what is meant by "smaller", and thus when to prefer DELETE minus WHERE vs TRUNCATE
2. Update language to break the "infinite loop" between these pages' instructions :-)
3. Othher work as needed to achieve the above

[truncate]: https://www.cockroachlabs.com/docs/v2.1/truncate.html
[delete]: https://www.cockroachlabs.com/docs/v2.1/delete.html

rmloveland:
Fixes #4088.

Summary of changes:

- Update `DELETE` page to remove unnecessary `TRUNCATE` example, and add
  a note saying to prefer `TRUNCATE` for tables >= 1000 rows.

- Update `TRUNCATE` page to expand the initial description of its
  behavior slightly, and to add a note saying to prefer `DELETE` for
  tables with < 1000 rows.

- Various small copy edits: typo fix; link update; whitespace.

### DIFF
--- a/v2.2/delete.md
+++ b/v2.2/delete.md
@@ -10,7 +10,6 @@ The `DELETE` [statement](sql-statements.html) deletes rows from a table.
 
 {{site.data.alerts.callout_info}}To delete columns, see <a href="drop-column.html"><code>DROP COLUMN</code></a>.{{site.data.alerts.end}}
 
-
 ## Required privileges
 
 The user must have the `DELETE` and `SELECT` [privileges](privileges.html) on the table.
@@ -54,12 +53,11 @@ Deleting a row does not immediately free up the disk space. This is
 due to the fact that CockroachDB retains [the ability to query tables
 historically](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).
 
-If disk usage is a concern, there are two potential solutions. The
-first is to [reduce the time-to-live](configure-replication-zones.html)
-(TTL) for the zone, which will cause garbage collection to clean up
-deleted rows more frequently. Second, unlike `DELETE`,
-[truncate](truncate.html) immediately deletes the entire table, so
-consider if you can use `TRUNCATE` instead.
+If disk usage is a concern, the solution is to
+[reduce the time-to-live](configure-replication-zones.html) (TTL) for
+the zone by setting `gc.ttlseconds` to a lower value, which will cause
+garbage collection to clean up deleted objects (rows, tables) more
+frequently.
 
 ## Select performance on deleted rows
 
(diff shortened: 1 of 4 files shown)


rmloveland:
Direct links for easier reference (these are for the stable versions, but the "dev" versions have the same changes):

- [TRUNCATE](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/cdab9702c530492e093f177ffe39e0c7b68c325a/stable/truncate.html)
- [DELETE](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/cdab9702c530492e093f177ffe39e0c7b68c325a/stable/delete.html)

rmloveland:
> Remove double space!

Fixed, thanks @lhirata

v2.2/delete.md
```diff
@@ -58,7 +57,7 @@ If disk usage is a concern, there are two potential solutions. The
 first is to [reduce the time-to-live](configure-replication-zones.html)
 (TTL) for the zone, which will cause garbage collection to clean up
 deleted rows more frequently. Second, unlike `DELETE`,
-[truncate](truncate.html) immediately deletes the entire table, so
```
dt: Huh, I think maybe GH ate my comment. Maybe @vivekmenezes can chime in here, but AFAIK, TRUNCATE doesn't delete the on-disk data any sooner -- it just avoids _writing_ deletes for every row.

vivekmenezes: I think we want to says.

1. Normally, TRUNCATE should be used to delete all the table data, but TRUNCATE is implemented as a schema change which is not transactional. Therefore some data added within a short interval after the TRUNCATE might get deleted by the truncate.
2. Alternatively, DELETE can be use for transactional deletion of table data, but care must be taken because it will not work on very large tables because of the large transactions are not supported limitation.

dt: My point is that if you want the disk space back, _neither_ TRUNCATE nor DELETE are going to give it to you, but the current text seems to imply that TRUNCATE will which is what I was objecting to. To be fair, it is _better_ than `DELETE`s in terms of disk space since it doesn't write a tombstone on every row, but it still isn't going to free it up any sooner than the GC time.

rmloveland: Thanks David and Vivek for the info.  Sounds like neither statement is useful for reclaiming disk, so we need to remove that bit.

Just pushed a change (also copied below) to reflect that neither statement will give your disk space back (soon), and you must update your GC TTL settings.  Let me know what you think.

> If disk usage is a concern, the solution is to [reduce the time-to-live](configure-replication-zones.html) (TTL) for the zone by setting `gc.ttlseconds` to a lower value, which will cause garbage collection to clean up deleted objects (rows, tables) more frequently.

rmloveland:
Just spoke with David and Vivek IRL - merging this.

jseldess:
@rmloveland, while you're here, would you please mention on the drop database, drop table, and truncate pages that you can view the progress of these operations via `show jobs` or the jobs page of the admin ui? Issue: https://github.com/cockroachdb/docs/issues/4103

jseldess:
Whoops, @rmloveland, I just realized my last comment was post-merge. Could you open a follow-up?

PR #4096, opened 2018-11-26
DROP TABLE does not show up in SHOW JOBS

### LINKED ISSUE #3561: Document that DROP TABLE does not show up in SHOW JOBS
Related to [cockroach#27203](https://github.com/cockroachdb/cockroach/issues/27203).

A `DROP` or `TRUNCATE` will drop the table immediately; but data only gets deleted after the GC TTL interval.

Even though there is a background job, it does not appear in the output of `SHOW JOBS` (as of versions <= 2.1).

lhirata:
Closes #3561.

### DIFF
--- a/v2.2/show-jobs.md
+++ b/v2.2/show-jobs.md
@@ -35,7 +35,7 @@ The following fields are returned for each job:
 Field | Description
 ------|------------
 `job_id` | A unique ID to identify each job. This value is used if you want to control jobs (i.e., [pause](pause-job.html), [resume](resume-job.html), or [cancel](cancel-job.html) it).
-`job_type` | The type of job. Possible values: `SCHEMA CHANGE`, [`BACKUP`](backup.html), [`RESTORE`](restore.html), or [`IMPORT`](import.html).<br><br>Note: Although there is a background job for [`DROP`](drop-table.html) or [`TRUNCATE`](truncate.html), it does not appear in the output of `SHOW JOBS`.
+`job_type` | The type of job. Possible values: `SCHEMA CHANGE`, [`BACKUP`](backup.html), [`RESTORE`](restore.html), [`IMPORT`](import.html), [`DROP`](drop-table.html), or [`TRUNCATE`](truncate.html).
 `description` | The command that started the job.
 `user_name` | The user who started the job.
 `status` | The job's current state. Possible values: `pending`, `running`, `paused`, `failed`, `succeeded`, or `canceled`.
--- a/v2.2/show-jobs.md
+++ b/v2.2/show-jobs.md
@@ -35,7 +35,7 @@ The following fields are returned for each job:
 Field | Description
 ------|------------
 `job_id` | A unique ID to identify each job. This value is used if you want to control jobs (i.e., [pause](pause-job.html), [resume](resume-job.html), or [cancel](cancel-job.html) it).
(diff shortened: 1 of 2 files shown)


vivekmenezes:
FYI this limitation is going away for 2.2. It has already been fixed but not present in 2.1. I'm fine with this change.

knz:
ok thanks @lhirata do you already have an issue filed to update the doc for 2.2?

lhirata:
Thanks @knz and @vivekmenezes. I created an issue to track the v2.2 change but I think we can avoid having to update the docs later if I just don't add the new note in the first place. I'll remove it from the v2.2 docs and leave it in v2.1.

jseldess:
@lhirata, would you mind factoring https://github.com/cockroachdb/docs/issues/4103 into this PR? Basically, as of 2.2, DROP DATABASE/TABLE and TRUNCATE do show up in the SHOW JOBS output and in the Jobs page of the admin UI. We missed this because there wasn't a release note text.

Looks like the relevant PR also added more details to the show jobs output as well. PTAL.

lhirata:
@jseldess / @vivekmenezes - added the new `running_status` column to the v2.2 SHOW JOBS doc. Can you take another look?

v2.2/show-jobs.md
```diff
@@ -35,10 +35,12 @@ The following fields are returned for each job:
 Field | Description
 ------|------------
 `job_id` | A unique ID to identify each job. This value is used if you want to control jobs (i.e., [pause](pause-job.html), [resume](resume-job.html), or [cancel](cancel-job.html) it).
-`job_type` | The type of job. Possible values: `SCHEMA CHANGE`, [`BACKUP`](backup.html), [`RESTORE`](restore.html), or [`IMPORT`](import.html).
+`job_type` | The type of job. Possible values: `SCHEMA CHANGE`, [`BACKUP`](backup.html), [`RESTORE`](restore.html), [`IMPORT`](import.html), [`DROP TABLE`](drop-table.html), [`DROP DATABASE`](drop-database.html), or [`TRUNCATE`](truncate.html).
```
jseldess: Remove the drop and truncate values. They show up as schema change.

v2.2/show-jobs.md
```diff
@@ -35,10 +35,12 @@ The following fields are returned for each job:
...
 `description` | The command that started the job.
 `user_name` | The user who started the job.
 `status` | The job's current state. Possible values: `pending`, `running`, `paused`, `failed`, `succeeded`, or `canceled`.
+`running_status` | The job's detailed running status, which provides visibility into the progress of the dropping or truncating of tables (i.e., [`DROP TABLE`](drop-table.html), [`DROP DATABASE`](drop-database.html), or [`TRUNCATE`](truncate.html)). For dropping or truncating jobs, the detailed running status is determined by the status of the table at the earliest stage of the schema change. The job is completed when the GC TTL expires and both the table
```
jseldess: Does this status apply for any other types of jobs?

lhirata: Tried a BACKUP and it also has `NULL` in the `running-status` column. Seems like it is only for `DROP` and `TRUNCATE`.

vivekmenezes: yes only applies for DROP and TRUNCATE

vivekmenezes:
LGTM

Issue #3549, opened 2018-08-15
Overview of security and access control


mberhault:
We have basic mechanical docs on certs and users/privileges, but we're missing a clear overview of CockroachDB's security offerings. This should include:
- [ ] Encryption (TLS and At Rest)
- [ ] Authentication (TLS and passwords)
- [ ] Access Control (users, privileges, and roles)
      - See https://forum.cockroachlabs.com/t/admin-ui-authentication/661/4?u=bdarnell

jseldess:
Thanks, @mberhault. I'll see if I can get this written up for the 2.1 release, given that this is a regular source of confusion. @tim-o, do you agree?

tim-o:
Yes, definitely agree. Connecting to a cluster is a major point of confusion for customers.

jseldess:
Not sure there'll be bandwidth to handle this for 2.1. Will have to wait and see.

jseldess:
In addition to the "how to connect" page, we should write a page summarizing all CockroachDB's security options:

- Encryption (TLS and At Rest)
- Authentication (TLS and passwords)
- Access Control

jseldess:
From @tim-o:

On this doc:
https://www.cockroachlabs.com/docs/stable/create-user.html

This snippet:
```
Users with passwords can authenticate their access by entering their password at the command prompt instead of using their client certificate and key.

If we cannot find client certificate and key files matching the user, we fall back on password authentication.

cockroach sql --user=jpointsman
```

Is inaccurate. Should read something like:
```
If you wish users to authenticate with a password, you can, though users will still need to provide a client certificate and key in order to connect to CRDB.
```

See: https://github.com/cockroachdb/cockroach/issues/22537#issuecomment-364603744

mberhault:
The doc had it right: if a user has a client certificate and key matching their username in the certs-dir (eg: `client.jpointsman.crt` and `client.jpointsman.key`), it will perform certificate authentication. If not present, CockroachDB will fall back on password authentication.

A user **cannot** use password authentication when passing a client cert and key.

What prompted the proposed correction?

jseldess:
You're right, @mberhault. I was trying to consolidate related issues, but that issue isn't correct. I do think there's room to improve how we explain this as part of this project, however.

jseldess:
@lhirata, @rolandcrosby, here's the security docs project for 2.2. Please feel free to clarify and/or expand as needed.

Amruta-Ranade:
[...]

PR #4120, opened 2018-11-30
Security overview docs 

Amruta-Ranade:
Addresses #3549 

**Summary of changes:**
- Added a security overview doc
- Added a "How digital certificates work with CockroachDB" doc
- Added a Security section to the sidebar nav

**Proposed info arch. for the security docs:**
- The overview page gives a starting point for all security-related docs. 
- The "how digital certs work" page gives a conceptual overview of how digital certs work, how CRDB uses digital certs, and the four usage scenarios
- The `cockroach cert`, `openssl`, and `custom CA` docs give tactical details of how to use the certs.

**Structural review request:** @jseldess @rolandcrosby 
Please note these drafts are only meant for structural reviews for now. They need substantial edits. For example, the "how digital certs work" doc looks like a wall of text that I want to break up using diagrams and gifs. But I need initial feedback to know if I am on the right track before I put in additional work.

**(Note to self) Pending items:**
- After the custom CA PR is merged, cross-link relevant sections
- Technical/peer/editorial reviews
- Replicate changes to 2.1

Also cc @kannanlakshmi JFYI


Amruta-Ranade:
@bdarnell  @mberhault I might have gone into unnecessary details a bit in the authentication.md doc. Can you take a look please? Also the security-overview.md doc.

Amruta-Ranade:
___
*[v2.2/authentication.md, line 117 at r2](https://reviewable.io/reviews/cockroachdb/docs/4120#-LTReYwl01QMQ_sIuNof:-LTStZdJAOjO1bEqqrfk:b-n65vpd) ([raw file](https://github.com/cockroachdb/docs/blob/62cf59fabb33fe7dd02217e8f7d6df730d3ee26e/v2.2/authentication.md#L117)):*

"all" is a good way to make sure nothing is missed. Technically, we only need the names and addresses used to reach the node.
Some things users should definitely include:
- the values specified in `--listen-host` and `--advertise-host`, or the node hostname and fully qualified hostname if not specified
- any host addresses/names used to reach a specific node
- any load balancer addresses/names or DNS aliases through which the node could be reached
- localhost and local address if connections are made through the loopback device on the same host

As some point, this list becomes "all addresses and names", but most people rarely need everything. An example with a load balancer (or DNS alias) could simply be: given a load balancer in front of all nodes in the region, the LB address/name must be in the certificate.

</blockquote></details>

Oh, this is excellent input! Thank you :)

jseldess:
___
*[v2.2/authentication.md, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/4120#-LTZhHiqEzN35Dam7Uu8:-LTbEp068eDAOHxfPm8d:b2i7xaf) ([raw file](https://github.com/cockroachdb/docs/blob/bc33d4539e419ea4b9b3d5247bcd7c950d6e6ec1/v2.2/authentication.md#L7)):*

This is good material, but I think it probably belongs in a separate document ("security background"?) instead of the "authentication" document (which I would start with the section "using digital certificates with cockroachdb"). 
</blockquote></details>

I know we've gone back and forth on this, Amruta, but I've been waiting to look at how this feels on the page until now, and I agree with Ben. Most user will want to get right to the heart of the authentication in CockroachDB rather than background material. 

One idea is to tweak the "Security Overview" page to accommodate both an overview of CockroachDB security features and background on security. So the intro could be a concise statement of what's on the page and then there could be 3 h2s: 
- `Overview of security in CockroachDB`
- `Security features in CockroachDB`
- `General background on security` (you put all of the general conceptual content here)

Thoughts? Would you be willing to give this a try?

Amruta-Ranade:
___
*[v2.2/authentication.md, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/4120#-LTZhHiqEzN35Dam7Uu8:-LTcGt6IBacpxZAc1QFl:bdpy64b) ([raw file](https://github.com/cockroachdb/docs/blob/bc33d4539e419ea4b9b3d5247bcd7c950d6e6ec1/v2.2/authentication.md#L7)):*

I know we've gone back and forth on this, Amruta, but I've been waiting to look at how this feels on the page until now, and I agree with Ben. Most user will want to get right to the heart of the authentication in CockroachDB rather than background material. 

One idea is to tweak the "Security Overview" page to accommodate both an overview of CockroachDB security features and background on security. So the intro could be a concise statement of what's on the page and then there could be 3 h2s: 
- `Overview of security in CockroachDB`
- `Security features in CockroachDB`
- `General background on security` (you put all of the general conceptual content here)

Thoughts? Would you be willing to give this a try? 
</blockquote></details>

Sure! My only concern is that the third section would talk about conceptual info about authentication alone, and not all security features. I can't intuit if that would be okay - I can give it a try and see how that turns out.  Is that okay?

Amruta-Ranade:
@bdarnell @jseldess I have restructured the Overview and Authentication docs as discussed. I see the following issues with the new structure:

- The overview doc is authentication info-heavy. I can't really title it "How security features work" as I have done now, or "General background information about security" as per Jesse's suggestion, because the information pertains only to authentication and not to the other security features.
- Now we have three places where we give information about authentication: The overview doc, the authentication doc, and the create-security-certificates doc set. IMO, this is an ineffective information arch, where we make people click to different pages. 

My vote is for the former structure. As we discussed in our previous 1:1, it's okay to have a conceptual-heavy doc and then a how-to doc set (i.e., the create-security-certificates doc set). I like the one-stop flow of information that the earlier structure presented. I can also add a note saying if you know this conceptual info, skip to the "Using certs with CRDB" section.

Let me know what you think :)

Amruta-Ranade:
@jseldess Ready for your final review!
@mberhault @rolandcrosby Can you verify your requested changes and LGTM the PR if everything looks okay?

rolandcrosby:
@jseldess thanks for LGTM'ing, I concur.

PR #4123, opened 2018-11-30
Add secure version of performance tuning tutorial

### LINKED ISSUE #4119: Secure version of Performance Tuning tutorial
This is a gap, and a customer already asked for it.

jseldess:
- Add a secure version of the python script.
- Split the tutorial into 2 pages, secure and insecure.
- On each page, use front matter variables to auto-populate
  the correct security flag for cockroach client commands
  and the correct python script name.

Fixes #4119.


robert-s-lee:
We have insecure insurrections in other places.  Perhaps there should be a separate discussion on removing insecure or what the plan should be.  Would prefer to leave insecure for now.

PR #4127, opened 2018-12-03
Add Rust secure examples

couchand:
This adds the code for secure examples in Rust, in addition to making a few tweaks to the examples.  Namely, an import that had a spurious `self::` now refers directly to the imported `postgres` crate, and the function trait bound on the `execute_txn` method has been changed from `FnMut` to `Fn` to better reflect the semantics of this function.


sploiselle:
@couchand PTAL once TC puts up the AWS preview link

couchand:
Ok I updated it with a comment referring to the impending API changes.  Once that all settles I'll try to remember to come back and fix this up.

couchand:
TFTR @nvanbenschoten 

@sploiselle it's ready to go from my end, thanks!

Issue #3315, opened 2018-07-02
Transaction retry guidance for SQLAlchemy


rmloveland:
Our [client driver tutorials](https://www.cockroachlabs.com/docs/dev/build-a-go-app-with-cockroachdb.html) include code and guidance for handling client-side transaction retries. We should expand the corresponding "build an app" tutorials for the top ORMs to provide such guidance as well:

- [ ] [Hibernate](https://www.cockroachlabs.com/docs/dev/build-a-java-app-with-cockroachdb-hibernate.html)
- [x] [SQLAlchemy](https://www.cockroachlabs.com/docs/dev/build-a-python-app-with-cockroachdb-sqlalchemy.html)
- Others?

For details, see https://github.com/cockroachdb/cockroach/issues/26966

In addition, it'd be excellent to:
- [ ] Write more robust, stand-alone tutorials for Hibernate and SQLAlchemy, leveraging `movr` as the example application. 
- [ ] Cover transaction retries in the [production checklist](https://www.cockroachlabs.com/docs/dev/recommended-production-settings.html). See https://github.com/cockroachdb/docs/issues/3266. We may eventually want to add a `Development Checklist` or something similar.

jseldess:
Another forum post looking for this guidance: https://forum.cockroachlabs.com/t/error-transaction-is-too-large-to-complete-try-splitting-into-pieces/1890

rmloveland:
See also: #4008

awoods187:
See https://cockroachlabs.slack.com/archives/CBBGUKTA8/p1543505931005900

PR #4142, opened 2018-12-10
Update SQLAlchemy example(s) to use txn retries

rmloveland:
Addresses #3315.

Summary of changes:

- Update SQLAlchemy example code to use `run_transaction` function to
  perform txn retries automagically.

- Also addresses the issue that the previous code sample could not be
  run more than once.

- Add new 'Best Practices' section that describes what to do (and not
  do) when driving CockroachDB from SQLAlchemy, including:

  - Noting that one should not munge txn state inside `run_transaction`.

  - Showing some example code to address "transaction too large errors"
    by breaking updates into chunks.

  - Recommendation to use IMPORT for bulk work, not ORMs.

  - Recommendation to prefer query builder over ORM, if possible.


rmloveland:
Direct link to updated pagefor easier reference:

http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/ac178662ffc950a8c623ad2727e11eb78bf18575/stable/build-a-python-app-with-cockroachdb-sqlalchemy.html

_includes/v2.2/app/insecure/sqlalchemy-basic-sample.py
```diff
@@ -1,35 +1,108 @@
...
-# for the engine URL indicates that we are connecting to CockroachDB.
-engine = create_engine('cockroachdb://maxroach@localhost:26257/bank',
-                       connect_args = {
-                           'sslmode' : 'disable'
-                       })
-Session = sessionmaker(bind=engine)
+
+# Create an engine to communicate with the database. The
+# "cockroachdb://" prefix for the engine URL indicates that we are
+# connecting to CockroachDB using the 'cockroachdb-python' dialect.
```
bdarnell: I think I'd just call this `cockroachdb`, not `cockroachdb-python`. The repo is named `cockroachdb-python`, but users won't generally see the repo. They'll use `pip` or `requirements.txt` to install the package, and in that context it's just called `cockroachdb`.

_includes/v2.2/app/insecure/sqlalchemy-basic-sample.py
```diff
@@ -1,35 +1,108 @@
...
+for elem in res:
+    val = elem[0]
+    if val not in seen:
+        seen[val] = 1
+
+
+def create_random_accounts(sess, seen, n):
+    """Create N new accounts with random IDs and random account balances.
+
+    Note that we do some work to avoid account ID collisions by
```
bdarnell: This is an unusual pattern. Normally when using random IDs you just assume that you're using a large enough space that you don't bother checking for duplicates. I think we should remove the duplicate checking here and have the database generate conflict-free IDs (with unique_rowid, sequences, or uuids).

_includes/v2.2/app/insecure/sqlalchemy-basic-sample.py
```diff
@@ -1,35 +1,108 @@
...
+    inspecting the contents of the SEEN dictionary.
+    """
+    new_accounts = []
+    elems = iter(range(n))
+    for i in elems:
+        maybe_id = floor(random.random()*1_000_000)
```
bdarnell: This syntax is only supported in Python 3.6; we should call out the required version somewhere (or I think if we removed the underscores it could work in more versions).

_includes/v2.2/app/sqlalchemy-basic-sample.py
```diff
@@ -1,38 +1,111 @@
...
+
+# Create an engine to communicate with the database. The
+# "cockroachdb://" prefix for the engine URL indicates that we are
+# connecting to CockroachDB using the 'cockroachdb-python' dialect.
+# For more information, see
+# https://github.com/cockroachdb/cockroachdb-python.
+
+engine = create_engine(
+    'cockroachdb://maxroach@localhost:26257/bank',
+    connect_args={
```
bdarnell: Instead of having two files that differ only in this block, I'd suggest just having one with some sort of `if secure:` check around the create_engine call.

_includes/v2.2/app/sqlalchemy-large-txns.py
```diff
@@ -0,0 +1,59 @@
+from sqlalchemy import create_engine, Column, Float, Integer
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import sessionmaker
+from cockroachdb.sqlalchemy import run_transaction
+from random import random
+
+Base = declarative_base()
```
bdarnell: Why are you using the declarative interface here but not in the others? I would expect that if anything you'd be more likely to use the ORM in your main application but the lower-level (and probably more efficient) interfaces in your bulk loader.

v2.2/build-a-python-app-with-cockroachdb-sqlalchemy.md
```diff
@@ -54,17 +48,26 @@ Create a certificate and key for the `maxroach` user by running the following co
 $ cockroach cert create-client maxroach --certs-dir=certs --ca-key=my-safe-directory/ca.key
 ~~~
 
-## Step 3. Run the Python code
+## Step 4. Run the Python code
+
+The code below uses the [SQLAlchemy ORM](https://docs.sqlalchemy.org/en/latest/) to map Python objects and methods to SQL operations.
```
bdarnell: Just call it SQLAlchemy, not the SQLAlchemy ORM. We're not using the ORM half of SQLAlchemy here.

rmloveland: Fixed.

_includes/v2.2/app/sqlalchemy-basic-sample.py
rmloveland: Fixed by adding some conditional code to this sample and then deleting the "insecure" example file.

_includes/v2.2/app/insecure/sqlalchemy-basic-sample.py
rmloveland: Fixed by removing the underscores and adding a note that Python 3 is used for the example code.

_includes/v2.2/app/insecure/sqlalchemy-basic-sample.py
rmloveland: Fixed.

rmloveland:
Thank you for the reviews @bdarnell and @knz !

Ben, I have addressed all but two of your comments.  I spent some time trying to rewrite the code to follow your suggestions but I got stuck.  I know in principal that I should use more query building (via `Engine.execute` I think) but I was not able to get something working after about 90 minutes of trying, so I figured I should call for help.

I see a few options:

1. If either of you have a few minutes to patch the example code to address the remaining comments I would be thrilled to have the help.
2. If the deficiencies are not urgent enough that they should distract you from other matters, we can merge the PR as is and have room to make improvements later.  (Sorry if this option was implied by your LGTMs, I wanted to explicitly address your review comments Ben since you took the time to make them.)

What do you think?

bdarnell:
I'm fine with this getting merged with the bulk loading file unchanged. I think it's worth getting rid of the retry loop for duplicates in the account creation step. For demo purposes I think you can just choose random numbers out of a billion and then just delete the logic to check for duplicates. (using UUIDs would be even better but might be a more far-reaching change)

rmloveland:
> I'm fine with this getting merged with the bulk loading file unchanged. 

Cool, leaving it as is for now.

> I think it's worth getting rid of the retry loop for duplicates in the account creation step. For demo purposes I think you can just choose random numbers out of a billion and then just delete the logic to check for duplicates. (using UUIDs would be even better but might be a more far-reaching change)

Fixed in latest update - now the code (1) generates random numbers from a billion and (2) uses a SQL query when it needs to fetch random account IDs (instead of the now-deleted dictionary).  Thanks again for the review and feedback.

_includes/v2.2/app/sqlalchemy-basic-sample.py
```diff
@@ -1,38 +1,108 @@
...
+
+
+run_transaction(sessionmaker(bind=engine),
+                lambda s: create_random_accounts(s, 100))
+
+
+# Helper for getting random existing account IDs.
+
+def get_random_account_id():
+    res = engine.execute('SELECT id FROM accounts ORDER BY random() LIMIT 1;')
```
bdarnell: `ORDER BY random()` is expensive. I think you still want a local set for selecting accounts here. Create a set of all the account IDs created in `create_random_accounts`, just don't worry about filling it with any pre-existing ones.

rmloveland: That makes sense.  In the latest update I used a Python set to hold just the most recently created account IDs, and when it's time to shuffle money around we use only those (rather than selecting at random from the entire table).

jseldess:
LGTM!

PR #4153, opened 2018-12-10
Update the Production Checklist

### LINKED ISSUE #46: Add recommendations about file systems
So our product promises to behave nicely assuming the user is storing their files in a properly-behaved filesystem / persistent storage.

However as discussed in this paper: https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-pillai.pdf

operating systems have bugs in how they implement file systems, and sometimes the OS promises that data written by an application is actually written to disk while there is a bug which causes this data to not be saved. Or the OS promises something about the order in which I/O operations issued by the DB are performed on disk, and instead actually does things in a different order. This is Real Bad.

Our docs need to explain what we assume from the underlying storage and make recommendations about which file systems are known to work sufficiently well.

### LINKED ISSUE #3276: Recommend RAID 10 (not 5 or 6) when RAID is used
https://github.com/cockroachdb/cockroach/issues/26529#issuecomment-397636645

RAID can be used in several different modes with widely different performance characteristics. RAID 5 and 6 have poor performance for small writes, which has a severe impact on CockroachDB's performance. We should recommend RAID 10 instead, which has minimal performance impact (RAID 0 and 1 are also acceptable from a performance perspective).

### LINKED ISSUE #3936: Update and expand the production checklist
- [ ] Update our [basic hardware recommendations](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#basic-hardware-recommendations).
    - Instead of a blanket minimum 2GB RAM recommendations, can we provide min RAM guidelines that consider data size and other factors? Probably want general guidelines for either amount of memory per core or amount of memory per data stored. (From https://github.com/cockroachdb/docs/issues/3795) 
    - Add guidance on disk type, disk size, and node/disk ratio (See https://github.com/cockroachdb/docs/issues/4112) 
    - Add guidance on IOPs for disks. Can we provide some guidance or a formula to help users understand how to properly provision their disks? (From https://github.com/cockroachdb/docs/issues/3841)
    - Add recommendations around operating systems. (includes https://github.com/cockroachdb/docs/issues/4035)
    - Add recommendations around file systems. (See https://github.com/cockroachdb/docs/issues/46)
    - Add guidance around `nobarrier`. (See https://github.com/cockroachdb/docs/issues/3962)
    - Add guidance around RAID. (See https://github.com/cockroachdb/docs/issues/3276)
- [ ] Update our [cloud-specific hardware recommendations](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#cloud-specific-recommendations)
    - Per cloud, are the instance type recommendations still right? Can we provide better minimum and recommended specs? What are the updated specs for Cockroach Labs internal testing? 
    - For AWS specifically, we should update our hardware recommendations for EC2.  From https://github.com/cockroachdb/docs/issues/3852: `m5` and `m5d` instances use a new hypervisor that resolves the performance problem in https://heapanalytics.com/blog/engineering/clocksource-aws-ec2-vdso. Our old recommendation of `m3.large` is closest to `m5d.large`. However, I don't think we've tested that configuration. Our current main in-house use of EC2 is now on `m5.xlarge` instances with separate attached EBS volumes.
    - Check with Andy W re: performance between AWS and GCE, and if we want to make any recommendations.
- [x] Update [clock sync recommendations](https://www.cockroachlabs.com/docs/dev/recommended-production-settings.html#basic-hardware-recommendations). (From https://github.com/cockroachdb/docs/issues/3856)
    - We should make it clear that you only want one time sync program running on VMs where cockroach is running.
    -  We should still recommend using google or amazon clock service but with whatever time sync program they are already using. So if they're already using ntp, configure ntp to point to the google or amazon time server. Likewise for chrony.
    - As an advanced option, users who'd rather not use google or amazon can use chrony and enable client side leap smearing (unless they're using a clock source that does server-side smearing). There will not be a leap second before June 2019 at the earliest, so this is not high priority yet.
    - For the choice between google and amazon's time servers: Amazon's time server is only available within EC2, so hybrid environments should use google's instead. For reference, redhat has some docs on options for handling leap seconds: https://access.redhat.com/articles/15145

### LINKED ISSUE #3962: Document when to use nobarrier in production
In our TPC-C benchmarking tutorial, we recommend [mounting local ssd and disabling write cache flushing](https://www.cockroachlabs.com/docs/stable/performance-benchmarking-with-tpc-c.html#step-1-create-3-google-cloud-platform-gce-instances) to simulate bare metal deployment. Disabling write cache flushing is the `nobarrier` option on GCE. This option is safe, and boosts write performance, when suitably reliable power backup is in use. 

We should:

- [ ] Add this option to the [Production Checklist](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html), explaining when to consider it and what benefits it provides. Mainly, this is safe when on bare metal with power backup.
- [ ] We should also explicitly explain the dangers of using this without reliable power backup. This is likely the case when running on public clouds. We don't have enough info there to be confident.

### LINKED ISSUE #4112: Recommendations on disk type, disk size, and node/disk ratio
See https://forum.cockroachlabs.com/t/best-practice-for-mounting-disks/2216.

Summary: Better to use 1 larger disk per node than multiple smaller disk per node. We may not always rebalance optimally when nodes have multiple disks. 

Open questions:
- Which is better, SSD or EBS?
- How big should a disk be? See thread below:

> Andy W: my testing isnt complete but i think SSD > EBS generally speaking and larger than your workload by at least ~30% but i'll defer to engineering
>
> Andrei: As long as you don’t run out of disk space, I don’t think you necessarily need to leave much headroom. When doing capacity planning, the data size is the biggest question I’d say, because of our space amplification. And I wouldn’t know what guidance to give on amplification…

lhirata:
Summary of changes: Update the basic and cloud specific hardware recommendations.

Note: Will add the updates to the 2.1 version once the 2.2 is ready to go.

Closes #3936 (main tracking issue).
Closes #4112.
Closes #46.
Closes #3962.
Closes #3276.

### DIFF
--- a/v2.2/recommended-production-settings.md
+++ b/v2.2/recommended-production-settings.md
@@ -25,7 +25,9 @@ Term | Definition
 - Run each node on a separate machine. Since CockroachDB replicates across nodes, running more than one node per machine increases the risk of data loss if a machine fails. Likewise, if a machine has multiple disks or SSDs, run one node with multiple `--store` flags and not one node per disk. For more details about stores, see [Start a Node](start-a-node.html).
 
 - When deploying in a single datacenter:
+
     - To be able to tolerate the failure of any 1 node, use at least 3 nodes with the [`.default` 3-way replication factor](configure-replication-zones.html#view-the-default-replication-zone). In this case, if 1 node fails, each range retains 2 of its 3 replicas, a majority.
+
     - To be able to tolerate 2 simultaneous node failures, use at least 5 nodes and [increase the `.default` replication factor for user data](configure-replication-zones.html#edit-the-default-replication-zone) to 5. The replication factor for [important internal data](configure-replication-zones.html#create-a-replication-zone-for-a-system-range) is 5 by default, so no adjustments are needed for internal data. In this case, if 2 nodes fail at the same time, each range retains 3 of its 5 replicas, a majority.
 
 - When deploying across multiple datacenters in one or more regions:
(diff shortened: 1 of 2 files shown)


lhirata:
@drewdeally - link to view changes: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/207e6d14d9cad3854d521fa6972da56a6e8c85ba/dev/recommended-production-settings.html

lhirata:
Still to do: Drew to help with kernel settings and adding to the Performance Recommendations section.

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
-        {{site.data.alerts.callout_danger}}
-        {% include {{page.version.version}}/known-limitations/system-range-replication.md %}
-        {{site.data.alerts.end}}
+For best performance:
+
+- Use NVMe SSDs over HDDs.
+- To simulate a bare metal deployment where machines are directly connected to one physical disk each, create and mount a local SSD using a SCSI interface.
+- Use more powerful nodes. Adding more CPU is usually more beneficial than adding more RAM.
```
awoods187: I don't think this is true currently. We recommend 16 cpu nodes and adding more of them rather than using higher CPU nodes. We have some work in progress to address this for 2.2 but for now I'd say 16 cpu is the recommendation cc @nvanbenschoten

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+    {{site.data.alerts.callout_danger}}
+    {% include {{page.version.version}}/known-limitations/system-range-replication.md %}
+    {{site.data.alerts.end}}
 
 ### Cloud-specific recommendations
 
 Cockroach Labs recommends the following cloud-specific configurations based on our own internal testing. Before using configurations not recommended here, be sure to test them exhaustively.
 
+In general, we have found CockroachDB to perform better on AWS over GCE.
```
awoods187: We need to be careful about wording--we don't want the takeaway to be dont use GCP.

awoods187: Also It should be GCP not GCE

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+
 #### AWS
 
-- Use `m` (general purpose), `c` (compute-optimized), or `i` (storage-optimized) [instances](https://aws.amazon.com/ec2/instance-types/). For example, Cockroach Labs has used `m3.large` instances (2 vCPUs and 7.5 GiB of RAM per instance) for internal testing.
-- **Do not** use ["burstable" `t2` instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-instances.html), which limit the load on a single core.
-- Use [Provisioned IOPS SSD-backed (io1) EBS volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html#EBSVolumeTypes_piops) or [SSD Instance Store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html).
+- Use `m` (general purpose), `c` (compute-optimized), or `i` (storage-optimized) [instances](https://aws.amazon.com/ec2/instance-types/). For example, Cockroach Labs has used `m5d.xlarge` instances (4 vCPUs and 16 GiB of RAM per instance) for internal testing.
```
awoods187: We don't recommend I instances at all--it was much worse on testing.

awoods187: Also need to specify that the D in `m5d.xlarge` is SSD. Also, I'd use the C series and I'd use `c5d.4xlarge` to show the 16 cpu

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+        {{site.data.alerts.callout_danger}}
+        Do not use ["burstable" `t2` instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-instances.html), which limit the load on a single core.
+        {{site.data.alerts.end}}
+- Use [Provisioned IOPS SSD-backed (io1) EBS volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html#EBSVolumeTypes_piops) or [SSD Instance Store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html). We recommend using SSD.
```
awoods187: Probably need to discuss SSD vs. EBS. We use SSD with no barrier when we can tolerate disk corruption--when we can't we use EBS. EBS also need to have iOPs provisioned which can be very expensive

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+- When mounting SSD, we recommend using one larger disk per node instead of multiple smaller disks per node. CockroachDB may not always rebalance optimally when nodes have multiple disks.
+- We recommend using 16 cores, 32-64 GB memory each.
```
awoods187: be consistent in cores vs. cpus vs. vCPUs

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+    {{site.data.alerts.end}}
 - Use [Premium Storage](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/premium-storage) or local SSD storage with a Linux filesystem such as `ext4` (not the Windows `ntfs` filesystem). Note that [the size of a Premium Storage disk affects its IOPS](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/premium-storage#premium-storage-disk-limits).
 - If you choose local SSD storage, on reboot, the VM can come back with the `ntfs` filesystem. Be sure your automation monitors for this and reformats the disk to the Linux filesystem you chose initially.
-- **Do not** use ["burstable" B-series](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/b-series-burstable) VMs, which limit the load on a single core. Also, Cockroach Labs has experienced data corruption issues on A-series VMs and irregular disk performance on D-series VMs, so we recommend avoiding those as well.
 
 #### Digital Ocean
 
 - Use any [droplets](https://www.digitalocean.com/pricing/) except standard droplets with only 1 GB of RAM, which is below our minimum requirement. All Digital Ocean droplets use SSD storage.
 
 #### GCE
```
awoods187: GCP

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
 
-- Use `n1-standard` or `n1-highcpu` [predefined VMs](https://cloud.google.com/compute/pricing#predefined_machine_types), or [custom VMs](https://cloud.google.com/compute/pricing#custommachinetypepricing). For example, Cockroach Labs has used custom VMs (8 vCPUs and 16 GiB of RAM per VM) for internal testing.
-- **Do not** use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on a single core.
+- Use `n1-standard` or `n1-highcpu` [predefined VMs](https://cloud.google.com/compute/pricing#predefined_machine_types), or [custom VMs](https://cloud.google.com/compute/pricing#custommachinetypepricing). For example, Cockroach Labs has used custom VMs (8 vCPUs and 16 GiB of RAM per VM) for internal testing.
```
awoods187: We use `n1-standard-16` for performance testing

awoods187: Also worth noting that we found some benefits in using the Skylake series

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,89 @@ Term | Definition
...
+- Use [zone configs](configure-replication-zones.html) to increase the replication factor from 3 (the default) to 5 (across at least 5 nodes). This is especially recommended if you are using local disks with no RAID protection rather than a cloud providers' network-attached disks that are often replicated underneath the covers, because local disks have a greater risk of failure. You can do this for the [entire cluster](configure-replication-zones.html#edit-the-default-replication-zone) or for specific [databases](configure-replication-zones.html#create-a-replication-zone-for-a-database), [tables](configure-replication-zones.html#create-a-replication-zone-for-a-table), or [rows](configure-replication-zones.html#create-a-replication-zone-for-a-table-or-secondary-index-partition) (enterprise-only).
+    {{site.data.alerts.callout_danger}}
+    {% include {{page.version.version}}/known-limitations/system-range-replication.md %}
+    {{site.data.alerts.end}}
 
 ### Cloud-specific recommendations
 
 Cockroach Labs recommends the following cloud-specific configurations based on our own internal testing. Before using configurations not recommended here, be sure to test them exhaustively.
```
awoods187: We may want to link to the 2018 cloud computing report here

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,113 @@ Term | Definition
...
+    Do not use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on a single core.
+    {{site.data.alerts.end}}
 
-- Use `n1-standard` or `n1-highcpu` [predefined VMs](https://cloud.google.com/compute/pricing#predefined_machine_types), or [custom VMs](https://cloud.google.com/compute/pricing#custommachinetypepricing). For example, Cockroach Labs has used custom VMs (8 vCPUs and 16 GiB of RAM per VM) for internal testing.
-- **Do not** use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on a single core.
 - Use [Local SSDs](https://cloud.google.com/compute/docs/disks/#localssds) or [SSD persistent disks](https://cloud.google.com/compute/docs/disks/#pdspecs). Note that [the IOPS of SSD persistent disks depends both on the disk size and number of CPUs on the machine](https://cloud.google.com/compute/docs/disks/performance#optimizessdperformance).
+- `nobarrier` can be used with SSDs, but only if it has battery-backed write cache because, while this is faster, data can be corrupted in the event of a crash.
```
nvanbenschoten: "without one data can be corrupted ..."

Or something like that. We just need to make it clear that the corruption is only possible with this setting and without a battery-backed write cache.

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +39,113 @@ Term | Definition
...
 ### Basic hardware recommendations
 
-- Nodes should have sufficient CPU, RAM, network, and storage capacity to handle your workload. It's important to test and tune your hardware setup before deploying to production.
+Nodes should have sufficient CPU, RAM, network, and storage capacity to handle your workload. It's important to test and tune your hardware setup before deploying to production.
+
+#### CPU, memory, and storage recommendations
+
+- At a bare minimum, each node should have **2 GB of RAM and one entire core**.
+
+    More data, complex workloads, higher concurrency, and faster performance require additional resources; as a general rule of thumb, increase the number of cores and additional memory to match with a ratio of 4-8 GB of RAM per core, never exceeding a maximum of 16 GB. On larger socket systems, NUMA should be enabled.
```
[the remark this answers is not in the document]
drewdeally: Good point, as it should read ' for hypervisor configurations which have more than one socket, Numa best practices should be followed for optimal memory performance.'  However, this can be more complex as if 2 numa nodes appear in an OS, which is possible on large containers on cloud platforms, we would need to test it.  What would be the best practice for 2 numa nodes in one container, 1 CR or 2 CR numa bind to each node.   This would need to be tested.

v2.2/recommended-production-settings.md
bdarnell: But up to 16 cpus, adding more cpu is better than adding more ram. That's why we recommend `c` instances instead of `m` instances on aws. I'd recommend a range of 4-16 cpus per node.

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
-    - Use larger/more powerful nodes. Adding more CPU is usually more beneficial than adding more RAM.
+- The recommended Linux filesystem is [ext4](https://ext4.wiki.kernel.org/index.php/Main_Page).
+- Avoid using shared storage such as NFS, CIFS and CEPH storage.
+- The optimal configuration for striping more than one device is [RAID 10](https://en.wikipedia.org/wiki/Nested_RAID_levels#RAID_10_(RAID_1+0)). RAID 0 and 1 are also acceptable from a performance perspective.
+
+#### Performance recommendations
+
+For best performance:
+
+- For storage, SSD or NVMe devices will yield the best performance results. The usual volume size should be 300-500 GB.
```
Amruta-Ranade: nit: "Usual volume size should be ..." > "recommended volume size is ..."

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
+
+    Monitor IOPS for higher service times, if they exceed 1-5 ms, you will need to add more devices or expand the cluster to reduce the disk latency. To monitor IOPS, use tools such as `iostat` (part of `sysstat`).
```
Amruta-Ranade: nit: "Monitor IOPS for higher service times, if they exceed..." > "Monitor IOPS for higher service times. If the IOPS exceed ..."

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
+
+- The ideal configuration is 4-16 CPUs, 16-64 GB memory nodes (2-4 GB of memory per CPU).
+
+    To add more processing power (up to 16 CPUs), add more CPUs is better than adding more RAM. Otherwise, add more nodes rather than using higher CPU per node; higher CPUs will have NUMA implications. Our internal testing results indicate this is the sweet spot for OLTP workloads. It is a best practice to use uniform nodes so SQL performance is consistent.
```
Amruta-Ranade: nit: add more CPUs > adding more CPUs

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
+
+- To calculate IOPS, use [sysbench](https://github.com/akopytov/sysbench). If IOPS decreases, add more nodes to your cluster to add IOPS.
```
Amruta-Ranade: decreases > decrease
add IOPS > increase IOPS

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
-        {{site.data.alerts.callout_danger}}
-        {% include {{page.version.version}}/known-limitations/system-range-replication.md %}
-        {{site.data.alerts.end}}
+    This is especially recommended if you are using local disks with no RAID protection rather than a cloud providers' network-attached disks that are often replicated underneath the covers, because local disks have a greater risk of failure. You can do this for the [entire cluster](configure-replication-zones.html#edit-the-default-replication-zone) or for specific [databases](configure-replication-zones.html#create-a-replication-zone-for-a-database), [tables](configure-replication-zones.html#create-a-replication-zone-for-a-table), or [rows](configure-replication-zones.html#create-a-replication-zone-for-a-table-or-secondary-index-partition) (enterprise-only).
```
Amruta-Ranade: nit: underneath the covers > under the hood

v2.2/recommended-production-settings.md
```diff
@@ -39,54 +41,111 @@ Term | Definition
...
+    Do not use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on CPU resources.
+    {{site.data.alerts.end}}
 
-- Use `n1-standard` or `n1-highcpu` [predefined VMs](https://cloud.google.com/compute/pricing#predefined_machine_types), or [custom VMs](https://cloud.google.com/compute/pricing#custommachinetypepricing). For example, Cockroach Labs has used custom VMs (8 vCPUs and 16 GiB of RAM per VM) for internal testing.
-- **Do not** use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on a single core.
 - Use [Local SSDs](https://cloud.google.com/compute/docs/disks/#localssds) or [SSD persistent disks](https://cloud.google.com/compute/docs/disks/#pdspecs). Note that [the IOPS of SSD persistent disks depends both on the disk size and number of CPUs on the machine](https://cloud.google.com/compute/docs/disks/performance#optimizessdperformance).
+- `nobarrier` can be used with SSDs, but only if it has battery-backed write cache. Without one, data can be corrupted in the event of a crash.
 
 ## Security
```
Amruta-Ranade: Should we link to the updated security docs here?

PR #4160, opened 2018-12-12
document leaseholder rebalancing in architecture docs

### LINKED ISSUE #2568: Document leaseholder rebalancing
From https://github.com/cockroachlabs/support/issues/146#issuecomment-369276765:

* The current leaseholder keeps track of how many requests it receives from each locality as an exponentially weighted moving average.
* Periodically (every 10 minutes by default in large clusters, but more frequently in small clusters) each leaseholder considers whether it should rebalance where the ranges replicas are or transfer the lease to another existing replica.
* The leaseholder considers 3 primary inputs: the number of leases on each node, the number of requests from each locality, and the latency between localities.
* If all the nodes are in the same locality, the decision is made entirely on the basis of the number of leases on each node, trying to achieve a balance across all of them. We basically compare the current leaseholder's number of leases and the potential transferees' numbers of leases to the average number of leases on all the nodes in the cluster to determine which are underfull, overfull, or about right.
* If the nodes are in different localities, then we assign a "weight" to each existing replica's locality based on the fraction of requests that came from that locality or from similar localities
   * For example, if a bunch of requests were received by cockroach nodes in the locality "country=us,region=central" and forwarded to the leaseholder for some range, then a replica of that range in locality "country=us,region=west" would get partial credit (50% credit, to be precise, because the first locality tier is shared) for those requests, whereas a replica in "country=aus,region=central" would get no credit. A replica in "country=us,region=central" would get full credit for the requests.
* We combine the ratio of the weights of the source replica to each potential transferee with the latency between the source replica's locality and the potential transferee's locality to come up with an adjustment factor.
* The greater the disparity between "weights" and the higher the latency between localities, the more we'll favor the node from the locality with the larger weight. We then use that adjustment factor to adjust the underfull/overfull targets such that the favored node is more likely to be chosen as the lease location.

sploiselle:
Completes doc of #2568 (not closing issue because still want to incorporate setting into demo)

@a-robinson @benesch PTAL for tech edits, will clean up copy once it looks good

### DIFF
--- a/v2.1/architecture/replication-layer.md
+++ b/v2.1/architecture/replication-layer.md
@@ -78,6 +78,48 @@ This mechanism lets us avoid tracking leases for every range, which eliminates a
 
 Your table's meta and system ranges (detailed in the distribution layer) are treated as normal key-value data, and therefore have leases, as well. However, instead of using epochs, they have an expiration-based lease. These leases simply expire at a particular timestamp (typically a few seconds)––however, as long as the node continues proposing Raft commands, it continues to extend the expiration of the lease. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
 
+#### Leaseholder rebalancing
+
+Because CockroachDB serves reads from a range's leaseholder, it benefits your cluster's performance if the replica closest to the primary geographic source of traffic holds the lease. However, as traffic to your cluster shifts throughout the course of the day, you might want to dynamically shift which nodes hold leases.
+
+{{site.data.alerts.callout_info}}
+
+This feature is also called [Follow-the-Workload](../demo-follow-the-workload.html) in our documentation.
+
+{{site.data.alerts.end}}
+
+Periodically (every 10 minutes by default in large clusters, but more frequently in small clusters), each leaseholder considers whether it should transfer the lease to another replica by considering the following inputs:
+
+- Number of requests from each locality
+- Number of leases on each node
+- Latency between localities
+
+##### Intra-locality
+
+If all the replicas are in the same locality, the decision is made entirely on the basis of the number of leases on each node that contains a replica, trying to achieve a roughly equitable distribution of leases across all of them. This means the distribution isn't perfectly equal; it intentionally tolerates small deviations between nodes to prevent thrashing (i.e. excessive adjustments trying to reach an equilibrium).
+
+##### Inter-locality
+
+If replicas are in different localities, CockroachDB attempts to calculate which replica would make the best leaseholder, i.e. provide the lowest latency.
+
+To enable dynamic leaseholder rebalancing, a range's current leaseholder tracks how many requests it receives from each locality as an exponentially weighted moving average. This calculation results in the locality that has recently requested the range most often being assigned the greatest weight. If another locality then begins requesting the range very frequently, this calculation would shift to assign the second region the greatest weight.
+
+When checking for leaseholder rebalancing opportunities, the leaseholder correlates each requesting locality's weight (i.e. the proportion of recent requests) to the locality of each replica by checking how similar the localities are. For example, if the leaseholder received requests from gateway nodes in locality `country=us,region=central`, CockroachDB would assign the following weights to replicas in the following localities:
+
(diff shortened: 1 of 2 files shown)


sploiselle:
@rmloveland PTAL for copy edits

v2.1/architecture/replication-layer.md
```diff
@@ -78,6 +78,38 @@ This mechanism lets us avoid tracking leases for every range, which eliminates a
...
 
+#### Leaseholder Rebalancing
+
+Because CockroachDB serves reads from a range's leaseholder, it benefits your cluster's performance if the replica closest to the primary geographic source of traffic holds the lease. However, as traffic to your cluster shifts throughout the course of the day, you might want to dynamically shift which nodes hold leases.
+
+To attempt to choose an optimal node as the range's leaseholder, the current leaseholder keeps track of how many requests it receives from each locality as an exponentially weighted moving average, i.e., the more requests that come from a specific locality, the greater the value is, and will transfer the lease if doing so will reduce latency.
```
rmloveland: This is a long one.  Suggestion:  split at "i.e." and update slightly to something like (spitballing)

> This means that the more requests that come from a specific locality, the greater the average is, and the cluster will transfer the lease if doing so will reduce latency.

(Not sure if "the cluster" is transferring the lease but I'm suggesting adding whatever the subject is that's doing the verbing)

v2.1/architecture/replication-layer.md
```diff
@@ -78,6 +78,38 @@ This mechanism lets us avoid tracking leases for every range, which eliminates a
...
+
+Periodically (every 10 minutes by default in large clusters, but more frequently in small clusters), each leaseholder considers whether it should rebalance where the range's replicas are or transfer the lease to another existing replica by considering 3 primary inputs:
+
+	- The number of leases on each node
```
rmloveland: I think the indentation is preventing the bulleted list from rendering properly (I'm looking at http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/44d5ed7ce29ccfbfff49730edeb14628da5f959c/v2.2/architecture/replication-layer.html).

Suggest wording as "considering the following inputs"

v2.1/architecture/replication-layer.md
```diff
@@ -78,6 +78,38 @@ This mechanism lets us avoid tracking leases for every range, which eliminates a
...
+	- The number of requests from each locality
+	- The latency between localities
+
+**Intra-locality**
```
rmloveland: I'm guessing this is formatted as bold because our h4's show up as small words with thin lines that are hard to see, right?  Might be worth filing a docs site issue with Joe (but totes understood if you just need a bold section now, today).  Ah the joys of the semantic markup lifestyle

sploiselle: Yeah, I should just be the change I want to see in the world and take care of this.

v2.1/architecture/replication-layer.md
```diff
@@ -78,6 +78,38 @@ This mechanism lets us avoid tracking leases for every range, which eliminates a
...
+
+If all the replicas are in the same locality, the decision is made entirely on the basis of the number of leases on each node that contains a replica, trying to achieve a roughly equitable distribution of leases across all of them. This means the distribution isn't perfectly equal; it intentionally tolerates small deviations between nodes to prevent thrashing.
```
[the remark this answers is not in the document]
sploiselle: Dude, like this:

image

Also, done.

v2.1/architecture/replication-layer.md
sploiselle: There was a typo here. Thanks.

sploiselle:
@a-robinson I was going through another round of edits on this and realized I needed to re-org some of the info. Could I get you to give it a quick pass-through to make sure I didn't mangle any of the technical details? [Here's a link, if that helps](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/59747019ee006dd781a266f0fd32f857308bd19a/stable/architecture/replication-layer.html#leaseholder-rebalancing).

PR #4163, opened 2018-12-13
Document customizable savepoint names

### LINKED ISSUE #4008: [MH] enable customization of the name of the retry savepoint
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw5Jsp2TrY3it9Yr/recPmUvxxOK9LveyY
PM: @awoods187 
Eng: @bdarnell

rmloveland:
Fixes #4008.

Summary of changes:

- Added a new section *Customizing the savepoint name* to several
  pages (via include), which describes the new `force_savepoint_restart`
  session variable and makes clear that SAVEPOINT is still only used for
  transaction retries, and that we don't support nested
  transactions (yet!).

- Made the "this is not a generalized SAVEPOINT, it's only for retries"
  warning that lives on several pages an include.

- Removed all inline mentions of `cockroach_restart` *except* in actual
  code samples, since it's not part of the statement name.  E.g.,
  everywhere `SAVEPOINT cockroach_restart` is used in prose, it's been
  changed to just `SAVEPOINT`.

- Added the `force_savepoint_restart` session variable to the list of
  session vars.  While there, I discovered some broken table cells (and
  that we had two duplicate lists in different files) so I broke the
  list of session vars out into a single (HTML) include.  This had the
  following benefits:

  - HTML can be validated, allowing me to see and fix existing
    errors (due to error-prone Markdown table pipe syntax on big tables)

  - List of session vars can now be maintained in one place (the table
    has an addition column so that "See with SHOW?" and "Update with
    SET?" are both listed, to facilitate re-use).

### DIFF
--- a/_includes/v2.2/misc/savepoint-limitations.md
+++ b/_includes/v2.2/misc/savepoint-limitations.md
@@ -1,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
+CockroachDB's [`SAVEPOINT`](savepoint.html) implementation does not support nested transactions (i.e., subtransactions).  It is only used to handle [transaction retries](transactions.html#transaction-retries).
+{{site.data.alerts.end}}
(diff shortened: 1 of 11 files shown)


_includes/v2.2/misc/savepoint-limitations.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
+CockroachDB's [`SAVEPOINT`](savepoint.html) implementation does not support nested transactions (a.k.a. subtransactions).  It is only used to handle [transaction retries](transactions.html#transaction-retries).
```
jseldess: nit: `a.k.a. subtransactions` > `i.e., subtransactions`, though I'm not sure we even need this parenthetical.

rmloveland: Thanks for the review Jesse!

Fixed by changing to "i.e.".

I agree that we may not need the parenthetical from a writerly perspective but it's for SEO :-)  -- I'm trying to work in the text "subtransactions" so these pages appear in search results for that term since it seems to be a commonly used variant.

Issue #4128, opened 2018-12-03
Training updates for next classroom session


jseldess:
For an upcoming classroom training (spread over 2 days, 2 3-hour sessions per day), we need to fill some gaps in our "CockroachDB 101" training content:

- [x] Add Geo-partitioning training after [Locality and Replication Zones](https://www.cockroachlabs.com/docs/stable/training/locality-and-replication-zones.html). Can leverage https://www.cockroachlabs.com/docs/stable/performance-tuning.html and https://www.cockroachlabs.com/docs/stable/partitioning.html.
- [x] Update deck for [Locality and Replication Zones](https://www.cockroachlabs.com/docs/stable/training/locality-and-replication-zones.html) training show two-level locality, country/continent as level 2, region as level 2.
- [x] Create k8s training module. Probably repurpose https://www.cockroachlabs.com/docs/stable/orchestrate-a-local-cluster-with-kubernetes-insecure.html. Probably focus on helm.
- [x] Update data import lab and slides to cover IMPORT postgres/mysql.
- [x] Stress testing: Maybe create a module on TPC-C benchmarking, based on https://www.cockroachlabs.com/docs/stable/performance-benchmarking-with-tpc-c.html. 
- [x] Do a run-through of entire process on Windows.
    - What tools to install to do this successfully?
    - Install VirtualBox on Spencer's windows machine and test this out. 
    (Deemed not necessary, since they will have access to linux ec2 vms).
- [x] Break down schedule for 2 days.

We may need some content in terms of "client tools" as well. TBD.

@bdarnell, if you have other thoughts, please add them.

jseldess:
VirtualBox install guidance:

[Installing Cockroachdb in VirtualBox.docx.pdf](https://github.com/cockroachdb/docs/files/2668733/Installing.Cockroachdb.in.VirtualBox.docx.pdf)

bdarnell:
A few more things I noticed as I went through the training:
- The cluster upgrade page should be updated to discuss preserve_downgrade_option. The slides (but not the text) still have some references to upgrading from 1.1 to 2.0, which is no longer relevant.
- The backup/restore page has embedded a trial license, which I assume has expired. We may want a different way of handling this. 
- The order of locality, data import, and sql basics is weird. I think I'd prefer them in the reverse order (or data import, sql basics, then locality, if sql basics refers to data that was loaded in the import step)
[...]

PR #4214, opened 2018-12-27
Training updates

jseldess:
Addresses #4128.  

This PR updates our training content as follows:

- Adds geo-partitioning, Kubernetes, and TPC-C modules.
- Updates locality and replication zones and several other modules.
- Removes training for older versions.


jseldess:
TFTR, @bdarnell. I'll get back to this early in the new year.

bdarnell:
OK, for kubernetes, let's skip helm and just use the statefulset configs. We're not yet able to use helm for things like scaling and upgrades, so I think it's simpler to leave it out and just work with kubectl. 

- Don't link to the performance-optimized yaml file; the training should always use the basic version. Instead, at the end of the page, offer a link to https://www.cockroachlabs.com/docs/stable/kubernetes-performance.html
- Add some cleanup steps before `minikube delete`. They may be using minikube for other things and not want to destroy the whole thing:

  - `kubectl delete -f https://raw.githubusercontent.com/cockroachdb/cockroach/master/cloud/kubernetes/cockroachdb-statefulset.yaml`
  - `kubectl delete job.batch/cluster-init`

Issue #2044, opened 2017-10-18
Allow users to view query plans in the Web UI


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwINAgXA5on9pr2T/recijX3LQyP2EvCF8
PM: @piyush-singh 
Eng partners: @couchand, @vilterp 

PRs:

jseldess:
This has been moved to 2.2.

Amruta-Ranade:
From @celiala 's email:

**Logical Plans: Cluster Settings**

sql.metrics.statement_details.plan_collection.enabled
- Boolean, default is true
- Specifies whether to periodically save a logical plan for each fingerprint

sql.metrics.statement_details.plan_collection.period
- Duration, default is 5 minutes
- The time until a new logical plan is collected (e.g. how often we sample a new logical plan for a particular fingerprint)

Amruta-Ranade:
The logical plan is the recent sample but the latency is an average of all SQL statements forming the fingerprint.

Amruta-Ranade:
Reference read: https://www.depesz.com/2013/04/16/explaining-the-unexplainable/

jseldess:
@Amruta-Ranade, can you give a quick update on this project? I know it's at least partially done.

Amruta-Ranade:
Yes, the UI docs are up-to-date with the current implementation. Since I worked on this issue for an alpha release, I wanted to keep the issue open for tracking purposes (in case we tweak the implementation before the final release).

Amruta-Ranade:
Closing this issue since I am not anticipating any doc changes for logical plans for 19.1. Will open a follow-up issue in case something comes up.

celiala:
Hi @Amruta-Ranade  -

Fyi, the functionality will be the same, but based on customer feedback we're doing one more cosmetic change -- see design prototype below:

![logicalPlan-05-cautionExpand](https://user-images.githubusercontent.com/3051672/54754125-28612980-4bb9-11e9-90d2-bb6434a22ead.jpg)

cc @piyush-singh

Amruta-Ranade:
Yes, but I checked the Statements page doc and do not think there's a doc impact.

celiala:
Ah, agreed that there's no doc impact for the cosmetic change :)

PR #4227, opened 2019-01-09
Added content for Logical plans

Amruta-Ranade:
Addresses #2044

### DIFF
--- a/v2.2/admin-ui-statements-page.md
+++ b/v2.2/admin-ui-statements-page.md
@@ -60,9 +60,13 @@ Latency | The average service latency of the SQL statement (or multiple statemen
 
 ## Statement Details page
 
-The **Statement Details** page displays the details of the time, execution count, retries, rows returned, and latency by phase and by gateway node for the selected statement fingerprint.
+The **Statement Details** page displays the logical plan as well as the details of the time, execution count, retries, rows returned, and latency by phase and by gateway node for the selected statement fingerprint.
 
 <img src="{{ 'images/v2.2/admin_ui_statements_details_page.png' | relative_url }}" alt="CockroachDB Admin UI Statements Page" style="border:1px solid #eee;max-width:100%" />
+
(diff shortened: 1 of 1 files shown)


couchand:
Looks good.  One thing I think we should mention is the fact that the plan is randomly sampled, so it may or may not be representative of any particular execution of the statement.  Not sure the best way to say that.

Amruta-Ranade:
___
*[v2.2/admin-ui-statements-page.md, line 65 at r1](https://reviewable.io/reviews/cockroachdb/docs/4227#-LVsAzXR8vQvx2O92tmd:-LVx8B-2Cbjqe_piT_wp:b1yzey7) ([raw file](https://github.com/cockroachdb/docs/blob/18606f5e4bd6ad63f8ef4a8fbad81f0379c9062e/v2.2/admin-ui-statements-page.md#L65)):*
<details><summary><i>Previously, celiala wrote…</i></summary><blockquote>

Noting this so we don't forget to do this later:

Once we've implemented the final design for the Logical Plan, we'll want to replace this image with a new snapshot of the statements page (that shows a logical plan).

</blockquote></details>

Yup..made a note of it in my to-check list for the final release doc.

Amruta-Ranade:
Oh! I thought the statements having the same fingerprint would also have the same logical plan. Am I mistaken?

Amruta-Ranade:
Is it okay if I merge this PR for now, and we can rework the section once I get back from vacation? Or is this caveat a blocker?

piyush-singh:
For an Alpha, I think this is okay. There are some nuances - sampling plans for example like Andrew mentioned, and indeed, statements having the same fingerprint can have different logical plans (i.e. you can cause a join to use different join mechanisms based on cardinality of inputs, or you can just add an index to a table and the same exact query will now be planned differently). We can worry about those more subtly nuances for the full release docs.

PR #4271, opened 2019-01-16
Update description of a stored value

### LINKED ISSUE #3912: Confusing description of a stored value
At 
https://www.cockroachlabs.com/docs/stable/indexes.html#storing-columns it specifies values from the `FROM` clause, but I believe it should say values returned by the `SELECT` statement. As `FROM` lists the tables which values are retrieved by. This is what I believe it should say: 

---

Storing a column optimizes the performance of queries that retrieve its values but are not used for filtering. This is because indexing values are only useful when they're filtered, but if the index itself is mainly used to retrieve a small subset of the columns from the table, then storing the values along with the index can save a trip to the table to retrieve the values.

rmloveland:
Fixes #3912.

### DIFF
--- a/v2.1/indexes.md
+++ b/v2.1/indexes.md
@@ -66,7 +66,7 @@ When designing indexes, it's important to consider which columns you index and t
 
 ### Storing columns
 
-Storing a column optimizes the performance of queries that retrieve its values (i.e., in the `FROM` clause) but do not filter them. This is because indexing values is only useful when they're filtered, but it's still faster for SQL to retrieve values in the index it's already scanning rather than reaching back to the table itself.
+Storing a column optimizes the performance of queries that retrieve its values but are not used for filtering. This is because indexing values are only useful when they're filtered, but if the index itself is mainly used to retrieve a small subset of the columns from the table, then storing the values along with the index can save a trip to the table to retrieve the values.
 
 However, for SQL to use stored columns, queries must filter another column in the same index.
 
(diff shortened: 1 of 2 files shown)


v2.1/indexes.md
```diff
@@ -66,7 +66,7 @@ When designing indexes, it's important to consider which columns you index and t
 
 ### Storing columns
 
-Storing a column optimizes the performance of queries that retrieve its values (i.e., in the `FROM` clause) but do not filter them. This is because indexing values is only useful when they're filtered, but it's still faster for SQL to retrieve values in the index it's already scanning rather than reaching back to the table itself.
+Storing a column optimizes the performance of queries that retrieve its values but are not used for filtering. This is because indexing values are only useful when they're filtered, but if the index itself is mainly used to retrieve a small subset of the columns from the table, then storing the values along with the index can save a trip to the table to retrieve the values.
```
jseldess: I'm ok with this, but I don't think this rewrite is all that much clearer than what we had before and, grammatically, some of the changes are hard to parse. Would you mind trying to explain this again from scratch, differently, rather than adjusting the existing language?

Also, if we don't want to emphasize the `FROM` clause (and I'm not sure we don't), you should revisit the "store columns" bullet at the beginning of the Best Practices section.

rmloveland: > Would you mind trying to explain this again from scratch, differently, rather than adjusting the existing language?

Thanks for looking at this Jesse.  I'm going to close this PR since I mistakenly thought this might be a quick hit language update, and I don't have time to work on digging into this right now.   Sorry for the noise.

PR #4273, opened 2019-01-17
Update perf best practices with interleave caveats

### LINKED ISSUE #3397: Add Caveats regarding interleaves
We currently recommend interleave tables as a way to speed up performance: https://www.cockroachlabs.com/docs/stable/performance-best-practices-overview.html#interleave-tables

However, an important caveat is that interleaving only improves performance _if all operations on the table specify a single value on the parent column_. Any other query will result in a full table scan, and deletes in particular will become much slower.

This caveat recently caused us to recommend that a customer run a query hourly (`SELECT * FROM t LIMIT 1`, used to refresh table leases) that took upward of 9s to complete. 

This means that any query or mutation run on an interleaved child table with a specified interleave key will be faster, but all other queries and mutations will require scans. So this blurb in the original doc:

_Interleaving tables improves query performance by optimizing the key-value structure of closely related tables, attempting to keep data on the same key-value range if it's likely to be read and written together. This is particularly helpful if the tables are frequently joined on the columns that consist of the interleaving relationship._

Should probably be closer to:

_If you have two tables where all operations (selects, inserts, updates, and deletes) are performed on a single value shared between both tables, then interleaving the tables will lead to significant reductions in latency. However, note that after interleaving, operations that span multiple values or do not specify the interleaved parent ID will be significantly slower than they were prior to interleaving the tables._

rmloveland:
Fixes #3397.

### DIFF
--- a/v2.1/performance-best-practices-overview.md
+++ b/v2.1/performance-best-practices-overview.md
@@ -48,6 +48,13 @@ When a table is created, all columns are stored as a single column family. This 
 
 [Interleaving tables](interleave-in-parent.html) improves query performance by optimizing the key-value structure of closely related tables, attempting to keep data on the same key-value range if it's likely to be read and written together. This is particularly helpful if the tables are frequently joined on the columns that consist of the interleaving relationship.
 
+However, the above is only true for tables where all operations (e.g., [`SELECT`](selection-queries.html) or [`INSERT`](insert.html)) are performed on a single value shared between both tables. The following types of operations may actually become slower after interleaving:
+
+- Operations that span multiple values.
+- Operations that do not specify the interleaved parent ID.
+
+This happens because when data is interleaved, queries that work on the parent table(s) will need to "skip over" the data in interleaved children, which increases the read and write latencies to the parent in proportion to the number of interleaved values.
+
 ## Unique ID best practices
 
 A traditional approach for generating unique IDs is one of the following:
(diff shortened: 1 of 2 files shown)


v2.1/performance-best-practices-overview.md
```diff
@@ -48,6 +48,8 @@ When a table is created, all columns are stored as a single column family. This
 
 [Interleaving tables](interleave-in-parent.html) improves query performance by optimizing the key-value structure of closely related tables, attempting to keep data on the same key-value range if it's likely to be read and written together. This is particularly helpful if the tables are frequently joined on the columns that consist of the interleaving relationship.
 
+However, the above is only true for tables where all operations (i.e., [`SELECT`](selection-queries.html), [`INSERT`](insert.html), [`UPDATE`](update.html), and [`DELETE`](delete.html)) are performed on a single value shared between both tables.  After interleaving, operations that span multiple values, or that do not specify the interleaved parent ID, will be significantly slower than they were prior to interleaving the tables.
```
knz: I would write `(e.g., SELECT or INSERT)` instead of trying to list all of them. As it is now you're forgetting half a dozen already.

rmloveland: Fixing now, thanks.

Follow-up question: Does this information also apply to the following versions?

- 2.0.*
- 2.2

If so I'll add this information there as well.

rmloveland: Ok, updated this PR so this info is added to the 2.1 and 2.2 docs.  Raphael, if you confirm it applies to 2.0, I'll add it there as well.

rmloveland: > What does the word "significantly" mean? Significant according to which standard? Compared to what?

screen shot 2019-01-24 at 11 55 46 am

😂

Thanks for this additional feedback, Raphael.

I have attempted another revision that includes this information, rewritten slightly for clarity.  Please let me know what you think.

PR #4274, opened 2019-01-17
Note that truncated tables must be pre-split again

### LINKED ISSUE #3402: Clarify that truncated tables need to be pre-split again
When a table is truncated, it is essentially re-created in a single new empty range and the old ranges that used to constitute the table are garbage collected. One effect of this is that any pre-splitting that a user has performed on the old version of the table will not carry over to the new version. The new table will need to be pre-split again. We should make this detail clear in the SPLIT AT [documentation](https://www.cockroachlabs.com/docs/v2.0/split-at.html).

rmloveland:
Fixes #3402.

### DIFF
--- a/v2.0/split-at.md
+++ b/v2.0/split-at.md
@@ -51,6 +51,8 @@ the ranges that store tables or indexes:
   workload performance that results when automatic splits are unable to keep up
   with write traffic.
 
+Note that when a table is [truncated](truncate.html), it is essentially re-created in a single new empty range, and the old ranges that used to constitute the table are garbage collected. Any pre-splitting you have performed on the old version of the table will not carry over to the new version. The new table will need to be pre-split again.
+
 ## Examples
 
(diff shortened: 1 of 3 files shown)


rmloveland:
Thanks for the reviews Nathan and Jesse!

Jesse, thanks for the catch re: spacing.

Regarding:

> Should we mention this somewhere in the TRUNCATE documentation?

I think the [TRUNCATE docs](https://www.cockroachlabs.com/docs/dev/truncate.html) already handle this at the right level of abstraction by saying

> At a high level, it works by dropping the table and recreating a new table with the same name.

which implies that any operations on a previous table with the same name are null and void.  I think adding this level of detail only to SPLIT AT makes sense since you are only reading about SPLIT AT if you (think you) need it.

jseldess:
Makes sense, @rmloveland. Thanks for checking.

PR #4299, opened 2019-01-28
Fix table in null-handling.md

lucy-zhang:
The markdown table formatting was broken.


rmloveland:
@lucy-zhang I think we need an empty line between the end of the last table and the header "NULLs and arithmetic" (line 248)

jseldess:
Yep, @rmloveland was right. I made the change and also applied it to 2.2 docs.

jseldess:
Thanks for kicking this off, @lucy-zhang!

Issue #4253, opened 2019-01-15
sql: support CREATE STATISTICS with no columns and table refs


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/32981

From release notes:
> Added support for collecting table statistics on a default set of columns by calling `CREATE STATISTICS` with no columns specified. [#32981][#32981] {% comment %}doc{% endcomment %}

rmloveland:
Fixed by #4308.

Issue #4180, opened 2018-12-17
sql: add optional AS OF SYSTEM TIME with CREATE STATISTICS


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/32643

From release notes:
> Added support for `AS OF SYSTEM TIME` with the `CREATE STATISTICS` statement. [#32643][#32643] {% comment %}doc{% endcomment %}

rmloveland:
Addressed by #4308

PR #4308, opened 2019-01-30
Update CREATE STATISTICS doc with missing variants

### LINKED ISSUE #4241: Update stats statement docs with `CREATE STATISTICS foo FROM bar`
In version 2.2 you can issue the following statement, which
automatically figures out which columns to get stats on (specifically:
cols that are part of the primary key or an index).

    > CREATE STATISTICS employees_stats FROM employees;
    CREATE STATISTICS

However, our docs don't mention this yet.  So we should fix that. :-)

Estimated scope of work:

- Update [`CREATE STATISTICS`](create-statistics.html)
- Possibly other things (need to search docs for references to this statement)

rmloveland:
Fixes #4241, #4253, #4180.

Summary of changes:

- Update CREATE STATISTICS doc with the following variants:

  - CREATE STATISTICS foo FROM bar
  - CREATE STATISTICS __auto__ FROM bar
  - CREATE STATISTICS foo FROM bar AS OF SYSTEM TIME '-4h'

- Added information about automatic stats collection to the CREATE
  STATISTICS page (via an include)

- Updated instructions for deleting statistics, added to both CREATE
  STATISTICS and SHOW STATISTICS pages (via include)

- Updated railroad diagram and parameter descriptions for CREATE
  STATISTICS v2.2

- Updated sidebar and page titles to remove extraneous "experimental"
  text, since it's already clearly marked in the various documents via
  an include.

- Used a proper emdash (!) in a few places.

### DIFF
--- a/_includes/sidebar-data-v2.2.json
+++ b/_includes/sidebar-data-v2.2.json
@@ -733,7 +733,7 @@
                 ]
               },
               {
-                "title": "<code>CREATE STATISTICS</code> (Experimental)",
+                "title": "<code>CREATE STATISTICS</code>",
                 "urls": [
                   "/${VERSION}/create-statistics.html"
                 ]
@@ -1075,7 +1075,7 @@
                 ]
               },
               {
-                "title": "<code>SHOW STATISTICS</code> (Experimental)",
+                "title": "<code>SHOW STATISTICS</code>",
                 "urls": [
                   "/${VERSION}/show-statistics.html"
                 ]
--- a/_includes/v2.2/misc/automatic-statistics.md
+++ b/_includes/v2.2/misc/automatic-statistics.md
@@ -1,0 +1,27 @@
+<span class="version-tag">New in v2.2</span>: CockroachDB can generate table statistics automatically as tables are updated.
+
+To turn on automatic statistics collection:
+
+1. Run the following statement to enable the automatic statistics [cluster setting](cluster-settings.html):
+
+    {% include copy-clipboard.html %}
+    ~~~ sql
+    > SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=true
+    ~~~
+
+2. Restart the nodes in your cluster so it can generate statistics for all tables at startup (including read-only tables).
+
+To turn off automatic statistics collection:
+
+1. Run the following statement to disable the automatic statistics [cluster setting](cluster-settings.html):
+
+    {% include copy-clipboard.html %}
+    ~~~ sql
+    > SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=false
+    ~~~
+
+2. Look up what statistics were created by the automatic statistics generator using the [`SHOW STATISTICS`](show-statistics.html) statement.
+
+3. Delete the automatically generated statistics using the instructions in [delete statistics](create-statistics.html#delete-statistics).
+
+4. Restart the nodes in your cluster to clear the statistics caches.
--- a/_includes/v2.2/misc/delete-statistics.md
+++ b/_includes/v2.2/misc/delete-statistics.md
@@ -1,0 +1,15 @@
+To delete statistics for all tables in all databases:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> DELETE FROM system.table_statistics WHERE true;
+~~~
+
+To delete a named set of statistics (e.g, one named "my_stats"), run a query like the following:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> DELETE FROM system.table_statistics WHERE name = 'my_stats';
+~~~
+
+For more information about the `DELETE` statement, see [`DELETE`](delete.html).
--- a/_includes/v2.2/sql/diagrams/create_stats.html
+++ b/_includes/v2.2/sql/diagrams/create_stats.html
@@ -1,4 +1,4 @@
-<div><svg width="655" height="103">
+<div><svg width="627" height="103">
 <polygon points="9 17 1 13 1 21"></polygon>
 <polygon points="17 17 9 13 9 21"></polygon>
 <rect x="31" y="3" width="72" height="32" rx="10"></rect>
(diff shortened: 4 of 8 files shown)


rmloveland:
Direct links for (possibly) easier review:

- [CREATE STATISTICS](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/5d4d47208870373e9355b73f47956fc5d83cc079/dev/create-statistics.html) (several new things -- see summary of changes in commit message)
- [SHOW STATISTICS](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/5d4d47208870373e9355b73f47956fc5d83cc079/dev/show-statistics.html) (new section on how to delete stats)

And some specific questions:

-   Are there any functional differences in the operation of `CREATE
         STATISTICS foo FROM bar` and `CREATE STATISTICS __auto__ FROM
         bar`?  Based on very light manual testing, it seems not, except that the
    `__auto__` stats are updated automagically if you have the auto
    stats flag set?  (But please let me know if this is wrong)

-   Is the description of how the name `__auto__` is treated
    specially by the system accurate?  (in
    <https://github.com/cockroachdb/docs/pull/4308/files#diff-020dc10848f3aae7ded12a8608f1d84a>)

rytaft:
This is great, Rich! I have a few comments:

On the CREATE STATISTICS page:
- I think you can get rid of the warning saying "this is an experimental feature". I don't think the interface/output are likely to change. (Maybe you could call it a beta feature similar to the cost-based optimizer page...?)
- You can get rid of the item under "Considerations" saying that old stats must be deleted by the user. We now do this automatically (there are still some cases where a user may want to delete stats, though -- e.g. if they manually created a column stat that turned out not to be useful and is not one of the stats we refresh automatically)
- Under the Automatic Statistics section, you can take out the part about manually creating stats to avoid overloading the system. I've fixed that issue now. Users should still manually create stats for tables that they don't expect to update frequently, though, since I haven't yet added logic to do that automatically. Enabling automatic stats will currently only create statistics for tables that are updated regularly. (I'm going to be working on this next, so there's a chance I could also get this feature in before the branch is cut Monday...)
- You can take out all references to the name `__auto__`. Since we're using that name for automatic stats, I think it's better if users just aren't aware of it. They should create their own names instead. (I told you to include it before because of the potential for overloading the system -- but that's not a concern anymore).
- The name for the setting has changed. It's no longer `sql.defaults.experimental_automatic_statistics`. Now it's a cluster setting and called `sql.stats.experimental_automatic_collection.enabled`. (So you'll need to run `SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=true`)
- For the AOST example, I'd suggest using a more realistic example than 4h -- maybe 1m? The reason to use AOST in this case is just to avoid interfering with the production workload. There's no benefit to having stats that are really old.

On the SHOW STATISTICS page:
- I think you can also get rid of the experimental feature warning here.
- Otherwise everything looks good.

Sorry this is a lot of changes -- things keep changing since I'm actively working on this stuff! Let me know if any of this is confusing -- happy to discuss offline too.

rmloveland:
> On Jan 31, 2019, at 4:15 PM, rytaft <notifications@github.com> wrote:
> 
> This is great, Rich! I have a few comments:
> 
Thank you! And thanks for reviewing this PR.  Responses and a couple of questions below.
> On the CREATE STATISTICS page:
> 
> I think you can get rid of the warning saying "this is an experimental feature". I don't think the interface/output are likely to change. (Maybe you could call it a beta feature similar to the cost-based optimizer page…?)
Done.
> You can get rid of the item under "Considerations" saying that old stats must be deleted by the user. We now do this automatically (there are still some cases where a user may want to delete stats, though -- e.g. if they manually created a column stat that turned out not to be useful and is not one of the stats we refresh automatically)
Done.  I deleted that section altogether (but left the instructions for deleting stats where they were for reference when needed).
> Under the Automatic Statistics section, you can take out the part about manually creating stats to avoid overloading the system. I've fixed that issue now. Users should still manually create stats for tables that they don't expect to update frequently, though, since I haven't yet added logic to do that automatically. Enabling automatic stats will currently only create statistics for tables that are updated regularly. (I'm going to be working on this next, so there's a chance I could also get this feature in before the branch is cut Monday…)
I’ve removed the part about manually creating stats.

Should I still add the instruction that users should still manually create stats for tables they don’t expect to update frequently?  Or did the potential change you mentioned make it in?
> You can take out all references to the name __auto__. Since we're using that name for automatic stats, I think it's better if users just aren't aware of it. They should create their own names instead. (I told you to include it before because of the potential for overloading the system -- but that's not a concern anymore).
Done.  All references to __auto__ have been removed from the docs.
> The name for the setting has changed. It's no longer sql.defaults.experimental_automatic_statistics. Now it's a cluster setting and called sql.stats.experimental_automatic_collection.enabled. (So you'll need to run SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=true)
Done.  All references to this setting have been updated.
> For the AOST example, I'd suggest using a more realistic example than 4h -- maybe 1m? The reason to use AOST in this case is just to avoid interfering with the production workload. There's no benefit to having stats that are really old.
Done.  Updated to use ‘1m’ as you said.
> On the SHOW STATISTICS page:
> 
> I think you can also get rid of the experimental feature warning here.
Done.
> Otherwise everything looks good.
Yay, thanks!
> Sorry this is a lot of changes -- things keep changing since I'm actively working on this stuff! Let me know if any of this is confusing -- happy to discuss offline too.
> 
No worries, change is good.  I think other than my question above I’m not too confused about anything you wrote (at least that I’m aware of, but let me know 🙂)

rytaft:
Great -- thanks! Do you have links you want me to review again?

Regarding creating stats for tables that aren't updated frequently: I did get a PR in that creates stats for all tables (including read-only tables) when the nodes start up (https://github.com/cockroachdb/cockroach/pull/34519). So I think the main advice we should give to users now is that they should restart their servers after enabling the `sql.stats.experimental_automatic_collection.enabled` cluster setting. (Of course this probably doesn't matter if they are doing an upgrade -- presumably they would be cycling the machines anyway...)

rmloveland:
> On Feb 4, 2019, at 2:45 PM, rytaft <notifications@github.com> wrote:
> 
> Great -- thanks! Do you have links you want me to review again?
> 

Yes please -  Here is the updated link to CREATE STATS which also includes the information below re: restarting after the setting):

CREATE STATISTICS <http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/2b2e7c59273522975870f95c3fc0944a8d9d89c6/dev/create-statistics.html>

> Regarding creating stats for tables that aren't updated frequently: I did get a PR in that creates stats for all tables (including read-only tables) when the nodes start up (cockroachdb/cockroach#34519 <https://github.com/cockroachdb/cockroach/pull/34519>). So I think the main advice we should give to users now is that they should restart their servers after enabling the sql.stats.experimental_automatic_collection.enabled cluster setting. (Of course this probably doesn't matter if they are doing an upgrade -- presumably they would be cycling the machines anyway...)
> 
> 

OK, I updated the automatic stats instructions with the above information - should be visible on the page linked above.

rytaft:
Nice! Really minor: in the AOST example you call the stats `yesterday_employee_stats`. Since the time is just one minute ago, why not call them `employee_stats`?

Another minor suggestion: maybe make the info about automatic statistics more prominent? I'm going to enable automatic statistics by default for this alpha, so we may want to make it clear how to turn them on *and* off.  If someone wants to turn auto stats off and also get rid of all the stats, they will need to first set that cluster setting to false, then delete the stats (using the instructions you've provided), and then restart their machines to clear the stats caches.

rytaft:
Also one more thing: This command fails because `sql.stats.experimental_automatic_collection.enabled` is a cluster setting, not a session setting:
```
> SET sql.stats.experimental_automatic_collection.enabled=true
```
instead it should be:
```
> SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=true
```

rmloveland:
> On Feb 4, 2019, at 3:55 PM, rytaft wrote:
> 
> Nice! Really minor: in the AOST example you call the stats yesterday_employee_stats. Since the time is just one minute ago, why not call them employee_stats?
> 
Fixed.  Thanks for catching that.
> Another minor suggestion: maybe make the info about automatic statistics more prominent? I'm going to enable automatic statistics by default for this alpha, so we may want to make it clear how to turn them on and off. If someone wants to turn auto stats off and also get rid of all the stats, they will need to first set that cluster setting to false, then delete the stats (using the instructions you've provided), and then restart their machines to clear the stats caches.
> 

Great idea - I’ve bumped the “Automatic Statistics” section up a level on the CREATE STATS page so it has its own TOC link.  And on the main CBO page I’ve moved it to the top of the “Examples” section.

I’ve also:

- Updated the settings page to note that the default for the setting is now ‘true’.
- Added your instructions above for turning off automatic stats
- updated 'SET sql.stats.XXX' to ‘SET CLUSTER SETTING sql.stats.XXX’ as you noted in another comment in this thread (Thank you for this catch!)

Link to latest build of CREATE STATS page is here <http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/f249afa2c478a8875f21724884893d538a5826c6/dev/create-statistics.html>

rytaft:
LGTM! I'll let you know if for some reason the [final PR to enable auto stats](https://github.com/cockroachdb/cockroach/pull/34529) doesn't get merged before the alpha branch is cut, but otherwise we should be good to go.

rmloveland:
> On Feb 4, 2019, at 4:49 PM, rytaft wrote:
> 
> LGTM! I'll let you know if for some reason the final PR to enable auto stats <https://github.com/cockroachdb/cockroach/pull/34529> doesn't get merged before the alpha branch is cut, but otherwise we should be good to go.
> 
Hey Becca, looks like that one merged - did it make it into the alpha release?

If so I’ll pass this along for Docs team review now.

rytaft:
Yep! The alpha branch hasn't been cut yet, so it's definitely in. Thanks for your work on this!

PR #4320, opened 2019-02-04
Warning about rebalancing in a 3-node cluster with multiple stores/node

### LINKED ISSUE #4319: Multi-store nodes and rebalancing
Users have asked how rebalancing works when a node has multiple disks. We need to answer this in the docs. 

This PR is a start: https://github.com/cockroachdb/docs/pull/3488/files. And this forum post has most of what we need: https://forum.cockroachlabs.com/t/storage-on-multiple-disks/1328/5

jseldess:
Fixes #4319.


jseldess:
@bdarnell, @knz, I'm basing this on this helpful PR from an external contributor: https://github.com/cockroachdb/docs/pull/3488

However, I haven't yet included this bit:
> If you have similar setup, for the most automatic IO and ranges load balancing between drives, please use filesystem-level striping across the multiple disks (LVM, btrfs, linear RAID). This will increase read and write bandwidth just as well as using multiple stores in CockroachDB. However please note that on drive failure, the failure domain is the entire server, not only single disk on the server.

Do we want to make this a canonical recommendation in our docs for 3-node cluster with multiple stores per node?

jseldess:
TFTR, @knz. I'll apply to 2.2 as well.

PR #4325, opened 2019-02-06
Add doc on API stability

knz:
Here is a doc which does two things:

- it clarifies what external users can depend on
- it clarifies "experimental" vs "beta" vs "stable"

cc @BramGruneir @mjibson @bobvawter @jordanlewis for checking this is aligned with our actual practices so far and aspirations for the future.

cc @rmloveland for further editing and integration in the docs.

cc @bdarnell to sign off on the stability commitments.

cc @rolandcrosby @awoods187 to acknowledge the compatibility and visibility commitments.

### DIFF
--- a/v2.2/api-stability.md
+++ b/v2.2/api-stability.md
@@ -1,0 +1,360 @@
+---
+title: API Stability
+summary: Stability commitment of various APIs throughout the lifecycle of CockroachDB.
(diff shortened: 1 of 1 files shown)


knz:
I broke this PR... I have to re-create it. will do

Issue #2957, opened 2018-04-19
Document how to use crdb_internal


lnhsingh:
In trying to figure out how to read from the `information_schema` table, it was suggested by @benesch that it might be beneficial to teach people how to use `crdb_internal`.

Taken from slack:

```
My opinion is that it probably makes more sense to teach people to use crdb_internal
information_schema is technically "stable" but it has a bunch of random crap because it needs to describe any SQL database ever
And crdb_internal only has stuff that's actually relevant to Cockroach
```

benesch:
The pushback here will be that we don't want applications to rely on `crdb_internal`. If production applications query `crdb_internal` in we'll be locked in to backwards compatibility, and we've tried hard to avoid that so far. I think we could solve this with big "EXPERIMENTAL ONLY" callouts—it _is_ super useful for one off-introspections a broken cluster, where backwards compatibility concerns don't apply—but now that I think about it I am kind of nervous about making it easier to use `crdb_internal`.

Maybe the right answer is to ensure that absolutely everything useful in `crdb_internal` eventually has a stable interface. Was there something in particular you were trying to read from `information_schema`, @lhirata?

lhirata:
Interesting - will see what @bdarnell and @jseldess say too.

Re: `information_schema`, I was trying to remember how to read from it (for table constraints). Was able to figure it out by googling it. I ended up adding examples to the Information Schema doc, to hopefully make it easier for future me to figure out: #2958

benesch:
Ah, I see! Yeah, definitely a good idea to show people a representative example or two. But documenting all of its facets will be a lot of effort for little payoff, especially given that it's mostly documented elsewhere (e.g., in the Postgres docs and the SQL standard).

Also we have a much nicer `SHOW CONSTRAINTS` for the specific case of showing a table's constraints. Apologies if you already knew that!

jseldess:
Adding a comment from @bdarnell: 

> In general my thinking is that crdb_internal should remain internal-only and undocumented. we should take the stuff that's useful enough to document and put it behind SHOW commands or maybe make a separate `crdb_catalog` virtual schema with more stability/documentation guarantees
i also lean towards documenting information_schema instead of introducing yet another db-specific view of this data, although i don't know enough about what information_schema actually looks like to have an informed opinion here.

Closing this as a result. But I'll ping product on identifying useful parts of crdb_internal to put behind `SHOW` statements.

awoods187:
[...]

PR #4326, opened 2019-02-06
Add doc on API stability

knz:
Here is a doc which does two things:

- it clarifies what external users can depend on
- it clarifies "experimental" vs "beta" vs "stable"

cc @BramGruneir @mjibson @bobvawter @jordanlewis for checking this is aligned with our actual practices so far and aspirations for the future.

cc @rmloveland for further editing and integration in the docs.

cc @bdarnell to sign off on the stability commitments.

cc @rolandcrosby @awoods187 to acknowledge the compatibility and visibility commitments.

### DIFF
--- a/_includes/sidebar-data-v19.2.json
+++ b/_includes/sidebar-data-v19.2.json
@@ -2051,6 +2051,82 @@
             ]
           }
         ]
+      },
+      {
+          "title": "APIs and forward compatibility",
+          "items": [
+			  {
+				  "title": "Interface types",
+				  "urls": [
+					  "/${VERSION}/interface-types.html"
+				  ]
+			  },
+			  {
+				  "title": "Compatibility and programmability guarantees",
+				  "urls": [
+					  "/${VERSION}/compatibility-and-programmability-guarantees.html"
+				  ]
+			  },
+			  {
+			  	  "title": "Experimental feature lifecycle",
+			  	  "urls": [
+			  		  "/${VERSION}/experimental-feature-lifecycle.html"
+			  	  ]
+			  },
+			  {
+				  "title": "Overview of APIs and interfaces",
+				  "urls": [
+					  "/${VERSION}/overview-of-apis-and-interfaces.html"
+				  ]
+			  },
+			  {
+				  "title": "Programmability of command-line interfaces",
+				  "urls": [
+					  "/${VERSION}/programmability-of-command-line-interfaces.html"
+				  ]
+			  },
+			  {
+				  "title": "SQL APIs and compatibility",
+				  "items": [
+					  {
+						  "title": "Client-server protocol",
+						  "urls": [
+							  "/${VERSION}/client-server-protocol.html"
+						  ]
+					  },
+					  {
+						  "title": "SQL functional behavior guarantees",
+						  "urls": [
+							  "/${VERSION}/sql-functional-behavior-guarantees.html"
+						  ]
+					  },
+					  {
+						  "title": "Programmability of session variables",
+						  "urls": [
+							  "/${VERSION}/programmability-of-session-variables.html"
+						  ]
+					  },
+					  {
+						  "title": "Time and memory cost model",
+						  "urls": [
+							  "/${VERSION}/time-and-memory-cost-model.html"
+						  ]
+					  },
+					  {
+						  "title": "Introspection interfaces",
+						  "urls": [
+							  "/${VERSION}/introspection-interfaces.html"
+						  ]
+					  }
+				  ]
+			  },
+			  {
+				  "title": "Programmability of other interfaces",
+				  "urls": [
+					  "/${VERSION}/programmability-of-other-interfaces.html"
+				  ]
+			  }
+		   ]
       }
     ]
   },
--- a/_includes/v19.2/misc/experimental-warning.md
+++ b/_includes/v19.2/misc/experimental-warning.md
@@ -1,3 +1,3 @@
 {{site.data.alerts.callout_danger}}
-**This is an experimental feature**. The interface and output are subject to change.
+**This is an [experimental](experimental-feature-lifecycle.html) feature**. The interface and output are subject to change.
 {{site.data.alerts.end}}
--- a/v19.2/as-of-system-time.md
+++ b/v19.2/as-of-system-time.md
@@ -36,7 +36,7 @@ Format | Notes
 [`INT`](int.html) | Nanoseconds since the Unix epoch.
 negative [`INTERVAL`](interval.html) | Added to `statement_timestamp()`, and thus must be negative.
 [`STRING`](string.html) | A [`TIMESTAMP`](timestamp.html), [`INT`](int.html) of nanoseconds, or negative [`INTERVAL`](interval.html).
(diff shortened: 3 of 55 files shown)


knz:
I added the summary at the top.

knz:
> Put giant disclaimers on all of the individual SHOW pages? That seems pretty good.

knz:
> Put giant disclaimers on all of the individual SHOW pages? That seems pretty good.

That's a pretty good idea. I can go through the rest of the docs and add the annotations that indicate what is programmable and what is not, with references to the new page.

knz:
> as written will only ever be used or referenced by cockroach employees due to its length and depth. That is, my read here is that the audience for this doc is us, not users.

My hope is that PMs and sales eng will also know this, so that they can set customer expectations right and also discover areas where we need to better answer user needs by providing additional guarantees (mainly turn non-programmable or reserved things into public and programmable).

knz:
> We need a separate document that we might have a slim hope that a user will read. It needs to get to something a user can do quickly. Like it could say in the first or second paragraph: "here's a list of things that might change between versions (SHOW, etc...), instead use pg_catalog.blah".

I also agree with this and I hope our PMs will agree for such a high-level summary to become part of the extended release notes for the next release, with a reference to the detailed doc.

knz:
>  I don't think it's useful for our users to say "see we had this document describing public non-programmable behavior? you should have read it". This almost seems like a footgun we are giving out to our users.

I think this sentence is not accounting for the reality of professional software development and paid support contracts. My experience is the opposite: when working for an enterprise shop we were damn careful to read the doc to double-check the contract guarantees, because stepping outside of the doc would void the support contract.

I think this doc will help in that way. @robert-s-lee do you agree? (I will also ask our customers during my visit next week)

knz:
@piyush-singh I have added the section that was missing. I think the document is up-to-date now. Please go through the motion of accruing the required validations.

v19.2/api-stability.md
```diff
@@ -0,0 +1,586 @@
...
+  patch revisions.
+- there is no expectation of accurate or up-to-date documentation.
+
+## Stability phases for public and programmable interfaces
+
+New features or behavior changes in public/programmable interfaces are
+typically released first in status “experimental” then goes to “beta”
+then evolves to “stable”.
+
+Unless specified otherwise, a new feature that appears in version N in
```
rmloveland: I am reading this sentence as "everything in a new release is beta unless we explicitly document that it is stable".

However, our docs do not appear to currently follow this rule.  AFAICT, we do not explicitly document anywhere that a feature is stable.  We only explicitly note when a feature is unstable (beta/experimental), and remove the note when the feature graduates to stable.

Am I totally misreading this sentence? Please enlighten me if so.

knz: I think the intended meaning is "if the docs are silent, assume that the first release of a new feature is beta and that is automatically gets stabilized in the next release". Of course if the docs are explicit, that supersedes.

rmloveland: Got it - thanks for explaining.

v19.2/api-stability.md
```diff
@@ -0,0 +1,586 @@
...
+  we wish to have a period of time to gather feedback from users and
+  potentially make API changes accordingly. We thereby introduce
+  [intermediate **experimental** and **beta** stages](#stability-phases-for-public-and-programmable-interfaces) during which new
+  APIs may change more frequently or more radically.
+- Certain features of CockroachDB exist only as a mean to aid
+  development of CockroachDB itself and not meant to be consumed by
+  users. However, because CockroachDB is an open source project, users
+  often stumble upon these interfaces. To prevent users from accidentally
+  starting to rely on an internal interface without any stability
+  guarantee, we aim to call them out here explicitly as [**reserved**](#reserved-interfaces).
```
ericharmeling: > To prevent users from accidentally
  starting to rely on an internal interface without any stability
  guarantee, we aim to call them out here explicitly as [**reserved**](#reserved-interfaces).

@awoods187 
Given that the consensus is to reveal internal tables (#2957 and #4978), I do think in #5454 (the PR that fixes #2957 and #4978, a page on `crdb_internal`), if we are going to tell folks to use an internal table for scripting, we should have some kind of note about the stability of each table. I like the approach @knz has taken here.

v19.2/api-stability.md
```diff
@@ -0,0 +1,586 @@
...
+| `CANCEL`, `PAUSE`, `RESUME` inputs                                             | public and programmable                         | reserved                              |
+| `CANCEL`, `PAUSE`, `RESUME` success/error status                               | public and programmable                         | reserved                              |
+| `CANCEL`, `PAUSE`, `RESUME` outputs                                            | public and non-programmable                     | reserved                              |
+| `SCRUB` inputs and outputs                                                     | public and programmable in “experimental” phase | reserved                              |
+| Tabular data produced by set-generating functions also supported in PostgreSQL | public and programmable                         | public and programmable               |
+| Tabular data produced by set-generating functions specific to CockroachDB      | public and programmable                         | reserved                              |
+
+### SQL introspection interfaces
```
ericharmeling: @knz @awoods187 
Is this "public and programmable" vs. "reserved" distinction official for all of the `crdb_internal` tables listed here? In addition to listing out the tables in `crdb_internal` with name and description (see #5454), I'd like to have some kind of note about the stability of each table, very similar to the "public and programmable" vs. "reserved" distinction.

v19.2/api-stability.md
```diff
@@ -0,0 +1,586 @@
...
+| `crdb_internal.node_runtime_info`         | all                                                                                                                                                  | public and programmable               |
+| `crdb_internal.node_sessions`             | `node_id`, `session_id`, `user_name`, `client_address`, `application_name`, `active_queries`,`last_active_query`, `session_start`                    | public and programmable               |
+| `crdb_internal.node_sessions`             | other columns                                                                                                                                        | reserved                              |
+| `crdb_internal.node_statement_statistics` | all                                                                                                                                                  | public and programmable               |
+| `crdb_internal.partitions`                | all                                                                                                                                                  | reserved                              |
```
ericharmeling: @knz 
I'm seeing a `node_txn_stats` table here. Is this public/programmable?

knz:
As discussed with @ericharmeling I have split the new page into multiple small sub-pages, each covering just one topic.

Also as requested by @mjibson I have added the cross-references from the other doc pages, so that users have a greater chance to arrive to this new material.

As discussed with @piyush-singh I think this work is now maturing and I will motion to accelerate the review in view to make the content available in 19.2.

knz:
> o, my suggestion at this point would be that we open a new PR with the very high-level stuff from this PR (i.e. interface-types.md and overview-of-apis-and-interfaces.md and compatibility-and-programming-guarantees.md), and then start adding the docs specific to the interfaces/guarantees piece-by-piece in separate PR's 

Works for me. The 3 first pages you listed here are the most important anyway. (I would recommend also including the `experimental-feature-lifecycle.md` in there because it comes up a lot too)

jseldess:
@taroface, given the publication of https://www.cockroachlabs.com/docs/stable/api-support-policy.html, should this PR be closed?

taroface:
@jseldess I think so, but since this is a big PR, I'd like to confirm with @knz that there isn't some work in here that we might wish to repurpose for a future update or enhancement.

PR #4327, opened 2019-02-06
minor fixes

dbist:
some spelling mistakes
one command in restart cluster did not work for a local cluster until port was removed.
tpcc workload was pointing to bank command


dbist:
please review @jseldess

dbist:
addressed your comments @jseldess

PR #4332, opened 2019-02-06
Document join reordering by the optimizer

### LINKED ISSUE #4330: Document join reordering by the optimizer
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw5Jsp2TrY3it9Yr/recBO5KhQYDifJZJk
PM: @awoods187 
Eng: @andy-kimball 

As of recent 2.2 alpha, the optimizer will explore additional join
orderings if a (default-on) session variable is set.

Additionally, you can set a limit to the number of joins it will try
to reorder.

Estimated scope of work:

- Update CBO docs to mention that this is a thing, noting possible tradeoffs, etc.

- ~~Document the `experimental_reorder_joins` setting~~ (this is no longer a thing)

- Document the `experimental_reorder_joins_limit` setting

See also:

- [cockroach#34170](https://github.com/cockroachdb/cockroach/pull/34170)

- [cockroach#34549](https://github.com/cockroachdb/cockroach/pull/34549)

rmloveland:
Fixes #4330.

Summary of changes:

- Update 'Cost-based optimizer' page with a new section 'Join
  reordering', which discusses the new functionality, as well as
  tradeoffs involved in its operation.  The new
  `experimental_reorder_joins_limit` setting is described, along with
  some warnings to the user about not setting it too high, since that
  may lead to performance degradation.

- Update the session variables page to include the
  `experimental_reorder_joins_limit` variable.

### DIFF
--- a/v2.2/cost-based-optimizer.md
+++ b/v2.2/cost-based-optimizer.md
@@ -143,6 +143,25 @@ The query plan cache is still under development and has the following limitation
 - If you use the query plan cache in conjunction with table statistics, cached plans do not yet get invalidated when new statistics are created.
 {{site.data.alerts.end}}
 
+## Join reordering
+
+<span class="version-tag">New in v2.2</span>: The cost-based optimizer will explore additional join orderings in an attempt to find the lowest-cost execution plan for a query involving multiple joins, which can lead to significantly better performance in some cases.
+
+Because this process leads to an exponential increase in the number of possible execution plans for such queries, it's only used to reorder subtrees containing 4 or fewer joins by default.
+
+To change this setting, which is controlled by the `experimental_reorder_joins_limit` [session variable](set-vars.html), run the statement shown below.  To disable this feature, set the variable to `0`.
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET experimental_reorder_joins_limit = 6;
+~~~
+
+{{site.data.alerts.callout_danger}}
+We strongly recommend not setting this value higher than 8 to avoid performance degradation. If set too high, the cost of generating and costing execution plans can end up dominating the total execution time of the query.
+{{site.data.alerts.end}}
+
+For more information about the difficulty of selecting an optimal join ordering, see our blog post [An Introduction to Join Ordering](https://www.cockroachlabs.com/blog/join-ordering-pt1/).
+
 ## How to turn the optimizer off
 
 With the optimizer turned on, the performance of some workloads may change. If your workload performs worse than expected (e.g., lower throughput or higher latency), you can turn off the cost-based optimizer and use the heuristic planner.
--- a/_includes/v2.2/misc/session-vars.html
+++ b/_includes/v2.2/misc/session-vars.html
@@ -98,6 +98,18 @@
   </tr>
 
   <tr>
+   <td>
+    <code>experimental_reorder_joins_limit</code>
+   </td>
+   <td>Maximum number of joins that the optimizer will attempt to reorder when searching for an optimal query execution plan. For more information, see <a href="cost-based-optimizer.html#join-reordering">Join reordering</a>.</td>
+   <td>
+    <code>4</code>
+   </td>
+   <td>Yes</td>
+   <td>Yes</td>
+  </tr>
+
+  <tr>
    <td><code>force_savepoint_restart</code></td>
    <td>When set to <code>true</code>, allows the <a href="savepoint.html"><code>SAVEPOINT</code></a> statement to accept any name for a savepoint.</td>
    <td>


rmloveland:
For convenience, direct link to the new [join reordering](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/98fffc974965e241c6d6f240e27c1e1cb3ec8461/dev/cost-based-optimizer.html#join-reordering) section of the CBO page.

Specific questions:

- Is the mention of "join associativity" too specific?  I read the PR which maybe I shouldn't have since I didn't "understand" it
- Is it accurate to say the increase in the number of possible plans is exponential, as we do in this draft?
- I made up a max number for the setting (12), but what should the max recommended value be?
- Is there anything incorrect?
- Is there anything missing that you think should be stated here?

v2.2/cost-based-optimizer.md
```diff
@@ -143,6 +143,25 @@ The query plan cache is still under development and has the following limitation
...
+{% include copy-clipboard.html %}
+~~~ sql
+> SET experimental_reorder_joins_limit = 8;
+~~~
+
+{{site.data.alerts.callout_danger}}
+We strongly recommend not setting this value higher than 12 to avoid performance degradation. If set too high, the cost of generating and costing execution plans can end up dominating the total execution time of the query.
+{{site.data.alerts.end}}
+
+For more information about the difficulty of selecting an optimal join ordering (spoiler: it's NP-complete), see our blog post [An Introduction to Join Ordering](https://www.cockroachlabs.com/blog/join-ordering-pt1/).
```
justinj: I'm not actually sure it's NP-complete, it's definitely NP-hard.

v2.2/cost-based-optimizer.md
```diff
@@ -143,6 +143,25 @@ The query plan cache is still under development and has the following limitation
 - If you use the query plan cache in conjunction with table statistics, cached plans do not yet get invalidated when new statistics are created.
 {{site.data.alerts.end}}
 
+## Join reordering
+
+<span class="version-tag">New in v2.2</span>: The cost-based optimizer will explore additional join orderings in an attempt to find the lowest-cost execution plan for a query involving multiple joins. Specifically, it applies the rule of join associativity to generate even more possible join orderings to choose from, which can lead to significantly better performance in some cases.
+
+Because this process leads to an exponential increase in the number of possible execution plans for such queries, it's only used on queries with 4 or fewer joins by default, as determined by the `experimental_reorder_joins_limit` [session variable](set-vars.html).
```
justinj: It's not really "queries with 4 or fewer joins", it's "will reorder subtrees containing 4 or fewer joins." But I'm not sure if that's difficult to communicate here?

The difference is that if a query has, say, 8 joins, the subtree it has containing 4 *will* be reordered.

rmloveland:
> On Feb 6, 2019, at 5:24 PM, Justin Jeffrey wrote:
> 
> @justinj approved this pull request.
> 
> I'm biased towards being less specific, but if you want to include the word "associativity" you should also include "commutativity" since they work together to explore everything.
I removed the mention of "associativity" then, since having both of these terms means we should probably get into what they mean, which is really handled better in your blog post(s), for now.  It would def blow up our word count here and be possibly confusing for non-join-experts.

> It's actually worse, it follows the Catalan Numbers <https://en.wikipedia.org/wiki/Catalan_number>. But exponential is close enough for these purposes I think.
OK, leaving it at “exponential” then. 🙂
> 12 is super high, I would say going above 8 will begin to cause problems. Here's a rough graph I made showing the plan time for max depth+number of tables a couple weeks ago:
> 
OK, will update to say the limit should be 8.

Related: should we add a line that says “Most users should not change this setting” ?   It might be good to discourage casually tweaking this. (Or not? What do you think?)
>  <https://user-images.githubusercontent.com/409075/52377493-0c535f80-2a33-11e9-92f1-d951622e5bff.png>
> In v2.2/cost-based-optimizer.md <https://github.com/cockroachdb/docs/pull/4332#discussion_r254475852>:
> 
> > +<span class="version-tag">New in v2.2</span>: The cost-based optimizer will explore additional join orderings in an attempt to find the lowest-cost execution plan for a query involving multiple joins. Specifically, it applies the rule of join associativity to generate even more possible join orderings to choose from, which can lead to significantly better performance in some cases.
> +
> +Because this process leads to an exponential increase in the number of possible execution plans for such queries, it's only used on queries with 4 or fewer joins by default, as determined by the `experimental_reorder_joins_limit` [session variable](set-vars.html).
> +
> +To change the number of joins that will be reordered by the optimizer, set the variable as shown below.
> +
> +{% include copy-clipboard.html %}
> +~~~ sql
> +> SET experimental_reorder_joins_limit = 8;
> +~~~
> +
> +{{site.data.alerts.callout_danger}}
> +We strongly recommend not setting this value higher than 12 to avoid performance degradation. If set too high, the cost of generating and costing execution plans can end up dominating the total execution time of the query.
> +{{site.data.alerts.end}}
> +
> +For more information about the difficulty of selecting an optimal join ordering (spoiler: it's NP-complete), see our blog post [An Introduction to Join Ordering](https://www.cockroachlabs.com/blog/join-ordering-pt1/).
> I'm not actually sure it's NP-complete, it's definitely NP-hard.
> 
Ah, typo on my part since I didn’t really know the difference!  Updated to say “NP-hard”, and made it a link to Wikipedia article on that topic for the curious (which I will now go read 🙂)

> 
> In v2.2/cost-based-optimizer.md <https://github.com/cockroachdb/docs/pull/4332#discussion_r254476689>:
> 
> > @@ -143,6 +143,25 @@ The query plan cache is still under development and has the following limitation
>  - If you use the query plan cache in conjunction with table statistics, cached plans do not yet get invalidated when new statistics are created.
>  {{site.data.alerts.end}}
>  
> +## Join reordering
> +
> +<span class="version-tag">New in v2.2</span>: The cost-based optimizer will explore additional join orderings in an attempt to find the lowest-cost execution plan for a query involving multiple joins. Specifically, it applies the rule of join associativity to generate even more possible join orderings to choose from, which can lead to significantly better performance in some cases.
> +
> +Because this process leads to an exponential increase in the number of possible execution plans for such queries, it's only used on queries with 4 or fewer joins by default, as determined by the `experimental_reorder_joins_limit` [session variable](set-vars.html).
> It's not really "queries with 4 or fewer joins", it's "will reorder subtrees containing 4 or fewer joins." But I'm not sure if that's difficult to communicate here?
> 

If it’s what this setting refers to, I think we must explain it - what does “subtree” mean in this context from a user’s POV?  Do I need to look at EXPLAIN output, or is it possible to determine by visually inspecting a SQL query?
> The difference is that if a query has, say, 8 joins, the subtree it has containing 4 will be reordered.
> 
I can definitely add this info once we explain subtrees (as discussed above).

Finally, I’ve pushed the changes I’ve made that incorporate your feedback so far.

rmloveland:
Re: discussion above about describing "subtrees", I think that work can wait until after the current alpha release.  Leaving this as is for now for docs review.

rmloveland:
> I'd remove `(spoiler: it's NP-hard)`.

Fixed.

PR #4346, opened 2019-02-11
Follower reads

sploiselle:
@ajwerner PTAL!

### DIFF
--- a/v2.2/follower-reads.md
+++ b/v2.2/follower-reads.md
@@ -1,0 +1,110 @@
+---
+title: Follower Reads
+summary: To reduce latency for read queries, you can choose to have the closest node serve the request using the follower reads feature.
+toc: true
+---
+
+<span class="version-tag">New in v2.2:</span> To reduce latency for read queries, you can use the follower reads feature, which lets the closest replica serve the read request at the expense of only not guaranteeing that data is up to date.
+
+{{site.data.alerts.callout_danger}}
+The Follower reads feature is only available to [enterprise](https://www.cockroachlabs.com/product/cockroachdb/) users.
+{{site.data.alerts.end}}
+
+## What are Follower reads?
+
+Follower reads are a mechanism to let any replica of a range serve a read request, but are only available for read queries that are sufficiently in the past, i.e., using `AS OF SYSTEM TIME`. Currently, follower reads are available for any read operation at least 48 seconds in the past, though there is active work reduce that window. 
+
+In widely distributed deployments, using follower reads can reduce the latency of read operations (which can also increase throughput) by letting the replica closest to the gateway serve the request, instead of forcing the gateway to communicate with the leaseholder, which could be geographically distant.
+
+To future-proof this feature in your code, we've also included a convenience function (`experimental_follower_read_timestamp()`) that runs your queries at a time as close as possible to the present time while remaining safe for follower reads.
+
+## Settings
+
+### Enable/disable follower reads
+
+Use [`SET CLUSTER SETTING`](set-cluster-setting.html) to set `kv.closed_timestamp.follower_reads_enabled` to:
+
+- `false` to disable follower reads _(default)_
+- `true` to enable follower reads 
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING kv.closed_timestamp.follower_reads_enabled = true;
+~~~
+
+#### When to use follower reads
+
+Follower reads return consistent historical reads; currently a minimum of 48 seconds in the past, though we are actively working on reducing that number.
+
+As long as your `SELECT` operations can tolerate slightly outdated data, Follower reads can reduce read latencies and increase throughput.
+
+#### When not to use follower reads
+
+You should not use follower reads when you need up-to-date data.
+
+### Make follower read-compatible queries
+
+Any `SELECT` statement with an `AS OF SYSTEM TIME` value at least 48 seconds in the past can be served by any replica (i.e., can be a Follower Read).
+
+To simplify this calculation, we've added a convenience function that will always set the `AS OF SYSTEM TIME` value to the minimum required for follower reads, `experimental_follower_read_timestamp()`:
+
+``` sql
+SELECT ... FROM ... AS OF SYSTEM TIME experimental_follower_read_timestamp();
+```
+
+### Make follower read-compatible transactions
+
+You can set the `AS OF SYSTEM TIME` value for all operations in a read-only transaction:
+
+```sql
+BEGIN AS OF SYSTEM TIME experimental_follower_read_timestamp();
+
+SAVEPOINT cockroach_restart;
+
+SELECT ...
+SELECT ...
+
+COMMIT;
+```
+
+## How follower reads works
+
+In CockroachDB's general architecture, all reads are served by a range's [leaseholder](architecture/replication-layer.html#leases), which is a replica elected to coordinate all write operations. Because this node contains information about all of a range's writes, it can also serve reads for the range while still guaranteeing `SERIALIZABLE` isolation. With this architecture, the client might need to communicate with a machine that is far away, creating greater network latencies.
+
+However, if you were to lower the isolation requirements of an operation, it's possible to serve the read from _any_ replica, not only the leaseholder, given that the data can be sufficiently old. 
+
+To accomplish this in CockroachDB, we've created a mechanism to let you express that any node can serve the request (`kv.closed_timestamp.follower_reads_enabled`) and that a historical read is sufficient (`AS OF SYSTEM TIME`), given that the argument to `AS OF SYSTEM TIME` is sufficiently in the past (`experimental_follower_read_timestamp()`.
+
+For a more detailed explanation, you can also read the [follower reads RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20180603_follower_reads.md).
+
+### Reading from followers
+
+Each CockroachDB node tracks a property called its "closed timestamp", which means that no new writes can ever be introduced below that timestamp. The closed timestamp advances forward by some target interval behind the current time. If the replica receives a write at a timestamp less than its closed timestamp, it rejects the write.
+
+With follower reads enabled, any replica on a node can serve a read for a key as long as the time at which the operation is performed (i.e. the `AS OF SYSTEM TIME` value) is less or equal to the node's closed timestamp.
+
+### Determining which node to read from
+
+Every node keeps a record of its latency with all other nodes in the system. When a gateway node in cluster with follower reads enabled receives a request to read a key with a sufficiently old `AS OF SYSTEM TIME` value, it forwards the request to the closest node that contains a replica of the data––whether it be a follower or the leaseholder.
+
+### Interactions with long-running writes
+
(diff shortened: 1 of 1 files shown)


v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Follower Reads
+summary: To reduce latency for `SELECT` operations, you can choose to have the closest node serve the request using the "Follower Reads" feature.
```
ajwerner: something of a nit: Maybe you should just say read queries? It can also speed up backups

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+
+{{site.data.alerts.callout_danger}}
+The Follower Read feature is only available to [enterprise](https://www.cockroachlabs.com/product/cockroachdb/) users.
+{{site.data.alerts.end}}
+
+## Settings
+
+### Enable/disable Follower Reads
+
+Use [`SET CLUSTER SETTING`](set-cluster-setting.html) to set `kv.range_split.by_load_enabled` to:
```
ajwerner: wrong variable here, should say `kv.closed_timestamp.follower_reads_enabled`

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+
+- `false` to disable follower reads _(default)_
```
ajwerner: this is right for this alpha but I really want to change this to true for the release

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING kv.closed_timestamp.follower_reads_enabled = true;
+~~~
+
+#### When to use Follower Reads
+
+Follower Reads return consistent historical reads; currently a minimum of 48 seconds in the past, though we are actively working on reducing that number.
+
+As long as your `SELECT` operations can tolerate slightly outdated data, Follower Reads can reduce read latencies.
```
ajwerner: maybe add: `and increase throughput`

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+- `true` to enable follower reads 
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING kv.closed_timestamp.follower_reads_enabled = true;
+~~~
+
+#### When to use Follower Reads
+
+Follower Reads return consistent historical reads; currently a minimum of 48 seconds in the past, though we are actively working on reducing that number.
```
ajwerner: I feel like somewhere in this document it's worth explaining what a follower read is. 

I might also use a quite comment about how a follower read differs from a leaseholder (maybe call it transactional) read.
```
Unlike transactional reads which must be fulfilled by the current leaseholder for a range, follower reads can be served safely by any replica. In order to utilize safe and consistent follower reads users must be willing to sacrifice some amount of data freshness; follower reads only apply to historical read queries (using AS OF SYSTEM TIME) which are sufficiently in the past. Currently that bar is calculated to be 48 seconds though there is active work to lower that number. A function described below can be provided to AS OF SYSTEM TIME to request a timestamp which is as close to the present as possible yet safe for follower reads.
```

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+
+As long as your `SELECT` operations can tolerate slightly outdated data, Follower Reads can reduce read latencies.
+
+#### When not to use Follower Reads
+
+Long-running write operations can have negative interactions with Follower Reads because the Follower Read can either get blocked by the long-running write or cause it to be retried. However, this can be mitigated by placing Follower Reads in `HIGH` priority transactions and using [transaction retries](transactions.html#client-side-transaction-retries) for the write operation.
```
ajwerner: This is more of a gotcha than a time not to use follower reads. The time not to use follower reads is when you need fresh data.

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Make Follower Read-compatible queries
+
+Any `SELECT` statement with an `AS OF SYSTEM TIME` value at least 48 seconds in the past can be served by any replica (i.e., can be a Follower Read).
+
+To simplify this calculation, we've added a convenience function that will always set the `AS OF SYSTEM TIME` value to the minimum required for follower reads, `experimental_follower_read_timestamp()`:
+
+``` sql
+SELECT ... FROM ... AS OF SYSTEM TIME(experimental_follower_read_timestamp())
```
ajwerner: no parens needed

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+### Make Follower Read-compatible transactions
+
+You can set the `AS OF SYSTEM TIME` value for all operations in a read-only transaction:
+
+```sql
+BEGIN AS OF SYSTEM TIME(experimental_follower_read_timestamp());
```
ajwerner: no parens needed

v2.2/follower-reads.md
```diff
@@ -0,0 +1,98 @@
...
+
+However, if you were lower the isolation requirements of an operation, it's possible to serve the read from _any_ replica, not only the leaseholder. To accomplish this is CockroachDB, we've create a mechanism to let you express that a historical read is sufficient (`AS OF SYSTEM TIME`) and that any node can server the request (`kv.closed_timestamp.follower_reads_enabled`).
+
+### Reading from followers
+
+Each CockroachDB node tracks a property called its "closed timestamp", which means that no new writes can ever be introduced below that timestamp. The closed timestamp advances forward every 30 seconds. If the replica receives a write at a timestamp less than its closed timestamp, it rejects the write.
```
ajwerner: It's probably best not to overly repeat the current default value of the the cluster setting. Maybe here just mention that it advances forward based on a target delay behind the present. Also perhaps add a link to https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20180603_follower_reads.md in this section?

v2.2/follower-reads.md
ajwerner: I feel like your below section is good enough for this issue.

v2.2/follower-reads.md
sploiselle: Glad to change for release. The idea is to document the state of the world for the alpha users.

sploiselle:
@jseldess @rmloveland PTAL. If there's any way to get this edited today, we'll be able to get it in time for the release notes.

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
...
+SELECT ...
+
+COMMIT;
+```
+
+## How Follower reads works
+
+In CockroachDB's general architecture, all reads are served by a range's [leaseholder](architecture/replication-layer.html#leases), which is a replica elected to coordinate all write operations. Because this node contains information about all of a range's writes, it can also serve reads for the range while still guaranteeing `SERIALIZABLE` isolation. With this architecture, the client might need to communicate with a machine that is far away, creating greater network latencies.
+
+However, if you were lower the isolation requirements of an operation, it's possible to serve the read from _any_ replica, not only the leaseholder. To accomplish this is CockroachDB, we've create a mechanism to let you express that a historical read is sufficient (`AS OF SYSTEM TIME`) and that any node can serve the request (`kv.closed_timestamp.follower_reads_enabled`).
```
ajwerner: `s/we've create/we've created/`

ajwerner: This paragraph needs editing. Something about sufficiently old.

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Follower Reads
+summary: To reduce latency for read queries, you can choose to have the closest node serve the request using the "Follower reads" feature.
+toc: true
+---
+
+<span class="version-tag">New in v2.2:</span> To reduce latency for read queries, you can choose to have the closest node serve the request using the follower read  feature. For this to work, though, the response might not have the most recent version of the data, though the data will be consistent at the point in time the read is served from.
```
rmloveland: > For this to work, though, the response might not have the most recent version of the data, though the data will be consistent at the point in time the read is served from.

I would remove this sentence.  It's a bit confusing, and also the information is better expressed below (IMO).  I think having the intro be the single sentence "To reduce latency, ..." is fine.

rmloveland: > have the closest node serve the request using the follower read  feature

s/read/reads/

I'm not sure if the singular is also correct, but it caught my eye since the page title is "Follower reads".

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
...
+
+{{site.data.alerts.callout_danger}}
+The Follower Read feature is only available to [enterprise](https://www.cockroachlabs.com/product/cockroachdb/) users.
```
rmloveland: > The Follower Read feature

Same comment as above re: singular v. plural

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
...
+{{site.data.alerts.end}}
+
+## What are Follower reads?
+
+Follower reads are a mechanism to let any replica of a range serve a read request, but are only available for read queries that are sufficiently in the past, i.e., using `AS OF SYSTEM TIME`. Currently, follower reads are available for any read operation at least 48 seconds in the past, though there is active work reduce that window. 
+
+To future-proof this feature in your code, we've also included a convenience function that runs your queries at a time as close as possible to the present time while remaining safe for follower reads.
```
rmloveland: > To future-proof this feature in your code, we've also included a convenience function that runs your queries at a time as close as possible to the present time while remaining safe for follower reads.

> In widely distributed deployments, using follower reads can reduce the latency of read operations (which can also increase throughput) by letting the replica closest to the gateway serve the request, instead of forcing the gateway to communicate with the leaseholder, which could be geographically distant.

I think the order of these two paras should be reversed, on the argument that the info in the current 3rd para is more important/general than the 2nd.

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
...
+
+- `false` to disable follower reads _(default)_
+- `true` to enable follower reads 
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SET CLUSTER SETTING kv.closed_timestamp.follower_reads_enabled = true;
+~~~
+
+#### When to use Follower reads
```
rmloveland: Nitpicking, but I think we should call it "Follower Reads" with caps in all uses if we're using that feature name as a proper noun.  I would also be ok with lower-cased "follower reads" instead, as you've done elsewhere.  In other words IMO we should stick to only one usage or the other.

v2.2/follower-reads.md
```diff
@@ -0,0 +1,108 @@
...
+
+SAVEPOINT cockroach_restart;
+
+SELECT ...
+SELECT ...
+
+COMMIT;
+```
+
+## How Follower reads works
```
rmloveland: If "Follower Reads" is a proper noun and we want the plural variant, suggest this heading should be:

"How Follower Reads work"

(Same comment everywhere this term is used, so I don't annoy you with tons more comments. :-} )

v2.2/follower-reads.md
rmloveland: > if you were lower

"if you were to lower"

Issue #4323, opened 2019-02-05
CDC: Core implementation


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw5Jsp2TrY3it9Yr/recJE62qfNOgTJaxB
PM: @rolandcrosby 
Eng: @danhhz

PR #4351, opened 2019-02-11
CDC: Update cluster settings description

lhirata:
Reworded CDC section on the new cluster settings.

Related to #4323.

### DIFF
--- a/v2.2/change-data-capture.md
+++ b/v2.2/change-data-capture.md
@@ -108,17 +108,17 @@ The changefeed emits duplicate records 1, 2, and 3 before outputting the records
 
 ## Enable rangefeeds to reduce latency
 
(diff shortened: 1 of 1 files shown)


jseldess:
___
*[v2.2/change-data-capture.md, line 115 at r1](https://reviewable.io/reviews/cockroachdb/docs/4351#-LYTt_r5CsT1Vorh17Wy:-LYZmiEN16YzCyBuga5a:b-8mz4q7) ([raw file](https://github.com/cockroachdb/docs/blob/c6b35569c94b7b847b852e7cc263090ac1a6f2b1/v2.2/change-data-capture.md#L115)):*

Personally, I like this better if we delete the first two sentences entirely "If you are experiencing an issue, you can revert back to the previous behavior by setting `changefeed.push.enabled` to `false`. Note that this setting will be removed in a future release; if you have to use the fallback, please [file a Github issue](file-an-issue.html)."

If that's objectionable, then we shouldn't say that Rangefeeds require changefeed.push.enabled, because that's not quite true. Rangefeed-powered changefeeds (or End-to-end push changefeeds or something like that) need that setting.
</blockquote></details>

Dan's suggestion sounds fine to me.

danhhz:
Thanks, Lauren!

PR #4391, opened 2019-02-14
Add third-party tools sections; include Intellij docs

### LINKED ISSUE #4004: Test intelliJ ide
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw5Jsp2TrY3it9Yr/recLhELw2PiWauIWw
PM: @awoods187 
Eng: @BramGruneir

sploiselle:
Restructured docs around third-party tools to aid in visibility and extensibility. Added strawman description of our levels of support; don't know that it really exists, but think this works as an informal definition. Glad to change or tighten this up as other see fit.

These changes must also be propagated over to the 2.2 repo before merging.

Also added doc for using CockroachDB as a data source for Intellij.

@vilterp Could I get your eyes on this as an end-user of the tool to sanity check it for me?

@knz Don't necessarily need a deep technical review here, but you were the engineer on the issue for supporting these features so at least wanted to let you know this was incoming. I also know you've been working on some definitions of levels of support, so would be glad to get some insight into how we can align our communication around those things to customers. If there isn't an existing plan for talking about third-party tools, that's A-OK and we can table a discussion until later.

@jseldess Dug around and verified with @vilterp that Intellij doesn't support any kind of ORM-like class-relation mapping, so the use is limited to the doc I've linked to from Intellij. Would also love your eyes on this level of support.

After receiving these edits, will include PMs and copy edits.

Closes #4004


vilterp:
Glad to have this. Probably want to mention that these instructions work not only in IntelliJ idea, but in any IntelliJ IDE that has the Database tool window (GoLand, I think also RubyMine, probably PyCharm, etc). Not sure where definitive list is.

sploiselle:
@vilterp All of those IDEs are developed by JetBrains; is it common to refer to all of them as "IntelliJ" even though that isn't technically anywhere in their naming conventions?

sploiselle:
@rmloveland Can I get a copy edit on this?

PR #4405, opened 2019-02-21
remove extraneous link to MVCC explanation from txn arch docs

### LINKED ISSUE #4167: Redundant explanation of MVCC acronym
well down into the transaction layer documentation in the "Write Intents" section we expand the definition of MVCC as if it was the first time it was being introduced.  Perhaps this language should be used when the concept or acronym is first discussed earlier in the section... 

https://www.cockroachlabs.com/docs/stable/architecture/transaction-layer.html#write-intents

"These are essentially multi-version concurrency control values (also known as MVCC, which is explained in greater depth in the storage layer)"

For example the acronym is used in the paragraph above without this explanation.

sploiselle:
closes #4167

### DIFF
--- a/v2.1/architecture/transaction-layer.md
+++ b/v2.1/architecture/transaction-layer.md
@@ -114,7 +114,7 @@ The transaction record for a committed transaction remains until all its write i
 
 ### Write intents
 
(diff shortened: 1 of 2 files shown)


sploiselle:
🍪

PR #4406, opened 2019-02-21
expand considerations for decommissioning nodes

### LINKED ISSUE #3885: Improve decommissioning documentation
We need to improve documentation around the decommissioning process. 

We do not think customers should decommission multiple nodes at once.

Customers should never kill nodes that have not decommissioned. This causes irreparable data loss.

We will also likely invest time in learning about the [speed of the decommissioning process.](https://github.com/cockroachdb/cockroach/issues/31502)

sploiselle:
closes #3885

@awoods187 PTAL

@bdarnell The original issue had an ask from you for any additional input, so I'll just echo that here. If you don't have any feedback, A-OK.

### DIFF
--- a/v19.1/remove-nodes.md
+++ b/v19.1/remove-nodes.md
@@ -73,7 +73,11 @@ To successfully decommission a node, you need to first add a 6th node:
 
 ### Before you begin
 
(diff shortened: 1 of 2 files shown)


sploiselle:
@rmloveland Can I get a copy edit on this, please?

PR #4407, opened 2019-02-21
Release notes for v19.1.0-beta.20190225

jseldess:


Amruta-Ranade:
The release notes LGTM. 

Nit: Noticed that the version selector dropdown lists v19.1 after v2.0.7 and not as the latest version. Is this something we can fix?

jseldess:
Good catch, @Amruta-Ranade. I don't think there'll be time to fix that for the release today, but we should def do that soon.

Issue #4304, opened 2019-01-29
Use vCPU instead of CPU and define the terms


jseldess:
Our production checklist talks about CPUs, but users might not know if we're talking about CPU vs. vCPU vs core. In fact, we're talking about vCPUs. We should update the production checklist to use vCPUs, but we should also find the right place to define these terms. 

Helpful resource: https://docs.google.com/presentation/d/1r2y1lMs-6fhFtF4TiqCrrBbQcUmBakpeCFKW3QNOg_g/edit#slide=id.g469a4beebf_0_159

Issue #4287, opened 2019-01-23
workload load takes only one connection string


jseldess:
In our [Performance Benchmarking](https://www.cockroachlabs.com/docs/v2.1/performance-benchmarking-with-tpc-c.html#step-3-load-data-for-the-benchmark) tutorial, the `workload load tpcc` step for a small cluster is correct, using the connection URL for only one node. The corresponding step for a large cluster, however, shows using multiple connection URLs. This is wrong. The `load` command should point to a single node (we distributed the loading across nodes server-side, like in a `RESTORE`).

PR #4416, opened 2019-02-25
Update Production Checklist

### LINKED ISSUE #4301: Clarify minimum requirements: Production Checklist Doc Update
Re: [Production Checklist](https://cockroachlabs.com/docs//v2.1/recommended-production-settings.html)

## Issue Description
We talk about min CPU requirements. Do we mean 2 scheduling units (ie vCPUs/hyperthreads) or do we actually mean two physical cores (in which case we may have 4 vCPUs as the minimum)?

## Suggested Resolution 
Please clarify the minimum requirements. 

cc @jseldess

lhirata:
- Clarify that we are talking about vCPU instead of CPU.
- Fix connection string info in the workload instructions of benchmarking a large cluster.

Closes #4301, #4304, #4287.

### DIFF
--- a/v19.1/performance-benchmarking-with-tpc-c.md
+++ b/v19.1/performance-benchmarking-with-tpc-c.md
@@ -261,7 +261,7 @@ CockroachDB offers a pre-built `workload` binary for Linux that includes several
     ~~~ shell
     $  ./workload.LATEST fixtures load tpcc \
     --warehouses=10000 \
-    "postgres://root@<node1 address>?sslmode=disable postgres://root@<node2 address>?sslmode=disable postgres://root@<node3 address>?sslmode=disable [...space separated list]"
+    "postgres://root@<node1 address>?sslmode=disable"
     ~~~
 
     This command runs the TPC-C workload against the cluster. This will take at about an hour and loads 10,000 "warehouses" of data.
--- a/v19.1/recommended-production-settings.md
+++ b/v19.1/recommended-production-settings.md
@@ -43,15 +43,9 @@ For added context about CockroachDB's fault tolerance and automated repair capab
 
 ## Hardware
 
-### Terminology
-
-To properly select your cluster's hardware, it's important to review define basic hardware terminology:
-
-Term | Definition
------|------------
-**CPU** | The [physical central processing unit (CPU)](https://en.wikipedia.org/wiki/Central_processing_unit) that runs a CockroachDB server run on a (virtual) machine.</br>Also known as: central processor, main processor
-**vCPU** | Four virtual central processing units (vCPU) comprise a CPU. vCPUs increase a core's performance by running multiple tasks simultaneously. </br>Also known as: hyperthread, scheduling unit
-**Core** | A core is the basic unit for processing instructions in a CPU. A rule of thumb for cloud providers is that there are typically 2 vCPUs per core.
+{{site.data.alerts.callout_info}}
+Mentions of "CPU resources" refer to vCPUs, which are also known as hyperthreads.
+{{site.data.alerts.end}}
 
 ### Basic hardware recommendations
 
--- a/v19.1/recommended-production-settings.md
+++ b/v19.1/recommended-production-settings.md
@@ -63,7 +63,7 @@ Nodes should have sufficient CPU, RAM, network, and storage capacity to handle y
 
 - The ideal configuration is 4-16 vCPUs, 8-64 GB memory nodes (2-4 GB of memory per vCPU).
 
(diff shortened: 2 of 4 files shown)


jseldess:
___
*[v2.2/recommended-production-settings.md, line 131 at r2](https://reviewable.io/reviews/cockroachdb/docs/4416#-LZa7qUg7WlXx-XTi7zY:-LZa7qUg7WlXx-XTi7zZ:b6v78b1) ([raw file](https://github.com/cockroachdb/docs/blob/9aaa9bc40cbacde5031efa94e3418eb1f021dcd0/v2.2/recommended-production-settings.md#L131)):*
> ```Markdown
> 
>     {{site.data.alerts.callout_danger}}
>     Do not use `f1` or `g1` [shared-core machines](https://cloud.google.com/compute/docs/machine-types#sharedcore), which limit the load on vCPU resources.
> ```

@bdarnell, does it make sense to use `vCPU` over `CPU` in cases like this?

jseldess:
LGTM, but I'd like @bdarnell to confirm that it's ok to use vCPU so liberally. Also, there's an overlapping issue that you might as well tackle while you at this: https://github.com/cockroachdb/docs/issues/4304.

jseldess:
OK, thanks, @bdarnell. @lhirata, can you take another pass at this with Ben's guidance in mind?

v19.1/recommended-production-settings.md
```diff
@@ -43,25 +43,35 @@ For added context about CockroachDB's fault tolerance and automated repair capab
 
 ## Hardware
 
+### Terminology
```
lhirata: @bdarnell - can you review this new section?

v19.1/performance-benchmarking-with-tpc-c.md
```diff
@@ -261,7 +261,7 @@ CockroachDB offers a pre-built `workload` binary for Linux that includes several
     ~~~ shell
     $  ./workload.LATEST fixtures load tpcc \
     --warehouses=10000 \
-    "postgres://root@<node1 address>?sslmode=disable postgres://root@<node2 address>?sslmode=disable postgres://root@<node3 address>?sslmode=disable [...space separated list]"
+    "postgres://root@<node1 address>?sslmode=disable"
```
bdarnell: Don't you still want to include all the nodes in the tpcc command line?

v19.1/recommended-production-settings.md
```diff
@@ -43,25 +43,35 @@ For added context about CockroachDB's fault tolerance and automated repair capab
...
+
+To properly select your cluster's hardware, it's important to review define basic hardware terminology:
+
+Term | Definition
+-----|------------
+**CPU** | The [physical central processing unit (CPU)](https://en.wikipedia.org/wiki/Central_processing_unit) that runs a CockroachDB server run on a (virtual) machine.</br>Also known as: central processor, main processor
+**vCPU** | Four virtual central processing units (vCPU) comprise a CPU. vCPUs increase a core's performance by running multiple tasks simultaneously. </br>Also known as: hyperthread, scheduling unit
```
bdarnell: A CPU can have any number of vCPUs, not always 4. 

Hyperthreading increases a core's performance relative to non-hyperthreaded cores, although whether there is hyperthreading or not there are still vCPUs. It's just a question of whether there is more than one vCPU per core.

v19.1/recommended-production-settings.md
```diff
@@ -43,25 +43,35 @@ For added context about CockroachDB's fault tolerance and automated repair capab
...
+**Core** | A core is the basic unit for processing instructions in a CPU. A rule of thumb for cloud providers is that there are typically 2 vCPUs per core.
```
bdarnell: Not just cloud providers. Most server-class processors these days have two vCPUs per core.

v19.1/recommended-production-settings.md
bdarnell: I don't know that we need to define CPU or Core here, because we don't really use these terms. We mainly just need to say that when we talk about CPU resources we're talking about vCPUs, which are also known as hyperthreads.

lhirata: We mention cores later but if you think this is overkill, I can just include this info as a note.

v19.1/performance-benchmarking-with-tpc-c.md
lhirata: Based on the info in #4287, the load command should point to a single node, not a list. Am I misunderstanding?

bdarnell: Ah, got it. `fixtures load` just wants one URL but the `run` part uses more than one.

v19.1/recommended-production-settings.md
bdarnell: Where do we mention cores? The only place I see is in reference to "shared-core" machines which link to cloud-specific docs, which seems fine. We want to avoid using the word "core" anywhere (except to say that in most cases there are two vCPUs per core).

lhirata: Just the mention of "shared-core", so will move this to a note instead.

PR #4419, opened 2019-02-25
Update EXPORT

### LINKED ISSUE #3563: EXPORT docs notes from QA
* `EXPORT` doesn't show up in jobs -- it shows up in `SHOW QUERIES` and can be cancelled like other queries. Some day we might make jobs for them, but for now we need to let users know where to look.

From https://github.com/cockroachdb/cockroach/issues/25223:

> * `The `nullas` docs have a typo and read as if they’re still talking about IMPORT

> * Not a bug, but unicode emoji aren’t valid delimiters and that makes me sad :(

We should mention that only valid ascii chars can be used as csv delimiters.

> * The documentation of the GCS parameters doesn’t make sense, and appears incorrect.

>   * It equates not specifying AUTH with setting AUTH=default, which isn’t true in practice:
> ```
> root@:26257/defaultdb> EXPORT INTO CSV 'gs://cockroach-alex/csv' WITH delimiter=e'\u03B1' FROM SELECT * from startrek.quotes;
> pq: write to google cloud: googleapi: Error 403: 21965078311-compute@developer.gserviceaccount.com does not have storage.objects.create access to cockroach-alex/csv/n1.0.csv., forbidden
> root@:26257/defaultdb> EXPORT INTO CSV 'gs://cockroach-alex/csv?AUTH=default' WITH delimiter=e'\u03B1' FROM SELECT * from startrek.quotes;
> pq: expected settings value for cloudstorage.gs.default.key
> root@:26257/defaultdb> EXPORT INTO CSV 'gs://cockroach-alex/csv?AUTH=implicit' WITH delimiter=e'\u03B1' FROM SELECT * from startrek.quotes;
> pq: write to google cloud: googleapi: Error 403: 21965078311-compute@developer.gserviceaccount.com does not have storage.objects.create access to cockroach-alex/csv/n1.0.csv., forbidden
> ```

* Should mention that it is important to pick nullas values unlikely to appear in the exported data to avoid collusions.

lhirata:
- Update delimiter info to accept only ASCII
- Add `nullas` recommendation
- Add how to view exports
- Add how to delete exports

Closes #3563.

### DIFF
--- a/v2.1/export.md
+++ b/v2.1/export.md
@@ -55,7 +55,7 @@ You can control the [`EXPORT`](export.html) process's behavior using any of the 
 
 #### `delimiter`
 
-If not using comma as your column delimiter, you can specify another Unicode character as the delimiter.
+If not using comma as your column delimiter, you can specify another ASCII character as the delimiter.
 
 <table>
 	<tbody>
@@ -69,7 +69,7 @@ If not using comma as your column delimiter, you can specify another Unicode cha
 		</tr>
 		<tr>
 			<td><strong>Value</strong></td>
-			<td>The unicode character that delimits columns in your rows</td>
+			<td>The ASCII character that delimits columns in your rows.</td>
 		</tr>
 		<tr>
 			<td><strong>Example</strong></td>
@@ -94,7 +94,7 @@ Convert SQL *NULL* values so they match the specified string.
 		</tr>
 		<tr>
 			<td><strong>Value</strong></td>
-			<td>The string that should be used to represent <em>NULL</em> values</td>
+			<td>The string that should be used to represent <em>NULL</em> values. To avoid collisions, it is important to pick <code>nullas</code> values that does not appear in the exported data.</td>
 		</tr>
 		<tr>
 			<td><strong>Example</strong></td>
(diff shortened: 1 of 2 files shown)


v2.1/export.md
```diff
@@ -130,6 +130,24 @@ Convert SQL *NULL* values so they match the specified string.
 $ cockroach sql -e "SELECT * from bank.customers WHERE id>=100;" --format=csv > my.csv
 ~~~
 
+### View an export
```
dt: a *running* export

unlike jobs, show queries is just while it is running.

ditto below.

v2.1/export.md
```diff
@@ -94,7 +94,7 @@ Convert SQL *NULL* values so they match the specified string.
 		</tr>
 		<tr>
 			<td><strong>Value</strong></td>
-			<td>The string that should be used to represent <em>NULL</em> values</td>
+			<td>The string that should be used to represent <em>NULL</em> values. It is important to pick `nullas` values unlikely to appear in the exported data to avoid collusions.</td>
```
dt: I'm deeply worried about collusion, but right here I think it is collisions I'm more worried about. 

Also, I'd probably say we don't want to soften it with "unlikely". It is important to pick a `nullas` values that does not appear in the exported data if is important to distinguish it later.

Amruta-Ranade: nit: `nullas` should be formatted as `<code> </code>`

v2.2/export.md
```diff
@@ -94,7 +94,7 @@ Convert SQL *NULL* values so they match the specified string.
 		</tr>
 		<tr>
 			<td><strong>Value</strong></td>
-			<td>The string that should be used to represent <em>NULL</em> values</td>
+			<td>The string that should be used to represent <em>NULL</em> values. It is important to pick `nullas` values unlikely to appear in the exported data to avoid collusions.</td>
 		</tr>
```
Amruta-Ranade: Same as above

PR #4421, opened 2019-02-25
describe differences between bytes and strings

### LINKED ISSUE #609: sql: Clarify how BYTES and STRING types differ
Jesse:

> It looks like the `string` and `bytes` types are identical. Each seems to accept the same inputs. And Matt Jibson tells me the internal format is the same for them, too. Do you happen to know why we've kept them as distinct types instead of making one an alias for the other?

@tamird:

> They’re not identical. In particular, some functions behave differently for `string` and `bytes`. For example `LEFT(<thing of type bytes>, 5)` will return the 5 leftmost bytes, while `LEFT(<thing of type string>, 5)` will return the 5 leftmost characters. The result is the same if the string is comprised only of ASCII characters, but not if it contains 2,3, or 4-byte characters.

sploiselle:
closes #609

@knz Can you take a quick look at this? Hope I didn't butcher your explanation!

### DIFF
--- a/v19.1/bytes.md
+++ b/v19.1/bytes.md
@@ -86,6 +86,41 @@ odd number of them. Two conversion modes are supported:
 - Otherwise, the string is converted to a byte array that contains its
   UTF-8 encoding.
 
+### `STRING` vs. `BYTES`
+
+While both `STRING` and `BYTES` can appear to have similar behavior in many situations, one should understand their nuance before casting one into the other.
+
+`STRING` treats all of its data as characters, or more specificially, Unicode code points. `BYTES` treats all of its data as a byte string. This difference in implementation can lead to dramatically different behavior. For example, let's take a complex Unicode character such as ☃ ([the snowman emoji](https://emojipedia.org/snowman/)):
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT length('☃'::string);
+~~~
+
+~~~
+  length
++--------+
+       1
+~~~
+
+~~~ sql
+> SELECT length('☃'::bytes);
+~~~
+~~~
+  length
++--------+
+       3
+~~~
+
+In this case, [`LENGTH(string)`](functions-and-operators.html#string-and-byte-functions) measures the number of Unicode code points present in the string, whereas [`LENGTH(bytes)`](functions-and-operators.html#string-and-byte-functions) measures the number of bytes required to store that value. Each character (or Unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
+
+#### Translating literals to `STRING` vs. `BYTES`
+
+A literal entered through a SQL client will be translated into a different value based on the type:
+
++ `BYTES` give a special meaning to the pair `\x` at the beginning, and translates the rest by substituting pairs of hexadecimal digits to a single byte. For example, `\xff` is equivalent to a single byte with the value of 255. For more information, see [SQL Constants: String literals with character escapes](sql-constants.html#string-literals-with-character-escapes).
++ `STRING` does not give a special meaning to `\x`, so all characters are treated as distinct Unicode code points. For example, `\xff` is treated as a `STRING` with length 4 (`\`, `x`, `f`, and `f`).
+
 ## See also
 
 [Data Types](data-types.html)
--- a/v2.1/bytes.md
+++ b/v2.1/bytes.md
@@ -86,6 +86,41 @@ odd number of them. Two conversion modes are supported:
 - Otherwise, the string is converted to a byte array that contains
   its UTF-8 encoding.
 
+### `STRING` vs. `BYTES`
+
+While both `STRING` and `BYTES` can appear to have similar behavior in many situations, one should understand their nuance before casting one into the other.
+
+`STRING` treats all of its data as characters, or more specificially, Unicode code points. `BYTES` treats all of its data as a byte string. This difference in implementation can lead to dramatically different behavior. For example, let's take a complex Unicode character such as ☃ ([the snowman emoji](https://emojipedia.org/snowman/)):
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT length('☃'::string);
+~~~
+
+~~~
+  length
++--------+
+       1
+~~~
+
+~~~ sql
+> SELECT length('☃'::bytes);
+~~~
+~~~
+  length
++--------+
+       3
+~~~
+
+In this case, [`LENGTH(string)`](functions-and-operators.html#string-and-byte-functions) measures the number of Unicode code points present in the string, whereas [`LENGTH(bytes)`](functions-and-operators.html#string-and-byte-functions) measures the number of bytes required to store that value. Each character (or Unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
+
+#### Translating literals to `STRING` vs. `BYTES`
+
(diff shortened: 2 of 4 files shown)


v2.1/bytes.md
```diff
@@ -66,6 +66,33 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+~~~
+~~~
++-----------+-----------+
+| length(x) | length(y) |
++-----------+-----------+
+|         4 |         1 |
++-----------+-----------+
+~~~
+
+In this case, `LENGTH(bytes)` measures the number of bytes; whereas `LENGTH(string)` measures the number of unicode code points present in the string. Each character (or unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
```
knz: This is conflating multiple steps together unfortunately, to the point it becomes (slightly) inaccurate.

Maybe start with this example:

```
root@127.1.169.208:44415/defaultdb> select length('☃'::string);
  length
+--------+
       1
(1 row)


root@127.1.169.208:44415/defaultdb> select length('☃'::bytes);
  length
+--------+
       3
(1 row)

```

This will support your argument above.

Then *after* you make that argument, you should also talk about the process of translating a *literal* (what the client enters in the SQL syntax) to a STRING or BYTES *value* inside CockroachDB.

There, again different rules apply:

- BYTES will give a special meaning to the pair `\x` at the beginning, and translate the rest by substituting pairs of hexadecimal digits to a single byte. This is why `\xff` gives just 1 byte (with value 255).
- STRING does not give a special meaning to `\x` and just keeps every character (including `\` and `x`) in the result. This is why `\xff` gives 4 characters (`\`, `x`, `f`, `f`).

sploiselle:
@rmloveland Can I get a copy edit on this? Same content in each file, thankfully.

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
 `STRING` values can be cast explicitly to `BYTES`. This conversion
 always succeeds.
 
+### `STRING` vs. `BYTES`
+
+While both `STRING` and `BYTES` can appear to have similar behavior in many situations, one should understand their nuance before casting one into the other.
+
+`STRING` treats all of its data as characters, or more specificially, unicode code points. `BYTES` treats all of its data as a byte string. This difference in implementation can lead to dramatically different behavior. For example, let's take a complex ASCII character such as ☃ ([the snowman emoji](https://emojipedia.org/snowman/)):
```
rmloveland: sp: "specifically"

I don't understand, what is the "complex ASCII character ⛄️" This doesn't seem to be ASCII at all but rather Unicode (per https://emojipedia.org/snowman/)?  (Apologies if this was already discussed in engineering review, but I would expect some users to also find this confusing.)

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+
+{% include copy-clipboard.html %}
+~~~ sql
+SELECT length('☃'::string);
```
rmloveland: Nit: We usually have a `>` prompt char in front of SQL statements.

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+  length
++--------+
+       3
+~~~
+
+In this case, `LENGTH(string)` measures the number of unicode code points present in the string, where as `LENGTH(bytes)` measures the number of bytes required to store that value. Each character (or unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
+
+#### Translating literals to `STRING` vs. `BYTES`
+
+A literals entered through a SQL client will be translated into a different value based on the type:
```
rmloveland: I think this should be singular: "A literal entered ..."

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+~~~ sql
+SELECT length('☃'::bytes);
+~~~
+~~~
+  length
++--------+
+       3
+~~~
+
+In this case, `LENGTH(string)` measures the number of unicode code points present in the string, where as `LENGTH(bytes)` measures the number of bytes required to store that value. Each character (or unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
```
rmloveland: - Suggest linking to `LENGTH` built-in if possible.
- We should capitalize "Unicode" wherever it's used, as it's a proper noun.
- `s/where as/whereas/`
- We could link to more information about [code points](https://en.wikipedia.org/wiki/Code_point) but IDK if it's really necessary.

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+
+#### Translating literals to `STRING` vs. `BYTES`
+
+A literals entered through a SQL client will be translated into a different value based on the type:
+
++ `BYTES` give a special meaning to the pair `\x` at the beginning, and translate the rest by substituting pairs of hexadecimal digits to a single byte. For example, `\xff` is equivalent to a single byte with the value of 255.
++ `STRING` does not give a special meaning to `\x`, so all characters are treated as distinct unicode code points. For example, `\xff` is treated as a `STRING` with length 4 (`\`, `x`, `f`, and `f`).
```
rmloveland: Same comment as above re: caps for Unicode.

v19.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
...
+       3
+~~~
+
+In this case, `LENGTH(string)` measures the number of unicode code points present in the string, where as `LENGTH(bytes)` measures the number of bytes required to store that value. Each character (or unicode code point) can be encoded using multiple bytes, hence the difference in output between the two.
+
+#### Translating literals to `STRING` vs. `BYTES`
+
+A literals entered through a SQL client will be translated into a different value based on the type:
+
++ `BYTES` give a special meaning to the pair `\x` at the beginning, and translate the rest by substituting pairs of hexadecimal digits to a single byte. For example, `\xff` is equivalent to a single byte with the value of 255.
```
rmloveland: I think 'translate' should be plural here, i.e., "`BYTES` ... translates the rest ..."

knz:
LGTM

The explanation of the difference of interpretation of `\x` should link to the page on string and byte literals: https://www.cockroachlabs.com/docs/v2.1/sql-constants.html

v2.1/bytes.md
```diff
@@ -66,6 +66,41 @@ valid UTF-8 byte sequences; an error is reported otherwise.
 `STRING` values can be cast explicitly to `BYTES`. This conversion
 always succeeds.
 
+### `STRING` vs. `BYTES`
+
+While both `STRING` and `BYTES` can appear to have similar behavior in many situations, one should understand their nuance before casting one into the other.
+
+`STRING` treats all of its data as characters, or more specificially, Unicode code points. `BYTES` treats all of its data as a byte string. This difference in implementation can lead to dramatically different behavior. For example, let's take a complex Unicde character such as ☃ ([the snowman emoji](https://emojipedia.org/snowman/)):
```
rmloveland: Final typo: `s/Unicde/Unicode/`

Thanks for the updates!

jseldess:
@sploiselle, is this ready to merge?

Issue #791, opened 2016-10-27
sql: support full HLC precision timestamps in AS OF #9934


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/9934

As mentioned in `beta-2016-10-27` release notes: 

> `AS OF SYSTEM TIME` queries now support the decimal format returned by `cluster_logical_timestamp()` for maximum precision.

sploiselle:
Feature already documented.

PR #4431, opened 2019-02-25
describe decimal support for AOST

sploiselle:
@mjibson Are you a good resource to make sure this is technically sound/good?

Closest #791


mjibson:
We already document decimal handling at the end of the page (https://www.cockroachlabs.com/docs/v2.1/as-of-system-time.html#tech-note). We did it there instead of above because we in general don't want people to use that method. I think we can close that issue without any changes. Unless there's a reason to move the existing note up. I'm not sure if customers have been wanting it more visible.

sploiselle:
Closing per Matt's comment above.

PR #4432, opened 2019-02-25
Remove nanoseconds from INTERVAL docs

### LINKED ISSUE #4375: sql: remove nanoseconds from INTERVAL
PR: https://github.com/cockroachdb/cockroach/pull/34202

From release notes:
> [`INT`](../v2.2/int.html) values are now stored with microsecond precision instead of nanoseconds. Existing intervals with nanoseconds are no longer able to return their nanosecond part. An existing table `t` with nanoseconds in intervals of column `s` can round them to the nearest microsecond with `UPDATE t SET s = s + '0s'`. Note that this could potentially cause uniqueness problems if the interval is a primary key. [#34202][#34202] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #4375.

Summary of changes:

- Update `INTERVAL` docs with a new section on 'Precision', where we
  note that as of 19.1, nanoseconds are going away in favor of
  microseconds only

- Remove description of Golang-ish interval syntax which mentioned
  nanoseconds

- s/nano/micro/g (which may not have made sense in all cases, pending
  engineering review).

- Add new 'aliases' section to match other data types, where we note
  that you can specify intervals with string literals in some contexts.

- Move information about internal representation to the 'Size' section.

- Update Examples to use new output format.

### DIFF
--- a/v19.1/interval.md
+++ b/v19.1/interval.md
@@ -6,6 +6,9 @@ toc: true
 
 The `INTERVAL` [data type](data-types.html) stores a value that represents a span of time.
 
+## Aliases
+
+There are no aliases for the interval type. However, CockroachDB supports using uninterpreted [string literals](sql-constants.html#string-literals) in contexts where an `INTERVAL` value is otherwise expected.
 
 ## Syntax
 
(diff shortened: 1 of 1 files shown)


rmloveland:
Direct link to updated page is here: [INTERVAL](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/2b72d20560845e5b3596cf896ba47c61e35eb240/dev/interval.html).

PR #4434, opened 2019-02-26
Mark 1.0 as unsupported and 1.1 as soon-unsupported in docs

### LINKED ISSUE #4430: Mark 1.0 as unsupported in docs
@bdarnell says we support 2 version prior to stable. So currently that means 2.1, 2.0, and 1.1. 
For all earlier versions, we should put a disclaimer/warning all all pages.

jseldess:
Fixes #4430.

screen shot 2019-02-25 at 10 49 13 pm

### DIFF
--- a/_includes/unsupported-version-soon.md
+++ b/_includes/unsupported-version-soon.md
@@ -1,3 +1,3 @@
 {{site.data.alerts.callout_info}}
-Cockroach Labs supports the current <a href="../stable/install-cockroachdb.html">stable release</a> and two releases prior. Therefore, this version will no longer be supported after the Spring 2019 release.
+Cockroach Labs supports the current <a href="https://www.cockroachlabs.com/docs/stable/install-cockroachdb-mac.html">stable release</a> and two releases prior. Therefore, this version will no longer be supported after the Spring 2019 release.
 {{site.data.alerts.end}}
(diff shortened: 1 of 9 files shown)


jseldess:
@bdarnell, is this warning sufficient, or do we need to be more explicit about our rule of supporting 2 version prior to stable?

_includes/v1.0/misc/unsupported-version.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
+This version of CockroachDB is no longer supported. Please use a more recent version.
```
sploiselle: I think that including a link here to the install/download page for `/stable` would feel like we were trying to help users who're trying to use these docs.

bdarnell:
> is this warning sufficient, or do we need to be more explicit about our rule of supporting 2 version prior to stable?

It would be nice to document the support policy somewhere, but that can be a separate change. (I'm not sure where we'd put it currently)

Maybe we should add a warning to 1.1 too. Not quite as scary (yellow instead of red), but say that 1.1 will no longer be supported after the spring 2019 release.

_includes/v1.0/misc/unsupported-version.md
jseldess: Done.

jseldess:
@bdarnell, updated the unsupported callout and added a soon-unsupported callout to 1.1 docs. PTAL.

Issue #4255, opened 2019-01-15
sql: add validation to cascading zone configs


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/32861

From release notes:
> When creating a [replication zone](../v2.2/configure-replication-zones.html), if a field is set to `COPY FROM PARENT`, the field now inherits its value from its parent zone, but any change to the field in the parent zone no longer affects the child zone. [#32861][#32861] {% comment %}doc{% endcomment %}

PR #4446, opened 2019-02-28
Update Zone Config docs

### LINKED ISSUE #4062: sql, config: make zone configs cascade
PR: https://github.com/cockroachdb/cockroach/pull/30611

From release notes:
> Updated the output of [`SHOW ZONE CONFIGURATIONS`](../v2.2/show-zone-configurations.html). Also, unset fields in [zone configurations](../v2.2/configure-replication-zones.html) now inherit parent values. [#30611][#30611] {% comment %}doc{% endcomment %}

lhirata:
- Change warning to info callout. Changes to a zone will now cascade
- Add new COPY FROM PARENT value that allows a child zone to copy values 
from a parent zone. When this is used, changes will no longer cascade
- Update SQL diagrams

Closes #4062, #4255.


knz:
Yes I think your explanation is valid.

Just be minfdul that `.default` is overall less important. New zones inherit from their *parent* which is not necessarily `.default` (eg a partition zone inherits from the table).

There are two possible "values" stored in a zone:
- value is **absent**: the parent zone applies, or the grandparent if the parent setting is also absent, etc, up to `.default`
- value is **present**: the value applies.

To achieve these two cases you can use the following **methods**:

- Value becomes **present**:
  - when set explicitly (SET ...)
  - when COPY FROM PARENT is used during zone creation
  - when COPY FROM PARENT is used after zone creation

- Value becomes **absent**:
  - when omitted during zone creation
  - when DROP is used
  - (not documented, preserved for compatibility with 2.0/2.1) when value set to NULL 


Maybe this presentation (with a table) could be a good complement to the notes already added in the PR.

_includes/v19.1/known-limitations/system-range-replication.md
```diff
@@ -1 +1 @@
...
+Changes to the [`.default` cluster-wide replication zone](configure-replication-zones.html#edit-the-default-replication-zone) are automatically applied to existing replication zones, including pre-configured zones for important system ranges that must remain available for the cluster as a whole to remain available. The zones for these system ranges have an initial replication factor of 5 to make them more resilient to node failure. However, if you increase the `.default` zone's replication factor above 5, consider [increasing the replication factor for important system ranges](configure-replication-zones.html#create-a-replication-zone-for-a-system-range) as well.
```
bdarnell: This is unclear - the second sentence is an exception to the first (and it's not the only one - if you change `range_{min,max}_bytes` or `gc.ttl`, those also won't be applied to the system zones because they have default values for those fields).

Do we really need this callout any more? It was important when even system zones defaulted to 3 replicas. Now that they default to 5, it's saying that if you set your default replication factor to 7 or more you may need to do the same to the system zones, but we don't really recommend running with more than 5 replicas. 

I think there should be a mention on the main page for zone config docs that says that the system zones have their own zone configs that may override settings from `.default`, but we don't need to copy this warning everywhere zone configs are mentioned.

_includes/v2.1/known-limitations/system-range-replication.md
```diff
@@ -1 +1 @@
...
+<span class="version-tag">New in v2.1:</span> Changes to the [`.default` cluster-wide replication zone](configure-replication-zones.html#edit-the-default-replication-zone) are automatically applied to existing replication zones, including pre-configured zones for important system ranges that must remain available for the cluster as a whole to remain available. The zones for these system ranges have an initial replication factor of 5 to make them more resilient to node failure. However, if you increase the `.default` zone's replication factor above 5, consider [increasing the replication factor for important system ranges](configure-replication-zones.html#create-a-replication-zone-for-a-system-range) as well.
```
bdarnell: I don't think this is true - as far as I can see cascading zone configs are new in 19.1.

lhirata:
@jseldess + @knz and/or @bdarnell, can you do another pass on this please?

lhirata:
TFTRs!

PR #4451, opened 2019-03-02
Use script to add client-side redirects to prevent search result 404s

### LINKED ISSUE #4450: v2.2 search results are 404ing
After renaming v2.2 to v19.1, google search results
still contain links to docs with v2.2 in the url, and those links
are 404ing.

jseldess:
We recently renamed v2.2 to v19.1. However, google search results
still contain links to docs with v2.2 in the url, and those links
are 404ing. To prevent this, this PR adds client-side redirects
from v2.2 to v19.1 via a script. 

Fixes #4450.


bdarnell:
The problem with this is that the 2.2 URLs will all continue to exist as a separate copy of the site, so you could continue to see both 2.2 and 19.1 coexisting in search results indefinitely. I think we should either do nothing (leave the broken links as they are now and wait for google to sort it out) or do a proper redirect. 

I think we could also make google more responsive to things like this if our sitemap was better and included last-modified timestamps for each file.

jseldess:
Good points, @bdarnell. I've used a script to add client-side redirects instead and amended the PR description. PTAL.

jseldess:
@bdarnell, I'm seeing problems, so please hold off for now.

jseldess:
@bdarnell, opened a separate PR to do this via server-side redirects: https://github.com/cockroachdb/docs/pull/4464.

jseldess:
Closing in favor of #4464.

PR #4402, opened 2019-02-20
Add read in uncertainty interval to common errors

rmloveland:
Fixes #4401.

Summary of changes:

- Update 'Common Errors' page with a new 'read within uncertainty
  interval' section, which describes the error and how to avoid it

- Link from the above to other relevant docs (e.g., architecture docs on
  txn conflicts, AOST, etc.)

Issue #4401, opened 2019-02-20
Add read within uncertainty interval to common errors


rmloveland:
We should add `ReadWithinUncertaintyIntervalError` to our list of [Common errors](https://www.cockroachlabs.com/docs/v2.2/common-errors.html) so that users who encounter this error can find it via search.

Estimated scope of work:

- Update 'Common Errors' page with a new 'read within uncertainty interval' section, which describes the error and how to avoid it
- Link from the above to other relevant docs (e.g., architecture docs on txn conflicts, AOST, etc.)

See also:

- [cockroach#34517](https://github.com/cockroachdb/cockroach/issues/34517)
- #1775

PR #4456, opened 2019-03-04
Add read within uncertainty interval to 19.1 docs

rmloveland:
Follow-up to #4402.

Finishes off #4401.


rmloveland:
Jesse you are right!  It was in #4401 

Apologies for making the crazy.  Will just merge this.

> Are these placeholders for future work?

Yes, different retry error types I "overheard" mentioned on Slack.

PR #4462, opened 2019-03-05
Update PG dialect in Hibernate config examples

### LINKED ISSUE #4458: org.hibernate.dialect.PostgreSQLDialect is used in the hibernate example but is deprecated
Re: [Build a Java App with CockroachDB](https://cockroachlabs.com/docs//v19.1/build-a-java-app-with-cockroachdb-hibernate.html)

## Issue Description

In the hibernate.cfg.xml we specify `org.hibernate.dialect.PostgreSQLDialect` This has been deprecated.

## Suggested Resolution 

`org.hibernate.dialect.PostgreSQL95Dialect` should be used as CRDB advertises to clients that it is PostgreSQL 9.5 compliant.

rmloveland:
Fixes #4458.

Specifically, update all uses of the PG dialect variable in
`hibernate.cfg.xml` for the following versions:

- 2.1
- 19.1


rmloveland:
Jesse, FYI this change is a pretty straightforward text-replacement AFAICT.  I tested it locally with a Hibernate test app and it works fine.

Considered just merging it, but thought it might be better to get your eye on it to avoid any obvious errors I might have missed.

rmloveland:
Thanks for the review Jesse!

PR #4464, opened 2019-03-05
Server-side redirect of v2.2 urls to v19.1 via netlify

### LINKED ISSUE #4450: v2.2 search results are 404ing

jseldess:
This PR replaces https://github.com/cockroachdb/docs/pull/4451,
which used client-side redirects.

Fixes #4450.


jseldess:
Once approved, we'll need to merge this, wait for Netlify to update, and then test (and revert if there are any problems).

PR #4490, opened 2019-03-13
Added "Why are my deletes getting slower?" FAQ

tim-o:
Documenting @knz's explanation from a conversation with @gigatexal. This has come up a couple times with paying customers as well.

Also removed reference to gitter.


tim-o:
@jseldess feel free to delegate review. You were last editor here so didn't know who else to tag.

jseldess:
Thanks for this, @tim-o! I'll pass over to @rmloveland for full review. 

Some quick thoughts (for Rich, too):
- Need to port this to 19.1 faq as well.
- Probably want to directly mention the [`gc.ttlseconds` variable](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#replication-zone-variables). 
- Probably want to add related content to the `DELETE` docs.

rmloveland:
Thanks @tim-o for this contribution!  With your permission, I'd like to "steal" this PR and make some of the additional updates/edits mentioned by @jseldess.  Sound OK?

tim-o:
Yessir, go ham.

rmloveland:
> Yessir, go ham.

Cool.  Untagging other reviewers, will re-tag when ready.

jseldess:
@rmloveland, do you want me to take this over? I don't think there's much work to do to get this merged.

rmloveland:
Sorry @jseldess, kinda dropped this one.  Picking it back up now.

knz:
I am sorry I haven't reacted on this sooner. The content as-is is fine, however we may want to check whether the GC automatically picks up a lower TTL immediately. Last time I checked, it's necessary to restart the node after changing the value (although last time I checked was a long time ago, so this may have changed since then.)

rmloveland:
> we may want to check whether the GC automatically picks up a lower TTL immediately. Last time I checked, it's necessary to restart the node after changing the value 

It looks like a restart is not necessary, at least if `SHOW ZONE CONFIGURATIONS` is to be believed (see below).  Or is there another/better way to check for this?

```
root@localhost:26257/bank> CREATE TABLE kv (K INT, v INT);
CREATE TABLE

Time: 5.461ms

root@localhost:26257/bank> ALTER TABLE kv CONFIGURE ZONE USING gc.ttlseconds = 300;
CONFIGURE ZONE 1

Time: 8.051ms

root@localhost:26257/bank> SHOW ZONE CONFIGURATION FOR TABLE kv;
  zone_name |             config_sql               
+-----------+-------------------------------------+
  bank.kv   | ALTER TABLE kv CONFIGURE ZONE USING  
            |     range_min_bytes = 16777216,      
            |     range_max_bytes = 67108864,      
            |     gc.ttlseconds = 300,             
            |     num_replicas = 3,                
            |     constraints = '[]',              
            |     lease_preferences = '[]'         
(1 row)

Time: 1.69ms
```

knz:
SHOW only shows the configuration, it doesn't show what's actually running. The GC loop uses a timer expiration system. If you change the configuration to 5mn while it just started to wait for 25 hours, you may need to wait 24h+ before it notices the delay was decreased to 5mn.

The timer is reset when the node restarts, that's why I mentioned that.

This needs checking with either experiment or someone in the know. Things may have changed since I last studied this area.
-- 

knz:
@nvanbenschoten @ajwerner is my understanding correct? (see previous comment)

nvanbenschoten:
@knz I don't believe that is correct. The `replicaScanner` will run each Range through the `gcQueue`  once every 10 minutes, so a change in TTL will be picked up within that period.

rmloveland:
@knz given Nathan's response, are you OK with the content of this PR from a technical POV?

rmloveland:
Thanks @jseldess!

>  I'd move the main content from the general FAQ to either the Operational FAQs or the SQL FAQs. Not sure where it's best.

Moved it to the SQL FAQs.  Also took the opportunity to move another question about how to use CRDB as a key-value store to the SQL FAQs since it probably also belongs there.

> Also, we should probably call this out as a Bulk deletion best practices section in https://www.cockroachlabs.com/docs/stable/performance-best-practices-overview.html.

Done.  Added the new section *Bulk deletion best practices* with a link to the SQL FAQ entry.

> Might also want to add a callout/link to this DELETE example: https://www.cockroachlabs.com/docs/stable/delete.html#delete-specific-rows

Done.

### delete.md

>> For an explanation of why this happens, and for instructions showing how to iteratively delete rows in constant time, see [Why are my deletes getting slower over time?](frequently-asked-questions.html#why-are-my-deletes-getting-slower-over-time) from the FAQ.

> nit: I'd remove from the FAQ.

Done.

### frequently-asked-questions.md

> nit: Add comma after when you delete a row.

Done.

> I'd rephrase this:
>> the gc.ttlseconds zone configuration

> to:
>> the gc.ttlseconds variable in the applicable zone configuration

Done.

PS I applied all of the above changes to both the 19.1 and 19.2 docs

rmloveland:
Jesse Seldess <notifications@github.com> writes:

> v2.1/frequently-asked-questions.md, line 167 at r2 (raw file):
>
>     {% include {{ page.version.version }
>     }/faq/simulate-key-value-store.html %}
>
> ## Why are my deletes getting slower over time?
>
> Don't you want to move this to the sql faq?

Ah, forgot about 2.1.  It's so old!  Feels like years ago now. :-)

Fixed.

PR #4493, opened 2019-03-13
fix typos in default-value.md

lucy-zhang:
"vale" -> "value"


jseldess:
Thanks, @lucy-zhang. LGTM. Please just make same fix to the v19.1 version of the page.

PR #4497, opened 2019-03-14
SQL Dump Known Issue

tim-o:
Documenting that dump won't work without manual edits when interleaved tables are present.

I'm hoping this will avoid some consternation for someone trying to restore down the road. Better to inform them if basic functionality won't work as expected, especially if the fix is easy. This could free up the bulk IO team to work on things without a workaround.

@jseldess AFAIK we haven't actually referred to issues in the docs before, so let me know if this is alright with you.

### DIFF
--- a/_includes/v19.1/known-limitations/import-interleaved-table.md
+++ b/_includes/v19.1/known-limitations/import-interleaved-table.md
@@ -1,0 +1,1 @@
+After using [`cockroach dump`](sql-dump.html) to dump the schema and data of an interleaved table, the output must be edited before it can be imported via [`IMPORT`](import.html). See [#35462](https://github.com/cockroachdb/cockroach/issues/35462) for the workaround and more details.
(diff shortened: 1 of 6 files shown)


rolandcrosby:
Mitigates cockroachdb/cockroach#35462 (obviously; just wanted to link the issues)

jseldess:
This is great, @tim-o. Mind if I take it over to make some tweaks and apply to other versions? I think we should also add it to the known limitations page.

tim-o:
@jseldess sounds good to me

jdoupe:
#35462 references interleaved indexes, and the workaround specifically applies to them.  This (and existing) documentation refers to interleaved TABLES.  My understanding is that these two things are not quite the same, and the import doesn't explicitly handle either one.  And, furthermore, there doesn't seem to be a workaround for interleaved tables as it seems you can only create an interleaved table at creation time.

Please (please!) let me know if my understanding is incorrect.

rmloveland:
Hi @jdoupe !  Sorry for the slow response here.  Would you mind asking your question at https://forum.cockroachlabs.com?

I think it will get more expert eyes on it over there, and we're happy to make additional docs changes based on what comes to light from that discussion.

(For the record I don't know anything about this issue / how to answer your question but I don't want you to go unheeded - thanks for your patience)

PR #4501, opened 2019-03-14
Beta release draft for GSSAPI auth

Amruta-Ranade:
Will iterate on the content/info arch after Monday's beta.

### DIFF
--- a/v19.1/gssapi_authentication.md
+++ b/v19.1/gssapi_authentication.md
@@ -8,82 +8,96 @@ CockroachDB supports the Generic Security Services API (GSSAPI) with Kerberos au
 
 ## Configuring KDC
 
-This document assumes you have configured a Kerberos service principal name (SPN) for CockroachDB and generated a valid keytab file for CockroachDB as well.
+To use Kerberos authentication with CockroachDB, configure a Kerberos service principal name (SPN) for CockroachDB and generate a valid keytab file with the following specifications:
+
+- Set the SPN to `postgres`.
+- Create SPNs for all DNS addresses that a user would use to connect to your CockroachDB cluster (including any TCP load balancers between the user and the CockroachDB node) and ensure that the keytab contains the keys for every SPN you create.
 
 ## Configuring the CockroachDB node
 1. Copy the keytab file at a location accessible by the cockroach binary.
+
 2. [Create certificates](create-security-certificates.html) for internode and root user authentication:
 
-  {% include copy-clipboard.html %}
-  ~~~ shell
-  $ mkdir certs my-safe-directory
-  ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir certs my-safe-directory
+    ~~~
 
-  {% include copy-clipboard.html %}
-  ~~~ shell
-  $ cockroach cert create-ca \
-  --certs-dir=certs \
-  --ca-key=my-safe-directory/ca.key
-  ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach cert create-ca \
+    --certs-dir=certs \
+    --ca-key=my-safe-directory/ca.key
+    ~~~
 
-  {% include copy-clipboard.html %}
-  ~~~ shell
-  $ cockroach cert create-node \
-  localhost \
-  $(hostname) \
-  --certs-dir=certs \
-  --ca-key=my-safe-directory/ca.key
-  ~~~
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach cert create-node \
+    localhost \
+    $(hostname) \
+    --certs-dir=certs \
+    --ca-key=my-safe-directory/ca.key
+    ~~~
 
-  {% include copy-clipboard.html %}
-  ~~~ shell
-  $ cockroach cert create-client \
-  root \
-  --certs-dir=certs \
-  --ca-key=my-safe-directory/ca.key
-  ~~~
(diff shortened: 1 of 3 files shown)


v19.1/gssapi_authentication.md
```diff
@@ -0,0 +1,106 @@
...
+toc: true
+---
+
+CockroachDB supports the Generic Security Services API (GSSAPI) with Kerberos authentication as an [Enterprise feature](enterprise-licensing.html).
+
+## Configuring KDC
+
+To use Kerberos authentication with CockroachDB, configure a Kerberos service principal name (SPN) for CockroachDB and generate a valid keytab file with the following specifications:
+
+- Set the SPN to `postgres`.
```
mjibson: @roland Do you happen to know why this is required? I don't. When I changed our docker compose test to use `cockroach/` instead of `postgres/` it fails. We don't hard code the word `postgres` in our GSS code. So I *think* (?) that it must be coming from the psql client. If that's the case, then this step isn't actually required, and could instead be specifiable by whatever client they are using. This currently the only area where I'm not sure what's going on.

v19.1/gssapi_authentication.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Configuring the client
+
+{{site.data.alerts.callout_info}}
+The cockroach sql shell does not yet support GSSAPI authentication. You need to use a GSSAPI-compatible Postgres client, such as Postgres's `psql` client.
+{{site.data.alerts.end}}
+
+1. Install and configure your Kerberos client.
+2. Install the Postgres client (for example, postgresql-client-10 Debian package from postgresql.org).
+3. Get a Kerberos TGT from Active Directory using `kinit`.
```
mjibson: Should we remove "from Active Directory" here since they might not be using AD?

v19.1/gssapi_authentication.md
```diff
@@ -0,0 +1,106 @@
...
+      --listen-addr=0.0.0.0
+      ~~~
+
+5. Connect to CockroachDB as root using the root client certificate generated above:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach sql --certs-dir=certs
+    ~~~
+
+6. [Enable an enterprise license](enterprise-licensing.html#obtain-a-trial-or-enterprise-license-key).
```
mjibson: I'd include a note here too that they can skip this step, which will allow them to test GSS configuration and auth before having to have an enterprise license. We should list the expected error message in that case "GSS authentication requires an enterprise license", and that it means the GSS stuff worked correctly, all they need now is the license.

v19.1/gssapi_authentication.md
```diff
@@ -0,0 +1,106 @@
...
+
+To use Kerberos authentication with CockroachDB, configure a Kerberos service principal name (SPN) for CockroachDB and generate a valid keytab file with the following specifications:
+
+- Set the SPN to `postgres`.
+- Create SPNs for all DNS addresses that a user would use to connect to your CockroachDB cluster (including any TCP load balancers between the user and the CockroachDB node) and ensure that the keytab contains the keys for every SPN you create.
+
+## Configuring the CockroachDB node
+1. Copy the keytab file at a location accessible by the cockroach binary.
+
+2. [Create certificates](create-security-certificates.html) for internode and root user authentication:
```
mjibson: Does it make more sense to link to another part of the docs here, instead of duplicate how to create certs? I kind of suspect most people at this step already have their certs. Just a question, since you may already have a policy/decision about this.

v19.1/gssapi_authentication.md
rolandcrosby: It's specified by the client driver using the [`krbsrvname`](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-KRBSRVNAME) connection parameter. I believe any value should work as long as you specify the same value in the SPN and the connection string/URL. I wouldn't mind if we made it explicit that you can choose whatever you want and that `postgres` is just the default used by libpq, but it's not extremely important to me.

v19.1/gssapi_authentication.md
rolandcrosby: good catch, yes, "from the KDC" would be better.

v19.1/gssapi_authentication.md
mjibson: Thanks. We should probably change this to note that if you are testing with psql, it should be postgres, otherwise it can be whatever you configure.

v19.1/gssapi_authentication.md
mjibson: We built the feature specifically to allow for this use case so that people who are trying out cockroach can try out even the GSS stuff without a license, to determine if it'll work in their environment. That is, we had hoped it could be part of an early evaluation/POC phase and would assist in our sales process.

v19.1/gssapi_authentication.md
```diff
@@ -0,0 +1,106 @@
...
+      --listen-addr=0.0.0.0
+      ~~~
+
+5. Connect to CockroachDB as root using the root client certificate generated above:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach sql --certs-dir=certs
+    ~~~
+
+6. (Optional)[Enable an enterprise license](enterprise-licensing.html#obtain-a-trial-or-enterprise-license-key).
```
mjibson: Space after `(Optional)`?

Amruta-Ranade:
Merging this for now. Will iterate on the content next week.

PR #4503, opened 2019-03-14
Fix table partitioning docs

### LINKED ISSUE #4499: Define Table Partitions Doc Update
Re: [Define Table Partitions](https://cockroachlabs.com/docs//v2.1/partitioning.html)

## Issue Description
We should add an example of `SHOW ZONE CONFIGURATION FOR PARTITION` to this page. 
## Suggested Resolution

jseldess:
There were numerous problems in the partition by list example:

- It used only 2 datacenters, a bad pattern for CockroachDB.
- In each datacenter, there was only 1 node. This made partitioning
  non-functional because CockroachDB won't place more than one replica
  of a range on a single node.
- There were some bugs in the zone config settings.

This PR expands the example to 3 datacenters with 3 nodes in each. It also
converts the example to be runnable as a local cluster and adds
an example of how to view the zone config for a partition.

Fixes #4499.

Separately, we'll need to re-test and refresh the partition by range example and the sub-partition example.

### DIFF
--- a/_includes/v19.1/zone-configs/view-the-replication-zone-for-a-partition.md
+++ b/_includes/v19.1/zone-configs/view-the-replication-zone-for-a-partition.md
@@ -1,23 +1,16 @@
 {% include copy-clipboard.html %}
 ~~~ sql
-> SHOW ZONE CONFIGURATION FOR PARTITION north_america OF TABLE customers;
-~~~
-~~~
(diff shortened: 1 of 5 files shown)


jseldess:
TFTR, @rkruze. @Amruta-Ranade, I'll wait for your copy review as well.

PR #4508, opened 2019-03-16
Fix column name when selecting job by type.

mberhault:
As mentioned in the table above, the column is named `job_type`.


jseldess:
LGTM. Thanks, @mberhault. Would you apply this fix to the v19.1 version of the page as well, or would you like me to take this over?

mberhault:
Applied to v19.1 as well. PTAL.

PR #4537, opened 2019-03-19
Updated authentication.md

bartcone:
Multiple places referenced certs being signed by `ca.crt` which is not
correct. Updated to reference `ca.key` instead.

### DIFF
--- a/v2.1/authentication.md
+++ b/v2.1/authentication.md
@@ -76,7 +76,7 @@ A node must have the following files with file names as specified in the table:
 File name | File usage
 -------------|------------
 `ca.crt`     | CA certificate created using the `cockroach cert` command.
-`node.crt`   | Server certificate created using the `cockroach cert` command. <br><br> `node.crt` must have `CN=node` and the list of IP addresses and DNS names listed in `Subject Alternative Name` field. <br><br>Must be signed by `ca.crt`.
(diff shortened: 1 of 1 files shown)


mberhault:
CC: @Amruta-Ranade 

Thanks for the PR. I agree with you that `signed by <private key>` is more accurate. However, the aim of spelling out the certificate filename is to make it very clear which CA is responsible for which certificate. This becomes especially important in the `split CA certificates`.

We also care less about the CA key filename (the full path to the key is specified in the cockroach tooling) as opposed to the certificate filename (which is fixed).

An alternative phrasing would be: "... must be signed by the CA represented by `ca.crt`". While pedantic, it is accurate and keeps the emphasis on the specific CA identified by filename. The same would go for `client.node.crt` which was left unmodified.

Amruta-Ranade:
Thanks for the PR! Will look into this and get back to you.

jseldess:
@Amruta-Ranade, any updates here?

jseldess:
@Amruta-Ranade or @lhirata, can one of you take this?

jseldess:
@Amruta-Ranade, ping.

Amruta-Ranade:
Ah yes. Thanks for the reminder, Jesse.

Amruta-Ranade:
Addressed in #6313

PR #4565, opened 2019-03-22
CBO supports COLLATE expressions

### LINKED ISSUE #4178: opt: support COLLATE expressions
PR: https://github.com/cockroachdb/cockroach/pull/32500

From release notes:
> Queries involving `COLLATE` expressions are now supported by the cost-based optimizer. [#32500][#32500] {% comment %}doc{% endcomment %}

jseldess:
Fixes #4178.

### DIFF
--- a/v19.1/cost-based-optimizer.md
+++ b/v19.1/cost-based-optimizer.md
@@ -74,9 +74,10 @@ The cost-based optimizer supports most SQL statements. Specifically, the followi
 - `FILTER` clauses on [aggregate functions](functions-and-operators.html#aggregate-functions)
 - [Sequences](create-sequence.html)
 - [Views](views.html)
-- All [`SELECT`](select.html) statements that do not include window functions
-- All `UNION` statements that do not include window functions
-- All `VALUES` statements that do not include window functions
+- [`SELECT`](select.html) statements that do not include window functions
+- [`UNION`](selection-queries.html#union-combine-two-queries) clauses that do not include window functions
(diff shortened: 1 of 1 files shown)


jseldess:
@justinj, ping.

justinj:
I don't think `COLLATE` belongs in this list, `COLLATE` is a scalar expression more akin to things like `1 + 2`, the rest of these are statements.

justinj:
I don't think there should be any docs changes required for `COLLATE`.

jseldess:
@justinj, I was responding to this release note:

> Queries involving COLLATE expressions are now supported by the cost-based optimizer. [#32500][#32500] {% comment %}doc{% endcomment %}

Did I misunderstand?

justinj:
The release note is accurate, but the other things listed here aren't expressions, they're statements.

jseldess:
Well, there's also mention of `FILTER` clauses and Sequences, so it's not all statements. But if you don't think any updates are needed, I'll just close this.

justinj:
Ah, I didn't notice those, well the text above the list describes them as statements, so I would say those probably don't belong there either. Otherwise there is a very long list of features (basically all of SQL) that would need to be listed.

jseldess:
@rmloveland, I'll leave it to you to decide what to list here and what not. Perhaps it's only important to list what's NOT covered by CBO at this point?

rmloveland:
> @rmloveland, I'll leave it to you to decide what to list here and what not. Perhaps it's only important to list what's NOT covered by CBO at this point?

Good question, filed #4626 to address that (though probably after the 19.1 release).

PR #4567, opened 2019-03-22
Document tzdata dependency

### LINKED ISSUE #4433: Document `tzdata` as a dependency
Certain features of CockroachDB require the `tzdata` package (sometimes called `zoneinfo`) to be installed (specifically, the ability to use place names as time zone identifiers). We had thought this was installed by default in all major linux distributions but in https://github.com/cockroachdb/cockroach/issues/32415#issuecomment-467196857 an ubuntu user reported that it was missing. We should document this dependency (and recommend that it be kept up to date, since the time zone/DST rules change several times a year).

jseldess:
Fixes #4433.

### DIFF
--- a/v19.1/recommended-production-settings.md
+++ b/v19.1/recommended-production-settings.md
@@ -240,6 +240,22 @@ $ cockroach start --cache=.25 --max-sql-memory=.25 <other start flags>
 Avoid setting `--cache` and `--max-sql-memory` to a combined value of more than 75% of a machine's total RAM. Doing so increases the risk of memory-related failures.
 {{site.data.alerts.end}}
 
+## Dependencies
+
+The [CockroachDB binary for Linux](install-cockroachdb-linux.html) depends on the following libraries:
+
+Library | Description
+-----------|------------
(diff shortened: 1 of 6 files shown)


jseldess:
@bdarnell, is this sufficient, or do we need to explicitly call this out in the install pages and/or the deployment tutorials?

v19.1/recommended-production-settings.md
```diff
@@ -240,6 +240,10 @@ $ cockroach start --cache=.25 --max-sql-memory=.25 <other start flags>
 Avoid setting `--cache` and `--max-sql-memory` to a combined value of more than 75% of a machine's total RAM. Doing so increases the risk of memory-related failures.
 {{site.data.alerts.end}}
 
+## Time zone data
+
+Certain features of CockroachDB require the [`tzdata`](https://www.iana.org/time-zones) package (sometimes called `tz` or `zoneinfo`), for example, to support using place names as time zone identifiers. Therefore, ensure that this package is installed and kept up-to-date on all VMs where CockroachDB nodes are running.
```
bdarnell: "machines", not VMs - we don't know or care whether they're virtualized. 

This does not apply to docker-based deployments. The tzdata package is included in the docker image and is update with each patch release of cockroachdb.

jseldess: Done. I'll work on one more commit adding the `tzdata` error to the common errors page.

jseldess:
@bdarnell, PTAL again.

I'm not sure why the build is failing. The anchors it's complaining about are perfectly fine.

Issue #4471, opened 2019-03-07
Encryption at Rest: Validating and getting to GA


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw5Jsp2TrY3it9Yr/reccas5UqyT0V2RID?blocks=hide
PM: @rolandcrosby 
Eng: @mberhault 

@Amruta-Ranade, there are no feature changes. Just removing the experimental status, once Marc is ready.

mberhault:
Some small additions since last time:
https://github.com/cockroachdb/cockroach/pull/35234: `cockroach debug encryption-active-key ` which outputs the current encryption key ID (without needing to know it)
https://github.com/cockroachdb/cockroach/pull/35506: adds the `rocksdb.encryption.algorithm` per-store metric

mberhault:
I don't think we make much mention of the performance impact of encryption-at-rest.
In general, we will see an extra 5-10% CPU utilization on modern CPUs (in the last 10 years or so). This extra utilization depends a lot on load of course.
A "normal" sql load will have higher cpu utilization for writes (no choice there) and will often not change for reads as rocksDB will have all this cached.
An extreme example is a full backup which reads all the data in the cluster.

Here is an example of cpu utilization over 4 days. Before the big dip (firewall mishap), everything is in plain text. The daily increases in CPU are full backups. The hourly (smaller) increases in CPU are incremental backups.

We can see a slight increase in CPU usage overall (half read/half write workload) but a much greater CPU utilization increase when Full backups are running.

The machines at m5.large (2vcpu) which is the lowest recommended.

We do not have good benchmarks to show the impact of encryption at rest, but we should at least mention "higher cpu utilization". We should also mention (and probably measure) the impact of backups on cluster performance.

jseldess:
Removing "experimental" label in https://github.com/cockroachdb/docs/pull/4531.

Amruta-Ranade:
Closed via #4577

PR #4577, opened 2019-03-26
Encryption at rest updates

Amruta-Ranade:
Addresses #4471 

Changes made:
- Added doc for `debug encryption-status`
- Updated the encryption doc with link to the debug command doc
- Added a known issue for higher CPU usage
- Updated the sidebar nav
- Removed the `experimental` label

To do: Figure out how to make the `rocksdb.encryption.algorithm` metric info show up on the Custom Charts doc


v19.1/encryption.md
```diff
@@ -10,18 +10,9 @@ Data encryption and decryption is the process of transforming plaintext data to
 
 CockroachDB uses TLS 1.2 for inter-node and client-node [authentication](authentication.html) as well as setting up a secure communication channel. Once the secure channel is set up, all inter-node and client-node network communication is encrypted using a [shared encryption key](https://en.wikipedia.org/wiki/Transport_Layer_Security) as per the TLS 1.2 protocol. This feature is enabled by default for all secure clusters and needs no additional configuration.
 
-## Encryption at Rest (Experimental)(Enterprise)
+## Encryption at Rest (Enterprise)
```
mberhault: yay! GA!

v19.1/encryption.md
```diff
@@ -188,6 +180,8 @@ Information about keys is written to [the logs](debug-and-error-logs.html), incl
 * Active/old key information at startup.
 * New key information after data key rotation.
 
+Alternatively, you can use the [`cockroach debug encryption-status`](debug-encryption-status.html) command to view information about a store's encryption algorithm and store key.
```
mberhault: this should say `encryption-active-key`

v19.1/debug-encryption-active-key.md
```diff
@@ -0,0 +1,35 @@
...
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach start --store=cockroach-data --enterprise-encryption=path=cockroach-data,key=aes-128.key,old-key=plain --insecure --certs-dir=certs
+~~~
+
+View the encryption algorithm and store key:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach debug encryption-active-key cockroach-data
```
mberhault: Do we want a sample output? If so, here's one:
```
$ cockroach debug encryption-active-key cockroach-data
AES128_CTR:be235c29239aa84a48e5e1874d76aebf7fb3c1bdc438cec2eb98de82f06a57a0
```

Amruta-Ranade:
@mberhault Do we also want to document the `encryption-status` command? If yes, can you point me to the PR?

v19.1/debug-encryption-active-key.md
Amruta-Ranade: Hehe.. just asked you if we want to document the command :)

mberhault:
Seems we have the same question. I'm not convinced we need `encryption-status`, but it can be added if we want to be thorough. The PR was https://github.com/cockroachdb/cockroach/pull/28582
It's definitely a debug command in the sense that it just dumps a bit list of details: key metadata (never the key contents), and information about per-file encryption.

v19.1/debug-encryption-active-key.md
Amruta-Ranade: Added the output.

Amruta-Ranade:
As I understand it, the `encryption-active-key` command is the user-friendly command. `encryption status` seems more oriented towards the developers, and also displays the data keys, which I am not sure if we want the users to easily access. And we don't document all debug commands - just a select few. I agree that we probably shouldn't document it.

mberhault:
Sounds good. Anyone curious will probably just check the list of available commands.

LGTM again. Thanks for the changes!

jseldess:
___
*[v19.1/debug-encryption-active-key.md, line 2 at r4](https://reviewable.io/reviews/cockroachdb/docs/4577#-Laz_lU7D4214a5jV34y:-LazcOxvBJ-53MGyNMxR:bb7vl8z) ([raw file](https://github.com/cockroachdb/docs/blob/b4065aa802cc6a199c4b87b1ec3cb3c031d0b72f/v19.1/debug-encryption-active-key.md#L2)):*

This should be title case
</blockquote></details>

Ah, good catch. That's true for the other encryption debug page you added recently, too, Amruta.

Issue #4177, opened 2018-12-17
sql: automatically add index for fk constraint on empty table


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwCRxl2NcZZqWlu3/recW0rpfQpLQy0nB1
PM: @awoods187 
Eng: @lucy-zhang 

PR: https://github.com/cockroachdb/cockroach/pull/32234

From release notes:
> An `ALTER TABLE` statement to add a foreign key constraint now automatically creates the necessary index if the referencing table is empty and the index does not already exist. [#32234][#32234] {% comment %}doc{% endcomment %}

rmloveland:
Fixed by #4599

PR #4599, opened 2019-03-28
Automatically index fk columns on an empty table

### LINKED ISSUE #4134: Automatically add the needed index during alter table add fk constraint on empty tables
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwCRxl2NcZZqWlu3/recW0rpfQpLQy0nB1
PM: @awoods187 
Eng: @lucy-zhang

rmloveland:
Fixes #4134, #4177.

Summary of changes:

- Update ADD CONSTRAINT doc's section on adding foreign key constraints
  to remove the "you must create an index" language and add some
  "indexes are created for you" language

- Update tables in ADD CONSTRAINT to use new style

- A few small link and copy edits

### DIFF
--- a/v19.1/add-constraint.md
+++ b/v19.1/add-constraint.md
@@ -6,9 +6,9 @@ toc: true
 
 The `ADD CONSTRAINT` [statement](sql-statements.html) is part of `ALTER TABLE` and can add the following [constraints](constraints.html) to columns:
 
-- [`CHECK`](check.html)
-- [Foreign key](foreign-key.html)
-- [`UNIQUE`](unique.html)
+- [`UNIQUE`](#add-the-unique-constraint)
+- [`CHECK`](#add-the-check-constraint)
+- [Foreign key](#add-the-foreign-key-constraint-with-cascade)
 
 {{site.data.alerts.callout_info}}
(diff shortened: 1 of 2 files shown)


rmloveland:
Direct link to updated section:

- [ADD CONSTRAINT > Add the foreign key constraint with CASCADE](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/5922b9b8f24ae1e9bc107b463dc5467ef45c1f27/v19.1/add-constraint.html#add-the-foreign-key-constraint-with-cascade)

rmloveland:
> This looks fine as is, but I wonder if we should be more explicit that adding a foreign key for a non-empty non-indexed table will fail. Do we say that anywhere else?

I think it's implied on the [Foreign key constraint page](https://www.cockroachlabs.com/docs/v19.1/foreign-key.html) where we say among other things that

> Foreign key columns must be indexed

and

> To meet the index requirement when adding the Foreign Key constraint to an existing table, if the columns you want to constraint are not already indexed, use [`CREATE INDEX`](create-index.html) to index them and only then use the [`ADD CONSTRAINT`](add-constraint.html) statement to add the Foreign Key constraint to the columns.

To make it explicit, I've updated this PR in 2 places:

1. Added a small not to the ADD CONSTRAINT page saying 

> Adding a foreign key for a non-empty, non-indexed table will fail, since foreign key columns must be indexed. For more information about the requirements for creating foreign keys, see [Rules for creating foreign keys](foreign-key.html#rules-for-creating-foreign-keys).

2. Updated the 'Foreign Keys' page list of requirements for creating FKs to note that we generate the index for you when you add the FK constraint to an empty table.  Looking at it prompted me to do so.

PR #4603, opened 2019-03-29
ci: Deeplink to changed files in PR comments.

rolandcrosby:
This change to the `pull-request` CI script attempts to determine
which Markdown files have changed using `git diff`, and adds deep
links to the generated HTML output when commenting on Github.

I'm not sure exactly where this script runs, so this may not work in production, but I've verified that the parts that I changed produce the expected output in isolation.


rolandcrosby:
OK, I think I got this working now.

PR #4608, opened 2019-04-01
19.1 updates

jseldess:
- Remove "experimental" from name of auto stats cluster settings 
- Rename experimental_reorder_joins_limit session variable
- Increase required Go version for building from source

### DIFF
--- a/_includes/v19.1/misc/automatic-statistics.md
+++ b/_includes/v19.1/misc/automatic-statistics.md
@@ -6,7 +6,7 @@ If you need to turn off automatic statistics collection:
 
     {% include copy-clipboard.html %}
     ~~~ sql
-    > SET CLUSTER SETTING sql.stats.experimental_automatic_collection.enabled=false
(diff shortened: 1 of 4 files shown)


jseldess:
@rolandcrosby, interestingly, only 1 of 4 changed files were picked up here. The `_includes` is pretty impossible to pick up, cause you'd have to search for the files where it's included. But the install pages probably should've been picked up. I suspect they were missed cause they're `.html` rather than `.md` files?

rolandcrosby:
Yeah I should have been more specific in my CI change PR - it only tries to
link to .md files whose path does not start with an underscore. So
_includes is skipped, as is any .html file. I can change the script today
to include .html files as well.

jseldess:
Thanks, @rolandcrosby. Don't worry about includes. Those files aren't rendered directly. They're included in other files that get rendered. But it'd be great to capture any `.html` files. There are only a few.

Issue #4369, opened 2019-02-12
distsql: rip out the experimental lookup join code


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/34142

From release notes:
> The `experimental_force_lookup_join` [session variable](../v2.2/set-vars.html) has been removed. [#34142][#34142] {% comment %}doc{% endcomment %}

jseldess:
Done.

rmloveland:
Reopening based on convo in https://github.com/cockroachdb/docs/pull/4614

grepping in `v19.1` folder reveals:

```
ack 'experimental_force_lookup_join'
experimental-features.md
22:| `experimental_force_lookup_join`    | `'off'`       | Indicates whether the planner should try and plan a lookup join where the left side is scanned and index lookups are done on the right side. Will emit a warning if a lookup join cannot be planned.                                                                                                    |

joins.md
146:    > SET experimental_force_lookup_join = true;

show-vars.md
83:| experimental_force_lookup_join | off            |
```

PR #4614, opened 2019-04-01
Update zigzag join in experimental features

### LINKED ISSUE #2974: Planning Zigzag Joins
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw8IfZMg9Up8SP33/recSXt518uUYSnN8c
PM: @awoods187 
Eng Partners: @andy-kimball

rmloveland:
Fixes #2974.

Summary of changes:

- Update description of zigzag join session var on 'Experimental
  Features' page.

- Update session var name to `experimental_enable_zigzag_join` wherever
  it is used.


rmloveland:
> I know this is experimental, but do we want to call it out as a distinct join algorithm here? https://www.cockroachlabs.com/docs/v19.1/joins.html#join-algorithms

I don't think so.  Over on #2974 Andy W. said

> This is still experimental and we don't recommend it for production use yet. I think this should just be in the list of experimental features

so I went for the "minimum viable change". :-)

> We do callout lookup joins there, which is also marked as experimental...

That will be going away via #4369 since apparently experimental lookup joins are gone (notified by Radu)

jseldess:
Makes sense, @rmloveland. Thanks!

PR #4615, opened 2019-04-01
Document RENAME CONSTRAINT and fact that ALTER TABLE commands can be combined

### LINKED ISSUE #4520: sql: rearchitecture ALTER TABLE RENAME, add support for renaming constraints
PR: https://github.com/cockroachdb/cockroach/pull/35091

From release notes:
> CockroachDB now supports `ALTER TABLE ... RENAME CONSTRAINT`. Only indexes that are not depended on by views can be renamed. [#35091][#35091] {% comment %}doc{% endcomment %}
>
> The [`RENAME COLUMN`](../v19.1/rename-column.html) command can now be used alongside other table commands in a single [`ALTER TABLE`](../v19.1/alter-table.html) statement. This makes it possible to, for example, atomically add a computed column based on an existing column, and rename the columns so that the computed column "replaces" the original column. [#35091][#35091] {% comment %}doc{% endcomment %}

jseldess:
Fixes #4520.


jseldess:
One other questions, @knz. Should we adjust any of these schema change limitations? https://www.cockroachlabs.com/docs/dev/online-schema-changes.html#limitations

knz:
thank you!

knz:
> One other questions, @knz. Should we adjust any of these schema change limitations? cockroachlabs.com/docs/dev/online-schema-changes.html#limitations

The limitations remain the same. However you can perhaps highlight that a single ALTER TABLE statement with multiple sub-commands (something that's possible today) can become a handy substitute to a BEGIN...COMMIT block with multiple ALTER statements inside (which is not possible).

I don't know if that would fit in the limitation section, or a different section, though.

jseldess:
___
*[v19.1/known-limitations.md, line 35 at r4](https://reviewable.io/reviews/cockroachdb/docs/4615#-LbYT3rR08FxcyNZZySN:-LbYUOqs3t6_4DShKxpj:b712wug) ([raw file](https://github.com/cockroachdb/docs/blob/98b42b8cef718f433a060475e108a215cc68f3d5/v19.1/known-limitations.md#L35)):*

This entire paragraph is not relevant any more. The limitation has been lifted in both 2.1.x and 19.1.
</blockquote></details>

Thanks, @knz. Removed. I'll check the rest of the limitations while adding new 19.1 limitations in a separate PR.

PR #4618, opened 2019-04-02
Remove extra dot in Linux copy command

mattayes:


mattayes:
Thanks for the quick merge @jseldess!

jseldess:
My pleaseure, @mattayes. Thanks for helping out!

PR #4623, opened 2019-04-03
Edit ApplicationName param for JDBC

### LINKED ISSUE #4294: JDBC "application_name" parameter does not work because its invalid
Using the latest version of cockroach `2.1.4` and the postgres JDBC driver `42.2.5` there seems to be difficulty passing `application_name`.

This seems to be because the correct parameter as defined by ~postgres~ the JDBC driver (in contrast to the actual setting supported by the postgres server) is actually `ApplicationName` not `application_name` see: https://jdbc.postgresql.org/documentation/head/connect.html

Therefore, specifically for JDBC examples, places where we suggest this: 
```
jdbc:postgresql://host:5432/db?application_name=foo
```
should actually be this...
```
jdbc:postgresql://host:5432/db?ApplicationName=foo
```

i have seen the incorrect parameter in a number of official and unofficial examples i believe.

BTW, this impacts "App" filtering on the "Statements" screen in our UI.

lhirata:
- Searched for mentions of `application_name` for JDBC and didn't find anything to clean up.
- Added a note to Connection Parameters to call out `ApplicationName` for JDBC.

Closes #4294.

### DIFF
--- a/_includes/v19.1/app/BasicSample.java
+++ b/_includes/v19.1/app/BasicSample.java
@@ -24,6 +24,7 @@ public class BasicSample {
         props.setProperty("sslrootcert", "certs/ca.crt");
         props.setProperty("sslkey", "certs/client.maxroach.pk8");
         props.setProperty("sslcert", "certs/client.maxroach.crt");
+        props.setProperty("ApplicationName", "roachtest");
 
         Connection db = DriverManager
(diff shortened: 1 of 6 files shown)


jseldess:
Tim agreed to this change in another channel, so let's go ahead and merge, @lhirata.

Issue #3184, opened 2018-05-22
windows: Linux builds don't work on windows filesystems (via docker/WSL)


ssmun:
I tried to test CockroachDB and followed instructions on this link: [https://www.cockroachlabs.com/docs/stable/start-a-local-cluster-in-docker.html#os-windows](url)

> ERROR: failed to start server: could not create temp storage: could not open rocksdb instance: IO error: While fsync: a directory: Invalid argument

Log file:

> I180522 20:22:08.011632 1 util/log/clog.go:1134  [config] file created at: 2018/05/22 20:22:08
> I180522 20:22:08.011632 1 util/log/clog.go:1134  [config] running on machine: roach1
> I180522 20:22:08.011632 1 util/log/clog.go:1134  [config] binary: CockroachDB CCL v2.0.2 (x86_64-unknown-linux-gnu, built 2018/05/21 14:55:20, go1.10)
> I180522 20:22:08.011632 1 util/log/clog.go:1134  [config] arguments: [/cockroach/cockroach start --insecure]
> I180522 20:22:08.011632 1 util/log/clog.go:1134  line format: [IWEF]yymmdd hh:mm:ss.uuuuuu goid file:line msg utf8=✓
> W180522 20:22:08.011625 1 cli/start.go:909  RUNNING IN INSECURE MODE!
> 
> - Your cluster is open for any client that can access <all your IP addresses>.
> - Any user, even root, can log in without providing a password.
> - Any user, connecting as root, can read or write any data in your cluster.
> - There is no network encryption nor authentication, and thus no confidentiality.
> 
> Check out how to secure your cluster: https://www.cockroachlabs.com/docs/v2.0/secure-a-cluster.html
> I180522 20:22:08.046850 1 server/config.go:330  available memory from cgroups (8.0 EiB) exceeds system memory 1.9 GiB, using system memory
> W180522 20:22:08.066892 1 cli/start.go:812  Using the default setting for --cache (128 MiB).
>   A significantly larger value is usually needed for good performance.
>   If you have a dedicated server a reasonable setting is --cache=.25 (495 MiB).
> I180522 20:22:08.067224 1 server/config.go:330  available memory from cgroups (8.0 EiB) exceeds system memory 1.9 GiB, using system memory
> W180522 20:22:08.067282 1 cli/start.go:825  Using the default setting for --max-sql-memory (128 MiB).
>   A significantly larger value is usually needed in production.
[...]

PR #4628, opened 2019-04-03
Be louder about experimental status of windows build

danhhz:
This "Install CockroachDB on Windows" page has a warning box but the
bolded text is "Native CockroachDB on Windows requires Windows 8 or
higher" and the experimental status is a bit hidden in the rest of the
box. This binary is almost entirely untested, so I'd suggest we swap the
bolded text to be about the experimental status of the binary.
Otherwise, I worry that users may form a bad initial impression of our
performance/stability/maturity (and perhaps don't bother contacting us
as a result).

### DIFF
--- a/v19.1/install-cockroachdb-windows.html
+++ b/v19.1/install-cockroachdb-windows.html
@@ -17,7 +17,7 @@ key: install-cockroachdb.html
 <div id="download-the-binary-windows" class="install-option">
   <h2>Download the Binary</h2>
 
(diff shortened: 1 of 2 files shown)


danhhz:
cc @bdarnell in case you have any thoughts about the phrasing of the part about testing. I tried strengthening the language here from "has not been extensively tested by Cockroach Labs" to "is largely untested by Cockroach Labs".

jseldess:
Thanks, @danhhz. I'm ok with this, but note that there are some other issues around windows that we have yet to resolve. For example, this issue suggest that we should remove docker and kubernetes options for windows and encourage running in a linux vm: https://github.com/cockroachdb/docs/issues/3184#issuecomment-461506063.

@rolandcrosby is helping with this, so he may have some opinions, too.

danhhz:
Gotcha. I'm not too tied to this. Just found myself filing an issue about it and realized that it was literally less work to PR than to finish writing up the issue. Happy to merge as an interim solution if that makes sense or throw it away if it doesn't.

jseldess:
I'll merge for you, @danhhz. Thank you!

PR #4631, opened 2019-04-04
Update CBO page `EXPLAIN(OPT)` info

### LINKED ISSUE #4348: Update uses of EXPLAIN(OPT) to reflect simpler default output
Via @andy-kimball comment on #4340:

> The default output of EXPLAIN(OPT) has changed to show much less detail. To show more verbosity, you can use EXPLAIN(OPT,VERBOSE) or EXPLAIN(OPT,TYPES). You'll need to update the sample output accordingly.

Estimated scope of work:

- [Cost-based optimizer](http://www.cockroachlabs.com/docs/v2.2/cost-based-optimizer.html) (the "View query plan" section)
- [`EXPLAIN (OPT)` docs](https://www.cockroachlabs.com/docs/stable/explain.html#opt-option)
- Possibly other examples

rmloveland:
Fixes #4348.

Summary of changes:

- Remove explicit use of `EXPLAIN(OPT)` from CBO page; all usage should
  be on `EXPLAIN` page

- Add quick explanations of the `OPT, *` options, with direct links to
  each example on the `EXPLAIN` page

### DIFF
--- a/v19.1/cost-based-optimizer.md
+++ b/v19.1/cost-based-optimizer.md
@@ -27,38 +27,9 @@ The most important factor in determining the quality of a plan is cardinality (i
 
 ## View query plan
 
-To see whether a query will be run with the cost-based optimizer, run the query with [`EXPLAIN (OPT)`](explain.html). The `OPT` option displays a query plan tree, along with some information that was used to plan the query. If the query is unsupported (i.e., it returns an error message that starts with e.g., `pq: unsupported statement` or `pq: aggregates with FILTER are not supported yet`), the query will not be run with the cost-based optimizer and will be run with the legacy heuristic planner.
+To see whether a query will be run with the cost-based optimizer, run the query with [`EXPLAIN (OPT)`](explain.html#opt-option). The `OPT` option displays a query plan tree, along with some information that was used to plan the query.
 
-For example, the following query (which uses [CockroachDB's TPC-H data set](https://github.com/cockroachdb/cockroach/tree/b1a57102d8e99b301b74c97527c1b8ffd4a4f3f1/pkg/workload/tpch)) returns the query plan tree, which means that it will be run with the cost-based optimizer:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> EXPLAIN (OPT) SELECT l_shipmode, avg(l_extendedprice) from lineitem GROUP BY l_shipmode;
-~~~
-
-~~~
-                                     text
-+-----------------------------------------------------------------------------+
-group-by
-├── columns: l_shipmode:15(string!null) avg:17(float)
-├── grouping columns: l_shipmode:15(string!null)
-├── stats: [rows=700, distinct(15)=700]
-├── cost: 1207
-├── key: (15)
-├── fd: (15)-->(17)
-├── prune: (17)
-├── scan lineitem
-│    ├── columns: l_extendedprice:6(float!null) l_shipmode:15(string!null)
-│    ├── stats: [rows=1000, distinct(15)=700]
-│    ├── cost: 1180
-│    └── prune: (6,15)
-└── aggregations [outer=(6)]
-└── avg [type=float, outer=(6)]
-└── variable: l_extendedprice [type=float, outer=(6)]
-(16 rows)
-~~~
-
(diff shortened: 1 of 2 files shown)


v19.1/cost-based-optimizer.md
```diff
@@ -27,38 +27,20 @@ The most important factor in determining the quality of a plan is cardinality (i
 
 ## View query plan
 
-To see whether a query will be run with the cost-based optimizer, run the query with [`EXPLAIN (OPT)`](explain.html). The `OPT` option displays a query plan tree, along with some information that was used to plan the query. If the query is unsupported (i.e., it returns an error message that starts with e.g., `pq: unsupported statement` or `pq: aggregates with FILTER are not supported yet`), the query will not be run with the cost-based optimizer and will be run with the legacy heuristic planner.
+To see whether a query will be run with the cost-based optimizer, run the query with [`EXPLAIN (OPT)`](explain.html#opt-option). The `OPT` option displays a query plan tree, along with some information that was used to plan the query.
```
justinj: Can we emphasize that `explain (opt)` isn't really intended for use by users (except to detect if a plan is running with CBO vs. HP)? The common case should be that they just use `explain`, and if that's missing information users need we should add it there.

v19.1/cost-based-optimizer.md
```diff
@@ -27,38 +27,20 @@ The most important factor in determining the quality of a plan is cardinality (i
...
-└── aggregations [outer=(6)]
-└── avg [type=float, outer=(6)]
-└── variable: l_extendedprice [type=float, outer=(6)]
-(16 rows)
-~~~
+<span class="version-tag">New in v19.1</span>: The default output of `EXPLAIN (OPT)` has been simplified from previous versions. Also, there are several new options that provide more information about how the optimizer is planning a query:
 
-In contrast, queries that are not supported by the cost-based optimizer return errors that begin with the string `pq: unsupported statement: ...` or specific messages like `pq: aggregates with FILTER are not supported yet`. Such queries will use the legacy heuristic planner instead of the cost-based optimizer.
+- [`EXPLAIN (OPT)`](explain.html#opt-option) shows a query plan tree.
+- [`EXPLAIN (OPT, VERBOSE)`](explain.html#opt-verbose-option) shows cost details.
```
justinj: this actually shows a ton of internal stuff in addition to cost. my bias is always towards not documenting such things but I'll defer to your judgement :)

v19.1/cost-based-optimizer.md
```diff
@@ -27,38 +27,20 @@ The most important factor in determining the quality of a plan is cardinality (i
...
+- [`EXPLAIN (OPT, TYPES)`](explain.html#opt-types-option) shows cost and type details.
+- [`EXPLAIN (OPT, ENV)`](explain.html#opt-env-option) shows all details used by the optimizer to generate the query plan, including:
```
justinj: I'm similarly mildly against publicizing this feature, since I see its value as being an escape hatch for *us* to get all the information we need from customers about a bad plan, and without that particular use case I don't see what value a customer would get out of it. Like, for instance, I don't think this page really explains why someone would ever want to use it.

v19.1/cost-based-optimizer.md
rmloveland: > Can we emphasize that explain (opt) isn't really intended for use by users (except to detect if a plan is running with CBO vs. HP)?

I think I've achieved roughly this by removing all of the bulleted list items that describe the different `EXPLAIN (OPT)` variants, ending the section with the "if a query is unsupported" statement since that's the use case we want to emphasize.

I didn't add more content in the spirit of "in all other cases, just use EXPLAIN" since it would spend more words to say "use this feature in this one case but not otherwise" which would paradoxically serve to emphasize it even more, I think. :-)

FTR these options are documented over at [`EXPLAIN`](https://www.cockroachlabs.com/docs/dev/explain.html#opt-option), but extremely briefly - mostly just to note their existence.

v19.1/cost-based-optimizer.md
rmloveland: Pulled from the doc per the previous comment.

v19.1/cost-based-optimizer.md
rmloveland: Also removed.

PR #4633, opened 2019-04-04
Remove beta from CBO; add join hint consideration

### LINKED ISSUE #4509: Updates to Optimizer Docs
We recently added cross join hints (https://github.com/cockroachdb/cockroach/issues/35695). As such we need to update the docs:
https://www.cockroachlabs.com/docs/dev/cost-based-optimizer.html
This is no longer true:
`CockroachDB does not support a CROSS <hint> JOIN`

We should also add a bullet under additional considerations to the effect of:
"Hints should be reconsidered with each new release of the CBO. We are constantly adding to our capabilities and the hints specified in an older version of the CBO may cause decreased performance with future releases."

Also, I think we can remove beta from the CBO:
```Warning:
This is a beta feature. It is currently undergoing continued testing. Please file a Github issue with us if you identify a bug.```

rmloveland:
Fixes #4509.


rmloveland:
@awoods187 FYI gonna bump this directly into docs team review since it's such a small change.

Issue #4570, opened 2019-03-24
sql: Expand automatic statistics documentation


awoods187:
Document the knobs users can set to adjust automatic statistics collection (introduced in this [pr](https://github.com/cockroachdb/cockroach/pull/36085)) either in the main issue https://www.cockroachlabs.com/docs/v19.1/cost-based-optimizer.html#table-statistics or via cluster settings/session variables.

New settings:
sql.stats.automatic_collection.fraction_stale_rows 
sql.stats.automatic_collection.min_stale_rows

jseldess:
@awoods187, we've already updated the cluster settings page to cover these new settings: https://www.cockroachlabs.com/docs/dev/cluster-settings.html. Is that enough, or do you think they're important enough to call out in the cost-based optimizer page as well?

awoods187:
I think it might be nice to have one sentence that says "you can configure automatic statistics using these cluster settings X and Y" with a link to the cluster settings page.

jseldess:
@rmloveland, are you touching the cost-based optimizer docs still for 19.1? If so, can you factor this in?

rmloveland:
> are you touching the cost-based optimizer docs still for 19.1? If so, can you factor this in?

Yes and yes.

rmloveland:
Fixed by #4635

Issue #4517, opened 2019-03-18
sql: create stats after a schema change


rmloveland:
PR: https://github.com/cockroachdb/cockroach/pull/35252

From release notes:
> Schema changes now trigger automatic statistics collection for the affected table. [#35252][#35252] {% comment %}doc{% endcomment %}

rmloveland:
Fixed by #4635

PR #4635, opened 2019-04-04
Document more auto stats features & knobs

### LINKED ISSUE #4455: Revise optimizer table stats info with more context re: auto vs. non-auto
In #4441, @jseldess asks (rightly) regarding our current docs on table stats, why a user would want to manually generate stats given that we are now defaulting to auto-stats.

Estimated scope of work:

- Revise table stats section of CBO page with more high-level guidance and context re: auto-stats vs. manual stats vs. how it all works, and why (if ever) you would manually generate stats given our current defaults

See also:

- #4441
- #4379

rmloveland:
Fixes #4455, #4570, #4517.

Summary of changes:

- Add more context re: auto stats in general, and deemphasize turning it
  off, tweaking the settings, or running CREATE STATS manually

- Note that auto stats update after schema changes

- Add docs on cluster settings for throttling auto stats


rmloveland:
@rytaft

Some of this PR is factual and some more conceptual.  In addition to adding facts about settings I hope to provide more context around why users should mostly leave auto stats alone unless they are troubleshooting / know what they are doing.

However what I have here may not be the right way to think about it (or express it).  It may be making too strong a claim about leaving things alone, since there may be use cases that are pretty common for disabling / throttling stats, or running CREATE STATS manually - I'm honestly not sure, so I welcome any feedback.  Thanks!

RaduBerinde:
Hi @rmloveland - I noticed there is another page which mentions how to delete statistics, but it doesn't mention the "restart to clear cache part": https://www.cockroachlabs.com/docs/stable/create-statistics.html#delete-statistics

Should we update that too? Or just have it link to the CBO section?

rmloveland:
Thanks @RaduBerinde , created #4872 to fix.

PR #4644, opened 2019-04-08
Allow big tables to scroll right if needed

### LINKED ISSUE #4643: Cluster Settings Doc Update
Re: [Cluster Settings](https://cockroachlabs.com/docs//v19.1/cluster-settings.html)

The `Description` column on this page does not fit horizontally and there's no scroll in sight. Even if I maximize the window on my wide-ass monitor, I still can't see the whole things; it's quite hard to read.

rmloveland:
Tweak the CSS to use automatic overflow.

Fixes #4643.


andreimatei:
It is better. When I maximize the window on my monitor, I can read the whole description. But the width of the description field still seems way too little to me.
For example, here's how it looks then the windows is half of my screen. I wish we could make the other cols narrower.

Screenshot 2019-04-08 15 04 21

rmloveland:
> I wish we could make the other cols narrower

I agree.  I think the culprits are `server.heap_profile.system_memory_threshold_fraction` and friends, since the max length setting determines that column's width.

To (perhaps) improve things even more, I'll remove the TOC from this page in a forthcoming commit, since it's taking up space on the right but only providing a couple of (IMO) not so interesting headings.

PR #4645, opened 2019-04-08
Add S3 / GCE examples

### LINKED ISSUE #3910: Add import instructions for GCS and S3
Our import docs currently include examples for Azure storage (their equivalent of S3). But given most of our managed customers will be on GCE and AWS, and one of the first things they'll try to do is import, we should have instructions on how to import from S3 and Google Cloud Storage. This is primarily aimed at managed customers on GCP and AWS, but would apply to non-managed customers as well. 

cc: @mberhault

lhirata:
Closes #3910.

### DIFF
--- a/_includes/v2.1/misc/external-urls.md
+++ b/_includes/v2.1/misc/external-urls.md
@@ -4,12 +4,12 @@
 
 | Location                                                    | Scheme      | Host                                             | Parameters                                                                 |
 |-------------------------------------------------------------+-------------+--------------------------------------------------+----------------------------------------------------------------------------|
-| Amazon S3                                                   | `s3`        | Bucket name                                      | `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`                               |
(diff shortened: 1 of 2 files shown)


kannanlakshmi:
I am happy to review this, but I am not an import expert and we should probably get @rolandcrosby's eyes on these!

lhirata:
@kannanlakshmi mostly added you since you filed the issue. Will add @rolandcrosby too!

Amruta-Ranade:
@mberhault I think we addressed that in https://github.com/cockroachdb/docs/pull/4600

mberhault:
Nope, that feature is in 19.1 only. We may want examples for it though.

lhirata:
@mberhault looks like it's a 2.1 update too (https://github.com/cockroachdb/cockroach/pull/33046), so added `SESSION_TOKEN` to the 2.1 docs and added it to the new 19.1 examples.

lhirata:
Removed AWS_SESSION_TOKEN from most examples, except the first. We have it listed in the Parameters table too.

lhirata:
@mberhault and/or @rolandcrosby, can you take another look at this please?

PR #4653, opened 2019-04-10
Add support for AS OF SYSTEM TIME clause in BEGIN and SET TRANSACTION

### LINKED ISSUE #4376: sql: add support for AS OF SYSTEM TIME clause in BEGIN and SET TRANSACTION
PR: https://github.com/cockroachdb/cockroach/pull/34305

From release notes:
> Added support for [`AS OF SYSTEM TIME`](../v2.2/as-of-system-time.html) clauses in [`BEGIN TRANSACTION`](../v2.2/begin-transaction.html) and [`SET TRANSACTION`](../v2.2/set-transaction.html) statements, which enables entire read-only transactions to be run against a historical timestamp. This functionality simplifies performing complex analytics against a consistent snapshot of historical data and eases the burden to use historical reads with ORMs which generally make modifying the syntax of generated `SELECT` statements difficult. [#34305][#34305] {% comment %}doc{% endcomment %}
>
> Also, the behavior of the `now()` [built-in function](../v2.2/functions-and-operators.html) inside of historical `SELECT ... AS OF SYSTEM TIME` queries now reflects the historical timestamp at which the query is being run rather than the current clock time when the statement is executed. [#34305][#34305] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Closes #4376

### DIFF
--- a/_includes/v19.1/sql/begin-transaction-as-of-system-time-example.md
+++ b/_includes/v19.1/sql/begin-transaction-as-of-system-time-example.md
@@ -1,0 +1,19 @@
+{% include copy-clipboard.html %}
+~~~ sql
+> BEGIN AS OF SYSTEM TIME '2019-04-09 18:02:52.0+00:00';
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM orders;
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM products;
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> COMMIT;
+~~~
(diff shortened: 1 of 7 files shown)


nvanbenschoten:
@ajwerner the one thing this might still be missing is a mention of what other effects `AS OF SYSTEM TIME` has on a transaction. Does it make the transaction read-only?

ajwerner:
Great point. Using `AS OF SYSTEM TIME` implies `READ ONLY`. It is illegal to specify `READ WRITE` with an AOST in either `BEGIN` Or `SET TRANSACTION`. It is legal to `BEGIN READ WRITE` and the `SET TRANSACTION AS OF SYSTEM TIME` which will set the transaction read only. Today you can already do `BEGIN READ WRITE` followed by `SET TRANSACTION READ ONLY` so this behavior shouldn’t be too surprising.

Amruta-Ranade:
@ajwerner I don't understand what "makes the transaction read-only" means. Can you elaborate on that please?

ajwerner:
The `BEGIN` and `SET TRANSACTION` statements allow the user to specify the read mode as `READ ONLY` or `READ WRITE`. A read only transaction cannot issue any statements which modify data. Historical queries are read only (it’s not clear what it would mean to write in the past). 

The following statement would be illegal:

```
BEGIN READ WRITE, AS OF SYSTEM TIME ...;
```

Let me know if you’d like more clarification.

PR #4674, opened 2019-04-22
Update ORM docs with foreign key note

### LINKED ISSUE #4669: Note that some ORMs like SQLAlchemy rely on foreign keys to do their join magic
Some ORMs like sqlalchemy rely on the existance of foreign keys to do a lot of their join magic:

https://docs.sqlalchemy.org/en/13/orm/join_conditions.html

So in cases where we have to remove FKs, that means we basically canât use joins in ORMs (natively) and instead have to do that work in the app or issue raw SQL

See also, workarounds for https://github.com/cockroachdb/cockroach/issues/36859

(h/t @nstewart for the info)

rmloveland:
Fixes #4669.

Summary of changes:

- Note that these ORMs rely on foreign keys to generate joins.  If you
  disable FKs, they won't build the joins for you.  For each ORM, also
  link to the likely API for emitting raw SQL.

### DIFF
--- a/v19.1/build-a-python-app-with-cockroachdb-sqlalchemy.md
+++ b/v19.1/build-a-python-app-with-cockroachdb-sqlalchemy.md
@@ -22,6 +22,10 @@ We have tested the [psycopg2 driver](http://initd.org/psycopg/docs/) and [SQLAlc
 The example code on this page uses Python 3.
 {{site.data.alerts.end}}
 
+{{site.data.alerts.callout_danger}}
+SQLAlchemy relies on the existence of [foreign keys](foreign-key.html) to generate [`JOIN` expressions](joins.html) from your application code. If you remove foreign keys from your schema, SQLAlchemy won't generate joins for you. As a workaround, you can [create a "custom foreign condition" by adding a `relationship` field to your table objects](https://stackoverflow.com/questions/37806625/sqlalchemy-create-relations-but-without-foreign-key-constraint-in-db), or do the equivalent work in your application.
+{{site.data.alerts.end}}
+
 ## Step 1. Install SQLAlchemy
 
 To install SQLAlchemy, as well as a [CockroachDB Python package](https://github.com/cockroachdb/cockroachdb-python) that accounts for some differences between CockroachDB and PostgreSQL, run the following command:
--- a/v19.1/build-a-go-app-with-cockroachdb-gorm.md
+++ b/v19.1/build-a-go-app-with-cockroachdb-gorm.md
@@ -44,7 +44,7 @@ $ go get -u github.com/jinzhu/gorm
 
 ## Step 3. Generate a certificate for the `maxroach` user
 
-Create a certificate and key for the `maxroach` user by running the following command.  The code samples will run as this user.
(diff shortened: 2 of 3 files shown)


rmloveland:
Nate, FYI I'm going to try to find reviewers for the Gorm and ActiveRecord pieces (I assume you're OK with reviewing the SQLAlchemy bit, but let me know if not).

v19.1/build-a-python-app-with-cockroachdb-sqlalchemy.md
```diff
@@ -22,6 +22,12 @@ We have tested the [psycopg2 driver](http://initd.org/psycopg/docs/) and [SQLAlc
 The example code on this page uses Python 3.
 {{site.data.alerts.end}}
 
+{{site.data.alerts.callout_danger}}
+SQLAlchemy relies on the existence of foreign keys to generate [`JOIN` expressions](joins.html) from your application code. If you remove foreign keys from your schema, SQLAlchemy won't generate joins for you. Instead, you'll have to issue raw SQL (via [`engine.execute()`](https://docs.sqlalchemy.org/en/13/core/connections.html?highlight=engine.execute#sqlalchemy.engine.Engine.execute)) or do the equivalent work in your application.
```
bdarnell: Instead of falling back to raw SQL, you can also use a "custom foreign condition" by adding a `relationship` field to your table objects.

https://stackoverflow.com/questions/37806625/sqlalchemy-create-relations-but-without-foreign-key-constraint-in-db

https://docs.sqlalchemy.org/en/13/orm/join_conditions.html#creating-custom-foreign-conditions

(the sqlalchemy docs are rather opaque about this so the SO answer is probably worth linking to instead)

BramGruneir:
GORM doesn't seem to require it, but you can add them: http://doc.gorm.io/associations.html#polymorphism

ActiveRecord doesn't seem to require them either, but they help: https://edgeguides.rubyonrails.org/association_basics.html

v19.1/build-a-python-app-with-cockroachdb-sqlalchemy.md
rmloveland: I've updated this section to point users to the custom foreign condition SO link and removed the (potentially confusing) link to the SQLAlchemy docs, since it might not be necessary to solve the user's problem.  PTAL.

rmloveland:
Thanks for the help parsing those magic scrolls @BramGruneir - if those ORMs don't require FKs, I'm inclined to drop those changes from this PR.  Do you think that's reasonable?

BramGruneir:
I agree.  No need to confuse people more.

v19.1/build-a-python-app-with-cockroachdb-sqlalchemy.md
nstewart: Thanks @bdarnell -- I tried following the docs and this still wasn't working (as you noticed, the docs weren't that clear). I gave up after that and just did the joins at the app level. Good to know there is a more comprehensive SO solution.

rmloveland:
> I agree. No need to confuse people more.

Removed in latest version.

rmloveland:
Since there have been no further comments, I'll pass this on for docs team review now.  Thanks for the help!

PR #4675, opened 2019-04-22
Added MSO security docs

### LINKED ISSUE #4039: Managed CockroachDB Security overview
We need to describe the security of a Managed CockroachDB cluster. 

This should leverage the work @lhirata is doing for https://github.com/cockroachdb/docs/issues/3549. It also needs to explain that a managed cluster is single-tenant (no shared machines/cpu) running in a VPC (no shared network). Firewall is only way in.

Amruta-Ranade:
Closes #4039

**Summary:**

- Created MSO security docs
- Deleted the user management doc and merged the content with the new managed-authorization doc
- Updated the sidebar nav

**Note for all reviewers:**
I have left out the 'how certs work' section and other conceptual info because I don't think MSO users need to worry about those details. However, I am open to including the details as well - let me know what you think. 

@jseldess After the IA discussion in our last 1:1, I thought it was better to make new documents for the managed security docs since I didn't find much content overlap with the CRDB docs. However, I am open to reuse the CRDB docs and use MSO-specific tags instead. Let me know what you prefer.

### DIFF
--- a/v19.1/managed-authentication.md
+++ b/v19.1/managed-authentication.md
@@ -1,0 +1,30 @@
+---
+title: Authentication
+summary: Learn about the authentication features for Managed CockroachDB clusters.
+toc: true
+build_for: [managed]
+---
+
+Managed CockroachDB uses TLS 1.2 digital certificates for inter-node authentication, [SSL modes](https://www.postgresql.org/docs/11/libpq-ssl.html) for node identity verification, and password authentication for client identity verification.
+
+## Node identity verification
+
+Managed CockroachDB uses the `verify-full` [SSL mode](https://www.postgresql.org/docs/11/libpq-ssl.html) by default to verify a node’s identity. This mode encrypts the data in-flight as well as verifies the identity of the CockroachDB node, thus ensuring a secure connection to your cluster. Using this mode prevents MITM (Man in the Middle) attacks, impersonation attacks, and eavesdropping.
+
+To connect securely to your cluster using the `verify-full` mode:
+
+1. Download the CA certificate and place it in the `certs` directory. The Certificate Authority (CA) certificate is the file that the client uses to verify the identity of the CockroachDB node.
+2. When connecting to the cluster, specify the path to the `certs` directory in the connection string. See [Connect to your cluster](managed-connect-to-your-cluster.html) for more details.
+
+Managed CockroachDB also supports the `require` SSL mode, although we do not recommend using it since it can make the cluster susceptible to MITM and impersonation attacks. For more information, see the "Protection Provided in Different Modes" section in PostgreSQL's [SSL Support](https://www.postgresql.org/docs/9.4/libpq-ssl.html) document.
+
+## Client identity verification
+
+Managed CockroachDB uses password authentication for verifying a client’s identity. If no password has been set up for a user, password authentication will always fail for that user and you won’t be able to connect to the cluster.
+
+To set a password for a SQL user, [use the Console](managed-authorization.html#use-the-console) or the  [`ALTER USER`](alter-user.html) command.
+
+## See also
+
+- [Client Connection Parameters](connection-parameters.html)
+- [Connect to Your Managed Cluster](managed-connect-to-your-cluster.html)
--- a/v19.1/managed-authentication.md
+++ b/v19.1/managed-authentication.md
@@ -9,7 +9,7 @@ Managed CockroachDB uses TLS 1.2 for inter-node and client-node communication, d
 
 ## Node identity verification
 
-Use the `verify-full` [SSL mode](https://www.postgresql.org/docs/11/libpq-ssl.html) to verify a node’s identity. This mode encrypts the data in-flight as well as verifies the identity of the CockroachDB node, thus ensuring a secure connection to your cluster. Using this mode prevents MITM (Man in the Middle) attacks, impersonation attacks, and eavesdropping.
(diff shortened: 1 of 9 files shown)


mberhault:
The links in the comment result in a 404. I had to use the `managed/dev/` prefix instead.

Amruta-Ranade:
@mberhault Oh! Hmm..the top link works fine, but the deep links don't. @rolandcrosby Any idea how I could fix this?

v19.1/managed-authorization.md
```diff
@@ -1,22 +1,36 @@
...
+Managed CockroachDB supports network authorization and user authorization.
+
+## Network authorization
+
+Managed CockroachDB requires you to authorize the networks that can access the cluster. This helps prevent denial-of-service and brute force password attacks.
+
+Authorize your application server’s network and your local machine’s network by adding the IP addresses in the CIDR notation using the [Networking page](managed-connect-to-your-cluster.html#authorize-your-network). If you change your location, you will need to authorize the new location’s network, else the connection from that network will be rejected.
+
+{{site.data.alerts.callout_info}}
+While developing and testing your application, you may whitelist `0.0.0.0/0`, which allows all networks. However, before moving into production, make sure you delete the `0.0.0.0/0` network since it allows anybody who uses your password to access the database and makes your application vulnerable to security breaches.
```
mberhault: "vulnerable to security breaches". That's a bit dramatic. I think the main concern currently is as mentioned above: denial-of-service (easy) and brute-force-attacks (not easy, but also turns into a DOS)

v19.1/managed-authentication.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Authentication
+summary: Learn about the authentication features for Managed CockroachDB clusters.
+toc: true
+build_for: [managed]
+---
+
+Managed CockroachDB uses TLS 1.2 digital certificates for inter-node authentication, [SSL modes](https://www.postgresql.org/docs/11/libpq-ssl.html) for node identity verification, and password authentication for client identity verification.
```
mberhault: This is always a bit weird. Certificates are not for a specific version of TLS. Instead, I think we should say that we use TLS 1.2 for client/node and node/node communication, and digital certificates for node/node authentication.

v19.1/managed-authentication.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Node identity verification
+
+Managed CockroachDB uses the `verify-full` [SSL mode](https://www.postgresql.org/docs/11/libpq-ssl.html) by default to verify a node’s identity. This mode encrypts the data in-flight as well as verifies the identity of the CockroachDB node, thus ensuring a secure connection to your cluster. Using this mode prevents MITM (Man in the Middle) attacks, impersonation attacks, and eavesdropping.
```
mberhault: This is more about recommending `verify-full` than using.
The only mode we cannot support is `disable` as that requires plaintext. All other modes with work just fine with our clusters.

v19.1/managed-authorization.md
Amruta-Ranade: Hmm..in that case, can I stop at "access the database" and remove the "makes your application..." part?

v19.1/managed-authentication.md
Amruta-Ranade: Is this better:

"Managed CockroachDB uses TLS 1.2 for inter-node and client-node communication, digital certificates for inter-node authentication, [SSL modes](https://www.postgresql.org/docs/11/libpq-ssl.html) for node identity verification, and password authentication for client identity verification."

v19.1/managed-authentication.md
Amruta-Ranade: This para talks about the SSL mode we use by default (`verify-full`) and what the user needs to do about it. I do include info about using the other supported mode (`require`) in the next para.

Also, I tried the `disable`, `allow`, and `prefer` modes and got the `unsupported sslmode` error.

mberhault: I know, I was talking more about the phrasing. We don't use and can't enforce which mode they use other than `disable`.
The errors come from libpq which does not support `allow` and `prefer` as it can't switch dynamically between TLS and plaintext. `disable` just won't work with our clusters.
If you try with `psql`, you should be able to connect with everything other than `disable`.

Amruta-Ranade: Ahh..okay. Does this sound better: "Use the `verify-full` SSL mode to verify a node's identity"

Amruta-Ranade:
___
*[v19.1/managed-authentication.md, line 12 at r1](https://reviewable.io/reviews/cockroachdb/docs/4675#-Ld5ckuAq3XeCcd2Xmfi-r1-12:-LdAvJTWBfWu0VqDO20Q:bwmrsvb) ([raw file](https://github.com/cockroachdb/docs/blob/eb862437e90a64804c216d90fed46dcb879d8a92/v19.1/managed-authentication.md#L12)):*

Ahh..okay. Does this sound better: "Use the `verify-full` SSL mode for node authentication."
</blockquote></details>

In-person discussion with Marc: "Yes" :)

Amruta-Ranade:
From Slack: "document k8s style setup where CA is stored as k8s secret"

mberhault:
I would say the k8s instructions can be done in a separate PR.

Amruta-Ranade:
From the MSO product council: We should have developer code snippets talking about how to use certs while building apps with different languages (Go, Python, etc.). 
cc @rmloveland @jseldess @vilterp @kannanlakshmi

jseldess:
@Amruta-Ranade, we already have this in our non-managed docs. For each language, we have secure and insecure code snippets. The secure versions use certs. For example: https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb.html#basic-statements

So we just need to understand how to reuse in managed docs.

PR #4685, opened 2019-04-24
CBO: document index selection w/leaseholder prefs

### LINKED ISSUE #4586: release-19.1: opt: Consider leaseholder preferences when selecting indexes
PR: https://github.com/cockroachdb/cockroach/pull/36123

From release notes:
> The [cost-based optimizer](../v19.1/cost-based-optimizer.html) will now try to select the index having a leaseholder preference that is closest. [#36123][#36123] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #4586.

Summary of changes:

- Update 'Cost-based optimizer' with a new section 'Preferring the
  nearest index > Example - leaseholder preferences' where we describe
  how index selection based on leaseholder prefs works at a high level,
  as well as some considerations.  Finally, add an example showing how
  to configure it

- Update 'Enterprise Licensing' page with a description of the feature
  and a link to docs


v19.1/cost-based-optimizer.md
```diff
@@ -328,6 +331,236 @@ $ cockroach sql --insecure --host=localhost --port=26259 --database=test -e 'EXP
...
+
+- Zone constraints are always respected (hard constraint), whereas lease preferences are taken into account as "additional information" -- as long as they don't contradict the zone constraints.
+
+- The optimizer doesn't consider the real-time location of leaseholders when selecting an index; it is pattern matching on the text values passed in the configuration (e.g., the [`ALTER INDEX`](alter-index.html) statements shown below). For the same reason, the optimizer only matches against the first locality in your `lease_preferences` array.
+
+- In some cases, it may not be possible for the optimizer to use an index that satisfies your leaseholder preferences, since [leaseholders can move](architecture/replication-layer.html#leaseholder-rebalancing)  to different nodes. Therefore, you should only use this feature if you’re confident you know where the leaseholders will end up based on your cluster's usage patterns. We recommend thoroughly testing your configuration to ensure the optimizer is selecting the index(es) you expect.
```
andy-kimball: It's always "possible" for the optimizer to use any index. It just may be that it decides to use an index based on leaseholder preferences, even though that index has moved to a different node/region, thus causing slower performance than expected.

v19.1/cost-based-optimizer.md
```diff
@@ -328,6 +331,236 @@ $ cockroach sql --insecure --host=localhost --port=26259 --database=test -e 'EXP
...
+
+> CREATE INDEX token_id_east_idx ON token (token_id) STORING (access_token, refresh_token);
+~~~
+
+Enter zone configurations to distribute replicas across the cluster as follows:
+
+- For the "East" index, store 2/5 replicas in the East, 2/5 in Central, and 1/5 in the West. Further, prefer that the leaseholders for that index live in the East or, failing that, in the Central region.
+- Follow the same replica and leaseholder patterns for each of the Central and West regions.
+
+The idea is that e.g. `token_id_east_idx` will have sufficient replicas and its leaseholder stored in the East, so that if a query comes in that accesses the columns covered by that index from the East gateway node, the optimizer will select `token_id_east_idx` for fast reads.
```
andy-kimball: Might want to also mention that you're deliberately putting 2 replicas in the preferred region so that if one replica goes down, the leaseholder will still stay in that region.

v19.1/cost-based-optimizer.md
```diff
@@ -328,6 +331,236 @@ $ cockroach sql --insecure --host=localhost --port=26259 --database=test -e 'EXP
...
+
+{% include copy-clipboard.html %}
+~~~ sql
+> ALTER TABLE token CONFIGURE ZONE USING
```
andy-kimball: One nit: the table itself is the primary index, so it's somewhat redundant to have the table with lease preferences in the west, and to also have a secondary index in the west. It makes for an understandable example, though. Maybe have a side-note explaining it?

v19.1/cost-based-optimizer.md
rmloveland: Rewrote to use this phrasing/info instead - PTAL.

v19.1/cost-based-optimizer.md
rmloveland: Updated to make this explicit.

v19.1/cost-based-optimizer.md
rmloveland: I added a note, let me know what you think.  Hopefully it's not too confusing.  We could also remove the `ALTER TABLE` if the note isn't working.

rmloveland:
Hey @andy-kimball what do you think of the latest updates?  If they look OK I'll pass this along for docs team review.  If not I'm happy to make any needed changes.

PR #4688, opened 2019-04-28
Geo partitioning tutorial

jseldess:
Add geo-partitioning tutorial to accompany new video demo. Also create new `Tutorials` section for such stand-alone demonstrations/tutorials, and move contents of `Guides > Explore Capabilities` as well. If not in this PR, then soon after, I'd like to find a way to clarify Guides vs. Tutorials, perhaps via a simple tooltip in the sidenav.


piyush-singh:
Barring some minor grammatical nits, :lgtm:! Most of my suggestions can probably wait until after Wednesday if there is a time crunch. The only one I would definitely recommend changing is adding locality flags to the `gen haproxy` commands.

PR #4701, opened 2019-05-01
Add cloud storage sink known limitation

### LINKED ISSUE #4686: CDC: Call out lack of key in cloud storage sink
See cockroachdb/cockroach#36994. In v19.1.0, cloud storage CDC sinks do not emit the record's keys when emitting deletes, so you can't actually tell what record is deleted. We plan to fix this shortly in a point release, but in the meantime we should call this out as a known limitation until that fix lands.

lhirata:
Closes #4686.


lhirata:
Bump @rolandcrosby / @Amruta-Ranade for review. Should be quick! Thanks

PR #4712, opened 2019-05-06
Clarifying some statements about indexes

thatnerd:
* Suggest that users look to `EXPLAIN` plans to determine when an index is being used
* Modified the language around when indexes will slow down a write operation; it's a little more ambiguous than was implied.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *~
-*.swp
+.*.swp
 .DS_Store
 generate/generate
 Railroad.jar
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.swp
 .DS_Store
 generate/generate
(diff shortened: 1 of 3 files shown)


thatnerd:
Will do. I think I need to review the purpose of those different directories.

jseldess:
Yeah, I know the directory structure isn't immediately clear. Our repo contains docs for all versions to-date. https://github.com/cockroachdb/docs/blob/master/CONTRIBUTING.md#docs-structure

thatnerd:
Added a commit that ports those lines (and only those lines) to v2.1 @jseldess

thatnerd:
Dammit, forgot I was still mid-pull-request. Added a change to .gitignore.

jseldess:
Thanks, @thatnerd! I'll merge for you.

PR #4804, opened 2019-05-14
Added PERSISTENT => true to php examples. 

tim-o:
Without this we'll cycle connections each query. See https://forum.cockroachlabs.com/t/throughput-not-increasing-when-doubling-nodes-3-to-6/2725/5


tim-o:
@jseldess d'oh. Added!

PR #4805, opened 2019-05-14
Update k8s tutorials

### LINKED ISSUE #4689: Upgrade CockroachDB via Helm
Our [current Helm instructions](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html#upgrade-the-cluster) do not use `helm` commands to upgrade a CockroachDB cluster. There was a specific reason for not using helm here, but it's been lost to history, it seems. 

- This note might touch on why we didn’t just use `helm upgrade`: https://github.com/helm/charts/tree/master/stable/cockroachdb#upgrading.
- This section also mentions problems with `helm upgrade` and StatefulSets: https://github.com/helm/charts/tree/master/stable/cockroachdb#scaling

Guidance from @DuskEagle:
> I think it will only work if you delete the `cluster-init.yaml` or `cluster-init-secure.yaml` before issuing the `helm upgrade` command.

jseldess:
- Use helm upgrade for adding and removing nodes and cluster upgrades
- Correct downgrade option for 19.1 k8s upgrade
- Wrap commands to be more visible

Fixes #4689.


jseldess:
@DuskEagle, sorry for including so many unrelated formatting changes in this PR. 

The files to look at for the `helm upgrade` changes are:
- `kubernetes-upgrade-cluster.md`
- `kubernetes-scale-cluster.md`
- `kubernetes-remove-nodes-secure.md`
- `kubernetes-remove-nodes-insecure.md`

jseldess:
@Amruta-Ranade, ready for your review.

PR #4813, opened 2019-05-15
Specify reasonable instance type for k8s deployments

### LINKED ISSUE #4811: Reasonable defaults for GKE node pool instance size in Kubernetes docs
Our [orchestrate CockroacdhDB with Kubernetes](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html) docs has users create a new GKE cluster with a default node pool. This node pool runs [n1-standard-1](https://cloud.google.com/compute/docs/machine-types) instances, which are woefully underprovisioned for running CockroachDB. 

We should have users create a node pool with a more reasonable default instance type, such as an n1-standard-4. This can be done by passing a `--machine-type` flag to the `gcloud container clusters create cockroachdb` CLI invocation, e.g.
```
gcloud container clusters create cockroachdb --machine-type n1-standard-4
```

jseldess:
Fixes #4811.

### DIFF
--- a/_includes/v19.1/orchestration/start-kubernetes.md
+++ b/_includes/v19.1/orchestration/start-kubernetes.md
@@ -16,14 +16,14 @@ Choose whether you want to orchestrate CockroachDB with Kubernetes using the hos
 
     {% include copy-clipboard.html %}
     ~~~ shell
-    $ gcloud container clusters create cockroachdb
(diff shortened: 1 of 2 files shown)


jseldess:
TFTR, @DuskEagle!

PR #4820, opened 2019-05-17
Topology patterns, v2

### LINKED ISSUE #4574: Topology Pattern Library
We have a [v1 of topology patterns](https://www.cockroachlabs.com/docs/stable/cluster-topology-patterns.html). We need to make sure we're presenting the right patterns with the right dimensions and guidance 

- Docs goal: Making it easier for users to know how to configure CockroachDB to get the latency and resiliency they require in a variety of contexts. 
- Marketing goal: Showing that CockroachDB can do things that other databases can't. Tying topology patterns into real customer stories.

RAIC:
- R (Responsible): Jesse
- A (Accountable): Jesse
- C (Consulted): Roko, Tim V, Robert, Ben, Core Eng team
- I (Informed): Rest of SE team, Jim W, Kurt H, MSO team, Support team, Nate, Spencer

jseldess:
- [x] Overview
- Single-region
    - [x] Development
    - [x] Basic Production
- Multi-region
    - [x] Geo-Partitioned Replicas
    - [x] Geo-Partitioned Leaseholders
    - [x] Pinned Secondary Indexes
    - [x] Follower Reads
    - [x] Follow-the-Workload

Fixes #4574.

Future improvements/considerations tracked in https://github.com/cockroachdb/docs/issues/4991.


jseldess:
___
*[v19.1/single-region-topologies.md, line 83 at r1](https://reviewable.io/reviews/cockroachdb/docs/4820#-LhakWw26vnMisrcbdUn:-Liit_617fN7tWTYfjf8:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0b8abd78f88492e5debb09471509571bb7a1446/v19.1/single-region-topologies.md#L83)):*

> ```
> Once each leaseholder has appended the write to its Raft log, it notifies its follower replicas
> ```

This isn't exactly true. The leaseholder will append the write to its log at the same time that it's waiting for followers to do so.

Also, you're mixing leaseholder and leader here. I doubt it's worth making the distinction in this paragraph.
</blockquote></details>

Done.

jseldess:
___
*[v19.1/single-region-topologies.md, line 93 at r1](https://reviewable.io/reviews/cockroachdb/docs/4820#-LhMda2P2UOHjGpDaeP4:-LiistsQ6lQlBO6Qcowo:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0b8abd78f88492e5debb09471509571bb7a1446/v19.1/single-region-topologies.md#L93)):*

The whole table doesn't become unavailable; only those portions with replicas on the failing machine. On the other hand, it's possible that an AZ+1 failure hits a critical system range and the whole cluster becomes unavailable. So I wouldn't talk about tables or secondary indexes here, just that there may be data unavailability during an AZ+1 failure. 
</blockquote></details>

Done.

jseldess:
@bdarnell, please take another look. I posted an open question for you on follow-the-workload. And I still need to factor in anti-patterns, though I may hold off on that for v3. I'll also hold off on improving the images.

jseldess:
___
*[v19.1/follower-reads.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/4820#-Linmy_cDfNfdkTGmepF:-Lispcgp-lUk5mClcf4A:b-v6s1o3) ([raw file](https://github.com/cockroachdb/docs/blob/7cd928dcf3779ba4a677a6c367bb9d199d9f9b0b/v19.1/follower-reads.md#L27)):*

This change looks like it belongs in another PR.
</blockquote></details>

Removed.

rkruze:
___
*[_includes/v19.1/topology-patterns/multi-region-cluster-setup.md, line 14 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-LisrLYk7IhnnaUigHye:-LisrLYk7IhnnaUigHyf:b-5afo3k) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/_includes/v19.1/topology-patterns/multi-region-cluster-setup.md#L14)):*
> ```Markdown
> - Region-specific app instances and load balancers
>     - Each load balancer redirects to CockroachDB nodes in its region.
>     - When CockroachDB nodes are unavailable in a region, the load balancer redirects to nodes in other regions.
> ```

If we mention this we should have a section someplace in the docs about how to setup `backup` servers in haproxy.

rkruze:
___
*[v19.1/topology-basic-production.md, line 50 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-LissGUfEtnkWz2ZaICm:-LissGUg9qv8JV8ZWF03:b-2pfbm4) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/v19.1/topology-basic-production.md#L50)):*
> ```Markdown
> #### Reads
> 
> Since all ranges, including leaseholder replicas, are in a single region, reads are very fast.
> ```

I think it would be good to talk about things in terms of latency. So here we could say that "reads will have low latency."

rkruze:
___
*[v19.1/topology-basic-production.md, line 64 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-LissoQ19OYbHXiSlOIS:-LissoQ19OYbHXiSlOIT:bnhuz60) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/v19.1/topology-basic-production.md#L64)):*
> ```Markdown
> #### Writes
> 
> Just like for reads, since all ranges are in a single region, writes can achieve consensus very fast.
> ```

"Writes will also have low latency."

rkruze:
___
*[v19.1/topology-basic-production.md, line 98 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-ListTYU1ODN98nuh3Yf:-ListTYU1ODN98nuh3Yg:b-agub0z) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/v19.1/topology-basic-production.md#L98)):*
> ```Markdown
> - Other system ranges contain information needed to allocate new table IDs and track the status of a cluster's nodes.
> 
> For the cluster as a whole to remain available, these system ranges must retain a majority of their replicas, so CockroachDB comes with [pre-configured replication zones](configure-replication-zones.html#create-a-replication-zone-for-a-system-range) for them with the replication factor increased to `5` by default. Based on the [cluster setup](#configuration) described above, with each node started with the `--locality` flag specifying its AZ, each system range's replicas are spread across 3 AZs. This means that one entire AZ can fail without causing any system range to lose consensus:
> ```

One caveat here is that I don't believe it goes to '5' replicas if there are only 3 machines in the cluster.

rkruze:
___
*[v19.1/topology-follower-reads.md, line 9 at r2](https://reviewable.io/reviews/cockroachdb/docs/4820#-LisW-6j5S8tUEITpAym:-LisuZtvCcJDAZPE0DZZ:bj2l2w0) ([raw file](https://github.com/cockroachdb/docs/blob/7cd928dcf3779ba4a677a6c367bb9d199d9f9b0b/v19.1/topology-follower-reads.md#L9)):*
> ```
> v19.1/topology-follower-reads.md
> ```
I often talk about follower reads being good for queries against tables which have fewer updates.

rkruze:
___
*[v19.1/topology-follower-reads.md, line 10 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-LisumCA0_vWCDjPUgE6:-LisumCA0_vWCDjPUgE7:bpt8ymh) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/v19.1/topology-follower-reads.md#L10)):*
> ```Markdown
> 
> - The table is read frequently but updated infrequently.
> - Reads must be fast, but writes can be slower.
> ```

I'm not sure this should be here, there is nothing about follower reads which slows down writes from any other topology.

rkruze:
___
*[v19.1/topology-follower-reads.md, line 13 at r3](https://reviewable.io/reviews/cockroachdb/docs/4820#-Lisv2xd-fRDsaILnpCh:-Lisv2xd-fRDsaILnpCi:b-tfsgtx) ([raw file](https://github.com/cockroachdb/docs/blob/cfc5777ef408f5d9d2f9f5aee61fdb1e13c7db43/v19.1/topology-follower-reads.md#L13)):*
> ```Markdown
> - Reads can be historical (48 seconds or more in the past).
> - Rows in the table **cannot** be tied to specific geographies (e.g., a reference table).
> - The table must remain available during a region failure.
> ```

How is this related to follower reads?

PR #4830, opened 2019-05-19
Update install-cockroachdb-mac.html

mdunegan:
I suggest changing this sentence to 1.) make it more inclusive, 2.) eliminate contractions: 

CHANGE: Copy the binary into your PATH so it's easy to execute cockroach commands from any shell:
TO: Copy the binary into your PATH so you can execute cockroach commands from any shell:


lhirata:
Thank you for this edit! Can you make the same edit on the `v19.1/install-cockroachdb-linux.md`  page?

PR #4873, opened 2019-06-04
Added Licensing FAQs

Amruta-Ranade:
To-Do: Update licensing terminology in the rest of the docs

### DIFF
--- a/_includes/sidebar-data-v19.1.json
+++ b/_includes/sidebar-data-v19.1.json
@@ -1928,6 +1928,12 @@
         ]
       },
       {
+        "title": "Licensing FAQs",
+        "urls": [
+          "/${VERSION}/licensing-faqs.html"
+        ]
+      },
+      {
         "title": "CockroachDB in Comparison",
         "urls": [
           "/${VERSION}/cockroachdb-in-comparison.html"
--- a/v19.1/licensing-faqs.md
+++ b/v19.1/licensing-faqs.md
@@ -49,7 +49,7 @@ Yes, employees and contractors can use your internal CockroachDB instance as a s
 
 ## What constitutes hosting CockroachDB as a service?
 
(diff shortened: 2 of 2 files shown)


knz:
note: some of the suggestions i've made in the additional faqs mention the phrase "please contact us". I recommend that the doc page contains a link to a contact form or some more details about _how_ to contact us.

piyush-singh:
We should link to our support channel (link: https://support.cockroachlabs.com/hc/en-us). Alternatively, if we feel that is too specific, we can link to the general contact page (https://www.cockroachlabs.com/contact/), but I would recommend the first option to avoid having requests in many different channels.

knz:
So a friend and I were trying to find this FAQ on the web site and we were stumped: it is not listed in the side bar alongside the other FAQs. Where can we find it?
-- 

knz:
I also used the "license" keyword with the search function and couldn't find anything.
-- 

Amruta-Ranade:
@knz 
Here's the link: https://www.cockroachlabs.com/docs/stable/licensing-faqs.html

knz:
So thank you for the link, that helps.

Meanwhile I find the page difficult to find:

- if I open the docs site on my mobile browser, it does not appear listed in the main table of contents (which I assume corresponds to the sidebar)
- the search function does not list it under the keyword "license"

Maybe there are ways to increase its discoverability?

Amruta-Ranade:
You'd have to navigate to FAQs > Licensing FAQs on the sidebar nav. As for search, I remember a convo about indexing taking time after a page is published, but I could be misremembering. cc @jseldess

knz:
So starting later today the mobile version of the site I was getting on my phone was eventually updated. I think I was hitting an outdated version from cloudflare or something. Now the page is easy to discover. Thank you again!

PR #4902, opened 2019-06-06
Update licensing-faqs.md

knz:
There are 3 situations, not two, in the last FAQ.


knz:
> apply this change to the 19.2 version of the page as well.
done.

(May I suggest to have the template for editing a page, linked to from the top right of the rendered doc pages, explain to a contributor that this is desired. I forget every time.)

jseldess:
Ah, I didn't realized you were editing from the browser. I don't want to make that any more difficult. It's fine for external users to edit a page and then leave it to us to port the change to other versions. I assumed you were working from the command line.

knz:
I usually do, but for very simple fixes like this one, I really like the convenience of the direct link to the github editor.

Issue #4852, opened 2019-05-29
Expand docs sidenav by default


jseldess:
This would make our categorizations a lot more intuitive.

jseldess:
This should probably wait until we kickoff our broader docs refresh with design.

luciaaaaaaa:
hey, following up on this as part of the redesign. 

Does this refer to the navigation sidenav? would like to explore alternative navigation patterns instead of expanding by default. 

One question I have around the sidenav is whether we think there is value in removing the automatic collapse that happens as I navigate through different sections. That way a user could expand all sections and Ctrl+F. Thoughts?

cc @lnhsingh

jseldess:
Yes, this is about the sidenav. Excited to get your input here. I don't want this issue to dictate our solution, especially if we change our navigation in more substantial ways.

luciaaaaaaa:
sounds good. thinking it would be helpful if we could change this issue into a descriptor of what the challenges you experience with the current navigation is. otherwise perhaps we should just close out?

jseldess:
Let's just close. I think the motive here is to give users a more intuitive sense of what's available. Our categories are collapsed by default, in contrast to a lot of other docs. So there are clicks involved in discovering content or verifying instincts. I know some users prefer the docs landing page, for example, since it lists a lot more topics.

PR #4906, opened 2019-06-07
[WIP] Expand top-level sections in sidenav

jseldess:
This is an experiment to see whether the sidenav works
better with top-level sections expanded by default.

I've done this via a css hack. The right way likely
involves changing the JS in _includes/sidebar.html.
As-is, top-level sections are expanded, but when you
click a page in the sidenav, it redraws with the
sections briefly collapsing due to something in the JS.

After confirming that the top-level Tutorails section
hasn't caused those pages to get more views than when
they were a sub-section of Guides, I've moved them back
under Guides as a new Evaluate section.

Also, I removed Guides > Contribute to conserve vertical
spaces, given those pages get very little traffic.

Addresses #4852.


rmloveland:
Feedback: I like this so so so so much.  Seeing more things without having to click is a big win IMO.

PR #4919, opened 2019-06-17
CDC Updates

### LINKED ISSUE #4785: Changefeeds Dashboard Doc Update
Re: [Changefeeds Dashboard](https://cockroachlabs.com/docs//v19.1/admin-ui-cdc-dashboard.html)

## Issue Description

The graph shows as 29.4 seconds, but the text below has nanoseconds.  

In the node view, the graph shows the maximum latency (in nanoseconds) of any running changefeed for the node.

## Suggested Resolution 

is the time in nano or seconds?

### LINKED ISSUE #4841: release-19.1: changefeedccl: fix DELETEs in cloud storage sinks
PR: https://github.com/cockroachdb/cockroach/pull/37328

From release notes:
> [`CHANGEFEED`s](../v19.1/create-changefeed.html) now accept a `key_in_value` option; this is automatically used for cloud storage sinks, making the primary key of deleted rows recoverable. [#37328][#37328] {% comment %}doc{% endcomment %}

### LINKED ISSUE #4869: Clarify how to compare two rows for happens-before
Roko: the timestamp with `updated` is a cluster global timestamp but looking at our docs it seems that in order to make this work you have to compare the “resolved” timestamp from all partitions?

Dan: With duplicates removed, an individual row is emitted in the same order as the transactions that updated it. However, this is not true for updates to two different rows, even two rows in the same table. Resolved timestamp notifications on every Kafka partition can be used to provide strong ordering and global consistency guarantees by buffering records in between timestamp closures.

Roko: it seems to imply `updated` is only on a partition basis and that `resolved` has to be used and compared to make this work.

Dan: If you want to compare two rows for happens-before, you can simply compare the updated timestamp. This works across tables, nodes, anything in the same cluster. The resolved timestamp comes into play if you want to know every row that changed at a certain time 

Ben: (with the caveat that happens-before is potentially ambiguous in a serializable (but not linearizable) environment)

Dan: (Which would require waiting for every kafka partition to be past some point of progress)

Roko: Ok, so there could be ambiguity of 500ms, right?

Ben: right, but as long as things were changed in overlapping transactions, it's fine. the ambiguity is just between disjoint transactions, where it often doesn't matter

### LINKED ISSUE #4917: CDC: Edit avro output
Client is seeing the following in the avro output: `{"id":{"long":1}}          {"after":{"office_dogs_avro":{"id":{"long":1},"first_name":{"string":"Mani 1- Updated 2019-Jun-07 2:38pm "},"last_name":null}}}` Is the `after` element to be expected?

Yeah, that's the format. The new row is always in "after", which lets us stick metadata at the top level.

In the docs it says it should show up as `{"id":1}    {"id":1,"name":{"string":"Petee H"}}`

lhirata:
- Update CDC Dashboard - Closes #4785.
- DELETEs in cloud storage (key_in_value) - Closes #4841.
- How to compare rows for happens-before - Closes #4869.
- Edit Avro output - Closes #4917.

### DIFF
--- a/_includes/v19.1/known-limitations/cdc.md
+++ b/_includes/v19.1/known-limitations/cdc.md
@@ -6,4 +6,4 @@ The following are limitations in the v19.1 release and will be addressed in the 
 - Partial or intermittent sink unavailability may impact changefeed stability; however, [ordering guarantees](change-data-capture.html#ordering-guarantees) will still hold for as long as a changefeed [remains active](change-data-capture.html#monitor-a-changefeed).
 - Changefeeds cannot be altered. To alter, cancel the changefeed and [create a new one with updated settings from where it left off](create-changefeed.html#start-a-new-changefeed-where-another-ended).
 - Additional target options will be added, including partitions and ranges of primary key rows.
-- In v19.1.0, when emitting deletes, [cloud storage sinks](create-changefeed.html#cloud-storage-sink) do not emit the record's keys; therefore, the deleted record is not identifiable.
+- In v19.1.0, when emitting deletes, [cloud storage sinks](create-changefeed.html#cloud-storage-sink) do not emit the record's keys; therefore, the deleted record is not identifiable. This has been fixed in v19.1.1 and above.
--- a/v19.1/create-changefeed.md
+++ b/v19.1/create-changefeed.md
@@ -96,6 +96,7 @@ Option | Value | Description
 `cursor` | [Timestamp](as-of-system-time.html#parameters)  | Emits any changes after the given timestamp, but does not output the current state of the table first. If `cursor` is not specified, the changefeed starts by doing an initial scan of all the watched rows and emits the current value, then moves to emitting any changes that happen after the scan.<br><br>When starting a changefeed at a specific `cursor`, the `cursor` cannot be before the configured garbage collection window (see [`gc.ttlseconds`](configure-replication-zones.html#replication-zone-variables)) for the table you're trying to follow; otherwise, the changefeed will error. With default garbage collection settings, this means you cannot create a changefeed that starts more than 25 hours in the past.<br><br>`cursor` can be used to [start a new changefeed where a previous changefeed ended.](#start-a-new-changefeed-where-another-ended)<br><br>Example: `CURSOR=1536242855577149065.0000000000`
 `format` | `json` / `experimental_avro` | Format of the emitted record. Currently, support for [Avro is limited and experimental](#avro-limitations). For mappings of CockroachDB types to Avro types, [see the table below](#avro-types). <br><br>Default: `format=json`.
 `confluent_schema_registry` | Schema Registry address | The [Schema Registry](https://docs.confluent.io/current/schema-registry/docs/index.html#sr) address is required to use `experimental_avro`.
+`key_in_value` | N/A | <span class="version-tag">New in v19.1</span>: Makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for these sinks (currently only [cloud storage sinks](#cloud-storage-sink)).
 
 #### Avro limitations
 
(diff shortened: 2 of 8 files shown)


v19.2/change-data-capture.md
```diff
@@ -69,9 +69,13 @@ The core feature of CDC is the [changefeed](create-changefeed.html). Changefeeds
     {"__crdb__": {"updated": "1532379923319195777.0000000000"}, "id": 4, "name": "Lucky"}
     ~~~
 
-- With duplicates removed, an individual row is emitted in the same order as the transactions that updated it. However, this is not true for updates to two different rows, even two rows in the same table. Resolved timestamp notifications on every Kafka partition can be used to provide strong ordering and global consistency guarantees by buffering records in between timestamp closures.
+- With duplicates removed, an individual row is emitted in the same order as the transactions that updated it. However, this is not true for updates to two different rows, even two rows in the same table.
 
-    Because CockroachDB supports transactions that can affect any part of the cluster, it is not possible to horizontally divide the transaction log into independent changefeeds.
+    To compare two different rows for happens-before, compare the "updated" timestamp. This works across anything in the same cluster (e.g., tables, nodes, etc.).
```
rmloveland: Might be useful to link to the wikipedia entry for happens-before to make it clear that it is an understood term of art: https://en.wikipedia.org/wiki/Happened-before

v19.2/create-changefeed.md
```diff
@@ -96,6 +96,7 @@ Option | Value | Description
...
 `format` | `json` / `experimental_avro` | Format of the emitted record. Currently, support for [Avro is limited and experimental](#avro-limitations). For mappings of CockroachDB types to Avro types, [see the table below](#avro-types). <br><br>Default: `format=json`.
 `confluent_schema_registry` | Schema Registry address | The [Schema Registry](https://docs.confluent.io/current/schema-registry/docs/index.html#sr) address is required to use `experimental_avro`.
+`key_in_value` | N/A | Makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for these sinks (currently only cloud storage).
```
rmloveland: Is it understood what we mean by "cloud storage" in this case?  Is there a list of what cloud storage we support for CDC that we could easily link to?

v19.1/create-changefeed.md
```diff
@@ -96,6 +96,7 @@ Option | Value | Description
...
 `format` | `json` / `experimental_avro` | Format of the emitted record. Currently, support for [Avro is limited and experimental](#avro-limitations). For mappings of CockroachDB types to Avro types, [see the table below](#avro-types). <br><br>Default: `format=json`.
 `confluent_schema_registry` | Schema Registry address | The [Schema Registry](https://docs.confluent.io/current/schema-registry/docs/index.html#sr) address is required to use `experimental_avro`.
+`key_in_value` | N/A | <span class="version-tag">New in v19.1</span>: Makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for these sinks (currently only cloud storage).
```
rmloveland: Oh, it looks like we do have a linkable "cloud storage sinks" thing.  Mind adding a link here?

_includes/v19.1/known-limitations/cdc.md
```diff
@@ -6,4 +6,4 @@ The following are limitations in the v19.1 release and will be addressed in the
 - Partial or intermittent sink unavailability may impact changefeed stability; however, [ordering guarantees](change-data-capture.html#ordering-guarantees) will still hold for as long as a changefeed [remains active](change-data-capture.html#monitor-a-changefeed).
 - Changefeeds cannot be altered. To alter, cancel the changefeed and [create a new one with updated settings from where it left off](create-changefeed.html#start-a-new-changefeed-where-another-ended).
 - Additional target options will be added, including partitions and ranges of primary key rows.
-- In v19.1.0, when emitting deletes, [cloud storage sinks](create-changefeed.html#cloud-storage-sink) do not emit the record's keys; therefore, the deleted record is not identifiable.
+- In v19.1.0, when emitting deletes, [cloud storage sinks](create-changefeed.html#cloud-storage-sink) do not emit the record's keys; therefore, the deleted record is not identifiable. This has been fixed in v19.1.1.
```
rmloveland: In situations like this I like to say a little more explicitly "in versions XXX and above" (if true) so the content continues to make sense even if it hangs around for a little while after 19.1.1 doesn't matter that much anymore.

Issue #4571, opened 2019-03-25
Add TypeORM to list of supported 3rd party tools


rmloveland:
We work with TypeORM now.

For details, see:

https://github.com/cockroachdb/cockroach/issues/22298

rmloveland:
Fixed in #4923

PR #4923, opened 2019-06-17
Update Client Drivers page

### LINKED ISSUE #4851: Client drivers page needs various updates
The [Client drivers](https://cockroachlabs.com/docs/stable/install-client-drivers.html) page could stand to be updated.  There are a few nits that once addressed would make it much easier to use when choosing which driver/ORM to use.

- The TOC calls it "Supported Drivers", the page title is "Install Client Drivers", and the link from the outline on the docs homepage is "Client drivers".  "Client drivers" is the title I think we should use for all three.
- It says "recommended drivers" in the table of language/driver combos, but I think we should drop "recommended" and instead list all of the drivers and ORMs that work for each language, and let the developer choose which one to use based on their app's needs and how useful our tutorial for that driver is for them.  Also, the links to the driver pages should instead link to our tutorial for that particular driver (which will in turn link to the driver homepages for downloads, docs, etc.)
- There should be a 'See Also' list at the bottom of the page that links to common stuff like: connection params, txns, etc.

Related:

- #3585
- #1795 (maybe)

rmloveland:
... including mention of TypeORM support.

Summary of changes:

- Update TOC, homepage, and individual pages to use same title: 'Client
  Drivers'

- Break out list of drivers and links to docs pages into a separate
  include so that it can be used on 'Client Drivers' and 'Build an App'
  pages, since the lists were nearly identical (and are now fully so)

- All of the above applies to 19.1 and 19.2

Fixes #4851, #4571


rmloveland:
@Amruta-Ranade FYI AFAICT there is no eng review needed on this one, docs changes/refactoring includes only.

rmloveland:
Screenshot 2019-06-17 15 27 27

Build is not really broken, Github is having issues that are not allowing build to complete.

rmloveland:
> The build is failing due to some link problems:

Dang, sorry -- was flying a bit blind without a locally working `htmltest` due to recent build config changes.  Will fix this now (obviously :-) ), and will see if I can figure out how to update the configs so we have both faster local builds and working tests locally.

Should be fixed now.

> How about adding an h2 (## Tutorials) above the include so See also isn't the only thing in the TOC?

Fixed for both 19.1 and 19.2

>  the build and app overview page and the client drivers page are virtually identical. This is nothing new. Should we consolidate and, say, just redirect from client drivers to the build and app overview page?

I think it would be better not to (yet), since it would create a potentially jarring experience if you are navigating the docs using the left nav.  If we redirect, users will be bounced from **SQL > Client Drivers** to a totally different section of the docs for **Get Started > Build an App**.  We could def look at reorganizing in future though if we really want to avoid the duplication.

Thanks for the review Jesse!

Issue #4872, opened 2019-06-04
Unify stats deletion procedures across DELETE and main CBO pages


rmloveland:
Quoth @RaduBerinde in #4635:

> I noticed there is another page which mentions how to delete statistics, but it doesn't mention the "restart to clear cache part": https://www.cockroachlabs.com/docs/stable/create-statistics.html#delete-statistics

> Should we update that too? Or just have it link to the CBO section?

Estimated scope of work:

- Resolve inconsistency noted by Radu above
- ... ideally, by revamping this information so it's the same in multiple places

Could be as simple as linking to the CBO section, but an include that makes the info available wherever it is used is probably the right thing.

rmloveland:
Fixed in #4925

PR #4925, opened 2019-06-17
Update docs on automatic stats refresh rate

### LINKED ISSUE #4809: Need more info about automatic stats settings
The automatic stats docs (https://www.cockroachlabs.com/docs/v19.1/cost-based-optimizer.html#controlling-automatic-statistics) mention two of the settings used to control automatic stats: `sql.stats.automatic_collection.fraction_stale_rows` and `sql.stats.automatic_collection.min_stale_rows`, but they don't give enough info to really understand how they work. One thing that would be helpful would be a direct link to those settings on the settings page, but it would also help to add a bit more info in the "Controlling automatic statistics" section. Some points we should emphasize:
- The target number of stale rows in a particular table before a refresh is actually equal to `target_stale_rows = fraction_stale_rows * <size of table> + min_stale_rows`.
- We don't necessarily perform a refresh when exactly that target number of rows are stale, because the decision to start a refresh is probabilistic. After each mutation operation (INSERT, UPDATE, or DELETE), the probability of a refresh is `P(refresh) = (# rows updated, inserted, deleted)/target_stale_rows`. The result is that on average, we will perform a refresh after `target_stale_rows` are stale, but it may be higher or lower. 
- To ensure that we always have stats and they aren't *too* stale, we also always perform a refresh if there are no stats or if it's been a long time since the last refresh (a "long time" is defined as 2x the average time between refreshes over the last 4 refreshes).

I don't think we necessarily need all this detail, but I've put it here so you can judge what to include or not. Please let me know if you have any questions -- thanks!
cc @RaduBerinde

rmloveland:
... and node restarts after stats deletion to clear caches.

Summary of changes:

- Add a new subsection to CBO page, 'Controlling statistics refresh
  rate', where we describe the cases when stats are refreshed in more
  detail.

- To match the structure of the above, we break the instructions for
  deleting stats into a new section 'Turning off statistics'

- Finally, tweak stats deletion instructions on CBO page and CREATE
  STATS page so both say that nodes must be restarted
  post-stats-deletion to clear caches.

Fixes #4809, #4872.

### DIFF
--- a/v19.1/cost-based-optimizer.md
+++ b/v19.1/cost-based-optimizer.md
@@ -67,10 +67,24 @@ The cost-based optimizer can often find more performant query plans if it has ac
 
 For best query performance, most users should leave automatic statistics enabled with the default settings. The information provided in this section is useful for troubleshooting or performance tuning by advanced users.
 
-To control how often the automatic statistics jobs run on your cluster, adjust the following [cluster settings](cluster-settings.html). They define the target number of rows in a table that should be stale before statistics on that table are refreshed.
+#### Controlling statistics refresh rate
 
-- `sql.stats.automatic_collection.fraction_stale_rows`
-- `sql.stats.automatic_collection.min_stale_rows`
+Statistics are refreshed in the following cases:
+
+1. When there are no statistics.
+2. When it's been a long time since the last refresh, where "long time" is defined according to a moving average of the time across the last several refreshes.
+3. After each mutation operation ([`INSERT`](insert.html), [`UPDATE`](update.html), or [`DELETE`](delete.html)), the probability of a refresh is calculated using a formula that takes the [cluster settings](cluster-settings.html) shown below as inputs. These settings define the target number of rows in a table that should be stale before statistics on that table are refreshed.  Increasing either setting will reduce the frequency of refreshes.  In particular, `min_stale_rows` impacts the frequency of refreshes for small tables, while `fraction_stale_rows` has more of an impact on larger tables.
+
+| Setting                                              | Default Value | Details                                                                              |
+|------------------------------------------------------+---------------+--------------------------------------------------------------------------------------|
+| `sql.stats.automatic_collection.fraction_stale_rows` |           0.2 | Target fraction of stale rows per table that will trigger a statistics refresh       |
+| `sql.stats.automatic_collection.min_stale_rows`      |           500 | Target minimum number of stale rows per table that will trigger a statistics refresh |
+
+{{site.data.alerts.callout_info}}
+Because the formula for statistics refreshes is probabilistic, you should not expect to see statistics update immediately after changing these settings, or immediately after exactly 500 rows have been updated.
+{{site.data.alerts.end}}
+
+#### Turning off statistics
 
 If you need to turn off automatic statistics collection, follow the steps below:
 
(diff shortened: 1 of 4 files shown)


rmloveland:
@rytaft Thanks for the details.  I tried to split the difference language-wise so we're not too constrained if the implementation of refresh rate needs to change.  PTAL and let me know what you think.  I am happy to add more detail if you think that's too vague.  I guess my sense is that if users want to mess with these settings they should really know what they're doing (at least that's what the docs say currently) so I'm loathe to document exact formulas. But I defer to your expertise!

@RaduBerinde I snuck in the update to fix #4872 (stats deletion & node restart to clear cache) as well since it's in the same area, mind taking a look?

rytaft:
Thanks for this update, @rmloveland! I'm not too worried about the implementation of the refresh rate changing, since in that case we'll need to change those cluster settings and will update the docs anyway (although we could change the defaults). Regardless, I do think this is a good level of detail. The one thing I'd suggest is to add a sentence or two about how changing the settings qualitatively affects the refresh rate. For example, you could mention that increasing either `min_stale_rows` or `fraction_stale_rows` reduces the frequency of refreshes. `min_stale_rows` in particular impacts the frequency of refreshes for small tables, while `fraction_stale_rows` has more of an impact on larger tables.

Maybe @RaduBerinde also has an opinion about the level of detail...

v19.1/cost-based-optimizer.md
```diff
@@ -67,10 +67,20 @@ The cost-based optimizer can often find more performant query plans if it has ac
...
-- `sql.stats.automatic_collection.min_stale_rows`
+Statistics are refreshed in the following cases:
+
+1. When there are no statistics.
+2. When it's been a long time since the last refresh, where "long time" is defined according to a moving average of the time across the last several refreshes.
+3. After each mutation operation ([`INSERT`](insert.html), [`UPDATE`](update.html), or [`DELETE`](delete.html)), the probability of a refresh is calculated using a formula that takes the [cluster settings](cluster-settings.html) shown below as inputs. These settings define the target number of rows in a table that should be stale before statistics on that table are refreshed.
+
+| Setting                                              | Details                                                                              |
+|------------------------------------------------------+--------------------------------------------------------------------------------------|
+| `sql.stats.automatic_collection.fraction_stale_rows` | Target fraction of stale rows per table that will trigger a statistics refresh       |
```
RaduBerinde: Do we want to include the default values here? Seems like it would be useful, but there's the downside that they could become stale. Not sure what the general practice is for cluster settings.

rmloveland: Our general practice for cluster settings (thus far) is to try really hard to point users to the (generated, up-to-date) cluster settings page, unless we really need to mention the settings.  Since we need them here in this case, I'd rather not include the default values here for the reason you said: staleness / the risk of the values being incorrect.

rmloveland:
Thanks for the reviews @rytaft and @RaduBerinde !  I hope my latest updates incorporate your feedback correctly, please let me know if you think further updates are needed.

rytaft:
Looks good, @rmloveland! I think you might just want to add a bit to this sentence:
> you should not expect to see statistics update immediately after changing these settings

In addition to expecting a refresh after changing the settings, users might also expect a refresh after a certain number of rows have changed. For example, If a table is empty, users might expect a refresh after exactly 500 rows have been inserted (the default for `min_stale_rows`). But the refresh could actually happen before or after that.

rmloveland:
rytaft <notifications@github.com> writes:

> Looks good, @rmloveland! I think you might just want to add a bit to
> this sentence:
>
>     you should not expect to see statistics update immediately after
>     changing these settings
>
> In addition to expecting a refresh after changing the settings, users
> might also expect a refresh after a certain number of rows have
> changed. For example, If a table is empty, users might expect a
> refresh after exactly 500 rows have been inserted (the default for
> min_stale_rows). But the refresh could actually happen before or after
> that.

Thanks Becca!  Updated to note that 500 rows will not necessarily
trigger the refresh.  In order for that sentence to even make sense, I
went ahead and added the default values for each setting to this page
instead of just relying on the cluster settings page (so Radu I guess I
lied before, sorry. :-} ).

PTAL

rytaft:
LGTM thanks!

PR #4929, opened 2019-06-18
Fix em-dashes in cost-based query optimizer docs

akshayjshah:
The documentation for the cost-based query optimizer uses `&emdash;`
instead of `&mdash;`, which isn't correct. Luckily, the fix is simple :)


akshayjshah:
`CONTIBUTING.md` suggests that I assign this to @jseldess, but it doesn't look like I have the requisite permissions. Sorry!

PR #4955, opened 2019-06-20
Describe schema caching vs. locality-aware indexes

### LINKED ISSUE #4672: CBO: Explain the role schema caching plays in locality-aware index prefs
Re: [Cost-Based Optimizer](https://cockroachlabs.com/docs//v19.1/cost-based-optimizer.html)

This is for the locality-aware optimizer section - https://www.cockroachlabs.com/docs/dev/cost-based-optimizer.html#preferring-the-nearest-index

## Issue Description
Add a note highlighting the role caching plays here when suggesting users "verify this feature is working as expected"

## Suggested Resolution 

According to @andy-kimball,  if you're running a true geo-distributed scenario across multiple DCs, then it could take time to go fetch schema the first time we plan a query but thereafter schema should be cached. 

Note that if you have 10 nodes, you may see 10 queries with high latency due to cache misses before the latencies drop. This can lead to misleadingly high latencies that persist for some time in the web ui.

cc @piyush-singh @awoods187

rmloveland:
Fixes #4672.

### DIFF
--- a/v19.1/cost-based-optimizer.md
+++ b/v19.1/cost-based-optimizer.md
@@ -291,6 +291,14 @@ ALTER INDEX postal_codes@idx_apac CONFIGURE ZONE USING constraints='["+region=ap
 
 To verify this feature is working as expected, we'll query the database from each of our local nodes as shown below. Each node has been configured to be in a different region, and it should now be using the index pinned to that region.
 
+{{site.data.alerts.callout_info}}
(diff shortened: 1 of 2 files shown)


rmloveland:
Hopefully a quick review Andy - this is mostly an edited transcription of Nate's description in #4672 of info he got from you.  Want to make sure I didn't introduce a subtle error in translation.

PR #4956, opened 2019-06-20
resolve outstanding issues with convertflow CSS

sploiselle:


jseldess:
I think this LGTM, but I'm not sure what I exactly I should be checking. Can you provide guidance?

maria-toft:
Functionality: signing up and the scrolling/stickiness on the right-hand side of the page looks good. Placement and look and feel looks great. This looks good to go. Thanks so much for your help @sploiselle.

@jseldess heads-up that the CTA itself is just a test. The signing up for the newsletter will live at the footer as we discussed.

css/layouts/_footer.scss
```diff
@@ -1,6 +1,8 @@
 section.footer {
 	background: $cl_gray_footer;
 	padding: 0 0 20px;
+	z-index: 100;
```
j-low: nit: if possible, using as small as possible z-index values is ideal since-- in absence of some extant map of z-index values-- it minimizes the opportunity for hard-to-debug z-index conflicts to arise.  It is much easier to know if a display issue of a container is z-index related if incrementing by small values than large ones (e.g. if a `div` is being hidden unexpectedly and you are unsure if the reason is a z-index conflict or not, bumping the `div`'s z-index to 100 will definitely reveal the answer when everything in the site CSS in z-index < 10 vs. if z-indexes are of arbitrarily large values and therefore your arbitrarily large z-index test-value may or may not exceed the z-index of some unknown element that could be causing a conflict).

sploiselle:
@jseldess Sorry about that; can you check out what the implementation looks like [here](http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/18ab6de651e9402702ecf66fa1b43235086a68f9/v19.1/demo-automatic-cloud-migration.html#watch-the-demo)?

css/layouts/_footer.scss
sploiselle: Moved down to 1 and still shows the desired behaviour; thanks for catching that, Joe.

jseldess:
Thanks, @sploiselle. That page LGTM!

PR #4979, opened 2019-06-26
Fix various typos in some training docs

### LINKED ISSUE #4943: CRDB online training typos
1. https://www.cockroachlabs.com/docs/stable/training/locality-and-replication-zones.html

On step 3 after starting nodes 7,8,9 the docs state "You started nodes 7, 8, and 9 in the East region."

They should say "You started nodes 7, 8, and 9 in the **West** region." as the datacenter is us-west.

Additionally, on Step 7 the configure zone SQL statements are missing the **./** on both SQL statements.

2. https://www.cockroachlabs.com/docs/stable/training/planned-maintenance.html

the note below is from step 2, the word mot should be **not**

image

rmloveland:
Fixes #4943.

### DIFF
--- a/v19.1/training/locality-and-replication-zones.md
+++ b/v19.1/training/locality-and-replication-zones.md
@@ -173,7 +173,7 @@ Add 6 more nodes, this time using the [`--locality`](../configure-replication-zo
     --background
     ~~~
 
-    You started nodes 7, 8, and 9 in the East region.
+    You started nodes 7, 8, and 9 in the West region.
 
 ## Step 4. Write data and verify data distribution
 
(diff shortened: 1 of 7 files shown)


rmloveland:
@mattcrdb let me know if this fixes your issue.  Thanks for finding (and filing)!

mattcrdb:
Additional typo, this one raises an error in the training module:
https://www.cockroachlabs.com/docs/stable/training/under-replication-troubleshooting.html

On step 2 the instruction states to reduce the amount of time the cluster waits before considering a node dead to just 1 minute.

The minimum time to do so is 1 minute and 15 seconds, both the instruction and the code snippet should be changed. 

`./cockroach sql --insecure --host=localhost:26257 --execute="SET CLUSTER SETTING server.time_until_store_dead = '1m0s';"`

should be 

`./cockroach sql --insecure --host=localhost:26257 --execute="SET CLUSTER SETTING server.time_until_store_dead = '1m15s';"`

rmloveland:
Thanks Matt, missed that one.

Issue #4810, opened 2019-05-15
De-emphasize SAVEPOINT retries


bdarnell:
Re: [Transactions](https://cockroachlabs.com/docs//v19.1/transactions.html)

## Issue Description

The SAVEPOINT-based retry protocol is best considered an advanced optimization. It was more important in CRDB 1.x, but these days most apps are fine without it as long as they have a simple retry loop. 

Implementing savepoint retries correctly is fairly subtle and most application writers shouldn't try; it should be left to framework developers, and users of frameworks that don't support this should be fine without it. 

## Suggested Resolution 

At a minimum, remove "strongly recommended" from the savepoint option in "client-side intervention". Move the "client-side transaction retries" separate to a separate "advanced client-side retries" page and in this document only talk about basic non-savepoint retries.

awoods187:
@jseldess I think this should be a p-0 since its one of the most confusing aspects of this problem to implement. If users only need a retry loop, without any fancy adaptation, it could unblock many use cases. It doesn't solve those customers who are unwilling to add retry loops, but it does lower the burden of doing so, which may be enough.

jseldess:
Interesting! Does this mean that we should re-factor all of our "build an app" code samples to NOT emphasize SAVEPOINT? That's probably one of the most visible representations of our historical guidance.

awoods187:
I think so--but will let @bdarnell weigh in

bdarnell:
Yeah, we should remove those from the "build an app" samples. Maybe keep them around as examples on the "advanced retries" page, but not on the main docs. The idea is that if you're using SQLAlchemy, for example, you can take advantage of the built-in savepoint retries, but if your framework doesn't do it for you you shouldn't worry about it.

jseldess:
OK, I'd say this needs to be considered a larger "docs project" then. @awoods187, we'll need to review and reprioritize this with Rich. My understanding is, once he's back in office, he'll finish up https://github.com/cockroachdb/docs/issues/4621. Prior to this discussion, the next project was either Hibernate retries or https://github.com/cockroachdb/docs/issues/4693. We can't do all of it this quarter, let alone for 19.2.

jseldess:
Closing in favor of https://github.com/cockroachdb/docs/issues/5035.

PR #4996, opened 2019-07-01
Deemphasize SAVEPOINT retries (1/n)

rmloveland:
Summary of changes:

- Split discussion of client-side retries roughly in two:

  - Application developers should use an application-level retry loop
    with exponential backoff

  - Library authors should use SAVEPOINT and friends

- For application developers, update 'Transactions' page with example in
  pseudocode showing how to implement an application-level retry loop
  with exponential backoff.

- For library authors, break out discussion of client-side txn retries
  using the SAVEPOINT protocol into a separate page, "Advanced
  Client-side Transaction Retries".

Addresses the non-code-sample-update parts of #4810.

### DIFF
--- a/v19.1/advanced-client-side-transaction-retries.md
+++ b/v19.1/advanced-client-side-transaction-retries.md
@@ -1,0 +1,69 @@
+---
+title: Advanced Client-side Transaction Retries
+summary: Advanced client-side transaction retry features for library authors
+toc: true
+---
+
+This page has instructions for authors of [database drivers and ORMs](install-client-drivers.html) who would like to implement client-side retries in their database driver or ORM for maximum efficiency and ease of use by application developers.
+
+{{site.data.alerts.callout_info}}
+If you are an application developer who needs to implement an application-level retry loop, see the [Client-side intervention example](transactions.html#client-side-intervention-example).
+{{site.data.alerts.end}}
+
+## Overview
+
+To improve the performance of transactions that fail due to contention, CockroachDB includes a set of statements (listed below) that let you retry those transactions. Retrying transactions using these statements has the following benefits:
+
+1. When you use savepoints, you "hold your place in line" between attempts. Without savepoints, you're starting from scratch every time.
+2. Transactions increase their priority each time they're retried, increasing the likelihood they will succeed. This has a lesser effect than #1.
+
+## How transaction retries work
+
+A retryable transaction goes through the process described below, which maps to the following SQL statements:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> BEGIN;                                  -- #1
+> SAVEPOINT cockroach_restart;            -- #2
+-- ... various transaction statements ... -- #3
+> RELEASE SAVEPOINT cockroach_restart;    -- #5 (Or #4, ROLLBACK, in case of retry error)
+> COMMIT;
+~~~
+
+1. The transaction starts with the [`BEGIN`](begin-transaction.html) statement.
+
+2. The [`SAVEPOINT`](savepoint.html) statement declares the intention to retry the transaction in the case of contention errors. Note that CockroachDB's savepoint implementation does not support all savepoint functionality, such as nested transactions. It must be executed after [`BEGIN`](begin-transaction.html) but before the first statement that manipulates a database.
+
+3. The statements in the transaction are executed.
+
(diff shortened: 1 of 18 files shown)


rmloveland:
@bdarnell, the meat of this is in 'Transactions' and 'Advanced client-side transaction retries'.  The other pages are mostly fixing up links.

As noted above, this is the prose portion of the necessary changes (AFAICT), and leaves the code samples to be updated in a later PR.

I'm especially interested in your thoughts on the "simple retry loop" pseudocode, since it will inform how I update the code samples going forward.  

For example, the basic structure of the pseudocode made sense to me in a Perl script I wrote for testing and seems like it would broadly apply to more dynamic languages, but the test program I wrote in Java didn't seem like much of a win LOC-wise over implementing SAVEPOINT once the backoff logic is added.  But that might be a discussion for the later PR.

rmloveland:
@Amruta-Ranade ping

v19.1/transactions.md
```diff
@@ -147,72 +147,50 @@ Your application should include client-side retry handling when the statements a
...
+2. **Most users, such as application authors**: Abort the transaction using the [`ROLLBACK`](rollback-transaction.html) statement, and then reissue all of the statements in the transaction. For an example, see the [Client-side intervention example](#client-side-intervention-example).
+3. **Advanced users, such as library authors**: Use the [`SAVEPOINT`](savepoint.html) statement to create retryable transactions. Retryable transactions can improve performance because their priority is increased each time they are retried, making them more likely to succeed the longer they're in your system. For instructions showing how to do this, see [Advanced Client-Side Transaction Retries](advanced-client-side-transaction-retries.html).
+
+#### Client-side intervention example
+
+The Python-like pseudocode below shows how to implement an application-level retry loop; it does not require your driver or ORM to implement [advanced retry handling logic](advanced-client-side-transaction-retries.html), so it can be used from any programming language or environment. In particular, your retry loop must:
```
ericharmeling: This is cool... using Python or pseudocode (which Python pretty much is IMO) is a great pedagogical resource, as it can be informative for developers of all walks and talks. How often do we use pseudocode in the docs?

v19.1/advanced-client-side-transaction-retries.md
```diff
@@ -0,0 +1,70 @@
...
+
+5. Once the transaction executes all statements without encountering contention errors, execute [`RELEASE SAVEPOINT`](release-savepoint.html) to commit the changes. If this succeeds, all changes made by the transaction become visible to subsequent transactions and are guaranteed to be durable if a crash occurs.
+
+    In some cases, the [`RELEASE SAVEPOINT`](release-savepoint.html) statement itself can fail with a retry error, mainly because transactions in CockroachDB only realize that they need to be restarted when they attempt to commit. If this happens, the retry error is handled as described in step 4.
+
+## Customizing the savepoint name
+
+{% include {{ page.version.version }}/misc/customizing-the-savepoint-name.md %}
+
+## Examples
```
ericharmeling: We talked earlier today in our PM 1:1 about prioritizing Go, Java, Python, and JavaScript as the languages of choice for application developer docs... Do we have a list of application-level features/areas that are covered by current example applications? Maybe by language? (e.g. "Advanced Client-Side Transaction Retries: Go [NA], Python [check], Java [check], Node [ ]"). I can see how that kind of list would be helpful when we start looking for ways to extend and improve the example application code that we currently have.

v19.1/transactions.md
rmloveland: I think this is the first time I've done it.  In part it was helpful to me while writing, too, since it created a nice feedback loop with the description.  Also I was wondering what I would want if I had to implement this, and a prose description would be necessary but probably not sufficient.

v19.1/advanced-client-side-transaction-retries.md
rmloveland: I have a list in #5035, take a look and let me know what you think.  I think that list is just for this feature though.  If you mean "do we have a matrix-y thing that describes what features are covered by each example?" then I think the answer is no - at least I didn't find anything in [this quick search](https://github.com/cockroachdb/docs/issues?utf8=✓&q=is%3Aissue+is%3Aopen+code+sample).  Feel free to open a new issue!

PR #5000, opened 2019-07-02
Document ballast file as production best practice

### LINKED ISSUE #4932: Document ballast file as production best practice
If a store's disk fills up, it's impossible to bring that node back online without finding some files to delete (or resizing the partition). It's often possible to delete log files to free up space, but if the data lives on its own partition, there may not be any data that is freely deletable. If this only happens to one node you can wipe it and recover from its peers, but this may not be safe if more than one node fills up at once. 

A partial remedy to this situation is to use a ballast file (a large file that is unused and exists solely to give you something to delete in emergency situations). This file can be created using `cockroach debug ballast` or standard tools like `dd`. We should suggest the use of ballast files in our production best practices doc.

jseldess:
Fixes #4932.

### DIFF
--- a/_includes/sidebar-data-v19.1.json
+++ b/_includes/sidebar-data-v19.1.json
@@ -1589,6 +1589,12 @@
             ]
           },
           {
+            "title": "<code>cockroach debug ballast</code>",
+            "urls": [
+              "/${VERSION}/debug-ballast.html"
+            ]
+          },
+          {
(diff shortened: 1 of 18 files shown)


jseldess:
Opened another cockroach issue to suggest moving documented `debug` commands out of `debug`: https://github.com/cockroachdb/cockroach/issues/38662.

PR #5001, opened 2019-07-02
Add warning that nodelocal doesn't work in multi-node clusters

### LINKED ISSUE #4982: Add warning that nodelocal doesn't work in multi-node clusters
dt: i’m guessing they used `nodelocal:///` in a multi-node cluster, and the missing sst is in the extern dir on one of the other nodes (since `nodelocal` means each node writes to it’s own, local disk)
ron: they definitely used `nodelocal`
dt: `nodelocal` does *not* work in multi-node clusters.
ron: Gotcha. So they should back up to s3 then? Or something else just not node local
dt: in general, yes — the only way to backup, restore or import in multi-node clusters is to use some cloud store location. in this specific case, it is only 3 SSTs and they could _probably_ just track down the missing file by hand by looking in the local store dir on each node
dt: i just read though the docs and it isn’t all that clear https://www.cockroachlabs.com/docs/v19.1/backup.html#example-file-urls
dt: people have been making similar mistakes since BACKUP was released though
dt: see https://github.com/cockroachdb/cockroach/issues/19158, including this:
```The scariest part of this is that a user might run a BACKUP ... nodelocal:// that completes successfully and think their data is backed up and later lose data because their backup isn't usable. Ideally we should make this mistake harder to make.```

### PR DESCRIPTION
**Pull request by lhirata**
Closes #4982.

### DIFF
--- a/_includes/v19.1/misc/external-urls.md
+++ b/_includes/v19.1/misc/external-urls.md
@@ -12,7 +12,7 @@
 | S3-compatible services&nbsp;[<sup>5</sup>](#considerations) | `s3`        | Bucket name                                      | `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`&nbsp;[<sup>6</sup>](#considerations) (optional), `AWS_ENDPOINT` |
 
 {{site.data.alerts.callout_danger}}
-If you write to `nodelocal` storage in a multi-node cluster, individual data files will be written to the `extern` directories of arbitrary nodes. Reading from `nodelocal` will fail unless all the files are available in the same place on all the nodes.
+If you write to `nodelocal` storage in a multi-node cluster, individual data files will be written to the `extern` directories of arbitrary nodes.  Therefore, unless the [`--external-io-dir`](start-a-node.html#general) is configured to point to an NFS mount or other network-backed, shared storage, using `nodelocal` in a multi-node cluster will likely not work as intended.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dt** [comment]
we could explicitly say something like "thus, unless the `--external-io-dir` is configured to point to an NFS mount or other network-backed, shared storage, using `nodelocal` in a multi-node cluster will likely not work as intended."

File: _includes/v19.1/misc/external-urls.md [hunk 1]
```diff
@@ -12,7 +12,7 @@
 | S3-compatible services&nbsp;[<sup>5</sup>](#considerations) | `s3`        | Bucket name                                      | `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`&nbsp;[<sup>6</sup>](#considerations) (optional), `AWS_ENDPOINT` |
 
 {{site.data.alerts.callout_danger}}
-If you write to `nodelocal` storage in a multi-node cluster, individual data files will be written to the `extern` directories of arbitrary nodes. Reading from `nodelocal` will fail unless all the files are available in the same place on all the nodes.
+If you write to `nodelocal` storage in a multi-node cluster, individual data files will be written to the `extern` directories of arbitrary nodes.  Therefore, unless the [`--external-io-dir`](start-a-node.html#general) is configured to point to an NFS mount or other network-backed, shared storage, using `nodelocal` in a multi-node cluster will likely not work as intended.
```
dt: should that read "... `--external-io-dir` *flag* is configured ..."  or is that implied?

PR #5018, opened 2019-07-04
Known limitation: Requests to a restarted node in need of snapshots can hang

### LINKED ISSUE #4957: Known limitation: Restarted node in need of snapshots can be wedged for long time
Need to document this [known limitation](https://github.com/cockroachdb/cockroach/issues/37906) and explain the relevant workaround/response.

jseldess:
Requests to a restarted node in need of snapshots can hang.

Fixes #4957.

### DIFF
--- a/v19.1/known-limitations.md
+++ b/v19.1/known-limitations.md
@@ -8,6 +8,21 @@ This page describes newly identified limitations in the CockroachDB {{page.relea
 
 ## Unresolved limitations
 
+### Requests to restarted node in need of snapshots may hang
+
+When a node is offline, the [Raft logs](architecture/replication-layer.html#raft-logs) for the ranges on the node get truncated. When the node comes back online, it therefore often needs [Raft snapshots](architecture/replication-layer.html#snapshots) to get many of its ranges back up-to-date. While in this state, requests to a range will hang until its snapshot has been applied, which can take a long time.  
+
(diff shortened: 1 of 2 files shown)


jseldess:
___
*[v19.1/known-limitations.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/5018#-LjM9rSIECu1utVLRu4y:-LjRV_9xBRVFzjBQLO3K:b-k13xb3) ([raw file](https://github.com/cockroachdb/docs/blob/2d8669afb8fe6e6c357c5121ccaa9a62954e0115/v19.1/known-limitations.md#L15)):*
<details><summary><i>Previously, tbg (Tobias Grieger) wrote…</i></summary><blockquote>

Add a note to verify that the network interfaces support a multiple of the configured bandwidth to make sure foreground traffic isn't impacted, and to make sure to reset to the original value after the problem has resolved.
</blockquote></details>

In Slack, you clarified that this is about the physical network of the client, i.e., "if you have trouble sending 50mb/s over your network, don’t set the setting to 50mb/s or probably even 25mb/s". 

But I'm not understanding how this relates to the problem of a restarted node getting too far behind and hanging. 

Also, just so I understand, as of 19.1.3, is the workaround via `kv.snapshot.recovery.max_rate` still relevant to mention?

jseldess:
___
*[v19.1/known-limitations.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/5018#-LjM9rSIECu1utVLRu4y:-LjSbpnoDOmqoSuArfqx:b-465i9i) ([raw file](https://github.com/cockroachdb/docs/blob/2d8669afb8fe6e6c357c5121ccaa9a62954e0115/v19.1/known-limitations.md#L15)):*
<details><summary><i>Previously, tbg (Tobias Grieger) wrote…</i></summary><blockquote>

Sorry, should've clarified that that was a comment about the cluster setting workaround. If you set `max_rate` too high (where "too high" means "it will saturate your network"), that'll be untested territory and it may or may not be fine.
</blockquote></details>

Thanks, Tobi. Done.

PR #5026, opened 2019-07-08
Add vectorization to experimental features list

### LINKED ISSUE #4953: Add sql.defaults.experimental_vectorized setting to experimental features list
```
sql.defaults.experimental_vectorize                      | off                     | e            | default experimental_vectorize mode [off = 0, on = 1, always = 2]
```

Customer asked about the dangers of using this in production. Not clear. We should add to https://www.cockroachlabs.com/docs/dev/experimental-features.html with some details.

I answered:
> In the meantime, we just posted a blog on vectorizing the merge joiner that might add some clarity: https://www.cockroachlabs.com/blog/vectorizing-the-merge-joiner-in-cockroachdb/
>
> I think the experimental status pertains the underlying vectorization engine more generally. You can track the work left to be done in this GitHub issue, if you like: https://github.com/cockroachdb/cockroach/issues/36507
>
>I see this at the end:
> > the bottleneck becomes not the joiner itself but the speed at which data can be read from the disk. This is still important as a more optimized merge joiner can free up CPU cycles to be used elsewhere, on another query perhaps.

rmloveland:
... along with some context about what the feature is, how it works (by
linking to George's blog post), and a list of current limitations.

Fixes #4953.


rmloveland:
@asubiotto hoping for your review since you had commented on the issue for this.  In particular, please let me know:

- If you have a different / better description of the feature
- If you can think of other limitations users should be aware of before turning it on.  Also, should the language be more or less scary than what I have so far? :-) 
- If anything is missing or incorrect in general

rmloveland:
Thanks for the review Alfonso!

Jordan, I'll let this sit a bit before sending along to Docs team in case you would like some changes.  Thanks!

PR #5028, opened 2019-07-09
Secure app examples

ericharmeling:
Added the following to the "Build an App with CockroachDB" examples:
- C++ secure application example
- Clojure secure application example

### DIFF
--- a/_includes/v19.1/app/insecure/basic-sample.clj
+++ b/_includes/v19.1/app/insecure/basic-sample.clj
@@ -7,10 +7,6 @@
               :dbname "bank"
               :host "localhost"
               :port "26257"
-              :ssl true
-              :sslmode "require"
-              :sslcert "certs/client.maxroach.crt"
-              :sslkey "certs/client.maxroach.key.pk8"
               :user "maxroach"})
 
 (defn test-basic []
--- a/_includes/v19.1/app/insecure/basic-sample.cpp
+++ b/_includes/v19.1/app/insecure/basic-sample.cpp
@@ -10,7 +10,7 @@ using namespace std;
 int main() {
   try {
     // Connect to the "bank" database.
-    pqxx::connection c("dbname=bank user=maxroach sslmode=require sslkey=certs/client.maxroach.key sslcert=certs/client.maxroach.crt port=26257 host=localhost");
+    pqxx::connection c("postgresql://maxroach@localhost:26257/bank");
 
     pqxx::nontransaction w(c);
 
--- a/_includes/v19.1/app/insecure/txn-sample.clj
+++ b/_includes/v19.1/app/insecure/txn-sample.clj
@@ -7,10 +7,6 @@
               :dbname "bank"
               :host "localhost"
               :port "26257"
-              :ssl true
-              :sslmode "require"
-              :sslcert "certs/client.maxroach.crt"
-              :sslkey "certs/client.maxroach.key.pk8"
               :user "maxroach"})
 
 ;; The transaction we want to run.
--- a/_includes/v19.1/app/insecure/txn-sample.cpp
+++ b/_includes/v19.1/app/insecure/txn-sample.cpp
@@ -59,7 +59,7 @@ void executeTx(
 
 int main() {
   try {
-    pqxx::connection c("dbname=bank user=maxroach sslmode=require sslkey=certs/client.maxroach.key sslcert=certs/client.maxroach.crt port=26257 host=localhost");
+    pqxx::connection c("postgresql://maxroach@localhost:26257/bank");
 
     executeTx(&c, [](pqxx::dbtransaction *tx) {
           transferFunds(tx, 1, 2, 100);
--- a/v19.1/build-a-c++-app-with-cockroachdb.md
+++ b/v19.1/build-a-c++-app-with-cockroachdb.md
@@ -12,15 +12,32 @@ We have tested the [C++ libpqxx driver](https://github.com/jtv/libpqxx) enough t
 
 ## Before you begin
 
-Make sure you have already [installed CockroachDB](install-cockroachdb.html).
+{% include {{page.version.version}}/app/before-you-begin.md %}
 
 ## Step 1. Install the libpqxx driver
 
 Install the C++ libpqxx driver as described in the [official documentation](https://github.com/jtv/libpqxx).
 
-{% include {{ page.version.version }}/app/common-steps.md %}
+{{site.data.alerts.callout_info}}
+If you are running macOS, you need to install version 4.0.1 or higher of the libpqxx driver.
+{{site.data.alerts.end}}
 
-## Step 5. Run the C++ code
+<section class="filter-content" markdown="1" data-scope="secure">
+
+## Step 2. Create the `maxroach` user and `bank` database
+
+{% include {{page.version.version}}/app/create-maxroach-user-and-bank-database.md %}
+
+## Step 3. Generate a certificate for the `maxroach` user
+
+Create a certificate and key for the `maxroach` user by running the following command.  The code samples will run as this user.
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach cert create-client maxroach --certs-dir=certs --ca-key=my-safe-directory/ca.key
+~~~
+
+## Step 4. Run the C++ code
 
 Now that you have a database and a user, you'll run code to create a table and insert some rows, and then you'll run code to read and update values as an atomic [transaction](transactions.html).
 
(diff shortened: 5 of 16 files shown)


_includes/v19.1/app/basic-sample.clj
```diff
@@ -3,8 +3,14 @@
...
+(def db-spec {:dbtype "postgresql"
+              :dbname "bank"
+              :host "localhost"
+              :port "26257"
+              :ssl true
+              :sslmode "require"
+              :sslcert "certs/client.maxroach.crt"
+              :sslkey "certs/client.maxroach.key.pk8"
               :user "maxroach"
               :password ""})
```
rmloveland: This `:password` field is not needed.  Perhaps we should remove it just to make it a little clearer that we are not using pw-based auth.

_includes/v19.1/app/insecure/basic-sample.clj
```diff
@@ -0,0 +1,31 @@
...
+    (j/insert! conn :accounts {:id 1 :balance 1000})
+    (j/insert! conn :accounts {:id 2 :balance 250})
+
+    ;; Print out the balances.
+    (println "Initial balances:")
+    (->> (j/query conn ["SELECT id, balance FROM accounts"])
+         (map println)
+         doall)
+
+    ;; The database connection is automatically closed by with-db-connection.
```
rmloveland: It's not clear this comment is really needed, since in Lisp world `WITH-FOO` always implies that `FOO` is a resource of some kind that will be auto-closed if the body to be executed fails.  The comment also pushes the closing parens onto their own line, which is also not generally good Lisp style.  (But for the record these issues are not really hurting anything, so feel free to leave this stuff in.  I'm just being a Lisp nerd because I get so few chances to do so here.)

_includes/v19.1/app/insecure/txn-sample.clj
```diff
@@ -0,0 +1,43 @@
+(ns test.test
+  (:require [clojure.java.jdbc :as j]
+            [test.util :as util]))
+
+;; Define the connection parameters to the cluster.
+(def db-spec {:subprotocol "postgresql"
```
rmloveland: Here we use `:subprotocol` and friends, but in the other example we use `:dbtype`, `:host`, etc.  They both work but I think we should standardize (IDK which is better, if either)

_includes/v19.1/app/insecure/txn-sample.clj
```diff
@@ -0,0 +1,43 @@
...
+              :subname "//localhost:26257/bank"
+              :user "maxroach"
+              :password ""})
```
rmloveland: Is this `:password` key needed for the insecure examples?  If it's not I think we should drop it here also, since we're still not doing pw-based auth (in fact I don't think CRDB really does pw-based auth, IIRC)

_includes/v19.1/app/txn-sample.clj
```diff
@@ -3,8 +3,14 @@
             [test.util :as util]))
 
 ;; Define the connection parameters to the cluster.
-(def db-spec {:subprotocol "postgresql"
-              :subname "//localhost:26257/bank"
+(def db-spec {:dbtype "postgresql"
```
rmloveland: I'm deducing that the reason `:dbtype` and friends were not used here is that perhaps something didn't work in sekuritay-land did not like `:subprotocol`, etc.?  (I'm being too lazy to find the answer myself, sorry)

v19.1/build-a-clojure-app-with-cockroachdb.md
```diff
@@ -91,6 +107,100 @@ lein run
...
+{% include copy-clipboard.html %}
+~~~ clojure
+{% include {{ page.version.version }}/app/insecure/txn-sample.clj %}
+~~~
+
+Run with:
+
+{% include copy-clipboard.html %}
+~~~ shell
+lein run
```
rmloveland: We prepend shell commands with `$` (the copy-paste button automagically removes them)

_includes/v19.1/app/insecure/basic-sample.cpp
```diff
@@ -0,0 +1,41 @@
+// Build with g++ -std=c++11 basic-sample.cpp -lpq -lpqxx
```
rmloveland: I think we should update this to use a named output file, e.g. `g++ -std=c++11 basic-sample.cpp -lpq -lpqxx -o basic-sample`.  That way you have a binary called `basic-sample` which is slightly nicer.  (Actually we should have a real project with a Makefile and stuff but that is future work obviously!)

_includes/v19.1/app/insecure/txn-sample.cpp
```diff
@@ -0,0 +1,76 @@
+// Build with g++ -std=c++11 txn-sample.cpp -lpq -lpqxx
```
rmloveland: Same comment as above except `g++ -std=c++11 txn-sample.cpp -lpq -lpqxx -o txn-sample`

_includes/v19.1/app/insecure/txn-sample.cpp
```diff
@@ -0,0 +1,76 @@
...
+  while (true) {
+    try {
+      pqxx::subtransaction s(tx, "cockroach_restart");
+      fn(&s);
+      s.commit();
+      break;
+    } catch (const pqxx::pqxx_exception& e) {
+      // Swallow "transaction restart" errors; the transaction will be retried.
+      // Unfortunately libpqxx doesn't give us access to the error code, so we
+      // do string matching to identify retriable errors.
```
rmloveland: We spell this as "retryable" everywhere else.

v19.1/build-a-c++-app-with-cockroachdb.md
```diff
@@ -12,15 +12,32 @@ We have tested the [C++ libpqxx driver](https://github.com/jtv/libpqxx) enough t
...
-Make sure you have already [installed CockroachDB](install-cockroachdb.html).
+{% include {{page.version.version}}/app/before-you-begin.md %}
 
 ## Step 1. Install the libpqxx driver
 
 Install the C++ libpqxx driver as described in the [official documentation](https://github.com/jtv/libpqxx).
 
-{% include {{ page.version.version }}/app/common-steps.md %}
+{{site.data.alerts.callout_info}}
+If you are running macOS, you need to install version 4.0.1 of the libpqxx driver.
```
rmloveland: Perhaps this should say "4.0.1 or higher" ?  I was able to use the one currently available from Homebrew, which is 6.4.5, and it worked great.

rmloveland:
Oh @ericharmeling one last thing.  Would you mind squashing this before merging?  It's a bit hard to read the history otherwise

_includes/v19.1/app/basic-sample.clj
ericharmeling: You're right... If you try to use the wrong key type (not the .pk8), `lein` with give you an error that you need to specify a password. I don't think this is worth mentioning though. I'm sure you ran into the same type of thing when you created the secure Java example.

_includes/v19.1/app/insecure/txn-sample.clj
ericharmeling: Agreed! I just opted for not touching the insecure examples, but I will update to standardize. Same for the C++ examples. Most of my syntax is based on some API reference or other than I found.

_includes/v19.1/app/insecure/basic-sample.clj
ericharmeling: Let's take it out then!

_includes/v19.1/app/insecure/basic-sample.cpp
ericharmeling: Sounds good to me! I think there's a lot of work in general to be done on the application examples to make projects a little more realistic and language-specific (like what you did for the Java example). Makefiles, for example, for C++ and C (which don't yet exist...) examples.

_includes/v19.1/app/insecure/txn-sample.clj
ericharmeling: I ran into some "password required" errors, but I just ran through both examples again with the password parameter and everything worked. Must have been for the secure example when I was using the wrong type of key. I'll remove that parameter from both!

_includes/v19.1/app/insecure/txn-sample.cpp
ericharmeling: Right on.

_includes/v19.1/app/insecure/txn-sample.cpp
ericharmeling: OK  - I'll change this.

_includes/v19.1/app/txn-sample.clj
ericharmeling: I tried out using `:dbtype` and friends and they worked for the insecure example! I'll change this to make it standard across the insecure and secure examples.

v19.1/build-a-c++-app-with-cockroachdb.md
ericharmeling: Yes, I will do that. I installed 4.0.0 for some reason(?), and it didn't have this mac-specific bug fixed yet: https://github.com/Homebrew/legacy-homebrew/issues/23627. Your language is more accurate.

v19.1/build-a-clojure-app-with-cockroachdb.md
ericharmeling: Ah OK. I will add those back in.

ericharmeling:
> Oh @ericharmeling one last thing. Would you mind squashing this before merging? It's a bit hard to read the history otherwise

For sure!

PR #4321, opened 2019-02-05
Explain txn retry errors.

knz:
This follows a discussion last week through various support channels.

The docs are good on what to do in clients when the error occurs, but  we don't explain the how and why of these errors very well.

This PR fixes that (initial draft co-authored with justin).

@andreimatei @bdarnell please have a look at the content. Remember this needs to be accurate but remain high-level (this is not for an audience interested in the crdb internals).

@rmloveland please take over the PR for further edits and integrations in the rest of the docs site. I think it should be linked from various places: the SQL FAQs, the page on transaction contention, and in all pages where the client-side retry loop is documented.

PR #5033, opened 2019-07-09
Add `ReadWithinUncertainty...` to Common Errors

### LINKED ISSUE #4331: Add docs on contention errors (ReadWithinUncertaintyIntervalError). Why they exist and how to handle and reduce them
In https://github.com/cockroachdb/cockroach/issues/34517, it became apparent that receiving contention errors like ReadWithinUncertaintyIntervalError can be quite confusing to our users. 

As discussed in https://github.com/cockroachdb/cockroach/issues/34517, this process is working by design. We should explain that retry errors are always possible and users need to be prepared to address them via retry loops in their application, if they do not already have one. Note that retry loops are good hygiene for any application even if they were to not use CRDB. (Let's make sure this point itself is a well covered in docs). We should also explain how CRDB has worked to reduce the number of these errors. 

The first select in any transaction can return an uncertainty error. 

https://github.com/cockroachdb/cockroach/issues/34517 reminded us that the second select can also cause an uncertainty error. The uncertainty window has decreased between the first select and the second but it is not zero and therefore can trigger this occurrence. While it will be rare to see a transaction with the same select twice, foreign key checks can create an implicit second select which can also demonstrate this behavior. 

The longer a transaction goes, the less likely it is for a user to experience this error. This is because while the amount of concurrent data can increase, it is significantly outweighed by the reduction in the uncertainty window. 

To address these contention errors users can:
- Redesign their schema and queries to reduce contention (beneficial for performance as well)
- Read at historical time stamps using AS OF SYSTEM TIME

rmloveland:
Fixes #4331 (along with #4402 that already did most of the work).

### DIFF
--- a/v19.1/common-errors.md
+++ b/v19.1/common-errors.md
@@ -69,6 +69,8 @@ To understand how transactions work in CockroachDB, and why transaction retries 
 
 ### read within uncertainty interval
 
+(Error string includes: `ReadWithinUncertaintyIntervalError`)
+
 Uncertainty errors can occur when two transactions which start on different gateway nodes attempt to operate on the same data at close to the same time. The uncertainty comes from the fact that we cannot tell which one started first - the clocks on the two gateway nodes may not be perfectly in sync.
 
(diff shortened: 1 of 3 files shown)


rmloveland:
> Once this merges, please add a comment to #4321 and then close it.

Done.  Thanks for the review!

Issue #5042, opened 2019-07-11
Single-region Kubernetes on EKS


jseldess:
[Our single-region Kubernetes tutorial](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html#step-1-start-kubernetes) currently covers only GKE for hosted Kubernetes offerings. We should expand to include EKS. 

The process shouldn't be all that different from EKS, just with an AWS cli: https://docs.aws.amazon.com/eks/latest/userguide/getting-started-eksctl.html.

jseldess:
Notes:

- Insecure with configs and helm work fine. I'll start with those.
- Secure with standard configs doesn't work due to some issue with Kubernetes-created CSRs on EKS.  Need to use these different configs to use certs generated outside of Kubernetes. I'll do this second.
- Secure with helm won't work because it uses standard configs. We will probably need distinct helm charts to GKE and EKS. Opened this issue: https://github.com/cockroachdb/cockroach/issues/38847. Secure with helm on EKS is therefore out-of-scope for now.

jseldess:
cc: @kannanlakshmi, @DuskEagle, and @keith-mcclellan, who may be interested in this progress.

PR #5047, opened 2019-07-12
Single-region insecure cluster on EKS tutorial

jseldess:
How to run insecure on EKS using configs and helm.

Running secure will be handled in a separate PR. 
- Secure with standard configs doesn't work due to some issue with Kubernetes-created CSRs on EKS. Will need to use different configs to use certs generated outside of Kubernetes. 
- Secure with helm won't work because it uses standard configs. We will probably need distinct helm charts to GKE and EKS. Opened this issue: cockroachdb/cockroach#38847. Secure with helm on EKS is therefore out-of-scope for now.

Addresses #5042.

### DIFF
--- a/_includes/v19.1/orchestration/kubernetes-scale-cluster.md
+++ b/_includes/v19.1/orchestration/kubernetes-scale-cluster.md
@@ -2,6 +2,7 @@ The Kubernetes cluster contains 4 nodes, one master and 3 workers. Pods get plac
 
 1. Add a worker node:
     - On GKE, [resize your cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/resizing-a-cluster).
+    - On EKS, resize your [Worker Node Group](https://docs.aws.amazon.com/eks/latest/userguide/update-stack.html).
     - On GCE, resize your [Managed Instance Group](https://cloud.google.com/compute/docs/instance-groups/).
     - On AWS, resize your [Auto Scaling Group](https://docs.aws.amazon.com/autoscaling/latest/userguide/as-manual-scaling.html).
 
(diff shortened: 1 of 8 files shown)


jseldess:
A few issues found during testing that are probably independent of these instructions:

When I started a cluster via helm on EKs, the `cockroach` version was 2.1.1:

```
~$ kubectl run cockroachdb -it --image=cockroachdb/cockroach:v19.1.2 --rm --restart=Never -- node status --insecure --host=my-release-cockroachdb-public
  id |                                     address                                     | build  |            started_at            |            updated_at            | is_available | is_live
+----+---------------------------------------------------------------------------------+--------+----------------------------------+----------------------------------+--------------+---------+
   1 | my-release-cockroachdb-0.my-release-cockroachdb.default.svc.cluster.local:26257 | v2.1.1 | 2019-07-12 17:33:24.152488+00:00 | 2019-07-12 17:42:28.706545+00:00 | true         | true
   2 | my-release-cockroachdb-1.my-release-cockroachdb.default.svc.cluster.local:26257 | v2.1.1 | 2019-07-12 17:33:27.48001+00:00  | 2019-07-12 17:42:27.548286+00:00 | true         | true
   3 | my-release-cockroachdb-2.my-release-cockroachdb.default.svc.cluster.local:26257 | v2.1.1 | 2019-07-12 17:39:02.015971+00:00 | 2019-07-12 17:42:29.033669+00:00 | true         | true
(3 rows)
```

Don't understand why that would be given the chart says 19.1.2: https://github.com/helm/charts/blob/master/stable/cockroachdb/values.yaml#L8

Also, the helm config doesn't seem to have the `app` label, which we rely on to delete resources with one fell swoop (and maybe for other reasons). In the absence of that, I'll probably have to update the "Stop the cluster" section for helm.

DuskEagle:
I don't see a reason why we couldn't add an `app` label to our Helm charts. It seems a lot of other Helm charts either have an `app` or a `app.kubernetes.io/name` label.

DuskEagle:
I think the v2.1.1 happened because you didn't run `helm repo update` before starting. Could you add that to the instructions?

Issue #5035, opened 2019-07-10
Deemphasize SAVEPOINT retries in code samples


rmloveland:
Richard Loveland (rmloveland) commented:

This issue is to handle the "update all the code samples" portion of #4810.  (PR #4996 will address #4810 at a prose/pseudocode level.)

List of samples that need updating is below.  This may need to be broken down into sub-lists for prioritization.  According to our website analytics, the highest page views are for Go, Java, Python, and Node.js, followed at some distance by the rest.

- [x] Go (GORM)
  - [x] insecure (https://github.com/cockroachdb/docs/pull/5084)
  - [x] secure (#5163)

- [x] Go (pq) (N/A since we provide a lib to handle retries)
- [x] Java (JDBC)
  - [x] insecure (#5088)
  - [x] secure (#5172)

- [x] Java (Hibernate)
  - [x] insecure (#5537)
  - [x] secure (#5537)

- [x] Python (psycopg2)
  - [x] insecure (https://github.com/cockroachdb/docs/pull/5085)
  - [x] secure (#5173)
- [X] Python (SQLAlchemy) (N/A since we provide a SQLAlchemy adapter to handle retries)

- [x] Node.js (pg)
- [ ] Node.js (sequelize)
- [ ] Ruby (pg)
- [ ] Ruby (ActiveRecord)
- [ ] C++
- [ ] C#
- [ ] Clojure
- [ ] PHP
- [ ] Rust

Jira Issue: DOC-290

PR #5049, opened 2019-07-15
Update GORM insecure example with app retry loop

rmloveland:
... and update the surrounding descriptive text to match the new code.

Addresses #5035.

### DIFF
--- a/_includes/v19.1/app/insecure/gorm-sample.go
+++ b/_includes/v19.1/app/insecure/gorm-sample.go
@@ -1,0 +1,126 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"math"
+	"math/rand"
+	"time"
+
+	// Import GORM-related packages.
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+)
+
+// Account is our model, which corresponds to the "accounts" database table.
+type Account struct {
+	ID      int `gorm:"primary_key"`
+	Balance int
+}
+
+func main() {
+	// Connect to the "bank" database as the "maxroach" user.
+	const addr = "postgresql://maxroach@localhost:26257/bank?sslmode=disable"
+	db, err := gorm.Open("postgres", addr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	db.LogMode(false)
+
+	// Automatically create the "accounts" table based on the Account
+	// model.
+	db.AutoMigrate(&Account{})
+
+	// Insert two rows into the "accounts" table.
+	db.Create(&Account{ID: 1, Balance: 1000})
+	db.Create(&Account{ID: 2, Balance: 250})
+
+	printBalances(db)
+
+	// Transfer funds between accounts. To handle any possible
+	// transaction retry errors, we add a retry loop with exponential
+	// backoff to the transfer logic (see below).
+
+	var maxRetries = 3
+	var amount = 100
+
+	var fromAccount Account
+	var toAccount Account
+
+	db.First(&fromAccount, 1)
+	db.First(&toAccount, 2)
+
+retryLoop:
+	for retries := 0; retries <= maxRetries; retries++ {
+
+		if retries == maxRetries {
+			fmt.Errorf("hit max of %d retries, aborting", retries)
+			break retryLoop
+		}
+
+		err := transferFunds(db, fromAccount, toAccount, amount)
+
+		if err != nil {
+			// If there's an error, we try again after sleeping an
+			// increased amount of time, a.k.a. exponential backoff.
+			fmt.Println("error: ", err)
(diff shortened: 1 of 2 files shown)


rmloveland:
@bdarnell this is a first crack at some actual code.  Was hoping you'd take a look and validate the general approach.  I am not at all sure this is "the right way" to do this in Go!

Notably, GORM apparently does not expose PG error codes (according to a GH issue linked in the code sample), but if you know a way to get at them I'd love to have that check in the sample.  Right now it just retries on any type of GORM DB error running the "transfer funds" query.

bdarnell:
The one thing I'd want to see is to print the error with `%T` in addition to `%v` to see what type it is. If the underlying error is a `pq.Error`, you could cast it and get the error code that way. From the log in pastebin, it's not obviously wrapped in a way that would prevent that.

And as a last resort, we could always look for the string `restart transaction:`. That would be a little ugly but better than nothing.

rmloveland:
> The one thing I'd want to see is to print the error with %T in addition to %v to see what type it is. If the underlying error is a pq.Error, you could cast it and get the error code that way. From the log in pastebin, it's not obviously wrapped in a way that would prevent that.

That does appear to be the case.  I'll try that approach, thanks.

~~~ go
if err := db.Exec("SELECT crdb_internal.force_retry('1s':::INTERVAL)").Error; err != nil {
	fmt.Printf("db.Error type: %T \n", err)
	// prints "db.Error type: *pq.Error"
}
~~~

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,126 @@
...
+		err := fmt.Errorf("account %d balance %d is lower than transfer amount %d", fromAccount.ID, fromAccount.Balance, amount)
+		return err
+	}
+
+	// We would like to check for the 40001 error code below, but it
+	// isn't clear how to get at it from GORM. According to
+	// https://github.com/jinzhu/gorm/issues/17, GORM does not expose
+	// the PG error codes. Therefore, we retry on any error.
+
+	db.Exec("BEGIN")
```
bdarnell: Transaction statements like BEGIN are special and can't be executed on a plain `DB` object. You have to use `txn := db.Begin()` and do the rest of the statements on `txn` instead of `db` (and similarly for the `Commit` and `Rollback` methods).

rmloveland: > Transaction statements like BEGIN are special and can't be executed on a plain DB object. You have to use txn := db.Begin() and do the rest of the statements on txn instead of db (and similarly for the Commit and Rollback methods).

I tried changing everything to use `txn := db.Begin()` and `txn.Foo()` everywhere, but the code no longer works in that I could no longer see the explicit `BEGIN` etc. statements being passed to the DB, nor could I cast the Error object and get the return code.  I pasted the code and output to https://pastebin.com/ELtLu5yk

rmloveland:
> It looks like you forgot to push your new changes.

Weird, I thought I force-pushed everything the other day.

OK, just tried again a few times and it worked in the terminal but the web page for this PR is not updating, nor is a new docs build kicking off AFAICT.  Possible Github bug?

``` shell
$ date
Wed Jul 24 13:51:00 2019
$ git push --verbose --force origin 20190710-moar-savepoint-reduction 
Pushing to git@github.com:cockroachdb/docs
To github.com:cockroachdb/docs
 = [up to date]        20190710-moar-savepoint-reduction -> 20190710-moar-savepoint-reduction
updating local tracking ref 'refs/remotes/origin/20190710-moar-savepoint-reduction'
Everything up-to-date
```

_includes/v19.1/app/insecure/gorm-sample.go
bdarnell: `db.Exec("BEGIN")` will work as long as you only have one thread, but it's not safe with multiple threads because of the way connection pools work (unless gorm's connection pool is significantly different from `database/sql`'s). So we do need to get to the bottom of this.

It looks like you are successfully casting the error to `pq.Error`; if that was failing you'd get a panic.

rmloveland:
Still now accepting pushes -- Closing in favor of #5084.  That PR should have all of the updates based on your feedback (minus the `txn := db.Begin()` recommendations, which I still haven't figured out how to get working - details in pastebin in previous comment).

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: > db.Exec("BEGIN") will work as long as you only have one thread, but it's not safe with multiple threads because of the way connection pools work (unless gorm's connection pool is significantly different from database/sql's). So we do need to get to the bottom of this.

OK.  I'll revise to use `txn = db.Begin()` and friends and see if I can find the incantations it wants to make it work with retry errors.

PR #5050, opened 2019-07-15
Fix kv storage examples on all interleaving pages

### LINKED ISSUE #4642: Key value storage example shows all keys on same line
On https://www.cockroachlabs.com/docs/stable/interleave-in-parent.html#key-value-storage-example all of the example keys are on the same line:
Screenshot of Safari (9-04-19, 6-15-25 AM)

I think the intention was for each key to be on a separate line. That's how it shows in GitHub's [markdown rendering](https://github.com/cockroachdb/docs/blob/4ca8a8a51773b03d1ab82aeb18f4d4741c65d930/v19.1/interleave-in-parent.md), but not in the published version.

Screenshot of Safari (9-04-19, 6-17-14 AM)

rmloveland:
Fixes #4642.

### DIFF
--- a/v1.0/interleave-in-parent.md
+++ b/v1.0/interleave-in-parent.md
@@ -141,16 +141,16 @@ It can be easier to understand what interleaving tables does by seeing what it l
   (2, 1003, 70.00);
 ~~~
 
-Using an illustrative format of the key-value store (keys are represented in colors; values are represented by `-> value`), the data would be written like this:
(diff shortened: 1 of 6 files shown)


rmloveland:
@jseldess this is just a formatting fix, but it's for a user-filed issue (#4642) so I figured I'd do it while I'm looking at INTERLEAVE due to some other work.  PTAL.

Issue #5062, opened 2019-07-16
Update demo and workload flags for new workload(s) in 19.2


ericharmeling:
- docs/v19.2/cockroach-demo.md
- docs/v19.2/cockroach-workload.md

AFAIK `movr` is the only newly-whitelisted workload/demo for 19.2...

ericharmeling:
#5075

Issue #5061, opened 2019-07-16
MovR Documentation


ericharmeling:
We need to create a documentation section that introduces the MovR application and dataset, and provides instructions for setting up the database and tables, to run the application.

This page will include the following:

- A general overview of the MovR application and dataset.
- Instructions on using `cockroach demo` and `cockroach workload` for MovR.
- Instructions on manual setup (perhaps first on a simple local cluster, and then on a larger, distributed GCP cluster).

_Originally posted by @ericharmeling in https://github.com/cockroachdb/docs/issues/2849#issuecomment-509316569_

See also https://github.com/cockroachdb/movr/issues/90 and https://github.com/cockroachdb/docs/issues/5062

ericharmeling:
See #5075

Issue #5041, opened 2019-07-11
SQL Statement examples with MovR (First Round)


ericharmeling:
After #5040, update [SQL statement](https://www.cockroachlabs.com/docs/stable/sql-statements.html) examples to use MovR dataset.

Let's start with the statements used in #5040.

_Originally posted by @ericharmeling in https://github.com/cockroachdb/docs/issues/2849#issuecomment-509316569_

ericharmeling:
First commit: https://github.com/ericharmeling/docs/commit/7b0447de99affc2e70228866e2707060866adadb

Issue #5040, opened 2019-07-11
Learn CockroachDB SQL with MovR


ericharmeling:
Update first example: [Learn CockroachDB SQL](https://www.cockroachlabs.com/docs/dev/learn-cockroachdb-sql.html)

I'm thinking we leave the original example as-is, but change the title to "Learn CockroachDB SQL with a Running Cluster", place it under "Learn CockroachDB SQL" in the ToC, with new "Overview" and a "Learn CockroachDB SQL with Temporary, In-memory Cluster" sections.

@rmloveland Thoughts?

I've done this in my [movr-update](https://github.com/ericharmeling/docs/tree/movr-update) branch:
https://github.com/ericharmeling/docs/commit/33e5ec87da0aaa26b1b9b932292709c5c9e178ec 

Feel free to check that out to see what I'm talking about!

Another idea: we could use those section switches that are used throughout the docs to switch "Secure" and "Insecure" examples (e.g. `<section class="filter-content" markdown="1" data-scope="movr">`).

_Originally posted by @ericharmeling in https://github.com/cockroachdb/docs/issues/2849#issuecomment-509316569_

ericharmeling:
PR: #5216

PR #5075, opened 2019-07-19
First round of movr updates

ericharmeling:
This is the very first round of `movr` doc updates. These changes are all for 19.2 doc.

This PR includes the following:
- A section dedicated to explaining the MovR example (movr.md) #5061
- The `movr` command added to the `cockroach demo` and `cockroach workload` command documentation for 19.2 (cockroach-demo.md, cockroach-workload.md) #5062
- Updates to "Learn CockroachDB SQL", to use the `movr` database instead of other generic examples (learn-cockroachdb-sql-movr.md) #5040
- Updates to a few SQL statement example sections, including INSERT (insert.md), SHOW TABLES (show-tables.md), SHOW COLUMNS (show-columns.md), DROP TABLE (drop-table.md), and CREATE TABLE (create-table.md) #5041 ... Many more of these to come!

Note that I've also changed/added several other include files where necessary. I suspect there are plenty of changes that need to be made to this PR upon review. No rush!


rmloveland:
Can you please squash and force push at your leisure?  Will make review easier for me.
(Hoping you don't see this or act on it until after your vacation :-} )

_includes/sidebar-data-v19.2.json
```diff
@@ -43,7 +43,7 @@
           {
             "title": "Learn CockroachDB SQL",
             "urls": [
-              "/${VERSION}/learn-cockroachdb-sql.html"
+              "/${VERSION}/learn-cockroachdb-sql-movr.html"
```
rmloveland: We should not do this IMO for several reasons: (1) we throw away all our PageRank on the existing page built up over several years, (2) the existing page becomes an orphan since AFAICT it's still in the site.  We should instead update the existing "Learn CRDB SQL" page with the MovR content IMO

_includes/v19.2/sql/movr-start.md
```diff
@@ -0,0 +1,6 @@
+- Run [`cockroach demo movr`](cockroach-demo.html) to open an interactive SQL shell to a temporary, in-memory cluster with the `movr` database preloaded and set as the [current database](sql-name-resolution.html#current-database).
+- Use [`cockroach workload`](cockroach-workload.html):
+ 1. Start up a [secure](secure-a-cluster.html) or [insecure](start-a-local-cluster.html) local cluster.
+ 1. Run `cockroach workload init movr` with the appropriate flags and [connection string](connection-parameters.html) to initialize and populate the `movr` database on your running cluster.
```
rmloveland: Could we show the exact command string here for at least the simplest case?  And keep the links for people who need something more complicated

_includes/v19.2/sql/movr-start.md
```diff
@@ -0,0 +1,6 @@
...
+ 1. Open an interactive SQL shell to the cluster with the [`cockroach sql`](use-the-built-in-sql-client.html) command.
```
rmloveland: Same comment as in some other places re: could we show the exact `$ cockroach sql --insecure ...` command?  Would make it easier for users to copypasta and get through the instructions

_includes/v19.2/sql/movr-start.md
```diff
@@ -0,0 +1,6 @@
...
+ 1. Set `movr` as the [current database](sql-name-resolution.html#current-database) for the session with the [`SET`](set-vars.html) statement.
```
rmloveland: Same comment as before re: could we show the explicit command for copypasta?  also, I think it's easiest to `> USE movr;` instead of `SET ...`

_includes/v19.2/sql/movr-statements-no-data.md
```diff
@@ -0,0 +1,10 @@
+The following examples use MovR, a fictional vehicle-sharing application, to demonstrate CockroachDB SQL statements. For more information about the MovR example application, see [MovR: A Global Ride-sharing App](movr.html).
```
rmloveland: I think we should standardize on either ride-sharing or vehicle-sharing.  AFAICT it's really more "vehicle sharing" since there is no drivers table, etc.  Is that correct?

_includes/v19.2/sql/movr-statements.md
```diff
@@ -0,0 +1,5 @@
+The following examples use MovR, a fictional vehicle-sharing application, to demonstrate CockroachDB SQL statements. For more information about the MovR example application and dataset, see [MovR: A Global Ride-sharing App](movr.html).
+
+To follow along with the example statements, do one of the following:
+
+{% include {{page.version.version}}/sql/movr-start.md %}
```
rmloveland: Is it necessary to have the double-layer of includes here?  Maybe I'm not understanding the whole structure of this.
General comment on the include strategy: I wonder also if we could get away with having one `movr-statements.md` include rather than having one each for with/without data.  Maybe there's a way to write it up as a choice, like "in case A, do this; in case B, do this"?

v19.2/create-table.md
```diff
@@ -103,140 +105,132 @@ In CockroachDB, every table requires a [primary key](primary-key.html). If one i
...
-| logon      | primary    |   false    |            1 | rowid       | ASC       |  false  |  false   |
-+------------+------------+------------+--------------+-------------+-----------+---------+----------+
+  users      | primary    |   false    |            1 | rowid       | ASC       |  false  |  false
 (1 row)
 ~~~
 
 ### Create a table (primary key defined)
 
-In this example, we create a table with three columns. One column is the [`PRIMARY KEY`](primary-key.html), another is given the [`UNIQUE` constraint](unique.html), and the third has no constraints. The `PRIMARY KEY` and column with the `UNIQUE` constraint are automatically indexed.
+In this example, we create the `users` table, but with some column [constraints](constraints.html). One column is the [`PRIMARY KEY`](primary-key.html), another is given the [`UNIQUE` constraint](unique.html), and another has both a `UNIQUE` constraint and a [`CHECK`](check.html) constraint that limits the length of the string. The `PRIMARY KEY` and column with the `UNIQUE` constraint are automatically indexed.
```
rmloveland: I think our convention is to refer to these things in prose without upper-case or monospace as a "primary key", "unique constraint", "check constraint", etc., except in cases where we are directly referring to SQL statement text.  Although I was not able to find this in [our style guide](https://github.com/cockroachdb/docs/blob/master/STYLE.md), so maybe we need an update there.  But I'm pretty sure this is our convention based on convos I've had during PR reviews.

v19.2/create-table.md
```diff
@@ -103,140 +105,132 @@ In CockroachDB, every table requires a [primary key](primary-key.html). If one i
...
-| logoff     | logoff_user_email_key |   false    |            2 | user_id     | ASC       |  false  |   true   |
-+------------+-----------------------+------------+--------------+-------------+-----------+---------+----------+
-(3 rows)
+table_name |   index_name   | non_unique | seq_in_index | column_name | direction | storing | implicit
++------------+----------------+------------+--------------+-------------+-----------+---------+----------+
+users      | primary        |   false    |            1 | id          | ASC       |  false  |  false
+users      | users_name_key |   false    |            1 | name        | ASC       |  false  |  false
+users      | users_name_key |   false    |            2 | id          | ASC       |  false  |   true
+users      | users_dl_key   |   false    |            1 | dl          | ASC       |  false  |  false
+users      | users_dl_key   |   false    |            2 | id          | ASC       |  false  |   true
```
rmloveland: This table appears to be a little out of alignment. #nitpicking

v19.2/create-table.md
```diff
@@ -264,74 +258,108 @@ In this example, we use `ON DELETE CASCADE` (i.e., when row referenced by a fore
...
-|            | REFERENCES customers (id) ON DELETE CASCADE,                             |
-|            |                                                                          |
-|            |     INDEX orders_auto_index_fk_customer_id_ref_customers (customer_id    |
-|            | ASC),                                                                    |
-|            |                                                                          |
-|            |     FAMILY "primary" (id, customer_id)                                   |
-|            |                                                                          |
-|            | )                                                                        |
-+------------+--------------------------------------------------------------------------+
+> SHOW CREATE vehicles;
```
rmloveland: This could be slightly more specific as `SHOW CREATE TABLE` but IDK if it matters.  Either way it works!

v19.2/create-table.md
```diff
@@ -264,74 +258,108 @@ In this example, we use `ON DELETE CASCADE` (i.e., when row referenced by a fore
...
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM vehicles;
+~~~
+
+~~~
+                   id                  |  city   | type |               owner_id               | creation_time | status | current_location | ext
++--------------------------------------+---------+------+--------------------------------------+---------------+--------+------------------+------+
+  fc6f7a8c-4ba9-42e1-9c37-7be3c906050c | seattle | NULL | 26da1fce-59e1-4290-a786-9068242dd195 | NULL          | NULL   | NULL             | NULL
```
rmloveland: Interesting, so the DB schema doesn't auto-fill a creation time.  Huh.  Not material to this review, just a possible MovR code/DB update for the future, perhaps?  TBH it strikes me as odd that this field, status, and current location are allowed to be NULL.  Particularly location and status.

v19.2/drop-table.md
```diff
@@ -62,52 +65,82 @@ DROP TABLE
...
+  users
+  vehicle_location_histories
+  vehicles
+(5 rows)
 ~~~
 
 ### Remove a table and dependent objects with `CASCADE`
 
-In this example, a view depends on the table being dropped. Therefore, it's only possible to drop the table while simultaneously dropping the dependent view using `CASCADE`.
+In this example, a foreign key from a different table references the table being dropped. Therefore, it's only possible to drop the table while simultaneously dropping the dependent foreign key constraint using `CASCADE`.
```
rmloveland: It might be nice to make the first mention of "foreign key" into a link to the Foreign Key Constraint doc.

v19.2/drop-table.md
```diff
@@ -62,52 +65,82 @@ DROP TABLE
...
-| table_name         |
-+--------------------+
-| accounts           |
-| user_accounts_view |
-+--------------------+
-(2 rows)
+pq: "users" is referenced by foreign key from table "vehicles"
 ~~~
 
+You can use the [`SHOW CREATE`](show-create.html) statement to see how the columns in a table are created, including data types, default values, indexes, and constraints.
```
rmloveland: It might be nice to preface this with some context, something like "To see how 'users' is referenced from 'vehicles', use the `SHOW CREATE` ..."

v19.2/drop-table.md
```diff
@@ -116,15 +149,41 @@ DROP TABLE
...
-(0 rows)
+          table_name
++----------------------------+
+  rides
+  user_promo_codes
+  vehicle_location_histories
+  vehicles
+(4 rows)
+~~~
+
```
rmloveland: It might be good to add some explication here between these statements like "Now, we can verify the FK has been removed by running SHOW CREATE again and seeing that the XXX_fk is gone" ... or similar.

v19.2/insert.md
```diff
@@ -67,68 +67,81 @@ key. Using `ON CONFLICT` is therefore more flexible.
...
-All of the examples below assume you've already created a table `accounts`:
-
-{% include copy-clipboard.html %}
-~~~ sql
-> CREATE TABLE accounts(
-    id INT DEFAULT unique_rowid() PRIMARY KEY,
-    balance DECIMAL
-);
-~~~
+{% include {{page.version.version}}/sql/movr-statements.md %}
```
rmloveland: Left comments on `movr-statements.md` already, but wanna reiterate question of whether we can revamp to add "logic" to the text so we don't need `movr-statements-no-data.md`.

v19.2/insert.md
```diff
@@ -67,68 +67,81 @@ key. Using `ON CONFLICT` is therefore more flexible.
...
-   1 | 10000.50
-(1 row)
+                   id                  |   city   |       name       |           address           | credit_card
++--------------------------------------+----------+------------------+-----------------------------+-------------+
+  00000000-0000-4000-8000-000000000000 | new york | Robert Murphy    | 99176 Anderson Mills        | 8885705228
+  051eb851-eb85-4ec0-8000-000000000001 | new york | James Hamilton   | 73488 Sydney Ports Suite 57 | 8340905892
+  0a3d70a3-d70a-4d80-8000-000000000002 | new york | Judy White       | 18580 Rosario Ville Apt. 61 | 2597958636
+  0f5c28f5-c28f-4c00-8000-000000000003 | new york | Devin Jordan     | 81127 Angela Ferry Apt. 8   | 5614075234
+  147ae147-ae14-4b00-8000-000000000004 | new york | Catherine Nelson | 1149 Lee Alley              | 0792553487
+  c28f5c28-f5c2-4000-8000-000000000026 | new york | Petee            | 101 5th Ave                 | 1234567890
```
rmloveland: Movr comment, not docs comment: we should update the credit card handling to be more realistic at some point, hashed, elliptic curved, or "just use Stripe", or whatever the correct behavior should be - not plaintext.

v19.2/insert.md
```diff
@@ -439,204 +471,195 @@ IDs:
...
 ~~~
 
 The printed values would look like:
 
 ~~~
 IDs:
-{ id: '190019066706952193' }
-{ id: '190019066706984961' }
+{ id: 'cdd379e3-2d0b-4622-8ba8-4f0a1edfbc8e' }
+{ id: '4224b360-b1b0-4e4d-aba2-a35c64cdf404' }
```
rmloveland: All of these code sample updates LGTM at first blush.  After I get through the initial prose/doc review, I'll take a second pass through where I check all the SQL statements and code snippets.  Gosh we need automated testing :-)

v19.2/learn-cockroachdb-sql-movr.md
```diff
@@ -0,0 +1,381 @@
+---
+title: Learn CockroachDB SQL
```
rmloveland: I don't understand why this is a new file named `learn-cockroachdb-sql-movr.md` and not an update to the existing `learn-cockroachdb-sql.md`?  Aha, it looks like you pinged me in #5040 but I didn't get an email, weird.  My opinion is that we should do the minimal update to edit the existing "Learn CockroachDB SQL" page to make it "movr-compatible", since I think making "Learn CockroachDB SQL" a family of pages with different options for starting the database is adding complexity to the "user journey", and using "cockroach demo" vs. spinning up a local cluster is sort of a separate issue we should decide how to address / standardize in our docs IMO.  But let's discuss!  We should def get Jesse's opinion when he returns from vacation, I'm curious about his thoughts also.

v19.2/learn-cockroachdb-sql-movr.md
```diff
@@ -0,0 +1,381 @@
...
+summary: Learn some of the most essential CockroachDB SQL statements on a local cluster.
+toc: true
+build_for: [standard, managed]
+---
+
+This tutorial walks you through some of the most essential CockroachDB SQL statements, using the `movr` dataset. For a complete list of supported SQL statements and related details, see [SQL Statements](sql-statements.html). For more information about the `movr` dataset and the MovR example application, see [MovR: A Global Ride-sharing App](movr.html).
```
rmloveland: These 2 finishing sentences would be easier to scan if they were each on their own line at the end of this intro

v19.2/learn-cockroachdb-sql-movr.md
```diff
@@ -0,0 +1,381 @@
...
+
+{{site.data.alerts.callout_info}}
+CockroachDB aims to provide standard SQL with extensions, but some standard SQL functionality is not yet available. See our [SQL Feature Support](sql-feature-support.html) page for more details.
```
rmloveland: I think this note was already here, but I've filed https://github.com/cockroachdb/docs/issues/5125 to try to remove it

v19.2/learn-cockroachdb-sql.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Learn CockroachDB SQL
-summary: Learn some of the most essential CockroachDB SQL statements.
+title: Learn CockroachDB SQL with a Running Cluster
```
rmloveland: Hm, from looking at the auto-built test site, it looks like this file is an orphan from the TOC now.  I left a comment above on the TOC JSON that is basically: I don't think we should add a new "Learn CRDB SQL MovR" page, I think we should edit the existing page with the MovR updates.  Reasoning is explained in that comment.

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+MovR is a fictional vehicle-sharing company created to demonstrate CockroachDB's features.
+
+## Overview
+
+The MovR example consists of three components:
+
+- The `movr` database, which contains several tables that hold vehicle, user, and ride data. The `movr` database is built into `cockroach demo` and `cockroach workload`.
+- The `movr` dataset, which consists of rows of data that populate tables in the `movr` database. The `movr` dataset is built into `cockroach demo` and `cockroach workload`.
+- The MovR application is a fully-functional vehicle-sharing application, written in Python. All of MovR application source code is open-source, and available on the [movr](https://github.com/cockroachdb/movr) GitHub repository.
```
rmloveland: All of the mentions (or at least the first mention) of `cockroach demo` and `cockroach workload` could link to the docs for those commands.

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+## The `movr` Database
+
+Three tables in the `movr` database store most of the user, vehicle, and ride data for MovR:
```
rmloveland: It may be correct that they store "most of" the data, but perhaps for simplicity of introducing this it could be glossed over?  Genuinely asking because I don't actually know movr that well.  Alternatively, if we do keep it in, IMO we should say where "the rest of it" is.

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+Each table has a composite primary key of `city` and `id`, with `city` being first in the key.
+
+<img src="{{ 'images/v19.2/geo-partitioning-schema.png' | relative_url }}" alt="Geo-partitioning schema" style="max-width:100%" />
+
+The `movr` database also contains the `promo_codes`, `user_promo_codes`, and `vehicle_location_histories` tables.
+
+## Using `cockroach` Commands to Generate Schemas and Data for MovR
+
+Use `cockroach` commands to load the `movr` database and dataset that are built into the CockroachDB binary.
```
rmloveland: More of a style thing, but I'd probably edit this slightly to say "Use the following cockroach commands ..." and then make the following 2 items a bulleted list.

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+[`cockroach demo movr`](cockroach-demo.html) opens a temporary, in-memory cluster with the `movr` database preloaded and set as the [current database](sql-name-resolution.html#current-database):
+
+[`cockroach workload init movr`](cockroach-workload.html) loads the `movr` database and some sample data to a running cluster.
+
+## Using the MovR Application
+
+The workflow for MovR is as follows:
```
rmloveland: I was a little confused by this sentence until I read the steps that follow.  It might be a little clearer to say something like "at a high level, the movr app works as follows" (or something much better than that obv :-} )

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+5. The user ends the ride and releases the vehicle, which is an `UPDATE` of a row in the `vehicles` table and an `UPDATE` of a row in the `rides` table:
+
+    ~~~ sql
+    > UPDATE vehicles SET status = 'available' WHERE ...
+    ~~~
+
+    ~~~ sql
+    > UPDATE rides SET end_address = <value> ...
+    ~~~
```
rmloveland: This section is great!  The mapping of use cases to SQL is so easy to follow / understand.

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Extended Examples
+
+MovR scales as a geo-partitioned application and database that you can deploy using Docker images across a multi-region cluster deployment. For a full example, see [Geo-Partitioning for Fast Reads and Writes in a Multi-Region Cluster](demo-geo-partitioning.html).
```
rmloveland: This could possibly be a bit clearer if the latter 2 sentences were each on their own line for easier scanning, as well as revised so their clause structures approximately match a "for more info about X, see Y" formulation, e.g.,
"For an example showing how to tune geo-partitioning for fast reads and writes ..., see ..."
"For a tutorial showing how to tune CockroachDB's performance, see ..."
Writer nerd thing: This type of phrasing is taken from the MSFT style guide 3rd ed. which says
"Information about why the link is interesting should precede the link itself."
Bonus: the reader knows quickly whether they need to read to the end of the sentence or not.
(PS I love that book, it's kinda weird and dated but very cool IMO)

v19.2/movr.md
```diff
@@ -0,0 +1,85 @@
...
+
+You can also learn more about performance tuning in CockroachDB with the MovR example: [Performance Tuning](performance-tuning.html).
```
rmloveland: I think we should add a 'Related Topics' section to the bottom of this page similar to the one found on most other pages.  It could link to various other movr-related things

_includes/sidebar-data-v19.2.json
ericharmeling: Both totally legitimate reasons to just update the file with the same name rather than create an additional file and have two separate files! I'll fix this.

_includes/v19.2/sql/movr-start.md
ericharmeling: Sure thing!

_includes/v19.2/sql/movr-start.md
ericharmeling: Sure thing! Question: Should we be providing copy-paste commands for users trying to start a secure cluster, or just for insecure clusters? We would have to make the assumption that folks have already created certs, etc, or we would have to provide sample commands for all of those steps...

_includes/v19.2/sql/movr-start.md
ericharmeling: Done!

_includes/v19.2/sql/movr-statements-no-data.md
ericharmeling: I 100% agree on this. It is, in fact, a vehicle-sharing app. I confirmed this with Nate last week. I plan to make this global change a part of the PR.

_includes/v19.2/sql/movr-statements.md
ericharmeling: The intro sentence really only applies to SQL statement examples (e.g. insert.md), but the movr-start.md instructions apply to all examples where we use the MovR dataset (e.g. learn-cockroachdb-sql.md). If we could use conditional includes, that would be amazing. Candidly, I don't know how to do that... I'll follow up offline about this.

v19.2/create-table.md
ericharmeling: Makes sense to me!

ericharmeling: Do we typically keep the reference links to other doc pages though?

v19.2/create-table.md
ericharmeling: Great catch!

v19.2/create-table.md
ericharmeling: I agree! There are a lot of gaps in the data generator like this... I am compiling a list of improvements to the dataset/generator. Setting the default values as non-null, generated values is definitely among on that list.

v19.2/drop-table.md
ericharmeling: Done.

v19.2/drop-table.md
ericharmeling: Done.

v19.2/create-table.md
ericharmeling: Done.

v19.2/insert.md
ericharmeling: Sure thing - let's talk about this.

v19.2/insert.md
ericharmeling: Agreed. I'll add this to the notes.

v19.2/learn-cockroachdb-sql-movr.md
ericharmeling: OK. I'll go ahead and remove it from here.

v19.2/insert.md
ericharmeling: Sounds good! +1

v19.2/drop-table.md
ericharmeling: Done!

v19.2/learn-cockroachdb-sql-movr.md
ericharmeling: I think you are right on this... We should be aiming to simplify docs, especially "Learn CockroachDB" docs and others of that ilk. I'll simplify this just as you suggest.

v19.2/learn-cockroachdb-sql-movr.md
ericharmeling: Done.

v19.2/learn-cockroachdb-sql.md
ericharmeling: Done.

v19.2/movr.md
ericharmeling: This was shamelessly copied from https://www.cockroachlabs.com/docs/stable/demo-geo-partitioning.html ! :)

v19.2/movr.md
ericharmeling: Done!

v19.2/movr.md
ericharmeling: Good call. I removed that "most of" part. There is a sentence further down that lists the other tables (currently) in the database.

v19.2/movr.md
ericharmeling: Done!

v19.2/movr.md
ericharmeling: Done.

v19.2/movr.md
ericharmeling: Done. This is also just awesome general writing advice! Thanks!

I'll have to check out this book...

v19.2/movr.md
ericharmeling: Done!

_includes/v19.2/sql/movr-statements.md
rmloveland: SGTM

v19.2/create-table.md
rmloveland: Yes - I am only referring to changing the text.

_includes/v19.2/sql/movr-start.md
rmloveland: As we discussed, my vote is for simple insecure-only, since if you're using secure we can't really know what your connection string is.

jseldess:
___
*[v19.2/movr.md, line 13 at r3](https://reviewable.io/reviews/cockroachdb/docs/5075#-LmF8H0WDGeuR63JA-Yn:-LmF8H0WDGeuR63JA-Yo:bbx1bt5) ([raw file](https://github.com/cockroachdb/docs/blob/b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/movr.md#L13)):*
> ```Markdown
> The MovR example consists of three components.
> 
> - The `movr` database, which contains several tables that hold vehicle, user, and ride data. The `movr` database is built into [`cockroach demo`](cockroach-demo.html) and [`cockroach workload`](cockroach-workload.html).
> ```

On another review, I think this first bullet isn't necessary. Focusing on this as a dataset, which implies a database, and an application, seems clearer.

jseldess:
___
*[_includes/sidebar-data-v19.2.json, line 126 at r2](https://reviewable.io/reviews/cockroachdb/docs/5075#-Lll_8rS7_MZhAzfYqsh:-LmF8UcB9s90yCQCwo32:bqvct4i) ([raw file](https://github.com/cockroachdb/docs/blob/0cf4fe269ac9d714ad93eb0d5e01eeed96809384/_includes/sidebar-data-v19.2.json#L126)):*

Valid point, but despite it also being a static dataset, I think "Sample Applications" focuses on the fact that it is based on a fictional application. I like it how you have it, and I like imagining that we might add more sample applications as well. Maybe show it in the sidenav as `MovR: vehicle-sharing app" or something?
</blockquote></details>

Thinking about this more. We could also name this section just "Workloads" and consider featuring more of all of the workloads and datasets built into `cockroach workload` and `cockroach demo` (separately, of course). I'm thinking YCSB and TPCC would be particularly useful to call out as workloads. Thoughts?

jseldess:
@ericharmeling, teamcity is failing due to these broken links:

```
[21:39:34][Step 1/1] ========================================================================
[21:39:34][Step 1/1] running in concurrent mode, this is experimental
[21:39:35][Step 1/1] b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/indexes.html
[21:39:35][Step 1/1]   hash does not exist --- b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/indexes.html --> create-table.html#create-a-table-primary-key-defined
[21:39:35][Step 1/1] b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/add-constraint.html
[21:39:35][Step 1/1]   hash does not exist --- b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/add-constraint.html --> create-table.html#create-a-table-primary-key-defined
[21:39:36][Step 1/1] ========================================================================
```

Let me know if you need any help resolving them.

ericharmeling:
> @ericharmeling, teamcity is failing due to these broken links:
> 
> ```
> [21:39:34][Step 1/1] ========================================================================
> [21:39:34][Step 1/1] running in concurrent mode, this is experimental
> [21:39:35][Step 1/1] b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/indexes.html
> [21:39:35][Step 1/1]   hash does not exist --- b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/indexes.html --> create-table.html#create-a-table-primary-key-defined
> [21:39:35][Step 1/1] b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/add-constraint.html
> [21:39:35][Step 1/1]   hash does not exist --- b1544f5348dc974a88daee7be262c0dcbb0db432/v19.2/add-constraint.html --> create-table.html#create-a-table-primary-key-defined
> [21:39:36][Step 1/1] ========================================================================
> ```
> 
> Let me know if you need any help resolving them.

Fixed the broken links... These happened because I removed the `CREATE TABLE` example that doesn't define a primary key in response to #4912.

PR #5079, opened 2019-07-23
Quick hyperlink reference for ENUM data type

NiteshKoushik:
Knowing ENUM data type of different database (MySQL and Postgres) helps the users to understand what would be the replaceable data type in cockroach DB

### DIFF
--- a/v19.1/migration-overview.md
+++ b/v19.1/migration-overview.md
@@ -40,7 +40,7 @@ Above a certain size, many data types such as [`STRING`](string.html)s, [`DECIMA
 
 ## Unsupported data types
 
-CockroachDB does not provide `ENUM` or `SET` data types.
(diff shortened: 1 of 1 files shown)


v19.1/migration-overview.md
```diff
@@ -40,7 +40,7 @@ Above a certain size, many data types such as [`STRING`](string.html)s, [`DECIMA
 
 ## Unsupported data types
 
-CockroachDB does not provide `ENUM` or `SET` data types.
+CockroachDB does not provide [`ENUM`](https://en.wikipedia.org/wiki/Enumerated_type) or `SET` data types.
```
jseldess: This link is to a wikipedia entry that is much broader than the SQL ENUM type, so it's not particularly helpful.

PR #5084, opened 2019-07-24
Update GORM insecure example with app retry loop

rmloveland:
... and update the surrounding descriptive text to match the new code.

Includes updates based on review comments in https://github.com/cockroachdb/docs/pull/5049, which for some reason stopped accepting updates.

Addresses #5035.

### DIFF
--- a/_includes/v19.1/app/insecure/gorm-sample.go
+++ b/_includes/v19.1/app/insecure/gorm-sample.go
@@ -1,0 +1,207 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"math"
+	"math/rand"
+	"time"
+
+	// Import GORM-related packages.
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+
+	// Necessary in order to check for transaction retry error codes.
+	"github.com/lib/pq"
+)
+
+// Account is our model, which corresponds to the "accounts" database
+// table.
+type Account struct {
+	ID      int `gorm:"primary_key"`
+	Balance int
+}
+
+// Functions of type `txnFunc` are passed as arguments to our
+// `runTransaction` wrapper that handles transaction retries for us
+// (see implementation below).
+type txnFunc func(*gorm.DB) error
+
+// This function is used for testing the transaction retry loop.  It
+// can be deleted from production code.
+var forceRetryLoop txnFunc = func(db *gorm.DB) error {
+
+	// The first statement in a transaction can be retried transparently
+	// on the server, so we need to add a dummy statement so that our
+	// force_retry statement isn't the first one.
+	if err := db.Exec("SELECT now()").Error; err != nil {
+		return err
+	}
+	// Used to force a transaction retry.  Can only be run as the
+	// 'root' user.
+	if err := db.Exec("SELECT crdb_internal.force_retry('1s'::INTERVAL)").Error; err != nil {
+		return err
+	}
+	return nil
+}
+
+func transferFunds(db *gorm.DB, fromID int, toID int, amount int) error {
+	var fromAccount Account
+	var toAccount Account
+
+	db.First(&fromAccount, fromID)
+	db.First(&toAccount, toID)
+
+	if fromAccount.Balance < amount {
+		return fmt.Errorf("account %d balance %d is lower than transfer amount %d", fromAccount.ID, fromAccount.Balance, amount)
+	}
+
+	fromAccount.Balance -= amount
+	toAccount.Balance += amount
+
+	if err := db.Save(&fromAccount).Error; err != nil {
+		return err
+	}
+	if err := db.Save(&toAccount).Error; err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
+	// Connect to the "bank" database as the "maxroach" user.
+	const addr = "postgresql://maxroach@localhost:26257/bank?sslmode=disable"
+	db, err := gorm.Open("postgres", addr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+
+	// Set to `true` and GORM will print out all DB queries.
+	db.LogMode(false)
+
+	// Automatically create the "accounts" table based on the Account
+	// model.
+	db.AutoMigrate(&Account{})
+
+	// Insert two rows into the "accounts" table.
+	var fromID = 1
+	var toID = 2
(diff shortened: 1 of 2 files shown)


rmloveland:
Ben, I think I've got this code sample working properly using the recommended `txn := db.Begin()` ... etc. syntax.  In latest commit I added some commented-out retry logic that I was using to test.  Didn't squash so the diff would be easy to see.

Based on my tests with a local cluster it is working properly with that update: casting GORM error to `pq.Error`, checking the code, and running through the loop again `maxRetries` times before bailing, while using the `txn.Foo()` statements.

At your convenience do you mind testing it and verifying that it works as it should?  Once it looks good to you I'll clean it up by removing the commented out retry stuff, and add in the secure example boilerplate, etc.

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+	// 3. Print account balances again to verify the transfer occurred.
+
+	// Print balances before transfer.
+	printBalances(db)
+
+	var amount = 100
+	var fromAccount Account
+	var toAccount Account
+
+	db.First(&fromAccount, 1)
```
bdarnell: These calls must be inside the transaction. Any time the transaction retries, we need to re-query the starting balances. (This applies to the other language examples too).

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+	deleteAccounts(db)
+}
+
+func transferFunds(db *gorm.DB, fromAccount Account, toAccount Account, amount int) error {
+	if fromAccount.Balance < amount {
+		return fmt.Errorf("account %d balance %d is lower than transfer amount %d", fromAccount.ID, fromAccount.Balance, amount)
+	}
+
+	var maxRetries = 3
+	for retries := 0; retries <= maxRetries; retries++ {
```
bdarnell: In general, you'd want to refactor the retry loop into a higher-order function instead of inlining it in every function that needs a transaction (where supported by the language). I think it's probably a good idea to do that in our examples.

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+		if retries == maxRetries {
+			return fmt.Errorf("hit max of %d retries, aborting", retries)
+		}
+
+		txn := db.Begin()
+		// The statement below is used for forcing transaction retries
```
bdarnell: Explain a little more about what's going on: The first statement in a transaction can be retried transparently on the server, so we need to add a dummy statement so that our `force_retry` statement isn't the first one.

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+		// for verifying the retry loop logic.  It is not necessary
+		// for production code.
+		// txn.Exec("SELECT now()")
+		if err := txn.Exec(
+			// The below statement is used for forcing transaction
+			// retries for verifying the retry loop logic.  It is not
+			// necessary for production code.
+			// `SELECT crdb_internal.force_retry('1s'::INTERVAL)`,
```
bdarnell: It's stylistically weird to insert this at this point. I'd put this in its own txn.Exec statement after the txn.Exec for `SELECT now()`.  This will also force a refactoring to put the error handling in one place (which again would mean making a runTransaction higher-order function)

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+				var sleepMs = math.Pow(2, float64(retries)) * 100 * (rand.Float64() + 0.5)
+				fmt.Printf("Hit 40001 transaction retry error, sleeping %s milliseconds\n", sleepMs)
+				time.Sleep(time.Millisecond * time.Duration(sleepMs))
+			} else {
+				return err
+			}
+		} else {
+			// Happy case.  All went well, so we commit and break out
+			// of the retry loop.
+			txn.Commit()
```
bdarnell: `txn.Commit` can return an error which must be handled. In particular it can return a retry error.

_includes/v19.1/app/insecure/gorm-sample.go
```diff
@@ -0,0 +1,158 @@
...
+		// The statement below is used for forcing transaction retries
+		// for verifying the retry loop logic.  It is not necessary
+		// for production code.
+		// txn.Exec("SELECT now()")
+		if err := txn.Exec(
+			// The below statement is used for forcing transaction
+			// retries for verifying the retry loop logic.  It is not
+			// necessary for production code.
+			// `SELECT crdb_internal.force_retry('1s'::INTERVAL)`,
+			`UPSERT INTO accounts (id, balance) VALUES
```
bdarnell: Users of an ORM don't generally want to write SQL queries by hand like this. Idiomatic use of GORM would be to either modify the Account objects and call Save on them or use gorm's Update method.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: Fixed these by moving them inside `transferFunds`, which is now a function argument to a `runTransaction` function very similar to the Python examples.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: Done.  It's pretty close to the Python version but may need some updates to be idiomatic Go.  PTAL.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: Updated with this explanation.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: This is all refactored into the `runTransaction` HOF now.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: Updated to check for 40001 errors on `txn.Commit()` and continue the retry loop if they're encountered.

_includes/v19.1/app/insecure/gorm-sample.go
rmloveland: Updated to modify the Account objects and call Save on them.

PR #5088, opened 2019-07-25
Update Java JDBC example to use app retry loop

rmloveland:
As well as a small tweak to the docs to link to the newer client-side
transaction retry recommendations.

Addresses #5035.


rmloveland:
Thanks for the review, Ben!

PR #5094, opened 2019-07-29
Reference link to know what Index is.

NiteshKoushik:
Knowing about Index will give a clear idea on - what index is and post that knowing primary and secondary index will help the users.

### DIFF
--- a/v19.1/performance-best-practices-overview.md
+++ b/v19.1/performance-best-practices-overview.md
@@ -158,7 +158,7 @@ However, the performance best practice is to use a `RETURNING` clause with `INSE
 
 ### Use secondary indexes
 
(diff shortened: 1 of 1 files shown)


v19.1/performance-best-practices-overview.md
```diff
@@ -158,7 +158,7 @@ However, the performance best practice is to use a `RETURNING` clause with `INSE
 
 ### Use secondary indexes
 
-You can use secondary indexes to improve the performance of queries using columns not in a table's primary key. You can create them:
+You can use secondary [indexes](https://www.cockroachlabs.com/docs/stable/indexes.html) to improve the performance of queries using columns not in a table's primary key. You can create them:
```
jseldess: This is a good change, but we use relative links. I'll handle this in a separate PR.

PR #5095, opened 2019-07-29
Removal of same statement as it is used multiple times

NiteshKoushik:
Removal Nodes - During decommission need know what Remove nodes means, but stating the same across the documents does not seems to be good for the end user. Instead of that, we can add a reference link near decommission which tells the same thing about Remove node.


jseldess:
Handled in separate PR.

PR #5096, opened 2019-07-29
Removal of same statement as it is used multiple times

NiteshKoushik:
Removal Nodes - During decommission need know what Remove nodes means, but stating the same across the documents does not seems to be good for the end user. Instead of that, we can add a reference link near decommission which tells the same thing about Remove node.


v19.1/view-node-details.md
```diff
@@ -258,11 +258,11 @@ You need to run the `node status` command on a live node to identify the other l
 
 ### Decommission nodes
 
-See [Remove Nodes](remove-nodes.html)
+See [remove nodes](remove-nodes.html)
```
jseldess: Title case is used because we listing a page title.

v19.1/view-node-details.md
```diff
@@ -258,11 +258,11 @@ You need to run the `node status` command on a live node to identify the other l
...
 
 ### Recommission nodes
 
-See [Recommission Nodes](remove-nodes.html#recommission-nodes)
+See [recommission nodes](remove-nodes.html#recommission-nodes)
```
jseldess: Title case is used because we listing a page title.

jseldess:
Most of these changes are helpful, but there a few that we don't want to make. I'll handle the helpful changes in a separate PR.

PR #5105, opened 2019-07-30
Topnav and footer updates

### LINKED ISSUE #5057: Nav bar dropdown for "Product" needs updating
The dropdown menu when you hover over "Product" in the navigation bar needs updating to make it consistent with the rest of the site

Recently, the dropdown menu for the rest of the site was updated so it now looks like this:
Screen Shot 2019-07-16 at 10 07 00 AM

The dropdown on the docs site is still the old version:
Screen Shot 2019-07-16 at 10 06 51 AM

Please note also that in the new version of the dropdown, the green header reading "CockroachDB" is going to be updated so it is regular black text, and is clickable, directing to this link: https://www.cockroachlabs.com/product
So in the updated dropdown on the docs site, please add these changes as well

Thank you!

jseldess:
Updates to match marketing site. 

Fixes #5057.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -67,6 +67,13 @@
               >Architecture Overview</a
             >
           </li>
+          <li>
+            <a
+              href="https://www.cockroachlabs.com/docs/stable/frequently-asked-questions.html"
+              >FAQ</a
+            >
+          </li>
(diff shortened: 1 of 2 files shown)


meagan-g:
@jseldess thanks, I reviewed it and this is good to go

PR #5107, opened 2019-07-30
Updating reference link to know the supporting database

NiteshKoushik:
Updating reference link to know the supporting database - MySQL and Postgres


jseldess:
These databases are very commonly known, so these links are unnecessary and would detract from the more meaningful links further down the page.

PR #5108, opened 2019-07-30
The word secondary indexes suites better than the word secondary indices

NiteshKoushik:
The word secondary indexes suites better than the word secondary indices. And users can connect with Index word than Indices


v19.1/interleave-in-parent.md
```diff
@@ -66,7 +66,7 @@ In general, reads, writes, and joins of values related through the interleave pr
 
 Fast deletes are available for interleaved tables that use [`ON DELETE CASCADE`](add-constraint.html#add-the-foreign-key-constraint-with-cascade).  Deleting rows from such tables will use an optimized code path and run much faster, as long as the following conditions are met:
 
-- The table or any of its interleaved tables do not have any secondary indices.
+- The table or any of its interleaved tables do not have any secondary indexes.
```
jseldess: We actually use both terms through the docs, so this is a larger consistency issue. I'll follow up separately.

PR #5114, opened 2019-07-31
Removal of highlighting SELECT in heading

NiteshKoushik:
Removal of highlighting SELECT key word from heading, as it does not fit better.


jseldess:
This formatting is intended.

PR #5116, opened 2019-07-31
Updating the font of Header

NiteshKoushik:
Updating the font of header instead of highlighting.


jseldess:
This formatting is intended.

PR #5117, opened 2019-07-31
Formatting the header font in 3 places

NiteshKoushik:
Formatting the header font in 3 places


jseldess:
This formatting is intended.

PR #5137, opened 2019-08-05
Referencing leaseholder

NiteshKoushik:
Referencing leaseholder


jseldess:
This is a good idea, but we use relative links in our docs. I'll make this change in a separate PR.

PR #5166, opened 2019-08-06
Remove option to open a PR directly from a docs PR

### LINKED ISSUE #5160: Remove "Edit this page" option from docs
We added the ability to open a PR directly from a docs page to lower the barrier to entry for external contributors. But it is not used much, and when it is used, it causes us a bit too much work for the following reasons:

- Small fixes across files (e.g., spelling, adding links) that could be batched logically into a single PR are always in distinct PRs. This increases the reviewer burden significantly.
- Most of these PRs are changing just the stable version of the docs, and we need to open separate PRs to apply the changes to the dev version as well.

jseldess:
Fixes #5160.

### DIFF
--- a/_includes/contribute-options.html
+++ b/_includes/contribute-options.html
@@ -6,7 +6,7 @@
       <span class="arrow-up"></span>
     </button>
     <ul class="dropdown-menu" aria-labelledby="dropdownMenu1">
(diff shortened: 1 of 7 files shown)


jseldess:
On further thought, I'm going to leave this feature in place. It doesn't get used much, and when it do, the edits are almost always valuable.

PR #5171, opened 2019-08-07
Piping to chmod for workload doesn't work

rkruze:
Changing the '|' to a ';' as the pipe does not change the permission of the workload file.


jseldess:
I'll take over this PR and make the change to 19.2 myself.

jseldess:
Actually found many other instances of this that needed updating.

PR #5208, opened 2019-08-12
MSO permissions docs

### LINKED ISSUE #4947: Permissions model V1
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwchkdQkRUw6hZ3z/recM9bU6TnKjRoSyg?blocks=hide
PM: @kannanlakshmi

Amruta-Ranade:
Closes #4947


v19.2/managed-console-access-management.md
```diff
@@ -0,0 +1,66 @@
...
+
+A Developer is a limited-access role. A Developer cannot invite Team Members to the Console or access databases.
+
+To access a database, you need to ask a Console Admin for details of a SQL user. To find out who's your Console Admin, check the Access page.
+
+### Console Admin
+
+A Console Admin is an all-access role. A Console Admin can perform the following tasks:
+
+- [Invite Team Members to CockroachDB Cloud]()
```
mberhault: There are no links here. Does that get automatically changed to in-page anchors jumps?

v19.2/managed-connect-to-your-cluster.md
```diff
@@ -7,77 +7,106 @@ build_for: [managed]
...
     The **Add Network** modal displays.
 
-    <img src="{{ 'images/v19.2/managed/add-network-modal.png' | relative_url }}" alt="Add network" style="border:1px solid #eee;max-width:100%" />
+    <img src="{{ 'images/v19.1/managed/add-network-modal.png' | relative_url }}" alt="Add network" style="border:1px solid #eee;max-width:100%" />
 
-3. Enter the public IP address of the machine in the **Network** field.
+4. Enter the public IPv4 address of the machine in the **Network** field.
 
-    The IP address should be written in Classless Inter-Domain Routing (CIDR) notation. For example:
+    You can use `0.0.0.0/0`, which allows all networks. Use this with caution; anybody who uses your password will be able to access the database, and your cluster will be more exposed if there's ever a security bug. The firewall is an extra layer of defense.
```
mberhault: I think I prefer the "prevent denial-of-service and brute force password attacks" rationale. But either way, it seems that we should stick to one phrasing for all whitelist warnings.

v19.2/managed-console-access-management.md
```diff
@@ -0,0 +1,66 @@
...
+
+<Screenshot here>
+
+## Roles
+
+Every CockroachDB Cloud user is either a Developer or a Console Admin.
+
+### Developer
+
+A Developer is a limited-access role. A Developer cannot invite Team Members to the Console or access databases.
```
mberhault: We should probably mention that a developer can do. This phrasing makes it sounds like it's nothing.

kannanlakshmi:
Looks good Amruta! My main question is if it makes sense to be using "Cockroach Cloud" terminology yet. We haven't announced that term and the rest of the docs still say "Managed CockroachDB." The plan is to make an external announcement about Cockroach Cloud in October. My preference would be to continue using the term "managed cockroachdb" till then.

Amruta-Ranade:
@kannanlakshmi Oh! I thought we are making the announcement this week! Sorry, my bad. Will use "Managed CockroachDB" for now.

kannanlakshmi:
That's my bad for not being clearer. Internally we use cloud but externally we won't use that till October. Sorry about the change and thanks!

Amruta-Ranade:
> In the same spirit, we should introduce a table of permissions for each role. It might be a bit early given the small set of actions a "developer" can do.

That was the initial plan, but considering the Developer doesn't have any permissions, the table was a bit sad (All yes'es in Admin column and crosses in Dev column). Will add a table when we implement Permissions v2.0 :)

mberhault:
> > In the same spirit, we should introduce a table of permissions for each role. It might be a bit early given the small set of actions a "developer" can do.
> 
> That was the initial plan, but considering the Developer doesn't have any permissions, the table was a bit sad (All yes'es in Admin column and crosses in Dev column). Will add a table when we implement Permissions v2.0 :)

Agreed.

PR #5216, opened 2019-08-13
Learn CRDB SQL MovR update

ericharmeling:
#5040
This PR includes the "Learn CockroachDB SQL" MovR update, originally a part of the #5075 PR. I broke it out into a different feature branch because of the conflicts with MSO...


ericharmeling:
@jseldess As you can see in the file diff, I crammed in two separate Learn CRDB SQL instructions... One that uses MovR, for the standard 19.2 docs, and one that does not use MovR (it is essentially identical to the 19.1 Learn CRDB SQL page), for managed.

PR #5222, opened 2019-08-14
updated insecure cluster rust driver sample code to display insecure …

mattcrdb:
Sample code download for rust driver on an insecure cluster was linking to the code for the secure one.

The confusion is coming previously versions of CRDB (v2.0 and below) where the insecure code file path is the same as the secure file path in future versions.


jseldess:
Fixes #5223.

PR #5226, opened 2019-08-15
SHOW EXPERIMENTAL RANGES changed to SHOW RANGES

ericharmeling:
All instances of EXPERIMENTAL RANGES in 19.2 docs have been changed to SHOW RANGES. This includes diagrams, file names, and references.


ericharmeling:
@jseldess This is ready to be reviewed again. See comments for details...

Also, I replied to this in a separate comment reply:

> Thanks for kicking this off. The `SHOW RANGES` page itself needs some work still.
> 
> Also, unfortunately, the `SHOW RANGES` outputs through all of the other pages should, in principle, be updated to show the `locality` field in the response. Ugh. That's a ton of manual work to reproduce the outputs. This is why we need to automate this eventually.
> 
> Let me know how you'd like to proceed here. I could accept just not showing locality in the outputs until we update those pages more thoroughly. Except the `show ranges` page itself; the example outputs need to be accurate there, at least.

There was (the relevant part of) my reply

>I added the locality output for this statement page for this PR....

>I mentioned in a comment on another PR (#5242) that we should make the SHOW RANGES locality output change universally to all SHOW RANGES statements as part of a separate PR. That output is actually a separate feature that was implemented when this statement was called SHOW EXPERIMENTAL_RANGES. So I should have done that before changing all of these statement names, but it will be slightly more time-consuming since I'll have to recreate all examples and get their SHOW RANGES output.

PR #5312, opened 2019-08-26
Remove `cockroach zone` and deprecate `cockroach user`

jseldess:
- Completely remove mention of `cockroach zone`.
- Reference `USER` sql statements instead of `cockroach user` and remove `cockroach user` page.


knz:
Gotcha. Reviewable had merged them that's why I was confused.

PR #5358, opened 2019-09-03
fix syntax highlighting with new version of Jekyll

sploiselle:
Reverts syntax highlighting behavior to state before #5239 

CC @benesch


benesch:
Thanks, Sean!

PR #5388, opened 2019-09-09
Adding a copy functionality

NiteshKoushik:
Adding a copy functionality by adding div class =  copy-clipboard


jseldess:
Thanks for finding this, @NiteshKoushik. The fix isn't quite complete, so I've opened https://github.com/cockroachdb/docs/pull/5428 to finish this off.

PR #5409, opened 2019-09-10
Delete 'Parallel Statement Execution'.

### LINKED ISSUE #4885: sql: remove parallel statement execution
PR: https://github.com/cockroachdb/cockroach/pull/35959

From release notes:
> Parallel statement execution has been removed because the performance improvement it provided was absorbed by transactional pipelining, which doesn't require client buy in. The `RETURNING NOTHING` syntax will still be accepted but becomes a no-op. [#35959][#35959] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #4885.


rmloveland:
@nvanbenschoten this is "just" deleting the parallel stmt execution page (and all references, etc.) from the 19.2 docs.

AFAICT from looking at [your PR](https://github.com/cockroachdb/cockroach/pull/35959), this is a straight-up removal.  Please let me know if you think any additional docs work is needed due to some nuance I don't understand on this.

nvanbenschoten:
The syntax is still supported, it just has no effect. So there are no real backward compat issues.

jseldess:
Ok. Thanks for that clarification, Nathan.

PR #5411, opened 2019-09-11
Licensing update 

### LINKED ISSUE #5402: Licensing FAQ: update language around free tier for educational orgs
Link: https://www.cockroachlabs.com/docs/stable/licensing-faqs.html#i-work-for-a-non-profit-as-a-student-or-in-academia-and-i-would-like-to-use-cockroachdb-do-i-need-a-license-to-access-enterprise-features

In the above question, we point out that CockroachDB can be provided for free to select institutions. We should clarify this language to point out that the free offering is only for our on-prem deployed version of CockroachDB. We can offer discounted pricing on CockroachCloud, but it will not be free. In both cases, contacting us is the correct next step.

cc @kannanlakshmi who is figuring out next steps for CockroachCloud after a request from a prospective customer

Amruta-Ranade:
Closes #5402

### DIFF
--- a/v19.2/licensing-faqs.md
+++ b/v19.2/licensing-faqs.md
@@ -77,4 +77,4 @@ You can fork any historical version of CockroachDB in your own project, as allow
 
 ## I work for a non-profit, as a student, or in academia, and I would like to use CockroachDB. Do I need a license to access enterprise features?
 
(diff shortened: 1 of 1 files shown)


kannanlakshmi:
*[a discussion](https://reviewable.io/reviews/cockroachdb/docs/5411#-LoVeUgvCDDtSnKWEYGo:-LoVeUgw7CjG6JySE2cp:bclqwqs) (no related file):*
LGTM! Thanks for the edits

---

Amruta-Ranade:
@jseldess Ready for another look.

PR #5412, opened 2019-09-11
Added info on span and index to SQL tuning with EXPLAIN

### LINKED ISSUE #5401: SQL Tuning with EXPLAIN Doc Update
Re: [SQL Tuning with EXPLAIN](https://cockroachlabs.com/docs//v19.2/sql-tuning-with-explain.html)

## Issue Description

When talking about the full table scan in the first second, I believe it would be helpful to point out the `spans ALL` as the clear indicator of a full table scan. Also a quick refresher on what the `<table>@<index>` syntax means would also be helpful. 

## Suggested Resolution 

Add information on what `span` and `<table>@<index>` means.

Amruta-Ranade:
Closes #5401

### DIFF
--- a/v19.2/sql-tuning-with-explain.md
+++ b/v19.2/sql-tuning-with-explain.md
@@ -1,14 +1,12 @@
 ---
-title: SQL Tuning with EXPLAIN
-summary: How to use EXPLAIN to identify and resolve SQL performance issues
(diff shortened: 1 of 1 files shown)


Amruta-Ranade:
That's strange..I didn't change the title or the setup section. Need to find out if anyone edited the file recently. Will report back.

Issue #5176, opened 2019-08-07
Update 19.2 code samples to note `crdb_internal.force_retry()` does not require root anymore


rmloveland:
See https://github.com/cockroachdb/cockroach/issues/39243

Estimated scope of work here is:

- Update any 19.2 code samples that make use of this function, since some of the comments note the old behavior (they can't just be copypasta'd directly from 19.1 in at least 1-2 cases)
- Update 19.1 code samples when and if  https://github.com/cockroachdb/cockroach/pull/39246 is backported in an upcoming 19.1.x point release
- Verify that our 19.2 docs (and 19.1.x as applicable) do not make mention of this function needing to be the root user (I don't think this is mentioned outside code samples, but IDK)

ericharmeling:
FYI: The entry for `force_retry` in the functions doc doesn't mention privileges: https://www.cockroachlabs.com/docs/v19.2/functions-and-operators.html#system-info-functions.

All of those function descriptions are really unhelpful though. That's a separate issue.

PR #5415, opened 2019-09-12
Update 19.2 Python code with app retry loop

rmloveland:
Also, delete the comments in the Python code that state that
`crdb_internal.force_retry()` must be run as root.  This restriction was
lifted in https://github.com/cockroachdb/cockroach/pull/39246.

The 19.1 Python code samples were updated in #5085 and #5173.  We then
forgot to port those changes to 19.2.  This commit remedies that
oversight.

Addresses part of #5176.

### DIFF
--- a/_includes/v19.2/app/basic-sample.py
+++ b/_includes/v19.2/app/basic-sample.py
@@ -1,37 +1,152 @@
-# Import the driver.
+#!/usr/bin/env python3
+
 import psycopg2
+import psycopg2.errorcodes
+import time
+import logging
+import random
 
-# Connect to the "bank" database.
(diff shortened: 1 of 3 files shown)


rmloveland:
Eric, this code has already had engineering review in the linked PRs where it was originally added to 19.1.  For some reason I / we forgot to port the changes to 19.2 also!  So here they are.  #OPAD

rmloveland:
Sorry for the late re-push, needed to remove a version-hardcoded link to the example Python code, replace it with the `{{page.version.version}}` magic words.  As long as the build is green (which it should be) it shouldn't change anything.

PR #5426, opened 2019-09-13
Changed kv.bulk_io_write.max_rate to correct value

rkruze:
kv.bulk_io_write.max_rate was shown as 8 EiB but when I show the setting on a newly installed cluster the value is 1 TiB.


jseldess:
As discussed on Slack, handling this in a separate PR with broader updates to our auto-generated docs for 19.2.

PR #5438, opened 2019-09-16
Truncate sql

### LINKED ISSUE #5399: Add more detail to TRUNCATE SQL Statement page
Customer request for doc improvement:

> The topic "Use TRUNCATE instead of DELETE to delete all rows in a table" should at least mention that TRUNCATE and DELETE have some differences that may make you prefer DELETE after all. (Some of these differences are discussed in the doc on partitioning)

ericharmeling:
Fixes #5399.

### DIFF
--- a/v19.2/performance-best-practices-overview.md
+++ b/v19.2/performance-best-practices-overview.md
@@ -26,7 +26,7 @@ For more information, see:
 
 ### Use `TRUNCATE` instead of `DELETE` to delete all rows in a table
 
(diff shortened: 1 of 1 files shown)


ericharmeling:
@jseldess This is in response to the customer feedback sent via HubSpot on 9/6/19.

ericharmeling:
@jseldess This info was already in the docs, on the [`TRUNCATE`](https://www.cockroachlabs.com/docs/stable/truncate.html) page. That note has been there since v1 of the docs: https://github.com/cockroachdb/docs/blame/master/v1.0/truncate.md#L9.

PR #5451, opened 2019-09-18
Add references to IMPORT INTO in relevant docs

### LINKED ISSUE #5180: Update docs to reference importing into an existing table
We now have 19.2 docs on using `IMPORT INTO` to import into an existing table, but we still have statements throughout the docs that it's not possible to import into an existing table. We need to find and update all of these. We also need to update any reference to import best practices, for example, https://www.cockroachlabs.com/docs/dev/performance-best-practices-overview.html#bulk-insert-best-practices and https://www.cockroachlabs.com/docs/dev/sql-faqs.html#how-do-i-bulk-insert-data-into-cockroachdb.

@lhirata, if you already have an issue to polish our import into docs, feel free to close this.

lhirata:
Closes #5180.

### DIFF
--- a/v19.2/import-into.md
+++ b/v19.2/import-into.md
@@ -1,47 +1,41 @@
 ---
 title: IMPORT INTO
-summary: Incrementally import CSV data into an existing CockroachDB table.
+summary: Import CSV data into an existing CockroachDB table.
 toc: true
 ---
 
-<span class="version-tag">New in v19.2:</span>The `IMPORT INTO` [statement](sql-statements.html) incrementally imports CSV data into an existing CockroachDB table. To create a table, use [`CREATE TABLE`](create-table.html).
+<span class="version-tag">New in v19.2:</span> The `IMPORT INTO` [statement](sql-statements.html) imports CSV data into an existing table. To create a table, use [`CREATE TABLE`](create-table.html).
 
 {{site.data.alerts.callout_success}}
 `IMPORT INTO` only works for existing tables. For information on how to import data into new tables, see [`IMPORT`](import.html).
 {{site.data.alerts.end}}
 
-{{site.data.alerts.callout_danger}}
-**This is an experimental feature and should not be used in production due to correctness issues.** See [Known limitations](#known-limitations) below for more information.
-{{site.data.alerts.end}}
-
 ## Required privileges
 
-Only members of the `admin` role can run `IMPORT`. By default, the `root` user belongs to the `admin` role.
+Only members of the `admin` role can run `IMPORT INTO`. By default, the `root` user belongs to the `admin` role.
 
 ## Synopsis
 
-~~~
-> IMPORT INTO table_name (column_name [, ...])
-         CSV DATA ('file_location' [, ...])
-         [ WITH <option> [= <value>] [, ...] ];
-~~~
+<div>
+  {% include {{ page.version.version }}/sql/diagrams/import_into.html %}
+</div>
 
 {{site.data.alerts.callout_info}}
-The table will be taken offline during the incremental import.
+While importing into an existing table, the table is taken offline.
 {{site.data.alerts.end}}
 
 ## Parameters
 
 Parameter | Description
 ----------|------------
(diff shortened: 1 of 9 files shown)


jseldess:
___
*[v19.2/insert.md, line 12 at r1](https://reviewable.io/reviews/cockroachdb/docs/5451#-LpExkcC4FW_EQXfMQ4F:-LpR-Pr0-6jWymgU4I4x:bq075kq) ([raw file](https://github.com/cockroachdb/docs/blob/1b8cc3a3754d8a3156b6349b87704b815e455827/v19.2/insert.md#L12)):*

When I talked to @rolandcrosby, he mentioned that it's kind of a grey area and maybe we could lay out that new in 19.2, you can import from CSV into an existing table and let the user decide what to do. Do you think specifying here (and where you commented elsewhere) "To insert bulk data from a CSV file into an existing table, use `IMPORT INTO`" differentiates itself enough from "inserting multiple rows"?
</blockquote></details>

Can you look into this more? I'd like us to understand when to tell people to use multi-row inserts vs import into. I suspect multi-row inserts will be for smaller batches, where you get better performance from a single batch (`insert into xxx values (a, 1), (b, 2), (c, 3);`) than from separate inserts (`insert into xxx values (a, 1);`, `insert into xxx values (b, 2);`, `insert into xxx values (c, 3);`. You should ask some engineers like David Taylor and probably Ben Darnell.

lhirata:
Note: SQL diagram also needs to be added. The SVG needs to be generated once https://github.com/cockroachdb/cockroach/pull/41214 is merged.

lhirata:
Might want to also add to Known Limitations that you can't use `IMPORT INTO` on interleaved tables.

jseldess:
@dt and @rolandcrosby, PTAL one more time.

jseldess:
@dt, @rolandcrosby, I'd really like to wrap this so. Can one of you please check the latest changes?

v19.2/import-into.md
```diff
@@ -184,23 +182,18 @@ Google Cloud:
...
 - Only CSV data importing into a single table is supported.
+- During incremental import, the table being imported into is taken offline.
+- After incremental import, [constraints](constraints.html) will be un-validated and need to be [re-validated](validate-constraint.html).
 - `IMPORT INTO` **cannot**:
     - Be used within a [transaction](transactions.html).
     - Replace a row.
     - Add a row that conflicts with a `UNIQUE` [index](indexes.html).
-- Currently, `IMPORT INTO` does not work on indexed tables.
-- `IMPORT` can sometimes fail with a "context canceled" error, or can restart itself many times without ever finishing. If this is happening, it is likely due to a high amount of disk contention. This can be mitigated by setting the `kv.bulk_io_write.max_rate` [cluster setting](cluster-settings.html) to a value below your max disk write speed. For example, to set it to 10MB/s, execute:
+    - Be used on [indexed](indexes.html) or [interleaved](interleave-in-parent.html) tables.
```
dt: We removed the no-indexes limitation, though the "replace a row" limitation above includes unique secondary indexes. The limitation is "no imported row may conflict with an existing row (in the primary or any unique secondary index) or the import will be rejected."

v19.2/import-into.md
```diff
@@ -184,23 +182,18 @@ Google Cloud:
 ## Known limitations
```
dt: probably should add "the table's configured GC TTL must be longer than the IMPORT will take to complete"

v19.2/experimental-features.md
```diff
@@ -125,6 +125,10 @@ This example uses the `users` table from our open-source, fictional peer-to-peer
 
 The [`SHOW RANGE ... FOR ROW`](show-range-for-row.html) statement shows information about a [range](architecture/overview.html#glossary) for a particular row of data. This information is useful for verifying how SQL data maps to underlying ranges, and where the replicas for a range are located.
 
+### Incremental import into an existig table
+
+The [`IMPORT INTO`](import-into.html) statement incrementally imports CSV data into an existing table.
```
dt: I don't think I understand what "incrementally" means in this context.

PR #5453, opened 2019-09-18
Vectorized Execution

### LINKED ISSUE #4790: Productionize vectorization prototype
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwchkdQkRUw6hZ3z/rec6XtFIj6grDDHAS?blocks=hide
PM: @awoods187 

Create a new section with the following information:
- [x] Document settings, link to cluster setting docs
- [x] Explain how vectorized execution works conceptually and mechanically similarly to other non-user facing architectural changes (e.g. parallel commits)

### LINKED ISSUE #5141: Update vectorized setting with new name and defaults
tl;dr: `sql.defaults.experimental_vectorize` is now
`sql.defaults.vectorize`, and it has some new defaults.

Via @yuzefovich comment on #4953:

A quick note: we have just merged a PR ([cockroachdb/cockroach#38777][PR]) that renamed `experimental_vectorize` to just `vectorize`. It also added a new possible mode of vectorization, and now all modes are:

- `off` - the vectorized engine is disabled, all queries go through the row execution engine.
- `auto` (which is now the default choice) - all queries consisting only of streaming operators (a streaming operator is such that doesn't require any buffering) are executed through the vectorized engine whereas all others are run through the row execution engine.
- `experimental_on` - all queries that are supported in the vectorized engine (both streaming and non-streaming) are run through it. It is "experimental" because we still don't have disk spilling (which means that large queries can get out of memory error and crash the node).
- `experimental_always` - absolutely all queries are forced to run through the vectorized engine. If the engine doesn't support the query, it errors out. The only exception is SET queries so that vectorize session variable can be changed.> 
[PR]: https://github.com/cockroachdb/cockroach/pull/38777

### LINKED ISSUE #5287: sql: use vectorized engine only when there is a fair amount of data
PR: https://github.com/cockroachdb/cockroach/pull/39329

From release notes:
> The new `vectorize_row_count_threshold` [cluster setting](../v19.2/cluster-settings.html) controls when the vectorized execution engine is used. [#39329][#39329] {% comment %}doc{% endcomment %}

### LINKED ISSUE #5288: sql: add explain(vec)
PR: https://github.com/cockroachdb/cockroach/pull/39280

From release notes:
> Added a `VEC` option to the [`EXPLAIN`](explain.html) command, which shows detailed plan information for vectorized plans. [#39280][#39280] {% comment %}doc{% endcomment %}

### LINKED ISSUE #5513: sql: EXPLAIN shows if a query will execute with vec/distsql
PR: https://github.com/cockroachdb/cockroach/pull/40652

From release notes:
> [`EXPLAIN`](../v19.2/explain.html) now has additional output that shows if a query will run with the vectorized execution engine, and if the query will be distributed across nodes. [#40652][#40652] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #4790.
Fixes #5141.
Fixes #5287.
Fixes #5288.
Fixes #5513.

This PR includes the following initial changes:
- Vectorized Query Execution page added
- Vectorized execution section removed from experimental features
- Vectorized execution section added to SQL Layer page

### DIFF
--- a/_includes/v19.2/misc/session-vars.html
+++ b/_includes/v19.2/misc/session-vars.html
@@ -309,6 +309,20 @@
 
   <tr>
    <td>
+    <code>vectorize</code>
+   </td>
+   <td>The vectorized execution engine mode. Options include <code>auto</code>, <code>experimental_on</code>, and <code>off</code>.
+     For more details, see <a href="vectorized-execution.html#configuring-vectorized-execution">Configuring vectorized execution for CockroachDB</a>.
+   </td>
+   <td>
+    <code>auto</code>
+   </td>
+   <td>Yes</td>
+   <td>Yes</td>
+  </tr>
+
+  <tr>
+   <td>
     <code>client_encoding</code>
    </td>
    <td>(Reserved; exposed only for ORM compatibility.)</td>
--- a/v19.2/architecture/sql-layer.md
+++ b/v19.2/architecture/sql-layer.md
@@ -78,6 +78,16 @@ Components of the physical plan are sent to one or more nodes for execution. On 
 
 Each processor uses an encoded form for the scalar values manipulated by the query. This is a binary form which is different from that used in SQL. So the values listed in the SQL query must be encoded, and the data communicated between logical processors, and read from disk, must be decoded before it is sent back to the SQL client.
 
+#### Vectorized query execution
+
+<span class="version-tag">New in v19.2:</span> If [vectorized execution](../vectorized-execution.html) is enabled, the physical plan is sent to nodes to be processed by the vectorized execution engine.
+
+Upon receiving the physical plan, the vectorized engine reads batches of table data [from disk](storage-layer.html) and converts the data from row format to columnar format. These batches of column data are stored in memory so the engine can access them quickly during execution.
+
+The vectorized engine uses specialized, precompiled functions that quickly iterate over the type-specific arrays of column data. The columnar output from the functions is stored in memory as the engine processes each column of data.
+
+After processing all columns of data in the input buffer, the engine converts the columnar output back to row format, and then returns the processed rows to the SQL interface. After a batch of table data has been fully processed, the engine reads the following batch of table data for processing, until the query has been executed.
+
 ### Encoding
 
 Though SQL queries are written in parsable strings, lower layers of CockroachDB deal primarily in bytes. This means at the SQL layer, in query execution, CockroachDB must convert row data from their SQL representation as strings into bytes, and convert bytes returned from lower layers into SQL data that can be passed back to the client.
--- a/v19.2/experimental-features.md
+++ b/v19.2/experimental-features.md
@@ -21,8 +21,7 @@ The table below lists the experimental session settings that are available.  For
 |-------------------------------------+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `experimental_force_split_at`       | `'off'`       | Indicates whether checks to prevent incorrect usage of [`ALTER TABLE ... SPLIT AT`](split-at.html) should be skipped.                                                                                                                                                                                   |
 | `experimental_enable_zigzag_join`   | `'off'`       | Indicates whether the [cost-based optimizer](cost-based-optimizer.html) will plan certain queries using a zig-zag merge join algorithm, which searches for the desired intersection by jumping back and forth between the indexes based on the fact that they share a sorted order in their key suffix. |
-| `experimental_serial_normalization` | `'rowid'`     | If set to `'virtual_sequence'`, make the [`SERIAL`](serial.html) pseudo-type optionally auto-create a sequence for [better compatibility with Hibernate sequences](https://forum.cockroachlabs.com/t/hibernate-sequence-generator-returns-negative-number-and-ignore-unique-rowid/).                    |
-| `experimental_vectorize`            | `'off'`       | Whether to run SQL queries in [vectorized execution mode](#vectorized-query-execution). `'off'` means never run queries with vectorized; `'on'` means run queries with vectorized when the query is supported; `always` means run everything with vectorized, even if it's not supported.              |
+| `experimental_serial_normalization` | `'rowid'`     | If set to `'virtual_sequence'`, make the [`SERIAL`](serial.html) pseudo-type optionally auto-create a sequence for [better compatibility with Hibernate sequences](https://forum.cockroachlabs.com/t/hibernate-sequence-generator-returns-negative-number-and-ignore-unique-rowid/).
 
 ## SQL statements
 
@@ -137,19 +136,16 @@ The table below lists the experimental SQL functions and operators available in 
 | [`experimental_strptime`](functions-and-operators.html#date-and-time-functions)  | Format time using standard `strptime` notation. |
 | [`experimental_uuid_v4()`](functions-and-operators.html#id-generation-functions) | Return a UUID.                                  |
 
-## Vectorized query execution
+## Vectorized execution on disk-spilling operations
 
-<span class="version-tag">New in v19.2</span>: When the `experimental_vectorize` setting is enabled, vectorized query execution uses a columnar orientation to dramatically increase query execution performance for supported OLTP Heavy/[Online analytical processing (OLAP)](https://en.wikipedia.org/wiki/Online_analytical_processing) queries. By default, CockroachDB uses a SQL execution engine organized in a row orientation, which offers good performance for [Online transaction processing (OLTP)](https://en.wikipedia.org/wiki/Online_transaction_processing) queries but can offer suboptimal performance for OLTP Heavy/OLAP queries.
+[Vectorized query execution](vectorized-execution.html) in CockroachDB is experimental for the following [disk-spilling operations](vectorized-execution.html#disk-spilling-operations):
 
-For an example of how vectorized query execution works on vectorizing a merge join, see the blog post [Vectorizing the merge joiner in CockroachDB](https://www.cockroachlabs.com/blog/vectorizing-the-merge-joiner-in-cockroachdb/).
+{% include {{page.version.version}}/sql/disk-spilling-ops.md %}
 
-{{site.data.alerts.callout_danger}}
-The vectorized execution engine has the following limitations:
+To turn vectorized execution on for all operations, do one of the following:
 
-- It does not yet monitor memory usage or spill to disk for large joins, so there is a risk that queries that perform large joins or other memory-intensive operations could cause the node to run out of memory and crash.
-
-- It is undertested compared to CockroachDB's existing row-oriented execution engine.
-{{site.data.alerts.end}}
+- Set the `sql.defaults.vectorize` [cluster setting](cluster-settings.html) to `experimental_on`.
+- Set the `vectorize` [session variable](set-vars.html) to `experimental_on`.
 
 ## See Also
 
@@ -158,3 +154,4 @@ The vectorized execution engine has the following limitations:
 - [`ALTER TABLE ... EXPERIMENTAL_AUDIT`](experimental-audit.html)
 - [`SHOW TRACE FOR SESSION`](show-trace.html)
 - [`SHOW RANGE ... FOR ROW`](show-range-for-row.html)
+- [Vectorized Query Execution](vectorized-execution.html)
--- a/v19.2/explain.md
+++ b/v19.2/explain.md
@@ -18,6 +18,8 @@ Using `EXPLAIN`'s output, you can optimize your queries by taking the following 
 
 - Avoid scanning an entire table, which is the slowest way to access data. You can avoid this by [creating indexes](indexes.html) that contain at least one of the columns that the query is filtering in its `WHERE` clause.
 
+- <span class="version-tag">New in v19.2:</span> By default, the [vectorized execution](vectorized-execution.html) engine is enabled for all [supported operations](vectorized-execution.html#disk-spilling-operations) and [data types](vectorized-execution.html#supported-data-types). If you are querying a table with a small number of rows, it might be more efficient to use row-oriented execution. The `vectorize_row_count_threshold` [cluster setting](cluster-settings.html) specifies the minimum number of rows required to use the vectorized engine to execute a query plan.
+
 You can find out if your queries are performing entire table scans by using `EXPLAIN` to see which:
 
 - Indexes the query uses; shown as the **Description** value of rows with the **Field** value of `table`
@@ -41,8 +43,9 @@ The user requires the appropriate [privileges](authorization.html#assign-privile
  `VERBOSE`          | Show as much information as possible about the query plan.
  `TYPES`            | Include the intermediate [data types](data-types.html) CockroachDB chooses to evaluate intermediate SQL expressions.
  `OPT`              | Display a query plan tree if the query will be run with the [cost-based optimizer](cost-based-optimizer.html). If it returns an "unsupported statement" error, the query will not be run with the cost-based optimizer and will be run with the heuristic planner.<br><br>To include cost details used by the optimizer in planning the query, use `OPT, VERBOSE`. To include cost and type details, use `OPT, TYPES`. To include all details used by the optimizer, including statistics, use `OPT, ENV`.
+ `VEC`              | Show detailed information about the [vectorized execution](vectorized-execution.html) plan for a query. If the table queried includes [unsupported data types](vectorized-execution.html#supported-data-types), an unhandled data type error is returned.
+ `preparable_stmt`  | The [statement](sql-grammar.html#preparable_stmt) you want details about. All preparable statements are explainable.
  `DISTSQL`          | Generate a URL to a [distributed SQL physical query plan tree](explain-analyze.html#distsql-plan-viewer).<br><br>{% include {{ page.version.version }}/sql/physical-plan-url.md %}
- `preparable_stmt` | The [statement](sql-grammar.html#preparable_stmt) you want details about. All preparable statements are explainable.
 
 {{site.data.alerts.callout_danger}}
 `EXPLAIN` also includes other modes besides query plans that are useful only to CockroachDB developers, which are not documented here.
@@ -54,13 +57,15 @@ Successful `EXPLAIN` statements return tables with the following columns:
 
  Column | Description
 -----------|-------------
-**Tree** | A tree representation showing the hierarchy of the query plan.
-**Field** | The name of a parameter relevant to the query plan node immediately above.
-**Description** | Additional information for the parameter in  **Field**.
+**Tree** | A tree representation of the hierarchy of the query plan.
+**Field** | The name of a property for the query plan.<br><br>The `distributed` and `vectorized` properties apply to the entire query plan. All other properties apply to the query plan node in the **Tree** column.
+**Description** | Additional information about the parameter in  **Field**.
 **Columns** | The columns provided to the processes at lower levels of the hierarchy. Included in `TYPES` and `VERBOSE` output.
 **Ordering** | The order in which results are presented to the processes at each level of the hierarchy, as well as other properties of the result set at each level. Included in `TYPES` and `VERBOSE` output.
 
 ## Examples
+
+The following examples use the [`startrek` example dataset](cockroach-demo.html#datasets). To follow along, you can use `cockroach demo startrek` to start a temporary, in-memory cluster with the `startrek` dataset preloaded.
 
 ### Default query plans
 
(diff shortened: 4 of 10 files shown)


ericharmeling:
@jordanlewis You can probably go ahead and review this. I'll later into adding a diagram after I knock out more release note doc items.

ericharmeling:
@jordanlewis As (briefly) discussed offline, this is good for technical review. After we work through technical accuracy/breadth, I'll have @rmloveland look over things.

_includes/v19.2/misc/session-vars.html
```diff
@@ -307,6 +307,20 @@
    <td>Yes</td>
   </tr>
 
+  <tr>
+   <td>
+    <code>vectorize</code>
+   </td>
+   <td>The vectorized execution engine mode. Options include <code>auto</code>, <code>experimental_on</code>, and <code>off</code>.
+     For more details, see <a href="vectorized-execution.html#configuring-vectorized-execution">Configuring vectorized execution for CockroachDB</a>.
```
rmloveland: Nice!  Glad you wrote up separate configuration instructions, would have been too complex here.

v19.2/architecture/sql-layer.md
```diff
@@ -78,6 +78,10 @@ Components of the physical plan are sent to one or more nodes for execution. On
 
 Each processor uses an encoded form for the scalar values manipulated by the query. This is a binary form which is different from that used in SQL. So the values listed in the SQL query must be encoded, and the data communicated between logical processors, and read from disk, must be decoded before it is sent back to the SQL client.
 
+<span class="version-tag">New in v19.2:</span> CockroachDB supports column-oriented ("vectorized") query execution on all [supported SQL operations](https://www.cockroachlabs.com/docs/stable/sql-feature-support.html). Column-oriented execution models contrast with row-oriented execution models, which can offer good performance for [Online transaction processing (OLTP)](https://en.wikipedia.org/wiki/Online_transaction_processing) queries, but suboptimal performance for [Online analytical processing (OLAP)](https://en.wikipedia.org/wiki/Online_analytical_processing).
```
rmloveland: In general, I think this paragraph is a too "light" on technical detail for the architecture docs.  I would like to see the technical content from `vectorized-execution.html#how-vectorized-execution-works` woven in here.  The architecture docs are explicitly for users who want to know "how stuff works".  Happy to chat more offline about this, I know weaving stuff into the architecture docs can be hard.
I think the text "supports vectorized .. on all supported SQL operations" is slightly misleading from a user POV, since the "Vectorized Query Execution" page then goes on to list all sorts of contexts in which it does not run stuff in vectorized like "streaming operations" etc etc.  It also reads phrasing-wise as a little tautological, "we support all the supported things". Not sure how to suggest a rephrasing in a way that makes sense, since I don't have much technical context here.  Maybe something like "the vectorized engine is enabled for many SQL queries that would benefit from columnarized execution", if that is in fact the case?  Happy to talk offline also if that would be useful.
I would way prefer the link texts here were just "OLTP" and "OLAP".  I think if you are this deep in the SQL layer of the architecture docs, you probably know what those things are.  Also, the linked wiki pages make it clear what they are anyway, whereas "online transaction processing" is still pretty opaque unless you already know what that is.
Further, I think we should link use of "column-oriented execution models" to something like https://en.wikipedia.org/wiki/Column-oriented_DBMS#Column-oriented_systems, and use of "row-oriented" to something like https://en.wikipedia.org/wiki/Column-oriented_DBMS#Row-oriented_systems.
I think the multi-clause sentence "Column-oriented execution models contrast with ..." is kinda hard to follow the logic of.  It should be broken into shorter, more declarative sentences. E.g., "Row-oriented execution models can offer good performance on [OLTP]() queries, but may not perform well on [OLAP]() queries.  Therefore we've added support for columnar ("vectorized" execution, which dramatically improves performance for OLAP ..."

v19.2/architecture/sql-layer.md
```diff
@@ -78,6 +78,10 @@ Components of the physical plan are sent to one or more nodes for execution. On
...
+
+For more information about CockroachDB's vectorized execution engine, see [Vectorized Query Execution](../vectorized-execution.html).
```
rmloveland: Since we just got done describing the vectorized execution, I don't think we need to say the words again here.  Suggest changing to "For more information, see Vectorized ...".
(Aside: I understand why this has a new page all its own right now, but I think we will need to revisit this in the next release (also known as "so far in the future it doesn't matter" :-} ))

v19.2/experimental-features.md
```diff
@@ -141,19 +140,9 @@ The table below lists the experimental SQL functions and operators available in
...
-For an example of how vectorized query execution works on vectorizing a merge join, see the blog post [Vectorizing the merge joiner in CockroachDB](https://www.cockroachlabs.com/blog/vectorizing-the-merge-joiner-in-cockroachdb/).
-
-{{site.data.alerts.callout_danger}}
-The vectorized execution engine has the following limitations:
-
-- It does not yet monitor memory usage or spill to disk for large joins, so there is a risk that queries that perform large joins or other memory-intensive operations could cause the node to run out of memory and crash.
-
-- It is undertested compared to CockroachDB's existing row-oriented execution engine.
-{{site.data.alerts.end}}
+[Vectorized query execution](vectorized-execution.html) in CockroachDB is experimental for non-streaming operations, including global sorts, hash joins, merge joins on non-unique columns, and window functions. By default, CockroachDB uses vectorized execution on streaming operations. You can turn vectorized execution on for all operations by setting the `sql.defaults.vectorize` [cluster setting](cluster-settings.html) or the `vectorize` [session variable](set-vars.html) to `experimental_on`.
```
rmloveland: I found this paragraph hard to read as someone who doesn't already know stuff about this feature.  Maybe it would be clearer as something like "By default, CockroachDB does not support vectorized execution for the following non-streaming operations: (bulleted list with line breaks).  To turn on experimental support in the vectorized execution engine for non-streaming operations, set (bulleted list of settings flags with line breaks)."  I would also like the streaming operations to link to the things they are talking about, e.g., merge joins link to merge join section of JOIN docs, etc.  In general the whole "streaming" vs "non-streaming" operations makes a lot of sense if you think about DistSQL all day, but it probably won't to many normal users (including me unless I need to care), so we should be way more explicit about what things go in what categories.

v19.2/explain.md
```diff
@@ -18,6 +18,8 @@ Using `EXPLAIN`'s output, you can optimize your queries by taking the following
 
 - Avoid scanning an entire table, which is the slowest way to access data. You can avoid this by [creating indexes](indexes.html) that contain at least one of the columns that the query is filtering in its `WHERE` clause.
 
+- <span class="version-tag">New in v19.2:</span> By default, `vectorize=auto`, meaning the column-oriented vectorized SQL execution engine is enabled for all streaming operations. If you are querying a table with a small number of rows, it might be more efficient to use row-oriented execution. The `vectorize_row_count_threshold` [cluster setting](cluster-settings.html) specifies the minimum number of rows required to use the vectorized engine to execute a query plan.
```
rmloveland: Same comment as above re: I don't think "streaming operations" is something we can expect most users to understand without explanation.  It would be more understandable IMO to avoid saying "supports streaming operations" and instead say something like "the vectorized engine is enabled for many SQL queries that would benefit from columnarized execution.  For details about queries supported by the vectorized engine, see '[Vectorized Execution]()'" (or some such).  And then list out all the streaming vs. non-streaming and which queries are what on the separate '[Vectorized Execution]()' page.

v19.2/explain.md
```diff
@@ -43,6 +45,7 @@ The user requires the appropriate [privileges](authorization.html#assign-privile
  `OPT`              | Display a query plan tree if the query will be run with the [cost-based optimizer](cost-based-optimizer.html). If it returns an "unsupported statement" error, the query will not be run with the cost-based optimizer and will be run with the heuristic planner.<br><br>To include cost details used by the optimizer in planning the query, use `OPT, VERBOSE`. To include cost and type details, use `OPT, TYPES`. To include all details used by the optimizer, including statistics, use `OPT, ENV`.
  `DISTSQL`          | Generate a URL to a [distributed SQL physical query plan tree](explain-analyze.html#distsql-plan-viewer).<br><br>{% include {{ page.version.version }}/sql/physical-plan-url.md %}
  `preparable_stmt` | The [statement](sql-grammar.html#preparable_stmt) you want details about. All preparable statements are explainable.
+ `VEC`             | Show detailed information for [vectorized](vectorized-execution.html) execution plans.
```
rmloveland: For some reason this breaks the HTML rendering of the table.

v19.2/explain.md
```diff
@@ -62,6 +65,8 @@ Successful `EXPLAIN` statements return tables with the following columns:
 
 ## Examples
 
+The following examples use the [`startrek` example dataset](cockroach-demo.html#datasets). To follow along, you can use `cockroach demo startrek` to start a temporary, in-memory cluster with the `startrek` dataset preloaded.
+
```
rmloveland: Nice touch here

v19.2/explain.md
```diff
@@ -72,15 +77,17 @@ By default, `EXPLAIN` includes the least detail about the query plan but can be
...
-   │        | order  | +season
-   └── scan |        |
-            | table  | episodes@primary
-            | spans  | ALL
-            | filter | season > 3
-(6 rows)
+    tree    |    field    |   description
++-----------+-------------+------------------+
+            | distributed | true
+            | vectorized  | false
```
rmloveland: Isn't it the case that vectorized execution is also farmed out/distributed across nodes? At least that's what I read on the 'Vectorized Query Execution' page (AIUI). Unfortunate terminology situation arising here.  FYI @jordanlewis @awoods187 for future thinking about these terms (unless I'm totes misunderstanding this?)
Also, AFAICT we have not documented these fields anywhere, so we should do that.  Also the definition of "Field" needs to be updated since it says "name of a parameter relevant to the query plan node immediately above", but there is no plan node above these.

v19.2/explain.md
```diff
@@ -252,88 +285,107 @@ To include all details used by the optimizer, including statistics, use `OPT, EN
...
+  │
+  └ Node 1
+    └ *colexec.sortOp
+      └ *colexec.selGTInt64Int64ConstOp
+        └ *colexec.colBatchScan
+(5 rows)
+~~~
+
+The output shows the different functions that will be used to process each batch of column-oriented data.
+
```
rmloveland: AFAICT these are the names of actual Go functions right?  If so I think we should call these "internal functions"

v19.2/explain.md
```diff
@@ -396,12 +450,14 @@ If there were an index on `v`, CockroachDB would be able to avoid scanning the e
...
-------+-------+-------------
- scan |       |
-      | table | kv@v
-      | spans | /4-/6
-(3 rows)
+  tree |    field    | description
++------+-------------+-------------+
+       | distributed | false
+       | vectorized  | false
+  scan |             |
```
rmloveland: What does it mean when both of the above are false?  I think we should EXPLAIN this (lol get it)

v19.2/set-vars.md
```diff
@@ -64,11 +64,9 @@ default database for the current session:
 ~~~
 
 ~~~
+  database
 +----------+
-| database |
-+----------+
-| bank     |
-+----------+
```
rmloveland: I'm glad you're updating these examples opportunistically, but it reminds me that gosh I really miss the old wrapped tables output.

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Vectorized Query Execution
+summary: The CockroachDB vectorized SQL query execution engine processes query plans using a column-oriented model to improve performance.
+toc: true
+---
+
+ <span class="version-tag">New in v19.2:</span> CockroachDB supports column-oriented ("vectorized") query execution on all [supported SQL operations](https://www.cockroachlabs.com/docs/stable/sql-feature-support.html).
```
rmloveland: Same comments as above re: "I think the text "supports vectorized .. on all supported SQL operations" is slightly misleading from a user POV, since the "Vectorized Query Execution" page then goes on to list all sorts of contexts in which it does not run stuff in vectorized like "streaming operations" etc etc.  It also reads phrasing-wise as a little tautological, "we support all the supported things". Not sure how to suggest a rephrasing in a way that makes sense, since I don't have much technical context here.  Happy to talk offline also if that would be useful."  Maybe something like "the vectorized engine is enabled for many SQL queries that would benefit from columnarized execution", if that is in fact the case?

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+ Many SQL databases, including MySQL, PostgreSQL, and all versions of CockroachDB released prior to 19.2, execute [query plans](https://en.wikipedia.org/wiki/Query_plan) one row of table data at a time. The CockroachDB vectorized execution engine dramatically improves performance over row-oriented execution by processing each component of a query plan on type-specific batches of column data. Column-oriented execution models contrast with row-oriented execution models, which can offer good performance for [online transaction processing (OLTP)](https://en.wikipedia.org/wiki/Online_transaction_processing) queries, but suboptimal performance for [online analytical processing (OLAP)](https://en.wikipedia.org/wiki/Online_analytical_processing) queries.
```
rmloveland: The opening sentence would not lose any important information, and would be easier to read as, "Many SQL databases execute query plans one row at a time"
I think we should link use of "column-oriented execution models" to something like https://en.wikipedia.org/wiki/Column-oriented_DBMS#Column-oriented_systems, and use of "row-oriented" to something like https://en.wikipedia.org/wiki/Column-oriented_DBMS#Row-oriented_systems.
I think the multi-clause sentence "Column-oriented execution models contrast with ..." is kinda hard to follow the logic of.  It should be broken into shorter, more declarative sentences. E.g., "Row-oriented execution models can offer good performance on [OLTP]() queries, but may not perform well on [OLAP]() queries.  Therefore we've added support for columnar ("vectorized" execution, which dramatically improves performance for OLAP ..."

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Configuring vectorized execution
+
+By default, vectorized execution is enabled in CockroachDB for all supported "streaming" SQL operations (i.e. operations that do not require [memory buffering](https://en.wikipedia.org/wiki/Data_buffer)). Non-streaming operations include global [sorts](query-order.html), [window functions](window-functions.html), [hash joins](joins.html#hash-joins), and [merge joins](joins.html#merge-joins) on non-unique columns. Merge joins on columns that are guaranteed to have one row per value, also known as "key columns", are considered streaming operations.
```
rmloveland: I'm glad we are linking to the actual things here (joins, etc.)  I think we should go further and be very explicit about 1. defining "streaming" 2. defining "non-streaming" 3. listing all "streaming" SQL operations with links to their reference pages 4. listing "non-streaming" SQL operations with links to their reference pages.  Some of that content is here but I think it should be moved out of here into a separate section called something like "Supported Query Types" and be much more explicit in the form of lists of things.

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+You can turn vectorized execution on or off for all operations with the `sql.defaults.vectorize` [cluster setting](cluster-settings.html) or the `vectorize` [session variable](set-vars.html). The following options are supported for both the cluster setting and session variable:
```
rmloveland: I think "you can turn v.e. (so tired) on or off for all operations" is not exactly correct, since there is also the default auto mode which means "on, sometimes".  Perhaps this can say something like "To configure v.e., use the following settings: ..."  And the settings should be in a bulleted list for easier reading/scanning IMO.

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+Option | Description
+----------|------------
+`auto` | Instructs CockroachDB to use the column-oriented vectorized SQL execution engine on streaming operations. By default, `vectorize=auto`.
```
rmloveland: Suggest using bold "**Default**" to show this is the default, to follow the pattern elsewhere in our docs (e.g., [BEGIN params](https://www.cockroachlabs.com/docs/stable/begin-transaction.html#parameters))

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+`experimental_on` | Turns vectorization on for all supported operations. We do not recommend using this setting in production environments, as it can lead to memory issues. See [Known Limitations](vectorized-execution.html#known-limitations) for more information.
```
rmloveland: This "for all supported operations" language (here and elsewhere) is just highly confusing to me.  If something will lead to memory issues or a node crashing, then it's not really supported is it?  My mental model is this: If I were reading docs about something non-CRDB, say the Linux kernel, and it said "this module works with all supported kernel configurations" but then also said "ps if you enable this module, kernel could OOM your processes or crash your machine altogether" I would be *extremely* wary of whether it was in fact "supported".  It would really mess with my trust in the rest of the docs to even have that language in there.  Do you know what I mean?  Am I just crazy?  I think this is a larger conversation cc @jordanlewis @awoods187

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+`off` | Turns vectorization off for all operations.
```
rmloveland: Suggest rewording all of the above to use "Turns on X." or "Turns off X." as the first sentence for easier reading.

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+You can verify whether will use the execution engine or not with the [`EXPLAIN`](explain.html) statement.
+
+For more information about configuring cluster settings, see [Cluster Settings](cluster-settings.html). For more information about setting session variables, see [`SET` &lt;session variable&gt;](set-vars.html).
+
+### Setting the row threshold for vectorized execution
+
+The efficiency of column-oriented execution increases with the number of rows processed. If you are querying a table with a small number of rows, it is currently more efficient to use row-oriented execution.
```
rmloveland: I think we should standardize on one usage everywhere in our documentation for this.  Otherwise it's going to be hard to find stuff about vectorized.  We could either say "vectorized" or say "column-oriented" or say "column-oriented ('vectorized')" but I don't think we should mix and match all 3.  The first mention of vectorized on each page where it appears could be "vectorized (column-oriented) execution" and thereafter just use "vectorized".

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+With `vectorize=auto`, the `vectorize_row_count_threshold` [cluster setting](cluster-settings.html) specifies the minimum number of rows required to use the vectorized engine to execute a query plan. If the number of rows being processed falls below the row count threshold, CockroachDB uses the row-oriented execution engine instead. By default, `vectorize_row_count_threshold=1000`.
```
rmloveland: This would make more sense to me with a structure like the following:

rmloveland: Also, is the `vectorize_row_count_threshold` cluster setting still respected when `vectorize=experimental_on`?  or does it ignore `v_r_c_t` altogether when `vectorize=experimental_on` since presumably the latter means "fuck it we're going five blades", right?  It's implied here (I think?) but we should make it explicit.

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+## How vectorized execution works
+
+When you issue a SQL query in CockroachDB, the gateway node (i.e. the node from which you issue the query) [parses the query and creates a physical plan](architecture/sql-layer.html#sql-parser-planner-executor) for execution on each node. If vectorized execution is enabled, the physical plan is sent to each node, to be processed by the vectorized execution engine.
```
rmloveland: Everything from "if vectorized is enabled ..." should be integrated with the architecture docs (as I stated in a previous comment).

v19.2/vectorized-execution.md
```diff
@@ -0,0 +1,58 @@
...
+
+To see a detailed view of the vectorized execution plan for a query, you can use the [`EXPLAIN(VEC)`](explain.html#vec-option) statement. For detailed examples of vectorized query execution for hash and merge joins, see the blog posts [40x faster hash joiner with vectorized execution](https://www.cockroachlabs.com/blog/vectorized-hash-joiner/) and [Vectorizing the merge joiner in CockroachDB](https://www.cockroachlabs.com/blog/vectorizing-the-merge-joiner-in-cockroachdb/).
+
+## Known Limitations
+
+The vectorized execution engine has the following limitations:
+
+- It is not as extensively tested as CockroachDB's existing row-oriented execution engine.
+- There is a risk that queries that perform large joins or other memory-intensive operations could cause the node to run out of memory and crash. As such, vectorized execution is experimental for non-streaming operations, including global sorts, hash joins, merge joins on non-unique columns, and window functions. We do not recommend using vectorized execution for non-streaming operations in production environments, as these operations can buffer an unlimited number of rows before they can start producing output.
```
rmloveland: As mentioned previously, we should capture in one place (perhaps a new "Supported query types" section) what "streaming" and "non-streaming" mean, and all of the SQL they represent, with links to all such statements.  Perhaps then those could be linked from here via the term "streaming".

v19.2/vectorized-execution.md
rmloveland: Sorry, something got dropped here.
Structure like:
- By default, vectorized works on rows with 1000 rows or more
- You can change this setting using field X in cluster settings
- ... something about why you would ever do this is necessary, I think

v19.2/explain.md
```diff
@@ -72,18 +77,22 @@ By default, `EXPLAIN` includes the least detail about the query plan but can be
...
 
-The first column shows the tree structure of the query plan; a set of properties is displayed for each node in the tree. Most importantly, for scans, you can see the index that is scanned (`primary` in this case) and what key ranges of the index you are scanning (in this case, a full table scan). For more information on indexes and key ranges, see the [example](#find-the-indexes-and-key-ranges-a-query-uses) below.
+The `tree` column shows the tree structure of the query plan. In the `field` and `description` columns, a set of properties is shown for each node in the tree. Most importantly, for scans, the `table` property shows you which index is scanned (`primary` in this case) and the `spans` property shows you what key ranges of the index you are scanning (in this case, a full table scan). For more information on indexes and key ranges, see the [example](#find-the-indexes-and-key-ranges-a-query-uses) below.
+
+With the `distributed` property, the output also shows you that the query plan will be distributed to multiple nodes on the cluster. The `vectorized` property shows that the plan will be executed with the row-oriented execution engine, and not the vectorized engine.
```
rmloveland: Nit: "vectorized engine" could be a link

ericharmeling:
@jseldess This is ready for your review.

ericharmeling:
@awoods187 As you can see, this has undergone pretty extensive review, and Jesse still needs to review it. So no pressure or rush to review (but as always, your feedback is welcome)!

jseldess:
___
*[v19.2/vectorized-execution.md, line 68 at r7](https://reviewable.io/reviews/cockroachdb/docs/5453#-Ls3GrEt8dWnumpykv1c:-Ls3GrEt8dWnumpykv1d:b-g4rovv) ([raw file](https://github.com/cockroachdb/docs/blob/f2d7471a5ea0d5f5b12e58fa5ff6cc256b8f1465/v19.2/vectorized-execution.md#L68)):*
> ```Markdown
> ### Queries with constant `NULL` comparisons
> 
> The vectorized execution engine does not support queries that contain comparisons against constant a `NULL`. This includes `NOT NULL` permutations in generic [selection query](selection-queries.html) comparisons and [`CASE`](scalar-expressions.html#simple-case-expressions) expressions, in addition to other projection operators on constant `NULL` values.
> ```

nit: "constant a `NULL`" > "a constant `NULL`", I think

PR #5454, opened 2019-09-18
Virtual Schema docs

### LINKED ISSUE #2957: Document how to use crdb_internal

### LINKED ISSUE #4978: Document how to capture the status of partitioning
As discussed in https://github.com/cockroachdb/docs/issues/2957, we want to document internal tables for partitioning use cases. 

It is not documented but users can run this query for tables: 
```SQL
Select * from crdb_internal.partitions 
join crdb_internal.tables on partitions.table_id=tables.table_id 
where tables.name='mytable'
```
It is also not documented but users can run this query for databases: 
```SQL
Select partitions.name 
from crdb_internal.partitions 
join crdb_internal.tables on partitions.table_id=tables.table_id 
where database_name='mydb
```
We should add this to our various docs including https://www.cockroachlabs.com/docs/v19.1/demo-geo-partitioning.html#step-3-provision-vms

ericharmeling:
Fixes #2957. 
Fixes #4978. 

This PR includes the following changes:
- Added page on `crdb_internal`
- Updated some terminology on `information_schema` page
- Added page on virtual schemas

Note that each entry in the table of `crdb_internal` tables (new in this PR) notes the stability of the table for programmatic purposes. We need a separate doc that details the API stability policy for all interfaces (i.e. explains what "public", "programmable", and "reserved" all mean, so that part of #4326). Then we can start adding notes about stability for each CRDB interface (as is done here).

### DIFF
--- a/_includes/sidebar-data-v19.2.json
+++ b/_includes/sidebar-data-v19.2.json
@@ -1621,9 +1621,27 @@
             ]
           },
           {
-            "title": "Information Schema",
-            "urls": [
-              "/${VERSION}/information-schema.html"
+            "title": "Virtual Schemas",
+            "items":
+            [
+              {
+                "title": "Overview",
+                "urls": [
+                  "/${VERSION}/virtual-schemas.html"
+                ]
+              },
+              {
+                "title": "<code>crdb_internal</code>",
+                "urls": [
+                  "/${VERSION}/crdb-internal.html"
+                ]
+              },
+              {
+                "title": "<code>information_schema</code>",
+                "urls": [
+                  "/${VERSION}/information-schema.html"
+                ]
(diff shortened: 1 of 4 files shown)


ericharmeling:
Closing in favor of https://github.com/cockroachdb/docs/pull/8909.

PR #5464, opened 2019-09-23
Merge join doc clarification

### LINKED ISSUE #4443: Add nuance to join choice algorithm
In our docs today we say `By default, CockroachDB uses merge joins whenever possible because they are more performant than hash joins, computationally and in terms of memory.`  https://www.cockroachlabs.com/docs/dev/joins.html

`Whenever possible` is kind of misleading, I think it was supposed to say whenever the results are ordered for a merge. Technically merge join is always possible if you resort both sides but we don't do that

The sentence that follows implies that we can only do merge joins when both sides are already ordered, in which case "whenever possible" is correct. This was true with the HP, but this needs updating with more nuance.

ericharmeling:
Fixes #4443.

Wording clarified on merge joins to reflect nuance of join choice algorithm.

### DIFF
--- a/v19.2/joins.md
+++ b/v19.2/joins.md
@@ -104,7 +104,7 @@ CockroachDB supports the following algorithms for performing a join:
 
 ### Merge joins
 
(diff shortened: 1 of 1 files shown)


ericharmeling:
Merging this. @awoods187, let me know if this isn't sufficient to resolve #4443.

awoods187:
lgtm

PR #5472, opened 2019-09-24
Clarify hardware recs for throughput

### LINKED ISSUE #4711: Clarify hardware recs for throughput
Re: [Cluster Topology Patterns](https://cockroachlabs.com/docs//v19.1/cluster-topology-patterns.html)

## Issue Description
The following statement can be found on this page: 
Performance: Adding nodes for more processing power and/or storage typically increases throughput. For example, with five nodes and a replication factor of 3, each range has 3 replicas, with each replica on a different node. In this case, there will only be 1-2 replicas on each nod, leaving additional storage and bandwidth available.

## Suggested Resolution 
The last statement is not clear on what it is trying to convey. Maybe something like: 

Performance: CockroachDB is a horizontally scalable database in which adding nodes increases the overall throughput of the system. For example, when an additional node is added to the cluster CockroachDB will automatically balance out the ranges from the pre-existing nodes to the new nodes to take advantage of the additional throughput and capacity of the new node.

jseldess:
Increase optimal vCPUs to 32.

Update GCP and AWS recs as well. Holding off on Azure until
the upcoming cloud report, at which point we should revisit
our hardware recs more broadly.

Fixes #4711.

### DIFF
--- a/v19.1/recommended-production-settings.md
+++ b/v19.1/recommended-production-settings.md
@@ -25,7 +25,7 @@ Also keep in mind some basic topology recommendations:
 
 - When deploying across multiple availability zones:
     - To be able to tolerate the failure of 1 entire AZ in a region, use at least 3 AZs per region and set `--locality` on each node to spread data evenly across regions and AZs. In this case, if 1 AZ goes offline, the 2 remaining AZs retain a majority of replicas.
-    - To be able to tolerate the failure of 1 entire region, use at least 3 regions. 
+    - To be able to tolerate the failure of 1 entire region, use at least 3 regions.
 
 ## Hardware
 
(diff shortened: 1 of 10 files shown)


v19.1/recommended-production-settings.md
```diff
@@ -39,19 +39,19 @@ Nodes should have sufficient CPU, RAM, network, and storage capacity to handle y
...
+- At a bare minimum, each node should have **2 vCPUs and 2 GB of RAM**. More data, complex workloads, higher concurrency, and faster performance require additional resources.
 
     {{site.data.alerts.callout_danger}}
     Avoid "burstable" or "shared-core" virtual machines that limit the load on CPU resources.
     {{site.data.alerts.end}}
 
-- The ideal configuration is 4-16 vCPUs, 8-64 GB memory nodes (2-4 GB of memory per vCPU).
+- To optimize for throughput, use larger nodes, up to 16 vCPUs and 64 GB of RAM. Based on internal testing results, 16 vCPUs is the sweet spot for OLTP workloads.
+
+    To increase throughput further, add more nodes to the cluster instead of increasing node size; higher vCPUs will have NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access)(non-uniform memory access) implications.
```
bdarnell: I'd remove this note about NUMA. We haven't explored this but we have no reason to believe it's a major concern at this point. The immediate reason why going to 32 vCPUs and beyond loses efficiency is simple mutex contention.

bdarnell: Also for the record, on AWS m5/c5 nodes, you can go up to 48 vCPUs in a single NUMA group, so NUMA isn't a concern on this platform until you get to 72 vCPUs.

PR #5478, opened 2019-09-25
Tell Netlify to redirect to our custom 404 page

### LINKED ISSUE #4908: Docs 404 page not working
When you call up a non-existing page for which we don't have a redirect in place, we no longer show our custom 404 page, e.g., this file for 19.1 docs: https://github.com/cockroachdb/docs/blob/master/v19.1/404.md.

Example of when you should get that 404 page (client-driver.html instead of the correct client-drivers-html): https://www.cockroachlabs.com/docs/stable/install-client-driver.html.

jseldess:
Fixes #4908.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -10,3 +10,4 @@ build_for: [standard, managed]
 /docs/v2.2/* /docs/v19.1/:splat 301
 /docs/releases/ /docs/releases/index.html 200!
 /docs /docs/stable/ 301
+/docs/* /docs/stable/404.html 404


_redirects
```diff
@@ -10,3 +10,4 @@ build_for: [standard, managed]
 /docs/v2.2/* /docs/v19.1/:splat 301
 /docs/releases/ /docs/releases/index.html 200!
 /docs /docs/stable/ 301
+/docs/* /docs/stable/404.html 404
```
jseldess: @mjibson, is this ok, or should I set up such redirects for the 404 page in each version directory?

mjibson: This is fine with me.

PR #5508, opened 2019-10-01
Fix command to remove data dir on Windows

### LINKED ISSUE #5167: Fix command to remove data dir on Windows
A user pointed out a potential error in [docker setup instructions for windows](https://www.cockroachlabs.com/docs/stable/start-a-local-cluster-in-docker.html#os-windows).

The user states that `rm -rf cockroach-data` is not a windows command. They mention that in other sections  we mention Powershell and offer up this input that the correct command may be:

`Remove-Item C:\Users\<username>\cockroach-data -recurse`

jseldess:
Fixes #5167.


BramGruneir:
Besides the extra tag, lgtm

PR #5529, opened 2019-10-02
Changefeed cursors must be in quotes

### LINKED ISSUE #5502: Specify that changefeed cursors need to be in quotes
Our docs don't currently indicate that, when using the `WITH cursor` option, you need to put single quotes around the cursor. The example on the `CREATE CHANGEFEED` page is incorrect as well:

> Example: CURSOR=1536242855577149065.0000000000

An example of correct usage would look like:

`CREATE CHANGEFEED FOR TABLE foo INTO 'experimental-https://example.com' WITH cursor = '1568920310686945044.0000000000';`

jseldess:
Fixes #5502.


jseldess:
@rolandcrosby, such a minor change, I'm just going to merge.

PR #5535, opened 2019-10-02
Timestamp precision doc

### LINKED ISSUE #5005: sql: Add precision to timestamp objects, with 6 as the default precision value
PR: https://github.com/cockroachdb/cockroach/pull/37920

From release notes:
> Implemented SQL support for timestamp objects to have an optional precision value. [#37920][#37920] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5005.

- Added doc about specifying precision for timestamp objects
- Updated timestamp example output


jseldess:
@ericharmeling, this product change has been reverted in 19.2. Can you please revert the docs change as well?

ericharmeling:
I’m on it - see https://github.com/cockroachdb/docs/pull/5827

PR #5543, opened 2019-10-02
cockroach sql --watch flag

### LINKED ISSUE #5520: cli/sql: new parameter --watch to repeat the -e statements
PR: https://github.com/cockroachdb/cockroach/pull/40594

From release notes:
> [`cockroach sql`](../v19.2/cockroach-commands.html) now supports the `--watch` flag. When used to specified a duration, the client will repeat the statement(s) specified by `-e` until an error occurs. This is intended for simple monitoring scenarios during development and testing. [#40594][#40594] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5520. 

Added `--watch` flag description and example.

### DIFF
--- a/v19.2/use-the-built-in-sql-client.md
+++ b/v19.2/use-the-built-in-sql-client.md
@@ -74,6 +74,7 @@ Flag | Description
 <a name="sql-flag-format"></a> `--format` | How to display table rows printed to the standard output. Possible values: `tsv`, `csv`, `table`, `raw`, `records`, `sql`, `html`.<br><br>**Default:** `table` for sessions that [output on a terminal](#session-and-output-types); `tsv` otherwise<br /><br />This flag corresponds to the `display_format` [client-side option](#client-side-options).
 `--safe-updates` | Disallow potentially unsafe SQL statements, including `DELETE` without a `WHERE` clause, `UPDATE` without a `WHERE` clause, and `ALTER TABLE ... DROP COLUMN`.<br><br>**Default:** `true` for [interactive sessions](#session-and-output-types); `false` otherwise<br /><br />Potentially unsafe SQL statements can also be allowed/disallowed for an entire session via the `sql_safe_updates` [session variable](set-vars.html).
 `--set` | Set a [client-side option](#client-side-options) before starting the SQL shell or executing SQL statements from the command line via `--execute`. This flag may be specified multiple times, once per option.<br><br>After starting the SQL shell, the `\set` and `unset` commands can be use to enable and disable client-side options as well.  
+`--watch` | Repeat the SQL statements specified with `--execute` or `-e` until a SQL error occurs or the process is terminated. `--watch` applies to all `--execute` or `-e` flags in use.<br />You must also specify an interval at which to repeat the statement, followed by a time unit. For example, to specify an interval of 5 seconds, use `5s`.<br /><br /> Note that this flag is intended for simple monitoring scenarios during development and testing. See the [example](#repeat-a-sql-statement) below.
 
 ### Client connection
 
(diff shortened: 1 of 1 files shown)


v19.2/use-the-built-in-sql-client.md
```diff
@@ -740,6 +741,61 @@ Time: 2.426534ms
...
+
+Use the `--execute` flag to create a table from the command line:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach sql --insecure \
+--execute="CREATE TABLE t1 (id UUID DEFAULT gen_random_uuid(), name STRING)"
+~~~
+
+With the `--watch` flag, you can repeat the statements specified with the `--execute` flag periodically until an error occurs:
```
rmloveland: I would be more explicit here about what we mean by "an error occurs".  For example, do we mean: SQL error?  UNIX process kill?  Both/either?  If both/either, maybe some text like "until a SQL error occurs, or the process is killed.  In this example we kill the process by typing ctrl-c."  or something like that.  Also, I would include the text from the code PR that "this is intended for simple monitoring scenarios during development and testing", since the motivation for this added flag was not clear to me until I went to the CRDB commit.

v19.2/use-the-built-in-sql-client.md
```diff
@@ -740,6 +741,61 @@ Time: 2.426534ms
...
+--watch 3s
+~~~
+
+~~~
+INSERT 3
+INSERT 3
+INSERT 3
+INSERT 3
+INSERT 3
+^C
```
rmloveland: This box is typically used to show output in our docs, but in this case the ctrl-C is shell input.  As mentioned above, I would explicitly describe the ctrl-c scenario in prose.

v19.2/use-the-built-in-sql-client.md
```diff
@@ -74,6 +74,7 @@ Flag | Description
...
 `--set` | Set a [client-side option](#client-side-options) before starting the SQL shell or executing SQL statements from the command line via `--execute`. This flag may be specified multiple times, once per option.<br><br>After starting the SQL shell, the `\set` and `unset` commands can be use to enable and disable client-side options as well.  
+`--watch` | Repeat the SQL commands specified with `--execute` or `-e` until a SQL error occurs or the process is terminated. You must also specify an interval at which to repeat the statement. Valid [time units](https://en.wikipedia.org/wiki/Orders_of_magnitude_(time)) are `ns`, `us`, `ms`, `s`, `m`, and `h`. See the [example](#repeat-a-sql-statement) below.<br /><br /> Note that this flag is intended for simple monitoring scenarios during development and testing.
```
knz: 1) it's possible to have multiple uses of `-e`; the `--watch` applies to all of them not just one

2) I would not put a hyperlink for the units and instead phrase it as: "... specify a time interval between executions of the statements; a time interval is a number followed by a time unit, for example 's' for seconds". Let the user figure out the other units on their own.

(We do not have a fixed set of units today, we're dependent on another package for this, and we probably don't want to commit on a particular set of units. I'd say just `s` is guaranteed.)

v19.2/use-the-built-in-sql-client.md
```diff
@@ -740,6 +741,63 @@ Time: 2.426534ms
...
+With the `--watch` flag, you can repeat the statements specified with the `--execute` flag periodically, until a SQL error occurs or the process is terminated:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach sql --insecure \
+--execute="INSERT INTO t1 (name) VALUES ('a'), ('b'), ('c')" \
+--watch 3s
+~~~
+
+In this example we kill the process with Ctrl+C.
```
knz: not a great example. Applications that perform INSERTs want to insert different data every time, and that's not going to work.

Also the feature is really intended for _monitoring_ so it would be interesting to have a monitoring-oriented example.

For example perhaps something like `--watch 1h -e 'select count(*)'` to monitor table size, or `--watch 1s -e 'select sum(count) from crdb_internal.node_statement_statistics'` for the number of queries run on a given node.

v19.2/use-the-built-in-sql-client.md
```diff
@@ -74,6 +74,7 @@ Flag | Description
...
 `--set` | Set a [client-side option](#client-side-options) before starting the SQL shell or executing SQL statements from the command line via `--execute`. This flag may be specified multiple times, once per option.<br><br>After starting the SQL shell, the `\set` and `unset` commands can be use to enable and disable client-side options as well.  
+`--watch` | Repeat the SQL commands specified with `--execute` or `-e` until a SQL error occurs or the process is terminated. `--watch` applies to all `--execute` or `-e` flags in use.<br />You must also specify an interval at which to repeat the statement, followed by a time unit. For example, to specify an interval of 5 seconds, use `5s`.<br /><br /> Note that this flag is intended for simple monitoring scenarios during development and testing. See the [example](#repeat-a-sql-statement) below.
```
knz: s/commands/statements/

PR #5544, opened 2019-10-02
Document JDBC driver cursor support

### LINKED ISSUE #5448: Document improved streaming behavior of JDBC driver
In CockroachDB 19.2, we now support streaming behavior in JDBC.  This is a way to do pagination from the driver.

Specifically, 2 new things work that did not work before:

1. set autocommit = false
2. setting a rowsize fetch limit

Did some searching on the PGJDBC docs website, and it looks like this feature is explained at [Getting results based on a cursor][cursor].

Estimated scope of work:

- Update [Build a Java App with CockroachDB][jdbc] with a section explaining that we support this behavior, with a code sample

See also:

- #3743
- [Improve streaming behavior on JDBC driver (Airtable)][at]

[cursor]: https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor
[at]: https://airtable.com/tblD3oZPLJgGhCmch/viw4fl7JawCP1kOdq/rec7rOTING8w7YyoL?blocks=hide
[jdbc]: https://www.cockroachlabs.com/docs/stable/build-a-java-app-with-cockroachdb.html

rmloveland:
Fixes #5448.

### DIFF
--- a/v19.2/build-a-java-app-with-cockroachdb.md
+++ b/v19.2/build-a-java-app-with-cockroachdb.md
@@ -258,6 +258,14 @@ try (PreparedStatement pstmt = connection.prepareStatement("INSERT INTO accounts
 }
 ~~~
 
+### Retrieve large data sets in chunks using cursors
+
+<span class="version-tag">New in v19.2:</span> CockroachDB now supports the Postgres wire-protocol cursors for implicit transactions and explicit transactions executed to completion. This means the [PGJDBC driver](https://jdbc.postgresql.org) can use this protocol to stream queries with large result sets. This is much faster than [paginating through results in SQL using `LIMIT .. OFFSET`](selection-queries.html#paginate-through-limited-results).
+
+For instructions showing how to use cursors in your Java code, see [Getting results based on a cursor](https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor) from the PGJDBC documentation.
+
+Note that interleaved execution (partial execution of multiple statements within the same connection and transaction) is not supported when [`Statement.setFetchSize()`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setFetchSize-int-) is used.
+
 ## What's next?
 
 Read more about using the [Java JDBC driver](https://jdbc.postgresql.org/).


rmloveland:
@mjibson I wrote a little test program to try this out based on the PGJDBC docs example.  Really cool!

I think this doc update is pretty close to right, but please let me know if:
- something is incorrect
- we should include some additional info that is not here

rmloveland:
Thanks Matt!

rmloveland:
@mjibson please take another look, I added some info about interleaved execution not being supported when `setFetchSize()` is set on the statement.

v19.2/build-a-java-app-with-cockroachdb.md
```diff
@@ -258,6 +258,21 @@ try (PreparedStatement pstmt = connection.prepareStatement("INSERT INTO accounts
 }
 ~~~
 
+### Retrieve large data sets in chunks using cursors
+
+<span class="version-tag">New in v19.2:</span> CockroachDB now supports the Postgres wire-protocol cursors for implicit transactions and explicit transactions executed to completion. This means the [PGJDBC driver](https://jdbc.postgresql.org) can use this protocol to stream queries with large result sets. This is much faster than [paginating through results in SQL using `LIMIT .. OFFSET`](selection-queries.html#paginate-through-limited-results).
+
+For instructions showing how to use cursors in your Java code, see [Getting results based on a cursor](https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor) from the PGJDBC documentation.
+
+Note that interleaved execution (partial execution of multiple statements within the same connection and transaction) is not supported when [`Statement.setFetchSize()`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setFetchSize-int-) is used. For example, the following will not work:
```
mjibson: I would not include the code example until we have one that is known to not work. This is because I think it is more misleading than helpful. It may cause people to think they can't execute multiple statements in the same txn.

rmloveland: That makes sense.  I've removed it.

PR #5553, opened 2019-10-03
CRDB interface stability docs

ericharmeling:
This commit includes the following new pages:
- Overview of CockroachDB Interfaces
- Interface Types
- Stability Guarantees
- The Lifecycle of Experimental, Beta, and Stable Features

This commit is a modified version of @knz's API stability doc (#4326)

This commit will also inform #5454, which *could* form a basis for how we document API stability for each interface.

### DIFF
--- a/_includes/sidebar-data-v19.2.json
+++ b/_includes/sidebar-data-v19.2.json
@@ -1950,6 +1950,35 @@
             ]
           }
         ]
+      },
+      {
+          "title": "Programming Interfaces",
+          "items": [
+			  {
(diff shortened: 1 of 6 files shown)


jseldess:
@ericharmeling, @knz, should we leave this open, or has this initiative evolved in other places (internal wiki) and we need to restart on the docs side when it's prioritized?

ericharmeling:
> has this initiative evolved in other places (internal wiki) and we need to restart on the docs side when it's prioritized?

I believe so. Closing this PR for now.

Issue #5446, opened 2019-09-18
MSO rebranding to CC


Amruta-Ranade:
Tasks:

- [x] Search through all docs for mentions of "MSO" and change to "CC"

- [x] Change filenames and URLs. Use redirects if required. Probably need just one server-side redirect.

- [x] Change MSO docs selector in Docs site top menu

- [ ] Update screenshots

Later (not user-facing):
- [x] Rename `make managed` to `make cloud` and adjust TC as necessary

Amruta-Ranade:
Closed via #5559

PR #5559, opened 2019-10-07
Rebranding MSO to CC in docs

Amruta-Ranade:
Partially addresses #5446


Amruta-Ranade:
@jseldess I need help with changing the top nav product selector and redirects.

Amruta-Ranade:
1. Put client-side redirects for all changed files (redirect from: old filename.html)
2. config_managed.yml
3. _includes /topnav
4. Server-side redirects: Jesse's taking care of it
5. Change links in CC console
6. Jesse to look into the required updates for the build process.

jseldess:
@mjibson, I'd really like your review of the last 4 commits. As you'll see, we're threading the managed cockroachdb -> cockroachcloud rebranding through our docs and build process. I think my changes cover everything needed for local and PR builds, but I'm not entirely sure what else needs to change for the final publishing step post-merge. I see this in the Publish TeamCity config:

Screen Shot 2019-10-10 at 5 07 19 PM

I think we need to change `managed` to `cockroachcloud` in that config right before we merge this PR, is that right? I'm assuming that we're still publishing to GitHub Pages and the integration with Netlify is somehow automated, but I'd love further insight.

Makefile
```diff
@@ -42,22 +42,25 @@ extra-config := $(if $(JEKYLLCONFIG),$(comma)$(JEKYLLCONFIG))
...
+.PHONY: cockroachdb
+cockroachdb: jekyll-action := serve --port 4000
+cockroachdb: bootstrap
+	bundle exec jekyll $(jekyll-action) --incremental --config _config_base.yml,_config_cockroachdb.yml,_config_cockroachdb_local.yml$(extra-config) $(JEKYLLFLAGS)
 
 .PHONY: standard
-standard: jekyll-action := serve --port 4000
-standard: bootstrap
-	bundle exec jekyll $(jekyll-action) --incremental --config _config_base.yml,_config_standard.yml,_config_standard_local.yml$(extra-config) $(JEKYLLFLAGS)
+standard: cockroachdb
```
benesch: Intentional to keep "standard" as an alias for "cockroachdb" but not "managed" as an alias for "cockroachcloud"?

mjibson:
I've twiddled with the _redirects file and the version selector now seems to work. @jseldess this should be ready-to-merge. If you change the _redirects file (like in #5607), this will probably need a rebase and manual merge, which I can do. But maybe do another spot check to make sure I didn't miss anything?

mjibson:
Rebased with the new redirect from #5607.

jseldess:
@mjibson, if you go here and click the version selector, it opens a new window: https://deploy-preview-5559--cockroachdb-docs.netlify.com/docs/cockroachcloud/stable/. Any ideas?

mjibson:
Ergh well it worked before the rebase. Fixing again.

mjibson:
Well it uh works now so...I have no idea what's going on. But 8 ball says signs point to yes.

PR #5562, opened 2019-10-07
There are 2 RocksDB instances per store, not 3

### LINKED ISSUE #5561: Update storage layer re: # of instances of RocksDB
On https://www.cockroachlabs.com/docs/stable/architecture/storage-layer.html we say the following:

> This data is stored as key-value pairs on disk using RocksDB, which is treated primarily as a black-box API. Internally, each store contains three instances of RocksDB:
> - One for the Raft log
> - One for storing temporary distributed SQL data
> - One for all other data on the node

However, we do not have a separate instance for the Raft log.

rmloveland:
Fixes #5561


irfansharif:
LGTM

PR #5566, opened 2019-10-07
Updated sslmode flag for app samples

### LINKED ISSUE #4762: Update `build an app` docs
The `build an app with CockroachDB` docs currently have `sslmode=require`. We recommend `sslmode=verify-full` for MSO. Update the docs to match.

Amruta-Ranade:
Closes #4762

### DIFF
--- a/_includes/v19.1/app/BasicExample.java
+++ b/_includes/v19.1/app/BasicExample.java
@@ -38,7 +38,7 @@ public class BasicExample {
         ds.setUser("maxroach");
         ds.setPassword(null);
         ds.setSsl(true);
-        ds.setSslMode("require");
+        ds.setSslMode("verify-full");
         ds.setSslCert("certs/client.maxroach.crt");
         ds.setSslKey("certs/client.maxroach.key.pk8");
         ds.setReWriteBatchedInserts(true); // add `rewriteBatchedInserts=true` to pg connection string
--- a/_includes/v19.1/app/TxnSample.java
+++ b/_includes/v19.1/app/TxnSample.java
@@ -103,7 +103,7 @@ public class TxnSample {
         // Connect to the 'bank' database.
         Properties props = new Properties();
         props.setProperty("user", "maxroach");
-        props.setProperty("sslmode", "require");
+        props.setProperty("sslmode", "verify-full");
         props.setProperty("sslrootcert", "certs/ca.crt");
         props.setProperty("sslkey", "certs/client.maxroach.key.pk8");
         props.setProperty("sslcert", "certs/client.maxroach.crt");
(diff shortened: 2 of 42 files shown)


rmloveland:
@Amruta-Ranade have you checked if all of these examples still work with `sslmode=verify-full`?  It's possible that some of them might fail, but right now we don't have any automated way of checking that.

PS Sorry to jump in on this PR to bug you, I just saw it go by in an email and got curious.

Amruta-Ranade:
@rmloveland Thanks for jumping in! 

As I understand it, `sslmode=verify-full` is what we initially intended when we said `sslmode=require`. `sslmode=require` takes care of in-flight encryption but is not meant to check for certs (and thus does not verify servers and clients). Since our secure mode needs CA certs, node certs, and client certs, we need the mode that checks for all three. That mode is `sslmode=verify-full`. So using `sslmode-require` gives the false sense of security that we are using certs for auth when, in fact, we aren't. This doc explains it better than I just did: https://www.postgresql.org/docs/9.1/libpq-ssl.html

If our code samples fail with `sslmode-verify-full`, that could be cause for concern IMO. That said, I will be happy to hold off the PR until 19.2 is done and we have the bandwidth to test all samples. Let me know what you think :)

rmloveland:
Amruta Ranade <notifications@github.com> writes:

> @rmloveland Thanks for jumping in!

Thanks for starting the work on this!  I look forward to us having better more secure examples for sure.

> As I understand it, sslmode=verify-full is what we initially intended when we said sslmode=require. sslmode=require takes care of in-flight encryption but is not meant to check for certs (and thus does not verify servers and clients). Since our secure mode needs CA certs, node certs, and client certs, we need the mode that checks for all three. That mode is sslmode=verify-full. So using sslmode-require gives the false sense of security that we are using certs for auth when, in fact, we aren't. This doc explains it better than I just did: https://www.postgresql.org/docs/9.1/libpq-ssl.html

Thanks for explaining this.  I am totes on board with the security benefits that this change will bring!  I just don't want us to break our examples for users.

> If our code samples fail with sslmode-verify-full, that could be cause for concern IMO. That said, I will be happy to hold off the PR until 19.2 is done and we have the bandwidth to test all samples. Let me know what you think :)

I think we should hold off until we have bandwidth to test all code samples with `sslmode=verify-full`.  Otherwise we risk breaking some of the examples.

jseldess:
@Amruta-Ranade, I'm ok holding off until post-19.2, contrary to what I mentioned earlier on Slack. I think Rich is right that we should test everything before merging this.

jseldess:
@Amruta-Ranade, I'm going to go ahead and close this PR since things have evolved quite a bit in the meantime, both in CC and the docs.

PR #5568, opened 2019-10-07
Force index selection for DELETE and UPDATE statements

### LINKED ISSUE #5547: Expand on index hints to UPDATE and DELETE
Airtable entry:
https://airtable.com/tblD3oZPLJgGhCmch/viw4fl7JawCP1kOdq/reczqfWIqUUiIV9MH?blocks=hide

Github issue:
https://github.com/cockroachdb/cockroach/pull/40715

ericharmeling:
Fixes #5547.

Changes include:
- New sections in DELETE and UPDATE statement pages
- New examples in DELETE and UPDATE statement pages
- Callout on Indexes and Table Expressions pages

### DIFF
--- a/v19.2/delete.md
+++ b/v19.2/delete.md
@@ -81,6 +81,32 @@ If you are deleting a large amount of data using iterative `DELETE ... LIMIT` st
 
 For an explanation of why this happens, and for instructions showing how to iteratively delete rows in constant time, see [Why are my deletes getting slower over time?](sql-faqs.html#why-are-my-deletes-getting-slower-over-time).
 
+## Force index selection for deletes
+
+<span class="version-tag">New in v19.2:</span> By using the explicit index annotation (also known as "index hinting"), you can override [CockroachDB's index selection](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/) and use a specific [index](indexes.html) for deleting rows of a named table.
+
+{{site.data.alerts.callout_info}}
+Index selection can impact [performance](performance-best-practices-overview.html), but does not change the result of a query.
+{{site.data.alerts.end}}
+
+The syntax to force a specific index for a delete is:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> DELETE FROM table@my_idx;
+~~~
+
+This is equivalent to the longer expression:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> DELETE FROM table@{FORCE_INDEX=my_idx};
+~~~
+
+To view how the index hint modifies the [query plan](cost-based-optimizer.html#view-query-plan) that CockroachDB follows for deleting rows, use an [`EXPLAIN (OPT)`](explain.html#opt-option) statement. To see all indexes available on a table, use [`SHOW INDEXES`](show-index.html).
+
+For examples, see [Delete with index hints](#delete-with-index-hints).
+
 ## Examples
 
 {% include {{page.version.version}}/sql/movr-statements.md %}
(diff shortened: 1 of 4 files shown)


ericharmeling:
@justinj ... before I put Rich on review, can you check to make sure everything I say is correct and the examples make sense? :)

v19.2/delete.md
```diff
@@ -81,6 +81,32 @@ If you are deleting a large amount of data using iterative `DELETE ... LIMIT` st
...
+This is equivalent to the longer expression:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> DELETE FROM table@{FORCE_INDEX=my_idx};
+~~~
+
+To view how the index hint modifies the [query plan](cost-based-optimizer.html#view-query-plan) that CockroachDB follows for deleting rows, use an [`EXPLAIN (OPT)`](explain.html#opt-option) statement. To see all indexes available on a table, use [`SHOW INDEXES`](show-index.html).
+
+For examples of index hinting for deletes, see the [Delete on a specific index](#delete-on-a-specific-index) example.
```
rmloveland: We could use the word "example" less without losing anything.  This could be shorter as "For examples ..., see 'Delete on a specific index'."

v19.2/delete.md
```diff
@@ -218,6 +244,72 @@ To sort and return deleted rows, use a statement like the following:
...
++-------------------------------------------------------------------------------------------------------+
+  delete users
+   └── select
+        ├── scan users@users_name_city_idx
+        └── filters
+             └── id IN ('3d70a3d7-0a3d-4000-8000-00000000000c', '70a3d70a-3d70-4400-8000-000000000016')
+(5 rows)
+~~~
+
+The optimizer still scans the newly-created `users_name_city_idx` index when performing the delete. Although scanning the table on this index could still be the most efficient, you could want to assess the performance difference between using `users_name_city_idx` and an index on the `id` column, as you are performing a delete with a filter on the `id` column.
```
rmloveland: I think this is easier to understand as 'may' rather than 'could', as in "you may want to assess the performance difference".

v19.2/indexes.md
```diff
@@ -31,7 +31,7 @@ To create the most useful secondary indexes, you should also check out our [best
 
 Because each query can use only a single index, CockroachDB selects the index it calculates will scan the fewest rows (i.e., the fastest). For more detail, check out our blog post [Index Selection in CockroachDB](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/), which will show you how to use the [`EXPLAIN`](explain.html) statement for your query to see which index is being used.
 
-To override CockroachDB's index selection, you can also force [queries to use a specific index](table-expressions.html#force-index-selection) (also known as "index hinting").
+To override CockroachDB's index selection, you can also force queries [to use a specific index](table-expressions.html#force-index-selection) (also known as "index hinting"). Currently, index hinting is supported for [`SELECT`](select-clause.html#select-from-a-specific-index), [`DELETE`](delete.html#force-index-selection-for-deletes), and [`UPDATE`](update.html#force-index-selection-for-updates) statements.
```
rmloveland: I'd lose "Currently" since IMO it's unnecessary.  That something is currently the case is always implied by the fact that these docs track CRDB version numbers.

v19.2/table-expressions.md
```diff
@@ -87,6 +87,8 @@ For example:
 
 {% include {{page.version.version}}/misc/force-index-selection.md %}
 
+You can also force index selection for [`DELETE`](delete.html#delete-on-a-specific-index) and [`UPDATE`](update.html#update-on-a-specific-index) statements.
+
```
rmloveland: I see what you did there since the preceding snippet is also used in the SELECT docs.  Maybe this could be an info callout so it's harder to miss on this page?

v19.2/update.md
```diff
@@ -38,6 +38,32 @@ Parameter | Description
...
+This is equivalent to the longer expression:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> UPDATE table@{FORCE_INDEX=my_idx} SET ...
+~~~
+
+To view how the index hint modifies the [query plan](cost-based-optimizer.html#view-query-plan) that CockroachDB follows for updating rows, use an [`EXPLAIN (OPT)`](explain.html#opt-option) statement. To see all indexes available on a table, use [`SHOW INDEXES`](show-index.html).
+
+For examples of index hinting for updates, see the [Update on a specific index](#update-on-a-specific-index) example.
```
rmloveland: Same comment as above re: We could use the word "example" less without losing anything ("For examples, ... see 'Update on a specific index'.")

v19.2/update.md
```diff
@@ -476,6 +502,59 @@ IDs:
...
+        │    ├── scan users
+        │    │    └── flags: force-index=primary
+        │    └── filters
+        │         └── name = 'Michael Brown'
+        └── projections
+             └── const: 'Michael Brown (there are two)'
+(9 rows)
+~~~
+
+As you can see from the output of the `EXPLAIN` statement, the plan changed quite a bit when you forced the index selection.
```
rmloveland: It changed, but is this better or worse than before?  It's not at all clear to me as a non-expert at reading EXPLAINs.  The narrative structure/flow of this seems to suggest to me that the second result is something "better" (maybe?).  But the text content seems to be saying "it depends on which of these is better, but we ain't saying".  If the latter is the case I think we should be more explicit about that.

rmloveland:
@ericharmeling mostly LGTM but I was quite confused about one new section's goals re: what it's telling the reader to do (if anything?).

PR #5576, opened 2019-10-08
pg_collation_for and COLLATION FOR

### LINKED ISSUE #5285: sql: Add pg_collation_for info function
PR: https://github.com/cockroachdb/cockroach/pull/39198

From release notes:
> Added the `COLLATION FOR` syntax and `pg_collation_for()` [built-in function](functions-and-operators.html). [#39198][#39198] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5285.

- Entry for COLLATION FOR added to functions and operators doc page
- New example for pg_collation_for and COLLATION FOR added to COLLATE page

### DIFF
--- a/v19.2/collate.md
+++ b/v19.2/collate.md
@@ -71,13 +71,12 @@ The sort will now honor the `de` collation that treats *ä* as *a* in alphabetic
 > SELECT * FROM de_names ORDER BY name;
 ~~~
 ~~~
+    name
 +----------+
-|   name   |
-+----------+
-| Backhaus |
-| Bär      |
-| Baz      |
-+----------+
+  Backhaus
+  Bär
+  Baz
+(3 rows)
 ~~~
 
 ### Order by non-default collation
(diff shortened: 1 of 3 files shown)


v19.2/collate.md
```diff
@@ -128,6 +129,47 @@ You can also use casting to remove collations from values.
...
++----------+
+  Backhaus
+  Baz
+  Bär
+(3 rows)
+~~~
+
+### Show collation for strings
+
+<span class="version-tag">New in v19.2:</span> You can use the `pg_collation_for` [built-in function](functions-and-operators.html#string-and-byte-functions), or its alternative [syntax form](functions-and-operators.html#special-syntax-forms) `COLLATION FOR`, to return the collation of a collated string.
```
rmloveland: The couple things:
1. I think this should say "locale name of a collated string".  According to the logic test I'm seeing at https://github.com/cockroachdb/cockroach/pull/39198/files this function returns the "locale name of a collated string".  However here we're saying the "collation of a collated string". Is that the same thing?  In any case, if I don't know what collation is it's a bit of a circular definition.
2. Re: the new syntax form, pretty sure the SQL grammar needs to be updated also.  Feel free to ping me if you need help with that.  I haven't done it in quite a while so we can learn/remember together.

PR #5578, opened 2019-10-08
Update docs on epoch- vs. expiration-based leases

### LINKED ISSUE #5563: Update docs on epoch based leases
The [replication layer docs](https://www.cockroachlabs.com/docs/stable/architecture/replication-layer.html#epoch-based-leases-table-data) state that epoch based leases help us avoid "tracking leases for every range".

We should clarify exactly **how** epoch based leases help us avoid tracking leases for every range, since it isn't obvious from the current description.

rmloveland:
Fixes #5563.

### DIFF
--- a/v19.1/architecture/replication-layer.md
+++ b/v19.1/architecture/replication-layer.md
@@ -69,13 +69,15 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
 
 #### Epoch-based leases (table data)
 
-To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. When the node disconnects, the epoch is considered changed, and the node immediately loses all of its leases.
+To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the the epoch is considered changed. This causes the node to immediately lose all of its leases.
 
-This mechanism lets us avoid tracking leases for every range, which eliminates a substantial amount of traffic we would otherwise incur. Instead, we assume leases do not expire until a node loses connection.
+Because leases don't expire until a node disconnects from a cluster, leaseholders do not have to individually renew their own leases. Tying lease lifetimes to node liveness in this way lets us eliminate a substantial amount of traffic and Raft processing we would otherwise incur, while still tracking leases for every range.
 
 #### Expiration-based leases (meta and system ranges)
 
-Your table's meta and system ranges (detailed in the distribution layer) are treated as normal key-value data, and therefore have leases, as well. However, instead of using epochs, they have an expiration-based lease. These leases simply expire at a particular timestamp (typically a few seconds)––however, as long as the node continues proposing Raft commands, it continues to extend the expiration of the lease. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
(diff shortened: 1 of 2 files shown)


v19.1/architecture/replication-layer.md
```diff
@@ -69,13 +69,15 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
 
 #### Epoch-based leases (table data)
 
-To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. When the node disconnects, the epoch is considered changed, and the node immediately loses all of its leases.
+To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the the epoch is considered changed. This causes the node to immediately lose all of its leases.
 
 This mechanism lets us avoid tracking leases for every range, which eliminates a substantial amount of traffic we would otherwise incur. Instead, we assume leases do not expire until a node loses connection.
```
irfansharif: The language around "lets us avoid tracking leases for every range" is actually what originally tripped me up. We still _do_ track leases for every range, but we're smarter about it. By assuming leases don't expire until a node loses connection, each leaseholder does not have to individually renew their own lease. By tying lease lifetimes to the node itself (that holds many many leaseholders), we eliminate a substantial amount of traffic and Raft processing we would otherwise incur. I'd include something to this effect here instead. Everything else LGTM.

rmloveland:
irfan sharif <notifications@github.com> writes:

> The language around "lets us avoid tracking leases for every range" is actually what originally tripped me up. We still do track leases for every range, but we're smarter about it. By assuming leases don't expire until a node loses connection, each leaseholder does not have to individually renew their own lease. By tying lease lifetimes to the node itself (that holds many many leaseholders), we eliminate a substantial amount of traffic and Raft processing we would otherwise incur. I'd include something to this effect here instead. Everything else LGTM.

Ah, I missed/forgot that that was the crux of the issue.  Replaced that paragraph with a rephrasing of what you said above that I hope makes things clearer.  PTAL.

v19.1/architecture/replication-layer.md
```diff
@@ -69,13 +69,15 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
...
+A table's meta and system ranges (detailed in the [distribution layer](distribution-layer.html#meta-ranges)) are treated as normal key-value data, and therefore have leases just like table data.
+
+However, unlike table data, system ranges cannot use epoch-based leases because that would create a circular dependency: system ranges are already being used to implement epoch-based leases for table data. Therefore, system ranges must use expiration-based leases. Expiration-based leases expire at a particular timestamp (typically after a few seconds). However, as long as a node continues proposing Raft commands, it continues to extend the expiration of its leases. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
```
irfansharif: (_extreme_) nit: I wouldn't say "system ranges **must** use expiration-based leases", I'd just say "system ranges use expiration-based leases instead".

irfansharif:
Everything LGTM. Thanks!

rmloveland:
irfan sharif writes:

> (extreme) nit: I wouldn't say "system ranges must use expiration-based leases", I'd just say "system ranges use expiration-based leases instead".

Updated to say that.  And I appreciate the clarification!  I would prefer not to use "must" except when it's the truth.

Thanks for the thorough review!

rmloveland:
Thanks Eric!

PR #5582, opened 2019-10-09
Use one space after a period in architecture docs

rmloveland:
Updates the architecture docs for versions 19.1 and 19.2 to only use one space after a period, per our style guide.


rmloveland:
Eric, I created this PR because I wanted to kick off a new site build to cause it to redeploy and maybe "fix" it.  It's not urgent but would like to get it in at some point.

rmloveland:
Thanks @ericharmeling !  The "fix" I was referring to was to try to fix the badness from #5584 by kicking off a new site build, but it has since been resolved (for the time being - probably a CDN issue).

PR #5590, opened 2019-10-10
Updated Copy feature for the code snippets

NiteshKoushik:
Updated Copy feature for the code snippets in 4 places for the development version


jseldess:
We don't use the copy feature in CLI synopses.

PR #5591, opened 2019-10-10
Updated documents, so that proper scripts can be selected and copied

NiteshKoushik:
Updated documents, so that proper scripts can be selected and copied

### DIFF
--- a/v19.2/training/sql-basics.md
+++ b/v19.2/training/sql-basics.md
@@ -277,16 +277,22 @@ Suppose Carl wants to withdraw $250 from his checking account. First, check that
 
 Here's how that would look (do not run this example yet):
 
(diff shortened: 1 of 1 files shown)


jseldess:
___
*[v19.2/training/sql-basics.md, line 280 at r1](https://reviewable.io/reviews/cockroachdb/docs/5591#-LquvUfPBnebYYRHffOW:-LquvUfPBnebYYRHffOX:brl4zmd) ([raw file](https://github.com/cockroachdb/docs/blob/51899dec4aca01c30a4ce153d256354ef325ab60/v19.2/training/sql-basics.md#L280)):*
> ```Markdown
> Here's how that would look (do not run this example yet):
> 
> ```

We don't want this here. The note above say we don't want you to run this.

PR #5600, opened 2019-10-11
Updating document and providing copy feature

NiteshKoushik:
Updating document and providing copy feature for 11 code snippets for the Dev version v19.2.0

### DIFF
--- a/v19.2/view-node-details.md
+++ b/v19.2/view-node-details.md
@@ -21,63 +21,74 @@ Subcommand | Usage
 
 List the IDs of active and inactive nodes:
 
+{% include copy-clipboard.html %}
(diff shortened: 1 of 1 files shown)


jseldess:
Appreciate this effort, @NiteshKoushik, but we don't want copy-to-clipboard for every code block. These are just a quick "synopsis" of the command variations.  Making them copyable doesn't help much. All the examples on the page have appropriate copy-paste buttons. 

I see that you've opened a number of PRs with these types of changes. If you wouldn't mind, instead, please track down all of the problems you see and open an issue for us to consider. That way, we can guide you to make changes we'd like and avoid spending your time on other changes.

PR #5601, opened 2019-10-11
Document Update

NiteshKoushik:
Changes done in 8 place, by providing copy feature for the scripts for the document version v19.2.0


jseldess:
We don't want the copy feature in these cases.

PR #5608, opened 2019-10-14
SET DEFAULT cascading action implicitly uses NULL

### LINKED ISSUE #5290: sql: allow cascading action when default is set implicitly to null
PR: https://github.com/cockroachdb/cockroach/pull/39136

From release notes:
> Columns without an explicit default value now support foreign keys with the `SET DEFAULT` action, in the same way as the `SET NULL` and `SET DEFAULT NULL` cases. [#39136][#39136] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5290. 

- Added note to parameters.
- Added example.
- Updated all example outputs, with minor corrections.


ericharmeling:
@jordanlewis Can you quickly check this for technical accuracy. Thanks!

jordanlewis:
Technically looks good. Stylistically I'd question the choice to call this out as "new in 19.2" - while strictly correct, I think that the PR is actually a bugfix rather than a new feature so it feels odd to me to call it out specially.

ericharmeling:
> Technically looks good. Stylistically I'd question the choice to call this out as "new in 19.2" - while strictly correct, I think that the PR is actually a bugfix rather than a new feature so it feels odd to me to call it out specially.

Awesome. Thank you. I removed that note.

v19.2/foreign-key.md
```diff
@@ -45,7 +45,7 @@ Multiple-column (composite) foreign keys only accept null values in the followin
 
 For more information about composite foreign keys, see the [composite foreign key matching](#composite-foreign-key-matching) section.
 
-Note that allowing null values in either your foreign key or referenced columns can degrade their referential integrity, since any key with a null value is never checked against the referenced table. To avoid this, you can use a [`NOT NULL` constraint](not-null.html) on foreign keys when [creating your tables](create-table.html). 
+Note that allowing null values in either your foreign key or referenced columns can degrade their referential integrity, since any key with a null value is never checked against the referenced table. To avoid this, you can use a [`NOT NULL` constraint](not-null.html) on foreign keys when [creating your tables](create-table.html).
```
rmloveland: ~This is just a preference, but I would prefer style changes to happen in separate PRs.  Preferably batched, as in "This PR removes all trailing whitespace at EOL in the 19.1 docs" or whatever.

v19.2/foreign-key.md
```diff
@@ -105,7 +105,7 @@ Parameter | Description
...
 `ON DELETE SET NULL` / `ON UPDATE SET NULL` | When a referenced foreign key is deleted or updated, respectively, the columns of all rows referencing that key will be set to `NULL`. The column must allow `NULL` or this update will fail.
-`ON DELETE SET DEFAULT` / `ON UPDATE SET DEFAULT` | When a referenced foreign key is deleted or updated, respectively, the columns of all rows referencing that key are set to the default value for that column. If the default value for the column is null, this will have the same effect as `ON DELETE SET NULL` or `ON UPDATE SET NULL`. The default value must still conform with all other constraints, such as `UNIQUE`.
+`ON DELETE SET DEFAULT` / `ON UPDATE SET DEFAULT` | When a referenced foreign key is deleted or updated, the columns of all rows referencing that key are set to the default value for that column. <br/><br/> If the default value for the column is null, or if no default value is provided and the column does not have a [`NOT NULL`](not-null.html) constraint, this will have the same effect as `ON DELETE SET NULL` or `ON UPDATE SET NULL`. The default value must still conform with all other constraints, such as `UNIQUE`.
```
rmloveland: ~Nice update. Also like the added line break for readability.

v19.2/foreign-key.md
```diff
@@ -251,12 +251,11 @@ The update to the referenced table returns an error because `id = 1001` is refer
...
 ~~~
+   id  |         email
 +------+------------------------+
-|  id  |         email          |
-+------+------------------------+
-| 1001 | a@co.tld               |
-| 1111 | info@cockroachlabs.com |
-+------+------------------------+
+  1001 | a@co.tld
+  1111 | info@cockroachlabs.com
```
rmloveland: ~Same comment as above re: batching style changes.  It makes the reviews more costly in terms of time & attention.

v19.2/foreign-key.md
```diff
@@ -437,17 +431,16 @@ Insert some records into the referencing table:
 
 {% include copy-clipboard.html %}
 ~~~ sql
-> SELECT * FROM customers_3;
+> SELECT * FROM orders_3;
```
rmloveland: ~This looks like a bugfix.  Is this a bug in the previous instructions?

PR #5625, opened 2019-10-15
fix leftover managed text

mjibson:


css/layouts/_version-nav.scss
```diff
@@ -67,7 +67,7 @@
           content: 'Managed CRDB';
```
jseldess: Can you correct this as well? Just change it to `CRC`?

mjibson: fixed

PR #5642, opened 2019-10-17
Added note to 19.1 docs about turning off range merging for manual splits

### LINKED ISSUE #4960: SPLIT AT Doc Update
Re: [SPLIT AT](https://cockroachlabs.com/docs//v19.1/split-at.html)

In 19.1 it should be called out that one can't manually split a table unless they turn off the merge queue by `SET CLUSTER SETTING kv.range_merge.queue_enabled = false`.

In 19.2 I believe the above limitation goes away and the new `with expiration` option should be documented. Introduced in #38004 by @jeffrey-xiao.

ericharmeling:
Fixes #4960.


bdarnell:
We should probably have a companion upgrade checklist: when upgrading to 19.2, check your cluster settings:
- If you disabled the range merge queue in order to use manual splits, consider turning it back on since manual splits no longer conflict with the range queue
- etc.

ericharmeling:
@bdarnell Thanks for the suggestion Ben! I updated the PR:
- Added a note about turning merging back on when upgrading (in SPLIT AT and in Range Merges docs)
- Removed the sections about "When to turn merging on/off", as they don't really add any information or apply in 19.2

I'm happy to add/remove any more information.

bdarnell:
I was thinking the note about removing the setting in upgrading would go in https://www.cockroachlabs.com/docs/dev/upgrade-cockroach-version.html

ericharmeling:
> I was thinking the note about removing the setting in upgrading would go in https://www.cockroachlabs.com/docs/dev/upgrade-cockroach-version.html

Looking at that page, I'm not sure exactly where putting a checklist like that should go in the current structure of the page. It might be best to create a new section at the bottom of the page (something like "Post-upgrade Checklist"), with some suggestions about changing settings for performance optimization and new features in 19.2. That probably should be a separate issue and PR. That separate issue might also inform the issue that @jseldess opened about finalizing an upgrade: #5646

@jseldess Thoughts?

jseldess:
If this action needs to be taken after finalizing the upgrade, I agree that we could either add it to https://www.cockroachlabs.com/docs/dev/upgrade-cockroach-version.html#step-5-finish-the-upgrade, or create a subsequent step.

ericharmeling:
@bdarnell 

> We should probably have a companion upgrade checklist: when upgrading to 19.2, check your cluster settings:
>
> - If you disabled the range merge queue in order to use manual splits, consider turning it back on since manual splits no longer conflict with the range queue
> - etc.

A couple follow-up questions:
- Is there a reason why reseting `kv.range_merge.queue_enabled` is not automated as a part of the upgrade finalization process?
- What other cluster settings (the "etc." you mentioned above) need checking/updating post-upgrade?
- Are there any other post-upgrade steps that you know of that we are not documenting on [Upgrade to CockroachDB v19.2](https://www.cockroachlabs.com/docs/dev/upgrade-cockroach-version.html)?

bdarnell:
> Is there a reason why reseting kv.range_merge.queue_enabled is not automated as a part of the upgrade finalization process?

Because they might have turned it off for other reasons. We should be very cautious about automatically undoing user actions. 

> What other cluster settings (the "etc." you mentioned above) need checking/updating post-upgrade?

Good question. I don't know if there are any, but the "known limitation" list might be a good place to look. Any known limitation which A) had a workaround and B) is resolved in 19.2 would be worth listing here to let users know that the workaround can be removed.

ericharmeling:
Thanks @bdarnell. I opened a new issue for the update to the [Upgrade to CockroachDB v19.2](https://www.cockroachlabs.com/docs/dev/upgrade-cockroach-version.html) page, as I think it's out-of-scope for this PR (see #5672). I will open a separate PR to resolve that separate issue this week.

If you (or @andreimatei) are fine with the other changes in this PR, I can pass it off to @rmloveland to review for merging.

jseldess:
@bdarnell, you ok with this PR if we follow-up with a post-upgrade checklist?

Issue #5651, opened 2019-10-18
CockroachCloud docs -- search crawlers


jess-edwards:
When I search "CockroachCloud docs" on Google, none of the cloud docs are coming up. 

My guess is that some of the "make canonical" rules that we established when we published docs for Managed CockroachDB are still in place and are suppressing CockroachCloud pages. With CockroachCloud using a distinct product name from CockroachDB, I'd expect crawlers and users to experience much less confusion than when it was Managed CockroachDB vs CockroachDB.

It could also be the case that we need to ask Google to re-index CockroachCloud docs now that the name change has gone through.

A non-exhaustive list of pages I would expect to see in search results: 
1. [Sign up for a CockroachCloud cluster](https://www.cockroachlabs.com/docs/cockroachcloud/stable/cockroachcloud-sign-up-for-a-cluster.html)
2. [CockroachCloud docs](https://www.cockroachlabs.com/docs/cockroachcloud/stable/
)  <--- What do you think of retitling this page to "CockroachCloud Docs" instead of "CockroachDB Docs", and is that possible?
3. [Connect to your CockroachCloud Cluster](https://www.cockroachlabs.com/docs/cockroachcloud/stable/cockroachcloud-connect-to-your-cluster.html)
4. [CockroachCloud Security](https://www.cockroachlabs.com/docs/cockroachcloud/stable/cockroachcloud-security-overview.html)

cc @jseldess

jess-edwards:
CC @jeffreyroach who may be able to help here. He's currently tinkering with the sitemap.

Amruta-Ranade:
Yep..this is a known issue. @jseldess is going to help me fix this before beta.

jseldess:
I think it's as simple as removing this condition: https://github.com/cockroachdb/docs/blob/master/_includes/head.html#L16. I'll look into it further.

jseldess:
#2 is a bit tricky because we're using the same index.md file to generate both the cockroachdb docs landing page and the cockroachcloud docs landing page. I need to figure something else out.

jseldess:
OK, Google is now indexing CockroachCloud docs, though there aren't enough of them to really catch most keywords at this time. 

Next, I'll work on the cloud landing page. Then I think this is resolved for now.

jess-edwards:
@jseldess looks like Google is catching the CockroachCloud docs landing page! 

It hasn't caught much else yet.

[...]

PR #5652, opened 2019-10-19
Remove cockroachcloud canonical reference to cockroachdb docs

jseldess:
Address #5651.


jseldess:
Just going to merge this so we can see if it has any effect in next few days.

PR #5658, opened 2019-10-21
FOR UPDATE docs

### LINKED ISSUE #5516: parser: handle FOR UPDATE [ OF tables ]
PR: https://github.com/cockroachdb/cockroach/pull/40962

From release notes:
> Added parsing for `FOR UPDATE [ OF tables ]` form. [#40962][#40962] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5516.

- Updated selection query diagram
- Updated selection query parameters
- Added Locking with FOR UPDATE section to PostgreSQL compatibility page
- Updated Oracle Migration Guide


v19.2/porting-postgres.md
```diff
@@ -101,3 +101,11 @@ In CockroachDB, no such modulo is performed.
 ~~~ sql
 SELECT 1::int << (x % 64)
 ~~~
+
+### Locking with `FOR UPDATE`
+
+<span class="version-tag">New in v19.2:</span> The `FOR UPDATE` [locking clause](sql-grammar.html#locking_clause) is supported in [selection queries](selection-queries.html#parameters) for database migration compatibility.
+
+In CockroachDB, the `FOR UPDATE` clause places an advisory lock on data for performance optimization. It does not provide stronger, PostgreSQL-standard locking guarantees for uses unrelated to performance, such as applications that require locks to protect data from concurrent access altogether.
```
knz: If I understand correctly, this paragraph will be good for 20.1 but is not yet true as of 19.2.

For 19.2 the sentence would be 

> In CockroachDB, the `FOR UPDATE` clause is a no-op, because CockroachDB only supports SERIALIZABLE isolation and does not need locking to order concurrent transactions relative to each other. It does not provide stronger, PostgreSQL-....

knz: "PostgreSQL-compatible" instead of "PostgreSQL-standard"

knz: You can even drop the first part since that is explained in the next paragraph below. Instead just start "It does not provide stronger, ..."

ericharmeling:
> Good but let's make a note to update this when 20.1/dev docs are forked.

Thanks! See #5676.

PR #5663, opened 2019-10-22
Remove references to older training pages

### LINKED ISSUE #5496: Remove references to old training pages
In conjunction with the launch of our online training platform and first interactive online course, we need to move the current content at https://www.cockroachlabs.com/docs/stable/training/ into our docs, or otherwise figure out what to do with it. Mostly, the pages are tutorials, some duplicative of pages already in the docs, others valuable additions.

jseldess:
Fixes #5496.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -10,37 +10,21 @@
         </li>
         <ul class="footer-sub-nav">
           <li>
+            <a href="https://www.cockroachlabs.com/product/">CockroachDB</a>
+          </li>
+          <li>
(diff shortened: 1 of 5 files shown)


thatnerd:
LGTM. 

Checked the following (somewhat ad-hoc) & found everything as expected:
• Checked the header & footer
• Searched for the string "training" in the downloaded preview landing page
• Verified that /training is still there & reachable with the url
  • Verified that the training nav on the left still works
• Ran a grep for `training` in v19.2/ (excluding the training directory) and found it only in an example directory

Issue #4655, opened 2019-04-11
Window Functions Doc Update 


mjibson:
Re: [Window Functions](https://cockroachlabs.com/docs//v2.1/window-functions.html)

I was reading over our window function doc and noticed a few things.

1) Most of the example queries can be converted into aggregate queries with `GROUP BY`, and have no need to use window functions. Worse, the performance of the window functions is much worse than the equivalent group by (6s -> 1.5s in one case, 4x slower).
2) There's no documentation about the `WINDOW w AS (PARTITION BY blah)` syntax besides it shows up in a few examples.
3) I think the query under "Total number of riders, and total revenue" is wrong since I get different results for just `select sum(revenue) from ride` which should return the same total revenue, but doesn't. Also if you just run the inner select it spits out multiple rows per user with identical revenues. Summing those seems wrong.

If we have examples on this page, I think they should only be things that aren't achievable using group by and aggregate functions. The only example that genuinely needed and used window functions was the one printing a row number, which is a thing that can easily be done by an application.

### Suggestions

Here's some ideas:

1) Vehicles by popularity per city. This shows the number of rides per vehicle per city and the % of rides that vehicle made in that city.

```
SELECT
    *,
    count / sum(count) OVER (PARTITION BY city)
        AS "pct rides in this city"
FROM
    (
        SELECT
            rides.city, type, count(*)
        FROM
            rides JOIN vehicles ON rides.vehicle_id = vehicles.id
        GROUP BY
            rides.city, type
    )
ORDER BY
    city, count DESC;
```

2) Top 5 riders by revenue per city:

```
SELECT
    *
FROM
    (
        SELECT
            *, rank() OVER (PARTITION BY city ORDER BY revenue DESC)
        FROM
            (
                SELECT
                    users.city, name, sum(revenue) AS revenue
                FROM
                    users JOIN rides ON users.id = rides.rider_id
                GROUP BY
                    users.city, name
            )
    )
WHERE
    rank <= 5
ORDER BY
    city, rank;
```

3) Top 5 rides by number of rides per city. This one is interesting because sometimes a city will have 6 results instead of 5 because there are multiple people with the same number of rides, so they tie.

```
SELECT
    *
FROM
    (
        SELECT
            *, rank() OVER (PARTITION BY city ORDER BY rides DESC)
        FROM
            (
                SELECT
                    users.city,
                    name,
                    sum(revenue) AS revenue,
                    count(*) AS rides
                FROM
[...]

PR #5669, opened 2019-10-22
Window frame docs

### LINKED ISSUE #5295: sql: add support for optional frame exclusion
PR: https://github.com/cockroachdb/cockroach/pull/28262

From release notes:
> CockroachDB now supports an optional frame exclusion clause in the specification of window frames. [#28262][#28262] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5295.
Informs #4655.

- Added window definition section
- Added window_definition diagram and parameters
- Added extended opt_frame_clause diagram and parameters
- Updated all output for Window Functions doc page
- Updated description of window functions to include OVER and WINDOW clause description

@mjibson I just saw #4655, but I agree that we need to update this page ASAP.

### DIFF
--- a/v19.2/window-functions.md
+++ b/v19.2/window-functions.md
@@ -12,23 +12,76 @@ For a complete list of supported window functions, see [Functions and Operators]
 All [aggregate functions][aggregate-functions] can also be used as [window functions][window-functions]. For more information, see the [Examples](#examples) below.
 {{site.data.alerts.end}}
 
-{{site.data.alerts.callout_info}}
-The examples on this page use the `users`, `rides`, and `vehicles` tables from our open-source fictional peer-to-peer vehicle-sharing application, [MovR](movr.html).
-{{site.data.alerts.end}}
+## Window definitions
+
+Window frames are defined in [`OVER` clauses](sql-grammar.html#over_clause) or [`WINDOW` clauses](sql-grammar.html#window_clause).
+
+### Syntax
+
+<div class="filters clearfix">
+  <button style="width: 15%" class="filter-button" data-scope="basic">Basic</button>
+  <button style="width: 15%" class="filter-button" data-scope="expanded">Expanded</button>
+</div>
+
+<br><br>
+
+<div class="filter-content" markdown="1" data-scope="basic">
+
+<div>
+  {% include {{ page.version.version }}/sql/diagrams/window_definition.html %}
+</div>
+
+### Parameters
+
+Parameter | Description
+----------|-------------
+`window_name` | The name of the new window frame.
+`opt_existing_window_name` | An optional name of an existing window frame, defined in a different window definition.
+`opt_partition_clause`  | An optional `PARTITION BY` clause.
+`opt_sort_clause` | An optional `ORDER BY` clause. See [Ordering Query Results](query-order.html) for details.
+`opt_frame_clause`  | An optional frame clause, which contains a frame boundary and/or an `EXCLUDE` clause.
+
+</div>
+
+<div class="filter-content" markdown="1" data-scope="expanded">
+
+<div>
+  {% include {{ page.version.version }}/sql/diagrams/window_definition.html %}
+</div>
+
+**opt_frame_clause ::=**
+
+<div>
+  {% include {{ page.version.version }}/sql/diagrams/opt_frame_clause.html %}
+</div>
+
+### Parameters
+
+Parameter | Description
+----------|-------------
+`window_name` | The name of the new window frame.
+`opt_existing_window_name` | An optional name of an existing window frame, defined in a different window definition.
+`opt_partition_clause`  | An optional `PARTITION BY` clause.
+`opt_sort_clause` | An optional `ORDER BY` clause. See [Ordering Query Results](query-order.html) for details.
+`frame_bound` | An optional frame boundary.<br>Valid start boundaries include `UNBOUNDED PRECEDING`, `<offset> PRECEDING`, and `CURRENT ROW`.<br>Valid end boundaries include `UNBOUNDED FOLLOWING`, `<offset> FOLLOWING`, and `CURRENT ROW`.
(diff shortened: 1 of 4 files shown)


mjibson:
Moving my review to the engineer who actually implemented these since I still don't really know how they work.

ericharmeling:
@mjibson Thanks!!

v19.2/window-functions.md
```diff
@@ -12,23 +12,76 @@ For a complete list of supported window functions, see [Functions and Operators]
...
+  {% include {{ page.version.version }}/sql/diagrams/window_definition.html %}
+</div>
+
+### Parameters
+
+Parameter | Description
+----------|-------------
+`window_name` | The name of the new window frame.
+`opt_existing_window_name` | An optional name of an existing window frame, defined in a different window definition.
+`opt_partition_clause`  | An optional `PARTITION BY` clause. See [Define Table Partitions](partitioning.html) for details.
```
rmloveland: I don't think the `PARTITION BY` clause we use in window functions is the same as the `PARTITION BY` syntax we introduced for geo-distributed clusters.  The window function `PARTITION BY` syntax is [also supported by PG](https://www.postgresql.org/docs/9.1/tutorial-window.html).

 Pretty sure we overloaded this phrase for the geo-distributed use case.

tl;dr: the link to "Define Table Partitions" can be removed

  @yuzefovich please shout if you disagree with the above just in case I'm misremembering somehow.

v19.2/window-functions.md
```diff
@@ -38,112 +91,31 @@ Its operation can be described as follows (numbered steps listed here correspond
 
 1. The outer `SELECT DISTINCT(city) ... FROM rides` creates a "virtual table" on which the window functions will operate.
 2. The window function `SUM(revenue) OVER ()` operates on a window frame containing all rows of the query output.
-3. The window function `SUM(revenue) OVER (PARTITION BY city)` operates on several window frames in turn; each frame contains the `revenue` columns for a different city (Amsterdam, Boston, L.A., etc.).
+3. The window function `SUM(revenue) OVER (PARTITION BY city)` operates on several window frames in turn; each frame contains the `revenue` columns for a different city [partition](partitioning.html) (Amsterdam, Boston, L.A., etc.).
```
rmloveland: I think my comment above re: this is not that `PARTITION BY` also applies here (but we can wait for Yahor to confirm).  It's unfortunate that movr is fundamentally geo-distributed and this particular syntax is overloaded since it makes it quite confusing IMO.

v19.2/window-functions.md
yuzefovich: Yes, you're right - those two are different unrelated things.

v19.2/window-functions.md
```diff
@@ -38,112 +91,31 @@ Its operation can be described as follows (numbered steps listed here correspond
...
 
 <img src="{{ 'images/v19.2/window-functions.png' | relative_url }}" alt="Window function diagram" style="border:1px solid #eee;max-width:100%" />
 
-## Caveats
+### Caveats
 
-The most important part of working with window functions is understanding what data will be in the frame that the window function will be operating on. By default, the window frame includes all of the rows of the partition. If you order the partition, the default frame includes all rows from the first row in the partition to the current row. In other words, adding an `ORDER BY` clause when you create the window frame (e.g., `PARTITION BY x ORDER by y`) has the following effects:
+The most important part of working with window functions is understanding what data will be in the frame that the window function will be operating on. By default, the window frame includes all of the rows of the partition. If you order the partition, the default frame includes all rows from the first row in the partition of the current row. In other words, adding an `ORDER BY` clause when you create the window frame (e.g., `PARTITION BY x ORDER by y`) has the following effects:
```
rmloveland: I think the previous use of "to" was correct?  As in "the window frame extends from row A to row B"?  Unless I'm fundamentally misunderstanding something, which is totes possible!

v19.2/window-functions.md
```diff
@@ -38,112 +91,31 @@ Its operation can be described as follows (numbered steps listed here correspond
...
 
 - It makes the rows inside the window frame ordered.
 - It changes what rows the function is called on - no longer all of the rows in the window frame, but a subset between the "first" row and the current row.
 
-Another way of saying this is that you can run a window function on either:
+Another way of saying this is that you can run a window function on:
```
rmloveland: I think this is still an either/or case unless I was wrong the first time around and something has changed?  If so please let me know.  But if not, removing "either" makes it seem like this is AND instead of XOR

v19.2/window-functions.md
ericharmeling: Ahh, OK. That makes more logical sense. Thanks Rich!

v19.2/window-functions.md
```diff
@@ -38,112 +91,31 @@ Its operation can be described as follows (numbered steps listed here correspond
...
-|       | vehicle_id) REFERENCES vehicles (city, id),                              |
-|       |     INDEX rides_auto_index_fk_vehicle_city_ref_vehicles (vehicle_city    |
-|       | ASC, vehicle_id ASC),                                                    |
-|       |     FAMILY "primary" (id, city, vehicle_city, rider_id, vehicle_id,      |
-|       | start_address, end_address, start_time, end_time, revenue),              |
-|       |     CONSTRAINT check_vehicle_city_city CHECK (vehicle_city = city)       |
-|       | )                                                                        |
-+-------+--------------------------------------------------------------------------+
-~~~
+{{site.data.alerts.callout_info}}
```
rmloveland: This could be a green "tip" box (they're rare!)

v19.2/window-functions.md
```diff
@@ -263,30 +229,30 @@ To see which customers have taken the most rides while generating the most reven
...
+  Judy White       |              15 |              818.00
+  Tony Ortiz       |              15 |              724.00
+  Cindy Medina     |              14 |              740.00
+  Tyler Dalton     |              13 |              786.00
+  Samantha Coffey  |              13 |              758.00
+  Nicole Mcmahon   |              13 |              696.00
 (10 rows)
 ~~~
 
+Note that in the query above, a `WINDOW` clause defines the window frame, and the `OVER` clauses reuse the window frame.
```
rmloveland: Noice

PR #5670, opened 2019-10-23
Add homepage front-matter field

### LINKED ISSUE #5651: CockroachCloud docs -- search crawlers

jseldess:
This uses the homepage_title field from the relevant
config as the index.html h1. CockroachCloud Docs for
CockroachCloud, and CockroachDB Docs for CockroachDB.

Fixes #5651.

### DIFF
--- a/v19.1/index.md
+++ b/v19.1/index.md
@@ -2,6 +2,7 @@
 title: CockroachDB Docs
 summary: CockroachDB user documentation.
 toc: true
+homepage: true
 contribute: false
 build_for: [cockroachdb, cockroachcloud]
 cta: false
--- a/v19.2/index.md
+++ b/v19.2/index.md
@@ -2,6 +2,7 @@
 title: CockroachDB Docs
 summary: CockroachDB user documentation.
 toc: true
+homepage: true
 contribute: false
 build_for: [cockroachdb, cockroachcloud]
 cta: false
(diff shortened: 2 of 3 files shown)


jseldess:
@Amruta-Ranade, please check the h1 on CockroachCloud docs vs. CockroachDB docs. I think I've fixed the problem.

Amruta-Ranade:
Perfect! Thanks so much!

Issue #5636, opened 2019-10-17
sql: new columns in crdb_internal.zones table


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/40493

From release notes:
> The [`SHOW PARTITIONS`](../v19.2/show-partitions.html) statement now displays inherited as well as explicitly set zone configurations for partitions. Also, it is now possible to retrieve full zone configurations via the `full_config_sql` column in the `crdb_internal.zones` table, and the new `zone_id` and `subzone_id` columns in the `crdb_internal.partitions` table point to partitions' corresponding zone configurations in `crdb_internal.zones`. [#40493][#40493] {% comment %}doc{% endcomment %}

ericharmeling:
I changed the name of this issue to cover what hasn't already been filed/documented (the `crdb_internal` table).
See #5364 and #5541 - these covered the inherited constraints issue.

Several duplicates of this issue have also popped up (#5530, #5531, #5532).

rmloveland:
@ericharmeling FYI I'm addressing the `crdb_internal.zones` part of this issue as part of #4730.

rmloveland:
@ericharmeling I've addressed the following bits re: `crdb_internal.zones` in #5679 

> Also, it is now possible to retrieve full zone configurations via the full_config_sql column in the crdb_internal.zones table, and the new zone_id and subzone_id columns in the crdb_internal.partitions table point to partitions' corresponding zone configurations in crdb_internal.zones

I think the only thing left is matching up `crdb_internal.partitions` with `crdb_internal.zones`, but it looks like that work may happen via #4978 ?

tl;dr: can we close this?  or do we need it to stay open to track the `zones <-> partitions` connection work?

ericharmeling:
@rmloveland I think we can close this.

`crdb_internal.zones` is documented in #5679, and we direct users to query `crdb_internal.partitions` for partition information for "programmatic" purposes in #5722. 

And yes, we plan to go into much more detail about all `crdb_internal` tables with #4978.

PR #5679, opened 2019-10-23
Document how to check constraint conformance etc.

### LINKED ISSUE #4730: Better insight into constraint conformance
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwchkdQkRUw6hZ3z/reckmMLBqVHuJjlVn?blocks=hide
PM: @awoods187, @piyush-singh 
Eng: @andreimatei

rmloveland:
Specifically, add docs for the new system tables that provide insight
into:

- Whether data is under-replicated or unavailable

- Violations of data placement requirements

- Which localities have data that is at risk in the event of node
  failure

Fixes #4730.

Addresses #2957, #5636.


v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+|------------------+-------------------------------+---------------------------------------------------------------------------------------------------------|
+| zone_id          | [`INT8`](int.html)            | The ID of the [replication zone](configure-zone.html).                                                  |
+| subzone_id       | [`INT8`](int.html)            | The ID of the subzone (i.e., [partition](partition-by.html)).                                           |
+| type             | [`STRING`](string.html)       | The type of zone configuration that was violated, e.g., `constraint`.                                   |
+| config           | [`STRING`](string.html)       | The YAML key-value pair used to configure the zone, e.g., `+region=europe-west1`.                       |
+| report_id        | [`INT8`](int.html)            | The ID of the [report](#system-reports_meta) that generated this data.                                  |
+| violation_start  | [`TIMESTAMP`](timestamp.html) | The time when the violation occurred, in UTC.                                                           |
```
andreimatei: we screwed up here; this should have been `timestamptz` and then you don't have to say anything about UTC. I've sent a PR to fix it and will backport. I'd change the docs to reflect that.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+{% include copy-clipboard.html %}
+~~~ sql
+SHOW COLUMNS FROM system.replication_stats;
+~~~
+
+| Column name             | Data type          | Description                                                                            |
+|-------------------------+--------------------+----------------------------------------------------------------------------------------|
+| zone_id                 | [`INT8`](int.html) | The ID of the [replication zone](configure-zone.html).                                 |
+| subzone_id              | [`INT8`](int.html) | The ID of the subzone (i.e., [partition](partition-by.html)).                          |
+| report_id               | [`INT8`](int.html) | The ID of the [report](#system-reports_meta) that generated this data.                 |
```
andreimatei: I think it'd be useful to introduce some terms applicable to all these tables: a "report" is generated periodically and it contains a number of "report entries" (one per zone).

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| total_ranges            | [`INT8`](int.html) | Total [ranges](architecture/overview.html#architecture-range) in the cluster.          |
```
andreimatei: Not in the cluster! In the zone that this report entry refers to (and all the child ranges that don't get their own report entry cause they haven't overwritten the respective attributes).

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| unavailable_ranges      | [`INT8`](int.html) | Unavailable ranges.                                                                    |
```
andreimatei: for all the following fields I think it'd be useful to somehow make it clear that these counts are relative to the zone/subzone that this report entry is talking about

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| under_replicated_ranges | [`INT8`](int.html) | [Under-replicated](admin-ui-replication-dashboard.html#under-replicated-ranges) ranges. |
+| over_replicated_ranges  | [`INT8`](int.html) | Over-replicated ranges.                                                                |
+
+### system.replication_critical_localities
+
+The `system.replication_critical_localities` table shows which localities in your cluster have data that is at risk in the event of node failures in your cluster.
```
andreimatei: Like I was saying above, I'd explain what a critical locality is more and explain that it's not necessarily bad to have them.
I'd also find a way to work the term "locality tier" into this discussion.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+{% include copy-clipboard.html %}
+~~~ sql
+SHOW COLUMNS FROM system.replication_critical_localities;
+~~~
+
+| Column name    | Data type               | Description                                                                                                             |
+|----------------+-------------------------+-------------------------------------------------------------------------------------------------------------------------|
+| zone_id        | [`INT8`](int.html)      | The ID of the [replication zone](configure-zone.html).                                                                  |
+| subzone_id     | [`INT8`](int.html)      | The ID of the subzone (i.e., [partition](partition-by.html)).                                                           |
+| locality       | [`STRING`](string.html) | The human-readable name of the critical [locality](configure-replication-zones.html#zone-config-node-locality).         |
```
andreimatei: human-readable as opposed to what?

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+{{site.data.alerts.end}}
+
+## Conformance reporting tables
+
+The following new and updated tables are available for verifying constraint conformance.
+
+- [`system.replication_stats`](#system-replication_stats) shows information about whether data is under-replicated, over-replicated, or unavailable.
+- [`system.replication_constraint_stats`](#system-replication_constraint_stats) shows a list of violations to any data placement requirements you've configured, along with information about whether the cluster is actively recovering from those violations.
+- [`system.replication_critical_localities`](#system-replication_critical_localities) shows which localities in your cluster have data that is at risk in the event of node failures.
+- [`system.reports_meta`](#system-reports_meta) lists the IDs and times when the reports ran that generated the data in the `system.replication_*` tables.
```
andreimatei: I think better terminology is to switch from say that a report "runs", to saying that a report is "produced" at a particular time, and that rows from the other tables are entries in one of these reports.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| report_id      | [`INT8`](int.html)      | The ID of the [report](#system-reports_meta) that generated this data.                                                  |
+| at_risk_ranges | [`INT8`](int.html)      | The [ranges](architecture/overview.html#architecture-range) that are at risk of data loss as of the time of the report. |
+
+{{site.data.alerts.callout_info}}
+If you have not [defined any localities](configure-replication-zones.html#zone-config-node-locality), this report will not return any results. It only reports on localities that have been explicitly defined.
+{{site.data.alerts.end}}
+
+### system.replication_constraint_stats
+
+The `system.replication_constraint_stats` table lists violations to any data placement requirements you've configured, along with information about whether the cluster is actively recovering from those violations.
```
andreimatei: again I'd scratch the thing about recovering

v19.2/verify-constraint-conformance.md
andreimatei: s/violation occurred/violation was detected

And say that its NULL if violating_ranges is 0.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+### system.reports_meta
+
+{% include copy-clipboard.html %}
+~~~ sql
+SHOW COLUMNS FROM system.reports_meta;
+~~~
+
+| Column name | Data type                     | Description                                                                                      |
+|-------------+-------------------------------+--------------------------------------------------------------------------------------------------|
+| id          | [`INT8`](int.html)            | The ID of the report run that generated the data available in the `system.replication_*` tables. |
```
andreimatei: again s/ID of the report run that generated the data/ID of the report that this entry is part of

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| generated   | [`TIMESTAMP`](timestamp.html) | When the report was generated, in UTC.                                                           |
+
+### crdb_internal.zones
+
+The `crdb_internal.zones` table is useful for:
+
+- Matching up data returned from the various conformance reports with the human-readable names of the databases and tables, indexes, and partitions where that data lives.
+- Viewing your cluster's zone configurations in various formats: YAML, SQL, etc.
```
andreimatei: I'd switch the order of these two bullets

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+~~~ sql
+SHOW COLUMNS FROM crdb_internal.zones;
+~~~
+
+| column_name         | data_type               | description                                                                                                          |
+|---------------------+-------------------------+----------------------------------------------------------------------------------------------------------------------|
+| zone_id             | [`INT8`](int.html)      | The ID of the [replication zone](configure-zone.html).                                                               |
+| subzone_id          | [`INT8`](int.html)      | The ID of the subzone (i.e., [partition](partition-by.html)).                                                        |
+| target              | [`STRING`](string.html) | The "object" that the constraint is being applied to, e.g., `PARTITION us_west OF INDEX movr.public.users@primary`.  |
+| range_name          | [`STRING`](string.html) | The human-readable printed representation of the range. Unimplemented.                                               |
```
andreimatei: unimplemented? I think this is filled in for zones like the default one that don't refer to a database/table/...

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+| database_name       | [`STRING`](string.html) | The [database](show-databases.html) where the `target`'s data is located.                                            |
+| table_name          | [`STRING`](string.html) | The [table](show-tables.html) where the `target`'s data is located.                                                  |
+| index_name          | [`STRING`](string.html) | The [index](show-index.html) where the `target`'s data is located.                                                   |
+| partition_name      | [`STRING`](string.html) | The [partition](show-partitions.html) where the `target`'s data is located.                                          |
+| raw_config_yaml     | [`STRING`](string.html) | The YAML you used to [configure this replication zone](configure-replication-zones.html).                            |
+| raw_config_sql      | [`STRING`](string.html) | The SQL you used to [configure this replication zone](configure-replication-zones.html).                             |
+| raw_config_protobuf | [`BYTES`](bytes.html)   | The protobuf representation of the YAML used to [configure this replication zone](configure-replication-zones.html). |
```
andreimatei: I think some of these columns are hidden, aren't they? (i.e. they don't show up in select *). I wouldn't document those.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+       55 | PARTITION europe_west OF INDEX movr.public.rides@rides_auto_index_fk_city_ref_users            | NULL       | movr          | rides      | rides_auto_index_fk_city_ref_users            |                       9  
+       55 | PARTITION europe_west OF INDEX movr.public.rides@rides_auto_index_fk_vehicle_city_ref_vehicles | NULL       | movr          | rides      | rides_auto_index_fk_vehicle_city_ref_vehicles |                       9  
+       55 | PARTITION europe_west OF INDEX movr.public.rides@rides_auto_index_fk_vehicle_city_ref_vehicles | NULL       | movr          | rides      | rides_auto_index_fk_vehicle_city_ref_vehicles |                       9  
+       55 | PARTITION europe_west OF INDEX movr.public.rides@rides_auto_index_fk_vehicle_city_ref_vehicles | NULL       | movr          | rides      | rides_auto_index_fk_vehicle_city_ref_vehicles |                       9  
+(30 rows)
+~~~
+
+### Find out which databases and tables have at-risk ranges
+
+By default, this geo-distributed demo cluster will not have any at-risk ranges. A range is considered "at-risk" if its data could be lost in case of a node failure.
```
andreimatei: s/node failure/<something else>

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+
+To force the cluster into a state where some ranges are at risk, issue the following statement, which tells it to store 9 copies of each range underlying the `rides` table (by default [it stores 3](configure-replication-zones.html#num_replicas)).
+
+{% include copy-clipboard.html %}
+~~~ sql
+ALTER TABLE rides CONFIGURE ZONE USING num_replicas=9;
+~~~
+
+Once the statement above runs, the cluster will rebalance so that it's storing 9 replicas of each range underlying the `rides` table. This puts that data at risk since you are storing 9 copies of the table's data on only 9 nodes.  If a node fails, you will be left with 8 replicas, and there will be no additional nodes to upreplicate the range to to fulfill the `num_replicas=9` requirement.
```
andreimatei: > If a node fails, you will be left with 8 replicas, and there will be no additional nodes to upreplicate the range to to fulfill the `num_replicas=9` requirement.

I don't think this is right. There might be a misunderstanding about what makes a locality be critical. It's not about whether the replication factor could still be met. In a 3 localities setup, 3 nodes per locality, and everything replicated 9x, no locality is critical. You can lose any one locality without losing quorum for any ranges.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+
+The following new and updated tables are available for verifying constraint conformance.
+
+- [`system.replication_stats`](#system-replication_stats) shows information about whether data is under-replicated, over-replicated, or unavailable.
+- [`system.replication_constraint_stats`](#system-replication_constraint_stats) shows a list of violations to any data placement requirements you've configured, along with information about whether the cluster is actively recovering from those violations.
+- [`system.replication_critical_localities`](#system-replication_critical_localities) shows which localities in your cluster have data that is at risk in the event of node failures.
+- [`system.reports_meta`](#system-reports_meta) lists the IDs and times when the reports ran that generated the data in the `system.replication_*` tables.
+- [`crdb_internal.zones`](#crdb_internal-zones) can be used with the tables above to figure out the database and table names where the non-conforming or at-risk data is located.
+
+To configure how often the conformance reports run, adjust the XXX cluster setting.
```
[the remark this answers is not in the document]
rmloveland: Fixed, thanks.  Added the example of tweaking the cluster setting as well.

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+summary: Verify that your cluster's data replication, data placement, and zone configurations are working as expected.
+keywords: availability zone, zone config, zone configs, zone configuration, constraint, constraints, 
+toc: true
+---
+
+<span class="version-tag">New in v19.2:</span> Several new and updated tables (listed below) are available to help you verify your cluster's constraint conformance. For example, you can check the following types of constraints:
+
+- See what data is under-replicated or unavailable
+- Show localities that are at risk in the event of node failures
+- See if any of your cluster's data placement constraints are being violated, and whether the cluster is recovering from those violations.
```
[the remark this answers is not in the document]
rmloveland: IDK - waiting and running a query was my guess.  Scratched it instead!

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+
+## Conformance reporting tables
+
+The following new and updated tables are available for verifying constraint conformance.
+
+- [`system.replication_stats`](#system-replication_stats) shows information about whether data is under-replicated, over-replicated, or unavailable.
+- [`system.replication_constraint_stats`](#system-replication_constraint_stats) shows a list of violations to any data placement requirements you've configured, along with information about whether the cluster is actively recovering from those violations.
+- [`system.replication_critical_localities`](#system-replication_critical_localities) shows which localities in your cluster have data that is at risk in the event of node failures.
+- [`system.reports_meta`](#system-reports_meta) lists the IDs and times when the reports ran that generated the data in the `system.replication_*` tables.
+- [`crdb_internal.zones`](#crdb_internal-zones) can be used with the tables above to figure out the database and table names where the non-conforming or at-risk data is located.
```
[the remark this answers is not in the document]
rmloveland: Hm, I don't understand.  From my user POV, I don't really know what "zone IDs" mean unless I can get at the table names etc. where the data "is located" (from my POV) that I can access from SQL.  I only introduced the use of `crdb_internal.zones` because it allowed me to map from zone IDs to databases and table names.

Are you saying any or all of:

1. the support for this mapping is not correct?
2. we should not try to do the mapping in this doc?  (i.e., remove use of `crdb_internal.zones`?)
3. something completely else (?)

v19.2/verify-constraint-conformance.md
```diff
@@ -0,0 +1,565 @@
...
+
+To configure how often the conformance reports run, adjust the XXX cluster setting.
+
+Only members of the `admin` role can access these tables. By default, the `root` user belongs to the `admin` role. For more information about users and roles, see [Manage Users](authorization.html#create-and-manage-users) and [Manage Roles](roles.html).
+
+{{site.data.alerts.callout_info}}
```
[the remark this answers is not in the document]
rmloveland: > That's not true (is it?!?)

Sorry, I hallucinated some text there.

Fixed (I think) by updating to use an edited version of your comment.

v19.2/verify-constraint-conformance.md
rmloveland: Where can I learn more about what is a "critical locality"? (other than your previous comment)

grepping the code & RFCs for "critical locality" returned one result from your PR adding these reports :-) 

Despite the existence of [Configure Replication Zones](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html) and friends, according to the Googles we have not really documented those terms in a public-facing way yet. :-/

(I did find lots of "locality tier" comments in the code so I'll read them and learn something about that in the meantime.)

v19.2/verify-constraint-conformance.md
rmloveland: Updated to use that terminology

v19.2/verify-constraint-conformance.md
rmloveland: Removed all references to recovering here and in other places

v19.2/verify-constraint-conformance.md
rmloveland: Switched.

v19.2/verify-constraint-conformance.md
rmloveland: Ah, that was my guess based on the fact that I was only seeing `NULL` for that column.  I've removed the word "unimplemented" but did not try to add any more explanation.  Do you think anything else is needed beyond calling it "the human-readable printed representation"?

v19.2/verify-constraint-conformance.md
rmloveland: I'll remove the `raw_*` columns but FWIW they do show up in the default output of

`SELECT * FROM crdb_internal.zones;`

It returns the columns

- zone_id
- subzone_id
- target
- range_name
- database_name
- table_name
- index_name
- partition_name
- raw_config_yaml
- raw_config_sql
- raw_config_protobuf
- full_config_yaml
- full_config_sql 

(This is in version `  CockroachDB CCL v19.2.0-rc.1 (x86_64-apple-darwin14, built 2019/10/14 18:28:24, go1.12.5)  `)

knz:
> So I'd be inclined to do something to make you happy sooner rather than later, but the question is what. That's why I was asking tour opinion about future views (or perhaps even future virtual tables) in the system database. Cause if that's a possibility, then it does give us a migration path and so people can probably just start using these names and we don't have to warn them.

The thing is, the UX design for this feature was simply not done. It's not that it was incorrect or that we were lacking the proper directions etc. I'd argue that it was _simply not done_, as in, nobody really cared about our API stability guidelines, the experimental-beta release process, etc.

There's a process failure right there, and I am hard at work to figure out what to do about this so it does not happen in the future.

**In the meantime** the proper thing to do is to acknowledge that the work was rushed and **our current standard way to acknowledge this** is to mark the thing as "experimental".

But I don't have a *quick answer* for you as to what's going to happen *afterwards* i.e. how to make it non-experimental. There's something to be said about publicizing `crdb_internal` because there is a huge ask about this. We also have huge precedent about introducing SHOW porcelain, and I am sincerely surprised this has not happened here from the get go. Oh well.

There are **multiple** things to be said **against** publicizing `system`, which I will come back to (later).

So in the meantime, let's keep it at the warning and the disclaimer.

knz:
> The thing is, if indeed we don't want people using the system.replication_foo names, then I don't know if we should have pages and pages of documentation about them. 

That's actually a legitimate concern. If @awoods187 wasn't so keen on pushing docs on this new feature, so it can be marketed/advertised as a flagship new feature, I'd also recommend *not documenting it* — because it was rushed and it's not polished enough yet.

But I do also acknowledge that there is so much push from sales and support to have _something_ like this feature that it'd be silly to be completely silent. I think @rmloveland did fantastic work here and all the content in these pages will naturally adapt (with a simple substitution) when we have the adequate `crdb_internal` views / tables / built-ins.

rmloveland:
@andreimatei are you ok with this doc after the latest round of changes?

If so I'll pass along to Docs team for review so we can have it merged by next week's release.

If not, I'm happy to make any necessary updates.

rmloveland:
@ericharmeling I think the engineering review is winding down, this should be ready for a look sometime Monday AM. 🤞

rmloveland:
Thanks Jesse!

Created https://github.com/cockroachdb/docs/issues/5831 to track adding linkage, etc.

rmloveland:
> I just saw "Query Replication Reports" used in a document as the name of the feature (which is "replication reports"). The fact that "query" here was supposed to be a verb and not a noun got lost.

Bummer!  Understandable error, though.

> What do you think about renaming this to just Replication Reports? It looks like not all pages in these guides have actions in their titles.

I'm fine with it.  @jseldess do you have any objection as long as we put in a redirect from the current name?

> Btw, do you think there's another place besides Guides in which we should list these tables?

As far as the tables specifically, I'm not sure - I don't think we document system tables anywhere.  

However, I just took a look through some other docs, and I think we should write something about these reports on the main [Configure Replication Zones](https://www.cockroachlabs.com/docs/v19.2/configure-replication-zones.html) page.  Maybe a new section on that page that talks about the replication reports and links to the longer doc.  What do you think of that approach?

rmloveland:
Created https://github.com/cockroachdb/docs/issues/6857 to work on the things mentioned in my previous note.  Please comment over there if you have thoughts/opinions.

PR #5689, opened 2019-10-25
Add note about quotes in unquoted CSV fields

### LINKED ISSUE #5404: Add note about keytab updates to GSSAPI authentication page
A customer using GSSAPI asked whether it's necessary to restart CockroachDB nodes when replacing server keytabs. From my testing, this does not appear to be necessary: keytabs can be replaced without needing to restart the server. We should document this on the page about GSSAPI authentication.

### LINKED ISSUE #5405: importccl: force csv import to fail on bad usage of quotes in csv fields
PR: https://github.com/cockroachdb/cockroach/pull/40424

From release notes:
> Quotes are no longer accepted inside unquoted CSV fields. [#40424][#40424] {% comment %}doc{% endcomment %}

jseldess:
Fixes #5405.

### DIFF
--- a/v19.2/import-into.md
+++ b/v19.2/import-into.md
@@ -68,7 +68,9 @@ For instructions and working examples showing how to migrate data from other dat
 Before using `IMPORT INTO`, you should have:
 
 - An existing table to incrementally import into (use [`CREATE TABLE`](create-table.html)).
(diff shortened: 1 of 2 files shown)


spaskob:
Sounds good, thank you

PR #5710, opened 2019-10-28
Querying partitioned data

### LINKED ISSUE #4796: Make it easy to focus queries onto partitioned data
Background: https://airtable.com/tblD3oZPLJgGhCmch/viwchkdQkRUw6hZ3z/recKgsWJj2YKGBh50?blocks=hide

PM: @awoods187

ericharmeling:
Fixes #4796. 

- Added "Querying partitions" include file.
- Included this file as a section in [Define Table Partitions](https://www.cockroachlabs.com/docs/dev/partitioning.html) and [PARTITION BY](https://www.cockroachlabs.com/docs/dev/partition-by.html) pages.

Still a WIP - please provide feedback.

### DIFF
--- a/_includes/v19.2/sql/querying-partitions.md
+++ b/_includes/v19.2/sql/querying-partitions.md
@@ -1,0 +1,163 @@
+## Querying partitions
+
+Similar to [indexes](indexes.html), partitions can improve query performance by limiting the numbers of rows that a query must scan. In the case of [geo-partitioned data](topology-geo-partitioned-replicas.html), partitioning can limit a query scan to data in a specific region.
+
+### Filtering on an indexed column
+
+If you filter the query of a partitioned table on a [column in the index directly following the partition prefix](indexes.html#indexing-columns), the [cost-based optimizer](cost-based-optimizer.html) creates a query plan that scans each partition in parallel, rather than performing a costly sequential scan of the entire table.
+
+For example, suppose that the tables in the [`movr`](movr.html) database are geo-partitioned by region, and you want to query the `users` table for information about a specific user.
+
+Here is the `CREATE TABLE` statement for the `users` table:
+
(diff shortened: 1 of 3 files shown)


awoods187:
Overall this PR looks really good. I don't think we should be using `Explain (opt, verbose)` because its output can be extremely long and overwhelming. We pretty much only want to recommend this for troubleshooting when you are working with a CRL. We should instead show the regular `EXPLAIN` which also shows the plans being scanned in parallel. cc @RaduBerinde

ericharmeling:
@awoods187 I updated this PR to just use the standard `EXPLAIN` statement.

Also, added @RaduBerinde for an engineering review. :)

awoods187:
lgtm

ericharmeling:
@RaduBerinde reminder to review :)

RaduBerinde:
___
*[_includes/v19.2/sql/querying-partitions.md, line 79 at r1](https://reviewable.io/reviews/cockroachdb/docs/5710#-Lt0Ki7NCilTzcxYCtu8:-Lt0rOdu-8i0wqemvQ0v:b215iqy) ([raw file](https://github.com/cockroachdb/docs/blob/ca8670f3e84eae1fd1952eb03b2b4f4bfb18f7ed/_includes/v19.2/sql/querying-partitions.md#L79)):*

Added an additional example EXPLAIN statement:

```
EXPLAIN SELECT * FROM users WHERE id='00000000-0000-4000-8000-000000000000' 
AND city IN ('amsterdam','boston','los angeles','new york','paris','rome','san francisco','seattle','washington dc');
```
</blockquote></details>

I meant we can add `city IN ('amsterdam','boston','los angeles','new york','paris','rome','san francisco','seattle','washington dc')` as a CHECK constraint in the schema, which would do the same without having to specify it on each query.

PR #5735, opened 2019-10-30
Performance Page Update: TPC-C 100k

awoods187:
Here is a commit with the updated performance page information for 100k. It will need to replace links with the blog post (when its ready) and with links to the 100k reproduction steps (when ready).


jseldess:
Resolved a conflict and tweaked a little. Merging once this passes TC.

PR #5741, opened 2019-10-30
Added CC troubleshooting page

### LINKED ISSUE #5596: CC: Document unhelpful error messages
Can add them to the troubleshooting guide.
https://docs.google.com/document/d/1_t4ZanNWAohp5QmqYBEr1vuRbHW5kKXmzF55gG18QKE/edit

Amruta-Ranade:
Closes #5596

### DIFF
--- a/v19.2/cockroachcloud-troubleshooting-page.md
+++ b/v19.2/cockroachcloud-troubleshooting-page.md
@@ -18,7 +18,7 @@ Error: x509: certificate signed by unknown authority
 Failed running "sql"
 ~~~
 
-**Solution:** Check if you are using the right cluster name in the connection string.
+**Solution:** Check if you are using the right cluster name in the [connection string](cockroachcloud-connect-to-your-cluster.html#step-3-generate-the-connection-string).
 
 ### Network not authorized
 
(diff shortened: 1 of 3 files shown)


kannanlakshmi:
Nice, LGTM! Thank for finding these errors!

Amruta-Ranade:
cc @jeremycbloom

v19.2/cockroachcloud-troubleshooting-page.md
```diff
@@ -0,0 +1,51 @@
...
+This page describes common CockroachCloud errors and their solutions.
+
+## Connection errors
+
+### Wrong connection string
+
+The following error is displayed on the terminal if you use the incorrect connection string while trying to connect to a cluster:
+
+~~~ shell
+Error: x509: certificate signed by unknown authority
```
chrisseto: I wouldn't say that I'm particularly well versed in the errors generated when connecting to a cluster but this error sounds like it would come from not having the correct `certs` directory specified?

The error below, ` i/o timeout`, could come from copying the connection string incorrectly as well.

Amruta-Ranade: Nope..the error you get if the path is incorrect is:

~~~
Error: open amruta-192-gs-ca.crt: no such file or directory
Failed running "sql"
~~~

But this could be confusing as well. Adding this scenario to the doc.

The errors that rise out of copying connection string incorrectly are tricky. It depends on what part of the connection string they got wrong. I also don't anticipate the users copying the connection string incorrectly since they just need to press the copy button and we will copy it correctly for them.

Amruta-Ranade: Done :)

PR #5762, opened 2019-10-31
Fix inaccurate TIMESTAMPTZ description

solongordon:
The timestamp docs specified that our TIMESTAMPTZ datatype stores a time
zone offset. This is not true. It always stores the timestamp in UTC.
This is explicitly stated further down on the same page: "However, it is
conceptually important to note that TIMESTAMPTZ does not store any time
zone data."

For some reason this was correct in the v2.0 docs and earlier but
incorrect in later docs.

### DIFF
--- a/v19.2/timestamp.md
+++ b/v19.2/timestamp.md
@@ -1,10 +1,10 @@
 ---
 title: TIMESTAMP
-summary: The TIMESTAMP data type stores a date and time pair in UTC, whereas TIMESTAMPTZ stores a date and time pair with a time zone offset from UTC.
+summary: The TIMESTAMP data type stores a date and time pair in UTC.
 toc: true
 ---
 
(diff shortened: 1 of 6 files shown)


v19.2/timestamp.md
```diff
@@ -1,10 +1,10 @@
 ---
 title: TIMESTAMP
-summary: The TIMESTAMP data type stores a date and time pair in UTC, whereas TIMESTAMPTZ stores a date and time pair with a time zone offset from UTC.
+summary: The TIMESTAMP data type stores a date and time pair in UTC.
 toc: true
 ---
 
-The `TIMESTAMP` [data type](data-types.html) stores a date and time pair in UTC, whereas `TIMESTAMPTZ` stores a date and time pair with a time zone offset from UTC.
+The `TIMESTAMP` [data type](data-types.html) stores a date and time pair in UTC.
```
solongordon: Could also say "The TIMESTAMP and TIMESTAMPTZ data types store a date and time pair in UTC." if that's clearer.

jseldess:
I'll go ahead and merge this and open a follow-up PR to address my nits.

PR #5774, opened 2019-11-04
Note that zigzag joins are on by default in 19.2

### LINKED ISSUE #4881: opt: turn on zigzag joins by default
PR: https://github.com/cockroachdb/cockroach/pull/35911

From release notes:
> Zigzag joins will now be planned by the optimizer in certain cases. They were previously available via the `experimental_enable_zigzag_join` session variable and can still be disabled by setting that value to `false`. [#35911][#35911] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #4881.


_includes/v19.2/misc/session-vars.html
```diff
@@ -95,6 +95,18 @@
    <td>Yes</td>
   </tr>
 
+  <tr>
+   <td>
+    <code>enable_zig_zag_join</code>
+   </td>
+   <td>Indicates whether the <a href="cost-based-optimizer.html">cost-based optimizer</a> will plan certain queries using a zig-zag merge join algorithm, which searches for the desired intersection by jumping back and forth between the indexes based on the fact that they share a sorted order in their key suffix. </td>
```
justinj: nit: it's not really their key suffix, it's like...the prefix after you remove any constant columns. there's probably a more succinct way to express that.

Maybe something like:
> ... after constraining indexes, they share an ordering.

LGTM besides that

rmloveland: Updated to use that language (I think).  PTAL.

rmloveland:
@ericharmeling just want a second set of eyes on this one before merging.  I've made the edit Justin asked for so I'm gonna call it good from a technical review POV.

PR #5775, opened 2019-11-04
Remove list of supported SQL from CBO docs

### LINKED ISSUE #5284: opt: support CREATE VIEW
PR: https://github.com/cockroachdb/cockroach/pull/39169

From release notes:
> [`CREATE VIEW`](../v19.2/create-view.html) now supports all queries supported by the optimizer, including those containing [correlated subqueries](../v19.2/subqueries.html#correlated-subqueries). [#39169][#39169] {% comment %}doc{% endcomment %}
>
> Mutations are no longer allowed inside [views](../v19.2/views.html). [#39169][#39169] {% comment %}doc{% endcomment %}

rmloveland:
Since almost everything is planned by the CBO in 19.2, maintaining an explicit list was getting unwieldy and is probably no longer necessary.

Fixes #5284.

### DIFF
--- a/v19.1/delete.md
+++ b/v19.1/delete.md
@@ -81,6 +81,30 @@ If you are deleting a large amount of data using iterative `DELETE ... LIMIT` st
 
 For an explanation of why this happens, and for instructions showing how to iteratively delete rows in constant time, see [Why are my deletes getting slower over time?](sql-faqs.html#why-are-my-deletes-getting-slower-over-time).
 
+## Force index selection for deletes
+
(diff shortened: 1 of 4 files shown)


rmloveland:
@RaduBerinde there isn't much to see here but I'd appreciate your 👍 to verify the approach of:

1. Removing the list of things the optimizer supports, since it's now a matter of what very few things it doesn't support
2. Choosing not to update the views documentation re: mutations not being allowed, since we did not state anywhere that mutations were allowed previously (i.e. it was undocumented behavior AFAICT).  Please correct me if we should say something here though

Issue #4696, opened 2019-04-30
Extend Kubernetes tutorials to cover using a custom CA


jseldess:
This relates closely to the topic of [adding known production guidance to our Kubernetes tutorials](https://github.com/cockroachdb/docs/issues/5459), but let's leave this as a stand-alone project because it's more involved.

jseldess:
MSO needs guidance on using cockroach CA cert to access cockroach from a separate Kubernetes cluster. Is this the same topic?

jseldess:
Resource: https://github.com/cockroachdb/cockroach/tree/master/cloud/kubernetes/bring-your-own-certs

jseldess:
Also: https://rinokadijk.github.io/vault-cockroach/

jseldess:
@taroface, here's another Kubernetes topic we should put on your backlog.

PR #5787, opened 2019-11-05
update K8s deploy docs

### LINKED ISSUE #5761: Vet/update Kubernetes single-cluster deployment tutorials
https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html

Will test: 

- Hosted GKE
- Hosted EKS
- Secure & Insecure

taroface:
Fixes #5761 

Couldn't fully verify the EKS steps using custom CA. May be related to [this issue](https://github.com/cockroachdb/cockroach/issues/42173), or [prometheus.yaml](https://github.com/cockroachdb/cockroach/blob/master/cloud/kubernetes/prometheus/prometheus.yaml) and [alertmanager.yaml](https://github.com/cockroachdb/cockroach/blob/master/cloud/kubernetes/prometheus/alertmanager.yaml) using the Kubernetes CA, which isn't supported on EKS. I will work through this in #4696.


taroface:
Thanks for catching this. It's a provisional update I had that slipped through before we had our conversation about how to approach this. For now, I'm reverting to the way it was before - hardcoding "v19.1.0" - and have replaced all the liquid snippets (some of which were there initially) in that doc with `v19.1.0`.

PR #5789, opened 2019-11-05
Apply unsupported version only to v1.0 and v1.1 pages

jseldess:


rmloveland:
w00t!  Would be great to discuss the path towards deleting these altogether.  Maybe we could zip up a directory of Markdown files for v1* and stick it in the Internet Archive or similar.

jseldess:
@mberhault, that was just plain sloppy. Now we're consistently using single-quotes (in that file, at least).

PR #5790, opened 2019-11-05
Disallow column references in ROWS/RANGE 

### LINKED ISSUE #5514: opt,sql: disallow column references in ROWS/RANGE
PR: https://github.com/cockroachdb/cockroach/pull/40832

From release notes:
> Column references are no longer allowed in `ROWS` and `RANGE` clauses in [window functions](../v19.2/window-functions.html). [#40832][#40832] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5514.

Added warning to [Window Function Syntax section](https://www.cockroachlabs.com/docs/dev/window-functions#parameters) stating that variables are no longer valid in ROW/RANGE clauses.

Do we want an example too? Not sure how much attention we want to bring to this, as our behavior before this change was undocumented and not postgres-like.

### DIFF
--- a/v19.2/window-functions.md
+++ b/v19.2/window-functions.md
@@ -68,6 +68,11 @@ Parameter | Description
 
 </div>
 
+
+{{site.data.alerts.callout_danger}}
+<span class="version-tag">New in v19.2:</span> For PostgreSQL compatibility, variables are no longer valid arguments in window frame `ROW`/`RANGE` clauses.
+{{site.data.alerts.end}}
+
 ## How window functions work
 
 At a high level, window functions work by:


ericharmeling:
> :lgtm: but I don't really think this is necessarily worth documenting (but I'm biased towards not documenting things)
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/5790#-:-LsxDp-vAXAK3EAdaC__:b-g6nkxj)_ status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @justinj)

I kind of agree... I'm going to close this and the issue.

PR #5797, opened 2019-11-06
Add transaction retry info to production checklist

### LINKED ISSUE #3266: mention tx retries in Production Checklist
I think this point is worth mentioning in prod checklist because it differs from how postgres works. Postgres uses locks and the client doesn't have to do a whole lot of retrying (small/simple apps may not need to do any), whereas roach will try and fail and leave retries to the client.

This is imo very crucial, but not clear from the production checklist. 

Additional details: https://forum.cockroachlabs.com/t/insert-failed-on-transaction/831/6

I suggest to add something of this sort:
> CockroachDB uses optimistic concurrency control instead of table or row locks. When several transactions compete for the same resource, all but one will fail. The client must implement a retry mechanism to avoid failures in production. See https://www.cockroachlabs.com/docs/stable/transactions.html#client-side-transaction-retries

to the production checklist at https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html

rmloveland:
Fixes #3266.

### DIFF
--- a/v19.1/recommended-production-settings.md
+++ b/v19.1/recommended-production-settings.md
@@ -513,3 +513,7 @@ When running CockroachDB on Kubernetes, making the following minimal customizati
 * Configure CPU and memory [resource requests and limits](kubernetes-performance.html#resource-requests-and-limits).
 
 For more information and additional customization suggestions, see our full detailed guide to [CockroachDB Performance on Kubernetes](kubernetes-performance.html).
+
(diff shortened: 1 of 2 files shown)


rmloveland:
@jseldess are you OK with these edits?

jseldess:
Yes, LGTM now, Rich. Thank you!

rmloveland:
Thanks Jesse!

Issue #5800, opened 2019-11-06
Update CBO nearest index selection example


rmloveland:
Richard Loveland (rmloveland) commented:

Prompted by [Update CBO nearest index selection example · Issue #4605 · cockroachdb/docs](https://github.com/cockroachdb/docs/issues/4605)  we are updating the [CBO nearest index preference docs](https://www.cockroachlabs.com/docs/v19.2/cost-based-optimizer.html#preferring-the-nearest-index) (for 19.1 and 19.2) to clarify exactly what users need to do to make sure the optimizer will use the right index(es).

However in addition to clarifying the requirements in prose, we should update the example to make it even clearer.  Right now the example could lead one to believe that you can blindly `SELECT * FROM ...` but in fact you have to make sure that every column you want to query using the replicated index is part of that index.  (@awoods187 hit this; for details, see [https://github.com/cockroachdb/cockroach/issues/35923)](https://github.com/cockroachdb/cockroach/issues/35923)).

Jira Issue: DOC-362

PR #5801, opened 2019-11-06
Clarify requirements for CBO using "nearest" index

### LINKED ISSUE #4605: cbo: clarify that to use the nearest index every column the query uses must be part of the index
As discovered in https://github.com/cockroachdb/cockroach/issues/35923#issuecomment-478128117, to use the nearest index every column the query uses must be part of the index. 

We should modify the example to have a third column in the table and then limit the query to only the things in the index. We should also add this explicit warning to the user here so that they know how to plan their queries and indexes.

rmloveland:
Fixes #4605.

Should be followed up ASAP with #5800.


rmloveland:
@awoods187 I hope this at least clears up the requirements for the feature to work while we get in #5800 ASAP.  PTAL at your convenience.

rmloveland:
@ericharmeling small one but FYI since you'll be closer to CBO going forward post-19.2 release

PR #5803, opened 2019-11-06
Document schema change DDL "partial commit" error

### LINKED ISSUE #4896: sql: use a CockroachDB-specific error code for schg errs upon COMMIT
PR: https://github.com/cockroachdb/cockroach/pull/37791

From release notes:
> CockroachDB now reports a dedicated SQL error code in the particular situation where a multi-statement transaction is committed but one of the schema change (DDL) operations failed. As in previous versions, CockroachDB leaves the transaction partly committed and partly aborted in that case; the new error code aims to facilitate the recognition of this case in client code. [#37791][#37791] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #4896.

### DIFF
--- a/_includes/v19.1/known-limitations/schema-change-ddl-inside-multi-statement-transactions.md
+++ b/_includes/v19.1/known-limitations/schema-change-ddl-inside-multi-statement-transactions.md
@@ -1,0 +1,52 @@
+Schema change [DDL](https://en.wikipedia.org/wiki/Data_definition_language#ALTER_statement) statements that run inside a multi-statement transaction with non-DDL statements can fail at [`COMMIT`](commit-transaction.html) time, even if other statements in the transaction succeed.  This leaves such transactions in a "partially committed, partially aborted" state that may require manual intervention to determine whether the DDL statements succeeded.
+
+If such a failure occurs, CockroachDB will emit the Postgres error code `40003`, `"statement completion unknown"`.
+
+{{site.data.alerts.callout_danger}}
+If you must execute schema change DDL statements inside a multi-statement transaction, we **strongly recommend** checking for this error code and handling it appropriately every time you execute such transactions.
+{{site.data.alerts.end}}
+
+This error will occur in various scenarios, including but not limited to:
+
+- Creating a unique index fails because values aren't unique.
+- The evaluation of a computed value fails.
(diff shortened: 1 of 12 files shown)


rmloveland:
@knz and @lucy-zhang,

Would really appreciate both of your eyes on this as it's a bit complex from a user's POV.  Feel free to redirect / choose one or the other of you as primary reviewer if that is more appropriate though.

Questions/comments:

1. Is this limitation/behavior in 19.1 as well, except that we return the PG 40003 error code?  If so I'm thinking we need to include a similar-but-different set of edits to the 19.1 docs.  Do y'all agree?

2. I made a small edit to the limitation called "no schema changes inside transactions", which is clearly not accurate since you can run schema changes inside transactions.  Hence I did a `s/no schema/schema/` there.

3. Since this is a bit lengthy/complex (including its own example), I chose to add it under its own heading instead of packing it in under the already existing "Schema changes inside transactions" limitation.  However an argument could be made that this info should go there.  What do you think?

4. Is the warning about "strongly recommending" the user to check for the error code too strong or just right?

5. Is there additional specific action the user should take beyond a general "when this error occurs, you should go and verify your data is what you think it is"

rmloveland:
@knz and @lucy-zhang, would it be better for me to remove one of you from the review to unstick things?

I can stick with Raphael as the primary reviewer as he is the author of https://github.com/cockroachdb/cockroach/pull/37791.  Does that work?

lucy-zhang:
Sorry for the delay. I'm taking a look now.

I don't have a preference on whether either of us should be the primary reviewer, but I'll try to review comprehensively in any case. knz knows more about the pg error codes than I do.

lucy-zhang:
> I think it'd be better to have one of the above as an example, and then a list of these three cases (the two listed above, plus the case in the example where evaluation of a computed value fails) as the most common scenarios, so that users know to look out for all of them. (There may be other sources of this error, but I can't think of any right now.)

There actually are other sources of this error...I had written the above before remembering the other causes listed in knz's table of schema changes linked above.

rmloveland:
@ericharmeling FYI the updates to 19.2 and 20.1 are identical.

The update to 19.1 is almost identical to the first 2, save that we emit a different (PG) error code.

So it's about 1/3 as much review as it looks like.

I will be OOO next week (and this will keep until then if needed), but I will be working tomorrow and can update/merge/etc. as needed then if you do end up getting to this.

PR #5805, opened 2019-11-06
delete: change fictional name

andreimatei:
I think Michael Brown was a bit inappropriate, particularly given that we were deleting him.


rmloveland:
Yikes! Thanks for this PR.

rmloveland:
PS LGTM if you don't want to wait to merge.

PR #5809, opened 2019-11-07
Update follower reads docs to use SET TRANSACTION

### LINKED ISSUE #5613: Follower reads should mention SET TRANSACTION syntax
https://www.cockroachlabs.com/docs/stable/follower-reads.html doesn't currently mention https://www.cockroachlabs.com/docs/v19.1/set-transaction.html#use-the-as-of-system-time-option, which is an important syntax to enable follower reads for ORMs. We should include it in that doc.

rmloveland:
... to make the feature easier to access from drivers and ORMs.

Fixes #5613.


rmloveland:
TFTR Nathan!

PR #5812, opened 2019-11-08
K8s prod requirements

### LINKED ISSUE #5459: Update single-region Kubernetes tutorials to reflect known prod requirements
Ahead of a full Kubernetes-specific production checklist, we need to update our single-region Kubernetes tutorials to offer the following important production guidance:

- Document recommended memory adjustments. See https://github.com/cockroachdb/cockroach/issues/35622 for full details, but the summary is that our k8s templates set `--cache` and `--max-sql-memory` to 25%, on the assumption that kubernetes sets the cgroup memory limit appropriately. That's not true, so kubernetes deployments commonly exceed their memory limits and crash. CockroachDB can tolerate nodes being OOM killed, but it's not good for performance and it's not something that should happen under normal usage. Until we implement an automatic fix for this, you can work around it in one of two ways:
    - Increase the memory allocated to the cockroach containers so they are allowed to use all of the machine's memory
    - Adjust [this line](https://github.com/cockroachdb/cockroach/blob/856ba9108f112f85d406bbe88d2208651859336e/cloud/kubernetes/v1.7/cockroachdb-statefulset-secure.yaml#L213) to reflect the amount of memory available. Replace both uses of 25% (in the `--cache` and `--max-sql-memory` flags) with amounts appropriate for the container's memory allocation. For example, if you're allocating 8GB of memory to the container, use `--cache 2GB` `--max-sql-memory 2GB`.
    - (Note from Jesse): We need to think through handling this via Helm as well. 

- Document resizing of Kubernetes storage volumes.
    - As of Kubernetes 1.11, the ability to resize persistent volume claims is on by default in the major cloud providers. Given that our configuration files default to such a small disk size, I think this is worth including a small section on in our Kubernetes docs.
    - Upstream documentation: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims

taroface:
Fixes #5459

- For the memory adjustments, I documented the second workaround that was listed in the issue. I was able to execute the steps for both Helm and config, but wasn't sure how to verify that the new cache and SQL memory size values were actually in effect (the Admin UI doesn't appear to reflect these limits). Will possibly need a technical review. If we also want to describe the first workaround, I will need some more technical context on this.
- For expanding the persistent volume claims, I was able to increase the disk sizes in production (GKE) even after manually setting `allowVolumeExpansion` to `false`. I documented the requirement to set this to `true` anyway, but could use clarity from an engineer on why that might have happened. Also, because I wasn't able to test this by changing the boolean, it's not clear to me whether the value has to be `true` before the pod is created, or only before the PVC is resized. The official k8s docs imply the former, but an OpenShift doc explicitly states that the pod has to have been created from a `StorageClass` where this value is `true`. I thought I should go with the first-party docs.

### DIFF
--- a/_includes/v19.1/orchestration/kubernetes-expand-disk-size.md
+++ b/_includes/v19.1/orchestration/kubernetes-expand-disk-size.md
@@ -85,7 +85,7 @@ You can expand certain [types of persistent volumes](https://kubernetes.io/docs/
     ~~~
 	</section>
 
-4. If the volume contains a file system, its capacity will not have changed. Confirm this by viewing the persistent volume claim:
+4. Check the capacity of the persistent volume claim:
 
 	<section class="filter-content" markdown="1" data-scope="helm">
     {% include copy-clipboard.html %}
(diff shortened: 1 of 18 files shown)


taroface:
@bdarnell Would you be able to review the updates to the memory adjustments? Also see my initial comment on this.

PR #5827, opened 2019-11-11
Revert "- Add doc about specifying precision for timestamp objects"

ericharmeling:
This reverts commit 4d287bdc7e354c3879ccede62211de51d0ad45e6 (PR #5535).

See https://github.com/cockroachdb/cockroach/pull/42285#issuecomment-551292412


ericharmeling:
@rohany FYI

Issue #4863, opened 2019-06-04
Complete Index pull request (hot ranges related)


thatnerd:
[Pull request here](https://github.com/cockroachdb/docs/pull/4846)

* May need to incorporate additional information from [Ben's blog post](https://docs.google.com/document/d/1vNCVtOzDzF7cKpVypUcJpu04_rswlk_IeqWgpTMtR0Q/edit#)

DoD
* All content around indexes is in 1-2 pull requests and those PRs are approved.

rmloveland:
@thatnerd can I steal this issue?  I would like to shepherd that PR.  I will add this issue to the upcoming Education sprint

jseldess:
@rmloveland, Ben's [blog post on choosing index keys](https://www.cockroachlabs.com/blog/how-to-choose-db-index-keys/) is now live. It'd be great for you to review that and incorporate insights/guidance as appropriate.

rmloveland:
@jseldess roger that

rmloveland:
Related: Interesting potential confirmation from customer POV of "composite PKs are faster than UUIDs" in the forum at https://forum.cockroachlabs.com/t/performance-of-insert-statements/2949

thatnerd:
Looks like I missed your original request to steal this. Glad you took it on @rmloveland

PR #5829, opened 2019-11-11
Update unique ID section of perf best practices

rmloveland:
Specifically:

- Updated 'Unique ID Best Practices' section with a table showing pros/cons of each approach for generating PKs / unique IDs: multi-column, UUID, serial ints

- Added a new approach describing how to get good performance with composite PKs

- Expanded our description of how hotspots can be caused by monotonically increasing index keys or table primary keys

Related to #4863


rmloveland:
@ajwerner these are the same changes that went into the 19.1 docs in #4863 , which you reviewed

I forgot to add those changes to the 19.2 docs at that time, sadly.

Do you mind taking a look and seeing if this content still basically makes sense?  Happy to update as needed.

rmloveland:
Thanks Andrew!  FYI just pushed a non-change to kick off the stuck build, sorry for noise

PR #5830, opened 2019-11-11
Fix typo on transaction layer page

rmloveland:


rmloveland:
@ericharmeling a quick typo fix

PR #5842, opened 2019-11-12
performance page: remove aurora comparison

awoods187:
We are removing the comparison vs. aurora from the performance page.


v19.2/performance.md
```diff
@@ -17,23 +17,17 @@ This document is about CockroachDB’s performance on benchmarks. For guidance o
...
-
-To learn more about our comparison with Amazon Aurora, see this [blog post](https://www.cockroachlabs.com/blog/cockroachdb-2dot1-performance/).
-{% comment %}replace link with 100k blog post{% endcomment %}
+                                   | CockroachDB       
+-----------------------------------|-------------------
+Max Throughput                     | 1,245,462 tpmC    
+Max Warehouses with Max Efficiency | 100,000 Warehouses
+Max Number of Rows                 | 49.8B             
+Max Unreplicated Data              | 8TB               
+Machine type                       | 81 c5d.9xlarge
```
irfansharif: Maybe pull out 81 into a separate row?

v19.2/performance.md
```diff
@@ -17,23 +17,17 @@ This document is about CockroachDB’s performance on benchmarks. For guidance o
...
-Max Warehouses with Max Efficiency | 100,000 Warehouses| 1,000 Warehouses
-Max Number of Rows                 | 49.8B             | 0.499B         
-Max Unreplicated Data              | 8TB               | 0.08TB         
-Machine type                       | 81 c5d.9xlarge    | 2 r3.8xl         
-
-Unlike Amazon Aurora, CockroachDB achieves this performance in [`SERIALIZABLE` isolation](demo-serializable.html), the strongest isolation level in the SQL standard. Like many other databases, Aurora selectively degrades isolation levels for performance, leaving workloads susceptible to fraud and data loss.
-
-To learn more about our comparison with Amazon Aurora, see this [blog post](https://www.cockroachlabs.com/blog/cockroachdb-2dot1-performance/).
-{% comment %}replace link with 100k blog post{% endcomment %}
+                                   | CockroachDB
```
nvanbenschoten: Does this header make sense to include anymore?

v19.2/performance.md
nvanbenschoten: +1

v19.2/performance.md
awoods187: it screws up the styling to remove it completely. I've made it Metrics/CockroachDB Results

v19.2/performance.md
awoods187: ack and done

awoods187:
We have those details but it would be a bit confusing on the machine types and nodes because we switched to using AWS and we switched CPU sizes. We used over hundred gcp nodes on 50 k but only 81 larger aws nodes on 100k.

awoods187:
I think we should merge as is to get this addressed and update with additional information in the future.

ajwerner:
> I think we should merge as is to get this addressed and update with additional information in the future.

I agree. We ran the 50k and 10k runs at different times on different versions. We should perhaps run them again and on 19.2 to have a reasonable table but that’s follow up work.

jseldess:
OK, merging this as-is. FWIW, I ran 1K and 10K on aws as part of the process of writing the repro steps, so I think I have that data. But not 50k.

PR #5866, opened 2019-11-14
Fix Jekyll plugin ordering

benesch:
The FlavorSelector plugin needs to run before the JekyllVersions plugin.
We were getting lucky previously with the right order, but For some
reason enabling the sitemap plugin caused us to get unlucky.

Use more granular plugin priorities to enforce exactly the right
ordering.


jseldess:
Thank you, @benesch! LGTM!

Issue #5889, opened 2019-11-19
CC: Client-side load balancing not required


Amruta-Ranade:
Based on https://support.cockroachlabs.com/access/unauthenticated?return_to=https%3A%2F%2Fcockroachdb.zendesk.com%2Fagent%2Ftickets%2F4106

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/5894

PR #5894, opened 2019-11-20
CC: Added FAQ for load balancers

Amruta-Ranade:
Addresses #5889


mattcrdb:
I think it's worth mentioning that the management console will generate a different connection url for a different selected region. Other than that lgtm. 👍

Amruta-Ranade:
> I think it's worth mentioning that the management console will generate a different connection url for a different selected region. Other than that lgtm. 👍

Oh, right. Fixed. TFTR!

PR #5899, opened 2019-11-20
Add note about property print.key flag

### LINKED ISSUE #5861: Edit CDC examples
https://www.cockroachlabs.com/docs/stable/change-data-capture#create-a-changefeed-connected-to-kafka

Tim V: i'd like to create a standard java library that easily converts CDC responses into java objects.  This is generally pretty simple and common stuff.  the problem i'm having is that our basic json message doesn't seem to me to be proper JSON and thus, makes this much more complex to deal with.  for example we emit something like [1] {"after": {"id": 1, "name": "Petee"}} but standard parsers treat this string as invalid because it starts with [1] instead of the traditional { to start a json object.  furthermore after is both a key and value here which makes things difficult.  what about a format that looks more like this {"keys": [1], "value": {"type":"after","content":{"id": 1, "name": "Petee"}}}

Our usage examples have `property print.key=true`, which prints both the key and value, which are separate fields. If you remove that flag, the changefeed will only print the value.

Dan says he'd be okay with simply removing the `--property print.key=true` from our examples and updating the output to match 

Roland: note that examples of deletes get really confusing if you omit keys

lnhsingh:
Closes #5861.

### DIFF
--- a/v20.1/change-data-capture.md
+++ b/v20.1/change-data-capture.md
@@ -213,7 +213,7 @@ Changefeed progress is exposed as a high-water timestamp that advances as the ch
     ~~~
             job_id       |  job_type  |                              description                               | ... |      high_water_timestamp      | error | coordinator_id
     +--------------------+------------+------------------------------------------------------------------------+ ... +--------------------------------+-------+----------------+
-      383870400694353921 | CHANGEFEED | CREATE CHANGEFEED FOR TABLE office_dogs2 INTO 'kafka://localhost:9092' | ... | 1537279405671006870.0000000000 |       |              1
+      383870400694353921 | CHANGEFEED | CREATE CHANGEFEED FOR TABLE office_dogs INTO 'kafka://localhost:9092' | ... | 1537279405671006870.0000000000 |       |              1
     (1 row)
     ~~~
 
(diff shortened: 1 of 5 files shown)


lnhsingh:
Change viewable under this section of the examples: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/91768503653c05f613311aafd86a5cf1ce68aae0/v20.1/change-data-capture.html#create-a-changefeed-connected-to-kafka

danhhz:
My takeaway from tim's confusion was that we shouldn't be printing the key (though I like the idea of a callout to point out that you can). The examples still feel pretty clear to me without the keys. @rolandcrosby what are your thoughts here?

lnhsingh:
Ok, in that case, I can swap the examples to not print the key and change the note to say "you can use print.key". I'll wait for @rolandcrosby 's feedback

lnhsingh:
@danhhz or @rolandcrosby -- can i get a LGTM from one of you if it LGTY?

danhhz:
I'm still not convinced that it's better to include the keys in most examples, I would have done something else for deletes. However, I don't think it makes a big enough difference either way that it's worth hashing it out and potentially rewriting this PR. So, I'm going to disagree and commit here, but I'll still defer to roland to give the LGTM

lnhsingh:
@danhhz reworked it since the examples don't show a delete anyway and this is a little more streamline from a docs perspective (only one note include across both json/avro examples)

v20.1/change-data-capture.md
```diff
@@ -540,17 +541,18 @@ In this example, you'll set up a changefeed for a single-node cluster that is co
...
     --bootstrap-server=localhost:9092 \
-    --property print.key=true \
     --from-beginning \
     --topic=office_dogs
     ~~~
 
     ~~~ shell
-    {"id":{"long":1}}	{"after":{"office_dogs":{"id":{"long":1},"name":{"string":"Petee"}}}}
-    {"id":{"long":2}}	{"after":{"office_dogs":{"id":{"long":2},"name":{"string":"Carl"}}}}
+    {"after":{"office_dogs2":{"id":{"long":1},"name":{"string":"Petee H"}}}}
```
danhhz: Where's this office_dogs2 coming from? I see a CREATE CHANGEFEED FOR office_dogs2, but I don't see where we create that table

lnhsingh: Good catch - that's a typo from me running a second changefeed. It should be `office_dogs` still.

PR #5939, opened 2019-11-26
Update drop-index.md

dbist:


dbist:
based on the (following)[https://github.com/cockroachdb/cockroach/pull/42413], `CASCADE` is not required.

PR #6079, opened 2019-11-27
Minor tweaks to roadmap script

jseldess:
- Get 5 airtable records at a time
- More efficient querystring parameters

### DIFF
--- a/scripts/issues-from-release-notes.py
+++ b/scripts/issues-from-release-notes.py
@@ -67,7 +67,8 @@ with open("../releases/" + release_notes) as file:
             try:
                 pr_num = line[line.rfind("[")+2:line.rfind("]")]
                 url = "https://api.github.com/repos/cockroachdb/cockroach/pulls/" + str(pr_num)
-                req = requests.get(url)
+                headers = {"Authorization": "token " + args.github_access_token}
+                req = requests.get(url, headers=headers)
                 resp = req.json()
                 title = resp["title"]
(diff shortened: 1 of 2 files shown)


knz:
The change looks good.

Here are a couple of suggestions for the rest of the program:
- readability:
  - extract the body of this loop: https://github.com/cockroachdb/docs/blob/master/scripts/issues-from-roadmap.py#L43-L61
    into a different function
  - extract the body of this action: https://github.com/cockroachdb/docs/blob/master/scripts/issues-from-roadmap.py#L91-L105
    into a different function
- maintainability/extensibility:
  - use a Python dictionary/map for the user/label mapping at https://github.com/cockroachdb/docs/blob/master/scripts/issues-from-roadmap.py#L64
  - ditto for the milestones at https://github.com/cockroachdb/docs/blob/master/scripts/issues-from-roadmap.py#L84
  - make the selection of Airtable view configurable e.g. via a command-line parameter. This will cause less headache if you ever want to experiment with multiple/different Airtable views.
  - Consider using the Github API using a Python library, e.g. https://pygithub.readthedocs.io/en/latest/introduction.html
  - Consider using the Airtable API using a Python library, e.g. https://github.com/josephbestjames/airtable.py or https://github.com/gtalarico/airtable-python-wrapper

jseldess:
Thanks very much, @knz! I'll take up your suggestions in a follow-up PR.

PR #6114, opened 2019-12-05
Update pagination docs to use keyset / seek method

### LINKED ISSUE #3743: Make client-side pagination a true recommendation
I heard from k yesterday that they thought pagination via offset/limit was slow and were looking for recommendations on what to do for large result sets.

I spoke with @jordanlewis and he recommended:
"Doing “client-side pagination” by retrieving a set of records with a limit, and then checking the index key of the last row, using that as an index constraint, and then running the query again
this is normal - not specific to cockroach"

Our current docs: https://www.cockroachlabs.com/docs/stable/selection-queries.html#limiting-row-count-and-pagination actually suggest the wrong solution for pagination.

The reason is that "offset doesn’t know anything smart - it has to get the same data as before, just skips the first n. You have to participate as a client by remembering the index key of the last result set you saw."

We should also reference that some databases offer a feature called cursors to do this. @andy-kimball mentioned that "cursors are generally not a great architecture as well b/c they force server to keep state. client side is way to go for pagination. server pagination just doesn't scale well"

Jordan also mentioned that "you run into trouble with a scale out system like cockroach - if your load balancer moves you to a different server for example the cursor will be lost"

Andy also mentioned that "sql server has cursors, and we spent a lot of time recommending customers not use them"

This way we can make it clear that client-side is the way to go and its not just because we don't have cursors.

rmloveland:
Summary of changes:

- Explain difference between keyset pagination and LIMIT/OFFSET
- Show examples of the former being fast and the latter being slow
- Show how to use EXPLAIN to check why the difference exists
- Add warning to LIMIT/OFFSET docs recommending keyset pagination
- ... all of the above for 19.1, 19.2, 20.1 docs

Fixes #3743

### DIFF
--- a/v19.1/limit-offset.md
+++ b/v19.1/limit-offset.md
@@ -26,6 +26,10 @@ as part of [`INSERT`](insert.html) or [`UPSERT`](upsert.html).
 `OFFSET` restricts the operation to skip the first `offset_value` number of rows.
 It is often used in conjunction with `LIMIT` to "paginate" through retrieved rows.
 
+{{site.data.alerts.callout_danger}}
+Using `LIMIT`/`OFFSET` to implement pagination can be very slow for large tables.  We recommend using [keyset pagination](selection-queries.html#pagination-example) instead.
+{{site.data.alerts.end}}
+
 For PostgreSQL compatibility, CockroachDB also supports `FETCH FIRST
 limit_val ROWS ONLY` and `FETCH NEXT limit_val ROWS ONLY` as aliases
 for `LIMIT`. If `limit_val` is omitted, then one row is fetched.
--- a/v19.1/selection-queries.md
+++ b/v19.1/selection-queries.md
@@ -337,26 +337,167 @@ LIMIT 5;
 
 ### Paginate through limited results
 
-If you want to limit the number of results, but go beyond the initial set, use `OFFSET` to proceed to the next set of results. This is often used to paginate through large tables where not all of the values need to be immediately retrieved.
+To iterate through a table one "page" of results at a time (also known as pagination) there are two options, only one of which is recommended:
+
+- Keyset pagination (**fast, recommended**)
+- `LIMIT` / `OFFSET` pagination (slow, not recommended)
+
+Keyset pagination (also known as the "seek method") is used to fetch a subset of records from a table quickly. It does this by restricting the set of records returned with a combination of `WHERE` and [`LIMIT`](limit-offset.html) clauses. To get the next page, you check the value of the column in the `WHERE` clause against the last row returned in the previous page of results.
+
+The general pattern for keyset pagination queries is:
 
 {% include copy-clipboard.html %}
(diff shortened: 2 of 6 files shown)


rmloveland:
@jordanlewis and @rafiss 

Since I think this pagination content is "app-dev" related:

Who's the best engineer to ask for a review?  Lots of opinions on the linked issue's thread.

rafiss:
I can do a review of this (at least from an app dev perspective).

rmloveland:
Thanks for the review @rafiss - I addressed your feedback in every case but one, where I had another question.

rmloveland:
kena <notifications@github.com> writes:

> So I think this conversation should be grounded properly by refering
> to what users usually expect, that is SQL cursors.
> CockroachDB does not support these but they are useful to set
> expectations right.
>
> A cursor is not much more than a long-standing transaction where the
> client can request more data from the same query through multiple
> (separate) statements.
>
> Importantly, cursors operate over a snapshot of the database at the
> moment the cursor is opened (it's possible to open non-consistent
> cursors but it's rarely used and also not always supported).
>
> So folk used to cursor already expect that pagination operates at a
> snapshot of the database established when the query starts.
>
> Bringing this back to CockroachDB, this means that it is a-OK to
> provide users with a feature that "anchors" the paginated read at some
> point in time, possibly slightly in the past (as long as it's not
> further in the past than the most recent statement in the same
> session).
>
> My opinion remains that proper pagination should be done using AOST
> with a follower read timestamp: this guarantees there won't be
> concurrent updates (and thus no risk of txn retries during the
> pagination), and also enables reads to access other replicas than the
> leaseholder. If necessary we can provide a feature to make such a
> paginated statement automatically wait for a small delay before
> starting the read, to ensure it will capture all the latest writes by
> the same session (or direct users to use pg_sleep() before their AOST
> query to achieve the same manually).

Thanks for providing this context Raphael.  I'll update this PR, then,
to show how to paginate with AOST and a follower read timestamp (or
~equivalent, since that's an enterprise feature) to mimic cursor-like
functionality.

rafiss:
I agree about recommending `AS OF SYSTEM TIME`. It also would not hurt to include a note that without `AS OF SYSTEM TIME`, the application should be aware that there could be missing/duplicated data during pagination.

rmloveland:
@rafiss thanks again for your review and comments.  I think the stuff you mentioned is in the latest push.

@knz since you shared some specific opinions, would you mind taking a look at the updates I just pushed?  [here is a direct link to the patch](https://github.com/cockroachdb/docs/pull/6114/commits/b7c31eeff17dc35d02c03609f9f3f9ccdabf2534).

I tried to capture the gist of what you said without bringing follower reads into it since those are an enterprise feature - went with '-1m' instead, since if I'm reading it right the current timestamp returned by `experimental_follower_read_timestamp()` is about 48s.

jseldess:
@rmloveland, looks like you can merge now?

PR #6115, opened 2019-12-05
Kerberos / GSSAPI docs initial update based on customer engagement

keith-mcclellan:
This format is all wrong but it has the additional info required to get started

### DIFF
--- a/v19.2/gssapi_authentication.md
+++ b/v19.2/gssapi_authentication.md
@@ -44,7 +44,7 @@ Copy the resulting keytab to the client machine. If you are connecting from mult
 
 ### MIT KDC
 
-In MIT KDC, you can't map a service principal to an SPN with a different user name, so you will need to create a service principal that includes the SPN for your client.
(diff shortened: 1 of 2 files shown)


v19.2/gssapi_authentication.md
```diff
@@ -103,16 +132,43 @@ The `cockroach sql` shell does not yet support GSSAPI authentication. You need t
 {{site.data.alerts.end}}
 
 1. Install and configure your Kerberos client.
+
+In CentOS/RHEL - `yum install krb5-user`
```
Amruta-Ranade: Is this specific for CentOS/RHEL? Do we have instructions for other systems?

keith-mcclellan: should be `apt-get install krb5-user`

jseldess:
@Amruta-Ranade, @keith-mcclellan, any updates on this PR?

v19.2/gssapi_authentication.md
Amruta-Ranade: Done.

PR #6116, opened 2019-12-05
make netlify/serve the same as `make cockroachdb`

mjibson:


mjibson:
We could do that, yes.

PR #6120, opened 2019-12-09
Update geo-partitioning.md

dbist:
add `use movr;` to step 5 and fix nit in step 9.


jseldess:
@dbist, before merging, please apply these updates to the 20.1 version as well.

PR #6124, opened 2019-12-10
cmd to delete cockroachdb-client-secure is missing

dbist:
Instructions mention an option to delete client pod but if one chooses to skip the step, there's an additional pod lingering in the clean up step. This adds an explicit step to delete the pod.


jseldess:
Thanks, @dbist! @taroface, LGTM, but can you please check this? If all's well, we'll need to apply this change to 20.1.

taroface:
LGTM

dbist:
thanks @jseldess added v20.1 changes, keep forgetting to include them when I use contribute and modify page options on the site.

PR #6130, opened 2019-12-10
Correct 19.2 auto-gen docs

### LINKED ISSUE #6108: 19.2 and stable documentation has documentation from master
Related: https://github.com/cockroachdb/cockroach/issues/42936

`isodow` was only added to master in https://github.com/cockroachdb/cockroach/pull/41784. It should not be available in 19.2 (the proof is in the backport https://github.com/cockroachdb/cockroach/pull/42957). However, it is in the documentation for 19.2 and stable:
* https://www.cockroachlabs.com/docs/v19.2/functions-and-operators.html#date-and-time-functions
* https://www.cockroachlabs.com/docs/stable/functions-and-operators.html#date-and-time-functions
* (correctly) not in 19.1: https://www.cockroachlabs.com/docs/v19.1/functions-and-operators.html#date-and-time-functions

Why is this showing up in the 19.2 and master documentation? Can we remove this?

jseldess:
Fixes #6108.

### DIFF
--- a/_includes/v19.2/sql/aggregates.md
+++ b/_includes/v19.2/sql/aggregates.md
@@ -1,160 +1,304 @@
 <table>
 <thead><tr><th>Function &rarr; Returns</th><th>Description</th></tr></thead>
 <tbody>
-<tr><td><a name="array_agg"></a><code>array_agg(arg1: <a href="bool.html">bool</a>) &rarr; <a href="bool.html">bool</a>[]</code></td><td><span class="funcdesc"><p>Aggregates the selected values into an array.</p>
+<tr><td><code>cume_dist() &rarr; <a href="float.html">float</a></code></td><td><span class="funcdesc"><p>Calculates the relative rank of the current row: (number of rows preceding or peer with current row) / (total rows).</p>
 </span></td></tr>
-<tr><td><a name="array_agg"></a><code>array_agg(arg1: <a href="bytes.html">bytes</a>) &rarr; <a href="bytes.html">bytes</a>[]</code></td><td><span class="funcdesc"><p>Aggregates the selected values into an array.</p>
+<tr><td><code>dense_rank() &rarr; <a href="int.html">int</a></code></td><td><span class="funcdesc"><p>Calculates the rank of the current row without gaps; this function counts peer groups.</p>
 </span></td></tr>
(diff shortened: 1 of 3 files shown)


_includes/v19.2/sql/aggregates.md
```diff
@@ -1,160 +1,304 @@
 <table>
 <thead><tr><th>Function &rarr; Returns</th><th>Description</th></tr></thead>
 <tbody>
-<tr><td><a name="array_agg"></a><code>array_agg(arg1: <a href="bool.html">bool</a>) &rarr; <a href="bool.html">bool</a>[]</code></td><td><span class="funcdesc"><p>Aggregates the selected values into an array.</p>
+<tr><td><code>cume_dist() &rarr; <a href="float.html">float</a></code></td><td><span class="funcdesc"><p>Calculates the relative rank of the current row: (number of rows preceding or peer with current row) / (total rows).</p>
 </span></td></tr>
-<tr><td><a name="array_agg"></a><code>array_agg(arg1: <a href="bytes.html">bytes</a>) &rarr; <a href="bytes.html">bytes</a>[]</code></td><td><span class="funcdesc"><p>Aggregates the selected values into an array.</p>
```
otan: why are there so many blank <a> tags in 20.1 :o

jseldess: You mean the empty `[]`s? I think that's just indicating that the argument/result can be an array of values...

otan: Ooh Nah I meant the <a name ="func" ></a> stuff. Not a biggie. Apparently we should be using <a id> though.

jseldess: Oh, right. I'm not sure what those are about. I suspect they're their so we have anchors to link to for specific functions? @ericharmeling, @rmloveland, do you know?

rmloveland: Correct - those are there so we can link directly to e.g., `array_agg` rather than make users search for the function on the [Functions and Operators](https://www.cockroachlabs.com/docs/dev/functions-and-operators.html) page.

nvanbenschoten:
@jseldess I believe this PR broke `_includes/v19.2/sql/aggregates.md`. It now lists window functions instead of aggregate functions. `docgen` still seems to work correctly, so something must have gone wrong in translation.

For my own understanding, what is the process for updating these pages? Are you manually running `docgen` and copying over the resulting markdown files?

jseldess:
@nvanbenschoten, oh, thanks for catching this. The process is simpler. We just copy the contents of the relevant file from the `cockroach` repo (in this case, from the release-19.2 branch) to the relevant file in the `docs` repo. I must've copied something wrong. I can fix in the morning.

PR #6132, opened 2019-12-10
Clarify go compiler version requirements for all releases.

### LINKED ISSUE #6097: Be more prescriptive about Go versions in build-from-source
Re: [Install CockroachDB](https://cockroachlabs.com/docs//v2.0/install-cockroachdb.html)

## Issue Description

CockroachDB 2.0 does not work correctly when built with Go 1.11, so the 2.0 build-from source instructions should specify that Go 1.10.x must be used, not "1.10 or newer". (cockroachdb/cockroach#42913)

For newer versions, I think we should be more prescriptive about the versions used: Each release of cockroachdb has a corresponding version of Go that is used for all official builds. Building from source with a newer version of go is risky and production users should prefer to use the tested version.

| CockroachDB version | Go version |
| --- | --- |
| 2.0 | 1.10 |
| 2.1 | 1.10 |
| 19.1 | 1.11 |
| 19.2 | 1.12 |

andreimatei:
Fixes #6097

### DIFF
--- a/v19.1/install-cockroachdb-linux.html
+++ b/v19.1/install-cockroachdb-linux.html
@@ -130,7 +130,10 @@ key: install-cockroachdb.html
         </tr>
         <tr>
           <td>Go</td>
-          <td>Version 1.11.6 or higher is required.</td>
+          <td>Version 1.11.6+. Older go versions might work by doing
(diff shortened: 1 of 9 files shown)


jseldess:
Will merge and address nits in follow-up.

Issue #6123, opened 2019-12-10
Update 20.1 roadmap issues with new details from Airtable


jseldess:
We created [this script](https://github.com/cockroachdb/docs/blob/master/scripts/issues-from-roadmap.py) to auto-generate GitHub issues for 20.1 roadmap records in Airtable, using the GitHub and Airtable APIs. 

In the meantime, a valuable additional detail was added to the Airtable records that we need in our GitHub issues: "GitHub Tracking Issue". If populated, this field in the Airtable record links to the relevant GitHub issue in the `cockroach` repo. 

It'd be wonderful to adapt our script, or write a new script, to update all of [these GitHub issues](https://github.com/cockroachdb/docs/issues?q=is%3Aopen+is%3Aissue+milestone%3A20.1+label%3AC-roadmap) with the "GitHub Tracking Issue" in the corresponding Airtable record, if it exists. 

Here's an [API reference for the relevant Airtable view](https://airtable.com/apppcLIR8IFy1QDcA/api/docs#curl/table:epics:fields).

otan:
Resolved by https://github.com/cockroachdb/docs/pull/6278!

PR #6138, opened 2019-12-10
scripts: update airtable entry on issue create/update

otan:
Resolves: #6123

This works by searching a github issue based on title, and updating if necessary.

I haven't run this on every record yet, and there's rate limiting from github which rate limits us to 20 searches per minute, but otherwise it seems to work for the one record I tried.

Also added `--dry_run` mode so that it doesn't make records but does all the processing we want.


otan:
See https://github.com/cockroachdb/docs/pull/6156

PR #6145, opened 2019-12-10
Clarify known limitation on schema changes within transactions

### LINKED ISSUE #4905: Clarify known limitation on schema changes within transactions
This [known limitation](https://www.cockroachlabs.com/docs/stable/online-schema-changes.html#no-schema-changes-within-transactions) states that arranging DML statements so they come after DDL statements should work, but it contradicts a few of [these examples](https://www.cockroachlabs.com/docs/stable/online-schema-changes.html#examples-of-statements-that-fail). And @joshimhoff confirmed that you can't, for example, add a column and then update its value within a single transaction.

rafiss:
fixes #4905


jseldess:
Re-running the preview deploy in Netlify. Not sure why it failed.

PR #6155, opened 2019-12-13
Add docker images to releases list

### LINKED ISSUE #5792: release: add docker image to release table
In the [main docs page on releases](https://www.cockroachlabs.com/docs/releases/#testing-releases) users cannot tell that we have docker images available. At least one user didn't want to use a beta because they thought we didn't offer docker images for our alphas/betas.
image

This information is available on individual [release pages](https://www.cockroachlabs.com/docs/releases/v19.2.0-rc.3.html):
image

We should add a docker image link to the main page. 

@bdarnell also notes that our windows binaries have fallen off the main page. 

There are also some ideas in [this thread](https://cockroachlabs.slack.com/archives/C4A9ALLRL/p1573000844273400) about how to avoid cluttering this main page.

jseldess:
Fixes #5792.

Haven't added Kubernetes or helm to the release page yet, mostly because fontawesome doesn't provide icons.


bdarnell:
This makes the page really wide (there's horizontal scrolling in the desktop layout unless the browser is at least 1400 pixels wide, and there's awkward wrapping at anything less than 1600), and it's even worse when you get down to the unstable releases. I don't think adding the docker image name here is the right answer, and I don't think the problem is urgent enough to need a quick fix like this.

I think if we want a quick fix we should consider trimming this table down to just the date and link to the per-release page, and then we have all the room we need for per-platform downloads, docker/helm instructions, etc. Or with a little more work, I'd put a platform-selector at the top of the page, and then you'd see the links for your chosen platform in each row of the table.

jseldess:
Great feedback, @bdarnell. I added "platform" filters to simplify the page. Still need to figure out why the Docker icon isn't rendering. I think we might just need a newer version of the font awesome css file. In any case, let me know if this is closer to what you're envisioning.

awoods187:
LGTM

jseldess:
@j-low, I've not been able to figure out how to upgrade font awesome so as to get the Docker icon to show up:

Screen Shot 2019-12-16 at 10 09 27 PM

Would you mind spending a few mins on this to see if you can get it to work? Feel free to take over the PR to do so.

j-low:
@jseldess sure I can take a look either later today or tomorrow.  I'll push updates to this PR.

jseldess:
TFTR, @j-low.

Issue #5953, opened 2019-11-26
Enable SQL Savepoints


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/recxI0iMnCxdfpS2h

Description: This feature is KV supporting SQL's adding Savepoint syntax.

A Savepoint is a marker within a transaction that allows rolling back to a specific point in a transaction. At the moment, we use Savepoint keyword for rolling back only to the beginning of a transaction in a way that a client can begin again, helping users handle transaction retry errors.

The problem is CRDB's lack of compatibility with ORMs. For application developers depending on ORMs like Django, SQLAlchemy and Hibernate, this represents a significant barrier to adoption. More specifically, Savepoint support is required for passing third-party test suites.

Team: John Kendall, Andrew Kimball, Andrew Woods, Raphael 'kena' Poss, Andrei Matei, Bilal Akhtar

Github Tracking Issue: https://github.com/cockroachdb/cockroach/issues/10735 https://github.com/cockroachdb/cockroach/issues/16491 https://github.com/cockroachdb/cockroachdb-python/issues/25 https://github.com/cockroachdb/docs/issues/5953

rmloveland:
Estimated scope of work:

- Update SAVEPOINT docs to mention the support for nesting, add examples showing how
- Search for other mentions of the "no nested transactions" limitation and see if / how to update them
- Remove from known limitations page, if this is mentioned there

jseldess:
@rmloveland, there are a bunch of savepoint-related updates in the next beta release notes. Pasting them here rather than opening distinct issues.

```
- CockroachDB now collects separate sets of metrics for usage of [`SAVEPOINT`](../v20.1/savepoint.html): one set for regular SQL savepoints and one set for uses dedicated to CockroachDB's client-side transaction retry protocol. [#45566][#45566] {% comment %}doc{% endcomment %}
- SQL [savepoints](../v20.1/savepoint.html) are now supported. `SAVEPOINT <foo>`, `RELEASE SAVEPOINT <foo>`, and `ROLLBACK TO SAVEPOINT <foo>` now work. `SHOW SAVEPOINT STATUS` can be used to inspect the current stack of active savepoints. [#45566][#45566]
[...]

PR #6156, opened 2019-12-13
scripts: add ability to update issues, add Github Tracking Issue

### LINKED ISSUE #6123: Update 20.1 roadmap issues with new details from Airtable

otan:
Resolves #6123. Example: #5953.

This works by searching a github issue based on title, and updating if necessary. Run `--check_existing` if you wish to search issues before adding -- it is slower due to the rate limit so is off by default.

Also add the Github Tracking Issue from Airtable to Github.

I haven't run this on every record yet, and there's rate limiting from github which rate limits us to 30 searches per minute, but otherwise it seems to work for the one record I tried.

Also added `--dry_run` mode so that it doesn't make records but does all the processing we want.

Cleaned up some stuff too! Let me know if you have questions.

### DIFF
--- a/scripts/issues-from-roadmap.py
+++ b/scripts/issues-from-roadmap.py
@@ -5,6 +5,7 @@ Example use: python3 issues-from-roadmap.py --milestone=20.1
 
 Requirements:
   - Requests: http://docs.python-requests.org/en/master/
+  - ratelimiter: https://pypi.org/project/ratelimiter/
   - A GitHub personal access token.
   - An Airtable developer key.
 """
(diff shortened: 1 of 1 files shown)


scripts/issues-from-roadmap.py
```diff
@@ -13,6 +14,18 @@
...
+from ratelimiter import RateLimiter
+
+# Map Airtable product areas to GitHub labels and writers.
+epic_area_to_tag_and_writer = {
+    "KV": ("A-kv-storage", "rmloveland"),
+    "SQL": ("A-sql", "ericharmeling"),
+    "Observability": ("A-observability", "taroface"),
+    "IAM & Security": ("A-iam-security", "Amruta-Ranade"),
+    "Bulk I/O": ("A-io", "lnhsingh"),
+    "I/O": ("A-io", "lnhsingh"),
```
otan: (this seemed to have changed, so I put both in!)

Issue #4093, opened 2018-11-21
local htmltest runs no longer pass in a post-managed-docs world


rmloveland:
Sorry if there is an issue filed against this behavior already, a search for "is:open htmltest" didn't find it!

tl;dr: `htmltest` no longer passes when run locally, so we're flying blind until we push to Github and see the green check (I think?  Happy to be wrong about this, maybe I'm doing something wrong locally).

Relates to #3301, #3833.

Full version:

If you run `make standard` and your build output directory contains `_site/docs`, local runs of `htmltest -s` return the following:

    htmltest -s
    Skipping the checking of external links.
    htmltest started at 03:32:28 on _site
    ========================================================================
    running in concurrent mode, this is experimental
    docs/releases/v2.1.0.html
      target does not exist --- docs/releases/v2.1.0.html --> ../managed/v2.1/
      target does not exist --- docs/releases/v2.1.0.html --> ../managed/v2.1/managed-sign-up-for-a-cluster.html
    docs/v2.1/frequently-asked-questions.html
      target does not exist --- docs/v2.1/frequently-asked-questions.html --> ../managed/v2.1/
    docs/v2.2/frequently-asked-questions.html
      target does not exist --- docs/v2.2/frequently-asked-questions.html --> ../managed/v2.1/
    ========================================================================
    âââ failed in 2.196259095s
    4 errors in 1962 documents

If you then run `make managed` and your build output directory contains both `_site/docs` and `_site/managed`, local runs of `htmltest -s` return a giant list of errors that I will attach below.

[foo.txt](https://github.com/cockroachdb/docs/files/2605750/foo.txt)

rmloveland:
It occurs to me that we could get this case to pass (`make standard`) by adding some logic to create a few dummy pages in the `_site/managed` folder, but that's probably not the Right Thing (?).

otan:
took a brief look at this, what does this pass in on atm?
it does seem like genuine issues are being hit.

otan:
also i think the two related issues are unrelated and separate!

rmloveland:
> took a brief look at this, what does this pass in on atm?

@otan sorry, I don't understand exactly what you are asking.  Do you mean what args are getting passed to `htmltest` when it's running locally?

At least in my usage, it's getting run basically as follows:

```
$ cd /path/to/cockroachdb/docs
$ make standard-build # generates _site/docs directory with HTML files
$ htmltest -s
# broken link output here
```

otan:
[...]

PR #6157, opened 2019-12-13
config: make htmltest -s work locally

otan:
Potentially resolves https://github.com/cockroachdb/docs/issues/4093.

Make htmltest -s work locally by removing the /docs prefix.

Also add a make command for executing htmltest (`make htmltest`)

Not sure how to test if this breaks building locally though - please pull and see if it works!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,10 @@ cockroachcloud-build: bootstrap
 cockroachcloud: jekyll-action := serve --port 4001
 cockroachcloud: cockroachcloud-build
 
+.PHONY: htmltest
+htmltest: cockroachdb-build cockroachcloud-build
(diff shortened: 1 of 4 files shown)


rmloveland:
@mjibson and @jseldess - added y'all as reviewers because I do not understand the site build / deployment well enough to assess this.  

In particular there are some changes to the `baseurl` and how we do redirects in the netlify build that seem like they might break stuff?  But I don't know, hence this message.

mjibson:
The config and build files should absolutely not be changed. You can click on the `deploy/netlify` preview link in the build checks here to see that it does not work.

otan:
image
looks like alternate PR would better solve this problem :)

PR #6160, opened 2019-12-16
change htmltest DocumentConcurrencyLimit to 16

### LINKED ISSUE #3833: Running `make test` locally kills my mac
I ran `make test` to check links locally. I got a panic and pretty much all applications on my machine froze. I force quit everything, but still have to force reboot. 

@mjibson thinks it may be because `hmtltest` opens up too many files at once.

otan:
Resolves #3833.

The current default of 128 has a tendency to reach "too many files at
once" when attempting to `os.Stat` on local setups. I don't think the
speedup is too important, so setting to a lower setting of 16.

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -8,3 +8,4 @@ IgnoreURLs:
   - http://localhost:4000
 IgnoreInternalEmptyHash: true
 TestFilesConcurrently: true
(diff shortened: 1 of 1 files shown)


mjibson:
While this LGTM, htmltest changes should be reviewed by the docs team.

PR #6170, opened 2019-12-17
Use cli command name for files and h1s

### LINKED ISSUE #6168: Open a SQL Shell to a Temporary Cluster Doc Update
Re: [Open a SQL Shell to a Temporary Cluster](https://cockroachlabs.com/docs//v19.2/cockroach-demo.html)

## Issue Description
The H1 on this page makes no mention of Cockroach Demo, which is what the whole page is about. It's easy enough to from Google Search, but nearly impossible to identify using search from within docs.

## Suggested Resolution 
Could "Cockroach Demo" be added to the H1?

jseldess:
These pages are reference docs, though the filenames and h1s
make them seem like task-oriented topics.

Fixes #6168.


rmloveland:
I would be *personally* cool with having these different page names (which are way better for the record!), however with my work hat on I tend to lean against it for a few reasons (listed in order of importance):

- Isn't this going to break many years' worth of emailed links, bookmarks, browser histories, etc.?  If we set up 302 redirects for all of the old renamed pages that could work (but then we'd have like 30 markdown files hanging around that were just redirects, so 🤷‍♂ )

- I think this is worth discussing with SEO magicians since if we break links it could also hurt our search rankings.  (I honestly don't understand this area well tho)

- Is this also going to break things our site search?  I guess we can regenerate the indexes or something in that case.

jseldess:
Thanks, @rmloveland. I checked with @jess-edwards on this, and she was ok with the change, but I'll chat with @jeffreyroach as well. 

I put redirects in place already, so this shouldn't result in broken links. 

For site search, yes, we can ask them to regenerate the indexes pretty easily.

jseldess:
@rmloveland, for the redirects, this doesn't result in a bunch of empty markdown files. I've used the `redirect-from` front-matter field. I think you're thinking of `redirect-to`.

rmloveland:
Great - thanks for answering my questions @jseldess!  In that case I am definitely way into this change, thanks for doing it. :-)

jseldess:
Synced with Jeffrey. He's good with these changes, too. I'll fix the conflict and port the changes to 20.1.

PR #6228, opened 2019-12-19
Correct Step 3 --store grammar error.

florence-crl:
removed "the"


florence-crl:
Found grammar error while doing tutorial for onboarding.

PR #6235, opened 2019-12-20
Update redirects to use correct YAML var name

rmloveland:
According to the documentation for the Jekyll redirect
plugin (https://github.com/jekyll/jekyll-redirect-from) we are using,
the variable name to use in the YAML header is

  `redirect_from`

Prior to this change, we had used the variable name

  `redirect-from`

which is apparently not understood by the Jekyll plugin.

This was causing our redirects to break.

Note: Unfortunately, this redirection plugin does not work for anchors
to specific locations on the page, so that visiting e.g.,

https://www.cockroachlabs.com/docs/stable/create-security-certificates.html#create-the-certificate-and-key-pair-for-a-client

redirects to the top-level page URL

https://www.cockroachlabs.com/docs/v19.2/cockroach-cert.html

which means the user is not directed to the specific location implied by
the anchor.

### DIFF
--- a/v19.2/cockroach-cert.md
+++ b/v19.2/cockroach-cert.md
@@ -2,7 +2,7 @@
 title: cockroach cert
 summary: A secure CockroachDB cluster uses TLS for encrypted inter-node and client-node communication.
 toc: true
-redirect-from: create-security-certificates.html
+redirect_from: create-security-certificates.html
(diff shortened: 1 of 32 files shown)


taroface:
LGTM!

jseldess:
Thanks for catching and fixing this, @rmloveland!

PR #6322, opened 2020-01-10
Update cluster troubleshooting with node liveness

### LINKED ISSUE #4475: Explain node liveness
Explain the causes of node liveness, how to diagnose that its the problem, and what to do as a result.

rmloveland:
Fixes #4475.

Summary of changes:

- Add new 'Node liveness' section to the 'Troubleshoot cluster setup'
  page, including:

  - What it is

  - How the system checks for it

  - Causes of common problems with node liveness (overloaded disk /
    busted network connectivity)

### DIFF
--- a/v19.2/cluster-setup-troubleshooting.md
+++ b/v19.2/cluster-setup-troubleshooting.md
@@ -57,7 +57,7 @@ Other services may be running on port 26257 or 8080 (CockroachDB's default `--li
 
   If you change the port, you will need to include the `--port=<specified port>` flag in each subsequent cockroach command or change the `COCKROACH_PORT` environment variable.
 
-### Networking issues
+### Single-node networking issues
 
 Networking issues might prevent the node from communicating with itself on its hostname. You can control the hostname CockroachDB uses with the [`--listen-addr` flag](cockroach-start.html#networking).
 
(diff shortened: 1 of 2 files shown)


rmloveland:
Hey @ajwerner figured you might be a good candidate based on your recent comment on https://github.com/cockroachdb/cockroach/issues/19699 and linked PR.

My main questions/comments for you are:

1. Does this change miss listing any other common reasons why node liveness can get behind?

2. Is your change in https://github.com/cockroachdb/cockroach/pull/39172 going to be backported to  anything <19.2 ?

3. Related to (2), I'm planning to copy this change to 19.2 docs once it gets through review.  I will also copy to 19.1 etc. depending on your answer to (2).

4. Should we say anything more specific re: what "maxing" out disk throughput is that would generally cause this?  E.g., should we say "try to keep IO below X% of disk max throughput" (etc)?

5. Is there anything else obvious that seems missing / wrong in this PR to you?  That I might not even know to ask about?

rmloveland:
Additional question:

- Looking at one of Andrei's recent liveness PRs, it seems like users should care about `server.time_until_store_dead`.  Should we add a link to that setting here with a few words?

rmloveland:
Thanks @ajwerner - based on your answers to questions 4 and 5 I'll work on updating the PR to point the user to the Admin UI dashboards and remove the reference to using a SQL query, since it's probably not going to be that helpful (if I'm understanding what you wrote correctly) to provide them with data from `crdb_internal.gossip_liveness` and expect them to duplicate the calculations manually that our code is running to generate the stuff that's already in the UI.

> The last update will also show up in /#/reports/nodes in the admin UI.

Cool, I'll add that as well.

rmloveland:
@ajwerner the latest change updates the docs to point users to the pages in the Admin UI you mentioned for command commit latency, etc.  In each new section I added a line "Expected values for a healthy cluster: XXX".  Can you advise about what those values should be in each case?  Approximations are OK, I think, since we weren't providing anything at all before.

rmloveland:
@ajwerner friendly ping on this, particularly

>  Can you advise about what those values should be in each case? Approximations are OK, I think, since we weren't providing anything at all before.

If determining those expected values would be a lot of work / thinking that you don't have bandwidth for right now, I can drop the **Expected values** bits.  I think even without them, this change will be an improvement, since before we had ~nothing.

Issue #6241, opened 2019-12-24
Incorrect description of behaviour with clock skews greater than threshold


irfansharif:
Looking at https://www.cockroachlabs.com/docs/dev/operational-faqs.html#what-happens-when-node-clocks-are-not-properly-synchronized, we have the following:

> For this reason, when a node detects that its clock is out of sync with at least half of the other nodes in the cluster by 80% of the maximum offset allowed (500ms by default), it spontaneously shuts down. This avoids the risk of violating serializable consistency and causing stale reads and write skews, but it's important to prevent clocks from drifting too far in the first place by running NTP or other clock synchronization software on each node.

Even with clock skew greater than the upperbound, we never violate serializable consistency. We may however violate single-key linearizability, which is what I believe "causing stale reads and write skews" alludes to. 

+cc @nvanbenschoten

irfansharif:
I'm also not sure if we _can_ have write skews. Thinking about it for 30s tells me no, we can't, given we don't violate serializability.

nvanbenschoten:
> I'm also not sure if we can have write skews. Thinking about it for 30s tells me no, we can't, given we don't violate serializability.

That's exactly correct.

jseldess:
@irfansharif, would you mind taking a swing at a quick rewrite of the paragraph above?

irfansharif:
@jseldess: #6329.

PR #6329, opened 2020-01-13
Fix description for behaviour under clock skew greater than thr…

irfansharif:
Looking at cockroachlabs.com/docs/dev/operational-faqs.html we have the
following:

    > For this reason, when a node detects that its clock is out of sync
    with at least half of the other nodes in the cluster by 80% of the
    maximum offset allowed (500ms by default), it spontaneously shuts
    down. This avoids the risk of violating serializable consistency and
    causing stale reads and write skews, but it's important to prevent
    clocks from drifting too far in the first place by running NTP or
    other clock synchronization software on each node.

Even with clock skew greater than the upperbound, we never violate
serializable consistency. We may however violate single-key
linearizability, which is what I believe "causing stale reads and write
skews" alludes to.

Fixes https://github.com/cockroachdb/docs/issues/6241.

---

I'm not sure how to check to see I'm not missing anything. I simply searched and replaced for "write skew".


rmloveland:
> I'm not sure how to check to see I'm not missing anything. I simply searched and replaced for "write skew".

This is where I would start also.  If there are other "long tail" usages we can fix them as we find them.  Thanks for this PR!

jseldess:
Thanks, @irfansharif!

PR #6335, opened 2020-01-14
Django doc

apantel:


_includes/v20.1/app/django-basic-sample/views.py
```diff
@@ -0,0 +1,99 @@
...
+        return HttpResponse("python/django", status=200)
+
+def retry_on_exception(num_retries=3, on_failure=HttpResponse(status=500), delay_=0.5, backoff_=1.5):
+    def retry(view):
+        def wrapper(*args, **kwargs):
+            delay = delay_
+            for i in range(num_retries):
+                try:
+                    return view(*args, **kwargs)
+                except IntegrityError as e:
```
rafiss: can we modify this so it only retries if the error's SQL state is 40001 (i.e. SerializationFailure)? that's what our hibernate example does.

_includes/v20.1/app/django-basic-sample/views.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+import time
+
+from .models import *
+
+class PingView(View):
+    def get(self, request, *args, **kwargs):
+        return HttpResponse("python/django", status=200)
+
+def retry_on_exception(num_retries=3, on_failure=HttpResponse(status=500), delay_=0.5, backoff_=1.5):
```
rafiss: maybe we could include a comment or pydoc to explain this a bit: both why we need it, and also document the function parameters/units.

PR #6342, opened 2020-01-15
add note about ndjson and document issue 43784

dbist:


jseldess:
Thanks, @dbist. The second bullet is clearly a limitation. Thank you for calling it out. But is the first bullet a limitation or just something we need to document better?

Also, would you please open a quick docs issue with a quick description of the problems you're fixing and where we learned about them?

dbist:
@jseldess filed [6350](https://github.com/cockroachdb/docs/issues/6350)

PR #6343, opened 2020-01-15
Follow-up to --max-sql-memory default change

jseldess:
Add "changed in v20.1" callout, and tweak language in a few places. 

Separately, we still need to test various local cluster tutorials to make sure this changes doesn't cause OOM errors.


taroface:
LGTM, thanks for sharing these updates with me @jseldess.

PR #6359, opened 2020-01-23
Django ORM "Build an app" tutorial + Code

ericharmeling:


ericharmeling:
@Amruta-Ranade 

The Django ORM support announcement has been pushed out a week already, due to lack of docs. I'm going to merge this PR this afternoon, to avoid blocking the announcement again. Hopefully you can get to it by then, but if not, we can always do a post-publish review.

Thanks! :)

jseldess:
@ericharmeling, can you please create a docs issue to add an insecure version? Also, can we port this to 19.2 and possibly 19.1, or are we dependent on 20.1 for some reason? @awoods187?

awoods187:
We can support 19.2 and 20.1. There are a number of things that don't work in 19.2 (due to bugs/missing features in CRDB that can't be backported) but they are well documented on the readme. I don't think we can go to 19.1.

jseldess:
OK, thanks, @awoods187. @ericharmeling, we should also create an issue to enhance the tutorial to call out the bugs/limitations in the readme. And we should probably open yet another issue to enhance it with any Django-specific best practices we know about (like we've done for [SQLAlchemy](https://www.cockroachlabs.com/docs/dev/build-a-python-app-with-cockroachdb-sqlalchemy.html#best-practices))

ericharmeling:
@jseldess see:
#6366 
#6367
#6368

PR #6386, opened 2020-01-28
update to get ticket for db user on client

keith-mcclellan:
Docs were mistakenly telling the user to get a ticket for the postgres user on the client rather than the db user.


jseldess:
Thanks, @keith-mcclellan. Can you please apply this change to the 20.1 version as well before merging?

PR #6394, opened 2020-01-29
Update enable-node-map.md

dbist:


dbist:
@jseldess do we also want to change the following line?

### Step 1. Ensure the CockroachDB Version is 2.0 or higher 
to 2.1 or higher?

jseldess:
@dbist, sure, can you make that change?

dbist:
@jseldess not quite sure I understand, I did make the change in my PR.

PR #6408, opened 2020-01-31
Fix SQL errors in Follower Reads topology page

### LINKED ISSUE #6383: "Follower Reads topology" syntax error + query error
The original issue fielded from a customer was that the query on [this](https://www.cockroachlabs.com/docs/stable/topology-follower-reads.html) page shows:
```sql
SELECT code FROM postal_codes
WHERE id = 5
AS OF SYSTEM TIME experimental_follower_read_timestamp();
```
It should read:
```sql
SELECT code FROM postal_codes
AS OF SYSTEM TIME experimental_follower_read_timestamp()
WHERE id = 5;
```

Additionally, this create table:
```sql
CREATE TABLE postal_codes (
    id INT PRIMARY KEY,
    code STRING,
);
```
Should not have a comma after `code STRING`.

Lastly, the query assumes we've inserted data into this table. However, it doesn't seem like we've inserted any data to support the experimental follower read  timestamp query.

rmloveland:
Fixes #6383.

### DIFF
--- a/v19.1/topology-follower-reads.md
+++ b/v19.1/topology-follower-reads.md
@@ -45,8 +45,15 @@ Assuming you have a [cluster deployed across three regions](#cluster-setup) and 
 ~~~ sql
 > CREATE TABLE postal_codes (
     id INT PRIMARY KEY,
-    code STRING,
+    code STRING
 );
+~~~
+
+Insert some data:
+
+{% include copy-clipboard.html %}
+~~~ sql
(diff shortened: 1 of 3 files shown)


rmloveland:
@mattcrdb this fixes the syntax errors.  I did not load any data because generating a data set, etc., would take quite a bit more effort right now.  But at least this fixes the SQL syntax errors (AFAICT).  PTAL.

rmloveland:
> we should add a simple insert of a few rows, including one with the id referenced in the select, so copy/pasting these steps actually works.

@jseldess just added an `INSERT` to make the selection queries work.  PTAL.

rmloveland:
@jseldess ok to merge this?  I'm fairly confident in the change, manually tested it the other day.

rmloveland:
Thanks Jesse!

PR #6421, opened 2020-02-04
update GCP load balancer restrictions

### LINKED ISSUE #6389: GCP internal load balancer limitations updates
In [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes-multi-cluster.html#exposing-dns-servers), we mention that Google's internal load balancers do not support clients in one region using a load balancer in another region.

As of [December 2019](https://cloud.google.com/blog/products/networking/new-features-for-l4-internal-load-balancer), this is no longer true. GCP internal load balancers now support [global access](https://cloud.google.com/load-balancing/docs/internal/setting-up-internal#ilb-global-access), meaning clients in one region can connect to an internal load balancer in another region. We should update our multiregion docs to note this, and provide an example of how to enable this on these load balancers.

cc @taroface

taroface:
Fixes #6389. 

In the interest of expediency, I added 1) a link to the new GCP global access instructions and 2) a note that the tutorial will be revised to utilize global access (which I plan to split into a separate issue).

### DIFF
--- a/v19.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md
+++ b/v19.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md
@@ -47,9 +47,11 @@ Kubernetes 1.8 or higher is required.
 
 #### Exposing DNS servers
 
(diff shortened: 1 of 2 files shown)


taroface:
@rmloveland Let me know if you think this is too much of a band-aid solution. I just didn't want to hold up making the basic correction.

PR #6432, opened 2020-02-05
Fix docs about CDC correctness issue

### LINKED ISSUE #6430: Fix CDC docs about correctness issues
Related to #5025

Correctness issue is in 19.1, fixed in 19.2. Docs are currently switched/incorrect.

```
mvardi: Our docs indicate a violation of ordering guarantees on cloud storage sinks, can someone shed some light on why this happens? Question from a customer.
awerner: It has to do with how we named files and what happens when the changefeed restarts internally. It was fixed in 19.2.
```

lnhsingh:
Closes #6430.

### DIFF
--- a/_includes/v19.1/cdc/correctness-warning.md
+++ b/_includes/v19.1/cdc/correctness-warning.md
@@ -1,0 +1,7 @@
+{{site.data.alerts.callout_danger}}
+**This is an experimental feature.** The interface and output are subject to change.
+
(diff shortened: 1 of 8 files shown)


danhhz:
feel free to ignore the drive by review, but my 2c:
- we should mention that this is unlikely. it currently sounds like it happens all the time, but it's likely rare in practice
- this only affects resolved timestamps. if you're not using resolved timestamps, you're fine. i think it's worth calling that out explictly

> Maybe it should instead say at the end "This is fixed in v19.2 and beyond", but I'm not sure if we do that or not.
- I'm in favor of saying this, yes. it sounds much worse as a totally unfixed issue rather than something we can't backport

lnhsingh:
@ericharmeling - can you sign off here if you think the edits are sufficient? ty!

PR #6435, opened 2020-02-05
Recursive CTE docs

### LINKED ISSUE #5956: [SH] WITH RECURSIVE CTE (common table expression) support
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/rec06cPoJb6dHNk5a

Description: CockroachDB supports using RECURSIVE (WITH). This means that a query can refer to its own output. A very simple example is this query to sum the integers from 1 through 100:

WITH RECURSIVE t(n) AS (
    VALUES (1)
  UNION ALL
    SELECT n+1 FROM t WHERE n < 100
)
SELECT sum(n) FROM t;

The general form of a recursive WITH query is always a non-recursive term, partnered with a UNION ALL.  This is a recursive term that can contain a reference to the query's own output. CockroachDB will support both tree structured data and more general graphs as well. Note, CockroachDB will not support recursive CTEs with UNION.

CockroachDB does not support recursive CTEs. This can be confusing to developers because CRDB does support single and multi-use CTEs. In particular, recursive CTEs can reduce the burden on the developer when working with graph, hierarchical or tree-structured data. It can shorten the lines of code needed to express the same idea by a large amount. In addition, recursive CTEs are commonly used in third-party tooling. 

Team: Andrew Woods, Jordan Lewis, Radu Berinde, Justin Jaffray

Github Tracking Issue: https://github.com/cockroachdb/cockroach/issues/21085 https://github.com/cockroachlabs/support/issues/88 https://github.com/cockroachlabs/support/blob/master/dreamworks/with_recursive.md https://github.com/cockroachdb/cockroach/issues/21085  https://github.com/cockroachdb/docs/issues/5956

ericharmeling:
Recursive CTE docs...

Fixes #5956.

- Updated 20.1 `delete`, `insert`, `select`, `update`, `upsert`, and `with_clause` diagrams.
- Added new section and examples to 20.1 "Common Table Expressions" doc page.

**Note** 
I did not update the `stmt_block` diagram (the primary SQL grammar chart), due to an issue with the [Railroad Diagram Generator](https://bottlecaps.de/rr/ui) API. This will block this PR from being merged, as other diagrams reference this chart. See #6472 for more details.

### DIFF
--- a/v20.1/common-table-expressions.md
+++ b/v20.1/common-table-expressions.md
@@ -173,6 +173,121 @@ For example:
 
 In this single query, you define two CTE's and then reuse them in a table join.
 
+## Recursive common table expressions
+
+<span class="version-tag">New in v20.1:</span> CockroachDB supports [recursive common table expressions](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#Common_table_expression). Recursive common table expressions are common table expressions that contain subqueries that refer to their own output.
+
+Recursive CTE definitions take the following form:
+
+~~~ sql
+WITH RECURSIVE <cte name> (<columns>) AS (
+    <initial subquery>
+  UNION ALL
+    <recursive subquery>
+)
+<query>
+~~~
+
+To write a recursive CTE:
+
+1. Add the `RECURSIVE` keyword directly after the `WITH` operator in the CTE definition, and before the CTE name.
+1. Define an initial, non-recursive subquery. This subquery defines the initial values of the CTE.
+1. Add the `UNION ALL` keyword after the initial subquery.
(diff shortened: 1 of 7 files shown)


v20.1/common-table-expressions.md
```diff
@@ -173,6 +173,121 @@ For example:
...
+    <initial subquery>
+  UNION ALL
+    <recursive subquery>
+)
+<query>
+~~~
+
+To write a recursive CTE:
+
+1. Add the `RECURSIVE` keyword directly after the `WITH` operator in the CTE definition, and before the CTE name.
```
Amruta-Ranade: The numbering seems off here but renders okay.

ericharmeling:
TFTR @Amruta-Ranade !

PR #6490, opened 2020-02-06
Correct release note about cross joins

### LINKED ISSUE #6489: Incorrect release note about cross joins in explain output
From release notes:

> The output of EXPLAIN now shows joins where there are no equality columns as "cross" instead of "hash". The hash join can be very expensive and should be avoided. #43239

From @robert-s-lee: 

> don’t we mean that the cross join can be very expensive and should be avoided

jseldess:
Fixes #6489


RaduBerinde:
LGTM, thanks!

PR #6492, opened 2020-02-07
New section advisories with per-advisory pages

### LINKED ISSUE #6385: Need new section for product advisories
Discussed with @dbist @tim-o @jseldess previously.

**Advisories** are different from **release notes** because they summarize situations that users have to acknowledge and act upon even without upgrading.

These typically cover security disclosures, but not exclusively: it's also customary to report severe regressions (hopefully we'll have those only rarely) and known risks of data loss.

Sales and other people refer to advisories via an "advisory code" and it is expected that each advisory gets its own page with a URL containing its code, for easy access, linking and retrieval.

The list of all advisories published so far, sorted in descending chronological order, should also be easily accessible, visually close to the "new releases" section.

I (@knz) would recommend a new page "Advisories" linked just underneath "Release notes" in the last sections of the sidebar, and containing a table to the individual advisory pages.

In the future (when the org gets bigger and these things get managed by different folk) we could even envision an automated data pull from the external advisory database into that doc page. This is what Apple, Red Hat, IBM etc do.

knz:
Fixes #6385.

### DIFF
--- a/_data/advisories.yml
+++ b/_data/advisories.yml
@@ -1,0 +1,24 @@
+- advisory: 44299
+  summary: Schema changes may cause cluster unavailability
+  versions: 19.1.0-19.1.7, 19.2.0-19.2.3
+  date: Feb 12, 2020
+- advisory: 44348
+  summary: Data leak in statement details
+  versions: 2.1.0-2.1.11, 19.1.0-19.1.7, 19.2.0-19.2.3
+  date: Feb 12, 2020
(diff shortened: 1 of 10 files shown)


dbist:
@knz in regards to the individual advisories, it makes sense to add a published date when the customer communication went out, possibly linked back to the advisory index page too.

knz:
The index page already has it; I'll add it to the individual pages.

PR #6509, opened 2020-02-10
Update GC TTL info in Storage Layer docs

### LINKED ISSUE #6505: Storage layer garbage collection wrong gc.ttl
https://www.cockroachlabs.com/docs/stable/architecture/storage-layer.html#garbage-collection states: `the garbage collection period is 24 hours, but it can be set at the cluster, database, or table level through replication zones.`.

The default garbage collection period is 25 hours (90000 seconds).

rmloveland:
Fixes #6505.

### DIFF
--- a/v19.1/architecture/storage-layer.md
+++ b/v19.1/architecture/storage-layer.md
@@ -57,7 +57,7 @@ Using these tools, you can get consistent data from your database as far back as
 
 ### Garbage collection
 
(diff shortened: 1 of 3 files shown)


rmloveland:
@mattcrdb take a look - FWIW I removed the hardcoded value, so this page should never be wrong (about this) again.

mattcrdb:
Looks great!

Thank you

PR #6511, opened 2020-02-10
Edit CDC examples to show more than one table in a changefeed

### LINKED ISSUE #6109: CDC: recommend creating as few changefeeds as possible
After setting up changefeeds on our telemetry cluster out to S3, I noticed that some of the changefeeds had failed. Upon inspecting the cluster, we noticed errors on the changefeeds that indicated the changefeed process needed more memory than was available to the cluster. @danhhz pointed out that setting up separate changefeed processes for each table would consume more memory than setting up a single changefeed with a comma separated list of tables. This is not currently obvious in the documentation - I would expect these two methods to be equivalent.

We should add guidance that tells users to try to create as few changefeeds as possible to avoid this issue. cc @rolandcrosby and @lnhsingh

lnhsingh:
Closes #6109

### DIFF
--- a/_includes/v19.2/known-limitations/cdc.md
+++ b/_includes/v19.2/known-limitations/cdc.md
@@ -1,6 +1,7 @@
 The following are limitations in the v19.2 release and will be addressed in the future:
 
 - Changefeeds only work on tables with a single [column family](column-families.html) (which is the default for new tables).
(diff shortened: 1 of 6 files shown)


lnhsingh:
cc @mwang1026

PR #6513, opened 2020-02-10
Online primary key changes

### LINKED ISSUE #5955: [MH] Online primary key changes
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/reckXe8OwbdVi8WzH

Description: In CockroachDB, a primary key is essentially a special index. This feature allows developers to designate a different unique index as the  primary index. This will not require a bulk re-write of all data but it will require an update to any existing secondary indexes. This change will occur online and require zero downtime. 

Primary keys cannot be changed within CRDB. Instead, users must reimport there data via CSV/bulk import and setup a new schema or use CREATE TABLE AS. This is particularly problematic for geo-partitioned tables and interleaved tables but it is basic SQL functionality expected from any database.

Team: Andrew Woods, Roland Crosby, Jordan Lewis, Lucy Zhang

Github Tracking Issue: https://github.com/cockroachdb/cockroach/issues/19141 https://github.com/cockroachdb/docs/issues/2991  https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20180413_alter_primary_key.md https://github.com/cockroachdb/docs/issues/5955

ericharmeling:
Fixes #5955.


v20.1/alter-primary-key.md
```diff
@@ -0,0 +1,150 @@
...
+You can add a column and change the primary key with a couple `ALTER TABLE` statements:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> ALTER TABLE users ADD COLUMN id UUID DEFAULT gen_random_uuid();
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> ALTER TABLE users ALTER PRIMARY KEY USING COLUMN (id);
```
rohany: /s/COLUMN/COLUMNS/

v20.1/alter-primary-key.md
```diff
@@ -0,0 +1,150 @@
...
+~~~
+
+If you still plan to filter queries on the `name` column, you can add an index on the column to improve performance:
```
rohany: unless you drop the old primary key (upcoming work), the primary key change is going to basically demote the old primary key into a secondary index, so there will already be an index on name post the primary key change.

v20.1/alter-primary-key.md
```diff
@@ -0,0 +1,150 @@
...
+{% include copy-clipboard.html %}
+~~~ sql
+> SHOW CREATE TABLE users;
+~~~
+
+~~~
+  table_name |                      create_statement
+-------------+-------------------------------------------------------------
+  users      | CREATE TABLE users (
+             |     id UUID NOT NULL DEFAULT gen_random_uuid(),
```
rohany: there would still be an index on id here

rohany:
I'd be more explicit about the fact that the old primary key becomes a secondary index after the primary key change, rather than just mentioning that it is there after code sample of the primary key change executing.

Not sure where to put this, but I think we should mention somewhere that this is an expensive operation that performs a number of rewrites of indexes.

v20.1/alter-primary-key.md
```diff
@@ -0,0 +1,157 @@
...
+ <span class="version-tag">New in v20.1:</span> The `ALTER PRIMARY KEY` [statement](sql-statements.html) is a subcommand of [`ALTER TABLE`](alter-table.html) that can be used to change the [primary key](primary-key.html) of a table.
+
+{{site.data.alerts.callout_danger}}
+This is currently an experimental feature. The interface and output are subject to change.
+
+To enable online primary key changes, set the `experimental_enable_primary_key_changes` [session variable](experimental-features.html#session-variables) to `true`.
+{{site.data.alerts.end}}
+
+{{site.data.alerts.callout_info}}
+When you change a primary key with `ALTER PRIMARY KEY`, the old primary key index becomes a secondary index. This helps optimize the performance of queries that still filter on the old primary key column, but it makes `ALTER PRIMARY KEY` an expensive operation, as it performs rewrites of multiple indexes.
```
rohany: The rewrite of the old primary key into a secondary index isn't the reason why alter primary key is expensive. alter primary key is expensive just because it has to rewrite a new index + multiple indexes other than the old primary key being rewritten. Do we have a precedent of linking to rfc's/internal docs for discussion of more detailed information? If so, we could show the rules for which indexes in the table get rewritten.

rohany:
LGTM. Yeah it doesn't seem like we need to go into that much detail about how and why indexes get rewritten.

ericharmeling:
@Amruta-Ranade Ping on review :)

ericharmeling:
> Sorry for the late review!

No problem!

PR #6572, opened 2020-02-19
CDC: experimental_avro + envelope=key_only + updated is now rejected

### LINKED ISSUE #6182: ccl/changefeedccl: format=experimental_avro, envelope=key_only, and updated is now rejected
PR: https://github.com/cockroachdb/cockroach/pull/41793

From release notes:
> The combination of the [`CHANGEFEED`](../v20.1/change-data-capture.html) options `format=experimental_avro`, `envelope=key_only`, and `updated` is now rejected. This is because the use of `key_only` prevents any rows with updated fields from being emitted, which renders the `updated` option meaningless. [#41793][#41793] {% comment %}doc{% endcomment %}

lnhsingh:
Changes:
- Added note to `CREATE CHANGEFEED` > Options section

Closes #6182.

### DIFF
--- a/v20.1/create-changefeed.md
+++ b/v20.1/create-changefeed.md
@@ -99,7 +99,7 @@ Option | Value | Description
 `key_in_value` | N/A | Makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for these sinks (currently only [cloud storage sinks](#cloud-storage-sink)).
 
 {{site.data.alerts.callout_warning}}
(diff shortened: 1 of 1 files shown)


nvanbenschoten:
> Do you have a preference since I'll be pinging you to review? If it's easier, I can push the other changes to this PR and lyk when the whole thing is ready for review.

I think I'd prefer that, but I'd go with whatever is easiest for you. I'll also be out for the next few days, so the review might be a little delayed.

lnhsingh:
> I think I'd prefer that, but I'd go with whatever is easiest for you. I'll also be out for the next few days, so the review might be a little delayed.

Ok, will do! I'll add to this PR for you to review once you're back. Thanks!

lnhsingh:
Merging. Will open another PR for the remaining `WITH diff` changes.

PR #6573, opened 2020-02-19
Re-structure "Tutorials" section of sidenav

jseldess:
Categorize tutorials as either primary about
"Database Operations" or "App Development".


jseldess:
@ericharmeling, interested in your thoughts here. We'll likely restructure the sidenav more dramatically in the coming quarters, but this gives us some level of categorization in the meantime.

Issue #6571, opened 2020-02-19
Document new approach to locking and queueing in the key value layer


rmloveland:
We need to update the architecture documentation to discuss the new approach to locking and queuing in the key-value layer.

AFAICT this new approach is enabled by the new concurrency control package/lock table referred to by [this PR][wip_sfu] (and [this one][sumeer_locktable], and probably others).

Estimated scope of work:
  
- [ ] Update the architecture docs (probably the [storage layer][storage layer]) to describe the new concurrency manager / how locking and queuing works

- [ ] Review other areas of the architecture docs to ensure they reference (or do not reference) this new layer, as appropriate

[wip_sfu]: https://github.com/cockroachdb/cockroach/pull/43775
[sumeer_locktable]: https://github.com/cockroachdb/cockroach/pull/43740
[storage layer]: https://www.cockroachlabs.com/docs/dev/architecture/storage-layer.html

jseldess:
Thanks for being proactive here, @rmloveland. Is this not reflected on the KV product roadmap? I'm surprised we don't already have a docs issue for it.

rmloveland:
> Is this not reflected on the KV product roadmap? 

I did some searching in AT but was not able to find anything beyond the SQL-level description of the work.

Issue #2836, opened 2018-04-02
Testing: Add automated test to detect git merge conflict markers


rmloveland:
We should create an automated test to prevent PRs from being merged which contain git merge conflict markers, e.g.,

```
<<<<<<< HEAD

... stuff

=======

... stuff

>>>>>>> abc123... 
```

Relates to #2832

jseldess:
We have closed this issue because it is more than 3 years
old. If this issue is still relevant, please add a comment and re-open
the issue. Thank you for your contribution to CockroachDB docs!

rmloveland:
Def still valid AFAICT, see also #3301

PR #6671, opened 2020-02-24
Document the 'SELECT ... FOR UPDATE' statement

### LINKED ISSUE #6000: Enable “SELECT FOR UPDATE” Statement
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/recnmogESIC5147Jf

Description: This feature is implementing SELECT FOR UPDATE statement, making it usable by application developers. At the moment, SELECT FOR UPDATE explicitly does not do anything when executed. In its implemented state, SELECT FOR UPDATE provides one exclusive access when executing an UDPATE statement.

The important problem is retry errors because our customers consistently express it as a pain point. As of September 2019, 4.5% of our customers’ long-running clusters experience transaction retry errors. Application developers are accustomed to other databases, such as Oracle, that cause less cases of contention that are the root cause of these errors.

As a result, their applications are not setup to handle the additional retry errors that are inherent to our design. Customers have the burden of added application logic to handle this case. While this expected behavior and best practices for handling is well-documented (https://www.cockroachlabs.com/docs/stable/advanced-client-side-transaction-retries.html
), it’s still a pain point for customers because this type of change feels invasive.

Team: John Kendall, Nathan VanBenschoten

Github Tracking Issue: https://github.com/cockroachdb/docs/issues/6000

rmloveland:
Fixes #6000.

Summary of changes:

- Add docs for the `SELECT ... FOR UPDATE` statement, including an
  example showing its intended use.

- Update other pages re: selection queries to mention the use of `SELECT
  FOR UPDATE` to reduce contention/retries and improve
  throughput/latency.

- Update Transactions page and Perf Best Practices page to mention
  `SELECT FOR UPDATE` as a possible solution to contention/retry issues.

- Add `SELECT FOR UPDATE` to SQL Statements and SQL Feature Support
  pages.

- Add docs for new `enable_implicit_select_for_update` session and
  cluster setting.


rmloveland:
Related issue to document KV layer changes: #6571

rmloveland:
Nathan this PR is meant to document the user-facing changes related to `SELECT FOR UPDATE`.  We will document the KV-layer changes separately via #6571.

Please take a look at your convenience and let me know if this is missing any pertinent info that SQL users should have re: SFU, or if anything is incorrect.  Also, if you have suggestions re: how to better explain this feature, I'm happy to update.

rmloveland:
Also, thanks for the review Nathan!  (Since I forgot to say that :-} )

rmloveland:
> This is rendering with what looks like merge conflict messages - <<<<<<< HEAD and ====== and >>>>>>> Recursive CTE docs. I think there's also two diagrams showing when there should be one.

This IS there, my bad.  I went and looked at the include.  It was introduced in a different PR :-/

I will make the fix in a separate PR since I want to drive some discussion of one of my pet issues #2836

PR #6675, opened 2020-02-25
Document nested SAVEPOINTs

### LINKED ISSUE #5953: Enable SQL Savepoints

rmloveland:
Fixes #5953.

Summary of changes:

- Update SAVEPOINT page with more examples of "general" savepoints, in addition to the existing "retry" savepoints

- Add a new SHOW SAVEPOINT STATUS statement page, and add it to the sidebar

- Update RELEASE SAVEPOINT docs to clarify "retry" savepoints vs. "general", and update the example

### DIFF
--- a/_includes/sidebar-data-v20.1.json
+++ b/_includes/sidebar-data-v20.1.json
@@ -1351,6 +1351,12 @@
                 ]
               },
               {
+                "title": "<code>SHOW SAVEPOINT STATUS</code>",
+                "urls": [
+                  "/${VERSION}/show-savepoint-status.html"
+                ]
+              },
+              {
                 "title": "<code>SHOW TABLES</code>",
                 "urls": [
                   "/${VERSION}/show-tables.html"
--- a/_includes/v20.1/misc/customizing-the-savepoint-name.md
+++ b/_includes/v20.1/misc/customizing-the-savepoint-name.md
@@ -1,7 +1,5 @@
- Set the `force_savepoint_restart` [session variable](set-vars.html#supported-variables) to `true` to enable using a custom name for the restart savepoint (for example, because you are using an ORM that wants to use its own names for savepoints).
+Set the `force_savepoint_restart` [session variable](set-vars.html#supported-variables) to `true` to enable using a custom name for the [retry savepoint](advanced-client-side-transaction-retries.html#retry-savepoints).
 
-Once this variable is set, the [`SAVEPOINT`](savepoint.html) statement will accept any name for the savepoint, not just `cockroach_restart`. This allows compatibility with existing code that uses a single savepoint per transaction as long as that savepoint occurs before any statements that access data stored in non-virtual tables.
+Once this variable is set, the [`SAVEPOINT`](savepoint.html) statement will accept any name for the retry savepoint, not just `cockroach_restart`. In addition, it causes every savepoint name to be equivalent to `cockroach_restart`, therefore disallowing the use of [nested transactions](transactions.html#nested-transactions).
 
-{{site.data.alerts.callout_danger}}
-The `force_savepoint_restart` variable changes the semantics of CockroachDB savepoints so that `RELEASE SAVEPOINT <your-custom-name>` functions as a real commit. Note that the existence of this variable and its behavior does not change the fact that CockroachDB savepoints can only be used as a part of the transaction retry protocol.
-{{site.data.alerts.end}}
+This feature exists to support applications that want to use the [advanced client-side transaction retry protocol](advanced-client-side-transaction-retries.html), but cannot customize the name of savepoints to be `cockroach_restart`.  For example, this may be necessary because you are using an ORM that requires its own names for savepoints.
--- a/_includes/v20.1/misc/savepoint-limitations.md
+++ b/_includes/v20.1/misc/savepoint-limitations.md
@@ -1,3 +1,0 @@
-{{site.data.alerts.callout_danger}}
-CockroachDB's [`SAVEPOINT`](savepoint.html) implementation does not support nested transactions (i.e., subtransactions).  It is only used to handle [transaction retries](transactions.html#transaction-retries).
-{{site.data.alerts.end}}
--- a/_includes/v20.1/sql/diagrams/rollback_transaction.html
+++ b/_includes/v20.1/sql/diagrams/rollback_transaction.html
@@ -1,22 +1,17 @@
-<div><svg width="518" height="68">
-         
-         <polygon points="9 17 1 13 1 21"></polygon>
-         <polygon points="17 17 9 13 9 21"></polygon>
-         <rect x="31" y="3" width="92" height="32" rx="10"></rect>
-         <rect x="29" y="1" width="92" height="32" class="terminal" rx="10"></rect>
-         <text class="terminal" x="39" y="21">ROLLBACK</text>
-         <rect x="163" y="35" width="38" height="32" rx="10"></rect>
-         <rect x="161" y="33" width="38" height="32" class="terminal" rx="10"></rect>
-         <text class="terminal" x="171" y="53">TO</text>
-         <rect x="221" y="35" width="98" height="32" rx="10"></rect>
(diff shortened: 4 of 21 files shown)


rmloveland:
Raphael, these doc changes are based on what I found on your `knz/20191208-sql-savepoints` branch thanks to https://github.com/cockroachdb/cockroach/pull/43051

(I realize I'm a bit early.  Trying to get ahead of the curve for the 20.1 release.)

If you think there are too many things that will eventually change based on your in-flight PR or the scheduled follow-up work you mentioned over there, we can hold off for a bit.

knz:
1) note that the RFC mentions rolling back DDL but this is not yet implemented

2) in your changes you say "rollback operates on non-DDL statements". I think this does not convey the intention properly. The proper description is "rollback operates on any statement" and SEPARATELY "known limitation: DDL rollbacks are not supported yet" (they will be in the future).

rmloveland:
> this means that some of the things I commented here and elsewhere in this PR were mistaken, and some of the changes in this PR need to be reverted. I'm going to populate a few examples in the RFC (and the internal crdb test suite) and then we can revisit this docs PR.

OK thanks @knz .  I will pause on this for a bit and revisit when you tell me the RFC examples and test suite are ready.

rmloveland:
(PS thanks @andreimatei for your review.  I look forward to updating this soon with your feedback.)

rmloveland:
Latest updates:

- Rebase onto master - sorry for noise, was getting far behind and had to fix the merge conflict
- Add an example showing how to use ROLLBACK TO SAVEPOINT to handle error condition in sub-transaction
- Fix up various mentions of save points sprinkled throughout the docs that used to say we did not have support or only retry-protocol support to note that we now have general support.

RFAL.

rmloveland:
@knz and/or @andreimatei  - any further comments on this?  I'd love to be able to move this to the next step: docs team review, and then publish.  Thanks!

rmloveland:
@andreimatei do you have any additional changes you'd like to this PR?

If not, I'd like to pass this along for the Docs team to review.

rmloveland:
@jseldess if you have other tasks taking priority, I think this PR is in reasonable shape to merge already - it's had a bunch of review passes.

(Sorry to ping, just trying to get everything in by EOW for the release.)

knz:
> why and when users should use a retry savepoint vs. a savepoint for nested transactions. 

in all fairness to Rich, we don't know either!

We're providing nested transactions because users told us they want them, but we didn't really bother to check _why_ they were needed. 

In the blog post that I wrote about them, I concluded (and reminded the reader) that nested transactions are actually a poor abstraction and should be *discouraged* in new apps. 

In fact, I would prefer if our doc site would not promote their use too much.

I don't think we can properly phrase *any* explanation that sounds like "If you have situation X, then you should use a nested transaction." There is no good situation X that fits this model.

To me the only thing we know are true are:

- we want users to _always_ implement a retry loop on the outer transaction using the restart protocol. That's the same in 20.1 as it was before.
- *for those users who are troubleshooting legacy code using nested transactions*, we need to present *reference-level* docs that explain what is available in CockroachDB for whom wants to know. But we don't want to advertise it really. Especially not to new users.

rmloveland:
> Even if there was truth to them, I'd leave for them for an opinion piece rather than reference docs.

Thanks Andrei.  That makes sense to me.  For now I'll discard that proposed edit, then.

> For now, I'd prefer we keep this simple and find a way to say why nested savepoints are useful. 

OK, I'll work on adding something here, probably 1-2 sentences in the 'Transactions > Nested transactions' section for now.

> I think our special savepoint approach, which we're now calling "retry savepoint" in the docs, is fine.

Cool.

knz:
> In particular, given that they're the only way to recover from errors inside a txn, I think that alone makes them valuable.

You make it sound like recovering from errors is desirable. I don't think so (I don't know of any compelling real world use case where doing so inside a txn is the right  approach). Moreover there are multiple reasons why doing so with nested txns is a bad idea.

Anyway the point remains that the restart protocol (not nested txns) is just as special and important in 20.1 as in 19.x and before and that needs to be conveyed somehow. We should not emphasize nested txns in any way that eclipses the restart protocol.




-- 

jseldess:
___
*[_includes/v20.1/misc/customizing-the-savepoint-name.md, line 1 at r7](https://reviewable.io/reviews/cockroachdb/docs/6675#-M5c8tjr55ecx3PFQ4dm:-M5el7Ne6o9XWyXGc25V:b-rk7do7) ([raw file](https://github.com/cockroachdb/docs/blob/ca7b9fe784f5825e5eda37390585f1afb1f8d679/_includes/v20.1/misc/customizing-the-savepoint-name.md#L1)):*

Updated!
</blockquote></details>

@rmloveland, just realized we need to call this and a few other features out as "New in v20.1".

rmloveland:
Thanks for reviewing that last addition, Andrei!

All, since it's Friday and this needs to be ready for Monday, and it's had many rounds of review, I'm going to merge it now.

Any additional work that is needed, please let's file follow-up issues and we can work on improving it as we go.

Thanks again Raphael, Andrei, Lauren, and Jesse for your reviews and help on this!

jseldess:
___
*[_includes/v20.1/misc/customizing-the-savepoint-name.md, line 1 at r7](https://reviewable.io/reviews/cockroachdb/docs/6675#-M5c8tjr55ecx3PFQ4dm:-M5gmIKu2w2tHe4M4CBq:bg8hdds) ([raw file](https://github.com/cockroachdb/docs/blob/ca7b9fe784f5825e5eda37390585f1afb1f8d679/_includes/v20.1/misc/customizing-the-savepoint-name.md#L1)):*

We got this `force_savepoint_restart` session var in 19.1.  Commit that added it was 7166e8f (fixing #4008).

But perhaps I am misunderstanding? Are you referring to something else in here? Sorry if so and I'm not seeing it.
</blockquote></details>

I'm just talking about the overall ability to use savepoints with custom names. Isn't that new in 20.1? If that variable was there in 19.1, I'm confused...

rmloveland:
> I'm just talking about the overall ability to use savepoints with custom names. Isn't that new in 20.1? If that variable was there in 19.1, I'm confused...

Ah! OK, I see what you mean.

This particular variable was definitely there in 19.1.  But it is for using custom savepoint names with "restart" transactions.

But now we support "real" savepoints.  But we also decided to still support "restart" savepoints as well, to not break existing user integrations.

So if you *don't* toggle this variable, you get "real" savepoints with arbitrary names, but "restart" savepoints must be named `cockroach_restart`.

But if you *do* set this variable, you *don't* get "real" savepoints with arbitrary names anymore, but now "restart" savepoints can have names other than `cockroach_restart`.

What is new in 20.1 is the support for "real" savepoints.

This confusion is caused by what Andrei alluded to: our previous "hack" to re-use SAVEPOINT but have it mean something different.  Specifically, we now "overload" the name, because we have to support our old hack going forward, and "real" savepoints as understood by everyone else in SQL world.

I will make another PR that attempts to add the "new in v20.1" marker to the SAVEPOINT doc in a non-confusing way.  Sorry about the confusion, this whole area is kinda confusing tbh :-/

PR #6691, opened 2020-02-26
Uplevel Driver/ORM support in Developer Guide

ericharmeling:
Upleveled Driver/ORM support overview in "Develop" section.

I also removed this from the SQL Reference section, as this isn't reference material.

### DIFF
--- a/_includes/sidebar-data-v19.2.json
+++ b/_includes/sidebar-data-v19.2.json
@@ -60,6 +60,12 @@
             ]
           },
           {
+            "title": "Supported Drivers and ORMs",
+            "urls": [
+              "/${VERSION}/install-client-drivers.html"
+            ]
+          },
+          {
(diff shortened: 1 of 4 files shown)


ericharmeling:
@awoods187 FYI

awoods187:
I love pulling this up--this is great. I like Jesse's idea but I think I'd need to see it in practice.

ericharmeling:
> Should we consider making this action-oriented, like the rest of the topics in this guide: `Install a Postgres Client`? 

This is a great idea. We currently have no instructions for specific drivers/ORMs outside of their Hello World tutorials. It might be best to decouple that from this PR though, as I'll probably do that for all of them and then make those instructions into includes that we add to the Hello World examples. Thoughts @jseldess ?

jseldess:
SGTM!

PR #6717, opened 2020-03-02
CC upgrade policy

### LINKED ISSUE #6520: CC: Upgrade policy
Reference: https://cockroachlabs.atlassian.net/wiki/spaces/OKRAS/pages/214565035/CockroachCloud+-+Cross+Dept?focusedCommentId=310215283#comment-310215283

Amruta-Ranade:
Closes #6520


emily-horing:
OK here are my thoughts:

Amruta-Ranade:
> OK here are my thoughts:
> 
> **Minor version section:**
> 
> * let's write out "generally available" instead of GA
Changed it to "major" instead to maintain consistency.
> * "before" can be more specific -- maybe "when the upgrade begins and again once it's fully applied" or something
The complete statement reads "before and after their clusters are upgraded..."
> * remove "via email" -- we are hoping to provide the notification in the UI soon, so worth just saying "notify" without being so specific.
Done.
> 
> **Major version section:**
> 
> * "if you are a CockroachCloud Admin, you will receive an email notifying you _when_ a new major release is available."
Reworded it to make it less clunky.
> * I'm thinking we don't need to explain the email this much -- most of the info will end up in the email itself. I would just abbreviate to say "customers will be able to opt-in to the new release and will be upgraded within 3 business days of opting in. If you decide not to opt-in, you will remain on the lowest supported version of CRDB until it reaches EOL"
I think the explanation is important for now since this is a new process and we are educating the users about it.
> 
> **EOL section**
> 
> * "if you are running the lowest supported version" instead of "If you are running older, supported versions" --> there is only 1 version that reaches EOL at a time
Reworded
> * I would use "to the subsequent" instead of "to the latest"; I found it confusing to read since this relates to when there is an even newer version being introduced, and we are just upgrading them by one release. (so if you are running 19.1, we will email to get you to upgrade to 19.2, not to 20.1)
Ahh..I didn't know that. I assumed you'd be upgraded to the latest release. 
Can they ask to be upgraded to the latest version though? 

> * "no later than" --> "at least" . idk why this sounds more forgiving to me :)
Yeah, that sounded weird to me as well but thought it might be necessary legalese. Fixed.
> * "the older version" --> "that version"
Fixed.
> 
> **Rollbacks**
> 
> * i would make this header the same size as the EOL, since it only applies to major upgrades
Good point. Fixed.
> * We should indicate that 7 days is the default (for 20.1 we will have to let people change the finalization period to deal with a schema change issue, so the 7 day time frame won't be enforced for now)
Can you open a ticket and assign it to me about revisiting this section after 20.1 is released and we start upgrading clusters? 
> LMK if any of this doesn't make sense!

emily-horing:
Everything lgtm except for the section on major upgrade actions!

I do want to educate users, but I don't think I feel comfortable with the description of the options because we don't know for sure yet that these are the options delivered in this specific manner. Could we reframe it so that we can make it clear what the user actions are without specifically calling out what the email will look like & what actions will be called in the email? My concern is that it won't be ready to work the way we want and people will be even more confused having read this beforehand.

Here's what I'm thinking:
"CC admins will be notified when a new major release becomes available, and will be able to take one of the following actions:
1. opt in to the new release -- [same text as 'upgrade me now']
2. request a reminder -- [same text as 'remind me']
3. opt in to the release for when the first point release is available -- [same text as 'wait for point release']
4. Do nothing -- [same text as 'ok got it'] "

We can update the policy to be more specific when we build out the UI side of the upgrades work, which we will definitely want to provide more guidance on!

Amruta-Ranade:
> Everything lgtm except for the section on major upgrade actions!
> 
> I do want to educate users, but I don't think I feel comfortable with the description of the options because we don't know for sure yet that these are the options delivered in this specific manner. Could we reframe it so that we can make it clear what the user actions are without specifically calling out what the email will look like & what actions will be called in the email? My concern is that it won't be ready to work the way we want and people will be even more confused having read this beforehand.
> 
> Here's what I'm thinking:
> "CC admins will be notified when a new major release becomes available, and will be able to take one of the following actions:
> 
> 1. opt in to the new release -- [same text as 'upgrade me now']
> 2. request a reminder -- [same text as 'remind me']
> 3. opt in to the release for when the first point release is available -- [same text as 'wait for point release']
> 4. Do nothing -- [same text as 'ok got it'] "
> 
> We can update the policy to be more specific when we build out the UI side of the upgrades work, which we will definitely want to provide more guidance on!

Ahhh..that's good to know. I assumed this was the finalized implementation and will be live before 20.1. Okay, in that case, we can keep it vague and just remove the table/actions for now. Once we implement it, I can add it back. What do you think?

emily-horing:
Ah sorry that's my bad, I should have been more clear about the customer communication stuff being WIP.
I think as long as we make it clear that major upgrades are *opt-in* and the customer will need to act to get upgraded that is sufficient for now to share publically.

Amruta-Ranade:
> Ah sorry that's my bad, I should have been more clear about the customer communication stuff being WIP.
> I think as long as we make it clear that major upgrades are _opt-in_ and the customer will need to act to get upgraded that is sufficient for now to share publically.

No worries! Can you check and LMK if if looks okay now?

emily-horing:
LGTM! Thanks Amruta 😃 

I'll create an issue about changes to the policy (specifically for major upgrades and rollbacks) for us to track as we nail down more details.

joshimhoff:
I'm concerned about merging this now, since we are committing to features we haven't implemented yet. To be more specific, I'm concerned about committing to two biz days for minor releases + to having email notifications ready by April. These are certainly our goals, but can we wait until we have the features built to document? I'm not sure I understand docs workflows well; maybe merging this doesn't lead to it being visible to customers right away?

Amruta-Ranade:
> I'm concerned about merging this now, since we are committing to features we haven't implemented yet. To be more specific, I'm concerned about committing to two biz days for minor releases + to having email notifications ready by April. These are certainly our goals, but can we wait until we have the features built to document? I'm not sure I understand docs workflows well; maybe merging this doesn't lead to it being visible to customers right away?

Once this PR is merged, it will be visible to the customers. I share your concerns -- I am not sure why we want to publish the policy before we implement it. I just know internal stakeholders have been asking for it. But I would be totally okay with publishing a non-specific policy for now and then adding the specifics in a follow-up PR. @emily-horing, thoughts?

emily-horing:
Yeah, I can understand the concern. Frankly the most important piece to me is the EOL stuff because that notification needs to be done ASAP with 20.1 coming in April or we can't reasonably enforce it. 

If we want to generalize the policy further until we have more concrete capabilities I am ok with that, though I do think it would be great if we could put a cap on the minor version upgrades window (maybe 4 days instead of 2? @joshimhoff you can lmk what might make sense here) since that has already come up with a customer when I spoke with them about the policy.

joshimhoff:
> Frankly the most important piece to me is the EOL stuff because that notification needs to be done ASAP with 20.1 coming in April or we can't reasonably enforce it.

Can we just publish the note about only supporting two major versions at a time now? And add the rest in a month or so? I still feel we shouldn't have an external SLA on minor version upgrade timeline until we have the automation. We are close to having the automation, so I don't see why we can't wait a bit longer.

Happy to tak in person too.

v19.2/cockroachcloud-upgrade-policy.md
```diff
@@ -0,0 +1,33 @@
...
+---
+
+This page describes the upgrade policy for CockroachCloud.
+
+CockroachCloud supports the latest major version of CockroachDB and the version immediately preceding it. Support for these versions includes minor version updates and security patches.
+
+## Minor Version Upgrades
+"Minor" versions (or "point" releases) are stable, backward-compatible improvements to the major versions of CockroachDB.
+
+CockroachCloud automatically upgrades all clusters to the latest supported minor version (for example, v19.2.1 → v19.2.2) within the 2 business days after a minor version release. Starting April, the [CockroachCloud Admin(s)](cockroachcloud-console-access-management.html#console-admin) will be notified before and after their clusters are upgraded to the latest minor version.
```
Amruta-Ranade: @joshimhoff Should we remove this para? ("CC automatically upgrades...minor version). 
@emily-horing Or would you prefer we increase the cap (from 2 days to 4 days) but retain the para?

v19.2/cockroachcloud-upgrade-policy.md
```diff
@@ -0,0 +1,33 @@
...
+
+{{site.data.alerts.callout_danger}}
+Single-node clusters will experience some downtime during cluster maintenance.
+{{site.data.alerts.end}}
+
+## Major Version Upgrades
+
+"Major" version releases contain new functionality and potentially backward-incompatible changes to CockroachDB (for example, v19.2.x → v20.1.x).
+
+If you are a [CockroachCloud Admin](cockroachcloud-console-access-management.html#console-admin), you will receive an email notification for each major version release. The email will have instructions on how to opt-in to have your clusters upgraded to the new version.
```
Amruta-Ranade: @joshimhoff Are you okay with leaving this para as is (since this is more for 20.1) or do you want it removed?

v19.2/cockroachcloud-upgrade-policy.md
emily-horing: Can we just say "CockroachCloud upgrades all clusters to the latest supported minor version" and leave it at that? That way customers know that they are always getting point releases but we don't have to commit to anything yet.

joshimhoff: This SGTM. Thanks!

v19.2/cockroachcloud-upgrade-policy.md
joshimhoff: This is okay to leave in! Thanks.

v19.2/cockroachcloud-upgrade-policy.md
Amruta-Ranade: IMO, that would cause more confusion. Like, it can set the expectation that we upgrade the clusters as soon as the version is released. I would rather give them a timeframe or not mention it at all. However, I might be getting paranoid here. I am okay with what y'all decide.

Amruta-Ranade: Okay..going with Emily's suggestion.

joshimhoff: Can we also leave out the part about email notifications? Again, this is the goal, but I don't want to commit to a deadline in case we need to cut scope.

v19.2/cockroachcloud-upgrade-policy.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Auto-upgrades after current version EOL
+
+As CockroachDB introduces new major versions, older versions reach their End of Life (EOL) on CockroachCloud. If you are running a CockroachDB version nearing EOL, you will be notified at least one month before that version’s EOL that your clusters will be auto-upgraded on the EOL date. You should request an upgrade to a newer CockroachDB version during this timeframe to avoid being force-upgraded.
+
+### Rollback support
+
+The default finalization period for each major upgrade is 7 days from upgrade completion. You can request a rollback of a major upgrade via a [support ticket](https://support.cockroachlabs.com/hc/en-us).
```
joshimhoff: "You can request a rollback so long as finalization hasn't been done"?

v19.2/cockroachcloud-upgrade-policy.md
emily-horing: The minor upgrade automation will be the first to be completed (AFAIK) so this will also be the first thing we can update, which makes me more comfortable letting it be vague for now. Thanks Amruta!!

emily-horing:
Re: EOL - yes, the major release date is roughly the same as the EOL date but the difference is that we support the last TWO versions of CRDB, not one. So on April 6, 20.1 will be released and 19.1 will reach EOL. Customers already on 19.2 have the option to opt into 20.1, but if you were running 19.1 you will be forced to upgrade to 19.2. 

Re: rollback - finalization period is basically limbo between being on one version and another. So if you upgrade from 19.2 --> 20.1, with a 7 day finalization period, your upgrade will complete (all nodes will be running 20.1) and then you have 7 days in which you can rollback (ie. there is a path to downgrade). After the finalization happens, you cannot rollback. You would have to restore from a 19.2 backup to get back on 19.2. CRL cannot roll you back either, so it's not a self-service v. support ticket question.

Hope this clarifies!

emily-horing:
"Auto upgrades after current version EOL" title --> let's remove "current" because it implies that the recent version is reaching EOL. We can replace with "running" or something or just remove the adjective.

With that fix this LGTM!

In the v2 when we update this we should maybe link to the SLA on "high availability" in the 1 node downtime warning. Not a blocker.

PR #6767, opened 2020-03-05
Remove merge conflict marker from `SELECT` diagram

rmloveland:
Relates to #2836.


rmloveland:
Eric, requesting your review since I think this conflict got in via 58aa589.  No blame, just want to make sure that the diagram is the correct one you intended.

If what I have here is wrong, it might be easier to close this PR and for you to submit a new, correct diagram.  But let me know what you think!

ericharmeling:
@rmloveland 

Thank you for catching this! I definitely should have caught this when merging that PR.

This looks correct, but to be sure, I just generated a new diagram and opened a new PR to merge that into the 20.1 docs. I'll add you as a reviewer.

PR #6784, opened 2020-03-09
Added precision sections for TIME and TIMESTAMP data types

### LINKED ISSUE #6197: sql: add precision support for TIMESTAMP/TIMESTAMPTZ (v2)
PR: https://github.com/cockroachdb/cockroach/pull/42580

From release notes:
> Introduce precision support for [`TIMESTAMP` and `TIMESTAMPTZ`](../v20.1/timestamp.html), supporting precisions from 0 to 6 inclusive. Previous versions of `TIMESTAMP` and `TIMESTAMPTZ` defaulted to 6 units of precision. Note that if you downgrade while having a precision set, you will have full precision (6) again, but if you re-upgrade you will find your precisions truncated again. [#42580][#42580] {% comment %}doc{% endcomment %}

### LINKED ISSUE #6209: sql: add precision support for TIME/TIMETZ
PR: https://github.com/cockroachdb/cockroach/pull/42668

From release notes:
> Added new support for precision for `TIME` types (e.g. `TIME(3)` will truncate to milliseconds). Previously this would raise syntax errors. [#42668][#42668] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #6209. 
Fixes #6197.

### DIFF
--- a/v20.1/time.md
+++ b/v20.1/time.md
@@ -63,6 +63,16 @@ A `TIME` column supports values up to 8 bytes in width, but the total storage si
 
 A `TIMETZ` column supports values up to 12 bytes in width, but the total storage size is likely to be larger due to CockroachDB metadata.
 
+## Precision
+
+<span class="version-tag">New in v20.1:</span> CockroachDB supports precisions from 0 (seconds) to 6 (microseconds) for `TIME`/`TIMETZ` values. By default, `TIME`/`TIMETZ` values have a precision of 6 (microseconds).
+
+For example, specifying a `TIME` value as `TIME(3)` truncates the time precision to milliseconds.
+
+{{site.data.alerts.callout_info}}
+If you downgrade to a version of CockroachDB that does not support precision for `TIME`/`TIMETZ` values, all `TIME`/`TIMETZ` values previously specified with precision will be stored with full precision.
+{{site.data.alerts.end}}
+
 ## Example
 
 {% include copy-clipboard.html %}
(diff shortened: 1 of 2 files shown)


v20.1/time.md
```diff
@@ -63,6 +63,16 @@ A `TIME` column supports values up to 8 bytes in width, but the total storage si
 
 A `TIMETZ` column supports values up to 12 bytes in width, but the total storage size is likely to be larger due to CockroachDB metadata.
 
+## Precision
```
otan: wondering if we should do examples for this in the below section (e.g. something like https://sourcegraph.com/github.com/cockroachdb/cockroach@793d487107c6a8f61070ef0027e6d0574edb9e07/-/blob/pkg/sql/logictest/testdata/logic_test/timestamp#L85). otherwise, LGTM!

v20.1/time.md
```diff
@@ -63,7 +63,19 @@ A `TIME` column supports values up to 8 bytes in width, but the total storage si
 
 A `TIMETZ` column supports values up to 12 bytes in width, but the total storage size is likely to be larger due to CockroachDB metadata.
 
-## Example
+## Precision
+
+<span class="version-tag">New in v20.1:</span> CockroachDB supports precisions from 0 (seconds) to 6 (microseconds) for `TIME`/`TIMETZ` values. By default, `TIME`/`TIMETZ` values have a precision of 6 (microseconds).
```
Amruta-Ranade: Should we define what `precision` means or is that common knowledge? 
Also, is `precisions` the correct term? (It might be; I am unsure about it because I don't understand what precision means).
The example in the following section doesn't help me understand how/why/when to use it either.

(Sorry if all of this is common knowledge, in which case, please excuse my ignorance)

Issue #6189, opened 2019-12-18
Expose session ID to clients


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/41622

From release notes:
> The ID of the current session is now available via a `session_id` variable. Session IDs are also now shown in [`SHOW QUERIES`](../v20.1/show-queries.html) results. [#41622][#41622] {% comment %}doc{% endcomment %}

PR #6787, opened 2020-03-09
Updated example output in Manage Long-Running Queries

ericharmeling:
I missed this page when updating `SHOW QUERIES` example output in #6764, for #6189.


ericharmeling:
TFTR @Amruta-Ranade !

PR #6800, opened 2020-03-10
document new diagnostics.reporting cluster setting

### LINKED ISSUE #6378: Diagnostics Reporting Doc Update
Re: [Diagnostics Reporting](https://cockroachlabs.com/docs//v19.2/diagnostics-reporting.html)

## Issue Description
Need to update docs to show include turning off `diagnostics.reporting.send_crash_reports` which also sends cluster information to 3rd party. 

Currently, it only tells users to diable `SET CLUSTER SETTING diagnostics.reporting.enabled = false;`

## Suggested Resolution 

Add a new line to read:

`SET CLUSTER SETTING diagnostics.reporting.send_crash_reports = false;`

taroface:
Fixes #6378.

Also added to Cluster Settings table for 20.1. It was only in 19.2

@roncrdb is there any reason `diagnostics.reporting.send_crash_reports` shouldn't have been included in 20.1?

### DIFF
--- a/_includes/v20.1/sql/settings/settings.md
+++ b/_includes/v20.1/sql/settings/settings.md
@@ -10,6 +10,7 @@
 <tr><td><code>diagnostics.reporting.enabled</code></td><td>boolean</td><td><code>true</code></td><td>enable reporting diagnostic metrics to cockroach labs</td></tr>
 <tr><td><code>diagnostics.reporting.interval</code></td><td>duration</td><td><code>1h0m0s</code></td><td>interval at which diagnostics data should be reported</td></tr>
 <tr><td><code>diagnostics.sql_stat_reset.interval</code></td><td>duration</td><td><code>1h0m0s</code></td><td>interval controlling how often SQL statement statistics should be reset (should be less than diagnostics.forced_sql_stat_reset.interval). It has a max value of 24H.</td></tr>
+<tr><td><code>diagnostics.reporting.send_crash_reports</code></td><td>boolean</td><td><code>true</code></td><td>send crash and panic reports</td></tr>
 <tr><td><code>enterprise.license</code></td><td>string</td><td><code></code></td><td>the encoded cluster license</td></tr>
 <tr><td><code>external.graphite.endpoint</code></td><td>string</td><td><code></code></td><td>if nonempty, push server metrics to the Graphite or Carbon server at the specified host:port</td></tr>
 <tr><td><code>external.graphite.interval</code></td><td>duration</td><td><code>10s</code></td><td>the interval at which metrics are pushed to Graphite (if enabled)</td></tr>
--- a/v19.2/diagnostics-reporting.md
+++ b/v19.2/diagnostics-reporting.md
@@ -33,18 +33,23 @@ To make sure that absolutely no diagnostic details are shared, you can set the e
 
 ### After cluster initialization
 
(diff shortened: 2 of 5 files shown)


taroface:
@roncrdb friendly ping for review!

rmloveland:
I don't know anything about this particular setting, but in 20.1 we do have a (new) distinction between "public" and "non-public" cluster settings, which may have something to do with what you are seeing. 

I just checked in 20.1-beta.4 (using `SHOW ALL CLUSTER SETTINGS`, which shows everything, including "non-public") and it looks like this setting is non-public in 20.1 (see `public` column's value).

```
> SHOW ALL CLUSTER SETTINGS;

| variable                                 | value | setting_type | public | description                  |
|------------------------------------------+-------+--------------+--------+------------------------------|
| diagnostics.reporting.send_crash_reports | true  | b            | false  | send crash and panic reports |
```

I have no insight into why / why not for this particular one, though.

rmloveland:
@taroface - Looking at this PR a bit more, I would sync with someone in engineering about these settings as well.

Specifically, to verify if it's actually necessary to toggle *both* of these flags.

Based on what I'm seeing, it looks like you don't need to do that? 

It looks like if you turn off `diagnostics.reporting.enabled` then it would not send the crash reports (since that's also part of diagnostics reporting?) so there is no need to also toggle the (now-hidden in 20.1) crash report setting.

Dug in a bit and based on this comment from #6378

> worth noting @knz has confirmed that the env variable COCKROACH_SKIP_ENABLING_DIAGNOSTIC_REPORTING will disable panic/crash reports as well.

And this comment from the code in `pkg/util/log/crash_reporting.go`

> 	// DiagnosticsReportingEnabled wraps "diagnostics.reporting.enabled".
	//
	// "diagnostics.reporting.enabled" enables reporting of metrics related to a
	// node's storage (number, size and health of ranges) back to CockroachDB.
	// Collecting this data from production clusters helps us understand and improve
	// how our storage systems behave in real-world use cases.
	//
	// Note: while the setting itself is actually defined with a default value of
	// `false`, it is usually automatically set to `true` when a cluster is created
	// (or is migrated from a earlier beta version). This can be prevented with the
	// env var COCKROACH_SKIP_ENABLING_DIAGNOSTIC_REPORTING.
	//
	// Doing this, rather than just using a default of `true`, means that a node
	// will not errantly send a report using a default before loading settings.

It looks like you don't need to toggle the `send_crash_reports` flag, since it's covered when you disable diagnostic reporting using the `enable_*` flag.

Suggest verifying this with @knz since `git blame` shows he made the "public" vs. "non-public" settings change.

If the above hypothesis is correct, these changes don't need to be made in the 20.1 docs (and perhaps not even in 19.2 docs?  If the settings work the same way - IDK)

taroface:
Hi @knz, please see above conversation. This PR documents setting `diagnostics.reporting.send_crash_reports` to `false` in addition to `diagnostics.reporting.enabled` as previously documented. However, the former appears to be a non-public cluster setting. As such, should it be documented, or can our previous guidance (set `diagnostics.reporting.enabled` to `false`) stand?

knz:
I'm sorry I'm not the right person to ask for guidance here any more -- I paged out on this knowledge a long time ago. Deferring to @dt and @rohany for guidance instead.

taroface:
@dt @rohany Would either of you be able to comment on the above?

rohany:
Hey, sorry for the delay. I don't think you need to set both to disable reports. Looking at some code (`pkg/util/log/crash_resporting.go:478`) shows that if `diagnostics.reporting.enabled` is false then we don't send crash reports. I think that `send_crash_reports` is a internal flag that is superseded by `diagnostics.reporting.enabled`.

taroface:
Okay, closing this PR per comment above. Thank you @rmloveland for the detective work!

PR #6820, opened 2020-03-12
Hash-sharded indexes

### LINKED ISSUE #6518: sql: hash sharded indexes
This work was done by an intern on the KV team, but is actually within SQL. We need to consider this part of the 20.1 roadmap list of items.

Description:

> Currently, in CockroachDB, write workloads that are sequential on a particular key will cause a hotspot on a single range if there's any sort of index with the said key as a prefix.
> This PR alleviates this problem by allowing users to shard their indexes into a specific number of buckets by creating a computed shard column under the hood and prefixing the index with this shard column in order to distribute sequential traffic uniformly over the index's ranges.

Here is the RFC https://github.com/cockroachdb/cockroach/pull/42924
Here is the PR https://github.com/cockroachdb/cockroach/pull/42922

Also, @ericharmeling since the intern is no longer working at CRDB @ajwerner is the expert to go to with questions. @ajwerner can you provide any additional information?

### LINKED ISSUE #6730: sql: native support for creating hash sharded indexes
PR: https://github.com/cockroachdb/cockroach/pull/42922

From release notes:
> Added native support of hash-sharded indexes with new `USING HASH WITH BUCKET_COUNT = <n>` syntax for indices (including the primary index of a table). This feature allows users to easily relieve write hot-spots caused by sequential insert patterns at the cost of scan time for queries over the hashed dimension. [#42922][#42922] {% comment %}doc{% endcomment %}

### LINKED ISSUE #6731: sql: enable primary key change into a hash sharded primary key
PR: https://github.com/cockroachdb/cockroach/pull/44993

From release notes:
> Added support for primary key changes into hash sharded indexes. [#44993][#44993] {% comment %}doc{% endcomment %}

### LINKED ISSUE #6732: sql: disable interleaved hash sharded indexes
PR: https://github.com/cockroachdb/cockroach/pull/44996

From release notes:
> Disabled creating a hash sharded index that is also interleaved. [#44996][#44996] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #6732. 
Fixes #6731.
Fixes #6730. 
Fixes #6518.

- Updated diagrams to include USING HASH clause
- Added Hash-sharded indexes section to Indexes page
- Added USING HASH parameter descriptions to ALTER PRIMARY KEY, CREATE INDEX pages
- Added hash-sharding examples to CREATE INDEX, CREATE TABLE pages
- Added experimental_enable_hash_sharded_indexes row to experimental features table
- Updated performance guidance
- Added note about interleaving/hash-shards


jseldess:
___
*[v20.1/indexes.md, line 52 at r1](https://reviewable.io/reviews/cockroachdb/docs/6820#-M2Jrms6DmsB4b4syP4Q:-M2i307y2G7OLOyV2zfG:blpajpq) ([raw file](https://github.com/cockroachdb/docs/blob/7ea793eb08c305e42fdee38ba0f6474015d4a3ca/v20.1/indexes.md#L52)):*

Fair enough. The wording on this is super tricky. I updated it. Please re-review.

A couple things you'll notice:
- I've opted in favor of "indexes on sequential keys" where possible.
- I've distinguished between "sequences", which are a very specific type of incrementing integer, and data that is "sequential in nature", which includes sequences, but also other data that is incremented, like `TIMESTAMP` values. 

LMK what you think.
</blockquote></details>

This is great. I don't know if we have time, but a tutorial on this might be helpful, with a schema that includes sequential keys without a hash-sharded index and then with, showing improved performance. Actually, we should probably add this to the SQL Best Practices page as part of this PR, as a first step. Can you do that?

PR #6825, opened 2020-03-13
Decommissioned nodes are considered ready

### LINKED ISSUE #6589: server: declare node ready while decommissioning
PR: https://github.com/cockroachdb/cockroach/pull/43889

From release notes:
> A node no longer declares itself unready through the [`/health?ready=1`](../v20.1/monitoring-and-alerting.html#health-ready-1) endpoint while in the process of decommissioning. It continues to declare itself unready while draining. [#43889][#43889] {% comment %}doc{% endcomment %}

jseldess:
Update health endpoint doc to clarify that a node is considered
unready when it can no longer accept new SQL connectsions, for
example, when shutting down. This excludes decomissioning since
a decommissioned node, although absent of data, can still serve
as a SQL gateway.

Fixes #6589.

### DIFF
--- a/v19.1/remove-nodes.md
+++ b/v19.1/remove-nodes.md
@@ -12,13 +12,11 @@ For information about temporarily stopping a node (e.g., for planned maintenance
 
 ### How it works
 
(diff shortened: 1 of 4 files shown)


jseldess:
@andreimatei, did I get this right? What happens during decommissioning *before* leases and replicas are transferred? With that understanding, I could add a "Changed in v20.1" callout, if you think it'd be helpful.

PR #6839, opened 2020-03-16
Update cockroach-start-single-node.md

dbist:


dbist:
I have v20.1 patch also but for some reason I can't push it here.

Issue #5885, opened 2019-11-18
Move "Hello, World!" code samples into separate repos


rmloveland:
Richard Loveland (rmloveland) commented:

Currently, our "hello world" style code samples (such as for [Java][java], [Python][python], [Go][golang], etc.) are stored directly in the cockroachdb/docs git repo.

This is not ideal, since it means that developers who want to get the code have to do non-fun things like:

- Copy and paste code, which can introduce errors, or cause dependency issues
- Clone the whole docs repo and troll through its various folder structures to find the code they want

These actions can be tedious and/or error-prone, when you really just want to `git clone`.

To resolve this, we should store each of our "hello world" code samples in their own separate git repos.

Estimated scope of work for this issue is:

1. Come up with a consistent naming scheme for these repos so they are easily recognizable and discoverable.  Also, they should sort nicely in the Github UI's "list of repos" view, while staying grouped by language.  For example, a naming scheme that meets these requirements could be something like 'hello-world-java-jdbc', 'hello-world-java-hibernate', 'hello-world-python-psycopg2', 'hello-world-python-sqlalchemy', and so on. (This is the real work)
2. Create a git repo for each "hello world" code sample
3. Populate it with everything needed to build and run the code
4. Add a README documenting how to do the above

The work of integrating these code samples from external repos into the docs site proper, and testing that integration, etc., are not part of this issue.  Those tasks are tracked in the linked issues below.

See also:

- #3203 which is saying basically the same thing as this issue, but its focus is on how to pull the external repo code samples into the docs, and testability from a docs website POV, which are related but somewhat orthogonal to this task, which is "get them each into separate repos"
- #3943, which also mentions separate repos but is also more about testing/automation from a docs team POV
- #5865 Developer Guide

[java]: https://www.cockroachlabs.com/docs/stable/build-a-java-app-with-cockroachdb.html
[golang]: https://www.cockroachlabs.com/docs/stable/build-a-go-app-with-cockroachdb-gorm.html
[python]: https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb-sqlalchemy.html

Jira Issue: DOC-371

ericharmeling:
I think Rich already did this for Go, Java, Python, and Ruby samples.

@rmloveland can you confirm? Are there any samples that still need their own repo?

rmloveland:
Everything that has its own repo can be viewed here:

https://github.com/cockroachlabs?q=hello-world&type=&language=

- Go/pq
[...]

PR #6840, opened 2020-03-16
Add links to git repos for 'Hello World' apps

rmloveland:
Addresses #5885.

Summary of changes:

- Add links to git repos for 'Hello World' apps for Go, Java, and Python

- Because the git repos show "insecure" connections, the doc updates
  only cover the insecure examples.  The secure examples get a note
  letting the user know about the git repo.

- Propagate the above changes to versions: 19.1, 19.2, 20.1


rmloveland:
@ericharmeling this feels a bit "hacky", in the sense that the git repos are written to connect to insecure clusters (because that's much, much easier), but our docs pages default to secure clusters.   I've tried to split the difference by adding links to the insecure example code even if the docs show a secure example.  If you have better ideas for how to approach this, I'm happy to hear.

PR #6843, opened 2020-03-17
Update advisory 43870

knz:


jseldess:
Thanks, @knz! Let's please make sure to give @Amruta-Ranade time to review since she'll need to connect this with other docs updates.

dbist:
@knz my opinion is because it is yet another behavioral change that requires a new release, to treat this as a net-new advisory instead of updating the existing but given changes are not major and really only affecting core users, we can probably let this slide. LGTM.

jseldess:
___
*[advisories/a43870.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/6843#-M2hLSHc9_ebvFaT8wJL:-M2hLSHc9_ebvFaT8wJM:b-eu5dsu) ([raw file](https://github.com/cockroachdb/docs/blob/e9a6f27fa5ca78c6397f65e3f687212a5dd2d58c/advisories/a43870.md#L29)):*
> ```Markdown
> 
> Starting in v19.2.5 and v20.1, no license is required any more to add
> more users to the `admin` role using the GRANT statement. Users are
> ```

GRANT should be `GRANT`, with `GRANT` as a link to the relevant doc page.

jseldess:
___
*[advisories/a43870.md, line 52 at r1](https://reviewable.io/reviews/cockroachdb/docs/6843#-M2hLbyp2a3Xx1iqVg-h:-M2hLbyq7PWh_U-QuPyu:b77xjp2) ([raw file](https://github.com/cockroachdb/docs/blob/e9a6f27fa5ca78c6397f65e3f687212a5dd2d58c/advisories/a43870.md#L52)):*
> ```Markdown
> 
> Alternatively, additional users can be added to the `admin` group
> via an INSERT statement, for example:
> ```

INSERT should be `INSERT`, with `INSERT` as a link to the relevant doc page. 

@amruta-ranade, let's try to catch these types of issues as a part of the writer review. Can you open a follow-up PR to make these tweaks?

knz:
it's ok I made the mistake I can fix it. I 'll send the PR.

PR #6858, opened 2020-03-18
Update secure-a-cluster.md

dbist:

PR #6640, opened 2020-02-19
document Admin UI security update

taroface:
Fixes #6331 
Fixes #6517
Fixes #6580
Fixes #6617 

This PR does the following on 19.2 and 19.1:

- Adds an "Admin UI access" subheading to Admin UI Overview page explaining why we limit access & listing the restricted Admin UI sections
  - They are called "secure areas" to emphasize that this is a security feature
  - The list of "privileged information" in the table may not be complete
- Describes the two status-quo workarounds for core users provided in the [doc](https://docs.google.com/document/d/1EHxEpsWXM--A-Dq-0YCxwWEJqpDJq2KVdIP_QOeBdNo/edit#) 
- Adds an info callout & link back to "Admin UI access" on the relevant Admin UI pages
- Updates tutorials in which we suggest checking any of the restricted Admin UI pages
  - These tutorials generally emphasize speed (of starting, monitoring, testing a cluster) over conceptual thoroughness. As such, I specifically use the SQL workaround that doesn't require an enterprise license.

Should the above also apply to 2.1?

PR #6859, opened 2020-03-18
Update admin-ui-overview.md

dbist:


dbist:
@jseldess according to this [thread](https://cockroachdb.slack.com/archives/CP4D9LD5F/p1581935937148700?thread_ts=1581577209.135700&cid=CP4D9LD5F), the admin UI mitigation for non-admin users requires server restart. I did not see the same set of instructions on v20.1

jseldess:
@Amruta-Ranade, @taroface, whoever takes https://github.com/cockroachdb/docs/pull/6858 should take this as well.

taroface:
@jseldess @Amruta-Ranade 

Amruta has taken #6858, but I might be better positioned to address this update, since I did the Admin UI security updates and this step needs to be added to a number of different docs (see: #6640). If you both don't mind, I can self-assign this.

Amruta-Ranade:
Oh..hmm..Then do you mind taking both issues? I think dividing them up might cause something to slip through the cracks.

taroface:
@Amruta-Ranade Sure! I will take both. Sorry for any confusion.

taroface:
I opened #6958 based on this change. Closing this one!

PR #6868, opened 2020-03-19
Update secure-a-cluster.md

dbist:


dbist:
no equivalent 20.1 entry.

Amruta-Ranade:
> no equivalent 20.1 entry.

Thanks for fixing the error! There's a 20.1 doc that also needs to be updated. LMK if you'd prefer me to do it!

dbist:
go right ahead @Amruta-Ranade

jseldess:
@Amruta-Ranade, I'm actually just opening a PR to fix this page for a different reason, so I'll make this change as part of that PR. Sorry for the churn here.

PR #6872, opened 2020-03-19
dbist-issue-6855

dbist:


v19.2/build-a-go-app-with-cockroachdb.md
```diff
@@ -16,7 +16,49 @@ We have tested the [Go pq driver](https://godoc.org/github.com/lib/pq) enough to
 
 ## Before you begin
 
-{% include {{page.version.version}}/app/before-you-begin.md %}
```
jseldess: @dbist, thanks for kicking this off. These steps are in the wrong place. They should be down in step 3. I'll close this PR and take care of this in a separate PR.

PR #6999, opened 2020-03-26
Fixes to the 20.1-beta3 release notes

knz:
Two commits here:

1) correctness fixes, mostly in the security updates section. cc @Amruta-Ranade this may inform your work.

2) re-sort the release notes per product area, so that related items are also visually close to each other. (The second commit merely shuffles the notes around; it does not modify them.)

I think moving forward we should apply ad-hoc re-structuring every time a single section exceeds 10 items.


knz:
I think this time is exceptional anyway - such detailed structuring should only be needed when we have 50+ changes to describe, which in the last year happened only twice (around each release).

knz:
@jseldess some things that I noticed as I was going through the change:

- I see we have more and more affinity between release note groups (which cover things of interest to users) and the eng teams. I believe this is due to the org becoming increasingly better at allocating groups of people to user-important problems. 

  This may hint that we could use team membership as one of the variables to hint RN categories (heuristically)

- There are more and more cases where subsequent PRs in the same release alter previous behavior and thus require an update to a previous release note.

  This may hint at the need for a new "queue" system where we'd be able to have engineers edit+merge their release notes ahead of them entering the doc pipeline.

jseldess:
Good points. I've added them to the tracking issue for the lunch and learn.

Issue #6017, opened 2019-11-26
Pause/restart for schema change jobs 


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/recUpBr66N0LAlKD7

Description: Allow for pause and restart for schema change jobs.

Schema changes are an important feature and selling point for using CockroachDB in production. We should make sure the UX of schema changes is good by covering all use cases that are considered schema changes and adding them as a job.

Team: Michael Wang

Github Tracking Issue: https://github.com/cockroachdb/cockroach/issues/22652#issuecomment-366315564 https://github.com/cockroachdb/docs/issues/4015

lnhsingh:
Docs changes:
- Need to remove note on PAUSE JOB / RESUME JOB that says "you cannot pause schema changes"
- Add schema changes to the list of jobs you can pause / resume

jseldess:
From beta.3 release notes:

> - Schema changes are now scheduled and run fully like other [jobs](../v20.1/show-jobs.html): they now can be [canceled](../v20.1/cancel-job.html), [paused](../v20.1/pause-job.html), and [resumed](../v20.1/resume-job.html). Some other UI differences come with this implementation change; notably, all schema changes now have an associated job, failed schema changes are now rolled back within the "Reverting" phase of the same job, and GC for dropped indexes and tables is deferred to a later job. [#45870][#45870]

PR #7009, opened 2020-03-26
Schema change jobs

### LINKED ISSUE #5044: Document or caveat schema change rollbacks
@lucy-zhang pointed out that some schema changes are not completely rolled back when run inside a transaction that fails. For example, it's possible that an ALTER TABLE RENAME COLUMN operation would succeed despite other statements in a transaction being rolled back. We now have an error code to alert users that this may occur, but this has not yet been added to the docs; we should add a "known limitation" to the schema change docs in the meantime.

### LINKED ISSUE #6894: sql, jobs: add schema change job resumer
PR: https://github.com/cockroachdb/cockroach/pull/45870

From release notes:
> Non-running [`jobs`](../v20.1/show-jobs.html) are now considered for adoption in randomized order instead of in a determistic order by creation time, to avoid potential deadlocks when schema change jobs need to execute in a specific order. This is a preemptive change, not a bug fix, but it affects all jobs. [#45870][#45870] {% comment %}doc{% endcomment %}

lnhsingh:
- All schema changes are now run as jobs and can be paused, resumed, and 
canceled (Closes #6894, #6017)
- Clarify what happens if a schema change fails (Closes #5044)


lnhsingh:
@lucy-zhang - friendly review nudge. if you aren't the correct reviewer, please lmk! thanks

PR #7035, opened 2020-03-31
new logo

anagio:
New logo


anagio:
> Thanks, Peter! The logo itself looks nice, but the sizing and placement are slightly different than on the marketing site. Same is true of the topnav menus. Since we are trying to keep these in sync, and maintain the feeling that it's one seamless site, we should probably fix these issues?
> 
> Long-term, we can look at having the docs site diverge from the marketing site.
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7035#-:-M3lJ0LW43nECiylQTGx:b-jrjvnh)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @jseldess)

I see the width of the docs menu is 1530px and on www it's 1500 which should be used? Size will be adjusted

jseldess:
@anagio, I'm not sure which is right. Can you ask @jess-edwards? I think I'd defer to the what's on the marketing site.

jess-edwards:
@jseldess @anagio As long as they match, I don't have a preference. If it's easier to update docs to match the marketing site, please do. If the inverse is easier, that works, too.

jseldess:
@anagio, just bubbling this back to the surface. Sounds like it's fine to make the docs match the marketing site.

PR #7038, opened 2020-03-31
RESTORE ... AS OF SYSTEM TIME 

### LINKED ISSUE #6897: backupccl: allow picking a backup with AS OF SYSTEM time
PR: https://github.com/cockroachdb/cockroach/pull/45368

From release notes:
> [`RESTORE`](../v20.1/restore.html) now allows using `AS OF SYSTEM TIME` to pick a target backup from a larger list of incremental backups. [#45368][#45368] {% comment %}doc{% endcomment %}

lnhsingh:
`AS OF SYSTEM TIME` can now be used for any restore, not just of backups with revision history

Closes #6897.

### DIFF
--- a/v20.1/restore.md
+++ b/v20.1/restore.md
@@ -34,6 +34,10 @@ You can restore:
 
 Because this process is designed for disaster recovery, a full cluster restore can only be run on a target cluster with no databases or tables.
 
+{{site.data.alerts.callout_info}}
+When you do a full cluster restore, it will restore the [enterprise license](enterprise-licensing.html) of the cluster you are restoring from. If you want to use a different license in the new cluster, make sure to [update the license](enterprise-licensing.html#set-a-license) _after_ the restore is complete.
+{{site.data.alerts.end}}
+
 #### Databases
 
 To restore a database, the database cannot already exist in the target cluster. Restoring a database will create the database and restore all of its tables and views. By default, tables and views are restored into a database with the name of the database from which they were backed up. However, also consider:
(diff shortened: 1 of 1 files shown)


v20.1/restore.md
```diff
@@ -34,6 +34,10 @@ You can restore:
 
 Because this process is designed for disaster recovery, a full cluster restore can only be run on a target cluster with no databases or tables.
 
+{{site.data.alerts.callout_info}}
+When you do a full cluster restore, it will restore the [enterprise license](enterprise-licensing.html) of the cluster you are restoring from. If you want to use a different license in the new cluster, make sure to [update the license](enterprise-licensing.html#set-a-license).
```
dt, line 38: nit: might help to clarify that that is update the license _after the restore is completed_

v20.1/restore.md
```diff
@@ -84,7 +88,7 @@ Full backup + <br>incremental backups | If the full backup and incremental backu
 
 ### Point-in-time restore
 
-If the full or incremental backup was taken [with revision history](backup.html#backups-with-revision-history), you can restore the data as it existed at the specified point-in-time within the revision history captured by that backup.
+<span class="version-tag">New in v20.1</span> You can restore data as it existed at a specified point-in-time for any full or incremental backup by using the [`AS OF SYSTEM TIME`](as-of-system-time.html) clause.
```
dt, line 91: This is not _quite_ accurate: you still need to have used `with revision_history` when backing up if you want to restore to just any old time, however now, if you're backing up incrementally every X time, you can restore to any X using `AS OF SYSTEM TIME`.

You used to be able to do the same thing when each incremental backup was separate by just specifying which one, e.g. say you backup every hour so you have `base`, `1am`, `2am`, `3am`, `4am`, ... and you want to restore to `3am` -- in the old UX, you'd have separate backups for each hour and you'd run `RESTORE ... FROM base, 1am, 2am, 3am`. In the new UX, all those backups are in one big `base` folder and while you _could_ specify them individually to RESTORE, it is probably easier to just say what time you want, e.g. `RESTORE FROM base AS OF SYSTEM TIME ..:03:..`.

Note that that requires you specify the exact time of the 3am backup, which can be found in the `end_time` of the `SHOW BACKUP` output, or, if you just say the time you want, the error will tell you the nearest adjacent times.

Put a bit more formally:
If you're taking periodic incremental backups, for example hourly, you can restore to a specific one, i.e. to 6am or 7am etc, by specifying the end time of that backup (visible in SHOW BACKUP) in the `AS OF SYSTEM TIME` option to RESTORE. Additionally, you can restore to an _arbitrary time_ between the start and end time of the backup if the backup was created with the `revision_history` option.

Issue #6890, opened 2020-03-19
sql: Role management, CREATEROLE privilege


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/44232

From release notes:
> Modified the [`CREATE ROLE`](../v20.1/create-role.html) statement to add the `CREATEROLE` option. Also added the `ALTER ROLE` statement and the `CREATEROLE` privilege. It is managed by choosing options when creating and altering a role. Example: `CREATE ROLE role WITH CREATEROLE;` [#44232][#44232] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/7061

Issue #6882, opened 2020-03-19
sql: user account expiration (role LOGIN option and password expiry via VALID UNTIL option)


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/45541

From release notes:
> User and role principals can now be prevented from logging in using the `NOLOGIN` attribute, which can be set using [`ALTER USER/ROLE`](../v20.1/alter-user.html). [#45541][#45541] {% comment %}doc{% endcomment %}
>
> [`SHOW USERS`](../v20.1/show-users.html) now shows all roles/users that have `LOGIN` enabled. [#45541][#45541] {% comment %}doc{% endcomment %}
>
> Added the `VALID UNTIL` option to indicate passwords are only valid until a given timestamp. `VALID UNTIL` only affects `PASSWORDS` - if passwords are disabled, `VALID UNTIL` has no effect. Also, you can login even with expired password if passwords are not required to login-- same as PostgreSQL). [#45541][#45541] {% comment %}doc{% endcomment %}
>
> Added the `LOGIN` option to enable login for roles/users. When using [`CREATE USER`](../v20.1/create-user.html) syntax, `LOGIN` is enabled by default whereas when using [`CREATE ROLE`](../v20.1/create-role.html), `LOGIN` is not enabled by default. [#45541][#45541] {% comment %}doc{% endcomment %}
>
> Any [role](../v20.1/roles.html) created prior to v20.1 will be able to log into clusters started with `--insecure`, unless/until they are given the `NOLOGIN` attribute. For secure clusters, roles cannot log in by virtue of not having a password nor a client certificate. [#45541][#45541] {% comment %}doc{% endcomment %}
>
> The password field (used exclusively for password-based authentication) can now be configured to have an expiration date using the `VALID UNTIL` attribute, which can be set with [`ALTER USER/ROLE`](../v20.1/alter-user.html). Note that the attribute sets an expiration date for the password, not the user account. This is consistent with PostgreSQL. [#45541][#45541] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/7061

Issue #6878, opened 2020-03-19
cli: validation rule for principal names was extended to support periods


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/45575

From release notes:
> The validation rule for principal names was extended to support periods and thus allow domain name-like principals. For reference, the validation rule is currently the regular expression `^[\p{Ll}0-9_][---\p{Ll}0-9_.]+$`, and limited to a size of 63 UTF-8 bytes (larger usernames are rejected with an error); for comparison, PostgreSQL allows many more characters and truncates at 63 characters silently. [#45575][#45575] {% comment %}doc{% endcomment %}
>
> [Usernames](../v20.1/create-user.html) can now contain periods, for compatibility with certificate managers that require domain names to be used as usernames. [#45575][#45575] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/7061

Issue #6876, opened 2020-03-19
sql: release note for ZONECONFIG


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/45201

From release notes:
> Admin users can now [grant](../v20.1/grant.html) `ZONECONFIG` to non-admin users on specific SQL objects (databases/tables). When set, the user is authorized to modify that object's [zone configuration](../v20.1/configure-replication-zones.html) and decide data placement on specific nodes or groups thereof. [#45201][#45201] {% comment %}doc{% endcomment %}

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/7061

PR #7061, opened 2020-04-02
IAM doc updates

### LINKED ISSUE #7034: Update authorization best practices
Source: Raphael on Slack:

- at the end of the first, "create users or roles with specific privileges based..."
- remove the second bullet entirely
- keep the 3rd as-is

Amruta-Ranade:
Closes #7034 #6586 #6605 #6882 #6890 #6876 #5960 #6622 #6804 #6878 

Summary of changes:
- Updated authorization best practices based on Raphael's comment on Slack
- Added `ALTER ROLE` doc
- Added content for enabling password for the `root` user
- Added content for disabling password authentication
- Documented login privilege parameters
- Added content for password validity
- Added content for `CREATEROLE`
- Added content for `zoneconfig`
- Updated username requirements

cc @aaron-crl for visibility


Amruta-Ranade:
The broken link is due to a SQL diagram error I think. Might need help looking into it.

RichardJCai:
Some general comments.
* In the ALTER ROLE / USER, CREATEROLE privilege is missing.
* This is a preference but I would prefer the options be capitalized - ie refer always use `CREATEROLE` instead of `createrole`
* Do you think it would be a good idea to reuse use the same documents for the ROLE vs USER versions of the syntaxes since it's effectively the same? Could possibly even use a redirect from the user verison to the ROLE version and really drive home the fact that users = roles. On this topic I would like a note saying that the only difference between the whole ROLE vs USER syntax is that CREATE USER has login enabled by default while CREATE ROLE has login disabled by default. `CREATE USER is now an alias for CREATE ROLE. The only difference is that when the command is spelled CREATE USER, LOGIN is assumed by default, whereas NOLOGIN is assumed when the command is spelled CREATE ROLE.` pg has this noted in there docs - https://www.postgresql.org/docs/current/sql-createuser.html

Amruta-Ranade:
> Some general comments.
> 
> * In the ALTER ROLE / USER, CREATEROLE privilege is missing.
Good catch. Fixed.

> * This is a preference but I would prefer the options be capitalized - ie refer always use `CREATEROLE` instead of `createrole`
Me too. I don't know why it's lowercase in some cases in legacy content. I followed the conventions in each doc for now but will follow up in another PR to make  capitalization the default.

> * Do you think it would be a good idea to reuse use the same documents for the ROLE vs USER versions of the syntaxes since it's effectively the same? Could possibly even use a redirect from the user verison to the ROLE version and really drive home the fact that users = roles. Also it be easier to update in the future instead of managing two versions of essentially the same command. On this topic I would like a note saying that the only difference between the whole ROLE vs USER syntax is that CREATE USER has login enabled by default while CREATE ROLE has login disabled by default. `CREATE USER is now an alias for CREATE ROLE. The only difference is that when the command is spelled CREATE USER, LOGIN is assumed by default, whereas NOLOGIN is assumed when the command is spelled CREATE ROLE.` pg has this noted in there docs - https://www.postgresql.org/docs/current/sql-createuser.html
Yep. I need to get to the authentication updates first, but this docs reorg is on my to-do list. I opened an issue to track this: https://github.com/cockroachdb/docs/issues/7063

Amruta-Ranade:
> Looks good overall - just left some notes in the previous comment.

TFTR!!

v20.1/alter-user.md
```diff
@@ -33,7 +35,10 @@ table td:first-child {
...
+`password` | Let the user [authenticate their access to a secure cluster](authentication.html#client-authentication) using this new password. Passwords should be entered as [string literal](sql-constants.html#string-literals). For compatibility with PostgreSQL, a password can also be entered as an [identifier](#change-password-using-an-identifier), although this is discouraged. <br><br>To prevent a user from using [password authentication](authentication.html#client-authentication) and to mandate [certificate-based client authentication](authentication.html#client-authentication), [set the password as `NULL`](#prevent-a-user-from-using-password-authentication).
+`valid until` | The date and time (in the [`timestamp`](timestamp.html) format) after which the password is not valid.
+`login`/`nologin` | The `login` parameter allows a user to login with one of the [client authentication methods](authentication.html#client-authentication). [Setting the parameter to `nologin`](#change-login-privileges-for-a-user) prevents the user from logging in using any authentication method.
```
RichardJCai, line 40: NOLOGIN only disables a logining in using a password - this is really unintuitive but this is pg's behaviour and we decided to mimic it. I think the reason for copying this behaviour was because we decided the other login methods have their own ways of disabling login - ie revoking certificates.

v20.1/authorization.md
```diff
@@ -28,6 +28,8 @@ By default, a new user belongs to the `public` role and has no privileges other
 ### `root` user
```
RichardJCai, line 28: I'm just including it here since this line doesn't show in the diff, but theres a block stating 
```
Use the CREATE USER and DROP USER statements to create and remove users, the ALTER USER statement to add or change a user's password, the GRANT <privileges> and REVOKE <privileges> statements to manage the user’s privileges, and the SHOW USERS statement to list users.
```

Can you update this to say `the ALTER USER statement to add or change a users password and role options`

v20.1/create-role.md
```diff
@@ -38,6 +40,10 @@ Roles can only be created by superusers, i.e., members of the `admin` role. The
 | Parameter | Description |
 ------------|--------------
 `name` | The name of the role you want to create. Role names are case-insensitive; must start with either a letter or underscore; must contain only letters, numbers, or underscores; and must be between 1 and 63 characters.<br><br>Note that roles and [users](create-user.html) share the same namespace and must be unique.
+`password` | Let the role [authenticate their access to a secure cluster](authentication.html#client-authentication) using this password. Passwords must be entered as [string](string.html) values surrounded by single quotes (`'`).<br><br>Password creation is supported only in secure clusters.
+`valid until` | The date and time (in the [`timestamp`](timestamp.html) format) after which the password is not valid.
+`login`/`nologin` | Allow or disallow a role to login with one of the [client authentication methods](authentication.html#client-authentication). <br><br>By default, the parameter is set to `nologin` for the `CREATE ROLE` statement.
```
RichardJCai, line 45: Same comment as before - only disables password logins.

v20.1/create-user.md
```diff
@@ -25,6 +25,8 @@ The `CREATE USER` [statement](sql-statements.html) creates SQL users, which let
 
 The user must have the `INSERT` and `UPDATE` [privileges](authorization.html#assign-privileges) on the `system.users` table.
```
RichardJCai, line 26: This is no longer true - having CREATEROLE privilege is sufficient - this is true to all CREATE/ALTER/DROP role/user commands

v20.1/create-user.md
```diff
@@ -62,22 +69,36 @@ Usernames are case-insensitive; must start with a letter, number, or underscore;
...
+### Set password validity
+
+The following statement sets the date and time after which the password is not valid:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> CREATE USER carl VALID UNTIL '2021-01-01';
+~~~
 
 ### Manage users
```
RichardJCai, line 103: ```After creating a user, you can use the ALTER USER statement to add or change the user's password and the DROP USER statement to the remove users.```
In manage users, you can mention ALTER USER can update role options.

v20.1/drop-role.md
```diff
@@ -19,6 +19,8 @@ The `DROP ROLE` [statement](sql-statements.html) removes one or more SQL roles.
 
 Roles can only be dropped by super users, i.e., members of the `admin` role.
```
RichardJCai, line 20: This is no longer true - having CREATEROLE is sufficient.

RichardJCai: Ditto for drop user - DELETE privilege is not required on system.users. 
In general permissions on system.users are no longer required for these RBAC commands.

v20.1/show-roles.md
```diff
@@ -18,7 +18,7 @@ The `SHOW ROLES` [statement](sql-statements.html) lists the roles for all databa
 
 ## Required privileges
 
-The user must have the [`SELECT`](select-clause.html) [privilege](authorization.html#assign-privileges) on the system table.
+The role must have the [`SELECT`](select-clause.html) [privilege](authorization.html#assign-privileges) on the system table.
```
RichardJCai, line 21: This one you actually need select on system.users table

RichardJCai: Ditto for system.roles - I believe `system` should be replaced with `system.users`

RichardJCai:
> I think the only main open topic at this point is what should be done about NOLOGIN. I'm pretty sure that what you described in this PR is good, and is also what PostgreSQL does. If our code is different (what Richard seems to think, and this comes as a surprise to me), we need to take action on that one way or another.

@knz Oh oops I had a brain fart - I mixed up NOLOGIN with VALID UNTIL. VALID UNTIL only applies to passwords, NOLOGIN applies to every auth method. My bad please disregard what I wrote about that, I'll edit it out.

Amruta-Ranade:
@knz TFTR! I added a small commit for the "username can now contain periods" change. LMK if someone else should review it.

PR #7082, opened 2020-04-05
[WIP] Reposition feedback widget

jseldess:
An experiment to see if we can better expose the docs feedback widget by making it always visible in the lower-right, like in [MongoDB's docs](https://docs.mongodb.com/stitch/users/). I'm interested in seeing if an approach like this drives more engagement (positive and negative submissions). The styling/design needs work, of course.

Also, this PR removes the marketing footer content from docs, where it's not really needed and takes up a lot of space.


jseldess:
cc @luciaaaaaaa, @anagio

Issue #7067, opened 2020-04-03
Automate the testing of SQL commands


jseldess:

jseldess:
Dup of https://github.com/cockroachdb/docs/issues/4705

PR #7093, opened 2020-04-06
Use correct `foo@foo_idx` syntax in examples

### LINKED ISSUE #7070: example is missing table name@index
[This example](https://www.cockroachlabs.com/docs/stable/online-schema-changes.html#examples-of-statements-that-fail) incorrectly has this statement:
```
 SELECT * from foo_idx;
```
It should be
```
 SELECT * from foo@foo_idx;
```

The results incorrectly has:
```
ERROR:  relation "foo_idx" does not exist
```
It should be
```
pq: index "foo_idx2" not found
```

rmloveland:
Fixes #7070.

Summary of changes:

- Update example on 'Online Schema Changes' page that was not using the
  correct syntax for running a selection query against a specific
  index.

- (Funnily enough, the error message produced was the same either way,
  which is probably why this slipped in.  That, and lack of #7067.)

### DIFF
--- a/v19.1/online-schema-changes.md
+++ b/v19.1/online-schema-changes.md
@@ -141,7 +141,7 @@ The following statements fail due to [limited support for schema changes within 
   BEGIN;
   SAVEPOINT cockroach_restart;
   CREATE INDEX foo_idx ON foo (id, name);
-  SELECT * from foo_idx;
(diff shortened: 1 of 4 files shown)


rmloveland:
Thanks Florence!

PR #7109, opened 2020-04-09
Add docs on the protected timestamp subsystem

### LINKED ISSUE #6003: Safer backups independent from ttl zone config
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/recXZ8sKfgRGaKe6Y

Description: Enables capturing a snapshot at a point in time that garbage collection won't collect from.

Garbage collection has no guards for how far back it can go in time. We lack a mechanism for defining a time to not garbage collect from.

Team: John Kendall, Andrew Werner

Github Tracking Issue:

rmloveland:
Fixes #6003.

Summary of changes:

- Add new section 'Storage Layer > GC > Protected Timestamps', which
  describes what they are and how they work at a high level.

### DIFF
--- a/v20.1/architecture/storage-layer.md
+++ b/v20.1/architecture/storage-layer.md
@@ -57,7 +57,24 @@ Using these tools, you can get consistent data from your database as far back as
 
 ### Garbage collection
 
-CockroachDB regularly garbage collects MVCC values to reduce the size of data stored on disk. To do this, we compact old MVCC values when there is a newer MVCC value with a timestamp that's older than the garbage collection period. The garbage collection period can be set at the cluster, database, or table level by configuring the [`gc.ttlseconds` replication zone variable](../configure-replication-zones.html#gc-ttlseconds).  For more information about replication zones, see [Configure Replication Zones](../configure-replication-zones.html).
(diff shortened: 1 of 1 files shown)


ajwerner:
This looks great. I wonder if we should document that the enforcement of the GC TTL is independent of the actual removal of data (potentially you were going to address that separately with the strict gc ttl change).

v20.1/architecture/storage-layer.md
```diff
@@ -57,7 +57,24 @@ Using these tools, you can get consistent data from your database as far back as
...
+- [Online schema changes](../online-schema-changes.html)
+
+Protected timestamps ensure the safety of historical data while also enabling shorter [GC TTLs](../configure-replication-zones.html#gc-ttlseconds). A shorter GC TTL means that fewer previous MVCC values are kept around. This can help lower query execution costs for workloads which update rows frequently throughout the day, since [the SQL layer](sql-layer.html) has to scan over previous MVCC values to find the current value of a row.
+
+##### How protected timestamps work
+
+Protected timestamps work by creating *protection records*, which are stored in an internal system table. When a long-running job such as a backup wants to protect data at a certain timestamp from being garbage collected, it creates a protection record associated with that data and timestamp.
+
+Upon successful creation of a protection record, the MVCC values for the specified data at timestamps less than or equal to the protected timestamp will not be garbage collected. When the job that created the protection record finishes its work, it removes the record, allowing the garbage collector to run on the formerly protected values.
```
ajwerner, line 77: Maybe add another sentence saying that there exists a background job to reconcile protection records with running jobs to defend against bugs where jobs fail to clean the records up themselves?

rmloveland: I think we should leave this out for now.  I was wondering whether to mention the jobs subsystem at all here, but we at least had to mention it in the context of how backups etc. get run (IMO).  However, I'd rather not get further into the jobs subsystem's workings than this here.

Issue #1821, opened 2017-08-14
Code blocks: Copy to clipboard button overlaps code


jseldess:
[Reported by](https://github.com/cockroachdb/docs/pull/1763#issuecomment-322284639) from @jess-edwards: The "copy" button in code snippets cuts off longer lines. How can we improve this UX?

cc: @kuanluo 

![screen shot 2017-08-14 at 4 51 06 pm](https://user-images.githubusercontent.com/15893490/29291191-cc433342-8110-11e7-9220-822d66003b93.png)

kuanluo:
@jseldess @jess-edwards Can you point me to where this happens? I poked around, and saw longer lines usually have this behavior:
![image](https://user-images.githubusercontent.com/1486483/29320772-8404802e-8195-11e7-8d31-bba015123602.png)
(From https://www.cockroachlabs.com/docs/stable/upgrade-cockroach-version.html)

jseldess:
Yeah, as you've found, it happens wherever we've used copy to clipboard and a line is long enough to be obscured by the "Copy" text and icon. In principle, I think it's ok for long lines to continue past the right margin. The "Copy" text is just pushing the right margin further in. 

A possible solution would be moving the "copy" text and icon outside of the code block, just to the right. I can't remember why we didn't go with this in the first place...

kuanluo:
> Yeah, as you've found, it happens wherever we've used copy to clipboard and a line is long enough to be obscured by the "Copy" text and icon. In principle, I think it's ok for long lines to continue past the right margin. The "Copy" text is just pushing the right margin further in.

Agreed. The screenshot you attached should have the same behavior. That might be a bug? 

> A possible solution would be moving the "copy" text and icon outside of the code block, just to the right. I can't remember why we didn't go with this in the first place...

Might be an overall content width issue. We can reopen the convo with Division Of and see what we can do?

luciaaaaaaa:
@jseldess I think its okay for the copy button to cover some of it (since the user will have to scroll anyways to see the whole line. 

Show the scroll bar when the user is scrolling or on hover would help clarify there is more past the right margin.
Removing the icon from the copy button will also help

luciaaaaaaa:
Link to designs in Figma: https://www.figma.com/file/6XcxaUdad2RkjpEqnVHKAG/Education-docs-redesign?node-id=2228%3A12946

PR #7163, opened 2020-04-15
Add max height and vertical scrollbar to code samples

yingzhuchin:
Still need to add option to expand, but throwing this up for some quick feedback.

### DIFF
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -3406,9 +3406,8 @@ pre, code {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   line-height: 25px;
+  max-height: 320px;
   background-color: #F7F7F7;
-  font-size: 14px;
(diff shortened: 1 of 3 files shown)


yingzhuchin:
I just want to call out that if the scrollbar were to be moved to the right of the copy button, then in certain situations the button will obscure part of the code text, like so:

Screen Shot 2020-04-16 at 12 02 04 PM

This is taken from https://kafka-tutorials.confluent.io/filter-a-stream-of-events/ksql.html, section 2 (Get Confluent Platform).

yingzhuchin:
Hi @jseldess, I found this UX issue: https://github.com/cockroachdb/docs/issues/1821, and so just wanted to double-check with you that we indeed do want to move the vertical scrollbar to the right of the Copy button, which may cause the button to obscure code (see my previous comment).

yingzhuchin:
Closing in favor of docs redesign project.

Issue #6920, opened 2020-03-19
kv: support unreplicated locks, hook up SELECT FOR UPDATE end-to-end


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/45701

From release notes:
> [`SELECT FOR UPDATE`](../v20.1/select-for-update.html) now hooks into a new leaseholder-only locking mechanism. This allows the feature to be used to improve performance of [transactions](../v20.1/transactions.html) that read, modify, and write contended to rows. Similarly, [`UPDATE`](../v20.1/update.html) statements now use this new mechanism by default, meaning that their performance under contention is improved. This is only enabled for `UPDATE` statements that can push their filter all the way into their key-value scan. To determine whether an `UPDATE` statement is implicitly using `SELECT FOR UPDATE` locking, look for a `locking strength` field in the `EXPLAIN` output for the statement. [#45701][#45701] {% comment %}doc{% endcomment %}

rmloveland:
Fixed in #7176

Issue #6903, opened 2020-03-19
sql/opt/exec: add implicit SELECT FOR UPDATE support for UPDATE statements


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/45159

From release notes:
> [`UPDATE`](../v20.1/update.html) statements now acquire locks using the `FOR UPDATE` locking mode during their initial row scan, which improves performance for contended workloads. This behavior is configurable using the `enable_implicit_select_for_update` [session variable](../v20.1/set-vars.html) and the `sql.defaults.implicit_select_for_update.enabled` [cluster setting](../v20.1/cluster-settings.html). [#45159][#45159] {% comment %}doc{% endcomment %}

rmloveland:
Fixed by #7176

PR #7176, opened 2020-04-16
Document viewing locking strength with EXPLAIN

### LINKED ISSUE #6699: Update docs to mention that DELETE and UPSERT also use the locking mechanism created for SELECT FOR UPDATE
These statements may get implicit `SELECT FOR UPDATE` locking support by the 20.1 release.

Estimated scope of work:

- Update docs for the `enable_implicit_select_for_update` setting to also mention `DELETE` and `UPSERT`

See also:

- #6000
- #6671

rmloveland:
Fixes #6699, #6903, #6920.

Summary of changes:

- Add a new section to EXPLAIN docs called 'Find out if a statement is
  using SELECT FOR UPDATE locking', with an example showing how to
  check.

- List out the SQL statements that have support enabled by default for
  SFU locking.

### DIFF
--- a/_includes/v20.1/misc/session-vars.html
+++ b/_includes/v20.1/misc/session-vars.html
@@ -107,6 +107,7 @@
 
   <tr>
    <td>
+    <a name="enable-implicit-select-for-update"></a>
     <code>enable_implicit_select_for_update</code>
    </td>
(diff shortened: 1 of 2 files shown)


rmloveland:
@nvanbenschoten in case it helps your review, some context for the structure of only putting this on the `EXPLAIN` page is:

If you are using `UPDATE`, `UPSERT`, etc. and they are now magically faster under contention, you don't need to care. So we don't put any mention of locking on those pages.

However, if you are using `EXPLAIN`, you might be debugging or trying to figure something out, in which case we mention it here (with links to affected statements), since we must in any case mention the new `EXPLAIN` output field.

I do have a question as well:

- Do you think we should link to the session var for toggling SFU locking from this EXPLAIN doc section?  Since folks are possibly testing/troubleshooting stuff here?  My sense was "maybe" but I'm not sure if/when users should ever toggle it unless they are pretty expert, and/or working with our Support team.  My sense is that linking it from here will tend to slightly "encourage" its use due to higher visibility.

Issue #5111, opened 2019-07-30
Document environment variables


robert-s-lee:
Exalate commented:

Re: [Cockroach Commands](https://cockroachlabs.com/docs//v19.1/cockroach-commands.html)

## Issue Description

customers sometimes have set CockroachDB recognized environment variable that inadvertently used.  For example:

In 19.1.3 this:

``` 
cockroach sql --host $(hostname).cockroach --certs-dir /certs/dir 
```

Fails with:

``` 
Error: invalid argument "hostname" for "--host" flag: address hostname:tcp://127.0.0.1:26257: too many colons in address 
Failed running "sql" 
```

## Suggested Resolution 

Document environment variable and where they could be used within CockroachDB

jseldess:
@robert-s-lee, we do mentioned env variables in the docs already, albeit at a high level: https://www.cockroachlabs.com/docs/stable/cockroach-commands.html#environment-variables. Do you think we need some specific examples?

robert-s-lee:
We should document which env variables we honor.  One customer somehow had a env setting that was used as a part of connection string.   A document showing a list of env variables and where/how they are used would have been useful in the troubleshooting.

jseldess:
Community user also asked for better docs here: https://forum.cockroachlabs.com/t/what-is-the-environment-variable-for-http-addr-flag/3498/2.

PR #7177, opened 2020-04-16
Document inconsistent time zone env var

### LINKED ISSUE #6928: Document inconsistent time zone environment variable for testing
PR: https://github.com/cockroachdb/cockroach/pull/45640

From release notes:
> CockroachDB now refuses to start if named time zones are not properly configured. It is possible to override this behavior for testing purposes, with the understanding that doing so will cause incorrect SQL results and other inconsistencies, using the environment variable `COCKROACH_INCONSISTENT_TIME_ZONES`. [#45640][#45640] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #6928.

Summary of changes:

- Document a new env var `COCKROACH_INCONSISTENT_TIME_ZONES`, which
  allows you to start CRDB even if timezones are not properly
  configured (and thus things will break).  For testing only.

### DIFF
--- a/v20.1/cockroach-start.md
+++ b/v20.1/cockroach-start.md
@@ -195,6 +195,14 @@ Field | Description
 `clusterID` | The ID of the cluster.<br><br>When trying to join a node to an existing cluster, if this ID is different than the ID of the existing cluster, the node has started a new cluster. This may be due to conflicting information in the node's data directory. For additional guidance, see the [troubleshooting](common-errors.html#node-belongs-to-cluster-cluster-id-but-is-attempting-to-connect-to-a-gossip-network-for-cluster-another-cluster-id) docs.
 `nodeID` | The ID of the node.
 
+## Environment variables
+
+The following environment variables change the behavior of `cockroach start`:
+
+| Variable                            | Description                                                                                                                                                         |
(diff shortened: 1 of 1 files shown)


knz:
Are you sure you want to document this?

jseldess:
@knz, this was exposed in release notes. I assume most if not all things in release notes are suitable for docs. Do you see this otherwise? This might be a point of clarification for the upcoming release notes lunch and learn.

knz:
I think this setting is only useful for users who are deploying cockroachdb inside containers (other than our own) or on Windows. I wouldn't emphasize it in the main `cockroach start` docs.

rmloveland:
TBH I'm fine with not documenting it and just leaving it in the release notes, if we think it's sufficiently edge-casey (which I personally think it is).

Any objection to me closing this, Raphael?


(Aside: FWIW I put it in `cockroach start` because that seemed like the only peg to hang it on, but I think this var should be documented (if at all) as part of #5111 - agree that putting it on `cockroach start` gives it an emphasis that makes it seem "ok to use" when it mostly isn't.)

knz:
ok to add this to the list in  #5111 instead.

PR #7219, opened 2020-04-24
Mark savepoints as 'New in v20.1' in more places

rmloveland:
Addresses  #5953.

Summary of changes:

- Add 'new in v20.1' tags and appropriate verbiage to the following
  parts of the docs that were missed in some previous work in this area:

  - RELEASE SAVEPOINT
  - ROLLBACK
  - SAVEPOINT
  - 'SQL Statements'

### DIFF
--- a/v20.1/release-savepoint.md
+++ b/v20.1/release-savepoint.md
@@ -4,7 +4,7 @@ summary: Commit a nested transaction.
 toc: true
 ---
 
-The `RELEASE SAVEPOINT` statement commits the [nested transaction](transactions.html#nested-transactions) starting at the corresponding `SAVEPOINT` statement using the same savepoint name, including all its nested sub-transactions.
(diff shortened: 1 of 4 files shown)


rmloveland:
> Hmm, wouldn't it make more sense to callout nested transactions and custom savepoint names as new?

Sorry, I don't understand.

Nested transactions are called out as "new in 20.1" already:

https://www.cockroachlabs.com/docs/v20.1/transactions.html#nested-transactions

Custom savepoint names (via the cluster setting) were new in 19.1:

https://www.cockroachlabs.com/docs/v19.1/advanced-client-side-transaction-retries.html#customizing-the-savepoint-name

rmloveland:
Hey Jesse, any objections to merging this?  I think it will make it a little clearer to users of savepoints that these bits of functionality are updated for 20.1, thanks to our new support for "real" savepoints.

PR #7222, opened 2020-04-27
missing backslash breaks the command

dbist:


dbist:
I'll provide a v19.2 patch soon

PR #7239, opened 2020-04-28
Index table column text; bump Algolia max record size; clean up comments (Fixes 7107)

yingzhuchin:
This change does a few things:
- Ask Algolia to start indexing text between \<td\> tags (doing this instead of \<table\> to avoid exceeding Algolia's record size restrictions)
- Bump our Algolia max record size from 10KB to 20KB, since we upgraded to a paid plan
- Clean up some existing code comments for better clarity

### DIFF
--- a/jekyll-algolia-dev/lib/jekyll-algolia.rb
+++ b/jekyll-algolia-dev/lib/jekyll-algolia.rb
@@ -26,7 +26,7 @@ module Jekyll
     # config - A hash of Jekyll config option (merge of _config.yml options and
     # options passed on the command line)
     #
-    # The gist of the plugin works by instanciating a Jekyll site,
+    # The gist of the plugin works by instantiating a Jekyll site,
     # monkey-patching its `write` method and building it.
     def self.init(config = {})
       # Monkey patch Jekyll and external plugins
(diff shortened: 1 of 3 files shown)


yingzhuchin:
@jseldess, happy to report that the test is successful, and we are now returning search results for text in formatted tables! See below for example of new behavior:

Screen Shot 2020-04-28 at 1 58 53 PM

yingzhuchin:
> @jseldess, happy to report that the test is successful, and we are now returning search results for text in formatted tables! See below for example of new behavior:
> 
> Screen Shot 2020-04-28 at 1 58 53 PM

Contrast this to behavior without current change:
Screen Shot 2020-04-28 at 3 10 15 PM

Before: 1 search result, no table text included
After: 2 search results, table text included

yingzhuchin:
One more example.

Before:
Screen Shot 2020-04-28 at 3 11 26 PM

After:
Screen Shot 2020-04-28 at 3 38 42 PM

The newly-included search result is also a page that pulls in another file from the `_includes` folder.

lnhsingh:
Nice job! 🎉

PR #7247, opened 2020-04-29
Add updated node capacity recommendations

### LINKED ISSUE #7242: Update recommended production settings re: per-node data size / number of ranges
We need to provide users with a recommendation about how much data can be stored per node (both data size, and number of ranges).

Per [this comment by @knz](https://github.com/cockroachdb/cockroach/issues/48085#issuecomment-621113588) we can say that:

> Today we have good reports from other users using less than 1TB per node or about 15000 replicas. It's possible to store more in a cluster by increasing the number of nodes. However, there may be some issues when the total number of ranges reaches 100k-200k or above.

We should update our [Recommended Production Settings](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html) to include this information.

Estimated scope of work:

- Verify which versions of CockroachDB this comment applies to
- Update production settings page (for each relevant version of CRDB) to include the above info

Related issues/PRs:

- Document production limits: #1830 (and an old but possibly still interesting PR: #1908)
- #5971, #7173

rmloveland:
Fixes #7242.

Summary of changes:

- Add a new subsection, 'Recommended capacity per node', that lays out
  recommendations for:

  - Max storage per node

  - Max # of ranges per node

- Link to the above from the 'Hardware > Storage' subsection

### DIFF
--- a/v20.1/recommended-production-settings.md
+++ b/v20.1/recommended-production-settings.md
@@ -47,7 +47,7 @@ Nodes should have sufficient CPU, RAM, network, and storage capacity to handle y
 
 - Avoid using shared storage such as NFS, CIFS, and CEPH storage.
 
-- For the best performance results, use SSD or NVMe devices. The recommended volume size is 300-500 GB.
+- For the best performance results, use SSD or NVMe devices. The recommended volume size is 300-500 GB. Some users have reported success with larger volumes of data per node; for more information, see [Recommended capacity per node](#recommended-capacity-per-node).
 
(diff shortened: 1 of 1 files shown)


rmloveland:
@knz please let me know if we should include more people as reviewers on this, and whom to ask.  There were a lot of opinions shared in the internal discussion.

knz:
I am not a good reviewer for this.
In fact I don't know anyone who has an informed authoritative opinion.

Deferring to @petermattis on this one.

knz:
I stand corrected it looks like @entin can help you here.
-- 

rmloveland:
Thanks @knz. I'm actually going to close this.  Based on the amount of complex internal discussions still ongoing, I think trying to drive this from a docs PR is "not even wrong".

Sorry for the noise.

PR #7259, opened 2020-05-01
Tweak savepoint language on upgrade finalization

rmloveland:
This change:

- Brings the language on the 'Upgrade CockroachDB' page in line with how
  we are talking about nested transactions elsewhere

- Corrects an erroneous reference to changing retry savepoint names for
  ORM usage, which was first made available in 19.1.


rmloveland:
Thanks Raphael!  Have a great weekend!

PR #7265, opened 2020-05-03
Fix typo on upgrade to 20.1 instructions

lbguilherme:
`Once all nodes have been upgraded to v20.1, any schema changes still running will stop making progress, [...]` -> `Until all nodes have been upgraded to v20.1, any schema changes still running will stop making progress, [...]`

Schema changes are disallowed during the rolling upgrade, not after it.


jseldess:
Thanks for this fix, @lbguilherme, and sorry for the delay. In the meantime, we clarified this language to say: "Once all nodes are running v20.1, but before the upgrade has been finalized."

PR #7273, opened 2020-05-04
Document that inverted indexes work on arrays

### LINKED ISSUE #6919: sql: add inverted indexes on arrays
PR: https://github.com/cockroachdb/cockroach/pull/45157

From release notes:
> The [inverted index](../v20.1/inverted-indexes.html) implementation now supports indexing [array](../v20.1/array.html) columns. This permits accelerating containment queries (`@>` and `<@`) on array columns by adding an index to them. [#45157][#45157] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #6919.

Summary of changes:

- Update the 'Inverted indexes' page to mention support for SQL `ARRAY`
  type.  Add links to array docs.

- Update the `ARRAY` data type reference to mention this as well.

- Update the 'Functions and operators' page to use anchors so we can
  link directly to the containment operators `>@` and `@>` being
  mentioned.

### DIFF
--- a/v20.1/inverted-indexes.md
+++ b/v20.1/inverted-indexes.md
@@ -1,17 +1,21 @@
 ---
 title: Inverted Indexes
-summary: Inverted indexes improve your database's performance and usefulness by helping SQL locate schemaless data in a JSONB column.
+summary: Inverted indexes improve your database's performance and usefulness by helping SQL locate data in JSONB or ARRAY columns.
 toc: true
 ---
 
-Inverted indexes improve your database's performance by helping SQL locate the schemaless data in a [`JSONB`](jsonb.html) column.
+Inverted indexes store mappings from values within a container column (such as a [`JSONB`](jsonb.html) document) to the row that holds that value. They are used to speed up containment searches, e.g., "show me all of the rows from this table which have a JSON column that contains the key-value pair `{"location":"NYC"}`". Inverted indexes are commonly used in [document retrieval systems](https://en.wikipedia.org/wiki/Document_retrieval).
+
+CockroachDB stores the contents of the following data types in inverted indexes:
+
+- [JSONB](jsonb.html)
+- <span class="version-tag">New in v20.1:</span> [Arrays](array.html)
 
 {{site.data.alerts.callout_success}}For a hands-on demonstration of using an inverted index to improve query performance on a <code>JSONB</code> column, see the <a href="demo-json-support.html">JSON tutorial</a>.{{site.data.alerts.end}}
 
-
 ## How do inverted indexes work?
 
-Standard [indexes](indexes.html) work well for searches based on prefixes of sorted data. However, schemaless data like [`JSONB`](jsonb.html) cannot be queried without a full table scan, since it does not adhere to ordinary value prefix comparison operators. `JSONB` needs to be indexed in a more detailed way than what a standard index provides. This is where inverted indexes prove useful.
+Standard [indexes](indexes.html) work well for searches based on prefixes of sorted data. However, data types like [`JSONB`](jsonb.html) or [arrays](array.html) cannot be queried without a full table scan, since they do not adhere to ordinary value prefix comparison operators. `JSONB` in particular needs to be indexed in a more detailed way than what a standard index provides. This is where inverted indexes prove useful.
 
 Inverted indexes filter on components of tokenizable data. The `JSONB` data type is built on two structures that can be tokenized:
 
@@ -48,11 +52,11 @@ An inverted index for this object would have an entry per component, mapping it 
 "cars" : "Honda"
 ~~~
 
-This lets you to search based on subcomponents.
+This lets you search based on subcomponents.
 
 ### Creation
 
-You can use inverted indexes to improve the performance of queries using `JSONB` columns. You can create them:
+You can use inverted indexes to improve the performance of queries using `JSONB` or `ARRAY` columns. You can create them:
 
 - At the same time as the table with the `INVERTED INDEX` clause of [`CREATE TABLE`](create-table.html#create-a-table-with-secondary-and-inverted-indexes).
 - For existing tables with [`CREATE INVERTED INDEX`](create-index.html).
@@ -64,7 +68,7 @@ You can use inverted indexes to improve the performance of queries using `JSONB`
 
 ### Selection
 
-If a query contains a filter against an indexed `JSONB` column that uses any of the supported operators, the inverted index is added to the set of index candidates.
+If a query contains a filter against an indexed `JSONB` or `ARRAY` column that uses any of the supported operators, the inverted index is added to the set of index candidates.
 
 Because each query can use only a single index, CockroachDB selects the index it calculates will scan the fewest rows (i.e., the fastest). For more detail, check out our blog post [Index Selection in CockroachDB](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/).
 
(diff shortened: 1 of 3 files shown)


v20.1/inverted-indexes.md
```diff
@@ -4,14 +4,18 @@ summary: Inverted indexes improve your database's performance and usefulness by
 toc: true
 ---
 
-Inverted indexes improve your database's performance by helping SQL locate the schemaless data in a [`JSONB`](jsonb.html) column.
+An inverted index stores a mapping from a value to its location in a schemaless data structure.  Inverted indexes are commonly used in [document retrieval systems](https://en.wikipedia.org/wiki/Document_retrieval).
```
jordanlewis, line 7: I don't think this is exactly correct. An inverted index accelerates containment searches into columns that are container types. I'm struggling to phrase it. But the key thing is that it specifically works for containment searches.

If you want to stick to the mapping phrasing, you could say that it stores a mapping from a value within a container column to the row that has that container, or something. That's not great either though!

I do think that "schemaless" should be removed from this though - I don't think it's all that accurate.

v20.1/inverted-indexes.md
```diff
@@ -4,14 +4,18 @@ summary: Inverted indexes improve your database's performance and usefulness by
...
 
+- [JSONB](jsonb.html)
+- <span class="version-tag">New in v20.1:</span> [Arrays](array.html)
+ 
+{{site.data.alerts.callout_success}}For a hands-on demonstration of using an inverted index to improve query performance on a <code>JSONB</code> column, see the <a href="demo-json-support.html">JSON tutorial</a>.{{site.data.alerts.end}}
 
 ## How do inverted indexes work?
 
-Standard [indexes](indexes.html) work well for searches based on prefixes of sorted data. However, schemaless data like [`JSONB`](jsonb.html) cannot be queried without a full table scan, since it does not adhere to ordinary value prefix comparison operators. `JSONB` needs to be indexed in a more detailed way than what a standard index provides. This is where inverted indexes prove useful.
+Standard [indexes](indexes.html) work well for searches based on prefixes of sorted data. However, schemaless data like [`JSONB`](jsonb.html) or [arrays](array.html) cannot be queried without a full table scan, since they do not adhere to ordinary value prefix comparison operators. `JSONB` in particular needs to be indexed in a more detailed way than what a standard index provides. This is where inverted indexes prove useful.
```
jordanlewis, line 18: I'm just not sure schemaless is the right term.

v20.1/inverted-indexes.md
```diff
@@ -4,14 +4,18 @@ summary: Inverted indexes improve your database's performance and usefulness by
...
 
 Inverted indexes filter on components of tokenizable data. The `JSONB` data type is built on two structures that can be tokenized:
```
jordanlewis, line 20: Tokenizable is better than schemaless...

v20.1/inverted-indexes.md
```diff
@@ -83,7 +87,10 @@ Tables are not locked during index creation thanks to CockroachDB's [schema chan
 Indexes create a trade-off: they greatly improve the speed of queries, but slightly slow down writes (because new values have to be copied and sorted). The first index you create has the largest impact, but additional indexes only introduce marginal overhead.
 
 ### Comparisons
-Currently, inverted indexes only support equality comparisons using the `=` operator. If you require comparisons using `>`, `<=`, etc., you can create an index on a computed column using your JSON payload, and then create a regular index on that. So if you wanted to write a query where the value of "foo" is greater than three, you would:
+
+#### JSONB
+
+Currently, inverted indexes on `JSONB` columns only support equality comparisons using the [`=` operator](functions-and-operators.html#operator-equals). If you require comparisons using [`<`](functions-and-operators.html#operator-less-than)), [`<=`](functions-and-operators.html#operator-less-than-or-equal), etc., you can create an index on a computed column using your JSON payload, and then create a regular index on that. So if you wanted to write a query where the value of "foo" is greater than three, you would:
```
jordanlewis, line 93: This is incorrect (and was before your changes) AFAICT. Only the @> variants (containment) are supported. You can't do a normal equality search using an inverted index at all.

v20.1/inverted-indexes.md
rmloveland: I tried rephrasing this intro part quite a bit using a simplified (?) version of your comment "it stores a mapping from a value within a container column ...".  PTAL.

v20.1/inverted-indexes.md
rmloveland: Updated all mentions of "schemaless" to use "tokenizable" or in some cases just say "data like JSON ..."

v20.1/inverted-indexes.md
rmloveland: Ack!  Updated to say

> Inverted indexes on `JSONB` columns support the following comparison operators:
> - "is contained by": [`<@`](functions-and-operators.html#operator-is-contained-by)
> - "contains": [`@>`](functions-and-operators.html#operator-contains)

v20.1/inverted-indexes.md
```diff
@@ -1,17 +1,21 @@
 ---
 title: Inverted Indexes
-summary: Inverted indexes improve your database's performance and usefulness by helping SQL locate schemaless data in a JSONB column.
+summary: Inverted indexes improve your database's performance and usefulness by helping SQL locate data in JSONB or ARRAY columns.
 toc: true
 ---
 
-Inverted indexes improve your database's performance by helping SQL locate the schemaless data in a [`JSONB`](jsonb.html) column.
+Inverted indexes store mappings from values within a column (such as a [`JSONB`](jsonb.html) document) to the row that holds that value.They are used to speed up containment searches, e.g., "show me all of the rows from this table which have a JSON column that contains the key-value pair `{"location":"NYC"}`". Inverted indexes are commonly used in [document retrieval systems](https://en.wikipedia.org/wiki/Document_retrieval).
```
jordanlewis, line 7: This is much better. I still think we need an adjective on `column` to show that it's not all columns. These columns contain tokens, or items. Container datatypes? Composite datatypes?

nit: `value.They` needs a space in between

v20.1/inverted-indexes.md
```diff
@@ -83,7 +87,15 @@ Tables are not locked during index creation thanks to CockroachDB's [schema chan
...
 
 ### Comparisons
-Currently, inverted indexes only support equality comparisons using the `=` operator. If you require comparisons using `>`, `<=`, etc., you can create an index on a computed column using your JSON payload, and then create a regular index on that. So if you wanted to write a query where the value of "foo" is greater than three, you would:
+
+#### JSONB
+
+Inverted indexes on `JSONB` columns support the following comparison operators:
+
+- "is contained by": [`<@`](functions-and-operators.html#operator-is-contained-by)
+- "contains": [`@>`](functions-and-operators.html#operator-contains)
```
jordanlewis, line 96: Crap, I actually forgot about a case. You can also use equality, but only when you've reached into the json document with the `->` operator. It's equivalent to using `@>`. Here's an example:

```
select * from a where j @>  '{"foo": "1"}';
```

This is equivalent to:

```
select * from a where j ->'foo' = '"1"';
```

I can provide more examples if you like.

@RaduBerinde are there more operators that are currently accelerated using inverted indexes?

RaduBerinde: No, as far as I can tell, only `@>` is (and we have rules that convert other expressions to use this operator when possible, like `j ->'foo' = '"1"'` to `j @> '{"foo": "1"}'`).

v20.1/inverted-indexes.md
rmloveland: Makes sense - I went with "container column" since it seems simplest of the available options - "composite" sounds a bit more complex and I'd rather not get too much into explaining these here in an overview (at least, for the scope of this change - I think this discussion points to a need to refactor this page to be more comprehensive at some point going forward)

Fixed the period spacing typo, thanks for that!

v20.1/inverted-indexes.md
rmloveland: Cool!  Updated this section to note the possibility of using `=` in this specific case, and to add both of these examples to show that it's really just `@>` again

rmloveland:
Hey @lnhsingh I'm gonna merge this now (since I'll be OOO until 25 May) but do you mind reviewing it and posting feedback, improvements etc. (if any)?  

I will get the emails with your comments, and will follow up with them after I'm back from vacation.

Want to make sure to get this info in front of 20.1 users, but realize it may need some polish.

PR #7292, opened 2020-05-06
GROUP BY supports grouping by aliases

### LINKED ISSUE #6193: opt: support grouping by aliases
PR: https://github.com/cockroachdb/cockroach/pull/42447

From release notes:
> It is now supported to specify selection target aliases as `GROUP BY` columns. Note that the `FROM` columns take precedence over the aliases, which are only used if there are no columns with those names in the current scope. [#42447][#42447] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #6193.

- Added note and example about grouping by aliases to the Select Clause page.


ericharmeling:
TFTR @lnhsingh !

PR #7293, opened 2020-05-06
Docs for non- vs. partitioned indexes and tables

### LINKED ISSUE #7031: sql: change messaging for non-partitioned indexes on partitioned table
PR: https://github.com/cockroachdb/cockroach/pull/46223

From release notes:
> Previously, when creating a non-partitioned index on a partitioned table with the `sql_safe_updates` [session variable](../v20.1/set-vars.html) set to `true`, CockroachDB would error out. CockroachDB now sends a NOTICE stating that creating a non-partitioned index on a partitioned table is not performant. [#46223][#46223] {% comment %}doc{% endcomment %}

rmloveland:
Fixes #7031.

Summary of changes:

- Update 'Define table partitions' to note that CockroachDB 20.1 sends a
  notice to the client warning about possible perf degradation when you
  create a non-partitioned index on a partitioned table.

### DIFF
--- a/v20.1/partitioning.md
+++ b/v20.1/partitioning.md
@@ -109,9 +109,11 @@ The primary key discussed above has two drawbacks:
 - It does not enforce that the identifier column is globally unique.
 - It does not provide fast lookups on the identifier.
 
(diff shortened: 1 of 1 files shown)


rmloveland:
Thanks, Oliver!

PR #6402, opened 2020-01-30
improve decommission node guidance

taroface:
Fixes #4915.

I took a small leap and added similar guidance for the "5-node cluster with 5-way replication for a specific table" example.

I did not create the new section heading (4-node cluster with 3-way replication) suggested in the issue, as I felt it would send a mixed signal in terms of how that section is conceptually organized.

Issue #4915, opened 2019-06-17
Decommission Nodes doc improvement


robert-s-lee:
Re: [Decommission Nodes](https://www.cockroachlabs.com/docs/v19.1/remove-nodes.html#3-node-cluster-with-3-way-replication)

## Issue Description

@ron pointed out decommission does not work on 3 node setup and @knz confirmed this behavior.  This is also pointed out in the doc.  

- on 3 node setup, the command starts but never completes.   

```
$ cockroach quit --decommission --insecure

  id | is_live | replicas | is_decommissioning | is_draining
+----+---------+----------+--------------------+-------------+
   1 |  true   |       21 |        true        |    false
(1 row)
..............
.......
```
- AdminUI marked the node as a suspect.

<img width="234" alt="Screen Shot 2019-06-17 at 9 05 16 AM" src="https://user-images.githubusercontent.com/6315124/59607556-8c748a80-90e1-11e9-8581-11dcf90e3022.png">

- once the cluster is in this state, when a 4th node is added, the suspect flag does not go away

```
$ cockroach node status --all --insecure
  id |      address      |  build  |            started_at            |            updated_at            | is_available | is_live | replicas_leaders | replicas_leaseholders | ranges | ranges_unavailable | ranges_underreplicated | live_bytes | key_bytes | value_bytes | intent_bytes | system_bytes | gossiped_replicas | is_decommissioning | is_draining
+----+-------------------+---------+----------------------------------+----------------------------------+--------------+---------+------------------+-----------------------+--------+--------------------+------------------------+------------+-----------+-------------+--------------+--------------+-------------------+--------------------+-------------+
   1 | rsl13.local:26257 | v19.1.1 | 2019-06-17 01:05:13.931882+00:00 | 2019-06-17 13:16:31.781711+00:00 |         true |    true |                0 |                     0 |      0 |                  0 |                      0 |   52233558 |   -854130 |    53087688 |            0 |            0 |                 0 |        true        |       false
   2 | rsl13.local:26259 | v19.1.1 | 2019-06-17 01:05:14.010574+00:00 | 2019-06-17 13:16:31.892999+00:00 |         true |    true |                7 |                     7 |     20 |                  0 |                      0 |  107981872 |    122926 |   107903922 |            0 |         9330 |                20 |       false        |       false
[...]

PR #7304, opened 2020-05-07
revise node decommissioning guidance and semantics

### LINKED ISSUE #6935: cli: deprecated `quit --decommission`
PR: https://github.com/cockroachdb/cockroach/pull/45903

From release notes:
> The `--decommission` flag for [`cockroach quit`](../v20.1/cockroach-quit.html) is now deprecated. It will be removed altogether in a next stable release. Deployments should use [`cockroach node decommission`](../v20.1/cockroach-node.html) followed by either `cockroach quit` or an equivalent form of server shut down. [#45903][#45903] {% comment %}doc{% endcomment %}

### LINKED ISSUE #7384: Better motivate usage of `recommissioning` nodes
Looking at https://cockroachlabs.com/docs//v20.1/remove-nodes.html, it's unclear to me what the real motivation is for using recommissioning. See this [internal](https://cockroachlabs.slack.com/archives/CA8H95NUU/p1590087247056500) thread for some discussion on the matter.

Specifically, it's unclear what's the benefit of recommissioning a fully decommissioned node, instead of starting afresh. I don't think there is any, aside from re-using NodeIDs, which I'd assume as an operator, I wouldn't much care for. Yet, our only [example](https://www.cockroachlabs.com/docs/stable/remove-nodes.html#recommission-nodes) for recommissioning a node is resurrecting a fully decommissioned node. I think the real motivation for having this recommissioning thing is to have an out of bound cancellation + undo mechanism for an ongoing decommissioning process.

### LINKED ISSUE #7422: release-20.1: misc drain/quit improvements
PR: https://github.com/cockroachdb/cockroach/pull/47692

From release notes:
> The default value of the parameter `--drain-wait` for [`cockroach quit`](../v20.1/cockroach-quit.html) has been increased from 1 minute to 10 minutes, to give more time for nodes with thousands of ranges to migrate their leases away. [#47692][#47692] {% comment %}doc{% endcomment %}
>
> The commands [`cockroach quit`](../v20.1/cockroach-quit.html) and `cockroach node drain` now report a "work remaining" metric on their standard error stream. The value reduces until it reaches `0` to indicate that the graceful shutdown has completed server-side. An operator can now rely on `cockroach node drain` to obtain confidence of a graceful shutdown prior to terminating the server process. [#47692][#47692] {% comment %}doc{% endcomment %}
>
> The time that [`cockroach quit`](../v20.1/cockroach-quit.html) waits client-side for the node to drain (remove existing clients and push range leases away) is now configurable via the command-line flag `--drain-wait`. Note that separate server-side timeouts also apply separately. Check the `server.shutdown.*` cluster settings for details. [#47692][#47692] {% comment %}doc{% endcomment %}
>
> It is now possible to drain a node without shutting down the process, using the `cockroach node drain` command. This makes it easier to integrate with service managers and orchestration: it now becomes safe to issue `cockroach node drain` and then separately stop the service via a process manager or orchestrator. Without this new mode, there is a risk to misconfigure the service manager to auto-restart the node after it shuts down via `quit`, in a way that's surprising or unwanted. The new command `node drain` also recognizes the new `--drain-wait` flag. [#47692][#47692] {% comment %}doc{% endcomment %}
>
> Added a new cluster setting, `server.shutdown.lease_transfer_wait`, that allows you to configure the server shutdown timeout period for transferring range leases to other nodes. Previously, the timeout period was not configurable and was set to 5 seconds, and the phase of server shutdown responsible for range lease transfers would give up after 10000 attempts of transferring replica leases away. The limit of 10000 attempts has been removed, so that now only the maximum duration `server.shutdown.lease_transfer_wait` applies. [#47692][#47692] {% comment %}doc{% endcomment %}
>
> [`cockroach quit`](../v20.1/cockroach-quit.html) now prints progress details on its standard error stream, even when `--logtostderr` is not specified. Scripts that wish to ignore this output can redirect the standard error stream. [#47692][#47692] {% comment %}doc{% endcomment %}

taroface:
Fixes #6935.
Fixes #7384.
Fixes #7422.

- Updated node decommissioning guidance to use `cockroach node decommission` over `cockroach quit --decommission` in 20.1, 19.2, 19.1.
- Introduced technical/semantic distinction between "decommissioning" and "decommissioned" nodes. Made some related corrections.
- Updated node decommissioning guidance to explicitly require stopping a node after decommissioning.
- Documented `cockroach node drain` and `--drain-wait` flag.
- Removed references to `cockroach quit` in 20.1 node decommissioning doc. Added `cockroach quit` deprecation announcement in 20.1.
- Corrected node recommissioning guidance.
- Added "restarting a decommissioned node" section.
- Updated & corrected how this is represented by the Admin UI.
- New Admin UI flows and images for 20.1.


_includes/v20.1/faq/planned-maintenance.md
```diff
@@ -1,4 +1,4 @@
...
+By default, if a node stays offline for more than 5 minutes, the cluster will consider it dead and will rebalance its data to other nodes. Before [temporarily stopping nodes](../../cockroach-quit.html) for planned maintenance (e.g., upgrading system software), if you expect any nodes to be offline for longer than 5 minutes, you can prevent the cluster from unnecessarily rebalancing data off the nodes by increasing the `server.time_until_store_dead` [cluster setting](cluster-settings.html) to match the estimated maintenance window.
 
 For example, let's say you want to maintain a group of servers, and the nodes running on the servers may be offline for up to 15 minutes as a result. Before shutting down the nodes, you would change the `server.time_until_store_dead` cluster setting as follows:
```
tbg, line 3: For the "resetting back to default" (further below, sorry, can't comment there) I think we'd want to advertise `RESET CLUSTER SETTING x` instead.

v19.1/admin-ui-cluster-overview-page.md
```diff
@@ -53,9 +53,9 @@ Down Since | How long the node has been down.
 
 ### Decommissioned Nodes
 
-Nodes that have been decommissioned for permanent removal from the cluster are listed in the **Decommissioned Nodes** table.
+Nodes that have been decommissioned for removal from the cluster are listed in the **Decommissioned Nodes** table.
 
-When you decommission a node, CockroachDB lets the node finish in-flight requests, rejects any new requests, and transfers all range replicas and range leases off the node so that it can be safely shut down. See [Remove Nodes](remove-nodes.html) for more information.
+When you initiate the [decommissioning process](remove-nodes.html#how-it-works) on a node, CockroachDB lets the node finish in-flight requests, rejects any new requests, and transfers all range replicas and range leases off the node so that it can be safely shut down.
```
tbg, line 58: Neither the old nor the new one are true. Decommissioning simply moves all data off the node (which includes moving all range leases). SQL sessions keep working forever, until the node is actually shut down. Shutting down the node gracefully will then let the node finish in-flight requests and drain clients etc.

v19.1/remove-nodes.md
```diff
@@ -1,26 +1,35 @@
...
 ---
 
-This page shows you how to decommission and permanently remove one or more nodes from a CockroachDB cluster. You might do this, for example, when downsizing a cluster or reacting to hardware failures.
+This page shows you how to decommission one or more nodes. Decommissioning a node removes it from the CockroachDB cluster.
 
-For information about temporarily stopping a node (e.g., for planned maintenance), see [Stop a Node](stop-a-node.html).
+You might do this, for example, when downsizing a cluster or reacting to hardware failures.
+
+{{site.data.alerts.callout_info}}
+In other scenarios, such as when [upgrading your cluster's version of CockroachDB](upgrade-cockroach-version.html) or performing planned maintenance (e.g., upgrading system software), you will want to temporarily [stop the node](stop-a-node.html) and restart it later.
```
tbg, line 12: Might wanna call out explicitly that that is *not* what decommissioning is for.

v19.1/remove-nodes.md
```diff
@@ -1,26 +1,35 @@
...
 
 ### How it works
 
-When you decommission a node, all range replicas on the node are transferred to other nodes.
+A node is considered to be decommissioned when it meets two criteria:
+
+1. The node has completed the decommissioning process.
+2. The node has been stopped and has not [updated its liveness record](cluster-setup-troubleshooting.html#node-liveness-issues) for 5 minutes.
+
+The decommissioning process transfers all range replicas on the node to other nodes. During and after this process, the node is considered "decommissioning" and continues to accept new SQL connections. Even without replicas, the node can still function as a gateway to route connections to relevant data. However, note that the [`/health?ready=1` monitoring endpoint](monitoring-and-alerting.html#health-ready-1) considers the node "unready" and returns a `503 Service Unavailable` status response code so load balancers stop directing traffic to the node. This behavior has been fixed in v20.1.
```
tbg, line 24: what has been fixed in v20.1?

v19.1/remove-nodes.md
```diff
@@ -1,26 +1,35 @@
...
 
-During and after decommissioning, the node continues to accept new SQL connections. Even without replicas, the node can still function as a gateway to route connections to relevant data. However, note that the [`/health?ready=1` monitoring endpoint](monitoring-and-alerting.html#health-ready-1) considers the node "unready" and returns a `503 Service Unavailable` status response code so load balancers stop directing traffic to the node. This behavior has been fixed in v20.1.
+After all range replicas have been transferred, it's typical to stop the node via a process manager or orchestration tool, or by sending `SIGTERM` manually. You can also use the [`cockroach quit`](stop-a-node.html) command to stop the node. When stopped, the node is drained of in-flight SQL connections and new SQL connections are rejected. At this point, the node is decommissioned.
```
tbg, line 26: Let's not mention `./cockroach quit` since we wish to get rid of it.

tbg: >  At this point, the node is decommissioned.

You say earlier that it needs to meet two criteria above, but it will take five minutes until it meets the second.

v19.1/remove-nodes.md
```diff
@@ -30,11 +39,15 @@ In this scenario, each range is replicated 3 times, with each replica on a diffe
...
 <div style="text-align: center;"><img src="{{ 'images/v19.1/decommission-scenario1.1.png' | relative_url }}" alt="Decommission Scenario 1" style="max-width:50%" /></div>
 
-If you try to decommission a node, the process will hang indefinitely because the cluster cannot move the decommissioned node's replicas to the other 2 nodes, which already have a replica of each range:
+If you try to decommission a node, the process will hang indefinitely because the cluster cannot move the decommissioning node's replicas to the other 2 nodes, which already have a replica of each range:
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/decommission-scenario1.2.png' | relative_url }}" alt="Decommission Scenario 1" style="max-width:50%" /></div>
 
-To successfully decommission a node, you need to first add a 4th node:
+{{site.data.alerts.callout_info}}
+Adding a 4th node to the cluster at this point will not enable the decommissioning process to complete. You can [recommission](#recommission-nodes) the node to return it to a healthy state.
```
tbg, line 47: Huh? It will, right? Did you want "now" instead of "not"?

tbg: You say the same thing just below so I'm confused

v19.1/remove-nodes.md
```diff
@@ -54,11 +67,15 @@ In this scenario, a [custom replication zone](configure-replication-zones.html#c
...
 
-If you try to decommission a node, the cluster will successfully rebalance all ranges but range 6. Since range 6 requires 5 replicas (based on the table-specific replication zone), and since CockroachDB will not allow more than a single replica of any range on a single node, the decommission process will hang indefinitely:
+If you try to decommission a node, the cluster will successfully rebalance all ranges but range 6. Since range 6 requires 5 replicas (based on the table-specific replication zone), and since CockroachDB will not allow more than a single replica of any range on a single node, the decommissioning process will hang indefinitely:
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/decommission-scenario3.2.png' | relative_url }}" alt="Decommission Scenario 1" style="max-width:50%" /></div>
 
-To successfully decommission a node, you need to first add a 6th node:
+{{site.data.alerts.callout_info}}
+Adding a 6th node to the cluster at this point will not enable the decommissioning process to complete. You can [recommission](#recommission-nodes) the node to return it to a healthy state.
```
tbg, line 75: Hmm same question here. Maybe I'm just misunderstanding how this renders (the HTML previews are 404)

v19.1/remove-nodes.md
```diff
@@ -69,39 +86,45 @@ To successfully decommission a node, you need to first add a 6th node:
...
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/before-decommission2.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/before-decommission1.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
-### Step 2. Decommission and remove the node
+### Step 3. Start the decommissioning process on the node
 
-SSH to the machine where the node is running and execute the [`cockroach quit`](stop-a-node.html) command with the `--decommission` flag and other required flags:
+SSH to any live node in the cluster and run the [`cockroach node decommission`](cockroach-node.html) command with the ID of the node to decommission:
```
tbg, line 115: This is a bit weird - we usually do SSH to the node because it's easiest from there to run commands against that node. But if you have the host, and it's reachable from wherever you are, then you can run the command from there instead via the `--host` flag. You specify that below, so it's weird that redundantly you ask users to ssh to a machine. I personally would focus on the `--host` version and avoid recommending to ssh to a machine that runs CRDB since that isn't even always possible (think k8s)

v19.1/remove-nodes.md
```diff
@@ -127,25 +150,47 @@ No more data reported on target nodes. Please verify cluster health before remov
...
 <div style="text-align: center;"><img src="{{ 'images/v19.1/after-decommission2.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
-Then view **Node List** on the **Overview** page and make sure all nodes but the one you removed are healthy (green):
+Then view **Node List** on the **Overview** page and make sure all nodes but the decommissioning node are healthy (green):
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/cluster-status-after-decommission1.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
+### Step 5. Stop the decommissioning node
+
+SSH to the machine running the node and execute the `cockroach quit` command:
```
tbg, line 169: Let's avoid cockroach quit.

v19.1/remove-nodes.md
```diff
@@ -127,25 +150,47 @@ No more data reported on target nodes. Please verify cluster health before remov
...
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/cluster-status-after-decommission2.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
-At this point, the node will no longer appear in timeseries graphs unless you are viewing a time range during which the node was live. However, it will never disappear from the **Decommissioned Nodes** list.
+At this point, the node is decommissioned and will no longer appear in timeseries graphs unless you are viewing a time range during which the node was live. However, it will never disappear from the **Decommissioned Nodes** list.
 
-Also, if the node is restarted, it will not accept any client connections, and the cluster will not rebalance any data to it; to make the cluster utilize the node again, you'd have to [recommission](#recommission-nodes) it.
+{{site.data.alerts.callout_info}}
+If you want to utilize a decommissioned node again, first [recommission](#recommission-nodes) the node to have the cluster rebalance data to the node. Then [restart](start-a-node.html) the node so that it accepts new SQL connections.
```
tbg, line 192: In the future this will be impossible unless the data dir is wiped (i.e. it counts as a new node). May want to incorporate this in the docs already (it doesn't hurt to do this today already)

v19.1/remove-nodes.md
```diff
@@ -154,32 +199,31 @@ Also, if the node is restarted, it will not accept any client connections, and t
...
 
 Open the Admin UI and select the **Node List** view. Note the ID of the node listed under **Dead Nodes**:
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/remove-dead-node1.png' | relative_url }}" alt="Decommission a single dead node" style="border:1px solid #eee;max-width:100%" /></div>
 
 ### Step 2. Mark the dead node as decommissioned
 
-SSH to any live node in the cluster and run the [`cockroach node decommission`](view-node-details.html) command with the ID of the node to officially decommission:
-
+SSH to any live node in the cluster and run the [`cockroach node decommission`](cockroach-node.html) command with the ID of the node to decommission:
```
tbg, line 214: Ditto

v19.1/remove-nodes.md
```diff
@@ -192,11 +236,15 @@ $ cockroach node decommission 4 --insecure --host=<address of live node>
...
+If you go back to the **Nodes List** page, in about 5 minutes, you'll see the node move from the **Dead Nodes** to the **Decommissioned Nodes** list. 
 
 <div style="text-align: center;"><img src="{{ 'images/v19.1/cluster-status-after-decommission2.png' | relative_url }}" alt="Decommission a single live node" style="border:1px solid #eee;max-width:100%" /></div>
 
-Also, if the node is ever restarted, it will not accept any client connections, and the cluster will not rebalance any data to it; to make the cluster utilize the node again, you'd have to [recommission](#recommission-nodes) it.
+At this point, the node is decommissioned and will no longer appear in timeseries graphs unless you are viewing a time range during which the node was live. However, it will never disappear from the **Decommissioned Nodes** list.
+
+{{site.data.alerts.callout_info}}
+If you want to utilize a decommissioned node again, first [recommission](#recommission-nodes) the node to have the cluster rebalance data to the node. Then [restart](start-a-node.html) the node so that it accepts new SQL connections.
```
tbg, line 246: Ditto

v19.1/remove-nodes.md
```diff
@@ -225,19 +277,19 @@ Select the **Replication** dashboard, and hover over the **Replicas per Store**
 
 ### Step 3. Decommission the nodes
 
-SSH to any live node in the cluster and run the [`cockroach node decommission`](view-node-details.html) command with the IDs of the nodes to officially decommission:
+SSH to any live node in the cluster and run the [`cockroach node decommission`](cockroach-node.html) command with the IDs of the nodes to decommission:
```
tbg, line 280: Ditto

v19.1/remove-nodes.md
```diff
@@ -263,27 +315,27 @@ Once the nodes have been fully decommissioned, you'll see a confirmation:
...
 
-<div style="text-align: center;"><img src="{{ 'images/v19.1/decommission-multiple7.png' | relative_url }}" alt="Decommission multiple nodes" style="border:1px solid #eee;max-width:100%" /></div>
-
-### Step 5. Remove the decommissioned nodes
-
-At this point, although the decommissioned nodes are live, the cluster will not rebalance any data to them, and the nodes will not accept any client connections. However, to officially remove the nodes from the cluster, you still need to stop them.
+### Step 5. Stop the decommissioning nodes
 
-For each decommissioned node, SSH to the machine running the node and execute the `cockroach quit` command:
+For each decommissioning node, SSH to the machine running the node and execute the `cockroach quit` command:
```
tbg, line 338: Ditto

v19.1/remove-nodes.md
```diff
@@ -299,36 +351,46 @@ $ cockroach quit --insecure --host=<address of decommissioned node>
...
 </div>
 
+In about 5 minutes, you'll see the nodes move to the **Decommissioned Nodes** list.
+
+<div style="text-align: center;"><img src="{{ 'images/v19.1/decommission-multiple7.png' | relative_url }}" alt="Decommission multiple nodes" style="border:1px solid #eee;max-width:100%" /></div>
+
+At this point, the nodes are decommissioned and will no longer appear in timeseries graphs unless you are viewing a time range during which the nodes were live. However, they will never disappear from the **Decommissioned Nodes** list.
+
+{{site.data.alerts.callout_info}}
+If you want to utilize a decommissioned node again, first [recommission](#recommission-nodes) the node to have the cluster rebalance data to the node. Then [restart](start-a-node.html) the node so that it accepts new SQL connections.
```
tbg, line 361: Ditto

v19.1/remove-nodes.md
```diff
@@ -394,6 +456,11 @@ $ cockroach node status --decommission --insecure --host=<address of any live no
 (5 rows)
 ~~~
 
+- `is_decommissioning` will be `true` for a node that is undergoing or has completed the decommissioning process. This indicates that replicas are being or have been transferred to other nodes.
+- `is_draining` will be `true` for a node that has been stopped. This indicates that SQL connections are no longer being accepted by the node.
```
tbg, line 460: Nodes may be stopped and `is_draining == false` as well. Just in case this may cause confusion.

v19.1/stop-a-node.md
```diff
@@ -4,9 +4,13 @@ summary: Learn how to temporarily stop a CockroachDB node.
 toc: true
 ---
 
-This page shows you how to use the `cockroach quit` [command](cockroach-commands.html) to temporarily stop a node that you plan to restart, for example, during the process of [upgrading your cluster's version of CockroachDB](upgrade-cockroach-version.html) or to perform planned maintenance (e.g., upgrading system software).
+This page shows you how to use the `cockroach quit` [command](cockroach-commands.html) to temporarily stop a node that you plan to restart.
```
tbg, line 7: Ditto about quit

tbg:
Looks good from a high level. I'm not sure I left the comments about cockroach quit in the right places (realized only later that there were multiple copies of the docs for the different versions), so please disregard them in that case - you are aware that we don't want to recommend quit in 20.1+ so if that's what the docs do that's great.

taroface:
TFTR, @tbg! I had one question for you above.

taroface:
@knz Tobias has reviewed the node decommissioning changes. 

Requesting your review as I've also folded your updates from cockroachdb/cockroach#47692 into this PR. In addition, I have updated the node recommissioning guidance as per [this thread](https://cockroachlabs.slack.com/archives/CA8H95NUU/p1590087247056500).

tbg:
> The 20.2 alpha behaves as you describe, so maybe the original doc was
incorrect on this. @jseldess <https://github.com/jseldess> do you know if
the currently documented behavior (decommissioning 1 node in a 3-node
cluster will hang, unless you recommission that node, add a 4th node, and
decommission again) was true of an earlier version? I'm able to simply add
the 4th node to complete the decommissioning.

At some point in the past, decommissioning multiple nodes at a time was a
terrible UX footgun because the nodes would immediately start draining
(prior to even being fully decommissioned) and so you could easily nuke the
cluster out from under you. I don't remember when we fixed this, but I
would think it had been fixed in 19.1 already.

jseldess:
Looks like that note was added to the 19.2 docs by @taroface: https://github.com/cockroachdb/docs/pull/6402 to address https://github.com/cockroachdb/docs/issues/4915. It looks like this behavior was reported by @robert-s-lee and confirmed by @knz. I'm not sure if @taroface or another writer tested this as well. I'll test across the last bunch of versions and report.

knz:
I'm looking at that other PR. In that PR I see two conflicting sentences:

>  If you try to decommission a node, the process will hang indefinitely because the cluster cannot move the decommissioning node's replicas to the other 2 nodes, which already have a replica of each range: 

This sentence is correct.

then the callout:

>  Adding a 4th node to the cluster at this point will neither enable the decommissioning process to complete nor change the **Suspect** node status. You can [recommission](#recommission-nodes) the node to return it to a healthy state. 

I believe this sentence is incorrect.

The thing is, I'm not sure. We have changed a few things in the code around this multiple times in the past. It may be that this sentence was correct at some point, and it is not any more (or vice-versa).

jseldess:
Yeah, in that issue, @robert-s-lee reported: `once the cluster is in this state, when a 4th node is added, the suspect flag does not go away`. I believe this is referring to the "Suspect" indication in the Admin UI. We may have misunderstood the issue here. I'll try to test soon.

jseldess:
On a side note, I'm not sure it make sense to categorize a decommissioning or decommissioned node as "suspect", which indicates something potentially fishy or unintentional.

knz:
> I'm not sure it make sense to categorize a decommissioning or decommissioned node as "suspect

I think it's because the UI only has 3 states "good" (green), "off" (red/absent) and  "yellow" (everything else)

It might make sense to add a new indicator color "node currently being removed" which will catch both decommissioning and nodes being shut down, and keep yellow for "suspect / don't know".

That's squarely UI design though, cc @dhartunian @piyush-singh

taroface:
> I think we may want to look for ways to clarify that, in the UI, a fully decommissioned node will continue to be shown as "suspect" until the node is stopped. That's counterintuitive; if we have a "Decommissioned Nodes" list, it should show nodes that have been decommissioned, whether they are live or not.

In the 20.1 UI, the Node List has a separate status for `DECOMMISSIONING` and `SUSPECT` (e.g., liveness not available and/or node was drained but is not decommissioning) nodes. However, the Node Status in the top module still conflates the two as "suspect", which is reflected in our [Cluster Overview docs](https://www.cockroachlabs.com/docs/v20.1/admin-ui-cluster-overview-page.html#cluster-overview-panel). 

image

Once a node is moved into "Recently Decommissioned Nodes" with a `DECOMMISSIONED` status (e.g., it is drained and shut down), it is no longer counted as "suspect" in the Node Status module.

Screen Shot 2020-06-15 at 12 31 21 PM

I think this doc should clarify that a `DECOMMISSIONING` node will be shown as "suspect" until it becomes `DECOMMISSIONED` and is moved to the list of Recently Decommissioned Nodes (or "Decommissioned Nodes" in earlier versions). (Will make that in the next update.)

v19.1/remove-nodes.md
jseldess: @taroface, do you have a docs issue to follow-up here? Looks like a relatively quick fix we should make, if I'm understanding.

taroface: Thanks! This is fixed in the v20.2 and v21.1 docs.

a-robinson: Thanks! Also, I mentioned it in the public Slack, but it seems like a mistake that the removal of support for running `recommission` on a decommissioned node wasn't mentioned in the v20.2.0 release notes: https://www.cockroachlabs.com/docs/releases/v20.2.0

The commit/PR that made the change had a "general change" release note attached to it: https://github.com/cockroachdb/cockroach/pull/50329

PR #7311, opened 2020-05-10
Document remaining 20.1 known limitations

### LINKED ISSUE #7277: Add known limitation for no index constraints with now()
As described in https://github.com/cockroachdb/cockroach/issues/26582 and https://github.com/cockroachdb/cockroach/issues/18836, the optimizer currently cannot constrain an index based on predicates with certain functions such as `now()`. This results in a full table scan for some queries. The workaround is to pass the correct date into the query as a parameter to a prepared query with a placeholder, which will allow the optimizer to constrain the index correctly. 

This should be documented as a known limitation for 20.1 and prior releases. Thanks!

### LINKED ISSUE #7323: Document remaining 20.1 known limitations
https://github.com/cockroachdb/cockroach/issues?q=is%3Aopen+is%3Aissue+label%3Adocs-known-limitation+-label%3Adocs-done

jseldess:
Fixes #7277.
Fixes #7323.

### DIFF
--- a/v20.1/known-limitations.md
+++ b/v20.1/known-limitations.md
@@ -8,6 +8,14 @@ This page describes newly identified limitations in the CockroachDB {{page.relea
 
 ## New limitations
 
+### Primary key changes and zone configs
+
+When you change a table's primary key with [`ALTER PRIMARY KEY`](alter-primary-key.html), any [zone configurations](configure-zone.html#create-a-replication-zone-for-a-table) for that table or its secondary indexes will no longer apply.
+
+As a workaround, recreate the zone configurations after changing the table's primary key.
+
+[Tracking Github Issue](https://github.com/cockroachdb/cockroach/issues/48254)
+
 ### `ROLLBACK TO SAVEPOINT` in high-priority transactions containing DDL
 
 Transactions with [priority `HIGH`](transactions.html#transaction-priorities) that contain DDL and `ROLLBACK TO SAVEPOINT` are not supported, as they could result in a deadlock. For example:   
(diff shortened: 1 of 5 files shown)


jseldess:
___
*[v20.1/known-limitations.md, line 102 at r1](https://reviewable.io/reviews/cockroachdb/docs/7311#-M6w3ENnAD2BJbhc6JTG:-M6w3ENo-sC_7SjVA5rG:bi4vugc) ([raw file](https://github.com/cockroachdb/docs/blob/abe5e96319259884a902766354446eec9b322a45/v20.1/known-limitations.md#L102)):*
> ```Markdown
> ~~~
> 
> As a workaround, pass the correct date into the query as a parameter to a prepared query with a placeholder, which will allow the optimizer to constrain the index correctly.
> ```

@rytaft, can you help me create an example demonstrating this workaround? I don't yet understand it. 

Once I have this, and your ok on the description above, I'll add this to early versions' known limitations as well.

jseldess:
@dt, please review the second commit, "Document dump limitation for collation".

rohany:
Jesse, can you add https://github.com/cockroachdb/cockroach/pull/48510

jseldess:
@RaduBerinde, please review the third commit, "Document set limitation with subquery".

jseldess:
@rohany, can you help me understand what I should add? Is there a related known limitation to document?

rohany:
Sorry, I should have linked the related issue (www.github.com/cockroach/cockroachdb/48254). The limitation is that indexes rebuilt because of a primary key change might lose the zone configurations applied to those indexes. Specifically, if we have a table `t (x int, y int, index i (y))` and we change the primary key to `x`, index i will lose any zone configs attached to it.

RaduBerinde:
"set limitation" commit LGTM

jseldess:
Thanks, @rohany. One more clarification: You say that `indexes rebuilt because of a primary key change might lose the zone configurations applied to those indexes`. "Might" is ambiguous. Do we know when this will and will not happen?

rohany:
The "might" was there because some indexes will appear to lose their zone configs, and some won't. I think we said in the alter primary key docs which indexes will get rebuilt as part of a primary key change, and all of the rebuilt indexes will lose their zone configs. The indexes that aren't rebuilt will keep their zone configs.

jseldess:
@rohany, please review the last commit.

v20.1/known-limitations.md
```diff
@@ -8,6 +8,14 @@ This page describes newly identified limitations in the CockroachDB {{page.relea
 
 ## New limitations
 
+### Primary key changes and zone configs
+
+If a table has a [zone configuration](configure-zone.html#create-a-replication-zone-for-a-table) and you change the table's primary key with [`ALTER PRIMARY KEY`](alter-primary-key.html), the zone configuration will no longer apply to the table.
```
rohany, line 13: I think this should be "if your indexes have a zone configuration ..."

jseldess: Reworded. PTAL.

rohany: LGTM thanks!

rohany:
Dump collate limitation lgtm

PR #7329, opened 2020-05-14
Flyway docs

### LINKED ISSUE #5356: sql: document Flyway
CockroachDB now works well with Flyway as of version 6.2.3.

Pre- 6.2.3 Flway:
Flyway creates one transaction per config file.  This can be problematic as CRDB doesn't always handle schema changes inside transactions well.

We should recommend that when using Flyway customers separate schema changes into separate config file.

Similarly, when using Flyway and services like K8s it can cause multiple current changes to occur at the sametime. We should recommend that CREATE TABLE (and other similar statements) should use IF NOT EXISTS to prevent this. Alternatively, you could do this in an init file in which the schema changes apply first before spinning up additional nodes. 

cc @BramGruneir for other points here

ericharmeling:
Fixes #5356. 

- Added a new page on using Flyway with CRDB. This page includes a simple tutorial, and some notes on transaction boundaries and retries.

### DIFF
--- a/_includes/sidebar-data-v20.1.json
+++ b/_includes/sidebar-data-v20.1.json
@@ -2093,6 +2093,12 @@
             ]
           },
           {
+            "title": "Flyway",
+            "urls": [
+              "/${VERSION}/flyway.html"
+            ]
+          },
+          {
             "title": "IntelliJ IDEA",
             "urls": [
               "/${VERSION}/intellij-idea.html"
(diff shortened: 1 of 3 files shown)


ericharmeling:
@vy-ton @awoods187

ericharmeling:
@awoods187 @vy-ton 
This is ready to merge from a technical point of view. Adding both of you to review to make sure we have the product messaging correct.

PR #7340, opened 2020-05-16
Move homebrew install instructions to 20.1 docs

jseldess:


v20.1/install-cockroachdb-mac.html
```diff
@@ -47,9 +47,9 @@ <h2>Use Homebrew</h2>
     </li>
   </ol>
 {{site.data.alerts.callout_info}}
-If you previously installed CockroachDB using a different method, you may need to remove the binary before you can run <code>brew install cockroach</code> or <code>brew upgrade cockroach</code>.
+If you previously installed CockroachDB via Homebrew before version v20.1, run <code>brew uninstall cockroach</code> before installing the new version. If you installed using a different method, you may need to remove the binary before installing via Homebrew.
```
bdarnell, line 50: This should say "before 19.2". We changed the way brew worked in that release so that you needed to uninstall first, but from 19.2 forward you can upgrade in place without uninstalling the old version.

jseldess: Whoops. Sorry about that. I'll fix that real quick.

PR #7368, opened 2020-05-20
Misc. BACKUP edits

### LINKED ISSUE #7112: backupccl: Full cluster restore has some zone config limitations
When restoring a Cluster backup, it first restores the data then restores the system tables. It restores the system table data as is. This means that it could restore Zone Configs that point to regions that don't have any active nodes in this new cluster.

For example, if your cluster backup has zone configs:
- `ALTER PARTITION europe_west OF INDEX movr.public.rides@primary CONFIGURE ZONE USING constraints = '[+region=europe-west1]';`
- `ALTER PARTITION us_east OF INDEX movr.public.rides@primary CONFIGURE ZONE USING constraints = '[+region=us-east1]';`
- `ALTER PARTITION us_west OF INDEX movr.public.rides@primary CONFIGURE ZONE USING constraints = '[+region=us-west1]';`

And the restored cluster doesn't have nodes with locality `region=us-west1`, the restored cluster will still have the zone config for `us-west1` for the `movr.public.rides@primary` index. In practice, the data won't be reshuffled to `us-west1` because that region doesn't exist. Instead, the data will be distributed as if that zone config doesn't exist.

This is counterintuitive in a sense--if a user tried to apply the `us-west1` config using `ALTER PARTITION` command it would fail, saying there are nodes with `region=us-west1`.

We should warn people in the docs that:
- We'll restore those zone configs, but that configs referencing regions/azs/etc. that don't exist will just be ignored and data will be distributed as normal
- To remediate, you can either (1) add nodes with that region, and the data will reshuffle [we should probably confirm this--we THINK this'll happen], or (2) remove the ignored zone configs and add new ones as required.

### LINKED ISSUE #7335: Incremental UX for locality aware backups example is unclear
https://www.cockroachlabs.com/docs/v20.1/backup-and-restore-advanced-options.html#create-an-incremental-locality-aware-backup

In this example, we show the syntax as `BACKUP TO 'gs://acme-co-backup/test-cluster' AS OF SYSTEM TIME '-10s' WITH revision_history;`

This is confusing as it doesn't have locality in there anywhere. We should add clarifications where:
- Example has locality flags in it
- Make it clear that you should try to include the same buckets every time (the only requirements we have is that you must always include the default bucket, but if say you included `default`, `us-west` and `us-east` in your base backup, and your next command was `default` and `us-west`, it'll work technically, but `us-east` data will now just be written to `default`. And on restore, you still have to specify EVERY bucket ever used, even if it was just for one incrmental in the chain)

lnhsingh:
- Clarify UX for locality-aware incremental backup / restore. Closes #7335.
- Add information about full cluster restore and zone configs. Closes #7112.

### DIFF
--- a/v20.1/backup-and-restore-advanced-options.md
+++ b/v20.1/backup-and-restore-advanced-options.md
@@ -131,20 +131,23 @@ Note that the list of URIs passed to [`RESTORE`][restore] may be different from 
 
 ### Create an incremental locality-aware backup
 
-To make an incremental locality-aware backup from a full locality-aware backup, the syntax is just like for [regular incremental backups](backup.html#create-incremental-backups):
+To create an incremental locality-aware backup from a full locality-aware backup, the syntax the same as it is for [regular incremental backups](backup.html#create-incremental-backups), i.e., if you backup to a destination already containing a full backup, an incremental backup will be appended to the full backup in a subdirectory:
 
 {% include copy-clipboard.html %}
 ~~~ sql
-> BACKUP TO \
-'gs://acme-co-backup/test-cluster' \
-AS OF SYSTEM TIME '-10s' WITH revision_history;
+> BACKUP TO
+	  ('s3://us-east-bucket?COCKROACH_LOCALITY=default', 's3://us-west-bucket?COCKROACH_LOCALITY=region%3Dus-west');
 ~~~
+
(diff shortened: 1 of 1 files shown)


lnhsingh:
@ericharmeling do you mind doing one last review for me since ryan is OOO?

Issue #7362, opened 2020-05-20
Docs state compatibility with Postgres 9.6 but our wire protocol advertises v9.5


vy-ton:
**Actual**
The [PostgreSQL Compatability page](https://www.cockroachlabs.com/docs/v20.1/postgresql-compatibility.html) states CockroachDB is compatible with version 9.6.

**Expected**
Our wire protocol actually advertises version 9.5
```
     flyway_1     | Database: jdbc:postgresql://cockroach:26257/defaultdb (PostgreSQL 9.5)
```

**Desired**
Docs should be updated to mention version 9.5

PR #7370, opened 2020-05-20
Change Postgres compatibility to version 9.5

vy-ton:
Addresses https://github.com/cockroachdb/docs/issues/7362


vy-ton:
@ericharmeling this should be a quick review

PR #7371, opened 2020-05-20
Document limitation when upgrading to v20.1.0

### LINKED ISSUE #7366: sql: dropping tables or databases on the 19.2 binary in a 19.2-20.1 mixed cluster orphans system.namespace2 entries
See https://github.com/cockroachdb/cockroach/issues/49092.

This will be solved in 20.1.1, but it's important to document for upgrades to 20.1.0.

jseldess:
Fixes #7366.

### DIFF
--- a/_includes/v20.1/known-limitations/dropping-renaming-during-upgrade.md
+++ b/_includes/v20.1/known-limitations/dropping-renaming-during-upgrade.md
@@ -1,0 +1,10 @@
+When upgrading from v19.2.x to v20.1.0, as soon as any node of the cluster has run v20.1.0, it is important to avoid dropping, renaming, or truncating tables, views, sequences, or databases on the v19.2 nodes. This is true even in cases where nodes were upgraded to v20.1.0 and then rolled back to v19.2.
+
+In this case, avoid running the following operations against v19.2 nodes:
+
+- [`DROP TABLE`](drop-table.html), [`TRUNCATE TABLE`](truncate.html), [`RENAME TABLE`](rename-table.html)
+- [`DROP VIEW`](drop-view.html)
+- [`DROP SEQUENCE`](drop-sequence.html), [`RENAME SEQUENCE`](rename-sequence.html)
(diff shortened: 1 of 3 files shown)


jseldess:
I've added emphasis, @lucy-zhang:

> If your cluster gets into this state, rolling all nodes back to v19.2 will not resolve the issue. Instead, you must follow the steps in this [known limitation](known-limitations.html#dropping-and-renaming-objects-during-an-upgrade-to-v20-1-0).

PTAL.

jseldess:
___
*[_includes/v20.1/known-limitations/dropping-renaming-during-upgrade.md, line 1 at r2](https://reviewable.io/reviews/cockroachdb/docs/7371#-M7ny3NE1XyIBoD3fUeG:-M7o1wN_COsC22HiOjjH:bt5ype0) ([raw file](https://github.com/cockroachdb/docs/blob/b0af9b1a80523a2012296b178e4a1e89f6ba9886/_includes/v20.1/known-limitations/dropping-renaming-during-upgrade.md#L1)):*
<details><summary><i>Previously, lucy-zhang (Lucy Zhang) wrote…</i></summary><blockquote>

Sorry, I should have been more clear. The problem is that once a 20.1.0 node has joined the cluster, the user can no longer safely drop/rename/truncate anything after that point (until they've finalized their upgrade), even if there are no longer any 20.1.0 nodes in the cluster. The problem is with the migration that the 20.1.0 node runs on startup. (This is what happened with the user that reported it: they tested with a single node running a 20.1.0 RC, then rolled back, then dropped a database on a pure 19.2 cluster, and still had this problem.)
</blockquote></details>

Just to make sure I understand: As soon as a node has run 20.1, whether it's still running 20.1 or has subsequently been rolled back to 19.2, performing drops/renames/etc against any node running 19.2 will cause this inconsistency, correct?

jseldess:
I'll remove the limitation from the 20.1 upgrade docs as soon as 20.1.1 is out.

PR #7381, opened 2020-05-21
Create Sample Apps section under Tutorials

jseldess:
With Hello World and Multi-Region sub-sections.

### DIFF
--- a/_includes/sidebar-data-v20.1.json
+++ b/_includes/sidebar-data-v20.1.json
@@ -112,85 +112,6 @@
             "urls": [
               "/${VERSION}/make-queries-fast.html"
             ]
-          },
-          {
-            "title": "'Hello World' Example Apps",
-              "items": [
-                  {
-                      "title": "Overview",
-                      "urls": [
-                          "/${VERSION}/hello-world-example-apps.html"
-                      ]
-                  },
-              {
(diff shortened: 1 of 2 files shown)


vy-ton:
Yes, we can do it in a follow up PR

jseldess:
___
*[v20.1/developer-guide-overview.md, line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/7381#-M7s31wa0UREWHsYphwg:-M7wWL-4E4CYh7D5__-U:b9vk9z3) ([raw file](https://github.com/cockroachdb/docs/blob/b2f2c3b6f6270c968997c0e9e075b15f890f945d/v20.1/developer-guide-overview.md#L22)):*

I'm unclear about exactly what this page should look like right now. Do you mean just leave the current https://www.cockroachlabs.com/docs/stable/hello-world-example-apps.html under Guides > Develop and rename it `Language Support Overview`? I like that idea, but I think it would involve renaming the file and restructuring/revising the page. Can you and Eric handle this in a follow-up PR, perhaps when you update the third-party tooling overview page?
</blockquote></details>

After more thought, the direction you're suggestion, which I think is great, does require a bunch more time and rework, so I'll merge this for now, and you and Eric can work out a solution.

PR #7392, opened 2020-05-24
Fix runtime cast error when using gorm

nadilas:


nadilas:
@rafiss I made your simplification and I found again the error:

```
errors.errorString{"pq: Could not complete operation in a failed transaction"}
```

nadilas:
Sorry, that took a little longer than expected 😄 

> That's strange. It works for me. What your versions for the following?
> 
> * CockroachDB
v20.1

> * lib/pq

github.com/lib/pq v1.5.2

> * jinzhu/gorm

github.com/jinzhu/gorm v1.9.12
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7392#-:-MA88laF8mHAFEMA-be6:b5sas4i)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nadilas)
> 
> _[_includes/v20.1/app/gorm-sample.go, line 150 at r2](https://reviewable.io/reviews/cockroachdb/docs/7392#-MA89-w1EXn9bQvF1cLZ:-MA89-w22D-bG99ZgYUs:b-pv1uix) ([raw file](https://github.com/cockroachdb/docs/blob/160ed2cdc66d34ec49d983726f7b2fbf9f2da0e2/_includes/v20.1/app/gorm-sample.go#L150)):_
> 
> > ```go
> > 			// handle retries appropriately.
> > 			pqErr, ok := err.(*pq.Error)
> > 			if !ok  {
> > ```
> 
> In order to preserve all the old logic, this should be:
> 
> ```
> if !ok || pqErr.Code != "40001" {
>   return err
> } else {
>   // Since this is a transaction retry error ...
>  ....
> }
> ```
> 
> _[_includes/v20.1/app/gorm-sample.go, line 172 at r2](https://reviewable.io/reviews/cockroachdb/docs/7392#-MA89YzX8xNEFKps_Ygd:-MA89YzX8xNEFKps_Yge:bnj7hl0) ([raw file](https://github.com/cockroachdb/docs/blob/160ed2cdc66d34ec49d983726f7b2fbf9f2da0e2/_includes/v20.1/app/gorm-sample.go#L172)):_
> 
> > ```go
> > 			if err := txn.Commit().Error; err != nil {
> > 				pqErr, ok := err.(*pq.Error)
> > 				if !ok {
> > ```
> 
> same as above

updated to preserve old logic

gilbsgilbs:
FWIW, newcomer to CockroachDB here and I also went through this. It was a little confusing to see a boilerplate from official docs crash like this, although not a hard fix.

_includes/v20.1/app/gorm-sample.go
```diff
@@ -146,7 +146,13 @@ func runTransaction(db *gorm.DB, fn txnFunc) error {
...
 			// handle retries appropriately.
-			pqErr := err.(*pq.Error)
+			pqErr, ok := err.(*pq.Error)
+			if !ok || pqErr.Code != "40001" {
+				// If it's not a retry error, it's some other sort of
+				// DB interaction error that needs to be handled by
+				// the caller.
+				return err
+			}
 			if pqErr.Code == "40001" {
```
gilbsgilbs, line 156: I don't think this conditional is useful as the previous one already catches all other cases.

gilbsgilbs:
I'd also add that there is no exponential back-off should the 40001 error occurs during the COMMIT statement. Not sure if this is intentional, but I think it's not.

nadilas:
> FWIW, newcomer to CockroachDB here and I also went through this. It was a little confusing to see a boilerplate from official docs crash like this, although not a hard fix.

@gilbsgilbs did you also have fun with the javascript boilerplate? Similar issue, but I haven’t gotten around to upload my fix yet.

_includes/v20.1/app/gorm-sample.go
nadilas: You are right, I should remove that wrapping

gilbsgilbs:
> did you also have fun with the javascript boilerplate?

Nope, haven't been that unlucky yet :smile: .

FYI, I ended up using roughly this implementation of `RunTransaction`:

```go
package main

import (
	"errors"
	"math"
	"math/rand"
	"time"

	"github.com/jinzhu/gorm"
	"github.com/lib/pq"
)


type TxnFunc func(*gorm.DB) error

var ErrMaxRetriesReached = errors.New(
	"maximum number of retries for CockroachDB transaction reached",
)

func shouldRetry(err error) bool {
	pqErr, isPqErr := err.(*pq.Error)
	return isPqErr && pqErr.Code == "40001"
}

func doRunTransaction(db *gorm.DB, fn TxnFunc) error {
	txn := db.Begin()

	if err := fn(txn); err != nil {
		txn.Rollback()
		return err
	}

	if err := txn.Commit().Error; err != nil {
		txn.Rollback()
		return err
	}

	return nil
}

func RunTransaction(db *gorm.DB, fn TxnFunc) error {
	for retryCount := 0; retryCount < 3; retryCount++ {
		err := doRunTransaction(db, fn)

		if err == nil {
			return nil
		}

		if !shouldRetry(err) {
			return err
		}

		sleepMs := math.Pow(2, float64(retryCount)) * 100 * (rand.Float64() + 0.5)
		time.Sleep(time.Duration(sleepMs) * time.Millisecond)
	}

	return ErrMaxRetriesReached
}
```

I think it is easier to follow than the one documented here and less error-prone. I was also able to write tests for my implementation (except for `Commit()` statement failures, which I'm unsure how to test).

rafiss:
the logic to do txn retries and error handling has been abstracted to https://github.com/cockroachdb/cockroach-go/blob/master/crdb/crdbgorm/gorm.go#L29 so we don't need this PR now.

thanks for bringing this to our attention though!

PR #7402, opened 2020-05-27
Change pg -> pq in Go example

sbward:
Go users can mistake "pg" to mean `go-pg` which is an ORM for Postgres. However, this article is about using "pq" which refers to the `lib/pq` SQL driver. To prevent ambiguity, this PR updates the article to use "pq" instead of "pg" throughout.


SWDevAngel:
@sbward Hi Sam. Thank you for contributing to CockroachDB this year. As a token of our appreciation we would like to send you a gift. Please DM me on our [community Slack](https://cockroa.ch/slack) `@lisa` so I can send you a link. (If you don’t want a gift, we also have a charitable contribution choice.) All orders need to be in by December 13, so please send this asap. :)

PR #7429, opened 2020-05-27
Misc. CDC edits

### LINKED ISSUE #5898: Changefeeds: add even more language about canceling changefeeds
The changefeeds have a nice Note about their cancellation behavior. Unfortunately this cancellation behavior has implication which extend to client driver usage which isn't always obvious. In particular, when users of client drivers encounter errors for a stream of rows, they tend to call rows.Close (e.g. https://golang.org/pkg/database/sql/#Rows.Close). Unfortunately, the PG wire protocol requires that the rows be consumed before the connection is again usable. In the case of a changefeed the rows are never consumed. It's critical that users instead close the connection (i.e. https://golang.org/pkg/database/sql/#Conn.Close). Otherwise the application will block forever on `Rows.Close()`.

cc @chrisseto  

 https://www.cockroachlabs.com/docs/stable/changefeed-for.html

### LINKED ISSUE #6700: CDC: Add examples with resolved timestamps
Update changefeed examples to include `resolved` timestamps for both kafka / cloud storage sink (possibly core changefeeds too).

cc: @mwang1026

### LINKED ISSUE #7105: Clarify that CDC / changefeeds are not intended to be a low latency pub sub mechanism
Our [CDC docs][1] say that it

> provides efficient, distributed, row-level change feeds into a configurable sink for downstream processing such as reporting, caching, or full-text indexing.

Although it's (perhaps?) implied by the text above, we should say explicitly that

> CDC is not a low latency pub sub mechanism
> CDC is for high-latency data export from the database to a data warehouse

This came up via internal discussion, so it's possible that some users will need this clarified / made more explicit as well.

[1]: https://www.cockroachlabs.com/docs/dev/change-data-capture.html

lnhsingh:
- Add information about canceling core changefeeds. Closes #5898.
- Add note that CDC is not a low-latency pub sub mechanism. Closes #7105.
- Updated CDC examples to include `WITH resolved`. Closes #6700.


lnhsingh:
hey @ajwerner, can you take a look? thanks!

lnhsingh:
@taroface ping on a review :)

PR #7437, opened 2020-05-28
Add script to handle admin tasks re: version bumps

rmloveland:
With each new major release of CockroachDB, there are a number of
administrative tasks that need to be accomplished on the docs website.
This includes, but is not limited to:

- Create a new `vXY.Z` folder, and copy content to it

- Copy includes to a new `_includes/vXY.Z` folder

- Copy images to a new versioned folder, and edit image link text in the
  referencing documents as needed to point to the right places

- Update the YAML config files with the new version info

- Create a new sidebar JSON file

- In the newly created XY.Z folder, remove the various "new in vXY.Z-1"
  callouts from the copied files.

Note: this list may not be exhaustive, see the code for details.

This script attempts to automate these tasks as much as possible.  In
addition to copying files around, editing files, etc., it also searches
for references to the previous version vX.Y.Z-1 in the newly copied
files that now serve as the documentation for vX.Y.Z, and shows those
references to the end user (as grep-like output), so they can be
edited (or not) as needed.

For example, to do all of the admin tasks around bumping the version of
CockroachDB from 20.1 to 20.2, and then surfacing the places where
additional edits may be needed, run:

    $ ./bump_major_version.py -d /path/to/docs -f 20.1 -t 20.2

### DIFF
--- a/scripts/bump_major_version.py
+++ b/scripts/bump_major_version.py
@@ -1,0 +1,322 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import argparse
+import shutil
+import yaml
+
+# With each new major release of CockroachDB, there are a number of
+# administrative tasks that need to be accomplished on the docs website.  This
+# includes, but is not limited to:
+
+# - Create a new `vXY.Z` folder, and copy content to it
+#
+# - Copy includes to a new `_includes/vXY.Z` folder
+#
+# - Copy images to a new versioned folder, and edit image link text in the
+# referencing documents as needed to point to the right places
+#
+# - Update the YAML config files with the new version info
+#
+# - Create a new sidebar JSON file
+#
+# - In the newly created XY.Z folder, remove the various "new in vXY.Z-1"
+# callouts from the copied files.
+#
+# Note: this list may not be exhaustive, see the code below for details.
+#
+# This script attempts to automate these tasks as much as possible.
+#
+# For example, to do all of the admin tasks around bumping the version of
+# CockroachDB from 20.1 to 20.2, and then surfacing the places where additional
+# edits may be needed, run:
+#
+# $ ./bump_major_version.py -d /path/to/docs -f 20.1 -t 20.2
+#
+# Note that running this command will generate a lot of output, since
+# it attempts to grep for the "old-versionish" things in the docs that
+# have been copied over into the new version.  These potential
+# references to the previous version will need to be manually reviewed
+# and edited by a human.
+#
+# Finally, note that this script is *only* for use when CockroachDB
+# does a major version release, since that is the only time we would
+# ever create a new version folder and copy files into it.  That means
+# this script would be used for a 20.1 -> 20.2 transition, but *not*
+# for a 20.1.1 to 20.1.2 transition.
+
+parser = argparse.ArgumentParser(
+    description="do admin tasks related to version bumps"
+)
+
+parser.add_argument("-f", "--from-version",
+                    required=True,
+                    help="previous version of CockroachDB")
+
+parser.add_argument("-t", "--to-version",
+                    required=True,
+                    help="next version of CockroachDB")
+
+parser.add_argument("-d", "--docs-directory",
+                    default=os.environ.get("COCKROACH_DOCS_REPO"),
+                    help="directory where you store 'cockroachdb/docs'")
+
+args = parser.parse_args()
+
+old_version = args.from_version
+new_version = args.to_version
+docs_directory = args.docs_directory
+
+# Every file we touch that has content in it will also be stored here.  At the
+# end we will iterate over the lines of each file and do some simple-minded
+# transformations (mostly regexes to check and bump version numbers, and remove
+# "New in vXX.1" tags).
+files_to_check = {}
+
+
+def version_string_name(version):
+    # Int -> String
+    return 'v' + str(version)
+
+
+old_version_name = version_string_name(old_version)
+new_version_name = version_string_name(new_version)
+
+
+def copy_files(type=None):
+    # Int Int -> IO! State!
+
+    marktree_p = False
+
+    if type == 'images':
+        subdirectory = 'images'
+    elif type == 'includes':
+        subdirectory = '_includes'
+        marktree_p = True
+    # This option is provided for clarity, but is not really needed.
+    elif type == 'docs':
+        subdirectory = ''
+        marktree_p = True
+    else:
+        subdirectory = ''
+        marktree_p = True
+
+    src_dir = os.path.join(docs_directory,
+                           subdirectory, old_version_name)
+    dest_dir = os.path.join(docs_directory,
+                            subdirectory, new_version_name)
+
+    if os.path.exists(dest_dir):
+        shutil.rmtree(dest_dir)
+
+    shutil.copytree(src_dir, dest_dir)
+
+    if marktree_p:
+        # Traverse down the directory, adding each to our list of
+        # content files to check later for references to the old
+        # version of CockroachDB.
+        marktree(dest_dir)
+
+
+def marktree(path):
+    # Path -> State!
+
+    # Recursively add the files in this subtree to our list of files to check
+    # later for mentions of the previous version number.
+    for dirpath, dirnames, filenames in os.walk(path):
+        for file in filenames:
+            f = os.path.join(dirpath, file)
+            if f not in files_to_check:
+                files_to_check[f] = 1
+
+
+def copy_sidebar_json():
+    # Int Int -> IO! State!
+
+    # Do it once for CockroachDB
+    src_json = os.path.join(docs_directory,
+                            '_includes',
+                            'sidebar-data-' + old_version_name + '.json')
+    dest_json = os.path.join(docs_directory,
+                             '_includes',
+                             'sidebar-data-' + new_version_name + '.json')
+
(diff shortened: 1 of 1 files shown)


rmloveland:
This was used to generate the changes in the first commit to https://github.com/cockroachdb/docs/pull/7439

rmloveland:
YZ, I figured you were the right person to ask if this is sane?  Happy to make any changes you would like to see.

yingzhuchin:
I'll take a look! First thought: I believe we also have to bump the version targeted by Vale, like so: https://github.com/cockroachdb/docs/pull/7451

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+    if os.path.exists(dest_dir):
+        shutil.rmtree(dest_dir)
+    shutil.copytree(src_dir, dest_dir)
+
+    # Traverse down the directory containing the "include files", adding each
+    # to our list of files to check later for references to the old version.
+    marktree(dest_dir)
+
+
+def copy_images(from_version, to_version):
```
yingzhuchin, line 121: Since `copy_files`, `copy_includes`, and `copy_images` are essentially performing the same steps, maybe you can condense them into one function with a third parameter passed along to modify `src_dir` and `dest_dir` as necessary.

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+                    required=True,
+                    help="next version of CockroachDB")
+
+parser.add_argument("-d", "--docs-directory",
+                    default=os.environ.get("COCKROACH_DOCS_REPO"),
+                    help="directory where you store 'cockroachdb/docs'")
+
+args = parser.parse_args()
+
+from_version = args.from_version
```
yingzhuchin, line 59: Could you call `version_string_name` here from parsed args, so that `from_version` and `to_version` are strings instead of ints? That way you don't have to repeatedly call `version_string_name` later on within functions.

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+    # Write the file out again
+    with open(file, 'w') as f:
+        f.write(yaml.dump(config))
+
+
+def copy_releases_page(from_version, to_version):
+    # Int Int -> IO! State!
+    old_releases_page = os.path.join(docs_directory, 'releases',
+                                     version_string_name(from_version) +
+                                     '.0.md')
```
yingzhuchin, line 300: Maybe a dumb question: Do we never do something like `from_version` "v19.2.6" --> `to_version` "v19.2.7"?

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+
+    config = yaml.load(filedata, Loader=yaml.BaseLoader)
+
+    new_version_name = version_string_name(to_version)
+
+    newdata = '''
+    name: {}
+    version: {}
+    docker_image: cockroachdb/cockroach-unstable
+    build_time: 2020/05/12 11:00:26 (go1.13.4)
```
yingzhuchin, line 258: Does this mean the script has to be updated on each use with similar hardcoded values?

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+    new_in_vXXX_string_01 = '<span class="version-tag">New in ' + \
+        version_string_name(version) + '</span>:'
+    new_in_vXXX_string_02 = '<span class="version-tag">New in ' + \
+        version_string_name(version) + '</span>'
+    changed_in_vXXX_string_00 = '<span class="version-tag">Changed in ' + \
+        version_string_name(version) + '</span>'
+    changed_in_vXXX_string_01 = '<span class="version-tag">Changed in ' + \
+        version_string_name(version) + ':</span>'
+
+    # Read in the file
```
yingzhuchin, line 178: Since we repeat this chunk and the "Write the file out again" chunk quite a bit, they might make sense as standalone functions to be called.

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+# addition to copying files around, editing files, etc., it also searches for
+# references to the previous version vX.Y.Z-1 in the newly copied files that
+# now serve as the documentation for vX.Y.Z, and shows those references to the
+# end user (as grep-like output), so they can be edited (or not) as needed.
+#
+# For example, to do all of the admin tasks around bumping the version of
+# CockroachDB from 20.1 to 20.2, and then surfacing the places where additional
+# edits may be needed, run:
+#
+# $ ./bump_major_version.py -d /path/to/docs -f 20.1 -t 20.2
```
yingzhuchin, line 39: Might be useful to document here that the script outputs a list of "old_versionish_things" for human consumption?

scripts/bump_major_version.py
rmloveland: I don't think so.  At least the `docker_image` field is always `cockroachdb/cockroach-unstable` at a time when this script would be run (early in a new version's release cycle).

Re: the `build_time` field, it looks like it has been getting copied forward from one release to the next in the YAML config for a while.  I don't know what uses it, but apparently it has been Fine So Far (tm).

(... until it's not? I have not seen any breakage from it in the past - famous last words)

scripts/bump_major_version.py
rmloveland: Good question! We should never need to do that.  This script is only appropriate for use in the case when we do a major version bump, e.g. 20.1 -> 20.2.  This is because a new major release is the only time when we need to create a new major version folder and copy a bunch of stuff into it.

The case you mention where 'v19.2.6' -> 'v19.2.7' involves a human tech writer editing the YAML file during a point release.

But I take this convo to mean that this script needs additional docs, so I will work on that.

rmloveland:
TFTR, YZ!  I tried to answer a couple questions between meetings (Tuesday = the heavy meeting day).  Will work on the other changes you suggested ASAP.

scripts/bump_major_version.py
yingzhuchin: Got it, thanks!

scripts/bump_major_version.py
```diff
@@ -0,0 +1,335 @@
...
+# "New in vXX.1" tags).
+files_to_check = {}
+
+
+def version_string_name(version):
+    # Int -> String
+    return 'v' + str(version)
+
+
+def make_new_version_folder(version):
```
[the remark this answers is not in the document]
rmloveland: Artifact of earlier development, I think!  No longer needed, deleted entirely.

scripts/bump_major_version.py
rmloveland: Done. Added a flag to `copy_files` that determines whether it's "regular" docs files, included files, or images that need to be dealt with.  Not sure if the arg-passing style is correct cuz I'm not really a Python-knower.  PTAL.

scripts/bump_major_version.py
rmloveland: Done.  There are now 2 effectively global vars, `old_version_name`, and `new_version_name`, that are created here and used everywhere instead.  They are not passed in as function arguments, so most functions take 0 or 1 args now, e.g., `copy_files(from_version, to_version)` becomes `copy_files()`, and so on.

scripts/bump_major_version.py
rmloveland: Added some more (and hopefully, better) notes in the introductory doc/comment that I think answer these questions.  PTAL.

scripts/bump_major_version.py
rmloveland: Have added a bit more explanation of that here.  Hope it's clearer? PTAL.

scripts/bump_major_version.py
```diff
@@ -0,0 +1,318 @@
...
+
+    marktree_p = False
+
+    if type == 'images':
+        subdirectory = 'images'
+    elif type == 'includes':
+        subdirectory = '_includes'
+        marktree_p = True
+    # This option is provided for clarity, but is not really needed.
+    elif type == 'docs':
```
yingzhuchin, line 98: Do we need to set `marktree_p` to true here as well?

scripts/bump_major_version.py
```diff
@@ -0,0 +1,318 @@
...
+        # We skip files that are not text, such as tarballs
+        return
+
+    # Replace the target string
+    filedata = filedata.replace(new_in_vXXX_string_00, '')
+    filedata = filedata.replace(new_in_vXXX_string_01, '')
+    filedata = filedata.replace(new_in_vXXX_string_02, '')
+    filedata = filedata.replace(changed_in_vXXX_string_00, '')
+    filedata = filedata.replace(changed_in_vXXX_string_01, '')
+
```
yingzhuchin, line 206: I think we need to `write_file` here

scripts/bump_major_version.py
yingzhuchin: Looks good, thank you!

yingzhuchin:
Other than the two new comments, everything looks good!

scripts/bump_major_version.py
rmloveland: Yes, thanks!  Here and in the default `else` case.  Scary working without tests - Thanks!!!

scripts/bump_major_version.py
rmloveland: Thank you!  Again, so scary lol

PR #7440, opened 2020-05-28
Stop building training and detailed SQL support pages

### LINKED ISSUE #4090: Update Detailed SQL Support doc with latest information as of v19.2
Relates to #1340 .

Estimated scope of work:

1. Update [Detailed SQL standard comparison](https://www.cockroachlabs.com/docs/stable/detailed-sql-support.html) to match the current reality
2. See what other work (if any) falls out of the above, and either do it or file issues/plan it for later

### LINKED ISSUE #7432: Stop building old training pages
People keep finding our old training pages, and we should delete them.  AFAIK nobody is maintaining or updating them anymore.

I did find [a PR to remove them from search results](https://github.com/cockroachdb/docs/pull/7409) but I think we should delete them altogether instead.

Since they are in our git history, we can always resurrect anything we need in there later.

But as usual, perhaps there is something I don't understand here. :-)  Do we keep these old training pages around because something important links to them?

jseldess:
- Added new top-level /archived directory and adjusted the
  configs to exclude files in that directory from the build.
- Moved the latest versions of training pages to the
  /archived directory for potential future reuse, etc.
  Deleted all other versions.
- Moved the latest version of the details SQL support page
  to the /archived directory. Deleted all other versions.
- Removed any remaining links to the archived pages and made
  cleaned up other site files.

Fixes #7432.
Fixes #4090.


jseldess:
TFTR, @rmloveland!

Issue #7443, opened 2020-05-28
Update node density docs


johnrk:
With @ajwerner 's larger range size change and testing ([report](https://docs.google.com/document/d/19ystASmq_14B9zzzsn0tQTu6LCahwh6bFyBfqjHEfS8/edit#))in 20.1, we can handle more storage per node. We should update our [production checklist docs](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html) to maintain one place for this type of info. Let's document four things:

1. Range size went from 64MiB to 512MiB in 20.1.
2. Users should do their own testing to arrive at their optimal storage per node because it varies by workload.
3. With our own internal testing using the [bank workload](https://www.cockroachlabs.com/docs/stable/cockroach-workload.html#bank-workload), we arrived at successfully storing 4.32 TiB logical data per node. 
4. Configuration and full results of our applying bank workload:

- 6x c5d.4xlarge nodes, gp2 EBS with 5TiB each

Results (could present it in a table):

- vCPU Per Node: 16

- Logical Data Per Node: 4.32 TiB

- Ram Per Node: 32 GiB

- Data Per Core: ~270 GiB / vCPU

-  Data Per Ram: ~135 GiB / GiB

PR #7173, opened 2020-04-16
Update default range size from 64 to 512 MiB

rmloveland:
Fixes #5971.

Summary of changes:

- Update all mentions of the default range size to the new value in
  CockroachDB v20.1: 512 MiB.

- Update 'Known Limitations' to mention that a single SQL statement can
  perform at most $new_range_size worth of updates.

Issue #5971, opened 2019-11-26
Increase maximum storage per node


jseldess:
Background: https://airtable.com/tblD3oZPLJgGhCmch/viw1DKmbKhg2MIECH/recO3yD744y8oNI3Y

Description: In 19.2, CockroachDB default range size is 64MiB and falls over at roughly 1TB per node. With this change, CockroachDB has a default range size that's higher than 64MiB and stores up to 10TB per node.

--

Customer Problem: Cost

The problem that his feature addresses is cost for customers with on-prem hosts that can store significantly more than 1TB. When running CRDB, they are forced to add more nodes to their deployments than necessary if Cockroach could support more data per node. Ultimately, customers have to pay more than necessary because they have to pay for more nodes.

Team: John Kendall, Andrew Werner, Andrew Kryczka

Github Tracking Issue: https://github.com/cockroachdb/cockroach/pull/31663 https://github.com/cockroachdb/docs/issues/5971

rmloveland:
Estimated scope of work:

- Review all mentions of default range size in docs
- Revise mentions to use an include wherever possible
- Update everything to note the new default range size, total cluster data size enabled by new limit
- Update production deployment docs re: storage, as/if needed

jseldess:
From release notes:

> New clusters will have a larger default range size of 512 MB, which will result in fewer ranges for the same amount of data. [#45209][#45209]

rmloveland:
@ajwerner as Jesse noted, the recent release note says 

> New clusters will have a larger default range size of 512 MB, which will result in fewer ranges for the same amount of data. [#45209][#45209]

Do we have updated guidance on the maximum amount of data to store in a cluster as well?  I note the wording of this release note makes sure to avoid mentioning that. :-)

rmloveland:
Update: i have been informed by Andrew that this is an active area of investigation.  Standing by to await the results of that work before proceeding here

PR #7449, opened 2020-06-01
Update mentions of `range_{min,max}_size`

### LINKED ISSUE #7367: Wrong default value of range_min_bytes and range_max_bytes
Re: [Configure Replication Zones](https://cockroachlabs.com/docs//v20.1/configure-replication-zones.html)

## Issue Description
In CockroachDB v20.1, it seems that the default range size is changed from 64/16 MB to 512/128 MB.
https://github.com/cockroachdb/cockroach/commit/660b3e7e0d05d31ab1cf6ec001f6f5121bd20fa6

However, in the Replication zone variables section of configure-replication-zones.html, it seems that the docs states the default value of range_min_bytes is '`16777216` (16MiB)'.

And, docs states the default value of range_max_bytes is '`67108864` (512MiB)'. (512MiB is correct, but it looks like the bytes 67108864 is incorrect.)

## Suggested Resolution
I think that the docs should state the default values like the following.

- range_min_bytes is '`134217728` (128MiB)'.
- range_max_bytes is '`536870912` (512MiB)'.

rmloveland:
Fixes #7367.

Related to #5971, #7173, #7443.

Summary of changes:

- Update all mentions of `range_min_size` and `range_max_size` to use
  the new default values of `134217728` and `536870912`, respectively.

- Fix a dangling prose mention of the previous value in MiB.


rmloveland:
Thank you Eric!

PR #7451, opened 2020-06-01
Bump doc version examined by Vale

yingzhuchin:


yingzhuchin:
Good question -- I defer to Jesse =) My gut says we run Vale on whatever is the current stable version, since that's what most users interact with.

PR #7477, opened 2020-06-06
20.1 install linux build from source

cargillb:
@jseldess  

Used the wiki (https://wiki.crdb.io/wiki/spaces/CRDB/pages/181338446/Getting+and+building+from+source) as my primary source paired with an attempt to build from source on a fresh install of Ubuntu. 

Another option could be to reference the wiki at some point in the documentation but I felt like the route I took was more in line with the rest of the documentation.


jseldess:
Sorry for the delay, @cargillb. I don't believe those additional dependencies are required when building from the source archive downloaded from docs, whereas they are when building from source in our repo. @bdarnell, can you confirm my understanding?

PR #7439, opened 2020-05-28
Site updates for v20.2

rmloveland:
Fixes #7427.

Summary of changes:

- Create a new v20.2 folder, and copy content to it

- Copy includes to a new _includes/v20.2 folder

- Copy images to a new versioned folder, and edit image link text in the
  referencing documents as needed to point to the right places

- Update the YAML config files with the new version info

- Create a new sidebar JSON file

- In the newly created 20.2 folder, remove the various "new in v20.1"
  callouts from the copied files.

Note: this list may not be exhaustive, see the PR for details.

PR #7484, opened 2020-06-08
Release notes v20.2.0 alpha.1 and v20.1.2

### LINKED ISSUE #7426: 20.2.0-alpha.1 release notes
See also:

- cockroachdb/cockroach#49853
- #7427

rmloveland:
Fixes #7426 

Depends on #7439

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -53,8 +53,8 @@ release_info:
     name: v20.2.0
     version: v20.2.0-alpha.1
     docker_image: cockroachdb/cockroach-unstable
-    build_time: 2020/05/12 11:00:26 (go1.13.4)
-    start_time: 2020-05-12 11:01:26.34274101 +0000 UTC
+    build_time: 2020/06/17 11:00:26 (go1.13.4)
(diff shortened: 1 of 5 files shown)


rmloveland:
> LGTM, with one more thing that I forget to mention: should there be updates to the following files as well?

> _config_base.yml (updated build date under 20.2)

I have no idea what this value should be.  In the other big "add 20.2" PR, I think this was the same value.  It appears to be a "magic value" that got put in but doesn't actually do anything?

> _data/releases.yml (release date listed under 20.2 release in "Testing releases")

OK!  I will update this (and the date in the actual release note) once we get an official go/no-go on this release (which AFAICT we don't have yet).

jseldess:
I think the 20.2 date in https://github.com/cockroachdb/docs/blob/master/_config_base.yml#L56 should just be manually update to match the release date. Those are used in some `cockroach start` output examples, I think.

rmloveland:
FYI - made a PR against the `release-notes.py` script so it doesn't generate its own `docker pull` command, and instead uses the YAML template logic we already have that ensures the correct invocation: https://github.com/cockroachdb/cockroach/pull/50113

Issue #5029, opened 2019-07-09
Docs pages have no scrollbar


RaduBerinde:
At least in Chrome (linux), the docs pages have no scrollbar. This can be very annoying for people who use that instead of a touchpad or scroll wheel.

![image](https://user-images.githubusercontent.com/16544120/60902792-39a07580-a23e-11e9-8235-d54242605888.png)

jseldess:
Users in the community have mentioned this UX issue as well:

<img width="529" alt="Screen Shot 2020-02-03 at 10 46 54 AM" src="https://user-images.githubusercontent.com/15893490/73667634-8fbd8600-4672-11ea-9ed0-402444cbf528.png">

jseldess:
Another example:

> Lack of a scroll bar is really annoying. Especially when dealing with very long pages like this one, for example: https://www.cockroachlabs.com/docs

PR #6420, opened 2020-02-04
fix scrollbar on chrome

anagio:
Makes scrollbar visible on Chrome

Fixes issue
https://github.com/cockroachdb/docs/issues/5029

Issue #1696, opened 2017-07-11
Show the sidebar when javascript is disabled


knz:
Currently if the browser has javascript disabled, the docs side is mostly fine. Except that the sidebar is completely absent.

This is because the site starts empty and then only the sidebar is populated in javascript.

This is incorrect; instead the basic pages should already include the side bar, and a javascript program should merely overwrite with the dynamic replacement.

jseldess:
@mjibson, once we're on Hugo, I vaguely remember we might have a built-in, non-JS sidebar option, is that right?

mjibson:
Yes, we think hugo will support our current stuff natively.

luciaaaaaaa:
Could you provide a screenshot of the issue? which sidebar is missing

knz:
here's the doc site with missing sidebar, notice how the sidebar on the left is missing.

![image](https://user-images.githubusercontent.com/642886/82432397-4ca3e900-9a90-11ea-8fc0-a1abe28358c5.png)

knz:
As I explained at the top, the issue here is technical:

> This is incorrect; instead the basic pages should already include the side bar, and a javascript program should merely overwrite with the dynamic replacement.

jseldess:
We have closed this issue because it is more than 3 years
old. If this issue is still relevant, please add a comment and reopen
the issue. Thank you for your contribution to CockroachDB docs!

PR #7487, opened 2020-06-09
Fix scrollbar on Chrome

yingzhuchin:
Makes scrollbar visible on Chrome

Fixes issue #5029

### DIFF
--- a/css/components/_toc.scss
+++ b/css/components/_toc.scss
@@ -30,12 +30,20 @@
   padding-right: 5px;
 }
 
+.toc-col-contents::-webkit-scrollbar {
+  display: none;
+}
+
 #toc-right {
   max-height: 85vh;
   padding-right: 15px;
   background: #ffffff;
   overflow-y: auto;
   margin-bottom: 1.5rem;
+}
+
+#toc-right::-webkit-scrollbar {
+  display: none;
 }
 
 #toc,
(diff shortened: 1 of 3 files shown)


yingzhuchin:
Hi @jseldess, I looked at http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/e23402969be2c33d336bc8ab2f663769b9114ec8/stable/architecture/transaction-layer.html, taken from your previous example on https://github.com/cockroachdb/docs/pull/6420. I'm on Chrome Version 83.0.4103.97 and I see one scrollbar even when the TOC is long. Does that match what you're seeing?

jseldess:
@yingzhuchin, scrollbars are still not showing up correctly for me. In this screenshot, for example, you can see there are two scrollbars next to the toc and when you scroll the main content, you see a third one all the way to the right (whereas I'd expect it directly to the right of the main body...

Screen Shot 2020-06-09 at 5 48 11 PM

yingzhuchin:
Could you let me know which browser and what version? I'm unable to reproduce unfortunately.

jseldess:
Sure: Screen Shot 2020-06-09 at 6 11 17 PM

yingzhuchin:
Hi @jseldess, the latest preview shows scrollbars working correctly for TOCs. What do you think? 

Unfortunately the main scrollbar cannot be between the page content and the TOC at this point, because not all pages have TOCs, and any TOCs that are present are generated dynamically by Javascript. See #1696.

PR #7480, opened 2020-06-08
Updated Active Record support docs

ericharmeling:
Fixes #7472.

PR #7504, opened 2020-06-11
Port #7480 to v20.2

rmloveland:


rmloveland:
Eric, I think this is the last (only?) of your recent PRs that needed to be ported forward to 20.2.  Mind taking a look and making sure I did not miss anything from your original PR? (#7480)

PR #7508, opened 2020-06-13
IMPORT INTO edits

### LINKED ISSUE #7270: Docs Improvement: Import Into
Import Into: https://www.cockroachlabs.com/docs/v19.2/import-into.html

It's unclear from the top of this page how Import Into works. As I understand it now based on some help from the Bulk IO team, `Import Into` acts like an append statement allowing you to add **new** rows onto a table from a CSV. 

I misunderstood this and took away from the page that Import Into can be used to insert data into an **existing** column or columns. For example, if it did work this way, if I wanted to add a column and fill it with data I would alter the table to add the new column then use` import into` to insert the data from CSV. 

I would suggest changing this page so that the intent of `import into` is more clear.

### LINKED ISSUE #7278: Import Into docs should mention interaction with foreign keys
Relates to https://github.com/cockroachdb/docs/issues/7270.

`IMPORT INTO` invalidates all foreign keys on the target table. This is not mentioned in the docs, but it's tricky enough and important enough that it should be.

lnhsingh:
- Clarify that IMPORT INTO can't be used to insert data into existing row. Closes #7270
- Add note that IMPORT INTO invalidates foreign keys. Closes #7278.

### DIFF
--- a/v19.2/import-into.md
+++ b/v19.2/import-into.md
@@ -10,6 +10,7 @@ toc: true
 
 - `IMPORT INTO` only works for existing tables. For information on how to import data into new tables, see [`IMPORT`](import.html).
 - `IMPORT INTO` cannot be used within a [transaction](transactions.html) or during a [rolling upgrade](upgrade-cockroach-version.html).
+- `IMPORT INTO` invalidates all [foreign keys](foreign-key) on the target table. To validate the foreign key(s), use the [`VALIDATE CONSTRAINT`](validate-constraint.html) statement.
 - `IMPORT INTO` cannot be used to insert data into a column for an existing row. To do this, use [`INSERT`](insert.html).
 
 ## Required privileges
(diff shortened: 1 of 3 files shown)


v19.2/import-into.md
```diff
@@ -10,7 +10,7 @@ toc: true
 
 - `IMPORT INTO` only works for existing tables. For information on how to import data into new tables, see [`IMPORT`](import.html).
 - `IMPORT INTO` cannot be used within a [transaction](transactions.html) or during a [rolling upgrade](upgrade-cockroach-version.html).
-- `IMPORT INTO` invalidates all [foreign keys](foreign-key) on the target table. To validate the foreign key(s), use the [`VALIDATE CONSTRAINT`](validate-constraint.html) statement.
+- `IMPORT INTO` invalidates all [foreign keys](foreign-ke.html) on the target table. To validate the foreign key(s), use the [`VALIDATE CONSTRAINT`](validate-constraint.html) statement.
```
taroface: `foreign-key.html` :)

lnhsingh: fixed!

PR #7537, opened 2020-06-20
Split client SQL and internal GRPC ports

Amruta-Ranade:
Updated the `cockroach start` doc with information about the new flags. 

Open questions:
- Should we update the Networking guidance in the Production Checklist now or wait until we improve this feature.
- Should we include an example showing how to use the new flags? If yes, I need help coming up with an example using the new flags. I tried looking at the `addr_validation_test.go` file but it confused me more.


Amruta-Ranade:
Chat with Jesse: We need to clarify why the user needs this and when to use it?

knz:
>  We need to clarify why the user needs this and when to use it?


Here's an explanation that can do *for now*:

> """A general security best practice is to use network firewalling to isolate certain services from each other, or client apps from certain privileged sevices. This is extremely commonplace in corporate networks.  To integrate CockroachDB in such an environment, the network operator can separate the TCP addr/ports used by CockroachDB for node-to-node and cluster administration traffic, which is privileged, from the TCP addr/ports used for SQL clients, and also separately that used by Web UI operators."""

Here's an example set-up corresponding to this:

image

Amruta-Ranade:
@knz TFTR! Can you help me come up with an example?

knz:
The diagram is the example

Amruta-Ranade:
> The diagram is the example

That's what's confusing me :) I can't figure out the port numbers to enter in the CLI command to set up the configuration in the diagram.

knz:
Oh I see. Actually, despite the title of the doc issue, the feature allows you to separate _addresses_ and not just ports.

The example would be compellling by just using different addresses, you don't need to change the port numbers (the default ports are likely fine once the addresses are separeate)

Amruta-Ranade:
Update after call with Raphael: We should not include an example in the cockroach-start reference doc. Instead, we should make this a separate docs project to update the production checklist doc after we work on the user story a bit more.

PR #7539, opened 2020-06-22
CC edit cluster

Amruta-Ranade:


v20.1/cockroachcloud-cluster-management.md
```diff
@@ -32,11 +28,26 @@ For each node, the page displays the node's `Name`, nested under its region.
 
 From the **Cluster Overview** page, you can connect to your cluster. For more information, see [Connect to Your CockroachCloud Cluster](cockroachcloud-connect-to-your-cluster.html).
 
-## Manage cluster configuration
+## Add or remove nodes from a cluster
+
+You can add or remove nodes from your cluster through the Console.
+
+{{site.data.alerts.callout_info}}
+At this time, we do not support scaling up a single-node cluster or scaling down to a single-node cluster.
```
kannanlakshmi, line 36: Can we add "if you want to scale down to 3 nodes or scale up your 1 node cluster, please open a support ticket at support.cockroachlabs.com.

v20.1/cockroachcloud-cluster-management.md
```diff
@@ -32,11 +28,26 @@ For each node, the page displays the node's `Name`, nested under its region.
...
+
+You can add or remove nodes from your cluster through the Console.
+
+{{site.data.alerts.callout_info}}
+At this time, you cannot use the Console to scale up a single-node cluster or scale down to a single-node cluster. For single-node cluster changes, contact [Support](https://support.cockroachlabs.com).
+{{site.data.alerts.end}}
+
+### Considerations
+
+- We recommend you add or remove nodes from a cluster when the cluster isn't experiencing heavy traffic.
```
chrisseto, line 41: Maybe a minor note as to why we recommend this? 
"Adding or removing nodes incurs a non-trivial amount of load on the cluster. Doing so while already experiencing load can result in degraded application performance or longer times for node modifications."

Amruta-Ranade: Done!

v20.1/cockroachcloud-cluster-management.md
Amruta-Ranade: Updated with the text we discussed on Slack

v20.1/cockroachcloud-cluster-management.md
```diff
@@ -32,11 +28,26 @@ For each node, the page displays the node's `Name`, nested under its region.
 
 From the **Cluster Overview** page, you can connect to your cluster. For more information, see [Connect to Your CockroachCloud Cluster](cockroachcloud-connect-to-your-cluster.html).
 
-## Manage cluster configuration
+## Add or remove nodes from a cluster
+
+You can add or remove nodes from your cluster through the Console.
+
+{{site.data.alerts.callout_info}}
+At this time, you cannot use the Console to scale up a single-node cluster or scale down to a single-node cluster. For single-node cluster changes, contact [Support](https://support.cockroachlabs.com).
```
rmloveland, line 36: I might change the second sentence to say "for these types of changes", since you've already described them in the first sentence.  Also, "single-node cluster changes" reads as ambiguous to me - it could refer to other changes to single-node clusters other than those described in the previous sentence.

v20.1/cockroachcloud-cluster-management.md
```diff
@@ -32,11 +28,26 @@ For each node, the page displays the node's `Name`, nested under its region.
...
+{{site.data.alerts.end}}
+
+### Considerations
+
+- We recommend you add or remove nodes from a cluster when the cluster isn't experiencing heavy traffic. Adding or removing nodes incurs a non-trivial amount of load on the cluster. Changing the cluster configuration during times of heavy traffic can result in degraded application performance or longer times for node modifications.
```
rmloveland, line 41: I might reverse the order of these 2 sentences so it's "problem -> solution"

v20.1/cockroachcloud-cluster-management.md
Amruta-Ranade: Yeah..I struggled with this phrasing. Thanks for the suggestion! Done!

v20.1/cockroachcloud-cluster-management.md
Amruta-Ranade: I went with the "state the action and then the reason" structure, but I don't mind the problem -> solution structure. Done!

PR #7540, opened 2020-06-22
Add docs for geospatial built-in functions

rmloveland:
... as well as whatever else was available on the main branch of
CockroachDB as of 6589106751d844bb6f01300c06d2872086de2632.


rmloveland:
@otan not much to review since you wrote most of these, more of an FYI

Issue #6951, opened 2020-03-23
Create how-to on shutting down a node


jseldess:
Jesse Seldess (jseldess) commented:

We are overloading the `cockroach quit` command to double as command reference and broader explanation of the process of shutting down a node. Instead, we should have a distinct page, `Stop a node`, that explains best practices and exactly what happens during the shut down process, and how to do it. 

Jira Issue: DOC-466

knz:
xref: https://github.com/cockroachdb/cockroach/pull/45149
xref: https://github.com/cockroachdb/cockroach/pull/46396

knz:
update on this: the new flag `--only-drain` to `cockroach quit` introduced in https://github.com/cockroachdb/cockroach/pull/45149 now reports a progress indicator, which can be used to determine when it's safe to stop the process.

taroface:
- Add links to this article from `cockroach-quit.md`, `planned-maintenance.md`.

taroface:
Also this thread: https://cockroachlabs.slack.com/archives/C01PHNMUFLN/p1635774345034000

PR #7542, opened 2020-06-22
update rolling upgrade docs with node drain, etc.

taroface:
Addresses: https://cockroachlabs.slack.com/archives/C65RYUY3Z/p1592851886246500

(Also reformats the `node-shutdown.md` include into HTML so that it works inside another list)


taroface:
> and only then send SIGTERM, wait 1 minute, then SIGKILL if necessary.

@knz Thanks for these clarifications. 

When is it necessary to send `SIGKILL`? Does this also mean that `SIGTERM` is "normally" the only required signal?

How are these signals handled by `systemctl stop <systemd config filename>` (this is the currently documented command on the upgrade doc)?

taroface:
@bdarnell Thank you for clarifying! Are the updated instructions accurate?

jseldess:
@bdarnell, ping.

taroface:
> > 1. Drain the node of SQL clients and [distributed SQL](architecture/sql-layer.html#distsql) queries:
> > ```
> 
> I think we want to mark this step as optional, if not remove it completely.

@bdarnell OK, noted! `cockroach node drain` is also listed as a required step for [node decommissioning](https://www.cockroachlabs.com/docs/v20.1/remove-nodes.html#step-5-stop-the-decommissioning-node). Do you mean that draining should always be optional before stopping a node, or only when stopping a node to upgrade?

bdarnell:
Doesn't decommissioning imply draining? I'm pretty sure it's completely redundant there. Anyway, yes, I'd remove `node drain` from the decommissioning process too. The only place I feel `node drain` is appropriate is for advanced users who are building it in to reusable infrastructure such as a k8s operator.

taroface:
> Doesn't decommissioning imply draining?

@bdarnell My understanding is that `node drain` is for draining SQL clients and distributed SQL queries, and that running `node decommission` mainly transfers ranges off the node, while continuing to accept SQL connections. Is it OK to decommission a node without draining SQL clients and DistSQL?

knz:
> My understanding is that node drain is for draining SQL clients and distributed SQL queries, and that running node decommission mainly transfers ranges off the node, while continuing to accept SQL connections.

Yes that understanding is correct. The two features are independent from each other.

taroface:
The latest updates to this PR:

- Remove `node drain` as a required step during decommissioning and upgrades
- Add a callout on the data unavailability / latency edge cases with SIGTERM + SIGKILL
- Replace instances of `pkill cockroach` with the node shutdown steps discussed here
  - (This command is still mentioned in `cockroach quit` docs, but since these are clearly being phased out in favor of #6951, I figure it's moot)

@bdarnell @knz Would you mind re-reviewing based on the latest conversation about our node shutdown guidance?

v20.1/remove-nodes.md
```diff
@@ -160,36 +160,14 @@ Even with zero replicas on a node, its [status](admin-ui-cluster-overview-page.h
...
-ok
-~~~
-
-Stop the node using one of the following methods:
+Drain and stop the node using one of the following methods:
 
 {% include {{ page.version.version }}/prod-deployment/node-shutdown.md %}
 
+{{site.data.alerts.callout_info}}
+In certain edge cases, stopping a node using signals can result in temporary data unavailability, latency spikes, uncertainty errors, ambiguous commit errors, or query timeouts. If you need maximum cluster availability during node decommissioning, you can run [`cockroach node drain`](cockroach-node.html) prior to node shutdown and actively monitor the draining process instead of automating it.
```
knz, line 168: ... stopping a node forcefully using SIGKILL or another signal than SIGTERM can result...

v20.1/remove-nodes.md
```diff
@@ -339,36 +317,14 @@ Even with zero replicas on a node, its [status](admin-ui-cluster-overview-page.h
...
-ok
-~~~
-
-Stop each node using one of the following methods:
+Drain and stop each node using one of the following methods:
 
 {% include {{ page.version.version }}/prod-deployment/node-shutdown.md %}
 
+{{site.data.alerts.callout_info}}
+In certain edge cases, stopping a node using signals can result in temporary data unavailability, latency spikes, uncertainty errors, ambiguous commit errors, or query timeouts. If you want to minimize these occurrences, you can run [`cockroach node drain`](cockroach-node.html) prior to node shutdown and monitor the draining process instead of automating it.
```
knz, line 325: ditto

v20.1/upgrade-cockroach-version.md
```diff
@@ -121,24 +121,14 @@ Note that this behavior is specific to upgrades from v19.2 to v20.1; it does not
...
-    ok
-    ~~~
-
-1. After the node is completely drained, stop the node:
+1. Drain and stop the node using one of the following methods:
     
     {% include {{ page.version.version }}/prod-deployment/node-shutdown.md %}
 
+    {{site.data.alerts.callout_info}}
+    In certain edge cases, stopping a node using signals can result in temporary data unavailability, latency spikes, uncertainty errors, ambiguous commit errors, or query timeouts. If you need maximum cluster availability during an upgrade, you can run [`cockroach node drain`](cockroach-node.html) prior to node shutdown and actively monitor the draining process instead of automating it.
```
knz, line 129: ditto

v20.1/remove-nodes.md
```diff
@@ -23,7 +23,7 @@ A node is considered to be decommissioned when it meets two criteria:
...
+After all range replicas have been transferred, the node can be drained of SQL clients, [distributed SQL](architecture/sql-layer.html#distsql) queries, and range leases, and then stopped. This can be done with a process manager or orchestration tool, or by sending `SIGTERM` manually. When stopped, the [`/health?ready=1` monitoring endpoint](monitoring-and-alerting.html#health-ready-1) starts returning a `503 Service Unavailable` status response code so that load balancers stop directing traffic to the node. At this point the node stops updating its liveness record, and after the duration configured via [`server.time_until_store_dead`](cluster-settings.html) is considered to be decommissioned.
```
knz, line 26: Regarding this bit: "When stopped, ..."

I think the paragraph should do a better job to distinguish the *initiation* of the graceful shutdown (a signal is sent), the *process* of shutting down/draining, and the *process termination*.

Then in that context, you can say "at some point during the process of shutting down, the monitoring endpoint starts reporting the node as non-ready, so that load balancers can redirect traffic".

The current text uses the verb "Stopped" which implicitly refers only to the end of the process. That's misleading.

v20.2/remove-nodes.md
```diff
@@ -23,7 +23,7 @@ A node is considered to be decommissioned when it meets two criteria:
...
+After all range replicas have been transferred, the node can be drained of SQL clients, [distributed SQL](architecture/sql-layer.html#distsql) queries, and range leases, and then stopped. This can be done with a process manager or orchestration tool, or by sending `SIGTERM` manually. When stopped, the [`/health?ready=1` monitoring endpoint](monitoring-and-alerting.html#health-ready-1) starts returning a `503 Service Unavailable` status response code so that load balancers stop directing traffic to the node. At this point the node stops updating its liveness record, and after the duration configured via [`server.time_until_store_dead`](cluster-settings.html) is considered to be decommissioned.
```
knz, line 26: see my note above

v20.2/remove-nodes.md
```diff
@@ -339,36 +317,14 @@ Even with zero replicas on a node, its [status](admin-ui-cluster-overview-page.h
...
-ok
-~~~
-
-Stop each node using one of the following methods:
+Drain and stop each node using one of the following methods:
 
 {% include {{ page.version.version }}/prod-deployment/node-shutdown.md %}
 
+{{site.data.alerts.callout_info}}
+In certain edge cases, stopping a node using signals can result in temporary data unavailability, latency spikes, uncertainty errors, ambiguous commit errors, or query timeouts. If you want to minimize these occurrences, you can run [`cockroach node drain`](cockroach-node.html) prior to node shutdown and monitor the draining process instead of automating it.
```
knz, line 325: see note above

v20.2/upgrade-cockroach-version.md
```diff
@@ -80,25 +80,15 @@ For each node in your cluster, complete the following steps. Be sure to upgrade
...
-    ~~~
-    
-1. After the node is drained, stop the node:
+  
+1. Drain and stop the node using one of the following methods:
     
     {% include {{ page.version.version }}/prod-deployment/node-shutdown.md %}
 
+    {{site.data.alerts.callout_info}}
+    In certain edge cases, stopping a node using signals can result in temporary data unavailability, latency spikes, uncertainty errors, ambiguous commit errors, or query timeouts. If you need maximum cluster availability during an upgrade, you can run [`cockroach node drain`](cockroach-node.html) prior to node shutdown and actively monitor the draining process instead of automating it.
```
knz, line 89: see note above

taroface:
@knz ping on review. thanks!

PR #7550, opened 2020-06-24
Update sidebar to better expose latest prod and testing releases

### LINKED ISSUE #7506: Reduce the number of releases shown on the /releases page
The number of releases shown on the /releases page is quite long, and I suspect that most users landing here are only interested in the latest version of the supported releases.

I have some suggestions to improve this:
* Move the very old releases to a separate /releases/all page
* For more recent releases, only include the latest patch version and only the test versions from the current release
* Tag old releases that are no longer supported as "EOL" or "unsupported"

### LINKED ISSUE #7547: testing releases are hard to find
Now that testing releases are no longer in the side nav they are quite hard to find:
image

To find it, one has to go from the left navigation releases->all releases to the right navigation testing releases to find it:
image

Switching navigation is unintuitive in this case and will make it hard for uses to find and try out our test releases.

I'd prefer adding testing releases back to the left navigation so that they are discoverable. 

I think this has contributed to reduced alpha usage:
image

https://cockroachlabs.looker.com/looks/263

jseldess:
- Create prod and testing sections.
- Expose latest release under each. Last 3 for prod. Only one
  for testing.
- Add intro to all prod patch release notes linking users
  to GA release notes for full feature summaries and upgrade
  instructions. With this, we can remove GA release notes
  from sidenav.

Fixes #7547.
Fixes #7506.

### DIFF
--- a/_includes/sidebar-releases.json
+++ b/_includes/sidebar-releases.json
@@ -3,45 +3,49 @@
   "is_top_level": true,
   "items": [
     {
-      "title": "v20.1.2",
-      "urls": [
-        "/releases/v20.1.2.html"
+      "title": "Production Releases",
+      "items": [
+        {
+          "title": "Latest v20.1",
+          "urls": [
+            "/releases/v20.1.2.html"
+          ]
+        },
+        {
+          "title": "Latest v19.2",
+          "urls": [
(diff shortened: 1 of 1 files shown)


awoods187:
LGTM!

PR #7551, opened 2020-06-24
Follow-up to https://github.com/cockroachdb/docs/pull/7550

jseldess:
Forgot to include this part in 7550:

Add intro to all prod patch release notes linking users
to GA release notes for full feature summaries and upgrade
instructions. With this, we can remove GA release notes
from sidenav.


awoods187:
LGTM

PR #7684, opened 2020-06-30
Add reference doc on transaction retry error codes

### LINKED ISSUE #4331: Add docs on contention errors (ReadWithinUncertaintyIntervalError). Why they exist and how to handle and reduce them

### LINKED ISSUE #5476: Transaction retry guidance and error reference
In https://github.com/cockroachdb/docs/issues/4331 we identified a need to add guidance around transaction retries. Unfortunately, this is insufficient. We need to be able to list all of the major transaction retry types as well as likely causes and most importantly, what to _do_ about these errors. 

Recently this came up for a customer in https://github.com/cockroachlabs/support/issues/334. We need a thorough audit of this information.

- [ ] ReadWithinUncertaintyIntervalError
- [ ]  Write Too Old
- [ ]  Forwarded timestamp (delete range)
- [ ]  Forwarded timestamp (iso=serializable)
- [ ]  Possible replay
- [ ]  Aynsc consensus failure
- [ ]  Transaction aborted error
- [ ]  Transaction push error (may not be possible)
- [ ]  TransactionRetryWithProtoRefreshError

And so on

rmloveland:
Fixes #4331.
Fixes #5476.

Summary of changes:

- Add a new 'Transaction retry error reference' page that lists as many
  of the retry error codes I could find.  For each error code, show:

  - What the error means

  - What to do about it

- Add links from other transaction- and troubleshooting-related pages in
  the docs to the new error reference page.


rmloveland:
Thanks for the reviews @andreimatei and @johnrk !

@andreimatei responding here since Reviewable's auth service is 404ing and otherwise erroring out for me (sorry for the clunkiness)

> You've said "accessing the same data" a couple of times so far. I think it'd be a good idea to suggest somehow that at least one of these accesses is a write; concurrent reads are fine.

Updated both of these mentions of "access the same data" to say "write to the same data".

> I think you want `RETRY_WRITE_TOO_OLD`?

I may have got this part wrong while trying to generate the error strings that get sent to the client.  I got the mapping from `RETRY_WRITE_TOO_OLD` -> `RestartsWriteTooOld` and the other mapping you were wondering about (`RETRY_SERIALIZABLE` -> `RestartsSerializable`) from https://github.com/cockroachdb/cockroach/blob/master/pkg/kv/kvclient/kvcoord/txn_coord_sender.go#L663

Are those the wrong mappings?  Or are you saying these are not the error message strings that are generated? 

I may have been looking in the wrong places for where these get put together in their final form before output. I thought these were built in https://github.com/cockroachdb/cockroach/blob/master/pkg/roachpb/errors.pb.go#L1147 ? Unfortunately the indirection from that function to a protobuf message is kinda hard for me to use to determine what the actual error messages to be printed will be.  If this is the wrong place to look, or if I'm reading stuff wrongly, please let me know.

> The linked page doesn't seem to explain anything about our auto-retry behavior. I think we should find a place to describe that, and explain that if you're able to batch your whole transaction and send it to crdb at once, we can generally swallow all these errors internally.

I have added a sentence after bullet 2 that says "In particular, if you are able to [send all of the statements in your transaction in a single batch](www.cockroachlabs.com/docs/v20.2/transactions.html#batched-statements), CockroachDB can usually automatically retry the entire transaction for you." (links to more info about the batching on the 'Transactions' page)

> I think you want `RETRY_SERIALIZABLE` ?

This is the mapping I found (discussed above with links to the code I was looking at), but I'm not sure if the right one.  Would appreciate any guidance on what these error message texts should be instead.

> It's not necessary for B to have committed; just for it to have already preformed the respective read.

Updated to say "Specifically, transaction _A_ tried to write a key that transaction _B_ had already read. _B_ was supposed to be serialized after _A_"

> I'd just leave "long-running". "Large" is ambiguous.

Removed this and all other uses of "large" in this doc.

> I'd note that here too (as well as case 2) ), the kicker from case 1) applies - there must have been another write to data that this transaction has already read and returned to the client. Otherwise the refresh would simply succeed. I'd also nod to this in the callout below. Long-running transaction are fine unless the respective data is also being written concurrently.

Thanks - I've sprinkled in the "another transaction did a write to data this one has read and returned" in cases 2 and 3.  Also in the callout, saying "if you absolutely must execute long-running transactions that execute concurrently with other transactions that are writing to the same data ..."

> I'd sprinkle the words "write skew" around somewhere for the anomaly that would be exhibited if these transaction were not restarted.

Added some words and link to our [write skew blog post](https://www.cockroachlabs.com/blog/what-write-skew-looks-like/)

> wrong name 

Similar to above, I found this mapping from `RETRY_ASYNC_WRITE_FAILURE` to ` RestartsAsyncWriteFailure` in the code files linked in the discussion of your earlier comment above.  If you can tell me the correct error message text that will appear to the user in this and those other cases, I would really appreciate it.  I don't seem to have dug it out correctly from the errors code

> I'd make this only slightly more specific that saying that we failed to replicate one of the txn's writes.

Updated to add "when some kind of problem with your cluster's operation occurs at the moment of a previous write in the transaction, causing us to fail to replicate one of the transaction's writes"

> ... and one of the operations is a write

Done.

> Is it? The first statement in the transaction is generally retried automatically, so, if anything, I think it's less likely than the others to return this (or, any) retriable error.

This info is taken from https://github.com/cockroachdb/cockroach/issues/34517#issuecomment-461077555  However, that was a while ago.  @bdarnell, has this changed since your comment in that discussion from Feb. 2019?

> I'm not sure what the "transaction queue" is :)

Me neither. :-)

Updated to say "timed out due to being pushed by other concurrent transactions"

> The "push" at play here is that same as the one in RETRY_SERIALIZABLE.

What should I update here in the doc (if anything)?  Are you saying the description of cases here should be the same as in the `RETRY_SERIALIZABLE` section?

> It's just running time that matters, I believe.

Updated to just say running time.

> But the thing is that it seems to me that, ever-since closed timestamps, we have a hard limit of ~5m on the duration of a transaction (because our implementation of this deadline is very dumb). In which case I think the docs should look differently. Let's see what comes out of cockroachdb/cockroach#36871 (comment).

Based on my understanding of what is written in that comment, we (currently, pending https://github.com/cockroachdb/cockroach/issues/51042) have a hard limit of ~5m.  Is that correct?  If so what do you think is the right thing to say to users here?

> s/tried to push/succeeding in pushing
> But actually pushing is not the right word here. We should say that B aborted A.

Ack - updated to say "B aborted A"

> If this error is caused by a deadlock, what the user should do is avoid producing deadlocks in their apps by making sure that the transactions involved acquire locks in the same order; deadlocks happen when different transactions try to acquire conflicting locks in reverse order from one another. We should give users more info on the deadlocks we detect to help them do that (you can even put a promise in these docs), but saying that deadlocks should be avoided and hinting at lock acquisition order I think is valuable regardless.

OK, updated to add the "avoid deadlocks" info as step 1.

rmloveland:
Thanks for the review Andrei!  Really appreciate it.

@bdarnell would you mind taking a pass over this before I send it through to docs team review?  I think the technical information is pretty much there AFAICT, but I would appreciate your feedback if you have opinions about how to frame it differently, or if you think something important is missing.

rmloveland:
Thanks Eric!  I am going to port these changes to the 20.1 docs as well.  Once the build is green with those changes, I'll merge.

Issue #6866, opened 2020-03-19
sql: undocumented third-party tooling [meta-issue]


awoods187:
This is a meta-issue for well-tested third-party tooling that is not currently documented or under-documented.

Drivers
- [x] [pgx](https://github.com/cockroachdb/docs/issues/7185) 

ORMs
- [ ] **[GORM](https://github.com/cockroachdb/docs/issues/6038)**
- [x] **[Spring Tutorial](https://github.com/cockroachdb/docs/issues/6046) (Java), [Retry Examples](https://github.com/cockroachdb/docs/issues/3238), and [application](https://github.com/cockroachdb/docs/issues/4904)**
- [ ] [Hibernate Tutorial](https://github.com/cockroachdb/docs/issues/5992)
- [ ] [Go-lang ORM](https://github.com/cockroachdb/docs/issues/6037) (Go)
- [ ] Buffalo ORM (Go)
- [x] [MyBatis](https://github.com/cockroachdb/docs/issues/6040) (Java)
- [ ] [Peewee](https://github.com/cockroachdb/docs/issues/6850) (Python) (This is currently linking to third-party tooling documentation without a tutorial)
- [ ] Django Improvements including [best practices](https://github.com/cockroachdb/docs/issues/6368) and [limitations](https://github.com/cockroachdb/docs/issues/6367) (Python)
- [x] [SQL Alchemy](https://github.com/cockroachdb/docs/issues/6044)
- [ ] [Sequelize](https://github.com/cockroachdb/docs/issues/4444) 

GUIs
- [ ] **Postico**
- [ ] **[Dbeaver updates](https://github.com/cockroachdb/docs/issues/6816)**
- [ ] [Meta-list from a previous blitz](https://github.com/cockroachdb/docs/issues/5260)
- [ ] dbglass
- [ ] PGWeb
- [ ] PSequel
- [ ] TablePlus
- [ ] Valentina studio

[Schema Migration](https://github.com/cockroachdb/docs/issues/6679)
- [x] **[Flyway](https://github.com/cockroachdb/docs/issues/5356)**
- [ ] **[Flowable](https://github.com/cockroachdb/docs/issues/6865)**
- [ ] [Liquibase](https://github.com/cockroachdb/docs/issues/5989)

Observability
- [ ] Add [Datadog](https://www.datadoghq.com/blog/monitor-cockroachdb-performance-metrics-with-datadog/)

Infrastructure/Other
[...]

PR #7698, opened 2020-07-02
Hello World tutorial for Go pgx

### LINKED ISSUE #7011: Install a Postgres Client Doc - let's recommend pgx instead of lib/pq
Re: [Install a Postgres Client](https://cockroachlabs.com/docs//v19.2/install-client-drivers.html)

Instead of recommending lib/pq as a Go driver, we should be recommending [pgx](https://github.com/jackc/pgx). I think everybody agrees it's much higher quality.

cc @rafiss

### LINKED ISSUE #7185: Update Go tutorial to use pgx driver
Re: [Build a Go App with CockroachDB the Go pg Driver](https://cockroachlabs.com/docs//v19.2/build-a-go-app-with-cockroachdb.html)

We should be recommending the `pgx` driver instead of `lib/pq`. It's better.

Separately, `pgx` can also be used directly as a library, not only as a driver for the Go `database/sql` package. We should also encourage that. There's native support for using `pgx` as a library in the `cockroach-go/crdb/crdbpgx` package.

ericharmeling:
Fixes #7185. 
Fixes #7011. 
Related to #6866.

This PR includes the following changes:
- Renamed lib/pq Hello World tutorial file from `build-a-go-app-with-cockroachdb.html` to `build-a-go-app-with-cockroachdb-pq.html`.
- Added new Hello World tutorial content to file named `build-a-go-app-with-cockroachdb.html`. This tutorial instructs users to set up and run a simple app that uses pgx to connect to CRDB. It is now the default Go tutorial.
- Added new code samples that use pgx to connect to CRDB, and crdbpgx to execute transactions.
- Updated `drivers.md` and `tooling.md` include files to list pgx at full support, and link to new tutorial.

These changes have only been made to the v20.1 docs. After this PR has been reviewed and the changes finalized, I'll make copies for v19.2 and v20.2.

Separately, I created the [hello-world-go-pgx](https://github.com/cockroachlabs/hello-world-go-pgx) repository for a single-file version of the code samples. Before merging this PR, I'll make that repo public.

@andreimatei 
Adding you for technical review, Andrei. As you can see, I adapted the lib/pq code samples (https://github.com/cockroachdb/docs/blob/master/_includes/v20.1/app/basic-sample.go and https://github.com/cockroachdb/docs/blob/master/_includes/v20.1/app/txn-sample.go) to use pgx. 

Note that I had to explicitly define `tls.Config.ServerName` in the connection configuration. Without this defined explicitly, the following error was reported for the secure sample:
```
error connecting to the database: failed to connect to `host=localhost user=maxroach database=bank`: failed to write startup message (tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config)
```
This step wasn't documented at https://pkg.go.dev/github.com/jackc/pgx?tab=doc or https://pkg.go.dev/github.com/cockroachdb/cockroach-go@v2.0.1+incompatible/crdb/crdbpgx?tab=doc, so it could be some kind of bug? Or I could have missed something?


ericharmeling:
@vy-ton FYI

rafiss:
The `ServerName` requirement is coming from the Go `tls` package: https://golang.org/pkg/crypto/tls/#Client

It looks like `pgx` allows full configuration of the TLS config, so you are right, it's required here. https://godoc.org/gopkg.in/jackc/pgx.v3#hdr-TLS

_includes/v20.1/app/insecure/basic-sample-pgx.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/jackc/pgx/v4"
+)
+
+func main() {
+	config, err := pgx.ParseConfig("postgresql://maxroach@localhost:26257/bank?sslmode=disable")
+	if err != nil {
+		log.Fatal("error configuring the database: ", err)
+	}
+
+	config.TLSConfig.ServerName = "localhost"
```
rafiss, line 17: well actually, I wouldn't expect ServerName to be required here. does this fail without specifying `ServerName`?

ericharmeling: You're right. In the insecure example, it doesn't fail. I'll remove that line from the insecure example.

ericharmeling:
@lnhsingh 
FYI
I'm going to copy these files over to the 19.2 and 20.2 directories after you've signed off on what's in 20.1.

PR #7736, opened 2020-07-15
Document how to get CRDB-spatial up and running

### LINKED ISSUE #7723: Document how to get 20.2 alpha version of CRDB with geospatial support up and running
We need installation docs so users can try out early versions of the geospatial capabilities we are planning for 20.2.

Right now (July 2020), the way to do this is to run a special build from within a Docker container as follows:

1. Start the Server

```
docker run -d --name=geo --hostname=geo-host -p 26257:26257 -p 8080:8080  -v "${PWD}/cockroach-data/roach1:/cockroach/cockroach-data"  otanatcockroach/cockroachdb-geospatial start-single-node --insecure
```

2. Connect from a SQL client

3. ???

4. Profit!

Estimated scope of work:

- Add a new section to the 20.2 docs called "Geospatial something something" where this can live
- Write a bit of overview / intro text
- Write the instructions

rmloveland:
Fixes #7723.

Summary of changes:

- Add a new tutorial page, 'Working with Spatial Data', with information
  about how to install it, import data, run some simple spatial queries.
  List some basic performance and troubleshooting info.  Make it clear
  that this page refers to an early alpha release of 20.2, and things
  will change.


rmloveland:
@awoods187 for general review

@otan and @rytaft - would be good to get your feedback and clarifications on the stuff in the 'Performance' section which is SUPER high-level / hand-wavey re: "our indexing is divide-the-space" and "perf is usually no worse than 2x PGIS".  Also, your feedback on anything else in this doc is appreciated as well, of course!

v20.2/spatial-data.md
```diff
@@ -0,0 +1,305 @@
...
+title: Working with Spatial Data
+summary: CockroachDB has special support for efficiently storing and querying spatial data.
+toc: true
+---
+
+<span class="version-tag">New in v20.2</span>: CockroachDB has special support for efficiently storing and querying spatial data. Supported spatial data types include:
+
+- Geometric objects such as points, lines, paths, and polygons in 2-dimensional space. These are projected onto the flat surface of a plane and are represented in SQL by the `GEOMETRY` data type.
+
+- Geographic objects, which are also made up of points, lines, polygons, etc., in 2-dimensional space. They are projected onto the surface of a sphere and are represented in SQL by the `GEOGRAPHY` data type. (Technically, they are projected onto a spheroid: "a sphere with a bulge"). This projection means that paths between geographic objects are not straight lines; they are curves, and so the distances between objects are calculated using [great circle math](https://en.wikipedia.org/wiki/Great-circle_distance).
```
otan, line 11: I would try and somehow word "Long/Lat" as the "x,y" coordinate representations here.

v20.2/spatial-data.md
```diff
@@ -0,0 +1,305 @@
...
+
+We recommend using `GEOMETRY` as much as possible for best performance, and converting to the `GEOGRAPHY` type only where needed by your application.
+
+In order to avoid full table scans, make sure to add indexes to any columns that are frequently accessed. You can check which queries are using which indexes using the [`EXPLAIN`](explain.html) statement. For more information about general query tuning (including index usage), see [Make queries fast](make-queries-fast.html).
+
+The syntax for adding an (inverted) index to a geometry column is:
+
+{% include copy-clipboard.html %}
+~~~ sql
+CREATE INDEX tornado_geom_idx ON tornadoes USING GIN (geom);
```
otan, line 41: thought it was using GiST (https://postgis.net/workshops/postgis-intro/indexing.html), maybe GIN works too.

v20.2/spatial-data.md
```diff
@@ -0,0 +1,305 @@
...
+~~~
+
+For more information about how inverted indexes work, see [Inverted indexes](inverted-indexes.html).
+
+Because CockroachDB is a scale-out, multi-node database, our spatial indexing strategy is based on a [space-filling curve](https://en.wikipedia.org/wiki/Space-filling_curve) design (also known as "divide the space"), rather than the [R-Tree](https://en.wikipedia.org/wiki/R-tree) data structure used by some other spatial databases (also known as "divide the objects"). Other databases that use a "divide the space" strategy include Microsoft SQL Server and MongoDB.
+
+We are still working on optimizing our performance for some common use cases. In most cases we expect to be within 2x of the performance of PostGIS for supported queries.
```
otan, line 48: I would probably support omitting this line at this point in time; we just have nowhere near enough benchmarks.

rmloveland:
Oliver Tan <notifications@github.com> writes:

> @otan commented on this pull request.
>
> Looking good as a guide for our first documentation! Comments below.

Thanks Oliver!

> I haven't actually tried the GUI tool they recommended for importing
> data, because I don't have windows. You may need to enumerate the
> shp2pgsql steps again here, and start from after the point where the
> data is loaded.

OK.  Updated to say "avoid the gui/shapefile stuff, just load this sql
we made for you".

> Also would mention we do not support SVG.

Mentioned!

> i wonder if the copyright google bit is required when using google
> images.

Added "(Map data &copy; 2020 Google)".

> I'm a little hesitant to this advice - it really depends on what
> you're going for.
>
> Instead, something like
> https://postgis.net/workshops/postgis-intro/geography.html#why-not-use-geography
> would make more sense.

Agreed!  But I think I'm gonna punt on this for now and just delete that
line.  The goal for this PR is "minimum get-up-and-running words".  I
probably messed up by even mentioning performance here (yet).  But I
suppose the indexing stuff is pretty general-purpose, harmless advice.

> I would try and somehow word "Long/Lat" as the "x,y" coordinate
> representations here.

Updated to add that "the X and Y coordinates of 2-dimensional points are
actually Longitude and Latitude values."

> thought it was using GiST
> (https://postgis.net/workshops/postgis-intro/indexing.html), maybe GIN
> works too.

At least on the Docker image I have, only `GIN` works; using `GiST`
blows up (output below).  Maybe when 20.2.0-alpha.2 is out?

```
> CREATE INDEX tornado_geom_idx ON tornadoes USING GIST (geom);
invalid syntax: statement ignored: at or near "gist": syntax error: unimplemented: this syntax
SQLSTATE: 0A000
DETAIL: source SQL:
CREATE INDEX tornado_geom_idx ON tornadoes USING GIST (geom)
                                                 ^
HINT: You have attempted to use a feature that is not yet implemented.

Please check the public issue tracker to check whether this problem is
already tracked. If you cannot find it there, please report the error
with details by creating a new issue.

If you would rather not post publicly, please contact us directly
using the support form.

We appreciate your feedback.
root@localhost:26257/tornadoes> select version();
select version();
                                                      version
-------------------------------------------------------------------------------------------------------------------
  CockroachDB CCL v20.2.0-alpha.1-931-gc433cb6fc3 (x86_64-unknown-linux-gnu, built 2020/07/06 18:09:04, go1.14.4)
(1 row)
```

> I would probably support omitting this line at this point in time; we
> just have nowhere near enough benchmarks.

Deleted!

rmloveland:
OMG the GitHub handling of email replies is criminal malpractice.  Sorry about that, Oliver!  Will reply via GUI from now on.

v20.2/spatial-data.md
```diff
@@ -0,0 +1,309 @@
...
+{% include copy-clipboard.html %}
+~~~ shell
+cd otan-scripts/geospatial/geospatial_sql
+~~~
+
+Load the SQL files into your CockroachDB cluster:
+
+{% include copy-clipboard.html %}
+~~~ shell
+cat *sql |cockroach sql --insecure --host=localhost --port=26257
```
otan, line 293: `cat otan-scripts/geospatial/geospatial_sql/*.sql | cockroach sql --insecure --host=localhost --port=26257`

should do it.

v20.2/spatial-data.md
awoods187: Looks like both will work but for now we should say GIN since GIST doesn't work

v20.2/spatial-data.md
```diff
@@ -0,0 +1,309 @@
...
+(Map data &copy; 2020 Google)
+
+#### Example 2. NYC data from the PostGIS tutorial
+
+Follow the steps below to load the SQL for the NYC data used in the [Introduction to PostGIS](https://postgis.net/workshops/postgis-intro/) tutorial.  CockroachDB can work with the tutorial up to Chapter 20, with the following exceptions:
+
+- Do not try to load Shapefiles via the GUI as shown in the tutorial.  Instead, follow the steps below to load the SQL data directly into CockroachDB.  (We have already converted the tutorial Shapefiles to SQL for you.)
+- We do not support GML or KML data.
+- We do not support SVG.
+
```
awoods187, line 274: may be worth stating we only support up to chapter 20

otan:
___
*[v20.2/spatial-data.md, line 36 at r2](https://reviewable.io/reviews/cockroachdb/docs/7736#-MCSTRSYA9FjAKjqpW-b:-MCluLFD0ahCEjVHv-ia:b-mockeb) ([raw file](https://github.com/cockroachdb/docs/blob/146878c8e7aba90a5647d18e5848cf72a1ef3895/v20.2/spatial-data.md#L36)):*

Yep! These are the only functions for which we currently have indexing support:
```
	st_covers
	st_coveredby
	st_contains
	st_containsproperly
	st_crosses
	st_dwithin
	st_dfullywithin
	st_equals
	st_intersects
	st_overlaps
	st_touches
	st_within
```
(you can also find this list here: https://github.com/cockroachdb/cockroach/blob/c78c517ad747b659bfee95de001f8a9f5b8da872/pkg/geo/geoindex/geoindex.go#L33).

I don't think we are planning to add others, so this is probably the final list. But @otan should correct me if I'm wrong.

We also have equivalent versions of all of these functions that will explicitly *not* use the index:
```
	_st_covers
	_st_coveredby
	_st_contains
	_st_containsproperly
	_st_crosses
	_st_dwithin
	_st_dfullywithin
	_st_equals
	_st_intersects
	_st_overlaps
	_st_touches
	_st_within
```
These underscore versions return the same result as the regular functions, but will not use a geospatial index.
</blockquote></details>

correct. in addition, geography only has these:
        _st_covers
	_st_coveredby	
	_st_dwithin
	_st_intersects

PR #7739, opened 2020-07-15
upgrade Production Checklist with "rules of thumb"

### LINKED ISSUE #6702: Recommended Linux distros and kernel versions
We need to provide a recommendation in our docs for which Linux distributions and kernel versions work well and are tested with CockroachDB.

This is necessary in case a given Linux kernel version has a bug in a system call we depend on (e.g., `fsync`).

This will make life easier for users (since they will have a vetted kernel version) and us (since we will be able to point to that recommendation if users are running on a kernel with a bug).

Regarding the current state: Just did a bunch of grepping in the docs repo, a web search, and read our [Recommended Production Settings][prod], [Operational FAQ][ops], and [Install CRDB on Linux][linux] pages.  Did not find anything about kernel versions.

The most specific we seem to get is mentioning specific C libraries etc. in the Ops FAQ here:

https://www.cockroachlabs.com/docs/v19.2/recommended-production-settings.html#dependencies

Estimated scope of work:

- Update [Recommended Production Settings][prod], [Operational FAQ][ops], and [Install CRDB on Linux][linux] with a recommended Linux kernel version
- (This can probably be an include for reusability)
- Investigate where else this information needs to be shared (Kubernetes pages?), and update as needed

See also:

- #4209
 - #2268

[prod]: https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html
[ops]: https://www.cockroachlabs.com/docs/stable/operational-faqs.html
[linux]: https://www.cockroachlabs.com/docs/stable/install-cockroachdb-linux.html

### LINKED ISSUE #7096: Update platform provisioning recommendations
@a-entin has raised concern that, in our [Production Checklist](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html), our minimum RAM-to-vcpu ratio is not suitable for even basic performance testing and that our recs across clouds are not consistent and likely not based on production-like experiences. 

We need to take another look at these docs, with eng and CS.

taroface:
Fixes #7096.
Fixes #6702.

- As Raphael noted, selecting the number of nodes for fault tolerance comes before selecting the number of CPUs per node. We don't currently have a docs reference specifically about determining node count for a production deployment, but this is in progress. I'm discussing with @jseldess how best to add this to our deployment guidance.
- On the storage capacity issue (60 GB -> 250 GB), I was unclear on whether this capacity equates to the max `store` size and whether that in turn equates to the total disk capacity. Since we advise operators to always leave a certain amount (listed in this update as 40%) free, should they account for this 40% when provisioning volumes so that they can benefit from the full supported 60/250 GB?
  - For now I've conflated the recommended max storage with the disk capacity, and left the 40% free capacity as a separate consideration.
  - 250 GB is a value I made up that gives 4 TB / 16 vCPUs, based on the node density test. Is this too ambitious?
- There is discrepancy throughout between MB/GB/TB (used in [Alex's doc](https://docs.google.com/document/d/1eYbjhoyY4HgpmPUilYBDhRCn1vOtXiV2RlbSog7ZXhU/edit#)) and MiB/GiB/TiB (used elsewhere in the published doc). Should the units all match?
- Also, I am unclear whether the I/O unit is MB/s or Mbps. 
- There is currently no information about the specific effects of under-provisioning CPU, RAM, or disk, as requested by @petermattis. @bdarnell any insight on this would be helpful!

### DIFF
--- a/v19.2/recommended-production-settings.md
+++ b/v19.2/recommended-production-settings.md
@@ -36,55 +36,53 @@ This hardware guidance is meant to be platform agnostic and can apply to bare-me
 | IOPS per vCPU | 500 | [Disk I/O](#disk-i-o)
 | MB/s per vCPU | 30 | [Disk I/O](#disk-i-o)
 
-{{site.data.alerts.callout_danger}}
 Before deploying to production, test and tune your hardware setup for your application workload. For example, read-heavy and write-heavy workloads will place different emphases on [CPU](#cpu-and-memory), [RAM](#cpu-and-memory), [storage](#storage), [I/O](#disk-i-o), and [network](#networking) capacity.
-{{site.data.alerts.end}}
 
 #### CPU and memory
 
-Each node should have at least **2 vCPUs** and at least **4 GB of RAM per vCPU**. More data, complex workloads, higher concurrency, and faster performance require additional resources. 
+Each node should have **at least 2 vCPUs**, and for best performance we recommend between 4 and 32 vCPUs per node. Provision **4 GiB of RAM per vCPU**.
 
-{{site.data.alerts.callout_danger}}
-Avoid "burstable" or "shared-core" virtual machines that limit the load on CPU resources.
-{{site.data.alerts.end}}
+- To optimize for throughput, use larger nodes with up to 32 vCPUs. To further increase throughput, add more nodes to the cluster instead of increasing node size. 
 
-- Select the fastest available processors. 
+      {{site.data.alerts.callout_info}}
+      Note that the benefits to having more RAM decrease as the number of vCPUs increases.
+      {{site.data.alerts.end}}
 
-- To optimize for throughput, use larger nodes with up to 32 vCPUs (the sweet spot for OLTP workloads, based on internal testing results). See [Disk I/O](#disk-i-o) for additional throughput guidelines.
+- To optimize for resiliency, use many smaller nodes instead of fewer larger nodes. Recovery from a failed node is faster when data is spread across more nodes.
 
-    To further increase throughput, add more nodes to the cluster instead of increasing node size. Nodes with more than 32 vCPUs will have [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access) (non-uniform memory access) implications leading to performance downgrades.
-
-    {{site.data.alerts.callout_info}}
-    Note that the benefits to having more RAM decrease as the number of vCPUs increases.
-    {{site.data.alerts.end}}
-
-- To optimize for resiliency, use many smaller nodes (e.g., 4 vCPUs per node) instead of fewer larger nodes. Recovery from a failed node is faster when data is spread across more nodes.
+- Avoid "burstable" or "shared-core" virtual machines that limit the load on CPU resources.
 
 - To ensure consistent SQL performance, make sure all nodes have a uniform configuration.
 
+{{site.data.alerts.callout_info}}
+Underprovisioning RAM results in reduced performance (due to reduced caching and increased spilling to disk), and in some cases can cause OOM crashes. Underprovisioning CPU generally results in poor performance, and in extreme cases can lead to cluster unavailability. For more information, see [capacity planning issues](cluster-setup-troubleshooting.html#capacity-planning-issues) and [memory issues](cluster-setup-troubleshooting.html#memory-issues).
+{{site.data.alerts.end}}
+
 #### Storage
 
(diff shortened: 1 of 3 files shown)


taroface:
> Storage generally uses MB/s, while networking generally uses Mbps.

@rmloveland Will it be OK if I change the "data" units under "Node desnity testing configuration" to GB/TB rather than GiB/TiB for consistency? Or would that make the test result incorrect?

rmloveland:
> @rmloveland Will it be OK if I change the "data" units under "Node desnity testing configuration" to GB/TB rather than GiB/TiB for consistency? Or would that make the test result incorrect?

I think they're close enough it doesn't matter very much.

Possibly relevant aside that does not require any action from you in this issue (just thinking out loud): I think our use of the `*iB` measurements is probably lost on most users anyway, and if we are now going to mix `TiB` and `TB` I think the mixing is strictly much worse than just choosing to use one or the other all the time.  My vote would be for the less-pedantic but more-common `TB` variant.  I will put something on our team meeting agenda to discuss.

v19.2/recommended-production-settings.md
```diff
@@ -23,42 +27,75 @@ Mentions of "CPU resources" refer to vCPUs, which are also known as hyperthreads
...
+| Capacity per vCPU | 60 GB | [Storage](#storage)
+| IOPS per vCPU | 500 | [Disk I/O](#disk-i-o)
+| MB/s per vCPU | 30 | [Disk I/O](#disk-i-o)
+
+Before deploying to production, test and tune your hardware setup for your application workload. For example, read-heavy and write-heavy workloads will place different emphases on [CPU](#cpu-and-memory), [RAM](#cpu-and-memory), [storage](#storage), [I/O](#disk-i-o), and [network](#networking) capacity.
 
 #### CPU and memory
 
-- At a bare minimum, each node should have **2 vCPUs and 2 GB of RAM**. More data, complex workloads, higher concurrency, and faster performance require additional resources.
+Each node should have **at least 2 vCPUs**, and for best performance we recommend between 4 and 32 vCPUs per node. Provision **4 GiB of RAM per vCPU**.
```
Amruta-Ranade, line 43: nit: missing comma after "for best performance"

Amruta-Ranade, line 43: nit/opinion: We use `GiB` here and `MB`, `GB` elsewhere. Is this intentional?

v19.2/recommended-production-settings.md
```diff
@@ -23,42 +27,75 @@ Mentions of "CPU resources" refer to vCPUs, which are also known as hyperthreads
...
+We recommend provisioning volumes with **60 GB per vCPU**. It's fine to have less storage per vCPU if your workload does not have significant capacity needs.
 
-- Avoid using shared storage such as NFS, CIFS, and CEPH storage.
+- The maximum recommended storage capacity per node is 2 TB, regardless of the number of vCPUs.
 
-- For the best performance results, use SSD or NVMe devices. The recommended volume size is 300-500 GB.
+- Use dedicated volumes for the CockroachDB [store](architecture/storage-layer.html). Do not share the store volume with any other I/O activity.
 
-    Monitor IOPS for higher service times. If they exceed 1-5 ms, you will need to add more devices or expand the cluster to reduce the disk latency. To monitor IOPS, use tools such as `iostat` (part of `sysstat`).
+    We suggest storing CockroachDB [log files](debug-and-error-logs.html) in a separate volume from CockroachDB data so that logging is not impacted by I/O throttling. This can cause unwanted behavior on the cluster.
```
Amruta-Ranade, line 69: nit/opinion: Should we elaborate on what we mean by "unwanted behavior on the cluster."

taroface: Since this was downgraded from a 'critical' to a 'suggested' point in review, perhaps it's best not to get into it. I removed the line instead.

taroface:
Per [this discussion](https://cockroachlabs.slack.com/archives/C65RYUY3Z/p1595951811034400) I'm going to use `iB` throughout this document for this update, to preserve internal consistency.

- The Cloud Console uses `iB`, and so will the Admin UI once embedded into the Console
- From what I understand of this discussion, CRDB generally (?) uses 1024^i

v19.2/recommended-production-settings.md
taroface: Answered in below comment!

jseldess:
Excited to see this over the finish line. Nice work, @taroface!

PR #7742, opened 2020-07-16
open email popup on advisories page

anagio:
Added a condition around the js to check if on the advisories page to always show the email popup 

prettier formatted the page


anagio:
Check the latest preview i adjusted the condition

anagio:
I made some small edits adding a new form on the advisories page can you take one more look, please. Thank you

PR #7776, opened 2020-07-22
Remove gitter-related artifacts

jseldess:
Since our Gitter account is no more, also removing all Gitter-related artifacts from the docs repo.


anagio:
lgtm

Issue #5922, opened 2019-11-21
add guidance for CPU requirements on single-cluster k8s deployment


taroface:
Our single-cluster Kubernetes deployment tutorials have guidance on defining memory requests and limits. We need to provide similar guidance on defining CPU resources. The following points (paraphrased from @ajwerner) need to be conveyed and summarized with production recommendations:

- ~The CPU resources specified for a pod may not necessarily match the number of physical cores on the Kubernetes VMs being used.~ 
- ~When fewer cores are requested for CRDB than the total available, you must also specify a GOMAXPROCS value other than the total. Otherwise the go runtime will still use the total number of cores.~
- ~The requested CPU resources and GOMAXPROCS value may not necessarily match. It's possible that different values are required for optimal performance.~
- ~It may be preferable not to specify CPU requests:~ https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit
- It is possible (per above link) to set default limits for the cluster

Questions:

- Should users specify a CPU request/limit for Kubernetes pods?
  - If yes, how should they determine the CPU request ~and GOMAXPROCS values? Also, how do they set GOMAXPROCS?~
  - ~If no, should they set a default CPU limit for the cluster?~

Relates to #5931

EDIT: Removed incorrect line that stated a certain GOMAXPROCS value causes issues.

DuskEagle:
> Should users specify a CPU request/limit for Kubernetes pods? 

That ultimately depends on the environment. It's simpler if they don't, but that's not going to be an option for everyone.

> If yes, how should they determine the CPU request and GOMAXPROCS values?

I don't know the answer to this, other than setting CPU requests as high as possible. For GOMAXPROCS, as I understand it that should be set to the same amount as the number of CPU units you're requesting for the pod. I don't know what it should be if you request a non-integer amount of CPU units.

>Also, how do they set GOMAXPROCS?

Add this stanza to the [CockroachDB container spec's env](https://github.com/cockroachdb/cockroach/blob/fdffc5fee97476b2ff4bc2efba159f6f2ea235f7/cloud/kubernetes/cockroachdb-statefulset.yaml#L138):
```
- name: GOMAXPROCS
   value: 4
```

>If no, should they set a default CPU limit for the cluster?

That sounds like something only a Kubernetes cluster administrator would be able to do. If a user is going to set a resource limit, setting it on the statefulset spec is the place to put it.

[...]

PR #7782, opened 2020-07-23
K8s multiregion eks

### LINKED ISSUE #4314: Write multi-region EKS (Amazon) Kubernetes tutorial
Our current multi-k8s deployment guidance focused on GKE and requires a stable public load balancing IP to hook together DNS servers from different clusters. As seen on our forum, this approach  might not be possible on AWS, and as @bdarnell has mentioned, there are probably other, better options for achieving what we want anyway. 

- We should identify the right way to load balancer across multiple k8s clusters, likely via a load balancer hostname that is resolvable and routable from all the clusters. 
- We should then create a separate tutorial for AWS, and consider updating the GKE tutorial and configs.

taroface:
Fixes #4314.
Relates to https://github.com/cockroachdb/cockroach/pull/51775, which adds `cockroachdb-statefulset-secure-eks.yaml`, `dnb-lb-eks.yaml`, and `configmap.yaml` to the `multiregion` repo.
Also addresses https://github.com/cockroachdb/cockroach/pull/51471.

- Draft EKS multi-region deployment (manual steps; no script)
- Under **Create inbound rules**, I'm not sure if the "Inter-node and load balancer-node communication" still fits for this rule on multi-region. It seems the load balancer points at CoreDNS on port 53, and CoreDNS provides the load balancer-node communication. Is this right?
- Since #5922 appears to still be unresolved, there is no guidance here on setting a CPU request and/or limit.
- Bob's [PR](https://github.com/cockroachdb/cockroach/pull/51471) implied (to me) that `--cache` and `--max-sql-memory` are now dynamically set in the `cockroach start` command, so I removed this from the multi- and single-region guides.
  - Also, I am not sure if the line "To avoid running out of memory when CockroachDB is not the only pod on a Kubernetes node, you must set memory limits explicitly. This is because CockroachDB does not detect the amount of memory allocated to its pod when run in Kubernetes." is still accurate after that PR. I have an instruction here to add a memory request & limit, but am not clear on whether there is still a risk of OOM, after Bob's changes, if you don't set limits.

### DIFF
--- a/_includes/v20.1/prod-deployment/insecure-test-load-balancing.md
+++ b/_includes/v20.1/prod-deployment/insecure-test-load-balancing.md
@@ -1,8 +1,12 @@
 CockroachDB offers a pre-built `workload` binary for Linux that includes several load generators for simulating client traffic against your cluster. This step features CockroachDB's version of the [TPC-C](http://www.tpc.org/tpcc/) workload.
 
+{{site.data.alerts.callout_info}}
+Be sure that you have [set up an inbound rule](deploy-cockroachdb-on-aws-insecure.html#step-2-configure-your-network) that allows traffic from the application to the load balancer. In this case, you will run the sample workload on one of your instances. The traffic source for your inbound rule should therefore be the **internal (private)** IP address of that instance. To find this, open the Instances section of the Amazon EC2 console and click on the instance.
+{{site.data.alerts.end}}
+
 {{site.data.alerts.callout_success}}For comprehensive guidance on benchmarking CockroachDB with TPC-C, see our <a href="https://www.cockroachlabs.com/guides/cockroachdb-performance/">Performance Benchmarking white paper</a>.{{site.data.alerts.end}}
 
-1. SSH to the machine where you want the run the sample TPC-C workload.
+1. SSH to the machine where you want to run the sample TPC-C workload.
 
     This should be a machine that is not running a CockroachDB node.
 
--- a/_includes/v20.1/orchestration/start-cockroachdb-insecure.md
+++ b/_includes/v20.1/orchestration/start-cockroachdb-insecure.md
@@ -8,15 +8,21 @@
     ~~~
 
     {{site.data.alerts.callout_danger}}
-    To avoid running out of memory when CockroachDB is not the only pod on a Kubernetes node, you must set memory limits explicitly. This is because CockroachDB does not detect the amount of memory allocated to its pod when run in Kubernetes. Specify this amount by adjusting `resources.requests.memory` and `resources.limits.memory` in `cockroachdb-statefulset.yaml`. Their values should be identical.
+    To avoid running out of memory when CockroachDB is not the only pod on a Kubernetes instance, you *must* set `resources.requests.memory` and `resources.limits.memory` to explicit values in the CockroachDB `containers` spec. This is because CockroachDB does not detect the amount of memory allocated to its pod when run in Kubernetes. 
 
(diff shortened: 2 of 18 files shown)


taroface:
@DuskEagle friendly ping on review!

taroface:
TFTR @DuskEagle !

_includes/v20.1/prod-deployment/insecure-test-load-balancing.md
```diff
@@ -1,8 +1,12 @@
 CockroachDB offers a pre-built `workload` binary for Linux that includes several load generators for simulating client traffic against your cluster. This step features CockroachDB's version of the [TPC-C](http://www.tpc.org/tpcc/) workload.
 
+{{site.data.alerts.callout_info}}
+Be sure that you have [set up an inbound rule](deploy-cockroachdb-on-aws-insecure.html#step-2-configure-your-network) that allows traffic from the application to the load balancer. In this case, you will run the sample workload on one of your instances. The traffic source for your inbound rule should therefore be the **internal (private)** IP address of that instance. To find this, open the Instances section of the Amazon EC2 console and click on the instance.
```
Amruta-Ranade, line 4: nit: Should "Instances" be "*Instances*" since we are referring to a UI element? If yes, need to update other places where it appears.

jseldess:
___
*[v20.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md, line 261 at r3](https://reviewable.io/reviews/cockroachdb/docs/7782#-MFkVDrNE2snz9qRYYHj:-MFkVDrO599nsyHJnOTq:b-t4oa4) ([raw file](https://github.com/cockroachdb/docs/blob/b053df54444baed416cbf03d8c6442bb2f79e9e5/v20.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md#L261)):*
> ```Markdown
>     ~~~ shell
>     kubectl create namespace <cluster-namespace> --context=<cluster-context>
> ```

We use `=` for this flag but not for `namespace`. This is a nit, but I'd prefer we be consistent about this throughout.

jseldess:
___
*[v20.1/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md, line 296 at r3](https://reviewable.io/reviews/cockroachdb/docs/7782#-MFkdxhA5aQYmjH-S68o:-MFkdxhA5aQYmjH-S68p:bez1ksl) ([raw file](https://github.com/cockroachdb/docs/blob/b053df54444baed416cbf03d8c6442bb2f79e9e5/v20.1/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md#L296)):*
> ```Markdown
> ### Create inbound rules
> 
> For each region, navigate to the Security Groups section of the [Amazon EC2 console](https://console.aws.amazon.com/ec2/) and locate the security group that enables communication between nodes int he cluster. It should have a name like `ClusterSharedNodeSecurityGroup`. [Add Custom TCP inbound rules](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html#adding-security-group-rule) to this security group to allow TCP communication on two ports:
> ```

int he > in the

jseldess:
___
*[v20.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md, line 296 at r3](https://reviewable.io/reviews/cockroachdb/docs/7782#-MFkdtG71cmLivgMSDRk:-MFkdtG71cmLivgMSDRl:bez1ksl) ([raw file](https://github.com/cockroachdb/docs/blob/b053df54444baed416cbf03d8c6442bb2f79e9e5/v20.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.md#L296)):*
> ```Markdown
> ### Create inbound rules
> 
> For each region, navigate to the Security Groups section of the [Amazon EC2 console](https://console.aws.amazon.com/ec2/) and locate the security group that enables communication between nodes int he cluster. It should have a name like `ClusterSharedNodeSecurityGroup`. [Add Custom TCP inbound rules](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html#adding-security-group-rule) to this security group to allow TCP communication on two ports:
> ```

int he > in the

PR #7783, opened 2020-07-24
Sidenav experiment

jseldess:
This is an experiment to test out some navigation approaches
emerging from the Docs UX redesign project by Lucia and Lauren.

In this experiment:

- I moved the entirely of the CC docs to the first section of
  a shared sidenav. To do this, I created a root-level directory
  called cockroachcloud and moved all of the cc-specific markdown
  files there, out of the 20.1 directory. CC is not versioned,
  so this type of move is inevitable in any case. This change
  makes it so that a local `make standard` build now builds both
  CC and CRDB docs. Our PR and live builds, however, still will
  build the docs the previous way. If we like the approach I'm
  taking, we'll need YZ's help to adapt the build process.
- The only change I made in the CC section is renaming "Quickstart"
  to "Deploy a Free Cluster" and "Create a Cluster" to "Deploy
  a Dedicated Cluster". I also removed the "Create Your Account"
  page as I think linking to the account creation page is
  enough; it's an easy workflow that doesn't need docs, I think.
- In the CockroachDB section, I moved Maintain under Deploy as
  Cluster Maintenance, and I move everything previously under
  Tutorials into other sections: Database Features tutorials
  are now under Get Started > Explore Database Features; Sample
  Apps tutorials are now under Develop > Sample Apps; Third-Party
  Database Tools tutorials are now under Develop > Third-Party
  Database Tools. This isn't perfect, but it's an attempt to
  do away with a "Tutorials" sections distinct from the guidance
  sections.
- Following CockroachCloud and CockroachDB, the sidenav has
  external links to Cockroach University and Blog. This is also
  based on the redesign proposal. In the new design, external
  links will be visually identified as such.
- I also made a css change so that the top-level of the sidenav
  is no longer uppercase, again to match the redesign proposal.

Thoughts/questions:
- In this approach, there are times we go 4 levels deep, which
  may be too much and isn't currently styled correctly in any
  case.
- I played around with making the first two sections focused on
  "guidance" and moving all reference material to a third top-level
  section, e.g.:
    CockroachCloud
    CockroachDB
    Reference
    Cockroach University
    Blog

    or

    CockroachCloud Guide
    CockroachDB Guide
    CockroachDB Reference
    ...

  I'm still not sure Cockroach U and Blog deserve a top-level placement
  in the sidenav, especially if we call them out prominently on a new
  Docs landing page.


lnhsingh:
I agree, I think going 4 deep in the sidenav is too much. I like your suggestion of pulling out Reference into it's own top-level nav, which eliminates some of the nested topics. The other topic that has deep nesting is Sample Apps, which we could pull out from under Develop. (Back up and Restore Data also is nested but that can be easily un-nested). I don't have a preference on the naming scheme (CRDB Reference vs Reference).

I think it's nice to include the links to Cockroach U and the Blog in the main nav since they'll be always visible. In the new design, the logo in the top left corner will be changed to say "Docs" (instead of linking to the Marketing site), so the Blog link is a quick way to get to the Marketing site in the new design.

Issue #7780, opened 2020-07-23
SIAM: Add example for `sql-addr` flag usage


Amruta-Ranade:

Amruta-Ranade:
Addressed in #7790

PR #7790, opened 2020-07-27
Added sql_addr example

Amruta-Ranade:
Addresses https://github.com/cockroachdb/docs/issues/7780


v20.1/cockroach-start.md
```diff
@@ -485,6 +485,14 @@ $ cockroach sql --insecure --port 26261
 (1 row)
 ~~~
 
+### Start a cluster with separate RPC and SQL networks
+
+Start a cluster with separate RPC and SQL networks to avoid bandwidth interference and add a level of isolation against network attacks as a measure of 'defense in depth'.
```
knz, line 490: Protection against bandwidth interference is really only relevant when using different addresses.

Amruta-Ranade: How about "Start a cluster with separate RPC and SQL networks to partition traffic and add a level of isolation against network attacks as a measure of 'defense in depth'."

dbist:
it would be a welcome addition to also add `--advertise-sql-addr` to this example as my understanding this flag optionally works in tandem with `--sql-addr`?

knz:
We discussed this and we think that there are no good reasons for users to use SQL advertise addr in most cases. We might even want to hide the flag in the command line help.


Artem Ervits <notifications@github.com> schreef op 3 augustus 2020 13:58:09 CEST:
>it would be a welcome addition to also add `--advertise-sql-addr` to
>this example as my understanding this flag optionally works in tandem
>with `--sql-addr`?
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/cockroachdb/docs/pull/7790#issuecomment-667982289

-- 

knz:
Ben has a different view that he explained here: https://github.com/cockroachdb/cockroach/issues/52266#issuecomment-668198779

Basically Ben suggests that we should document `--sql-advertise-addr` in parallel with `--advertise-addr`, in the same guides: whenever there's a need to use `--advertise-addr`, the same situation also requires `--sql-advertise-addr`.

Amruta-Ranade:
Okay, I am even more confused now. I don't think I have heard of the `--sql-advertise-addr` before. Is it the same as `--advertise-sql-addr`? Also, I really need clarity on what's most useful for majority of users the way the product is implemented as of now.

jseldess:
___
*[v20.1/cockroach-start.md, line 490 at r3](https://reviewable.io/reviews/cockroachdb/docs/7790#-MDyc0ZGBL9S0SSqProX:-MDyc0ZGBL9S0SSqProY:bfd9848) ([raw file](https://github.com/cockroachdb/docs/blob/cf3c9498f82baa6dfe0428fc6c874cfcd1deb1e7/v20.1/cockroach-start.md#L490)):*
> ```Markdown
> ### Start a cluster with separate RPC and SQL networks
> 
> In enterprise deployments, separating RPC and SQL networks to partition traffic adds a level of isolation against network attacks as a measure of "defense in depth". To achieve the partition, the first step is to ensure that the server where CockroachDB is running has two or more network interfaces, with different IP addresses.
> ```

@amruta-ranade, on another read, I dislike using jargon like "defense in depth" generally, as it assume a broadly shared vocabulary and knowledge. That said, in this case, I'm ok leaving it as long as we make it a link to https://en.wikipedia.org/wiki/Defense_in_depth_(computing).

knz:
> Okay, I am even more confused now. I don't think I have heard of the --sql-advertise-addr before. 

Sorry I was mistaken. You are right the spelling is `--advertise-sql-addr`.

> I really need clarity on what's most useful for majority of users the way the product is implemented as of now.

Here's the rule:

| Deployment | Single DC | Multi-DC |
|------------|-----------|----------|
| Combined RPC and SQL | `--listen-addr` | `--listen-addr --advertise-addr` |
| Separate RPC and SQL ports (more security) | `--listen-addr` `--sql-addr` | `--listen-addr` `--advertise-addr` `--sql-addr` `--advertise-sql-addr` |

The transition from single DC to multi-DC is "one level of fault tolerance higher"

The transition from "combined RPC and SQL" to "separate RPC and SQL" is a separate "one level higher" but this time it's security, not fault tolerance

I don't know what is "most common"

- If users want both security and fault tolerance they want split ports and multi-DC and need to use 4 flags.
- If they only need security and only use one DC they can use just 2 flags.
- If they only need fault tolerance and are not too serious on security they can use just 2 (other) flags.

Amruta-Ranade:
Thanks for the clarification. If I am reading this thread right, we should add `--advertise-sql-addr` to the example as per Artem's suggestion. Thoughts?

jseldess:
I don't think we should think we should overload this example. The type of context you're providing, @knz, should go into updating the Networking recommendations in the production checklist, though I know we're not doing that at this time. 

I think the current example is perfect fine for the case when only `--listen-addr` and `--sql-addr` are set and those addresses are advertised by default. 

I recommend that we add another quick example to the end:

> Now let's suppose that...<something about the case where you want to use different listen and advertise addresses for both rpc and sql traffic. 

If we can set describe that in as simple a way as possible, the example command will be super easy to add as well. 

@Amruta-Ranade, @knz, thoughts?

knz:
Yes, but only when it's strictly needed: when there are multiple nodes running on different address ranges or there's NAT in-between. It's not needed if there's e.g. VPC peering or the nodes are running on the same (local) network.

knz:
> I recommend that we add another quick example to the end:

> >   Now let's suppose that...<something about the case where you want to use different listen and advertise addresses for both rpc and sql traffic.

> If we can set describe that in as simple a way as possible, the example command will be super easy to add as well.

Yes I like that too

knz:
From a UX perspective, would it help if the product told the user "You can't omit `--advertise-sql-addr` if you already use `--advertise-addr`"?

Amruta-Ranade:
Discussion with @bdarnell:
- What we have documented (flags description + example for `sql-addr`) is fine for now. Not a good use of time to keep polishing the docs until the feature's functionality is expanded.
- Swap the order of the `sql-addr` and `advertise-sql-addr` flags in the table 
- Reword the example text to clarify that the flag is useful only when used with network access setup/firewall rules. @bdarnell to provide the text.

Amruta-Ranade:
@bdarnell Ready for your review!

PR #7793, opened 2020-07-27
Removed paragraph about SAVEPOINTs from run_transaction() recommendation

### LINKED ISSUE #6390: Update SQLAlchemy docs to remove incorrect info that SAVEPOINT is faster
In the [Build a Python app with SQLAlchemy](https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb-sqlalchemy.html) docs we say re: the `run_transaction()` wrapper method that:

> It will run your transactions to completion much faster than a naive implementation that created a fresh transaction after every retry error. Because of the way the CockroachDB dialect's driver structures the transaction attempts (using a [`SAVEPOINT`](savepoint.html) statement under the hood, which has a slightly different meaning in CockroachDB than in some other databases), the server is able to preserve some information about the previously attempted transactions to allow subsequent retries to complete more easily.

This is not actually correct.  Using an application-level retry loop that does not use `SAVEPOINT` is basically fine most of the time, and performs well if you add exponential backoff.

For context see the discussion in [this mailing list thread](https://groups.google.com/a/cockroachlabs.com/d/msg/eng/sej4jIqGAwo/0sEUSC06AwAJ), and also https://github.com/cockroachdb/cockroach/issues/5935

ericharmeling:
Fixes #6390.


ericharmeling:
@rafiss 
Friendly review reminder :)

PR #7797, opened 2020-07-28
Remove erroneous link to non-public setting

### LINKED ISSUE #7796: Fix broken link to non-public cluster setting on transaction retry error reference page
Specifically, we mention this setting, but it's non-public, and thus does not appear in the "normal" cluster settings page:

```
kv.closed_timestamp.target_duration                      | 3s                                   | d            | false  | if nonzero, attempt to provide closed timestamp notifications for timestamps trailing cluster time by approximately this duration
```

rmloveland:
... since there's no there there.

Fixes #7796.


rmloveland:
Eric, I think this is a pretty trivial PR (removing links from text) but Murphy says if I don't ask for a review something will go wrong

PR #7817, opened 2020-07-29
Fix logic error in timestamp ordering remark

rmloveland:
Summary of changes:

- In the section on `RETRY_WRITE_TOO_OLD`, we are talking about the
  ordering of two transactions A and B.  I had erroneously said that if
  B comes after A, then B has a lower timestamp than A.  In fact, if B
  comes after A, then B has a *higher* timestamp than A.  This PR fixes
  that error.

### DIFF
--- a/v20.1/transaction-retry-error-reference.md
+++ b/v20.1/transaction-retry-error-reference.md
@@ -69,7 +69,7 @@ TransactionRetryWithProtoRefreshError: ... RETRY_WRITE_TOO_OLD ...
 
 _Description_:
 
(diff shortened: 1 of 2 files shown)


rmloveland:
Eric this is a one-word change I hope you don't mind rubber stamping (mostly want another pair of eyes to avoid Murphy's law)

It's a fix for a logic error that @andreimatei mentioned [in this comment](https://github.com/cockroachdb/docs/pull/7684#pullrequestreview-449274824) that I dropped somehow, which was kindly pointed out again by @RaduBerinde (thanks!)

PR #7819, opened 2020-07-29
EXPORT and CDC file naming conventions and best practices

### LINKED ISSUE #7374: EXPORT and CDC file naming conventions and best practices
We ran into a customer who was confused why their EXPORTs were being overwritten. What they didn't know is that we don't have unique names per export, so they had to point to a different bucket.

Similarly, there's confusion about what the file naming convention is for CDC to cloud sinks and how we can parse that out. @BramGruneir has done some work clarifying it (and even writing a regex parser--many thanks!)

*Deliverables*
- Update EXPORT and CDC with file naming convention
- Best practices given the naming limitations--namely that EXPORT should be to a new bucket/directory

lnhsingh:
CDC changes:
- Created new include files for external URL content. Only updated the Example file URLs at the bottom.
- Added Files section to describe the naming conventions of the files that are emitted.

EXPORT changes:
- Added notes about giving each export a unique destination to avoid overwriting other exports.

Closes #7374.


v20.1/export.md
```diff
@@ -36,19 +36,19 @@ Only members of the `admin` role can run `EXPORT`. By default, the `root` user b
 
  Parameter | Description
 -----------|-------------
- `file_location` | Specify the URL of the file location where you want to store the exported CSV data.
+ `file_location` | Specify the [URL of the file location](#export-file-url) where you want to store the exported CSV data.<br><br>Note: Exports do not generate unique names, so each export should have a unique destination to avoid overwriting.
```
mwang1026, line 39: `Exports do not generate unique names` isn't clear to me -- since it does create unique file names *within* the export, just not *across* exports.

Otherwise lgtm

mwang1026:
LGTM

PR #7960, opened 2020-08-10
Generated functions (built-ins, aggregates, operators) doc update for v20.2 (alpha 3)

ericharmeling:
Generated functions (built-ins, aggregates, operators) update as of https://github.com/cockroachdb/cockroach/releases/tag/provisional_202008031850_v20.2.0-alpha.3.

@rmloveland FYI, as some of these might touch geospatial.


rmloveland:
Thanks for the heads up Eric.  FYI the build breakage is due to some anchor links I added to a few of the operators a while back since otherwise they are pretty difficult to find by doing an in-page search.

(In general I think our approach to BIFs needs an overhaul but that is a discussion for another time. :-} )

PR #7992, opened 2020-08-12
Test search synonyms

### LINKED ISSUE #7759: Look into creating synonyms in Algolia
Investigate whether Algolia supports defining synonyms for search terms. 

Example use case:
https://github.com/cockroachdb/docs/issues/7758#issuecomment-661110936

Additional use case:
https://cockroachlabs.slack.com/archives/G0108LXFWTZ/p1597239856001000 (spatial <-> geospatial)

yingzhuchin:
Fixes #7759

### DIFF
--- a/jekyll-algolia-dev/lib/jekyll/algolia/indexer.rb
+++ b/jekyll-algolia-dev/lib/jekyll/algolia/indexer.rb
@@ -284,6 +284,22 @@ module Jekyll
         nil
       end
 
+      def self.update_synonyms
+        index.save_synonym('CBO', {
+          objectID: 'CBO',
+          type: 'synonym',
+          synonyms: ['CBO', 'cost-based optimizer', 'cost based optimizer']
+        }, false)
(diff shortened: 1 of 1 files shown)


jseldess:
This is great, @yingzhuchin. We should document how this and how to get more synonyms added. 

cc @rmloveland, @lnhsingh

yingzhuchin:
Thanks, Jesse! Good call. I've opened https://github.com/cockroachdb/docs/issues/7994 to support "self-service" for adding new synonyms, and will make sure to document that when done.

jseldess:
@yingzhuchin, I noticed in the Algolia docs that synonyms can also be added via the Algolia dashboard https://www.algolia.com/doc/guides/managing-results/optimize-search-results/adding-synonyms/how-to/managing-synonyms-from-the-dashboard/#synonyms-in-the-dashboard. Can you share your thoughts about using the api over the dashboard?

Issue #6667, opened 2020-02-24
Developer Guide v2


rmloveland:
This issue tracks tasks we'd like to add to the Developer Guide for v2 (the version in #5865 represents "v1").

- [ ] Show examples of `UPSERT`, and when to use
- [ ] Update 'Insert data' page with more info about e.g., `INSERT`s when there are uniqueness conflicts (`INSERT ... ON CONFLICT` or the aforementioned `UPSERT`)
- [ ] Update the 'Schema Design' section of the 'Make Queries Fast' page with an illustrative example (like we have for SQL tuning)
- [ ] Add batch write and delete loops
- [ ] Connection pooling

ericharmeling:
All topics mentioned in the description of this issue are included in the proposed content plan for the Developer Guide v2 (see the [Developer Guide Overhaul Atlassian page](https://cockroachlabs.atlassian.net/wiki/spaces/PROD/pages/1076002864/Developer+Guide+Overhaul) and linked tracking docs).

We'll be tracking this project with more granular issues. For high-level planning view, see https://github.com/cockroachdb/docs/projects/63.

PR #7999, opened 2020-08-13
Large deletes guidance and TTL workarounds

### LINKED ISSUE #4818: Deleting large amounts of data
There are two opportunities here:
- [ ] For deleting all rows in a table, consistently recommend `TRUNCATE` instead of `DELETE` without a `WHERE` clause.
    - The [SQL Best Practices](https://www.cockroachlabs.com/docs/v19.1/performance-best-practices-overview.html#use-truncate-instead-of-delete-to-delete-all-rows-in-a-table) doc already says this. Can probably say it clearer.
    - Our [`TRUNCATE`](https://www.cockroachlabs.com/docs/stable/truncate.html) doc suggest that, when deleting all rows from a table, it's best to use `TRUNCATE` for larger tables and `DELETE` for smaller tables. Based on feedback from @petermattis, we should stop encouraging any use of `DELETE` without `WHERE` and always recommend `TRUNCATE`. 
    - Given the above, the [`DELETE`](https://www.cockroachlabs.com/docs/v19.1/delete.html) page should much more clearly point to `TRUNCATE` for removing all rows of a table. Probably needs to be a callout in intro or something. And the [Delete all rows](https://www.cockroachlabs.com/docs/v19.1/delete.html#delete-all-rows) example needs to be adjusted as well.
-  [ ] For deleting large chunks of data from a table, do we need any recommendations or best practices beyond what's already called out in the `DELETE` docs about [disk space and select performance](https://www.cockroachlabs.com/docs/v19.1/delete.html#disk-space-usage-after-deletes)?

### LINKED ISSUE #5592: Document batch DELETEs
A customer pointed out that our `DELETE` doc doesn't have any mention of how to best delete multiple rows. 

Screen Shot 2019-10-10 at 9 08 38 AM

This is a gap. 

- Our SQL best practices call out "Delete multiple rows" in the ["Multi-row DML best practices"](https://www.cockroachlabs.com/docs/stable/performance-best-practices-overview.html#multi-row-dml-best-practices) section, but don't link to a meaningful example. 
- Also on that page, there's no mention of this in the ["Bulk deletion best practices"](https://www.cockroachlabs.com/docs/stable/performance-best-practices-overview.html#bulk-deletion-best-practices) section, in contrast to the "Insert delete best practices" section.  

Suggestion:

1. Extend the [`DELETE`](https://www.cockroachlabs.com/docs/stable/delete.html) doc to include a "Performance best practices" section similar to on [`INSERT`](https://www.cockroachlabs.com/docs/stable/insert.html#performance-best-practices), and add an explicit "Delete multiple rows" example similar to what we have on [`INSERT`](https://www.cockroachlabs.com/docs/stable/insert.html#insert-multiple-rows-into-an-existing-table) and [`UPSERT`](https://www.cockroachlabs.com/docs/stable/upsert.html#upsert-multiple-rows).
2. Update the SQL Best Practices doc (issues described above).
3. Feature this ability in the ["Delete rows in a table"](https://www.cockroachlabs.com/docs/stable/learn-cockroachdb-sql.html#delete-rows-in-a-table section of "Learn CockroachDB", like we do for [`Insert rows into a table`](https://www.cockroachlabs.com/docs/stable/learn-cockroachdb-sql.html#insert-rows-into-a-table).

Sidenote: As a follow-up, we may want to investigate doing similar work for `UPDATE` and `UPSERT`.

### LINKED ISSUE #5647: sql: provide documentation on row-level ttl workarounds
We don't yet support [row-level TTL](https://github.com/cockroachdb/cockroach/issues/20239). This feature is useful in that it cuts down on the effort developers need to take to delete and manage data.

In the meantime, we should document workarounds such as using cron jobs to periodically delete certain data in a performance first manner. 

cc @rkruze who was asking about this.

### LINKED ISSUE #8137: sql: expose mvcc timestamps to SQL
PR: https://github.com/cockroachdb/cockroach/pull/51494

From release notes:
> Exposed the MVCC timestamp of each row as a system column on the table. This column is named `crdb_internal_mvcc_timestamp` and is accessible only in a limited set of contexts. [#51494][#51494] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #5592.
Fixes #4818. 
Fixes #5647. 
Fixes #8137.

Adding examples in other languages will likely be part of the larger [Dev Guide v2 initiative](https://github.com/cockroachdb/docs/issues/6667).

### DIFF
--- a/v19.2/delete.md
+++ b/v19.2/delete.md
@@ -10,6 +10,12 @@ The `DELETE` [statement](sql-statements.html) deletes rows from a table.
 
 {{site.data.alerts.callout_info}}To delete columns, see <a href="drop-column.html"><code>DROP COLUMN</code></a>.{{site.data.alerts.end}}
 
+## Performance best practices
+
+- To delete all rows in a table, use [`TRUNCATE`](truncate.html) instead of `DELETE`. Unless your table is small (less than 1000 rows), `TRUNCATE` will be more performant.
+
+- To delete a large number of rows, we recommend iteratively deleting batches of rows until all of the unwanted rows are deleted. For more information, see [batch deletes](#batch-deletes).
+
 ## Required privileges
 
 The user must have the `DELETE` and `SELECT` [privileges](authorization.html#assign-privileges) on the table.
@@ -33,7 +39,7 @@ table td:first-child {
  `common_table_expr` | See [Common Table Expressions](common-table-expressions.html).
  `table_name` | The name of the table that contains the rows you want to update.
  `AS table_alias_name` | An alias for the table name. When an alias is provided, it completely hides the actual table name.
-`WHERE a_expr`| `a_expr` must be an expression that returns Boolean values using columns (e.g., `<column> = <value>`). Delete rows that return   `TRUE`.<br><br/>__Without a `WHERE` clause in your statement, `DELETE` removes all rows from the table.__
+`WHERE a_expr`| `a_expr` must be an expression that returns Boolean values using columns (e.g., `<column> = <value>`). Delete rows that return `TRUE`.<br><br/>__Without a `WHERE` clause in your statement, `DELETE` removes all rows from the table. To delete all rows in a table, we recommend using [`TRUNCATE`](truncate.html) instead of `DELETE`.__
  `sort_clause` | An `ORDER BY` clause. <br /><br />See [Ordering of rows in DML statements](query-order.html#ordering-rows-in-dml-statements) for more details.
  `limit_clause` | A `LIMIT` clause. See [Limiting Query Results](limit-offset.html) for more details.
  `RETURNING target_list` | Return values based on rows deleted, where `target_list` can be specific column names from the table, `*` for all columns, or computations using [scalar expressions](scalar-expressions.html). <br><br>To return nothing in the response, not even the number of rows updated, use `RETURNING NOTHING`.
@@ -54,7 +60,7 @@ due to the fact that CockroachDB retains [the ability to query tables
 historically](https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/).
 
 If disk usage is a concern, the solution is to
-[reduce the time-to-live](configure-replication-zones.html) (TTL) for
+[reduce the garbage collection time-to-live](configure-replication-zones.html) (TTL) for
 the zone by setting `gc.ttlseconds` to a lower value, which will cause
 garbage collection to clean up deleted objects (rows, tables) more
 frequently.
@@ -65,7 +71,7 @@ Queries that scan across tables that have lots of deleted rows will
 have to scan over deletions that have not yet been garbage
 collected. Certain database usage patterns that frequently scan over
 and delete lots of rows will want to reduce the
-[time-to-live](configure-replication-zones.html) values to clean up
+[garbage collection time-to-live](configure-replication-zones.html) values to clean up
 deleted rows more frequently.
 
 ## Sorting the output of deletes
@@ -75,12 +81,6 @@ deleted rows more frequently.
 For more information about ordering query results in general, see
 [Ordering Query Results](query-order.html) and [Ordering of rows in
 DML statements](query-order.html#ordering-rows-in-dml-statements).
-
-## Delete performance on large data sets
-
-If you are deleting a large amount of data using iterative `DELETE ... LIMIT` statements, you are likely to see a drop in performance for each subsequent `DELETE` statement.
-
-For an explanation of why this happens, and for instructions showing how to iteratively delete rows in constant time, see [Why are my deletes getting slower over time?](sql-faqs.html#why-are-my-deletes-getting-slower-over-time).
 
 ## Force index selection for deletes
 
@@ -108,52 +108,154 @@ To view how the index hint modifies the query plan that CockroachDB follows for 
 
 For examples, see [Delete with index hints](#delete-with-index-hints).
 
+## Batch deletes
+
+To delete a large number of rows (i.e., tens of thousands of rows or more), we recommend iteratively deleting subsets of the rows that you want to delete, until all of the unwanted rows have been deleted. You can write a script to do this, or you can write a loop into your application.
+
+In the sections below, we provide guidance on batch deleting with the `DELETE` query filter [on an indexed column](#batch-delete-on-an-indexed-column) and [on a non-indexed column](#batch-delete-on-a-non-indexed-column). Filtering on an indexed column is both simpler to implement and more efficient, but adding an index to a table can slow down insertions to the table and may cause bottlenecks. Queries that filter on a non-indexed column must perform at least one full-table scan, a process that takes time proportional to the size of the entire table.
+
+### Batch delete on an indexed column
+
+For high-performance batch deletes, we recommending filtering the `DELETE` query on an [indexed column](indexes.html).
+
+{{site.data.alerts.callout_info}}
+Having an indexed filtering column can make delete operations faster, but it might lead to bottlenecks in execution, especially if the filtering column is a [timestamp](timestamp.html).
+{{site.data.alerts.end}}
+
+Each iteration of a batch-delete loop should execute a transaction containing a single `DELETE` query. When writing this `DELETE` query:
+
+- Use a `WHERE` clause to filter on a column that identifies the unwanted rows. If the filtering column is not the primary key, the column should have [a secondary index](indexes.html). Note that if the filtering column is not already indexed, it is not beneficial to add an index just to speed up batch deletes. Instead, consider [batch deleting on non-indexed columns](#batch-delete-on-a-non-indexed-column).
+- To ensure that rows are efficiently scanned in the `DELETE` query, add an [`ORDER BY`](query-order.html) clause on the filtering column.
+- Use a [`LIMIT`](limit-offset.html) clause to limit the number of rows to the desired batch size. To determine the optimal batch size, try out different batch sizes (1,000 rows, 10,000 rows, 100,000 rows, etc.) and monitor the change in performance.
+- Add a `RETURNING` clause to the end of the query that returns the filtering column values of the deleted rows. Then, using the values of the deleted rows, update the filter to match only the subset of remaining rows to delete. This narrows each query's scan to the fewest rows possible, and [preserves the performance of the deletes over time](#preserving-delete-performance-over-time). This pattern assumes that no new rows are generated that match on the `DELETE` filter during the time that it takes to perform the delete.
+
+For example, suppose that you want to delete all rows in the [`tpcc`](cockroach-workload.html#tpcc-workload) `new_order` table where `no_w_id` is less than `5`, in batches of 5,000 rows. To do this, you can write a script that loops over batches of 5,000 rows, following the `DELETE` query guidance provided above. Note that in this case, `no_w_id` is the first column in the primary index, and, as a result, you do not need to create a secondary index on the column.
+
+In Python, the script would look similar to the following:
+
+{% include copy-clipboard.html %}
+~~~ python
+#!/usr/bin/env python3
+
+import psycopg2
+import psycopg2.sql
+import os
+
+conn = psycopg2.connect(os.environ.get('DB_URI'))
+filter = 4
+lastrow = None
+
+while True:
+  with conn:
+    with conn.cursor() as cur:
+        if lastrow:
+            filter = lastrow[0]
+        query = psycopg2.sql.SQL("DELETE FROM new_order WHERE no_w_id <= %s ORDER BY no_w_id DESC LIMIT 5000 RETURNING no_w_id")
+        cur.execute(query, (filter,))
+        print(cur.statusmessage)
+        if cur.rowcount == 0:
+            break
+        lastrow = cur.fetchone()
+
+conn.close()
+~~~
+
+This script iteratively deletes rows in batches of 5,000, until all of the rows where `no_w_id <= 4` are deleted. Note that at each iteration, the filter is updated to match a narrower subset of rows.
+
+### Batch delete on a non-indexed column
+
+If you cannot index the column that identifies the unwanted rows, we recommend defining the batch loop to execute separate read and write operations at each iteration:
+
+1. Execute a [`SELECT` query](selection-queries.html) that returns the primary key values for the rows that you want to delete. When writing the `SELECT` query:
+    - Use a `WHERE` clause that filters on the column identifying the rows.
+    - Add an [`AS OF SYSTEM TIME` clause](as-of-system-time.html) to the end of the selection subquery, or run the selection query in a separate, read-only transaction with [`SET TRANSACTION AS OF SYSTEM TIME`](as-of-system-time.html#using-as-of-system-time-in-transactions). This helps to reduce [transaction contention](transactions.html#transaction-contention).
+    - Use a [`LIMIT`](limit-offset.html) clause to limit the number of rows queried to a subset of the rows that you want to delete. To determine the optimal `SELECT` batch size, try out different sizes (10,000 rows, 100,000 rows, 1,000,000 rows, etc.), and monitor the change in performance. Note that this `SELECT` batch size can be much larger than the batch size of rows that are deleted in the subsequent `DELETE` query.
+    - To ensure that rows are efficiently scanned in the subsequent `DELETE` query, include an [`ORDER BY`](query-order.html) clause on the primary key.
+
+2. Write a nested `DELETE` loop over the primary key values returned by the `SELECT` query, in batches smaller than the initial `SELECT` batch size. To determine the optimal `DELETE` batch size, try out different sizes (1,000 rows, 10,000 rows, 100,000 rows, etc.), and monitor the change in performance. Where possible, we recommend executing each `DELETE` in a separate transaction.
+
+For example, suppose that you want to delete all rows in the [`tpcc`](cockroach-workload.html#tpcc-workload) `history` table that are older than a month. You can create a script that loops over the data and deletes unwanted rows in batches, following the query guidance provided above.
+
+In Python, the script would look similar to the following:
+
+{% include copy-clipboard.html %}
+~~~ python
+#!/usr/bin/env python3
+
+import psycopg2
+import os
+import time
+
+conn = psycopg2.connect(os.environ.get('DB_URI'))
+
+while True:
+    with conn:
+        with conn.cursor() as cur:
+            cur.execute("SET TRANSACTION AS OF SYSTEM TIME '-5s'")
+            cur.execute("SELECT h_w_id, rowid FROM history WHERE h_date < current_date() - INTERVAL '1 MONTH' ORDER BY h_w_id, rowid LIMIT 20000")
+            pkvals = list(cur)
+    if not pkvals:
+        return
+    while pkvals:
+        batch = pkvals[:5000]
+        pkvals = pkvals[5000:]
+        with conn:
+            with conn.cursor() as cur:
+                cur.execute("DELETE FROM history WHERE (h_w_id, rowid) = ANY %s", (batch,))
+                print(cur.statusmessage)
+    del batch
+    del pkvals
+    time.sleep(5)
+
+conn.close()
+~~~
+
+At each iteration, the selection query returns the primary key values of up to 20,000 rows of matching historical data from 5 seconds in the past, in a read-only transaction. Then, a nested loop iterates over the returned primary key values in smaller batches of 5,000 rows. At each iteration of the nested `DELETE` loop, a batch of rows is deleted. After the nested `DELETE` loop deletes all of the rows from the initial selection query, a time delay ensures that the next selection query reads historical data from the table after the last iteration's `DELETE` final delete.
+
+### Batch-delete "expired" data
+
+CockroachDB does not support Time to Live (TTL) on table rows. To delete "expired" rows, we recommend automating a batch delete process using a job scheduler like `cron`.
+
+For example, suppose that every morning you want to delete all rows in the [`tpcc`](cockroach-workload.html#tpcc-workload) `history` table that are older than a month. To do this, you could use the example Python script that [batch deletes on the non-indexed `h_date` column](#batch-delete-on-a-non-indexed-column).
+
+To run the script with a daily `cron` job:
+
+1. Make the file executable:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ chmod +x cleanup.py
+    ~~~
+
+2. Create a new `cron` job:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ crontab -e
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ txt
+    30 10 * * * DB_URI='cockroachdb://user@host:26257/bank' cleanup.py >> ~/cron.log 2>&1
+    ~~~
+
+Saving the `cron` file will install a new job that runs the `cleanup.py` file every morning at 10:30 A.M., writing the results to the `cron.log` file.
+
+### Preserving `DELETE` performance over time
+
+CockroachDB relies on [multi-version concurrency control (MVCC)](architecture/storage-layer.html#mvcc) to process concurrent requests while guaranteeing [strong consistency](frequently-asked-questions.html#how-is-cockroachdb-strongly-consistent). As such, when you delete a row, it is not immediately removed from disk. The MVCC values for the row will remain until the garbage collection period defined by the [`gc.ttlseconds`](configure-replication-zones.html#gc-ttlseconds) variable in the applicable [zone configuration](show-zone-configurations.html) has passed. By default, this period is 25 hours.
+
+This means that with the default settings, each iteration of your `DELETE` statement must scan over all of the rows previously marked for deletion within the last 25 hours. If you try to delete 10,000 rows 10 times within the same 25 hour period, the 10th command will have to scan over the 90,000 rows previously marked for deletion.
+
+To preserve performance over iterative `DELETE` queries, we recommend taking one of the following approaches:
+
+- At each iteration, update the `WHERE` clause to filter only the rows that have not yet been marked for deletion. For an example, see [Batch-delete on an indexed filter](#batch-delete-on-an-indexed-column) above.
+- At each iteration, first use a `SELECT` statement to return primary key values on rows that are not yet deleted. Rows marked for deletion will not be returned. Then, use a nested `DELETE` loop over a smaller batch size, filtering on the primary key values. For an example, see [Batch delete on a non-indexed column](#batch-delete-on-a-non-indexed-column) above.
+- To iteratively delete rows in constant time, using a simple `DELETE` loop, you can [alter your zone configuration](configure-replication-zones.html#overview) and change `gc.ttlseconds` to a low value like 5 minutes (i.e., `300`), and then run your `DELETE` statement once per GC interval.
+
 ## Examples
 
 {% include {{page.version.version}}/sql/movr-statements.md %}
 
-### Delete all rows
-
-You can delete all rows from a table by not including a `WHERE` clause in your `DELETE` statement.
-
-{{site.data.alerts.callout_info}}
-If the [`sql_safe_updates`](cockroach-sql.html#allow-potentially-unsafe-sql-statements) session variable is set to `true`, the client will prevent the update. `sql_safe_updates` is set to `true` by default.
-{{site.data.alerts.end}}
-
-{% include copy-clipboard.html %}
-~~~ sql
-> DELETE FROM vehicle_location_histories;
-~~~
-
-~~~
-pq: rejected: DELETE without WHERE clause (sql_safe_updates = true)
-~~~
-
-You can use a [`SET`](set-vars.html) statement to set session variables.
-
-{% include copy-clipboard.html %}
-~~~ sql
-> SET sql_safe_updates = false;
-~~~
-
-{% include copy-clipboard.html %}
-~~~ sql
-> DELETE FROM vehicle_location_histories;
-~~~
-
-~~~
-DELETE 1000
-~~~
-
-{{site.data.alerts.callout_success}}
-Unless your table is small (less than 1000 rows), using [`TRUNCATE`][truncate] to delete the contents of a table will be more performant than using `DELETE`.
-{{site.data.alerts.end}}
-
-### Delete specific rows
-
-When deleting specific rows from a table, the most important decision you make is which columns to use in your `WHERE` clause. When making that choice, consider the potential impact of using columns with the [Primary Key](primary-key.html)/[Unique](unique.html) constraints (both of which enforce uniqueness) versus those that are not unique.
-
-#### Delete rows using Primary Key/unique columns
+### Delete rows using Primary Key/unique columns
 
 Using columns with the [Primary Key](primary-key.html) or [Unique](unique.html) constraints to delete rows ensures your statement is unambiguous&mdash;no two rows contain the same column value, so it's less likely to delete data unintentionally.
 
@@ -167,7 +269,7 @@ In this example, `code` is our primary key and we want to delete the row where t
 DELETE 1
 ~~~
 
-#### Delete rows using non-unique columns
+### Delete rows using non-unique columns
 
 Deleting rows using non-unique columns removes _every_ row that returns `TRUE` for the `WHERE` clause's `a_expr`. This can easily result in deleting data you didn't intend to.
 
@@ -206,7 +308,6 @@ To retrieve specific columns, name them in the `RETURNING` clause.
   study_piece_war        | {"type": "percent_discount", "value": "10%"}
   tv_this_list           | {"type": "percent_discount", "value": "10%"}
 (5 rows)
-
 ~~~
 
 #### Change column labels
@@ -317,36 +418,10 @@ If you provide an index hint (i.e. force the index selection) to use the primary
 - [`INSERT`](insert.html)
 - [`UPDATE`](update.html)
 - [`UPSERT`](upsert.html)
-- [`TRUNCATE`][truncate]
+- [`TRUNCATE`](truncate.html)
 - [`ALTER TABLE`](alter-table.html)
 - [`DROP TABLE`](drop-table.html)
 - [`DROP DATABASE`](drop-database.html)
 - [Other SQL Statements](sql-statements.html)
 - [Limiting Query Results](limit-offset.html)
 
-<!-- Reference Links -->
-
-[truncate]: truncate.html
-
-<!--
-
-SQL for example table:
-
-CREATE TABLE account_details (
-       account_id INT PRIMARY KEY,
-       balance FLOAT,
-       account_type VARCHAR
-);
-
-INSERT INTO account_details (account_id, balance, account_type) VALUES (1, 32000, 'Savings');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (2, 30000, 'Checking');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (3, 30000, 'Savings');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (4, 22000, 'Savings');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (5, 43696.95, 'Checking');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (6, 23500, 'Savings');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (7, 79493.51, 'Checking');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (8, 40761.66, 'Savings');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (9, 2111.67, 'Checking');
-INSERT INTO account_details (account_id, balance, account_type) VALUES (10, 59173.15, 'Savings');
-
--->
--- a/v20.2/delete.md
+++ b/v20.2/delete.md
@@ -33,7 +33,7 @@ table td:first-child {
  `common_table_expr` | See [Common Table Expressions](common-table-expressions.html).
  `table_name` | The name of the table that contains the rows you want to update.
  `AS table_alias_name` | An alias for the table name. When an alias is provided, it completely hides the actual table name.
-`WHERE a_expr`| `a_expr` must be an expression that returns Boolean values using columns (e.g., `<column> = <value>`). Delete rows that return   `TRUE`.<br><br/>__Without a `WHERE` clause in your statement, `DELETE` removes all rows from the table.__
+`WHERE a_expr`| `a_expr` must be an expression that returns Boolean values using columns (e.g., `<column> = <value>`). Delete rows that return   `TRUE`.<br><br/>__Without a `WHERE` clause in your statement, `DELETE` removes all rows from the table. To delete all rows in a table, we recommend using [`TRUNCATE`](truncate.html) instead of `DELETE`.__
  `sort_clause` | An `ORDER BY` clause. <br /><br />See [Ordering of rows in DML statements](query-order.html#ordering-rows-in-dml-statements) for more details.
  `limit_clause` | A `LIMIT` clause. See [Limiting Query Results](limit-offset.html) for more details.
  `RETURNING target_list` | Return values based on rows deleted, where `target_list` can be specific column names from the table, `*` for all columns, or computations using [scalar expressions](scalar-expressions.html). <br><br>To return nothing in the response, not even the number of rows updated, use `RETURNING NOTHING`.
@@ -76,12 +76,6 @@ For more information about ordering query results in general, see
 [Ordering Query Results](query-order.html) and [Ordering of rows in
 DML statements](query-order.html#ordering-rows-in-dml-statements).
 
-## Delete performance on large data sets
-
-If you are deleting a large amount of data using iterative `DELETE ... LIMIT` statements, you are likely to see a drop in performance for each subsequent `DELETE` statement.
-
-For an explanation of why this happens, and for instructions showing how to iteratively delete rows in constant time, see [Why are my deletes getting slower over time?](sql-faqs.html#why-are-my-deletes-getting-slower-over-time).
-
 ## Force index selection for deletes
 
 By using the explicit index annotation (also known as "index hinting"), you can override [CockroachDB's index selection](https://www.cockroachlabs.com/blog/index-selection-cockroachdb-2/) and use a specific [index](indexes.html) for deleting rows of a named table.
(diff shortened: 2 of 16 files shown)


v19.2/delete.md
```diff
@@ -109,52 +131,63 @@ To view how the index hint modifies the query plan that CockroachDB follows for
 
 For examples, see [Delete with index hints](#delete-with-index-hints).
 
-## Examples
+## Delete "expired" data
```
BramGruneir, line 134: This is going to perform a scan with the delete. This can be devastating to performance.  I would suggest an alternate 2 step approach.

First, issue a 

```sql
SELECT primaryKey 
  FROM tableName 
 WHERE timestamp < current_date() - INTERVAL '1 MONTH'
 LIMIT 10000 
 AS OF SYSTEM TIME -5s; 
```
(or follower reads if possible)

Then take the results and issue a delete

```sql
DELETE FROM tableName WHERE primaryKey IN [list of keys];
```

This minimizes contention from the scan and maximizes the speed of the deletes.

v20.2/delete.md
```diff
@@ -109,6 +109,60 @@ To view how the index hint modifies the query plan that CockroachDB follows for
...
+import psycopg2
+import sys
+import os
+
+conn = psycopg2.connect(os.environ.get('DB_URI'))
+batchsize = 10000
+
+while True:
+    with conn.cursor() as cur:
+        cur.execute("DELETE from bank WHERE crdb_internal_mvcc_timestamp < current_date() - INTERVAL '1 MONTH' LIMIT {0}".format(batchsize))
```
BramGruneir, line 133: again, this is not a good pattern, it will greatly interfere with normal logic.

rohany:
> I updated this guidance to be more of a simple note, with the recommendation to use an AOST query (per Bram's recommendation) when using crdb_internal_mvcc_timestamp.

looks good to me

ericharmeling:
@BramGruneir @bdarnell 

Friendly reminder on re-review :)

bdarnell (commented):
> Reviewed 2 of 7 files at r1, 4 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, @ericharmeling, and @rohany)

___
*[_config_cockroachdb_local.yml, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGi5OoA1FIpHG0gSADz:-MGi5OoA1FIpHG0gSAE-:b-bshl2n) ([raw file](https://github.com/cockroachdb/docs/blob/3e92be91e3c1b0def349ac2d7253289d62a05aee/_config_cockroachdb_local.yml#L7)):*
> ```YAML
> - "v2.1"
> - "v19.1"
> - "v19.2"
> ```

Is this change supposed to be included in this commit?
___
*[v19.2/delete.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/7999#-MF1wURb2S-z6ZWztGlm:-MGi7iLa0OL9sZ0lUbmU:b4txkb1) ([raw file](https://github.com/cockroachdb/docs/blob/cd29e0c9fcc1ffbac2687156673a464818970091/v19.2/delete.md#L17)):*
>> How many rows are needed for this to make sense?
> 
> I'm not sure, and I suspect that it depends on the schema. Should we make a general recommendation re: row count?

Yeah, my question was really "can we provide a general recommendation here?" We do batches of 5-10k below, so maybe say something like "To delete a large number of rows (tens of thousands or more)..."

> We don't really document kv.transaction.max_intents_bytes outside of the cluster settings reference. And where we do document the setting in use, "we strongly recommend against it in most cases." 

This guidance is evolving. See https://github.com/cockroachdb/cockroach/issues/54029

> I guess we could add this as a special case if this is something we want to recommend. A related and important question: how do bytes correspond to rows (i.e., how would someone determine the number of rows to delete from the max_intent_bytes size)? I imagine it is not the same for all rows/tables?

For inserts and deletes, it's roughly the encoded size of the primary key columns times the number of indexes (primary+secondary), plus the encoded size of the key columns in all secondary indexes (plus something proportional to value sizes for inverted indexes). Updates are sometimes the same size and sometimes smaller, depending on index configuration and which fields get updated. A few data types (such as collated strings) get more complicated here. 

> I suppose you could loop over shards with DELETE WHERE crdb_internal_ts_shard_n = {0}.format(i) from 1 to n, where i is the shard number and n is the number of shards. Is this what you are suggesting?

Something like that, but the question is whether the optimizer would be able to do the right thing there. 
___
*[v19.2/delete.md, line 115 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiHijJ6BBbkZ-Izy3g:-MGiHijJ6BBbkZ-Izy3h:b-p72rwy) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L115)):*
> ```Markdown
> To delete a large number of rows, we recommend looping through subsets of the rows that you want to delete until all of the unwanted rows have been deleted. You can write a script to do this, or you can write the loop into your application.
> 
> ### Batch-delete on an indexed column
> ```

I'd insert a paragraph above this about the tradeoffs of indexed vs non-indexed columns. (i.e. that indexing the filtering column will make the deletion much faster, but it may cause insertion bottlenecks, especially if it's a timestamp)
___
*[v19.2/delete.md, line 117 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiGC3A3_hIFhb2CbyF:-MGiGC3A3_hIFhb2CbyG:b3w42ij) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L117)):*
> ```Markdown
> ### Batch-delete on an indexed column
> 
> Each iteration of the batch loop should execute a single `DELETE` query. When writing the query:
> ```

"should execute a transaction containing a single `DELETE` statement"

Emphasize that each loop is a transaction; putting a lot of deletes in the same transaction is not good. 
___
*[v19.2/delete.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiIDO-BC30qO-WlYf7:-MGiIDO-BC30qO-WlYf8:b-itruhk) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L124)):*
> ```Markdown
> - Add a `RETURNING` clause to the end of the query that returns the filtering column values of the deleted rows. Then, using the values of the deleted rows, update the filter to match only the subset of remaining rows to delete. This narrows the scan to the fewest rows possible, and [preserves the performance of the deletes over time](#preserving-delete-performance-over-time).
> 
> For example, suppose you want to delete all rows in the [`tpcc`](cockroach-workload.html#tpcc-workload) `order_line` table where `ol_amount` is less than `5000`, in batches of 5,000 rows.
> ```

The RETURNING pattern assumes that new deletable rows are no longer being generated, which is generally true if you're using timestamps. But if it's not a timestamp, it's not clear that that's true. There's no guarantee that this process will ever complete if new small orders are continually inserted. 

This also feels like an unnatural use case - `ol_amount` is not normally indexed in TPCC, and it doesn't really make sense to delete orders based on their size. Can we pick an existing indexed column instead? 
___
*[v19.2/delete.md, line 145 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiJmf5BvrCxb0tN1cN:-MGiJmf5BvrCxb0tN1cO:b-lw0ubf) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L145)):*
> ```Markdown
> 
> conn = psycopg2.connect(os.environ.get('DB_URI'))
> filter = "< 5000"
> ```

Swapping `<` for `<=` makes this more complicated (using `str.format` for SQL queries is bad because of the possibility of sql injection; it should be avoided if at all possible); I'd make it `<=` always (you could either make it `<= 4999` or just change the problem definition to keep it `<= 5000`). 
___
*[v19.2/delete.md, line 155 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiK90W2q2rqq7IZndM:-MGiK90W2q2rqq7IZndN:b3g69uj) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L155)):*
> ```Markdown
>         cur.execute(query)
>         if cur.rowcount == 0:
>             sys.exit()
> ```

Calling `sys.exit` here is unusual and prevents the `conn.close()` at the end. This should be `break` instead. 
___
*[v19.2/delete.md, line 164 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiKYxR80fIUVIat2Wh:-MGiKYxR80fIUVIat2Wi:bwyib87) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L164)):*
> ```Markdown
> The script deletes 5,000 rows, working iteratively in batches of 5,000 until all of the rows where `ol_amount < 5000` are deleted. Note that at each iteration, the filter is updated to match a narrower subset of rows.
> 
> ### Batch-delete using an `AS OF SYSTEM TIME` query
> ```

I'd make this heading "batch delete on a non-indexed column"
___
*[v19.2/delete.md, line 171 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiLYEy5kxE0XwQ6PSl:-MGiLYEy5kxE0XwQ6PSm:bi3qmyh) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L171)):*
> ```Markdown
>     - Use a `WHERE` clause that filters on the column identifying the rows.
>     - Add an [`AS OF SYSTEM TIME` clause](as-of-system-time.html) to the end of the selection subquery, or run the selection query in a separate, read-only transaction with [`SET TRANSACTION AS OF SYSTEM TIME`](as-of-system-time.html#using-as-of-system-time-in-transactions). This can help reduce [transaction contention](transactions.html#transaction-contention).
>     - Use a [`LIMIT`](limit-offset.html) clause to limit the number of rows queried to the batch size. To determine the optimal batch size, try out different batch sizes (1,000 rows, 10,000 rows, 100,000 rows, etc.), and monitor the change in performance.
> ```

It's probably best for the `SELECT` statement to have a substantially higher limit than the `DELETE` (so you have a nested loop to do multiple deletes per select). 
___
*[v19.2/delete.md, line 190 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiL0I17I8UX8Plg__W:-MGiL0I17I8UX8Plg__X:biiqdrv) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L190)):*
> ```Markdown
> 
> conn = psycopg2.connect(os.environ.get('DB_URI'))
> filter = "< 5000"
> ```

This does not appear to be used. 
___
*[v19.2/delete.md, line 198 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiLEmR5KALbXiQ0pgd:-MGiLEmR5KALbXiQ0pge:bbjf1hq) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L198)):*
> ```Markdown
>         if cur.rowcount == 0:
>             sys.exit()
>         pkvals = tuple(eval(row[0][1:-1]) for row in cur)
> ```

`eval` is very dangerous and should not be used here (and it's probably not going to be correct, depending on what exactly is in the table). What exactly are you trying to do here?

I think you probably want to remove the parens in the `SELECT` clause and then this is something like `pkvals = [tuple(row) for row in cur]`
___
*[v19.2/delete.md, line 200 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiMhq08j6jMII-2B5w:-MGiMhq08j6jMII-2B5x:b-34zot7) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L200)):*
> ```Markdown
>         pkvals = tuple(eval(row[0][1:-1]) for row in cur)
>         conn.commit()
>         cur.execute("DELETE FROM order_line WHERE (ol_w_id, ol_d_id, ol_o_id, ol_number) IN %s", [pkvals])
> ```

Don't you need a `conn.commit()` after this? (and then I don't think you need the one at the end of the loop). 
___
*[v19.2/delete.md, line 217 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiQXZdD78bqa0imBO3:-MGiQXZdD78bqa0imBO4:b-7gutt) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L217)):*
> ```Markdown
> For example, suppose that every morning you want to delete all rows in the [`tpcc`](cockroach-workload.html#tpcc-workload) `history` table that are older than a month.
> 
> First, you need to [index the timestamped column](indexes.html) on which you are filtering (i.e., `h_date`):
> ```

This is optional, but requires different implementations as discussed above (and mentioned below). Mention those tradeoffs here instead of just "you need to index the column". 
___
*[v19.2/delete.md, line 288 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiRKfk9_jbBu1-BlG_:-MGiRKfk9_jbBu1-BlGa:b-t8sef7) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L288)):*
> ```Markdown
> - At each iteration, update the `WHERE` clause to filter only the rows that have not yet been marked for deletion. For an example, see [Batch-delete on an indexed filter](#batch-delete-on-an-indexed-column) above.
> - At each iteration, first use a `SELECT` statement to return primary key values on rows that are not yet deleted. Rows marked for deletion will not be returned. Then, use a `DELETE`, filtering on the primary key values. For an example, see [Batch-delete using an `AS OF SYSTEM TIME` query](#batch-delete-using-an-as-of-system-time-query) above.
> - To iteratively delete rows in constant time, using a simple `DELETE` loop, you can [alter your zone configuration](configure-replication-zones.html#overview) and change `gc.ttlseconds` to a low value like 5 minutes (i.e., `300`), and then run your `DELETE` statement once per GC interval. If you take this approach, we strongly recommend returning `gc.ttlseconds` to the default value after the deletion is complete.
> ```

I would not recommend changing the GC TTL back and forth. Pick a value that fits your needs and then stop touching it.

ericharmeling (commented):
Thanks for iterating on this Ben!

I believe that I've addressed all of your comments in the latest commit. Changes will only be made to v19.2 files until everything is finalized. 

One thing to note is that I've consolidated the two bottom script examples. There are now only two: Indexed filtering column and non-indexed filtering column. The TTL workaround section is just a simple set of cron commands that reuses the second script.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999#-:-MHCtOti59Fg8xcK-7zy:bqlmabp)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, @ericharmeling, and @rohany)

___
*[_config_cockroachdb_local.yml, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGi5OoA1FIpHG0gSADz:-MGih_rTDiMNJwkYGS57:b-pyrbqy) ([raw file](https://github.com/cockroachdb/docs/blob/3e92be91e3c1b0def349ac2d7253289d62a05aee/_config_cockroachdb_local.yml#L7)):*

Is this change supposed to be included in this commit?
</blockquote></details>

No. Removed.
___
*[v19.2/delete.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/7999#-MF1wURb2S-z6ZWztGlm:-MHBrfgE-AC8n2LYTS_m:b26l4n) ([raw file](https://github.com/cockroachdb/docs/blob/cd29e0c9fcc1ffbac2687156673a464818970091/v19.2/delete.md#L17)):*
> ... We do batches of 5-10k below, so maybe say something like "To delete a large number of rows (tens of thousands or more)..."

Added a parenthetical clause "(i.e., tens of thousands of rows or more)".

> This guidance is evolving. See cockroachdb/cockroach#54029....

As this guidance is evolving, I say we open a separate docs issue about updating this setting. Thoughts?

> Something like that, but the question is whether the optimizer would be able to do the right thing there.

I think this requires further investigation, and could be documented in a separate issue. Thoughts?
___
*[v19.2/delete.md, line 115 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiHijJ6BBbkZ-Izy3g:-MHCK0ZC4ig16fd9HLc3:b-t9k02c) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L115)):*
Done.

Also, I added a recommendation to follow the non-indexed filtering column pattern if the filtering column is a timestamp, because
> it may cause insertion bottlenecks, especially if it's a timestamp`. 
___
*[v19.2/delete.md, line 117 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiGC3A3_hIFhb2CbyF:-MHCK1Z203qJ83S-T4qA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L117)):*

"should execute a transaction containing a single `DELETE` statement"

Emphasize that each loop is a transaction; putting a lot of deletes in the same transaction is not good. 
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiIDO-BC30qO-WlYf7:-MHCK8i-A5_n-5kNCIGM:b-207972) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L124)):*
> The RETURNING pattern assumes that new deletable rows are no longer being generated, which is generally true if you're using timestamps. But if it's not a timestamp, it's not clear that that's true. I can update this bullet point to include a sentence like "This assumes that no new rows are generated that match the `DELETE` filter." 

I updated this bullet point to include a sentence about this assumption. Because using an indexed timestamp column as the filtering column can create bottlenecks (as you noted in the comment above), we probably shouldn't call that out in the context of this example.  The other option is to just revise the recommendation to not use the `RETURNING` pattern at all. Thoughts? 

> This also feels like an unnatural use case - ol_amount is not normally indexed in TPCC, and it doesn't really make sense to delete orders based on their size. Can we pick an existing indexed column instead?

Sure. I changed this to use `new_order` with `no_w_id` as the filtering column instead, noting that it is the first column in the primary key of the table.
___
*[v19.2/delete.md, line 145 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiJmf5BvrCxb0tN1cN:-MHByFfLDOt0q0gr0Y17:b-16i3t3) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L145)):*
> I'd make it <= always (you could either make it <= 4999 or just change the problem definition to keep it <= 5000).

Done.

> (using str.format for SQL queries is bad because of the possibility of sql injection; it should be avoided if at all possible)

I updated the string formatting used here to be SQL-injection safe, following the recommendations from the [psycopg docs](https://www.psycopg.org/docs/usage.html).
___
*[v19.2/delete.md, line 155 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiK90W2q2rqq7IZndM:-MHBxoMmBNp9Kr6ZJ_8Y:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L155)):*

Calling `sys.exit` here is unusual and prevents the `conn.close()` at the end. This should be `break` instead. 
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 164 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiKYxR80fIUVIat2Wh:-MHBxvmZ6d4Aoz5yGTGo:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L164)):*

I'd make this heading "batch delete on a non-indexed column"
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 171 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiLYEy5kxE0XwQ6PSl:-MHDQc3m99taWLZ38_O5:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L171)):*

It's probably best for the `SELECT` statement to have a substantially higher limit than the `DELETE` (so you have a nested loop to do multiple deletes per select). 
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 190 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiL0I17I8UX8Plg__W:-MHByw9cCZpY_sO0mk2m:bss4a2b) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L190)):*

This does not appear to be used. 
</blockquote></details>

Removed.
___
*[v19.2/delete.md, line 198 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiLEmR5KALbXiQ0pgd:-MHDFNn_C1Z1so8q1RUv:bfmterj) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L198)):*

`eval` is very dangerous and should not be used here (and it's probably not going to be correct, depending on what exactly is in the table). What exactly are you trying to do here?

I think you probably want to remove the parens in the `SELECT` clause and then this is something like `pkvals = [tuple(row) for row in cur]`
</blockquote></details>

I was trying to pass the tuples as a string, but was doing so incorrectly. I simplified this portion with `pkvals = tuple(cur)`, and passed `pkvals` to `cur.execute()` as a tuple parameter, as shown in the [psycopg2 docs](https://www.psycopg.org/docs/usage.html#tuples-adaptation).
___
*[v19.2/delete.md, line 200 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiMhq08j6jMII-2B5w:-MHCL-LX-d073ZMLhe8D:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L200)):*

Don't you need a `conn.commit()` after this? (and then I don't think you need the one at the end of the loop). 
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 288 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiRKfk9_jbBu1-BlG_:-MHBzb0mBNOoedzmkLXR:b-53h0ej) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L288)):*

I would not recommend changing the GC TTL back and forth. Pick a value that fits your needs and then stop touching it. 
</blockquote></details>

Removed that last sentence.

bdarnell (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, @ericharmeling, and @rohany)

___
*[_config_cockroachdb_local.yml, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGi5OoA1FIpHG0gSADz:-MHMZGBCED1I6AIYDs0a:b-nyz3iu) ([raw file](https://github.com/cockroachdb/docs/blob/3e92be91e3c1b0def349ac2d7253289d62a05aee/_config_cockroachdb_local.yml#L7)):*

No. Removed.
</blockquote></details>

Looks like it's still there, at least in reviewable. 
___
*[v19.2/delete.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/7999#-MF1wURb2S-z6ZWztGlm:-MHMZOWVBSK6yp8MAtr7:b-nc983v) ([raw file](https://github.com/cockroachdb/docs/blob/cd29e0c9fcc1ffbac2687156673a464818970091/v19.2/delete.md#L17)):*
> As this guidance [on `kv.transaction.max_intents_bytes`] is evolving, I say we open a separate docs issue about updating this setting. Thoughts?

Yes, we just need to make sure that we update this page when we either increase the default or advise people to increase the setting on their own. (Specifically, the update will be that the "tens of thousands" guideline will increase when `kv.transaction.max_intents_bytes` is increased, so more users will be able to just do the straightforward thing instead of going through a more complex pagination process).

> I think this [hash-sharded indexes] requires further investigation, and could be documented in a separate issue. Thoughts?

This was much more speculative on my part; I think I'd just drop it instead of making a separate issue for it. 
___
*[v19.2/delete.md, line 115 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiHijJ6BBbkZ-Izy3g:-MHMbwn83Bk8mKsXvzIQ:b-5app7h) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L115)):*

Done.

Also, I added a recommendation to follow the non-indexed filtering column pattern if the filtering column is a timestamp, because
> it may cause insertion bottlenecks, especially if it's a timestamp`. 
</blockquote></details>

Good, but I think that it's also common for non-timestamp columns that are used for GC to create insertion bottlenecks, and on the other hand sometimes you need an index to make your queries fast even if it creates a bottleneck (and hash-sharding can help here). I'd say something like this 

"Having an indexed filtering column can make delete operations faster, but it might lead to bottlenecks in execution, especially if the filtering column is a [timestamp](timestamp.html) (these bottlenecks can be reduced but not eliminated if a hash-sharded index[link] is used). If the filtering column is not already indexed for other reasons, it may not be a good idea to add an index just to speed up batch deletes. Consider [batch deleting on non-indexed columns](#batch-delete-on-a-non-indexed-column) instead."

Or maybe some guidance on the tradeoffs belongs here (after the sentence starting "in the sections below") instead of in either section. "Filtering on an indexed column is both simpler to implement and more efficient, but adding an index can slow down insertions to the table and may cause bottlenecks. Filtering on a non-indexed column must perform a full-table scan (possibly more than one), so it takes time proportional to the size of the entire table."
___
*[v19.2/delete.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiIDO-BC30qO-WlYf7:-MHMmKGA2Hi43WNwQ20H:b-xisb3m) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L124)):*

> The RETURNING pattern assumes that new deletable rows are no longer being generated, which is generally true if you're using timestamps. But if it's not a timestamp, it's not clear that that's true. I can update this bullet point to include a sentence like "This assumes that no new rows are generated that match the `DELETE` filter." 

I updated this bullet point to include a sentence about this assumption. Because using an indexed timestamp column as the filtering column can create bottlenecks (as you noted in the comment above), we probably shouldn't call that out in the context of this example.  The other option is to just revise the recommendation to not use the `RETURNING` pattern at all. Thoughts? 

> This also feels like an unnatural use case - ol_amount is not normally indexed in TPCC, and it doesn't really make sense to delete orders based on their size. Can we pick an existing indexed column instead?

Sure. I changed this to use `new_order` with `no_w_id` as the filtering column instead, noting that it is the first column in the primary key of the table.
</blockquote></details>

OK, deleting `new_order` by `no_w_id` is a little more plausible. But I still feel like the most likely case for bulk deletion by an indexed column is going to be for timestamps (when the index is worthwhile because you use it in other queries), so I wouldn't shy away from that as an example.
___
*[v19.2/delete.md, line 145 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHN5wsh4cCdA9lef7mW:-MHN5wsh4cCdA9lef7mX:b-w4wlzb) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L145)):*
> ```Markdown
> 
> conn = psycopg2.connect(os.environ.get('DB_URI'))
> filter = "4"
> ```

You should be able to make this an integer instead of a string - remove the quotes here, make the placeholder `%d`, and call `execute(query, lastrow[0])` instead of formatting it into a string. 
___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHNArP_Cbrq3oFTxeZD:b-8vfcwr) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*
> ```Markdown
>         conn.commit()
>         while True:
>             cur.execute("DELETE FROM history WHERE h_w_id, rowid IN %s LIMIT 5000", (pkvals, ))
> ```

If you're using two different limits, you need two different loops: first you run the select and grab 20000 rows at once, then you loop over those 20000 rows in groups of 5000 and delete in those batches (so you don't actually use a LIMIT clause on the delete, you let the limit be implied by the size of the partial pkvals batch).

ericharmeling (commented):
@bdarnell 
Updated with your latest feedback. I had one question re: rewriting the nested loop in the non-indexed example - I'll wait for your advice on that before passing this off to writer review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999#-:-MHlalMq6-rft16v8BQb:btfx0pf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, and @rohany)

___
*[_config_cockroachdb_local.yml, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGi5OoA1FIpHG0gSADz:-MHSIj4z6juCDe43LYs1:b-657sqh) ([raw file](https://github.com/cockroachdb/docs/blob/3e92be91e3c1b0def349ac2d7253289d62a05aee/_config_cockroachdb_local.yml#L7)):*

Looks like it's still there, at least in reviewable. 
</blockquote></details>

Just checked this file out of master. Should be the same. :)
___
*[v19.2/delete.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/7999#-MF1wURb2S-z6ZWztGlm:-MHSJVXyEFRvADQBLHs8:bmjs13z) ([raw file](https://github.com/cockroachdb/docs/blob/cd29e0c9fcc1ffbac2687156673a464818970091/v19.2/delete.md#L17)):*
> Yes, we just need to make sure that we update this page when we either increase the default or advise people to increase the setting on their own. (Specifically, the update will be that the "tens of thousands" guideline will increase when kv.transaction.max_intents_bytes is increased, so more users will be able to just do the straightforward thing instead of going through a more complex pagination process).

Okay. When the `max_intents_bytes` default is updated, I will update this page. I created an issue so I don't forget. https://github.com/cockroachdb/docs/issues/8422
___
*[v19.2/delete.md, line 115 at r3](https://reviewable.io/reviews/cockroachdb/docs/7999#-MGiHijJ6BBbkZ-Izy3g:-MHSVGCuE7wOWOIzHOuy:be7jvbr) ([raw file](https://github.com/cockroachdb/docs/blob/4ea70b2af9d4da01a308bfcca0a1062892bd63e0/v19.2/delete.md#L115)):*
> "Having an indexed filtering column can make delete operations faster, but it might lead to bottlenecks in execution, especially if the filtering column is a timestamp (these bottlenecks can be reduced but not eliminated if a hash-sharded index[link] is used). If the filtering column is not already indexed for other reasons, it may not be a good idea to add an index just to speed up batch deletes. Consider batch deleting on non-indexed columns instead."

I weaved this advice into the note and the `DELETE` query guidance below.

>"Filtering on an indexed column is both simpler to implement and more efficient, but adding an index can slow down insertions to the table and may cause bottlenecks. Filtering on a non-indexed column must perform a full-table scan (possibly more than one), so it takes time proportional to the size of the entire table"

Super helpful. Thanks! Added guidance right below that second paragraph. I think it will be helpful to have the information at the top level of the batch deletes section, and in the subsection below.
___
*[v19.2/delete.md, line 145 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHN5wsh4cCdA9lef7mW:-MHSbJQR5vDu5M2ghTLd:b-3e64v2) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L145)):*
According to the [psycopg2 docs](https://www.psycopg.org/docs/usage.html):

> The variables placeholder must always be a %s, even if a different placeholder (such as a %d for integers or %f for floats) may look more appropriate.

So the placeholder needs to be %s, in all cases. 

Also, I updated `filter` to be an integer. However, passing an integer to `execute` will not work. On the same page of the psycopg2 docs:

> For positional variables binding, the second argument must always be a sequence, even if it contains a single variable (remember that Python requires a comma to create a single element tuple).

So I've used a single-value tuple here.
___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHWBU-jD_yVsmp1gwR9:b-30zgd1) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*
I'm a little confused about what should change here.

> If you're using two different limits, you need two different loops

There are already two different `while` loops: one top-level that selects 20000 rows at each iteration, and one (nested) that iteratively deletes those 20000 rows in batches of 5000.

> then you loop over those 20000 rows in groups of 5000 and delete in those batches (so you don't actually use a LIMIT clause on the delete, you let the limit be implied by the size of the partial pkvals batch)

This looks like a different pattern where there are not two different limits, which makes the loop a little more complex. I can update the second loop so that it does *not* use a `LIMIT`, but I'm not confident it is the best/most elegant advice:

```
def batch_delete(conn):
    while True:
        with conn.cursor() as cur:
            cur.execute("SET TRANSACTION AS OF SYSTEM TIME '-5s'")
            cur.execute("SELECT h_w_id, rowid FROM history WHERE h_date < current_date() - INTERVAL '1 MONTH' ORDER BY h_w_id, rowid LIMIT 17000")
            rowcount = cur.rowcount
            if rowcount == 0:
                break
            pkvals = tuple(cur)
            batches = int(rowcount/5000) + (rowcount%5000 > 0)
            conn.commit()
            i = 0
            j = 5000
            for n in range(batches):
                print(n)
                cur.execute("DELETE FROM history WHERE (h_w_id, rowid) IN %s", (pkvals[i:j], ))
                i = j
                j += 5000
                conn.commit()
            print("waiting")
            del pkvals
            time.sleep(5)
```

Please advise.

bdarnell (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, @ericharmeling, and @rohany)

___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHlhIKo51is7ihG8xBr:b1dls9w) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*
> There are already two different while loops: one top-level that selects 20000 rows at each iteration, and one (nested) that iteratively deletes those 20000 rows in batches of 5000.

Ah, right, I missed the inner loop because I was expecting a different implementation. The version currently in the branch is inefficient because it sends the entire 20000 pkvals to the server each time, and the server is going to re-process a lot of those before it gets to 5000 and can quit. So we should recommend something like what you have in this comment. Here's how I'd write it (note the use of `with conn` instead of explicit `commit` so it gets rolled back if an exception is raised):

```
def batch_delete(conn):
    while True:
        with conn:
            with conn.cursor() as cur:
                cur.execute("SET TRANSACTION AS OF SYSTEM TIME '-5s'")
                cur.execute("SELECT h_w_id, rowid FROM history WHERE h_date < current_date() - INTERVAL '1 MONTH' ORDER BY h_w_id, rowid LIMIT 17000")
                pkvals = list(cur)
        if not pkvals:
            return
        while pkvals:
            batch = pkvals[:5000]
            pkvals = pkvals[5000:]
            with conn:
                with conn.cursor() as cur:
                    cur.execute("DELETE FROM history WHERE (h_w_id, rowid) IN %s", (batch,))
                    print(cur.statusmessage)
        del batch
        del pkvals
        time.sleep(5)

```

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, and @rohany)

___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHq-sjT5dJDTGJzea6y:b-brv2wx) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*

> There are already two different while loops: one top-level that selects 20000 rows at each iteration, and one (nested) that iteratively deletes those 20000 rows in batches of 5000.

Ah, right, I missed the inner loop because I was expecting a different implementation. The version currently in the branch is inefficient because it sends the entire 20000 pkvals to the server each time, and the server is going to re-process a lot of those before it gets to 5000 and can quit. So we should recommend something like what you have in this comment. Here's how I'd write it (note the use of `with conn` instead of explicit `commit` so it gets rolled back if an exception is raised):

```
def batch_delete(conn):
    while True:
        with conn:
            with conn.cursor() as cur:
                cur.execute("SET TRANSACTION AS OF SYSTEM TIME '-5s'")
                cur.execute("SELECT h_w_id, rowid FROM history WHERE h_date < current_date() - INTERVAL '1 MONTH' ORDER BY h_w_id, rowid LIMIT 17000")
                pkvals = list(cur)
        if not pkvals:
            return
        while pkvals:
            batch = pkvals[:5000]
            pkvals = pkvals[5000:]
            with conn:
                with conn.cursor() as cur:
                    cur.execute("DELETE FROM history WHERE (h_w_id, rowid) IN %s", (batch,))
                    print(cur.statusmessage)
        del batch
        del pkvals
        time.sleep(5)

```
</blockquote></details>

@bdarnell 
Thank you!

A couple things:

- `execute()` doesn't seem to like a list of tuples as its second argument, so I kept `pkvals` a tuple (of tuples), not a list. Because we aren't trying to change the tuple, but rather replacing it, this runs just fine and doesn't seem to be a problem.
- I think the explicit `commit`s are needed (at least in the top loop) because the `SELECT` query is executed in a read-only transaction (as all AOST queries are). If we use `with conn` for both the top loop and the bottom loop, then the following error is returned:
`psycopg2.errors.ReadOnlySqlTransaction: cannot execute DELETE in a read-only transaction`

So I added `with conn` to the top of both loops, as you suggested above, but kept the explicit `commit` after the `SELECT` query in the top loop, to ensure that the `DELETE`s are run in a separate, read-write transaction. I also updated the indexed filtering column sample to use `with conn`, with no explicit `commit`s. 

Do these changes look good to you?

bdarnell (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, and @ericharmeling)

___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHqLuNC0U8IjkQApBS7:b-dxp7x5) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*

@bdarnell 
Thank you!

A couple things:

- `execute()` doesn't seem to like a list of tuples as its second argument, so I kept `pkvals` a tuple (of tuples), not a list. Because we aren't trying to change the tuple, but rather replacing it, this runs just fine and doesn't seem to be a problem.
- I think the explicit `commit`s are needed (at least in the top loop) because the `SELECT` query is executed in a read-only transaction (as all AOST queries are). If we use `with conn` for both the top loop and the bottom loop, then the following error is returned:
`psycopg2.errors.ReadOnlySqlTransaction: cannot execute DELETE in a read-only transaction`

So I added `with conn` to the top of both loops, as you suggested above, but kept the explicit `commit` after the `SELECT` query in the top loop, to ensure that the `DELETE`s are run in a separate, read-write transaction. I also updated the indexed filtering column sample to use `with conn`, with no explicit `commit`s. 

Do these changes look good to you?
</blockquote></details>

* Using tuples for `execute`: Fine by me, although the docs still lead me to believe that lists should be accepted ([example](https://www.psycopg.org/docs/usage.html#lists-adaptation)). Semantically lists would be (theoretically) preferred, but it's not worth worrying about.
* Explicit commits: This is because the indentation has become a little off. The `while pkvals:` block should be at the same level as the first `with conn:`. Right now you have the second `with conn:` running inside the first transaction block. Fix the indentation and you should be able to remove the explicit commit.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell and @BramGruneir)

___
*[v19.2/delete.md, line 199 at r5](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHNArP_Cbrq3oFTxeZC:-MHqT59-3Wm0_PNCjbrF:b-1cnok3) ([raw file](https://github.com/cockroachdb/docs/blob/9df1327713bd335c91911cbbcf496e21d5ea86f8/v19.2/delete.md#L199)):*

>  the docs still lead me to believe that lists should be accepted

You're right. The filter was using the wrong operator. Psycopg turns lists into SQL `ARRAY`s, so you need to use `ANY` instead of `IN`. Updated this.

> This is because the indentation has become a little off. 

Ah! You're right. Thank you for catching this. Fixed.

ericharmeling (commented):
Thanks a million for iterating on this @bdarnell. Passing off to writer review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999#-:-MHqX-Q7CT-0BEV6JpPw:b-r9c3mw)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell and @BramGruneir)

lnhsingh (commented):
A couple of small nits but otherwise LGTM! Nice job, very well-written

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999#-:-MHrrVS9Cuh5UuOZ6iye:bu5q34m)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, @ericharmeling, and @lnhsingh)

___
*[v19.2/delete.md, line 42 at r6](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHrfxDZCqJTUh4qhU21:-MHrfxD_3GY_u1YK-ul5:blc5b4y) ([raw file](https://github.com/cockroachdb/docs/blob/e35754bb228f5e6d42f0eaa4bbcd7fa7cbb7d4cb/v19.2/delete.md#L42)):*
> ```Markdown
>  `table_name` | The name of the table that contains the rows you want to update.
>  `AS table_alias_name` | An alias for the table name. When an alias is provided, it completely hides the actual table name.
> `WHERE a_expr`| `a_expr` must be an expression that returns Boolean values using columns (e.g., `<column> = <value>`). Delete rows that return   `TRUE`.<br><br/>__Without a `WHERE` clause in your statement, `DELETE` removes all rows from the table. To delete all rows in a table, we recommend using `TRUNCATE` instead of `DELETE`__
> ```

nit: Add period at the end / link to `truncate.html`
___
*[v19.2/delete.md, line 127 at r6](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHrkpCf6eeqVcFe24Vf:-MHrkpCf6eeqVcFe24Vg:b-dbvh8a) ([raw file](https://github.com/cockroachdb/docs/blob/e35754bb228f5e6d42f0eaa4bbcd7fa7cbb7d4cb/v19.2/delete.md#L127)):*
> ```
> Note that if the filtering column is not already indexed, it might not be a good idea to add an index just to speed up batch deletes.
> ```

Suggestion: Reword this sentence to be more active / stronger - something like `Note that if the filtering column is not already indexed, it is not beneficial to add an index just to speed up batch deletes.`

ericharmeling (commented):
TFTR @lnhsingh!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/7999#-:-MHvIyjqA-cKcPyimCa1:byxolix)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @BramGruneir, and @lnhsingh)

___
*[v19.2/delete.md, line 42 at r6](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHrfxDZCqJTUh4qhU21:-MHvJr4wBjdjq4-p26tT:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e35754bb228f5e6d42f0eaa4bbcd7fa7cbb7d4cb/v19.2/delete.md#L42)):*

nit: Add period at the end / link to `truncate.html`
</blockquote></details>

Done.
___
*[v19.2/delete.md, line 127 at r6](https://reviewable.io/reviews/cockroachdb/docs/7999#-MHrkpCf6eeqVcFe24Vf:-MHvKBVSDF49Tp-zxbnF:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e35754bb228f5e6d42f0eaa4bbcd7fa7cbb7d4cb/v19.2/delete.md#L127)):*

> ```
> Note that if the filtering column is not already indexed, it might not be a good idea to add an index just to speed up batch deletes.
> ```

Suggestion: Reword this sentence to be more active / stronger - something like `Note that if the filtering column is not already indexed, it is not beneficial to add an index just to speed up batch deletes.`
</blockquote></details>

Done.

PR #8001, opened 2020-08-14
mock up API endpoint docs

taroface:
Example use of:

- Full "API Reference" of all endpoints
- Individual "Hot Ranges" endpoint doc, with made-up stability guarantee
- Their location in the docs sidebar

@mjibson please note: I changed the h3s to h4s in api-reference.md, because the h3s were appearing in the right nav. Also, the right nav is creating a double scrollbar because there are too many items.

There is a callout at the top of the API Reference which states that all APIs are experimental. This is following a comment Ben made about the APIs (in comparison to SQL being our primary intended interface).

cc @lunevalex @kenliu

### DIFF
--- a/v20.2/hot-ranges.md
+++ b/v20.2/hot-ranges.md
@@ -9,44 +9,44 @@ The Hot Ranges endpoint returns ranges with active requests for a specified node
 
 This API will remain stable across minor (patch) releases within a major release.
 
-## Resource
+### Resource
 
 `GET /_status/hotranges`
 
-## Authorization
+### Authorization
 
 - TK token is required.
 - A user must have the `admin` role.
 
-## Request Parameters
+### Request Parameters
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
 | node_id | [string](#cockroach.server.serverpb.HotRangesRequest-string) |  | If left empty, hot ranges for all nodes/stores will be returned. |
 
-### Example Request
+#### Example Request
 
 ```
 code block
 ```
 
-## Response Parameters
+### Response Parameters
 
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
 | node_id | [int32](#cockroach.server.serverpb.HotRangesResponse-int32) |  | NodeID is the node that submitted all the requests. |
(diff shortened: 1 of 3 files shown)


mjibson:
My understanding was that the auto-generated files would be copied verbatim to the docs project and then included where needed. If you want to make changes to them (h3 -> h4), you should first do it in the cockroach repo and then copy those files without changes here. This makes keeping them up-to-date easier.

I was also under the impression that for the single endpoint pages, the request/response thing would be included from the auto-generated doc, just like the SQL diagrams. This allows the docs team to write whatever prose they want and then have the request/response stuff always correctly generated below that.

mjibson:
See #8053 for what I mean here.

taroface:
> My understanding was that the auto-generated files would be copied verbatim to the docs project and then included where needed. If you want to make changes to them (h3 -> h4), you should first do it in the cockroach repo and then copy those files without changes here. This makes keeping them up-to-date easier.
> 
> I was also under the impression that for the single endpoint pages, the request/response thing would be included from the auto-generated doc, just like the SQL diagrams. This allows the docs team to write whatever prose they want and then have the request/response stuff always correctly generated below that.

I agree with all the above. This PR was intended as an exercise to understand what we'd want to tweak on the `cockroach` template, based on the results of this mockup. I should have been clearer in asking for those tweaks. Will comment in #8053.

v20.2/hot-ranges.md
```diff
@@ -0,0 +1,83 @@
...
+- TK token is required.
+- A user must have the `admin` role.
+
+### Request Parameters
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| node_id | [string](#cockroach.server.serverpb.HotRangesRequest-string) |  | If left empty, hot ranges for all nodes/stores will be returned. |
+
+#### Example Request
```
taroface, line 27: @mjibson Before I review the current docgen update, can you let me know if the include on these individual pages needs to be a single unbroken block? I have added an `Example Request` and `Example Response` next to their corresponding parameters, which is a pretty standard API docs structure and IMO the most user-friendly.

If it's not possible for me to somehow break up the include to insert these sections, do you think that auto-generating and including an example request & response could actually be part of the template?

mjibson: If that is what you want then we will have to generate separate files for request and response. That's fine.

taroface: Yes, that would be great. And maybe a separate one for the special parameter types as well?

taroface:
Closing as this has served its purpose. Will use #8053 as a guideline for creating these docs in the future.

PR #8013, opened 2020-08-17
STRINGs remove trailing space

### LINKED ISSUE #7859: sql: truncate trailing spaces for char column types
PR: https://github.com/cockroachdb/cockroach/pull/50492

From release notes:
> The `CHAR` column [`type`](../v20.2/data-types.html) will now truncate the trailing space characters in line with Postgres. Existing stored `CHAR` entries with spaces at the end of the `CHAR` column [`type`](../v20.2/data-types.html) will no longer return rows with trailing space characters. Use the `LIKE` query to find and modify these rows. [#50492][#50492] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #7859.

### DIFF
--- a/v20.2/string.md
+++ b/v20.2/string.md
@@ -33,6 +33,10 @@ When inserting a string:
 - If the value is cast as a string with a length limit (e.g., `CAST('hello world' AS STRING(5))`), CockroachDB truncates to the limit.
 - If the value is under the column's length limit, CockroachDB does **not** add padding. This applies to `STRING(n)` and all its aliases.
 
+{{site.data.alerts.callout_info}}
+<span class="version-tag">New in v20.1:</span> To improve compatibility with PostgreSQL, CockroachDB truncates trailing space characters in `STRING` variables.
+{{site.data.alerts.end}}
+
 ## Syntax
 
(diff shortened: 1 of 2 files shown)


v20.2/string.md
```diff
@@ -33,6 +33,10 @@ When inserting a string:
 - If the value is cast as a string with a length limit (e.g., `CAST('hello world' AS STRING(5))`), CockroachDB truncates to the limit.
 - If the value is under the column's length limit, CockroachDB does **not** add padding. This applies to `STRING(n)` and all its aliases.
 
+{{site.data.alerts.callout_info}}
+<span class="version-tag">New in v20.1:</span> To improve compatibility with PostgreSQL, CockroachDB truncates trailing space characters in `STRING` variables.
```
otan, line 37: should only be CHAR column types, right?

ericharmeling: Hmm... I'm a little confused then. I thought CHAR was just an alias for STRING? That's how we document CHAR, at least: https://www.cockroachlabs.com/docs/stable/string.html#aliases

If they are two different data types, then we really need to update the docs.

ericharmeling:
@otan I'm closing this PR in favor of #8250.

PR #8015, opened 2020-08-17
Simplest, most manual breadcrumbs

yzdocs:
This works by going into every page we want to list on the sidenav and manually specifying their parents, in top-down hierarchical order.

If you want to see this in preview, I enabled breadcrumbs only for v20.1 pages under "GUIDES > Get Started". (Ignore styling ugliness for now of course.) The nice thing about this solution is it supports arbitrary/infinite sidenav hierarchies.

The other really nice thing is this solution will support breadcrumbs in Algolia, i.e. we should be able to get breadcrumbs in search results as well.

### DIFF
--- a/_data/breadcrumbs.yml
+++ b/_data/breadcrumbs.yml
@@ -1,0 +1,11 @@
+- title: GUIDES
+  breadcrumb: guides
+
+- title: Get Started
+  breadcrumb: get-started
+
+- title: Develop
+  breadcrumb: develop
+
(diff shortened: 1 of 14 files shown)


yingzhuchin:
An FYI for @anagio and @jseldess

yingzhuchin:
Screen Shot 2020-08-17 at 6 12 42 PM

Screenshot for posterity (for when previews expire)

PR #8039, opened 2020-08-21
Add docs for CockroachDB Spatial install

### LINKED ISSUE #7916: Document how to install a CockroachDB with spatial capabilities
Right now, we are punting to Docker for the 20.2 alphas.

For the 20.2 GA release, we will need to get clear on exactly what
this new install process entails. Since we are losing the "download
one binary and go" experience, we must document the necessary steps
while trying to keep it as simple as possible.

We also need to be clear in the docs about what exactly CockroachDB
will do if you run a `cockroach` binary that cannot find the C
libraries necessary for geospatial support.  E.g. will everything run
"fine" as long as you don't try any spatial stuff?  Will it return an
error if you try to load spatial data, or call a spatial function? Are
there situations where it will crash? Will monkeys fly out of your
nose? Etc.

rmloveland:
Fixes #7916.

Summary of changes:

- Add a new 'Install CockroachDB Spatial' page that lives alongside the
  regular installation instructions, but separately.  The reasoning is
  that there are many more fiddly steps that non-spatial users should
  ideally not have to mess with until this is in a more polished state.

- Update the 'Install CockroachDB' pages for Linux and Mac to add
  pointers to the spatial page.

- Add the new spatial install page to the sidebar (it is not an
  additional entry in the sidebar; instead it is yet another page that
  "hides" under 'Install CockroachDB').

### DIFF
--- a/_includes/sidebar-data-v20.2.json
+++ b/_includes/sidebar-data-v20.2.json
@@ -12,7 +12,8 @@
               "/${VERSION}/install-cockroachdb.html",
               "/${VERSION}/install-cockroachdb-mac.html",
               "/${VERSION}/install-cockroachdb-linux.html",
-              "/${VERSION}/install-cockroachdb-windows.html"
+              "/${VERSION}/install-cockroachdb-windows.html",
+              "/${VERSION}/install-cockroachdb-spatial.html"
             ]
           },
           {
--- a/v20.2/install-cockroachdb-linux.html
+++ b/v20.2/install-cockroachdb-linux.html
@@ -14,6 +14,10 @@ key: install-cockroachdb.html
 
 <p>See <a href="../releases/{{page.release_info.version}}.html" class="mac-releasenotes-download" id="mac-releasenotes-download-{{page.version.version}}" data-eventcategory="mac-releasenotes-download">Release Notes</a> for what's new in the latest release, {{ page.release_info.version }}. To upgrade to this release from an older version, see <a href="upgrade-cockroach-version.html">Cluster Upgrade</a>.</p>
 
+{{site.data.alerts.callout_success}}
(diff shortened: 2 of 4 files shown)


rmloveland:
Oliver since alpha.3 is not out yet these URLs will change, but this page has the meat of the instructions

Also FWIW installing anything on Windows does not appear to work (https://github.com/cockroachdb/docs/issues/8038) so that's why I have not done any Windows instructions here.

otan:
Sorry for delaying the review (accident!) -- but coincidentally enough, the alpha 3 zip files are available and should contain all the lib files!

rmloveland:
ok - I'll remove you as a reviewer for now and update with the new files - thanks Oliver!

rmloveland:
Oliver this has been tested and updated to use the recent alpha3 binaries for Mac and Linux

otan (approved):
LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8039#-:-MFg_PKYCf1uuMXXcsIc:b-85i648)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan and @rmloveland)

___
*[v20.2/install-cockroachdb-spatial.md, line 252 at r1](https://reviewable.io/reviews/cockroachdb/docs/8039#-MFg_KzmBxO4P-R_80gG:-MFg_KzmBxO4P-R_80gH:b46crst) ([raw file](https://github.com/cockroachdb/docs/blob/99ec9ac105497db106bb6505378c559a6ef64410/v20.2/install-cockroachdb-spatial.md#L252)):*
> ```Markdown
> 
> ~~~
> ERROR: st_isvalid(): geos: error during GEOS init: geos: cannot load GEOS from dir "/usr/local/lib/cockroach": failed to execute dlopen
> ```

This will give a better reason soon!

Issue #7947, opened 2020-08-07
Create Your CockroachCloud Cluster Doc Update 


andreimatei:
Re: [Create Your CockroachCloud Cluster](https://cockroachlabs.com/docs//v20.1/cockroachcloud-create-your-cluster.html)

The pricing is wrong. It says "per month" but it's really per hour. Quite the difference... Maybe we want to say "per hour, billed monthly" (if that's true).

Amruta-Ranade:
Excellent catch, Andrei! Thank you!

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/8045

Issue #7470, opened 2020-06-04
Add in FAQs or "create cluster" pricing page that we do not charge for data transfer costs


kannanlakshmi:
We get this question a lot on CockroachCloud - do we charge separately for data costs. The answer is we don't but we should explicitly say that, maybe under "FAQs"? https://cockroachlabs.slack.com/archives/CBE9QQC4S/p1591278383299000

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/8045

PR #8045, opened 2020-08-23
CC pricing updates

### LINKED ISSUE #7712: Add clarity around AWS/GCP pricing discrepancy
In our docs, we have a nice matrix that outlines a price comparison between choosing AWS and GCP for your CC cluster. However, we don't provide any opinion/explanation on the price discrepancy between the two providers. This leaves the user wondering "is GCP worse/less secure than AWS and that's why it's cheaper?" or assuming "GCP is always better because it's less expensive" -- both of which can be untrue.

We should clarify that they are the same offering and that the pricing fluctuation is a reflection of the provider's pricing strategies and not the underlying reliability, hardware, security etc. Might even be worth pointing to the latest Cloud Report to help guide users to the right cloud for them.

Amruta-Ranade:
Closes #7712 #7470 #7947

### DIFF
--- a/_includes/v20.1/cockroachcloud-pricing.md
+++ b/_includes/v20.1/cockroachcloud-pricing.md
@@ -1,0 +1,8 @@
+The choice of the cloud provider decides the price per node. For pricing comparison, refer to the following table:
+
+Hardware configuration	| GCP Pricing (per node, per hour)	| AWS Pricing (per node, per hour)
+----------|------------|------------
+`Option 1` (2 vCPU, 60 GB disk) |	$0.11	| $0.48
+`Option 2` (4 vCPU, 150 GB disk) | $0.69	| $0.83
+
+CockroachCloud does not charge you for data transfer costs.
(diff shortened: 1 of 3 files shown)


jseldess:
@Amruta-Ranade, in our 1:1, we briefly discussed adding a distinct docs page on pricing. I think it's good to keep the details in the create a cluster how-to, but we could use an include to have it in multiple places.

emily-horing:
@kannanlakshmi I understand if we don't want to give a reason like "differences in hardware costs" if that is only part of the story, however I think that there is real confusion and misinterpretation to think that maybe AWS is "better" because it is more expensive.

I think we could just say something like: 'Note there is no difference in quality, security (etc), between AWS and GCP as far as we know. Check out the cloud report for more details on performance comparisons"...

kannanlakshmi:
Fair enough, why don't we simply say, for deeper dive into performance differences between Cloud, consult our Cloud report and leave pricing totally out of it.

PR #7964, opened 2020-08-11
Support for ENUM types

ericharmeling:
Fixes #7751. 
Fixes #7626. 
Fixes #8154.
Fixes #8136.

Issue #7830, opened 2020-07-30
backupccl: allow for remapping backed up types to existing types


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/50349

From release notes:
> [`RESTORE`](../v20.2/restore.html) now remaps user-defined types within the backup to existing types in the cluster. [#50349][#50349] {% comment %}doc{% endcomment %}

lnhsingh:
I don't think any docs updates are needed. Closing.

PR #8052, opened 2020-08-24
Add info about restoring tables with UDTs

### LINKED ISSUE #7962: backup/restore Tables and User Defined Types (20.2)
Here's the original issue https://github.com/cockroachdb/cockroach/issues/48689
and the main PRs https://github.com/cockroachdb/cockroach/pull/50064 https://github.com/cockroachdb/cockroach/issues/50324

The big difference here is that now restore takes into account the current state of your cluster, in a sense
Nothing really changes for fullcluster and database restores
but for individual table backup, we collect all of the user defined types that the target table references (in columns or default expressions etc)
so if your table was `create table t (x int, y myenum)`, we'll include `myenum` in your backup
when restoring the backup, check to see if a type named `myenum` exists already
if not, then we just restore `myenum` from the backup
if it does exist, then we attempt to point all references to `myenum` in your back to the existing type `myenum` in the cluster.

lnhsingh:
Closes #7962, #7830.

To Do:

- [ ] Once [Eric's ENUM Types PR](https://github.com/cockroachdb/docs/pull/7964) is merged, add links to `ENUM`, `DROP TYPE` (is there `RENAME TYPE`?) - cc @ericharmeling

### DIFF
--- a/v20.2/restore.md
+++ b/v20.2/restore.md
@@ -112,7 +112,7 @@ To restore individual tables, the tables can not already exist in the [target da
 - If there is _not_ an existing type in the cluster with the same name, CockroachDB will create the user-defined type as it exists in the backup.
 - If there is an existing type in the cluster with the same name that is compatible with the type in the backup, CockroachDB will map the type in the backup to the type in the cluster.
 
-    In general, two types are compatible if they are the same kind (e.g., an enum is only compatible with other enums). Additionally, enums are only compatible if they have the same set of elements that have also added in the same order.
+    In general, two types are compatible if they are the same kind (e.g., an enum is only compatible with other enums). Additionally, enums are only compatible if they have the same ordered set of elements that have also been batched in the same way.
 
 - If there is an existing type in the cluster with the same name but it is _not_ compatible with the type in the backup, the restore will not succeed and you will be asked to resolve the name conflict. You can do this by either dropping or renaming the existing user-defined type.
 
--- a/v20.2/restore.md
+++ b/v20.2/restore.md
@@ -14,7 +14,7 @@ Because CockroachDB is designed with high fault tolerance, restores are designed
 
 You can restore:
 
--  [A full cluster](#full-cluster)
+- [A full cluster](#full-cluster)
 - [Databases](#databases)
 - [Tables](#tables)
 
@@ -67,7 +67,7 @@ The URL for your backup's locations must use the following format:
 
 You can restore:
 
--  [A full cluster](#full-cluster)
+- [A full cluster](#full-cluster)
 - [Databases](#databases)
 - [Tables](#tables)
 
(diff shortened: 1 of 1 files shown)


v20.2/restore.md
```diff
@@ -98,14 +98,20 @@ The target database must have not have tables or views with the same name as the
...
 - [Restore the table or view into a different database](#into_db).
 
-{{site.data.alerts.callout_info}}
-`RESTORE` only offers table-level granularity; it _does not_ support restoring subsets of a table.
-{{site.data.alerts.end}}
+<span class="version-tag">New in v20.2:</span> When restoring an individual table that references a user-defined type (e.g., `ENUM`), CockroachDB will first check to see if the type already exists. The CockroachDB restore will attempt the following for each user-defined type within a table backup:
+
+- If there is _not_ an existing type in the cluster with the same name, CockroachDB will create the user-defined type as it exists in the backup.
+- If there is an existing type in the cluster with the same name that is compatible with the type in the backup, CockroachDB will map the type in the backup to the type in the cluster.
```
rohany, line 113: we should define what compatibility means here. I would like the explanation to not get too much into implementation details to confuse readers, but we'll see. 

In general, two types are compatible if they are the same kind (eg enums are only compatible with other enums). Additionally, enums are only compatible if they have the same ordered set of elements. However, these elements must have also been added in the same order (due to how enums are implemented). I.e.
* `CREATE TYPE t1 AS ENUM ('hello', 'lauren')` and `CREATE TYPE t2 AS ENUM ('hello', 'lauren')` will be compatible
* `CREATE TYPE t1 AS ENUM ('hello', 'lauren')` and `CREATE TYPE t2 AS ENUM ('lauren', 'hello')` won't be compatible
* `CREATE TYPE t1 AS ENUM ('hello', 'lauren')` and `CREATE TYPE t2 AS ENUM ('hello'); ALTER TYPE t2 ADD VALUE ('lauren')` won't necessarily be compatible

rohany:
> Per this, why is the third example not necessarily compatible? Could the ADD VALUE ('lauren') append the new element to the beginning of the set (which would then be incompatible with ('hello', 'lauren')?

Yeah, this is the tricky part. The internal representation of `'lauren'` is different when both enum members were created up front vs when only `'hello'` is created up front and `'lauren'` is added later. This representation is stored durably as part of each row, so they aren't compatible.

rohany:
> Also, is "batched" ok to use here?

No I don't think thats quite right. I think just "created in the same way" should be fine. I think the examples would help catch some confusion in advance though.

rohany:
Linking to this section in the RFC might be useful https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20200331_enums.md#physical-layout

lnhsingh:
Thanks! Added the link and examples. Can you take a look?

rohany:
> This makes sense, especially with the example, though I wonder if there is a more precise term than "kind", since an enum is a data type. Something like "two user-defined types are compatible if they have the same data type"? Possible I'm not understanding correctly, though.

`Kind` is a term used to describe the "type" of a type

PR #8223, opened 2020-08-31
IO docs reorg

lnhsingh:
Previously, backup / restore / import / export docs were bloated and we're adding new backup features in 20.2 so this re-org is to make content more readable and easier to find the sidenav. To do this, IO docs were moved to the Deploy section as follows:

- Back up and Restore Data
    - Full and Incremental Backups (Take Full and Incremental Backups)
    - Backups with Revision History and Point-in-time Restore (Take Backups with Revision History and Restore from a Point-in-time)
    - Encrypted Backup and Restore (Take and Restore Encrypted Backups)
    - Locality-aware Backup and Restore (Take and Restore Locality-aware Backups)
    - Scheduled Backups (Set up a Backup Schedule) *
- File Storage for Bulk Operations
    - Cloud Storage (Use Cloud Storage for Bulk Operations)
    - Nodelocal Storage (Use Nodelocal Storage for Bulk Operations) *
    - Userlocal Storage (Use Userlocal Storage for Bulk Operations) *
    - Local File Server (Use a Local File Server for Bulk Operations)
- Stream Data Out of CockroachDB (Stream Data Out of CockroachDB Using Changefeeds)

For each of the above, the doc is either new or renamed, and redirects have been added and cross-references have been updated as needed. 

**Note:** there is no new content (content has been repurposed or minor update to make sense with the new organization), to limit the scope of this PR. Edits / additions to content / re-restructuring will come in future PRs (e.g., re-structuring the Use Cloud Storage for Bulk Operations page).

_* This has not been written yet and will be added in a future PR_

Closes #8050.

PR #8234, opened 2020-09-03
Separate example file URLs on Cloud Storage doc

lnhsingh:


jseldess:
Nice, @lnhsingh. For changes like this, if it's based on feedback or discussion, it's best to create an issue and resolve it with the pr.

lnhsingh:
Will do! This was a minor edit that Michael brought up when
reviewing #8223, so I should've linked it there as follow up.

PR #8242, opened 2020-09-03
Partial indexes

### LINKED ISSUE #7747: Partial indexes
https://cockroachlabs.atlassian.net/browse/CRDB-558

ericharmeling:
Fixes #7747. 

- Added a new page for Partial Indexes.
- Updated `CREATE INDEX` syntax diagram and parameters. 

This PR will likely fix some future "opt" release note issues.

### DIFF
--- a/_includes/sidebar-data-v20.2.json
+++ b/_includes/sidebar-data-v20.2.json
@@ -1949,6 +1949,12 @@
     		      "/${VERSION}/inverted-indexes.html"
     		    ]
     		  },
+          {
+            "title": "Partial Indexes",
+            "urls": [
+              "/${VERSION}/partial-indexes.html"
+            ]
+          },
     		  {
     		    "title": "Column Families",
     		    "urls": [
--- a/_includes/v20.2/sql/diagrams/create_index.html
+++ b/_includes/v20.2/sql/diagrams/create_index.html
@@ -1,91 +1,91 @@
-<div><svg width="692" height="632">
-<polygon points="11 17 3 13 3 21"></polygon>
-<polygon points="19 17 11 13 11 21"></polygon>
-<rect x="33" y="3" width="70" height="32" rx="10"></rect>
-<rect x="31" y="1" width="70" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="41" y="21">CREATE</text>
-<rect x="143" y="35" width="74" height="32" rx="10"></rect>
-<rect x="141" y="33" width="74" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="151" y="53">UNIQUE</text>
-<rect x="257" y="3" width="62" height="32" rx="10"></rect>
-<rect x="255" y="1" width="62" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="265" y="21">INDEX</text>
-<rect x="359" y="35" width="130" height="32" rx="10"></rect>
-<rect x="357" y="33" width="130" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="367" y="53">CONCURRENTLY</text>
+<div><svg width="686" height="698">
+<polygon points="9 17 1 13 1 21"></polygon>
+<polygon points="17 17 9 13 9 21"></polygon>
+<rect x="31" y="3" width="70" height="32" rx="10"></rect>
+<rect x="29" y="1" width="70" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="39" y="21">CREATE</text>
+<rect x="141" y="35" width="74" height="32" rx="10"></rect>
+<rect x="139" y="33" width="74" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="149" y="53">UNIQUE</text>
+<rect x="255" y="3" width="62" height="32" rx="10"></rect>
+<rect x="253" y="1" width="62" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="263" y="21">INDEX</text>
+<rect x="357" y="35" width="130" height="32" rx="10"></rect>
+<rect x="355" y="33" width="130" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="365" y="53">CONCURRENTLY</text>
 <a xlink:href="sql-grammar.html#opt_index_name" xlink:title="opt_index_name">
-<rect x="109" y="101" width="126" height="32"></rect>
-<rect x="107" y="99" width="126" height="32" class="nonterminal"></rect>
-<text class="nonterminal" x="117" y="119">opt_index_name</text>
+<rect x="105" y="101" width="126" height="32"></rect>
+<rect x="103" y="99" width="126" height="32" class="nonterminal"></rect>
+<text class="nonterminal" x="113" y="119">opt_index_name</text>
 </a>
-<rect x="109" y="145" width="34" height="32" rx="10"></rect>
-<rect x="107" y="143" width="34" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="117" y="163">IF</text>
-<rect x="163" y="145" width="48" height="32" rx="10"></rect>
-<rect x="161" y="143" width="48" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="171" y="163">NOT</text>
-<rect x="231" y="145" width="68" height="32" rx="10"></rect>
-<rect x="229" y="143" width="68" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="239" y="163">EXISTS</text>
+<rect x="105" y="145" width="34" height="32" rx="10"></rect>
+<rect x="103" y="143" width="34" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="113" y="163">IF</text>
+<rect x="159" y="145" width="48" height="32" rx="10"></rect>
+<rect x="157" y="143" width="48" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="167" y="163">NOT</text>
+<rect x="227" y="145" width="68" height="32" rx="10"></rect>
+<rect x="225" y="143" width="68" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="235" y="163">EXISTS</text>
 <a xlink:href="sql-grammar.html#index_name" xlink:title="index_name">
-<rect x="319" y="145" width="96" height="32"></rect>
-<rect x="317" y="143" width="96" height="32" class="nonterminal"></rect>
-<text class="nonterminal" x="327" y="163">index_name</text>
+<rect x="315" y="145" width="96" height="32"></rect>
+<rect x="313" y="143" width="96" height="32" class="nonterminal"></rect>
+<text class="nonterminal" x="323" y="163">index_name</text>
 </a>
-<rect x="455" y="101" width="40" height="32" rx="10"></rect>
-<rect x="453" y="99" width="40" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="463" y="119">ON</text>
+<rect x="451" y="101" width="40" height="32" rx="10"></rect>
+<rect x="449" y="99" width="40" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="459" y="119">ON</text>
 <a xlink:href="sql-grammar.html#table_name" xlink:title="table_name">
-<rect x="515" y="101" width="94" height="32"></rect>
-<rect x="513" y="99" width="94" height="32" class="nonterminal"></rect>
-<text class="nonterminal" x="523" y="119">table_name</text>
+<rect x="511" y="101" width="94" height="32"></rect>
+<rect x="509" y="99" width="94" height="32" class="nonterminal"></rect>
+<text class="nonterminal" x="519" y="119">table_name</text>
 </a>
-<rect x="114" y="287" width="64" height="32" rx="10"></rect>
-<rect x="112" y="285" width="64" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="122" y="305">USING</text>
+<rect x="110" y="287" width="64" height="32" rx="10"></rect>
+<rect x="108" y="285" width="64" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="118" y="305">USING</text>
 <a xlink:href="sql-grammar.html#name" xlink:title="name">
-<rect x="198" y="287" width="54" height="32"></rect>
-<rect x="196" y="285" width="54" height="32" class="nonterminal"></rect>
-<text class="nonterminal" x="206" y="305">name</text>
+<rect x="194" y="287" width="54" height="32"></rect>
+<rect x="192" y="285" width="54" height="32" class="nonterminal"></rect>
+<text class="nonterminal" x="202" y="305">name</text>
 </a>
-<rect x="292" y="255" width="26" height="32" rx="10"></rect>
-<rect x="290" y="253" width="26" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="300" y="273">(</text>
+<rect x="288" y="255" width="26" height="32" rx="10"></rect>
+<rect x="286" y="253" width="26" height="32" class="terminal" rx="10"></rect>
+<text class="terminal" x="296" y="273">(</text>
 <a xlink:href="sql-grammar.html#a_expr" xlink:title="a_expr">
-<rect x="358" y="255" width="64" height="32"></rect>
-<rect x="356" y="253" width="64" height="32" class="nonterminal"></rect>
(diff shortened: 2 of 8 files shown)


mgartner:
Not related to your changes, but I noticed that it isn't mentioned that inverted indexes can be created on ARRAY types here:
image

mgartner (commented):
Looks great!!!

I left a few comments. Other than those, I think the only other thing potentially missing is that the optimizer is not perfect in proving that some query filters imply partial index predicates. From the [RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20200507_partial_indexes.md#usage):

> Note that CRDB, like Postgres, will perform a best-effort attempt to prove that a query filter expression implies a partial index predicate. It is not guaranteed to prove implication of arbitrarily complex expressions.

In other words, false negatives are possible (where a filter _theoretically_ implies a predicate, but cannot be proven by the optimizer in _practice_). It should be very unlikely, but it is possible, and calling it out in our docs may help prevent confusion. Here's one example:

```
root@127.0.0.1:58391/defaultdb> CREATE TABLE t (a INT, b INT, c INT, INDEX (a) WHERE b = 1 OR c = 2 OR b = 3);
CREATE TABLE

Server Execution Time: 4.515ms
Network Latency: 1.406ms

root@127.0.0.1:58391/defaultdb> EXPLAIN SELECT a FROM t WHERE b IN (1, 3) OR c = 2;
    tree    |     field     |       description
------------+---------------+---------------------------
            | distribution  | full
            | vectorized    | false
  filter    |               |
   │        | filter        | (b IN (1, 3)) OR (c = 2)
   └── scan |               |
            | missing stats |
            | table         | t@primary
            | spans         | FULL SCAN
(8 rows)
```

ericharmeling (commented):
TFTR, @mgartner!

I think I addressed all of your comments. I also added a note about the false negatives. I don't think we need to point out an example, but I agree that adding a disclaimer note will be helpful. Those changes are all in the "mgartner feedback" commit.

re: docs unrelated to partial indexes, I've added some simple updates to separate commits. I'd prefer to have separate PRs for unrelated docs updates, especially if they are more involved, but separate commits for these small updates should be fine.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8242#-:-MGnU3Bn5edIcwsmNwUx:bb0tvye)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mgartner)

___
*[v20.2/partial-indexes.md, line 19 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-19:-MGnOQMlAy2Mc1HmgfTw:b-hr7zgp) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L19)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

I think this last bullet point is confusing. The advantage of partial indexes in regards to writes is the the overhead of writing to an index is only incurred for rows that must be added or removed from the partial index, whereas a non-partial index incurs this overhead for every row. For example, if we have an `INDEX (a) WHERE b = 'foo'`, and we `INSERT INTO t (a, b) VALUES (1, 'bar')`, there is no overhead of writing to the partial index because that row does not belong. 

Would something like below be more clear?


> With a partial index, write queries only incur the overhead of an index write when the row satisfies the predicate. This contrasts with full indexes, which incur the overhead of an index write for all rows when the indexed column is modified.
</blockquote></details>

Gotcha. I rewrote that last bullet, using a lot of your wording.
___
*[v20.2/partial-indexes.md, line 58 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-58:-MGnOTh66UI40XtLZT7m:b-9vctks) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L58)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

It should be possible to create them in a `CREATE TABLE` statement. Let me know if you ran into a case that didn't work.

```
root@127.0.0.1:58391/defaultdb> create table t (a int, index (a) where a > 0);
CREATE TABLE

Server Execution Time: 3.143ms
Network Latency: 998µs

root@127.0.0.1:58391/defaultdb> show create table t;
  table_name |               create_statement
-------------+------------------------------------------------
  t          | CREATE TABLE public.t (
             |     a INT8 NULL,
             |     INDEX t_a_idx (a ASC) WHERE a > 0:::INT8,
             |     FAMILY "primary" (a, rowid)
             | )
(1 row)

Server Execution Time: 5.263ms
Network Latency: 231µs
```
</blockquote></details>

Ah! Okay. Removed this note. Didn't run into any issues.
___
*[v20.2/partial-indexes.md, line 80 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-80:-MGnOwo72bIes3UiGfGT:b-4mhl9i) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L80)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

Ahh I see now that you have this correct below in "Known Limitations". I think this Note should include the "DO UPDATE" clarification or be removed.
</blockquote></details>

I updated the Known Limitations bullet and removed the note.
___
*[v20.2/partial-indexes.md, line 149 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-149:-MGndolt8U8oRop7BoMl:bz82vzp) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L149)):*
> there is no query plan in this section that takes advantage of `revenue` as an indexed column.

I'm not sure I understand what you mean by this. All of the queries filter on `revenue`. Do you mean "takes advantage of `city` as an indexed column"?

I added an example that includes the city in the filter clause.
___
*[v20.2/sql-feature-support.md, line 80 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8B-r1-80:-MGnVL6p8Da6hLnYfR9w:b-h2pkp9) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/sql-feature-support.md#L80)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

Unrelated to partial indexes but I noticed this "Multiple indexes per query" is marked planned. Since https://github.com/cockroachdb/cockroach/issues/2142 was fixed by https://github.com/cockroachdb/cockroach/pull/47094, there is a case where a single query can use multiple indexes (example below). This will be new in 20.2.

```
root@127.0.0.1:58391/defaultdb> create table t (k int primary key, a int, b int, index a_idx (a), index b_idx (b));
CREATE TABLE

Server Execution Time: 2.755ms
Network Latency: 810µs

root@127.0.0.1:58391/defaultdb> explain select k from t where a = 10 or b = 20;
          tree         |     field     | description
-----------------------+---------------+--------------
                       | distribution  | local
                       | vectorized    | false
  distinct             |               |
   │                   | distinct on   | k
   └── union all       |               |
        ├── index join |               |
        │    │         | table         | t@primary
        │    └── scan  |               |
        │              | missing stats |
        │              | table         | t@a_idx
        │              | spans         | [/10 - /10]
        └── index join |               |
             │         | table         | t@primary
             └── scan  |               |
                       | missing stats |
                       | table         | t@b_idx
                       | spans         | [/20 - /20]
(17 rows)

Server Execution Time: 154µs
Network Latency: 311µs
```
</blockquote></details>

I don't see a docs issue opened for this, so I'll just sneak in an update to this table into this PR, making this support "partial" (in a separate commit). 

I'd rather document updates for fully in separate PRs. I opened an issue for this: https://github.com/cockroachdb/docs/issues/8260.

mgartner (commented):
> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8242)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v20.2/partial-indexes.md, line 58 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-58:-MGsnTmKEUCKgaWDjxFx:be2snno) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L58)):*

Ah! Okay. Removed this note. Didn't run into any issues.
</blockquote></details>

Should the `CREATE TABLE` docs be updated so that the `index_def` syntax graph also has an `opt_where_clause` at the end?
___
*[v20.2/partial-indexes.md, line 149 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-149:-MGsiiTcEsX0-ep7xtWd:bsbvb39) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L149)):*

> there is no query plan in this section that takes advantage of `revenue` as an indexed column.

I'm not sure I understand what you mean by this. All of the queries filter on `revenue`. Do you mean "takes advantage of `city` as an indexed column"?

I added an example that includes the city in the filter clause.
</blockquote></details>

Sorry for creating confusion.

This is a step in the right direction. Before, all the spans were `FULL SCAN`s meaning that they scan the entire partial index. This is a useful case, but another useful example that I wanted to highlight would be a scan over just a small part of the partial index.

Now that you've added `city = 'new york'`, you can see that the span is constrained to `[/'new york' - /'new york']`. :heavy_check_mark: 

If you want to take it a step further to show all the indexed columns being used to constrain the spans, you can change the query filter to `WHERE city = 'new york' AND revenue >= 100 AND revenue < 150`. 

Because `revenue >= 100 AND revenue < 150` implies `revenue > 80`, the partial index can be used. But, it will need to still apply the revenue filter to remove rows where revenue is between 80 and 99 and 150 and +infinity. Luckily, the revenue column is the second indexed column, and the first indexed column, city, is constrained to a single value by the `city = 'new york'` filter. So the scan over the partial index would constrain both indexed columns, `city` and `revenue`, with the span `[/'new york'/100 - /'new york'/149]`.

I think this better shows the full potential of a 2-column partial index. But it's up to you if you want to include it. You may want to reword (or remove?) the `EXPLAIN SELECT city, revenue FROM rides WHERE revenue > 95;` example since that is similar to my suggested example, but it does a `FULL SCAN` because the first indexed column, `city`, is not constrained by the query filter—a crucial difference that may be worth calling out if both examples are on the page.

ericharmeling (commented):
@mgartner Thanks for iterating on this! Just updated the PR again. See the latest commit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8242#-:-MGysaXb-7qXpmUUdVYr:b3tufx8)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mgartner)

___
*[v20.2/partial-indexes.md, line 58 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-58:-MGy69YqDW17E9GMHP7b:bwwtwuj) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L58)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

Should the `CREATE TABLE` docs be updated so that the `index_def` syntax graph also has an `opt_where_clause` at the end?
</blockquote></details>

Good catch! Just added it.
___
*[v20.2/partial-indexes.md, line 149 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-149:-MGy6Cvl5Qeim4S7S53A:b-m2ldtu) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L149)):*
<details><summary><i>Previously, mgartner (Marcus Gartner) wrote…</i></summary><blockquote>

Sorry for creating confusion.

This is a step in the right direction. Before, all the spans were `FULL SCAN`s meaning that they scan the entire partial index. This is a useful case, but another useful example that I wanted to highlight would be a scan over just a small part of the partial index.

Now that you've added `city = 'new york'`, you can see that the span is constrained to `[/'new york' - /'new york']`. :heavy_check_mark: 

If you want to take it a step further to show all the indexed columns being used to constrain the spans, you can change the query filter to `WHERE city = 'new york' AND revenue >= 100 AND revenue < 150`. 

Because `revenue >= 100 AND revenue < 150` implies `revenue > 80`, the partial index can be used. But, it will need to still apply the revenue filter to remove rows where revenue is between 80 and 99 and 150 and +infinity. Luckily, the revenue column is the second indexed column, and the first indexed column, city, is constrained to a single value by the `city = 'new york'` filter. So the scan over the partial index would constrain both indexed columns, `city` and `revenue`, with the span `[/'new york'/100 - /'new york'/149]`.

I think this better shows the full potential of a 2-column partial index. But it's up to you if you want to include it. You may want to reword (or remove?) the `EXPLAIN SELECT city, revenue FROM rides WHERE revenue > 95;` example since that is similar to my suggested example, but it does a `FULL SCAN` because the first indexed column, `city`, is not constrained by the query filter—a crucial difference that may be worth calling out if both examples are on the page.
</blockquote></details>

Ahh. I see. I've updated the example again!

mgartner (approved):
:lgtm_strong: This looks great!

> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8242#-:-MGyxR9f8vfq4Vr94MSA:b7y6g1b)* status: :shipit: complete! 1 of 0 LGTMs obtained

___
*[v20.2/partial-indexes.md, line 149 at r1](https://reviewable.io/reviews/cockroachdb/docs/8242#-MGKWqtgL2FRXOjZbR8C-r1-149:-MGyxO_Q0Yixws9C-45N:b9bkkp1) ([raw file](https://github.com/cockroachdb/docs/blob/7a895d983a575bbac5158731a94f0002991c02f1/v20.2/partial-indexes.md#L149)):*

Ahh. I see. I've updated the example again!
</blockquote></details>

Looks great!

lnhsingh (commented):
:lgtm: - just a couple of nits. Nice job!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8242#-:-MHC4v9uBKJP_r2HUCJq:bes1k8x)* status: :shipit: complete! 2 of 0 LGTMs obtained (waiting on @ericharmeling and @lnhsingh)

___
*[v20.2/partial-indexes.md, line 3 at r3](https://reviewable.io/reviews/cockroachdb/docs/8242#-MHC1xW34v5OM8FzCYCL:-MHC1xW34v5OM8FzCYCM:b-444v7y) ([raw file](https://github.com/cockroachdb/docs/blob/871302997d05831a6e0c244e8d6b226a2a72d6b0/v20.2/partial-indexes.md#L3)):*
> ```Markdown
> ---
> title: Partial Indexes
> summary: Partial indexes
> ```

nit: Add a more descriptive summary. A variation of the first sentence or two of the doc is good
___
*[v20.2/partial-indexes.md, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/8242#-MHC2kip0U5GQ8Inagym:-MHC2kiq2116UcRPfibb:bcq0at7) ([raw file](https://github.com/cockroachdb/docs/blob/871302997d05831a6e0c244e8d6b226a2a72d6b0/v20.2/partial-indexes.md#L7)):*
> ```Markdown
> ---
> 
> <span class="version-tag">New in v20.2:</span> Partial indexes allow you to specify a subset of rows and columns to add to an [index](indexes.html). Partial indexes include the subset of rows in a table that evaluate to true on a boolean *predicate expression* (i.e. a `WHERE` filter) defined at [index creation](#creation).
> ```

nit: add `,` after `i.e.`
___
*[v20.2/partial-indexes.md, line 186 at r3](https://reviewable.io/reviews/cockroachdb/docs/8242#-MHC3u8w25SGDiT0Mv0e:-MHC3u8xB4-R8V8tuVSp:b3ezkd2) ([raw file](https://github.com/cockroachdb/docs/blob/871302997d05831a6e0c244e8d6b226a2a72d6b0/v20.2/partial-indexes.md#L186)):*
> ```Markdown
> ~~~
> 
> Note that query's `SELECT` statement queries all columns in the `rides` table, not just the indexed columns. As a result, an "index join" is required on both the primary index and the partial index.
> ```

`Note that query's` > `Note that the query's`

PR #8249, opened 2020-09-08
[FIX] Since v20.x follower reads are 4.8 seconds or more in the past not 48 seconds

onno-vos-dev:


jseldess:
Sorry for the extreme delay here, @onno-vos-dev.

@rmloveland, can you PTAL and see if we still need to make these changes?

rmloveland:
Hi @onno-vos-dev thanks for this PR - sorry for the long delay!

This appears to have been fixed already via #7995

PR #8257, opened 2020-09-09
Remove version switcher from docs homepage

### LINKED ISSUE #8063: Versioning
No version in CC articles

yingzhuchin:
Fixes #8063.

@lnhsingh, @Amruta-Ranade, @jseldess: This PR removes the version switcher from the Docs homepage, which I believe was requested by Amruta.

While making this change, I noticed that there are CRDB pages linked from the CC part of the sidenav, namely the pages under "Use Your Cluster" > "Migrate Data." Those will have the version switcher displayed, since they are actually CRDB pages and not CC pages. Is that fine? Note that this is similar to the issue affecting breadcrumbs. When the same pages are linked from multiple sidenav locations, breadcrumbs will not work as expected.

### DIFF
--- a/_includes/version-nav.html
+++ b/_includes/version-nav.html
@@ -1,20 +1,6 @@
 <div class="version-nav-wrapper">
   <div class="version-nav">
-    <!-- <a
-      class="version-nav__link version-nav__link--crdb {% unless site.cockroachcloud %}version-nav__link--active{% endunless %}"
-      href="{% if site.cockroachcloud %}{{ "/../stable/" | relative_url }}{% else %}{{ "/stable/" | relative_url }}{% endif %}"
-    >
-      <span class="version-nav__link-text version-nav__link-text--crdb" ></span>
-    </a>
-
-    <a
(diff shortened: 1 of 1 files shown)


yingzhuchin:
> While making this change, I noticed that there are CRDB pages linked from the CC part of the sidenav, namely the pages under "Use Your Cluster" > "Migrate Data." Those will have the version switcher displayed, since they are actually CRDB pages and not CC pages.

One way to get around this that I can think of: In the CC folder, have hard symlinks pointing to pages in CRDB. But in order for this to work, symlinked pages cannot use the {{ page.version.version }} macro to include files, because those break in the unversioned CC folder. In other words, all files included from pages that need to be in both CC and CRDB have to be version-neutral.

Amruta-Ranade:
Thanks, YZ! I think it's okay if the CRDB pages show the version selector even if they are linked from the CC nav.

yingzhuchin:
Sounds good, thanks Amruta! I will merge this PR today then.

lnhsingh (approved):

Issue #8076, opened 2020-08-25
Error page: 404 messaging


yingzhuchin:
Docs 404 page not working #4908

Issue #8082, opened 2020-08-25
404 illustration


yingzhuchin:

yingzhuchin:
I noticed the original illustration is no longer in the latest Figma: https://www.figma.com/file/6XcxaUdad2RkjpEqnVHKAG/Education-docs-redesign?node-id=3277%3A12

Are we waiting for an updated illustration to go with the new color scheme?

luciaaaaaaa:
This issue will require design work from Marketing, since they are developing new illustration styles cc @anagio

PR #8415, opened 2020-09-16
Fix docs custom 404 page and reword messaging

### LINKED ISSUE #4908: Docs 404 page not working

yingzhuchin:
Fixes #4908 and #8076.

The custom 404 image is a stretch goal so I haven't attempted it with this PR.

Please preview the new 404 page here: https://deploy-preview-8415--cockroachdb-docs.netlify.app/docs/stable/404.html

You will not be able to see the page above by going to a broken link via preview site, because the way Netlify redirects, it will send you from a preview broken link to the 404 page in production. But you **can** use the Netlify preview site to verify that broken links are being redirected to the 404 page in production, which is an improvement over current behavior (see #4908).

I think this is one PR that we should make live immediately instead of waiting for all style updates. This is because the current 404 functionality is broken (see #4908), so we are essentially fixing a bug that affects users today.

### DIFF
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -3689,4 +3689,22 @@ a.accordion-toggle {
   color: #497EDA;
   font-weight: 900; }
 
+.page-not-found {
+  width: 60%;
+  margin: 0 auto;
+  padding: 50px;
+  text-align: center; }
+
+#page-not-found-heading {
+  color: #212B36;
+  font-size: 20px;
+  letter-spacing: -0.01em;
+  line-height: 32px; }
+
+#page-not-found-body {
+  color: #637282;
+  font-size: 14px;
+  letter-spacing: 0.1px;
+  line-height: 22px; }
+
 /*# sourceMappingURL=customstyles.css.map */
--- a/css/customstyles.scss
+++ b/css/customstyles.scss
@@ -621,3 +621,24 @@ a.accordion-toggle {
   color: #497EDA;
   font-weight: 900;
 }
+
+.page-not-found {
+  width: 60%;
+  margin: 0 auto;
+  padding: 50px;
+  text-align: center;
+}
+
+#page-not-found-heading {
+  color: #212B36;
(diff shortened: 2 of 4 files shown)


jseldess (approved):
:lgtm: 

Thanks, @yingzhuchin!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8415#-:-MHO7_6H2xnq1cCSVNc3:bv69t65)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @anagio, @jseldess, @lnhsingh, and @luciaaaaaaa)

lnhsingh (commented):
:lgtm: and I agree that it makes sense to push this one live

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8415#-:-MHRJNwf4kAl0ZH7y3MF:b-lz890l)* status: :shipit: complete! 2 of 0 LGTMs obtained (waiting on @anagio, @lnhsingh, and @luciaaaaaaa)

luciaaaaaaa:
Wondering if we want to add the 'Return to home' button, as a way for the user to have a clear action to take from here. Otherwise, lgtm!

yingzhuchin:
> Wondering if we want to add the 'Return to home' button, as a way for the user to have a clear action to take from here. Otherwise, lgtm!

I think since the sidebar and top navigation bar are both still accessible on the 404 page, the user should be good on what to do next. My two cents!

yingzhuchin:
I'll merge this for now since I'm shifting focus, but we can add any additional asks to #8082 and pick up from there!

PR #8428, opened 2020-09-17
update Production Checklist numbers

taroface:
Updates numbers first added in #7739.

Thread: https://cockroachlabs.slack.com/archives/C013VDTRA30/p1600109555012800

- @a-entin You mentioned that versions prior to 20.1 have a 1 TB/node storage capacity limit, per conversations with Raphael. We previously had 2 TB documented. Just wanted to note that I've adjusted that number down for 19.2.
- Added back wording on 32 vCPUs as an OLTP "sweet spot" and clarified that this is not a hard limit.

### DIFF
--- a/v19.2/recommended-production-settings.md
+++ b/v19.2/recommended-production-settings.md
@@ -40,9 +40,13 @@ Before deploying to production, test and tune your hardware setup for your appli
 
 #### CPU and memory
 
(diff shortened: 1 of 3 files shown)


bdarnell (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8428)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @a-entin)

PR #8453, opened 2020-09-23
Fix incorrect Follower Reads interval

### LINKED ISSUE #8262: Outdated follower reads interval in Duplicate Indexes Topology doc
Re: [Duplicate Indexes Topology](https://cockroachlabs.com/docs//v20.1/topology-duplicate-indexes.html)

## Issue Description

in the Alternative section https://www.cockroachlabs.com/docs/v20.1/topology-duplicate-indexes.html#alternatives the seconds are 48 instead of 4.8.

> If reads from a table can be historical (48 seconds or more in the past), consider the Follower Reads pattern.
 
## Suggested Resolution

> If reads from a table can be historical (4.8 seconds or more in the past), consider the Follower Reads pattern.

rmloveland:
... in both 20.1 and 20.2 docs.

Fixes #8262.

### DIFF
--- a/v20.1/topology-duplicate-indexes.md
+++ b/v20.1/topology-duplicate-indexes.md
@@ -179,7 +179,7 @@ Because this pattern balances the replicas for the table and its secondary index
 
 ## Alternatives
 
-- If reads from a table can be historical (48 seconds or more in the past), consider the [Follower Reads](topology-follower-reads.html) pattern.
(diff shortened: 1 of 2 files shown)


rmloveland:
Two-line change Eric - just want another set of eyes to avoid Murphy's Law - thanks!

ericharmeling (approved):
LGTM!

PR #8456, opened 2020-09-23
Update Docker links on 'Working with Spatial Data'

### LINKED ISSUE #8441: Update docker image link in 'Working with Spatial Data'
Right now (Tuesday, September 22, 2020) the `docker pull` command in [Working with Spatial Data](https://www.cockroachlabs.com/docs/v20.2/spatial-data) is pointing at an out of date custom build.  Need to update it to point at the most recent 20.2 beta.

rmloveland:
Fixes #8441.

### DIFF
--- a/v20.2/spatial-data.md
+++ b/v20.2/spatial-data.md
@@ -32,14 +32,14 @@ Get the spatial build of CockroachDB:
 
 {% include copy-clipboard.html %}
 ~~~ shell
-docker pull otanatcockroach/cockroachdb-geospatial
+docker pull cockroachdb/cockroach-unstable:v20.2.0-beta.2
 ~~~
 
 Start the server:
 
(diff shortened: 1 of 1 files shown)


rmloveland:
Hey Oliver!

Trivial change (and I tested it) but wanted a second pair of eyes.  `beta.2` not out yet as of this writing but at least #itssomething

otan (approved):
wonder if we can just use a constant somewhere for these things :P

rmloveland:
> wonder if we can just use a constant somewhere for these things :P

Ideally yes.  In the past we never really put links to alpha/beta stuff in the docs except on the release notes page for that release.  I think I'll file an issue in the docs repo to create a YAML tag that we can refer to that means "latest testing release" (if no such issue exists).

Also, thanks for looking at it!

PR #8460, opened 2020-09-23
Batch deleting from tables with explicit foreign key actions

ericharmeling:
- Added note on batch deleting from tables with explicit foreign key actions

### DIFF
--- a/v20.2/delete.md
+++ b/v20.2/delete.md
@@ -108,6 +108,10 @@ To delete a large number of rows (i.e., tens of thousands of rows or more), we r
 
 In the sections below, we provide guidance on batch deleting with the `DELETE` query filter [on an indexed column](#batch-delete-on-an-indexed-column) and [on a non-indexed column](#batch-delete-on-a-non-indexed-column). Filtering on an indexed column is both simpler to implement and more efficient, but adding an index to a table can slow down insertions to the table and may cause bottlenecks. Queries that filter on a non-indexed column must perform at least one full-table scan, a process that takes time proportional to the size of the entire table.
 
+{{site.data.alerts.callout_danger}}
+Exercise caution when batch deleting rows from tables with foreign key constraints and explicit [`ON DELETE` foreign key actions](foreign-key.html#foreign-key-actions) (e.g., [tables that enable interleaved fast path deletes](interleave-in-parent.html#interleaved-fast-path-deletes)). To preserve `DELETE` performance on tables with foreign key actions, we recommend using smaller batch sizes, as additional rows updated or deleted due to `ON DELETE` actions can make batch loops significantly slower, and will not be reflected in the output of a `DELETE` query.
+{{site.data.alerts.end}}
+
 ### Batch delete on an indexed column
 
 For high-performance batch deletes, we recommending filtering the `DELETE` query on an [indexed column](indexes.html).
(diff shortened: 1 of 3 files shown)


ericharmeling:
@drewdeally

v20.2/delete.md
```diff
@@ -108,6 +108,10 @@ To delete a large number of rows (i.e., tens of thousands of rows or more), we r
 
 In the sections below, we provide guidance on batch deleting with the `DELETE` query filter [on an indexed column](#batch-delete-on-an-indexed-column) and [on a non-indexed column](#batch-delete-on-a-non-indexed-column). Filtering on an indexed column is both simpler to implement and more efficient, but adding an index to a table can slow down insertions to the table and may cause bottlenecks. Queries that filter on a non-indexed column must perform at least one full-table scan, a process that takes time proportional to the size of the entire table.
 
+{{site.data.alerts.callout_danger}}
+Exercise caution when batch deleting rows from tables with foreign key constraints and explicit [`ON DELETE` foreign key actions](foreign-key.html#foreign-key-actions) (e.g., [tables that enable interleaved fast path deletes](interleave-in-parent.html#interleaved-fast-path-deletes)). To preserve `DELETE` performance on tables with foreign key actions, we recommend using smaller batch sizes, as additional rows updated or deleted due to `ON DELETE` actions can make batch loops significantly slower, and will not be reflected in the output of a `DELETE` query.
```
bdarnell, line 112: I think I'd remove the `e.g.` about interleaved tables. 

If I understood @drewdeally 's comment correctly, if you're using interleaved tables and ON DELETE CASCADE, the child table's rows *will* be counted in the DELETE result, so you would see numbers higher than the batch size (I haven't verified this myself). This is basically a bug, but one that can't really be fixed while preserving the fast cascading delete for interleaved tables.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8460)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @drewdeally)

___
*[v20.2/delete.md, line 112 at r1](https://reviewable.io/reviews/cockroachdb/docs/8460#-MHwtp9zARUVOH_u0pV6-r1-112:-MI0RJg28PHjglbD6GuD:bigg57n) ([raw file](https://github.com/cockroachdb/docs/blob/9def3e044a47b22f22566692fdd91abf7a2eb15b/v20.2/delete.md#L112)):*

I think I'd remove the `e.g.` about interleaved tables. 

If I understood @drewdeally 's comment correctly, if you're using interleaved tables and ON DELETE CASCADE, the child table's rows *will* be counted in the DELETE result, so you would see numbers higher than the batch size (I haven't verified this myself). This is basically a bug, but one that can't really be fixed while preserving the fast cascading delete for interleaved tables. 
</blockquote></details>

Updated the note.

v20.2/delete.md
drewdeally: Correct, my recommendation is to reflect this in the documentation so the user is not surprised the results.

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8460#-:-MIKdlqKEBH60G2UcULY:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @bdarnell and @lnhsingh)

PR #8489, opened 2020-09-29
Docs redesign

anagio:


jseldess:
@anagio, just pushed some link updates for the cc landing page and the main landing page. However, as I note in my commit, there's not yet a way to differentiate between the main landing page and the crdb docs landing page. So for now, I used the v20.1/index.md file as the main landing page, but the links will have to be fixed again depending on the solution you find. Also, I'm worried that your branch is quite far behind master. It predates our 20.2 release, when 20.2 was main the stable version, and more recently, we added a v21.1 version as well.

PR #8523, opened 2020-10-01
Special collations

### LINKED ISSUE #1471: Document special collations
Our collation docs link to a list of languages that we support, but there's no mention of other modifiers. One useful modified locale is `en_u_ks_level2`, which is case-insensitive english (I think the `_u_ks_level2` modifier can be added to other languages too to use their case-sensitivity rules). I'm not sure if there are other useful modifiers that are supported by our collation package. These modifiers are standardized, but the references I've found don't seem to agree with each other. Unicode [TR 35](http://www.unicode.org/reports/tr35/#Locale_Extension_Key_and_Type_Data) looks like the authority on this, although it doesn't define the `level2` syntax. [`collate/option.go`](https://github.com/golang/text/blob/master/collate/option.go) is where these modifiers get parsed.

ericharmeling:
Fixes #1471. 
 
- Added docs on collation subtags.
- Added note on deterministic and non-deterministic support.

### DIFF
--- a/v19.2/collate.md
+++ b/v19.2/collate.md
@@ -23,7 +23,49 @@ Collated strings are important because different languages have [different rules
 
 ## Supported collations
 
-CockroachDB supports the collations provided by Go's [language package](https://godoc.org/golang.org/x/text/language#Tag). The `<collation>` argument is the BCP 47 language tag at the end of each line, immediately preceded by `//`. For example, Afrikaans is supported as the `af` collation.
+CockroachDB supports collations identified by [Unicode locale identifiers](http://cldr.unicode.org/core-spec#Identifiers). For example, `en-US` identifies US English, `es` identifies Spanish, and `fr-CA` identifies Canadian French. Collation names are case-insensitive, and hyphens and underscores are interchangeable.
+
+{{site.data.alerts.callout_info}}
+If a hyphen is used in a SQL query, the collation name must be enclosed in double quotes, as single quotes are used for SQL string literals.
+{{site.data.alerts.end}}
+
+A list of supported collations can be found in the `pg_catalog.pg_collation` table:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT collname from pg_catalog.pg_collation;
+~~~
+
+~~~
+       collname
+-----------------------
+  und
+  aa
+  af
+  ar
+...
+(95 rows)
+~~~
+
+CockroachDB supports standard aliases for the collations listed in `pg_collation`. For example, `es-419` (Latin American Spanish) and `zh-Hans` (Simplified Chinese) are supported, but they do not appear in the `pg_collations` table because they are equivalent to the `es` and `zh` collations listed in the table.
+
+CockroachDB also supports the following Unicode locale extensions:
+
+- `co` (collation type)
+- `ks` (strength)
+- `kc` (case level)
+- `kb` (backwards second level weight)
+- `kn` (numeric)
+- `ks` (strength)
+- `ka` (alternate handling)
+
+To use a locale extension, append `-u-` to the base locale name, followed by the extension. For example, `en-US-u-ks-level2` is case-insensitive US English. The `ks` modifier changes the "strength" of the collation, causing it to treat certain classes of characters as equivalent (PostgreSQL calls these "non-deterministic collations"). Setting the `ks` to `level2` makes the collation case-insensitive (for languages that have this concept).
+
+For more details on locale extensions, see the [Unicode Collation Algorithm](https://unicode.org/reports/tr10/).
+
+## Collation versioning
+
+While changes to collations are rare, they are possible, especially in languages with a large numbers of characters (e.g., Simplifed and Traditional Chinese). CockroachDB updates its support with new versions of the Unicode standard every year, but there is currently no way to specify the version of Unicode to use. As a result, it is possible for a collation change to invalidate existing collated string data. To prevent collated data from being invalidated by Unicode changes, we recommend storing data in columns with an uncollated string type, and then using a [computed column](computed-columns.html) for the desired collation. In the event that a collation change produces undesired effects, the computed column can be dropped and recreated.
 
 ## SQL syntax
 
@@ -78,6 +120,30 @@ The sort will now honor the `de` collation that treats *ä* as *a* in alphabetic
   Bär
   Baz
 (3 rows)
+~~~
+
+### Specify collations with locale extensions
+
+{% include copy-clipboard.html %}
+~~~ sql
+> CREATE TABLE nocase_strings (greeting STRING COLLATE "en-US-u-ks-level2");
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> INSERT INTO nocase_strings VALUES ('Hello, friend.' COLLATE "en-US-u-ks-level2"), ('Hi. My name is Petee.' COLLATE "en-US-u-ks-level2");
+~~~
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM nocase_strings WHERE greeting = ('hi. my name is petee.' COLLATE "en-US-u-ks-level2");
+~~~
+
+~~~
+        greeting
++-----------------------+
+  Hi. My name is Petee.
+(1 row)
 ~~~
 
 ### Order by non-default collation
--- a/v20.1/collate.md
+++ b/v20.1/collate.md
@@ -23,7 +23,49 @@ Collated strings are important because different languages have [different rules
 
 ## Supported collations
 
-CockroachDB supports the collations provided by Go's [language package](https://godoc.org/golang.org/x/text/language#Tag). The `<collation>` argument is the BCP 47 language tag at the end of each line, immediately preceded by `//`. For example, Afrikaans is supported as the `af` collation.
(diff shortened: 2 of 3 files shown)


RaduBerinde (approved):
:lgtm: :100: :tada: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523#-:-MI_axVz5aKtjOayPikc:b-72oxhe)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @bdarnell)

bdarnell (commented):
I really got nerd-sniped on this one...

It's great that we have the `pg_collation` table; now that we have that I'd remove the talk about Go. 

In general it's time to make this more self-contained instead of linking out to external resources. For example, the link to "unicode language and locale identifiers"  includes a lot of stuff that is irrelevant (the `ca` calendar modifiers) and more importantly it includes some stuff that is relevant but not supported in Go (such as the `kf` case-first modifier [reference](https://github.com/golang/text/blob/a8b4671254579a87fadf9f7fa577dc7368e9d009/collate/option.go#L136)). 

I'm also seeing some more issues now that I've opened this can of worms. The `pg_collation` table shows the collation names containing hyphens instead of underscores, which requires using double quotes (not the usual sql single quotes) in the collation syntax. Can we change that table to use underscores? If not, we should probably use hyphens and double quotes throughout the docs to match. 

And versioning of collations are a potential nightmare. Collations rarely change, but now that i've found the [source data](https://github.com/unicode-org/cldr/tree/master/common/collation) I see that the chinese and japanese collations have both had changes in the last couple of years. So maybe we should add some best practices to always use uncollated strings for real columns and to only use collated string types via computed columns (and soon computed indexes).

So let me take a stab at the whole thing:

## Supported Collations

CockroachDB supports collations identified by [Unicode locale identifiers](http://cldr.unicode.org/core-spec#Identifiers). For example, `en-US` identifies US English, `es` is Spanish, and `fr-CA` is Canadian French. Collation names are case-insensitive, and hyphens and underscores are interchangeable. If a hyphen is used, the collation name must be enclosed in double quotes when used in SQL syntax (not single quotes, which are used for SQL string literals). 

A list of supported collations can be found in the `pg_catalog.pg_collation` table: `SELECT collname from pg_catalog.pg_collation;`). In addition, some aliases for these collations are also supported, although they do not appear in the table. For example, `es-419` (Latin American Spanish) and `zh-Hans` (Simplified Chinese) are supported, but they do not appear in the `pg_collations` table because they are equivalent to the basic `es` and `zh` collations. 

Some Unicode locale extensions are also supported. For example, the `ks` modifier changes the "strength" of a collation, causing it to treat certain classes of characters as equivalent (PostgreSQL calls theses "non-deterministic collations"). For example, setting `ks` to `level2` makes a collation case-insensitive (for languages that have this concept). To use one of these extensions, append `-u-` to the base locale name followed by the extension: `en-US-u-ks-level2` is case-insensitive US English.  Currently supported extensions are `co` (collation type), `ks` (strength), `kc` (case level), `kb` (backwards second level weight), `kn` (numeric), `ks` (strength), and `ka` (alternate handling). These extensions are defined in the [Unicode Collation Algorithm](https://unicode.org/reports/tr10/). 

## Collation Versioning

CockroachDB updates with new versions of the Unicode standard each year, and does not currently provide a mechanism to specify the version of Unicode in use. While changes to collations are rare, they are possible, especially in languages that use large numbers of characters such as Chinese. It is possible for a collation change to invalidate existing collated string data. For this reason we recommend storing data in columns with an uncollated string type, then using a computed column for the desired collation. This way, in the event a collation change has undesired effects, the computed column can be dropped and recreated. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523#-:-MI_aiIg49z6jEGVWVBG:b-jd9sil)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @bdarnell and @ericharmeling)

___
*[v19.2/collate.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/8523#-MI_dRlTBM7ETvNQTq1_:-MI_dRlTBM7ETvNQTq1a:bmxl159) ([raw file](https://github.com/cockroachdb/docs/blob/8be14c8671eef4450f26e61eb9f1042763110389/v19.2/collate.md#L28)):*
> ```Markdown
> CockroachDB supports the collation locales provided by Go's [language package](https://godoc.org/golang.org/x/text/language#Tag), using [BCP47-based](https://tools.ietf.org/html/bcp47) identifiers and extensions.
> 
> For example, the identifier `es-419` specifies Latin American Spanish, and `en_US_u_ks_level1` specifies case-insensitive American English.
> ```

`es-419` doesn't show up in `pg_collation` (although it appears to work). I think this is because latin american spanish uses the same collation rules as spain spanish. `fr-ca` might be a better example since canadian french does have different rules than in france. 

Down below you use `ks_level2` for case-insensitive instead of `ks_level1`. I think `level2` is the right answer here (level1 is both case- and accent-insensitive)

ericharmeling (commented):
TFTR @raduberinde !

And thank you for the thorough review and feedback @bdarnell !

Following your feedback, I've updated the 19.2 files for a second review (will propagate to other versions when approved). I basically just copyedited what you wrote above, and then updated the example following your advice. Fortunately, there are no other instances of local extensions in the docs, so we don't need to replace any other examples with double quotes and hyphens.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523#-:-MIzaDzBEZLqGgs0TJjS:b6jd8p9)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @bdarnell)

bdarnell (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @bdarnell)

lnhsingh (commented):
LGTM! One comment on a format suggestion 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523#-:-MJE1x0eETPoie12-TPM:b-5ex9c0)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @bdarnell, @ericharmeling, and @lnhsingh)

___
*[v19.2/collate.md, line 52 at r3](https://reviewable.io/reviews/cockroachdb/docs/8523#-MJE1OUu6QPCJ7SE_2Lu:-MJE1OUu6QPCJ7SE_2Lv:b-jwcy9h) ([raw file](https://github.com/cockroachdb/docs/blob/17c6a7f829f184870cb78195dd268541b557f504/v19.2/collate.md#L52)):*
> ```Markdown
> CockroachDB supports standard aliases for the collations listed in `pg_collation`. For example, `es-419` (Latin American Spanish) and `zh-Hans` (Simplified Chinese) are supported, but they do not appear in the `pg_collations` table because they are equivalent to the `es` and `zh` collations listed in the table.
> 
> CockroachDB also supports the following Unicode locale extensions: `co` (collation type), `ks` (strength), `kc` (case level), `kb` (backwards second level weight), `kn` (numeric), `ks` (strength), and `ka` (alternate handling). To use a locale extension, append `-u-` to the base locale name, followed by the extension. For example, `en-US-u-ks-level2` is case-insensitive US English. The `ks` modifier changes the "strength" of the collation, causing it to treat certain classes of characters as equivalent (PostgreSQL calls these "non-deterministic collations"). Setting the `ks` to `level2` makes the collation case-insensitive (for languages that have this concept).
> ```

Format suggestion: Make the Unicode locale extensions a bulleted list for better readability (this applies to the other versions too)
___
*[v19.2/collate.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/8523#-MJE1qlX-TkCZoAoDG4j:-MJE1qlX-TkCZoAoDG4k:b-2zntdi) ([raw file](https://github.com/cockroachdb/docs/blob/17c6a7f829f184870cb78195dd268541b557f504/v19.2/collate.md#L124)):*
> ```Markdown
> ~~~ sql
> > INSERT INTO nocase_strings VALUES ('Hello, friend.' COLLATE "en-US-u-ks-level2"), ('Hi. My name is Petee.' COLLATE "en-US-u-ks-level2");
> ```

I like this example lol :dog:

ericharmeling (commented):
TFTR, @lnhsingh !

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8523#-:-MJSc3vt2MmPa6mEa4Fs:btvgpor)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @bdarnell and @lnhsingh)

___
*[v19.2/collate.md, line 52 at r3](https://reviewable.io/reviews/cockroachdb/docs/8523#-MJE1OUu6QPCJ7SE_2Lu:-MJSc3Fv6w1wpUAxVV0q:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/17c6a7f829f184870cb78195dd268541b557f504/v19.2/collate.md#L52)):*

Format suggestion: Make the Unicode locale extensions a bulleted list for better readability (this applies to the other versions too)
</blockquote></details>

Done.
___
*[v19.2/collate.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/8523#-MJE1qlX-TkCZoAoDG4j:-MJSVwM6CEPRn8EdNzqF:b32vhw) ([raw file](https://github.com/cockroachdb/docs/blob/17c6a7f829f184870cb78195dd268541b557f504/v19.2/collate.md#L124)):*

I like this example lol :dog: 
</blockquote></details>

:)

Issue #7781, opened 2020-07-23
Make replication zone constraints error more visible in docs


florence-crl:
According to the documentation [here](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#replication-zone-variables) about the `constraints` variable:

> To prevent hard-to-detect typos, constraints placed on store attributes and node localities must match the values passed to at least one node in the cluster. If not, an error is signalled.

Please make the error more visible in the documentation by mentioning something like:

"The preferred way to prevent this error is to configure Replication Zones constraints after at least one active node in the cluster matches the passed in values. The `constraints` variable is evaluated when the ALTER...CONFIGURE ZONE statement is run."

zuliatowoade:
Hello, I will like to work on this issue. Can you please assign it to me? Thank you.

Amruta-Ranade:
Hi @zuliatowoade ..we are asking folks to work on one issue at a time. Feel free to check back after you work on #8018

dgabriel123:
Hello @Amruta-Ranade, may I work on this issue?

Amruta-Ranade:
@dgabriel123 Yep!

dgabriel123:
Thanks @Amruta-Ranade! By the way, there are 5 versions of the file. Should I edit all or just 19.2 + 20.1 + 20.2?

Amruta-Ranade:
Only 19.2, 20.1, 20.2 :)

dgabriel123:
> Only 19.2, 20.1, 20.2 :)

### OK thanks! I have a question about linking from an include file. I'm adding [`ALTER ... CONFIGURE ZONE`](configure-zone.html) to variables.md, but I think because the file is an include, the link is sending me to a 404 page when I click it in Preview mode. Does this URL change once the file is updated on the back end, or should I maybe use an absolute link (https://www.cockroachlabs.com/docs/v20.1/configure-zone) instead?

Sorry if this was confusing.

dgabriel123:
> Only 19.2, 20.1, 20.2 :)

thanks @Amruta-Ranade, ready for review.

florence-crl:
This docs issue was originally created due to zendesk [ticket 5880](https://cockroachdb.zendesk.com/agent/tickets/5880).

Amruta-Ranade:
Thanks for helping me understand the issue, @florence-crl !

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/8537

PR #8537, opened 2020-10-05
Replication Zone Constraints error

dgabriel123:
Updated the language for the 'constraints' variable. Updated in versions 19.2 20.1 and 20.2 (#7781 ).


Amruta-Ranade:
Can we reword the new text as:
"To prevent this error, make sure at least one active node is configured to match the constraint. For example, apply  `constraints = '[-region=west]'` only if you had set `--locality=region=west` for at least one node while starting the cluster."

cc @florence-crl

dgabriel123:
> Can we reword the new text as:
> "To prevent this error, make sure at least one active node is configured to match the constraint. For example, apply `constraints = '[-region=west]'` only if you had set `--locality=region=west` for at least one node while starting the cluster."
> 
> cc @florence-crl

all set with the copy update @Amruta-Ranade . Let me know if I missed anything.

Amruta-Ranade:
I made a mistake in the suggested text..it should be `constraints = '[+region=west]'` instead of `constraints = '[-region=west]'`. Sorry!

dgabriel123:
> I made a mistake in the suggested text..it should be `constraints = '[+region=west]'` instead of `constraints = '[-region=west]'`. Sorry!

No problem at all! Corrected.

PR #8543, opened 2020-10-06
v20.2 functions and operators refresh

ericharmeling:
Refreshing with latest functions/operators.

@rmloveland adding you to review, as most (if not all) of these new functions touch spatial features and will likely resolve issues currently assigned to you.


rmloveland (approved):
Eric this LGTM - modulo that the broken build is because I added some direct anchor links like "operator-less-than" which are being pointed to from elsewhere. I think if you leave those anchors in everything else LGTM

ericharmeling:
> Eric this LGTM - modulo that the broken build is because I added some direct anchor links like "operator-less-than" which are being pointed to from elsewhere. I think if you leave those anchors in everything else LGTM

TFTR @rmloveland ! I went ahead and changed those references to match the generated functions page. I don't think the manual work of adding in those anchors is worth the accuracy of the links on the operator page.

PR #8545, opened 2020-10-06
Clean up 'build_for' from front matter

### LINKED ISSUE #8276: Remove build_for from front matter
Since we [moved the CC docs into the same sidebar as CRDB docs](https://github.com/cockroachdb/docs/pull/8086), we no longer need the `build_for` field in the front matter of a doc. We should clean up the docs and remove this field from all docs.

Note: If a doc needs to be included in both sections, there should be two copies (one in `vX.x` dir and one in `cockroachcloud` dir) and both should be maintained for each product.

yingzhuchin:
Fixes #8276.


jseldess (approved):
LGTM

Thanks, @yingzhuchin!

lnhsingh (approved):

PR #8548, opened 2020-10-06
v20.2 session variables update

ericharmeling:
Fixes #8148.
Fixes #8319. 
Fixes #7614.

This PR adds the new v20.2 session variables to the list of exposed session variables in the v20.2 docs. 

The PR also includes the following fixes:

- Updates to the session variables in v20.1 and v20.2 that were added in previous CRDB versions, but were not previously documented (https://github.com/cockroachdb/cockroach/pull/37333, https://github.com/cockroachdb/cockroach/pull/35116, https://github.com/cockroachdb/cockroach/pull/42914, https://github.com/cockroachdb/cockroach/pull/48333)
- Changed `prefer_lookup_joins_for_fk` to `prefer_lookup_joins_for_fks`, to reflect the true name of the variable.

Adding @otan as primary engineering reviewer, but feel free to add anyone else.

PR #8549, opened 2020-10-06
Disallow full table scans

### LINKED ISSUE #8313: sql: add session/cluster settings to disallow full table/index scans
PR: https://github.com/cockroachdb/cockroach/pull/52278

From release notes:
> Added a new public [cluster setting](../v20.2/cluster-settings.html) `sql.defaults.disallow_full_table_scans.enabled` that defaults to false, which informs the session setting `disallow_full_table_scans`. If the session is configured to disallow full table/index scans, any user query that plans a full table/index scan is rejected. This does not apply to internal queries or queries over virtual schemas (e.g., `pg_catalog`). [#52278][#52278] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #8313.

See #8548 for updates to the session variables docs.

### DIFF
--- a/v20.2/explain.md
+++ b/v20.2/explain.md
@@ -18,6 +18,8 @@ Using `EXPLAIN`'s output, you can optimize your queries by taking the following 
 
 - Avoid scanning an entire table, which is the slowest way to access data. You can avoid this by [creating indexes](indexes.html) that contain at least one of the columns that the query is filtering in its `WHERE` clause.
 
+    <span class="version-tag">New in v20.2</span>: You can disable query plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html). When `disallow_full_table_scans=on`, attempting to execute a query with a plan that includes a full table scan will return an error.
+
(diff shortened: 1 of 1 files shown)


arulajmani (approved):

lnhsingh (commented):
One nit with layout but otherwise LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8549#-:-MJ2Zzv54GufyKPXNihU:b-upq4vz)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling and @lnhsingh)

___
*[v20.2/explain.md, line 21 at r1](https://reviewable.io/reviews/cockroachdb/docs/8549#-MJ2ZNEg1TQ8x5DHqNes:-MJ2ZNEg1TQ8x5DHqNet:brvvi1r) ([raw file](https://github.com/cockroachdb/docs/blob/b8a929e1da6d7ac65c942fb36d40cc8308b9ac8a/v20.2/explain.md#L21)):*
> ```Markdown
> - Avoid scanning an entire table, which is the slowest way to access data. You can avoid this by [creating indexes](indexes.html) that contain at least one of the columns that the query is filtering in its `WHERE` clause.
> 
>   <span class="version-tag">New in v20.2</span>: You can disable query plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html). When `disallow_full_table_scans=on`, attempting to execute a query with a plan that includes a full table scan will return an error.
> ```

Is this paragraph supposed to be nested under the bullet above? If so, I think the whole text block needs to indented, not just the first line

ericharmeling:
TFTR @lnhsingh !

Fixed the formatting.

lnhsingh:
Looks good!

PR #8612, opened 2020-10-08
Use include for enterprise feature callout, where possible

### LINKED ISSUE #7095: BACKUP doc "warning" prompt is not really a warning
Re: [BACKUP](https://cockroachlabs.com/docs//v19.2/backup.html)

## Issue Description

We have a loud, red "warning" prompt for enterprise features. We shouldn't, they're not that scary.

## Suggested Resolution 

Something like "Info" + perhaps a green background seems less aggressive.

jseldess:
Also confirmed that these callouts are not longer red warnings.

Fixes #7095.


jseldess:
@irfansharif, this isn't very contentious, so I'll just merge.

PR #8615, opened 2020-10-08
update rolling upgrades guidance

### LINKED ISSUE #6319: Proceeding from one node to another during a rolling upgrade
Our [rolling upgrade instructions](https://www.cockroachlabs.com/docs/stable/upgrade-cockroach-version.html#step-4-perform-the-rolling-upgrade) say:

> Upgrade only one node at a time, and wait at least one minute after a node rejoins the cluster to upgrade the next node. Simultaneously upgrading more than one node increases the risk that ranges will lose a majority of their replicas and cause cluster unavailability.
>
> Also, refrain from starting schema changes during the upgrade process. Schema changes are complex operations that involve coordination across nodes and can increase the potential for unexpected behavior during an upgrade.

In our [Kubernetes-specific upgrade instructions](https://www.cockroachlabs.com/docs/stable/orchestrate-cockroachdb-with-kubernetes.html#upgrade-the-cluster), we just say to change the Docker image in the helm chart or Statefulset, which triggers a restart of the pods sequentially, but without any controlled waiting or checks in-between.

A customer noticed this discrepancy and asked about the wait period, which lead to the following guidance from @andy-kimball:

1. Wait until `select count(*) FROM crdb_internal.gossip_nodes where is_live=false;` is zero. This is simple, and relies on the assumption is that if all your nodes are live, then there is very low risk of under-replicated ranges. During discussions, we couldn’t confirm any relevant cases where under-replicated ranges would occur on a fully-live cluster.
2. Shut down next node using `cockroach quit` or `SIGTERM`. There is some danger that you’d see higher query tail latencies if all leaseholders cannot be drained within a 5s period of time. If in practice you’re seeing this be a problem, we can make this timeout be configurable, and you could set it to 1m or 5m, or whatever.
3. Bring node back up after upgrade.
4. Proceed to next node by going back to step #1.

We should think about updating the upgrade instructions to reflect this guidance. We should also update the Kubernetes-specific instructions to follow a similar approach. The CC team is already doing something similar, according to @DuskEagle:

> Instead of using `select count(*) FROM crdb_internal.gossip_nodes where is_live=false;` to detect when a node is live, we are using a SQL query similar to the one used by `cockroach node status --ranges` to compute the underreplicated ranges in the cluster, and only proceeding when the number of underrepicated ranges is zero: `select sum((metrics->>'ranges.underreplicated')::DECIMAL)::INT AS ranges_underreplicated from crdb_internal.kv_store_status;`
>
> besides that, when updating the statefulset with a new version of CockroachDB, we also set the `spec.updateStrategy.rollingUpdate.partition` value: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions. If we have 3 pods in the statefulset, named crdb-0, crdb-1, and crdb-2, we will first set the partition value to 2. This means that only crdb-2 will be updated to the new Cockroach version.
>
> We then wait until the pod is healthy, and the number of underreplicated ranges is 0, and then reduce the partition number by 1. This causes crdb-1 to be updated to the new version. And so on.

taroface:
Fixes #6319.

- For non-orchestrated deployments, this uses guidance from @knz on determining that a node is ready to accept SQL and the next node can be stopped.
- For K8s deployments, uses the guidance from @DuskEagle in the linked issue.

@DuskEagle for the Kubernetes section, it seems the partition would have to be removed from the StatefulSet after the upgrade? Otherwise wouldn't any updates to the STS only affect a single pod? Can you advise on how this would be done?

Also, I am not 100% certain the Helm command is correct. Will need to test.

### DIFF
--- a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
+++ b/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
@@ -81,7 +81,7 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
         > \q
         ~~~
 
-1. Add a [partition](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#staging-an-update) to the `updateStrategy` defined in the StatefulSet. For a cluster with 3 pods, the partition value should be 2:
+1. Add a [partition](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#staging-an-update) to the `updateStrategy` defined in the StatefulSet. Only the pods numbered greater than or equal to the partition value will be updated. For a cluster with 3 pods (e.g., `cockroachdb-0`, `cockroachdb-1`, `cockroachdb-2`) the partition value should be 2:
 
     <section class="filter-content" markdown="1" data-scope="manual">
     {% include copy-clipboard.html %}
@@ -247,7 +247,7 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     > \q
     ~~~
 
-1. Decrement the partition value by 1:
+1. Decrement the partition value by 1 to allow the next pod in the cluster to update:
 
     <section class="filter-content" markdown="1" data-scope="manual">
     {% include copy-clipboard.html %}
@@ -270,8 +270,6 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     --set statefulset.updateStrategy.rollingUpdate.partition=1 \
     ~~~
     </section>
-
-    This allows the next pod in the cluster to update. 
 
 1. Repeat steps 4-8 until all pods have been restarted and are running the new image (the final partition value should be `0`).
 
--- a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
+++ b/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
@@ -101,7 +101,7 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     $ helm upgrade \
     my-release \
     cockroachdb/cockroachdb \
-    --set statefulset.updateStrategy.rollingUpdate.partition=2 \
+    --set statefulset.updateStrategy.rollingUpdate.partition=2
     ~~~
     </section>
 
--- a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
+++ b/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
@@ -35,7 +35,6 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
         ~~~
         </section>
 
-
     {% else %}
 
     1. Launch a temporary interactive pod and start the [built-in SQL client](cockroach-sql.html) inside it:
@@ -68,21 +67,45 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
 
     {% endif %}
 
-    2. Set the `cluster.preserve_downgrade_option` [cluster setting](cluster-settings.html):
+    1. Set the `cluster.preserve_downgrade_option` [cluster setting](cluster-settings.html):
 
         {% include copy-clipboard.html %}
         ~~~ sql
         > SET CLUSTER SETTING cluster.preserve_downgrade_option = '19.2';
         ~~~
 
-    3. Exit the SQL shell and delete the temporary pod:
+    1. Exit the SQL shell and delete the temporary pod:
 
         {% include copy-clipboard.html %}
         ~~~ sql
         > \q
         ~~~
 
-2. Kick off the upgrade process by changing the desired Docker image:
+1. Add a [partition](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#staging-an-update) to the `updateStrategy` defined in the StatefulSet. For a cluster with 3 pods, the partition value should be 2:
+
+    <section class="filter-content" markdown="1" data-scope="manual">
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ kubectl patch statefulset cockroachdb \
+    -p='{"spec":{"updateStrategy":{"type":"RollingUpdate","rollingUpdate":{"partition":2}}}}'
+    ~~~
+
+    ~~~
+    statefulset.apps/cockroachdb patched
+    ~~~
+    </section>
+
+    <section class="filter-content" markdown="1" data-scope="helm">
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ helm upgrade \
+    my-release \
+    cockroachdb/cockroachdb \
+    --set statefulset.updateStrategy.rollingUpdate.partition=2 \
+    ~~~
+    </section>
+
+1. Kick off the upgrade process by changing the desired Docker image:
 
     <section class="filter-content" markdown="1" data-scope="manual">
     {% include copy-clipboard.html %}
@@ -118,7 +141,7 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     ~~~
     </section>
 
-3. If you then check the status of your cluster's pods, you should see them being restarted:
+1. Check the status of your cluster's pods. You should see one of them being restarted:
 
     {% include copy-clipboard.html %}
     ~~~ shell
@@ -130,8 +153,7 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     NAME            READY     STATUS        RESTARTS   AGE
     cockroachdb-0   1/1       Running       0          2m
     cockroachdb-1   1/1       Running       0          2m
-    cockroachdb-2   1/1       Running       0          2m
-    cockroachdb-3   0/1       Terminating   0          1m
+    cockroachdb-2   0/1       Terminating   0          1m
     ...
     ~~~
     </section>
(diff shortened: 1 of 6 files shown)


v20.2/upgrade-cockroach-version.md
```diff
@@ -192,7 +196,18 @@ We recommend creating scripts to perform these steps instead of performing them
...
+1. After the node has rejoined the cluster, ensure that the node is ready to accept a SQL connection.
+
+    Unless there are tens of thousands of ranges on the node, it's usually sufficient to wait at least one minute. To be certain that the node is ready, run the following command:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    cockroach sql -e 'select 1'
+    ~~~
+
+    The command should complete. If it hangs, the node is not yet ready.
```
knz, line 208: The command always complete. The text should explain "The command will automatically wait to complete until the node is ready."

knz (approved):
Just a minor nit on my side. I'll let Joel evaluate the k8s guidance.

v20.2/upgrade-cockroach-version.md
```diff
@@ -177,11 +183,9 @@ We recommend creating scripts to perform these steps instead of performing them
     $ systemctl start <systemd config filename>
     ~~~
 
-1. Verify the node has rejoined the cluster through its output to `stdout` or through the [Admin UI](admin-ui-overview.html).
+1. Verify the node has rejoined the cluster through its output to [`stdout`](cockroach-start.html#standard-output) or through the [Admin UI](admin-ui-cluster-overview-page.html#node-status). 
 
-    {{site.data.alerts.callout_info}}
-    To access the Admin UI for a secure cluster, [create a user with a password](create-user.html#create-a-user-with-a-password). Then open a browser and go to `https://<any node's external IP address>:8080`. On accessing the Admin UI, you will see a Login screen, where you will need to enter your username and password.
-    {{site.data.alerts.end}}
+    The node should have at least 70% of the [memory usage](admin-ui-cluster-overview-page.html#node-details) you observed before stopping the node.
```
DuskEagle, line 188: I don't think this advice is sound, it depends too much on how long the node was running previously and what the workload was on the node before it was restarted. For example, if the node was running for long enough that the RocksDB/Pebble cache was full, but there was relatively little load on the cluster, the memory usage on the cluster could be expected to be much lower after starting the node again.

DuskEagle (changes requested):
Helm instructions LGTM.

> @DuskEagle for the Kubernetes section, it seems the partition would have to be removed from the StatefulSet after the upgrade? Otherwise wouldn't any updates to the STS only affect a single pod? Can you advise on how this would be done?

No, because the partition value ensures that any pods numbered **greater than or equal to** the partition value are updated. For example, if the partition value is 1, and we have pods `cockroachdb-0`, `cockroachdb-1`, and `cockroachdb-2`, then `cockroachdb-1` and `cockroachdb-2` will be updated. 

If the partition value is 0, then that implies all pods will be updated.

It is also fine to remove the partition value after the update is finished; it's just not necessary.

taroface (commented):
Got it. I added this clarification to the doc.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8615#-:-MJSzMapC3J6P7tIO7B2:bqy5xp5)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @DuskEagle and @knz)

___
*[v20.2/upgrade-cockroach-version.md, line 188 at r1](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00U-r1-188:-MJSzci-EN37zPqXGUvs:b-7i85n1) ([raw file](https://github.com/cockroachdb/docs/blob/2610845d3a37a60ed16b62b3bd55475165938614/v20.2/upgrade-cockroach-version.md#L188)):*
<details><summary><i>Previously, DuskEagle (Joel Kenny) wrote…</i></summary><blockquote>

I don't think this advice is sound, it depends too much on how long the node was running previously and what the workload was on the node before it was restarted. For example, if the node was running for long enough that the RocksDB/Pebble cache was full, but there was relatively little load on the cluster, the memory usage on the cluster could be expected to be much lower after starting the node again.
</blockquote></details>

Okay, will strike the memory usage from the guidance.
___
*[v20.2/upgrade-cockroach-version.md, line 208 at r1](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00U-r1-208:-MJT27WqEPXqyb457vQo:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2610845d3a37a60ed16b62b3bd55475165938614/v20.2/upgrade-cockroach-version.md#L208)):*

The command always completes. The text should explain "The command will automatically wait to complete until the node is ready."
</blockquote></details>

Done.

v20.2/upgrade-cockroach-version.md
```diff
@@ -192,7 +192,18 @@ We recommend creating scripts to perform these steps instead of performing them
 
     If you leave versioned binaries on your servers, you do not need to do anything.
 
-1. Wait at least one minute after the node has rejoined the cluster, and then repeat these steps for the next node.
+1. After the node has rejoined the cluster, ensure that the node is ready to accept a SQL connection.
+
+    Unless there are tens of thousands of ranges on the node, it's usually sufficient to wait at least one minute. To be certain that the node is ready, run the following command:
```
Amruta-Ranade, line 197: nit:"at least" seems contradictory to what the statement seems to convey.

_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
```diff
@@ -152,7 +173,107 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
     {{site.data.alerts.end}}
     </section>    
 
-4. This will continue until all of the pods have restarted and are running the new image. To check the image of each pod to determine whether they've all be upgraded, run:
+1. After the pod has been restarted with the new image, get a shell into the pod and start the CockroachDB [built-in SQL client](cockroach-sql.html):
```
Amruta-Ranade, line 176: The numbering seems off here but renders correctly ¯\_(ツ)_/¯

_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
```diff
@@ -68,21 +67,45 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
...
-    3. Exit the SQL shell and delete the temporary pod:
+    1. Exit the SQL shell and delete the temporary pod:
 
         {% include copy-clipboard.html %}
         ~~~ sql
         > \q
         ~~~
 
-2. Kick off the upgrade process by changing the desired Docker image:
+1. Add a [partition](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#staging-an-update) to the `updateStrategy` defined in the StatefulSet. Only the pods numbered greater than or equal to the partition value will be updated. For a cluster with 3 pods (e.g., `cockroachdb-0`, `cockroachdb-1`, `cockroachdb-2`) the partition value should be 2:
```
Amruta-Ranade, line 84: nit: Should this be "the `updateStrategy` field"?

_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md
```diff
@@ -68,21 +67,45 @@ Kubernetes knows how to carry out a safe rolling upgrade process of the Cockroac
...
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ helm upgrade \
+    my-release \
+    cockroachdb/cockroachdb \
+    --set statefulset.updateStrategy.rollingUpdate.partition=2 \
+    ~~~
+    </section>
+
+1. Kick off the upgrade process by changing the desired Docker image:
```
Amruta-Ranade, line 108: Suggestion: "desired Docker image" > "Docker image used in the CockroachDB StatefulSet "

Amruta-Ranade (approved):

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8615)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @Amruta-Ranade, @DuskEagle, and @knz)

___
*[_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md, line 84 at r2](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00P-r2-84:-MJYNjV8Ec2Qn-e6iEF0:b-bqi2nw) ([raw file](https://github.com/cockroachdb/docs/blob/357dc85be9184adab4474ed6bb6b420cf8a98c3a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md#L84)):*

nit: Should this be "the `updateStrategy` field"?
</blockquote></details>

Hmm good point. I'm just going to make this "update strategy" to avoid getting too far into the weeds.
___
*[_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md, line 108 at r2](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00P-r2-108:-MJYO0HeD8nWo3Tp85eG:b1natol) ([raw file](https://github.com/cockroachdb/docs/blob/357dc85be9184adab4474ed6bb6b420cf8a98c3a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md#L108)):*

Suggestion: "desired Docker image" > "Docker image used in the CockroachDB StatefulSet "
</blockquote></details>

Perfect! Done.
___
*[_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md, line 176 at r2](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00P-r2-176:-MJYO3YpBXJyoKsaLffb:bpqk4ih) ([raw file](https://github.com/cockroachdb/docs/blob/357dc85be9184adab4474ed6bb6b420cf8a98c3a/_includes/v20.1/orchestration/kubernetes-upgrade-cluster.md#L176)):*

The numbering seems off here but renders correctly ¯\_(ツ)_/¯
</blockquote></details>

Yeah, Markdown automatically uses correct numbering regardless of the number used. I think we settled on using `1.` throughout ordered lists as a docs practice. IMO this helps keep the numbering clear as steps are added and removed in updates.
___
*[v20.2/upgrade-cockroach-version.md, line 197 at r2](https://reviewable.io/reviews/cockroachdb/docs/8615#-MJ9ZZM-QaM0QG4Yz00U-r2-197:-MJYOevJ4MKhAcEUltLZ:bi2qzy1) ([raw file](https://github.com/cockroachdb/docs/blob/357dc85be9184adab4474ed6bb6b420cf8a98c3a/v20.2/upgrade-cockroach-version.md#L197)):*

nit:"at least" seems contradictory to what the statement seems to convey. 
</blockquote></details>

Good point ... I removed this.

Issue #8513, opened 2020-09-30
release-20.2: backport ST_S2Covering


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/54433

From release notes:
> Implemented `ST_S2Covering`, which returns the S2 coverings used for indexing for geometry and geography types. [#54433][#54433] {% comment %}doc{% endcomment %}

rmloveland:
Fixed by #8617 which has usage examples, etc.

Issue #8377, opened 2020-09-15
builtins: implement ST_Collect aggregate


jseldess:
Jesse Seldess (jseldess) commented:

PR: https://github.com/cockroachdb/cockroach/pull/53645

From release notes:
> Implemented the geometry aggregate [builtin](../v20.2/functions-and-operators.html#built-in-functions] `ST_Collect`. [#53645][#53645) {% comment %}doc{% endcomment %}

Jira Issue: DOC-757

Issue #8356, opened 2020-09-15
sql: support specifying custom index parameters for geospatial indexes


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/52800

From release notes:
> Introduced the `s2_max_level`, `s2_level_mod`, and `s2_max_cells` storage parameters for modifying the S2 parameters for indexing [`GEOMETRY` and `GEOGRAPHY`](../v20.2/spatial-data.html) data types in an inverted index. [#52800][#52800] {% comment %}doc{% endcomment %}

rmloveland:
Fixed in #8617

Issue #8172, opened 2020-08-26
parser: allow USING GIST syntax for creating inverted indexes


lnhsingh:
PR: https://github.com/cockroachdb/cockroach/pull/51516

From release notes:
> Added support for the `USING GIST` syntax to create an [inverted index](../v20.2/inverted-indexes.html) on `GEOMETRY` and `GEOGRAPHY` columns. [#51516][#51516] {% comment %}doc{% endcomment %}

ericharmeling:
@rmloveland This looks like a spatial feature?

rmloveland:
yes, snagged

rmloveland:
Fixed in #8617

PR #8617, opened 2020-10-09
Add docs on spatial indexes

### LINKED ISSUE #8563: Document Geospatial Indexing

rmloveland:
Fixes #8563.

Addresses #7835, #7900, #7901, #7902, #7903, #8172, #8356, #8377, #8513, #8567.

Summary of changes:

- Add a new page 'Spatial Indexes' that:

  - Describes what they are and how they work at high and lower(ish)
    levels

  - Provides some examples of usage, including how to tune them to
    generate tighter or looser coverings with 'CREATE INDEX', as well as
    exploring potential coverings with 'st_s2covering'.

- Add links to/from 'CREATE INDEX', 'Indexes', and 'Inverted Indexes'
  pages.

- Also, add lots of cross-linking among various spatial pages,
  esp. where indexing is mentioned in the text thereof.


rmloveland:
Hi Becca, Oliver, Sumeer:

Sorry to ping all of you - wasn't sure if it would be enough to get just one person's eyes - I think this doc could benefit from each of your perspectives, since it's kind of trying to be the all-singing all-dancing spatial indexes page (for now).

In addition to pointing out any errors, would appreciate suggestions on how to make this better so it's more helpful to users trying to understand stuff.

Thanks!

otan:
___
*[v20.2/spatial-indexes.md, line 130 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDPdB60V_hd6OTOEIP:-MJDPdB60V_hd6OTOEIQ:bo9nhzb) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L130)):*
> ```Markdown
> | `s2_max_level`   | 30                                            | The maximum level of S2 cell used in the covering. Allowed values: 1-30. Setting it to less than the default means that CockroachDB will be forced to generate coverings using larger cells.                                                |
> | `s2_max_cells`   | 4                                             | The maximum number of S2 cells used in the covering. Provides a limit on how much work is done exploring the possible coverings. Allowed values: 1-30. You may want to use higher values for odd-shaped regions such as skinny rectangles. |
> | `geometry_min_x` | Derived from SRID bounds, else `-(1 << 31)`   | Minimum X-value of the [spatial reference system](spatial-glossary.html#spatial-reference-system) for the object(s) being covered. This only needs to be set if you are using a custom [SRID](spatial-glossary.html#srid).                  |
> ```

This only needs to be set if the default bounds of the SRID is too large/small for the given data, or SRID = 0 and you wish to use a smaller range (unfortunately this is currently not exposed, but is viewable on https://epsg.io/3857).
By default, SRID = 0 assumes [-min int32, max int32] ranges.

otan (commented):
> Reviewed 2 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan, @rmloveland, @rytaft, and @sumeerbhola)

___
*[v20.2/spatial-indexes.md, line 12 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOLTUAeGepTulkVHS:-MJDOLTUAeGepTulkVHT:bryezxb) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L12)):*
> ```Markdown
> 
> - What spatial indexes are
> - How they works
> ```

nit: how they work
___
*[v20.2/spatial-indexes.md, line 27 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOT5tECCPJnZ43l6A:-MJDOT5tECCPJnZ43l6B:belcz9z) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L27)):*
> ```Markdown
> - Fast filtering of lists of shapes based on spatial predicate functions such as `ST_Contains`.
> 
> - Speeding up joins between spatial and non-spatial data.
> ```

not sure how spatial indexes achieve this -- mind elaborating?
___
*[v20.2/spatial-indexes.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOZJt7hoxWnY5RE4S:-MJDOZJt7hoxWnY5RE4T:bpjvmmj) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L33)):*
> ```Markdown
> - Their inner workings are specialized to operate on 2-dimensional (finite) `GEOMETRY` and `GEOGRAPHY` data types. They are stored by CockroachDB as a special type of [inverted index](inverted-indexes.html). For more details, see below.
> 
> - If needed, they can be tuned to store looser or tighter coverings of the shapes being indexed, depending on the needs of your application. Tighter coverings are more expensive to generate and store (and update), but can perform better in some cases because they return fewer false positives during the initial index lookup. However, most users will not need to change the default settings. For more information, see [Tuning spatial indexes](#tuning-spatial-indexes) below.
> ```

they can also perform worse because there's more to inverted index rows, so it's a bit of a tradeoff.
___
*[v20.2/spatial-indexes.md, line 111 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOu0iE0HKs5g0xeed:-MJDOu0iE0HKs5g0xeee:b-ot9mc7) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L111)):*
> ```Markdown
> 
> ~~~
> 'LINESTRING(-76.4955 42.4405,  -75.6608 41.4102,-73.5422 41.052, -73.929 41.707, -76.4955 42.4405)'
> ```

i think you want an SRID=4326 in front of this to ensure ST_S2Covering uses the right bounds.
this will use the default bounds, instead of the lat lng bounds, which may generate different results.

sumeerbhola (changes requested):
Generally looks good. Thanks for putting this together!

> Reviewed 1 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJEiQV370QqvQYN4dl1:b2p9atn)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan, @rmloveland, @rytaft, and @sumeerbhola)

___
*[v20.2/inverted-indexes.md, line 13 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEakMs9sawgpTdq5sr:-MJEakMs9sawgpTdq5ss:b6hov3x) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/inverted-indexes.md#L13)):*
> ```Markdown
> - [JSONB](jsonb.html)
> - [Arrays](array.html)
> - [Spatial data (finite `GEOMETRY` and `GEOGRAPHY` types)](spatial-indexes.html)
> ```

is "finite" something that needs to be said? I don't know if one can even represent an infinite plane in these types, let alone index them. 
___
*[v20.2/spatial-indexes.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEbvZq33eLEZZxNkL1:-MJEbvZq33eLEZZxNkL2:b-qjlxrm) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L31)):*
> ```Markdown
> Spatial indexes differ from other types of indexes as follows:
> 
> - Their inner workings are specialized to operate on 2-dimensional (finite) `GEOMETRY` and `GEOGRAPHY` data types. They are stored by CockroachDB as a special type of [inverted index](inverted-indexes.html). For more details, see below.
> ```

same question about needing to mention "finite".
___
*[v20.2/spatial-indexes.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOZJt7hoxWnY5RE4S:-MJEc3yv3fphAPIibXCG:b-pbrn5j) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L33)):*

they can also perform worse because there's more to inverted index rows, so it's a bit of a tradeoff.
</blockquote></details>

And more scans. We should definitely say there is a tradeoff that can make performance worse.
___
*[v20.2/spatial-indexes.md, line 39 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEcXW7-RnqEIY2ouuF:-MJEcXW7-RnqEIY2ouuG:b-4cb30r) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L39)):*
> ```Markdown
> ### Overview
> 
> There are two main approaches to building geospatial indexes in use:
> ```

nit: the "in use" seems unnecessary in the sentence.
___
*[v20.2/spatial-indexes.md, line 55 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEcthv10wUjLfRY_IO:-MJEcthv10wUjLfRY_IP:b-xpxdjp) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L55)):*
> ```Markdown
> - Allows advanced users to make a per-object tradeoff between index size and false positives during index creation. (See [Tuning spatial indexes](#tuning-spatial-indexes) below.)
> 
> Disadvantages of the "divide the space" approach include:
> ```

Do we typically try to run through the space of design options in our docs? I worry this will just confuse the user -- they just want it to work. We could just say:
Some of the reasons CockroachDB uses the "divide the space" approach are:
and list the reasons above, and cut out the disadvantages section.
(since we are no longer trying to argue for one over the other)
___
*[v20.2/spatial-indexes.md, line 57 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEdNJ45tlHhklxV6IR:-MJEdNJ52NRfONtn4o07:b-n2au6r) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L57)):*
> ```Markdown
> Disadvantages of the "divide the space" approach include:
> 
> - It does not support indexing infinite `GEOMETRY` types. Because the space is divided before it can be indexed, it must be finite. This means that CockroachDB's spatial indexing works for `GEOGRAPHY` (spherical) and for finite `GEOMETRY` (planar) objects, but not for infinite `GEOMETRY`.
> ```

This is confusing. I don't think there are infinite shapes. It is just that for GEOMETRY one needs to know the coordinates that will contain all shapes for optimal indexing. Can we just say something about this in a "how to use geometry indexing section". Something like:
The divide the space approach needs the bounding coordinates that will contain all or most of the shapes. For known SRIDs, CockroachDB automatically derives these bounding coordinates. For others, it uses the default ..., that can be changed. Note that if a rare shape falls outside these bounding coordinates, the database will behave correctly -- there will just be a performance loss in that such shapes may not be eliminated by the index lookup.
___
*[v20.2/spatial-indexes.md, line 58 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEe_q9DTe5u_UDYyvR:-MJEe_q9DTe5u_UDYyvS:b-j3kssd) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L58)):*
> ```Markdown
> 
> - It does not support indexing infinite `GEOMETRY` types. Because the space is divided before it can be indexed, it must be finite. This means that CockroachDB's spatial indexing works for `GEOGRAPHY` (spherical) and for finite `GEOMETRY` (planar) objects, but not for infinite `GEOMETRY`.
> - Includes more false positives in the index by default, which must then be filtered out during SQL execution. CockroachDB mitigates this by providing advanced users with the ability to [tune spatial indexes](#tuning-spatial-indexes).
> ```

We should not say this. There are many cases where coverings can be better than using a bounding box. 
___
*[v20.2/spatial-indexes.md, line 76 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEgJay4fCfyhQqZQ53:-MJEgJay4fCfyhQqZQ54:b-s6yb3s) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L76)):*
> ```Markdown
> S2 Cubed Sphere - 3D
> 
> When you index a spatial object, a covering is computed using some number of the cells in the quadtree. The number of covering cells can vary per indexed object by passing special arguments to `CREATE INDEX` that tell CockroachDB how many levels of s2 cells to use. The leaf nodes of the S2 quadtree are at level 30, and measure 1cm across the Earth's surface. This means that advanced users can tune the accuracy of indexes down to 1cm, depending on the shapes being covered. However, most users will get the best performance with the default settings. For more information, see [Tuning spatial indexes](#tuning-spatial-indexes) below.
> ```

I've suggested some changes below, which say up front that users are getting that 1cm of precision by default for GEOGRAPHY.

When you index a spatial object, a covering is computed using some number of the cells in the quadtree. The number of covering cells can vary per indexed object by passing special arguments to `CREATE INDEX` that tell CockroachDB how many levels of s2 cells to use. The leaf nodes of the S2 quadtree are at level 30, and for GEOGRAPHY measure 1cm across the Earth's surface. By default, GEOGRAPHY indexes use up to level 30, so get this level of precision. We also use S2 cell coverings, in a slightly different way, for GEOMETRY indexes. The precision one gets there is the bounding length of the GEOMETRY index divided by 4^30. For more information, see [Tuning spatial indexes](#tuning-spatial-indexes) below.
___
*[v20.2/spatial-indexes.md, line 78 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEhP9qDcXJZAHWQXNj:-MJEhP9qDcXJZAHWQXNk:b-zcidy8) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L78)):*
> ```Markdown
> When you index a spatial object, a covering is computed using some number of the cells in the quadtree. The number of covering cells can vary per indexed object by passing special arguments to `CREATE INDEX` that tell CockroachDB how many levels of s2 cells to use. The leaf nodes of the S2 quadtree are at level 30, and measure 1cm across the Earth's surface. This means that advanced users can tune the accuracy of indexes down to 1cm, depending on the shapes being covered. However, most users will get the best performance with the default settings. For more information, see [Tuning spatial indexes](#tuning-spatial-indexes) below.
> 
> CockroachDB stores spatial indexes as a special type of [inverted index](inverted-indexes.html). The spatial inverted index maps from a location to one or more shapes whose [coverings](spatial-glossary.html#covering) include that location. Since a location can be used in the covering for multiple shapes, and each shape can have multiple locations in its covering, there is a many-to-many relationship between locations and shapes.
> ```

The example is somewhat confusing since the keys are cells and not some arbitrary shapes. I'm not sure how to show that properly in the example. 
But in the preceding sentence we can say
The spatial inverted index maps from a location, which is a square cell in the quad-tree, to one or more ...

rmloveland (commented):
Thanks Sumeer!  And many thanks to you and Oliver for the reviews! (And to y'all and Becca for the very informative and educational RFC).

I think my latest update has addressed your feedback.  PTAL and let me know!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJS_v3B8f2pFQRFob8z:b95kqrq)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan, @rytaft, and @sumeerbhola)

___
*[v20.2/inverted-indexes.md, line 13 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEakMs9sawgpTdq5sr:-MJSBJF96gFE1qGW3p2M:btaev32) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/inverted-indexes.md#L13)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

is "finite" something that needs to be said? I don't know if one can even represent an infinite plane in these types, let alone index them. 
</blockquote></details>

Probably not - removed it!
___
*[v20.2/spatial-indexes.md, line 12 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOLTUAeGepTulkVHS:-MJSBT849dil4bJHmsKH:bu065vj) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L12)):*

nit: how they work
</blockquote></details>

HA - fixed thanks

(I guess at least it didn't say "how they workses my preciousssssss")
___
*[v20.2/spatial-indexes.md, line 27 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOT5tECCPJnZ43l6A:-MJSBb2P2uDmYUnihOZR:b1l23gw) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L27)):*

not sure how spatial indexes achieve this -- mind elaborating?
</blockquote></details>

Hmmm I think what I meant to write was "Speeding up joins that involve spatial data columns."  Does that sound more reasonable?

Not sure where I got the other idea.  Maybe low blood sugar?
___
*[v20.2/spatial-indexes.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEbvZq33eLEZZxNkL1:-MJSBwkE5EL5eHh71YGf:bss4a1y) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L31)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

same question about needing to mention "finite".
</blockquote></details>

Removed!
___
*[v20.2/spatial-indexes.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOZJt7hoxWnY5RE4S:-MJSCQ75EtkdIv4YLh3f:b-wm4g0o) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L33)):*
Updated to add

> ... Tighter coverings can also lead to worse performance due to more rows in the index, and more scans at the storage layer.  That's why we recommend that most users should not need to change the default settings. For more information, see ...

Is that an accurate representation?
___
*[v20.2/spatial-indexes.md, line 39 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEcXW7-RnqEIY2ouuF:-MJSCdYJ67gJI5gyAxto:bss4a1y) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L39)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

nit: the "in use" seems unnecessary in the sentence.
</blockquote></details>

Removed!
___
*[v20.2/spatial-indexes.md, line 55 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEcthv10wUjLfRY_IO:-MJSCr5Q9ibGXvk9YVii:bj8j86m) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L55)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

Do we typically try to run through the space of design options in our docs? I worry this will just confuse the user -- they just want it to work. We could just say:
Some of the reasons CockroachDB uses the "divide the space" approach are:
and list the reasons above, and cut out the disadvantages section.
(since we are no longer trying to argue for one over the other)

</blockquote></details>

I have updated it as you suggest, and removed the disadvantages part.

Agree we should avoid describing the design space too much.  I think I kept a bit too much of the RFC hanging around in this doc, and need to whittle more of it away.
___
*[v20.2/spatial-indexes.md, line 57 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEdNJ45tlHhklxV6IR:-MJSDlPRAxgqRq3Rxqx0:b-lv0n9s) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L57)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

This is confusing. I don't think there are infinite shapes. It is just that for GEOMETRY one needs to know the coordinates that will contain all shapes for optimal indexing. Can we just say something about this in a "how to use geometry indexing section". Something like:
The divide the space approach needs the bounding coordinates that will contain all or most of the shapes. For known SRIDs, CockroachDB automatically derives these bounding coordinates. For others, it uses the default ..., that can be changed. Note that if a rare shape falls outside these bounding coordinates, the database will behave correctly -- there will just be a performance loss in that such shapes may not be eliminated by the index lookup.
</blockquote></details>

Removed this whole section via your previous comment
___
*[v20.2/spatial-indexes.md, line 58 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEe_q9DTe5u_UDYyvR:-MJSDtr77EbGIlAqkyn4:b-lv0n9s) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L58)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

We should not say this. There are many cases where coverings can be better than using a bounding box. 
</blockquote></details>

Removed this whole section via your previous comment
___
*[v20.2/spatial-indexes.md, line 76 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEgJay4fCfyhQqZQ53:-MJSMP1CE8SMS2GInrhw:blb7et2) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L76)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

I've suggested some changes below, which say up front that users are getting that 1cm of precision by default for GEOGRAPHY.

When you index a spatial object, a covering is computed using some number of the cells in the quadtree. The number of covering cells can vary per indexed object by passing special arguments to `CREATE INDEX` that tell CockroachDB how many levels of s2 cells to use. The leaf nodes of the S2 quadtree are at level 30, and for GEOGRAPHY measure 1cm across the Earth's surface. By default, GEOGRAPHY indexes use up to level 30, so get this level of precision. We also use S2 cell coverings, in a slightly different way, for GEOMETRY indexes. The precision one gets there is the bounding length of the GEOMETRY index divided by 4^30. For more information, see [Tuning spatial indexes](#tuning-spatial-indexes) below.

</blockquote></details>

Thanks! I've updated the section with your text (very lightly edited)
___
*[v20.2/spatial-indexes.md, line 78 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEhP9qDcXJZAHWQXNj:-MJSMlEUAkZh0uM0widp:bcq9pxs) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L78)):*
<details><summary><i>Previously, sumeerbhola wrote…</i></summary><blockquote>

The example is somewhat confusing since the keys are cells and not some arbitrary shapes. I'm not sure how to show that properly in the example. 
But in the preceding sentence we can say
The spatial inverted index maps from a location, which is a square cell in the quad-tree, to one or more ...
</blockquote></details>

I have edited that sentence as you suggest - also removed the example, since it was pretty weak and as you said, confusing since it doesn't really show an accurate picture of what the text is saying.
___
*[v20.2/spatial-indexes.md, line 111 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOu0iE0HKs5g0xeed:-MJSZUmNEvkcWVaUiGZ5:b-9be0xm) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L111)):*

i think you want an SRID=4326 in front of this to ensure ST_S2Covering uses the right bounds.
this will use the default bounds, instead of the lat lng bounds, which may generate different results.
</blockquote></details>

Added, thanks!  Ran my image generator script using  the updated `'SRID=4326;LINESTRING(-76.4955 42.4405,  -75.6608 41.4102,-73.5422 41.052, -73.929 41.707, -76.4955 42.4405)'` 

The resulting coverings and images were unchanged AFAICT.
___
*[v20.2/spatial-indexes.md, line 130 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDPdB60V_hd6OTOEIP:-MJSAV5n7jpKlPSISfTp:b-kfpor0) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L130)):*
> This only needs to be set if the default bounds of the SRID is too large/small for the given data, or SRID = 0 and you wish to use a smaller range (unfortunately this is currently not exposed, but is viewable on https://epsg.io/3857).
> By default, SRID = 0 assumes [-min int32, max int32] ranges.

Thanks, updated the `geometry_min_x` description with this information.

rytaft (commented):
Looks great!

> Reviewed 18 of 20 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJSpLFr1v1uQRbRxNAQ:b9bkkp1)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan, @rmloveland, and @sumeerbhola)

___
*[v20.2/spatial-glossary.md, line 177 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSeZqWDnOoViDmdb77:-MJSeZqX1LezPWEw4zRZ:bw8x1tz) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-glossary.md#L177)):*
> ```Markdown
> <a name="s2"></a>
> 
> - _S2_: The [S2 Geometry Library](http://s2geometry.io) is a C++ code library for doing spherical geometry.  It models a sphere using a [quadtree](https://en.wikipedia.org/wiki/Quadtree) "divide the space" approach, and is used by CockroachDB.
> ```

[nit] Instead of "doing spherical geometry", maybe something like "performing spherical geometry computations"?
___
*[v20.2/spatial-indexes.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSjd9iC3E069Uw9CIr:-MJSjd9iC3E069Uw9CIs:bedg5sf) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L27)):*
> ```Markdown
> - Fast filtering of lists of shapes based on spatial predicate functions such as `ST_Contains`.
> 
> - Speeding up joins that involve spatial data columns.
> ```

It might help to include some example queries that can take advantage of a spatial index (or at least a link to another page with some examples). Let me know if you want some ideas.
___
*[v20.2/spatial-indexes.md, line 41 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSfd8J66muZC2u0OFf:-MJSfd8J66muZC2u0OFg:b-pmlo4q) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L41)):*
> ```Markdown
> There are two main approaches to building geospatial indexes:
> 
> - One approach is to "divide the objects". This works by inserting the objects into a tree (usually an [R-tree](https://en.wikipedia.org/wiki/R-tree)) whose shape depends on the data being indexed.
> ```

Perhaps mention that this is the approach that PostGIS uses
___
*[v20.2/spatial-indexes.md, line 106 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSiHcv0BMybka-JIVt:-MJSiHcv0BMybka-JIVu:bhjwfsj) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L106)):*
> ```Markdown
> Animated GIF of S2 Coverings - Levels 1 to 30
> 
> Below are the same images presented in a grid. You can see that as we turn up the `s2_max_cells` parameter, more work is done by CockroachDB to discover a tighter and tighter covering (that is, a covering using fewer and smaller cells). The covering for this particular shape reaches a reasonable level of accuracy when `s2_max_cells` reaches 10, and stops improving much past 12.
> ```

should "fewer and smaller cells" actually be "more and smaller cells"?

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 20 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJSrgasDn7SumgImG_6:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @otan, @rmloveland, and @sumeerbhola)

___
*[v20.2/spatial-indexes.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJDOZJt7hoxWnY5RE4S:-MJSpiqKB8usJUfJuzfq:b-47fafm) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L33)):*

Updated to add

> ... Tighter coverings can also lead to worse performance due to more rows in the index, and more scans at the storage layer.  That's why we recommend that most users should not need to change the default settings. For more information, see ...

Is that an accurate representation?
</blockquote></details>

looks good
___
*[v20.2/spatial-indexes.md, line 57 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEdNJ45tlHhklxV6IR:-MJSqk9bEx0bQ6605VOL:b-utz3gm) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L57)):*
> The divide the space approach needs the bounding coordinates that will contain all or most of the shapes. For known SRIDs, CockroachDB automatically derives these bounding coordinates. For others, it uses the default ..., that can be changed. Note that if a rare shape falls outside these bounding coordinates, the database will behave correctly -- there will just be a performance loss in that such shapes may not be eliminated by the index lookup.

The content in the sentence that starts with "Note ..." should be conveyed somewhere. Maybe where the doc describes the `geometry_min_x,...` parameters?

rmloveland (commented):
Thanks for the review Becca! And for the additional comments, Sumeer.

I think I have addressed the latest round of comments

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJTFIApCUqZDg3Ca_nb:bzohhk)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @otan, @rytaft, and @sumeerbhola)

___
*[v20.2/spatial-glossary.md, line 177 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSeZqWDnOoViDmdb77:-MJTCAdQ1JSYkTAYM24j:bu1labn) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-glossary.md#L177)):*

[nit] Instead of "doing spherical geometry", maybe something like "performing spherical geometry computations"?
</blockquote></details>

Much better - updated, thanks
___
*[v20.2/spatial-indexes.md, line 57 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEdNJ45tlHhklxV6IR:-MJTEnvHBcs9itRGoMgd:bgp74e5) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L57)):*
Oh nice catch!  I have added it to a note just above the s2 settings params that reads 

> If a shape falls outside the bounding coordinates determined by the `geometry_*` settings below, there will be a performance loss in that such shapes may not be eliminated by the index lookup, but the database will still return the correct answers.

Please let me know if I have introduced a subtle error in the re-wording, though
___
*[v20.2/spatial-indexes.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSjd9iC3E069Uw9CIr:-MJTCHni4INZgcLQnfE7:b-qo9rpq) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L27)):*
I agree, it would be very helpful!  We could add a few to the 'Examples' section on this page, and also link to them from the Joins page

If it's OK with you, I have filed a separate issue #8638 to work on adding more join info to this page separately.  Figure it will be better to get this published ASAP and then improve it since I have not played with spatial joins much at all yet.

> Let me know if you want some ideas.

Yes please!  Do you mind commenting on #8638  with some suggestions and ideas?  I would really appreciate it.
___
*[v20.2/spatial-indexes.md, line 41 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSfd8J66muZC2u0OFf:-MJTDPK37b-NjmFj4Z-p:bgrfpw0) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L41)):*

Perhaps mention that this is the approach that PostGIS uses
</blockquote></details>

Updated to add "This is the approach taken by [PostGIS](https://postgis.net)."
___
*[v20.2/spatial-indexes.md, line 106 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSiHcv0BMybka-JIVt:-MJTFBzS3U_na8WgVjsU:bsk0n5e) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L106)):*

should "fewer and smaller cells" actually be "more and smaller cells"?
</blockquote></details>

Fixed!  :sweat:

rytaft (approved):
:lgtm: Thanks!

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJTM13-DkOdmzAhHpjp:b-g2a5fb)* status: :shipit: complete! 1 of 0 LGTMs obtained (and 1 stale) (waiting on @otan, @rmloveland, and @sumeerbhola)

___
*[v20.2/spatial-indexes.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJSjd9iC3E069Uw9CIr:-MJTLXI61jNNvamK__he:behqle0) ([raw file](https://github.com/cockroachdb/docs/blob/5494887fb45bf35eb62d0ec71da0868b9a550733/v20.2/spatial-indexes.md#L27)):*

I agree, it would be very helpful!  We could add a few to the 'Examples' section on this page, and also link to them from the Joins page

If it's OK with you, I have filed a separate issue #8638 to work on adding more join info to this page separately.  Figure it will be better to get this published ASAP and then improve it since I have not played with spatial joins much at all yet.

> Let me know if you want some ideas.

Yes please!  Do you mind commenting on #8638  with some suggestions and ideas?  I would really appreciate it.
</blockquote></details>

Sounds good! Will do.

ericharmeling (commented):
Awesome work, Rich! The spatial indexes page is rad, (especially the details part). Very impressive, technical docs.

I had a few comments, some grammar-related, some include-file-related, some syntax-reference-related, which might demand some not-so-small changes, should you find my points valid and worth addressing...

Also, (I don't note this in my comments below), you might need to update the index_def.html diagram and the corresponding parameter description on the create-table.html page.

> Reviewed 7 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJXv9by7uC4uUp262TG:b-i5vxf0)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 2 stale) (waiting on @otan, @rmloveland, and @sumeerbhola)

___
*[_includes/v20.2/sql/spatial-support-new.md, line 1 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXpQ3t6c1lYF0Sr-9s:-MJXpQ3u3o8HBe5MP3dt:b-q3bf9o) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/_includes/v20.2/sql/spatial-support-new.md#L1)):*
> ```Markdown
> <span class="version-tag">New in v20.2</span>: CockroachDB supports efficiently storing and querying spatial data.
> ```

TBH, IDK if this needs to be an include file, on every page that covers a spatial feature. This seems more like an overview-page-type sentence (you actually have this sentence on spatial-features.html, but not in an include). It might be best to remove it from all pages except that overview page, and then add a version tag in its place.
___
*[v20.2/create-index.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXvkArEjldaEsp26kC:-MJXvkAs7foN39XCQPHh:bq59bzs) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/create-index.md#L31)):*
> ```
> <section>{% include {{ page.version.version }}/sql/diagrams/create_index.html %}</section>
> ```

I think spatial index syntax is covered by this diagram (`USING name`, right?), but it might still be worth regenerating it to make sure?
___
*[v20.2/create-index.md, line 46 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXvf2LDpFuBoKnHdn3:-MJXvf2LDpFuBoKnHdn4:b4dg9r5) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/create-index.md#L46)):*
> ```
> `USING name` 
> ```

This parameter description might need to be updated for spatial index syntax.
___
*[v20.2/indexes.md, line 70 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXq1Ud2REZ-vDP1Q4I:-MJXq1Ud2REZ-vDP1Q4J:bam7z78) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/indexes.md#L70)):*
> ```
> ## Spatial indexes
> ```

Inverted and partial indexes are special types of indexes that also have their own pages, but we don't call them out in their own level-2 header like this (we don't even call out inverted indexes at all on this page). It might be more consistent to have a short paragraph under "How do indexes work?", like we do for partial indexes. 

> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> For more information about spatial indexes, see [Spatial Indexes](spatial-indexes.html).

This include says nothing about spatial indexes, and I don't think it adds much information. I'd remove it, and add a sentence more specific to spatial indexes.

My suggestion (end of top-level of "How do indexes work?" section):

```
<span class="version-tag">New in v20.2</span>: To index [spatial data](spatial-data.html), CockroachDB uses *spatial indexes*. For more information about spatial indexes, see [Spatial Indexes](spatial-indexes.html).
```

Another idea would be to add this as a level 2 header to the inverted-indexes.html page, since spatial indexes are, in fact, inverted indexes.
___
*[v20.2/install-cockroachdb-spatial.md, line 10 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY4yNH6w2jBqq68756:-MJY4yNH6w2jBqq68757:b5tfz76) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/install-cockroachdb-spatial.md#L10)):*
> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Idk if this needs to be included here.
___
*[v20.2/spatial-data.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY599DEaaTXoSpecte:-MJY599DEaaTXoSpectf:bxfcq6k) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-data.md#L7)):*
> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Again, IDK how much this include adds to the page.
___
*[v20.2/spatial-data.md, line 9 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY53tJ42C1rP_Yyaot:-MJY53tJ42C1rP_Yyaou:b-97plgt) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-data.md#L9)):*
> ```
> spatial data types
> ```

[spatial](spatial-features.html) data types
___
*[v20.2/spatial-features.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXugiK6zGUQHwAXDOZ:-MJXugiL9rMgZv9boyDd:b-lgv6w0) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-features.md#L7)):*
> ```
> <span class="version-tag">New in v20.2</span>: CockroachDB supports efficiently storing and querying spatial data.
> ```

This seems like the best place for this kind of sentence.
___
*[v20.2/spatial-glossary.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXv4jp7zIk_QiGd31Z:-MJXv4jq1QlfUtiAcnRf:bralhll) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-glossary.md#L31)):*
> ```
> lie to the outside of
> ```

lie outside of
___
*[v20.2/spatial-indexes.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY5MA0CZde89qGGBo3:-MJY5MA0CZde89qGGBo4:b-buh2rw) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L7)):*
> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Again, IDK how much this include adds to this doc.
___
*[v20.2/spatial-indexes.md, line 9 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwX7l2EA5gm42WnX_:-MJXwX7l2EA5gm42WnXa:b-g71ncs) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L9)):*
> ```
> spatial data
> ```

`[spatial data](spatial-data.html)`
___
*[v20.2/spatial-indexes.md, line 21 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXw6te49T9dWHlTW-7:-MJXw6tf5oDsrxgwMczg:b-rjkflu) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L21)):*
> ```
> Its purpose in life is to 
> ```

Its purpose is to
___
*[v20.2/spatial-indexes.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwo_W6NzqyEyGmisp:-MJXwo_W6NzqyEyGmisq:b-5biqrt) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L31)):*
> ```
> Their inner workings are 
> ```

They are
___
*[v20.2/spatial-indexes.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwsnZBR9ASeqNuQcc:-MJXwsn_7u_LAy_KDDfv:b-vxdkex) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L31)):*
> ```
> For more details, see below.
> ```

For more details, see [below](#details).
___
*[v20.2/spatial-indexes.md, line 33 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXx3T73LwP9Wh65o8t:-MJXx3T73LwP9Wh65o8u:b-2muqjp) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L33)):*
> ```
> generate and store (and update)
> ```

generate, store, and update
___
*[v20.2/spatial-indexes.md, line 41 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxNe5DVkHkyH53-XK:-MJXxNe5DVkHkyH53-XL:bsgpt7i) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L41)):*
> ```
> "divide the objects". It works by inserting
> ```

"divide the objects" by inserting
___
*[v20.2/spatial-indexes.md, line 43 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxTaxEW6lteUIpRah:-MJXxTaxEW6lteUIpRai:b-n3bvwv) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L43)):*
> ```
> "divide the space" by creating
> ```
___
*[v20.2/spatial-indexes.md, line 50 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxoTJ0ej19yeyCb6d:-MJXxoTK-A-aIijiVAjJ:b-odnzyc) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L50)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
> Some of the reasons CockroachDB uses the "divide the space" approach are:
> - Easy to scale horizontally.
> - No balancing operations are required, unlike [R-tree indexes](https://en.wikipedia.org/wiki/R-tree).
> - Inserts require no locking.
> - Bulk ingest is simpler to implement than other approaches.
> - Allows advanced users to make a per-object tradeoff between index size and false positives during index creation. (See [Tuning spatial indexes](#tuning-spatial-indexes) below.)
> ```
</details>


CockroachDB uses the "divide the space" approach for the following reasons:
- It's easy to scale horizontally.
- It requires no balancing operations, unlike [R-tree indexes](https://en.wikipedia.org/wiki/R-tree).
- Inserts under this approach require no locking
- Bulk ingest under this approach is simpler to implement than under other approaches.
- It allows advanced users to make a per-object tradeoff between index size and false positives during index creation. (See [Tuning spatial indexes](#tuning-spatial-indexes) below.)
___
*[v20.2/spatial-indexes.md, line 56 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY-sKk5ic2ZeP2grav:-MJY-sKk5ic2ZeP2graw:b-irxy6r) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L56)):*
> ```Markdown
> 
> Whichever approach to indexing is used, when an object is indexed, a "covering" shape (i.e., a bounding box) is constructed that completely encompasses the indexed object. Index queries work by looking for containment or intersection between the covering shape for the query object and the indexed covering shapes. This retrieves false positives but no false negatives.
> ```

I might be missing something... but I really think we need some syntax reference documentation for creating spatial indexes. 

The only place I see anything re: full syntax for creating spatial indexes is in the "Performance" section of spatial-data.html and in the examples at the end of this page, which isn't how we document the syntax for most SQL features. It needn't be a full on diagram and parameters, but I think it would be helpful to just give the basic syntax overview.

Perhaps we should add a "Creating" level 3 header here before going into the "Details" section, that clears up a couple things that were not really explicitly stated, and that I really didn't quite understand after reading this page and the references to spatial indexes in the other pages (including `Indexes` and `CREATE INDEX`):

- Spatial indexes with `CREATE INDEX` require a `USING ??` clause, with the object data type option. (Is this accurate?)
- Spatial indexes are explicitly created as inverted indexes in `CREATE TABLE` statements, without a need for a `USING` clause. I assume that any time you create an inverted index on a geom or geog object, the index is spatial? 
- Even though spatial indexes are inverted indexes, you don't use `CREATE INVERTED INDEX` to create a spatial index, and when you create a spatial index with `CREATE INDEX ... USING` with a specific subset of options after `USING`, CRDB creates an inverted, spatial index.

The fact that geospatial indexes are inverted indexes makes the syntax pretty confusing, tbh.
___
*[v20.2/spatial-indexes.md, line 89 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXzHD18pDMw1eFkuh6:-MJXzHD206Vqub0JBnFH:b-1l27kf) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L89)):*
> ```
> 'st_s2covering'
> ```

`st_s2covering`
___
*[v20.2/spatial-indexes.md, line 232 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY-RrEDhrNjr_U2gqC:-MJY-RrEDhrNjr_U2gqD:bpf4i7y) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L232)):*
> ```
> Create a `GEOGRAPHY` index
> ```

## Create a spatial index on a `GEOGRAPHY` object

The example below shows how to create a spatial index on a `GEOGRAPHY` object using the default settings:

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 2 stale) (waiting on @otan, @rmloveland, and @sumeerbhola)

___
*[v20.2/spatial-indexes.md, line 57 at r1](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJEdNJ45tlHhklxV6IR:-MJYG4SYBzFsss3Gh9lI:b-47fafm) ([raw file](https://github.com/cockroachdb/docs/blob/1a563a0b0472b665b4befce5134e696cf57f3f76/v20.2/spatial-indexes.md#L57)):*

Oh nice catch!  I have added it to a note just above the s2 settings params that reads 

> If a shape falls outside the bounding coordinates determined by the `geometry_*` settings below, there will be a performance loss in that such shapes may not be eliminated by the index lookup, but the database will still return the correct answers.

Please let me know if I have introduced a subtle error in the re-wording, though
</blockquote></details>

looks good

rmloveland (commented):
Thanks for the review Eric!!!  I really appreciate your feedback.  I have tried to address everything you mentioned, including needing add'l syntax documentation for `CREATE INDEX`, and checking the diagrams.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MJd0zE20vgK2rZwH4xn:b-xyh129)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 2 stale) (waiting on @ericharmeling, @otan, @rmloveland, and @sumeerbhola)

___
*[_includes/v20.2/sql/spatial-support-new.md, line 1 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXpQ3t6c1lYF0Sr-9s:-MJcYex-7CiVMXNT09sY:bf8tezk) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/_includes/v20.2/sql/spatial-support-new.md#L1)):*

TBH, IDK if this needs to be an include file, on every page that covers a spatial feature. This seems more like an overview-page-type sentence (you actually have this sentence on spatial-features.html, but not in an include). It might be best to remove it from all pages except that overview page, and then add a version tag in its place.
</blockquote></details>

Makes sense to me.  I have deleted the include and removed all mentions of this line except from the overview page 'Spatial Features' and from the 'Install CockroachDB spatial' page.
___
*[v20.2/create-index.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXvkArEjldaEsp26kC:-MJci7Za02BSfyrtivoH:b3zzivw) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/create-index.md#L31)):*

> ```
> <section>{% include {{ page.version.version }}/sql/diagrams/create_index.html %}</section>
> ```

I think spatial index syntax is covered by this diagram (`USING name`, right?), but it might still be worth regenerating it to make sure?
</blockquote></details>

I regenerated the diagram for `CREATE INDEX` but according to `git status` the file contents of the diagram (for v20.2) were not changed, so I think you are correct that it is covered by `USING (name)`.
___
*[v20.2/create-index.md, line 46 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXvf2LDpFuBoKnHdn3:-MJccLeR3vP0oX-Y69IB:b-x8b36d) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/create-index.md#L46)):*

> ```
> `USING name` 
> ```

This parameter description might need to be updated for spatial index syntax.
</blockquote></details>

Nice catch, thanks - updated to add `gist` with a link to the spatial index page.
___
*[v20.2/indexes.md, line 70 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXq1Ud2REZ-vDP1Q4I:-MJcdE1r5U5Q9yAjyNPl:bwy8kbr) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/indexes.md#L70)):*

> ```
> ## Spatial indexes
> ```

Inverted and partial indexes are special types of indexes that also have their own pages, but we don't call them out in their own level-2 header like this (we don't even call out inverted indexes at all on this page). It might be more consistent to have a short paragraph under "How do indexes work?", like we do for partial indexes. 

> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> For more information about spatial indexes, see [Spatial Indexes](spatial-indexes.html).

This include says nothing about spatial indexes, and I don't think it adds much information. I'd remove it, and add a sentence more specific to spatial indexes.

My suggestion (end of top-level of "How do indexes work?" section):

```
<span class="version-tag">New in v20.2</span>: To index [spatial data](spatial-data.html), CockroachDB uses *spatial indexes*. For more information about spatial indexes, see [Spatial Indexes](spatial-indexes.html).
```

Another idea would be to add this as a level 2 header to the inverted-indexes.html page, since spatial indexes are, in fact, inverted indexes.
</blockquote></details>

Went with your suggested sentence at the end of the "How do indexes work?" section.

Agree re: adding more info to the 'Inverted Indexes' page.  It's in the plan to add more spatial index info to both 'Indexes' and 'Inverted Indexes' Real Soon (tm) (via #7902 and #7903)
___
*[v20.2/install-cockroachdb-spatial.md, line 10 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY4yNH6w2jBqq68756:-MJcfEnCEg36Iwb343K1:bss4a2b) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/install-cockroachdb-spatial.md#L10)):*

> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Idk if this needs to be included here.
</blockquote></details>

Removed.
___
*[v20.2/spatial-data.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY599DEaaTXoSpecte:-MJcfeos4Z_Z8P454Pqi:btr8juj) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-data.md#L7)):*

> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Again, IDK how much this include adds to the page.
</blockquote></details>

Removed this from all pages except the overview page 'Spatial Features'.
___
*[v20.2/spatial-data.md, line 9 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY53tJ42C1rP_Yyaot:-MJcfrGy9vuTugHlgd4T:btnidkz) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-data.md#L9)):*

> ```
> spatial data types
> ```

[spatial](spatial-features.html) data types
</blockquote></details>

Updated to add the link
___
*[v20.2/spatial-features.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXugiK6zGUQHwAXDOZ:-MJcfu9h5wek87E2gr4J:b-980o99) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-features.md#L7)):*

> ```
> <span class="version-tag">New in v20.2</span>: CockroachDB supports efficiently storing and querying spatial data.
> ```

This seems like the best place for this kind of sentence.
</blockquote></details>

Cool - left this one alone
___
*[v20.2/spatial-glossary.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXv4jp7zIk_QiGd31Z:-MJcg50g9K1jE-M7S9aE:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-glossary.md#L31)):*

> ```
> lie to the outside of
> ```

lie outside of
</blockquote></details>

Fixed
___
*[v20.2/spatial-indexes.md, line 7 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY5MA0CZde89qGGBo3:-MJcg6bl3lDU-qQjCeoW:bss4a1y) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L7)):*

> ```
> {% include {{page.version.version}}/sql/spatial-support-new.md %}
> ```

Again, IDK how much this include adds to this doc.
</blockquote></details>

Removed!
___
*[v20.2/spatial-indexes.md, line 9 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwX7l2EA5gm42WnX_:-MJcgFQvBjFnt60LzhVq:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L9)):*

> ```
> spatial data
> ```

`[spatial data](spatial-data.html)`
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 21 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXw6te49T9dWHlTW-7:-MJcgLENEG4YYjUSBjLq:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L21)):*

> ```
> Its purpose in life is to 
> ```

Its purpose is to
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwo_W6NzqyEyGmisp:-MJcgRbe5kKxbCXDL4Q2:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L31)):*

> ```
> Their inner workings are 
> ```

They are
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 31 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXwsnZBR9ASeqNuQcc:-MJcgSIrBTkCDNgiITNu:b-896fja) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L31)):*

> ```
> For more details, see below.
> ```

For more details, see [below](#details).
</blockquote></details>

Done!
___
*[v20.2/spatial-indexes.md, line 33 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXx3T73LwP9Wh65o8t:-MJcgf3d8-5DbhbkAass:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L33)):*

> ```
> generate and store (and update)
> ```

generate, store, and update
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 41 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxNe5DVkHkyH53-XK:-MJcglclBsBwCKCaICLv:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L41)):*

> ```
> "divide the objects". It works by inserting
> ```

"divide the objects" by inserting
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 43 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxTaxEW6lteUIpRah:-MJcgmPhBdY6GZv1-Khb:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L43)):*

> ```
> "divide the space" by creating
> ```

</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 50 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXxoTJ0ej19yeyCb6d:-MJch1Pb5gHr3Y88nsMc:babgvi2) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L50)):*

> ```
> Some of the reasons CockroachDB uses the "divide the space" approach are:
> - Easy to scale horizontally.
> - No balancing operations are required, unlike [R-tree indexes](https://en.wikipedia.org/wiki/R-tree).
> - Inserts require no locking.
> - Bulk ingest is simpler to implement than other approaches.
> - Allows advanced users to make a per-object tradeoff between index size and false positives during index creation. (See [Tuning spatial indexes](#tuning-spatial-indexes) below.)
> ```

CockroachDB uses the "divide the space" approach for the following reasons:
- It's easy to scale horizontally.
- It requires no balancing operations, unlike [R-tree indexes](https://en.wikipedia.org/wiki/R-tree).
- Inserts under this approach require no locking
- Bulk ingest under this approach is simpler to implement than under other approaches.
- It allows advanced users to make a per-object tradeoff between index size and false positives during index creation. (See [Tuning spatial indexes](#tuning-spatial-indexes) below.)

</blockquote></details>

Better - thanks!
___
*[v20.2/spatial-indexes.md, line 56 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY-sKk5ic2ZeP2grav:-MJcjCBQCVSHOo3K1uxB:b-mto4ts) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L56)):*
tl;dr: I have added a new section to the `CREATE INDEX` page called "Creating spatial indexes" that tries to meet some of this need.  PTAL!  
(Also I think `CREATE INVERTED INDEX ...` is a CockroachDB invention that is not in PG, and IMO we should not document it.)

> anything re: full syntax for creating spatial indexes is in the "Performance" section of spatial-data.html and in the examples at the end of this page, which isn't how we document the syntax for most SQL features. It needn't be a full on diagram and parameters, but I think it would be helpful to just give the basic syntax overview.

I agree - I hope the new section on `CREATE INDEX` helps to address this.  I tried to note there that what we show in the examples is what is supported.

> Spatial indexes with CREATE INDEX require a USING ?? clause, with the object data type option. (Is this accurate?)

Yes.

> Spatial indexes are explicitly created as inverted indexes in CREATE TABLE statements, without a need for a USING clause. I assume that any time you create an inverted index on a geom or geog object, the index is spatial?

You def need a `USING` clause.  AIUI only the syntax we show in the examples is supported.  There is no optional `CREATE INVERTED INDEX` form (edit: or maybe there is but I don't think we should doc it -- see below), and you must pass the `USING GIST(geom)` part.  I have added some language to this effect on the update to `CREATE INDEX`.  PTAL.

> Even though spatial indexes are inverted indexes, you don't use CREATE INVERTED INDEX to create a spatial index, and when you create a spatial index with CREATE INDEX ... USING with a specific subset of options after USING, CRDB creates an inverted, spatial index.

Yes.  Unfortunately I think the confusion is due to the fact that (AFAICT?) the `CREATE INVERTED INDEX` syntax is a "CockroachDBism" that is not in PG (based on looking at https://www.postgresql.org/docs/9.1/sql-createindex.html and https://www.postgresql.org/docs/current/textsearch-indexes.html).  I am not able to find anything in the PG docs except `CREATE INDEX foo ON bar USING GIN/GIST(mycolumn)`.  Nor in any of the other DBs with a quick Bing (ha) either.

> The fact that geospatial indexes are inverted indexes makes the syntax pretty confusing, tbh.

I agree.  Sorry for the confusion, I hope it is somewhat addressed (?) by my small update to `CREATE INDEX`. 

PS FWIW IMO we should rip out `CREATE INVERTED INDEX` if it is not in PG.  At least we should consider removing it from the docs.  However I am not a SQL expert, so maybe my understanding is wrong - I have filed a docs issue #8668  to discuss

PPS Actually it looks like maybe (?) we also support the `CREATE INVERTED INDEX ...` on these columns according to this PR, however I'd really rather not document it if it is our own thing that is not also in PG: https://github.com/cockroachdb/cockroach/pull/51516/files
___
*[v20.2/spatial-indexes.md, line 89 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJXzHD18pDMw1eFkuh6:-MJch95_9Xr-ogPLn5_a:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L89)):*

> ```
> 'st_s2covering'
> ```

`st_s2covering`
</blockquote></details>

Done
___
*[v20.2/spatial-indexes.md, line 232 at r4](https://reviewable.io/reviews/cockroachdb/docs/8617#-MJY-RrEDhrNjr_U2gqC:-MJci2iZ5R5gEK9uGttx:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/0da33b6e8b6c5b1c687e43d0c9a3e9738a60f269/v20.2/spatial-indexes.md#L232)):*

> ```
> Create a `GEOGRAPHY` index
> ```

## Create a spatial index on a `GEOGRAPHY` object

The example below shows how to create a spatial index on a `GEOGRAPHY` object using the default settings:
</blockquote></details>

Done

rmloveland:
Hey Eric, thanks again for your careful review of this PR.  Sorry that I seem to have missed some obvious things.  What do you think of the recent changes?  Do you feel like they address your feedback?  If not please let me know and I will try to fix.

(Also, sorry for forgetting to mention that I also ran the scripts to update the diagrams for `create_table` and `index_def` as you suggested in your opening comment.  It looks like no changes were generated for the 20.2 pages, so it seems they are up to date as well.)

ericharmeling (approved):
@rmloveland So sorry for not getting back on this sooner!

Thanks for carefully going through my nit-picking and lengthy review comments. :)

I just looked over all your follow-up comments and changes and everything LGTM now!

> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8617#-:-MK1fcO2CG16YtQHWJDi:bqa4qhk)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 2 stale) (waiting on @otan, @rmloveland, and @sumeerbhola)

PR #8622, opened 2020-10-10
Remove AWS cloudformation test deployment doc

### LINKED ISSUE #3870: Deploy a Test Cluster Doc Update - New workloads and K8s settings
Re: [Deploy a Test Cluster](https://cockroachlabs.com/docs//v2.0/deploy-a-test-cluster.html)

## Issue Description
Need to update docs page with changes to our cloudformation template.

## Suggested Resolution 
Updates
we now have TPC-C 10 and 100 workloads
image

You can move the k8s capacity independently of the cockroachdb capacity
image

The recommendation for the highest performance is to have an extra k8s node available so the workload generators don't contend for resources.

jseldess:
We're now encouraging uses to leverage CC for test deployments.

Closes #3870


jseldess:
Got LGTM from Nate in Slack.

PR #8633, opened 2020-10-12
FIX-8236 Provided example to specify columns with insert

### LINKED ISSUE #8236: SQL ref: provide example on how to specify columns with INSERT
https://www.cockroachlabs.com/docs/stable/insert.html#examples

In the examples for the INSERT command, it shows how to insert when providing all values in column order, like this:

```
INSERT INTO users VALUES
    ('c28f5c28-f5c2-4000-8000-000000000026', 'new york', 'Petee', '101 5th Ave', '1234567890');
```

But it doesn't show how you would insert a row specifying particular columns, like this:

```
INSERT INTO users (id, name) VALUES
    ('c28f5c28-f5c2-4000-8000-000000000026', 'Petee');
```

This would be helpful. 

The docs hint that the first example does specify columns, even though it doesn't:

> If you do not list column names, the statement will use the columns of the table in their declared order:

This implies to me that the section above would have shown column names, and the one below would specify without, but in fact the examples are both are inserting without column names.

pratikSethi:
Fixes #8236

Example provided for v19.2, v20.1, v20.2

### DIFF
--- a/v19.2/insert.md
+++ b/v19.2/insert.md
@@ -77,7 +77,7 @@ key. Using `ON CONFLICT` is therefore more flexible.
 
 {% include copy-clipboard.html %}
 ~~~ sql
-> INSERT INTO users VALUES
+> INSERT INTO users (id, city, name, address, credit_card) VALUES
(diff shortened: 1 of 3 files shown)


ericharmeling (approved):
LGTM. Thanks for the contribution, @pratikSethi !

PR #8634, opened 2020-10-12
Update SHOW RANGE FOR ROW to take indexed columns

### LINKED ISSUE #7821: sql: update the `SHOW RANGE FOR ROW` command to just take index cols
PR: https://github.com/cockroachdb/cockroach/pull/50479

From release notes:
> The [`SHOW RANGE FOR ROW`](../v20.2/show-range-for-row.html) statement now takes a tuple of the row's index columns instead of the full column set of the row. [#50479][#50479] {% comment %}doc{% endcomment %}

jseldess:
Fixes #7821.

### DIFF
--- a/v20.2/alter-database.md
+++ b/v20.2/alter-database.md
@@ -13,7 +13,7 @@ The `ALTER DATABASE` [statement](sql-statements.html) applies a schema change to
 Subcommand | Description
 -----------|------------
 [`CONFIGURE ZONE`](configure-zone.html) | [Configure replication zones](configure-replication-zones.html) for a database.
-[`CONVERT TO SCHEMA`](convert-to-schema.html) | <span class="version-tag">New in v20.2</span>: Convert a [database](create-database.html) to a [schema](sql-name-resolution.html#logical-schemas-and-namespaces).
+[`CONVERT TO SCHEMA`](convert-to-schema.html) | <span class="version-tag">New in v20.2</span>: Convert a [database](create-database.html) to a [schema](sql-name-resolution.html).
 [`RENAME`](rename-database.html) | Change the name of a database.
 
 ## Viewing schema changes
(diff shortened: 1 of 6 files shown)


rohany:
Looks good to me, but @solongordon should give it the final stamp :)

solongordon (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8634)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @jseldess, @rohany, and @solongordon)

___
*[v20.2/show-range-for-row.md, line 19 at r1](https://reviewable.io/reviews/cockroachdb/docs/8634#-MJSIh8I15Yp7Z2G_O6l:-MJSIh8I15Yp7Z2G_O6m:b-uc23sj) ([raw file](https://github.com/cockroachdb/docs/blob/ab4dd4fbe61f96ddec5d4db89db0ee195a4f71ee/v20.2/show-range-for-row.md#L19)):*
> ```Markdown
> ~~~
> SHOW RANGE FROM TABLE <tablename> FOR ROW (row, value, ...)
> SHOW RANGE FROM INDEX [ <tablename> @ ] <indexname> FOR ROW (row, value, ...)
> ```

Nit: I find the `(row, value, ...)` terminology kind of confusing. It seems to imply that the first element is a row and the second one is a value. Perhaps better to write something like `(value1, value2, ...)`?

jseldess:
___
*[v20.2/show-range-for-row.md, line 19 at r1](https://reviewable.io/reviews/cockroachdb/docs/8634#-MJSIh8I15Yp7Z2G_O6l:-MJULSqr9bemwOO9NfSG:b-z3o9n) ([raw file](https://github.com/cockroachdb/docs/blob/ab4dd4fbe61f96ddec5d4db89db0ee195a4f71ee/v20.2/show-range-for-row.md#L19)):*
<details><summary><i>Previously, solongordon (Solon) wrote…</i></summary><blockquote>

Nit: I find the `(row, value, ...)` terminology kind of confusing. It seems to imply that the first element is a row and the second one is a value. Perhaps better to write something like `(value1, value2, ...)`?
</blockquote></details>

I agree. I like your suggestion better. I got this syntax directly from the CLI help text in sql.y: https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/parser/sql.y#L4991. 

I'll change it here and also open a cockroach PR to change the help text.

PR #8664, opened 2020-10-14
Added note about long-running bulk insert transactions

### LINKED ISSUE #7815: sql/kv: warn users about long-running transactions
Customers have recently encountered problems with long-running transactions. Imagine a company that has a long-running bulk insert transaction (e.g., write transaction) that is causing contention for short-lived transactions (e.g., read + writes) that service their UI. CockroachDB does not handle long-running transactions well often resulting in [RETRY_COMMIT_DEADLINE_EXCEEDED](https://www.cockroachlabs.com/docs/v20.1/transaction-retry-error-reference.html#retry_commit_deadline_exceeded) or ["transaction deadline exceeded" and error code 40001](https://www.cockroachlabs.com/docs/v20.1/common-errors.html#transaction-deadline-exceeded) as well as generally poor performance. We counsel our customers to break these transactions into smaller batches so that they will complete quickly. 

Long-running transactions can be particularly problematic when they are write transactions that block shorter-lived transactions like the ones being surfaced by the UI above. We need to encourage companies to break up their bulk insert transaction into smaller transactions so that they won't cause contention, and therefore retries and poor performance on their other transactions. 

It's also fortuitous timing because we plan to add a transactions page in 20.2 which will surface long-running transactions making it easier to find and troubleshoot these types of problems.

ericharmeling:
Fixes #7815.

### DIFF
--- a/v20.2/insert.md
+++ b/v20.2/insert.md
@@ -9,11 +9,16 @@ The `INSERT` [statement](sql-statements.html) inserts one or more rows into a ta
 
 ## Performance best practices
 
-- To bulk-insert data into an existing table, batch multiple rows in one [multi-row `INSERT`](#insert-multiple-rows-into-an-existing-table) statement and do not include the `INSERT` statements within a transaction. Experimentally determine the optimal batch size for your application by monitoring the performance for different batch sizes (10 rows, 100 rows, 1000 rows).
+- To bulk-insert data into an existing table, batch multiple rows in one [multi-row `INSERT`](#insert-multiple-rows-into-an-existing-table) statement. Experimentally determine the optimal batch size for your application by monitoring the performance for different batch sizes (10 rows, 100 rows, 1000 rows). Do not include multi-row `INSERT` statements within an explicit transaction.
 
-    {{site.data.alerts.callout_info}}
+    {{site.data.alerts.callout_success}}
     You can also use the [`IMPORT INTO`](import-into.html) statement to bulk-insert CSV data into an existing table.
     {{site.data.alerts.end}}
+
+    {{site.data.alerts.callout_info}}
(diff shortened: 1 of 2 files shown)


ericharmeling:
@awoods187 Friendly ping on review. :)

awoods187:
LGTM @dt would you mind taking a look as well?

dt:
@awoods187 I / bulk-io aren't really the experts on INSERT performance and the SQL / OLTP write path. I'd defer to sql exec team on that I think.

awoods187:
Certainly for the bulk ones you are though? This is about the bulk inserts that result in long-running transactions and steering people away from that? I think it makes sense for you to look at that portion? 

@asubiotto will you also PTAL?

dt:
Not really: despite the label, "bulk" INSERTS are still just SQL INSERTS, and don't get any special bulk io treatment/consideration/etc, -- it is more up to the SQL experts to advise on how to use or not to use our SQL interface, including its INSERTS, transaction lifetimes, etc.

asubiotto (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8664#-:-MKFt-xIDYKHUyjIn-96:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187)

lnhsingh (commented):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8664#-:-MKG7EFZCNEEHSeqpLRJ:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @awoods187 and @lnhsingh)

PR #8656, opened 2020-10-14
No-op UPDATE ONLY/DELETE ONLY syntax

ericharmeling:
Fixes #8500. 

Note that this PR will conflict with the updates to `UPDATE` documentation in #8667.

PR #8667, opened 2020-10-14
UPDATE FROM syntax

### LINKED ISSUE #8219: sql: document UPDATE ... FROM syntax
PR https://github.com/cockroachdb/cockroach/pull/39202 added support for a new syntax for UPDATE queries, but was missing a release note, and is therefore not on the [docs page for UPDATE](https://www.cockroachlabs.com/docs/stable/update.html). (The new syntax has been present since v19.2.)

The semantics match that of Postgres: https://www.postgresql.org/docs/10/sql-update.html
> from_item
>
>    A table expression allowing columns from other tables to appear in the WHERE condition and update expressions. This uses the same syntax as the FROM Clause of a SELECT statement; for example, an alias for the table name can be specified. Do not repeat the target table as a from_item unless you intend a self-join (in which case it must appear with an alias in the from_item).

Another phrasing from the PR description:
> This change adds support for UPDATE ... FROM statements. The FROM clause tables are joined together with the target table and is used as input for the update. Furthermore, the RETURNING clause can reference any table in the FROM clause.

Creating this issue for tracking.

EDIT: Just saw an older issue that I accidentally duplicated (#5254). I'll keep this one open and close that one since this has more attention now. Also copying this note from there:

> When a FROM clause is present, what essentially happens is that the target table is joined to the tables mentioned in the from_list, and each output row of the join represents an update operation for the target table. When using FROM you should ensure that the join produces at most one output row for each row to be modified. In other words, a target row shouldn't join to more than one row from the other table(s). If it does, then only one of the join rows will be used to update the target row, but which one will be used is not readily predictable.
>
>    Because of this indeterminacy, referencing other tables only within sub-selects is safer, though often harder to read and slower than using a join.

ericharmeling:
Fixes #8219. 
Documents https://github.com/cockroachdb/cockroach/pull/39202.

Note that this PR will conflict with the updates to `UPDATE` documentation in #8656.

### DIFF
--- a/_includes/v19.2/sql/diagrams/update.html
+++ b/_includes/v19.2/sql/diagrams/update.html
@@ -1,118 +1,142 @@
-<div><svg width="666" height="724">
+<div><svg width="878" height="848">
 <polygon points="9 61 1 57 1 65"></polygon>
 <polygon points="17 61 9 57 9 65"></polygon>
-<rect x="51" y="47" width="56" height="32" rx="10"></rect>
-<rect x="49" y="45" width="56" height="32" class="terminal" rx="10"></rect>
+<rect x="51" y="47" width="58" height="32" rx="10"></rect>
+<rect x="49" y="45" width="58" height="32" class="terminal" rx="10"></rect>
 <text class="terminal" x="59" y="65">WITH</text>
 <a xlink:href="sql-grammar.html#common_table_expr" xlink:title="common_table_expr">
-<rect x="147" y="47" width="144" height="32"></rect>
-<rect x="145" y="45" width="144" height="32" class="nonterminal"></rect>
-<text class="nonterminal" x="155" y="65">common_table_expr</text>
+<rect x="149" y="47" width="148" height="32"></rect>
+<rect x="147" y="45" width="148" height="32" class="nonterminal"></rect>
+<text class="nonterminal" x="157" y="65">common_table_expr</text>
 </a>
-<rect x="147" y="3" width="24" height="32" rx="10"></rect>
-<rect x="145" y="1" width="24" height="32" class="terminal" rx="10"></rect>
-<text class="terminal" x="155" y="21">,</text>
-<rect x="351" y="47" width="74" height="32" rx="10"></rect>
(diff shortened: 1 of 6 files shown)


RaduBerinde (commented):
We are missing a description of the semantics of UPDATE FROM. Postgres [docs](https://www.postgresql.org/docs/9.5/sql-update.html) have this note (which applies to us as well):


> When a FROM clause is present, what essentially happens is that the target table is joined to the tables mentioned in the from_item list, and each output row of the join represents an update operation for the target table. When using FROM you should ensure that the join produces at most one output row for each row to be modified. In other words, a target row shouldn't join to more than one row from the other table(s). If it does, then only one of the join rows will be used to update the target row, but which one will be used is not readily predictable.
> 
> Because of this indeterminacy, referencing other tables only within sub-selects is safer, though often harder to read and slower than using a join.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8667#-:-MJhc23t6xfJvS_oZrOM:b-aeyr3s)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @RaduBerinde)

ericharmeling (commented):
TFTR, @raduberinde 

I added a section that summarizes the Postgres docs. Please re-review for accuracy.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8667#-:-MK0XUdH4h0kiclviVL2:bab37sn)* status: :shipit: complete! 0 of 0 LGTMs obtained

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8667#-:-MK0ZWRMAzz5XERIQC2P:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8667#-:-MK0h60t9hsaWqPX5eY8:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (and 1 stale) (waiting on @lnhsingh)

PR #8684, opened 2020-10-16
Change use of "an" to "a" in line 60

dilapitan:
The line in the docs: "In terms of the CAP theorem, this is **an CP** system...", change to "In terms of the CAP theorem, this is **a CP** system..."


Amruta-Ranade:
Can you update 20.2 docs too?

jseldess:
Thanks for this fix, @dilapitan. I've opened a new PR making this change across all versions of our docs: https://github.com/cockroachdb/docs/pull/11789. Sorry it took so long.

Issue #7294, opened 2020-05-06
Add an example for filtering on values in JSON columns


SpicyThrow:
On this page, https://www.cockroachlabs.com/docs/v19.2/jsonb.html another useful example would be to show how to filter on the values in the key-value pairs. Currently, this page only shows queries against the keys. 

Something like: 

`SELECT id, last_latitude, last_longitude, vehicle_info -> 'wear' FROM vehicles WHERE vehicle_info @> '{"wear":"damaged"}';`

The `@>` is illustrated on the page, but it's unclear that this can be used for filtering on the value.

dgabriel123:
Hey @Amruta-Ranade , I'd like to give this a try. I know JSON pretty well, but I may need some guidance if you have the bandwidth to help me with this. Thanks!

Amruta-Ranade:
I know nothing of JSON 😅.  But give it a try -- we can ask for help if we need it :)

dgabriel123:
> I know nothing of JSON 😅. But give it a try -- we can ask for help if we need it :)

Looking at the issue more, this is definitely more complicated than I expected 😅 I will opt out for now, thanks!

vipulgupta2048:
Hey @Amruta-Ranade saw your post on WTD, I can take this issue up. Will make a PR.

Amruta-Ranade:
Addressed in https://github.com/cockroachdb/docs/pull/8696

PR #8696, opened 2020-10-17
Add example for filtering key-value pairs

vipulgupta2048:
Add example for filtering JSONB values through key-value pairs 

Solves #7294 
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipulgupta2048@gmail.com>

### DIFF
--- a/v19.2/jsonb.md
+++ b/v19.2/jsonb.md
@@ -182,6 +182,20 @@ You can also use the `->>` operator to return `JSONB` field values as `STRING` v
 +-----------------------------+---------------------------+
 ~~~
 
+You can use the `@>` operator to filter the values in key-value pairs to return `JSONB` field values:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT user_profile->'first_name', user_profile->'location' FROM users WHERE user_profile @> '{"location":"NYC"}';
+~~~
+~~~
++-----------------------------+---------------------------+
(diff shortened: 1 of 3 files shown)


SpicyThrow:
I think this looks good, but someone with more JSON expertise than me should verify that this is correct.

Amruta-Ranade:
Got an internal LGTM 🎉
@vipulgupta2048 Can you update the 20.1 and 20.2 docs too?

vipulgupta2048:
Awesome @SpicyThrow @Amruta-Ranade Just a side thought, I like looking into the documentation pipelines and am always curious about them (part of the job) and I wanted to suggest if you folks have thought of going more into `docs as code` style for your documentation. I haven't looked at Jekyll for quite some time, but I feel the documentation's versioning and generation could be automated. Removing a lot of overhead on writers and contributors because it wasn't clear to me which version to edit and but issue had a link mentionining 19.2. Nonetheless, love to have a discussion on issue or something. 

Will update the PR for 20.1 and 20.2

Amruta-Ranade:
We have had discussions about it but never enough bandwidth to dive into the issue. We now have a product operations engineer onboard, so we might revisit this issue some time in the near future :)

vipulgupta2048:
That's great to hear, thank you for the prompt response @Amruta-Ranade More power to you folks! I will be especially interested to see that happen. I have updated the PR.

Issue #1870, opened 2017-08-30
Link from data type docs to relevant functions and operators


d4l3k:
I was looking for date manipulation functions and stumbled upon https://www.cockroachlabs.com/docs/stable/date.html, but what I really was looking for was https://www.cockroachlabs.com/docs/stable/functions-and-operators.html#date-and-time-functions. 

Might be a good idea to add a link from the data type to the functions that operate on it under the "See More" section.

This also applies to other data types and not just DATE.

Tawishi:
I would like to work on this issue.

Amruta-Ranade:
@Tawishi Sure. If you are participating in Hacktoberfest, I recommend waiting until at least Tuesday to open a PR. We haven't updated the repo to follow the [new rules](https://hacktoberfest.digitalocean.com/hacktoberfest-update) yet.

Tawishi:
Okay, till Tuesday then. :smiley:

shrey27tri01:
Hello! Is someone working on this issue?
If no, can I take this up? :)

Amruta-Ranade:
Let me check :)
@Tawishi Are you working on this? Feel free to open a PR now -- we marked the repo for hacktoberfest :)

Tawishi:
@Amruta-Ranade  Where are the data types to be linked?

Amruta-Ranade:
I think the work here is to add link to
functions and operators in https://www.cockroachlabs.com/docs/stable/functions-and-operators.html from data types in https://www.cockroachlabs.com/docs/stable/data-types.html

jseldess:
@Amruta-Ranade, @Tawishi, that's not really the solution we're looking for. The request is to link to relevant functions for each data type. For example, link the [array data type](https://www.cockroachlabs.com/docs/v20.1/array) doc to the [array functions](https://www.cockroachlabs.com/docs/v20.1/functions-and-operators.html#array-functions).

Let me open a quick PR to demonstrate how we want this done, @Tawishi.

jseldess:
@Tawishi, please emulate what I've done in this PR for the `ARRAY` type: https://github.com/cockroachdb/docs/pull/8623. You'll need to look through the other type docs and see if there are corresponding sections of the function doc to link to. In some cases, there probably won't be. For a few types, we might already be doing this. 

Please ping here if you have questions.

Tawishi:
[...]

PR #8701, opened 2020-10-19
Link data types to functions and operators

Tawishi:
Addresses #1870 
@jseldess Please review the changes and suggest more, if needed.

### DIFF
--- a/v20.1/bool.md
+++ b/v20.1/bool.md
@@ -78,5 +78,5 @@ Type | Details
 
 ## See also
 
-[Data Types](data-types.html)
-[Functions and Operators](https://www.cockroachlabs.com/docs/v20.1/functions-and-operators.html#bool-functions)
+* [Data Types](data-types.html)
+* [Functions and Operators](https://www.cockroachlabs.com/docs/v20.1/functions-and-operators.html#bool-functions)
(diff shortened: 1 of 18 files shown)


Tawishi:
Can _decimal, float, int_ types be linked to _maths and numeric_ functions and operators?

jseldess (commented):
This is great. Thank you, @tawishi. Just two requests:

1. Please apply the same changes to the v20.2 docs as well.
2. Perhaps we should add a "Functions" section to both the `UUID` and `INT` pages, both linking to: `functions-and-operators.html#id-generation-functions`?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8701#-:-MK1SmV631mE1CWCRSsi:b-4znxom)* status: :shipit: complete! 0 of 0 LGTMs obtained

Tawishi:
> This is great. Thank you, @Tawishi. Just two requests:
> 
> 1. Please apply the same changes to the v20.2 docs as well.
> 2. Perhaps we should add a "Functions" section to both the `UUID` and `INT` pages, both linking to: `functions-and-operators.html#id-generation-functions`?

I have made the necessary changes in v20.2 also. Please review and suggest changes, if needed.

Tawishi:
@jseldess Why is the ```TeamCity build``` failing ?

PR #8715, opened 2020-10-20
Database drops and renames in transactions

### LINKED ISSUE #8314: sql: use leased database descriptors, implement db schema changes in job
PR: https://github.com/cockroachdb/cockroach/pull/52975

From release notes:
> Renaming and dropping databases now satisfy the same transactional guarantees as online schema changes on tables: Changes will become visible if and only if the transaction commits, and a successful result indicates changes having propagated to the entire cluster. This fixes some long-standing problems with inconsistent database state being visible after renames and drops. (As a side effect of coherent caching for databases, database names can no longer be recycled in multiple schema changes in the same transaction.) [#52975][#52975] {% comment %}doc{% endcomment %}

ericharmeling:
Fixes #8314.
Fixes https://github.com/cockroachdb/docs/issues/8760.


lucy-zhang (changes requested):
A problem here is that our existing table and database schema change documentation (for rename and drop) is incorrect.

Neither database nor table renames/drops are transactional in a strict sense, but table schema changes have useful guarantees that database schema changes did not prior to 20.2. The problem specific to database schema changes is that it was possible to have the database rename/drop statement return a success result, but for the old state to still persist and be visible on other gateway nodes afterward. (This is https://github.com/cockroachdb/cockroach/issues/41894.) This is what's fixed in 20.2 by the associated PR, and now database and table renames/drops behave in the same way and are subject to the same limitations.

The problem was that https://github.com/cockroachdb/docs/pull/4442 seems to have introduced erroneous changes to the documentation. It wasn't correct at the time because the limitations under discussion at the time only applied to database schema changes, and the description of how renames/drops work is incorrect for both tables and databases.

So I think we need to get rid of these warnings entirely for 20.2, and put in the correct warnings for previous versions. Table renames/drops aren't strictly transactional, but they're part of the async schema changes framework (which in general guarantees that we don't return a success result until the schema change completes), and it'd be better to just point out that renames/drops are transactional schema changes and link to that page.

For the previous versions, I'll open a separate issue.

----

There is one other complication, which is mostly unrelated, but I don't know if it's documented anywhere else: We don't allow atomically swapping names in the same transaction. (See https://github.com/cockroachdb/cockroach/issues/54562.) This particular limitation has always existed for tables, but not for databases _until 20.2_, where it now does apply. The original old issue for this is https://github.com/cockroachdb/cockroach/issues/12123, which is erroneously linked in the existing docs but is unrelated to the problem being discussed in the docs.

I think the confusion arose because the OP of this forum thread https://forum.cockroachlabs.com/t/renaming-database-in-a-transaction/2397 was trying to do transactional name swaps, and the discussion turned to limitations of database renames in general.

Anyway, maybe we should document this as a new limitation for database renames. I'll also open an issue for that if you thin it's a good idea.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715#-:-MKbX9Cv0P90e_6o3bUd:byj5sl2)* status: :shipit: complete! 0 of 0 LGTMs obtained

lucy-zhang (commented):
I don't think I can edit comments in Reviewable but the end of the 3rd paragraph should say "and it'd be better to just point out that renames/drops are _online_ schema changes and link to that page." The page being https://www.cockroachlabs.com/docs/v20.1/online-schema-changes.html.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715#-:-MKbinXp2us3srUlzYeT:bx08atb)* status: :shipit: complete! 0 of 0 LGTMs obtained

lucy-zhang:
Issue for past versions: https://github.com/cockroachdb/docs/issues/8760

lucy-zhang:
Re: atomically swapping names, I just saw that in our online schema changes documentation, we do already say this:

> A table cannot be dropped and then recreated with the same name. This is not possible within a single transaction because DROP TABLE does not immediately drop the name of the table. As a workaround, split the DROP TABLE and CREATE TABLE statements into separate transactions.

https://www.cockroachlabs.com/docs/v20.1/online-schema-changes.html#limited-support-for-schema-changes-within-transactions

So I think for 20.2, we should expand this to indicate that this is true for databases, schemas, and user-defined types (enums). Also, we forbid any reuse of a name within a transaction, not just reuse after drop. (https://github.com/cockroachdb/cockroach/issues/54562 has some other examples: for instance, you can't rename `a` to `b` and then rename `c` to `a` in the same transaction.)

ericharmeling (commented):
Thanks for the detailed correction here, @lucy-zhang. Super helpful!

Please review the diff now (I squashed my edits into a single commit). This PR should now also address https://github.com/cockroachdb/docs/issues/8760.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715#-:-MKelsEv4tjMK1GmgmcR:b8rzq5v)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lucy-zhang)

ericharmeling:
@lucy-zhang Friendly ping on re-review. :)

lucy-zhang (approved):
> Reviewed 12 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v19.2/rename-database.md, line 29 at r2](https://reviewable.io/reviews/cockroachdb/docs/8715#-ML8ZHvm4V5PH9bZs9WU:-ML8ZHvm4V5PH9bZs9WV:b-39xydk) ([raw file](https://github.com/cockroachdb/docs/blob/876206cb9672e60044c9d9fae59e2f3853576e3e/v19.2/rename-database.md#L29)):*
> ```Markdown
> ## Limitations
> 
> If an `ALTER DATABASE ... RENAME` statement is issued on a single [gateway node](architecture/sql-layer.html#overview) and a successful result is returned, it is possible to observe the old database names in [transactions](transactions.html) on other gateway nodes for a short amount of time after the rename is executed. This issue is specific to databases, which have their metadata stored in an incoherent cache, unlike tables. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/41942). Note that this issue is [resolved in v20.2](https://github.com/cockroachdb/cockroach/pull/52975).
> ```

I don't think we should link to 41942. Its description is kind of misleading in that it describes the state of schema changes circa 2.1, even prior to the state of things we've been discussing. There's not really a good tracking issue for this, so I think we can leave it out.
___
*[v20.1/rename-database.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/8715#-ML8_4p1EDZ6goN14WLP:-ML8_4p1EDZ6goN14WLQ:bwow5ba) ([raw file](https://github.com/cockroachdb/docs/blob/876206cb9672e60044c9d9fae59e2f3853576e3e/v20.1/rename-database.md#L27)):*
> ```Markdown
> ## Limitations
> 
> If an `ALTER DATABASE ... RENAME` statement is issued on a single [gateway node](architecture/sql-layer.html#overview) and a successful result is returned, it is possible to observe the old database names in [transactions](transactions.html) on other gateway nodes for a short amount of time after the rename is executed. This issue is specific to databases, which have their metadata stored in an incoherent cache, unlike tables. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/41942). Note that this issue is [resolved in v20.2](https://github.com/cockroachdb/cockroach/pull/52975).
> ```

Same comment about the tracking issue.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lucy-zhang)

___
*[v19.2/rename-database.md, line 29 at r2](https://reviewable.io/reviews/cockroachdb/docs/8715#-ML8ZHvm4V5PH9bZs9WU:-MLDkoel-x1H4SjZCFLW:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/876206cb9672e60044c9d9fae59e2f3853576e3e/v19.2/rename-database.md#L29)):*
<details><summary><i>Previously, lucy-zhang (Lucy Zhang) wrote…</i></summary><blockquote>

I don't think we should link to 41942. Its description is kind of misleading in that it describes the state of schema changes circa 2.1, even prior to the state of things we've been discussing. There's not really a good tracking issue for this, so I think we can leave it out.
</blockquote></details>

Done.
___
*[v20.1/rename-database.md, line 27 at r2](https://reviewable.io/reviews/cockroachdb/docs/8715#-ML8_4p1EDZ6goN14WLP:-MLDkoDcCRrHTd8eExcU:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/876206cb9672e60044c9d9fae59e2f3853576e3e/v20.1/rename-database.md#L27)):*
<details><summary><i>Previously, lucy-zhang (Lucy Zhang) wrote…</i></summary><blockquote>

Same comment about the tracking issue.
</blockquote></details>

Done.

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8715#-:-MLJ0WjgEaZmOpp02X7D:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lnhsingh and @lucy-zhang)

PR #8718, opened 2020-10-20
Link to ALTER PRIMARY KEY to change bucket size of hash-sharded index

### LINKED ISSUE #7620: sql: don't copy old pk when changing bucket size on sharded index
PR: https://github.com/cockroachdb/cockroach/pull/45894

From release notes:
> Caused [primary key changes](../v20.2/alter-primary-key.html) to not create a copy of the old primary key if the primary key change only changes the bucket count of a [hash sharded index](../v20.2/indexes.html#hash-sharded-indexes). [#45894][#45894] {% comment %}doc{% endcomment %}

ericharmeling:
- Added short paragraph linking to `ALTER PRIMARY KEY` for changing bucket size of hash-sharded index.

### DIFF
--- a/v20.1/indexes.md
+++ b/v20.1/indexes.md
@@ -57,6 +57,8 @@ If you are working with a table that must be indexed on sequential keys, you sho
 
 To create a hash-sharded index, set the `experimental_enable_hash_sharded_indexes` [session variable](set-vars.html) to `on`. Then, add the optional [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) to a [`CREATE INDEX`](create-index.html) statement, to an [`INDEX` definition](sql-grammar.html#index_def) in a [`CREATE TABLE`](create-table.html) statement, or to an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement. When this clause is used, CockroachDB creates `n_buckets` computed columns, shards the index into `n_buckets` shards, and then stores each index shard in the underlying key-value store with one of the computed column's hash as its prefix.
 
+To change the bucket size of an existing hash-sharded primary key index, use an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement with a [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) that specifies the new bucket size and the existing primary key columns.
+
 {{site.data.alerts.callout_info}}
 Hash-sharded indexes cannot be [interleaved](interleave-in-parent.html).
 {{site.data.alerts.end}}
--- a/v20.2/indexes.md
+++ b/v20.2/indexes.md
@@ -63,6 +63,8 @@ If you are working with a table that must be indexed on sequential keys, you sho
 
 To create a hash-sharded index, set the `experimental_enable_hash_sharded_indexes` [session variable](set-vars.html) to `on`. Then, add the optional [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) to a [`CREATE INDEX`](create-index.html) statement, to an [`INDEX` definition](sql-grammar.html#index_def) in a [`CREATE TABLE`](create-table.html) statement, or to an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement. When this clause is used, CockroachDB creates `n_buckets` computed columns, shards the index into `n_buckets` shards, and then stores each index shard in the underlying key-value store with one of the computed column's hash as its prefix.
 
+To change the bucket size of an existing hash-sharded primary key index, use an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement with a [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) that specifies the new bucket size and the existing primary key columns.
+
 {{site.data.alerts.callout_info}}
 Hash-sharded indexes cannot be [interleaved](interleave-in-parent.html).
 {{site.data.alerts.end}}


ajwerner (commented):
Playing around with hash sharded indexed because of this lead me to https://github.com/cockroachdb/cockroach/pull/55766 which is nice. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8718#-:-MK6CNAn1mbhB8OxQ9Dn:bf3nu93)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajwerner and @ericharmeling)

___
*[v20.2/alter-primary-key.md, line 42 at r1](https://reviewable.io/reviews/cockroachdb/docs/8718#-MK6ByWd6xMzASVkokOH:-MK6ByWe7mmtHt-vzKWu:b-cxz30e) ([raw file](https://github.com/cockroachdb/docs/blob/8328ce0abb8edb375880338f98e12d8ad776afd9/v20.2/alter-primary-key.md#L42)):*
> ```Markdown
>  `index_params` | The name of the column(s) that you want to use for the primary key. These columns replace the current primary key column(s).
>  `opt_interleave` | You can potentially optimize query performance by [interleaving tables](interleave-in-parent.html), which changes how CockroachDB stores your data.<br>{{site.data.alerts.callout_info}}[Hash-sharded indexes](indexes.html#hash-sharded-indexes) cannot be interleaved.{{site.data.alerts.end}}
>  `USING HASH WITH BUCKET COUNT` | Creates a [hash-sharded index](indexes.html#hash-sharded-indexes) with `n_buckets` number of buckets.<br><br>When you use an `ALTER PRIMARY KEY` statement that only changes the bucket size of a hash-sharded primary key index, the old primary key (and the old bucket size) are permanently discarded.{{site.data.alerts.callout_info}}To enable hash-sharded indexes, set the `experimental_enable_hash_sharded_indexes` [session variable](set-vars.html) to `on`.{{site.data.alerts.end}}
> ```

Hmm, I find this confusing. 

1) What is the implication of ` the old primary key (and the old bucket size) are permanently discarded.`

2) I found `When you use an `ALTER PRIMARY KEY` statement that only changes the bucket size of a hash-sharded primary key index` to maybe imply that something like `ALTER TABLE foo ALTER PRIMARY KEY USING HAS WITH BUCKET_COUNT = 8` would work and keep the same columns, but it doesn't.

ericharmeling (commented):
@ajwerner 

> Playing around with hash sharded indexed because of this lead me to cockroachdb/cockroach#55766 which is nice.
Nice!

TFTR, Andrew.

After looking over https://github.com/cockroachdb/cockroach/pull/45894 again, and after reading your comments, I'm not so sure we want to document this behavior. It seems like doing so might raise more questions than necessary. Objections to just closing out #7620 with docs as-is?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8718#-:-MKFjJl49GiVHzuBvZS6:b-192lg6)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajwerner)

___
*[v20.2/alter-primary-key.md, line 42 at r1](https://reviewable.io/reviews/cockroachdb/docs/8718#-MK6ByWd6xMzASVkokOH:-MKFiVq_8tzUpwcLLuiU:b-cnmnsa) ([raw file](https://github.com/cockroachdb/docs/blob/8328ce0abb8edb375880338f98e12d8ad776afd9/v20.2/alter-primary-key.md#L42)):*
<details><summary><i>Previously, ajwerner wrote…</i></summary><blockquote>

Hmm, I find this confusing. 

1) What is the implication of ` the old primary key (and the old bucket size) are permanently discarded.`

2) I found `When you use an `ALTER PRIMARY KEY` statement that only changes the bucket size of a hash-sharded primary key index` to maybe imply that something like `ALTER TABLE foo ALTER PRIMARY KEY USING HAS WITH BUCKET_COUNT = 8` would work and keep the same columns, but it doesn't. 
</blockquote></details>

1. It does seem like a pretty loaded clause... It might be better just to say `the old primary key and bucket size are not retained`.
2. Is this behavior worth clarifying?

ajwerner:
Ah yeah, I think I should have gone back and looked at https://github.com/cockroachdb/cockroach/pull/45894. The point is that normally when you change the primary key, we retain a unique constraint over the old primary key columns but here you aren't changing that set of columns. I agree, no docs change is in order.

ericharmeling:
> Ah yeah, I think I should have gone back and looked at [cockroachdb/cockroach#45894](https://github.com/cockroachdb/cockroach/pull/45894). The point is that normally when you change the primary key, we retain a unique constraint over the old primary key columns but here you aren't changing that set of columns. I agree, no docs change is in order.

Sounds good. Removed the stuff about old pk keys, etc. But I kept the paragraph linking to `ALTER PRIMARY KEY` to change bucket size.

v20.1/indexes.md
```diff
@@ -57,6 +57,8 @@ If you are working with a table that must be indexed on sequential keys, you sho
 
 To create a hash-sharded index, set the `experimental_enable_hash_sharded_indexes` [session variable](set-vars.html) to `on`. Then, add the optional [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) to a [`CREATE INDEX`](create-index.html) statement, to an [`INDEX` definition](sql-grammar.html#index_def) in a [`CREATE TABLE`](create-table.html) statement, or to an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement. When this clause is used, CockroachDB creates `n_buckets` computed columns, shards the index into `n_buckets` shards, and then stores each index shard in the underlying key-value store with one of the computed column's hash as its prefix.
 
+To change the bucket size of an existing hash-sharded primary key index, use an [`ALTER PRIMARY KEY`](alter-primary-key.html) statement with a [`USING HASH WITH BUCKET_COUNT = n_buckets` clause](sql-grammar.html#opt_hash_sharded) that specifies the new bucket size.
```
ajwerner, line 60: I think I'd add that you need to specify the existing primary key columns too.

ajwerner (approved):
LGTM

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8718#-:-MKMA8eF1oACenYNgKLM:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajwerner, @ericharmeling, and @lnhsingh)

PR #8732, opened 2020-10-22
Fix docker pull command for stable releases

### LINKED ISSUE #8730: Error pulling Docker v20.1.8
Re: [What's New in v20.1.8](https://cockroachlabs.com/docs//releases/v20.1.8.html)

## Issue Description

```
% docker pull cockroachdb/cockroach-stable:v20.1.8
Error response from daemon: pull access denied for cockroachdb/cockroach-stable, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

jseldess:
In a few places, we were using `cockroach/cockroach-stable:<version>`.
`cockroach-stable` isn't valid. It should just be
`cockroach/cockroach:<version>`.

Fixes #8730.

### DIFF
--- a/releases/v19.1.7.md
+++ b/releases/v19.1.7.md
@@ -38,7 +38,7 @@ Get future release notes emailed to you:
 
 {% include copy-clipboard.html %}
 ~~~shell
-$ docker pull cockroachdb/cockroach-stable:v19.1.7
+$ docker pull cockroachdb/cockroach:v19.1.7
 ~~~
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


ericharmeling (approved):
LGTM

Should we also replace instances of `cockroachdb/cockroach-unstable` with `cockroachdb/cockroach` in production release pages (e.g., 19.1.6)?

jseldess (commented):
I don't *think* so. I think alpha/beta/rc versions will always have the `-unstable` prefix, even after the GA release of that version. @bdarnell, do you agree? 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8732#-:-MKGMQGx7vXqviJ2_yWb:b-k2tr23)* status: :shipit: complete! 0 of 0 LGTMs obtained

bdarnell:
Correct.

Issue #4962, opened 2019-06-24
Document workaround for MySQL FIELD function


rmloveland:
We should document a workaround for users who are using the [MySQL FIELD function](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_field) and want to [migrate to CockroachDB from MySQL](https://www.cockroachlabs.com/docs/stable/migrate-from-mysql.html).
 
According to [this forum post on Sorting using FIELD](https://forum.cockroachlabs.com/t/sorting-using-field/2865) there may be a workaround using `array_position`.

tim-o:
Zendesk ticket #[3462](https://cockroachdb.zendesk.com/agent/tickets/3462) has been linked to this issue.

deeptikorwar:
@Amruta-Ranade, can I work on this one?

deeptikorwar:
@rmloveland, I am an external contributor and I don't have access to this Zendesk ticket: [#3462](https://cockroachdb.zendesk.com/agent/tickets/3462)

Based on [this forum post on Sorting using FIELD](https://forum.cockroachlabs.com/t/sorting-using-field/2865), I'm planning to add a new section under "Considerations" in [migrate to CockroachDB from MySQL](https://www.cockroachlabs.com/docs/stable/migrate-from-mysql.html).

### FIELD function
The MYSQL FIELD function is not supported in CockroachDB. Instead, you can use the `array_position` function, which returns the index of the first occurrence of element in the array.

While MYSQL returns 0 when the element is not found, CockroachDB returns `NULL`. So if you are using the `ORDER BY` clause with the `array_position`  function, the caveat is that sort is applied even when the element is not found. As a workaround, you can use the COALESCE operator.

Example:
`SELECT array_position(ARRAY[4,1,3,2],1);`

  `array_position`
`------------------`
  `             2`
`(1 row)`

`SELECT * FROM table_a ORDER BY COALESCE(array_position(ARRAY[4,1,3,2],5),999);`

Will this work? Is there anything else that I can add on this workaround?

rmloveland:
Hy @deeptikorwar thanks for pulling this together.  I think that works.  Mind putting it in a PR so we can test and review?  Thanks!

rmloveland:
Fixed by #8751

PR #8751, opened 2020-10-24
Added workaround for FIELD function

deeptikorwar:
Addresses #4962

@rmloveland, I've added the workaround details in 20.1, but looks like the syntax is not quite right. 

I tried the example in 20.1  and although both `SELECT array_position(ARRAY[4,1,3,2],1);` and  `SELECT * FROM table_a ORDER BY 2` work, `SELECT * FROM table_a ORDER BY array_position(ARRAY[4,1,3,2],1);` doesn't.

### DIFF
--- a/v20.1/migrate-from-mysql.md
+++ b/v20.1/migrate-from-mysql.md
@@ -24,8 +24,6 @@ For more information about the case sensitivity of strings in MySQL, see [Case S
 
 The MYSQL `FIELD` function is not supported in CockroachDB. Instead, you can use the [`array_position`](functions-and-operators.html#array-functions) function, which returns the index of the first occurrence of element in the array.
 
-While MYSQL returns 0 when the element is not found, CockroachDB returns `NULL`. So if you are using the `ORDER BY` clause in a statement with the `array_position` function, the caveat is that sort is applied even when the element is not found. As a workaround, you can use the [`COALESCE`](functions-and-operators.html#conditional-and-function-like-operators) operator.
-
 Example usage:
 
 {% include copy-clipboard.html %}
@@ -39,6 +37,8 @@ Example usage:
                2
 (1 row)
 ~~~
+
+While MYSQL returns 0 when the element is not found, CockroachDB returns `NULL`. So if you are using the `ORDER BY` clause in a statement with the `array_position` function, the caveat is that sort is applied even when the element is not found. As a workaround, you can use the [`COALESCE`](functions-and-operators.html#conditional-and-function-like-operators) operator.
 
 {% include copy-clipboard.html %}
 ~~~ sql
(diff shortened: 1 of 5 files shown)


v20.1/migrate-from-mysql.md
```diff
@@ -20,6 +20,31 @@ MySQL strings are case-insensitive by default, but strings in CockroachDB are ca
 
 For more information about the case sensitivity of strings in MySQL, see [Case Sensitivity in String Searches](https://dev.mysql.com/doc/refman/8.0/en/case-sensitivity.html) from the MySQL documentation.  For more information about CockroachDB strings, see [`STRING`](string.html).
 
+### `FIELD` function
+
+The MYSQL `FIELD` function is not supported in CockroachDB. Instead, you can use the [`array_position`](functions-and-operators.html#array-functions) function, which returns the index of the first occurrence of element in the array.
+
+While MYSQL returns 0 when the element is not found, CockroachDB returns `NULL`. So if you are using the `ORDER BY` clause in a statement with the `array_position` function, the caveat is that sort is applied even when the element is not found. As a workaround, you can use the [`COALESCE`](functions-and-operators.html#conditional-and-function-like-operators) operator.
```
rmloveland, line 27: Can you please move this paragraph above the `COALESCE` example? (I have added another comment with 'X' to mark the spot)

That way the reader can read in a potentially nicer order:

1) intro
2) simple example
3) this paragraph with the more involved description of the workaround
4) the workaround

v20.1/migrate-from-mysql.md
```diff
@@ -20,6 +20,31 @@ MySQL strings are case-insensitive by default, but strings in CockroachDB are ca
...
+> SELECT array_position(ARRAY[4,1,3,2],1);
+~~~
+
+~~~
+  array_position
+------------------
+               2
+(1 row)
+~~~
+
```
rmloveland, line 42: X

rmloveland (changes requested):
This is really great @deeptikorwar !  Thank you for submitting this! I have requested only one change in the ordering, but otherwise it's very nice.

PS Sorry for the slow response on my end, we are in an especially busy time getting ready for the 20.2 release.

v20.1/migrate-from-mysql.md
deeptikorwar: Thank you.

rmloveland:
Hi @deeptikorwar, can you please change this PR so it only adds this info to the following versions:

- 19.2
- 20.1
- 20.2


We are releasing 20.2 very soon (this month), and our support policy (IIRC) is "current version plus last 2 releases".

Thanks!  This is really great, as soon as you change to affect just these versions, we can merge this immediately.

deeptikorwar:
Thanks @rmloveland. I've taken those files out.

rmloveland (approved):

rmloveland:
Thanks @deeptikorwar !  Very nice contribution.  Merging now :-)

PR #8754, opened 2020-10-26
Gracefully handle hanging backups 

### LINKED ISSUE #7930: Graceful Handling of Hanging Backups
When Backups hang silently, even if there is a legit reason that will be cleaned up soon, customers get nervous because it means a potential loss of data. Consistency in backups being taken gives them confidence that their data is safe.

We want to have some way for backups to succeed in light of hanging--overriding priority, etc. so that data is secure.

Fixes a known issue where backup performance can appear to be almost stuck in tables that have lots of intents.

Backup is almost unusably slow in some circumstances, and users do not have a good way of knowing what is causing this performance degradation. This is an attempt to address one known cause of backup performance degradation.

See this in registration, but also other tables as well.

Note: @pbardea added the current workaround to the [on-call runbook](https://cockroachlabs.atlassian.net/wiki/spaces/TS/pages/269156516/Bulk-IO+and+Schema+Change+Runbook).

cc: @mwang1026

### LINKED ISSUE #7943: document impact of transaction contention on BACKUPs
Our docs cover [transaction contention](https://www.cockroachlabs.com/docs/v20.1/performance-best-practices-overview.html#understanding-and-avoiding-transaction-contention) and mentions that this [BACKUPs can be affected](https://www.cockroachlabs.com/docs/dev/backup.html#performance) by this. 

However, we don't go into detail on what happens when a BACKUP is prevented by transaction contention, and how this can be prevented or resolved.

### LINKED ISSUE #8117: backupccl: avoid hanging forever on intents
PR: https://github.com/cockroachdb/cockroach/pull/51624

From release notes:
> The [`BACKUP`](../v20.2/backup.html) statement now takes priority over other transactions if its initial attempts to export a range fail. [#51624][#51624] {% comment %}doc{% endcomment %}

lnhsingh:
Closes #7930.
Closes #8117.
Closes #7943.


v20.2/backup.md
```diff
@@ -111,13 +111,19 @@ For an example of an incremental backup, see the [Create incremental backups](#c
 
 The `BACKUP` process minimizes its impact to the cluster's performance by distributing work to all nodes. Each node backs up only a specific subset of the data it stores (those for which it serves writes; more details about this architectural concept forthcoming), with no two nodes backing up the same data.
 
-For best performance, we also recommend always starting backups with a specific [timestamp](timestamp.html) at least 10 seconds in the past. For example:
+`BACKUP`, like any read, cannot export a range if it contains an unresolved intent. If CockroachDB hits an intent during iteration, CockroachDB will typically return a `WriteIntentError` and let the store resolve it and retry the request before replying to the sender.
```
dt, line 114: is "intent" too jargon-y/internal detail for the backup docs? if so, could say "pending / uncommitted write from another transaction" or something?

v20.2/backup.md
```diff
@@ -111,13 +111,19 @@ For an example of an incremental backup, see the [Create incremental backups](#c
...
+
+However, blocking intent resolution can mean a single transaction left open can cause backups to hang, sometimes for hours or even days in some observed cases. While you typically want bulk, background jobs like `BACKUP` to have as little impact on your foreground traffic as possible, it's more important for backups to actually complete (which maintains your [recovery point objective (RPO)](https://en.wikipedia.org/wiki/Disaster_recovery#Recovery_Point_Objective)). So while you typically would not want a `BACKUP` to immediately run at high priority and abort any transactions from OLTP foreground traffic as soon as it starts, you do, eventually, want it to do whatever it needs to do to finish.
```
dt, line 116: This paragraph taken from the commit message is mostly talking about the _pre_ 20.2 (mis)behavior as motivation for that commit -- "a single transaction left open can cause backups to hang, sometimes for hours or even days" is specifically _not_ a problem in 20.2 thanks to that change. I almost feel like we could elide the paragraph entirely and/or work the reason for invoking priority (RPO) into another sentence elsewhere.

v20.2/backup.md
```diff
@@ -111,13 +111,19 @@ For an example of an incremental backup, see the [Create incremental backups](#c
...
+
+We recommend always starting backups with a specific [timestamp](timestamp.html) at least 10 seconds in the past. For example:
 
 ~~~ sql
 > BACKUP...AS OF SYSTEM TIME '-10s';
 ~~~
 
-This improves performance by decreasing the likelihood that the `BACKUP` will be [retried because it contends with other statements/transactions](transactions.html#transaction-retries). However, because `AS OF SYSTEM TIME` returns historical data, your reads might be stale.
+This improves performance by decreasing the likelihood that the `BACKUP` will be [retried because it contends with other statements/transactions](transactions.html#transaction-retries). However, because `AS OF SYSTEM TIME` returns historical data, your reads might be stale. Taking backups with `AS OF SYSTEM TIME '-10s'` is a good best practice to reduce the number of still-running transactions you may encounter, but an abandoned or very long-running transaction can still hang your a backup.
```
dt, line 124: the end of this (""an abandoned or very long-running transaction can still hang your a backup") isn't strictly true anymore, and we might want to specify _why_ reducing the number of still-running transactions encountered is bad, i.e. that it is bad _for those transactions_ (because the backup takes priority and forces them to restart after it)

lnhsingh (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt and @mwang1026)

___
*[v20.2/backup.md, line 114 at r1](https://reviewable.io/reviews/cockroachdb/docs/8754#-MK_Bio4YBvb10RrkIGA-r1-114:-MKpMx4H2RKmI1kSSRdD:b-v1luvf) ([raw file](https://github.com/cockroachdb/docs/blob/352cd0a6b21ad96663c0febdbcb5581d4cf5cfe0/v20.2/backup.md#L114)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

I realize most of this is my own wording from the PR so my apologies for turning around and nitting it now:
- is "unresolved intent" too jargon-y/internal detail for the backup docs? if so, could say "pending  uncommitted write from another transaction" or something?

- "hits an intent" is also internal lingo I think -- I'd probably say "observes" or "encounters" for an external user?  similarly "store" might be a relatively "internal" concept -- it isn't _quite_ precise but "node" or "host" might be  a more broadly understood term.

actually on second thought, I'd probably cut that second sentence entirely -- in the commit message it came from it was contrasting a BACKUP read from a regular read but this is the backup docs so of course we're talking about backup reads, so talking about how regular reads behave may just confuse a reader. I might instead just say "However, unlike a normal reading transaction that will block until any uncommitted writes it encounters are resolved,  BACKUP will block only for a configurable duration before invoking priority to ensure it can complete on-time" or something like that?
</blockquote></details>

I think "unresolved intent" is ok for backup docs, we talk about write intents in other product docs. I did add a link to our Architecture docs though. Also revised the second sentence with your rec. Thanks!
___
*[v20.2/backup.md, line 116 at r1](https://reviewable.io/reviews/cockroachdb/docs/8754#-MK_Bio4YBvb10RrkIGA-r1-116:-MKp_ZdVDo5bEkO5pG8S:b-vg7syy) ([raw file](https://github.com/cockroachdb/docs/blob/352cd0a6b21ad96663c0febdbcb5581d4cf5cfe0/v20.2/backup.md#L116)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

This paragraph taken from the commit message is mostly talking about the _pre_ 20.2 (mis)behavior as motivation for that commit -- "a single transaction left open can cause backups to hang, sometimes for hours or even days" is specifically _not_ a problem in 20.2 thanks to that change. I almost feel like we could elide the paragraph entirely and/or work the reason for invoking priority (RPO) into another sentence elsewhere.
</blockquote></details>

Removed the first sentence and rearranged some of the following sentences to combine with the description above.
___
*[v20.2/backup.md, line 124 at r1](https://reviewable.io/reviews/cockroachdb/docs/8754#-MK_Bio4YBvb10RrkIGA-r1-124:-MKpae-G77_4HR_5dut0:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/352cd0a6b21ad96663c0febdbcb5581d4cf5cfe0/v20.2/backup.md#L124)):*
<details><summary><i>Previously, dt (David Taylor) wrote…</i></summary><blockquote>

the end of this (""an abandoned or very long-running transaction can still hang your a backup") isn't strictly true anymore, and we might want to specify _why_ reducing the number of still-running transactions encountered is bad, i.e. that it is bad _for those transactions_ (because the backup takes priority and forces them to restart after it)
</blockquote></details>

Done.

lnhsingh (commented):
@dt - mind taking another pass to make sure edits look good? Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754#-:-ML9RHkC801_1_OVm12y:b-l2dqi6)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt and @mwang1026)

mwang1026 (approved):
LGTM! Minor nit and q for david about whether we should mention the cluster setting to change the timeout away from the default

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754#-:-MLJCFab5wtHrkyROZLR:b-ncy5fo)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt, @lnhsingh, and @mwang1026)

___
*[v20.2/backup.md, line 124 at r1](https://reviewable.io/reviews/cockroachdb/docs/8754#-MK_Bio4YBvb10RrkIGA-r1-124:-MLJBoWHCzmYKsju7iXW:bmmdfcg) ([raw file](https://github.com/cockroachdb/docs/blob/352cd0a6b21ad96663c0febdbcb5581d4cf5cfe0/v20.2/backup.md#L124)):*

Done.
</blockquote></details>

v minor nit, that `since the backup will take priority and will force still-running transactions to restart after the backup is finished.` is technically New in v20.2

Not something major enough that I think you should reorg stuff / the new in v20.2 tags.
___
*[v20.2/backup.md, line 124 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLJCANu13h-KkNOynzb:-MLJCANu13h-KkNOynzc:b-qkcdih) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L124)):*
> ```
> default 1 minute
> ```

@dt should we reference the cluster setting here?

lnhsingh (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt)

___
*[v20.2/backup.md, line 124 at r1](https://reviewable.io/reviews/cockroachdb/docs/8754#-MK_Bio4YBvb10RrkIGA-r1-124:-MLKPsXk78QjseeV7fEl:bq0yoxd) ([raw file](https://github.com/cockroachdb/docs/blob/352cd0a6b21ad96663c0febdbcb5581d4cf5cfe0/v20.2/backup.md#L124)):*
<details><summary><i>Previously, mwang1026 (Michael Wang) wrote…</i></summary><blockquote>

v minor nit, that `since the backup will take priority and will force still-running transactions to restart after the backup is finished.` is technically New in v20.2

Not something major enough that I think you should reorg stuff / the new in v20.2 tags.
</blockquote></details>

I think it's ok to leave as-is since we discuss how the backup will take priority in the next paragraph
___
*[v20.2/backup.md, line 124 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLJCANu13h-KkNOynzb:-MLKPzWm95J6EcWrBzk8:bwb0s4t) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L124)):*
<details><summary><i>Previously, mwang1026 (Michael Wang) wrote…</i></summary><blockquote>

> ```
> default 1 minute
> ```

@dt should we reference the cluster setting here?
</blockquote></details>

When I chatted with David on slack, he mentioned that most users prob won't need to change the setting away from the default. Also, the cluster setting doesn't seem to be public (from what I can tell), so we'd probably want to expose it if we are documenting it.

taroface (changes requested):
I think some of the sequences being described here can be clearer. I have some suggestions, but also a few questions about things I still don't understand. No need to respond to these comments line-by-line if you get my overall gist on how the descriptions might be condensed & clarified.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754#-:-MLKlZaR4AOWqccNWwpr:bklpbf3)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt, @lnhsingh, and @taroface)

___
*[v20.2/backup.md, line 114 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLK_wruEAooeOna1Rn4:-MLK_wrv7bSafkAL1Mch:b-6mm182) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L114)):*
> ```
> cannot export a range if it contains
> ```

Nit: maybe "cannot export a range if the range contains" for absolute clarity

>```
> While you typically want
>```

Nit: suggest "While you will typically want" as this sounds slightly less conversational

>```
> So unlike a normal read transaction
>```

I suggest cutting "So" for the same reason.
___
*[v20.2/backup.md, line 124 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLKaoWjEO7kQN0IHgk7:-MLKaoWjEO7kQN0IHgk8:b-id7057) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L124)):*
> ```
> `BACKUP` will initially start asking individual ranges to backup but to skip if they encounter an intent
> ``` 

"initially start asking" -> "initially ask"

> ```
> , rather than attempt to resolve the intent and hold up the whole backup while doing so
> ```

IMHO this intention is clear enough from the description that follows, and can be cut for the sake of brevity

> ```
> If any range replies that it skipped, it is put at the back of the queue to come back to later. `BACKUP` will continue to work on other ranges, in the hopes that by the time it is done, the intent may have already resolved itself. When `BACKUP` revisits a range that it has reached the time limit for attempted backups (default 1 minute), it instead asks the range to backup but to _not_ skip if it hits an intent, and to try to resolve it instead.
> ```

Suggesting some line edits in the following vein, please adapt or disregard freely: 

"Any range that is skipped is placed at the end of the queue. When `BACKUP` has completed its initial pass and is revisiting ranges, it asks any range that did not resolve after 1 minute to backup and to attempt to resolve any intents that it encounters.

I may have garbled the facts here (sorry), partly because I can't tell what the 1 minute limit is for. Does it mean that every range (skipped or not) has 1 minute to resolve its intents (if any) and successfully backup? Would this then imply that if `BACKUP` goes through all the ranges in less than 1 minute, it will then wait until 1 minute has passed before asking the remaining ranges to resolve? 

> ```
> Additionally, the backup's transaction priority then is set to high, 
> ```

Does "high" need to have code ticks? Not sure how technical.

> ```
> so that during the intent resolution, other transactions will be aborted in order for the backup to finish.
> ```

Suggest: "which causes other transactions to abort until the intents are resolved and the backup is finished."

taroface (approved):
Really small nit I overlooked previously, but LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754#-:-MLTZUve9Rjnk597HBXc:b-8yhfip)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt and @lnhsingh)

___
*[v20.2/backup.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLTZRY95KllANi3uOCA:-MLTZRYA2v3S31_lSsAu:b-767nz2) ([raw file](https://github.com/cockroachdb/docs/blob/91f8cac6c4480708b77f9ae95c207361cc6b4929/v20.2/backup.md#L124)):*
> ```
> the backup's transaction priority then is set to `high`
> ```

Small nit: "then is" -> "is then"

Sorry I didn't see this before!

lnhsingh (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8754#-:-MLOIeiA-yY8jTrFvdL3:bz5mair)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dt and @taroface)

___
*[v20.2/backup.md, line 114 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLK_wruEAooeOna1Rn4:-MLOGIY8DrVOQemiQKDa:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L114)):*

> ```
> cannot export a range if it contains
> ```

Nit: maybe "cannot export a range if the range contains" for absolute clarity

>```
> While you typically want
>```

Nit: suggest "While you will typically want" as this sounds slightly less conversational

>```
> So unlike a normal read transaction
>```

I suggest cutting "So" for the same reason.
</blockquote></details>

Done.
___
*[v20.2/backup.md, line 124 at r2](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLKaoWjEO7kQN0IHgk7:-MLOHvK68--6I_TCMuea:b-zcuwwl) ([raw file](https://github.com/cockroachdb/docs/blob/126ab4211b76d7c2fe17d3da9e3691c070f29468/v20.2/backup.md#L124)):*

> ```
> `BACKUP` will initially start asking individual ranges to backup but to skip if they encounter an intent
> ``` 

"initially start asking" -> "initially ask"

> ```
> , rather than attempt to resolve the intent and hold up the whole backup while doing so
> ```

IMHO this intention is clear enough from the description that follows, and can be cut for the sake of brevity

> ```
> If any range replies that it skipped, it is put at the back of the queue to come back to later. `BACKUP` will continue to work on other ranges, in the hopes that by the time it is done, the intent may have already resolved itself. When `BACKUP` revisits a range that it has reached the time limit for attempted backups (default 1 minute), it instead asks the range to backup but to _not_ skip if it hits an intent, and to try to resolve it instead.
> ```

Suggesting some line edits in the following vein, please adapt or disregard freely: 

"Any range that is skipped is placed at the end of the queue. When `BACKUP` has completed its initial pass and is revisiting ranges, it asks any range that did not resolve after 1 minute to backup and to attempt to resolve any intents that it encounters.

I may have garbled the facts here (sorry), partly because I can't tell what the 1 minute limit is for. Does it mean that every range (skipped or not) has 1 minute to resolve its intents (if any) and successfully backup? Would this then imply that if `BACKUP` goes through all the ranges in less than 1 minute, it will then wait until 1 minute has passed before asking the remaining ranges to resolve? 

> ```
> Additionally, the backup's transaction priority then is set to high, 
> ```

Does "high" need to have code ticks? Not sure how technical.

> ```
> so that during the intent resolution, other transactions will be aborted in order for the backup to finish.
> ```

Suggest: "which causes other transactions to abort until the intents are resolved and the backup is finished."
</blockquote></details>

Thank you for the suggestions for making this more concise - it was very helpful. I took a stab at incorporating your feedback in here - PTAL. If I understand correctly, every range has 1 min to resolve intents. So yes, I think you're right that if a backup goes through all the ranges in less than a minute, it will have to wait until the end of the minute before the transaction priority is set to high and the ranges push through on resolving intents (rather than skipping)
___
*[v20.2/backup.md, line 124 at r3](https://reviewable.io/reviews/cockroachdb/docs/8754#-MLTZRY95KllANi3uOCA:-MLT_A0w3veCJC0NtDpC:b-jcw1b6) ([raw file](https://github.com/cockroachdb/docs/blob/91f8cac6c4480708b77f9ae95c207361cc6b4929/v20.2/backup.md#L124)):*

> ```
> the backup's transaction priority then is set to `high`
> ```

Small nit: "then is" -> "is then"

Sorry I didn't see this before!
</blockquote></details>

thanks for the catch!

PR #8758, opened 2020-10-26
Regenerated builtin file, with updated crdb_interal.force_retry text

### LINKED ISSUE #5538: Update generated docs for crdb_internal.force_retry
Currently the docs for `crdb_internal.force_retry()` state:

> This function is used only by CockroachDB's developers for testing purposes.

However this is not entirely accurate.  It is useful to developers when writing and testing app retry loops as described in #5035. Thanks to the fix for https://github.com/cockroachdb/cockroach/issues/39243, you also don't need to be root anymore to run it.

I'm not sure exactly what the doc should be, but maybe something like 

> Triggers a transaction retry.  Useful for testing purposes.

Or something even better :-}

To update this you will need to edit `pkg/sql/sem/builtins.go` in the CRDB repo. You'll need to edit the `Info` key shown below.

```golang
// If force_retry is called during the specified interval from the beginning
// of the transaction it returns a retryable error. If not, 0 is returned
// instead of an error.
// The second version allows one to create an error intended for a transaction
// different than the current statement's transaction.
"crdb_internal.force_retry": makeBuiltin(
    tree.FunctionProperties{
        Category: categorySystemInfo,
        Impure:   true,
    },
    tree.Overload{
        Types:      tree.ArgTypes{{"val", types.Interval}},
        ReturnType: tree.FixedReturnType(types.Int),
        Fn: func(ctx *tree.EvalContext, args tree.Datums) (tree.Datum, error) {
            minDuration := args[0].(*tree.DInterval).Duration
            elapsed := duration.MakeDuration(int64(ctx.StmtTimestamp.Sub(ctx.TxnTimestamp)), 0, 0)
            if elapsed.Compare(minDuration) < 0 {
                return nil, ctx.Txn.GenerateForcedRetryableError(
                    ctx.Ctx(), "forced by crdb_internal.force_retry()")
            }
            return tree.DZero, nil
        },
        Info: "This function is used only by CockroachDB's developers for testing purposes.",
    },
),
```

ericharmeling:
Fixes #5538.

Regenerated functions.md file, on https://github.com/cockroachdb/cockroach/releases/tag/v20.2.0-rc.3 + https://github.com/cockroachdb/cockroach/pull/55996/commits/ae19006f70604b9dc95f726498b5075dd44078ee

See https://github.com/cockroachdb/cockroach/pull/55996.

### DIFF
--- a/_includes/v20.2/sql/functions.md
+++ b/_includes/v20.2/sql/functions.md
@@ -442,9 +442,7 @@ historical reads against a time which is recent but sufficiently old for reads
 to be performed against the closest replica as opposed to the currently
 leaseholder for a given range.</p>
 <p>Note that this function requires an enterprise license on a CCL distribution to
-return a result that is less likely the closest replica. It is otherwise
-hardcoded as -4.8s from the statement time, which may not result in reading from the
-nearest replica.</p>
+return without an error.</p>
 </span></td></tr>
 <tr><td><a name="localtimestamp"></a><code>localtimestamp() &rarr; <a href="date.html">date</a></code></td><td><span class="funcdesc"><p>Returns the time of the current transaction.</p>
 <p>The value is based on a timestamp picked when the transaction starts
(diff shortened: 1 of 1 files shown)


lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8758#-:-MKfddzp-hm1K8vIwEu-:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lnhsingh)

PR #8766, opened 2020-10-27
Added section on transactional schema change limitations to Flyway tutorial

### LINKED ISSUE #7453: Flyway limitations
Following up from discussion with @rafiss, et al., we should make our limitations in Flyway support more explicit: If a migration with multiple schema changes fails at any point, the migration may be left in an incomplete partial state. Flyway can't handle rolling this back AFAIK, so manual intervention (both to determine the actual state and to fix it) would be required.

This applies even to single `ALTER TABLE` statements which include multiple schema changes, like `ALTER TABLE ... ALTER COLUMN a RENAME ..., ADD COLUMN b ...`.

There's no good way to get around this right now because we don't support transactional schema changes.

ericharmeling:
Fixes #7453.


ericharmeling:
@lucy-zhang Friendly ping on review. :)

lucy-zhang (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8766)* status: :shipit: complete! 0 of 0 LGTMs obtained

lnhsingh (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8766#-:-ML9QRWt7X62NkorKcMW:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lnhsingh)

PR #8777, opened 2020-10-28
Updates to SQL audit log docs

### LINKED ISSUE #3177: SQL Audit Logging Doc Update: Performance Implications
Re: [SQL Audit Logging](https://cockroachlabs.com/docs//v2.0/sql-audit-logging.html)

## Issue Description
We should explain what the performance penalty is for enabling this feature.

### LINKED ISSUE #3178: Document SQL audit log format
Re: [EXPERIMENTAL_AUDIT](https://cockroachlabs.com/docs//v2.0/experimental-audit.html)

## Suggested Resolution 
We should note that the log format is google's log format. So for the first item, it's not a date, it's a log level + a date. https://github.com/google/glog/blob/master/src/logging.cc#L1079

ericharmeling:
Fixes #3178.
Fixes #3177.

- Added "Performance considerations" section to 19.2, 20.1, and 20.2 EXPERIMENTAL AUDIT docs.
- Corrected log output field description in 19.2, 20.1, and 20.2 EXPERIMENTAL AUDIT docs.
- Updated example output in SQL Audit Logging pages, where output format has changed.


v19.2/experimental-audit.md
```diff
@@ -88,6 +88,26 @@ If your deployment requires particular lifecycle and access policies for audit l
 
 {% include {{ page.version.version }}/misc/schema-change-view-job.md %}
 
+## Performance considerations
+
+Enabling SQL audit logs will impact performance, as every query that causes a logging event must access the disk of the node on which audit logging is enabled.
```
knz, line 93: I think this text could do a better job of explaining there are different user stories, and that one should be careful to choose the features they need:

- from a security perspective, an "audit log" provides what is known as *non-repudiability*, the ability for an operator to know what users (especially malicious users) did with the confidence that *all* of what the user did is in the log. 

  Non-repudiability requires the writes to disk to be synchronous, because otherwise a malicious user could do something bad then immediately crash the node and then the evidence would be erased (in that case, the log is repudiable and cannot be used for audit purposes)

- separately, users may want/need a trace of all SQL statements executed, so that they can more readily troubleshoot application bugs or performance problems. These logs typically do not need to be complete, let alone non-repudiable. For that use case, the "audit" logs are overkill and detrimental (because of their performance impact).

For the security story, our current logging story is not exactly complete (we have more work to do).

For the SQL analysis / troubleshooting story, our current logging story has a bit _too many_ features -- in particular it's tempting to include the "experimental audit logging" as one of the logging systems that can be used for it, even though it was not designed for it.


Now I realize after writing all this that this general way to look at things is not the one that has been adopted in our doc around logging (yet). This may be a sign that a larger rework is in order. I do not know what you think would be the best first next step here.

ericharmeling: Excellent feedback, @knz. 

I think the best first step would be to:

- Go into a little more detail about the non-repudiation guarantee on the [`EXPERIMENTAL AUDIT` reference page](https://www.cockroachlabs.com/docs/stable/experimental-audit).

- Call out audit logging as a security feature (or as a logging feature for security), rather than a logging feature for troubleshooting in as many places as possible. 

I've added a commit that does both of these things, in the v19.2 docs (I will propagate these changes to v20.1 and v20.2 docs once I get your approval).

Note that the [SQL Audit Logging page](https://www.cockroachlabs.com/docs/stable/sql-audit-logging.html) is already listed as a [security feature in our security docs](https://www.cockroachlabs.com/docs/stable/security-overview.html#security-features-in-cockroachdb).

ericharmeling: @knz Ping on re-review (but no rush). :)

knz (approved):
> Reviewed 4 of 6 files at r1, 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8777)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v19.2/experimental-audit.md, line 93 at r1](https://reviewable.io/reviews/cockroachdb/docs/8777#-MKk0KWPEJj_FuveUlba-r1-93:-MLmJ-biA3yYk_raUoRT:bq5msbp) ([raw file](https://github.com/cockroachdb/docs/blob/d793dc32c6e0e9b30d18ecd3bc7a405f01d0ffc6/v19.2/experimental-audit.md#L93)):*

@knz Ping on re-review (but no rush). :)
</blockquote></details>

sorry I was on vacay last week.

Your proposal looks good. I like the overall changes to this PR.
___
*[v19.2/experimental-audit.md, line 97 at r2](https://reviewable.io/reviews/cockroachdb/docs/8777#-MLmIvAyCIQJcpaQKgaV:-MLmIvAyCIQJcpaQKgaW:bcjduny) ([raw file](https://github.com/cockroachdb/docs/blob/2cfb4e8bb9ffefb8baf17b3ca22f87d50f135b8b/v19.2/experimental-audit.md#L97)):*
> ```Markdown
> To ensure [non-repudiation](https://en.wikipedia.org/wiki/Non-repudiation) in audit logs, CockroachDB synchronously logs all of the activity of every user on a cluster in a way that is durable to system failures. Every query that causes a logging event must access the disk of the node on which audit logging is enabled. As a result, enabling SQL audit logs negatively impacts performance, and we recommend using SQL audit logs for security purposes only.
> 
> For debugging and troubleshooting on production clusters, the most performant way to log all queries is to turn on the [cluster-wide setting](cluster-settings.html) `sql.trace.log_statement_execute`:
> ```

I would say "For debugging and troubleshooting, ... refer to this page instead."

Then place these instructions in the page that talks about query logging. Otherwise it's a bit odd to see this piece of tehcnical advice on a page that is really about a different topic.

ericharmeling (commented):
TFTRs, @knz!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8777#-:-MLmTl6k9Bl9li2gpQYL:bjp1qew)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @knz)

___
*[v19.2/experimental-audit.md, line 93 at r1](https://reviewable.io/reviews/cockroachdb/docs/8777#-MKk0KWPEJj_FuveUlba-r1-93:-MLmLw3b7UqALIRn_Qg1:b-7f79vl) ([raw file](https://github.com/cockroachdb/docs/blob/d793dc32c6e0e9b30d18ecd3bc7a405f01d0ffc6/v19.2/experimental-audit.md#L93)):*

sorry I was on vacay last week.

Your proposal looks good. I like the overall changes to this PR.
</blockquote></details>

Ah! Sorry for the ping then! Hope you enjoyed the time off :)
___
*[v19.2/experimental-audit.md, line 97 at r2](https://reviewable.io/reviews/cockroachdb/docs/8777#-MLmIvAyCIQJcpaQKgaV:-MLmNuMZ50ojnI7zR_cw:b-ho5wpv) ([raw file](https://github.com/cockroachdb/docs/blob/2cfb4e8bb9ffefb8baf17b3ca22f87d50f135b8b/v19.2/experimental-audit.md#L97)):*

I would say "For debugging and troubleshooting, ... refer to this page instead."

Then place these instructions in the page that talks about query logging. Otherwise it's a bit odd to see this piece of tehcnical advice on a page that is really about a different topic.
</blockquote></details>

Agreed. Done.

lnhsingh (commented):
Minor nit but otherwise LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8777#-:-MLnStMC4qjobOXhkeaq:b-8eqglx)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @knz, and @lnhsingh)

___
*[v19.2/experimental-audit.md, line 56 at r3](https://reviewable.io/reviews/cockroachdb/docs/8777#-MLmxPYo7RouJDRe-OrT:-MLmxPYo7RouJDRe-OrU:b2d1tuw) ([raw file](https://github.com/cockroachdb/docs/blob/965724a2042f94b3792afb649c408e5216cae7d7/v19.2/experimental-audit.md#L56)):*
> ```Markdown
> ~~~
> 
> 1. Log level (`INFO`, `WARN`, `ERROR`, or `FATAL`) and date (in yymmdd format)
> ```

nit: `yymmdd` > `YYMMDD`

ericharmeling (commented):
TFTR @lnhsingh !

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8777#-:-MLySxL80QvfmbGDrc6u:bhjhpnt)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @knz)

___
*[v19.2/experimental-audit.md, line 56 at r3](https://reviewable.io/reviews/cockroachdb/docs/8777#-MLmxPYo7RouJDRe-OrT:-MLySwKC35JhdSZu6k6x:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/965724a2042f94b3792afb649c408e5216cae7d7/v19.2/experimental-audit.md#L56)):*

nit: `yymmdd` > `YYMMDD`
</blockquote></details>

Done.

PR #8806, opened 2020-11-02
add offline docs container

otan:


keith-mcclellan:
stashing this somewhere will be good if we ever need to do this again (we probably will)

OFFLINE-DOCS.md
```diff
@@ -0,0 +1,12 @@
+# Offline docs instructions
```
jlinder, line 1: This file only mentions the `./Dockerfile` file. What about the contents of `offline-build`?

offline-build/builder
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+image=cockroachdb/offline-docs
```
jlinder, line 5: Does this docker hub repo exist already?

OFFLINE-DOCS.md
```diff
@@ -0,0 +1,12 @@
...
+
+* Either:
+  * Build the container: `docker build -t offline-docs .`
+  * Pull the container: `docker pull offline-docs`.
```
jlinder, line 5: Should this read:

```suggestion
   * Pull the container: `docker pull cockroachdb/offline-docs`. 
```

jlinder (commented):
Why is this docker files at `Dockerfile` and  `offline-build/Dockerfile` appear to be the same. What is the intended difference?

OFFLINE-DOCS.md
otan: Done!

OFFLINE-DOCS.md
otan: I've deleted the root Dockerfile

offline-build/builder
otan: No -- I think I need you to create it.
Also, not sure about deployment process for this yet. This is copy-pasta from elsewhere.

otan:
duplicate items have been deleted :)

otan:
guess we have no takers :P

PR #8820, opened 2020-11-03
Horizontal scroll fix

### LINKED ISSUE #8056: Scrolling
The side-panel and the article section should have their own Scroll area.There should be no horizontal scrolling in these areas: Remove horizontal scrolling from content listing #6761

yingzhuchin:
Fixes #8056 

I've enabled horizontal scroll for some especially wide SQL Diagrams, but did not go through and do it for every single diagram, since that is time-consuming and I think we are moving away from diagrams in general. Let me know if we do still want to enable for every single diagram.

Example of especially wide diagrams: http://cockroach-docs-review.s3-website-us-east-1.amazonaws.com/ab24354caee902f77d6e061cfd0664b8dbc77032/v20.1/configure-zone.html


yingzhuchin:
@anagio, when the CSS is compiled here there's a lot of pure formatting changes (moving closing bracket to a new line), so the CSS diff ends up being huge. Is that fine with you? We can just review the SCSS files.

lnhsingh:
LGTM!

v2.0/create-table.md
```diff
@@ -36,13 +36,13 @@ The user must have the `CREATE` [privilege](privileges.html) on the parent datab
 
 **col_qualification ::=**
 
-<div>
+<div id="horizontal-scroll">
```
elkmaster, line 39: id ==> class

yingzhuchin: Thanks, good catch!

luciaaaaaaa:
Hi, homepage looks good to me.  I'm afraid we'll push the site live without horizontal scroll for the SQL diagrams and users wont be able to see them - they will think it's broken. 

@lnhsingh / @ericharmeling What is the timeline to remove SQL diagrams?

ericharmeling:
> What is the timeline to remove SQL diagrams?

We are targeting by end of Q4, but it could happen earlier.

CC @ianjevans

lnhsingh:
> I'm afraid we'll push the site live without horizontal scroll for the SQL diagrams and users wont be able to see them - they will think it's broken.

@luciaaaaaaa I _think_ it should be ok though since YZ did implement horizontal scroll for some of the wider SQL diagrams, just not all of them (iiuc)

yingzhuchin:
If there are specific diagrams for which you'd like to enable horizontal scrolling, just let me know.

yingzhuchin:
Based on discussion from our sync, I will merge this by end of today unless I hear otherwise!

luciaaaaaaa:
image

will the TOC and sidenav alignment issues in the screenshot be takled separately?

yingzhuchin:
> will the TOC and sidenav alignment issues in the screenshot be takled separately?

They should! This PR doesn't do anything to change sidenav alignment.

Issue #7816, opened 2020-07-29
Update "Secure" and "Insecure" language in example workflows


ericharmeling:
The workflow in many of our examples is split into "Secure" and "Insecure". We should update this language to emphasize that the "`--insecure`" `cockroach` mode is only meant for local development purposes, and should NEVER be used in production environments.

Suggested replacements:

- "Production" and "Development"
- "Secure" and "Testing"

@vy-ton @Amruta-Ranade @kenliu @thtruo 

Feel free to add to this list, or change it!

@taroface 
This will affect deployment docs as well.

vy-ton:
+1 to `Production` and `Development`

jseldess:
Thanks for tacking this, all. I think this is heading the right direction. I like using the production/testing terminology for deployment docs. I don't like that for our getting started docs and learning-centric tutorials because they are definitely not focused on production. 

Instead, I'd suggest we consider a different angle: Cloud vs. Local. Although we're not currently doing this, I think the right move is to default to having users leverage a free CC cluster. This supports our Product objective of "Make CockroachCloud the developer's database of choice" very well. The secondary option could be "Local", and there we could just provide guidance for a secure local cluster. For app dev tutorials, I think the best approach may be to just use `cockroach demo`, but I do wonder if persisting data is important to devs getting started. 

Two options is enough, but the third option we could provide very soon would be "In-Browser", leveraging the in-browser interactive tutorial approach I'm working on. 

cc @nstewart

vy-ton:
For tutorials and developer recipes, my angle is to focus on the application developer mindset when approaching certain developer tasks. Having the `secure vs insecure` distinction at the top level is misleading because oftentimes that distinction only applies to specific steps, such as connecting to the database. 

Most third-party tool tutorials have the following introduction
```
Before you begin
1. Install CockroachDB.
2. Start up a secure or insecure local cluster.
3. Choose the instructions that correspond to whether your cluster is secure or insecure:
```
For starting the cluster (2), what options should we highlight? `Cloud vs local` would make sense here. But for the remaining content on the page, there is no or little difference between `cloud vs local` so "choose the instructions" (3) no longer applies.

For docs, I hope all code samples could one-day run in-browser.

Amruta-Ranade:
+1 for `Cloud v. local`. 
@ericharmeling and I chatted about the app tutorials and making the instructions platform-agnostic:

1. Install CockroachDB.
[...]

PR #8826, opened 2020-11-04
Comment out CRC content in Python tutorials as part of #7816

ianjevans:
Per discussion with Andy, Vy, and Eric, we are removing CRC instructions until CRC Free Tier is available and just going with secure demo clusters. I'm leaving the updates Jesse made to the Python tutorials in, but commenting them out so we can revise it later.

### DIFF
--- a/_includes/v20.1/app/create-a-database.md
+++ b/_includes/v20.1/app/create-a-database.md
@@ -25,6 +25,7 @@
 
 </section>
 
+{% comment %}
 <section class="filter-content" markdown="1" data-scope="cockroachcloud">
 
 1. If you haven't already, [download the CockroachDB binary](install-cockroachdb.html).
@@ -55,3 +56,4 @@
     ~~~
 
 </section>
+{% endcomment %}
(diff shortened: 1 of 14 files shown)


ericharmeling (commented):
These changes look great! 

A couple more related things that we should probably do for this example:

- The help text in `_includes/<version>/app/python/psycopg2/example.py` could be updated so that CockroachCloud isn't mentioned. These changes should also be propagated to https://github.com/cockroachlabs/hello-world-python-psycopg2/blob/master/example.py.

- The comments in the `engine` definition in `_includes/<version>/app/python/sqlalchemy/example.py` could be updated so that CockroachCloud isn't mentioned. These changes should also be propagated to https://github.com/cockroachlabs/hello-world-python-sqlalchemy/blob/master/example.py.

-  The comments in the `engine` definition in `_includes/<version>/app/python/sqlalchemy/sqlalchemy-large-txns.py` could be updated so that CockroachCloud isn't mentioned. (This file isn't located in the https://github.com/cockroachlabs/hello-world-python-sqlalchemy repo).

- (Unrelated clean-up) The `_includes/<version>/app/txn-sample.py` file can be deleted from 19.2, 20.1, and 20.2 docs.

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8826#-:-MLJNxW0DS20di4dpN-x:b420s28)* status: :shipit: complete! 0 of 0 LGTMs obtained

ericharmeling (approved):
LGTM!

PR #8828, opened 2020-11-04
Documented the `--accept-sql-without-tls` flag

### LINKED ISSUE #8473: Support opt-in non-TLS SQL client conns in secure clusters
PR: https://github.com/cockroachdb/cockroach/pull/54198
- Backport includes:
    - 2/2 commits from "pgwire: accept non-TLS client conns safely in secure mode": https://github.com/cockroachdb/cockroach/pull/53991
    - 1/1 commits from "cli: allow SQL commands to use password authn in more cases": https://github.com/cockroachdb/cockroach/pull/53994

From release notes:

- A new experimental flag `--accept-sql-without-tls` has been introduced for [`cockroach start`](../v20.2/cockroach-start.html) and [`cockroach-start-single-node`](../v20.2/cockroach-start-single-node.html). When specified, a secure node will also accept secure SQL connections without TLS. When this flag is enabled:

  - Node-to-node connections still use TLS: the server must still be started with `--certs-dir` and valid [TLS cert configuration](../v20.2/authentication.html) for nodes. 
  - Client [authentication](../v20.2/authentication.html) (spoof protection) and [authorization](../v20.2/authorization.html) (access control and privilege escalation prevention) are performed by CockroachDB as usual, subject to the HBA configuration (for authentication) and SQL privileges (for authorization). 
  - Transport-level security (integrity and confidentiality) for client connections must then be provided by the operator outside of CockroachDB—for example, by using a private network or VPN dedicated to CockroachDB and its client app(s). 
  - The flag only applies to the SQL interface. TLS is still required for the HTTP endpoint (unless `--unencrypted-localhost-http` is passed) and for the RPC endpoint.

		To introduce this feature into an existing cluster: 

		<ol><li>Ensure the cluster ugprade is finalized.</li>
		<li>Set up the HBA configuration to reject `host` connections for any network other than the one that has been secured.</li>
		<li>Add the command-line flag and restart the nodes. Note that even when the flag is supplied, clients can still negotiate TLS and present a valid TLS certificate to identify themselves (at least under the default HBA configuration).</li></ol>
    Finally, this flag is experimental and its ergonomics will likely change in a later version. <a href="https://github.com/cockroachdb/cockroach/pull/54198">#54198</a>

Related release note:

> - It is now possible to use password authentication over non-TLS connections with `cockroach` client CLI commands that use only SQL, e.g., [`cockroach sql`](../v20.2/cockroach-sql.html) or [`cockroach node ls`](../v20.2/cockroach-node.html). [#54198][#54198]

Amruta-Ranade:
Closes #8473


bdarnell (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8828)* status: :shipit: complete! 0 of 0 LGTMs obtained

bdarnell:
If we're going to document this option, we should probably also talk about `sslmode=require`. That's going to be more secure than `--accept-sql-without-tls` and has the same advantage of not needing a CA certificate. It's not clear to me when you'd prefer to turn off TLS completely instead of just turning off CA verification. (@knz, any thoughts?)

Amruta-Ranade:
> If we're going to document this option, we should probably also talk about `sslmode=require`. That's going to be more secure than `--accept-sql-without-tls` and has the same advantage of not needing a CA certificate. It's not clear to me when you'd prefer to turn off TLS completely instead of just turning off CA verification. (@knz, any thoughts?)

If `sslmode=require` and password is a better option, should we even document this new flag?

knz:
1. in cases where users *can* use TLS at all, then `sslmode=require` (or `verify-ca`) with password is strictly better than no TLS. This was true as early as CockroachDB v1.1 when we started supporting passwords.

2. the reason why this feature exists is that some users have explicitly said they *cannot* (or do not want to) use TLS *at all*. In those cases, there was no solution until v20.2, and now there is. 

I don't know if we want to spend a lot of documentation space for this, however I agree with the callout: if TLS is acceptable, users should use it and sslmode=require is strictly better than sslmode=disable (and sslmode=verify-ca is even better, even if they don't use client certs).

knz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8828)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @Amruta-Ranade and @rmloveland)

___
*[v20.2/authentication.md, line 71 at r1](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmGw_I2j6Cn5KcN_Ot:-MLmGw_I2j6Cn5KcN_Ou:b-krtdc0) ([raw file](https://github.com/cockroachdb/docs/blob/6f768f3bf9d91bed97c8131d8a97e96c041ce775/v20.2/authentication.md#L71)):*
> ```Markdown
>    Note that the client still needs the CA certificate to validate the nodes' certificates.
> 
>    A less secure option is to set the experimental [`--accept-sql-without-tls` flag](cockroach-start.html#security) when starting the cluster. Setting the flag allows you to connect to the cluster without validating the nodes' certificates.
> ```

I would make this a separate section, and explain as follows:

```
NEW IN v20.2

For deployments where transport security is already handled at the infrastructure level (e.g. IPSec with DMZ), and TLS-based transport security is not possible or not desirable, CockroachDB now supports delegating transport security to the infrastructure with the new flag `--accept-sql-without-tls`.

With this flag, SQL clients can establish a session over TCP without a TLS handshake. They still need to present valid authentication credentials, for example a password in the default configuration. Different authentication schemes can be further configured as per `server.host_based_authentication.configuration`.
```

(No need to say "less secure option". Instead outline that in this case the responsibility for security is transferred from cockroachdb into the infrastructure.)

Amruta-Ranade:
> 1. in cases where users _can_ use TLS at all, then `sslmode=require` (or `verify-ca`) with password is strictly better than no TLS. This was true as early as CockroachDB v1.1 when we started supporting passwords.
> 2. the reason why this feature exists is that some users have explicitly said they _cannot_ (or do not want to) use TLS _at all_. In those cases, there was no solution until v20.2, and now there is.
> 
> I don't know if we want to spend a lot of documentation space for this, however I agree with the callout: if TLS is acceptable, users should use it and sslmode=require is strictly better than sslmode=disable (and sslmode=verify-ca is even better, even if they don't use client certs).

In that case, we should change "This experimental flag allows you to connect to the cluster using a SQL user's password without [validating the nodes' certificates](authentication.html#client-authentication)." > "This experimental flag allows you to connect to the cluster using a SQL user's password without [using TLS authentication](authentication.html)."

Amruta-Ranade (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8828)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @knz and @rmloveland)

___
*[v20.2/authentication.md, line 71 at r1](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmGw_I2j6Cn5KcN_Ot:-MLmZxaY4NhFiqET9yns:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/6f768f3bf9d91bed97c8131d8a97e96c041ce775/v20.2/authentication.md#L71)):*

I would make this a separate section, and explain as follows:

```
NEW IN v20.2

For deployments where transport security is already handled at the infrastructure level (e.g. IPSec with DMZ), and TLS-based transport security is not possible or not desirable, CockroachDB now supports delegating transport security to the infrastructure with the new flag `--accept-sql-without-tls`.

With this flag, SQL clients can establish a session over TCP without a TLS handshake. They still need to present valid authentication credentials, for example a password in the default configuration. Different authentication schemes can be further configured as per `server.host_based_authentication.configuration`.
```

(No need to say "less secure option". Instead outline that in this case the responsibility for security is transferred from cockroachdb into the infrastructure.)
</blockquote></details>

Done.

knz (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8828)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @Amruta-Ranade and @rmloveland)

___
*[v20.2/cockroach-start.md, line 94 at r2](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmb_bo54ASKNOyyLnM:-MLmb_bp9u5oS1wfjWSS:bv3c35o) ([raw file](https://github.com/cockroachdb/docs/blob/9ffb242dbc70f65343b09b2b805231536df62858/v20.2/cockroach-start.md#L94)):*
> ```Markdown
> `--certs-dir` | The path to the [certificate directory](cockroach-cert.html). The directory must contain valid certificates if running in secure mode.<br><br>**Default:** `${HOME}/.cockroach-certs/`
> `--insecure` | **Note:** The `--insecure` flag is intended for non-production testing only. It **will be deprecated** and replaced with new secure options in v21.1.<br><br>Run in insecure mode. If this flag is not set, the `--certs-dir` flag must point to valid certificates.<br><br><strong>Note the following risks:</strong> An insecure cluster is open to any client that can access any node's IP addresses; any user, even `root`, can log in without providing a password; any user, connecting as `root`, can read or write any data in your cluster; and there is no network encryption or authentication, and thus no confidentiality.<br><br>**Default:** `false`
> `--accept-sql-without-tls` | This experimental flag allows you to connect to the cluster using a SQL user's password without [validating the nodes' certificates](authentication.html#client-authentication). When connecting to the cluster using a [connection string](connection-parameters.html), you can omit the `sslmode` parameter from the string. When connecting using the built-in SQL client, [use the `--insecure` flag with the `cockroach sql` statement](cockroach-sql.html#client-connection).
> ```

1. without validating the **client**'s certificate.

2. please remove "when connecting using the built-in SQL client..."; that is in fact incompatible.
___
*[v20.2/cockroach-start-single-node.md, line 67 at r2](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmbmCn1dIeWvt5UsWF:-MLmbmCo0VnDolDJnYSA:b-7ropln) ([raw file](https://github.com/cockroachdb/docs/blob/9ffb242dbc70f65343b09b2b805231536df62858/v20.2/cockroach-start-single-node.md#L67)):*
> ```Markdown
> `--certs-dir` | The path to the [certificate directory](cockroach-cert.html). The directory must contain valid certificates if running in secure mode.<br><br>**Default:** `${HOME}/.cockroach-certs/`
> `--insecure` | Run in insecure mode. If this flag is not set, the `--certs-dir` flag must point to valid certificates.<br><br><strong>Note the following risks:</strong> An insecure cluster is open to any client that can access any node's IP addresses; any user, even `root`, can log in without providing a password; any user, connecting as `root`, can read or write any data in your cluster; and there is no network encryption or authentication, and thus no confidentiality.<br><br>**Default:** `false`
> `--accept-sql-without-tls` | This experimental flag allows you to connect to the cluster using a SQL user's password without [validating the nodes' certificates](authentication.html#client-authentication). When connecting to the cluster using a [connection string](connection-parameters.html), you can omit the `sslmode` parameter from the string. When connecting using the built-in SQL client, [use the `--insecure` flag with the `cockroach sql` statement](cockroach-sql.html#client-connection).
> ```

ditto

Amruta-Ranade (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8828)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @knz and @rmloveland)

___
*[v20.2/cockroach-start.md, line 94 at r2](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmb_bo54ASKNOyyLnM:-MLmesqw1bKKM-9NSWq0:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/9ffb242dbc70f65343b09b2b805231536df62858/v20.2/cockroach-start.md#L94)):*

1. without validating the **client**'s certificate.

2. please remove "when connecting using the built-in SQL client..."; that is in fact incompatible.
</blockquote></details>

Done.
___
*[v20.2/cockroach-start-single-node.md, line 67 at r2](https://reviewable.io/reviews/cockroachdb/docs/8828#-MLmbmCn1dIeWvt5UsWF:-MLmetS60zCEjTTOCCHI:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/9ffb242dbc70f65343b09b2b805231536df62858/v20.2/cockroach-start-single-node.md#L67)):*

ditto
</blockquote></details>

Done.

v20.2/cockroach-start-single-node.md
```diff
@@ -64,6 +64,7 @@ Flag | Description
 -----|-----------
 `--certs-dir` | The path to the [certificate directory](cockroach-cert.html). The directory must contain valid certificates if running in secure mode.<br><br>**Default:** `${HOME}/.cockroach-certs/`
 `--insecure` | Run in insecure mode. If this flag is not set, the `--certs-dir` flag must point to valid certificates.<br><br><strong>Note the following risks:</strong> An insecure cluster is open to any client that can access any node's IP addresses; any user, even `root`, can log in without providing a password; any user, connecting as `root`, can read or write any data in your cluster; and there is no network encryption or authentication, and thus no confidentiality.<br><br>**Default:** `false`
+`--accept-sql-without-tls` | This experimental flag allows you to connect to the cluster using a SQL user's password without [validating the client's certificate](authentication.html#client-authentication). When connecting using the built-in SQL client, [use the `--insecure` flag with the `cockroach sql` statement](cockroach-sql.html#client-connection).
```
jseldess, line 67: nit: `cockroach sql` statement > `cockroach sql` command

Also, if this is experimental, it should be added to the experimental features page and also called out as such on the authentication page.

@Amruta-Ranade, can you handle this as a follow-up?

PR #8850, opened 2020-11-07
Fix #4477 List WITH under SQL statements 

### LINKED ISSUE #4477: List WITH under SQL statements
Re: [SQL Statements](https://cockroachlabs.com/docs//v2.1/sql-statements.html)

## Issue Description

The WITH statement isn't listed on this page or the SQL statements section of the sidebar. This sometimes [leads users to believe that we don't support it](https://stackoverflow.com/questions/55047689/sum-of-differences-over-a-jsonb-array-field-in-cockroachdb). Instead, it's documented under SQL Syntax -> Common Table Expressions, which is not necessarily a name that users know. 

## Suggested Resolution 

Refer to the WITH statement (by that name, not just as CTE) in the lists of statements. It's not quite a regular statement but it looks enough like one that that's how users think of it.

Shiwanee:
Moved the topic Common Table expressions from under SQL Syntax to SQL Statements and updated it to include WITH


jseldess:
Thanks for this work, @Shiwanee, and sorry for the delay.

@ericharmeling, can you PTAL and, if necessary, port these changes to a new PR to get it over the finish line?

ericharmeling (commented):
@shiwanee Thanks for doing this. I really like these changes. This makes our WITH documentation more consistent with [the PostgreSQL docs](https://www.postgresql.org/docs/current/queries-with.html).

I have just a couple of comments regarding style, but I can handle those on my end. I'm going to have to port your changes to a new PR anyways, because there are now conflicts with your forked branch and the master branch.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8850#-:-MkbZKz8BAnlgb_fMpcW:bajihdw)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @Shiwanee)

PR #8863, opened 2020-11-09
rename Admin UI -> DB Console

taroface:
Renamed Admin UI to DB Console in v20.2 docs.

- "Admin UI" is now called "DB Console" in all docs
- URLs and image filenames replace `admin-ui` with `ui` (which is more generic and future-proof)
- Added redirects from "Admin UI" to "DB Console" pages.


Amruta-Ranade (approved):
LGTM! 🎉

thtruo (approved):
Nice one @taroface !!! All outstanding issues mentioned on Slack have been addressed. This LGTM!

PR #8832, opened 2020-11-04
Add jekyll-remote-includes plugin. Fixes #7500

ianjevans:

PR #8864, opened 2020-11-09
Use jekyll-remote-include plugin

### LINKED ISSUE #7500: Enable remote file includes in jekyll build
Related to #7066. 

Jekyll [requires all included files to be local](https://jekyllrb.com/docs/includes/). We should look into make a custom plugin that enables including remote files. This would make including code samples from external repos (with automated testing!) possible

Here is a discussion on this topic, with some possible solutions to work off of:
https://stackoverflow.com/questions/14958294/including-external-files-in-a-jekyll-template

yingzhuchin:
Fixes #7500 

Related to https://github.com/cockroachdb/docs/pull/8832 (see conversation there)


ericharmeling (approved):
This looks great! Thanks @yingzhuchin. This is going to be very useful for sample-based developer tutorials moving forward.

We'll need to finish the work consolidating the hello-world tutorial workflows before we can propagate the remote include changes to other hello-world tutorials with corresponding, stand-alone repos. Currently, most of the tutorials include two separate code samples (one for secure and one for insecure).

@jseldess and @ianjevans got the psycopg2 tutorial in order, and I believe @ianjevans is working on Java right now. Next up would be golang, node, and ruby.

I'll open an issue to track other work (Spring, multi-region, etc) that could benefit from these changes.

yingzhuchin:
> This looks great! Thanks @yingzhuchin. This is going to be very useful for sample-based developer tutorials moving forward.

Great! Thanks, @ianjevans, for pinpointing the plugin needed! If we are going to merge the change to the Python tutorial, I think we should delete all `_includes/<version>/app/python/psycopg2/` folders as well. I'm happy to do that and merge. Just let me know.

ericharmeling:
> I think we should delete all `_includes/<version>/app/python/psycopg2/` folders as well (unless they are used somewhere else also?). 

That would be great! Thanks! I don't think they are included elsewhere, but if they are, the TeamCity build will fail and let us know before we can merge.

jseldess:
This is really exciting progress. Thank you, @yingzhuchin, @ianjevans, @ericharmeling!

PR #8868, opened 2020-11-10
Update Vale to run on latest stable

yingzhuchin:
@rmloveland, any chance you would consider adding this step to `bump_major_version.py`? 😄


jseldess:
LGTM. Thank you, @yingzhuchin! 

I know we need to figure out whether we want to take this linting seriously or not.

PR #8871, opened 2020-11-10
Fix a small nit in 20.2 only

dbist:


taroface:
LGTM, thanks @dbist .

taroface (approved):

PR #8878, opened 2020-11-10
Fix error on CockroachDB in Comparison page

yingzhuchin:
On [this page](https://www.cockroachlabs.com/docs/dev/cockroachdb-in-comparison.html), Spanner values are mistakenly shown for Yugabyte. 

Not sure who should review, but GitHub recommended Rich =)


jseldess (approved):
LGTM

Thanks for the fix. This page is owned by @jaymce and Product Marketing, but this is an obvious error.

PR #8879, opened 2020-11-10
Add back homebrew install instructions for 20.2

### LINKED ISSUE #8858: Test and update homebrew install instructions for 20.2 release
Due to the big changes in our install procedure to accomodate adding shared libraries that enable [spatial feature support](https://www.cockroachlabs.com/docs/v20.2/spatial-features.html), we need to verify that `brew install` still works properly for users before adding it back into the docs post-20.2.0.

(It is probably going to be deleted from the install instructions for the immediate 20.2.0 release, since we have not tested it with `brew` at all yet as of Monday, November 9, 2020).

Estimated scope of work for this issue:

- Test whether installing CockroachDB via brew has working spatial support.  This will entail installing it with brew after the 20.2.0 release drops, and seeing if it works.
- If it works, add the `brew` installation docs back
- If it does not work, file an issue against CockroachDB to get the `brew` installation of spatial stuff working

See also:

- cockroachdb/cockroach#56435

rmloveland:
Fixes #8858.

Summary of changes:

- Uncomment homebrew install instructions

- Update instructions to mention that this is now v20.2 (not 19.2)

- Note that I manually tested the `brew install`-ed CockroachDB and
  verified that spatial features work.

### DIFF
--- a/v20.2/install-cockroachdb-mac.html
+++ b/v20.2/install-cockroachdb-mac.html
@@ -16,14 +16,6 @@ key: install-cockroachdb.html
 
 {% include cockroachcloud/use-cockroachcloud-instead.md %}
 
-<!--
-
-Commented out by Rich Loveland on Monday, November 9, 2020
-
-We will add this back once the issues around the 20.2.0 brew installation for spatial libraries are resolved.
-
-For details, see https://github.com/cockroachdb/docs/issues/8858
-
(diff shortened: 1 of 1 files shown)


rmloveland:
Hopefully a trivial review, just uncommenting some existing content that I had commented out until the `brew install` path was working again.

ericharmeling (approved):
LGTM!

PR #8890, opened 2020-11-11
Use updated 404 page on v20.2

yingzhuchin:
We were using the updated 404 page on v20.1. We now need to use it in v20.2 since Netlify always redirects 404s to "stable" version.

Related to #8082


jseldess (approved):
LGTM

lnhsingh (approved):

PR #8893, opened 2020-11-11
update default resources

taroface:
Update documented CPU/memory requests & limits to address https://github.com/cockroachdb/cockroach-operator/issues/234. Verified that these settings work with `n1-standard-4`.


piyush-singh (approved):
LGTM

PR #8979, opened 2020-11-20
Recommend disabling swap

### LINKED ISSUE #7969: Recommend disabling swap on production machines
https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html

In the "CPU and memory" hardware requirements section, it would be useful to add something like the following:

> We recommend minimizing the use of virtual memory swapfiles when deploying CockroachDB into a production environment.  CockroachDB manages its own caches (configured via the `--cache` and `--max-sql-memory` flags), independently of the operating system.  Over-allocating memory on production machines can lead to unexpected (and hard to reproduce) performance hits when memory pages have to be read back into memory.  Disabling swap can be accomplished by executing `sudo swapoff --all` or removing swap devices from `/etc/fstab`.  Operators that wish to leave swap enabled are encouraged to set the `vm.swappiness` sysctl to 0, in order to minimize opportunistic page-outs.

jseldess:
Fixes #7969.

### DIFF
--- a/v19.2/recommended-production-settings.md
+++ b/v19.2/recommended-production-settings.md
@@ -42,7 +42,7 @@ Before deploying to production, test and tune your hardware setup for your appli
 
 Each node should have **at least 2 vCPUs**. For best performance, we recommend at least 4 vCPUs per node. Provision **4 GiB of RAM per vCPU**.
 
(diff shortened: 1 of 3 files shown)


nvanbenschoten (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/8979#-:-MMbldFg3khkrI_Fj1AX:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @jseldess)

PR #8984, opened 2020-11-22
Fix command to decommission a dead node

### LINKED ISSUE #8865: 20.2 decommission dead node example is incorrect
https://www.cockroachlabs.com/docs/v20.2/remove-nodes.html#step-1-mark-the-dead-node-as-decommissioned

The example shown is incorrect:
```
cockroach node decommission --self --certs-dir=certs --host=<address of node to decommission>
```

That would try to connect to the dead node and decommission itself. But since the node is dead, it will fail to connect. Instead, the command should be like this:

```
cockroach node decommission <id of the dead node> --certs-dir=certs --host=<address of an alive node>
```

jseldess:
Fixes #8865.


sheaffej:
LGTM

Issue #7957, opened 2020-08-09
[Security] Bump JDBC driver recommendation to >= 42.2.13


bobvawter:
We recommend 42.2.12 as of this date.  The .13 release includes a fix for CVE-2020-13692, which is a pretty nifty attack that can be exploited by having the pg driver decode XML.

ericharmeling:
@bobvawter FYI, we just updated our docs to recommend the JDBC driver version we use in our CI nightlies: https://github.com/cockroachdb/docs/pull/9008

On 12/1/20, we were using 42.2.9. It would be nice to recommend a fixed version that we test (see https://github.com/cockroachdb/docs/pull/9008#discussion_r533490943). Do you have a recommendation for a specific version?

rmloveland:
Bump to P-1 b/c it's a security thing and we don't want this to hang out

rafiss:
We plan to do this in February. It will involve updating our roachtest nighty test to use this version.

rafiss:
I discussed with @ericharmeling that the docs can automatically pull the version number from here: https://github.com/cockroachdb/examples-orms/blob/f3a95b3a6deeff828f745f856b739e028118851a/java/hibernate/build.gradle#L16

ericharmeling:
This is resolved by https://github.com/cockroachdb/docs/pull/10756.

We pull the versions now from [`roachtest`](https://github.com/cockroachdb/cockroach/tree/master/pkg/cmd/roachtest) now. The latest version we've tested is 42.2.19.

PR #9008, opened 2020-11-30
Liquibase v4.2

### LINKED ISSUE #8957: Document Liquibase full support and new version
Recommended version should be `4.2.1` once the Liquibase team releases it.

* Call out that Liquibase transactions have been disabled

ericharmeling:
Fixes #8957.

### DIFF
--- a/_includes/v20.1/app/hibernate-basic-sample/build.gradle
+++ b/_includes/v20.1/app/hibernate-basic-sample/build.gradle
@@ -12,5 +12,5 @@ repositories {
 
 dependencies {
     implementation 'org.hibernate:hibernate-core:5.4.19.Final'
-    implementation 'org.postgresql:postgresql:42.2.14'
+    implementation 'org.postgresql:postgresql:42.2.9'
 }
--- a/v20.2/liquibase.md
+++ b/v20.2/liquibase.md
@@ -13,7 +13,7 @@ For detailed information about using Liquibase, see the [Liquibase documentation
 Before you begin the tutorial, do the following:
 
 1. [Install CockroachDB](install-cockroachdb.html), and [start a secure cluster](secure-a-cluster.html). When starting your cluster, make sure that you generate cluster certificates, create the `bank` database, and create the `max` user.
-1. Download and install a Java Development Kit. Liquibase supports Java 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8.
+1. Download and install a Java Development Kit. Liquibase supports JDK versions 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8, but you can follow along with any JDK version 11+.
 
 ## Step 1. Download and install Liquibase
 
@@ -98,7 +98,7 @@ To install the driver for Liquibase:
 
     {% include copy-clipboard.html %}
     ~~~ shell
-    $ cp ~/Downloads/postgresql-42.2.18.jar liquibase-4.2.0-bin/lib/
+    $ cp ~/Downloads/postgresql-42.2.9.jar liquibase-4.2.0-bin/lib/
     ~~~
 
 {{site.data.alerts.callout_success}}
@@ -238,6 +238,7 @@ To configure Liquibase properties:
 
 1. Add the following property definitions to the file:
 
+    {% include copy-clipboard.html %}
     ~~~ yml
     changeLogFile: changelog-main.xml
     driver: org.postgresql.Driver
--- a/v20.2/liquibase.md
+++ b/v20.2/liquibase.md
@@ -13,7 +13,7 @@ For detailed information about using Liquibase, see the [Liquibase documentation
 Before you begin the tutorial, do the following:
 
 1. [Install CockroachDB](install-cockroachdb.html), and [start a secure cluster](secure-a-cluster.html). When starting your cluster, make sure that you generate cluster certificates, create the `bank` database, and create the `max` user.
-1. Download and install a Java Development Kit. Liquibase supports JDK versions 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8, but you can follow along with any JDK version 11+.
+1. Download and install a Java Development Kit. Liquibase supports JDK versions 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8, but you can follow along with any JDK version 8+.
 
 ## Step 1. Download and install Liquibase
 
@@ -515,4 +515,3 @@ If you run into problems, please file an issue on the [Liquibase issue tracker](
 + [Learn CockroachDB SQL](learn-cockroachdb-sql.html)
 + [Build a Spring App with CockroachDB and JDBC](build-a-spring-app-with-cockroachdb-jdbc.html)
 + [Build a Spring App with CockroachDB and JPA](build-a-spring-app-with-cockroachdb-jpa.html)
-
--- a/v20.2/liquibase.md
+++ b/v20.2/liquibase.md
@@ -6,8 +6,6 @@ toc: true
 
 This page walks you through a series of simple database schema changes using the [Liquibase](https://www.liquibase.org/get-started/how-liquibase-works) command-line tool and the [CockroachDB SQL shell](cockroach-sql.html).
 
-We have tested Liquibase enough to claim **beta-level** support.
-
 For detailed information about using Liquibase, see the [Liquibase documentation site](https://docs.liquibase.com/home.html).
 
 ## Before you begin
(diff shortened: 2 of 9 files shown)


v20.2/liquibase.md
```diff
@@ -6,22 +6,20 @@ toc: true
...
-
 For detailed information about using Liquibase, see the [Liquibase documentation site](https://docs.liquibase.com/home.html).
 
 ## Before you begin
 
 Before you begin the tutorial, do the following:
 
 1. [Install CockroachDB](install-cockroachdb.html), and [start a secure cluster](secure-a-cluster.html). When starting your cluster, make sure that you generate cluster certificates, create the `bank` database, and create the `max` user.
-1. Download and install a Java Development Kit. Liquibase supports Java 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 11.
+1. Download and install a Java Development Kit. Liquibase supports Java 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8.
```
vy-ton, line 16: Is there a reason we lowered to JDK 8?

ericharmeling: JDK 8 is much more popular than JDK 11, so it's likely than developers looking at our docs will be using JDK 8. Also, we use JDK 8 in both of our Spring tutorials. So it would be good to be consistent across our docs tutorials. Also, Liquibase docs point to a JDK 8 installation: https://docs.liquibase.com/concepts/installation/installation-linux-unix-mac.html.

So I think JDK 8 would be a good move for this tutorial. But I'm not super familiar with the Java community. @ianjevans and/or @kai-niemi might have opinions on the topic as well, seeing as they have both have a lot of experience with the Java community.

kai-niemi: I think recommending "JDK 1.8 (or later)" for demo's is best for both consistency and compability. JDK 8 is still widely adopted and offers better compability with the rest of the community frameworks/libraries (like Spring, Hibernate, Liquibase etc) for years to come. A lot of things happened after JDK 8 both in terms of the platform and also Oracle's re-licensing. While both the language and platform features have evolved significantly since JDK8, none of the demos depend on that (little reason to). Hence you could use JDK11+ and still run all demos but its not mandatory.

vy-ton:
> To avoid running into issues with incomplete transactions, we recommend setting the `runInTransaction` attribute on each of your changesets to `"false"`, as demonstrated throughout this tutorial.

We've disabled transactions by default so users no longer need to set this

v20.2/liquibase.md
```diff
@@ -100,7 +98,7 @@ To install the driver for Liquibase:
 
     {% include copy-clipboard.html %}
     ~~~ shell
-    $ cp ~/Downloads/postgresql-42.2.6.jar liquibase-3.10.2-bin/lib/
+    $ cp ~/Downloads/postgresql-42.2.18.jar liquibase-4.2.0-bin/lib/
```
vy-ton, line 101: Just checked our CI nightlies, which runs against `42.2.9`. Can we be consistent and updated the pgjdbc recommended version to be in sync?

v20.2/liquibase.md
```diff
@@ -243,7 +241,7 @@ To configure Liquibase properties:
     ~~~ yml
```
vy-ton, line 241: Missing `Copy` clipboard

vy-ton:
Reviewed, looks great Eric!

ericharmeling (commented):
TFTR @vy-ton !

@ianjevans This should be good for writer review. :)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9008#-:-MNUkjLT8-JhP9UTEi-e:b-ra54qq)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans and @vy-ton)

___
*[v20.2/liquibase.md, line 16 at r1](https://reviewable.io/reviews/cockroachdb/docs/9008#-MNPdhxwkidcEsNjwzAs-r1-16:-MNUm-P20ZL7KfKJ9Wis:b4zle20) ([raw file](https://github.com/cockroachdb/docs/blob/86f02f5525f19783945cb1ca2c76acd74c7a5c25/v20.2/liquibase.md#L16)):*
<details><summary><i>Previously, kai-niemi (Kai Niemi) wrote…</i></summary><blockquote>

I think recommending "JDK 1.8 (or later)" for demo's is best for both consistency and compability. JDK 8 is still widely adopted and offers better compability with the rest of the community frameworks/libraries (like Spring, Hibernate, Liquibase etc) for years to come. A lot of things happened after JDK 8 both in terms of the platform and also Oracle's re-licensing. While both the language and platform features have evolved significantly since JDK8, none of the demos depend on that (little reason to). Hence you could use JDK11+ and still run all demos but its not mandatory.
</blockquote></details>

Thanks Kai! Updated the text here a bit. This is good guidance for other Java docs as well.
___
*[v20.2/liquibase.md, line 101 at r1](https://reviewable.io/reviews/cockroachdb/docs/9008#-MNPdhxwkidcEsNjwzAs-r1-101:-MNUn7kJCdiuuqEOSsnJ:b-prxzon) ([raw file](https://github.com/cockroachdb/docs/blob/86f02f5525f19783945cb1ca2c76acd74c7a5c25/v20.2/liquibase.md#L101)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Just checked our CI nightlies, which runs against `42.2.9`. Can we be consistent and updated the pgjdbc recommended version to be in sync?
</blockquote></details>

Changed to 42.2.9 here, and on the tooling page, the client driver page, and the tutorial page.

Is there a reason why we are using 42.2.9 in our CI nightlies (like are there any known bugs in the later versions)? We have recommended using the latest current version on our docs page for a while now.
___
*[v20.2/liquibase.md, line 241 at r1](https://reviewable.io/reviews/cockroachdb/docs/9008#-MNPdhxwkidcEsNjwzAs-r1-241:-MNUn5-90o8piiXx6lUj:b-o7vuhc) ([raw file](https://github.com/cockroachdb/docs/blob/86f02f5525f19783945cb1ca2c76acd74c7a5c25/v20.2/liquibase.md#L241)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Missing `Copy` clipboard
</blockquote></details>

Good catch! Done.

v20.2/liquibase.md
```diff
@@ -6,22 +6,20 @@ toc: true
...
-
 For detailed information about using Liquibase, see the [Liquibase documentation site](https://docs.liquibase.com/home.html).
 
 ## Before you begin
 
 Before you begin the tutorial, do the following:
 
 1. [Install CockroachDB](install-cockroachdb.html), and [start a secure cluster](secure-a-cluster.html). When starting your cluster, make sure that you generate cluster certificates, create the `bank` database, and create the `max` user.
-1. Download and install a Java Development Kit. Liquibase supports Java 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 11.
+1. Download and install a Java Development Kit. Liquibase supports JDK versions 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8, but you can follow along with any JDK version 11+.
```
ianjevans, line 16: A little confusing. Wouldn't it be?
```suggestion
1. Download and install a Java Development Kit. Liquibase supports JDK versions 8+. In this tutorial, we use [AdoptOpenJDK](https://adoptopenjdk.net/) 8, but you can follow along with any JDK version 8+.
```

_includes/v20.1/app/hibernate-basic-sample/build.gradle
```diff
@@ -12,5 +12,5 @@ repositories {
 
 dependencies {
     implementation 'org.hibernate:hibernate-core:5.4.19.Final'
-    implementation 'org.postgresql:postgresql:42.2.14'
+    implementation 'org.postgresql:postgresql:42.2.9'
```
ianjevans, line 15: FWIW, the latest version is 42.2.18, and that's what I used in the Java hello world examples.

ianjevans (approved):
Approved with some potential minor change suggestions.

_includes/v20.1/app/hibernate-basic-sample/build.gradle
ericharmeling: Right on. I think we just want to make sure we're in sync with whichever version the official test suite is running, as that's going to be quite a bit more rigorous than our example apps. I ran this Liquibase project on 4.2.18 as well without issues. And I think we've used 42.2.14 elsewhere with no issues.

@rafiss Perhaps we should look into updating the JDBC driver that we are using on our tests? (unless there's a specific reason why we are using 42.2.9)

ericharmeling:
TFTR @ianjevans !

rafiss:
The version we test with can change -- the only reason it's pinned is so that the code/tests don't change and cause instability unless we explicitly bump the version. our goal is to keep those versions up to date at least once per release.

ericharmeling:
@rafiss Per https://github.com/cockroachdb/docs/issues/7957, it looks like 42.2.13 might be a good start.

PR #9012, opened 2020-12-01
Remove duplicate release note from v20.2.2 notes

### LINKED ISSUE #9011: Duplicate release note item found between 20.2.1 and 20.2.2
We were alerted by a user to the fact that there is a duplicate issue in both the v20.2.1 and v20.2.2 release notes:

> is there a reason that the encryption at rest issue (#56678) is in the release notes for both v20.2.1 and v20.2.2 ?

I have the following hypotheses about how this happened:

1. Itâs possible that one of the SHAs I used to generate the v20.2.2 release notes () was not the right one, though I did use the one that (AFAICT) was marked as the v20.2.1 release.  However we have to find this SHA manually by looking at our github repo's git tags page so it's highly possible I made an error.

2. Itâs also possible (but Iâm not sure?) that some SHA we release as version 20.x.y is not necessarily guaranteed to be the exact SHA that is tagged in Github as release-20.x.y, because it gets changed or not re-tagged or whatever due to some extraordinary circumstances (or even a simple SHA change due to finding a bug - don't really know for sure).  Running our release notes script on the wrong "previous SHA" will result in incorrect output.

In either case I should have confirmed the previous release's SHA with the release manager.

The fix in this case is to remove the line mentioning cockroachdb issue 56678.

(To make sure I found this and other such issues in these release notes, I also wrote a script that can be used to do this check in future, if needed: https://gist.github.com/rmloveland/d429c0af36ef650d1219eb232b0eb4f9)

rmloveland:
Fixes #9011.


taroface (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9012#-:-MNTwLnV7U3uMOtQtaaJ:b-uwkowd)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @rmloveland)

PR #9050, opened 2020-12-03
COPY FROM docs

### LINKED ISSUE #9022: sql: Document COPY statement
CockroachDB has supported the COPY statement for the past few releases, but it is not documented.

Code changes:
https://github.com/cockroachdb/cockroach/pull/8756
https://github.com/cockroachdb/cockroach/pull/51890

Other docs issues:
https://github.com/cockroachdb/docs/issues/546
#3058 
https://github.com/cockroachdb/docs/issues/1676
https://github.com/cockroachdb/docs/issues/1831
https://github.com/cockroachdb/docs/issues/8156
https://github.com/cockroachdb/docs/issues/8155

ericharmeling:
Fixes #9022.

Please review this PR carefully for accuracy.

This statement was fairly difficult to document, for the following reasons:

- We don't support `COPY FROM` in the CockroachDB SQL shell. As a result, I had to use a different client (`psql`) to create a simple example. And I'm not sure how useful this example is...
- We don't support most of the PostgreSQL [`COPY FROM` syntax](https://www.postgresql.org/docs/current/sql-copy.html) (e.g., `WHERE`, `PROGRAM`, etc.). We have several different issues tracking some of the individual, unsupported options and clauses (e.g., https://github.com/cockroachdb/cockroach/issues/54580), but we don't have all of them, so it's a little hard to point to tracking issues or to create "known limitations" for syntax that we might not intend to support. There is a catch-all issue (https://github.com/cockroachdb/cockroach/issues/41608), but it only lists some of the unsupported syntax.
- We recommend *not* using `COPY FROM` to import data into CockroachDB, which makes these docs a little superfluous.
- Some of the `COPY FROM` syntax that we allow doesn't really make sense in the context of copying data *into* CockroachDB. For example, for `COPY FROM ... DESTINATION = filename`, I'm not sure what `DESTINATION = filename` refers to, and it's not documented in the Postgres docs, so I omitted the option.
- I wasn't able to copy data from a file using `COPY FROM` in the `psql` client. Do we support copying from files?
- There is no help text, and I couldn't really find any examples, so my example and the syntax documented here is based on https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/parser/sql.y#L2861.

### DIFF
--- a/v20.2/copy-from.md
+++ b/v20.2/copy-from.md
@@ -1,0 +1,122 @@
+---
+title: COPY FROM
+summary: Copy data from a third-party client to CockroachDB.
+toc: true
+---
+
+The `COPY FROM` statement copies data from third-party clients to tables in your cluster.
+
+{{site.data.alerts.callout_info}}
+CockroachDB currently only supports `COPY FROM` statements issued from third-party clients, for compatibility with PostgreSQL drivers and ORMs. `COPY FROM` statements cannot be issued from the [`cockroach` SQL shell](cockroach-sql.html). To copy data from a file to your cluster, we recommend using an [`IMPORT`](import.html) statement instead.
+{{site.data.alerts.end}}
+
+## Syntax
+
+~~~
+COPY table_name [( <colnames...> )] FROM STDIN [WITH <option> = <value>]
+~~~
+
+### Parameters
+
+Parameter | Description
+-----------|-------------
+`table_name` | The name of the table to which to copy data.
+`<colnames>` | The name(s) of the column(s) to which to copy data.
+`WITH <option> = value` | Specify one or more [copy options](#options).
+
+### Options
+
+Option | Description
+-----------|-------------
+`BINARY` | Copy data from binary format.<br>If not specified, CockroachDB copies in plaintext format. Note that CSV format is not yet supported.
+
+## Required privileges
+
+Only members of the `admin` role can run `COPY` statements. By default, the `root` user belongs to the `admin` role.
+
+## Known limitations
+
+### `COPY FROM` statements are not supported in the CockroachDB SQL shell
+
+{% include {{ page.version.version }}/known-limitations/copy-from-clients.md %}
+
+### `COPY` syntax not supported by CockroachDB
+
+{% include {{ page.version.version }}/known-limitations/copy-syntax.md %}
+
+## Example
+
+The following example copies data from the PostgreSQL [`psql` client](https://www.postgresql.org/docs/current/app-psql.html) into a demo CockroachDB cluster. To follow along, make sure that you have [PostgreSQL](https://www.postgresql.org/download/) installed.
+
+Run [`cockroach demo`](cockroach-demo.html) to start a temporary, in-memory cluster with the [`movr` database](movr.html) preloaded:
+
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach demo
+~~~
+
+Take note of the `(sql/tcp)` connection string listed under `Connection parameters` in the welcome message of the demo cluster's SQL shell:
+
+~~~
+# Connection parameters:
+...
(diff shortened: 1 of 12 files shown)


knz:
This is owned by Bulk I/O now. Most of the special options are related to the new `cockroach nodelocal` and `cockroach userfile` commands and thus are used "internally" and not meant for direct use by end-users.

@dt may be able to route the review of this PR to the appropriate team member?

dt:
@knz bulkIO doesn't own normal SQL INSERTs done via COPY FROM, which is what it looks like all the content here is in relation to? We do, as you point out, have two special-cases where COPY FROM can be used to interact with bulk-io's file management features, but a quick skim of the content changes here looks like it is just about SQL inserts using COPY?

knz:
Well my understanding is that COPY support was only supported for the benefit of `cockroach dump` and the `nodelocal` / `usefile` facilities, which are all bulk i/o features.

I am pretty confident that the SQL experience team has not really spent time studying / maintaining COPY otherwise. Maybe @rafiss has another opinion?

in any case, I am not the suitable reviewer, and Eric needs a reviewer.

dt:
`dump` doesn't emit COPY, it emits INSERTs. 
The special-case target table names for `nodelocal` and `userfile` were added to the existing SQL support for COPY last year but SQL has supported inserting via COPY since 2017: https://github.com/cockroachdb/cockroach/pull/8756

ericharmeling:
@rafiss Pinging you for a review.

rafiss:
will get to this by eow

rafiss (approved):
finally had time to give this a close read.

looks great! thank you

ericharmeling:
@rafiss Sweet. Thanks!

v20.2/copy-from.md
```diff
@@ -0,0 +1,122 @@
...
+
+~~~
+COPY 2
+~~~
+
+In the demo cluster's shell, query the `users` table for the rows that you just inserted:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM users WHERE id IN ('8a3d70a3-d70a-4000-8000-00000000001d', '9eb851eb-851e-4800-8000-00000000001e');
```
ianjevans, line 104:
```suggestion
SELECT * FROM users WHERE id IN ('8a3d70a3-d70a-4000-8000-00000000001d', '9eb851eb-851e-4800-8000-00000000001e');
```

ianjevans: Removing the shell prompt, so it can be copy/pastable.

ianjevans (approved):
Approving, but you might want to remove the prompt from the SQL statement so users can run the query unmodified.

ericharmeling (commented):
TFTR @ianjevans !

See my comment below. Pretty much all of our SQL examples have that `>` prompt.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9050#-:-MS3yI1J358RDCTER48H:bx83rlf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans)

___
*[v20.2/copy-from.md, line 104 at r1](https://reviewable.io/reviews/cockroachdb/docs/9050#-MNf07l6MoImDLzg4hZd-r1-104:-MS3wER1-yDECugyAtlr:b-vvhhet) ([raw file](https://github.com/cockroachdb/docs/blob/2263f16e30ec86e01486a613801da433df6230f4/v20.2/copy-from.md#L104)):*
<details><summary><i>Previously, ianjevans (Ian Evans) wrote…</i></summary><blockquote>

Removing the shell prompt, so it can be copy/pastable. 
</blockquote></details>

When you "copy" from the HTML, the prompt is actually removed from what is copied to the clipboard. Throughout our SQL syntax docs, we have the `>` prompt in front of most (if not all) copyable SQL examples.

Issue #9031, opened 2020-12-02
usability issues with JPA hello world tutorial


dianacarroll:
Exalate commented:

I am working through the JPA tutorial and have a few things I noticed that I think would improve it.

1) The README.md steps and the steps on the docs page are out of sync:
* docs have you choose your own user name, the readme specifies maxroach
* The docs have you start "cockroach demo", the readme has you start a secure cluster, with a link to https://www.cockroachlabs.com/docs/stable/secure-a-cluster

2) I believe that demo mode, by nature, means that if I stop and restart cockroach, my schema changes disappear. Although that is well-documented elsewhere, I think it would be helpful to mention here as well, because it's a non-intuitive way for a DB to work. Maybe just a link to what demo mode is.

3) I was confused about what port to configure in the .cfg file. The instructions say "Where {port} is the port number on which the CockroachDB demo cluster is listening." I looked at the CRDB startup output to find the port (which I think is the only way to do that?) However, the actual output when starting cockroach looks like this:

```
# Connection parameters:
#   (console) http://127.0.0.1:58979
#   (sql)     postgres://root:admin@?host=%2Fvar%2Ffolders%2F56%2Fd9tz0kvd59s3vhrw67lqvhkr0000gp%2FT%2Fdemo923542301&port=26257
#   (sql/tcp) postgres://root:admin@127.0.0.1:58981?sslmode=require
```
there are two different ports it shows: 58979 (for "console" and "sql/tcp") and 26257 (for "sql"). I used the wrong one (26257) and nothing worked. It would be helpful if the instructions were more explicit, such as something like "use the port as specified when you started cockroach in the sql/tcp connection parameter".

(Is that a CRDB bug?)

Jira Issue: DOC-841

dianacarroll:
Wait! The port issue is user error. I see that at the very beginning of the exercise, it explains where to get the port.  I had lost that from my brain by the time I got to the end where you actually edit the config file. (Might still be nice to have something like "where port is the port number you took note of in step 1 when starting cockroach" for people of limited brain such as myself.)

ianjevans:
@ericharmeling I can add some clarification text to the `start-cockroachdb.md` include to address points 2 & 3, which will cascade to the other tutorials. Just a heads up.

ianjevans:
It may not necessarily be a goal to have the code project README exactly match the tutorials. E.g. we may want simplified instructions in the README that work with minimal or no code changes.

[...]

PR #9051, opened 2020-12-04
Add clarification text to the start instructions per Diana's feedback.

ianjevans:
Fixes the doc side of #9031.


ericharmeling (approved):
LGTM - thanks!

PR #9058, opened 2020-12-07
Deprecate interleaved tables notice

### LINKED ISSUE #8891: Document intention to deprecate interleaved tables
After https://github.com/cockroachdb/cockroach/issues/56579 closes, we should add interleaved tables as a deprecation

ericharmeling:
Fixes #8891.


lucy-zhang (approved):
I think "de-interleave" might be too jargony and not have an obvious meaning for users as a verb. Maybe something like "convert an interleaved table to a non-interleaved table," while verbose, would be clearer in some contexts. But I'll leave that for the docs team.

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9058#-:-MO9RGoTBUrSU3ldLlLm:bo7gaur)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

ericharmeling (commented):
> "de-interleave" might be too jargony

Yeah I think you're right. Updated to be more specific.

Also updated the 21.1 docs. If this deprecation is planned for 21.1, we'll just remove this entire page (and redirect to a notice on a different page) when that deprecation is merged.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9058#-:-MOCWhv-7uliAjmX2i_P:bslszrl)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lucy-zhang)

ianjevans (approved):

PR #9064, opened 2020-12-08
Remote include fragment

### LINKED ISSUE #8973: Add ability to remote_include file fragments.
The next step after including the entire contents of a remote file is to only include fragments of a remote file. This should greatly simplify including tested code snippets in the docs, and possibly enable use to include the SQL syntax directly from the cockroachdb/cockroach repo.

To do this, we can extend the `remote_include` library with optional begin and end tokens. I've forked `remote_include` with an initial version at [ianjevans/jekyll-remote-include,](https://github.com/ianjevans/jekyll-remote-include) as the [netrics/jekyll-remote-include](https://github.com/netrics/jekyll-remote-include) project seems to gone quiet (last commit in 2019, an uncommented open PR, etc.).

ianjevans:
Fixes #8973.


jseldess (approved):
This LGTM, but I think we should move your remote include fork into the cockroachdb GitHub account rather than leave it on your personal account. Thoughts?

Also, can you think about if/how to document this in YZ's wiki docs on CI/CD and/or Jekyll?

https://cockroachlabs.atlassian.net/wiki/spaces/PROD/pages/464453705/Docs+CI+CD
https://cockroachlabs.atlassian.net/wiki/spaces/PROD/pages/452264584/Jekyll

Issue #4838, opened 2019-05-20
CSV import: be explicit about column ordering


rolandcrosby:
When importing data from CSV files, the order of the columns in the input file needs to match the column order in the schema defined by the user. A customer ran into problems with this and pointed out that this requirement isn't explicit in the documentation. The fact that we allow import of CSVs with headers (via the `WITH skip='1'` option) may also give users the wrong impression about how IMPORT works.

rolandcrosby:
We should ensure that this gets updated on both the "Migrate from CSV" page and the IMPORT documentation.

jseldess:
@lnhsingh, @mwang1026, is this still a doc need?

mwang1026:
Still relevant and is a small-ish change. We can figure out when to prioritize.

lnhsingh:
Just change the column order in the CREATE TABLE statement. Same with IMPORT INTO (order in SQL statement should match the file)

PR #9200, opened 2020-12-15
Add column order callout

gemma-shay:
Resolves issue #4838. Added a note that column order must be the same in user's SQL statement and file being imported to the Migrate from CSV and IMPORT statement docs.

### DIFF
--- a/_includes/v20.1/misc/csv-import-callout.md
+++ b/_includes/v20.1/misc/csv-import-callout.md
@@ -1,0 +1,3 @@
+{{site.data.alerts.callout_info}}
+The column order in your schema must match the column order in the file being imported.
+{{site.data.alerts.end}}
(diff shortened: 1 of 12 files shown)


mwang1026:
What's there LGTM. Do we have something in the `IMPORT INTO` docs? That one is more interesting because you can do `IMPORT INTO tablename (a,b,c) CSV 'file'` where (a,b,c) can be reordered to what's in your CSV file.

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9200#-:-MP61NsUCDTypspLCDRF:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lnhsingh and @mwang1026)

Issue #7800, opened 2020-07-28
For SHOW RANGES, explain why start_key and end_key can be both set to NULL 


florence-crl:
A Customer was confused when they saw `NULL` for both `start_key` and `end_key` for a given range:
```
root@:26257/defaultdb> select * from [show ranges from database tpcc] where range_id = 63;
  table_name | start_key | end_key | range_id | range_size_mb | lease_holder | lease_holder_locality | replicas | replica_localities
-------------+-----------+---------+----------+---------------+--------------+-----------------------+----------+---------------------
  history    | NULL      | NULL    |       63 |     11.375361 |            1 |                       | {1}      | {""}
```

Reply to the customer:
> The `NULL, NULL` output you were seeing when utilizing the `SHOW RANGES` query is actually somewhat expected, as the `SHOW RANGES` will strip the table identifiers, and if there aren't any splits within the table prefix then the range `[/Table/TableID, /Table/TableID+1)` will just display as `[NULL,NULL)`.

Could some note with the above explanation be added to this page: https://www.cockroachlabs.com/docs/v20.1/show-ranges.html?

Thanks!

PR #9202, opened 2020-12-16
Doc edit for Show Ranges with null start and end keys #7800

chriscasano:
This is a small fix to update an explanation for SHOW RANGES when null start and end keys return.  The jsonb.md change here can be discarded as it's part of another pull request.  I tried rebasing and couldn't get it out of my branch.

Closes https://github.com/cockroachdb/docs/issues/7800


jseldess:
LGTM, @chriscasano. Thank you! I'll just merge this as-is and port your change back to 20.2, the current stable version. I'll leave the jsonb change as well. If it conflicts with another PR you're working on, you can resolve that conflict there.

PR #9209, opened 2020-12-17
Remove hubspot form for docs feedback

jseldess:
Previously, when a user answered "No" to the "Was this page helpful?"
prompt at the bottom of a page, we would display a HubSpot form
to collect more detailed negative feedback. Now, clicking "No"
simply displays a thank you message, the same as for "Yes".

We are removing the HubSpot form because the Marketing team is
moving off of HubSpot to Marketo. Eventually, we should reimplement
some form or mechanism to continue collecting more detailed negative
feedback.

Clicks on "Yes" and "No" continue to be tracked via Google Analytics
for more quantitative uses.


anagio (approved):
lgtm

PR #9216, opened 2020-12-18
update Docker startup steps

### LINKED ISSUE #9185: Use docker volumes in docker instructions
In the [Start a Cluster in Docker (Insecure)](https://www.cockroachlabs.com/docs/stable/start-a-local-cluster-in-docker-mac.html) instructions, we provide instructions for starting three nodes in docker containers all writing through to the host filesystem. There is a bug (see cockroachdb/cockroach#57192, [docker/for-mac#5124](https://github.com/docker/for-mac/issues/5124#issuecomment-745393800)) in the current default Docker for Mac implementation of file sharing. The bug causes missing files and permanent crashes of the nodes.

Disabling the gRPC FUSE file sharing option in Docker for Mac is one workaround. Another workaround, recommended by a Docker engineer, is to use docker volumes: https://github.com/docker/for-mac/issues/5124#issuecomment-745211978 . This is also expected to be more performant.

I think we should update the documentation to instruct users to create a separate volume per container:
```
docker volume create roach1
docker volume create roach2
docker volume create roach3
```

And then update the `-v` flag to pass the volume identifer:
```
-v "roach1:/cockroach/cockroach-data" \
```

taroface:
Fixes #9185.

### DIFF
--- a/_includes/v19.2/start-in-docker/mac-linux-steps.md
+++ b/_includes/v19.2/start-in-docker/mac-linux-steps.md
@@ -11,6 +11,23 @@ We've used `roachnet` as the network name here and in subsequent steps, but feel
 
 ## Step 2. Start the cluster
 
+1. Create a [Docker volume](https://docs.docker.com/storage/volumes/) for each container:
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    docker volume create roach1
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    docker volume create roach2
+    ~~~
+
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    docker volume create roach3
+    ~~~
+
 1. Start the first node:
 
     {% include copy-clipboard.html %}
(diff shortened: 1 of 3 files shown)


taroface:
@jbowens I went as far back as 19.2 to make this fix, but let know if that wasn't necessary.

Also, these articles are used in our Linux tutorials (e.g. https://www.cockroachlabs.com/docs/v20.2/start-a-local-cluster-in-docker-linux) as well. Can the same updates apply to Linux, or should we break this out into a Mac-specific article?

jbowens:
Looks good! I think it's good idea to make the change all the way back to 19.2.

I think one more thing we should change is Step 6: Stop the cluster. Instead of telling them to `rm -rf cockroach-data`, we should tell them to remove the Docker volumes:

```
docker volume rm roachdb1 roachdb2 roachdb3
```

taroface:
TFTR!

jbowens (approved):
lgtm!

mikeCRL (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9216#-:-MP6Jgw37ju7_BFj9_Xr:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mikeCRL)

jseldess:
@taroface, I'll go ahead and merge this. If there's any work left, you can follow-up with another PR.

PR #9260, opened 2021-01-04
Use include for Cockroach U pydev callout

jseldess:
Also remove an extraneous </br> tag that was
breaking the callout in some cases.


ericharmeling (approved):
Awesome. Thanks for doing this!

Issue #8482, opened 2020-09-28
scheduled backup: add note that we don't accept AOST


mwang1026:
Schedules are created to maintain particular RPO (e.g. every hour).  We will set appropriate AOST when schedule executes to meet that RPO.

a-robinson:
The language added for this doesn't make much sense to me. https://github.com/cockroachdb/docs/pull/8467 added the following text:

> Schedules are created to maintain a particular recovery point objective (RPO) (e.g., an hour); therefore, you cannot set the AS OF SYSTEM TIME when creating a scheduled backup. CockroachDB will set the appropriate AS OF SYSTEM TIME when the schedule executes to meet the specified RPO.

But this is either unclear or very surprising behavior. Does it mean that if you ask for hourly backups that cockroach will use `AS OF SYSTEM TIME '-1h'`? Or that it'll use `AS OF SYSTEM TIME '-1d'` if you do daily backups? That would be very surprising as a user. But I can't figure out what else it means, since I can't figure else what would be the "specified RPO" given that no other time periods can be specified by the user.

a-robinson:
@dt am I just being daft or does this not make sense to you either?

dt:
Yeah, it could maybe be a little clearer.

When a scheduled backup is run, it backups up as of the system time at which it was scheduled to run, regardless of what time it _is_ run_. If a schedule for, say, hourly backups on the hour has a next-run-time of 4am, and at 4:00:24 the scheduler poller wakes up and notices a there's something to run, it'll run it with AOST of 4AM. That is what I think the original wording was trying to get it -- if the schedule is hourly on the hour, the system times will be hourly on the hour too.

a-robinson:
Ah, thanks! That makes more sense. I'll put up a PR real quick with wording that seems clearer to me, but of course feel free to take it or leave it :)

PR #9317, opened 2021-01-08
Attempt to clarify statement on AS OF SYSTEM TIME in scheduled backups

a-robinson:
As discussed on #8482

I can imagine wanting to not mention anything about an "internal polling mechanism", but without that a reader seems very likely to not be clear about why a backup doesn't start right away, or roughly how long of a delay to expect (although that latter concern isn't addressed super well here either).

@dt @lnhsingh


dt (approved):

a-robinson:
In case it isn't clear, I have no merge privileges, so this is waiting on someone to push the big green button.

lnhsingh:
Sorry for the delay -- thank you for opening this, Alex!!

PR #9345, opened 2021-01-15
New fragment include file for CRC sql commands.

ianjevans:
This is a fragment file with begin/end tokens for referring to the commands to start a SQL shell for the dedicated and free-tier of CRC.

### DIFF
--- a/_includes/cockroachcloud/connect_to_crc.md
+++ b/_includes/cockroachcloud/connect_to_crc.md
@@ -1,0 +1,17 @@
+<!-- Fragment include sections for connecting to CockroachCloud -->
+
+<!-- BEGIN CRC dedicated sql -->
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach sql \
+--url='postgres://<username>:<password>@<global host>:26257/<database>?sslmode=verify-full&sslrootcert=<path to the CA certificate>'
+~~~
+<!-- END CRC dedicated sql -->
+
+<!-- BEGIN CRC free sql -->
+{% include copy-clipboard.html %}
+~~~ shell
+$ cockroach sql \
+--url='postgres://<username>:<password>@<global host>:26257/<database>?sslmode=verify-full&sslrootcert=<path to the CA certificate>&options=--cluster=<cluster_name>'
+~~~
+<!-- END CRC free sql -->


ericharmeling (commented):
These fragments look great.

General question: does the standard Jekyll `include` allow including fragments from the same file? I've not done this before from a local include file.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9345#-:-MRQ1LdREVGCtDcNDXJu:b-d7veqv)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans and @lnhsingh)

___
*[_includes/cockroachcloud/connect_to_crc.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/9345#-MRQ0pKE4CEV0UStMR4d:-MRQ0pKFDe4kAN5SXtUK:b-5m7842) ([raw file](https://github.com/cockroachdb/docs/blob/b4d9ac3c91c29dd49e39fad7f918b73a566c8b68/_includes/cockroachcloud/connect_to_crc.md#L15)):*
> ```
> &options=--cluster=<cluster_name>
> ```

Unrelated to this PR:

IIUC, this parameter is one of the main differences between connecting to a local cluster and a CC cluster. We should add this to the [Connection Parameters](https://www.cockroachlabs.com/docs/stable/connection-parameters) page ahead of Free Tier launch for sure. 

(copied from https://github.com/cockroachdb/docs/pull/9346#pullrequestreview-571286268).

ianjevans:
> General question: does the standard Jekyll `include` allow including fragments from the same file? I've not done this before from a local include file.

Not from the same file, no. I think with local includes you're supposed to break up each part into individual files. For a handful of fragments that might work, but in this case we might end up with a dozen different but similar connection strings for CRC.

> > ```
> > &options=--cluster=<cluster_name>
> > ```
> 
> Unrelated to this PR:
> 
> IIUC, this parameter is one of the main differences between connecting to a local cluster and a CC cluster. We should add this to the [Connection Parameters](https://www.cockroachlabs.com/docs/stable/connection-parameters) page ahead of Free Tier launch for sure.

Definitely. I'll add that as part of the connection PR.

lnhsingh (commented):
The commands LGTM, but I have a more general fragment noob question-- where will these fragments be used and what's the difference between a fragment vs. include?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9345#-:-MRRqGIG7HH-nsJUC_eO:blgoifh)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans)

ianjevans:
> The commands LGTM, but I have a more general fragment noob question-- where will these fragments be used and what's the difference between a fragment vs. include?
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9345#-:-MRRqGIG7HH-nsJUC_eO:blgoifh)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans)

These will be used in the new Dev Guide v2 topics, but can be used anywhere.

A local `{% include somefile %}` includes the entire file. So if we wanted to have different commands for free tier and dedicated, we'd need (at least) 2 different include files. By using a fragment `remote_include` we can specify begin/end tokens surrounding just the commands, all within a single file. This is cleaner, and makes e.g. updating similar commands easier since they're all in one place.

The main use case of fragment `remote_include` tags is to only pull in specific sections of code from a larger source file. That's impossible with local includes.

But this demonstrates a different use case, for pulling in smaller sections of a common file without cluttering up the includes dir.

PR #9454, opened 2021-01-21
First draft of the general connection pooling info.

ianjevans:


ericharmeling (commented):
This is an excellent start to our connection pooling docs. Awesome work. 

I had just a few comments (mostly grammatical, qualitative feedback that you can take or leave).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9454#-:-MRfXv9K53ummeGIs9bI:b-vctivr)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans, @rafiss, and @vy-ton)

___
*[v21.1/connection-pooling.md, line 27 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfkvp65BZGmizy9YcY:-MRfkvp7E0TjzXrp83vB:bgvxkpz) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L27)):*
> ```
> ... is higher than ...
> ```

Is this always the case? Should we say "... can be higher..."?
___
*[v21.1/connection-pooling.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfXgm8DHrEbrHOHbbx:-MRfXgm8DHrEbrHOHbby:b30u29f) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L29)):*
> ```
> This seems counterintuitive. Shouldn't a pool maintain more threads so each operation can immediately get a connection to the database server, and then let the system process these threads in parallel?
> ```

nit:
I like the voice here, but I think you can probably remove these sentences. Not sure this bit adds any information.
___
*[v21.1/connection-pooling.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRflVAHAQ7Wlx5Wz8cw:-MRflVAI5ExYg3gIe_8j:bjm96hl) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L29)):*
> ```
> time slice
> ```

Might want to link to a definition here... https://en.wikipedia.org/wiki/Preemption_(computing)#Time_slice ...

also, nit:
Is the verb "timeslice" or "time-slice"?
___
*[v21.1/connection-pooling.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfloxk68EAJ8Oz7r-U:-MRfloxk68EAJ8Oz7r-V:b-oooz0s) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L29)):*
> ```
> Each processor core, however, can only
> ```

nit:
"Each processor core can only"
___
*[v21.1/connection-pooling.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfmPyj5KrBaZ0sWfu-:-MRfmPyj5KrBaZ0sWfu0:bev8ux0) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L31)):*
> ```
> Processor performance isn't the only factor, however.
> ```

nit:
You can probably remove this sentence.
___
*[v21.1/connection-pooling.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfnEIJA97fvNS5aY4t:-MRfnEIJA97fvNS5aY4u:buqd6i4) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L33)):*
> ```
>  However like
> ```

nit:
"However, like"
___
*[v21.1/connection-pooling.md, line 35 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfnQa3ByT4bxm9sFFK:-MRfnQa3ByT4bxm9sFFL:bn2bc9e) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L35)):*
> ```
> Network latency is typically the least important factor in determining the connection pool size.
> ```

If this is the case, should we mention network latency at all?
___
*[v21.1/connection-pooling.md, line 37 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfnUdSAbdqVTqZOD80:-MRfnUdSAbdqVTqZOD81:b11ly5x) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L37)):*
> ```
> HikariCP project
> ```

Should probably link to https://github.com/brettwooldridge/HikariCP.
___
*[v21.1/connection-pooling.md, line 40 at r1](https://reviewable.io/reviews/cockroachdb/docs/9454#-MRfnnRC5WJHZ-EGHCbf:-MRfnnRC5WJHZ-EGHCbg:b-dbyhls) ([raw file](https://github.com/cockroachdb/docs/blob/80a73dc7bdc77d34ec1030eba8ff8a1beb767ed6/v21.1/connection-pooling.md#L40)):*
> ```
> ~~~
> connections = ((number of cores * 2)) + effective spindle count)
> ~~~
> ```

I like having a new line for this, but does it need to be in a code block? I honestly don't know our convention for formulas.

ianjevans:
> Is this always the case? Should we say "... can be higher..."?

I think it is always higher if the number of threads exceeds the number of cores, due to time-slicing. But just in case I changed it to "is typically higher".

> > ```
> > Network latency is typically the least important factor in determining the connection pool size.
> > ```
> 
> If this is the case, should we mention network latency at all?

I think we should if only to avoid pedants who will say, "what about network latency?". Maybe something to ask @rafiss  about.

ericharmeling (commented):
> I think we should if only to avoid pedants who will say, "what about network latency?". 

Makes sense to me!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9454#-:-MRtrVj682mW16Gunb0o:be4v0et)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans, @rafiss, and @vy-ton)

ericharmeling (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9454#-:-MRtrks71EFdADw4XoB6:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ianjevans, @rafiss, and @vy-ton)

a-entin:
I'm not sure how # spindles can be correlated to the connection pool size. I doubt it can be, credibly that is.

Specifically, the formula suggests the faster the storage (# connections is proportional to # spindles; more spindles generally means faster IO), the bigger the pool size can be. This maybe it's incorrect; the opposite is true. The reason overcommitting connections to CPU works is because during query processing CPU has to wait for IO (disk and network ios). Means it can process another query in the meantime. The faster the IO, the better is CPU utilization and consequently less benefit for running 2  queries on 1 core. In a theoretical case when 0 IO is needed, there is no benefit in having more than 1 sql exe thread per core.

I suggest removing the disk and network IO references altogether (adds confusion, not actionable by user, not entirely true technically; there are no "spindles" in many types of storage e.g. cloud storage) and use a simple formula consistent with our sizing guidance - the connection pool size across all apps should not exceed 4 * the total cores (vcpus) in the cluster, presuming connections are uniformly distributed across all nodes (in case nodes are diff size, not recommended, but then - proportionally to vcpus per node)

We don't need to talk about active connections in relation to connection pool size. We know the number of cores in the cluster. That's the input for connection pool sizing. (Fwiw, the active connections stuff is for cluster sizing, which the task of figuring the number of cores and nodes required... a related yet different exercise).

ianjevans:
> I'm not sure how # spindles can be correlated to the connection pool size. I doubt it can be, credibly that is.
> 
> Specifically, the formula suggests the faster the storage (# connections is proportional to # spindles; more spindles generally means faster IO), the bigger the pool size can be. This maybe it's incorrect; the opposite is true. The reason overcommitting connections to CPU works is because during query processing CPU has to wait for IO (disk and network ios). Means it can process another query in the meantime. The faster the IO, the better is CPU utilization and consequently less benefit for running 2 queries on 1 core. In a theoretical case when 0 IO is needed, there is no benefit in having more than 1 sql exe thread per core.

I think the HikariCP docs explain this in more detail, but it is strongly implied that the increasing the number of spindles increases latency in the storage system.
https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing

> I suggest removing the disk and network IO references altogether (adds confusion, not actionable by user, not entirely true technically; there are no "spindles" in many types of storage e.g. cloud storage) and use a simple formula consistent with our sizing guidance - the connection pool size across all apps should not exceed 4 * the total cores (vcpus) in the cluster, presuming connections are uniformly distributed across all nodes (in case nodes are diff size, not recommended, but then - proportionally to vcpus per node)
> 
> We don't need to talk about active connections in relation to connection pool size. We know the number of cores in the cluster. That's the input for connection pool sizing. (Fwiw, the active connections stuff is for cluster sizing, which the task of figuring the number of cores and nodes required... a related yet different exercise).

This brings up a problem in talking about connection pool sizing. How do we talk about sizing the connection pool at the _application layer_ with HikariCP when the connection pool size maximum is across all applications that use the _cluster_? In other words, how do I, an application developer, know how to size my connection pool when I don't know what other applications are hitting the same database? If I were configuring PgBouncer as a proxy in between my application and the database, that would be easy. But that is more an operator configuration since it's a deployment scenario.

vy-ton:
> This brings up a problem in talking about connection pool sizing. How do we talk about sizing the connection pool at the application layer with HikariCP when the connection pool size maximum is across all applications that use the cluster? In other words, how do I, an application developer, know how to size my connection pool when I don't know what other applications are hitting the same database?

I believe you would still size for your application needs. If there are multiple applications, which is something we see with users with many microservices, then you might want want to bring in something like PGBouncer to act as the funnel to the database. However, I think we can treat this as a separate use case which might fit better in operator docs.

PR #9530, opened 2021-01-29
Revising based on Alex Entin's feedback.

ianjevans:


a-entin (approved):
Looks good, thanks! Only comment - using the term "core" vs "vcpu". VERY loosely speaking they are kinda the same, yet one relates to bare metal platforms while the other to virtualized platforms. We just need consistency across the doc. We are using "vcpus" in some parts of the doc.

ericharmeling (approved):
LGTM!

PR #9692, opened 2021-02-02
remove incorrect/outdated callout

taroface:
Addresses [this](https://github.com/cockroachdb/docs/pull/7304#discussion_r568754657) comment.

Our docs state that the new recommissioning behavior begins as of v20.2, so I applied this change to v20.2 and v21.1.


jseldess (approved):
LGTM

PR #9693, opened 2021-02-02
Add information on connection validation to connection pool topic.

### LINKED ISSUE #9691: Document "validation" of connections in a connection pool
We have seen a few issues where customers are confused why connections in their application become broken. This can happen in normal operation, and is expected in CockroachCloud where we perform minor version upgrades and other rolling restarts. When using a connection pool. users must be sure that the pool is checking that a connection is valid before trying to use it

Check out the HikariCP docs for some language we might want to use: https://github.com/brettwooldridge/HikariCP#configuration-knobs-baby (Look for "isValid" and "connectionTestQuery"

Each connection pool tool/library might do things a bit differently. For example, hikaricp would do this automatically using the JDBC `Connection.isValid()` API — whenever the app borrows a connection from the pool, it first checks if the connection is still valid.

But we should add docs that specify that users must make sure their connection pool does something like this.

ianjevans:
Fixes #9691.

### DIFF
--- a/v21.1/connection-pooling.md
+++ b/v21.1/connection-pooling.md
@@ -36,6 +36,12 @@ connections = (number of cores * 4)
 
 Many workloads perform best when the number of connections was between 2 and 4 times the number of CPU cores in the cluster.
 
+## Validating connections in a pool
+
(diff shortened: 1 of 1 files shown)


ericharmeling (approved):
Looks good!

rafiss (approved):
thanks for the quick update!

PR #9712, opened 2021-02-04
add redirects

anagio:
add redirects for old top level pages


jseldess:
LGTM. We'll change the links on dockerhub directly, but merging this in the meantime. Thanks, @anagio.

PR #9714, opened 2021-02-04
Fix containers names when cluster stopping

akss1:
There is no container named cockroachdb-0, but there is a container named cockroachdb-3.


rmloveland (approved):
LGTM, thanks @akss1 !!!1!

PR #9721, opened 2021-02-05
update minikube start defaults

taroface:
Adjusted the minikube CPU and memory defaults (7 CPU and 26 GB) to work with the resource requests (2 CPU and 8 GB) we prepopulate in the [example.yaml](https://github.com/cockroachdb/cockroach-operator/blob/master/examples/example.yaml) for the Operator. This updates the [local Kubernetes deployment guide](https://www.cockroachlabs.com/docs/v20.2/orchestrate-a-local-cluster-with-kubernetes.html) for 20.2 and 21.1.

(Also removed the restart count for a cockroach pod in the example deploy output)

### DIFF
--- a/_includes/v20.2/orchestration/local-start-kubernetes.md
+++ b/_includes/v20.2/orchestration/local-start-kubernetes.md
@@ -16,7 +16,7 @@ Feature | Description
 
     {{site.data.alerts.callout_info}}Make sure you install <code>minikube</code> version 0.21.0 or later. Earlier versions do not include a Kubernetes server that supports the <code>maxUnavailability</code> field and <code>PodDisruptionBudget</code> resource type used in the CockroachDB StatefulSet configuration.{{site.data.alerts.end}}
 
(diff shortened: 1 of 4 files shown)


taroface:
@keith-mcclellan Would you mind weighing in on these values? I do not know if they make any sense at all. However, they were roughly the minimums that worked for me as I inched up from 3 CPUs and 15 GB for minikube.

keith-mcclellan:
This still seems really heavy - we may need a minikube specific example.yaml for local deployments if this is something we want to support. our kind deployments don't require nearly this much memory fwiw so this is probably a config issue.  @taroface is this something we need to focus on?  I haven't used minikube in a while but I could set it back up.

taroface:
@keith-mcclellan Per our convo I'll see if there's a memory issue on my end first.

taroface:
I need to do some retesting on this, so will reopen when that's done.

PR #9699, opened 2021-02-02
export: clarify file naming

shermanCRL:
Per a support issue, EXPORT file naming changes in 20.2 can lead to unexpected outcomes.

PR #9735, opened 2021-02-09
typo

shermanCRL:


shermanCRL:
This is one more tweak on https://github.com/cockroachdb/docs/pull/9699. Merging without review because I believe it’s trivial enough.

PR #9736, opened 2021-02-09
experiment with Jekyll version [do not merge]

shermanCRL:
This is an experiment, do not merge. Wondering if a traditional Jekyll version will speed up the Netlify deploy.


shermanCRL:
The Netlify checks went from 5 minutes down to 2, which is good news, but doing this properly will require some CI updates. Will wait for @yingzhuchin ’s return before proceeding.

PR #9768, opened 2021-02-09
fix CC All Releases Latest badge css

gemma-shay:
Custom CSS was breaking the Latest badge. Differences didn't show up in the local build but the TeamCity online preview seems accurate.


lnhsingh:
LGTM, thanks!

Issue #9774, opened 2021-02-10
Performance Benchmarking with TPC-C: workload fixtures import tppcc fails


ciklista:
Re: [Performance Benchmarking with TPC-C](https://cockroachlabs.com/docs//v20.2/performance-benchmarking-with-tpcc-small.html)

## Issue Description
When trying to replicate the performance benchmark with TPC-C, the following command fails (Step 3):
```
./workload fixtures import tpcc \
--warehouses 2500 \
"postgres://root@<address of any CockroachDB node>:26257?sslmode=disable"
```
The Error is:
```
Error: importing fixture: importing table order: pq: parsing parameters --warehouses=1000 --replicate-static-columns=false --seed=1 --deprecated-fk-indexes=false --families=false --fks=true --interleaved=false: unknown flag: --replicate-static-columns
```
## Suggested Resolution 
According to the [docks of the workload binary](https://www.cockroachlabs.com/docs/stable/cockroach-workload.html), the command should be
```
 ./workload init tpcc --warehouses 2500 "postgres://root@<address of any CockroachDB node>:26257?sslmode=disable"
```
I am currently testing with that command

ciklista:
It works as expected with the suggested solution

jseldess:
Thanks, @ciklista. @rmloveland, can you please test the changes in the PR and then approve and merge and close this?

rmloveland:
Thanks @ciklista for reporting this issue!

@jseldess thanks for the ping, gonna check in with folks from engineering on this to verify any changes so we don't introduce any additional subtle issues.

@aayushshah15 I have a few questions (my memory is you were the primary person running these - please redirect if not):

1. it seems possible that this issue (args to latest `workload` changing from what's in the v20.2 doc) could be caused by the fact that the directions tell the user to grab whatever is the latest build of the separate `workload` binary, and the flags to that binary may have changed out from under the doc.   Does that assessment seem correct? (I can also test this but I wanted to verify with you before spinning up a whole cluster of machines - also I do not currently have AWS access so it will take some time for me to investigate)

2. If yes to question 1 above, is there a stable/pinned version of `workload` at some specific SHA we could point users at who are trying to replicate these instructions as of about the v20.2.0 release date?

PR #9775, opened 2021-02-10
Fixing command to prepare tpc-c benchmark

ciklista:
Proposing a fix for issue https://github.com/cockroachdb/docs/issues/9774.

According to the [docs for the workload tool](https://www.cockroachlabs.com/docs/dev/cockroach-workload.html), there is no `workload fixtures import` command, so I am proposing to use `workload init ` instead. 
This is also used in the [docs for local testing](https://github.com/cockroachdb/docs/blame/master/v20.2/performance-benchmarking-with-tpcc-local.md).


ciklista:
fixed with https://github.com/cockroachdb/docs/pull/9935

PR #9790, opened 2021-02-11
Remove hubspot and heap assets from docs

jseldess:
Marketing doesn't use either of these systems anymore.
They migrated to Marketo.

For later follow-up:
- There's still hubspot-related styling in place. Search for "hubspot" here:
  https://github.com/cockroachdb/docs/blob/master/css/customstyles.css


anagio (approved):
g2g

jseldess:
TFTR and the commits, @anagio.

PR #9798, opened 2021-02-11
update PVC and PV deletion step

### LINKED ISSUE #9719: K8s Operator | Update Guidance on Deleting PV and PVCs
In our existing README and docs, we recommend our users to delete all pv and pvcs. 

After speaking with @keith-mcclellan , this risks customers deleting non-CRDB pv and pvcs.

Instead, let's guide users to specify deleting their pv and pvcs. 
_Delete persistent volume:_
`kubectl delete -f pv <name of persistent volume>`

_Delete persistent volume claim:_
`kubectl delete -f pvc <name of persistent volume claim>`

In addition, there may be a more efficient way to delete all crdb pv and pvcs in a single command without having to specify the name of each. 

After taking a look at the [official K8s Persistent Volumes docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming), we may need to have users update their Reclaim policy as well to "Delete"

@keith-mcclellan, any additional thoughts here?

taroface:
Fixes #9719.

- As discussed with @keith-mcclellan, this removes the command to specifically delete PVCs and PVs (on Operator and Manual deployment docs), and links instead to K8s documentation.
  - Also does the above for secret deletion.
- There is a red callout with an explicit warning about deleting PVs.
- This updates 21.1 and 20.2 docs (secure and insecure).

### DIFF
--- a/v21.1/orchestrate-cockroachdb-with-kubernetes-insecure.md
+++ b/v21.1/orchestrate-cockroachdb-with-kubernetes-insecure.md
@@ -103,7 +103,7 @@ To start your CockroachDB cluster, you can either use our StatefulSet configurat
 
 To shut down the CockroachDB cluster:
 
-1. Delete all of the resources you created, including the logs:
+1. Delete the resources you created, including the logs:
 
     {{site.data.alerts.callout_danger}}
     This does not include deleting the persistent volumes that were attached to the pods. If you want to delete the persistent volumes and free up the storage used by CockroachDB, be sure you have a backup copy of your data. Data **cannot** be recovered once the persistent volumes are deleted. For more information, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/#persistent-volumes).
--- a/v21.1/orchestrate-cockroachdb-with-kubernetes-insecure.md
+++ b/v21.1/orchestrate-cockroachdb-with-kubernetes-insecure.md
@@ -103,12 +103,16 @@ To start your CockroachDB cluster, you can either use our StatefulSet configurat
 
 To shut down the CockroachDB cluster:
 
-1. Delete all of the resources you created, including the logs and remote persistent volumes:
+1. Delete all of the resources you created, including the logs:
+
+    {{site.data.alerts.callout_danger}}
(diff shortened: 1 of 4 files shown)


taroface:
@keith-mcclellan Friendly ping for review!

keith-mcclellan:
@taroface we also should link out to manually clean up the secrets too.  Otherwise this appears to be correct

taroface:
@keith-mcclellan Is this another situation where we'd want to avoid having a `kubectl delete secrets` example directly in the doc? Or is it safe to include that example?

keith-mcclellan:
I would probably do the same as deleting the PVCs thanks!

v21.1/orchestrate-cockroachdb-with-kubernetes-insecure.md
```diff
@@ -103,12 +103,16 @@ To start your CockroachDB cluster, you can either use our StatefulSet configurat
 
 To shut down the CockroachDB cluster:
 
-1. Delete all of the resources you created, including the logs and remote persistent volumes:
+1. Delete all of the resources you created, including the logs:
+
+    {{site.data.alerts.callout_danger}}
+    This does not include deleting the persistent volumes that were attached to the pods. If you want to delete the persistent volumes and free up the storage used by CockroachDB, be sure you have a backup copy of your data. Data **cannot** be recovered once the persistent volumes are deleted. For more information, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/#persistent-volumes).
```
mikeCRL, line 109: I am not sure why "all of the resources you created" would not include data on persistent volumes attached to the pods. This might make sense to a person who's gotten to this point, but I'm new to CRDB in a Kubernetes environment, so the meaning is not obvious to me. So LGTM, if you think this meets the needs of the intended audience after considering this feedback and maybe you can clarify this distinction for me, regardless.

taroface: Thanks for this! It's potentially confusing for everyone IMO. I think I'll amend it by deleting the word "all".

Issue #9434, opened 2021-01-21
Top Nav height 


luciaaaaaaa:
Issue: Top nav has 10px more in height than top nav in the marketing site. 
Fix: Top nav height should be max 90 px

Issue #9432, opened 2021-01-21
Top Nav height increases when I scroll down 


luciaaaaaaa:
Issue: The top nav height increases when I scroll down. This behaviour does not occur on the marketing site. 

Fix: Height of the top nav should not change 

I have a reference video which was too big to upload here. Will post into Slack 

![image](https://user-images.githubusercontent.com/48258790/105380519-f9180b00-5bdb-11eb-9384-776a4b68c26a.png)

lnhsingh:
Related to #9434

anagio:
sent a fix to make sidebar and top navs all go mobile at the same time.

Issue #9426, opened 2021-01-21
Top Nav cuts off 


luciaaaaaaa:
Notice how the left side margin is cut off as I scroll down 

![image](https://user-images.githubusercontent.com/48258790/105380299-b81ff680-5bdb-11eb-961e-1e4a3ca0e168.png)

PR #9799, opened 2021-02-12
new menu new break points

### LINKED ISSUE #9243: Rollover popup menus hinder effective use of the web pages
Re: [Hello World Example Apps](https://cockroachlabs.com/docs//v20.2/hello-world-example-apps.html)

## Issue Description

The dropdown menus ("Why CockroachDB", "Product", and "Resources") hinder effective use of the web pages, for two reasons:

1. The menus display by simply moving a pointer over them, instead of clicking on them.  While many in the the Web UI community accept this, please reconsider.  Users generally move their pointer in order to, well, simply move it; to get it to another part of the screen.  It violates the "don't surprise" principle of UI design to have menus display as a side-effect of a user acting on their intention to simply move their pointer.  This is exacerbated when the menu hides the very content they intend to move their pointer to, which brings up a second reason:
1. The menus are too large, with too much white space, and have no visible border.  These effects combine to make it hard to distinguish between the menu itself and the background web page.  Given that the user probably did not consciously cause the menu to appear, the user has to 1: recognize that they are looking at a page that contains a menu instead of the expected content, 2: discern the boundaries of the menu, and 3: figure out where to move their pointer to make the menu go away.  It interrupts their "flow" by creating a kind of "puzzle" that they have to solve before they can continue what they intended to do.

## Suggested Resolution 

Action to display:

- Display drop-down menus by clicking on them, not by rolling over them.  
- If you wish to keep the rollover functionality, reduce the active "trigger" area the menu words themselves, because moving the pointer in the white space that is well above the menu words also causes the menus to display.

Menu design/content:

- Reduce the size of the menus
- Remove large blank areas
- Make menus more conspicuous and distinct (perhaps by having a border)

Example:

dislike rollover menus A 2020-12-23

dislike rollover menus B 2020-12-23

### LINKED ISSUE #9312: CSS responsive breakpoint is too large
Docs site jumps to the mobile layout around <1200px. Which on a laptop, means just a little less than full screen:

image

anagio:
new nav with data files - closes #9243 #9434 #9426 #9432 
new break points - closes #9312


lnhsingh:
The edit looks good on stable/dev. Are all of the TC errors are due to the nav not being updated for 19.2?

jseldess:
LGTM, too, but we need to fix the TC failures, which are related to the new topnav yaml approach. The links look OK in the test build, but htmltest see them as broken: https://teamcity.cockroachdb.com/buildConfiguration/Internal_Docs_GitHubCi/2671350?buildTab=log&focusLine=1174&linesState=455

@anagio, any ideas? We could make the urls absolute to fix this. Or we could tell htmltest to skip them. I can show you how to do that, if you want, but it's maybe risky to do that.

jseldess:
Looks like we need to resolve a layout conflict as well.

anagio:
Working on issues I will get an update out by tomorrow

anagio:
@jseldess fixed links and build passed!

bdarnell:
Confirmed that this fixes #9856 for me.

jseldess (commented):
Thanks, @anagio! This LGTM now.

jseldess (approved):

jseldess:
Merging.

Issue #9770, opened 2021-02-09
use traditional version on Jekyll gem


shermanCRL:
(Creating this issue for future reference, after @yingzhuchin has returned)

In the Gemfile, Jekyll is specified as a commit version on GitHub, instead a traditional release version. As a result, the [Netlify checks](https://app.netlify.com/sites/cockroachdb-docs/deploys/6021e71fb86fa1000803bb23) are expensive, requiring gem rebuilds. In [this experiment](https://github.com/cockroachdb/docs/pull/9736), using a traditional version reduced the execution time from ~5 minutes to ~2.

Looking at ci/README.md, I see that updating gem versions also requires updating the Docker image. So the immediate task is:

- Update the Dockerfile with the updated Jekyll gem version, as prescribed in the README

Further optimizations to consider after the above:

- Use a more recent Alpine image in the Dockerfile, which will get us a more recent Ruby
  - It is desirable that the CI Ruby version is close to the default version on Mac, that developers use
- Use an actual [Ruby Docker image](https://hub.docker.com/_/ruby)

Further consideration: requiring a Docker image update when gems are updated is a performance optimization, but imho, makes things harder than they should be. I’d rather have a gem update ‘just work’. If we wish to also optimize the Docker image, make it optional.

jseldess:
This was address in a PR by @ianjevans: https://github.com/cockroachdb/docs/pull/9804/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR8. 

The docker image is just for TeamCity. We need to think about moving all the way over to Netlify for CI/CD: https://github.com/cockroachdb/docs/issues/7050 once @yingzhuchin is back.

PR #9804, opened 2021-02-13
Multiple content filters set as query parameters

### LINKED ISSUE #9800: Topic content filters can't be embedded
For a topic that uses `filters` to show/hide sections, you can't use multiple filter groups.
```
<div class="filters clearfix">
  <button class="filter-button" data-scope="sql">SQL</button>
  <button class="filter-button" data-scope="go">Go</button>
  <button class="filter-button" data-scope="java">Java</button>
  <button class="filter-button" data-scope="python">Python</button>
</div>

...

<div class="filters clearfix">
  <button class="filter-button" data-scope="local">Local</button>
  <button class="filter-button" data-scope="cockroachcloud">CockroachCloud</button>
</div>
```

This is because the filter show/hide logic in `js/customscripts.js` assumes only one filter can be active at a time. If a user selects a filter by clicking a `filter-button` element, the page URL is modified with the selected `data-scope` name as the anchor.

The end result is that writers need to split topics based on one of the filter criteria. E.g. in the above example, splitting the topic into a local and cloud topic each with filters for the language. This results in duplication of content across topics.

Since the filters are already logically grouped by `<div>` tags around the buttons, it should be possible to show/hide the `data-scope` content by group, which would allow writers to use multiple filters within a single topic.

The URL rewriting logic could use URL parameters instead of anchor names to specify which filters were selected. For example, `https://www.cockroachlabs.com/docs/stable/connect-to-the-database.html?filters=cockroachcloud;&filters=python;`

ianjevans:
Fixes #9800.

This change modifies our custom JS to allow multiple content filters within a documentation topic.

This would allow us to document 2 or more levels of filtered content within the same topic. For example, a developer topic that has a main filter for Local vs. CockroachCloud, and further filtering by programming language.

```
<div class="filters clearfix">
  <button class="filter-button" data-scope="local">Local</button>
  <button class="filter-button" data-scope="cockroachcloud">CockroachCloud</button>
</div>
...
<div class="filters clearfix">
  <button class="filter-button" data-scope="sql">SQL</button>
  <button class="filter-button" data-scope="go">Go</button>
  <button class="filter-button" data-scope="java">Java</button>
  <button class="filter-button" data-scope="python">Python</button>
</div>
...
<section class="filter-content" markdown="1" data-scope="sql">
...
<div class="filter-content" markdown="1" data-scope="local">
...
</div>
<div class="filter-content" markdown="1" data-scope="cockroachcloud">
...
</div>
</section>
```

Currently it works with our existing single-level content filters, but I will do a test topic with multiple filter groups within the topic.

In order to store multiple filters selections, I had to change how the current filter is set to use URL query parameters instead of URL anchors.

This will affect, but not break, existing URLs and bookmarks that use the old filter trigger of a URL anchor. For example, this URL:
`https://www.cockroachlabs.com/docs/dev/connect-to-the-database.html#go`
changes to:
`https://www.cockroachlabs.com/docs/dev/connect-to-the-database.html?filters=go` after this PR is merged.

URLs and bookmarks will not break, but just won't select the correct filter when the page loads.


gemma-shay:
Tested with two sets of filters and it works perfectly unless the page is reloaded. If you have the first tab in one section open and a non-first tab in another open, the first tab's content will disappear on reload. Doesn't happen if only first tabs or only non-first tabs are selected on the page.
Screen+Recording+2021-02-16+at+10 00 22+AM

ericharmeling (commented):
It looks like this change works out really nicely with our current filters. Great work.

As you've noted, the only issue is that the (currently filter-created) URL anchors don't break, making it hard to detect and fix any of the existing links to filtered pages... 
To resolve this, you could:

1. Manually update the references to use parameters instead of URL anchors. This would require us to search for all `class="filter-content"` instances and then update their corresponding references. Or perhaps look somewhere where this is logged (I didn't see it in the TC build though), and then manually update based on the logged refs.

2. Add a function that replaces the URL with one that uses just the single hash in the event that there is just one filter query parameter. 

3. Decide that this is not an issue and just merge the changes. I'm really not sure how often we are sending people to the URL anchors for the filtered pages.

Also, might want to loop in @j-low for review as well.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9804#-:-MTfGx2g0WiBAStGxZ6C:b-e5was8)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay and @jseldess)

ianjevans:
Gemma: Thanks for testing! I pushed a new commit to the branch that I think
should fix the issue you found.

ianjevans:
Eric: If we wanted to fix the hash issue with a brute force global
search/replace, I would think grep-ing through the source for relative
links that have # anchors would be fairly straightforward.

I could also probably write a URL rewrite function to change anchor filters
to query params, though that doesn't fix the existing links.

ianjevans:
> Eric: If we wanted to fix the hash issue with a brute force global search/replace, I would think grep-ing through the source for relative links that have # anchors would be fairly straightforward. 

Derp, the big problem is that we have existing real anchor links, and no way to distinguish between filter hash links and anchor links. Yuck.

ericharmeling:
> > Eric: If we wanted to fix the hash issue with a brute force global search/replace, I would think grep-ing through the source for relative links that have # anchors would be fairly straightforward.
> 
> Derp, the big problem is that we have existing real anchor links, and no way to distinguish between filter hash links and anchor links. Yuck.

Per our discussion during this past week's team meeting, I think it'll be fine to just go forward with the PR.

ericharmeling (approved):

ianjevans:
While I was updating the CSS I also updated the Gemfile and CI builder image to use an explicit, newer version of Jekyll (4.2), related to #9770. That ticket includes additional suggested fixes (e.g. updating the Alpine image) that YZ should review so this PR shouldn't be considered a fix.

PR #9819, opened 2021-02-18
create-changfeed.md: add new sasl_mechanism Kafka URL parameter

stevendanna:
A new kafka URL parameter, `sasl_mechanism` was added in

https://github.com/cockroachdb/cockroach/commit/5f96c9132fc16576bdf4d1b9b51bb943ab8732bc

which allows users to specify the SASL mechanism used during
authentication.


lnhsingh:
Change has been added to the docs

PR #9788, opened 2021-02-10
CLI import command

gemma-shay:
-Add new `cockroach import` statement to cockroach commands page
-Create new `cockroach import` page
-Add callout to `cockroach userfile upload` page suggesting when to use `cockroach import` instead

Closes: #9139

PR #9839, opened 2021-02-19
CLI Import privileges fix

gemma-shay:
Missed @adityamaru's last comment before merging https://github.com/cockroachdb/docs/pull/9788. 

Change privileges from `admin` to `CREATE` on default database.


adityamaru:
LGTM! Except I think we should replace `default database` with `defaultdb`. This is the database that we start all clusters with and that is where this CLI will import.

lnhsingh:
Once you make the small fix Aditya suggested, LGTM

PR #9840, opened 2021-02-22
rel url paths

anagio:
set rel url for links

### DIFF
--- a/_layouts/homepage.html
+++ b/_layouts/homepage.html
@@ -69,7 +69,7 @@ layout: default
         <h3 class="mt-3">Get Started</h3>
         <div class="landing-column-content">
           <ul>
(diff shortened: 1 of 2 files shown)


jseldess (approved):
Thanks, @anagio. LGTM. Do we need to use that syntax whenever we use relative links in html hrefs, as opposed to Markdown links? 

If so, do we also need to do this on the other two landing pages?

https://github.com/cockroachdb/docs/blob/master/cockroachcloud/index.md
https://github.com/cockroachdb/docs/blob/master/v20.2/index.md (and for all the other versions)

anagio:
The upside to writing them with the `relative` filter is we see what the full relative path should be from /docs path down over using ../some-file, but using the filter means writing a bit more code for a link. Ideally we separate the html parts out of the markdown into a template and add front matter to the index markdown for specific content. Not sure its worth updating all older versions unless they're breaking.

PR #9855, opened 2021-02-23
Add meta description into head.html

jseldess:
This was removed at some point. It leverages the summary front-matter
field as the meta description.


jseldess:
Got approval from Dan Kelly in another channel. Merging because all of our docs are missing meta descriptions.

PR #9892, opened 2021-02-25
Added Working with Spatial page to the Develop ToC

ericharmeling:
@jseldess 
@rmloveland

### DIFF
--- a/_includes/sidebar-data-v20.2.json
+++ b/_includes/sidebar-data-v20.2.json
@@ -338,6 +338,12 @@
                     ]
                   }
                 ]
+              },
+              {
+                "title": "Working with Spatial Data",
+                "urls": [
+                  "/${VERSION}/spatial-data.html"
+                ]
               }
             ]
           },
--- a/_includes/sidebar-data-v20.2.json
+++ b/_includes/sidebar-data-v20.2.json
@@ -340,7 +340,7 @@
                 ]
               },
               {
-                "title": "Working with Spatial Data",
+                "title": "Work with Spatial Data",
                 "urls": [
                   "/${VERSION}/spatial-data.html"
                 ]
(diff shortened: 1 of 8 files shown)


ianjevans (approved):

jseldess:
Thanks, @ericharmeling! Question though: Why does this belong specifically under Reading data? Should it be parallel to that, or is that awkward?

ianjevans:
@ericharmeling I didn't realize you were still working on this, you can rerequest a review when you're done pushing updates.

ericharmeling:
@jseldess 

Great questions. I should say that Rich and I discussed how we might incorporate the existing spatial docs in the Develop ToC yesterday, and this PR is a result of that discussion. 

Our original plan was in two parts:
1. Simply move the existing "Working with Spatial Data" page into the "Develop" docs (this PR).
2. Incorporate spatial data into the "Develop" docs in a more integrated fashion (i.e., add a "Spatial" section to the relevant pages of the guide, such as "Query Data"), and then update the "Working with Spatial Data" page to fit a little better in the workflow of the "Develop" docs or remove the page entirely.

This plan is flexible though.

In response to your questions:

> Why does this belong specifically under Reading data?

"Working with Spatial Data", and really any workflow that deals with spatial data in CockroachDB primarily consists of two steps:
  1. Importing data
  2. Querying data

Although importing data is technically a "write" operation, Rich and I agreed that spatial data is primarily queried (i.e., "read").

> Should it be parallel to that, or is that awkward?

I think it would be awkward. I would prefer to limit the number of items at the top level of "Develop", especially if we can logically place them under one of the existing chapters. 

After reading your questions, I updated the PR. I took out one of the examples in "Working with Spatial Data" and put it in a new page called "Query Spatial Data". This page does not have explicit steps on importing data, but instead points users to the "Migrate from Shapefiles" page, which is (correctly, IMO) under "Migrate". The original "Working with Spatial Data" now serves as more of a reference page with an example, and remains under "Reference->SQL->Spatial Features" with all of the other reference pages for spatial.

@ianjevans Apologies for the churn - this should be ready for another review.

ianjevans (approved):

jseldess (approved):
LGTM

Issue #9886, opened 2021-02-24
Users can configure survivability goals for a database


rmloveland:
This issue tracks all work related to multi-region survival goals.

See also:

- [x] #9115
- [ ] #9116
- [ ] #9271
- [x] #9298
- [x] #9610

rmloveland:
Calling this one done via #9936 - we have docs for configuring survival goals

Issue #9884, opened 2021-02-24
The REGIONAL table locality pattern supports table-level homing


rmloveland:
This issue tracks all work related to making sure we have docs on table-level homing (aka `REGIONAL BY ROW`, or whatever we end up finally calling it).

See also:

- #9301
- #9303

Issue #9709, opened 2021-02-03
The REGIONAL table locality pattern supports row-level homing


rmloveland:
See also:

- [ ] #9265
- [ ] #9266
- [ ] #9288
- [x] #9301
- [ ] #9303

rmloveland:
The following pages complete the work for this issue IMO:

- [multi-region overview's section on regional by row tables](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview#regional-by-row-tables)
- [`ALTER TABLE ... SET LOCALITY REGIONAL BY ROW`](https://www.cockroachlabs.com/docs/v21.1/set-locality#regional-by-row)
- [Regional tables](https://www.cockroachlabs.com/docs/v21.1/regional-tables.html) page that replaced the old geo-partitioning stuff
- [multi-region latency demo / tutorial that makes heavy use of `REGIONAL BY ROW`](https://www.cockroachlabs.com/docs/v21.1/demo-low-latency-multi-region-deployment)

@awoods187 please re-open if you disagree 

(PS gonna be closing out some roadmap docs issues this week for the release next Tuesday so I may ping you on a few as FYIs or with questions)

Issue #9708, opened 2021-02-03
Users can configure the active regions for a database


rmloveland:
See also:

- #9299
- #9300
- #9309
- #9557
- #9598

rmloveland:
All subtasks done, closing this one out!

PR #9900, opened 2021-02-26
Break out smaller topics from multiregion overview

rmloveland:
... with the goal of increasing ease of understanding and findability
from searches.

Summary of changes:

- Break out 'Choosing a multi-region configuration' from the bottom of
  the original overview page; also, add a "locality vs. survival"
  feature matrix-y table to make it easier (?) to understand

- Add a "When to use ZONE vs REGION" page for easier searching etc.

- Add a "When to use REGIONAL vs GLOBAL" page, also for easier searching
  etc.

Addresses #8958, #9708, #9709, #9884, #9886.

### DIFF
--- a/_includes/sidebar-data-v21.1.json
+++ b/_includes/sidebar-data-v21.1.json
@@ -621,6 +621,24 @@
                 "urls": [
                   "/${VERSION}/multiregion-overview.html"
                 ]
+              },
+              {
+                "title": "Choosing a multi-region configuration",
+                "urls": [
+                  "/${VERSION}/choosing-a-multi-region-configuration.html"
+                ]
+              },
+              {
+                "title": "When to use ZONE vs. REGION Survival Goals",
+                "urls": [
+                  "/${VERSION}/when-to-use-zone-vs-region-survival-goals.html"
+                ]
+              },
+              {
+                "title": "When to use REGIONAL vs. GLOBAL tables",
+                "urls": [
+                  "/${VERSION}/when-to-use-regional-vs-global-tables.html"
+                ]
               }
             ]
           },
--- a/v21.1/choosing-a-multi-region-configuration.md
+++ b/v21.1/choosing-a-multi-region-configuration.md
@@ -1,0 +1,55 @@
+---
+title: Choosing a multi-region configuration
+summary: Learn how to use CockroachDB's improved multi-region user experience.
+toc: true
+---
+
+This page has high-level information about how to configure a [multi-region cluster's](multiregion-overview.html) [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality).
+
+The options for configuring your multi-region cluster include:
+
+- _Change nothing_: Using the [default settings](multiregion-overview.html#default-settings), you get:
+  - Zone survival (the default).
+  - Low-latency reads and writes from a single region.
+  - Low-latency stale reads from all other regions.
+
+- _Change only [survival goals](multiregion-overview.html#survival-goals)_: This configuration is useful for single-region apps that need higher levels of survival. In this configuration, you move from availability zone (AZ) survival to get:
+  - Region survival.
+  - Low-latency reads from a single region.
+  - Low-latency stale reads from all other regions.
+  - Higher-latency writes from all regions (due to region survival).
+
+- _Change only [table locality](multiregion-overview.html#table-locality)_: This is useful for multi-region apps that require different read/write latency guarantees for different tables in the database, and are not concerned with surviving a region failure. In this configuration, you get:
+  - Zone survival (the default).
+  - Low-latency reads and writes from all regions.
+
+- _Change both [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality)_: This is useful for multi-region apps that want a high level of survival. In this configuration, you move from zone survival and get:
+  - Region survival.
+  - Low-latency reads from all regions.
+  - Higher-latency writes from all regions (due to region survival).
+
+The table below offers another view of how the configuration options described above map to:
+
+- The performance characteristics of specific survival goal/table locality combinations.
+- The types of applications that can benefit from each combination.
+
+| <div style="width: 200px;"></div> locality &#8595; survival &#8594; | `ZONE`                                                                                                                                 | `REGION`                                                                                                                                                                                                             |
+|---------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `REGIONAL BY TABLE`                                                 | Low-latency for single-region writes and multi-region stale reads.                                                                     | Single-region writes are higher latency than for ZONE, as at least one additional region must be consulted for each write.  Stale multi-region reads are of comparable latency to ZONE survival.                     |
+|                                                                     | For single-region apps that can accept risk of region failure.                                                                         | For single-region apps that must survive region failure.                                                                                                                                                             |
+| `REGIONAL BY ROW`                                                   | Low-latency consistent multi-region reads & writes for rows which are homed in specific regions.                                       | Low-latency consistent multi-region reads.  Low-latency single-region writes if you are writing to a row from its home region.                                                                                       |
+|                                                                     | For multi-region apps which read/write individual rows of the table from a specific region, and can accept the risk of region failure. | Same as for ZONE survival, but for apps that must survive a region failure.                                                                                                                                          |
+| `GLOBAL`                                                            | Low-latency multi-region reads. Writes are higher latency than reads.                                                                  | Low-latency multi-region reads.  Writes are higher latency than reads. There should be minimal difference in write latencies between ZONE and REGION survival due to a new custom non-blocking transaction protocol. |
+|                                                                     | For multi-region apps that need low-latency reads of a "read-mostly" table.                                                            | For multi-region apps that need low-latency reads of a "read-mostly" table, and the ability to survive region failures.                                                                                              |
+
+{{site.data.alerts.callout_info}}
+Different databases and tables within the same cluster can each use different combinations of the settings above.
+{{site.data.alerts.end}}
+
+## See also
+
+- [Multi-region overview](multiregion-overview.html)
+- [When to use `REGIONAL` vs. `GLOBAL` tables](when-to-use-regional-vs-global-tables.html)
+- [When to use `ZONE` vs. `REGION` survival goals](when-to-use-zone-vs-region-survival-goals.html)
+- [Topology Patterns](topology-patterns.html)
+- [Disaster Recovery](disaster-recovery.html)
--- a/v21.1/multiregion-overview.md
+++ b/v21.1/multiregion-overview.md
@@ -231,23 +231,13 @@ To optimize read access to the data in a table from any region (that is, globall
 ALTER TABLE <table> SET LOCALITY GLOBAL;
 ~~~
 
-## Choosing a multi-region configuration
+## Next steps
 
-In this section we will try to explain how to change [the default settings](#default-settings) of your multi-region cluster to meet your application's needs.
-
-The options for changing your multi-region cluster's configuration include:
-
-- _Change nothing_: Using the [default settings](#default-settings), you get zone survival, low-latency single-region writes, and multi-region stale reads.
-
-- _Change only [survival](#survival-goals)_: In this configuration, you upgrade from availability zone (AZ) to get region survival, single-region writes, and multi-region stale reads. This configuration is useful for single-region apps that need higher levels of survival.
-
-- _Change only [table locality](#table-locality)_: In this configuration, you accept the default zone survival, and optimize for low-latency multi-region reads and writes. This is useful for multi-region apps that want low-latency writes and are not concerned with surviving a region failure.
-
-- _Change both [survival](#survival-goals) and [table locality](#table-locality)_: In this configuration, you upgrade from zone to region survival, and optimize for low-latency multi-region reads. Single-region writes can be made low-latency with the right settings, but multi-region writes are higher-latency because they have to move across the network. This is useful for multi-region apps that want a high level of survival.
-
-Finally, note that different databases within the same cluster can each use different combinations of the settings above.
++ [Choosing a multi-region configuration](choosing-a-multi-region-configuration.html)
 
 ## See also
 
+- [When to use `ZONE` vs. `REGION` survival goals](when-to-use-zone-vs-region-survival-goals.html)
+- [When to use `REGIONAL` vs. `GLOBAL` tables](when-to-use-regional-vs-global-tables.html)
 - [Topology Patterns](topology-patterns.html)
 - [Disaster Recovery](disaster-recovery.html)
--- a/v21.1/when-to-use-regional-vs-global-tables.md
+++ b/v21.1/when-to-use-regional-vs-global-tables.md
@@ -1,0 +1,34 @@
+---
+title: When to use REGIONAL vs. GLOBAL tables
+summary: Learn how to use CockroachDB's improved multi-region user experience.
+toc: false
+---
+
+<span class="version-tag">New in v21.1:</span> [_Table Localities_](multiregion-overview.html#table-locality) tell CockroachDB how to optimize access to a table's data in a multi-region cluster.  CockroachDB uses the table locality setting to determine how to optimize access to the table's data from that locality.
+
+The following table localities are available:
+
+- `REGIONAL`
+- `GLOBAL`
+
+Use [`REGIONAL` tables](multiregion-overview.html#regional-by-row-tables) if:
+
+- Your application requires low-latency reads and writes from a single region (either at the [row level](multiregion-overview.html#regional-by-row-tables) or the [table level](multiregion-overview.html#regional-tables)).
+- Access to the table's data can be slower (higher latency) from other regions.
+
+Use [`GLOBAL` tables](multiregion-overview.html#global-tables) if:
+
+- Your application has a "read-mostly" table of reference data that is rarely updated, and that needs to be available to all regions.
+- You can accept that writes to the table will incur higher latencies from any given region, since writes use a novel non-blocking transaction protocol that uses a timestamp "in the future" (documentation forthcoming).
+
+{{site.data.alerts.callout_success}}
+For more information about how to choose an overall multi-region configuration, see [Choosing a multi-region configuration](choosing-a-multi-region-configuration.html).
+{{site.data.alerts.end}}
+
+## See also
+
+- [Multi-region Overview](multiregion-overview.html)
+- [Choosing a multi-region configuration](choosing-a-multi-region-configuration.html)
+- [When to use `ZONE` vs. `REGION` survival goals](when-to-use-zone-vs-region-survival-goals.html)
+- [Topology Patterns](topology-patterns.html)
+- [Disaster Recovery](disaster-recovery.html)
--- a/v21.1/when-to-use-zone-vs-region-survival-goals.md
+++ b/v21.1/when-to-use-zone-vs-region-survival-goals.md
@@ -1,0 +1,33 @@
+---
+title: When to use ZONE vs. REGION survival goals
+summary: Learn how to use CockroachDB's improved multi-region user experience.
+toc: false
+---
+
+<span class="version-tag">New in v21.1:</span> [_Survival Goals_](multiregion-overview.html#survival-goals) dictate how many simultaneous failure(s) a [multi-region database](multiregion-overview.html) can survive.  All tables within the same database operate with the same survival goal. Each database is allowed to have its own survival goal setting.
+
+Allowed survival goals include:
+
+- `ZONE` (default)
+- `REGION`
+
+Set a [`ZONE` survival goal](multiregion-overview.html#surviving-zone-failures) if:
+
+- You can accept a single node failure up to an entire zone failure. If multiple zones fail in the same region, the database may become unavailable.
+
+Set a [`REGION` survival goal](multiregion-overview.html#surviving-region-failures) if:
+
+- The database must remain available, even if a region goes down.
+- You can accept the performance cost: write latency will be increased by at least as much as the round-trip time to the nearest region. Read performance will be unaffected.
+
+{{site.data.alerts.callout_success}}
+For more information about how to choose a multi-region configuration, see [Choosing a multi-region configuration](choosing-a-multi-region-configuration.html).
+{{site.data.alerts.end}}
+
+## See also
+
++ [Multi-region overview](multiregion-overview.html)
++ [Choosing a multi-region configuration](choosing-a-multi-region-configuration.html)
++ [When to use `REGIONAL` vs `GLOBAL` tables](when-to-use-regional-vs-global-tables.html)
+- [Topology Patterns](topology-patterns.html)
+- [Disaster Recovery](disaster-recovery.html)


rmloveland:
Adam & Nathan: FYI this is kind of a "intermediate step" PR with the goal of breaking some stuff out of the giant overview page so we have bones to hang stuff on as I have started writing SQL reference pages so I have something to link to from upcoming tutorials.   The PR summary has the details - It's a combo of edits and a couple new pages for people who are just trying to find "when to choose A vs. B".

Would appreciate any feedback - esp. if the advice is wrong or needs clarifying! Thanks!

ajstorm (changes requested):
Nice to see the documentation take form.  

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUhxIeF6BOQi73EVbp8:b-7fu0va)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nvanbenschoten and @rmloveland)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhxO7e3_PXgjwQQ1co:-MUhxO7e3_PXgjwQQ1cp:b-cvu1zq) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L9)):*
> ```Markdown
> This page has high-level information about how to configure a [multi-region cluster's](multiregion-overview.html) [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality).
> 
> The options for changing your multi-region cluster's configuration include:
> ```

This section is great.  I wonder though if it might be better to frame this more from the user's perspective instead of from the system perspective.  For example, we could rewrite the above, and the first bullet point below as:

"The options for tailoring your multi-region cluster's configuration include:

If your application predominantly reads/writes from a single region, requires fast stale reads from the other regions, and must only tolerate zone failures, the default configuration is sufficient."

Then subsequent bullets would start with the customer use-case, and finish with what needs changing.  For someone new to the system, it might be a more natural way of consuming the information.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 13 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhzWXY03Cn5ANFDZKw:-MUhzWXY03Cn5ANFDZKx:b-jjifh3) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L13)):*
> ```Markdown
> - _Change nothing_: Using the [default settings](multiregion-overview.html#default-settings), you get zone survival, low-latency single-region writes, and multi-region stale reads.
> 
> - _Change only [survival goals](multiregion-overview.html#survival-goals)_: In this configuration, you upgrade from availability zone (AZ) to get region survival, single-region writes, and multi-region stale reads. This configuration is useful for single-region apps that need higher levels of survival.
> ```

I think we should avoid the work "upgrade" here, as the survivability gets better, but the latency will get worse.  Some more neutral words might be "move", "switch", "convert", "transition".
___
*[v21.1/choosing-a-multi-region-configuration.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhz2bpCF6qeSTli6Db:-MUhz2bpCF6qeSTli6Dc:btjbb45) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L15)):*
> ```Markdown
> - _Change only [survival goals](multiregion-overview.html#survival-goals)_: In this configuration, you upgrade from availability zone (AZ) to get region survival, single-region writes, and multi-region stale reads. This configuration is useful for single-region apps that need higher levels of survival.
> 
> - _Change only [table locality](multiregion-overview.html#table-locality)_: In this configuration, you accept the default zone survival, and optimize for low-latency multi-region reads and writes. This is useful for multi-region apps that want low-latency writes and are not concerned with surviving a region failure.
> ```

I worry that this last line could be a bit misleading. For example, GLOBAL tabled don't guarantee low-latency writes, so we could be setting the wrong expectations here.  What about something more neutral, like "This is useful for multi-region application that require different read/write latency guarantees for different tables in the database (see below for details)"?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhztoa5y9Rf6PMNIGJ:-MUhztoa5y9Rf6PMNIGK:b-2ykj3m) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L17)):*
> Single-region writes can be made low-latency with the right settings, but multi-region writes are higher-latency because they have to move across the network.

This might be confusing without additional context.  We might want to defer this detail to the table below.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 26 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi-NjiBAM2rIdczoSX:-MUi-NjiBAM2rIdczoSY:bn2vcqo) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L26)):*
> ```Markdown
> | locality &#8595; survival &#8594; | `ZONE`                                                                                 | `REGION`                                                                                            |
> |-----------------------------------+----------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------|
> | `REGIONAL BY TABLE`               | Low-latency single-region writes, multi-region stale reads.                            | Single-region writes, multi-region stale reads.                                                     |
> ```

Nit: The comma here is throwing me off.  I think what we mean on the ZONE side is "Low-latency for single region writes and multi-region stale reads".  On the REGION side, how about "Single region writes are higher latency than with ZONE survival, as at least one additional region must be consulted for each write.  Stale reads are of comparable latency to ZONE survival."
___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUi0FUZ1lOXErIWUC96:b-950vni) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*
> ```Markdown
> | `REGIONAL BY TABLE`               | Low-latency single-region writes, multi-region stale reads.                            | Single-region writes, multi-region stale reads.                                                     |
> |                                   | For single-region apps that can accept risk of region failure.                         | For single-region apps that must survive region failure.                                            |
> | `REGIONAL BY ROW`                 | Low-latency multi-region reads & writes.                                               | Low-latency multi-region reads. Low-latency single-region writes with the right settings.           |
> ```

I think the first column here could do with some clarification, because as it reads, it sounds like the best option (low latency for both reads and writes).  We should add here that the latency guarantees are only for rows which live in the targeted region, and only if the read/write access is also within that region.  For the right column, can we be more specific about what we mean by "with the right settings"?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi1GP67WfyCkihALnk:-MUi1GP67WfyCkihALnl:b-wesk2) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L29)):*
> For multi-region apps that need low-latency writes, can accept risk of region failure. 

I think we need something more nuanced here.  Something like "For multi-region applications which read/write individual rows of the table from a specific region" and we might want to provide an example.  Similarly, here:

> For multi-region apps that want a high level of survival.

We should probably say something like "Same requirements, but requires the ability to survive a region failure."
___
*[v21.1/choosing-a-multi-region-configuration.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi2BdVCnA3TFJQO0mY:-MUi2BdVCnA3TFJQO0mZ:b-5r414o) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L30)):*
> ```Markdown
> | `REGIONAL BY ROW`                 | Low-latency multi-region reads & writes.                                               | Low-latency multi-region reads. Low-latency single-region writes with the right settings.           |
> |                                   | For multi-region apps that need low-latency writes, can accept risk of region failure. | For multi-region apps that want a high level of survival.                                           |
> | `GLOBAL`                          | Low-latency multi-region reads. High-latency writes.                                   | Low-latency multi-region reads.  High-latency writes due to cross-region replication.               |
> ```

Nit: We might want to reword "High-latency writes", as they sound pretty scary.  Maybe we can say something like "writes are higher latency as they're prioritized lower than reads" (not even sure about this, but maybe @nvanbenschoten has a good thought here).  Also, do we think we need to call out the "cross-region replication"?  Won't this be a smaller impact to latency than writing into the future? 
___
*[v21.1/choosing-a-multi-region-configuration.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi6GIAEr2gNQSG76fs:-MUi6GIAEr2gNQSG76ft:b-9e9cxu) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L31)):*
> high level of survival.

Nit: "...and the ability to survive region failures"?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 34 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi6mOb-XpZCmkay91b:-MUi6mOb-XpZCmkay91c:b-md06en) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L34)):*
> ```Markdown
> 
> Different databases within the same cluster can each use different combinations of the settings above.
> ```

database and tables?
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 7 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi7v-u5mlGJxkZjJiU:-MUi7v-u5mlGJxkZjJiV:bui53xu) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L7)):*
> ```Markdown
> ---
> 
> <span class="version-tag">New in v21.1:</span> [_Table Localities_](multiregion-overview.html#table-locality)tell CockroachDB how to optimize access to a table's data in a multi-region cluster.  CockroachDB uses the table locality setting to determine how to optimize access to the table's data from that locality.
> ```

Nit: Space before "tell".
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 16 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi83MI0PfbwEyQ1z7b:-MUi83MI0PfbwEyQ1z7c:b-mhhvv5) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L16)):*
> ```Markdown
> Use [`REGIONAL` tables](multiregion-overview.html#regional-by-row-tables) if:
> 
> - Your application requires low-latency reads and writes from a single region.
> ```

Nit: Might benefit from adding (either at a table level, or for individual rows of a table) at the end of this line.
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi8cfS7Fgg1cXvfv-w:-MUi8cfS7Fgg1cXvfv-x:b-rh83ed) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L23)):*
> ```Markdown
> 
> - Your application has a "read-mostly" table of reference data that is rarely updated, and that needs to be available to all regions.
> - You can accept that writes to the table will incur higher latencies from any given region, since writes have to be replicated across every region to make the global low-latency reads possible.
> ```

Nit: I think this is a good way of thinking about things from a customer perspective, but it's not 100% technically accurate.  For example, if you're in ZONE survivability, and you're writing to the PRIMARY region, the higher latency isn't caused by replication, but rather, is a result of the fact that we're writing into the future.  I know that this is far more nuance than we want customers to consume, but I worry that the wording as-is will cause confusion for super-users.  
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi9qMU5vh1A3hxbkcV:-MUi9qMU5vh1A3hxbkcW:b-bla6k1) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-zone-vs-region-survival-goals.md#L17)):*
> ```Markdown
> 
> - You can accept that if multiple zones fail in the same region, the database may be unavailable.
> - You need better performance: low-latency single-region writes, and multi-region stale reads.
> ```

Nit: The "low-latency single-region writes" part has some caveats (i.e. it only applies to REGIONAL tables, and even then, only in specific regions).  Wondering if we want to drop it from here.

rmloveland (commented):
Thanks for the review Adam! I think almost every item of feedback is addressed, I just had 1-2 questions below.  PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUp8hZdEBOX3gT_nJVU:bsg4v8f)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm and @nvanbenschoten)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhxO7e3_PXgjwQQ1co:-MUp1AFYDxByhZSfwfI1:b-6wzd59) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L9)):*

This section is great.  I wonder though if it might be better to frame this more from the user's perspective instead of from the system perspective.  For example, we could rewrite the above, and the first bullet point below as:

"The options for tailoring your multi-region cluster's configuration include:

If your application predominantly reads/writes from a single region, requires fast stale reads from the other regions, and must only tolerate zone failures, the default configuration is sufficient."

Then subsequent bullets would start with the customer use-case, and finish with what needs changing.  For someone new to the system, it might be a more natural way of consuming the information.
</blockquote></details>

Changed the intro to say "The options for configuring your multi-region cluster include:"

Re: changing the sub-topics, I would prefer not to rewrite them for now since they are copied from something we already merged - or should I say I'd rather not rewrite them ... yet.  I need to get more stuff up on the site and cross-link it around.  As we get closer to the release and more things are filled in, I will do a review of everything and revise based on the (hopefully) better, simpler understandings/explanations we have at that time.  Right now I want to work a bit more breadth-first.  But I totes agree they can/will be made better.  Hope that's not too annoying!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 13 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhzWXY03Cn5ANFDZKw:-MUp1lWJ4bNh5vfOYaXg:b3oyl6j) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L13)):*

I think we should avoid the work "upgrade" here, as the survivability gets better, but the latency will get worse.  Some more neutral words might be "move", "switch", "convert", "transition".
</blockquote></details>

Good idea - went with "move"
___
*[v21.1/choosing-a-multi-region-configuration.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhz2bpCF6qeSTli6Db:-MUp2-KA8EBO-YRZ5oBY:b4nhycr) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L15)):*

I worry that this last line could be a bit misleading. For example, GLOBAL tabled don't guarantee low-latency writes, so we could be setting the wrong expectations here.  What about something more neutral, like "This is useful for multi-region application that require different read/write latency guarantees for different tables in the database (see below for details)"?
</blockquote></details>

Updated to that language!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhztoa5y9Rf6PMNIGJ:-MUp221N0lY07QPIPncn:btp634c) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L17)):*

> Single-region writes can be made low-latency with the right settings, but multi-region writes are higher-latency because they have to move across the network.

This might be confusing without additional context.  We might want to defer this detail to the table below.
</blockquote></details>

I agree - I'd rather leave it for now since these are alpha docs anyway, and this is language we already merged in another PR.  As more stuff gets written I will have more context to link to, and more ways of saying stuff.  And as above I def agree we should do what you are saying!  Will end up revising this again as we get later into cycle.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 26 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi-NjiBAM2rIdczoSX:-MUp2hjn1vpMdpSpkNVh:b-tyi8u1) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L26)):*

Nit: The comma here is throwing me off.  I think what we mean on the ZONE side is "Low-latency for single region writes and multi-region stale reads".  On the REGION side, how about "Single region writes are higher latency than with ZONE survival, as at least one additional region must be consulted for each write.  Stale reads are of comparable latency to ZONE survival."
</blockquote></details>

Updated to use that language, thanks!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUp4-DQ71WToc8hJuZ6:b-qrmb1d) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*

I think the first column here could do with some clarification, because as it reads, it sounds like the best option (low latency for both reads and writes).  We should add here that the latency guarantees are only for rows which live in the targeted region, and only if the read/write access is also within that region.  For the right column, can we be more specific about what we mean by "with the right settings"?
</blockquote></details>

Hmm I actually do not know what "the right settings" are, this is something Nathan typed in a comment somewhere.  Do you know what they are?  Sorry, guess I should have asked this already.  Is the answer "RBR + when you are writing to a row whose home region is also the DB's primary region"?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 29 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi1GP67WfyCkihALnk:-MUp350m0ijyFM8l8OM6:b-mi4kce) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L29)):*

> For multi-region apps that need low-latency writes, can accept risk of region failure. 

I think we need something more nuanced here.  Something like "For multi-region applications which read/write individual rows of the table from a specific region" and we might want to provide an example.  Similarly, here:

> For multi-region apps that want a high level of survival.

We should probably say something like "Same requirements, but requires the ability to survive a region failure."

</blockquote></details>

Updated to use approx that language!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi2BdVCnA3TFJQO0mY:-MUp4vBK1XDT6kZCpgFi:b73ni3v) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L30)):*

Nit: We might want to reword "High-latency writes", as they sound pretty scary.  Maybe we can say something like "writes are higher latency as they're prioritized lower than reads" (not even sure about this, but maybe @nvanbenschoten has a good thought here).  Also, do we think we need to call out the "cross-region replication"?  Won't this be a smaller impact to latency than writing into the future? 
</blockquote></details>

How about "writes are higher latency than reads" without mentioning prioritization (yet), since we don't have docs on non-blocking transactions yet?  (I have read the RFC but only once so far so not 100% familiar)

Re: cross-region replication, IDK - I am happy to use different language.  I guess psychologically we need to give the reader something that is "different" in the REGION survival column that may explain any write latency differences (if any?).  

If there is no difference in write latency between ZONE and REGION we could say something like "there should be minimal difference in write latencies between ZONE and REGION survival due to a new custom non-blocking transaction protocol" or similar and leave it at that, and come back and link arch docs once written

Nathan, what do you think?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi6GIAEr2gNQSG76fs:-MUp3D_d4mnZz5ILPxd9:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L31)):*

> high level of survival.

Nit: "...and the ability to survive region failures"?
</blockquote></details>

Updated!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 34 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi6mOb-XpZCmkay91b:-MUp3EgE3wFp7EpLS7Hk:biai55g) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L34)):*

database and tables?
</blockquote></details>

Yes! Updated to say "different DBs and tables"
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 7 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi7v-u5mlGJxkZjJiU:-MUp6C4F6C7ViVAKwAvb:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L7)):*

Nit: Space before "tell".
</blockquote></details>

Fixed!
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 16 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi83MI0PfbwEyQ1z7b:-MUp6KH2AU2kDIGa6IfA:b-6h9vrp) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L16)):*

Nit: Might benefit from adding (either at a table level, or for individual rows of a table) at the end of this line.
</blockquote></details>

Added!

FYI also removed bullet 3 about "controlling where data is located" since I understand from some comments by Nathan elsewhere that this is Not A Thing (tm) - everything is everywhere.  Still drilling this into my brain
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi8cfS7Fgg1cXvfv-w:-MUp6gu_1ZGeb3FhwsXb:b-af3hlz) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L23)):*

Nit: I think this is a good way of thinking about things from a customer perspective, but it's not 100% technically accurate.  For example, if you're in ZONE survivability, and you're writing to the PRIMARY region, the higher latency isn't caused by replication, but rather, is a result of the fact that we're writing into the future.  I know that this is far more nuance than we want customers to consume, but I worry that the wording as-is will cause confusion for super-users.  
</blockquote></details>

Agreed!  As discussed in another comment we should have docs for non-blocking transactions later in the cycle.  What do you think about:

" ... since writes use a novel non-blocking transaction protocol that uses a timestamp "in the future" (documentation forthcoming)."
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi9qMU5vh1A3hxbkcV:-MUp7Lv60XqER3Q0WqSD:b5q6gjf) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-zone-vs-region-survival-goals.md#L17)):*

Nit: The "low-latency single-region writes" part has some caveats (i.e. it only applies to REGIONAL tables, and even then, only in specific regions).  Wondering if we want to drop it from here.
</blockquote></details>

Dropped it - can let the description of the higher latency cost for SURVIVE REGION below do the talking

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm and @nvanbenschoten)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhztoa5y9Rf6PMNIGJ:-MUp9fqn8JcOm6VV2tZF:b-53ss84) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L17)):*

I agree - I'd rather leave it for now since these are alpha docs anyway, and this is language we already merged in another PR.  As more stuff gets written I will have more context to link to, and more ways of saying stuff.  And as above I def agree we should do what you are saying!  Will end up revising this again as we get later into cycle.
</blockquote></details>

actually belay that - we discuss what the "right settings" are below in a different comment.  will update this based on that and/or defer as you suggest once we figure that out

ajstorm (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nvanbenschoten)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhxO7e3_PXgjwQQ1co:-MUpvdA332Sa36MHK6Ad:bb73djm) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L9)):*

Changed the intro to say "The options for configuring your multi-region cluster include:"

Re: changing the sub-topics, I would prefer not to rewrite them for now since they are copied from something we already merged - or should I say I'd rather not rewrite them ... yet.  I need to get more stuff up on the site and cross-link it around.  As we get closer to the release and more things are filled in, I will do a review of everything and revise based on the (hopefully) better, simpler understandings/explanations we have at that time.  Right now I want to work a bit more breadth-first.  But I totes agree they can/will be made better.  Hope that's not too annoying!
</blockquote></details>

SGTM
___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUpwN3RAnKNx9kv-7ez:b-d9x8q3) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*

Hmm I actually do not know what "the right settings" are, this is something Nathan typed in a comment somewhere.  Do you know what they are?  Sorry, guess I should have asked this already.  Is the answer "RBR + when you are writing to a row whose home region is also the DB's primary region"?
</blockquote></details>

@nvanbenschoten  for "right settings" are we referring to the optimizations that Becca put in?  Something else?
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi8cfS7Fgg1cXvfv-w:-MUpxl6k-oGJQdR_Plyf:b-bk5ip0) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/when-to-use-regional-vs-global-tables.md#L23)):*

Agreed!  As discussed in another comment we should have docs for non-blocking transactions later in the cycle.  What do you think about:

" ... since writes use a novel non-blocking transaction protocol that uses a timestamp "in the future" (documentation forthcoming)."
</blockquote></details>

Works for me.

nvanbenschoten (commented):
Looking good!

> Reviewed 2 of 5 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUtN-kL7dI1PBXzXid0:br2sbr6)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm and @rmloveland)

___
*[_includes/sidebar-data-v21.1.json, line 628 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3g_K8eipy-dub93G:-MUt3g_K8eipy-dub93H:b-35yqda) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/_includes/sidebar-data-v21.1.json#L628)):*
> ```JSON
>                 "title": "Choosing a multi-region configuration",
>                 "urls": [
>                   "/${VERSION}/multiregion-overview.html"
> ```

This link is still going to the overview. Took me a few clicks to realize.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUtHZ-BAYFIDYh6gnbJ:b-py93yn) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*
> Is the answer "RBR + when you are writing to a row whose home region is also the DB's primary region"?

I think this is what I meant, and it relates to Adam's initial comment on this thread.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi2BdVCnA3TFJQO0mY:-MUtHk5m6pApqHLTJMOR:b-s4g66x) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L30)):*
Even just "higher-latency instead of "high-latency" SGTM.

> If there is no difference in write latency between ZONE and REGION we could say something like "there should be minimal difference in write latencies between ZONE and REGION survival due to a new custom non-blocking transaction protocol" or similar and leave it at that, and come back and link arch docs once written

:thumbsup: 
___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3zeK3lWuHIg-bQrn:-MUt3zeK3lWuHIg-bQro:bjp9acn) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L9)):*
> ```Markdown
> This page has high-level information about how to configure a [multi-region cluster's](multiregion-overview.html) [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality).
> 
> The options for configuring your multi-region cluster include:
> ```

nit: the spacing seems a little compressed between this line and the next. Not sure if there's anything to do.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 11 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3uH-D8JIvWpSBEvU:-MUt3uH-D8JIvWpSBEvV:bt4w74t) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L11)):*
> ```Markdown
> The options for configuring your multi-region cluster include:
> 
> - _Change nothing_: Using the [default settings](multiregion-overview.html#default-settings), you get zone survival, low-latency single-region writes, and multi-region stale reads.
> ```

"single-region reads and write"

Same thing in the next option.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 11 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt5lkD9Z06JDxmz5Cw:-MUt5lkD9Z06JDxmz5Cx:bxbya4l) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L11)):*
> ```Markdown
> The options for configuring your multi-region cluster include:
> 
> - _Change nothing_: Using the [default settings](multiregion-overview.html#default-settings), you get zone survival, low-latency single-region writes, and multi-region stale reads.
> ```

The "low-latency/high-latency" vs. "single-region/multi-region" distinctions are subtle. I wonder if there's anything better we could say here to distinguish them. Should we be using "from a single region"/"from all regions" to drive home the idea that we're talking about access performance? So something like:

```
you get zone survival, low-latency reads and writes from a single region, and low-latency stale reads from all other regions

you get region survival, low-latency reads from a single region, higher-latency writes from a single region, and low-latency stale reads from all other regions

... low-latency reads and writes from all regions

... low-latency reads from all regions and higher-latency writes from all regions
```
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4d_Y4VPOAKabcipg:-MUt4d_Y4VPOAKabciph:b-veh5b6) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*
> ```Markdown
> - _Change only [table locality](multiregion-overview.html#table-locality)_: In this configuration, you accept the default zone survival, and optimize for low-latency multi-region reads and writes. This is useful for multi-region apps that require different read/write latency guarantees for different tables in the database, and are not concerned with surviving a region failure.
> 
> - _Change both [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality)_: In this configuration, you upgrade from zone to region survival, and optimize for low-latency multi-region reads.  Multi-region writes are higher-latency because they have to move across the network. This is useful for multi-region apps that want a high level of survival.
> ```

Should we use "move" instead of "upgrade" here as well?
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4lfDClCmY4E6ZPWJ:-MUt4lfDClCmY4E6ZPWK:b-eb1qd4) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*
> ```Markdown
> - _Change only [table locality](multiregion-overview.html#table-locality)_: In this configuration, you accept the default zone survival, and optimize for low-latency multi-region reads and writes. This is useful for multi-region apps that require different read/write latency guarantees for different tables in the database, and are not concerned with surviving a region failure.
> 
> - _Change both [survival goals](multiregion-overview.html#survival-goals) and [table locality](multiregion-overview.html#table-locality)_: In this configuration, you upgrade from zone to region survival, and optimize for low-latency multi-region reads.  Multi-region writes are higher-latency because they have to move across the network. This is useful for multi-region apps that want a high level of survival.
> ```

It's confusing that we mention writes being slower in this option, but not in the "Change only survival goals" option.

We should either hint that writes are slower (independent of locality) for REGION survivability across this overview, or ignore it.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 24 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt6v9V6c3r9cIr_GW-:-MUt6v9WBate3LeyAa40:bkuojfm) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L24)):*
> ```Markdown
> - The types of applications that can benefit from each combination.
> 
> | locality &#8595; survival &#8594; | `ZONE`                                                                                                                                 | `REGION`                                                                                                                                                                                         |
> ```

nit: the first column is getting compressed
___
*[v21.1/multiregion-overview.md, line 240 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUtI-5w9ItYHI3rj2XH:-MUtI-5w9ItYHI3rj2XI:b6zufbw) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/multiregion-overview.md#L240)):*
> ```Markdown
> ## See also
> 
> - [When to use `REGIONAL` vs. `GLOBAL` tables](when-to-use-regional-vs-global-tables.html)
> ```

Should we reverse this order? We talk about ZONE vs. REGION before REGIONAL vs. GLOBAL everywhere else.
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 16 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUtMFCaA2tJpJmXtji4:-MUtMFCbAudLDivNZBEl:b-kff90) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/when-to-use-zone-vs-region-survival-goals.md#L16)):*
> ```Markdown
> Set a [`ZONE` survival goal](multiregion-overview.html#surviving-zone-failures) if:
> 
> - You can accept that if multiple zones fail in the same region, the database may be unavailable.
> ```

It would be worth mentioning what we _can_ survive with this setting - a single node failure up to an entire zone failure.

rmloveland (commented):
Thanks Nathan!  I think I fixed all the things. PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUtz5Su8YXqullEyP1V:b7pdgkq)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm and @nvanbenschoten)

___
*[_includes/sidebar-data-v21.1.json, line 628 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3g_K8eipy-dub93G:-MUtq6IH5SMmLBgwlAzb:b-ly4np3) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/_includes/sidebar-data-v21.1.json#L628)):*

This link is still going to the overview. Took me a few clicks to realize.
</blockquote></details>

Ah! Fixed
___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUhxO7e3_PXgjwQQ1co:-MUtwm5C2serWXyUpKGk:bfbz85i) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L9)):*

SGTM
</blockquote></details>

(FYI ended up changing subtopics mostly as you requested - see below)
___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUtxJlGDj-HbkFifl6t:bmz57iu) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*

> Is the answer "RBR + when you are writing to a row whose home region is also the DB's primary region"?

I think this is what I meant, and it relates to Adam's initial comment on this thread.
</blockquote></details>

Updated to say just that! (RBR + when you are writing to row whose home is also primary)
___
*[v21.1/choosing-a-multi-region-configuration.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi2BdVCnA3TFJQO0mY:-MUtxqO8AexgQFmGk3f1:b-atafzz) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L30)):*

Even just "higher-latency instead of "high-latency" SGTM.

> If there is no difference in write latency between ZONE and REGION we could say something like "there should be minimal difference in write latencies between ZONE and REGION survival due to a new custom non-blocking transaction protocol" or similar and leave it at that, and come back and link arch docs once written

:thumbsup: 
</blockquote></details>

Cool - left at "higher-latency"

Added the bit about "minimal difference in write latencies between ZONE and REGION survival due to a new custom non-blocking transaction protocol" 
___
*[v21.1/choosing-a-multi-region-configuration.md, line 9 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3zeK3lWuHIg-bQrn:-MUtqKnvBA56d8Kbt-Kh:bdfn0an) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L9)):*

nit: the spacing seems a little compressed between this line and the next. Not sure if there's anything to do.
</blockquote></details>

agreed, filed an issue against the recent site redesign: https://github.com/cockroachdb/docs/issues/9924
___
*[v21.1/choosing-a-multi-region-configuration.md, line 11 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt3uH-D8JIvWpSBEvU:-MUtsZsaBN3KXrsmwo8W:b-bu91xm) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L11)):*

"single-region reads and write"

Same thing in the next option.
</blockquote></details>

Updated
___
*[v21.1/choosing-a-multi-region-configuration.md, line 11 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt5lkD9Z06JDxmz5Cw:-MUtts_1DPfwKNV8jJjt:bf4e2qi) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L11)):*

The "low-latency/high-latency" vs. "single-region/multi-region" distinctions are subtle. I wonder if there's anything better we could say here to distinguish them. Should we be using "from a single region"/"from all regions" to drive home the idea that we're talking about access performance? So something like:

```
you get zone survival, low-latency reads and writes from a single region, and low-latency stale reads from all other regions

you get region survival, low-latency reads from a single region, higher-latency writes from a single region, and low-latency stale reads from all other regions

... low-latency reads and writes from all regions

... low-latency reads from all regions and higher-latency writes from all regions
```
</blockquote></details>

I like this way of framing it better.  I revised each of the bullets in this section to use this language, and also to use a nested list for better scanning, e.g.

- change only survival goals: ... you get:
   - region survival
   - ...

Let me know what you think!

(Also, Adam - this means I ended up taking your suggestion after all to move the "this is useful for apps doing X" to the front, yay)
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4d_Y4VPOAKabcipg:-MUttvwmB3waf-ewVFgd:bjhrbs9) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*

Should we use "move" instead of "upgrade" here as well?
</blockquote></details>

Yes, updated.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4lfDClCmY4E6ZPWJ:-MUtuMID89ifv6gGYltM:b-bm0zqy) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*

It's confusing that we mention writes being slower in this option, but not in the "Change only survival goals" option.

We should either hint that writes are slower (independent of locality) for REGION survivability across this overview, or ignore it.
</blockquote></details>

I agree.  It's much clearer in the revised version because with everything in list form it's more scannable.  

I added the hint to both of the REGION survival options.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 24 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt6v9V6c3r9cIr_GW-:-MUtv1a-8B0fPQFrG3gI:b-2353nk) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L24)):*

nit: the first column is getting compressed
</blockquote></details>

Oh man these narrow layouts!!!

Added a rather yucky div tag with inline CSS based on something I saw done in docs#9833

seems to be fixed now (tm)

yay markdown!!!!
___
*[v21.1/multiregion-overview.md, line 240 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUtI-5w9ItYHI3rj2XH:-MUty0076xcwa5muZQvp:b-6ps2z8) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/multiregion-overview.md#L240)):*

Should we reverse this order? We talk about ZONE vs. REGION before REGIONAL vs. GLOBAL everywhere else.
</blockquote></details>

Agreed - swapped 'em!
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 16 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUtMFCaA2tJpJmXtji4:-MUtyHBv7uQO965nE-oi:bn45ogi) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/when-to-use-zone-vs-region-survival-goals.md#L16)):*

It would be worth mentioning what we _can_ survive with this setting - a single node failure up to an entire zone failure.
</blockquote></details>

OK! Changed to "- You can accept a single node failure up to an entire zone failure. If multiple zones fail in the same region, the database may become unavailable."

nvanbenschoten (approved):
:lgtm:

> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUuFzIw7z4TaP7ju9Fd:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajstorm, @nvanbenschoten, and @rmloveland)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUuFPQV7N1Eeke3pBGe:b-6g5uds) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*

Updated to say just that! (RBR + when you are writing to row whose home is also primary)
</blockquote></details>

Actually, this doesn't have to do with the primary region at all. It's just if you are reading from or writing to a row from its home region.

In other words, the only different between this and the ZONE version is that writes are a little slower.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4lfDClCmY4E6ZPWJ:-MUuFGiTBgJRJB6DWvnk:b-j45s0z) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*

I agree.  It's much clearer in the revised version because with everything in list form it's more scannable.  

I added the hint to both of the REGION survival options.
</blockquote></details>

This looks great!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 18 at r3](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUuF4Fo5JxXOHKT3Yk1:-MUuF4Fo5JxXOHKT3Yk2:b-ahx46w) ([raw file](https://github.com/cockroachdb/docs/blob/08d7130fb2fef8bd4007a04af5ac5dec6810d849/v21.1/choosing-a-multi-region-configuration.md#L18)):*
> ```Markdown
> - _Change only [survival goals](multiregion-overview.html#survival-goals)_: This configuration is useful for single-region apps that need higher levels of survival. In this configuration, you move from availability zone (AZ) survival to get:
>   - Region survival.
>   - Low-latency reads and writes from a single region.
> ```

~and writes~

rmloveland (commented):
Thanks Nathan!  Fixed the one thing about row's home vs. DB primary.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUuIRwq3cojyEp4ysYg:b-7kujxv)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @ajstorm and @nvanbenschoten)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 28 at r1](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUi0FUZ1lOXErIWUC95:-MUuHuC_2hs33NYN-zWN:b-lro2p4) ([raw file](https://github.com/cockroachdb/docs/blob/e3863d96335ee8b62ad75d30ca8f4aef6dc892f4/v21.1/choosing-a-multi-region-configuration.md#L28)):*

Actually, this doesn't have to do with the primary region at all. It's just if you are reading from or writing to a row from its home region.

In other words, the only different between this and the ZONE version is that writes are a little slower.
</blockquote></details>

Updated to say "from its home region".  Gosh I'm gonna start getting this right soon, I swear.
___
*[v21.1/choosing-a-multi-region-configuration.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUt4lfDClCmY4E6ZPWJ:-MUuIQcqCEAAwkjCnDHH:b-ucfknh) ([raw file](https://github.com/cockroachdb/docs/blob/95d931c761077ae78f4253d46517859a3ef97135/v21.1/choosing-a-multi-region-configuration.md#L17)):*

This looks great!
</blockquote></details>

Thanks!
___
*[v21.1/choosing-a-multi-region-configuration.md, line 18 at r3](https://reviewable.io/reviews/cockroachdb/docs/9900#-MUuF4Fo5JxXOHKT3Yk1:-MUuIP4EDfvzdM4B7Vnf:b-896fja) ([raw file](https://github.com/cockroachdb/docs/blob/08d7130fb2fef8bd4007a04af5ac5dec6810d849/v21.1/choosing-a-multi-region-configuration.md#L18)):*

~and writes~
</blockquote></details>

Done!

nvanbenschoten (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUuRGlJ9thhLU7A2irb:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajstorm, @lnhsingh, and @nvanbenschoten)

ajstorm (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MUyzwel-5J5ZkIfwtwc:bg7igkv)* status: :shipit: complete! 2 of 0 LGTMs obtained (waiting on @ajstorm, @lnhsingh, and @nvanbenschoten)

lnhsingh (commented):
<img class="emoji" title=":lgtm:" alt=":lgtm:" align="absmiddle" src="https://reviewable.io/lgtm.png" height="20" width="61"/> but added some comments to annoy you, since you do that to me :) 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MVHXeVeDyFSklR1u-Mq:b-5inl44)* status: :shipit: complete! 3 of 0 LGTMs obtained (waiting on @ajstorm, @nvanbenschoten, and @rmloveland)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 36 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHUSOf0JR0M821ezci:-MVHUSOf0JR0M821ezcj:b-wz8uju) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/choosing-a-multi-region-configuration.md#L36)):*
> ```Markdown
> - The types of applications that can benefit from each combination.
> 
> | <div style="width: 200px;"></div> locality &#8595; survival &#8594; | `ZONE`                                                                                                                                 | `REGION`                                                                                                                                                                                                             |
> ```

This might be a little easier to read if the info was in a bulleted list within the table cell (or if the locality headers merged across the two cells), but I assume you laid out the table the way you did so you didn't have to use HTML to create the list. 
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 4 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHWXwf8q229VSZYVdb:-MVHWXwf8q229VSZYVdc:bwdmgn3) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-regional-vs-global-tables.md#L4)):*
> ```Markdown
> title: When to use REGIONAL vs. GLOBAL tables
> summary: Learn how to use CockroachDB's improved multi-region user experience.
> toc: true
> ```

For pages that don't have any headings other than See Also, should we remove the TOC? You don't have to change anything here, but maybe this is something I should bring up for the style guide
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 15 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHX4G91xX4dQtj8kEt:-MVHX4G91xX4dQtj8kEu:b-tpg0h3) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-regional-vs-global-tables.md#L15)):*
> ```Markdown
> 
> Use [`REGIONAL` tables](multiregion-overview.html#regional-by-row-tables) if:
> ```

Part of me thinks that this info would be better in a table, that way you could combine the info above this ("The following table localities are available") and when you should use each type of table.
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 14 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHXaDCEPNfx2smdUex:-MVHXaDCEPNfx2smdUey:b-dlfghw) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-zone-vs-region-survival-goals.md#L14)):*
> ```Markdown
> - `REGION`
> 
> Set a [`ZONE` survival goal](multiregion-overview.html#surviving-zone-failures) if:
> ```

Same comment here about maybe presenting this info as a table, but feel free to ignore!


<!-- Sent from Reviewable.io -->

rmloveland (commented):
Thanks Lauren!  Your comments could never annoy me (well, probably) :-)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9900#-:-MVHofcA08o_ozzCKe4Z:b-8tazfl)* status: :shipit: complete! 3 of 0 LGTMs obtained (waiting on @ajstorm, @lnhsingh, and @nvanbenschoten)

___
*[v21.1/choosing-a-multi-region-configuration.md, line 36 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHUSOf0JR0M821ezci:-MVHnXlH8K3aOl5PVqRd:b6abvbx) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/choosing-a-multi-region-configuration.md#L36)):*

This might be a little easier to read if the info was in a bulleted list within the table cell (or if the locality headers merged across the two cells), but I assume you laid out the table the way you did so you didn't have to use HTML to create the list. 
</blockquote></details>

Yes that's correct!  I really wish Markdown (or whatever this is) had sane tables.  Sorry it's so hard to read.
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 4 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHWXwf8q229VSZYVdb:-MVHnxKF6HKGZo3RStbx:b-s8h278) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-regional-vs-global-tables.md#L4)):*

For pages that don't have any headings other than See Also, should we remove the TOC? You don't have to change anything here, but maybe this is something I should bring up for the style guide
</blockquote></details>

I agree we should remove it - done here and on the other page where this was the case
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 15 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHX4G91xX4dQtj8kEt:-MVHoG5k1p23uCsNh6Vu:bguxbc6) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-regional-vs-global-tables.md#L15)):*

Part of me thinks that this info would be better in a table, that way you could combine the info above this ("The following table localities are available") and when you should use each type of table.
</blockquote></details>

All else being equal I tend to agree but I'm reluctant b/c using tables is so hard in markdown!
___
*[v21.1/when-to-use-zone-vs-region-survival-goals.md, line 14 at r4](https://reviewable.io/reviews/cockroachdb/docs/9900#-MVHXaDCEPNfx2smdUex:-MVHo_3xA96flTqrlPhw:b-8d1ran) ([raw file](https://github.com/cockroachdb/docs/blob/eef0de44eb6d7ee1e89fc5cfaf857b5004f712cc/v21.1/when-to-use-zone-vs-region-survival-goals.md#L14)):*

Same comment here about maybe presenting this info as a table, but feel free to ignore!
</blockquote></details>

I never ignore your comments!

But I am reluctant for the reason stated in my reply above

PR #9909, opened 2021-03-02
Add clean target and ensure deps are created in vendor.

### LINKED ISSUE #9908: Add a clean target and move `bundle install` to the `vendor` target in the Makefile
The `bootstrap` target won't actually make sure the gems are installed as long as the `vendor` directory exists, even if the gem dependencies changed in `Gemfile`/`Gemfile.lock`.

To fix this, we can add a `clean` target that removes any existing `vendor` directory, and move the `bundle install` command to the `vendor` target to make sure it is run after installing the `bundler` gem.

We should also make sure the `PATH` is compatible with `rbenv`.

ianjevans:
Fixes #9908.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@
 # Note that we also amend the path to include /usr/local/opt/ruby/bin, which
 # is where Homebrew will install a more up-to-date version of Ruby on macOS.
 export GEM_HOME := vendor
-export PATH := $(GEM_HOME)/bin:/usr/local/opt/ruby/bin:$(PATH)
+export PATH := $(GEM_HOME)/bin:$(HOME)/.rbenv/shims:/usr/local/opt/ruby/bin:$(PATH)
 # HACK: Make has a fast path and a slow path for command execution,
 # but the fast path uses the PATH variable from when make was started,
 # not the one we set on the previous line. In order for the above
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -62,7 +62,10 @@ test:
 
 vendor:
 	gem install bundler
+	bundle install
 
 bootstrap: Gemfile Gemfile.lock | vendor
-	bundle install
 	touch $@
+
+clean:
```
jseldess, line 70: This is nice. When do we think we'll want to run `make clean`? Another thing people should do now and again is `rm -rf _site`. If you don't you can see some weirdness in generated files (css, etc.). Should that be a part of `clean` as well?

ianjevans: Good idea on removing `_site`. I'll push a new commit.

I think you'd run `make clean` whenever you need to reset your Jekyll environment. That could be because we updated the dependencies in `Gemfile`, Homebrew/rbenv updated the Ruby version, or because you're seeing an error when trying to build locally.

jseldess (approved):
LGTM, with a nit

ericharmeling (approved):
LGTM

ianjevans:
@jseldess Sorry, I am not clear on what the nit is. Just checking before a I squash/merge.

Makefile
```diff
@@ -20,7 +20,7 @@
 # Note that we also amend the path to include /usr/local/opt/ruby/bin, which
```
jseldess, line 20: Sorry. My comment wasn't left earlier for some reason. Just wonder if we should update this comment to mention the rbenv change.

PR #9910, opened 2021-03-02
rename demo empty flag

tharun208:
Rename `--empty` flag in `demo` command with `--empty-database`

Resolves https://github.com/cockroachdb/cockroach/issues/47757


knz:
No, let's not do this.

Issue #9808, opened 2021-02-16
Update docs to mention our usage of TLS 1.3


piyush-singh:
Piyush Singh commented:

A customer reached out asking about when we will use TLS 1.3 because our [page on encryption](https://www.cockroachlabs.com/docs/stable/encryption.html) incorrectly states that we use TLS 1.2. We should update this to say that:
- We accept TLS 1.2 and 1.3 for client connections
- Inter-node communications will always use TLS 1.3

[Slack thread](https://cockroachlabs.slack.com/archives/CGA9F858R/p1613503866096200) for additional details.

Jira Issue: DOC-985

piyush-singh:
A quick search has turned up several additional areas where we call out TLS 1.2:
- https://www.cockroachlabs.com/docs/v20.2/authentication.html
- https://www.cockroachlabs.com/docs/cockroachcloud/authentication.html
- https://www.cockroachlabs.com/docs/cockroachcloud/security-overview.html
- https://www.cockroachlabs.com/docs/v20.2/security-overview.html
- https://www.cockroachlabs.com/docs/cockroachcloud/free-faqs.html
- https://www.cockroachlabs.com/docs/cockroachcloud/frequently-asked-questions.html

Issue #7968, opened 2020-08-11
RBAC Docs Updates


Amruta-Ranade:
Based on the following docs/issues, update RBAC docs across the docset:
- https://docs.google.com/document/d/17eY5UJnExEQ-uPnA5EyGH14i3h3szzHIzxbOfdvmRCc/edit#heading=h.gs4zsc68k3h
- https://github.com/cockroachdb/cockroach/pull/51856
- https://github.com/cockroachdb/docs/issues/7063

Issue #7063, opened 2020-04-02
Establish equivalency between the keywords `ROLE` and `USER`


Amruta-Ranade:
Amruta Ranade commented:

This involves checking SQL diagrams, examples, and overall content for `CREATE ROLE`/`CREATE USER`, `SHOW ROLES`/`SHOW USERS`, `DROP ROLE`/`DROP USER`, `ALTER ROLE/ALTER USER`

Jira Issue: DOC-483

Issue #7022, opened 2020-03-30
Adjust SQL diagram to show `ROLE` as an alias for `USER`.


Amruta-Ranade:
Reference: https://github.com/cockroachdb/docs/pull/7003

PR #9913, opened 2021-03-02
Align, reorganize, update, and reframe CREATE USER, CREATE ROLE, and related content on users/roles

mikeCRL:
Addresses:

#7968 | RBAC Docs Updates
#7063 | Establish equivalency between the keywords `ROLE` and `USER`
#9808 | Update docs to mention our usage of TLS 1.3

- Clarifies across create-role.md, create-user.md, authorization.md that CREATE ROLE/USER are equivalent, with the one exception of the NOLOGIN option. 
- Points readers to thinking about users and roles as "roles"—specifically, users as a special type of role: one with login privileges.
- Leaves open the option to use 'USER' commands like `CREATE USER` but with the awareness that this is shorthand for `CREATE ROLE` + `LOGIN` option.
- Rewrites bulk of summaries and intros on CREATE ROLE and CREATE USER pages. 
- Resolves inconsistencies across CREATE ROLE and CREATE USER pages.
- Converts often-repeated considerations like naming guidelines into a reusable include.
- Maintains separate copies of user and role examples, in part because the problem to solve that ROLE examples all have responses with NOLOGIN complicates a resolution, and we're leaving the door open to using CREATE USER.
- Various other changes in response to `@knz` review (see comments).

For reference, see the PostgresQL framing of roles/users at:

- https://www.postgresql.org/docs/8.1/user-manag.html
    > A role can be thought of as either a database user, or a group of database users, depending on how the role is set up.
- https://www.postgresql.org/docs/8.1/sql-createuser.html
    > CREATE USER is now an alias for CREATE ROLE, which see for more information. The only difference is that when the command is spelled CREATE USER, LOGIN is assumed by default, whereas NOLOGIN is assumed when the command is spelled CREATE ROLE.

In review: Confirm with @piyush-singh that the role vs. user framing works for us, and what knowledge we have about how clients may use CREATE ROLE / CREATE USER based on their existing SQL knowledge and any tooling. (Will some opt to use both commands, depending on context, and others want to standardize on CREATE ROLE; i.e. is it therefore OK to leave our docs open to either approach without being opinionated, as we do here?)

* Edit: To summarize Piyush's response to the above, generally yes, but can be opinionated in some ways, e.g. not needing to refer to "role/user" everywhere; just use "role".

Follow-up TODOs in subsequent issues/PRs:

1. Align diagrams. Add note to Alter user/role about equivalency.
    - #7968 | Update CREATE ROLE, CREATE USER, ALTER USER SQL diagrams
    - #7022 | Adjust SQL diagram to show `ROLE` as an alias for `USER`.
2. Create issue capturing all `@knz` feedback marked for follow-up in comments.
3. Document differences found in role naming vs. PostgreSQL https://github.com/cockroachdb/docs/issues/10010


mikeCRL:
@piyush-singh Can I ask you to take a first crack at reviewing this, primarily that we're aligned on the approach described in my [summary comment/question](https://github.com/cockroachdb/docs/pull/9913#issue-583166943). Thank you.

v20.2/authorization.md
```diff
@@ -5,40 +5,42 @@ toc: true
...
+There is no technical distinction between a SQL role or user in CockroachDB. A role/user can:
 
-- be permitted to log in to the [SQL shell](cockroach-sql.html).
+- be permitted to log in to the [SQL shell](cockroach-sql.html) or other applications that connect to a database.
 - be granted [privileges](#privileges) to specific actions and database objects.
 - be a member of other users/roles, inheriting their privileges.
 - have other users/roles as members that inherit its privileges.
 
-We refer to these as "roles" when they are created for managing the privileges of their member "users" and not for logging in directly, which is typically reserved for "users".
+We refer to these as "roles" when they are created for managing the privileges of their member "users". We often refer to roles that are enabled to direclty log in to a database using the [built-in SQL shell](cockroach-sql.html) or an application as "users".
```
mikeCRL, line 23: TODO: Fix typo in `direclty`

mikeCRL (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9913)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @piyush-singh and @taroface)

___
*[v20.2/authentication.md, line 71 at r6](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVMVPUX6vOJLXipWPVJ:-MVcE0jE4yspC-r_NEAB:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2bcbff0854038fb4bb10d227ceb9b602ececf14b/v20.2/authentication.md#L71)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

Is it worth mentioning that this requires the `CREATELOGIN` privilege or an admin account?
</blockquote></details>

Done.
___
*[v20.2/authorization.md, line 18 at r7](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVMsHlF5OAIzJP3S8iL:-MVcE4a6-YLf0ZTAG66K:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/c88a9001737ae5eb619e59f85f20ace87a2d5510/v20.2/authorization.md#L18)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

This phrasing is a bit awkward, and I think some meaning may be getting lost here. Perhaps "can be used by applications that connect to the database"?
</blockquote></details>

Done.
___
*[v20.2/authorization.md, line 37 at r7](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVMsKn62wxwGpl0OPkI:-MVcE7wx9HItRt98OMNv:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/c88a9001737ae5eb619e59f85f20ace87a2d5510/v20.2/authorization.md#L37)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

We should probably either explain what `NOLOGIN` does here or link to the docs where this is explained to avoid confusion.
</blockquote></details>

Done.
___
*[v20.2/create-role.md, line 11 at r6](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVMqKzy3LKiXMzE7MdA:-MVcE8iQ-_UEftzT6td-:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/2bcbff0854038fb4bb10d227ceb9b602ececf14b/v20.2/create-role.md#L11)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

>by its members, by their members

nit: I found this wording slightly confusing and initially thought this was a typo with the same words repeated. Took a reread or two to understand the intent here. I can't think of any elegant way to phrase this to clarify though.
</blockquote></details>

Done.

mikeCRL:
@piyush-singh Thanks for the great edits and sorry for my delay in following up. I've implemented changes based on all your comments, and they inspired some further, similar minor edits beyond that.

(Not sure whether to select `Dismiss review` or `Re-request review`.)

jseldess:
Hey @mikeCRL and @piyush-singh, any updates on this. Are we close to being able to merge?

mikeCRL:
> Hey @mikeCRL and @piyush-singh, any updates on this. Are we close to being able to merge?

@jseldess Discussed with Piyush earlier this morning, who confirmed that going forward I should 're-submit for review'. I've done that and it's on his queue to confirm my edits. Also assigned to @taroface for review. I believe it is in a mergeable state, but doubtful these sign-offs can occur before the internal doc release notes.

piyush-singh (approved):
> Reviewed 1 of 4 files at r8, 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9913)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mikeCRL and @taroface)

___
*[v20.2/create-role.md, line 30 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVw4cxr7ocV7_-BZ4Qh:-MVw4cxr7ocV7_-BZ4Qi:b-kl8g4c) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L30)):*
> ```Markdown
> 
> - To create other roles, a role must have the [`CREATEROLE`](#create-a-role-that-can-create-other-roles-and-manage-authentication-methods-for-the-new-roles) parameter set.
> - To add the `LOGIN` capability for other roles so that they may log in as users, a role must also have the [`CREATELOGIN`](#create-a-role-that-can-create-other-roles-and-manage-authentication-methods-for-the-new-roles) parameter set. 
> ```

good catch!

taroface (changes requested):
A few suggestions!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9913#-:-MVwQKHTCVzXe0BTPPFH:b7vh466)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mikeCRL)

___
*[_includes/v20.2/sql/role-considerations.md, line 3 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvq7dq0Bds-s9QwCg8:-MVvq7drCsZIFOVBRPP4:bbj232s) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/_includes/v20.2/sql/role-considerations.md#L3)):*
> ```Markdown
> - Role names (also known as usernames):
>     - Are case-insensitive.
>     - Can contain only letters, numbers, underscores, and periods. When a period is present, quotes around the name are required in SQL commands.
> ```

I think that it would be clearest to cite the actual characters after a few of these:

underscores (`_`)
quotes (`"`) <-- single, double, or either?
___
*[_includes/v20.2/sql/role-considerations.md, line 6 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvqo8U3rXoVnS9NteB:-MVvqo8U3rXoVnS9NteC:b-7xu39b) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/_includes/v20.2/sql/role-considerations.md#L6)):*
> ```Markdown
>     - Must start with a letter or underscore.
>     - Can be 1 to 63 characters long.
>     - Must be unique. Note that roles and users share the same namespace.
> ```

Having trouble understanding the second sentence. Does it imply that intended "roles" and "users" can't have an identical name (since they're identical? but that sounds like a semantic issue, separate from sharing namespace). Perhaps it would help to spell out the implications here?
___
*[v20.2/authentication.md, line 7 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvuRRM9Dt0gfbzcl1H:-MVvuRRM9Dt0gfbzcl1I:bflgr8a) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authentication.md#L7)):*
> ```Markdown
> ---
> 
> Authentication refers to the act of verifying the identity of the other party in communication. CockroachDB requires TLS 1.3 digital certificates for inter-node authentication and accepts TLS 1.2 and TLS 1.3 certificates for client-node authentication. A Certificate Authority (CA) is required, as well as keys and certificates for nodes, clients, and, optionally, the DB Console. This document discusses how CockroachDB uses digital certificates and gives a [conceptual overview](#background-on-public-key-cryptography-and-digital-certificates) of public key cryptography and digital certificates.
> ```

This isn't part of your edits, but I don't think we capitalize Certificate Authority elsewhere.

suggested link DB Console, if applicable: https://www.cockroachlabs.com/docs/stable/ui-overview.html#db-console-security
___
*[v20.2/authentication.md, line 249 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwFz0REKUflvQskZK7:-MVwFz0REKUflvQskZK8:benpsw9) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authentication.md#L249)):*
> ```Markdown
> As mentioned above, CockroachDB supports the [TLS 1.3 and TLS 1.2](https://en.wikipedia.org/wiki/Transport_Layer_Security) security protocols, which take advantage of both symmetric (to encrypt data in flight) as well as asymmetric encryption (to establish a secure channel as well as **authenticate** the communicating parties).
> 
> Authentication refers to the act of verifying the identity of the other party in communication. CockroachDB uses TLS 1.3 digital certificates for inter-node authentication, and your choice of TLS 1.2 and TLS 1.3 certificates for client-node authentication. These authentication methods require a Certificate Authority (CA) as well as keys and certificates for nodes, clients, and, optionally, the DB Console.
> ```

Same comment as above.
___
*[v20.2/authorization.md, line 22 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwG8C_2sk-iJ-GhQI3:-MVwG8C_2sk-iJ-GhQI4:brcljbj) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L22)):*
> ```Markdown
> - be granted [privileges](#privileges) to specific actions and database objects.
> - be a member of other users/roles, inheriting their privileges.
> - have other users/roles as members that inherit its privileges.
> ```

Since we discussed this recently, should the items in this list be capitalized and not end in periods? Or is the above usage actually consistent with our overall style (I can't tell)?
___
*[v20.2/authorization.md, line 38 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwN_7k67w7WSmNtyLh:-MVwN_7k67w7WSmNtyLi:b598r8g) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L38)):*
> ```Markdown
> You can use the [`CREATE USER`](create-user.html) and [`DROP USER`](drop-user.html) statements to create and remove users, the [`ALTER USER`](alter-user.html) statement to add or change a user's password and role options, and the [`SHOW USERS`](show-users.html) statement to list users.
> 
> The statements `CREATE ROLE`, `DROP ROLE`, `ALTER ROLE`, AND `SHOW ROLES` are equivalent to these, respectively, with the exception of the default `NOLOGIN` setting added with `CREATE ROLE`.
> ```

Suggest linking these SQL commands to their respective docs.
___
*[v20.2/authorization.md, line 40 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwNkoMBFkQPluHBRfR:-MVwNkoMBFkQPluHBRfS:bcsfrnh) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L40)):*
> ```Markdown
> The statements `CREATE ROLE`, `DROP ROLE`, `ALTER ROLE`, AND `SHOW ROLES` are equivalent to these, respectively, with the exception of the default `NOLOGIN` setting added with `CREATE ROLE`.
> 
> Use the [`GRANT <privileges>`](grant.html) and [`REVOKE <privileges>`](revoke.html) statements to manage the user’s privileges. 
> ```

Low-priority, but we discussed placeholder characters in #docs-sync and I think the outcome was to use `{ }` where applicable. Not sure how official that is!
___
*[v20.2/create-role.md, line 8 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPuizBBEj4kRuRqD-:-MVwPuizBBEj4kRuRqD0:b4tf83e) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L8)):*
> ```
> There is no distinct "user" entity in CockroachDB, however, a role with
> ```

Suggest splitting this into two sentences for readability:

> ```
> There is no distinct "user" entity in CockroachDB. However, a role with
> ```
___
*[v20.2/create-role.md, line 16 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPE_oErTYp4awXthh:-MVwPE_oErTYp4awXthi:bwrbrzl) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L16)):*
> ```Markdown
>  `CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the `NOLOGIN` option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the `LOGIN` option to achieve the same result as `CREATE USER`.
> 
>  Prior to CockroachDB v20.1, Role-Based Access Control (RBAC) was an enterprise feature, and `CREATE ROLE` created an entity type called a role that was distinct from a user. As of v20.1, an updated version of this feature is freely available in CockroachDB core, and for enhanced Postgres compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements to refer to this new, singular entity type.
> ```

We call it "role-based access management" (lowercase) elsewhere, so not sure if it's worth being consistent. It is confusing since it clearly does not match the RBAC acronym...
___
*[v20.2/create-role.md, line 27 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPZaaBZuumKMaFnzz:-MVwPZabCg7zOzlFV0tl:b-eutu9p) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L27)):*
> ```Markdown
> ## Required privileges
> 
> Unless a role is a member of the admin role, additional [priviliges](#parameters) are required to manage other roles.
> ```

admin role -> `admin` role

Typo on "privileges"
___
*[v20.2/create-user.md, line 9 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQ3fN4jAwgqwkVYWB:-MVwQ3fN4jAwgqwkVYWC:benpsw9) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L9)):*
> ```Markdown
> The `CREATE USER` [statement](sql-statements.html) creates a SQL role that can be used to log in to a database.
> 
>  There is no distinct "user" entity in CockroachDB, however, a role with the [`LOGIN`](#parameters) option enabled can log in to the SQL shell and is often called a user.
> ```

Same comment as above.
___
*[v20.2/create-user.md, line 16 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQ9MsEicPwy5AD4G8:-MVwQ9MsEicPwy5AD4G9:benpsw9) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L16)):*
> ```Markdown
>  `CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the [`NOLOGIN`](#parameters) option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the `LOGIN` option to achieve the same result as `CREATE USER`.
> 
>  Prior to CockroachDB v20.1, Role-Based Access Control (RBAC) was an enterprise feature, and `CREATE ROLE` created an entity type called a role that was distinct from a user. As of v20.1, an updated version of this feature is freely available in CockroachDB core, and for enhanced Postgres compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements to refer to this new, singular entity type.
> ```

Same comment as above.
___
*[v20.2/create-user.md, line 27 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQCi08I5XRFKnaTau:-MVwQCi08I5XRFKnaTav:benpsw9) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L27)):*
> ```Markdown
> ## Required privileges
> 
> Unless a role is a member of the admin role, additional [priviliges](#parameters) are required to manage other roles.
> ```

Same comment as above.

mikeCRL (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9913#-:-MVwomBY1XJ8nEmNUY7j:b-814g8k)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @piyush-singh and @taroface)

___
*[_includes/v20.2/sql/role-considerations.md, line 3 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvq7dq0Bds-s9QwCg8:-MVwhfJwB9V4rHaU1gb1:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/_includes/v20.2/sql/role-considerations.md#L3)):*

I think that it would be clearest to cite the actual characters after a few of these:

underscores (`_`)
quotes (`"`) <-- single, double, or either?
</blockquote></details>

Done.
___
*[_includes/v20.2/sql/role-considerations.md, line 6 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvqo8U3rXoVnS9NteB:-MVwhhQj9DW3esP-ewJ5:b-ly37xz) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/_includes/v20.2/sql/role-considerations.md#L6)):*

Having trouble understanding the second sentence. Does it imply that intended "roles" and "users" can't have an identical name (since they're identical? but that sounds like a semantic issue, separate from sharing namespace). Perhaps it would help to spell out the implications here?
</blockquote></details>

Done. Dropped the mention of roles and users as separate entities here; I think it's confusing and a vestige of the prior state where these differed.
___
*[v20.2/authentication.md, line 7 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVvuRRM9Dt0gfbzcl1H:-MVwiNt_6xM4u2JZxn5h:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authentication.md#L7)):*

This isn't part of your edits, but I don't think we capitalize Certificate Authority elsewhere.

suggested link DB Console, if applicable: https://www.cockroachlabs.com/docs/stable/ui-overview.html#db-console-security
</blockquote></details>

Done.
___
*[v20.2/authentication.md, line 249 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwFz0REKUflvQskZK7:-MVwiwlYD6gVetW-odnh:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authentication.md#L249)):*

Same comment as above.
</blockquote></details>

Done.
___
*[v20.2/authorization.md, line 22 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwG8C_2sk-iJ-GhQI3:-MVwj63O9ozzrleHoUAD:b-vmi87m) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L22)):*

Since we discussed this recently, should the items in this list be capitalized and not end in periods? Or is the above usage actually consistent with our overall style (I can't tell)?
</blockquote></details>

I think keeping the periods here is consistent with our style because each bullet completes the sentence that is introduced before the list.
___
*[v20.2/authorization.md, line 38 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwN_7k67w7WSmNtyLh:-MVwjhR-66oyW7BHVLN8:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L38)):*

Suggest linking these SQL commands to their respective docs.
</blockquote></details>

Done.
___
*[v20.2/authorization.md, line 40 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwNkoMBFkQPluHBRfR:-MVwmNd-6yqYix7jElmj:btuodpt) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/authorization.md#L40)):*

Low-priority, but we discussed placeholder characters in #docs-sync and I think the outcome was to use `{ }` where applicable. Not sure how official that is!
</blockquote></details>

Good catch. I'd like to keep this as-is for now because there are 27 instances across 20.2 using `GRANT <privileges>` or `GRANT <roles>` and just one starting with `GRANT {`. I will bring this up on Slack again.
___
*[v20.2/create-role.md, line 8 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPuizBBEj4kRuRqD-:-MVwmlmo2CH-sWuAOitC:bults0) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L8)):*

> ```
> There is no distinct "user" entity in CockroachDB, however, a role with
> ```

Suggest splitting this into two sentences for readability:

> ```
> There is no distinct "user" entity in CockroachDB. However, a role with
> ```

</blockquote></details>

I agree, and I've dropped the "however", too, given that the idea of a user is still essential and the word is unnecessary here.
___
*[v20.2/create-role.md, line 16 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPE_oErTYp4awXthh:-MVwnnOn5KP0Yj9PRLid:b-vu8jtq) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L16)):*

We call it "role-based access management" (lowercase) elsewhere, so not sure if it's worth being consistent. It is confusing since it clearly does not match the RBAC acronym...
</blockquote></details>

Changing to role-based access control, since we seem to arbitrarily swap between them in the docs.
___
*[v20.2/create-role.md, line 27 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwPZaaBZuumKMaFnzz:-MVwoA5_Ddyb9WlLmklv:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L27)):*

admin role -> `admin` role

Typo on "privileges"
</blockquote></details>

Done.
___
*[v20.2/create-role.md, line 30 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVw4cxr7ocV7_-BZ4Qh:-MVwoI9z7kBbq1lluEFz:b-ucfknh) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-role.md#L30)):*
<details><summary><i>Previously, piyush-singh (Piyush Singh) wrote…</i></summary><blockquote>

good catch!
</blockquote></details>

Thanks!
___
*[v20.2/create-user.md, line 9 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQ3fN4jAwgqwkVYWB:-MVwoNPv-A-I1D3ZToT0:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L9)):*

Same comment as above.
</blockquote></details>

Done.
___
*[v20.2/create-user.md, line 16 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQ9MsEicPwy5AD4G8:-MVwoP-t5YUgKjUBvKg-:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L16)):*

Same comment as above.
</blockquote></details>

Done.
___
*[v20.2/create-user.md, line 27 at r9](https://reviewable.io/reviews/cockroachdb/docs/9913#-MVwQCi08I5XRFKnaTau:-MVwoij-4-GbSKxEWhPQ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b5f9d1ddb8652656fc1033c60ba39de89c47088d/v20.2/create-user.md#L27)):*

Same comment as above.
</blockquote></details>

Done.

taroface (approved):
LGTM!!

v20.2/authorization.md
```diff
@@ -5,40 +5,43 @@ toc: true
 redirect_from: [create-and-manage-users.html, roles.html]
 ---
 
-User authorization is the act of defining access policies for authenticated CockroachDB users. CockroachDB allows you to create, manage, and remove your cluster's SQL [users](#sql-users) and assign SQL-level [privileges](#assign-privileges) to the users. Additionally, you can use [role-based access management (RBAC)](#roles) for simplified user management.
+User authorization is the act of defining access policies for authenticated CockroachDB users. CockroachDB allows you to create, manage, and remove your cluster's SQL [users](#users-and-roles) and assign SQL-level [privileges](#assign-privileges) to them. Additionally, you can use [role-based access control (RBAC)](#roles) for simplified user management.
```
knz, line 8: This paragraph is very confused and probably inaccurate

- creating and removing users and roles are related to authenticatuon, not authorization
- the term "access policy" is related to both authentication and authorization, so it's confusing to see it here

What *is* related to authorization is the mapping of roles to SQL privileges, and the use of role membership to group users that share privileges. However, that idea is not even mentioned here! It should really be the first thing mentioned on this page.

The term "RBAC" was originally introduced inside documentation because of a product-level fad that the term would help drive sell—i.e. it is *not* a technical thing and does *not* have explanatory power. (The fact that no other database uses "RBAC" to structure its reference docs on authorization should be a strong hint.) We made a *mistake* to use it in the *reference* documentation - it should have been segregated to the "major release notes" and perhaps the main feature/comparison page. I think it's time to correct this mistake.

v20.2/authorization.md
```diff
@@ -5,40 +5,43 @@ toc: true
...
 
 {{site.data.alerts.callout_info}}
- Role-based access management (RBAC) is no longer an enterprise feature and is now freely available in the core version of CockroachDB. Also, for enhanced PostgreSQL compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements. Note that even though the keywords are now interchangeable, it is still helpful to understand the distinction between the concepts (a "user" refers to an individual database user and a "role" refers to a group of database users).
+ Role-based access control (RBAC) is no longer an enterprise feature and is now freely available in the core version of CockroachDB. Also, for enhanced PostgreSQL compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements as the two represent the same type of entity. A role can be used as a group of database roles/users or as an individual user. For more information, see [Users and roles](#users-and-roles).
```
knz, line 11: I think you can drop the callout - this change happened in 19.2 or 20.1

v20.2/authorization.md
```diff
@@ -5,40 +5,43 @@ toc: true
...
 {{site.data.alerts.end}}
 
 ## Users and roles
```
knz, line 14: This section and the next section belong to a separate page "User and role management" which is a separate concern from authentication and authorization.

The "authorization" page should be limited to the actual authorization matters:
- which SQL privileges exist
- how they can be granted and revoked
- how to group users in roles in practice to achieve various use cases

v20.2/create-role.md
```diff
@@ -1,45 +1,52 @@
...
 
 {{site.data.alerts.callout_info}}
- <code>CREATE ROLE</code> is no longer an enterprise feature and is now freely available in the core version of CockroachDB. Also, since the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements for enhanced Postgres compatibility, `CREATE ROLE` is now an alias for [`CREATE USER`](create-user.html).
+ `CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the [`NOLOGIN`](#parameters) option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the [`LOGIN`](#parameters) option to achieve the same result as `CREATE USER`.
+
+ Prior to CockroachDB v20.1, role-based access control (RBAC) was an enterprise feature, and `CREATE ROLE` created an entity type called a role that was distinct from a user. As of v20.1, an updated version of this feature is freely available in CockroachDB core, and for enhanced Postgres compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements to refer to this new, singular entity type.
```
knz, line 16: You can drop the second paragraph since in 20.2 there was no change here.

v20.2/create-role.md
```diff
@@ -1,45 +1,52 @@
...
 {{site.data.alerts.end}}
 
+For more information, see the [Authorization](authorization.html) documentation on roles and users.
```
knz, line 19: I would use 2 separate hyperlinks here: authorization for privilege management and role membership, and user/role management for creating/dropping roles

v20.2/create-role.md
```diff
@@ -1,45 +1,52 @@
...
+table td:first-child {
+    min-width: 225px;
+}
+</style>
+
 | Parameter | Description |
 ------------|--------------
-`name` | The name of the role you want to create. Role names are case-insensitive; must start with either a letter or underscore; must contain only letters, numbers, periods, or underscores; and must be between 1 and 63 characters.<br><br>Note that roles and [users](create-user.html) share the same namespace and must be unique.
+`name` | The name of the role you want to create. See the [Considerations](#considerations) section for important naming guidelines.
 `CREATELOGIN`/`NOCREATELOGIN` | Allow or disallow the role to manage authentication using the `WITH PASSWORD`, `VALID UNTIL`, and `LOGIN/NOLOGIN` parameters. <br><br>By default, the parameter is set to `NOCREATELOGIN` for all non-admin roles.
```
knz, line 48: The list of possible role options should be part of the authorization page, and referred to here. I think it's confusing to explain this in the "create role" page when really it's an authorization matter and users can inherit these options via role membership.

v20.2/create-user.md
```diff
@@ -1,30 +1,34 @@
...
+ 
+ You can [assign privileges](authorization.html#assign-privileges) to the user and [set it as a member](grant-roles.html) of other roles, inheriting their privileges.
 
 {{site.data.alerts.callout_info}}
- Since the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements for enhanced Postgres compatibility, `CREATE USER` is now an alias for [`CREATE ROLE`](create-role.html).
+ `CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the [`NOLOGIN`](#parameters) option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the [`LOGIN`](#parameters) option to achieve the same result as `CREATE USER`.
+
+ Prior to CockroachDB v20.1, role-based access control (RBAC) was an enterprise feature, and `CREATE ROLE` created an entity type called a role that was distinct from a user. As of v20.1, an updated version of this feature is freely available in CockroachDB core, and for enhanced Postgres compatibility, the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements to refer to this new, singular entity type.
```
knz, line 16: ditto you can remove this paragraph

v20.2/create-user.md
```diff
@@ -1,30 +1,34 @@
...
 {{site.data.alerts.end}}
 
+For more information, see the [Authorization](authorization.html) documentation on roles and users.
```
knz, line 19: ditto hyperlinks

v20.2/create-user.md
```diff
@@ -39,7 +43,7 @@ table td:first-child {
 
  Parameter | Description
 -----------|-------------
-`user_name` | The name of the user you want to create.<br><br>Usernames are case-insensitive; must start with a letter, number, or underscore; must contain only letters, numbers, or underscores; and must be between 1 and 63 characters.
+`user_name` | The name of the user you want to create. See the [Considerations](#considerations) section for important naming guidelines.
 `CREATELOGIN`/`NOCREATELOGIN` | Allow or disallow the user to manage authentication using the `WITH PASSWORD`, `VALID UNTIL`, and `LOGIN/NOLOGIN` parameters. <br><br>By default, the parameter is set to `NOCREATELOGIN` for all non-admin users.
```
knz, line 47: ditto reference to role options

v20.2/schema-design-overview.md
```diff
@@ -83,7 +83,7 @@ For guidance on using temporary objects, see [Temporary Tables](temporary-tables
 
 ## Controlling access to objects
 
-CockroachDB supports both user-based and role-based access control. With roles, or with direct assignment, you can grant a [SQL user](authorization.html#sql-users) the [privileges](authorization.html#privileges) required to view, modify, and delete database schema objects.
+CockroachDB supports both user-based and role-based access control. With roles, or with direct assignment, you can grant a [SQL user](authorization.html#users) the [privileges](authorization.html#privileges) required to view, modify, and delete database schema objects.
```
knz, line 86: The hyperlinks should point to separate pages for user/role management and authorization.

v20.2/show-grants.md
```diff
@@ -7,8 +7,8 @@ toc: true
 
 The `SHOW GRANTS` [statement](sql-statements.html) lists one of the following:
 
-- The [roles](authorization.html#sql-users) granted to [users](authorization.html#sql-users) in a cluster.
-- The [privileges](authorization.html#assign-privileges) [granted](grant.html) to [users](authorization.html#sql-users) on [databases](create-database.html), [schemas](create-schema.html), [tables](create-table.html), or [user-defined types](enum.html).
+- The [roles](authorization.html#roles) granted to [users](authorization.html#users) in a cluster.
```
knz, line 10: ditto hyperlinks

v20.2/show-grants.md
```diff
@@ -33,7 +33,7 @@ SHOW GRANTS ON ROLE [<roles...>] [FOR <users...>]
 Parameter    | Description
 -------------|-----------------------------------------------------------------------------------------------------
 `targets`    | A comma-separated list of database, schema, table, or user-defined type names.<br><br>{{site.data.alerts.callout_info}}To list the privilege grants for all tables in the current database, you can use `SHOW GRANTS ON TABLE *`.{{site.data.alerts.end}}
-`users`      | A comma-separated list of the [users](authorization.html#sql-users) whose privileges or roles you want to show.
+`users`      | A comma-separated list of the [users](authorization.html#users) whose privileges or roles you want to show.
```
knz, line 36: ditto hyperlink

knz (changes requested):
Good step forward and I am glad this is being tackled at last.

I think the biggest comment here is that the current structure of the doc around authorization mistakenly conflate managing the set of users and roles, which is primarily an authentication matter, and managing privileges and role membership, which is the true authorization matter.

They should be properly separated and the hyperlinks fixed accordingly.

Additionally the page on user/role management (and authentication) should introduce and define the term "authentication principal", which is used in Enterprise deployments and which we will increasingly use in later versions of CockroachDB.

_includes/v20.2/sql/role-considerations.md
```diff
@@ -0,0 +1,12 @@
+- Role names (also known as usernames):
+    - Are case-insensitive.
+    - Can contain only letters, numbers, underscores (`_`), and periods. When a period is present, single or double quotes around the name are required in SQL commands.
```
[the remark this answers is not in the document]
mikeCRL: @knz How do you define "Valid SQL identifier"? Valid per the SQL Standard?

_includes/v20.2/sql/role-considerations.md
```diff
@@ -0,0 +1,12 @@
...
+    - Must start with a letter or underscore.
```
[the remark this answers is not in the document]
mikeCRL: I had only tested this without quotes and it failed, so I thought this was accurate.

I now see it works with quotes. I will incorporate that detail into the broader point about valid names and quotes.

_includes/v20.2/sql/role-considerations.md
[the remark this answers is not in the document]
knz: per the postgres lexical rules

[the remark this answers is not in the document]
mikeCRL: @knz OK thanks.

## Naming details edit

I've reworked this in light of our comments, further testing, and to match the sensible [Postgres docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html) paradigm of splitting into what's allowed _without quotes_ vs. _with quotes_. I include 'with quotes' first, because we recommend always using quotes, and this makes it easier to see the short list of character usage that's lost without quotes. How's this?

(Note: I'm temporarily removing the 63-character limit mention; I have a follow-up question about that.)

> - Role names (also known as usernames) are case-insensitive and must be unique.
> - When surrounded by quotes in SQL statements (always recommended)—single or double quotes, depending on the statement—role names:
>     - Can contain letters, underscores, digits, periods, and dashes. Letters may include `a`-`z`, those with diacritical marks, and non-Latin letters*.
>     - Can begin with a letter, underscore, or digit.
> - When referenced without quotes, role names:
>     - Cannot contain periods or dashes.
>     - Cannot begin with a digit.
>     - Cannot match the name of a SQL keyword.

\* I'm provisionally adding this for your review, in case it's accurate, because it is specified in the Postgres docs, though I'm actually not sure which characters it's referring to.

## Length - characters vs bytes

Upon further testing, I think we are limiting name length by bytes and not characters, and that some characters are using two bytes each.

Testing a username that comprises 35 instances of `å` (with or without quotes) gets me this:

```
ERROR: username "ååååååååååååååååååååååååååååååååååå" is too long
```

I am wondering if this is a bug, and we should be able to store `å` using a single byte, or if it's expected. And regardless, whether we should add a caveat to the docs.

Side note: the hint in the SQL shell displayed when the name is too long does match your overall summary of requirements of characters that are allowed with quotes: "Usernames are case insensitive, must start with a letter, digit or underscore, may contain letters, digits, dashes, periods, or underscores, and must not exceed 63 characters." 

* However
   * the "63 characters" point is now in question. 
   * that this can be displayed when not using quotes (e.g. via `CREATE USER ååååååååååååååååååååååååååååååååååå`') and yet refers to character usage that won't work without quotes, may be confusing for users.

## Note on unicode

Just FYI, we appear to allow unicode characters via statements like `CREATE ROLE "\U+AB3E;"`.

That format isn't documented in Postgres, and I'm not sure how you'd string together multiple characters. We do not appear to support the similar Postgres format [described here](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#identifiers-and-key-words#:~:text=A%20variant%20of%20quoted%20identifiers%20allows%20including,Unicode%20characters%20identified%20by%20their%20code%20points.).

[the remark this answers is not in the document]
knz: > How's this?

probably fine. would like to see it in context though.

> I am wondering if this is a bug

Short answer: the UX is inconsistent and that's the bug. We can either choose 63 bytes or 63 characters (both would be possible in the code), but we need to ensure that the implementation matches the description/doc.

Which one *should* it be? I don't know. I believe we need a limit of 63 bytes for pure ASCII names, but if the name is pure ASCII then 63 bytes is also == 63 characters. Maybe start by documenting the 63 byte limit, and we can discuss later about extending the limit in the code to 63 characters.

> we appear to allow unicode characters via statements like CREATE ROLE "\U+AB3E;".

that's a happy accident, however probably not one we want to document yet, in light of this:

> That format isn't documented in Postgres, and I'm not sure how you'd string together multiple characters. We do not appear to support the similar Postgres format described here.

I would say we should aim to do what postgres accepts here. If we're not there yet, we need to file a followup issue in the crdb repo, and document the mismatch as a known limitation.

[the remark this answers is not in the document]
mikeCRL: Documenting the current state, plus any differences with Postgres, sounds good to me. Created an issue https://github.com/cockroachdb/docs/issues/10010 and included one other difference that I noticed:

> Difference in how quoted names are treated: always case insensitive in CRDB vs. case sensitive _when created using quotes_ in PostgreSQL.

v20.2/authentication.md
```diff
@@ -4,15 +4,15 @@ summary: Learn about the authentication features for secure CockroachDB clusters
 toc: true
 ---
 
-Authentication refers to the act of verifying the identity of the other party in communication. CockroachDB requires TLS 1.2 digital certificates for inter-node and client-node authentication, which require a Certificate Authority (CA) as well as keys and certificates for nodes, clients, and (optionally) the DB Console. This document discusses how CockroachDB uses digital certificates and also gives [conceptual overview](#background-on-public-key-cryptography-and-digital-certificates) of public key cryptography and digital certificates.
+Authentication refers to the act of verifying the identity of the other party in communication. CockroachDB requires TLS 1.3 digital certificates for inter-node authentication and accepts TLS 1.2 and TLS 1.3 certificates for client-node authentication. A certificate authority (CA) is required, as well as keys and certificates for nodes, clients, and, optionally, the DB Console. This document discusses how CockroachDB uses digital certificates and gives a [conceptual overview](#background-on-public-key-cryptography-and-digital-certificates) of public key cryptography and digital certificates.
```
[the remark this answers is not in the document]
mikeCRL: In my latest commit, I have dropped that line and replaced it with a different summary of what the doc contains, per your suggestion.

I will log follow-up issue to add the clarity that different CAs are used for different authentication channels, the limitations and plans for that, etc.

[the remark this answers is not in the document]
mikeCRL: https://github.com/cockroachdb/docs/issues/9808, which links back to  https://cockroachlabs.slack.com/archives/CGA9F858R/p1613504688098200?thread_ts=1613503866.096200&cid=CGA9F858R. Earlier in the thread, it's agreed that this has been the case since v20.2.

v20.2/authentication.md
```diff
@@ -44,49 +44,51 @@ CockroachDB offers the following methods for client authentication:
...
-   $ cockroach sql --certs-dir=certs --user=jpointsman
-   ~~~
+    Example:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach sql --certs-dir=certs --user=jpointsman
+    ~~~
 
-- **Password authentication**, which is available to users and roles who you've created passwords for. Password creation is supported only in secure clusters.
+- **Password authentication**, which is available to users and roles who you've created passwords for. Password creation is supported only in secure clusters. While users can use passwords to authenticate without supplying client certificates and keys, we recommend using certificate-based authentication whenever possible.
```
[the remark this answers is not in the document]
mikeCRL: Done. Note that I added the same link to `**Password authentication without TLS**`, as well.

v20.2/authentication.md
```diff
@@ -44,49 +44,51 @@ CockroachDB offers the following methods for client authentication:
...
 
-   Example:
-   {% include copy-clipboard.html %}
-   ~~~ shell
-   $ cockroach sql --certs-dir=certs --user=jpointsman
-   ~~~
+    Example:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach sql --certs-dir=certs --user=jpointsman
```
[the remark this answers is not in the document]
mikeCRL: Thanks. Saving for a follow-up issue.

v20.2/authentication.md
```diff
@@ -44,49 +44,51 @@ CockroachDB offers the following methods for client authentication:
...
```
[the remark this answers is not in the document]
mikeCRL: Thanks. Saving for a follow-up issue.

v20.2/authentication.md
```diff
@@ -44,49 +44,51 @@ CockroachDB offers the following methods for client authentication:
...
     # To exit, type: \q.
     #
     Enter password:
-  ~~~
+    ~~~
 
-   Note that the client still needs the CA certificate to validate the nodes' certificates.
+    Note that the client still needs the CA certificate to validate the nodes' certificates.
+
+    To create a user with a password, use the `WITH PASSWORD` clause of [`CREATE ROLE`](create-role.html) OR `CREATE USER`(create-user.html). To add a password to an existing user, use the [`ALTER ROLE`](alter-role.html) OR [`ALTER USER`](alter-user.html) statement. Note that the user performing these actions [must have the `CREATEROLE` and `CREATELOGIN` permissions](create-role.html#create-a-role-that-can-create-other-roles-and-manage-authentication-methods-for-the-new-roles) or be a member of the `admin` role.
```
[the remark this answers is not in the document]
mikeCRL: Done.

v20.2/authentication.md
```diff
@@ -44,49 +44,51 @@ CockroachDB offers the following methods for client authentication:
...
 
-   Example:
-   {% include copy-clipboard.html %}
-   ~~~ shell
-   $ cockroach sql --user=jpointsman --insecure
-   ~~~
+    Example:
+    {% include copy-clipboard.html %}
+    ~~~ shell
+    $ cockroach sql --user=jpointsman --insecure
```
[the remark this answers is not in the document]
mikeCRL: Removed the flag. Will call for considering adding connection URL(s) in the follow-up issue.

v20.2/authorization.md
mikeCRL: Thanks for the helpful context and analysis.

I've updated the intro to convey the key points you suggested around roles and privileges.

I will include the other reorg (moving creation of users/roles) in a follow-up issue.

v20.2/authorization.md
mikeCRL: Done.

v20.2/authorization.md
mikeCRL: Makes sense. Bringing this into a follow-up issue which will include creating that new page and any necessary reorg.

v20.2/create-role.md
mikeCRL: Done.

v20.2/create-role.md
mikeCRL: Updated the reference to the Authorization page; saving the second change for the follow-up issue.

v20.2/create-role.md
mikeCRL: I see your point. I am logging this in the follow-up issue.

v20.2/create-user.md
mikeCRL: Done.

v20.2/create-user.md
mikeCRL: Updated the reference to the Authorization page; saving the second for the follow-up issue.

v20.2/create-user.md
mikeCRL: Including this in the follow-up issue.

v20.2/schema-design-overview.md
mikeCRL: Including in the follow-up issue.

v20.2/show-grants.md
mikeCRL: Including in follow-up issue.

v20.2/show-grants.md
mikeCRL: Including in follow-up issue.

mikeCRL:
@knz Thanks for the review. I made various changes, as noted in my replies, in a follow up commit that I just pushed. Please review these changes.

As a next step, I think it would be best for me to log the rest of your feedback in a follow-up issue. (Sorry for the repeated replies to that effect, but I wanted to ensure everything was addressed or will be captured.) I will also include your [initial comment](https://github.com/cockroachdb/docs/pull/9913#pullrequestreview-614100354) on the overall recommendations.

knz (approved):
> Reviewed 1 of 4 files at r12, 3 of 5 files at r13, 1 of 1 files at r15, 1 of 1 files at r16.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9913)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @knz, @mikeCRL, @piyush-singh, and @taroface)

___
*[_includes/v20.2/sql/role-considerations.md, line 3 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUowseVHXHi5bNBOHpt-r12-3:-MW3yaQZ1Apzj8YsCKAv:b-6cflco) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/_includes/v20.2/sql/role-considerations.md#L3)):*
<details><summary><i>Previously, mikeCRL (Mike Lewis) wrote…</i></summary><blockquote>

Documenting the current state, plus any differences with Postgres, sounds good to me. Updated in a recent commit, and created an issue https://github.com/cockroachdb/docs/issues/10010 where I also included one other difference that I noticed:

> Difference in how quoted names are treated: always case insensitive in CRDB vs. case **sensitive** _when created using quotes_ in PostgreSQL.
</blockquote></details>

ok thanks
___
*[_includes/v20.2/sql/role-considerations.md, line 4 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUowseVHXHi5bNBOHpt-r12-4:-MW3ygHS-fjH013h_6Gy:bshevsc) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/_includes/v20.2/sql/role-considerations.md#L4)):*
> ```
> A name cannot be the null (0 byte) character
> ```

I think we don't accept the null byte anywhere (since it's neither digit, letter, dash, period or underscrore) so I don't fully understand why you need the extra mention.
___
*[v20.2/authentication.md, line 7 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUnlFKdkqY0Kk-sxJ2W-r12-7:-MW3yscl5_3sq7SS7PnX:bxxxkag) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/v20.2/authentication.md#L7)):*
<details><summary><i>Previously, mikeCRL (Mike Lewis) wrote…</i></summary><blockquote>

I should have added this in my top comment / PR description. Now updated.

It's from https://github.com/cockroachdb/docs/issues/9808, which links back to the originating Slack thread  https://cockroachlabs.slack.com/archives/CGA9F858R/p1613504688098200?thread_ts=1613503866.096200&cid=CGA9F858R. Earlier in the thread, it's agreed that this has been the case since v20.2.
</blockquote></details>

yep that checks out thanks
___
*[v20.2/authentication.md, line 50 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUnlFKdkqY0Kk-sxJ2W-r12-50:-MW3ywy1-abPJfB5pLNq:bghhkid) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/v20.2/authentication.md#L50)):*
<details><summary><i>Previously, mikeCRL (Mike Lewis) wrote…</i></summary><blockquote>

Thanks. Saving for a follow-up issue.
</blockquote></details>

did you have an issue number already?
___
*[v20.2/authentication.md, line 53 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUnlFKdkqY0Kk-sxJ2W-r12-53:-MW3z0342ioBhjQkSVnV:b-k73z76) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/v20.2/authentication.md#L53)):*
<details><summary><i>Previously, mikeCRL (Mike Lewis) wrote…</i></summary><blockquote>

Done. Note that I added the same link to `**Password authentication without TLS**`, as well.
</blockquote></details>

thanks
___
*[v20.2/authorization.md, line 8 at r12](https://reviewable.io/reviews/cockroachdb/docs/9913#-MUtv1tJQA4IyXCh5HNr-r12-8:-MW3zGlx3YnJt4nYKWmw:b5w2f0z) ([raw file](https://github.com/cockroachdb/docs/blob/e318d4f117d77ebc2c4aaa3257a70c7304c698aa/v20.2/authorization.md#L8)):*
<details><summary><i>Previously, mikeCRL (Mike Lewis) wrote…</i></summary><blockquote>

Thanks for the helpful context and analysis.

I've updated the intro to convey the key points you suggested around roles and privileges.

I will include the other reorg (moving creation of users/roles) in a follow-up issue.
</blockquote></details>

cool that's already better

mikeCRL:
@knz 

> I think we don't accept the null byte anywhere (since it's neither digit, letter, dash, period or underscrore) so I don't fully understand why you need the extra mention.

I thought it might be important since the previous doc said "1 to 63 characters" instead of less awkwardly just referring to a 63 character max, and I think Postgres' docs mentioned something along these lines. However, I'm going to drop this before merging. Seems like it is really is an edge case that would quickly be discovered if tested, and a needless distraction to virtually all readers.

> did you have an issue number already?

No. Will be creating that today. Will reply here and update the description.

Thanks again for these quick and thorough reviews.

PR #9923, opened 2021-03-03
Changefeed authentication changes

### LINKED ISSUE #9453: Changefeed Authentication changes
Doc update for - Kafka Changefeed SASL SCRAM-SHA-256 & SASL-SCRAM-SHA-512 support

In 21.2 we are adding support for SASL-SCRAM-SHA-256 and SASL-SCRAM-SHA-512 via a new sasl_mechanism parameter. This is in addition to PLAIN authentication, which we already supported.

Changes to be made: 

**In https://www.cockroachlabs.com/docs/stable/create-changefeed.html#query-parameters**
Under "sasl_enabled" query parameter:
- Right now: "If true, use SASL/PLAIN to authenticate."
- This should be updated to convey that the user can set the authentication protocol through the sasl_mechanism parameter  and that PLAIN is the defualt sasl_mechanism
- On a different note: tls_enabled=true must be set to use SASL, in addition to sasl_enabled, it would be good to make that clear here as well because we don't say that currently

**In https://www.cockroachlabs.com/docs/stable/create-changefeed.html#query-parameters**
Add a "sasl_mechanism" query parameter to the table:
- sink type: kafka
- type: string, options either "SCRAM-SHA-256", "SCRAM-SHA-512", or "PLAIN"
- We should mention that the kafka cluster needs to be configured for the corresponding authentication protocol, and we should link to the confluent page on SCRAM authentication, similarly to how we do in the "sasl_enabled" section - https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_scram.html

**In https://www.cockroachlabs.com/docs/stable/create-changefeed.html#parameters**
We should maybe also update the "Example of a Kafka sink URI" example to include the change:
- Right now: "'kafka://broker.address.com:9092?topic_prefix=bar_&tls_enabled=true&ca_cert=LS0tLS1CRUdJTiBDRVJUSUZ&sasl_enabled=true&sasl_user=petee&sasl_password=bones'"
- New: "'kafka://broker.address.com:9092?topic_prefix=bar_&tls_enabled=true&ca_cert=LS0tLS1CRUdJTiBDRVJUSUZ&sasl_enabled=true&sasl_user=petee&sasl_password=bones&sasl_mechanism=SASL-SCRAM-SHA-256'"

gemma-shay:
Add SASL-SCRAM query params and update example
closes #9453


lnhsingh (commented):
Minor nit, but otherwise LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9923#-:-MV3f9330UXiD5c66FQq:b7nnk2t)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay)

___
*[v20.2/create-changefeed.md, line 82 at r1](https://reviewable.io/reviews/cockroachdb/docs/9923#-MV3eeUt0Q2RDwMOTb0B:-MV3eeUu3EBdOJ0ABFsa:bo4flmm) ([raw file](https://github.com/cockroachdb/docs/blob/87e958376541fe0f5636bb99cc674b3e4cbae648/v20.2/create-changefeed.md#L82)):*
> ```Markdown
> `client_cert`      | [Kafka](#kafka)                               | [`STRING`](string.html)             | The base64-encoded Privacy Enhanced Mail (PEM) certificate. This is used with `client_key`.
> `client_key`       | [Kafka](#kafka)                               | [`STRING`](string.html)             | The base64-encoded private key for the PEM certificate. This is used with `client_cert`.
> `sasl_enabled`     | [Kafka](#kafka)                               | [`BOOL`](bool.html)                 | If `true`, the authentication protocol can be set to SCRAM or PLAIN using the `sasl_mechanism` parameter (see below). You must have `tls_enabled` set to `true` to use SASL. <br><br> **Default:** `false`
> ```

nit: not sure if you the the `(see below)` since it's directly below and not linked.

amruss (approved):
LGTM!

Issue #9065, opened 2020-12-08
Developer Guide v2: Write data


ericharmeling:
This chapter covers topics about writing data to a CockroachDB cluster:

- [ ] https://github.com/cockroachdb/docs/issues/9859
- [ ] https://github.com/cockroachdb/docs/issues/9860
- [ ] https://github.com/cockroachdb/docs/issues/9861
- [ ] https://github.com/cockroachdb/docs/issues/9862
- [ ] https://github.com/cockroachdb/docs/issues/9863
- [ ] https://github.com/cockroachdb/docs/issues/9864
- [ ] https://github.com/cockroachdb/docs/issues/9865

See [tracking doc](https://docs.google.com/document/d/1yHtJqNgdM04xzA7q5rE-mfXg_FkToSnRHed-3gCwHn8/edit#heading=h.ja5q237n5sb8).

ericharmeling:
Closing this out, as we will be tracking individual issues.

PR #9926, opened 2021-03-03
Bulk-update data

### LINKED ISSUE #9863: Developer Guide v2: Bulk-update Data
Part of https://github.com/cockroachdb/docs/issues/9065.

*from planning doc*

>Misc notes:
Requires code snippet.
https://www.cockroachlabs.com/docs/v20.2/delete#batch-deletes

ericharmeling:
Fixes #9863. 

Related to #9065.

### DIFF
--- a/_includes/sidebar-data-v20.2.json
+++ b/_includes/sidebar-data-v20.2.json
@@ -271,6 +271,12 @@
                 ]
               },
               {
+                "title": "Bulk-update Data",
+                "urls": [
+                  "/${VERSION}/bulk-update-data.html"
+                ]
+              },
+              {
                 "title": "Delete Data",
                 "urls": [
                   "/${VERSION}/delete-data.html"
--- a/_includes/sidebar-data-v21.1.json
+++ b/_includes/sidebar-data-v21.1.json
@@ -271,6 +271,12 @@
                 ]
               },
               {
+                "title": "Bulk-update Data",
+                "urls": [
+                  "/${VERSION}/bulk-update-data.html"
+                ]
+              },
+              {
                 "title": "Delete Data",
                 "urls": [
                   "/${VERSION}/delete-data.html"
--- a/v20.2/bulk-update-data.md
+++ b/v20.2/bulk-update-data.md
@@ -1,0 +1,114 @@
+---
+title: Bulk-update Data
+summary: How to update a large amount of data in a cluster
+toc: true
+---
+
+To update multiple rows in a table, you can use a single [`UPDATE` statement](update.html), with a `WHERE` clause that filters the rows you want to update.
+
+To update a large number of rows (i.e., tens of thousands of rows or more), we recommend iteratively updating subsets of the rows that you want to update, until all of the rows have been updated. You can write a script to do this, or you can write a loop into your application.
+
+This page provides guidance on writing batch-update loops with a pattern that executes `SELECT` and `UPDATE` statements at different levels of a nested loop.
+
+{{site.data.alerts.callout_danger}}
+Exercise caution when batch-updating rows from tables with foreign key constraints and explicit [`ON UPDATE` foreign key actions](foreign-key.html#foreign-key-actions). To preserve `UPDATE` performance on tables with foreign key actions, we recommend using smaller batch sizes, as additional rows updated due to `ON UPDATE` actions can make batch loops significantly slower.
+{{site.data.alerts.end}}
+
+## Before you begin
+
+Before reading this page, do the following:
+
+- [Install CockroachDB](install-cockroachdb.html).
+- [Start a local cluster](secure-a-cluster.html), or [create a CockroachCloud cluster](../cockroachcloud/create-your-cluster.html).
+- [Install a Postgres client](install-client-drivers.html).
+
+    For the example on this page, we use the `psycopg2` Python driver.
+- [Connect to the database](connect-to-the-database.html).
+- [Insert data](insert-data.html) that you now want to update.
+
+    For the example on this page, we load a cluster with the `movr` database and data from [`cockroach workload`](cockroach-workload.html).
+
+## Write a batch-update loop
+
+1. At the top level of a loop in your application, or in a script, execute a [`SELECT`](selection-queries.html) query that returns a large batch of primary key values for the rows that you want to update. When defining the `SELECT` query:
+    - Use a `WHERE` clause to filter on columns that identify the rows that you want to update. This clause should also filter out the rows that have been updated by previous iterations of the nested `UPDATE` loop:
+        - For optimal performance, the first condition of the filter should evaluate the last primary key value returned by the last `UPDATE` query that was executed. This narrows each `SELECT` query's scan to the fewest rows possible, and preserves the performance of the row updates over time.
+        - Another condition of the filter should evaluate column values persisted to the database that signal whether or not a row has been updated. This prevents rows from being updated more than once, in the event that the application or script crashes and needs to be restarted. If there is no way to distinguish between an updated row and a row that has not yet been updated, you might need to [add a new column to the table](add-column.html) (e.g., `ALTER TABLE ... ADD COLUMN updated BOOL;`).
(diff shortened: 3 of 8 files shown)


ericharmeling:
@bdarnell 

Adding you for review of the Python code. I followed the non-indexed column pattern from our [batch delete docs](https://www.cockroachlabs.com/docs/stable/bulk-delete-data.html#batch-delete-on-a-non-indexed-column). It runs, but I'm not very confident in its efficiency.

bdarnell (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9926)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v21.1/bulk-update-data.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy3TAP8ve6Wrg_J_wM:-MUy3TAP8ve6Wrg_J_wN:bqd81q4) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L31)):*
> ```Markdown
>     For the example on this page, we load a cluster with the `movr` database and data from [`cockroach workload`](cockroach-workload.html).
> 
> ## Write a batch-update loop
> ```

There's a danger with this pattern: if the script crashes and you restart it, you'll update the first rows more than once. (and the second run will process newer rows that wouldn't have been included the first time) You need some way in the database to tell which rows have already been processed. This tends to be application-dependent (for the "discount" example here, you'd want a `discount_applied` boolean column on the table, with corresponding application changes to apply the discount to new records and set this flag accordingly).  
___
*[v21.1/bulk-update-data.md, line 43 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy55KvE5LJYHbkJ7B4:-MUy55KvE5LJYHbkJ7B5:b-ci14kz) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L43)):*
> ```Markdown
> ## Example
> 
> Suppose that you have recorded hundreds of thousands of [MovR](movr.html) rides in a cluster loaded with the [`movr`](cockroach-workload.html#movr) database. And suppose that you want to apply a discount to all of the rides that cost more than 70¤. To do this, you need to update all of the rows in the `rides` table where the `revenue` column is greater than `70`. You can create a script that loops over the data and updates the relevant rows in batches, following the query guidance provided above.
> ```

This use case doesn't make a lot of sense - changing the price of trips that have already happened (and presumably already been charged)? 
___
*[v21.1/bulk-update-data.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy9dNrBV2QxTpyGPlg:-MUy9dNrBV2QxTpyGPlh:b-1ayoy9) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L64)):*
> ```Markdown
>                 cur.execute("SET TRANSACTION AS OF SYSTEM TIME '-5s'")
>                 if len(usedpks) > 0:
>                     cur.execute("SELECT id FROM rides WHERE id != ANY %s AND revenue > 70 ORDER BY id LIMIT 10000", (usedpks,))
> ```

Passing `usedpks` back into the DB gets expensive as that set grows. You're ordering by ID so you just want `where ID > %s` with the last ID you processed (the bulk delete page probably needs a similar change if it's doing the same thing).

ericharmeling (commented):
TFTR @bdarnell !

I believe that I've addressed your comments below in the latest commit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9926#-:-MV2t51n6kOS8sbZyio2:b-3ywm40)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell)

___
*[v21.1/bulk-update-data.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy3TAP8ve6Wrg_J_wM:-MV2Su76DBPAFKfxOiER:b-x44dl1) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L31)):*

There's a danger with this pattern: if the script crashes and you restart it, you'll update the first rows more than once. (and the second run will process newer rows that wouldn't have been included the first time) You need some way in the database to tell which rows have already been processed. This tends to be application-dependent (for the "discount" example here, you'd want a `discount_applied` boolean column on the table, with corresponding application changes to apply the discount to new records and set this flag accordingly).  
</blockquote></details>

I've added instructions to add a new `discounted` boolean column. The top-level select query now filters on this column, and the nested update queries update the `discounted` and `revenue` columns in the same statement . 

This removes the need to filter on the primary key (or a set of "used" primary key values) in the top-level select query, which simplifies tracking the columns that still need to be updated.
___
*[v21.1/bulk-update-data.md, line 43 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy55KvE5LJYHbkJ7B4:-MV2G-fUBF6uNnjR0oxb:b-hep856) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L43)):*

This use case doesn't make a lot of sense - changing the price of trips that have already happened (and presumably already been charged)? 
</blockquote></details>

You're right. I updated the language and the use case a little bit to make the example a little more reasonable. TLDR: applied the discount when you charged the user, but you haven't yet updated the `rides` table.

Hopefully this example now works for pedagogical purposes, but I'm open to reconsidering it.
___
*[v21.1/bulk-update-data.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy9dNrBV2QxTpyGPlg:-MV2DS8yABfAUQObk89c:b-b30brs) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L64)):*

Passing `usedpks` back into the DB gets expensive as that set grows. You're ordering by ID so you just want `where ID > %s` with the last ID you processed (the bulk delete page probably needs a similar change if it's doing the same thing). 
</blockquote></details>

You're right. I removed the `usedpks` set. 

I would use the `where id > s%`, but filtering on the new `discounted` boolean column makes it so we don't need to track the rows that have already been updated by primary key.

The non-indexed pattern on the bulk delete page doesn't actually use a `usedpks` set. It just returns all of the filtered rows that haven't yet been deleted.

bdarnell (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9926)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell and @ericharmeling)

___
*[v21.1/bulk-update-data.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy9dNrBV2QxTpyGPlg:-MVIc0P98pMmZzERfDAk:bb906wl) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L64)):*

You're right. I removed the `usedpks` set. 

I would use the `where id > s%`, but filtering on the new `discounted` boolean column makes it so we don't need to track the rows that have already been updated by primary key.

The non-indexed pattern on the bulk delete page doesn't actually use a `usedpks` set. It just returns all of the filtered rows that haven't yet been deleted.
</blockquote></details>

You still want the `where id > %s` clause - without it, the query will get slower over time since it has to skip over all the already-processed rows every time.

ericharmeling (commented):
TFTR @bdarnell !

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9926#-:-MVJDhCL7cLhmW2QHCa8:b-7b0vls)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell)

___
*[v21.1/bulk-update-data.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/9926#-MUy9dNrBV2QxTpyGPlg:-MVIoG9y78Bk0ClMGUtU:bz3vudb) ([raw file](https://github.com/cockroachdb/docs/blob/19cfb9b1f206efc29c353f65622e3b50afa962bd/v21.1/bulk-update-data.md#L64)):*

You still want the `where id > %s` clause - without it, the query will get slower over time since it has to skip over all the already-processed rows every time. 
</blockquote></details>

Makes sense! Done.

ianjevans (approved):

PR #9935, opened 2021-03-05
Update benchmarking docs to use built-in workload

### LINKED ISSUE #5067: References to workload should use `cockroach workload` instead of a separate bin
As `workload` has changed, we've found it increasingly hard to keep the binary backward-compatible with all prior releases of CockroachDB. This results in errors like we see in https://forum.cockroachlabs.com/t/workload-fixture-import-failing-on-version-error/2927.

To address this, we began packaging `workload` inside of `cockroach` distributions, which ensures that it stays in-sync with the version of Cockroach that is being used. Users can access this by running the `cockroach workload` subcommand. We should update all documentation to use `cockroach workload ...` instead of `workload ...` to ensure that users don't continue to run into issues like the one linked above when trying to follow our documentation.

Another option is to update the docs to pull specific versions of `workload` depending on the corresponding release instead of always pulling the latest `workload` binary. I don't think this is quite as good of a solution, but we can consider it.

cc. @danhhz

### LINKED ISSUE #6325: Update tpc-c docs to use fixtures import
@keith-mcclellan noticed that a few tpc-c docs still don't use fixtures import. Fixtures import is significantly faster and we should consider updating the remaining docs to use it.

10 warehouses doesn't use it while 1000 and up do here:
https://www.cockroachlabs.com/docs/v19.2/performance-benchmarking-with-tpc-c-10-warehouses.html

The performance benchmarking page also doesn't use it.

https://www.cockroachlabs.com/docs/v19.2/training/performance-benchmarking.html#main-content

### LINKED ISSUE #9774: Performance Benchmarking with TPC-C: workload fixtures import tppcc fails

jseldess:
Previously, we were using the external `workload` binary.

Unfortunately, `cockroach workload fixtures import`, which
we use to more quickly get tpcc data into a cluster,
is undocumented. We'll handle that separately.

Fixes #5067
Fixes #9774
Fixes #6325

### DIFF
--- a/v20.2/performance-benchmarking-with-tpc-c-10-warehouses.md
+++ b/v20.2/performance-benchmarking-with-tpc-c-10-warehouses.md
@@ -1,6 +1,0 @@
----
-title: Performance Benchmarking with TPC-C
-summary: Learn how to benchmark CockroachDB against TPC-C 13k on a local cluster
-toc: true
-toc_not_nested: true
----
--- a/v20.2/performance-benchmarking-with-tpc-c-100k-warehouses.md
+++ b/v20.2/performance-benchmarking-with-tpc-c-100k-warehouses.md
@@ -1,6 +1,0 @@
----
-title: Performance Benchmarking with TPC-C
-summary: Learn how to benchmark CockroachDB against TPC-C with 81 nodes on `c5d.9xlarge` machines
-toc: true
-toc_not_nested: true
----
--- a/v20.2/performance-benchmarking-with-tpc-c-10k-warehouses.md
+++ b/v20.2/performance-benchmarking-with-tpc-c-10k-warehouses.md
@@ -1,6 +1,0 @@
----
-title: Performance Benchmarking with TPC-C
-summary: Benchmark CockroachDB against TPC-C with 15 nodes on `c5d.4xlarge` machines
-toc: true
-toc_not_nested: true
----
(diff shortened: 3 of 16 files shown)


nvanbenschoten (approved):
Looks great! Thanks Jesse.

> Reviewed 16 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9935#-:-MV31DQG1d8j3uXHhi1e:bqwqqao)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @jseldess)

jseldess:
TFTR, @nvanbenschoten.

PR #9958, opened 2021-03-09
Increase whitespace between paragraphs

### LINKED ISSUE #9849: Increase whitespace for paragraphs and lists
Paragraphs do not have enough white space between them ([example](https://www.cockroachlabs.com/docs/stable/schema-design-table.html#column-definition-examples)) . Similarly, there's not enough space between bullet text and it's (unbulleted) sub-text ([example](https://www.cockroachlabs.com/docs/stable/schema-design-table.html#primary-key-best-practices))

jseldess:
Fixes #9849.


ericharmeling (approved):
The changes and the preview look good to me. In the preview, things look waaayy better. 

One question: was `10px` the spacing before the redesign? Have you tried `8px` or `9px`?

taroface:
Looks great. I even feel like there could be slightly more paragraph spacing. There is a fair amount of `line-height` within the paragraphs, and perhaps a bigger margin between paragraphs could help readability. It's (very) slightly hard to parse list items:

image

However, this could just mean that list items should have additional spacing between them.

mikeCRL:
> One question: was `10px` the spacing before the redesign? Have you tried `8px` or `9px`?

Funny, I was going to recommend 11px or 12px. But 10px is fine - certainly better than what we have now.

I see that this change is also being applied to lists, which are gaining the extra space between list items, which IMHO doesn't look great. However, I find there to be a couple existing issues with lists that this is merely amplifying: the line height (space between lines of the same bullet) is too big and there's too much padding after the lists themselves. For example:

Screen Shot 2021-03-09 at 2 15 48 PM

As a first iteration to solve the current style problem, I'd gladly merge this as this and improve the other items later.

jseldess:
I actually ended up increasing the paragraph margin to 10px. I think @taroface is right that it helps. As a follow-up, I'll see if I can tighten the padding between list elements.

taroface:
I think the new `margin-bottom` looks good. I agree with @mikeCRL that the `line-height` value (distance between each line) feels too big and is interfering with readability, both inside and outside of lists.

PR #9963, opened 2021-03-10
Increase spacing around filters

### LINKED ISSUE #9889: Filter buttons don't have enough top/bottom margin
When the content filter buttons are rendered, there isn't enough margin above and below so the content seems squeezed.

Screen Shot 2021-02-24 at 4 11 19 PM
Screen Shot 2021-02-24 at 4 11 28 PM

Adding 5-10 px margins above and below the `filters` class should fix this.

jseldess:
Fixes #9889


jseldess:
My one concern is that there is now space between filters and code blocks, whereas there was none before. 

Examples from live site:
https://www.cockroachlabs.com/docs/cockroachcloud/quickstart.html#step-3-use-the-built-in-sql-client
https://www.cockroachlabs.com/docs/stable/insert-data.html#insert-rows

If we want to retain that, we may need to add a special class for those cases and make sure we always use that class for filters about code blocks. Happy to take a swing at that if we think it's important.

luciaaaaaaa:
Hi - this looks okay. 

One note (this is not directly related to tabs) 
The padding between tabs and the text above of it, is greater than the padding between code blocks and text below them. Ideally we would increase the padding between the code-block and text below so that it is very clear which step it belongs to. 

Here are my wireframes with spacing specifications https://www.figma.com/file/6XcxaUdad2RkjpEqnVHKAG/Education-docs-redesign?node-id=4105%3A8784

jseldess:
Thanks, @luciaaaaaaa! I've adjusted the spacing above and below the filters to match your designs. However, I haven't increase the spacing following a code block to 30px because it feels disproportionate, especially when code blocks follow other code blocks, but also when a step follows. 

Please take another look.

jseldess:
Based on Slack conversation with @luciaaaaaaa, I increase the margin below code blocks to 15px to match the margin below tabs.

Issue #9864, opened 2021-02-23
Developer Guide v2: Delete Data


ericharmeling:
Part of https://github.com/cockroachdb/docs/issues/9065.

*from planning doc*

>Replacing: https://www.cockroachlabs.com/docs/v20.2/delete-data.html 
Delete rows on a condition (Level 2 - page section)
https://www.cockroachlabs.com/docs/v20.2/delete#delete-rows-using-primary-key-unique-columns 
https://www.cockroachlabs.com/docs/v20.2/delete#delete-rows-using-non-unique-columns 
Return deleted rows (Level 2 - page section)
Misc notes:
Requires code snippet.
https://www.cockroachlabs.com/docs/v20.2/delete#return-deleted-rows  
https://www.cockroachlabs.com/docs/v20.2/delete#sorting-the-output-of-deletes 
Advanced operations  (Level 2 - page section)
Force index selection (Level 3 - page section)
Misc notes:
Requires code snippet.
https://www.cockroachlabs.com/docs/v20.2/delete#force-index-selection-for-deletes

PR #9998, opened 2021-03-16
Dev guide v2 Delete Data updates

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/9864

### DIFF
--- a/v21.1/delete-data.md
+++ b/v21.1/delete-data.md
@@ -81,12 +81,12 @@ For more information about how to use the built-in SQL client, see the [`cockroa
 
 {% include copy-clipboard.html %}
 ~~~ go
-// tx is a *sql.Tx from "database/sql"
+// 'db' is an open database connection
 
 tsOne := "2021-03-17 14:00:00"
 tsTwo := "2021-03-17 15:00:00"
 
-if _, err := tx.Exec("DELETE FROM vehicle_location_histories WHERE timestamp BETWEEN $1 AND $2", tsOne, tsTwo); err != nil {
+if _, err := db.Exec("DELETE FROM vehicle_location_histories WHERE timestamp BETWEEN $1 AND $2", tsOne, tsTwo); err != nil {
   return err
 }
 return nil
@@ -104,9 +104,10 @@ String tsOne = "2021-03-17 14:00:00";
 String tsTwo = "2021-03-17 15:00:00";
 
 try (Connection connection = ds.getConnection()) {
-    connection.createStatement()
-            .executeUpdate("DELETE FROM vehicle_location_histories WHERE timestamp BETWEEN '" + tsOne
-                    + "' AND '" + tsTwo + "'");
+    PreparedStatement p = connection.prepareStatement("DELETE FROM vehicle_location_histories WHERE timestamp BETWEEN ? AND ?");
+    p.setString(1, tsOne);
+    p.setString(2, tsTwo);
+    p.executeUpdate();
 
 } catch (SQLException e) {
     System.out.printf("sql state = [%s]\ncause = [%s]\nmessage = [%s]\n", e.getSQLState(), e.getCause(),
@@ -128,7 +129,6 @@ tsTwo = '2021-03-17 15:00:00'
 with conn.cursor() as cur:
     cur.execute(
         "DELETE FROM vehicle_location_histories WHERE timestamp BETWEEN %s AND %s", (tsOne, tsTwo))
-conn.commit()
 ~~~
 
 </section>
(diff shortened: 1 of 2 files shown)


v21.1/delete-data.md
```diff
@@ -78,38 +122,99 @@ try (Connection connection = ds.getConnection()) {
...
 
 </section>
 
-## Delete multiple rows
+<section class="filter-content" markdown="1" data-scope="go">
 
-You can delete multiple rows from a table in several ways:
+{% include copy-clipboard.html %}
+~~~ go
+// tx is a *sql.Tx from "database/sql"
```
rafiss, line 166: we could also use a `*sql.DB` here, which matches the other examples that don't use an explicit transaction

v21.1/delete-data.md
```diff
@@ -78,38 +122,99 @@ try (Connection connection = ds.getConnection()) {
...
-    DELETE FROM accounts WHERE id IN (1, 2, 3, 4, 5);
-    ~~~
+String codeOne = "0_explain_theory_something";
+String codeTwo = "100_address_garden_certain";
+String codeThree = "1000_do_write_words";
 
-- Using [`TRUNCATE`](truncate.html) instead of [`DELETE`](delete.html) to delete all of the rows from a table, as recommended in our [performance best practices](performance-best-practices-overview.html#use-truncate-instead-of-delete-to-delete-all-rows-in-a-table).
+try (Connection connection = ds.getConnection()) {
+    connection.createStatement().executeUpdate(
+            "DELETE from promo_codes WHERE code IN('" + codeOne + "','" + codeTwo + "','" + codeThree + "')");
```
rafiss, line 192: it's risky to put in raw strings like this into SQL statements, since it allows for SQL injection. of course, it doesn't matter for this toy example, but we might as well use a best practice here so the reader doesn't copy+paste something dangerous.

```
PreparedStatement p = connection.prepareStatement("DELETE from promo_codes WHERE code IN(?, ?, ?)");
p.setString(1, codeOne);
p.setString(2, codeTwo);
p.setString(3, codeThree);
p.executeUpdate(); 
```

(it also seems unrealistic to write real JDBC code that way, since a framework on top would take care of this, but at least this way is "safe")

rafiss (changes requested):
minor java comment

v21.1/delete-data.md
rafiss: this is my personal preference, and also would be a big project, but i wonder if we could think about using [JDBI](https://jdbi.org/) for stuff like this -- where we want to run raw SQL with Java. it's a lightweight library that helps with query building and stuff. wouldn't say this is a big priority though.

ericharmeling (commented):
TFTR @rafiss !

Updated the code samples with your feedback.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/9998#-:-MW58vmC0Qt1MRRU3lQA:b-r58bw8)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

___
*[v21.1/delete-data.md, line 166 at r3](https://reviewable.io/reviews/cockroachdb/docs/9998#-MVvkakcHrmh27Y3KLiw-r3-166:-MW5QgEXDvo3mHhojC9i:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/b010a2384804a74b04e1d1149539fd465fcdfc1e/v21.1/delete-data.md#L166)):*

we could also use a `*sql.DB` here, which matches the other examples that don't use an explicit transaction
</blockquote></details>

Done.
___
*[v21.1/delete-data.md, line 192 at r3](https://reviewable.io/reviews/cockroachdb/docs/9998#-MVvkakcHrmh27Y3KLiw-r3-192:-MW5G83n6cfu_PEmesTD:bs5t0ew) ([raw file](https://github.com/cockroachdb/docs/blob/b010a2384804a74b04e1d1149539fd465fcdfc1e/v21.1/delete-data.md#L192)):*
> it's risky to put in raw strings like this into SQL statements, since it allows for SQL injection.

That makes sense (the python and golang examples do this correctly for this reason).

I've updated the Java samples.

> it also seems unrealistic to write real JDBC code that way, since a framework on top would take care of this, but at least this way is "safe"

Yeah.... it makes me wonder how useful even providing these low-level driver examples actually is.

> i wonder if we could think about using JDBI for stuff like this

That seems like a bigger project. But just skimming their docs and looking at the sample you gave makes me think it could be worth investing some time into, just to make these kinds of simple examples a little more succinct. I've made an issue: https://github.com/cockroachdb/docs/issues/10014

ianjevans (approved):

PR #10040, opened 2021-03-23
Fix docker pull command for 21.1 beta

jseldess:


jseldess:
Looks like we forgot to remove the `--prod-release` flag when generating the notes: https://cockroachlabs.atlassian.net/wiki/spaces/ENG/pages/73105625/Release+process#Releaseprocess-Generatereleasenotes

lnhsingh (approved):
Thanks for the fix!

PR #10041, opened 2021-03-23
Remove extra line break in tab name

### LINKED ISSUE #9918: VPC Peering / AWS PrivateLink tabs are misaligned
Observed on https://www.cockroachlabs.com/docs/cockroachcloud/connect-to-your-cluster.html#establish-vpc-peering-or-aws-privatelink 

Screen Recording 2021-03-02 at 07 00 PM

cc @gemma-shay IIRC you made an adjustment to smooth that out. Looks like it's back - could we tee up a fix here?

mikeCRL:
Fixes #9918

### DIFF
--- a/cockroachcloud/connect-to-your-cluster.md
+++ b/cockroachcloud/connect-to-your-cluster.md
@@ -57,7 +57,7 @@ CockroachCloud requires you to authorize the networks that can access the cluste
 VPC peering is only available for GCP clusters, and AWS PrivateLink is only available for AWS clusters.
 
 <div class="filters clearfix">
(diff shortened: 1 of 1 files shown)


gemma-shay (approved):
LGTM!

jseldess:
@mikeCRL, @gemma-shay, the tabs now look a bit odd on hover:

Screen Shot 2021-03-25 at 12 34 54 AM

Can we remove `style="width: 15%"` from each button? I think that would resolve this.

mikeCRL:
@jseldess That did resolve the issue here. Thanks. Merging with that change.

PR #10157, opened 2021-03-24
Modify Gemfile for Ruby 3.0.0 users

yingzhuchin:
https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

> The following libraries are no longer bundled gems or standard libraries. Install the corresponding gems to use these features.
sdbm
webrick
net-telnet
xmlrpc

As people start to use the latest Ruby for local builds, they might run into issues with our current Gemfile. Per above, Ruby 3 no longer bundles certain gems, including `webrick`, which is needed by Jekyll (https://jekyllrb.com/docs/configuration/webrick/).

Therefore install `webrick` explicitly for now. This will not affect users or CI running older Ruby versions. Our TeamCity Docker image and Netlify are both using older Ruby versions.

We will also explicitly install `rss` for now because @jseldess (who is running Ruby 3) saw dependency issues for that. In Ruby 3 (again per above link), `rss` is now a bundled gem instead of a default gem. We *shouldn't* need to explicitly install a bundled gem, but doing it for now to unblock Jesse while we continue to investigate.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,9 @@ gem "jekyll-redirect-from", "~> 0.15"
 gem "jekyll-sitemap", "~> 1.3.1"
 gem "liquid-c", "~> 4.0.0"
 gem "redcarpet", "~> 3.5"
+gem "rss"
 gem "sassc", "~> 2.2"
+gem "webrick"
 
 group :jekyll_plugins do
     gem "jekyll-include-cache"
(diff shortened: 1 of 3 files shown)


jseldess:
Thank you, @yingzhuchin. LGTM.

jseldess (approved):

jseldess:
@yingzhuchin, any chance we can get this merged? I know it's just for me at the moment, but it would make my life easier.

yingzhuchin:
> @yingzhuchin, any chance we can get this merged? I know it's just for me at the moment, but it would make my life easier.

Yes, will do today!

PR #10166, opened 2021-03-24
remove some unused fonts and reorder how js loads

anagio:
optimize loading of assets
cleanup fonts


anagio:
- Theme blue css imported to our main css file
- remove extra market form js
- remove vwo visual website optimizer
- move some js to footer
- add preconnect, dns-prefetch and preload attributes to some header link tags
- replace document.write with better js
- remove unused fonts
- remove fontawesome and replacing with png icons
- add font-display fallback
- load large header images though Imgix auto compressing and sizing

_layouts/default.html
```diff
@@ -53,7 +51,202 @@
 
           {% include_cached footer.html %}
 
-
+
```
yingzhuchin, line 54: This move from `search.html` could be confusing from a contributing standpoint. Can you explain more the enhancement trade-off here?

_includes/head.html
```diff
@@ -24,30 +23,27 @@
 };
 </script>
 
-{% include_cached vwo.html %}
```
yingzhuchin, line 27: We can remove this from the `_includes` folder then?

_includes/marketo-install.html
```diff
@@ -1,6 +1,5 @@
 <form id="release-mktoForm_1082_{{ include.uid }}" class="mkto-install-form"></form>
```
yingzhuchin, line 1: Are we using this file anywhere?

yingzhuchin, line 1: nvm, found it!

yingzhuchin (commented):
Had a few comments, LGTM otherwise!

_includes/head.html
anagio: Yes we can going to remove it

_layouts/default.html
anagio: ok algolia js is back with the search html file

Issue #10138, opened 2021-03-23
release-21.1: sql: make jobs that involve dropping enum values non-cancellable


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/61733

From release notes:
> `TYPE SCHEMA CHANGE` jobs which include a `DROP VALUE` in them are now cancellable. All other jobs remain non-cancellable. [#61733][#61733] {% comment %}doc{% endcomment %}

Issue #10137, opened 2021-03-23
release-21.1: sql: move `ALTER TYPE ... DROP VALUE` behind a feature flag


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/61723

From release notes:
> [`ALTER TYPE ... DROP VALUE`](../v21.1/alter-type.html) is gated behind a feature flag. The [session setting](../v21.1/set-vars.html) is called `enable_drop_enum_value` and the corresponding [cluster setting](../v21.1/cluster-settings.html) is called `sql.defaults.drop_enum_value.enabled`. [#61723][#61723] {% comment %}doc{% endcomment %}

Issue #10060, opened 2021-03-23
sql: add support for dropping enum values


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/58688

From release notes:
> The [`ALTER TYPE ... DROP VALUE ...`](../v21.1/alter-type.html) statement has been added to drop values from an [`ENUM`](../v21.1/enum.html) type. The statement drops the provided value if it isn't used as a value in any table's row. The use of this syntax is gated behind the `sql_safe_updates` [session variable](../v21.1/set-vars.html), as it is susceptible to the value being used in a table expression (such as a computed column). [#58688][#58688] {% comment %}doc{% endcomment %}

PR #10172, opened 2021-03-24
ALTER TYPE ... DROP VALUE

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/10138.
Fixes https://github.com/cockroachdb/docs/issues/10060.
Fixes https://github.com/cockroachdb/docs/issues/10137.

### DIFF
--- a/_includes/v21.1/sql/generated/diagrams/alter_type.html
+++ b/_includes/v21.1/sql/generated/diagrams/alter_type.html
@@ -1,4 +1,4 @@
-<div><svg width="799" height="355">
+<div><svg width="839" height="399">
 <polygon points="11 17 3 13 3 21"></polygon>
 <polygon points="19 17 11 13 11 21"></polygon>
 <rect x="33" y="3" width="62" height="32" rx="10"></rect>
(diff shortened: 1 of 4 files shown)


arulajmani (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10172)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v21.1/alter-type.md, line 39 at r1](https://reviewable.io/reviews/cockroachdb/docs/10172#-MWj9e2N9rXwBO0E22kX:-MWj9e2OCOWjg5laU6DP:bj50jtd) ([raw file](https://github.com/cockroachdb/docs/blob/caf9786eba0575432c9ddd5f930c300272a59a2c/v21.1/alter-type.md#L39)):*
> ```Markdown
> 
> - You can only reference a user-defined type from the database that contains the type.
> - You can only [cancel](cancel-job.html) `ALTER TYPE`[schema change jobs](online-schema-changes.html) that drop values. All other `ALTER TYPE` schema change jobs are non-cancellable.
> ```

I wouldn't call this a limitation, it is by design -- all other jobs are expected to finish very quickly, but drop value jobs can take an arbitrarily long time. To that effect, these jobs are cancellable. 

One known limitation of `ALTER TYPE .. DROP VALUE` is that if the value is used in a table expression, such as a computed column or default expressions or view expressions, those expressions may stop working after the drop. Worth documenting I think. 
___
*[v21.1/cancel-job.md, line 20 at r1](https://reviewable.io/reviews/cockroachdb/docs/10172#-MWjA4OFEBuveImnAmlz:-MWjA4OFEBuveImnAmm-:bt3r3v6) ([raw file](https://github.com/cockroachdb/docs/blob/caf9786eba0575432c9ddd5f930c300272a59a2c/v21.1/cancel-job.md#L20)):*
> ```Markdown
> 
> - When an enterprise [`RESTORE`](restore.html) is canceled, partially restored data is properly cleaned up. This can have a minor, temporary impact on cluster performance.
> - You can only cancel [`ALTER TYPE`](alter-type.html) [schema change jobs](online-schema-changes.html) that drop values. All other `ALTER TYPE` schema change jobs are non-cancellable.
> ```

See my comment about limitations above.

ericharmeling (commented):
TFTR @arulajmani !

I think I addressed your comments. If everything looks good, I'll pass it over to writer review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10172#-:-MWz1ydFAKO3N0PbVl6I:b-syy0gh)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @arulajmani)

___
*[v21.1/alter-type.md, line 39 at r1](https://reviewable.io/reviews/cockroachdb/docs/10172#-MWj9e2N9rXwBO0E22kX:-MWz-Im13KzNa2kmF4m4:brbe21g) ([raw file](https://github.com/cockroachdb/docs/blob/caf9786eba0575432c9ddd5f930c300272a59a2c/v21.1/alter-type.md#L39)):*
> I wouldn't call this a limitation, it is by design -- all other jobs are expected to finish very quickly, but drop value jobs can take an arbitrarily long time. To that effect, these jobs are cancellable.

Right on. If the text is correct, then I'll just make it a note at the top of the page. 

> One known limitation of ALTER TYPE .. DROP VALUE is that if the value is used in a table expression, such as a computed column or default expressions or view expressions, those expressions may stop working after the drop. Worth documenting I think.


I had a note in the `DROP VALUE` parameter description for this, but I agree that it is more of a limitation. I added it to the limitations and trimmed the parameter description a bit.
___
*[v21.1/cancel-job.md, line 20 at r1](https://reviewable.io/reviews/cockroachdb/docs/10172#-MWjA4OFEBuveImnAmlz:-MWz-SnF7bDiRtUu7W2P:b-gqza0s) ([raw file](https://github.com/cockroachdb/docs/blob/caf9786eba0575432c9ddd5f930c300272a59a2c/v21.1/cancel-job.md#L20)):*
<details><summary><i>Previously, arulajmani (Arul Ajmani) wrote…</i></summary><blockquote>

See my comment about limitations above. 
</blockquote></details>

Removed this from the list of limitations.

ianjevans (approved):

PR #10179, opened 2021-03-25
Update arrow directions in sidebar

### LINKED ISSUE #9433: Menu navigation arrows in the left nav are confusing
Arrows in left nav are confusing and makes it difficult to know if there are more items within a section.

I think when a section is closed they should be facing right, when a section is open the should face down (similar to how the Mac Finder works).

Screen Shot 2021-01-21 at 11 41 56 AM

Screen Shot 2021-01-21 at 11 46 25 AM

jseldess:
When section is closed, arrow now points to right.
When section is open, arrow now points down.

Fixes #9433


luciaaaaaaa:
This looks good to me!

jseldess:
Thanks, @luciaaaaaaa. There's probably a more elegant way to achieve this result, but I'll merge for now. @anagio, let me know if you have suggestions.

PR #10180, opened 2021-03-25
Make install options more obviously options

### LINKED ISSUE #10174: Improve the installation docs
## Issue Description
All first time users need to download the CRDB binary to connect to their cluster (whether on CRDB or CC, this is usually a required step), but if you are not an experienced developer our download docs are extremely confusing. [This page](https://www.cockroachlabs.com/docs/v20.2/install-cockroachdb-mac.html) makes it seem like you need to follow every single step listed, but in reality, you would just be downloading the binary a bunch of times in different ways. In addition to running into this problem personally when I first interviewed at CRL, we've recently seen CC interviewees who are Hackathon students get totally tripped up by this page. 

## Suggested Resolution 
Add a sentence at the top of the page "Use one of these methods to install the CockroachDB binary". Label each download method as an option, eg "Option 1: Install with Homebrew", "Option 2: Download the Binary", "Option 3: Use Kuberenets", etc.

Doing this will make it much clearer that all of these steps are not required and are in fact different variations of the same thing.

jseldess:
Add an h2 clarifying that all options on the page
are, in fact, options. Only worth doing this for
21.1, 20.2, and 20.1, though.

Fixes #10174

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,9 +11,12 @@ gem "jekyll-sitemap", "~> 1.3.1"
 gem "liquid-c", "~> 4.0.0"
 gem "redcarpet", "~> 3.5"
 gem "sassc", "~> 2.2"
+gem "rss"
 
 group :jekyll_plugins do
     gem "jekyll-include-cache"
     gem 'jekyll-algolia', "~> 1.0", path: "./jekyll-algolia-dev"
     gem 'jekyll-remote-include', "~> 1.1.3", github: 'ianjevans/jekyll-remote-include'
   end
+
+gem "webrick", "~> 1.7"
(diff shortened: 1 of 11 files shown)


emily-horing (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10180)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @jseldess)

___
*[v20.1/install-cockroachdb-linux.html, line 19 at r1](https://reviewable.io/reviews/cockroachdb/docs/10180#-MWe7Vu1BCm8OU-_kZgM:-MWe7Vu1BCm8OU-_kZgN:b-g7vd5j) ([raw file](https://github.com/cockroachdb/docs/blob/b058a51d2ef070fe19beaec2bcdaf52e5391cdb2/v20.1/install-cockroachdb-linux.html#L19)):*
> ```HTML
> {% include cockroachcloud/use-cockroachcloud-instead.md %}
> 
> <h2>Install options</h2>
> ```

Is it possible to add one line beneath this saying something like "Use one of the options below to download CockroachDB"
Thanks for updating this jesse!

jseldess:
Done. PTAL, @emily-horing.

PR #10188, opened 2021-03-26
Reference new connection pooling topic for prod checklists

### LINKED ISSUE #6706: Add production guidance on connection pooling
The CockroachCloud docs make recommendations around [connection pooling](https://www.cockroachlabs.com/docs/cockroachcloud/stable/cockroachcloud-production-checklist.html#use-a-small-pool-of-persistent-connections). We should vet those recommendations, update and expand them if necessary, and expose them in the CockroachDB [Production Checklist](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html) as well, with adjustments as needed.

There's a general sense from SEs that it's more expensive to open connections than to maintain them and that a larger default pool (e.g., 50 or 100) per app instance may be best.

Once possible factor is password vs. cert-based auth. CC intentionally uses a cpu heavy hash algorithm (bcrypt) to make it difficult to brute force a password match.

jseldess:
Fixes #6706

### DIFF
--- a/v20.2/recommended-production-settings.md
+++ b/v20.2/recommended-production-settings.md
@@ -250,11 +250,9 @@ Environment | Featured Approach
 
 ## Connection pooling
 
-Multiple active connections to the database enable efficient use of the available database resources. However, creating new authenticated connections to the database is CPU and memory-intensive and also adds to the latency since the application has to wait for database to authenticate the connection.
+Creating the appropriate size pool of connections is critical to gaining maximum performance in an application. Too few connections in the pool will result in high latency as each operation waits for a connection to open up. But adding too many connections to the pool can also result in high latency as each connection thread is being run in parallel by the system. The time it takes for many threads to complete in parallel is typically higher than the time it takes a smaller number of threads to run sequentially.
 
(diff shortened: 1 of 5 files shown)


v20.2/recommended-production-settings.md
```diff
@@ -250,11 +250,9 @@ Environment | Featured Approach
...
 
-Connection pooling helps resolve this dilemma by creating a set of authenticated connections that can be reused to connect to the database. To determine the size of the maximum connection pool, our recommendation is to start your testing with a pool size of  `(core_count * 2) + ssd_count)`, where `core_count` is the total number of cores in the cluster, and `ssd_count` is the total number of SSDs in the cluster. The optimal pool size will vary based on your workload and application, and may be smaller than that.
-
-In addition to setting a maximum connection pool size, the idle connection pool size must also be set. Our recommendation is to set the idle connection pool size equal to the maximum pool size. While this uses more application server memory, this allows there to be many connections when concurrency is high, without having to dial a new connection for every new operation.
```
jseldess, line 257: @ianjevans, this is the only bit I'm wondering if we need to retain, perhaps even in the main connection pooling topic. Thoughts?

ianjevans: The second paragraph could definitely be added to the main connection pooling topic.

The recommendations in the first paragraph are superseded by the new formula outlined in the main connection pooling topic, per @a-entin.

jseldess:
Thanks for the fixes, @ianjevans! Please merge when ready.

ianjevans (approved):

Issue #9040, opened 2020-12-02
Document "log-driven closed timestamps"


rmloveland:
Richard Loveland (rmloveland) commented:

In order to enable non-blocking transactions (#9038), we will also be adding something called "log-driven closed timestamps".  AIUI these will update our existing closed timestamps to remove some current limitations.

There may not be docs work here, but we do mention closed timestamps in the Follower Reads docs, so we may need to at least update that.

See also:

- #6793
- https://github.com/cockroachdb/cockroach/pull/56675

Jira Issue: DOC-844

rmloveland:
Fixed in #11762

PR #10190, opened 2021-03-26
Add architecture docs on non-blocking transactions

### LINKED ISSUE #9704: Document non-blocking transactions
Non-blocking transactions provide the foundation for Multiregion `GLOBAL` tables.  They are a variant of CockroachDB's standard read-write transaction protocol that permit low-latency, global reads of read-mostly and read-only (excluding maintenance events) data.

See also:

- https://github.com/cockroachdb/cockroach/pull/52745
- https://github.com/cockroachdb/docs/issues/9301
- https://github.com/cockroachdb/docs/issues/9268
- https://github.com/cockroachdb/docs/issues/9269

rmloveland:
Fixes #9704.
See also #9040.

Summary of changes:

- Add a new section to the 'Transaction Layer' doc: 'Non-blocking
  transactions', where we provide a high-level overview of the operation
  and benefits of these types of transactions, followed by some slightly
  lower-level (but still handwavy) description of how they work.


rmloveland:
Hi Nathan!

I had some additional questions re: this PR & this topic:

- Is anything glaring missing?
- Would this be better presented in another order?
- Any obvious errors?
- I chose not to update the "time and HLC" section earlier on the page at all, since conceptually (at least AIUI from the RFC) the synthetic timestamp is still a timestamp; therefore we don't yet in that earlier section need to really care about the details of non-blocking transactions and synthetic timestamps, etc.  Do you agree with that?
- I noted in the RFC there was mention of dropping the clock uncertainty to something lower than the current default.  I punted on that (so far) in this PR.  Is there something we should do in the docs for v21.1 around max uncertainty settings?  (maybe in this PR? or some other one?)

nvanbenschoten (commented):
Sorry for the delay on this, Rich! It's exciting to see this make its way into the docs. The big thing missing here is that a discussion of the "commit wait" stage that transactions that commit at a future time need to perform. We'll also want to mention that a reader of one of these future-time writes can have their timestamp bumped into the future as well if the write falls in their uncertainty window. And so a conflicting reader may also need to commit-wait.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190#-:-MXhkVNpAIJM7E4CICau:b-8xf7p7)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rmloveland)

___
*[v21.1/architecture/transaction-layer.md, line 371 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhj9M754aataaTgC7C:-MXhj9M754aataaTgC7D:bopryxx) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L371)):*
> ```Markdown
> ## Non-blocking transactions
> 
> <span class="version-tag">New in v21.1:</span> CockroachDB supports low-latency, global reads of read-mostly data in [multi-region clusters](../multiregion-overview.html) using _non-blocking transactions_: an extension of the [standard read-write transaction protocol](#overview) that allows a transaction to perform [locking](#concurrency-control) in a manner such that contending reads by other transactions can avoid waiting on its locks.
> ```

`s/allows a transaction/allows a writing transaction/`
___
*[v21.1/architecture/transaction-layer.md, line 388 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjShY02l53DfaVh-8:-MXhjShY02l53DfaVh-9:bqt6a40) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L388)):*
> ```Markdown
> Non-blocking transactions are implemented via _non-blocking ranges_. Every non-blocking range has the following properties:
> 
> - Any transaction that writes to it is marked as a non-blocking transaction.
> ```

I would re-word this to say that any transaction that writes to one of these ranges has their write timestamp pushed into the future.
___
*[v21.1/architecture/transaction-layer.md, line 389 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjbNn02Ce0L3XP1Bm:-MXhjbNn02Ce0L3XP1Bn:b-xwkpxe) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L389)):*
> ```Markdown
> 
> - Any transaction that writes to it is marked as a non-blocking transaction.
> - It is able to propagate a [closed timestamp](../follower-reads.html#how-follower-reads-work) in the future of present time, also known as a _synthetic timestamp_. Synthetic timestamps are not bound to a specific clock value, and are treated especially by the [hybrid logical clock mechanism](#time-and-hybrid-logical-clocks) to provide consistency guarantees. Essentially, the HLC waits until it advances past the synthetic timestamp on its own, or advances due to updates from other, non-synthetic ("standard") timestamps.
> ```

I don't think we should mention synthetic timestamps. They're an implementation detail. Referencing "future time" throughout these docs should be enough.
___
*[v21.1/architecture/transaction-layer.md, line 389 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjxrv6dc3DXre5HzX:-MXhjxrv6dc3DXre5HzY:bv6fh2o) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L389)):*
> ```
> Essentially, the HLC waits
> ```

We should break this into a separate bullet point that discusses that transactions the commit with future time commit timestamps need to "commit wait", which means that they have to wait until the HLC clock advances past their commit timestamp.

rmloveland (commented):
Thanks for the review Nathan! No rush was needed :-) 

I tried to address your feedback below - in particular added things about commit-wait, and made sure to capture your comment here re: conflicting readers maybe having to commit-wait, too

PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190#-:-MYBOHfBCCK_-kJHC962:b-hqnvu4)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nvanbenschoten)

___
*[v21.1/architecture/transaction-layer.md, line 371 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhj9M754aataaTgC7C:-MYBGYIZ18DchtfKtkKB:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L371)):*

`s/allows a transaction/allows a writing transaction/`
</blockquote></details>

Updated!
___
*[v21.1/architecture/transaction-layer.md, line 388 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjShY02l53DfaVh-8:-MYBGhU4B4oGOyBCA1gw:b-50rmxi) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L388)):*

I would re-word this to say that any transaction that writes to one of these ranges has their write timestamp pushed into the future.
</blockquote></details>

Revised to use that phrasing
___
*[v21.1/architecture/transaction-layer.md, line 389 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjbNn02Ce0L3XP1Bm:-MYBGqPUCg0txfp8VH_k:b7ducb3) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L389)):*

I don't think we should mention synthetic timestamps. They're an implementation detail. Referencing "future time" throughout these docs should be enough.
</blockquote></details>

Revised all mentions of 'synthetic timestamp' to say something like 'timestamp in the future' or 'future timestamp'
___
*[v21.1/architecture/transaction-layer.md, line 389 at r1](https://reviewable.io/reviews/cockroachdb/docs/10190#-MXhjxrv6dc3DXre5HzX:-MYBOXmo14quPxbycgMv:b-aji96b) ([raw file](https://github.com/cockroachdb/docs/blob/b68b153a398981323f7b5ddd86bc3a8dd0e962f6/v21.1/architecture/transaction-layer.md#L389)):*

> ```
> Essentially, the HLC waits
> ```

We should break this into a separate bullet point that discusses that transactions the commit with future time commit timestamps need to "commit wait", which means that they have to wait until the HLC clock advances past their commit timestamp.
</blockquote></details>

I added two new points:

- one about transactions that write to the range needing to commit wait as you mention in this comment
- another about transactions that read from one of the future writes to this range also maybe needing to commit-wait (per your comment at the top of this review)

I have tried to revise the language to fit the list format of "every non-blocking range has the following properties", but I may have introduced some technical errors, or maybe just made it confusing somehow - Please let me know what you think, suggestions welcome!

awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nvanbenschoten and @rmloveland)

___
*[v21.1/architecture/transaction-layer.md, line 375 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYFh2XV6mY6hD9bcPgP:-MYFh2XV6mY6hD9bcPgQ:b-2vfptj) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L375)):*
> ```Markdown
> The non-blocking transaction protocol and replication scheme differ from standard read-write transactions as follows:
> 
> - Non-blocking transactions use a replication scheme over the [ranges](overview.html#terms) they operate on that allows all followers in these ranges to serve consistent (non-stale) reads.
> ```

Is followers the term we want to use here? Is it too close to `follower reads` the feature?
___
*[v21.1/architecture/transaction-layer.md, line 394 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYFhTnc80bzr4T1NWYT:-MYFhTnc80bzr4T1NWYU:bidkceg) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L394)):*
> ```Markdown
> 
> As a result of the above properties, all replicas in a non-blocking range are expected to be able to serve transactionally-consistent reads at the present. This means that all follower replicas in a non-blocking range (which includes all [non-voting replicas](replication-layer.html#non-voting-replicas)) implicitly behave as "consistent read replicas", which are exactly what they sound like: read-only replicas that always have a consistent view of the range's current state.
> ```

I think an example transaction would really help this out a lot. One transaction is writing to the table while another is reading for the table and walk through what happens.

nvanbenschoten (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187 and @rmloveland)

___
*[v21.1/architecture/transaction-layer.md, line 375 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYFh2XV6mY6hD9bcPgP:-MYM6yJaCGZ1Z-_K6ksz:b-xgobsl) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L375)):*

Is followers the term we want to use here? Is it too close to `follower reads` the feature?
</blockquote></details>

This is an important conversation to have. Thanks for bringing it up. 

We need to be conscious about how we talk about these concepts. My current thinking is that, if anything, we should be changing how we talk about `follower reads` to avoid confusion and be more precise. Reads from GLOBAL tables are follower reads! They're just not stale like traditional "follower reads".

I think the way to look at this is that there are two different considerations here: the consistency level of the read and the replica that serves the read (leaseholder replica vs. follower/any replica). We were previously conflating these considerations because it was mostly harmless, but with GLOBAL tables and soon with bounded staleness, we're going to need to be more precise here for any of this to make sense to readers.

|      | Strong Reads | Stale Reads (further split into "exact" and "bounded" staleness)     |
| :---        |    :----:   |          ---: |
|Only From Leaseholder      |SELECT      | N/A |
|From Nearest Replica [2] | SELECT on GLOBAL table       | SELECT with AS OF SYSTEM TIME follower_read_timestamp() [1]|

[1] what we currently mean when we say "follower read"
[2] what I think we'll want to start meaning when we say "follower read"

Consistency level refers to the kind of guarantee a read has about which writes that causally preceded it will be observed. The gold standard here is "strong consistency", meaning that the read will observe any prior write. Weaker consistency levels sacrifice this guarantee. Spanner's docs do a very good job presenting this in https://cloud.google.com/spanner/docs/reads and in https://cloud.google.com/spanner/docs/timestamp-bounds, and I think we'll want to do something similar, especially when we introduce bounded staleness.

Which replica a read is served from is a mostly separate concern. Usually, staleness is required to allow a read to be served from a follower replica, but this is not always the case. For instance, GLOBAL tables make this possible even for strong reads. So the table locality and the consistency level both dictate whether a query can be served from a nearby follower replica or not, but they don't define what it means for a query to be a "follower read". If they did, then we'd need a new word for "read that is served from nearest replica".

Another interesting data point is that TiDB has a concept of "follower reads" which mean something even different: https://docs.pingcap.com/tidb/stable/follower-read. In TiDB, all reads ("follower" or otherwise) are strong reads, but this is because "follower reads" still require quickly talking to the leader, they just don't actually read the data from the leader. This is because TiDB cares a lot less about multi-region, where the focus is latency, and a lot more about HTAP, where the focus is resource isolation. So "the main benefits of Follower Read are to isolate read requests from write requests in the cluster and to increase overall read throughput." But all that goes to say that TiDB's use of "follower read" is only in reference to who serves the read and has no bearing on read consistency level.
___
*[v21.1/architecture/transaction-layer.md, line 389 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNA1-20aqMireRG4c4:-MYNA1-20aqMireRG4c5:blzxx5t) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L389)):*
> ```Markdown
> 
> - Any transaction that writes to this range has its write timestamp pushed into the future.
> - The range is able to propagate a [closed timestamp](../follower-reads.html#how-follower-reads-work) in the future of present time. These future time commit timestamps are not bound to a specific clock value. Instead, they are treated especially by the [hybrid logical clock mechanism](#time-and-hybrid-logical-clocks) to provide consistency guarantees.
> ```

I'm not sure that we need to say anything past the first sentence here.
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAJeR1lrj-DXGpLgD:-MYNAJeR1lrj-DXGpLgE:b-95hd2m) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*
> ```Markdown
> - The range is able to propagate a [closed timestamp](../follower-reads.html#how-follower-reads-work) in the future of present time. These future time commit timestamps are not bound to a specific clock value. Instead, they are treated especially by the [hybrid logical clock mechanism](#time-and-hybrid-logical-clocks) to provide consistency guarantees.
> - A transaction that writes to this range and commits with a future time commit timestamp needs to wait until the HLC advances past its commit timestamp. This process is known as _"commit-wait"_. Essentially, the HLC waits until it advances past the future timestamp on its own, or it advances due to updates from other timestamps.
> - A transaction that reads from a future-time write to this range can have its commit timestamp bumped into the future as well, if the write falls in its uncertainty window. Such transactions (a.k.a. "conflicting readers") may also need to commit-wait.
> ```

"reads from" may be ambiguous. Consider just saying "reads a future time-write"
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAQtX8dHdxaV3w0as:-MYNAQtX8dHdxaV3w0at:b-vg8q08) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*
> ```Markdown
> - The range is able to propagate a [closed timestamp](../follower-reads.html#how-follower-reads-work) in the future of present time. These future time commit timestamps are not bound to a specific clock value. Instead, they are treated especially by the [hybrid logical clock mechanism](#time-and-hybrid-logical-clocks) to provide consistency guarantees.
> - A transaction that writes to this range and commits with a future time commit timestamp needs to wait until the HLC advances past its commit timestamp. This process is known as _"commit-wait"_. Essentially, the HLC waits until it advances past the future timestamp on its own, or it advances due to updates from other timestamps.
> - A transaction that reads from a future-time write to this range can have its commit timestamp bumped into the future as well, if the write falls in its uncertainty window. Such transactions (a.k.a. "conflicting readers") may also need to commit-wait.
> ```

s/in its/in the read's uncertainty window/
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAVSkE9f41E8SW25y:-MYNAVSkE9f41E8SW25z:b-87hnah) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*
> ```Markdown
> - The range is able to propagate a [closed timestamp](../follower-reads.html#how-follower-reads-work) in the future of present time. These future time commit timestamps are not bound to a specific clock value. Instead, they are treated especially by the [hybrid logical clock mechanism](#time-and-hybrid-logical-clocks) to provide consistency guarantees.
> - A transaction that writes to this range and commits with a future time commit timestamp needs to wait until the HLC advances past its commit timestamp. This process is known as _"commit-wait"_. Essentially, the HLC waits until it advances past the future timestamp on its own, or it advances due to updates from other timestamps.
> - A transaction that reads from a future-time write to this range can have its commit timestamp bumped into the future as well, if the write falls in its uncertainty window. Such transactions (a.k.a. "conflicting readers") may also need to commit-wait.
> ```

We may want to remind readers that the size of this uncertainty window is dictated by the maximum clock offset configured for the cluster.

awoods187:
Discussed with @rmloveland today that we can split this into two PRs. one to merge above and a separate one with an example

rmloveland (commented):
Thanks for the review Andy!  As mentioned, I filed an issue to add an explanatory example of these transactions to the docs.

Thanks for the additional feedback Nathan!  I think your comments have been addressed in the latest update.  I also filed another issue re: stale vs. strong follower reads based on the discussion below

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190#-:-MYkRcO90ELnU4BlIaNr:b4j11q7)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187 and @nvanbenschoten)

___
*[v21.1/architecture/transaction-layer.md, line 375 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYFh2XV6mY6hD9bcPgP:-MYkOf_5EAS6coMZ7M1m:b-pagqph) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L375)):*

This is an important conversation to have. Thanks for bringing it up. 

We need to be conscious about how we talk about these concepts. My current thinking is that, if anything, we should be changing how we talk about `follower reads` to avoid confusion and be more precise. Reads from GLOBAL tables are follower reads! They're just not stale like traditional "follower reads".

I think the way to look at this is that there are two different considerations here: the consistency level of the read and the replica that serves the read (leaseholder replica vs. follower/any replica). We were previously conflating these considerations because it was mostly harmless, but with GLOBAL tables and soon with bounded staleness, we're going to need to be more precise here for any of this to make sense to readers.

|      | Strong Reads | Stale Reads (further split into "exact" and "bounded" staleness)     |
| :---        |    :----:   |          ---: |
|Only From Leaseholder      |SELECT      | N/A |
|From Nearest Replica [2] | SELECT on GLOBAL table       | SELECT with AS OF SYSTEM TIME follower_read_timestamp() [1]|

[1] what we currently mean when we say "follower read"
[2] what I think we'll want to start meaning when we say "follower read"

Consistency level refers to the kind of guarantee a read has about which writes that causally preceded it will be observed. The gold standard here is "strong consistency", meaning that the read will observe any prior write. Weaker consistency levels sacrifice this guarantee. Spanner's docs do a very good job presenting this in https://cloud.google.com/spanner/docs/reads and in https://cloud.google.com/spanner/docs/timestamp-bounds, and I think we'll want to do something similar, especially when we introduce bounded staleness.

Which replica a read is served from is a mostly separate concern. Usually, staleness is required to allow a read to be served from a follower replica, but this is not always the case. For instance, GLOBAL tables make this possible even for strong reads. So the table locality and the consistency level both dictate whether a query can be served from a nearby follower replica or not, but they don't define what it means for a query to be a "follower read". If they did, then we'd need a new word for "read that is served from nearest replica".

Another interesting data point is that TiDB has a concept of "follower reads" which mean something even different: https://docs.pingcap.com/tidb/stable/follower-read. In TiDB, all reads ("follower" or otherwise) are strong reads, but this is because "follower reads" still require quickly talking to the leader, they just don't actually read the data from the leader. This is because TiDB cares a lot less about multi-region, where the focus is latency, and a lot more about HTAP, where the focus is resource isolation. So "the main benefits of Follower Read are to isolate read requests from write requests in the cluster and to increase overall read throughput." But all that goes to say that TiDB's use of "follower read" is only in reference to who serves the read and has no bearing on read consistency level.
</blockquote></details>

Thanks for this context Nathan - I have filed an issue https://github.com/cockroachdb/docs/issues/10377 to get some of this info into the docs
___
*[v21.1/architecture/transaction-layer.md, line 389 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNA1-20aqMireRG4c4:-MYkOvMp3DXWlN-C4laP:b-lvm236) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L389)):*

I'm not sure that we need to say anything past the first sentence here.
</blockquote></details>

Removed everything after the first sentence
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAJeR1lrj-DXGpLgD:-MYkP2yx3w5hvEzbwsKf:b-896fja) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*

"reads from" may be ambiguous. Consider just saying "reads a future time-write"
</blockquote></details>

Done!
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAQtX8dHdxaV3w0as:-MYkP78P-IHoo5CnV9rB:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*

s/in its/in the read's uncertainty window/
</blockquote></details>

Done
___
*[v21.1/architecture/transaction-layer.md, line 391 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYNAVSkE9f41E8SW25y:-MYkPw-r8k1wJmaFFBBS:b-c8hz0u) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L391)):*

We may want to remind readers that the size of this uncertainty window is dictated by the maximum clock offset configured for the cluster.
</blockquote></details>

Added some words about that, and linked to description of max clock offset elsewhere in this doc.  Also updated that other section to link to the flag on `cockroach start` where it can be set.
___
*[v21.1/architecture/transaction-layer.md, line 394 at r2](https://reviewable.io/reviews/cockroachdb/docs/10190#-MYFhTnc80bzr4T1NWYT:-MYkQ5Kh3lkNCFmG_xfD:b-e8dxb9) ([raw file](https://github.com/cockroachdb/docs/blob/78b3a1811c708b1507b3d5bd97fa5fbf757e365f/v21.1/architecture/transaction-layer.md#L394)):*

I think an example transaction would really help this out a lot. One transaction is writing to the table while another is reading for the table and walk through what happens. 
</blockquote></details>

Yes!  Filed https://github.com/cockroachdb/docs/issues/10378

nvanbenschoten (approved):
:lgtm_strong: this looks great!

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190#-:-MYkUJpo06OuQYjlZWcF:bdu9uj3)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187)

lnhsingh (approved):
:lgtm: nice job!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10190#-:-MYqL7_O64U27uAm4zL_:bxpgx20)* status: :shipit: complete! 2 of 0 LGTMs obtained (waiting on @awoods187)

PR #10215, opened 2021-03-30
Merge range merges info into architecture docs

### LINKED ISSUE #5846: Range merges page should offer more explicit guidance to users
The [Range Merges][merge] page tells me that this feature exists, and why it's good.  However it also tells me how to turn it off, but without making it clear why I should ever do that?  Based on reading the page, merging ranges is only ever good.

Estimated scope of work for this issue:

- Small option (quick & dirty): We should update this page to provide guidance to users about whether they should ever mess with this, or perhaps a note saying "you probably shouldn't touch this" (my guess is 99% should never touch this, and only then if on the advice of CRL support/engineering)
- Large option (slower but better): This information should live on the [Distribution Layer][dist] page in the architecture docs.  It should probably be a new section following **Range Splits**.  The cluster setting should probably not be mentioned.  The separate "Range Merges" page should be deleted.

[merge]: https://www.cockroachlabs.com/docs/v19.2/range-merges.html
[dist]: https://www.cockroachlabs.com/docs/stable/architecture/distribution-layer.html

rmloveland:
Fixes #5846.

Summary of changes:

- Update the 'Distribution Layer' docs with a new section, *Range
  Merges*, to hold (most of) the contents of the formerly separate
  'Range Merges' page.

  - In addition to a light copy edit of the content, we discard the
    instructions for turning off range merges, since (AFAICT) that isn't
    really a thing most users should be doing in the v21.1+
    timeline.  (In fact, based on a quick check it's apparently no
    longer a public cluster setting.)

- Update the former standalone 'Range Merges' page to redirect to the
  new section of the distribution layer docs mentioned above.

### DIFF
--- a/v21.1/architecture/distribution-layer.md
+++ b/v21.1/architecture/distribution-layer.md
@@ -167,9 +167,46 @@ All of these updates to the range descriptor occur locally on the range, and the
 
 ### Range splits
 
-By default, CockroachDB attempts to keep ranges/replicas at 512 MiB. Once a range reaches that limit we split it into two smaller ranges (composed of contiguous key spaces).
+By default, CockroachDB attempts to keep ranges/replicas at the default range size (currently 512 MiB). Once a range reaches that limit, we split it into two smaller ranges (composed of contiguous key spaces).
 
 During this range split, the node creates a new Raft group containing all of the same members as the range that was split. The fact that there are now two ranges also means that there is a transaction that updates `meta2` with the new keyspace boundaries, as well as the addresses of the nodes using the range descriptor.
+
+### Range merges
+
+By default, CockroachDB automatically merges small ranges of data together to form fewer, larger ranges (up to the default range size). This can improve both query latency and cluster survivability.
+
+#### How range merges work
+
+As [described above](#overview), CockroachDB splits your cluster's data into many ranges. For example, your cluster might have a range for customers whose IDs are between `[1000, 2000)`. If that range grows beyond the default range size, the range is [split into two smaller ranges](#range-splits).
+
+However, as you delete data from your cluster, a range might contain far less data than the default range size. Over the lifetime of a cluster, this could lead to a number of small ranges.
+
+To reduce the number of small ranges, your cluster can have any range below a certain size threshold try to merge with its "right-hand neighbor", i.e., the range that starts where the current range ends. Using our example above, this range's right-hand neighbor might be the range for customers whose IDs are between `[2000, 3000)`.
+
+If the combined size of the small range and its neighbor is less than the maximum range size, the ranges merge into a single range. In our example, this will create a new range of keys `[1000, 3000)`.
+
+{{site.data.alerts.callout_info}}
+When ranges merge, the left-hand-side (LHS) range consumes the right-hand-side (RHS) range.
+{{site.data.alerts.end}}
+
+#### Why range merges improve performance
+
+##### Query latency
+
+Queries in CockroachDB must contact a replica of each range involved in the query. This creates the following issues for clusters with many small ranges:
+
+- Queries incur a fixed overhead in terms of processing time for each range they must coordinate with.
+- Having many small ranges can increase the number of machines your query must coordinate with. This exposes your query to a greater likelihood of running into issues like network latency or overloaded nodes.
+
+By merging small ranges, CockroachDB can greatly reduce the number of ranges involved in queries, thus reducing query latency.
+
+##### Survivability
+
+CockroachDB automatically rebalances the distribution of ranges in your cluster whenever nodes come online or go offline.
+
+During rebalancing, it's better to replicate a few larger ranges across nodes than many smaller ranges. Replicating larger ranges requires less coordination and often completes more quickly.
+
+By merging smaller ranges together, your cluster needs to rebalance fewer total ranges. This ultimately improves your cluster's performance, especially in the face of availability events like node outages.
 
 ## Technical interactions with other layers
 
--- a/v21.1/range-merges.md
+++ b/v21.1/range-merges.md
@@ -2,57 +2,7 @@
 title: Range Merges
 summary: To help improve performance, CockroachDB can automatically merge small ranges of data together.
 toc: true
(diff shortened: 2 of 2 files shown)


aayushshah15 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10215#-:-MX3WW-KBAd_I9yUcqKO:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @rmloveland)

___
*[v21.1/architecture/distribution-layer.md, line 189 at r1](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX3W_3E0NcOkzkDU0Br:-MX3W_3E0NcOkzkDU0Bs:b-1z3yql) ([raw file](https://github.com/cockroachdb/docs/blob/92e5cbe4e2210623c94def518516c8b9b41869ea/v21.1/architecture/distribution-layer.md#L189)):*
> ```Markdown
> 
> A range will only attempt to merge with its right-hand neighbor. It will not attempt to merge with its left-hand neighbor (i.e. the previous range that ends where the current range starts).
> ```

Another way to phrase this is: "the left-hand side range consumes the right-hand side range". Feel free to add it if you think that makes it clearer (cause people could interpret "merges with its right-hand neighbor" differently).

rmloveland (commented):
TFTR Aayush!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10215#-:-MX3oQjaEn_XXTIWAcwZ:br6a4x1)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @aayushshah15)

___
*[v21.1/architecture/distribution-layer.md, line 189 at r1](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX3W_3E0NcOkzkDU0Br:-MX3oFs12WU3SwmHpGuZ:b-vgtozg) ([raw file](https://github.com/cockroachdb/docs/blob/92e5cbe4e2210623c94def518516c8b9b41869ea/v21.1/architecture/distribution-layer.md#L189)):*
<details><summary><i>Previously, aayushshah15 (Aayush Shah) wrote…</i></summary><blockquote>

Another way to phrase this is: "the left-hand side range consumes the right-hand side range". Feel free to add it if you think that makes it clearer (cause people could interpret "merges with its right-hand neighbor" differently).
</blockquote></details>

Thanks - I think that is much clearer.  Also all of the code I saw uses the LHS/RHS terminology too (as I'm sure you know well!).

Updated to the clearer (IMO) single sentence 

"When ranges merge, the left-hand-side (LHS) range consumes the right-hand-side (RHS) range."

lnhsingh (commented):
An annoying nit, but :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10215#-:-MX8WwEw0xCJY49f1q0r:b4zf0n7)* status: :shipit: complete! 1 of 0 LGTMs obtained (and 1 stale) (waiting on @aayushshah15 and @rmloveland)

___
*[v21.1/range-merges.md, line 5 at r2](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX8QP3hCR8c0rf_bW3x:-MX8QP3hCR8c0rf_bW3y:b3mhvff) ([raw file](https://github.com/cockroachdb/docs/blob/1df48e46f904298b9d6d0e036ef3c0705a4fb760/v21.1/range-merges.md#L5)):*
> ```Markdown
> summary: To help improve performance, CockroachDB can automatically merge small ranges of data together.
> toc: true
> redirect_to: v21.1/architecture/distribution-layer.html#range-merges
> ```

Is there a reason to keep this page around vs. add a `redirect_from: v21.1/range-merges.html` on the Distribution Layer doc?
___
*[v21.1/architecture/distribution-layer.md, line 184 at r2](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX8Un_7AXXHoVFKn-rD:-MX8Un_7AXXHoVFKn-rE:bfaytfe) ([raw file](https://github.com/cockroachdb/docs/blob/1df48e46f904298b9d6d0e036ef3c0705a4fb760/v21.1/architecture/distribution-layer.md#L184)):*
> ```
> i.e. 
> ```
nit: `i.e.,`

rmloveland (commented):
Thanks Lauren!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10215#-:-MXD4vpY0CQBmekOK3nF:b-jia9os)* status: :shipit: complete! 1 of 0 LGTMs obtained (and 1 stale) (waiting on @aayushshah15 and @lnhsingh)

___
*[v21.1/range-merges.md, line 5 at r2](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX8QP3hCR8c0rf_bW3x:-MXD42OHAEA3GivLoFcm:b-zaoxjy) ([raw file](https://github.com/cockroachdb/docs/blob/1df48e46f904298b9d6d0e036ef3c0705a4fb760/v21.1/range-merges.md#L5)):*

Is there a reason to keep this page around vs. add a `redirect_from: v21.1/range-merges.html` on the Distribution Layer doc?
</blockquote></details>

Unfortunately yes there is a reason! 

I tried using the `redirect_from` command you shared and  it does not seem to work.  Nor does `redirect_from: ../range-merges.html`. I think there is some weirdness around the fact that this doc is in the `architecture/` directory, and that Jekyll plugin does not expect a non-flat directory structure.  Since the `redirect_to` apparently works, I went with it.

Someday (tm) it would be good to unravel the special-casing of some docs being in nested directories, but ... not today :-) 
___
*[v21.1/architecture/distribution-layer.md, line 184 at r2](https://reviewable.io/reviews/cockroachdb/docs/10215#-MX8Un_7AXXHoVFKn-rD:-MXD4u3DBtmenw0iSiwo:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/1df48e46f904298b9d6d0e036ef3c0705a4fb760/v21.1/architecture/distribution-layer.md#L184)):*

> ```
> i.e. 
> ```
nit: `i.e.,`
</blockquote></details>

Fixed!

PR #10252, opened 2021-04-06
Cache remote include

ianjevans:
This PR enables caching `remote_include` content for local builds, and adds some new `Makefile` targets for generating output without caching and enabling profiling.


yingzhuchin:
Do we know if others on the team might want to use this, or would you be the only one? Maybe we should do a demo during docs weekly and take a poll? Asking because Rich mentioned he would likely not use a separate build command when it was discussed on the last ticket.

ianjevans:
I'm not sure if it matters that the build targets are used infrequently. The `nocache` target is more for troubleshooting and smoke-testing, and same thing with the `profile` target.

If it's likely that we're going to increase our use of `remote_include` in the future (and I think it is likely), caching will help speed up local builds while not affecting CI builds. But I also feel it's important to be able to turn off caching to make sure you're seeing locally what the CI builds would see.

yingzhuchin:
I see your point. A counterpoint would be that infrequently used targets could confuse contributors, and are less likely to be deprecated when they go out of date. Let's clarify with comments, merge but still demo to make sure everyone knows what these are? What do you think?

Makefile
```diff
@@ -53,6 +53,14 @@ cockroachdb: bootstrap
 .PHONY: standard
 standard: cockroachdb
 
+.PHONY: nocache
```
yingzhuchin, line 56: This is a little bit misleading because it implies the build won't use any caches, when really we're doing away with just the jekyll-include cache but still using other caches (as the `--incremental` flag indicates, plus also `jekyll-include-cache`). Can we rename to clarify?

_config_cockroachdb_no_cache.yml
```diff
@@ -0,0 +1,2 @@
+remote_include:
```
yingzhuchin, line 1: Same here -- can we rename the file to clarify it's just no Jekyll-remote cache?

Makefile
yingzhuchin: In addition to renaming we could also add a code comment to explain exactly what's not being cached.

yingzhuchin (commented):
Left comments. Otherwise LGTM!

Makefile
ianjevans: Done.

_config_cockroachdb_no_cache.yml
ianjevans: Done.

yingzhuchin (approved):
Looks good!

PR #10275, opened 2021-04-07
Update docs to note default value of crdb_region

### LINKED ISSUE #10200: Explain what happens to the crdb_region column when you migrate from REGIONAL BY TABLE to REGIONAL BY ROW
When you move a table from `REGIONAL BY TABLE` to `REGIONAL BY ROW` but have not yet set a new value for the `crdb_region` column, it is given the value of the primary region by default.

E.g.,

```sql
create table foo(i int);
insert into foo values (1), (2), (3);
alter table foo set locality regional by table in primary region;
alter table foo set locality regional by row;
select crdb_region from foo;
```

This needs to be explained somewhere.  Possibly in:

- [`ALTER TABLE ... SET LOCALITY`](https://www.cockroachlabs.com/docs/v21.1/set-locality.html)
- [Regional by row table high-level description]()?  (probably not, should probably add to the SQL reference doc above)
- somewhere else???

rmloveland:
Fixes #10200.

Explain that when you move a table from `REGIONAL BY TABLE` to `REGIONAL
BY ROW` but have not yet set a new value for the `crdb_region` column,
it is given the value of the primary region by default.

### DIFF
--- a/v21.1/set-locality.md
+++ b/v21.1/set-locality.md
@@ -98,6 +98,8 @@ To add a new row to a regional by row table, you must choose one of the followin
 
 This is necessary because every row in a regional by row table must have a home region.
 
+If you do not set a home region for a row in a regional by row table, it defaults to the value returned by the built-in function `gateway_region()`. If the value returned by `gateway_region()` does not belong to the multi-region database the table is a part of, the home region defaults to the database's primary region.
+
 For more information about how this table locality works, see [Regional by row tables](multiregion-overview.html#regional-by-row-tables).
 
 Note that you can use a name other than `crdb_region` for the hidden column by using the following statements:


v21.1/set-locality.md
```diff
@@ -96,7 +96,7 @@ To add a new row to a regional by row table, you must choose one of the followin
     INSERT INTO {table} (crdb_region, ...) VALUES ('us-east-1', ...);
     ~~~
 
-This is necessary because every row in a regional by row table must have a home region.
+This is necessary because every row in a regional by row table must have a home region.  Note that until you set a home region for a row in a regional by row table, it defaults to the value of the primary region.
```
otan, line 99: not entirely true; it defaults to the `gateway_region()`, which if the gateway_region does not belong to the multi-region database then defaults to the PRIMARY REGION (hence, `DEFAULT default_to_primary_region(gateway_region())`).

rmloveland: Hi Oliver!

Thanks for the correction.  I have updated it using approx. your words as follows.  PTAL

> Note that until you set a home region for a row in a regional by row table, it defaults to the value returned by the built-in function `gateway_region()`. If the value returned by `gateway_region()` does not belong to the multi-region database the table is a part of, the home region defaults to the database's primary region.

otan: I would replace `Note that until you set a home region for a row` with `If you do not set a home region for a row`, because for me "Note that until" implies for me implies you should change this later, but that's a nit and me reading too far into it...

otan (approved):

v21.1/set-locality.md
rmloveland: I like your suggested language better - more clear and direct!  Updated it to that.

Thanks again Oliver! Passing on to docs team reviewer

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10275#-:-MYH-sY_3KDOGiQA6WOg:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @otan)

Issue #9551, opened 2021-01-29
timeutil: allow certain timezone names to be case insensitive


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/57250

From release notes:
> Previously, timezones had to be entered in the same case as they were stored on the system. Now, timezone names can be case-insensitive provided they match well-known zone names according to [Go's `time/tzdata` package](https://golang.org/src/time/tzdata/tzdata.go). [#57250][#57250] {% comment %}doc{% endcomment %}

PR #10290, opened 2021-04-08
Updated 21.1 set time zone guidance

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/9551.


ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10290)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan)

___
*[v21.1/set-vars.md, line 185 at r1](https://reviewable.io/reviews/cockroachdb/docs/10290#-MXmc8zuSCmvOMFDK3EV-r1-185:-MY5zpNG4ksbLif0-TQo:beg4voa) ([raw file](https://github.com/cockroachdb/docs/blob/a0652c2315c1457a364810df9ccb45ca839f6ca9/v21.1/set-vars.md#L185)):*

might be worth giving an example of `east of the specified timezone`, i.e. say `-7` is equivalent to `GMT+7`.
</blockquote></details>

So confusing lol.

I updated the wording, with that additional example, but please re-review for accuracy.

ianjevans (approved):

Issue #9076, opened 2020-12-09
sql: allow string concat with any non-array type


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/55611

From release notes:
> Concatenation between a non-null argument and a null argument is now typed as string concatenation, whereas it was previously typed as array concatenation. This means that the result of `NULL || 1` will now be `NULL` instead of `{1}`. To preserve the old behavior, the null argument can be casted to an explicit type. [#55611][#55611] {% comment %}doc{% endcomment %}

PR #10294, opened 2021-04-08
Allow string concat with any non-array type; Update NULL concatenation docs

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/9076.

### DIFF
--- a/v21.1/null-handling.md
+++ b/v21.1/null-handling.md
@@ -526,6 +526,10 @@ failed to satisfy CHECK constraint (discount <= price)
 
 <span class="version-tag">New in v21.1</span> Concatenation between a non-`NULL` value and a `NULL` value results in a `NULL` value.
 
+{{site.data.alerts.callout_info}}
+In CockroachDB v20.2 and earlier, for all values other than [`STRING`](string.html), concatenation between a non-`NULL` value and a `NULL` value results in an [`ARRAY`](array.html) of the non-`NULL` value's type. To return an `ARRAY` of a specific type from a `NULL` concatenation in CockroachDB v21.1 and later, [cast](data-types.html#data-type-conversions-and-casts) the `NULL` value to an `ARRAY`.
+{{site.data.alerts.end}}
+
 For example:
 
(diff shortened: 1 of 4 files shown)


v21.1/null-handling.md
```diff
@@ -521,3 +521,33 @@ failed to satisfy CHECK constraint (price > 0)
 ~~~
 failed to satisfy CHECK constraint (discount <= price)
 ~~~
+
+## NULLs and concatenation with other types
+
+<span class="version-tag">New in v21.1</span> Concatenation between a non-`NULL` value and a `NULL` value results in a `NULL` value.
```
rafiss, line 527: I wonder if we should explicitly call out here that in v20.2 and earlier, that concatenation between a non-`NULL` value and a `NULL` value results in a `NULL` value would result in an array.

and that if you still need this behavior you can cast NULL to an array type.

rafiss (approved):
lgtm, but left a comment in case you think it's worth adding more detail

ericharmeling (commented):
TFTR @rafiss !

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10294#-:-MY67aMz4qABooJ4YD2s:b-6dtb4o)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rafiss)

___
*[v21.1/null-handling.md, line 527 at r1](https://reviewable.io/reviews/cockroachdb/docs/10294#-MXnCj943XEXKnJr10B_-r1-527:-MY67a49BY4GmXhVwUJK:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/da1716cc521d72e3d1ad8e261b6356c089596ad6/v21.1/null-handling.md#L527)):*

I wonder if we should explicitly call out here that in v20.2 and earlier, that concatenation between a non-`NULL` value and a `NULL` value results in a `NULL` value would result in an array.

and that if you still need this behavior you can cast NULL to an array type.
</blockquote></details>

Done.

ianjevans (approved):

PR #10302, opened 2021-04-13
Change /stable and /dev redirects from 200 to 301

### LINKED ISSUE #10236: Remove content duplication in /stable and /dev
To improve SEO results, we want to remove content duplication in our /stable and /dev folders, which are symlinks to specific numbered version folders. This could improve traffic to docs by 3x.

Approach: Remove the directory symlinks such that /stable and /dev folders no longer appear in the built _site. Instead rely on Netlify redirects to resolve /stable and /dev in URLs.

yingzhuchin:
Fixes #10236 (?)

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -10,7 +10,8 @@
 <meta name="robots" content="nocache">
 {% endif %}
 <title>{{ page.title }} | {{ site.site_title }}</title>
-<link rel="canonical" href="{{ page.canonical | absolute_url }}">
+{% assign page_url = page.url | relative_url | split: "/" %}
+<link rel="canonical" href="{{ site.versions["stable"] | absolute_url | append: "/" | append: page_url[3] }}">
 <link rel="shortcut icon" href="{{ 'images/favicon.png' | relative_url }}" type="image/png">
 
 {% if page.comparison == true %}<link rel="stylesheet" type="text/css" href="{{ 'css/select2.min.css' | relative_url }}">{% endif %}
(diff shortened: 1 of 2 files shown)


_includes/head.html
```diff
@@ -10,7 +10,7 @@
 <meta name="robots" content="nocache">
 {% endif %}
 <title>{{ page.title }} | {{ site.site_title }}</title>
-<link rel="canonical" href="{{ page.canonical | absolute_url }}">
```
jseldess, line 13: Don't we want the canonical url to alway point to the "stable" version of the docs? With this change, the canonical is always matching the current docs version...

yingzhuchin: I thought Dan said we would want to only reference the numbered versions. I might have misunderstood! I can change back.

jseldess: Yes, I think that's right, but only the canonical number versions, so the equivalent of /stable. Right now, that would make it would always be v20.2. That's my understanding at least. Good to check with Dan.

yingzhuchin: I see, you're saying we want all versions of the same page to now have canonical links to `v20.2`, which we then update to `v21.1` and so on. If we have time on tomorrow's call to clarify with Dan that would be great.

yingzhuchin: Done

jseldess:
This LGTM!. Let's review with Dan and Peter later today. Thank you, @yingzhuchin!

PR #10322, opened 2021-04-13
Partial inverted index support in 21.1.

### LINKED ISSUE #9087: sql: add support for partial inverted indexes
PR: https://github.com/cockroachdb/cockroach/pull/54376

From release notes:
> It is now possible to create partial inverted indexes. The optimizer will choose to scan partial inverted indexes when the partial index predicate is implied and scanning the inverted index has the lowest estimated cost. [#54376][#54376] {% comment %}doc{% endcomment %}

ianjevans:
Fixes #9087

### DIFF
--- a/v21.1/inverted-indexes.md
+++ b/v21.1/inverted-indexes.md
@@ -145,7 +145,7 @@ If you require comparisons using [`<`](functions-and-operators.html#supported-op
 
 ## Partial inverted indexes
 
-<span class="version-tag">New in v21.1:</span> You can create a [partial](partial-indexes.html) inverted index, an inverted index on a subset of `JSON`, `ARRAY`, or geospatial container column data. Just like partial indexes that use non-container datatypes, create a partial inverted index by including a clause that evaluates to true on a boolean predicate, like a `WHERE` clause.
+<span class="version-tag">New in v21.1:</span> You can create a [partial](partial-indexes.html) inverted index, an inverted index on a subset of `JSON`, `ARRAY`, or geospatial container column data. Just like partial indexes that use non-container data types, create a partial inverted index by including a clause that evaluates to true on a boolean predicate, like a `WHERE` clause.
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -323,7 +323,7 @@ SELECT * FROM users WHERE user_profile -> 'online' = 'true';
 Time: 2ms total (execution 2ms / network 0ms)
 ~~~
 
-Using index hinting with the partial inverted index.
+Now, use index hinting with the partial inverted index.
 
 ~~~ sql
 SELECT * FROM users@idx_online_users WHERE user_profile->'online' = 'true' AND user_profile->'location' = '"NYC"';
--- a/v21.1/inverted-indexes.md
+++ b/v21.1/inverted-indexes.md
@@ -158,7 +158,6 @@ CREATE TABLE test (
 
 ## Known limitations
 
-- CockroachDB does not support partitioning inverted indexes. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/43643).
 - CockroachDB does not support [index hinting](table-expressions.html#force-index-selection) for partial inverted indexes.
 
 ## Example
--- a/v21.1/inverted-indexes.md
+++ b/v21.1/inverted-indexes.md
@@ -143,9 +143,23 @@ If you require comparisons using [`<`](functions-and-operators.html#supported-op
 - "is contained by": [`<@`](functions-and-operators.html#supported-operations)
 - "contains": [`@>`](functions-and-operators.html#supported-operations)
 
+## Partial inverted indexes
+
+<span class="version-tag">New in v21.1:</span> You can create a [partial](partial-indexes.html) inverted index, an inverted index on a subset of `JSON`, `ARRAY`, or geospatial container column data. Just like partial indexes that use non-container datatypes, create a partial inverted index by including a clause that evaluates to true on a boolean predicate, like a `WHERE` clause.
+
+{% include copy-clipboard.html %}
+~~~ sql
+CREATE TABLE test (
+  id INT,
+  data JSONB,
+  INVERTED INDEX idx_data(data) WHERE id > 10
+);
+~~~
+
 ## Known limitations
 
-CockroachDB does not support partitioning inverted indexes. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/43643).
+- CockroachDB does not support partitioning inverted indexes. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/43643).
+- CockroachDB does not support [index hinting](table-expressions.html#force-index-selection) for partial inverted indexes.
 
 ## Example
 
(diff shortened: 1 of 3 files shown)


v21.1/inverted-indexes.md
```diff
@@ -143,9 +143,23 @@ If you require comparisons using [`<`](functions-and-operators.html#supported-op
...
+  id INT,
+  data JSONB,
+  INVERTED INDEX idx_data(data) WHERE id > 10
+);
+~~~
+
 ## Known limitations
 
-CockroachDB does not support partitioning inverted indexes. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/43643).
+- CockroachDB does not support partitioning inverted indexes. For details, see [tracking issue](https://github.com/cockroachdb/cockroach/issues/43643).
```
mgartner, line 161: This is supported in 21.1. The inverted index has to be multi-column so that there is a non-container column to partition by.

v21.1/inverted-indexes.md
```diff
@@ -143,9 +143,23 @@ If you require comparisons using [`<`](functions-and-operators.html#supported-op
...
+- CockroachDB does not support [index hinting](table-expressions.html#force-index-selection) for partial inverted indexes.
```
mgartner, line 162: This is supported too! Did you run into a case where it didn't work as expected?

mgartner (approved):
Everything looks great! Just left comments on the limitations mentioned, which shouldn't be limitations. If you ran into problems when trying them, there might be a bug, so let me know!

v21.1/inverted-indexes.md
ianjevans: Hmm, this is on the latest `release-21.1`:

```
CREATE TABLE users (
    profile_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    last_updated TIMESTAMP DEFAULT now(),
    user_profile JSONB,
    INVERTED INDEX user_details (user_profile)
  );
INSERT INTO users (user_profile) VALUES
    ('{"first_name": "Lola", "last_name": "Dog", "location": "NYC", "online" : true, "friends" : 547}'),
    ('{"first_name": "Ernie", "status": "Looking for treats", "location" : "Brooklyn"}'),
    ('{"first_name": "Carl", "last_name": "Kimball", "location": "NYC", "breed": "Boston Terrier"}'
  );
CREATE INVERTED INDEX idx_online_users ON users(user_profile) WHERE user_profile -> 'online' = 'true';
```

Then try to query the partial inverted index:

```
SELECT * FROM users@idx_online_users;
```

which gives:

```
ERROR: index "idx_online_users" is a partial inverted index and cannot be used for this query
```

And:

```
SELECT * FROM users@{FORCE_INDEX=idx_online_users};
```

which also gives:

```
ERROR: index "idx_online_users" is a partial inverted index and cannot be used for this query
```

Am I doing something wrong?

mgartner: Your `SELECT` queries don't contain a `WHERE` clause that implies the partial index predicate. You're querying for all the rows in the table, but `idx_online_users` doesn't have an entry for every row in the table, so it can't answer your query.

This behavior is the same for non-inverted partial indexes too:

```sql
CREATE TABLE t (a INT, INDEX idx (a) WHERE a > 0);
CREATE TABLE

SELECT * FROM t@idx;
ERROR: index "idx" is a partial index that does not contain all the rows needed to execute this query

SELECT * FROM t@idx WHERE a > 0;
  a
-----
(0 rows)
```

Also, we don't support forcing an unconstrained scan on an inverted index, partial or not:

```sql
CREATE TABLE t (j JSON, INVERTED INDEX idx (j));
CREATE TABLE

SELECT * FROM t@idx;
ERROR: index "idx" is inverted and cannot be used for this query

SELECT * FROM t@idx WHERE j @> '1';
  j
-----
(0 rows)
```

The scan would be unconstrained because there's no `WHERE` clause that can be used to constrain the index scan. In theory, we could support this, but in practice it doesn't make much sense - an inverted index can have many more entries than rows, so if you're querying all rows then scanning a non-inverted index, like the primary index, is always better than scanning an inverted index.

So in order to force a **partial inverted index** you'd need to include a `WHERE` filter that (1) implies the partial index predicate and (2) constrains the inverted index scan. If both are not satisfied, you'll get an error.

A final demonstration:

```sql
CREATE TABLE t (j JSON, INVERTED INDEX idx (j) WHERE j->'a' = '1');
CREATE TABLE

-- This WHERE filter satisfies (1), but not (2).
SELECT * FROM t@idx WHERE j->'a' = '1';
ERROR: index "idx" is a partial inverted index and cannot be used for this query

-- This WHERE filter satisfies (2), but not (1).
SELECT * FROM t@idx WHERE j->'b' = '2';
ERROR: index "idx" is a partial inverted index and cannot be used for this query

-- This WHERE filter satisfies (1) and (2).
SELECT * FROM t@idx WHERE j->'a' = '1' AND j->'b' = '2';
  j
-----
(0 rows)
```

**TLDR;** If you update your query to something like below, forcing the index should work:

```sql
SELECT * FROM users@idx_online_users WHERE user_profile->'online' = 'true' AND user_profile->'location' = '"NYC"'
```

ianjevans: Ah, ok. I think I get it now.

ianjevans: Removed the bogus limitation, and added examples for both partial and partial inverted indexes to the index forcing topic.

ericharmeling (approved):
Nice! I just had two minor minor nits.

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10322#-:-MYLDm4u5TnUOuqUztMK:b-mgw6g7)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187, @ianjevans, @kevin-v-ngo, and @mgartner)

___
*[v21.1/inverted-indexes.md, line 148 at r3](https://reviewable.io/reviews/cockroachdb/docs/10322#-MYLDOQ6D5aO0SkmPi3D:-MYLDOQ6D5aO0SkmPi3E:b7yo4pe) ([raw file](https://github.com/cockroachdb/docs/blob/8be065cd06d610dc6a5fe32ed2cf2c93ccdf2338/v21.1/inverted-indexes.md#L148)):*
nit:

> ```
> datatypes
> ```

data types
___
*[v21.1/inverted-indexes.md, line 326 at r3](https://reviewable.io/reviews/cockroachdb/docs/10322#-MYLDcLS-E61sEN7sPNz:-MYLDcLS-E61sEN7sPO-:b-frx0n5) ([raw file](https://github.com/cockroachdb/docs/blob/8be065cd06d610dc6a5fe32ed2cf2c93ccdf2338/v21.1/inverted-indexes.md#L326)):*
nit:

> ```
> Using index hinting with the partial inverted index.
> ```

Now, use index hinting with the partial inverted index.

mgartner (approved):
LGTM

kevin-v-ngo (commented):
Looks good to me!

kevin-v-ngo (approved):

awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10322)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @ianjevans, and @mgartner)

___
*[v21.1/inverted-indexes.md, line 303 at r4](https://reviewable.io/reviews/cockroachdb/docs/10322#-MYeLE-ZCikx9AtG97rv:-MYeLE-ZCikx9AtG97rw:bqdz8zs) ([raw file](https://github.com/cockroachdb/docs/blob/77b4a9bdac1a1d2b780783fbfc82a275f6a3aaa4/v21.1/inverted-indexes.md#L303)):*
> ```Markdown
> ~~~
> 
> ### Create a table with a partial inverted index on a JSONB column
> ```

Should there be a discussion here or elsewhere about when to use a partial inverted index on a jsonb column vs. using a computed column/virtual compute column and indexing that column with a standard index?

Issue #9163, opened 2020-12-10
release-20.2: sql: prohibit ALTER TABLE SET DATA TYPE requiring conversion/validation


lnhsingh:
PR: https://github.com/cockroachdb/cockroach/pull/56629

From release notes:
> [`ALTER TABLE ... SET DATA TYPE ...`](../v20.2/alter-column.html) is no longer available for operations that involve conversion (e.g. [`STRING`](../v20.2/string.html) to [`TIMESTAMPTZ`](../v20.2/timestamp.html)) or precision/width truncation (e.g. [`INT(4)`](../v20.2/int.html) -> `INT(2)`) due to a bug in validation. These features are already gated by the [session variable](../v20.2/set-vars.html) `enable_experimental_alter_column_type_general` - setting this session variable is now a no-op. [#56629][#56629] {% comment %}doc{% endcomment %}

PR #10329, opened 2021-04-14
prohibit ALTER TABLE SET DATA TYPE requiring conversion/validation

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/9163.

IIUC https://github.com/cockroachdb/cockroach/pull/56629 just reverts `SET TYPE` to 20.1 behavior, correct?


otan (approved):
yes you are right

ianjevans (approved):

PR #10332, opened 2021-04-14
Rebalancing will happen when primary region is set

### LINKED ISSUE #10328: Make sure "PRIMARY REGION" docs are clear on what happens to replica and leaseholder placement when you do this
We must capture the following behavior explicitly in the docs to avoid surprising users (right now it is implicit).

Given a cluster with multiple regions, which includes a database for which no primary region has been set:

- Before a primary region is added to the database, its replicas will be spread as broadly as possible for resiliency
- When a primary region is added to the database:
  - All tables will be REGIONAL BY TABLE (the default - this is mentioned in the [Multi-region overview](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview))
  - As such, all tables will have all of their voting replicas and leaseholders moved to the primary region

Again, this is kind of "already there" by implication (how else would we be "optimizing access" to that data except by shuffling things around?), but it might startle users who see their cluster rebalancing and wonder why.

Estimated scope of work:

- Update PRIMARY REGION doc to say this explicitly
- Maybe update the overview page?  I kind of hate to get into low-level detail about ranges n stuff in an introduction, but maybe
- Something else?

rmloveland:
Fixes #10328.

### DIFF
--- a/v21.1/add-region.md
+++ b/v21.1/add-region.md
@@ -46,7 +46,12 @@ ALTER DATABASE foo PRIMARY REGION "us-east1";
 ALTER DATABASE PRIMARY REGION
 ~~~
 
-For more information, see [Database regions](multiregion-overview.html#database-regions).
+Given a cluster with multiple regions, any databases in that cluster that have not yet had their primary regions set will have their replicas spread as broadly as possible for resiliency. When a primary region is added to one of these databases:
+
+- All tables will be [`REGIONAL BY TABLE`](set-locality.html#regional-by-table) in the primary region by default.
+- This means that all such tables will have all of their voting replicas and leaseholders moved to the primary region. This process is known as [rebalancing](architecture/replication-layer.html#leaseholder-rebalancing).
+
+For more information about cluster regions and database regions, see [Cluster regions](multiregion-overview.html#cluster-regions) and [Database regions](multiregion-overview.html#database-regions).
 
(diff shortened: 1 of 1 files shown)


rmloveland:
Adam let me know if you'd rather I ask someone else to review this, pinged you b/c you mentioned it but happy to ask elsewhere

v21.1/add-region.md
```diff
@@ -46,7 +46,12 @@ ALTER DATABASE foo PRIMARY REGION "us-east1";
 ALTER DATABASE PRIMARY REGION
 ~~~
 
-For more information, see [Database regions](multiregion-overview.html#database-regions).
+Given a cluster with multiple regions, any databases in that cluster that have not yet had their primary regions set will have their replicas spread as broadly as possible for resiliency. When a primary region is added to one of these databases:
+
+- All tables will be [`REGIONAL BY TABLE`](set-locality.html#regional-by-table) by default.
```
ajstorm, line 51: We might want to say here "REGIONAL BY TABLE IN PRIMARY REGION" or REGIONAL BY TABLE in the primary region.

ajstorm (approved):
Other than the one comment, LGTM.

v21.1/add-region.md
rmloveland: Updated - thanks for the review!

PR #10352, opened 2021-04-19
Use /stable in CC as relative_url links

### LINKED ISSUE #9701: Update versioned links to use /stable/ in CC
When cross-referencing CRDB docs, CockroachCloud docs should use the `/stable/` versions of links instead of `/vX.X/`. This way, links do not have to be manually updated whenever the `stable` version is updated.

Note: Some time after the v20.2 release, the `/stable/` alias seemed to stop working in TeamCity. _Existing_ links using `/stable/` still seem to work, but any _new_ links throw an error in the build log.

yingzhuchin:
Fixes #9701


lnhsingh (approved):
Nice! LGTM

Issue #10069, opened 2021-03-23
pgwire: set options based on "options" URL parameter


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/59621

From release notes:
> CockroachDB now recognizes the `options` URL parameter. The `options` parameter specifies session variables to set at connection start. This is treated the same as defined in the [PostgreSQL docs](https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-PARAMKEYWORDS). [#59621][#59621] {% comment %}doc{% endcomment %}

PR #10354, opened 2021-04-19
options=-c support in connection URLs

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/10069.

### DIFF
--- a/v21.1/connection-parameters.md
+++ b/v21.1/connection-parameters.md
@@ -100,7 +100,7 @@ Parameter | Description
 `-c <session_variable>=<value>` | <span class="version-tag">New in v21.1:</span> Sets a [session variable](set-vars.html) for the SQL session.
 
 {{site.data.alerts.callout_info}}
-Note that some drivers require space characters to be properly encoded in URL connection strings. For example, spaces in [a JDBC connection string](https://jdbc.postgresql.org/documentation/head/connect.html#connection-parameters) must specified as `%20`.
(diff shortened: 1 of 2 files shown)


v21.1/connection-parameters.md
```diff
@@ -88,7 +88,20 @@ Parameter | Description | Default value
...
+
+CockroachDB supports the following `options` parameters. After the first `options` parameter is specified, any additional parameters in the same connection string must be separated by a space.
+
+Parameter | Description
+----------|-------------
+`--cluster=<cluster name>` | Specifies the cluster name when connecting to [CockroachCloud free-tier clusters](connect-to-the-database-cockroachcloud.html#connect).
+`-c <session_variable>=<value>` | <span class="version-tag">New in v21.1:</span> Sets a [session variable](set-vars.html) for the SQL session.
+
+{{site.data.alerts.callout_info}}
+Note that some drivers require space characters to be properly encoded in URL connection strings. For example, spaces in [a JDBC connection string](https://jdbc.postgresql.org/documentation/head/connect.html#connection-parameters) must specified as `%20`.
```
ianjevans, line 103: Isn't it more than just spaces? E.g. `options=--cluster=<cluster name>` would require the second equals sign to be URL encoded.

ianjevans (approved):
LGTM, but we might need to clarify about URL encoding all parameters.

v21.1/connection-parameters.md
ericharmeling: You're right - it could be more than just spaces. I've updated the description to use spaces just as an example.

IDK about equal signs though... At least in the case of the JDBC driver (https://jdbc.postgresql.org/documentation/head/connect.html#connection-parameters), the equal signs don't need to be URL-encoded. Only spaces do.

PR #10379, opened 2021-04-20
Update create-security-certificates-openssl.md

dbist:
tested this on a RHEL based machine and `CN=` now has a blank before equal.


rmloveland (approved):
LGTM

(PS also how DARE THEY change an output format we were grepping :-) )

rmloveland:
Also, thank you @dbist !!!

PR #10388, opened 2021-04-21
Revert "Updated SHOW ZONE CONFIGURATIONS diagram, parameters, and example output"

ericharmeling:
Reverts cockroachdb/docs#10366


ericharmeling:
It looks like https://github.com/cockroachdb/docs/pull/10387 might need to merge before this will build properly?

PR #10390, opened 2021-04-21
Replace explicit /stable links with named version

### LINKED ISSUE #9701: Update versioned links to use /stable/ in CC

jseldess:
Using a Liquid snippet

This PR also updates a lot of CC links that were referencing
outdated versions of CRDB docs (e.g., v20.1).

Fixes #9701


lnhsingh (approved):

yingzhuchin:
This should also take care of #9701 

Thanks Jesse!

PR #10393, opened 2021-04-21
Document delete behavior

HonoreDB:
This isn't new behavior, but we had some questions on it recently and did some investigation, and it seems worth calling out.

### DIFF
--- a/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
+++ b/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
@@ -96,6 +96,10 @@ The `kv.closed_timestamp.target_duration` [cluster setting](cluster-settings.htm
     Resolved timestamp notifications on every Kafka partition can be used to provide strong ordering and global consistency guarantees by buffering records in between timestamp closures. Use the "resolved" timestamp to see every row that changed at a certain time.
 
     The complexity with timestamps is necessary because CockroachDB supports transactions that can affect any part of the cluster, and it is not possible to horizontally divide the transaction log into independent changefeeds. For more information about this, [read our blog post on CDC](https://www.cockroachlabs.com/blog/change-data-capture/).
+    
+## Delete messages
+
+Deleting a row will result in a changefeed message giving the primary key of the deleted row and a null value. If an attempt is made to delete a row that never existed, changefeed behavior is undefined to allow for optimizations at the storage layer: you may or may not get a delete message. Similarly, creating and deleting a row within the same transaction will never result in an insert message, but may result in a delete message.
 
 ## Avro schema changes
 
--- a/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
+++ b/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
@@ -99,7 +99,13 @@ The `kv.closed_timestamp.target_duration` [cluster setting](cluster-settings.htm
     
 ## Delete messages
 
-Deleting a row will result in a changefeed message giving the primary key of the deleted row and a null value. If an attempt is made to delete a row that never existed, changefeed behavior is undefined to allow for optimizations at the storage layer: you may or may not get a delete message. Similarly, creating and deleting a row within the same transaction will never result in an insert message, but may result in a delete message.
+Deleting a row will result in a changefeed message giving the primary key of the deleted row and a null value. With default options, deleting the row with primary key `5` will look like
+
+~~~ shell
+[5] {"after": null}
+~~~
+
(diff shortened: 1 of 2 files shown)


v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
```diff
@@ -96,6 +96,10 @@ The `kv.closed_timestamp.target_duration` [cluster setting](cluster-settings.htm
     Resolved timestamp notifications on every Kafka partition can be used to provide strong ordering and global consistency guarantees by buffering records in between timestamp closures. Use the "resolved" timestamp to see every row that changed at a certain time.
 
     The complexity with timestamps is necessary because CockroachDB supports transactions that can affect any part of the cluster, and it is not possible to horizontally divide the transaction log into independent changefeeds. For more information about this, [read our blog post on CDC](https://www.cockroachlabs.com/blog/change-data-capture/).
+    
+## Delete messages
+
+Deleting a row will result in a changefeed message giving the primary key of the deleted row and a null value. If an attempt is made to delete a row that never existed, changefeed behavior is undefined to allow for optimizations at the storage layer: you may or may not get a delete message. Similarly, creating and deleting a row within the same transaction will never result in an insert message, but may result in a delete message.
```
amruss, line 102: Maybe an example of what a delete message looks like, something like {"id": 1, "name": null}

I would also just add a sentence that makes it clear what the impact of the undefined behavior. Something like - "In some situations you may get a delete message for an ID that never existed" then the rest - "If an attempt is made to delete a row that never existed, changefeed behavior is undefined to allow for optimizations at the storage layer: you may or may not get a delete message."

For the last part- I would add an explanation of what guarentees we have around transaction, something like -  "We only propagate the last change to a row in a transaction through a changefeed", then "Creating and deleting a row within the same transaction will never result in an insert message, but may result in a delete message."

jseldess:
Thanks, @HonoreDB and @amruss! Should we have @kathancox take this over if more iteration is needed? cc @lnhsingh

v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
HonoreDB: Added, thanks!

jseldess:
@kathancox, can you take over this PR? I think you may want to do just a bit of copyediting and then figure out if we need to apply the change to 20.2 docs as well.

kathancox:
@HonoreDB Could you take a look over the small edits and let me know if you have any questions? The changes were "giving" and "look like" to align with "output" used throughout the page. And, "For example", to make it clear that the "situations" you then mention are examples of this. Thanks so much!

lnhsingh (commented):
A couple of small edit suggestions (which apply to both versions of the doc). Once edited / sign off received from Aaron, feel free to merge!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10393#-:-MZOCWbvD2BsxJWm9QEy:ba0qmsl)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @HonoreDB, and @kathancox)

___
*[v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md, line 102 at r3](https://reviewable.io/reviews/cockroachdb/docs/10393#-MZOBilRD9bST_UFY2PA:-MZOBilS-wzouPJsFjYx:b-n3c4f6) ([raw file](https://github.com/cockroachdb/docs/blob/eedd6cf63475fbaa89da3f72a55b92b4276fbc31/v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md#L102)):*
> ```Markdown
> ## Delete messages
> 
> Deleting a row will result in a changefeed outputting the primary key of the deleted row and a null value. With default options, deleting the row with primary key `5` will output:
> ```

I think the second sentence should start with "For example, with default options, deleting the row..."
___
*[v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md, line 108 at r3](https://reviewable.io/reviews/cockroachdb/docs/10393#-MZOC6_6-NSdOavTgHEb:-MZOC6_6-NSdOavTgHEc:b-bt6klg) ([raw file](https://github.com/cockroachdb/docs/blob/eedd6cf63475fbaa89da3f72a55b92b4276fbc31/v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md#L108)):*
> ```Markdown
> ~~~
> 
> In some unusual situations you may receive a delete message for a row without first seeing an insert message. For example, if an attempt is made to delete a row that does not exist; changefeed behavior is undefined to allow for optimizations at the storage layer, therefore, you may or may not get a delete message. Similarly, if there are multiple writes to a row within a single transaction, only the last one will propagate to a changefeed. This means that creating and deleting a row within the same transaction will never result in an insert message, but may result in a delete message.
> ```

The second sentence in this paragraph is confusing with the semicolon. Could this sentence structure be simplified? Maybe something like "For example, if an attempt is made to delete a row that does not exist, you may or may not get a delete message because the changefeed behavior is undefined to allow for optimizations at the storage layer."

HonoreDB:
Thanks, looks good to me!

PR #10410, opened 2021-04-22
use stable links

taroface:


jseldess:
@taroface, is this PR still valid? If so, we aren't using `/stable/` in internal links anymore. Please see the syntax that YZ used in this PR: https://github.com/cockroachdb/docs/pull/10557/files.

jseldess:
Basically, `../{{site.versions["stable"]}}/sql-statements.html` instead of `../stable/sql-statements.html`.

Issue #10243, opened 2021-04-06
Removing text from compare page


meagan-g:
The subheadline of this page: https://www.cockroachlabs.com/docs/stable/cockroachdb-in-comparison.html# indicates you can click on feature but you can't.

It would be great if we could remove the text that says: `and click CockroachDB answers to view related documentation.`

olgaromashko:
@jseldess can i work on this issue? :)

jseldess:
@olgaromashko, sure. Please be sure to update the page in both the v20.2 and v21.1 directories.

olgaromashko:
@jseldess I opened a PR, if anything needs to be edited, let me know :)

PR #10445, opened 2021-04-28
removed text that says "and click CockroachDB answers to view related documentation." in v20.2 & v21.1

olgaromashko:
@jseldess issue is done! if anything needs to be edited, let me know :) 
issue: #10243


jseldess (approved):
LGTM. Thanks, @olgaromashko!

PR #10450, opened 2021-04-28
Correct canonical url for non-versioned docs pages

jseldess:
On current site, non-versioned docs pages have incorrect canonical url resulting in 404s.


_includes/head.html
```diff
@@ -11,7 +11,11 @@
 {% endif %}
 <title>{{ page.title }} | {{ site.site_title }}</title>
 {% assign page_url = page.url | relative_url | split: "/" %}
+{% if page.version %}
 <link rel="canonical" href="{{ site.versions["stable"] | absolute_url | append: "/" | append: page_url[3] }}">
```
yzdocs, line 15: Just a nit but could we perhaps indent within the conditional tags? Like on line 34. I know it's not consistently done even within this file so no worries if we'd rather do as part of a larger clean-up in another PR.

yzdocs (approved):
LGTM otherwise!

_includes/head.html
jseldess: Of course. I've indented within those tags now, but please free free to clean up the file further. I'm not sure how we want to handle empty lines, etc.

PR #10453, opened 2021-04-28
draft OpenShift deployment doc

taroface:
Draft OpenShift deployment doc using the source: https://github.com/cockroachdb/cockroach-operator/blob/master/docs/openshift/openshift-dev.md


v21.1/deploy-cockroachdb-with-kubernetes-openshift.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Step 3. Start CockroachDB
+
+1. When the Operator is ready, click **View Operator** to navigate to the **Installed Operators** page.
+
+1. In the **CockroachDB Operator** tile, click **Create instance**.
+
+	<img src="{{ 'images/v21.1/cockroachdb-operator-instance-openshift.png' | relative_url }}" alt="OpenShift OperatorHub" style="border:1px solid #eee;max-width:100%" />
+
+1. Make sure **CockroachDB Version** is set to an Operator-supported version. For a list of compatible image names, see the [Operator deployment definition](https://github.com/cockroachdb/cockroach-operator/blob/master/manifests/operator.yaml) on GitHub.
```
keith-mcclellan, line 91: This is awkward, and realistically we want to direct people to use the cockroachDBVersion api and not the image.name anyway on Red Hat.  We should instead maybe say something like `Please make sure that cockroachDBVersion is set to a valid version of CockroachDB, like v21.1.0 or v20.2.8.  This will use the official CockroachDB images that are certified by Red Hat.  Alternatively, customers can also set spec.image.name to pull a CockroachDB image from a custom location.

keith-mcclellan: fwiw the cockroachDBVersion api was built after I wrote the original readme so I should have clarified this with you earlier - sorry about that!

v21.1/deploy-cockroachdb-with-kubernetes-openshift.md
```diff
@@ -0,0 +1,329 @@
...
+	~~~
+
+## Step 4. Create a secure client pod
+
+To use the CockroachDB SQL client, first launch a secure pod running the `cockroach` binary.
+
+This can be defined with the following YAML, which mounts the Operator's generated certificates:
+
+{{site.data.alerts.callout_success}}
+`spec.containers.image` should refer to the CockroachDB version you specified when [starting the nodes](#step-3-start-cockroachdb).
```
keith-mcclellan, line 121: We should probably change this direction slightly to clarify this - in some cases OpenShift shows the image SHA instead of the image tag, and the customer will need to use the SHA.  Again, this is something we didn't know when we wrote up the README

v21.1/deploy-cockroachdb-with-kubernetes-openshift.md
```diff
@@ -0,0 +1,329 @@
...
+oc port-forward service/crdb-tls-example-public 8080
+~~~
+
+~~~
+Forwarding from [::1]:8080 -> 8080
+~~~
+
+1. Go to <a href="https://localhost:8080/" data-proofer-ignore>https://localhost:8080</a> and log in with the username and password you created earlier.
+
+    {% include {{ page.version.version }}/misc/chrome-localhost.md %}
```
keith-mcclellan, line 288: Could we add alternative instructions on how to download the ca.crt and make it trusted instead?

keith-mcclellan (changes requested):
In general this looks great. There are a couple of things that carried forward from the README that we probably should fix that we could have caught earlier.  Other than that, just little nits.

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10453)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @johnrk, @keith-mcclellan, and @rail)

___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 91 at r1](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r1-91:-MZTgYQp1j8udIG2u50U:b1o4py9) ([raw file](https://github.com/cockroachdb/docs/blob/d1f48f75f57b692a4be042554681737373b9151b/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L91)):*
<details><summary><i>Previously, keith-mcclellan (Keith) wrote…</i></summary><blockquote>

fwiw the cockroachDBVersion api was built after I wrote the original readme so I should have clarified this with you earlier - sorry about that!
</blockquote></details>

I'd prefer not to call out specific version numbers, since these will quickly be superseded by newer versions.

@rail suggested generating a separate YAML of just the supported versions, which seems like a good solution (thread: https://cockroachlabs.slack.com/archives/C0165QKD78F/p1619716393242200?thread_ts=1619461588.173000&cid=C0165QKD78F). I know it's not ideal, but I think it's best to use the current link until we have that YAML. This section is just assuming that someone wants to quickly deploy using the OpenShift UI fields.

Also, I will cover the `cockroachDBVersion` field in the broader guide on the CRDs. I can also link relevant sections from this deployment guide  to that page when it's ready.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 121 at r1](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r1-121:-MZTM1To-HeOpUFBD7Db:b9yazmk) ([raw file](https://github.com/cockroachdb/docs/blob/d1f48f75f57b692a4be042554681737373b9151b/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L121)):*
<details><summary><i>Previously, keith-mcclellan (Keith) wrote…</i></summary><blockquote>

We should probably change this direction slightly to clarify this - in some cases OpenShift shows the image SHA instead of the image tag, and the customer will need to use the SHA.  Again, this is something we didn't know when we wrote up the README
</blockquote></details>

Where would OpenShift display the image SHA? I should probably clarify that as well.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 288 at r1](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r1-288:-MZTOZZE2lgKiJ5xNDqW:b-729z82) ([raw file](https://github.com/cockroachdb/docs/blob/d1f48f75f57b692a4be042554681737373b9151b/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L288)):*
<details><summary><i>Previously, keith-mcclellan (Keith) wrote…</i></summary><blockquote>

Could we add alternative instructions on how to download the ca.crt and make it trusted instead?
</blockquote></details>

Can you help me understand how to do this? Does it involve exec-ing into one of the pods with the ca.crt as an argument?

keith-mcclellan:
1) your call there, as it stands its a bit confusing is all. I agree that Rail's idea has merit.
2) It will show it listed as the image.name in one of the cockroachdb pods, customer will see something like `{some-repo}/cockroachdb/cockroach:SADKH123123241!@#!@#413aDSAS` instead of a clean tag. They need to use that string if they're running in an offline mode and can't pull from the red hat registry directly.
3) Thats about right.  We can backlog this if you want and I can walk you through at least one way to do it when I get back from my vacation.  There's nothing wrong with whats there now.

v21.1/deploy-cockroachdb-with-kubernetes-openshift.md
```diff
@@ -0,0 +1,330 @@
...
+1. On secure clusters, [certain pages of the DB Console](ui-overview.html#db-console-access) can only be accessed by `admin` users.
+
+	Start the CockroachDB [built-in SQL client](cockroach-sql.html) from the client pod:
+
+	{% include copy-clipboard.html %}
+	~~~ shell
+	oc exec -it crdb-client-secure -- ./cockroach sql --certs-dir=/cockroach/cockroach-certs/ --host=crdb-tls-example-public
+	~~~
+
+1.  Assign `roach` to the `admin` role (you only need to do this once):
```
mlewis1, line 262: Should there be a step before this where we create the user `roach` and set a password? Then this line would make sense in making reference to the username, and the later line would also make sense:

> log in with the username and password you created earlier.

Looks like we have the same issue here https://www.cockroachlabs.com/docs/v20.2/orchestrate-a-local-cluster-with-kubernetes.html#step-4-access-the-db-console probably based on the way the example was ported over from here https://www.cockroachlabs.com/docs/v20.2/secure-a-cluster.html#step-3-use-the-built-in-sql-client where we *do* have a step to `CREATE USER max WITH PASSWORD 'roach';`

Although in this case, we're using `roach` as the user, not the password.

An aside, I've never liked using `max` as a user name because it also has the meaning 'maximum' and could be confusing to see in a code block.

mlewis1 (changes requested):
Left some very small suggestions and one comment. Otherwise LGTM.

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10453)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @johnrk, @keith-mcclellan, @mikeCRL, @mlewis1, and @rail)

___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 61 at r2](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r2-61:-M_MX0BZ-fpicW-mjkBx:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e8d6667e35de33615a774a03783fc3b006632220/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L61)):*
<details><summary><i>Previously, mlewis1 wrote…</i></summary><blockquote>

```suggestion
1. Navigate to your OpenShift web console and click **OperatorHub**.
```
</blockquote></details>

Done.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 63 at r2](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r2-63:-M_MX3I08dKMsszZal5A:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e8d6667e35de33615a774a03783fc3b006632220/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L63)):*
<details><summary><i>Previously, mlewis1 wrote…</i></summary><blockquote>

```suggestion
1. Enter "cockroach" in the search box. There are two tiles called **CockroachDB Operator**. Find the tile _without_ the `Marketplace` label (which requires a subscription).
```
</blockquote></details>

Done.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 71 at r2](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r2-71:-M_MX50P1Tg_5lLq8y9Y:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e8d6667e35de33615a774a03783fc3b006632220/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L71)):*
<details><summary><i>Previously, mlewis1 wrote…</i></summary><blockquote>

```suggestion
1. Confirm that the Operator is running:
```
</blockquote></details>

Done.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 121 at r2](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r2-121:-M_MXgP7DU-pVnst1oNd:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e8d6667e35de33615a774a03783fc3b006632220/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L121)):*
<details><summary><i>Previously, mlewis1 wrote…</i></summary><blockquote>

```suggestion
`spec.containers.image` should match the **Image** value that is displayed on the **Pods** page under **Containers** when you select a CockroachDB pod. OpenShift may display the image SHA instead of the tag. In this case, you should use the SHA for `spec.containers.image`.
```
</blockquote></details>

Done.
___
*[v21.1/deploy-cockroachdb-with-kubernetes-openshift.md, line 262 at r2](https://reviewable.io/reviews/cockroachdb/docs/10453#-MZPtxd3cO33Cf2G7Dp9-r2-262:-M_MZCjPEmwdjP6HUeZZ:beyytkj) ([raw file](https://github.com/cockroachdb/docs/blob/e8d6667e35de33615a774a03783fc3b006632220/v21.1/deploy-cockroachdb-with-kubernetes-openshift.md#L262)):*
<details><summary><i>Previously, mlewis1 wrote…</i></summary><blockquote>

Should there be a step before this where we create the user `roach` and set a password? Then this line would make sense in making reference to the username, and the later line would also make sense:

> log in with the username and password you created earlier.

Looks like we have the same issue here https://www.cockroachlabs.com/docs/v20.2/orchestrate-a-local-cluster-with-kubernetes.html#step-4-access-the-db-console probably based on the way the example was ported over from here https://www.cockroachlabs.com/docs/v20.2/secure-a-cluster.html#step-3-use-the-built-in-sql-client where we *do* have a step to `CREATE USER max WITH PASSWORD 'roach';`

Although in this case, we're using `roach` as the user, not the password.

An aside, I've never liked using `max` as a user name because it also has the meaning 'maximum' and could be confusing to see in a code block.
</blockquote></details>

Yep, user `roach` is created in the `kubernetes-basic-sql.md` include, which was a little buried here.

Good point about `max` btw.

PR #10464, opened 2021-04-29
Added note to confirm RESTORE does not support newer version backups

### LINKED ISSUE #10374: release-21.1: backupccl: prevent restoring a newer backup
PR: https://github.com/cockroachdb/cockroach/pull/62398

From release notes:
> `RESTORE` now can't [restore](../v21.1/take-full-and-incremental-backups.html) `BACKUP`s made by newer versions of CockroachDB. [#62398][#62398] {% comment %}doc{% endcomment %}

kathancox:
Closes #10374 

Added note to `RESTORE` page to confirm that it doesn't support restoring backups made by newer versions.

### DIFF
--- a/v21.1/restore.md
+++ b/v21.1/restore.md
@@ -19,6 +19,10 @@ You can restore:
 - [Tables](#tables)
 
 {{site.data.alerts.callout_info}}
+`RESTORE` cannot restore backups made by newer versions of CockroachDB.
+{{site.data.alerts.end}}
+
+{{site.data.alerts.callout_info}}
 `RESTORE` is a blocking statement. To run a restore job asynchronously, use the `DETACHED` option. See the [options](#options) below.
 {{site.data.alerts.end}}
 


kathancox:
@dt Can you confirm in your review whether this note needs to be on 20.2 as well? I wasn't sure if backups were made by 21.1 and then `RESTORE` ran on 20.2 whether that would be an issue.. Or, whether this just applies to `RESTORE` 21.1+

dt (approved):
20.2 doesn't _enforce_ a similar limitation so it is indeed 21.1 only, though it is probably a Very Bad Idea to restore a newer backup on any older versions

lnhsingh (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10464#-:-MZUJNkh6MeEZq2OQmhz:b593w57)* status: :shipit: complete! 1 of 0 LGTMs obtained

PR #10531, opened 2021-05-11
Added note on TLS encryption

### LINKED ISSUE #4699: Document that cloud storage SDKs use TLS
A customer noted that we don't explicitly mention whether CockroachDB's connections to external cloud storage systems are encrypted. We should add a note to the table of cloud storage providers indicating that the S3, Google, Azure, and HTTPS cloud storage providers all use TLS under the hood.

kathancox:
Added a note on TLS encryption for in transit and at rest following the cloud storage options table.

Closes #4699

### DIFF
--- a/v20.2/use-cloud-storage-for-bulk-operations.md
+++ b/v20.2/use-cloud-storage-for-bulk-operations.md
@@ -36,7 +36,7 @@ If your environment requires an HTTP or HTTPS proxy server for outgoing connecti
 {{site.data.alerts.end}}
 
 {{site.data.alerts.callout_info}}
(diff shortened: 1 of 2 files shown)


mwang1026 (commented):
What's written lgtm. I think one thing we might consider adding is mention of the client side encryption that we now offer (at the time the docs issue was written we didn't have that feature)

Something like `CRDB also provides client-side encryption of backup data via (link to the backup encryption docs)`

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10531#-:-M_RMiZ53suXnoB1ZSu9:bjkbuin)* status: :shipit: complete! 0 of 0 LGTMs obtained

kathancox:
@mwang1026 good point, have added this to the note.

jseldess (approved):
:lgtm: 

Separate issue, but @kathancox, we should probably look at conventions around how many callouts to include in a section, on a page, etc. Perhaps there are patterns we can use that would lead to less callout overload. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10531#-:-M_SUDh58p47_1LKxy_Q:b-6l8mog)* status: :shipit: complete! 1 of 0 LGTMs obtained

kathancox:
Thanks Jesse. Yes, I would like to figure out some best practices when it comes to that—notes start becoming a bit meaningless at a certain number.

PR #10535, opened 2021-05-11
Included known limitation for interleaved table 21.1

### LINKED ISSUE #10353: release-21.1: backupccl: require explicit option to backup interleaved tables
https://github.com/cockroachdb/cockroach/pull/63501

This should be listed as a known limitation in the BACKUP docs, as well as in the Upgrade docs, as folks that upgrade inadvertently may have their BACKUPs suddenly fail unless they have that option.

Currently, asking team whether we should backport the option as a no-op so that upgrade path is:
- Update your BACKUP processes to include that option
- Upgrade and hunky dory

kathancox:
Added a known limitation to backup docs 21.1 so it's clear that a BACKUP will fail without the option and restores won't work that backup to future versions with interleaves tables.

Closes #10353


jseldess (commented):
Looks good, but can you please also include this in the 21.1 Known Limitations page? https://github.com/cockroachdb/docs/blob/master/v21.1/known-limitations.md

Also, for limitations, as you'll see on that page, we like to include a GH tracking issue. Do we have one? That also makes it easier for us to track down when limitations are removed.

v21.1/backup.md
```diff
@@ -250,6 +250,8 @@ job_id             |  status   | fraction_completed | rows | index_entries | byt
 
 {% include {{ page.version.version }}/known-limitations/backup-interleaved.md %}
 
+[Tracking GitHub Issue](https://github.com/cockroachdb/cockroach/issues/52009)
```
jseldess, line 253: nit: Why don't you put this into the include so you don't have to maintain it twice?

kathancox: Ahhh thanks Jesse, good point. Done!

jseldess (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10535#-:-M_XW9Az5w6HxdkdQnxn:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @jseldess and @kathancox)

PR #10555, opened 2021-05-12
Add docs for spatial storage of Z/M coordinates

### LINKED ISSUE #9706: Document spatial storage of Z/M shapes
Storage of Z/M coordinates for shapes. We estimate this work will occur for the storage layer, and propagation up through built-in functions.  Index acceleration will probably not be part of the work.

rmloveland:
Fixes #9706.

Summary of changes:

- Update the following pages to describe that we also support Z and M
  coords:
  - POINT
  - LINESTRING
  - POLYGON
  - MULTIPOINT
  - MULTILINESTRING
  - MULTIPOLYGON
  - GEOMETRYCOLLECTION

- For the GEOMETRY data type, we update its entry in the 'Spatial
  Glossary' to point to the above docs, which give specifics

### DIFF
--- a/_includes/v21.1/spatial/zmcoords.md
+++ b/_includes/v21.1/spatial/zmcoords.md
@@ -1,0 +1,27 @@
+<span class="version-tag">New in v21.1:</span> You can also store a `{{page.title}}` with the following additional dimensions:
+
+- A third dimension coordinate `Z` (`{{page.title}}Z`).
+- A measure coordinate `M` (`{{page.title}}M`).
+- Both a third dimension and a measure coordinate (`{{page.title}}ZM`).
+
+The `Z` and `M` dimensions can be accessed or modified using a number of [built-in functions](functions-and-operators.html#spatial-functions), including:
+
+- `ST_Z`
+- `ST_M`
+- `ST_Affine`
+- `ST_Zmflag`
+- `ST_MakePoint`
+- `ST_MakePointM`
+- `ST_Force3D`
+- `ST_Force3DZ`
+- `ST_Force3DM`
+- `ST_Force4D`
+- `ST_Snap`
+- `ST_SnapToGrid`
+- `ST_RotateZ`
+- `ST_AddMeasure`
+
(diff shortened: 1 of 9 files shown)


_includes/v21.1/spatial/zmcoords.md
```diff
@@ -0,0 +1,24 @@
...
+- `ST_Force3D`
+- `ST_Force3DZ`
+- `ST_Force3DM`
+- `ST_Force4D`
+- `ST_Snap`
+- `ST_SnapToGrid`
+- `ST_RotateZ`
+- `ST_AddMeasure`
+
+Note that these functions are not index-accelerated.
```
otan, line 24: i would say "Note that the Z/M dimension is not index accelerated when using spatial predicates - indexing is still based on the 2D coordinate system. Some spatial functions ignore the Z/M dimension, with transformations discarding the Z/M value.", or something of the ilk as well.

_includes/v21.1/spatial/zmcoords.md
```diff
@@ -0,0 +1,24 @@
+<span class="version-tag">New in v21.1:</span> You can also store a `{{page.title}}` with the following additional dimensions:
+
+- A third dimension coordinate `Z` (`{{page.title}}Z`).
+- A measure coordinate `M` (`{{page.title}}M`).
+- Both a third dimension and a measure coordinate (`{{page.title}}ZM`).
+
+The `Z` and `M` dimensions are accessible using a number of [built-in functions](functions-and-operators.html#spatial-functions), including:
```
otan, line 7: I would replace "accessible" with "Z and M dimensions can be accessed or modified using a number of ..."

_includes/v21.1/spatial/zmcoords.md
rmloveland: Makes sense, I took a shot at it using a slightly edited version of what you said! PTAL

_includes/v21.1/spatial/zmcoords.md
rmloveland: Updated!

ericharmeling (approved):
LGTM!

Issue #10052, opened 2021-03-23
multiregionccl: gate ADD REGION under ccl / enterprise license


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/61169

From release notes:
> `ALTER DATABASE ... ADD REGION` now requires an enterprise license.[#61169][#61169] {% comment %}doc{% endcomment %}

PR #10560, opened 2021-05-13
Multiregion things require an enterprise license

rmloveland:
This is the analogue to #10052 (ALTER DATABASE ... ADD REGION) where we
mark various high-level pages describing multi-region capabilities with
the "enterprise feature" tag so users know about it up front.


ericharmeling (approved):
LGTM!

PR #10580, opened 2021-05-17
Update install-cockroachdb-mac.html

tbg:
```
$ curl https://binaries.cockroachdb.com/cockroach-v20.2.9.darwin-10.9-amd64.tgz | tar -xJ
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0xz: (stdin): File format not recognized
```

Haven't tried the linux page but maybe it has the same issue.

### DIFF
--- a/v20.2/install-cockroachdb-mac.html
+++ b/v20.2/install-cockroachdb-mac.html
@@ -55,7 +55,7 @@ If you previously installed CockroachDB via Homebrew, run <code>brew uninstall c
         <svg data-eventcategory="mac-binary-step1-button" id="copy-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 12"><style>.st0{fill:#A2A2A2;}</style><title>icon/buttons/copy</title><g id="Mask"><path id="path-1_1_" class="st0" d="M4.9 4.9v6h6v-6h-6zM3.8 3.8H12V12H3.8V3.8zM2.7 7.1v1.1H.1S0 5.5 0 0h8.2v2.7H7.1V1.1h-6v6h1.6z"/></g></svg>
         <svg id="copy-check" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 10"><style>.st1{fill:#54B30E;}</style><path id="path-1_2_" class="st1" d="M3.8 9.1c-.3 0-.5-.1-.6-.2L.3 6C0 5.7-.1 5.2.2 4.8c.3-.4.9-.4 1.3-.1L3.8 7 10.6.2c.3-.3.9-.4 1.2 0 .3.3.3.9 0 1.2L4.4 8.9c-.2.1-.4.2-.6.2z"/></svg>
       </div>
-      <div class="highlight"><pre class="highlight"><code data-eventcategory="mac-binary-step1"><span class="nv language-shell mac-binary-step1" id="mac-binary-step1-{{ page.version.version }}" data-eventcategory="mac-binary-step1">$ </span>curl https://binaries.cockroachdb.com/cockroach-{{page.release_info.version}}.darwin-10.9-amd64.tgz | tar -xJ</code></pre></div>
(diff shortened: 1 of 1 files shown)


rmloveland:
Thanks for the patch Tobi!  It works on my machine (tm) as well

Strangely, the existing command also works on my machine, but maybe I have a different `tar` version because I'm still on macOS 10.15.7 (Catalina)?

What is your `tar` version?

What does `uname -a` say?

```
$ curl https://binaries.cockroachdb.com/cockroach-v20.2.9.darwin-10.9-amd64.tgz | tar -xJ
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 52.7M  100 52.7M    0     0  10.9M      0  0:00:04  0:00:04 --:--:-- 11.4M
$ echo $?
0
$ tar --version
bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.6
$ uname -a
Darwin loveland-MacBook-Pro 19.6.0 Darwin Kernel Version 19.6.0: Mon Apr 12 20:57:45 PDT 2021; root:xnu-6153.141.28.1~1/RELEASE_X86_64 x86_64
```

Paging @jseldess b/c I'm hesitant to touch this page near release day without another set of eyes - but Tobi's change to the `tar` flags also works fine for me.  Does it work for you?

tbg:
@rmloveland I'm actually on Linux :-)

```
$ tar --version
tar (GNU tar) 1.30
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.

$ tar --help
[...]
       -J, --xz
              Filter the archive through xz(1).
```

You can reproduce by using `tar` from the builder image:

```
curl https://binaries.cockroachdb.com/cockroach-v20.2.9.darwin-10.9-amd64.tgz | build/builder.sh tar -xJ
```

jseldess:
@tbg, did you get that error on Mac or on Linux? I'm confused. Current docs work fine for me on Mac. 

I think we should review and refresh these commands here and everywhere we use them. There's actually a project to put these types of commands directly in the CC Console, led by @emily-horing, that @knz is helping with. Let's hold of on merging this PR for the moment. As Rich says, too many 21.1 changes going in for the GA release today.

tbg:
I'm on linux and got the error on linux. I'm fine holding off here. However the file is clearly gzipped:

> file cockroach-v21.1.0-beta.5.linux-amd64.tgz 
cockroach-v21.1.0-beta.5.linux-amd64.tgz: gzip compressed data, original size modulo 2^32 194082816 gzip compressed data, reserved method, ASCII, has CRC, extra field, encrypted, from FAT filesystem (MS-DOS, OS/2, NT), original size modulo 2^32 194082816

so the `-z` flag seems appropriate (unlike the `-J` flag, which denotes xz compression, and I don't think it's compatible with gzip though it does purportedly provide better results).

jseldess:
Ah, ok, thanks for the clarification, @tbg. Your initial comment said, "Haven't tried the linux page but maybe it has the same issue.", so I was confused.

tbg:
I meant that I didn't try the linux download (the issue above occurs when
following the OSX steps on linux). It's definitely not something many
people will od.

knz:
The J flag was incorrect from the get go and should never have been used.

It appears to work on Macos because Apple baked in automatic detection in their non- standard tar command. But that is an accident and even someone using a custom installed tar via e.g homebrew would run into issues.

We use tgz so the command should use "-xz". That would work on all platforms.

I said so much to Emily for the UI dialog but we need to update the docs throughout regardless.

jseldess:
Thanks, @knz. I'll make that change then.

jseldess:
Closing in favor or https://github.com/cockroachdb/docs/pull/10629. I'll ping you all there.

PR #10614, opened 2021-05-19
inline css

anagio:
test critical css


lnhsingh:
@anagio can I close out this PR?

PR #10620, opened 2021-05-20
Included note on cdc kafka version incompatibility

### LINKED ISSUE #10597: Document CDC backwards incompatible change in 21.1
The upgrade to the sarama library in facbd60 included a change to the default Version assumed by sarama from 0.8.2.0 to 1.0.0.

The result of this is that we no longer support Kafka < 1.0 (Confluent Platform < 4.0).

All of the impacted versions are EOL according to:

https://docs.confluent.io/platform/current/installation/versions-interoperability.html

NB: We are working on a workaround for users on these older versions, but at the moment we are simply incompatible with them.

kathancox:
Added note in releases doc, `CREATE CHANGEFEED` kafka section, and streaming docs. 

@stevendanna let me know if you feel more detail is needed here! Thanks

Closes #10597

### DIFF
--- a/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
+++ b/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
@@ -308,6 +308,10 @@ In this example, you'll set up a changefeed for a single-node cluster that is co
 
 3. Download and extract the [Confluent Open Source platform](https://www.confluent.io/download/) (which includes Kafka).
 
+{{site.data.alerts.callout_info}}
+Currently, [changefeeds](stream-data-out-of-cockroachdb-using-changefeeds.html) connected to [Kafka versions < v1.0](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) are not supported in CockroachDB v21.1.
+{{site.data.alerts.end}}
+
 4. Move into the extracted `confluent-<version>` directory and start Confluent:
 
     {% include copy-clipboard.html %}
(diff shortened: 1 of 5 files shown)


v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
```diff
@@ -308,6 +308,10 @@ In this example, you'll set up a changefeed for a single-node cluster that is co
 
 3. Download and extract the [Confluent Open Source platform](https://www.confluent.io/download/) (which includes Kafka).
 
+{{site.data.alerts.callout_info}}
+Currently, [changefeeds](stream-data-out-of-cockroachdb-using-changefeeds.html) connected to [Kafka versions < v1.0](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) are not supported in CockroachDB v21.1.
+{{site.data.alerts.end}}
```
stevendanna, line 313: I leave it to you, but I think we could probably avoid this callout here.  If the user is downloading confluent.io for the first time via these instructions, they are unlikely to be downloading an End-of-Life version.

stevendanna (approved):
Looks good to me content wise! I left one minor comment.

v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md
kathancox: Yes, that's a v. good point and I agree. I probably went slightly overboard with the callouts :-) I think the other sections would suffice... Thanks for the comment!

jseldess (approved):
:lgtm:, but @kathancox, would you mind improving the intro sentence on the known limitations page? https://deploy-preview-10620--cockroachdb-docs.netlify.app/docs/v21.1/known-limitations.html#change-data-capture. It summarizes change data capture, but doesn't intro the list of limitations in any way. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10620#-:-MaTi-SBA7w5QZ0kGQtw:b-pnv3qc)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stevendanna)

PR #10623, opened 2021-05-20
Backported to the example file URL changes to 20.1

### LINKED ISSUE #8603: v20.1 BACKUP Doc Update misleading example name
Re: [BACKUP](https://cockroachlabs.com/docs//v20.1/backup.html)

## Issue Description

In section https://www.cockroachlabs.com/docs/v20.1/backup.html#example-file-urls, the example file 'employees.sql' is a bit misleading as the output of BACKUP is not a sql dump file (where the extension for that file would be indeed '.sql') but instead is a set of .sst files in a more complex dir structure.

## Suggested Resolution 

I suggest we change the name from 'employees.sql' to 'cluster-name' or 'database-name'.

kathancox:
Updated the example file URLs to match those updated in v20.2.

Closes #8603


kathancox:
@jseldess The deploy link here doesn't show the changes propagated to the 20.1 `BACKUP` page.... wondering if it has something to do with the separation of the files that I've messed up

jseldess:
Yes, @kathancox, there must be a 20.2 version of the include file that you need to edit. Check `_includes/v20.2/misc/external-urls.md...

jseldess (approved):
LGTM. Thanks, @kathancox!

PR #10628, opened 2021-05-20
Update restore.md

dbist:
admin role appears a few times throughout this page and most of the time it is capitalized, same as in the BACKUP docs. Lauren is out, requesting review from you Jesse.


ianjevans (approved):

jseldess (approved):
LGTM

PR #10633, opened 2021-05-21
Linking changed doc content to the preview build.

ianjevans:
Modifies the GitHub action to link changed documentation topics to the Netlify preview site for the PR.

We can anticipate the preview site URL using the PR number and the changed file. Non-Markdown files and Markdown files that are in directories that begin with an underscore (e.g. `_includes/v21.1/file.md`) are not linked.

I also added a step to update any existing changed file comment so the action doesn't create multiple comments.

### DIFF
--- a/.github/workflows/changed_files.yml
+++ b/.github/workflows/changed_files.yml
@@ -2,7 +2,7 @@ name: List files changed as GitHub comment
 
 on:
   pull_request
-      
+
 jobs:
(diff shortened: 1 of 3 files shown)


yzdocs (approved):
Looks great! Thanks for tackling.

yzdocs:
Should we revert the two test comments?

PR #10636, opened 2021-05-23
add known limitation for intent buildup

erikgrinaker:
Users often experience slow or hung backups or other operations as a
consequence of intent buildup, and frequently contact our support staff
to resolve these issues.

This patch adds a section to our known limitations with tips for how to
diagnose and work around these issues.

This is a draft, once reviewed I will add a corresponding section for 21.1.


v20.2/known-limitations.md
```diff
@@ -194,6 +194,30 @@ Before increasing this value, however, verify that you will not end up saturatin
...
+To verify that intents may be causing an issue, visit the DB Console and create a custom chart for the metric `intentcount`. This will show the number of intents present over time. The following query can also be used to get intent counts by range:
+
+{% include copy-clipboard.html %}
+~~~ sql
+> SELECT * FROM (SELECT start_pretty, end_pretty, range_id, crdb_internal.range_stats(start_key)->'intent_count' AS intent_count FROM crdb_internal.ranges_no_leases) WHERE intent_count != '0'
+~~~
+
+To force cleanup of intents, either of the following methods can be used:
+
+* Do a high-priority scan of the table, which will resolve intents as it runs. Note that this may abort any conflicting transactions that are currently running. If the table has indexes, these can be cleaned by changing `<table>` into `<table>@<index>`. The key ranges in the intent query above will output numeric table identifiers that can be entered as `[<tableid>]` instead of the table name.
```
otan, line 210: do we need to detail `[table_id]@[idx_id]`?

otan (commented):
does this apply to v20.1 too?

also wonder if this is the right page for it, but i will defer this to the docs team.

thanks for getting on this!

v20.2/known-limitations.md
erikgrinaker: May as well, update.

erikgrinaker:
>does this apply to v20.1 too?

I'm not sure, we haven't focused too much on it since it's no longer supported. Looked at a few support issues, and they were mostly for 20.2. The one I found for 20.1 appeared to have a different cause (an ongoing, long-running query). I don't suppose it hurts to add it for 20.1 either though.

> also wonder if this is the right page for it, but i will defer this to the docs team.

@jseldess Have an opinion on where this belongs?

jseldess (commented):
This is great. Thanks for writing it up, @erikgrinaker! I left some minor copy edit suggestions. 

Otherwise, do we have a GitHub issue to link to? If so, please add that at the end as `[Tracking GitHub Issue](https://github.com/cockroachdb/cockroach/issues/<number>)`.

I think this is the right place. The only other place that might make sense is here: https://www.cockroachlabs.com/docs/v21.1/backup#known-limitation. If you want to expose it there, however, you will need to create this as an `_include` file and included it in both places, for easier future maintenance. Let me know what you think and I can help.

Also, please apply these changes to the v21.1 docs as well. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10636#-:-MajqdEdD9f6ZjruO8ve:b-gjp2gz)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @erikgrinaker, @jreut, and @otan)

___
*[v20.2/known-limitations.md, line 201 at r2](https://reviewable.io/reviews/cockroachdb/docs/10636#-MajoJJiES0nr-6w2_lp:-MajoJJiES0nr-6w2_lq:bebd90h) ([raw file](https://github.com/cockroachdb/docs/blob/b0b179368e49638fa03018b1711bdf4edac5c208/v20.2/known-limitations.md#L201)):*
> ```Markdown
> Due to known bugs, transactions do not always clean up their [write intents](architecture/transaction-layer.html#write-intents) (newly written values) on commit or rollback. Garbage collection is also rather slow to react to them. This can cause the amount of unresolved write intents to build up over time. While this isn't necessarily a problem in itself, some operations do not handle large amounts of intents well. In particular, backups and queries that touch large numbers of values may become very slow and appear to hang.
> 
> To verify that intents may be causing an issue, visit DB Console → Advanced Debug → Custom Chart and create a chart for the metric `intentcount`. This will show the number of intents present over time. The following query can also be used to get intent counts by range:
> ```

I'd suggest a slight rewrite:

```
To verify that intents may be causing an issue, open the [**Custom Chart** debug page](ui-custom-chart-debug-page.html) in the DB Console, and create a chart for the `intentcount` metric. This will show the number of intents present over time. The following query can also be used to get intent counts by range:
```
___
*[v20.2/known-limitations.md, line 205 at r2](https://reviewable.io/reviews/cockroachdb/docs/10636#-MajnHo_18uR-db7fRMg:-MajnHo_18uR-db7fRMh:b-nrm2wq) ([raw file](https://github.com/cockroachdb/docs/blob/b0b179368e49638fa03018b1711bdf4edac5c208/v20.2/known-limitations.md#L205)):*
> ```Markdown
> ~~~ sql
> > SELECT * FROM (SELECT start_pretty, end_pretty, range_id, crdb_internal.range_stats(start_key)->'intent_count' AS intent_count FROM crdb_internal.ranges_no_leases) WHERE intent_count != '0'
> ```

Need to add a closing semicolon.
___
*[v20.2/known-limitations.md, line 210 at r2](https://reviewable.io/reviews/cockroachdb/docs/10636#-MajndxJ1zR_Yx6E3CTq:-MajndxK1vrqHYLmq9X1:b-y9dwge) ([raw file](https://github.com/cockroachdb/docs/blob/b0b179368e49638fa03018b1711bdf4edac5c208/v20.2/known-limitations.md#L210)):*
> ```Markdown
> To force cleanup of intents, either of the following methods can be used:
> 
> * Do a high-priority scan of the table, which will resolve intents as it runs. Note that this may abort any conflicting transactions that are currently running. If the table has indexes, these can be cleaned by changing `<table>` into `<table>@<index>`. Numeric table and/or index identifiers (as output e.g. by the intent query above) can be used instead of names by placing them in brackets, e.g. `[<table-id>]` or `[<table-id>]@[<index-id>]`.
> ```

nit: We generally use `-` for Markdown bullets, though I'm sure we're not 100% consistent.

Also, I would change `(as output e.g. by the intent query above)` to `(e.g., as output by the intent query above)`.
___
*[v20.2/known-limitations.md, line 212 at r2](https://reviewable.io/reviews/cockroachdb/docs/10636#-MajnY9DApVp8y_K91Gj:-MajnY9DApVp8y_K91Gk:bo8d0qa) ([raw file](https://github.com/cockroachdb/docs/blob/b0b179368e49638fa03018b1711bdf4edac5c208/v20.2/known-limitations.md#L212)):*
> ```Markdown
> * Do a high-priority scan of the table, which will resolve intents as it runs. Note that this may abort any conflicting transactions that are currently running. If the table has indexes, these can be cleaned by changing `<table>` into `<table>@<index>`. Numeric table and/or index identifiers (as output e.g. by the intent query above) can be used instead of names by placing them in brackets, e.g. `[<table-id>]` or `[<table-id>]@[<index-id>]`.
> 
> ```

I think we want this code block to be nested under the bullet above? If so, you need to indent the whole thing four spaces.
___
*[v20.2/known-limitations.md, line 217 at r2](https://reviewable.io/reviews/cockroachdb/docs/10636#-MajpV5BBKY2MdZ0lGC-:-MajpV5C99bQg0yKyZqz:b-w94qou) ([raw file](https://github.com/cockroachdb/docs/blob/b0b179368e49638fa03018b1711bdf4edac5c208/v20.2/known-limitations.md#L217)):*
> ```Markdown
> ~~~
> 
> * Manually enqueue the range for garbage collection. Go into the DB Console → Advanced Debug → Enqueue Range. Select "Queue: gc", enter the range ID as output by the intent query above, check "SkipShouldQueue", and press "Submit". The operation will succeed on the leaseholder node and error on the others -- this is expected.
> ```

I'd suggest a slight revision:

```
- Manually enqueue the range for garbage collection. In the DB Console, open the [**Advanced Debug** page](ui-debug-pages.html), scroll down to **Tracing and Profiling Endpoints**, and click **Run a range through an internal queue**. then select **Queue: gc**, enter the range ID as output by the intent query above, check **SkipShouldQueue**, and click **Submit**. The operation will succeed on the leaseholder node and error on the others; this is expected.
```

erikgrinaker (commented):
Thanks for looking it over @jseldess, the suggestions are great! Will update the PR over the weekend.

> The only other place that might make sense is here: https://www.cockroachlabs.com/docs/v21.1/backup#known-limitation.

Yeah, given the frequency with which this occurs, we should call it out here as well. The text is a bit verbose, so perhaps a small note here with a link to the full text in the known limitations section might make sense.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10636#-:-Mam_4ssBHkpUm2Uog5O:b-yf83ts)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @erikgrinaker, @jreut, and @otan)

jseldess:
Thanks, @erikgrinaker. I think it's actually better to use an include and embed this content in both places. However, I'd want to introduce subheadings in the Known limitations section of the Backup page. If you're comfortable with it, I can take over the PR and set up the include once you've editing the main known limitation. That ok with you?

erikgrinaker:
> If you're comfortable with it, I can take over the PR and set up the include once you've editing the main known limitation. That ok with you?

Sounds great, thank you @jseldess! I've applied your suggestions, feel free to make any further edits as you see fit and merge this.

jseldess (approved):

PR #10639, opened 2021-05-24
CC Multi-Region 

gemma-shay:
Closes: https://cockroachlabs.atlassian.net/browse/CC-3813?atlOrigin=eyJpIjoiMDAyMTA0ZDkxZTJmNDU5NTkwYjk4YTc3OTQzNTAyMmIiLCJwIjoiaiJ9


gemma-shay:
CC: @jseldess

kannanlakshmi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay)

___
*[cockroachcloud/cluster-management.md, line 91 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoINEc29O-tKqtS5UL:-MaoINEc29O-tKqtS5UM:b-7kbq0m) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L91)):*
> ```Markdown
> ### Considerations
> 
> - You can add or remove up to three regions at a time. The maximum number of regions is seven, and two-region clusters are not allowed because they would be unable to survive a single region failure.
> ```

Can we break this first bullet further. 
1. You can add or remove upto 3 regions at a time. 
2. In the UI, we support upto 9 regions in a cluster (not 7) -- something that says that we do support more than 9 regions, but in the UI itself, it will be a max of 9. 
3. We don't allow 2 region clusters...for the reasons you mention above. 
___
*[cockroachcloud/cluster-management.md, line 92 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoInoI7PxFiJZ38K3B:-MaoInoI7PxFiJZ38K3C:bfp6mwu) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L92)):*
> ```Markdown
> 
> - You can add or remove up to three regions at a time. The maximum number of regions is seven, and two-region clusters are not allowed because they would be unable to survive a single region failure.
> - We recommend you use the same number of nodes in each region of your cluster for best performance and stability.
> ```

Its same number of nodes as well as multiples of the AZ for # of nodes in one region. These are two separate concepts that need to be explained further. Note that the number of AZs we're using by default is 3 in GCP, and 4 in AWS. We need to explain this further -- we recommend the same number of nodes across regions for balanced performance. 
we recommend nodes in the multiples of the AZs per region for better balancing of data. Maybe we can even give an example here. 
@joshimhoff might be able to provide better language here or just more details. 
___
*[cockroachcloud/cluster-management.md, line 95 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoLBxo6czSkXKOdK5w:-MaoLBxo6czSkXKOdK5x:ba4lghi) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L95)):*
> ```Markdown
> - When a region is removed from a cluster, access to the database from that region will no longer be as fast.
> - Adding or removing regions incurs a non-trivial amount of load on the cluster. Changing the cluster configuration during times of heavy traffic can result in degraded application performance or longer times for node modifications. We recommend you add or remove regions from a cluster when the cluster isn't experiencing heavy traffic.
> ```

We also require 3 or more nodes per region in a multi-region cluster. 
___
*[cockroachcloud/cluster-management.md, line 106 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoJr-42c8-BJ-Ky5_a:-MaoJr-42c8-BJ-Ky5_b:b-nu4qd0) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L106)):*
> ```Markdown
> 1. Click **Add a region**.
> 
>     If you have [VPC peering](network-authorization.html) enabled, the IP range will be automatically populated for added regions.
> ```

This is just got GCP right? Might be worth splitting this section up into AWS v GCP since there are slight differences in the # of zones and privatelink / VPC peering UX. You could also just specify the cloud provider if its specific to a Cloud provider. 
___
*[cockroachcloud/create-your-cluster.md, line 42 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKAqD8IKT81K9BsdA:-MaoKAqD8IKT81K9BsdB:b-e6tgnp) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L42)):*
> ```Markdown
> 
> Some regions in GCP and AWS might not be displayed in the **Regions** list. We run CockroachCloud in EKS and GKE - the managed Kubernetes offerings for AWS and GCP respectively - and support all regions that the offerings are available in. If a particular region is not available on the CockroachCloud console, that is due to the cloud provider not supporting the managed Kubernetes offering in that region. See list of [EKS regions](https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/) and list of [GKE regions](https://cloud.google.com/about/locations/) for details.
> ```

This isn't entirely true -- there are some regions we don't support because of capacity constraints. I believe you updated our FAQs on this topic. I'd recommend either removing this paragraph or updating it to reflect the same answer as the FAQ. 
___
*[cockroachcloud/create-your-cluster.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKMWY60DUA409FnYt:-MaoKMWY60DUA409FnYu:b-i2to5b) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L64)):*
> ```Markdown
> 
> You cannot create a two-node or two-region cluster because two-replica configurations are less reliable than a single replica.
> ```

The reasoning for 2 node and 2 region are slightly different. 2 node is as you call out : 2 replicas don't give you any more availability than 1 node. 2 regions will certainly not survive a region failure. Its sort of the same concept but the level at which the availability works/doesn't work is different (node level v region level). I would spell that out more clearly.
___
*[cockroachcloud/create-your-cluster.md, line 67 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKpb74SoUMi_STA5h:-MaoKpb74SoUMi_STA5i:bg5xa8s) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L67)):*
> ```Markdown
> {{site.data.alerts.end}}
> 
> Currently, you can add a maximum of 50 nodes to your cluster. For larger configurations, [contact us](https://support.cockroachlabs.com/hc/en-us/requests/new).
> ```

This is 150 nodes per https://cockroachlabs.atlassian.net/browse/CC-3859

gemma-shay (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @joshimhoff and @kannanlakshmi)

___
*[cockroachcloud/cluster-management.md, line 91 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoINEc29O-tKqtS5UL:-Mb7EiKxA_p414Ak6gqh:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L91)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

Can we break this first bullet further. 
1. You can add or remove upto 3 regions at a time. 
2. In the UI, we support upto 9 regions in a cluster (not 7) -- something that says that we do support more than 9 regions, but in the UI itself, it will be a max of 9. 
3. We don't allow 2 region clusters...for the reasons you mention above. 
</blockquote></details>

Done.
___
*[cockroachcloud/cluster-management.md, line 92 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoInoI7PxFiJZ38K3B:-Mb7jrRZ59OQgUUEWQ2R:bz83234) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L92)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

Its same number of nodes as well as multiples of the AZ for # of nodes in one region. These are two separate concepts that need to be explained further. Note that the number of AZs we're using by default is 3 in GCP, and 4 in AWS. We need to explain this further -- we recommend the same number of nodes across regions for balanced performance. 
we recommend nodes in the multiples of the AZs per region for better balancing of data. Maybe we can even give an example here. 
@joshimhoff might be able to provide better language here or just more details. 
</blockquote></details>

Adding this in add/remove nodes section. I know the considerations sections are getting big and have a lot of crossover, will see if I can improve.
___
*[cockroachcloud/cluster-management.md, line 95 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoLBxo6czSkXKOdK5w:-Mb7Zs9lAILfZJpnGUe7:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L95)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

We also require 3 or more nodes per region in a multi-region cluster. 
</blockquote></details>

Done.
___
*[cockroachcloud/cluster-management.md, line 106 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoJr-42c8-BJ-Ky5_a:-Mb7_x9f5ufDxbAW0quo:bp6an6) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/cluster-management.md#L106)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

This is just got GCP right? Might be worth splitting this section up into AWS v GCP since there are slight differences in the # of zones and privatelink / VPC peering UX. You could also just specify the cloud provider if its specific to a Cloud provider. 
</blockquote></details>

Yes, I think this is just a note for GCP and PrivateLink connections have to get set up after the region is added. Not sure what other differences are in the UX here. I think if it's just the data distribution over AZs, that info should stay in the **Considerations** section (to be updated in another commit), and the instructions here can stay unified. This could change later with https://cockroachlabs.atlassian.net/browse/CC-4296.
___
*[cockroachcloud/create-your-cluster.md, line 42 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKAqD8IKT81K9BsdA:-Mb7czfZ-TFulD2aSnNE:b-jz4q7i) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L42)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

This isn't entirely true -- there are some regions we don't support because of capacity constraints. I believe you updated our FAQs on this topic. I'd recommend either removing this paragraph or updating it to reflect the same answer as the FAQ. 
</blockquote></details>

Current FAQ has the same language: https://www.cockroachlabs.com/docs/cockroachcloud/frequently-asked-questions.html#why-cant-i-use-certain-regions-in-aws-and-gcp. Going to remove paragraph but leave the temporarily disabled region info. Should I open an issue to update the FAQ?
___
*[cockroachcloud/create-your-cluster.md, line 64 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKMWY60DUA409FnYt:-Mb7icmnA96uIlKw5wQo:bmbcw9t) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L64)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

The reasoning for 2 node and 2 region are slightly different. 2 node is as you call out : 2 replicas don't give you any more availability than 1 node. 2 regions will certainly not survive a region failure. Its sort of the same concept but the level at which the availability works/doesn't work is different (node level v region level). I would spell that out more clearly.
</blockquote></details>

Done. Moved the 2-region warning to the region section.
___
*[cockroachcloud/create-your-cluster.md, line 67 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKpb74SoUMi_STA5h:-Mb7ikgL9SmY8aun_U8x:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L67)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

This is 150 nodes per https://cockroachlabs.atlassian.net/browse/CC-3859 
</blockquote></details>

Done.

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay, @joshimhoff, and @kannanlakshmi)

___
*[cockroachcloud/cluster-management.md, line 73 at r2](https://reviewable.io/reviews/cockroachdb/docs/10639#-Mb7nxJODsj_UxuXOX-J:-Mb7nxJODsj_UxuXOX-K:b-sqaymp) ([raw file](https://github.com/cockroachdb/docs/blob/bcb4685071682aca8b18bd0d4e8e1d6122282297/cockroachcloud/cluster-management.md#L73)):*
> ```Markdown
> - We recommend you use the same number of nodes in each region of your cluster for best performance and stability.
> - GCP clusters use 3 Availability Zones (AZs). For balanced data distribution and best performance, we recommend using a multiple of 3 nodes (e.g., 3, 6, or 9 nodes).
> - AWS clusters use 4 AZs. For balanced data distribution and best performance, we recommend using a multiple of 4 nodes (e.g., 4, 8, or 12 nodes).
> ```

I don't think it's necessarily 4. I think we deploy the number of zones in the region, which can be 3 and can also be 4. @duskeagle in case I have it wrong. Note we do plan to change this so it is always 3 (like on GCP); it's the starter project of a recent hire. Also the plan is for the UI to warn in case you don't scale by multiple of zones. That maybe allows us to be vaguer in these docs?

joshimhoff (commented):
Exciting docs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639#-:-Mb7ouwCEnSNoBy5yFLM:b9x2cya)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay, @joshimhoff, and @kannanlakshmi)

gemma-shay:
@joshimhoff 
Thoughts on adding zone config management info discussed [here](https://cockroachlabs.atlassian.net/browse/CC-4185) to the [planning section](https://deploy-preview-10639--cockroachdb-docs.netlify.app/docs/cockroachcloud/cluster-management?filters=dedicated#recommendations)?

kannanlakshmi (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @chrisseto, @gemma-shay, @joshimhoff, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbHHLuW0ynJKVDmOajc:b5dw05p) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
> ```Markdown
> - Multi-region clusters require at least 3 nodes per region.
> - We don't allow 2-region clusters because they are less stable and would be unable to survive a single region failure.
> - We don't allow 2-node clusters because 2-replica configurations are less reliable than a single replica.
> ```

This might be my own ignorance but are they less reliable than a single replica or as unreliable as a single replica? Also, we say above that MR clusters require at least 3 nodes per region, this seems repetitive, because if you don't allow less than 3 nodes/region then you of course don't allow 2 nodes/ region. Consider combining this and the first bullet with an explanation. 
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 8 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHI6jfENmjjrNR2y2S:-MbHI6jfENmjjrNR2y2T:b5z8mwq) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L8)):*
> ```Markdown
> - You can have a maximum of 9 regions per cluster through the console. If you need to add more regions, [contact us](https://support.cockroachlabs.com).
> 
> ### Recommendations
> ```

nit: should these be bullet points like the requirements above? 
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 10 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHuK13A4Z5jF7b_2r:-MbHHuK13A4Z5jF7b_2s:b-7qs1bv) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L10)):*
> ```Markdown
> ### Recommendations
> 
> We recommend you use the same number of nodes in each region of your cluster for best performance and stability. GCP clusters use 3 Availability Zones (AZs), and AWS clusters use either 3 or 4 AZs. For balanced data distribution and best performance, we recommend using a number of nodes that is a multiple of the number of AZs. For example, a GCP cluster with 3 AZs would have balanced data distribution with 3, 6, or 9 nodes per region.
> ```


@joshimhoff 3 or 4 AZs for AWS isn't helpful for users though, which is it? Can we be opinionated here? Is the problem that we haven't made the change to use all 4 AZs per region?

@gemma-shay I still think these are two separate points 1. same number of nodes in each region 2. # of nodes should be a multiple of 3 or 4. Combining them this way seems confusing to me and its easy to miss the nuance. 
___
*[cockroachcloud/create-your-cluster.md, line 42 at r1](https://reviewable.io/reviews/cockroachdb/docs/10639#-MaoKAqD8IKT81K9BsdA:-MbHJ4zV82MWcoFSgAnh:b34gy90) ([raw file](https://github.com/cockroachdb/docs/blob/d63f8e81311cff7305037dbef56313eb24c7c148/cockroachcloud/create-your-cluster.md#L42)):*
<details><summary><i>Previously, gemma-shay (Gemma Shay) wrote…</i></summary><blockquote>

Current FAQ has the same language: https://www.cockroachlabs.com/docs/cockroachcloud/frequently-asked-questions.html#why-cant-i-use-certain-regions-in-aws-and-gcp. Going to remove paragraph but leave the temporarily disabled region info. Should I open an issue to update the FAQ?
</blockquote></details>

Seems like the FAQ is up to date. This is what I was referring to 
https://www.cockroachlabs.com/docs/cockroachcloud/frequently-asked-questions.html#why-are-certain-regions-in-aws-and-gcp-not-available

kannanlakshmi:
@gemma-shay should we get Rich or someone from the MR/ docs team review this as well?

gemma-shay (commented):
Rich is OOO until June 13 unfortunately. I requested Chris Seto on the engineering side, and I think Jesse/Ian are able to do a docs review. Anyone else from MR who should be looped in?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639#-:-MbHLLn5EeMgMp-ZK9DT:buz9b49)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @chrisseto, @gemma-shay, @joshimhoff, and @kannanlakshmi)

kannanlakshmi:
Ah that's plenty then, I just wanted someone on the DB or docs side to look but Ian or Jesse is more than enough. Thank you!

gemma-shay (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @chrisseto, @gemma-shay, @joshimhoff, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbcTILm3sId54KpzYTG:bxi6cxd) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

This might be my own ignorance but are they less reliable than a single replica or as unreliable as a single replica? Also, we say above that MR clusters require at least 3 nodes per region, this seems repetitive, because if you don't allow less than 3 nodes/region then you of course don't allow 2 nodes/ region. Consider combining this and the first bullet with an explanation. 
</blockquote></details>

Combined first two bullets. According to Rich's MR docs, 2-replica configurations are less reliable than a single replica. Who can I ask to confirm this?
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 8 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHI6jfENmjjrNR2y2S:-MbcTbIS9u3E7ZmgP1jV:bsoxws9) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L8)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

nit: should these be bullet points like the requirements above? 
</blockquote></details>

Was considering going with more of a paragraph format for the whole thing, but I think you're right.
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 10 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHuK13A4Z5jF7b_2r:-MbcTsK97hDuf1DPS5jf:b-of9bxj) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L10)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>


@joshimhoff 3 or 4 AZs for AWS isn't helpful for users though, which is it? Can we be opinionated here? Is the problem that we haven't made the change to use all 4 AZs per region?

@gemma-shay I still think these are two separate points 1. same number of nodes in each region 2. # of nodes should be a multiple of 3 or 4. Combining them this way seems confusing to me and its easy to miss the nuance. 
</blockquote></details>

Separated and commented out AZ warning for now.

kannanlakshmi (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @chrisseto, @gemma-shay, @joshimhoff, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbcUxRDDfYscNW9GqXT:b5bp3uf) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
<details><summary><i>Previously, gemma-shay (Gemma Shay) wrote…</i></summary><blockquote>

Combined first two bullets. According to Rich's MR docs, 2-replica configurations are less reliable than a single replica. Who can I ask to confirm this?
</blockquote></details>

Okay Rich knows better than me, so he is probably right :)

gemma-shay:
cc @ianjevans for visibility

chrisseto (approved):
LGTM!

_includes/cockroachcloud/planning-your-cluster.md
```diff
@@ -0,0 +1,17 @@
+### Requirements
+
+- Multi-region clusters require at least 3 nodes per region because 2-replica configurations are less reliable than a single replica.
+- We don't allow 2-region clusters because they are less stable and would be unable to survive a single region failure.
+- You can have a maximum of 9 regions per cluster through the console. If you need to add more regions, [contact us](https://support.cockroachlabs.com).
```
ianjevans, line 5: Do we refer to the CC console just as the "console"? I see some places where we call it "CockroachCloud Console" and others as the "Console."

So maybe first instance on a page, call it out as CockroachCloud Console, and thereafter just Console.

jseldess (changes requested):
@ianjevans, as part of your info architecture project, I think we need to take a hard look at the way we're organizing topics for CC. I'm not sure a single "Cluster Management" page is useful anymore, for example. Planning your cluster, Adding and removing nodes, Adding and removing regions: These might all warrant separate pages, for better findability, etc. But that would require us having a way to organize content for Serverless vs. Dedicated.   

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639#-:-Mbrs8uk0cz7sFcXpHlH:b-751aur)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay, @joshimhoff, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbrssOf6ey_w4eZCln0:bfllpoa) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
<details><summary><i>Previously, kannanlakshmi (Lakshmi Kannan) wrote…</i></summary><blockquote>

Okay Rich knows better than me, so he is probably right :) 
</blockquote></details>

@gemma-shay, can you point me to where our docs say 2 replicas is less reliable than 1? I don't think that's the case. I think it's simply not more reliable. With 1 replica, if you lose it, you lose your cluster. With 2 replicas, if you lose either replica, you lose the cluster. You need to go to 3 replicas to be able to survive the loss of a single replica. In any case, I'm worried that we're being too implicit in these first two bullets. I'd suggest this rewrite:

```
- Multi-region clusters must contain at least 3 regions. With the default replication factor of 3, this ensures that data spread across regions can survive the loss of one region.
- Each region of a multi-region cluster must contain at least 3 nodes. With the default replication factor of 3, this ensures that data located entirely in a region can survive the loss of one node in that region.
``` 
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 15 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-Mbrr3f9DoX7Xy2lGjF2:-Mbrr3f9DoX7Xy2lGjF3:b-a1t4sz) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L15)):*
> ```Markdown
> - We recommend you use the same number of nodes in each region of your cluster for best performance and stability.
> - When scaling up your cluster, it is generally more effective to increase node size up to 16 vCPUs before adding more nodes. For example, if you have a 3-node cluster with 2 vCPUs per node, consider scaling up to 8 vCPUs before adding a fourth node. For most production applications, we recommend at least 4 to 8 vCPUs per node.
> - Adding or removing regions or nodes incurs a non-trivial amount of load on the cluster. Changing the cluster configuration during times of heavy traffic can result in degraded application performance or longer times for node modifications. We recommend you add or remove regions or nodes from a cluster when the cluster isn't experiencing heavy traffic.
> ```

Can we start with the last sentence, since that's the recommendation and the rest is supporting explanation?
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 16 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbrrHhq7OU8iBh1Uzkw:-MbrrHhq7OU8iBh1Uzkx:b5b2me3) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L16)):*
> ```Markdown
> - When scaling up your cluster, it is generally more effective to increase node size up to 16 vCPUs before adding more nodes. For example, if you have a 3-node cluster with 2 vCPUs per node, consider scaling up to 8 vCPUs before adding a fourth node. For most production applications, we recommend at least 4 to 8 vCPUs per node.
> - Adding or removing regions or nodes incurs a non-trivial amount of load on the cluster. Changing the cluster configuration during times of heavy traffic can result in degraded application performance or longer times for node modifications. We recommend you add or remove regions or nodes from a cluster when the cluster isn't experiencing heavy traffic.
> - Before removing nodes from a cluster, ensure that the reduced disk space will be sufficient for the existing and anticipated data. Before removing regions from a cluster, be aware that access to the database from that region will no longer be as fast.
> ```

I'd make the second sentence a separate bullet, since removing nodes and removing regions are two distinct actions. 
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 18 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-Mbrrb6F1rIeqD-xZImA:-Mbrrb6F1rIeqD-xZImB:b55uxd6) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L18)):*
> ```Markdown
> - Before removing nodes from a cluster, ensure that the reduced disk space will be sufficient for the existing and anticipated data. Before removing regions from a cluster, be aware that access to the database from that region will no longer be as fast.
> - If you have changed the [replication factor](../{{site.versions["stable"]}}/configure-zone.html) for a cluster, you might not be able to remove nodes from the cluster. For example, suppose you have a 5-node cluster and you had previously changed the replication factor from its default value of 3 to 5. Now if you want to scale down the cluster to 3 nodes, the decommissioning nodes operation to remove nodes from the cluster might fail. To successfully remove nodes from the cluster, you will have to change the replication factor back to 3.
> - Similarly, you may want to adjust your cluster's [zone configurations](../{{site.versions["stable"]}}/configure-zone.html) before or after changing your cluster topology in order to achieve your [multi-region goals](../{{site.versions["stable"]}}/multiregion-overview.html). For example, if you are removing a region, and you pin some data to that region via zone configurations, the operation to remove that region might fail. For this specific case, you may need to adjust the zone configurations ahead of running remove region.
> ```

As of 21.1 customers mostly don't mess with zone configs for multi-region. They use our new multi-region declarations for "GLOBAL" and "REGIONAL" tables and we do the zone config and other magic behind the scenes. For that reason, I think we might want to avoid referencing zone configs in this bullet. @awoods187, since @rmloveland is OOO, do you have revision suggestions here?
___
*[cockroachcloud/cluster-management.md, line 97 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbrvMUs0Qn4Q-fxlxWv:-MbrvMUs0Qn4Q-fxlxWw:b7dgofe) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/cockroachcloud/cluster-management.md#L97)):*
> ```Markdown
> ## Add or remove regions from a cluster
> 
> You can add or remove up to three regions at a time through the Console. See the [Planning your cluster](#planning-your-cluster) section of this page for cluster requirements and recommendations before proceeding.
> ```

nit: I'd shorted "See the [Planning your cluster] section of this page for..." to "See [Planning your cluster] for...

gemma-shay (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay, @ianjevans, @joshimhoff, @jseldess, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbrzdP63xox5Fj_NfXO:budjddk) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*

@gemma-shay, can you point me to where our docs say 2 replicas is less reliable than 1? I don't think that's the case. I think it's simply not more reliable. With 1 replica, if you lose it, you lose your cluster. With 2 replicas, if you lose either replica, you lose the cluster. You need to go to 3 replicas to be able to survive the loss of a single replica. In any case, I'm worried that we're being too implicit in these first two bullets. I'd suggest this rewrite:

```
- Multi-region clusters must contain at least 3 regions. With the default replication factor of 3, this ensures that data spread across regions can survive the loss of one region.
- Each region of a multi-region cluster must contain at least 3 nodes. With the default replication factor of 3, this ensures that data located entirely in a region can survive the loss of one node in that region.
``` 
</blockquote></details>

I thought it was on the Topology Patterns page, but just revisited it and realized I misinterpreted Rich's warning then repeated it here: https://www.cockroachlabs.com/docs/cockroachcloud/create-your-cluster.html#step-4-select-the-number-of-nodes and took that as evidence. My bad, will change both instances :facepalm:.
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r4](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbCZiDeKZ5EOrPfxC8L-r4-5:-Mbs3ejg-noTTsKg_tdb:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/dc95120895c0e129f5cd9710f812dd529bd4b1b0/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
<details><summary><i>Previously, ianjevans (Ian Evans) wrote…</i></summary><blockquote>

Do we refer to the CC console just as the "console"? I see some places where we call it "CockroachCloud Console" and others as the "Console."

So maybe first instance on a page, call it out as CockroachCloud Console, and thereafter just Console.
</blockquote></details>

Done.
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 15 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-Mbrr3f9DoX7Xy2lGjF2:-Mbs4exZDmefmIzPCVjp:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L15)):*

Can we start with the last sentence, since that's the recommendation and the rest is supporting explanation?
</blockquote></details>

Done.
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 16 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbrrHhq7OU8iBh1Uzkw:-Mbs4dSO0ZJVIVPwzGAA:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L16)):*

I'd make the second sentence a separate bullet, since removing nodes and removing regions are two distinct actions. 
</blockquote></details>

Done.

jseldess:
___
*[_includes/cockroachcloud/planning-your-cluster.md, line 5 at r3](https://reviewable.io/reviews/cockroachdb/docs/10639#-MbHHLuW0ynJKVDmOajb:-MbuU6Gd8bzbNtN4yo6z:b-wzrl2r) ([raw file](https://github.com/cockroachdb/docs/blob/d12da565c2f8d2360a6b20b11b5a73f873274b0d/_includes/cockroachcloud/planning-your-cluster.md#L5)):*
<details><summary><i>Previously, gemma-shay (Gemma Shay) wrote…</i></summary><blockquote>

I thought it was on the Topology Patterns page, but just revisited it and realized I misinterpreted Rich's warning then repeated it here: https://www.cockroachlabs.com/docs/cockroachcloud/create-your-cluster.html#step-4-select-the-number-of-nodes and took that as evidence. My bad, will change both instances :facepalm:.
</blockquote></details>

Thanks, @gemma-shay. I'm wondering now whether we should go further and at least reference the new multi-region table patterns. @kannanlakshmi, @awoods187, @gemma-shay, what do you think about this revision? Another layer of complexity is that we're not describing how nodes line up with AZs. I wonder if we need to explain that up front as well?

```
- Multi-region clusters must contain at least 3 regions to ensure that data replicated across regions can survive the loss of one region. For example, this applies to internal system data that is important for overall cluster operations as well as tables with the [`GLOBAL`](../stable/global-tables.html) table locality or the [`REGIONAL BY TABLE`](../stable/regional-tables.html#regional-tables) table locality and `REGION` survival goal. 
-  Each region of a multi-region cluster must contain at least 3 nodes to ensure that data located entirely in a region can survive the loss of one node in that region. For example, this applies to tables with the [`REGIONAL BY ROW`](../stable/regional-tables.html#regional-by-row-tables) table locality. 
```

kannanlakshmi:
I do think its worth linking at least to the new patterns, yes.

jseldess:
@gemma-shay, looking at the CC docs more globally, I'm noticing that the "planning" content, which is very important for multi-region, doesn't seem to be reference from the page on creating a cluster. We need to at least reference those details. Even better would be to find a way for users to have it on the page. When does this need to go live? If we have time, I'd like to propose some structural changes.

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10639)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187, @gemma-shay, @ianjevans, @joshimhoff, @jseldess, and @kannanlakshmi)

___
*[_includes/cockroachcloud/planning-your-cluster.md, line 18 at r5](https://reviewable.io/reviews/cockroachdb/docs/10639#-Mbrrb6F1rIeqD-xZImA:-Mc9sbnT6k1nwNNeOh6g:bx4sgqa) ([raw file](https://github.com/cockroachdb/docs/blob/e0da75308e7255b68a64077963dfafcd7326a86b/_includes/cockroachcloud/planning-your-cluster.md#L18)):*

As of 21.1 customers mostly don't mess with zone configs for multi-region. They use our new multi-region declarations for "GLOBAL" and "REGIONAL" tables and we do the zone config and other magic behind the scenes. For that reason, I think we might want to avoid referencing zone configs in this bullet. @awoods187, since @rmloveland is OOO, do you have revision suggestions here?
</blockquote></details>

My opinion is that we should remove the bullet altogether and see how far users can get without it.  Mentioning zone configs requires exposing a whole world of additional complexity to the user in the docs, and we have thus far avoided "here's how you do multi-region thing X via zone config" in the v21.1 docs.

(Apologies if I'm missing context on this, just getting back from PTO)

gemma-shay:
@jseldess Open to structural changes now, and can definitely get them into the follow-up PR if we run out of time.

awoods187:
I'm aligned with not referencing zone configurations.

kannanlakshmi (approved):

jseldess (approved):
LGTM, once conflict is resolved

PR #10642, opened 2021-05-24
Restore /stable and /dev as canonical links

yzdocs:
Restores /stable and /dev as canonical links instead of using versioned URLs.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -10,12 +10,7 @@
   <meta name="robots" content="nocache">
 {% endif %}
 <title>{{ page.title }} | {{ site.site_title }}</title>
-{% assign page_url = page.url | relative_url | split: "/" %}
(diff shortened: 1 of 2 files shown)


jseldess (approved):
I think this LGTM. 

But I think we may need to follow up and revert some, but not all, of the links I changed in here as well: https://github.com/cockroachdb/docs/pull/10390. I can help with that. 

Do you want anyone else to review this? Peter, perhaps? 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10642#-:-MaUrH-76VwKg4JjI5UJ:b-wyrpfc)* status: :shipit: complete! 0 of 0 LGTMs obtained

yzdocs:
Sure, added Peter as a reviewer.

PR #10644, opened 2021-05-24
Change client-side redirects to server-side

jseldess:
This is for site SEO. This is the first of several PRs
transforming client-side to server-side redirects. Also
updating internal links and, where relevant, removing old pages
we were already redirecting away from.


_redirects
```diff
@@ -5,12 +5,69 @@ layout: null
...
 /docs/v2.2/* /docs/v19.1/:splat 301!
-/docs/releases/ /docs/releases/index.html 200!
-/docs/advisories/ /docs/advisories/index.html 200!
-/docs/*/contribute-to-cockroachdb.html https://wiki.crdb.io/wiki/spaces/CRDB/pages/73204033/Contributing+to+CockroachDB 301!
-/docs/*/deploy-a-test-cluster.html /docs/cockroachcloud/quickstart.html 301!
+/docs/releases/ /docs/releases/index.html 301!
+/docs/advisories/ /docs/advisories/index.html 301!
+/docs/advisories/advisories.html /docs/advisories/index.html 301!
+
+# Renamed pages
```
yzdocs, line 12: No need for exclamation points for renamed static files

PR #10674, opened 2021-05-26
Generate sitemap that excludes all versions except stable

### LINKED ISSUE #10671: Generate our own sitemap
Instead of using the `Jekyll-sitemap` plugin. This is because we want only canonical (/stable) links to show up in the sitemap, and using the plugin to exclude other versions drastically slows our build generation time.

yzdocs:
Fixes #10671

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ source "https://rubygems.org"
 
 gem "jekyll", "~> 4.2"
 gem "jekyll-redirect-from", "~> 0.15"
-gem "jekyll-sitemap", "~> 1.3.1"
 gem "liquid-c", "~> 4.0.0"
 gem "redcarpet", "~> 3.5"
 gem "rss"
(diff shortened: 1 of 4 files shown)


jseldess (approved):
LGTM!

jseldess:
@yzdocs, one question: In the new sitemap, the `urlset` looks like this:

`<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">`

On the live site, it looks like this:

`<urlset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">`

Is this an issue?

yzdocs:
@jseldess I don't think so. For this PR I am simply following Google's guidelines on generating an XML sitemap: https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap#xml

The additional specifications in the plugin-generated sitemap is used for validation: https://www.sitemaps.org/protocol.html

jseldess:
Thanks, @yzdocs!

PR #10677, opened 2021-05-26
Restore 'key' values needed by version switcher

yzdocs:
With #10644, we removed `key` values from some pages, which breaks the version switcher (see screenshot below). The `key` value is how the version switcher ties together pages that have undergone name changes.

Screen Shot 2021-05-26 at 3 14 08 PM


_redirects
```diff
@@ -39,16 +39,8 @@ layout: null
 /docs/v19.1/explore-the-admin-ui.html /docs/v19.1/admin-ui-overview.html 301
 /docs/v19.2/explore-the-admin-ui.html /docs/v19.2/admin-ui-overview.html 301
 /docs/v20.1/explore-the-admin-ui.html /docs/v20.1/admin-ui-overview.html 301
-/docs/v20.2/explore-the-admin-ui.html /docs/v20.2/admin-ui-overview.html 301
```
yzdocs, line 42: admin-ui-overview.html does not exist under v20.2 and v21.1 folders. For those two versions the file is named ui-overview.html (added below on lines 42 and 43).

_redirects
```diff
@@ -39,16 +39,8 @@ layout: null
...
-/docs/v21.1/explore-the-admin-ui.html /docs/v21.1/admin-ui-overview.html 301
-/docs/v1.1/architecture/index.html /docs/v1.1/architecture/overview.html 301
```
yzdocs, line 44: The existing "redirect_from: index.html" does not work anyway (possibly due to ambiguity with index.html at folder top-level).

jseldess (approved):
Whoops. Thanks for these fixes, @yzdocs.

LGTM

PR #10680, opened 2021-05-27
Add more explicit guidance on accessing the DB Console

### LINKED ISSUE #10679: SQL playground is missing guidance on accessing the DB Console
Feedback from @ajstorm:

> I was just playing around with it a bit and was wondering if there’s some way we could ensure that the user info for the DB Console ends up visible on the screen when the user starts?  Right now you have to scroll up, and also interpret the GRANT statement to know how to login to the console.  Perhaps a URL to the console (like we provide with demo would help).

In fact, we provide the DB Console credentials in the intro, but it's very easy to miss, and we don't explicitly guide users on how to open the DB Console: You have to click the "DB Console" tab in the interactive terminal. We can't give them a URL because it is dynamic based on the ephemeral VM.

jseldess:
Fixes #10679

### DIFF
--- a/v20.2/sql-playground.md
+++ b/v20.2/sql-playground.md
@@ -1,12 +1,14 @@
 ---
 title: SQL Playground
-summary: The SQL Playground gives you an interactive SQL shell to a single-node CockroachDB cluster. 
+summary: The SQL Playground gives you an interactive SQL shell to a single-node CockroachDB cluster.
 toc: false
 ---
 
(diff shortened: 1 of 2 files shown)


ajstorm (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10680#-:-Majg4JtDSFW9RbKgm4L:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @jseldess)

PR #8480, opened 2020-09-28
Indexes no longer needed on fk-constrained columns

ericharmeling:
Fixes #7861.
Fixes #8122. 
Related to #5967.

This PR removes language around auto-generated FK indexes on FK columns (i.e., referencing columns). 

----

Note that there still exist a number of examples that use `movr` tables with auto-generated FK indexes:

- comment-on.md
- create-table.md
- delete.md
- demo-low-latency-multi-region-deployment.md
- drop-table.md
- explain.md
- make-queries-fast.md
- multi-region-database.md
- performance-tuning-insecure.md
- performance-tuning.md
- query-replication-reports.md
- show-backup.md
- show-index.md
- show-partitions.md
- show-range-for-row.md
- show-ranges.md
- show-zone-configurations.md
- temporary-tables.md
- import-movr.md
- partition-by-city.md
- create-a-replication-zone-for-a-secondary-index.md
- create-a-replication-zone-for-a-table-partition.md

These examples can be updated in a separate PR, following the removal of these obsolete indexes (see tracking issue: https://github.com/cockroachdb/cockroach/issues/54764). That follow-up work should resolve #5967.

PR #10681, opened 2021-05-27
Columns referenced by FK do not need to be indexed

ericharmeling:
Background: https://github.com/cockroachdb/cockroach/pull/50771#issuecomment-849701585

### DIFF
--- a/v20.2/foreign-key.md
+++ b/v20.2/foreign-key.md
@@ -24,28 +24,16 @@ For example, given an `orders` table and a `customers` table, if you create a co
 - Foreign key columns must use their referenced column's [type](data-types.html).
 - A foreign key column cannot be a [computed column](computed-columns.html).
 
-{{site.data.alerts.callout_info}}
-<span class="version-tag">New in v20.2:</span> Foreign key columns do not need to be indexed. In versions < v20.2, an index on referencing foreign key columns is required.
-{{site.data.alerts.end}}
-
 **Referenced Columns**
 
 - Referenced columns must contain only unique sets of values. This means the `REFERENCES` clause must use exactly the same columns as a [`UNIQUE`](unique.html) or [`PRIMARY KEY`](primary-key.html) constraint on the referenced table. For example, the clause `REFERENCES tbl (C, D)` requires `tbl` to have either the constraint `UNIQUE (C, D)` or `PRIMARY KEY (C, D)`.
 - In the `REFERENCES` clause, if you specify a table but no columns, CockroachDB references the table's primary key. In these cases, the `FOREIGN KEY` constraint and the referenced table's primary key must contain the same number of columns.
 - <span class="version-tag">New in v20.2:</span> By default, referenced columns must be in the same database as the referencing foreign key column. To enable cross-database foreign key references, set the `sql.cross_db_fks.enabled` [cluster setting](cluster-settings.html) to `true`.
-- Referenced columns must be [indexed](indexes.html). There are a number of ways to meet this requirement:
 
-    - You can create indexes explicitly using the [`INDEX`](create-table.html#create-a-table-with-secondary-and-inverted-indexes) clause of `CREATE TABLE`.
-    - You can rely on indexes created by the [`PRIMARY KEY`](primary-key.html) or [`UNIQUE`](unique.html) constraints.
-    - If an index on the referenced column does not already exist, CockroachDB automatically creates one. It's important to note that if you later remove the `FOREIGN KEY` constraint, this automatically created index _is not_ removed.
 
-    {{site.data.alerts.callout_success}}
-    Using the referenced columns as the prefix of an index's columns also satisfies the requirement for an index. For example, if you create foreign key that references the columns `(A, B)`, an index of columns `(A, B, C)` satisfies the requirement for an index.
-    {{site.data.alerts.end}}
-
-    {{site.data.alerts.callout_info}}
-     You can drop the index on the referenced columns if another index exists on the same columns and fulfills the indexing requirement described above.
-    {{site.data.alerts.end}}
+{{site.data.alerts.callout_info}}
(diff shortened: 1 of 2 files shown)


knz (commented):
LGTM from me, but I'm not the expert. Please involve someone from the schema team.

ajwerner (commented):
I think updating the docs here for correctness is great but I also think this is a place where we could help some people avoid some pain. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10681#-:-Mb8N-QbBrLdbZCoVMvh:b-8to8if)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling and @ianjevans)

___
*[v20.2/foreign-key.md, line 40 at r1](https://reviewable.io/reviews/cockroachdb/docs/10681#-Mb8Mk3O33vkkiwT6zvB:-Mb8Mk3O33vkkiwT6zvC:b-1qclx3) ([raw file](https://github.com/cockroachdb/docs/blob/54f73a85807ac69167d065f831a73651ba8bd45c/v20.2/foreign-key.md#L40)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
> - Referenced columns must be [indexed](indexes.html). There are a number of ways to meet this requirement:
>     - You can create indexes explicitly using the [`INDEX`](create-table.html#create-a-table-with-secondary-and-inverted-indexes) clause of `CREATE TABLE`.
>     - You can rely on indexes created by the [`PRIMARY KEY`](primary-key.html) or [`UNIQUE`](unique.html) constraints.
>     - If an index on the referenced column does not already exist, CockroachDB automatically creates one. It's important to note that if you later remove the `FOREIGN KEY` constraint, this automatically created index _is not_ removed.
> ```
</details>

Do we want to change this to state that it's a good idea (and, of course, remove the last line)?
___
*[v20.2/foreign-key.md, line 161 at r1](https://reviewable.io/reviews/cockroachdb/docs/10681#-Mb8MwSw1WOJ83KwN9bb:-Mb8MwSw1WOJ83KwN9bc:bw5lezu) ([raw file](https://github.com/cockroachdb/docs/blob/54f73a85807ac69167d065f831a73651ba8bd45c/v20.2/foreign-key.md#L161)):*
> ```Markdown
>     customer INT NOT NULL REFERENCES customers (id) ON DELETE CASCADE,
>     orderTotal DECIMAL(9,2),
>     INDEX (customer)
> ```

This index may not be required but it is a good idea...

ericharmeling (commented):
TFTR @ajwerner !

Made some updates following your feedback. PTAL.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10681#-:-Mb8XdIM9euu12bARQ7i:by53f4n)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajwerner and @ianjevans)

___
*[v20.2/foreign-key.md, line 40 at r1](https://reviewable.io/reviews/cockroachdb/docs/10681#-Mb8Mk3O33vkkiwT6zvB:-Mb8Xk0U1pBG86TTK7tY:b-62ke96) ([raw file](https://github.com/cockroachdb/docs/blob/54f73a85807ac69167d065f831a73651ba8bd45c/v20.2/foreign-key.md#L40)):*
<details><summary><i>Previously, ajwerner wrote…</i></summary><blockquote>

> ```
> - Referenced columns must be [indexed](indexes.html). There are a number of ways to meet this requirement:
>     - You can create indexes explicitly using the [`INDEX`](create-table.html#create-a-table-with-secondary-and-inverted-indexes) clause of `CREATE TABLE`.
>     - You can rely on indexes created by the [`PRIMARY KEY`](primary-key.html) or [`UNIQUE`](unique.html) constraints.
>     - If an index on the referenced column does not already exist, CockroachDB automatically creates one. It's important to note that if you later remove the `FOREIGN KEY` constraint, this automatically created index _is not_ removed.
> ```
Do we want to change this to state that it's a good idea (and, of course, remove the last line)?
</blockquote></details>

IIUC, all referenc**ed** columns will always be indexed, because of the `UNIQUE`/`PRIMARY` requirement stated above. So isn't information about indexing referenced columns kind of meaningless (unless it's a good idea to have referenced column(s) in two indexes?). I could be missing something though.

It's not guaranteed that referenc**ing** columns (i.e., "foreign key columns") are indexed. In my original doc changes (https://github.com/cockroachdb/docs/pull/8480), I removed [an equivalent index requirement](https://github.com/cockroachdb/docs/pull/8480/files#diff-b6122925623267ed9642c8fd4c06bb9bd9da6fdcd7b91024feb7117b3fb6d571L26) from the **Foreign Key Columns** rules list.

So maybe we should leave this be and just add a recommendation to index referencing foreign key columns to the ["Performance" section](https://www.cockroachlabs.com/docs/stable/foreign-key.html#performance) instead? I've made this change in the latest commit.
___
*[v20.2/foreign-key.md, line 161 at r1](https://reviewable.io/reviews/cockroachdb/docs/10681#-Mb8MwSw1WOJ83KwN9bb:-Mb8XxyT2E0gDJzaoj_h:bge0ewx) ([raw file](https://github.com/cockroachdb/docs/blob/54f73a85807ac69167d065f831a73651ba8bd45c/v20.2/foreign-key.md#L161)):*
<details><summary><i>Previously, ajwerner wrote…</i></summary><blockquote>

This index may not be required but it is a good idea...
</blockquote></details>

Added it back (and below)!

ajwerner (approved):
:lgtm:

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10681#-:-MbCJnPj1m0bclOAUrtb:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ianjevans)

ianjevans (approved):

PR #10685, opened 2021-05-28
Remove duplicated text in multiregion-overview.md

tyler-boyd:
Looks like whenever this was moved into an include, they used copy instead of cut.


jseldess (approved):
:lgtm: 

Thanks for the fix, @tyler-boyd!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10685#-:-MapV7Yo9eteaZJ5oczn:b-kh99xx)* status: :shipit: complete! 1 of 0 LGTMs obtained

PR #10755, opened 2021-06-08
clarify cockroach init usage

### LINKED ISSUE #10342: Explain that `cockroach init` must target the node(s) mentioned in `--join` lists
See this issue: https://github.com/cockroachdb/cockroach/issues/61621

If there are multiple nodes connecting together,

then an `init` command pointed to a node that's not listed in the `--join` list of other nodes will fail to initialize the cluster correctly.

The doc should have a call out that `init` should target a node that's listed in the `--join` list of all the other nodes.

(Or alternatively, a cluster can be initialized with just 1 node, and the other nodes added after initialization. In that case, the join flags are not restricted.)

cc @taroface

taroface:
Fixes #10342.

### DIFF
--- a/_includes/v20.2/prod-deployment/insecure-initialize-cluster.md
+++ b/_includes/v20.2/prod-deployment/insecure-initialize-cluster.md
@@ -6,7 +6,7 @@ On your local machine, complete the node startup process and have them join toge
 
     {% include copy-clipboard.html %}
     ~~~ shell
-    $ cockroach init --insecure --host=<address of any node>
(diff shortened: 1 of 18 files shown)


knz (approved):
> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10755)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @taroface)

ianjevans (approved):

PR #10759, opened 2021-06-09
v20.2.11 release notes

kathancox:


releases/v20.2.11.md
```diff
@@ -0,0 +1,128 @@
...
+- Improved transaction cleanup for disconnected clients to reduce intent buildup. [#65384][#65384]
+- [Scheduled backups](../v20.2/manage-a-backup-schedule.html) with [interleaved tables](../v20.2/interleave-in-parent.html) can now be created with the `include_deprecated_interleaves` option. [#65730][#65730]
+- Calling [`get_bit` or `set_bit`](../v20.2/functions-and-operators.html#string-and-byte-functions) on a byte array argument now goes to the correct index of the underlying bitstring, which matches the Postgres behavior. [#65787][#65787]
+- Previously, [`ALTER DATABASE ... CONVERT TO SCHEMA`](../v20.2/alter-database.html) could potentially leave the schema with invalid privileges thus causing the privilege descriptor to be invalid. [#65813][#65813]
+- Fixed a scenario in which a rapid sequence of splits could trigger a storm of Raft snapshots. This would be accompanied by log messages of the form "would have dropped incoming MsgApp, but allowing due to ...", which tended to occur as part of [`RESTORE`](../v20.2/restore.html)/[`IMPORT`](../v20.2/import.html) operations. [#65500][#65500]
```
mikeCRL, line 77:
```suggestion
- Fixed a scenario in which a rapid sequence of splits could trigger a storm of Raft snapshots. This would be accompanied by log messages of the form `would have dropped incoming MsgApp, but allowing due to ...`, which tended to occur as part of [`RESTORE`](../v20.2/restore.html)/[`IMPORT`](../v20.2/import.html) operations. [#65500][#65500]
```

releases/v20.2.11.md
```diff
@@ -0,0 +1,128 @@
...
+- Fixed the error classification for duplicate index names where the later index was a [`UNIQUE`](../v20.2/unique.html) index. [#64002][#64002]
+- Fixed the error classification for [`ALTER TABLE ... ADD CONSTRAINT ... UNIQUE`](../v20.2/alter-table.html) with the same name as an existing index. [#64002][#64002]
+- Fixed a bug in [`SHOW RANGES`](../v20.2/show-ranges.html) that misattributed localities to nodes when using multiple stores. [#66038][#66038]
+- Fixed a bug where [incremental backups](../v20.2/take-full-and-incremental-backups.html) with [`revision_history`](../v20.2/backup.html#options) would sometimes fail if a view was altered between incremental backups. [#66072][#66072]
+
+### Performance improvements
+
+- Fixed an issue where the [storage layer](../v20.2/architecture/storage-layer.html) would under some rare write-heavy workloads start off by writing SSTable files that were too small to manage effectively. The storage layer now creates fewer, larger files under those workloads, in line with expected behavior. [#65481][#65481]
```
mikeCRL, line 87:
```suggestion
- Fixed an issue where the [storage layer](../v20.2/architecture/storage-layer.html) would under some rare write-heavy workloads start off by writing `SSTable` files that were too small to manage effectively. The storage layer now creates fewer, larger files under those workloads, in line with expected behavior. [#65481][#65481]
```

mikeCRL: It's not clear to me if the files were too small for CRDB to manage effectively, or for operators to manage effectively, and what we might mean by effectively.

I'd want to better understand the kind of impact in order to determine whether this would be best reclassified as a bug fix. It starts with "Fixed an issue..." which is already a flag that it could be better as a bug fix, but I'm not sure.

releases/v20.2.11.md
```diff
@@ -0,0 +1,128 @@
...
+- Fixed an issue in the optimizer that prevented spatial predicates of the form `(column && value) = true` from being index-accelerated. These queries can now use a [spatial index](../v20.2/spatial-indexes.html), if one is available. [#65988][#65988]
```
mikeCRL, line 88: From reading the original issue, this seems like a bug, but the effect of fixing it is a performance improvement, so it's hard to classify. I lean 'bug fix' because it did not use the index, as we would have intended. If you feel that it's worth reconsidering, please see the original issue https://github.com/cockroachdb/cockroach/issues/65684 and feel free to make a call or ask Ian for an opinion.

mikeCRL (changes requested):
Looks great. Some suggestions here for your consideration.

Please reassign me when ready for a follow-up review, and meanwhile let me know if you want to discuss anything.

releases/v20.2.11.md
kathancox: Yes, it is a bug and the engineer confirmed this and that it made it more difficult for _CockroachDB_ to manage effectively with smaller files.

releases/v20.2.11.md
```diff
@@ -55,6 +55,7 @@ $ docker pull cockroachdb/cockroach-unstable:v20.2.11
 ### Bug fixes
 
 - [`BACKUP`](../v20.2/backup.html) no longer resolves intents one by one. This eliminates the need to run a high-priority query to clean up intents to unblock `BACKUP` in case of intent buildup. [#64931][#64931]
+- Fixed a bug that could cause backups to be slower. `LockTableWaiter` now checks the finalized transaction cache before pushing transactions to avoid duplicating work. [#64999][#64999]
```
kathancox, line 58: @mikeCRL This is the bug we were talking about earlier. Both this one and the top bug fix here (64931) are related to solving that original issue (59704). I assumed we would want to keep both here, but I'm unsure?

mikeCRL: The new text is good and keeping both sounds good to me. Thanks.

releases/v20.2.11.md
```diff
@@ -0,0 +1,129 @@
...
+## June 14, 2021
+
+- For a comprehensive summary of features in v20.2, see the [v20.2 GA release notes](v20.2.0.html).
+- To upgrade to v20.2, see [Upgrade to CockroachDB v20.2](../v20.2/upgrade-cockroach-version.html).
+
+Get future release notes emailed to you:
+
+{% include marketo.html %}
+
+Get future release notes emailed to you:
```
jseldess, line 16: Need to remove the duplicate content here.

releases/v20.2.11.md
```diff
@@ -0,0 +1,129 @@
...
+
+<section class="filter-content" data-scope="windows">
+{% include windows_warning.md %}
+</section>
+
+### Docker image
+
+{% include copy-clipboard.html %}
+~~~shell
+$ docker pull cockroachdb/cockroach-unstable:v20.2.11
```
jseldess, line 38: This is a path release of a stable version, so you need to change `cockroach-unstable:` to `cockroach:`. This should be done automatically by the script when you pass the right flag.

mikeCRL (approved):
LGTM!

PR #10776, opened 2021-06-14
Add authorization info for cc sql users

gemma-shay:
closes https://cockroachlabs.atlassian.net/browse/CC-4379

### DIFF
--- a/cockroachcloud/console-access-management.md
+++ b/cockroachcloud/console-access-management.md
@@ -15,6 +15,14 @@ To switch between the organizations:
 1. [Log in](https://cockroachlabs.cloud/).
 2. From the drop-down box in the top-right corner, select the Organization you want to access.
 
+## SQL users
+
+[Console Admins](#console-admin) can [create and manage SQL users](connect-to-your-cluster.html#step-2-create-a-sql-user). A SQL user can interact with a CockroachDB database using the built-in SQL shell or through an application.
+
+SQL users created in the Console have admin privileges on the database by default, but not the Console. Therefore, anyone with the username and password of a default SQL user has privileges for all database objects across the cluster.
+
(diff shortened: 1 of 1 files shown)


emily-horing (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10776)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay)

___
*[cockroachcloud/console-access-management.md, line 22 at r2](https://reviewable.io/reviews/cockroachdb/docs/10776#-McLVYm85bLTFhZVobCT:-McLVYm85bLTFhZVobCU:bcxfwg2) ([raw file](https://github.com/cockroachdb/docs/blob/8b2a5a08822bc7fc5e12928c08cfcb404e16a194/cockroachcloud/console-access-management.md#L22)):*
> ```Markdown
> [Console Admins](#console-admin) can [create and manage SQL users](connect-to-your-cluster.html#step-2-create-a-sql-user). A SQL user can interact with a CockroachDB database using the built-in SQL shell or through an application.
> 
> SQL users created in the Console have admin privileges on the database by default, but not the Console. Therefore, anyone with the username and password of a default SQL user has privileges for all database objects across the cluster.
> ```

I'm not sure I follow what "but not in console" means. SQL users don't interact with the console, this is a concept specific to the db. I find it confusing to relate them like that

gemma-shay:
@emily-horing I was trying to clarify that being a DB admin =/= being a Console admin, but I see what you mean, I'll remove that part. Do you think anything should be added to clarify that point, or is it clear now?

emily-horing:
>  Do you think anything should be added to clarify that point, or is it clear now?

I think it's fine to just remove that.

PR #10779, opened 2021-06-15
Test htmltest IgnoreURL fix

yzdocs:


.htmltest_nightly.yml
```diff
@@ -6,6 +6,7 @@ IgnoreDirs:
   - ^docs/v1.2
 IgnoreURLs:
   - http://localhost:4000
-  - https://github.com
+  - "https://github.com"
```
jseldess, line 9: Based on the example in the readme, this looks like the right format now: https://github.com/wjdp/htmltest#example

yzdocs: I also had to pull in a fix to our fork of `htmltest`: https://github.com/cockroachdb/htmltest/commit/e8e1406f1f25ac02c372da23f0cf50a09d77abfc

Testing now to see if that worked.

yzdocs: I think the Netlify build cache may be preventing the `htmltest` changes I mentioned above from being applied.

yzdocs: I've ruled out the build cache as a factor. It seems that the underlying code in `htmltest` isn't working as expected, even with the fix I mentioned trying to pull in above.

yzdocs: Nevermind, I blundered! Will update soon.

Issue #10455, opened 2021-04-29
sql: new multi-region ORM guidance


awoods187:
In general, we encourage DDL to be done outside of ORMs (especially DDL requiring special CRDB only features). The new multi-region features largely are contained to DDL (and the changes to DML are automatic like homing on insert). However, the one area in which it can be a bit tricky is for regional by row tables using the crdb_region column. How should ORMs map this entity as a user-defined ENUM? You can model your data this way from the start, but what about upgrading from a single-region to multi-region? It's likely that this will be mitigated by user-defined columns set as the crdb_region column https://www.cockroachlabs.com/docs/v21.1/set-locality#regional-by-row but we should still provide guidance for each supported ORM.

I think this likely falls into @ericharmeling area given his work with MovR and ORM expertise but @rmloveland should be aware of it and we are both available for support. cc @vy-ton  for awareness

nstewart:
> It's likely that this will be mitigated by user-defined columns set as the crdb_region column

One tricky thing about the user-defined column (also a problem with the hidden column) is that those columns still have to be of type `crdb_internal_region`, which we need to teach an ORM how to understand.

vy-ton:
For Hibernate, I followed [this tutorial](https://thorben-janssen.com/hibernate-enum-mappings/#customizedMapping) to map to an `ENUM` type

ericharmeling:
For SQLAlchemy, it looks like you can just use an ENUM as well, but you need to specify the right parameters to make sure that it's not attempting to use the native postgres ENUM type.

[The SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Enum.__init__) suggest it would look something like:

```
from sqlalchemy.types import Enum

crdb_internal_regions = Enum(regions_list, name='crdb_internal_regions', native_enum=False)
```

I think the crdb_internal_regions is just implemented as a string enum in cockroach... but I could be wrong? @otan

ericharmeling:
So here are two general approaches to representing the `crdb_internal_region` ENUM in an ORM **for DML queries**:

**1. Just use string-typed columns**

I have found this approach works fine in SQLAlchemy when reading/writing against a table with an existing `crdb_region` column:

- Add `crdb_region` to an existing table object as a string-typed column.
- Use string literals as `crdb_region` values in read/write operations.

[...]

PR #10796, opened 2021-06-16
Added guidance for scaling applications to multiple regions

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/10455.

This PR introduces a new page with guidance on scaling from single-region to multi-region. 

The page includes some general guidance on representing `crdb_region` columns in an ORM, for DML statements, with an example for SQLAlchemy. We can add other examples (notably, in Hibernate) after publishing this page.

### DIFF
--- a/_includes/sidebar-data-v21.1.json
+++ b/_includes/sidebar-data-v21.1.json
@@ -255,6 +255,12 @@
                     "urls": [
                       "/${VERSION}/spatial-indexes.html"
                     ]
+                  },
+                  {
+                    "title": "Scale to Multi-region",
+                    "urls": [
+                      "/${VERSION}/multiregion-scale-application.html"
+                    ]
                   }
                 ]
               }
--- a/v21.1/multiregion-scale-application.md
+++ b/v21.1/multiregion-scale-application.md
@@ -1,75 +1,77 @@
 ---
-title: Scale a Single-region Database Schema to Multiple Regions
-summary: Learn how to migrate a single-region application to a multi-region application.
+title: Scale to Multiple Regions
+summary: Learn how to scale a single-region application to a multi-region application.
 toc: true
 ---
 
-This page provides guidance for scaling an application built on a single-region database schema to multiple regions.
+This page provides guidance for scaling a single-region application to a multi-region application.
 
 Before reading this page, we recommend reviewing [CockroachDB's multi-region capabilities](multiregion-overview.html).
 
 ## Overview
 
-To scale to multiple regions:
+Scaling an application from a single region to multiple regions consists of:
 
-[1.](#scale-database-deployments) Scale the deployment of your database to multiple regions.
+- [Scaling the database](#scale-the-database), which includes transforming the database schema into a multi-region schema and adding new nodes to a CockroachDB cluster.
 
-[2.](#scale-application-deployments) Scale the deployment of your application to multiple regions.
+- [Scaling the application](#scale-the-application), which includes deploying the application in new regions and, if necessary, updating the application code to work with the multi-region database schema.
 
-[3.](#transform-a-single-region-schema-into-a-multi-region-schema) Migrate the database schema from a single-region schema to a multi-region schema.
+## Scale the database
 
-[4.](#upgrade-read-write-operations) If necessary, upgrade the read/write operations in the application to work with your new database schema.
+To scale a database from a single region to multiple regions:
 
-## Scale database deployments
+1. Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to the single region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
 
-In order to use CockroachDB's [multi-region capabilities](multiregion-overview.html), your CockroachDB cluster must have nodes deployed in multiple regions, with different [regional node localities](multiregion-overview.html#cluster-regions).
+    Setting the primary database region before adding new regional nodes to the cluster prevents CockroachDB from [rebalancing row replications](architecture/replication-layer.html#leaseholder-rebalancing) across all regions each time a node is added in a new region.
 
-For instructions on adding nodes to an existing cluster, see one of the following pages:
+    {{site.data.alerts.callout_info}}
+    Executing `ALTER` statements performs a [schema migration](online-schema-changes.html) on the cluster. If you are using a schema migration tool, you will need to execute these statements as raw SQL, as the [multi-region SQL syntax](multiregion-overview.html) is specific to CockroachDB.
 
-- For managed CockroachCloud deployments, see [Cluster Management](../cockroachcloud/cluster-management.html).
-- For orchestrated deployments, see [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](orchestrate-cockroachdb-with-kubernetes-multi-cluster.html).
-- For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
+    Here are some simple tutorials on executing schema migrations against CockroachDB clusters:
+    - [Migrate CockroachDB Schemas with Liquibase](liquibase.html).
+    - [Migrate CockroachDB Schemas with Flyway](flyway.html).
+    - [Execute SQL statements from a file](cockroach-sql.html#execute-sql-statements-from-a-file) and [Change and Remove Objects in a Database Schema](schema-design-update.html).
+    {{site.data.alerts.end}}
 
-## Scale application deployments
+1. Add nodes to the cluster in new regions.
 
-To take full advantage of the latency and resiliency improvements of a multi-region database deployment, you need to scale the application deployments to multiple regions as well. To limit the latency between the application and the database, each deployment of the application must communicate with the closest database deployment.
+    For instructions on adding nodes to an existing cluster, see one of the following pages:
+    - For managed CockroachCloud deployments, see [Cluster Management](../cockroachcloud/cluster-management.html).
+    - For orchestrated deployments, see [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](orchestrate-cockroachdb-with-kubernetes-multi-cluster.html).
+    - For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
+
+    {{site.data.alerts.callout_info}}
+    You must specify a [regional locality](cockroach-start.html#locality) for each node at startup. These regional localities are represented as [cluster regions](multiregion-overview.html#cluster-regions) in the cluster.
+    {{site.data.alerts.end}}
+
+1. Use an [`ALTER DATABASE ... ADD REGIONS`](add-region.html) statement to add the new regions to your database. Only cluster regions (i.e., regional localities specified at cluster startup) can be added as [database regions](multiregion-overview.html#database-regions).
+
+1. Use [`ALTER DATABASE ... SURVIVE ... FAILURE`](survive-failure.html) statements to set your database's [survival goals](multiregion-overview.html#survival-goals).
+
+1. Use [`ALTER TABLE ... SET LOCALITY`](set-locality.html) statements to set [table localities](multiregion-overview.html#table-locality) for each table.
+
+## Scale the application
+
+Scaling application deployments in multiple regions can greatly improve latency for the end-user of the application.
 
 For guidance on connecting to CockroachDB from an application deployment, see one of the following pages:
 
 - For connecting to managed, CockroachCloud deployments, see [Connect to Your CockroachCloud Cluster](../cockroachcloud/connect-to-your-cluster.html) and [Connect to the Database (CockroachCloud)](connect-to-the-database-cockroachcloud.html).
 - For connecting to a standard CockroachDB deployment, see [`cockroach sql`](cockroach-sql.html) and [Connect to the Database](connect-to-the-database.html).
 
-For details on configuring database connections for individual application deployments, consult your cloud provider's documentation. For an example using Google Cloud services, see [Multi-region Application Deployment](multi-region-deployment.html).
+To limit the latency between the application and the database, each deployment of the application should communicate with the closest database deployment. For details on configuring database connections for individual application deployments, consult your cloud provider's documentation. For an example using Google Cloud services, see [Multi-region Application Deployment](multi-region-deployment.html).
 
-## Transform a single-region schema into a multi-region schema
+{{site.data.alerts.callout_info}}
+A multi-region application deployment does not require a multi-region database deployment. Deploying a global application in multiple regions can yield significant latency benefits for the end user, even if you have not yet scaled your database in multiple regions. For an example, see [Reducing Multi-Region Latency with Follower Reads](https://www.cockroachlabs.com/blog/follower-reads/#:~:text=Deployment%202%3A%20Global%20Application%20Deployment%2C%20No%20Follower%20reads).
 
-To transform a single-region database schema into a multi-region database schema:
+If you do scale the application first, make sure that you reconfigure each application deployment to communicate with the closest database deployment after deploying the database in multiple regions.
+{{site.data.alerts.end}}
 
-- [Choose a multi-region configuration](choosing-a-multi-region-configuration.html).
-- Use `ALTER DATABASE` and `ALTER TABLE` statements to set the [database regions](multiregion-overview.html#database-regions), [survival goals](multiregion-overview.html#survival-goals), and [table localities](multiregion-overview.html#table-locality), accordingly to your selected multi-region configuration.
+### Update the application code for multi-region
 
-For SQL syntax reference documentation for the multi-region `ALTER DATABASE` statements, see the following pages:
+For most table localities, including the default locality `LOCALITY REGIONAL BY TABLE IN PRIMARY REGION`, *you do not need to update your application code after migrating your database schema for multi-region*. CockroachDB automatically optimizes queries against multi-region databases, based on the regional locality of the node executing the query, and on the multi-region configuration of the database. For more details, see [Regional Tables](regional-tables.html#regional-by-row-tables). For an extended example, see [Develop and Deploy a Global Application: Create a Multi-region Database Schema](movr-flask-database.html).
 
-- [`SET PRIMARY REGION`](set-primary-region.html)
-- [`ADD REGION`](add-region.html)
-- [`SET LOCALITY`](set-locality.html)
-- [`SURVIVE ... FAILURE`](survive-failure.html)
-
-Executing `ALTER` statements performs a [schema migration](online-schema-changes.html) on the cluster. As a best practice, we recommend performing all schema migrations (including multi-region transformations) with a schema migration tool.
-
-Note that, because the `ALTER` statements listed above are specific to CockroachDB syntax, you will have to execute these statements as raw SQL in schema migration tools.
-
-Here are some simple tutorials on executing schema migrations against CockroachDB clusters:
-
-- [Migrate CockroachDB Schemas with Liquibase](liquibase.html).
-- [Migrate CockroachDB Schemas with Flyway](flyway.html).
-- [Execute SQL statements from a file](cockroach-sql.html#execute-sql-statements-from-a-file) and [Change and Remove Objects in a Database Schema](schema-design-update.html).
-
-## Upgrade read/write operations
-
-For most multi-region configurations, including the default configuration (`LOCALITY REGIONAL BY TABLE IN PRIMARY REGION` and `SURVIVE ZONE FAILURE`), you do not need to update any client-side read/write operations after migrating your database schema for multi-region features. CockroachDB automatically optimizes queries against multi-region databases, based on the locality of the node executing the query, and on the multi-region configuration of the database.
-
-However, there are some scenarios in which you might need to update the read/write operations in your application's persistence layer. For example:
+However, there are some scenarios in which you might need to update the SQL operations in your application. For example:
 
 - If a table has a `REGIONAL BY ROW AS <custom_region_column>` table locality, and you want to explicitly insert regional values into a table, as shown in [Low Latency Reads and Writes in a Multi-Region Cluster](demo-low-latency-multi-region-deployment.html#configure-regional-by-row-tables).
 - If a table has a `REGIONAL BY ROW` locality, and you want to update the hidden `crdb_region` column of existing rows in the table based on some other column value, as shown in [Set the table locality to `REGIONAL BY ROW`](set-locality.html#set-the-table-locality-to-regional-by-row).
@@ -77,9 +79,9 @@ However, there are some scenarios in which you might need to update the read/wri
 
 In all of these scenarios, statements reference the column that tracks the region for each row in a `REGIONAL BY ROW` locality. This column can be a custom column of the built-in `ENUM` type `crdb_internal_region`, or it can be the default, hidden [`crdb_region` column](set-locality.html#crdb_region).
 
-### Representing `crdb_region` in ORMs
+#### Representing `crdb_region` in ORMs
 
-If your application's ORM framework supports enumerated data types, and you need to explicitly reference the `crdb_region` column (or a custom region column) in a read/write operation, you should do the following:
+If your application's ORM framework supports enumerated data types, and you need to explicitly reference the `crdb_region` column (or a custom region column) in a SQL operation in your application code, you should do the following:
 
 - Using the ORM framework's type library, create an enumerated type, with the type name `crdb_internal_region` and enumerated values for each of the existing regions.
 - Add a new region column to the relevant table mappings, with the new enumerated data type and the column name `crdb_region` (or the name of the custom region column of the table). Note that this column does not persist to the database; it is simply a mapping for the application.
--- a/v21.1/multiregion-scale-application.md
+++ b/v21.1/multiregion-scale-application.md
@@ -1,10 +1,10 @@
 ---
 title: Scale to Multiple Regions
-summary: Learn how to scale a single-region application to a multi-region application.
+summary: Learn how to scale a single-region application to multiple regions.
 toc: true
 ---
 
-This page provides guidance for scaling a single-region application to a multi-region application.
+This page provides guidance for scaling a single-region application to multiple regions.
 
 Before reading this page, we recommend reviewing [CockroachDB's multi-region capabilities](multiregion-overview.html).
 
@@ -12,45 +12,56 @@ Before reading this page, we recommend reviewing [CockroachDB's multi-region cap
 
 Scaling an application from a single region to multiple regions consists of:
 
-- [Scaling the database](#scale-the-database), which includes transforming the database schema into a multi-region schema and adding new nodes to a CockroachDB cluster.
+- [Scaling the database](#scale-the-database), which includes adding new nodes to a CockroachDB cluster in different regions, adding regions to the database schema, and optionally transforming the database schema to leverage multi-region table localities.
 
 - [Scaling the application](#scale-the-application), which includes deploying the application in new regions and, if necessary, updating the application code to work with the multi-region database schema.
 
 ## Scale the database
 
-To scale a database from a single region to multiple regions:
+### Step 1. Prep the database schema
 
-1. Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to the single region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
+Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to a region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
 
-    Setting the primary database region before adding new regional nodes to the cluster prevents CockroachDB from [rebalancing row replications](architecture/replication-layer.html#leaseholder-rebalancing) across all regions each time a node is added in a new region.
+Setting the primary region before adding new regional nodes to the cluster prevents CockroachDB from [rebalancing row replications](architecture/replication-layer.html#leaseholder-rebalancing) across all regions each time a node is added in a new region.
 
-    {{site.data.alerts.callout_info}}
-    Executing `ALTER` statements performs a [schema migration](online-schema-changes.html) on the cluster. If you are using a schema migration tool, you will need to execute these statements as raw SQL, as the [multi-region SQL syntax](multiregion-overview.html) is specific to CockroachDB.
+{{site.data.alerts.callout_info}}
+Executing `ALTER` statements performs a [schema migration](online-schema-changes.html) on the cluster. If you are using a schema migration tool, you will need to execute these statements as raw SQL, as the [multi-region SQL syntax](multiregion-overview.html) is specific to CockroachDB.
 
-    Here are some simple tutorials on executing schema migrations against CockroachDB clusters:
-    - [Migrate CockroachDB Schemas with Liquibase](liquibase.html).
-    - [Migrate CockroachDB Schemas with Flyway](flyway.html).
-    - [Execute SQL statements from a file](cockroach-sql.html#execute-sql-statements-from-a-file) and [Change and Remove Objects in a Database Schema](schema-design-update.html).
-    {{site.data.alerts.end}}
+Here are some simple tutorials on executing schema migrations against CockroachDB clusters:
 
-1. Add nodes to the cluster in new regions.
+- [Migrate CockroachDB Schemas with Liquibase](liquibase.html)
+- [Migrate CockroachDB Schemas with Flyway](flyway.html)
+- [Migrate CockroachDB Schemas with Alembic](alembic.html)
+- [Execute SQL statements from a file](cockroach-sql.html#execute-sql-statements-from-a-file) and [Change and Remove Objects in a Database Schema](schema-design-update.html)
+{{site.data.alerts.end}}
 
-    For instructions on adding nodes to an existing cluster, see one of the following pages:
-    - For managed CockroachCloud deployments, see [Cluster Management](../cockroachcloud/cluster-management.html).
-    - For orchestrated deployments, see [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](orchestrate-cockroachdb-with-kubernetes-multi-cluster.html).
-    - For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
+### Step 2. Scale the cluster deployment
 
-    {{site.data.alerts.callout_info}}
-    You must specify a [regional locality](cockroach-start.html#locality) for each node at startup. These regional localities are represented as [cluster regions](multiregion-overview.html#cluster-regions) in the cluster.
-    {{site.data.alerts.end}}
+Scale the cluster by adding nodes to the cluster in new regions.
 
-1. Use an [`ALTER DATABASE ... ADD REGIONS`](add-region.html) statement to add the new regions to your database. Only cluster regions (i.e., regional localities specified at cluster startup) can be added as [database regions](multiregion-overview.html#database-regions).
+For instructions on adding nodes to an existing cluster, see one of the following pages:
 
-1. Use [`ALTER DATABASE ... SURVIVE ... FAILURE`](survive-failure.html) statements to set your database's [survival goals](multiregion-overview.html#survival-goals).
+- For managed CockroachCloud deployments, see [Cluster Management](../cockroachcloud/cluster-management.html).
+- For orchestrated deployments, see [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](orchestrate-cockroachdb-with-kubernetes-multi-cluster.html).
+- For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
 
-1. Use [`ALTER TABLE ... SET LOCALITY`](set-locality.html) statements to set [table localities](multiregion-overview.html#table-locality) for each table.
+{{site.data.alerts.callout_info}}
+For orchestrated and manual deployments, you must specify a [regional locality](cockroach-start.html#locality) for each node at startup. These regional localities are represented as [cluster regions](multiregion-overview.html#cluster-regions) in the cluster.
+{{site.data.alerts.end}}
+
+### Step 3. Scale the database schema
+
+Use an [`ALTER DATABASE ... ADD REGIONS`](add-region.html) statement to add the new regions to your database. Only cluster regions (i.e., regional localities specified at cluster startup) can be added as [database regions](multiregion-overview.html#database-regions).
+
+After you add new regions to the database schema, you can optionally configure the [survival goals](multiregion-overview.html#survival-goals) and [table localities](multiregion-overview.html#table-locality) of the multi-region database:
+
+- Add [`ALTER DATABASE ... SURVIVE ... FAILURE`](survive-failure.html) statements to set your database's [survival goals](multiregion-overview.html#survival-goals).
+
+- Add [`ALTER TABLE ... SET LOCALITY`](set-locality.html) statements to set [table localities](multiregion-overview.html#table-locality) for each table.
 
 ## Scale the application
+
+### Step 1. Scale application deployments
 
 Scaling application deployments in multiple regions can greatly improve latency for the end-user of the application.
 
@@ -67,54 +78,55 @@ A multi-region application deployment does not require a multi-region database d
 If you do scale the application first, make sure that you reconfigure each application deployment to communicate with the closest database deployment after deploying the database in multiple regions.
 {{site.data.alerts.end}}
 
-### Update the application code for multi-region
+### Step 2. *(Optional)* Update the application code for multi-region
 
 For most table localities, including the default locality `LOCALITY REGIONAL BY TABLE IN PRIMARY REGION`, *you do not need to update your application code after migrating your database schema for multi-region*. CockroachDB automatically optimizes queries against multi-region databases, based on the regional locality of the node executing the query, and on the multi-region configuration of the database. For more details, see [Regional Tables](regional-tables.html#regional-by-row-tables). For an extended example, see [Develop and Deploy a Global Application: Create a Multi-region Database Schema](movr-flask-database.html).
 
 However, there are some scenarios in which you might need to update the SQL operations in your application. For example:
 
 - If a table has a `REGIONAL BY ROW AS <custom_region_column>` table locality, and you want to explicitly insert regional values into a table, as shown in [Low Latency Reads and Writes in a Multi-Region Cluster](demo-low-latency-multi-region-deployment.html#configure-regional-by-row-tables).
-- If a table has a `REGIONAL BY ROW` locality, and you want to update the hidden `crdb_region` column of existing rows in the table based on some other column value, as shown in [Set the table locality to `REGIONAL BY ROW`](set-locality.html#set-the-table-locality-to-regional-by-row).
-- If a table has a `REGIONAL BY ROW` locality, and you want to filter a selection query based on the value of the region-tracking column.
+- If a table has a `REGIONAL BY ROW` locality, and you want to update the `crdb_region` value of existing rows in the table based on some other column value, as shown in [Set the table locality to `REGIONAL BY ROW`](set-locality.html#set-the-table-locality-to-regional-by-row).
+- If a table has a `REGIONAL BY ROW` locality, and you want to filter a [selection query](select-clause.html#filter-rows) based on the `crdb_region` value.
 
 In all of these scenarios, statements reference the column that tracks the region for each row in a `REGIONAL BY ROW` locality. This column can be a custom column of the built-in `ENUM` type `crdb_internal_region`, or it can be the default, hidden [`crdb_region` column](set-locality.html#crdb_region).
 
-#### Representing `crdb_region` in ORMs
+If you need to explicitly reference the region-tracking column in a SQL operation in your application code, you should do the following:
 
-If your application's ORM framework supports enumerated data types, and you need to explicitly reference the `crdb_region` column (or a custom region column) in a SQL operation in your application code, you should do the following:
+- Verify that the region-tracking column is visible to the ORM.
 
-- Using the ORM framework's type library, create an enumerated type, with the type name `crdb_internal_region` and enumerated values for each of the existing regions.
-- Add a new region column to the relevant table mappings, with the new enumerated data type and the column name `crdb_region` (or the name of the custom region column of the table). Note that this column does not persist to the database; it is simply a mapping for the application.
-- Reference the region column in read/write operations as needed.
+    To make a hidden column visible, use an [`ALTER TABLE ... ALTER COLUMN ... SET VISIBLE` statement](alter-column.html). By default, the `crdb_region` column created by CockroachDB is hidden.
+- Using your ORM framework, sync the mapping objects in your application to reflect the latest database schema with the region-tracking column(s).
+- Reference the region-tracking column in read/write operations as needed.
 
-For example, suppose that you have a single-region table that has just been transformed into a multi-region table. In the absence of an explicit, back-filling computed column for the hidden `crdb_region` column, the following code (written in Python, with [SQLAlchemy](https://www.sqlalchemy.org/)) updates the `crdb_region` values in rows that were inserted before the multi-region transformation was executed:
+For example, suppose that you have a single-region table called `users` that has just been transformed into a multi-region table with a `REGIONAL BY ROW` locality. When the application was first deployed, this table had no region-tracking column. During the multi-region database schema transformation, CockroachDB automatically created a hidden `crdb_region` column to track the region of each row.
 
-1. Create the `ENUM` type:
+In the absence of an explicit, back-filling computed column for the hidden `crdb_region` column, there is no way for CockroachDB to determine the region for old rows of data. The following steps update the `crdb_region` values in rows that were inserted before the multi-region transformation, based on the values of a `city` column:
+
+1. Make `crdb_region` visible in the relevant `REGIONAL BY ROW` table(s):
 
     {% include copy-clipboard.html %}
-    ~~~ python
-    from sqlalchemy.types import Enum
-    ...
-
-    region_list = list(tup[0] for tup in session.execute(text('SELECT region FROM [SHOW REGIONS]')).fetchall())
-    region_enum = Enum(*region_list, name='crdb_internal_region', create_type=False, native_enum=False)
+    ~~~ sql
+    ALTER TABLE users ALTER COLUMN crdb_region SET VISIBLE;
     ~~~
 
-2. Update the mapping in the application code:
+1. Update the table mappings in the application code (written in Python, with [SQLAlchemy](https://www.sqlalchemy.org/)):
 
     {% include copy-clipboard.html %}
     ~~~ python
-    crdb_region = Column('crdb_region', region_enum)
-    MyTable.append_column(crdb_region)
+    from models import Base
+
+    ...
+
+    Base.metadata.reflect(bind=self.engine, extend_existing=True, autoload_replace=True)
     ~~~
 
     {{site.data.alerts.callout_info}}
-    SQLAlchemy allows you to append columns to a table-mapped object with the `sqlalchemy.Table` class method [`append_column()`](https://docs.sqlalchemy.org/en/13/core/metadata.html#sqlalchemy.schema.Table.append_column). If your ORM framework does not support updating mapping objects dynamically, you might need to add the column to the table-mapping class definition and reinstantiate the object.
+    SQLAlchemy allows you to update all table mappings to reflect the database with the `sqlalchemy.schema.MetaData` class method [`reflect()`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.reflect). If your ORM framework does not support updating mapping objects dynamically, you might need to add the column to the table-mapping class definition as a `String`-typed column and reinstantiate the object.
     {{site.data.alerts.end}}
 
-3. Reference the column value as needed.
+1. Reference the column value as needed.
 
-    Here is an example function that updates the region value in a given table, using the values of a `city` column:
+    Here is an example function that updates the region value in a given table, using the values of the `city` column:
 
     {% include copy-clipboard.html %}
     ~~~ python
@@ -124,19 +136,12 @@ For example, suppose that you have a single-region table that has just been tran
     def update_region(engine, table, region, cities):
 
         def update_region_helper(session, table, region, cities):
-            query = table.update().where(column('city').in_(cities)).values({crdb_region: region})
+            query = table.update().where(column('city').in_(cities)).values({'crdb_region': region})
             session.execute(query)
 
         run_transaction(sessionmaker(bind=engine),
                         lambda session: update_region_helper(session, table, region, cities))
     ~~~
-
-If your ORM framework does not support `ENUM` data types, or if you run into type compatibility issues, you can alternatively try the following approach:
-
-- Add `crdb_region` (or the custom region-tracking column) to an existing table object as a *string*-typed column.
-- Validate the string input for `crdb_region` values, using one of the following approaches:
-    - Implement client-side string validation to check if the string input matches one of the allowed enumerated types.
-    - Add some exception handling to catch data errors (e.g., `sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation)` in SQLAlchemy).
 
 ## See also
 
--- a/v21.1/multiregion-scale-application.md
+++ b/v21.1/multiregion-scale-application.md
@@ -18,7 +18,7 @@ Scaling an application from a single region to multiple regions consists of:
 
 ## Scale the database
 
-### Step 1. Prep the database schema
+### Step 1. Prep the database
 
 Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to a region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
 
--- a/v21.1/multiregion-scale-application.md
+++ b/v21.1/multiregion-scale-application.md
@@ -12,17 +12,17 @@ Before reading this page, we recommend reviewing [CockroachDB's multi-region cap
 
 To scale to multiple regions:
 
-[1.](#scale-deployments) Scale both the deployment of your database and the deployment of your application to multiple regions.
+[1.](#scale-database-deployments) Scale the deployment of your database to multiple regions.
 
-[2.](#transform-a-single-region-schema-into-a-multi-region-schema) Migrate the database schema from a single-region schema to a multi-region schema, using a schema migration tool.
+[2.](#scale-application-deployments) Scale the deployment of your application to multiple regions.
 
-[3.](#upgrade-read-write-operations) If necessary, upgrade the read/write operations in the application to work with your new database schema.
+[3.](#transform-a-single-region-schema-into-a-multi-region-schema) Migrate the database schema from a single-region schema to a multi-region schema.
 
-## Scale deployments
+[4.](#upgrade-read-write-operations) If necessary, upgrade the read/write operations in the application to work with your new database schema.
 
-### Scale database deployments
+## Scale database deployments
 
-In order to use CockroachDB's [multi-region capabilities](multiregion-overview.html), your CockroachDB cluster must have nodes deployed in multiple regions, with different [regional localities](multiregion-overview.html#cluster-regions).
+In order to use CockroachDB's [multi-region capabilities](multiregion-overview.html), your CockroachDB cluster must have nodes deployed in multiple regions, with different [regional node localities](multiregion-overview.html#cluster-regions).
 
 For instructions on adding nodes to an existing cluster, see one of the following pages:
 
@@ -30,7 +30,7 @@ For instructions on adding nodes to an existing cluster, see one of the followin
 - For orchestrated deployments, see [Orchestrate CockroachDB Across Multiple Kubernetes Clusters](orchestrate-cockroachdb-with-kubernetes-multi-cluster.html).
 - For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
 
-### Scale application deployments
+## Scale application deployments
 
 To take full advantage of the latency and resiliency improvements of a multi-region database deployment, you need to scale the application deployments to multiple regions as well. To limit the latency between the application and the database, each deployment of the application must communicate with the closest database deployment.
 
--- a/v21.1/multiregion-scale-application.md
+++ b/v21.1/multiregion-scale-application.md
@@ -1,0 +1,142 @@
+---
+title: Scale a Single-region Database Schema to Multiple Regions
+summary: Learn how to migrate a single-region application to a multi-region application.
+toc: true
+---
+
+This page provides guidance for scaling an application built on a single-region database schema to multiple regions.
+
+Before reading this page, we recommend reviewing [CockroachDB's multi-region capabilities](multiregion-overview.html).
+
+## Overview
+
+To scale to multiple regions:
+
+[1.](#scale-deployments) Scale both the deployment of your database and the deployment of your application to multiple regions.
+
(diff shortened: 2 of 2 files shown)


awoods187 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj69DqDH6kGOLP0676:-Mcj69DqDH6kGOLP0677:b-a0vcrz) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L15)):*
> ```Markdown
> To scale to multiple regions:
> 
> [1.](#scale-deployments) Scale both the deployment of your database and the deployment of your application to multiple regions.
> ```

Im surprised to see this as one step. I think a more common pattern is to scale the app to multiple regions and then scale the db. Scaling the db also has some nuances to the order in which you do things (like adding the primary region before adding nodes with localities in other regions)
___
*[v21.1/multiregion-scale-application.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj5uqp29Nyth9JYDb0:-Mcj5uqp29Nyth9JYDb1:bs4kl9s) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L17)):*
> ```Markdown
> [1.](#scale-deployments) Scale both the deployment of your database and the deployment of your application to multiple regions.
> 
> [2.](#transform-a-single-region-schema-into-a-multi-region-schema) Migrate the database schema from a single-region schema to a multi-region schema, using a schema migration tool.
> ```

I like including the migration tool but it isn't strictly necessary. Should we allude to that more explicitly?
___
*[v21.1/multiregion-scale-application.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6fpeEeQWLdiFvIVq:-Mcj6fpeEeQWLdiFvIVr:boziql3) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L23)):*
> ```Markdown
> ## Scale deployments
> 
> ### Scale database deployments
> ```

the first step is to alter your database and add the primary region = the current single region you operate within. This will keep your read/write traffic the same once you add the other regions. if you didn't do this step first, crdb would try to spread replicas and leaseholders for diversity and availability, potentially negatively impact your existing traffic. However, you if add the primary region first, then add the other regions, no leaseholder or replicas will move holding perf steady. Then, you can upgrade tables to RBR or Global.
___
*[v21.1/multiregion-scale-application.md, line 25 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7YtcAvW4iH6n9riK:-Mcj6SER1wd2dOXDkheB:b8m58an) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L25)):*
> ```Markdown
> ### Scale database deployments
> 
> In order to use CockroachDB's [multi-region capabilities](multiregion-overview.html), your CockroachDB cluster must have nodes deployed in multiple regions, with different [regional localities](multiregion-overview.html#cluster-regions).
> ```

should this be node localities? or regional node localities? 
___
*[v21.1/multiregion-scale-application.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6Yn14yULBv4WckJJ:-Mcj6Yn14yULBv4WckJK:bb2tp10) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L33)):*
> ```Markdown
> - For manual deployments, see [`cockroach start`](cockroach-start.html) and [Manual Deployment](manual-deployment.html).
> 
> ### Scale application deployments
> ```

i think this should be listed first since the user would likely do it first
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-Mcj7L5Y0ValW7Tw-b3A:buq9jrx) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*
> ```Markdown
> - [Execute SQL statements from a file](cockroach-sql.html#execute-sql-statements-from-a-file) and [Change and Remove Objects in a Database Schema](schema-design-update.html).
> 
> ## Upgrade read/write operations
> ```

we've been calling these table localities 
___
*[v21.1/multiregion-scale-application.md, line 72 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7sJkBMBDw4a493My:-Mcj7sJkBMBDw4a493Mz:b-m2bvr0) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L72)):*
> ```Markdown
> For most multi-region configurations, including the default configuration (`LOCALITY REGIONAL BY TABLE IN PRIMARY REGION` and `SURVIVE ZONE FAILURE`), you do not need to update any client-side read/write operations after migrating your database schema for multi-region features. CockroachDB automatically optimizes queries against multi-region databases, based on the locality of the node executing the query, and on the multi-region configuration of the database.
> 
> However, there are some scenarios in which you might need to update the read/write operations in your application's persistence layer. For example:
> ```

I think its strange to jump over the cool auto-homing and list a bunch of examples of when you wouldn't do that without selling it/showing an example of it for inserts and selects

ericharmeling (commented):
TFTR @awoods187 !

I have some follow-up questions, for clarification. See below. (Apologies if I'm being a bit dense :laughing:.)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796#-:-McnpxFqBDt31sytCYL7:b-9o1rd2)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @rmloveland)

___
*[v21.1/multiregion-scale-application.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj69DqDH6kGOLP0676:-Mcnmf25D13ObHX_FjBC:bur7e5w) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L15)):*
> Im surprised to see this as one step.

Fair enough! I've split up the app and db deployment steps.

>  I think a more common pattern is to scale the app to multiple regions and then scale the db.

This doesn't make much sense to me. You configure which node to connect to when you deploy the application. If you scale your application first, and then you scale your database, you need to re-configure the application (and possibly even re-deploy it, depending on the orchestration tool) to connect to the closest db node after you have already deployed and configured the application once. 

> Scaling the db also has some nuances to the order in which you do things (like adding the primary region before adding nodes with localities in other regions)

So I think we might be talking about different things when we say "scaling the db" here. But maybe I'm misunderstanding what you're saying?

When I say "scale the database deployment", I just mean "add nodes to the cluster" (no multi-region syntax/schema changes included yet). After you add nodes with regional localities, then you can transform the schema for multi-region (i.e., use the multi-region syntax to add the database regions, etc.). I've made these two entirely separate steps, as one of the steps ("Scale database deployments") is about navigating orchestration tools/CockroachCloud to get a cluster deployed in multiple regions, and the other step ("Transform a single-region schema into a multi-region schema") is about executing schema changes in SQL.
___
*[v21.1/multiregion-scale-application.md, line 17 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj5uqp29Nyth9JYDb0:-McnmLa7-DlG5rHNrJMT:b57sdmd) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L17)):*

I like including the migration tool but it isn't strictly necessary. Should we allude to that more explicitly?
</blockquote></details>

I'll just remove it from here. The dedicated section ("Transform a single-region schema into a multi-region schema") mentions the migration tool.
___
*[v21.1/multiregion-scale-application.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6fpeEeQWLdiFvIVq:-McnxO243C9E3lTvtBx_:b-6gxvxl) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L23)):*

the first step is to alter your database and add the primary region = the current single region you operate within. This will keep your read/write traffic the same once you add the other regions. if you didn't do this step first, crdb would try to spread replicas and leaseholders for diversity and availability, potentially negatively impact your existing traffic. However, you if add the primary region first, then add the other regions, no leaseholder or replicas will move holding perf steady. Then, you can upgrade tables to RBR or Global.
</blockquote></details>

So I thought we might be talking about different things when we say "scale database deployments" (see my comment above), but now I'm not so sure.

I understand the workflow to be:

1. Scale database deployment (i.e., add nodes to the cluster with regional localities)
2. Scale application deployment (e.g., deploy Docker images in new regions, with configurations to connect to the closest db node)
3. Execute multi-region DDL statements (i.e., `ALTER DATABASE ... ADD PRIMARY REGION`, `ALTER DATABASE ... ADD REGION`, etc)
4. Update application, if necessary

It seems to me that you are suggesting the following workflow (but I could totally be misinterpreting what you are saying):

1. Scale the application (e.g., deploy Docker images in new regions, with configurations to connect to the *same*, single-region db node)
2. Execute a single `ALTER DATABASE ... ADD PRIMARY REGION` statement on a single-region CDB cluster
3. Scale the database deployment (i.e., add nodes to the cluster with regional localities)
4. Execute the rest of the multi-region DDL statements (`ALTER DATABASE ... ADD REGION`, etc.)
5. Update each application configuration to connect to the closest db node (e.g., update Docker images with new connection strings, and then re-deploy the images)
6. Update application, if necessary.

Is that correct?
___
*[v21.1/multiregion-scale-application.md, line 24 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mco9XuZ0sBUY-II38Ic:-Mco9XuZ0sBUY-II38Id:b-jhwu1p) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L24)):*
> the first step is to alter your database and add the primary region = the current single region you operate within. This will keep your read/write traffic the same once you add the other regions. 

A couple questions:

 - By "first step", do you mean before you have added any new nodes to your cluster (in new regions)?
- By "add other regions", do you mean "add other database regions" or "add other cluster regions" (i.e., "add new nodes to the cluster with other regional localities")? IIUC, the order of adding database regions is enforced in CRDB multi-region syntax. That is, you must add the primary database region before other database regions can be added.
___
*[v21.1/multiregion-scale-application.md, line 25 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7YtcAvW4iH6n9riK:-McnxUNJ5g4IwJXsTQGK:b-4zj5er) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L25)):*

should this be node localities? or regional node localities? 
</blockquote></details>

Updated to read "regional node localities".
___
*[v21.1/multiregion-scale-application.md, line 33 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6Yn14yULBv4WckJJ:-Mcnx-O99Av-EzXeK8eZ:b-upj58l) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L33)):*

i think this should be listed first since the user would likely do it first
</blockquote></details>

See my comments above.
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-McnzJ7p-lMuLr5k1qVE:b-wpo9c6) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*

we've been calling these table localities 
</blockquote></details>

Sorry - I'm not sure I know what you are referring to here.

I am trying to use the terminology used here: https://www.cockroachlabs.com/docs/stable/choosing-a-multi-region-configuration.html

But maybe I am missing something?
___
*[v21.1/multiregion-scale-application.md, line 72 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7sJkBMBDw4a493My:-Mcnzc161_YUStVLQIWQ:bp9nu2u) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L72)):*

I think its strange to jump over the cool auto-homing and list a bunch of examples of when you wouldn't do that without selling it/showing an example of it for inserts and selects
</blockquote></details>

Fair enough! I can add some examples here (or at least refer to some existing examples).

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 24 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mco9XuZ0sBUY-II38Ic:-McopA21BoJFccTjjGmW:b9gor8v) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L24)):*
Not Andy (obv) but added my understanding below re:

> By "first step", do you mean before you have added any new nodes to your cluster (in new regions)?

I think the answer is "yes": if you (1) add a bunch of new nodes at time T without changing your multi-region config (really, constraints), the cluster will do a bunch of work rebalancing among those nodes based on CRDB's "default" set of constraints C in place at time T, and then when you (2) add a primary region at a later time T+1, yielding a new "narrower" set of constraints C', the cluster AGAIN has to do a bunch of work rebalancing etc. to meet the new constraint set C' which is now in place at time T+1.   The fact that the constraints started out "wide" and got "narrow" results in a lot of data movement etc.

By contrast, if you add a primary region first thing at time T when you have one node, the initial set of constraints is already "narrowed" significantly, so any new constraints you add at time T+1, ..., etc when you add a bunch of new nodes will be validated against the already existing "narrower" constraints and the data will go where you want it more quickly without so much rebalancing (since you avoided the wide constraint -> narrow constraint change described above)

Basically it's like a constraint propagation system that will be faster if you narrow the set of constraints as soon as possible during cluster setup and before adding more nodes, so it has a narrower range of things to consider (thus resulting in less churn / data movement / network traffic / etc)

At least this is my understanding of it
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-McoshVyECWMiatlp10z:b-8o2606) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*

Sorry - I'm not sure I know what you are referring to here.

I am trying to use the terminology used here: https://www.cockroachlabs.com/docs/stable/choosing-a-multi-region-configuration.html

But maybe I am missing something?
</blockquote></details>

the phrase we have been using for the `SET LOCALITY FOO ...` stuff is table localities

e.g. I might call this section "set table localities"

and in the paragraph below I might say something like "for most ... configurations ..., you do not need to update the default table localities"

a reason for this obliqueness is in keeping with the declarative nature of SQL and also the multi-region stuff in general: it's not like a statement in Python or or ALGOL or whatever where I say "set x=5", it's more like operating a constraint propagation system: "make this table regional in us-west-1", and then the r/w perf characteristics of the table's performance fall out of what CRDB does under the hood based on the constraints we have given it, depending on where the client is accessing the table from.  

So we are not really "upgrading" or "updating read/write operations" since what we are really doing is at arm's length: changing the "locality" constraint of the table and then CRDB does "Magic(tm)" to make it so that reads / writes from specific regions have certain characteristics.

Basically it's more fuzzy/declarative and less like a "regular" programming thing

I hope that makes sense - the core of actionable info is in the first 3 sentences above, the rest is my possibly unhelpful faux explication

awoods187 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @ericharmeling, and @rmloveland)

___
*[v21.1/multiregion-scale-application.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj69DqDH6kGOLP0676:-MctQ9eZ-LP3aOsPy-PY:b-drlyk6) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L15)):*

> Im surprised to see this as one step.

Fair enough! I've split up the app and db deployment steps.

>  I think a more common pattern is to scale the app to multiple regions and then scale the db.

This doesn't make much sense to me. You configure which node to connect to when you deploy the application. If you scale your application first, and then you scale your database, you need to re-configure the application (and possibly even re-deploy it, depending on the orchestration tool) to connect to the closest db node after you have already deployed and configured the application once. 

> Scaling the db also has some nuances to the order in which you do things (like adding the primary region before adding nodes with localities in other regions)

So I think we might be talking about different things when we say "scaling the db" here. But maybe I'm misunderstanding what you're saying?

When I say "scale the database deployment", I just mean "add nodes to the cluster" (no multi-region syntax/schema changes included yet). After you add nodes with regional localities, then you can transform the schema for multi-region (i.e., use the multi-region syntax to add the database regions, etc.). I've made these two entirely separate steps, as one of the steps ("Scale database deployments") is about navigating orchestration tools/CockroachCloud to get a cluster deployed in multiple regions, and the other step ("Transform a single-region schema into a multi-region schema") is about executing schema changes in SQL.
</blockquote></details>

Well, it is a question of timing. You can actually get a lot of value out of deploying your app to multiple regions without deploying your db. You might do that for a few months before you go ahead and deploy your db (since it is so easy to do). With our new abstractions, it's also easy for the DB to deployed to multiple regions so you could choose to do it at the same time. However, as we will discuss below, you need to be careful how you do this. See this blog post for more details on app vs. db deployment https://www.cockroachlabs.com/blog/follower-reads/
___
*[v21.1/multiregion-scale-application.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6fpeEeQWLdiFvIVq:-MctQdktC_4j5Stn3u5p:ba28pos) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L23)):*

So I thought we might be talking about different things when we say "scale database deployments" (see my comment above), but now I'm not so sure.

I understand the workflow to be:

1. Scale database deployment (i.e., add nodes to the cluster with regional localities)
2. Scale application deployment (e.g., deploy Docker images in new regions, with configurations to connect to the closest db node)
3. Execute multi-region DDL statements (i.e., `ALTER DATABASE ... ADD PRIMARY REGION`, `ALTER DATABASE ... ADD REGION`, etc)
4. Update application, if necessary

It seems to me that you are suggesting the following workflow (but I could totally be misinterpreting what you are saying):

1. Scale the application (e.g., deploy Docker images in new regions, with configurations to connect to the *same*, single-region db node)
2. Execute a single `ALTER DATABASE ... ADD PRIMARY REGION` statement on a single-region CDB cluster
3. Scale the database deployment (i.e., add nodes to the cluster with regional localities)
4. Execute the rest of the multi-region DDL statements (`ALTER DATABASE ... ADD REGION`, etc.)
5. Update each application configuration to connect to the closest db node (e.g., update Docker images with new connection strings, and then re-deploy the images)
6. Update application, if necessary.

Is that correct?
</blockquote></details>

Yes, that is what I'm describing. Particualrly, you can also wait between 1 and 2 until your app needs it. If we want to do it all together, which I think does make sense, you could also do it this way:

Execute a single ALTER DATABASE ... ADD PRIMARY REGION statement on a single-region CDB cluster
Scale the database deployment (i.e., add nodes to the cluster with regional localities)
Execute the rest of the multi-region DDL statements (ALTER DATABASE ... ADD REGION, etc.)
Scale the application (e.g., deploy Docker images in new regions) and Update each application configuration to connect to the closest db node (e.g., update Docker images with new connection strings, and then re-deploy the images)
Update application, if necessary.

It may also be worth discussing how load balancing works. You could potentially deploy the app in multiple regions and have it talk to one load balancer that talks to one db node. Then you could do the load balancer in multiple regions talking to the db in multiple regions.
___
*[v21.1/multiregion-scale-application.md, line 24 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mco9XuZ0sBUY-II38Ic:-MctPnc44AhAwIS9mUMf:b-a6kf81) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L24)):*

Not Andy (obv) but added my understanding below re:

> By "first step", do you mean before you have added any new nodes to your cluster (in new regions)?

I think the answer is "yes": if you (1) add a bunch of new nodes at time T without changing your multi-region config (really, constraints), the cluster will do a bunch of work rebalancing among those nodes based on CRDB's "default" set of constraints C in place at time T, and then when you (2) add a primary region at a later time T+1, yielding a new "narrower" set of constraints C', the cluster AGAIN has to do a bunch of work rebalancing etc. to meet the new constraint set C' which is now in place at time T+1.   The fact that the constraints started out "wide" and got "narrow" results in a lot of data movement etc.

By contrast, if you add a primary region first thing at time T when you have one node, the initial set of constraints is already "narrowed" significantly, so any new constraints you add at time T+1, ..., etc when you add a bunch of new nodes will be validated against the already existing "narrower" constraints and the data will go where you want it more quickly without so much rebalancing (since you avoided the wide constraint -> narrow constraint change described above)

Basically it's like a constraint propagation system that will be faster if you narrow the set of constraints as soon as possible during cluster setup and before adding more nodes, so it has a narrower range of things to consider (thus resulting in less churn / data movement / network traffic / etc)

At least this is my understanding of it
</blockquote></details>

Good explanation Rich! That's exactly what I meant. 
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-MctQ0gu39AxU-12Lvh-:b-ff6jdg) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*

the phrase we have been using for the `SET LOCALITY FOO ...` stuff is table localities

e.g. I might call this section "set table localities"

and in the paragraph below I might say something like "for most ... configurations ..., you do not need to update the default table localities"

a reason for this obliqueness is in keeping with the declarative nature of SQL and also the multi-region stuff in general: it's not like a statement in Python or or ALGOL or whatever where I say "set x=5", it's more like operating a constraint propagation system: "make this table regional in us-west-1", and then the r/w perf characteristics of the table's performance fall out of what CRDB does under the hood based on the constraints we have given it, depending on where the client is accessing the table from.  

So we are not really "upgrading" or "updating read/write operations" since what we are really doing is at arm's length: changing the "locality" constraint of the table and then CRDB does "Magic(tm)" to make it so that reads / writes from specific regions have certain characteristics.

Basically it's more fuzzy/declarative and less like a "regular" programming thing

I hope that makes sense - the core of actionable info is in the first 3 sentences above, the rest is my possibly unhelpful faux explication 
</blockquote></details>

+1 to rich

ericharmeling:
@awoods187 @rmloveland

Thank you both for the detailed responses! I'll work in your feedback as soon as I can.

ericharmeling (commented):
TFTR @awoods187, @rmloveland 

I've updated the PR, and left some additional comments, after going over your feedback. PTAL at the changes and my new comments.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796#-:-MdO9W-6D-ePGaEU_llz:b-sf7ej1)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @rmloveland)

___
*[v21.1/multiregion-scale-application.md, line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj69DqDH6kGOLP0676:-MdMxQR94De8HUjzDDLQ:b6qk19m) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L15)):*
> You can actually get a lot of value out of deploying your app to multiple regions without deploying your db. 

Great point. I've added a note about the latency benefits of deploying the app in multiple regions, even if you stay with a single-region db. I also added a link to the relevant section of the blog post.
___
*[v21.1/multiregion-scale-application.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6fpeEeQWLdiFvIVq:-MdJ3XKr37Y8B60PyqSq:byk6oev) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L23)):*
> If we want to do it all together, which I think does make sense, you could also do it this way: 
> Execute a single ALTER DATABASE ... ADD PRIMARY REGION statement on a single-region CDB cluster
> Scale the database deployment (i.e., add nodes to the cluster with regional localities)
> Execute the rest of the multi-region DDL statements (ALTER DATABASE ... ADD REGION, etc.)
> Scale the application (e.g., deploy Docker images in new regions) and Update each application configuration to connect to the closest db node (e.g., update Docker images with new connection strings, and then re-deploy the images)
> Update application, if necessary.

Okay - I think it will be easier to parse if we place the steps in two separate sections, with the order that you suggested here. We can then add a note to suggest a more flexible application deployment approach, and point to the blog post (as I mentioned in the comment thread above this one).
___
*[v21.1/multiregion-scale-application.md, line 24 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mco9XuZ0sBUY-II38Ic:-MdMlpRW6Oq7ZbbSbbJ1:b2ohd73) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L24)):*

Good explanation Rich! That's exactly what I meant. 
</blockquote></details>

Thank you, Rich! Makes total sense. I've tried to summarize these performance implications in as clear and simple terms as I can.

I do have some thoughts on this though (or really, I just want to point some stuff out).

1. AFAICT, we don't currently explicitly document the performance implications of executing all multi-region DDL statements after adding new nodes (please correct me if I'm wrong!). I'm happy to make this page the first place where we do - but we should probably also update some of the existing reference docs (https://www.cockroachlabs.com/docs/v21.1/set-primary-region.html and https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html, in particular) to include a bit about the performance costs of adding nodes before setting the database(s)'s primary region.
2. The [multi-region tutorial](https://www.cockroachlabs.com/docs/stable/demo-low-latency-multi-region-deployment.html) does not demonstrate a proper "scaling" workflow. (And neither does the [scaling implementation in the python movr workload](https://github.com/cockroachdb/movr/blob/master/movr.py#L258)). The cluster in that tutorial starts as multi-region, so all the DDL is run back-to-back. I understand that scaling performance isn't really the purpose of the tutorial (it's more about comparing the performance with and without multi-region features), but I wonder if we want to create a tutorial, akin to the current multi-region tutorial, about scaling nodes *and* transforming the schema to use multi-region.
3. To set the database primary region when the cluster is single-region, you need to have set regional node localities for the nodes in the single-region cluster from the very beginning. UUIC, there is no way to add/change the locality of a node after it has been started, without shutting the node down and starting it up again (this something I brought up to Rich a while back). IDK if there is a technical reason for this, but IMO it greatly limits the ability of a single-region cluster to scale. A naive approach to resolving this would be to just make `--locality` a required flag. Something for engineering to consider.
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-MdMx7CuDkz_dq-uGezS:b-52bdhp) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*
>e.g. I might call this section "set table localities"

Maybe I'm misunderstanding what y'all are referring to, but this section isn't about setting localities. It's about changing your application code so that it works with the new multi-region database schema. It's not about changing the database schema; it's about changing the persistence layer functions in the code itself (i.e., the DML SQL statements) *after* the database schema has already been changed.

The lead-in sentence here acknowledges that you likely won't need to change anything if you have the default table localities.

> For most multi-region configurations, including the default configuration (`LOCALITY REGIONAL BY TABLE IN PRIMARY REGION` and `SURVIVE ZONE FAILURE`), you do not need to update any client-side read/write operations after migrating your database schema for multi-region features. CockroachDB automatically optimizes queries against multi-region databases, based on the locality of the node executing the query, and on the multi-region configuration of the database.

Perhaps the language I am using here is confusing. I've updated the language to be a little more explicit than just "client read/write operations". I also changed the "multi-region configuration" terminology to just focus on table localities.

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 24 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mco9XuZ0sBUY-II38Ic:-MdXcx3J90sVgAJPBk03:b7rpxl5) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L24)):*

Thank you, Rich! Makes total sense. I've tried to summarize these performance implications in as clear and simple terms as I can.

I do have some thoughts on this though (or really, I just want to point some stuff out).

1. AFAICT, we don't currently explicitly document the performance implications of executing all multi-region DDL statements after adding new nodes (please correct me if I'm wrong!). I'm happy to make this page the first place where we do - but we should probably also update some of the existing reference docs (https://www.cockroachlabs.com/docs/v21.1/set-primary-region.html and https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html, in particular) to include a bit about the performance costs of adding nodes before setting the database(s)'s primary region.
2. The [multi-region tutorial](https://www.cockroachlabs.com/docs/stable/demo-low-latency-multi-region-deployment.html) does not demonstrate a proper "scaling" workflow. (And neither does the [scaling implementation in the python movr workload](https://github.com/cockroachdb/movr/blob/master/movr.py#L258)). The cluster in that tutorial starts as multi-region, so all the DDL is run back-to-back. I understand that scaling performance isn't really the purpose of the tutorial (it's more about comparing the performance with and without multi-region features), but I wonder if we want to create a tutorial, akin to the current multi-region tutorial, about scaling nodes *and* transforming the schema to use multi-region.
3. To set the database primary region when the cluster is single-region, you need to have set regional node localities for the nodes in the single-region cluster from the very beginning. UUIC, there is no way to add/change the locality of a node after it has been started, without shutting the node down and starting it up again (this something I brought up to Rich a while back). IDK if there is a technical reason for this, but IMO it greatly limits the ability of a single-region cluster to scale. A naive approach to resolving this would be to just make `--locality` a required flag. Something for engineering to consider.
</blockquote></details>

Thanks for this valuable feedback Eric!

Re point 1 I filed #10867 

Re point 2, filed #10868 
  
___
*[v21.1/multiregion-scale-application.md, line 68 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj7L5Y0ValW7Tw-b39:-Mcovu3C6KIL4jiDhuY2:b2exyam) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L68)):*

the phrase we have been using for the `SET LOCALITY FOO ...` stuff is table localities

e.g. I might call this section "set table localities"

and in the paragraph below I might say something like "for most ... configurations ..., you do not need to update the default table localities"

a reason for this obliqueness is in keeping with the declarative nature of SQL and also the multi-region stuff in general: it's not like a statement in Python or or ALGOL or whatever where I say "set x=5", it's more like operating a constraint propagation system: "make this table regional in us-west-1", and then the r/w perf characteristics of the table's performance fall out of what CRDB does under the hood based on the constraints we have given it, depending on where the client is accessing the table from.  

So we are not really "upgrading" or "updating read/write operations" since what we are really doing is at arm's length: changing the "locality" constraint of the table and then CRDB does "Magic(tm)" to make it so that reads / writes from specific regions have certain characteristics.

Basically it's more fuzzy/declarative and less like a "regular" programming thing

I hope that makes sense - the core of actionable info is in the first 3 sentences above, the rest is my possibly unhelpful faux explication 
</blockquote></details>

PS if you have feedback on how we could improve the page you linked above

awoods187 (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/10796#-Mcj6fpeEeQWLdiFvIVq:-MeXcZGZ-It2HZ5i4gKH:bb73djm) ([raw file](https://github.com/cockroachdb/docs/blob/764f31374ebf1fd24550072d60b83bafbdf6d3d0/v21.1/multiregion-scale-application.md#L23)):*

> If we want to do it all together, which I think does make sense, you could also do it this way: 
> Execute a single ALTER DATABASE ... ADD PRIMARY REGION statement on a single-region CDB cluster
> Scale the database deployment (i.e., add nodes to the cluster with regional localities)
> Execute the rest of the multi-region DDL statements (ALTER DATABASE ... ADD REGION, etc.)
> Scale the application (e.g., deploy Docker images in new regions) and Update each application configuration to connect to the closest db node (e.g., update Docker images with new connection strings, and then re-deploy the images)
> Update application, if necessary.

Okay - I think it will be easier to parse if we place the steps in two separate sections, with the order that you suggested here. We can then add a note to suggest a more flexible application deployment approach, and point to the blog post (as I mentioned in the comment thread above this one).
</blockquote></details>

SGTM

ericharmeling:
@awoods187 @rmloveland 

Everything looking good from your points of view?

@ajstorm 

When you get a chance, it would be nice to have an engineering review on this PR as well. :)

rmloveland (approved):
lgtm yo

ericharmeling:
@ajstorm Friendly ping on review :)

ajstorm (changes requested):
Sorry for the delay.  Looks good overall.  Just a few comments (mostly focused on manually creating the enum).

> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796#-:-MfXtknpBQFNUyH22UaN:bjgtsrz)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187 and @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 3 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXnyp61BJZQuSGaUNo:-MfXnyp61BJZQuSGaUNp:b-l1ubxl) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L3)):*
> ```Markdown
> ---
> title: Scale to Multiple Regions
> summary: Learn how to scale a single-region application to a multi-region application.
> ```

Nit: This one is very much up to you, but this sentence might read a bit better if we ended it with "...application to multiple regions." so that it matches the title.
___
*[v21.1/multiregion-scale-application.md, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXoH97CAOmA5PiaZbA:-MfXoH97CAOmA5PiaZbB:b-rmrlqo) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L7)):*
> ```Markdown
> ---
> 
> This page provides guidance for scaling a single-region application to a multi-region application.
> ```

Nit: if we're going to change things above, we should also change them here.
___
*[v21.1/multiregion-scale-application.md, line 15 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXoROf310m746jpfEA:-MfXoROf310m746jpfEB:b5jhk5t) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L15)):*
> ```Markdown
> Scaling an application from a single region to multiple regions consists of:
> 
> - [Scaling the database](#scale-the-database), which includes transforming the database schema into a multi-region schema and adding new nodes to a CockroachDB cluster.
> ```

Nit: We might want to be more explicit here.  "...which includes adding new nodes to a CockroachDB cluster in different regions (if not present already), adding regions to the database, and optionally transforming the database schema to leverage multi-region table localities."
___
*[v21.1/multiregion-scale-application.md, line 17 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXpaa82Ay1C0CXhpqJ:-MfXpaa82Ay1C0CXhpqK:bu1aiz4) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L17)):*
> ```Markdown
> - [Scaling the database](#scale-the-database), which includes transforming the database schema into a multi-region schema and adding new nodes to a CockroachDB cluster.
> 
> - [Scaling the application](#scale-the-application), which includes deploying the application in new regions and, if necessary, updating the application code to work with the multi-region database schema.
> ```

Nit: I'd think we'd want to add "optionally" right before "updating the application code".  In most cases, no application code changes should be required.
___
*[v21.1/multiregion-scale-application.md, line 23 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXprLn9zef_Ky7FqpS:-MfXprLn9zef_Ky7FqpT:b-lxd6x8) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L23)):*
> ```Markdown
> To scale a database from a single region to multiple regions:
> 
> 1. Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to the single region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
> ```

Why "single region" here?  I'd think that we could change this to "one of the regions in which the cluster is deployed".  Then, the comment below about "setting the primary database region before", can call out the benefits to doing this before adding new regions.

After reading the discussion below, I agree about it's importance, but we might want to provide a bit more context up here, or change the wording slightly.  I could see this occurring in a few steps:

1) Prep the database (by adding the primary region)
2) Scale the cluster
3) Scale the database
4) Scale the application
___
*[v21.1/multiregion-scale-application.md, line 25 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXq5sk43cwq6-U0tlt:-MfXq5sk43cwq6-U0tlu:b439kua) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L25)):*
> ```Markdown
> 1. Use an [`ALTER DATABASE ... SET PRIMARY REGION`](set-primary-region.html) statement to set the database's [primary region](multiregion-overview.html#database-regions) to the single region in which the cluster is deployed. This region must have been specified as a [regional locality](cockroach-start.html#locality) at cluster startup.
> 
>     Setting the primary database region before adding new regional nodes to the cluster prevents CockroachDB from [rebalancing row replications](architecture/replication-layer.html#leaseholder-rebalancing) across all regions each time a node is added in a new region.
> ```

Nit: I'd drop "database" in "primary database region".
___
*[v21.1/multiregion-scale-application.md, line 44 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXrftl9oYXtmmy3gk3:-MfXrftl9oYXtmmy3gk4:b30k0ve) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L44)):*
> ```Markdown
> 
>     {{site.data.alerts.callout_info}}
>     You must specify a [regional locality](cockroach-start.html#locality) for each node at startup. These regional localities are represented as [cluster regions](multiregion-overview.html#cluster-regions) in the cluster.
> ```

Nit: "You" is only relevant here for orchestrated and manual deployments, right?  CC will set these localities automatically.
___
*[v21.1/multiregion-scale-application.md, line 49 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXrxxUCaqsX7kzzHWs:-MfXrxxUCaqsX7kzzHWt:b-4vs3mh) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L49)):*
> ```Markdown
> 1. Use an [`ALTER DATABASE ... ADD REGIONS`](add-region.html) statement to add the new regions to your database. Only cluster regions (i.e., regional localities specified at cluster startup) can be added as [database regions](multiregion-overview.html#database-regions).
> 
> 1. Use [`ALTER DATABASE ... SURVIVE ... FAILURE`](survive-failure.html) statements to set your database's [survival goals](multiregion-overview.html#survival-goals).
> ```

Nit: optionally...
___
*[v21.1/multiregion-scale-application.md, line 51 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXs0R361NvppjzuPtD:-MfXs0R361NvppjzuPtE:b9zasfb) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L51)):*
> ```Markdown
> 1. Use [`ALTER DATABASE ... SURVIVE ... FAILURE`](survive-failure.html) statements to set your database's [survival goals](multiregion-overview.html#survival-goals).
> 
> 1. Use [`ALTER TABLE ... SET LOCALITY`](set-locality.html) statements to set [table localities](multiregion-overview.html#table-locality) for each table.
> ```

Nit: optionally
___
*[v21.1/multiregion-scale-application.md, line 86 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXsWaN4wuf4PhzqgsR:-MfXsWaN4wuf4PhzqgsS:brf2dy8) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L86)):*
> ```Markdown
> If your application's ORM framework supports enumerated data types, and you need to explicitly reference the `crdb_region` column (or a custom region column) in a SQL operation in your application code, you should do the following:
> 
> - Using the ORM framework's type library, create an enumerated type, with the type name `crdb_internal_region` and enumerated values for each of the existing regions.
> ```

You can't create this type manually.  The database will do it for you automatically as new regions are added.  If you create it manually, bad stuff is going to happen.
___
*[v21.1/multiregion-scale-application.md, line 87 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXslnc9Ow_GlMFIhdt:-MfXslnc9Ow_GlMFIhdu:begj4qd) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L87)):*
> ```Markdown
> 
> - Using the ORM framework's type library, create an enumerated type, with the type name `crdb_internal_region` and enumerated values for each of the existing regions.
> - Add a new region column to the relevant table mappings, with the new enumerated data type and the column name `crdb_region` (or the name of the custom region column of the table). Note that this column does not persist to the database; it is simply a mapping for the application.
> ```

This column will be persisted.  Also, are we doing this only for REGIONAL BY ROW tables?  If so, we'd need a REGIONAL BY ROW AS statement.  

In general, I'm not following why the user would need to create this manually.  Can you walk me through the use case?
___
*[v21.1/multiregion-scale-application.md, line 100 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtBDsEqb4lWF9ahyB:-MfXtBDsEqb4lWF9ahyC:b-yanolo) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L100)):*
> ```Markdown
> 
>     region_list = list(tup[0] for tup in session.execute(text('SELECT region FROM [SHOW REGIONS]')).fetchall())
>     region_enum = Enum(*region_list, name='crdb_internal_region', create_type=False, native_enum=False)
> ```

This should be removed.  We can't create this enum manually.
___
*[v21.1/multiregion-scale-application.md, line 117 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtOIp0t80njAncsx0:-MfXtOIp0t80njAncsx1:b-3trgy0) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L117)):*
> ```Markdown
> 3. Reference the column value as needed.
> 
>     Here is an example function that updates the region value in a given table, using the values of a `city` column:
> ```

You should be able to reference this column automatically if you've defined the table as REGIONAL BY ROW.  Maybe we just need to walk through an example of that?
___
*[v21.1/multiregion-scale-application.md, line 136 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXteL7EyS7clOzRBn_:-MfXteL7EyS7clOzRBna:b4kxhgp) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L136)):*
> ```Markdown
> If your ORM framework does not support `ENUM` data types, or if you run into type compatibility issues, you can alternatively try the following approach:
> 
> - Add `crdb_region` (or the custom region-tracking column) to an existing table object as a *string*-typed column.
> ```

Can you walk me through the use case here?

ericharmeling (commented):
@ajstorm 

No worries re: the delay. Thanks for reviewing this and giving really great feedback!

I updated the page after going through your review. Please see my comments below re: the enum. (spoiler: I rewrote that whole section.)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796#-:-MfY1DKY7DxExoGXrmVo:b-r3juhn)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm and @awoods187)

___
*[v21.1/multiregion-scale-application.md, line 3 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXnyp61BJZQuSGaUNo:-MfY042R6bNIUwvPw_2P:btvpuu7) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L3)):*

Nit: This one is very much up to you, but this sentence might read a bit better if we ended it with "...application to multiple regions." so that it matches the title.
</blockquote></details>

Good shout. Done.
___
*[v21.1/multiregion-scale-application.md, line 7 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXoH97CAOmA5PiaZbA:-MfY0Oxw2udC65ltVz7V:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L7)):*

Nit: if we're going to change things above, we should also change them here.
</blockquote></details>

Done.
___
*[v21.1/multiregion-scale-application.md, line 15 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXoROf310m746jpfEA:-MfY0Thf1pZ7BYIHCD7G:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L15)):*

Nit: We might want to be more explicit here.  "...which includes adding new nodes to a CockroachDB cluster in different regions (if not present already), adding regions to the database, and optionally transforming the database schema to leverage multi-region table localities."
</blockquote></details>

Done.
___
*[v21.1/multiregion-scale-application.md, line 17 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXpaa82Ay1C0CXhpqJ:-MfY0gC550mrEkTQU9RG:b-9oozx2) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L17)):*

Nit: I'd think we'd want to add "optionally" right before "updating the application code".  In most cases, no application code changes should be required.
</blockquote></details>

I think this is covered by the "if necessary" parenthetical. Later on, there is a sentence that calls out that the application won't usually need updating. Also, I've added "Optional" to that dedicated "Update the application code" section header below.
___
*[v21.1/multiregion-scale-application.md, line 23 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXprLn9zef_Ky7FqpS:-MfY34kTEN8xlkBmJ_a0:bs65bif) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L23)):*
> Why "single region" here? I'd think that we could change this to "one of the regions in which the cluster is deployed". 

The assumption is that the cluster is single-region at this point (these are the steps to scale to multiple regions, from a single region). Your suggestion still holds though and would cover more cases (single region to multiple regions, multiple regions to more regions). Updated.

>Prep the database (by adding the primary region)
Scale the cluster
Scale the database
Scale the application

Thanks for the suggestion! Incorporated this into some section-restructuring changes.
___
*[v21.1/multiregion-scale-application.md, line 25 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXq5sk43cwq6-U0tlt:-MfY1BnSDEwgk7AvKO73:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L25)):*

Nit: I'd drop "database" in "primary database region".
</blockquote></details>

Done.
___
*[v21.1/multiregion-scale-application.md, line 44 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXrftl9oYXtmmy3gk3:-MfY1U6SDGMH3zdp6NQr:b-ageg3o) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L44)):*

Nit: "You" is only relevant here for orchestrated and manual deployments, right?  CC will set these localities automatically.
</blockquote></details>

Added "For orchestrated and manual deployments," to the beginning of that note.
___
*[v21.1/multiregion-scale-application.md, line 49 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXrxxUCaqsX7kzzHWs:-MfY22fa0NvUlwv0MIX4:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L49)):*

Nit: optionally...
</blockquote></details>

Done.
___
*[v21.1/multiregion-scale-application.md, line 51 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXs0R361NvppjzuPtD:-MfY23UnD1JNhROju7PY:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L51)):*

Nit: optionally
</blockquote></details>

Done.
___
*[v21.1/multiregion-scale-application.md, line 86 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXsWaN4wuf4PhzqgsR:-MfY9HpW-w8aI4JYT0R8:bkm6vy5) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L86)):*

You can't create this type manually.  The database will do it for you automatically as new regions are added.  If you create it manually, bad stuff is going to happen.
</blockquote></details>

This section is about creating a representation of the existing `crdb_region` type in the ORM, and not about creating the type and persisting it to the database. The guidance here is really only relevant if you need to reference `crdb_region` in a DML statement, and your ORM has no way of dynamically updating the mappings to the database schema, which I thought was the case but it turned out not to be. :)

I had to rewrite this after reading your feedback and doing a little more testing. It should be clearer now.
___
*[v21.1/multiregion-scale-application.md, line 87 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXslnc9Ow_GlMFIhdt:-MfY6qIpAc10MqeKIfp3:bt2okd3) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L87)):*
>This column will be persisted. 

IIUC, SQLAlchemy does not persist ORM mappings to the database unless you are hooked up to a schema migration tool that automatically executes DDL statements on behalf of the application. SQLAlchemy [does not have schema manipulation capabilities](https://docs.sqlalchemy.org/en/14/core/metadata.html#altering-database-objects-through-migrations) (e.g., you can't add a column to an existing table with SQLAlchemy alone). This limitation certainly influenced [my own multi-region scaling efforts](https://github.com/cockroachdb/movr/blob/master/movr.py#L258).

For other ORMs, this might not be the case though. Which makes providing any kind of ORM guidance here tricky, because it will likely vary from ORM to ORM. I think I've made the guidance little more general in my latest commit.

> In general, I'm not following why the user would need to create this manually. Can you walk me through the use case?

The idea is not to create the column/type manually, but to have some object representing the column in the application, so the application code can execute DML statements that reference the column, without having to use raw SQL. I've tried to list some instances earlier in this section:

> - If a table has a `REGIONAL BY ROW AS <custom_region_column>` table locality, and you want to explicitly insert regional values into a table, as shown in [Low Latency Reads and Writes in a Multi-Region Cluster](demo-low-latency-multi-region-deployment.html#configure-regional-by-row-tables).
> - If a table has a `REGIONAL BY ROW` locality, and you want to update the hidden `crdb_region` column of existing rows in the table based on some other column value, as shown in [Set the table locality to `REGIONAL BY ROW`](set-locality.html#set-the-table-locality-to-regional-by-row).
> - If a table has a `REGIONAL BY ROW` locality, and you want to filter a selection query based on the value of the region-tracking column.
In all of these scenarios, statements reference the column that tracks the region for each row in a `REGIONAL BY ROW` locality. This column can be a custom column of the built-in `ENUM` type `crdb_internal_region`, or it can be the default, hidden [`crdb_region` column](set-locality.html#crdb_region).

So I now know that there is a much much easier way to represent existing columns that I only just got working. I was having trouble doing this before because the `crdb_region` column was hidden, which was preventing SQLAlchemy from loading the database schema properly from the database.
___
*[v21.1/multiregion-scale-application.md, line 100 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtBDsEqb4lWF9ahyB:-MfYssmb9lhGiUBcqKIy:b-a9mtg) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L100)):*

This should be removed.  We can't create this enum manually.
</blockquote></details>

Again, this is meant just to reflect the column in the database.
___
*[v21.1/multiregion-scale-application.md, line 117 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtOIp0t80njAncsx0:-MfY92Wq6M8U3XR6rXtR:bijw95k) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L117)):*
>You should be able to reference this column automatically

I think that is the intention of this guidance. We aren't creating the column, we are just referencing it from the ORM.

If by "reference this column automatically" you are saying that you should be able to use the ORM libraries to update the local representation of the database columns dynamically, then you are describing the ideal situation (i.e., a dynamic sync from database to ORM). 

TBH, in my own local environment, I couldn't get this to work with SQLAlchemy (using [`reflect(extend_existing=True, autoload_replace=True)`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.reflect)) at first. Other ORMs surely have this kind of capability as well (even if they don't support an ORM to database sync, like SQLAlchemy).

However! I wasn't considering what effect the hidden column was having on the sync, and I didn't see https://github.com/cockroachdb/cockroach/pull/63052 (which has yet to be documented). After running `ALTER TABLE ... ALTER COLUMN ... SET VISIBLE`, the sync worked! 

So... I've updated the guidance here to include running this statement, and not creating a new column/type on the app side at all, but instead just loading the schema onto the mapping.
___
*[v21.1/multiregion-scale-application.md, line 136 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXteL7EyS7clOzRBn_:-MfYcsvz1r6LqWDQLa4p:b7vyepm) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L136)):*

Can you walk me through the use case here?
</blockquote></details>

So I relegated this to just a note, but the idea is that you are trying to reference the column and your ORM can't sync with the database. So you have to just create a manual, string-typed column with the name `crdb_region` and then reference that in your DML.

ajstorm (approved):
Thanks for addressing the comments.  :lgtm: 

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796#-:-MfZOg9C8a0bloiIYrkI:bvyk976)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187 and @ericharmeling)

___
*[v21.1/multiregion-scale-application.md, line 17 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXpaa82Ay1C0CXhpqJ:-MfZMh7QCLoYP0L-fo9L:b-8n327p) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L17)):*

I think this is covered by the "if necessary" parenthetical. Later on, there is a sentence that calls out that the application won't usually need updating. Also, I've added "Optional" to that dedicated "Update the application code" section header below.
</blockquote></details>

Yes, good point.
___
*[v21.1/multiregion-scale-application.md, line 23 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXprLn9zef_Ky7FqpS:-MfZNUur-bepeNDp0yAl:b9bkkpe) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L23)):*

> Why "single region" here? I'd think that we could change this to "one of the regions in which the cluster is deployed". 

The assumption is that the cluster is single-region at this point (these are the steps to scale to multiple regions, from a single region). Your suggestion still holds though and would cover more cases (single region to multiple regions, multiple regions to more regions). Updated.

>Prep the database (by adding the primary region)
Scale the cluster
Scale the database
Scale the application

Thanks for the suggestion! Incorporated this into some section-restructuring changes.
</blockquote></details>

Looks great.
___
*[v21.1/multiregion-scale-application.md, line 86 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXsWaN4wuf4PhzqgsR:-MfZO0eI317l5bQZuyQs:b-19z96f) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L86)):*

This section is about creating a representation of the existing `crdb_region` type in the ORM, and not about creating the type and persisting it to the database. The guidance here is really only relevant if you need to reference `crdb_region` in a DML statement, and your ORM has no way of dynamically updating the mappings to the database schema, which I thought was the case but it turned out not to be. :)

I had to rewrite this after reading your feedback and doing a little more testing. It should be clearer now.
</blockquote></details>

This look much better.  Thanks!
___
*[v21.1/multiregion-scale-application.md, line 100 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtBDsEqb4lWF9ahyB:-MfZOS9LBTlnH8mk6LnD:blbulw7) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L100)):*

Again, this is meant just to reflect the column in the database.
</blockquote></details>

Looks better now.
___
*[v21.1/multiregion-scale-application.md, line 117 at r3](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfXtOIp0t80njAncsx0:-MfZOXDB-J-FhVjp1rSN:b-dbab0j) ([raw file](https://github.com/cockroachdb/docs/blob/f0e8a60339a47b2aac68282019c9ffdc1d8f7400/v21.1/multiregion-scale-application.md#L117)):*

>You should be able to reference this column automatically

I think that is the intention of this guidance. We aren't creating the column, we are just referencing it from the ORM.

If by "reference this column automatically" you are saying that you should be able to use the ORM libraries to update the local representation of the database columns dynamically, then you are describing the ideal situation (i.e., a dynamic sync from database to ORM). 

TBH, in my own local environment, I couldn't get this to work with SQLAlchemy (using [`reflect(extend_existing=True, autoload_replace=True)`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.reflect)) at first. Other ORMs surely have this kind of capability as well (even if they don't support an ORM to database sync, like SQLAlchemy).

However! I wasn't considering what effect the hidden column was having on the sync, and I didn't see https://github.com/cockroachdb/cockroach/pull/63052 (which has yet to be documented). After running `ALTER TABLE ... ALTER COLUMN ... SET VISIBLE`, the sync worked! 

So... I've updated the guidance here to include running this statement, and not creating a new column/type on the app side at all, but instead just loading the schema onto the mapping.
</blockquote></details>

All good.  Thanks for taking a second look.
___
*[v21.1/multiregion-scale-application.md, line 21 at r4](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfZMwfQ2zKj3q5ToZsj:-MfZMwfQ2zKj3q5ToZsk:bwx03xq) ([raw file](https://github.com/cockroachdb/docs/blob/8d6e4d466de5b50763625e987f77868d660883a5/v21.1/multiregion-scale-application.md#L21)):*
> ```Markdown
> ## Scale the database
> 
> ### Step 1. Prep the database schema
> ```

Nit: I'd remove schema here.  We're not really modifying the database's schema at this point.  Well, to be fair, it depends on what you consider a schema to be, but I'd think we'd be find removing schema.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10796)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187)

___
*[v21.1/multiregion-scale-application.md, line 21 at r4](https://reviewable.io/reviews/cockroachdb/docs/10796#-MfZMwfQ2zKj3q5ToZsj:-Mfc6xQp6uyO6Cn-ikWD:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/8d6e4d466de5b50763625e987f77868d660883a5/v21.1/multiregion-scale-application.md#L21)):*

Nit: I'd remove schema here.  We're not really modifying the database's schema at this point.  Well, to be fair, it depends on what you consider a schema to be, but I'd think we'd be find removing schema.
</blockquote></details>

Done.

Issue #10479, opened 2021-04-30
recommend regional over zonal clusters in K8s deployment docs


udnay:
Yandu Oppacher (udnay) commented:

The current [multi-site guides](https://www.cockroachlabs.com/docs/v20.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster) use zonal clusters for the reference implementation. These should be regional clusters to ensure proper redundancy and matching SLA requirements within a region.

e.g.
```
regions = {
    'us-east1-b': 'us-east1',
    'us-west1-a': 'us-west1',
    'us-central1-a': 'us-central1',
}
```
Would only get a customer "single instance uptime"  SLA of `99.5` instead of `99.99` for regional workloads. There is similarly a lower SLA for zonal GKE clusters. We should update the docs to reflect this.

Jira Issue: DOC-1605

udnay:
We should also look through the README for [multi-region](https://github.com/cockroachdb/cockroach/tree/master/cloud/kubernetes/multiregion) in the CRDB repo which is also out of date. /cc @keith-mcclellan

udnay:
e.g. [External LB requirement](https://github.com/cockroachdb/cockroach/blob/master/cloud/kubernetes/multiregion/README.md#exposing-dns-servers-to-the-internet) is wrong if I understand what it's trying to do, you can [hit internal LB from any region](https://cloud.google.com/load-balancing/docs/internal#client_access) in google now

PR #10813, opened 2021-06-22
specify regional clusters in single-region GKE deployments

taroface:
Addresses #10479 for single-region deployments.

Also removed the callout about defining a default `gcloud` region/zone, since a regional cluster requires an explicit `--region` flag and it would send a mixed message to also define an unrelated default region.

The multi-region GKE doc needs a deeper overhaul. I didn't update it because parts of the Python script used in that doc appear to reference zone values, and I'm not sure whether they would be broken by specifying a regional cluster.

### DIFF
--- a/_includes/v20.1/orchestration/start-kubernetes.md
+++ b/_includes/v20.1/orchestration/start-kubernetes.md
@@ -5,6 +5,10 @@ Choose whether you want to orchestrate CockroachDB with Kubernetes using the hos
 - [Manual GCE](#manual-gce)
 - [Manual AWS](#manual-aws)
 
+{{site.data.alerts.callout_info}}
+The CockroachDB Kubernetes Operator is currently supported for GKE. You can also use the Operator on platforms such as [Red Hat OpenShift](deploy-cockroachdb-with-kubernetes-openshift.html) and [IBM Cloud Pak for Data](https://www.ibm.com/products/cloud-pak-for-data).
+{{site.data.alerts.end}}
+
 ### Hosted GKE
 
 1. Complete the **Before You Begin** steps described in the [Google Kubernetes Engine Quickstart](https://cloud.google.com/kubernetes-engine/docs/quickstart) documentation.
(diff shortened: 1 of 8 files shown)


udnay (approved):
> Reviewed 6 of 8 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10813)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @taroface)

___
*[_includes/v20.1/orchestration/start-kubernetes.md, line 37 at r2](https://reviewable.io/reviews/cockroachdb/docs/10813#-McpgsUH1q6IGxIjMMoM:-McpgsUH1q6IGxIjMMoN:bwyn8t4) ([raw file](https://github.com/cockroachdb/docs/blob/d2a29ed29a4d3554a15f62edbd1740bf9082a8be/_includes/v20.1/orchestration/start-kubernetes.md#L37)):*
> ```Markdown
>     ~~~
> 
>     This creates GKE instances and joins them into a single Kubernetes cluster named `cockroachdb`. The `--region` and `--num-nodes` flags specify a [regional three-zone cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-regional-cluster) with one node in each zone.
> ```

Maybe call out that the `--num-nodes` flag is how many nodes are wanted per zone. I guess you do in the example, probably sufficient.

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10813)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @udnay)

___
*[_includes/v20.1/orchestration/start-kubernetes.md, line 37 at r2](https://reviewable.io/reviews/cockroachdb/docs/10813#-McpgsUH1q6IGxIjMMoM:-MctJNaeEeMUE9ErawuY:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/d2a29ed29a4d3554a15f62edbd1740bf9082a8be/_includes/v20.1/orchestration/start-kubernetes.md#L37)):*
<details><summary><i>Previously, udnay (Yandu Oppacher) wrote…</i></summary><blockquote>

Maybe call out that the `--num-nodes` flag is how many nodes are wanted per zone. I guess you do in the example, probably sufficient.
</blockquote></details>

Done.

ianjevans (approved):

PR #10814, opened 2021-06-23
Add SSO to the enterprise features list

### LINKED ISSUE #10812: Include Single Sign On (SSO) on the Enterprise Feature list
Just noticed the SSO is not listed as an enterprise feature here: https://www.cockroachlabs.com/docs/v21.1/enterprise-licensing.html but we do call it out as being an enterprise feature:  https://www.cockroachlabs.com/docs/v21.1/sso.html.

jseldess:
Fixes #10812


jseldess:
PTAL, @chriscasano.

Also, I suspect that we weren't listing this feature in the enterprise list previously because it's marked as experimental. @thtruo, @aaron-crl, can you share your perspective on whether it's still experimental and whether or not it's ok to call it out in the enterprise features list? It does require an enterprise license, in the end.

aaron-crl:
Pinging @dhartunian as he lead this effort. To my knowledge the feature is still experimental.

dhartunian:
The reason for marking this experimental last year was that the API was built for a single enterprise client at the time and wasn't widely tested with many OAuth providers and configurations. We decided to mark the API as experimental in case we got more feedback from other customers that led to adjustments.

@thtruo if there aren't any pending asks here I'm happy with removing the experimental designation. It would involve some minimal code changes to remove the experimental labeling on cluster settings as well

thtruo:
Thanks for the context David. To be fully transparent, I haven't been close to this area since this shipped in 20.2 and don't have the latest feedback around what customers have been reporting back, if any. Furthermore, I believe Piyush would want to chime in on whether or not we remove the experimental designation. Perhaps the decision to remove the experimental label could wait when he returns? 
cc @mwang1026 

Having said that, it still feels relatively low risk to add SSO to the enterprise features list even when we maintain the experimental label, so this looks okay to me.

jseldess:
Thanks, all. I'll merge this.

PR #10842, opened 2021-06-24
Remove asciicast artifacts from docs repo

jseldess:
These haven't been used for several releases.


jseldess:
@mikeCRL, this is really just cleanup that I want you to be aware of. I'll merge. We can also revert if there's a need.

mikeCRL:
Thanks @jseldess. I got through a big chunk of it and saw no issues. Will
still read though the rest today.

PR #10844, opened 2021-06-24
document certificate rotation with Operator

### LINKED ISSUE #7773: Create documentation for rotating node certificates that do not use the k8s CA
When reading documentation for rotating node certs (https://www.cockroachlabs.com/docs/stable/rotate-certificates.html) [1], there is no guidance on how to rotate non-kubernetes CA based certificates (https://www.cockroachlabs.com/docs/v20.1/orchestrate-a-local-cluster-with-kubernetes.html#non-kubernetes-ca)[2]. The rotation documentation is not helpful if you created a node cert and pushed it into a k8s secret.

To better enable your developers, I would propose adding documentation to cover the gap described above. The steps I took to rotate my node certificates were:
1) overwrite the local node cert as described in [1].
2) delete the existing kubernetes secret for cockroachdb.node (something like "kubectl delete secret cockroachdb.node")
3) upload the new cert to k8s (kubectl create secret generic cockroachdb.node --from-file=certs).
4) roll cockroachdb-* pods.

There may be a more idiomatic way to rotate these but the above works. Thought I'd include it just in case it's helpful.

taroface:
Fixes https://github.com/cockroachdb/cockroach-operator/issues/413.
Fixes #7773.

- Added steps for manual cert rotation when using `cockroach cert`.


taroface:
@udnay Is it a high priority for us to document:

- rotation of the CA certificate (as here: https://www.cockroachlabs.com/docs/v20.2/rotate-certificates#rotate-the-ca-certificate)
- rotation of certs generated by the Operator (i.e., just with `tlsEnabled: true` and no custom node/client secrets defined)

udnay (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10844)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @taroface)

ianjevans (approved):

PR #10853, opened 2021-06-28
Clarify region survivability ups replication to 5

### LINKED ISSUE #10673: sql: clarify region survivability ups replication factor to 5
In https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html#surviving-region-failures we disucss surviving region failures but we don't specify that we up the replication factor to 5. We want to be careful about exposing this in too much detail but its valuable to document it and to explain that this will allow one node failure without moving the leaseholder out of the region to protect read latency.

rmloveland:
Fixes #10673

Summary of changes:

- Update 'Multiregion Overview' page with info re: replication factor in
  the REGION survival goal section

- Update the `num_replicas` and `num_voters` docs on the 'Configure
  Replication Zones' page to provide a bit more info re: the defaults in
  multi-region clusters with REGION survival

- Update the 'Replication Layer' page in the architecture docs to add a
  link from the non-voting replicas section to the REGION survival goals
  section of the 'Multi-region Overview' page


ajstorm (approved):
:lgtm: 

Only one nit about the reasons why we up the replication count.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10853#-:-MdO7xtf2M1f5f2Xjct9:b-nxf7ei)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @rmloveland)

___
*[v21.1/multiregion-overview.md, line 107 at r1](https://reviewable.io/reviews/cockroachdb/docs/10853#-MdO7Up63jrbT-VWfKpc:-MdO7Up63jrbT-VWfKpd:bxbmxu9) ([raw file](https://github.com/cockroachdb/docs/blob/77d36185edbac173230199911b9129b1d1ae05f4/v21.1/multiregion-overview.md#L107)):*
> ```Markdown
> The region level survival goal has the property that the database will remain fully available for reads and writes, even if an entire region goes down. This added survival comes at a cost: write latency will be increased by at least as much as the round-trip time to the nearest region. Read performance will be unaffected. In other words, you are adding network hops and making writes slower in exchange for robustness.
> 
> You can configure a database to survive region failures using the [`ALTER DATABASE ... SURVIVE REGION FAILURE` statement](survive-failure.html). This increases the [replication factor](configure-replication-zones.html#num_replicas) of all data in the database from 3 (the default) to 5; this is how CockroachDB is able to provide the resiliency characteristics described above.
> ```

Nit: this really allows us to survive region failures while still achieving a local quorum in the leaseholder's region (i.e. it's a performance issue, not a survivability issue).  I'd think we'd want to somehow reflect that above, but I'm not sure about the precise wording.

rmloveland (commented):
thanks Adam!  updated slightly based on your feedback - appreciate the detailed info!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10853#-:-MdSSbok-TC_55j3NkGw:b-n46pv4)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajstorm)

___
*[v21.1/multiregion-overview.md, line 107 at r1](https://reviewable.io/reviews/cockroachdb/docs/10853#-MdO7Up63jrbT-VWfKpc:-MdSSAZPEqYwSj1wKqtc:bhu1bxt) ([raw file](https://github.com/cockroachdb/docs/blob/77d36185edbac173230199911b9129b1d1ae05f4/v21.1/multiregion-overview.md#L107)):*

Nit: this really allows us to survive region failures while still achieving a local quorum in the leaseholder's region (i.e. it's a performance issue, not a survivability issue).  I'd think we'd want to somehow reflect that above, but I'm not sure about the precise wording.
</blockquote></details>

ah, thanks for explaining

updated to add " .. while maintaining a local quorum in the leaseholder's region for good performance"

ajstorm (commented):
:shipit: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10853#-:-MdSfNHT4_OWwC1dVNf0:bizlbj6)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale)

PR #10882, opened 2021-07-06
CC IP Allowlist limitation

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-4588
You can add up to 24 ip addresses via the allowlist. Use VPC peering or PrivateLink if you need to add more.


gemma-shay:
cc: @kannanlakshmi @logston Let me know if you'd rather the limit be documented as 20.

kannanlakshmi:
I am fine with 24 as that's the real limit, is there a reason you suggest 20?

gemma-shay:
@kannanlakshmi 20 was suggested in the comments here: https://cockroachlabs.atlassian.net/browse/SREOPS-2350

kannanlakshmi:
Got it, then 20 is fine. LGTM

ianjevans (approved):

PR #10883, opened 2021-07-06
TimeoutStopSecs should be TimeoutStopSec

gnat:
Typo. See: https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec=


jseldess:
Sorry for the delay, @gnat! And thanks for the fix.

@taroface, PTAL. If you approve the change, after merging this, please follow-up with a PR porting them to 21.2 docs as well.

taroface (approved):
LGTM. Thank you for the fix!

PR #10889, opened 2021-07-07
Table addition for differences between core/enterprise

### LINKED ISSUE #10793: Make clear the difference between Core and Enterprise changefeeds
People often get confused on the differences between the two, and the limitations of using each. We should create a chart or something that makes this clear on the 'Stream Data out of Cockroach using changefeeds page"

Differences that are top of mind:
Core changefeeds: Great for prototyping/ quick testing. Closes with SQL connection. 
Enterprise: Can only be used in CC dedicated or with an enterprise license. More robust and maintains connection. Distributed. Recommended for production use.

kathancox:
Closes #10793 

Added a comparison table to the top of the 20.2, 21.1 pages re enterprise & core changefeeds — replacing the bullet points

### DIFF
--- a/v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md
+++ b/v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md
@@ -12,8 +12,14 @@ While CockroachDB is an excellent system of record, it also needs to coexist wit
 
 The main feature of CDC is the changefeed, which targets an allowlist of tables, called the "watched rows". There are two implementations of changefeeds:
 
-- [Core changefeeds](#create-a-core-changefeed), which stream row-level changes to the client indefinitely until the underlying connection is closed or the changefeed is canceled.
-- [Enterprise changefeeds](#configure-a-changefeed-enterprise), where every change to a watched row is emitted as a record in a configurable format (`JSON` or Avro) to a configurable sink ([Kafka](https://kafka.apache.org/)).
+[Core changefeeds](#create-a-core-changefeed)                                                          | [Enterprise changefeeds](#configure-a-changefeed-enterprise)
+-------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------
+Useful for prototyping or quick testing.                                                               | Recommended for production use.
+Available in CockroachDB and CockroachCloud Dedicated.                                                 | Available in CockroachCloud Dedicated or with an [enterprise license](enterprise-licensing.html).
+Streams indefinitely until underlying SQL connection is closed.                                        | Maintains connection to configured sink.
+Create with [`EXPERIMENTAL CHANGEFEED FOR`](changefeed-for.html).                                      | Create with [`CREATE CHANGEFEED`](create-changefeed.html).
+Watches one or multiple tables in a comma-separated list.                                              | Emits every change to a "watched" row as a record in a configurable format (`JSON` or Avro) to a configurable sink ([Kafka](https://kafka.apache.org/)).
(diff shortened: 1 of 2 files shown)


kathancox:
@amruss re the qualification of CockroachCloud... I don't know if we feel adding "paid" to both instances would be more clear. I had initially used "Dedicated", but seems like this isn't in use/confirmed as yet.

amruss:
Hmm yeah I'm not sure what to do about the naming

v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md
```diff
@@ -12,8 +12,14 @@ While CockroachDB is an excellent system of record, it also needs to coexist wit
 
 The main feature of CDC is the changefeed, which targets an allowlist of tables, called the "watched rows". There are two implementations of changefeeds:
 
-- [Core changefeeds](#create-a-core-changefeed), which stream row-level changes to the client indefinitely until the underlying connection is closed or the changefeed is canceled.
-- [Enterprise changefeeds](#configure-a-changefeed-enterprise), where every change to a watched row is emitted as a record in a configurable format (`JSON` or Avro) to a configurable sink ([Kafka](https://kafka.apache.org/)).
+| [Core changefeeds](#create-a-core-changefeed)   | [Enterprise changefeeds](#configure-a-changefeed-enterprise) |
+--------------------------------------------------|-----------------------------------------------------------------|
+| Useful for prototyping or quick testing. | Recommended for production use. |
+| Available in CockroachDB and CockroachCloud Free Tier. | Available in CockroachCloud or with an [enterprise license](enterprise-licensing.html) in CockroachDB. |
```
amruss, line 18: Minor nit - "Available in CockroachDB and CockroachCloud Free Tier." for Core changefeeds feels like it maybe isn't on other Cockroach Cloud products? Maybe we could just say its available everywhere?

v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md
```diff
@@ -12,8 +12,14 @@ While CockroachDB is an excellent system of record, it also needs to coexist wit
...
+| Streams indefinitely until underlying SQL connection is closed. | Maintains connection to configured sink. |
+| Create with [`EXPERIMENTAL CHANGEFEED FOR`](changefeed-for.html). | Create with [`CREATE CHANGEFEED`](create-changefeed.html). |
+| Watches one or multiple tables in a comma-separated list. | Emits every change to a "watched" row as a record in a configurable format <br> (`JSON` or Avro) to a configurable sink  ([Kafka](https://kafka.apache.org/)). |
```
amruss, line 21: Core: "Watches one or multiple tables in a comma-separated list." 
- this is true for both 

Enterprise: "Emits every change to a "watched" row as a record in a configurable format(JSON or Avro) to a configurable sink (Kafka)."
- "Emits every change to a "watched" row as a record" is true for both
- the json/avro format and configurable sink are both true and unique to enterprise

amruss (changes requested):
I think there are a few clarifications needed here

v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md
kathancox: Yes, great points! I've incorporated these changes into the PR. On the naming, I think that this is changing soon and there will be more definition. I feel that for now the comparison to core changefeed being available everywhere might be OK and then I can come back in and change soon?

amruss (approved):
Thanks Kathryn, LGTM

rmloveland (approved):
LGTM!

PR #10890, opened 2021-07-07
Add call to action in CC Free FAQs

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-4586


kernfeld-cockroach:
Looks great! The only thing that I don't have the knowledge to review is `rel="noopener"`.

gemma-shay:
@kernfeld-cockroach thanks paul! Looking at other links in the docs, I believe "noopener" is correct but I updated the referral ID so we can tell when users sign up from this page.

ianjevans (approved):

PR #10894, opened 2021-07-07
New location of roachtest remote file.

ianjevans:
`master` broke due to a change in `cockroach`'s roachtest location.


gemma-shay (approved):
LGTM

PR #10917, opened 2021-07-14
Docs on how leases are transferred from dead nodes

### LINKED ISSUE #10913: doc request: describe how a lease transfer occurs after node failure
This issue is synthesizing a common request from the field. (See https://cockroachdb.zendesk.com/agent/tickets/7951 for a recent example.)

We currently have documentation about regular lease transfers here: https://www.cockroachlabs.com/docs/stable/architecture/replication-layer.html

But customers also are interested in the survivability of CockroachDB. Specifically, customers would like to know an **upper bound time** on how long a range will be unavailable after the leaseholder of the range suddenly dies.

https://github.com/cockroachdb/cockroach/pull/65195 is a recent PR that changes how a lease is transferred off of a dead node. It contains some bullet points about what is going on. This information should be included in our documentation.

```
    Wait 9s for liveness to expire
    Attempt to acquire the lease on a healthy node, this is rejected
    because lease acquisition can only happen on the raft leader
    The rejection unquiesces the range
    The nodes wait out the election timeout (3s) before trying to
    campaign
    A raft election occurs and a healthy node wins
    A lease transfer can now be processed by the newly elected leader
```

Also, the documentation should cover the special case of when the internal `liveness` range is colocated with some other lease for range "X". Does that make the lease transfer of "X" take even longer?

rmloveland:
Fixes #10913.

Summary of changes:

- Update 'Replication Layer' docs with a new section 'How leases are
  transferred from a dead node', which explains the process step by step
  at a high level.

  - Note: the steps vary slightly between the v20.* and v21.* series due
    to some recent changes in the CRDB PR linked from the fixed issue.

### DIFF
--- a/v20.2/architecture/replication-layer.md
+++ b/v20.2/architecture/replication-layer.md
@@ -69,7 +69,7 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
 
 #### Epoch-based leases (table data)
 
-To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to immediately lose all of its leases.
+To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to [lose all of its leases](#how-leases-are-transferred-from-a-dead-node) a few seconds later when the liveness record expires.
 
 Because leases do not expire until a node disconnects from a cluster, leaseholders do not have to individually renew their own leases. Tying lease lifetimes to node liveness in this way lets us eliminate a substantial amount of traffic and Raft processing we would otherwise incur, while still tracking leases for every range.
 
@@ -78,6 +78,23 @@ Because leases do not expire until a node disconnects from a cluster, leaseholde
 A table's meta and system ranges (detailed in the [distribution layer](distribution-layer.html#meta-ranges)) are treated as normal key-value data, and therefore have leases just like table data.
 
 However, unlike table data, system ranges cannot use epoch-based leases because that would create a circular dependency: system ranges are already being used to implement epoch-based leases for table data. Therefore, system ranges use expiration-based leases instead. Expiration-based leases expire at a particular timestamp (typically after a few seconds). However, as long as a node continues proposing Raft commands, it continues to extend the expiration of its leases. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
+
+#### How leases are transferred from a dead node
+
+When a node disconnects, the process by which each of its leases is transferred to a healthy node is as follows:
+
+1. The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds, and is heartbeated every 4.5 seconds. When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds.
+1. A healthy node attempts to acquire the lease. This is rejected because lease acquisition can only happen on the Raft leader, which the healthy node is not (yet). Therefore, a Raft election must be held.
+1. The rejected attempt at lease acquisition [unquiesces](../ui-replication-dashboard.html#replica-quiescence) ("wakes up") the range associated with the lease.
+1. What happens next depends on whether the lease is on [table data](#epoch-based-leases-table-data) or [meta ranges or system ranges](#expiration-based-leases-meta-and-system-ranges):
+    - If the lease is on [meta or system ranges](#expiration-based-leases-meta-and-system-ranges), the node that unquiesced the range waits out the Raft election timeout (3 seconds) before kicking off a campaign to try and win Raft leadership so it can become the leaseholder.
+    - If the lease is on [table data](#epoch-based-leases-table-data), the wait described above is skipped and an election is called immediately.
+1. The Raft election is held and a new leader is chosen from among the healthy nodes.
+1. The lease acquisition can now be processed by the newly elected Raft leader.
+
+This process should take no more than 9 seconds for liveness expiration plus the cost of 2 network roundtrips: 1 for Raft leader election, and 1 for lease acquisition. An additional 3 seconds' wait will occur if the lease is on a meta or system range.
+
+Finally, note that the process described above is lazily initiated: it only occurs when a new request comes in for the range associated with the lease.
 
 #### Leaseholder rebalancing
 
--- a/v21.1/architecture/replication-layer.md
+++ b/v21.1/architecture/replication-layer.md
@@ -83,7 +83,7 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
 
 #### Epoch-based leases (table data)
 
-To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to immediately lose all of its leases.
+To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to [lose all of its leases](#how-leases-are-transferred-from-a-dead-node) a few seconds later when the liveness record expires.
 
 Because leases do not expire until a node disconnects from a cluster, leaseholders do not have to individually renew their own leases. Tying lease lifetimes to node liveness in this way lets us eliminate a substantial amount of traffic and Raft processing we would otherwise incur, while still tracking leases for every range.
 
@@ -92,6 +92,23 @@ Because leases do not expire until a node disconnects from a cluster, leaseholde
 A table's meta and system ranges (detailed in the [distribution layer](distribution-layer.html#meta-ranges)) are treated as normal key-value data, and therefore have leases just like table data.
 
 However, unlike table data, system ranges cannot use epoch-based leases because that would create a circular dependency: system ranges are already being used to implement epoch-based leases for table data. Therefore, system ranges use expiration-based leases instead. Expiration-based leases expire at a particular timestamp (typically after a few seconds). However, as long as a node continues proposing Raft commands, it continues to extend the expiration of its leases. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
+
+#### How leases are transferred from a dead node
+
+When a node disconnects, the process by which each of its leases is transferred to a healthy node is as follows:
+
+1. The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds, and is heartbeated every 4.5 seconds. When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds.
+1. A healthy node attempts to acquire the lease. This is rejected because lease acquisition can only happen on the Raft leader, which the healthy node is not (yet). Therefore, a Raft election must be held.
+1. The rejected attempt at lease acquisition [unquiesces](../ui-replication-dashboard.html#replica-quiescence) ("wakes up") the range associated with the lease.
+1. What happens next depends on whether the lease is on [table data](#epoch-based-leases-table-data) or [meta ranges or system ranges](#expiration-based-leases-meta-and-system-ranges):
+    - If the lease is on [meta or system ranges](#expiration-based-leases-meta-and-system-ranges), the node that unquiesced the range checks if the Raft leader is alive according to the liveness record. If the leader is not alive, it kicks off a campaign to try and win Raft leadership so it can become the leaseholder.
+    - If the lease is on [table data](#epoch-based-leases-table-data), the "is the leader alive?" check described above is skipped and an election is called immediately. The check is skipped since it would introduce a circular dependency on the liveness record used for table data, which is itself stored in a system range.
+1. The Raft election is held and a new leader is chosen from among the healthy nodes.
+1. The lease acquisition can now be processed by the newly elected Raft leader.
+
+This process should take no more than 9 seconds for liveness expiration plus the cost of 2 network roundtrips: 1 for Raft leader election, and 1 for lease acquisition.
+
+Finally, note that the process described above is lazily initiated: it only occurs when a new request comes in for the range associated with the lease.
 
 #### Leaseholder rebalancing
 


rmloveland:
@lunevalex mind taking a look at this docs PR?  it's based on your notes in the recent changes in cockroachdb/cockroach#65195

@rafiss I would appreciate a look from you as well since you filed the docs issue! want to make sure you are happy with it  :-)

Note: the steps differ slightly in v20.2 and v21.1 pages due to the removal of the wait in Alex's PR - please correct anything I got wrong, thanks!

rmloveland:
PS one of the requests from the docs issue was an upper bound on the duration a range would be unavailable, so I totally handwaved the "This process should take no more than 15 seconds" line based on 9s + ... just wait longer for the $MAGIC ..., so please tell me something different to put there :-)

v20.2/architecture/replication-layer.md
```diff
@@ -69,7 +69,7 @@ To achieve this, each lease renewal or transfer also attempts to collocate them.
 
 #### Epoch-based leases (table data)
 
-To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to immediately lose all of its leases.
+To manage leases for table data, CockroachDB implements a notion of "epochs," which are defined as the period between a node joining a cluster and a node disconnecting from a cluster. To extend its leases, each node must periodically update its liveness record, which is stored on a system range key. When a node disconnects, it stops updating the liveness record, and the epoch is considered changed. This causes the node to immediately [lose all of its leases](#how-leases-are-transferred-from-a-dead-node).
```
lunevalex, line 72: the leases are actually transferred once liveness expires, we should not say immediately, as the expiration could take 9s.

v20.2/architecture/replication-layer.md
```diff
@@ -79,6 +79,21 @@ A table's meta and system ranges (detailed in the [distribution layer](distribut
 
 However, unlike table data, system ranges cannot use epoch-based leases because that would create a circular dependency: system ranges are already being used to implement epoch-based leases for table data. Therefore, system ranges use expiration-based leases instead. Expiration-based leases expire at a particular timestamp (typically after a few seconds). However, as long as a node continues proposing Raft commands, it continues to extend the expiration of its leases. If it doesn't, the next node containing a replica of the range that tries to read from or write to the range will become the leaseholder.
 
+#### How leases are transferred from a dead node
+
+When a node disconnects, the process by which each of its leases is transferred to a healthy node is as follows:
+
+1. The cluster waits 9 seconds for the dead node's liveness record, which is stored in a system range, to expire.
```
lunevalex, line 86: I would say a liveness record has an expiration time of 9s since last hearbeat, so the time you wait until expiration after a disconnect varies and on average would be 4.5s.

v20.2/architecture/replication-layer.md
```diff
@@ -79,6 +79,21 @@ A table's meta and system ranges (detailed in the [distribution layer](distribut
...
+2. A healthy node attempts to acquire the lease. This is rejected because lease acquisition can only happen on the Raft leader, which the healthy node is not (yet).  Therefore, an election must be held.
+3. The rejected attempt at lease acquisition unquiesces the system range holding the liveness record.
```
lunevalex, line 88: The range that is unquiesed is the range for which we are trying to acquire a lease which could be the liveness range if that was on the disconnected node.

v20.2/architecture/replication-layer.md
```diff
@@ -79,6 +79,21 @@ A table's meta and system ranges (detailed in the [distribution layer](distribut
...
+4. What happens next depends on whether the lease is on [table data](#epoch-based-leases-table-data) or [meta ranges or system ranges](#expiration-based-leases-meta-and-system-ranges):
+    - If the lease is on [table data](#epoch-based-leases-table-data), the node that unquiesced the range waits out the Raft election timeout (3 seconds) before kicking off a campaign to try and win Raft leadership so it can become the leaseholder.
```
lunevalex, line 90: It's the other way, we dont wait the 3s on epoch based leases, but do on expiration based leases.

v20.2/architecture/replication-layer.md
```diff
@@ -79,6 +79,21 @@ A table's meta and system ranges (detailed in the [distribution layer](distribut
...
+    - If the lease is on [meta or system ranges](#expiration-based-leases-meta-and-system-ranges), the "is the leader alive?" check described above is skipped, since it would introduce a circular dependency on the liveness record used for table data, which is itself stored in a system range.
+5. The Raft election is held and a new leader is chosen from among the healthy nodes.
+6. The lease transfer can now be processed by the newly elected Raft leader.
+
+This process should take no more than 15 seconds.
```
lunevalex, line 95: The process should be 
 - Liveness expiration (max of 9s)
 - Raft leader election (this should be 1 network roundtrip)
 - Lease transfer, this is a raft proposal (a proposal is typically committed in 1 round trip)

So this should be 9s + 2 RT, @nvanbenschoten does that math check out?

lunevalex (changes requested):

v20.2/architecture/replication-layer.md
rmloveland: Updated to say "causes the node to [lose all of its leases](#how-leases-are-transferred-from-a-dead-node) a few seconds later when the liveness record expires."

v20.2/architecture/replication-layer.md
rmloveland: Revised to "The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds.  When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 4.5 seconds."

v20.2/architecture/replication-layer.md
rmloveland: Updated to "The rejected attempt at lease acquisition unquiesces the range associated with the lease.  This range could be the one holding the liveness record, if that was on the disconnected node."

v20.2/architecture/replication-layer.md
rmloveland: Swapped 'em

rmloveland:
Thanks for the review @lunevalex - made the updates on both pages (20.2 and 21.1).  Will wait on the discussion re: how long this whole process should take.

aayushshah15:
Looks good to me. I have a suggestion though, could we explicitly talk about the fact that the process described here is lazily initiated (i.e. only when a new request comes in for the range in question). Talking about that part will change some of the back-of-the-envelope math here because the raft leadership election happens eagerly.

nvanbenschoten (commented):
This mostly looks good, though I left a few comments. I also agree with Aayush's suggestion about mentioning the lazy nature of all of this.

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10917#-:-Mf66i4X-83q0VMYjrCY:b-ai6s17)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lunevalex, @rafiss, and @rmloveland)

___
*[v20.2/architecture/replication-layer.md, line 86 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-86:-Mf658kC1P7N61hXi9pw:b-mov43c) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L86)):*

Revised to "The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds.  When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 4.5 seconds."
</blockquote></details>

I don't think this average is quite correct. The expiration time of a liveness record is 9s, but we heartbeat it every 4.5 seconds. So on average, a liveness record is 6.75s away from expiration at any given point in time.
___
*[v20.2/architecture/replication-layer.md, line 88 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-88:-Mf677NG-wEabY7JM2Y4:b-g7rxf1) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L88)):*
> ```
> This range could be the one holding the liveness record, if that was on the disconnected node.
> ```

I don't understand what we're trying to say here. My reading of these steps is that they are in the context of a specific range. So this sentence seems off.

I also wonder whether we should even mention quiescence at all. Is this discussed elsewhere in the docs?
___
*[v20.2/architecture/replication-layer.md, line 95 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-95:-Mf66damA8G54o3ZD-JC:b-vkvt6f) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L95)):*
<details><summary><i>Previously, lunevalex wrote…</i></summary><blockquote>

The process should be 
 - Liveness expiration (max of 9s)
 - Raft leader election (this should be 1 network roundtrip)
 - Lease transfer, this is a raft proposal (a proposal is typically committed in 1 round trip)

So this should be 9s + 2 RT, @nvanbenschoten does that math check out? 
</blockquote></details>

Yes, this math checks out.
___
*[v20.2/architecture/replication-layer.md, line 91 at r2](https://reviewable.io/reviews/cockroachdb/docs/10917#-Mf65kX44dweNTzz_2eZ:-Mf65kX44dweNTzz_2e_:b-kwhfe3) ([raw file](https://github.com/cockroachdb/docs/blob/d34411cb6e694117aad811cfc294c437a690dd04/v20.2/architecture/replication-layer.md#L91)):*
> ```Markdown
> 4. What happens next depends on whether the lease is on [table data](#epoch-based-leases-table-data) or [meta ranges or system ranges](#expiration-based-leases-meta-and-system-ranges):
>     - If the lease is on [meta or system ranges](#expiration-based-leases-meta-and-system-ranges), the node that unquiesced the range waits out the Raft election timeout (3 seconds) before kicking off a campaign to try and win Raft leadership so it can become the leaseholder.
>     - If the lease is on [table data](#epoch-based-leases-table-data), the "is the leader alive?" check described above is skipped, since it would introduce a circular dependency on the liveness record used for table data, which is itself stored in a system range.
> ```

Consider being more explicit here and saying that an election is called immediately.

Also, this discussion about circular dependencies seems like a bit of a distraction. Do you think it's helpful to justify why these ranges behave differently?
___
*[v20.2/architecture/replication-layer.md, line 93 at r2](https://reviewable.io/reviews/cockroachdb/docs/10917#-Mf6697rEEiOIs0KmMM5:-Mf6697rEEiOIs0KmMM6:blz7h1t) ([raw file](https://github.com/cockroachdb/docs/blob/d34411cb6e694117aad811cfc294c437a690dd04/v20.2/architecture/replication-layer.md#L93)):*
> ```Markdown
>     - If the lease is on [table data](#epoch-based-leases-table-data), the "is the leader alive?" check described above is skipped, since it would introduce a circular dependency on the liveness record used for table data, which is itself stored in a system range.
> 5. The Raft election is held and a new leader is chosen from among the healthy nodes.
> 6. The lease transfer can now be processed by the newly elected Raft leader.
> ```

Lease acquisition, not lease transfer. A lease transfer is cooperative (implies that the old leaseholder is still alive) while a lease acquisition is non-cooperative.

rmloveland (commented):
Thanks Aayush and Nathan!  I added a line to the bottom of the section

"Finally, note that the process described above is lazily initiated: it only occurs when a new request comes in for the range associated with the lease."

... and also tried to address all the other comments.  PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10917#-:-Mf8bYBf2lxC_lhx3JEx:bg4fr3p)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lunevalex, @nvanbenschoten, and @rafiss)

___
*[v20.2/architecture/replication-layer.md, line 86 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-86:-Mf8NWjS-58jCet_ZvFg:b5dwbdu) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L86)):*

I don't think this average is quite correct. The expiration time of a liveness record is 9s, but we heartbeat it every 4.5 seconds. So on average, a liveness record is 6.75s away from expiration at any given point in time.
</blockquote></details>

Updated to 

"The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds, and is heartbeated every 4.5 seconds.  When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds."
___
*[v20.2/architecture/replication-layer.md, line 88 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-88:-Mf8OGW0BeTnbEmF0CXi:b-hdsqpe) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L88)):*

> ```
> This range could be the one holding the liveness record, if that was on the disconnected node.
> ```

I don't understand what we're trying to say here. My reading of these steps is that they are in the context of a specific range. So this sentence seems off.

I also wonder whether we should even mention quiescence at all. Is this discussed elsewhere in the docs?
</blockquote></details>

Re: "this one could be holding the liveness record", I'm not sure where that sentence crept in from.  Maybe some code comments I read somewhere while researching?  Deleted it.

Re: quiescence, we have a "quiescent ranges" metric in the [replication dashboard](http://cockroachlabs.com/docs/v21.1/ui-replication-dashboard.html#replica-quiescence) in the UI, which I have linked to from the term; also added a parenthetical note that this means "wakes up"

"The rejected attempt at lease acquisition [unquiesces](../ui-replication-dashboard.html#replica-quiescence) ("wakes up") the range associated with the lease."
___
*[v20.2/architecture/replication-layer.md, line 95 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-95:-Mf8aZfw7yarpVXjK-9c:b-7ouxjv) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L95)):*

Yes, this math checks out.
</blockquote></details>

OK! For v21.1 I wrote the following:

"This process should take no more than 9 seconds for liveness expiration plus the cost of 2 network roundtrips: 1 for Raft leader election, and 1 for lease acquisition."

For v20.2 it looks like we still have the behavior that "the node that unquiesced the range waits out the Raft election timeout (3 seconds) before kicking off a campaign" which implies 9s + 3s + 2 RT, is that right for v20.2.* ?

Given the above, for v20.2 I wrote the following:

"This process should take no more than 9 seconds for liveness expiration plus the cost of 2 network roundtrips: 1 for Raft leader election, and 1 for lease acquisition.  An additional 3 seconds' wait will occur if the lease is on a meta or system range."
___
*[v20.2/architecture/replication-layer.md, line 91 at r2](https://reviewable.io/reviews/cockroachdb/docs/10917#-Mf65kX44dweNTzz_2eZ:-Mf8RPchBiQSNTHlobCe:b-i3lgkw) ([raw file](https://github.com/cockroachdb/docs/blob/d34411cb6e694117aad811cfc294c437a690dd04/v20.2/architecture/replication-layer.md#L91)):*

Consider being more explicit here and saying that an election is called immediately.

Also, this discussion about circular dependencies seems like a bit of a distraction. Do you think it's helpful to justify why these ranges behave differently?
</blockquote></details>

Re: being more explicit, done.

Re: discussing circular dependencies, I agree it's distracting.  Unfortunately I think it is necessary for the following reasons:

* we mention earlier on this page the distinction between epoch- and expiration-based leases.  Since we have introduced it above, we probably need to describe it here where we are talking about how the lease transfer works
* we already say in many places in the docs there is a thing called "system" ranges, which are special / different than "regular" ranges, and the motivating docs ticket in this case was caused by "a node is down and now you can't log in from any node because user/pass is on a system range", so since we have told people system ranges exist and are different we now probably have to say things about that difference when describing this lease transfer process (including explaining how we avoid the circular dependency)
___
*[v20.2/architecture/replication-layer.md, line 93 at r2](https://reviewable.io/reviews/cockroachdb/docs/10917#-Mf6697rEEiOIs0KmMM5:-Mf8X8d_09VwfbnJiuj1:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/d34411cb6e694117aad811cfc294c437a690dd04/v20.2/architecture/replication-layer.md#L93)):*

Lease acquisition, not lease transfer. A lease transfer is cooperative (implies that the old leaseholder is still alive) while a lease acquisition is non-cooperative.
</blockquote></details>

Fixed!

nvanbenschoten (approved):
:lgtm:

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10917#-:-Mf8tSLW3hoMH6fbl8SJ:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lunevalex and @rafiss)

taroface (approved):
LGTM with some suggestions! These are all really interesting details. All my comments on the one version should apply to the other.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10917#-:-MfA9C2V9Hp7EA1JMo_E:b-duhd52)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lunevalex, @rafiss, and @rmloveland)

___
*[v20.2/architecture/replication-layer.md, line 86 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-86:-MfA6W_q6UCq0dmFbgoT:b5zrrs) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L86)):*

Updated to 

"The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds, and is heartbeated every 4.5 seconds.  When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds."
</blockquote></details>

I suggest amending the end as follows (mostly to reiterate that the point of this is that the leases become available):

`the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds.` -> `the cluster must wait an average of 6.75 seconds for its liveness record to expire and for its leases to become available.`
___
*[v20.2/architecture/replication-layer.md, line 87 at r3](https://reviewable.io/reviews/cockroachdb/docs/10917#-MfA3vFn-yVsHZWwXrDR:-MfA3vFn-yVsHZWwXrDS:b-bq1b9j) ([raw file](https://github.com/cockroachdb/docs/blob/45e0a970ce5d3813651019b042ecb37d9326f75e/v20.2/architecture/replication-layer.md#L87)):*
> ```Markdown
> 
> 1. The dead node's liveness record, which is stored in a system range, has an expiration time of 9 seconds, and is heartbeated every 4.5 seconds. When the node dies, the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds.
> 2. A healthy node attempts to acquire the lease. This is rejected because lease acquisition can only happen on the Raft leader, which the healthy node is not (yet).  Therefore, an election must be held.
> ```

Maybe `an election` -> `a Raft election` or `an election for Raft leadership` (if accurate)

Also, can all the numbered items on this list be changed to `1.` per our general style?
___
*[v20.2/architecture/replication-layer.md, line 91 at r3](https://reviewable.io/reviews/cockroachdb/docs/10917#-MfA8TjUAaDFqDSU7wjw:-MfA8TjUAaDFqDSU7wjx:b-jgid84) ([raw file](https://github.com/cockroachdb/docs/blob/45e0a970ce5d3813651019b042ecb37d9326f75e/v20.2/architecture/replication-layer.md#L91)):*
> ```Markdown
> 4. What happens next depends on whether the lease is on [table data](#epoch-based-leases-table-data) or [meta ranges or system ranges](#expiration-based-leases-meta-and-system-ranges):
>     - If the lease is on [meta or system ranges](#expiration-based-leases-meta-and-system-ranges), the node that unquiesced the range waits out the Raft election timeout (3 seconds) before kicking off a campaign to try and win Raft leadership so it can become the leaseholder.
>     - If the lease is on [table data](#epoch-based-leases-table-data), the "is the leader alive?" check described above is skipped and an election is called immediately. The check is skipped since it would introduce a circular dependency on the liveness record used for table data, which is itself stored in a system range.
> ```

I've lost track of where the "is the leader alive?" check is. Suggest calling out the specific step here i.e. `...check described in step X`.

`an election is called immediately` -> `an election (step 5) is called immediately` or similar

Also, I am not sure the second sentence is needed here, unless we think the fact that the check is skipped raises a glaring question as to why. I feel the focus could be on the election being immediately called.

rmloveland (commented):
Thanks Ryan!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10917#-:-MfDlfdj9X4gbkw5w8kY:b7bqz4r)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @lunevalex, @rafiss, and @taroface)

___
*[v20.2/architecture/replication-layer.md, line 86 at r1](https://reviewable.io/reviews/cockroachdb/docs/10917#-MeaqxZ5227ppwq6apoe-r1-86:-MfDids52XK6D7ylKkSv:b-56nhhs) ([raw file](https://github.com/cockroachdb/docs/blob/ee5c686d5255fdd18677652c0985ab905585d615/v20.2/architecture/replication-layer.md#L86)):*

I suggest amending the end as follows (mostly to reiterate that the point of this is that the leases become available):

`the amount of time the cluster has to wait for the record to expire varies, but on average is 6.75 seconds.` -> `the cluster must wait an average of 6.75 seconds for its liveness record to expire and for its leases to become available.`
</blockquote></details>

thanks for this - but I think I'm going to avoid adding it b/c I don't want to say things yet in this step about leases becoming available, since I'm not sure that is strictly true at this step, and this has already had technical review.  I think it's OK that step 1 of this process doesn't remind us (yet) about what the ultimate goal is since it's in the section title.  Hope that's not being too fussy, just don't want to accidentally introduce a (possibly?) subtly incorrect statement
___
*[v20.2/architecture/replication-layer.md, line 87 at r3](https://reviewable.io/reviews/cockroachdb/docs/10917#-MfA3vFn-yVsHZWwXrDR:-MfDjIPZ6oBirmOHkO9l:bko5k7) ([raw file](https://github.com/cockroachdb/docs/blob/45e0a970ce5d3813651019b042ecb37d9326f75e/v20.2/architecture/replication-layer.md#L87)):*

Maybe `an election` -> `a Raft election` or `an election for Raft leadership` (if accurate)

Also, can all the numbered items on this list be changed to `1.` per our general style?
</blockquote></details>

went with "a Raft election"

changed to all 1's 
___
*[v20.2/architecture/replication-layer.md, line 91 at r3](https://reviewable.io/reviews/cockroachdb/docs/10917#-MfA8TjUAaDFqDSU7wjw:-MfDl7o6-gPuFU_sjU6g:b-gycfeo) ([raw file](https://github.com/cockroachdb/docs/blob/45e0a970ce5d3813651019b042ecb37d9326f75e/v20.2/architecture/replication-layer.md#L91)):*

I've lost track of where the "is the leader alive?" check is. Suggest calling out the specific step here i.e. `...check described in step X`.

`an election is called immediately` -> `an election (step 5) is called immediately` or similar

Also, I am not sure the second sentence is needed here, unless we think the fact that the check is skipped raises a glaring question as to why. I feel the focus could be on the election being immediately called.
</blockquote></details>

Aha!  That "is the leader alive?" check is referring to the new change in v21.1, but I mistakenly referenced it in this v20.2 page.  It isn't you who lost track, it's me!  Doing these cross-version edits with different behavior is confusing, I should have split this into separate PRs, sorry.

I have edited this to state instead

"If the lease is on [table data](#epoch-based-leases-table-data), the wait described above is skipped and an election is called immediately."

(and also remove the second sentence in this version since it makes less sense with this edit)

PR #10924, opened 2021-07-16
CC Connect improvements

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-4538


gemma-shay:
@emily-horing To align the Free and Dedicated Quickstart experiences step for step, creating a sql user and authorizing your network would need to be removed from the Dedicated guide. I just tested a new Dedicated cluster to see if this would be possible, but it looks like there is no default option and these need to be set up to proceed.
I think it might be best to leave these steps as-is, but let me know if you had something else in mind.

Amruta-Ranade (approved):
LGTM! Thanks, Gemma!!

emily-horing:
Makes sense @gemma-shay , just looking to see that the outcomes (ie sql statements in this case) are more aligned than before. This LGTM!

PR #10926, opened 2021-07-19
Syntax typo in UPDATE query example

emoriarty:
I've testing the UPDATE query and when using double quotes, the command returns the following error:
```sql
ERROR: column "eu-west" does not exist
```
Just changing double quotes by single does the fix.


jseldess:
Thanks, @emoriarty!

@rmloveland, PTAL.

rmloveland (approved):
LGTM, thanks @emoriarty - sorry for delay on my part

Issue #10222, opened 2021-04-01
Document using PGBouncer with CockroachDB


RichardJCai:
Richard Cai (RichardJCai) commented:

I created a rough tutorial on using CRDB with PGBouncer / HAProxy but I think the goal is to make a more formal doc on it.

Reference: https://github.com/RichardJCai/crdb-pgbouncer/blob/main/README.md

In general it seems CRDB is compatible with PGBouncer however there may not be much benefit to using it as CRDB idle connections are relatively cheap (compared to PG). We can provide guidance stating that it is likely unnecessary to use but may be beneficial if there is an extremely high number of connections relative to the hardware specs and it's best that they test it themselves to determine if it's necessary to use. 

Users can use PGBouncer if it provides application side benefits such as simplifying connection pooling.

Jira Issue: DOC-1077

RichardJCai:
cc @ericharmeling for awareness

ericharmeling:
@ianjevans

vy-ton:
@ericharmeling or @ianjevans 

The team discussed and we agreed that we should go ahead and add PgBouncer to the community tooling page with a link to https://dzone.com/articles/using-pgbouncer-with-cockroachdb. We can do this as a stopgap before investing in any more full fledged PgBouncer examples

PR #10930, opened 2021-07-20
Added PGBouncer to list of community-supported tools

ericharmeling:
Related to https://github.com/cockroachdb/docs/issues/10222.


ericharmeling:
@vy-ton Not sure if this should resolve https://github.com/cockroachdb/docs/issues/10222 or not...

vy-ton:
Let's keep #10222 for tracking but it can be deprioritized to P2

vy-ton (approved):

PR #10967, opened 2021-07-26
Adding known limitation for inverted index scans.

### LINKED ISSUE #10966: Add known limitation for inverted index scans on JSON columns with containment operators
From https://github.com/cockroachdb/cockroach/issues/55318

v20.2 and v21.1 have the following known limitation on inverted indexes:
CockroachDB cannot generate inverted index scans for statements with filters that have both JSON fetch values and containment operators. For example:

```
SELECT * FROM mytable WHERE j->'a' @> '{"b": "c"}';
```

This was possible in v20.1 and earlier, though the normalization rules used to convert the filters into JSON containment expressions would sometimes produce inequivalent expressions.

The workaround is to rewrite the statements to avoid using both JSON fetch values and containment operators. The following statement is equivalent to the non-index-accelerated statement above:

```
SELECT * FROM mytable WHERE j @> '{"a": {"b": "c"}}'
```

ianjevans:
Fixes #10966
Related to cockroachdb/cockroach#55318 

Adds a known-issue to v21.1 and v20.2 for statements that use inverted indexes and filters with both JSON fetch values and containment operators.

### DIFF
--- a/v20.2/known-limitations.md
+++ b/v20.2/known-limitations.md
@@ -181,7 +181,7 @@ Once restarted, monitor the Replica Quiescence graph on the [**Replication Dashb
 Resolved as of [v20.2.4](../releases/v20.2.4.html). See [#57789](https://github.com/cockroachdb/cockroach/pull/57789).
 {{site.data.alerts.end}}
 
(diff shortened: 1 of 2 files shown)


mgartner (approved):
Thank you! LGTM.

kathancox (approved):
LGTM!

PR #10970, opened 2021-07-26
Update v20.2 architecture docs re: range merges

### LINKED ISSUE #10577: Distribution Layer in Architecture does not mention range merges
We make no reference to range merging in https://www.cockroachlabs.com/docs/v20.2/architecture/distribution-layer.html.

We should do so in much the same way that we reference range splitting on the page.

rmloveland:
Fixes #10577

Summary of changes:

- Port range merges section from v21.1 Distribution Layer page to v20.2
  page

- Fix up all the links to point to the above

- Delete the old standalone 'Range Merges' page

- Finally, sneak in a little change to both the v20.2 and v21.1 pages:
  add 'Range merges' to the list of things that cause range descriptor
  updates


rmloveland:
Hi Nathan

this is mostly a "backport" of range merges info to the v20.2 dist layer docs per the linked issue  _except_ that I also added "range merges" to the list of things that causes a range descriptor to be updated in both the v20.2 and v21.1 docs (which ISTR you mentioned in another issue or I read somewhere?).  Anyway happy to make any updates as needed

nvanbenschoten (approved):

PR #10982, opened 2021-07-28
Options for recursive upload example

### LINKED ISSUE #10978: Add scripting example for DB cockroach userfile upload
In versions prior to 21.2 `userfile upload` can't recursively upload files. For example, in a backup where there is a collection of files with a `data` subdirectory this would mean uploading the files individually. 

Provide an example workaround with a script, something like: 
`BACKUP DATABASE to userfile` 
`userfile get`
`userfile upload (with script)`
`RESTORE DATABASE from userfile`

kathancox:
Closes #10978 

Added a **potential** example to the `cockroach userfile upload` page to provide an easier method to upload recursively in 21.1 and 20.2 where the `--recursive` flag is not going to be present. 

Bash script or piped command ideas included in this commit, not sure which would be preferable.

### DIFF
--- a/_includes/v21.1/userfile-examples/backup-userfile.md
+++ b/_includes/v21.1/userfile-examples/backup-userfile.md
@@ -4,8 +4,6 @@ We recommend starting backups from a time at least 10 seconds in the past using 
 Only database and table-level backups are possible when using `userfile` as storage. Restoring cluster-level backups will not work because `userfile` data is stored in the `defaultdb` database, and you cannot restore a cluster with existing table data.
 {{site.data.alerts.end}}
 
-#### Database and table
-
 When working on the same cluster, `userfile` storage allows for database and table-level backups.
 
 First, run the following statement to backup a database to a directory in the default `userfile` space:
@@ -18,6 +16,8 @@ This directory will hold the files that make up a backup; including the manifest
 
 {{site.data.alerts.callout_info}}
 When backing up from a cluster and restoring a database or table that is stored in your `userfile` space to a different cluster, you can run [`cockroach userfile get`](../{{site.versions["stable"]}}/cockroach-userfile-get.html) to download the backup files to a local machine and [`cockroach userfile upload --url {CONNECTION STRING}`](../{{site.versions["stable"]}}/cockroach-userfile-upload.html) to upload to the `userfile` of the alternate cluster.
+
+For an example, read [Upload a backup directory recursively](../{{site.versions["stable"]}}/cockroach-userfile-upload.html#upload-a-backup-directory-recursively).
 {{site.data.alerts.end}}
 
 In cases when your database needs to be restored, run the following:
--- a/v21.1/cockroach-userfile-upload.md
+++ b/v21.1/cockroach-userfile-upload.md
@@ -202,7 +202,7 @@ database-backup/
 
 If you should need to recover that data, or, apply it for testing purposes, you can now upload these files to a cluster with `cockroach userfile upload` and [`RESTORE`](restore.html).
 
-For the purposes of this examples, [`DROP`](drop.html) the database:
+For the purposes of this examples, [`DROP`](drop-database.html) the database:
 
 ~~~sql
 DROP DATABASE movr CASCADE;
--- a/v21.1/cockroach-userfile-upload.md
+++ b/v21.1/cockroach-userfile-upload.md
@@ -125,6 +125,181 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
 successfully uploaded to userfile://testdb.public.uploads/test-data.csv
 ~~~
 
+### Upload a backup directory recursively
+
+Currently in v21.1 and prior, it's [not possible to use `cockroach userfile upload` recursively](#known-limitation) when uploading a directory of files that also contains a subdirectory. When you need to upload a directory of files, such as a backup, to `userfile` storage, it is possible to programmatically upload your files.
+
+The following example uses `userfile` storage and its features to backup a database and then restore it to the cluster. When quick testing is required, or, to protect against accidental loss of data, this workflow provides a way to recover in those scenarios when using `userfile` as your storage option.
+
+After connecting to the cluster that contains the data you would like to backup, run the [`BACKUP`](backup.html) statement:
+
+~~~sql
+BACKUP DATABASE movr TO 'userfile://defaultdb.public.userfiles_$user/database-backup' AS OF SYSTEM TIME '-10s';
+~~~
+
+This will make a backup of the database under `database-backup` in the default `userfile` space.
+
+~~~
+job_id             |  status   | fraction_completed | rows | index_entries | bytes
+-------------------+-----------+--------------------+------+---------------+---------
+679645548255936513 | succeeded |                  1 | 2997 |          1060 | 492713
+(1 row)
+~~~
+
+Next, on the command line, use the following [`cockroach userfile get`](cockroach-userfile-get.html) command to fetch the files stored in the `userfile` storage and download them to your local directory. (Note that without passing a specific directory this command will fetch all files stored within the default user space in `userfile`.)
+
+~~~shell
+cockroach userfile get 'userfile://defaultdb.public.userfiles_$user/database-backup' --url {CONNECTION STRING}
+~~~
+
+The output will show the files downloading:
+
+~~~
+downloaded database-backup/BACKUP-CHECKPOINT-679645548255936513-CHECKSUM to database-backup/BACKUP-CHECKPOINT-679645548255936513-CHECKSUM (4 B)
+downloaded database-backup/BACKUP-CHECKPOINT-CHECKSUM to database-backup/BACKUP-CHECKPOINT-CHECKSUM (4 B)
+downloaded database-backup/BACKUP-STATISTICS to database-backup/BACKUP-STATISTICS (53 KiB)
+downloaded database-backup/BACKUP_MANIFEST to database-backup/BACKUP_MANIFEST (2.6 KiB)
+downloaded database-backup/BACKUP_MANIFEST-CHECKSUM to database-backup/BACKUP_MANIFEST-CHECKSUM (4 B)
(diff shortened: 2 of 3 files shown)


kathancox:
Hey @livlobo / @mwang1026, I've put together a v rough draft for a possible example for **_21.1 and prior_** re `--recursive` not being present with `userfile upload`. In the ^ link above there are two examples (marked with "NOTE" & "alternative), one using a bash script, one using a piped command. My scripting skills aren't the best :upside_down_face:. 

Also, lmk @livlobo if this was the kind of example workflow you were thinking of.

Thanks!

livlobo:
@kathancox - thanks for putting this together! 

The example LGTM. I think both example are equally clear - I don't have a preference :) 

I did notice the bash output has an error for the sub-dir though; so maybe the piped command is simpler/cleaner.

kathancox:
> I did notice the bash output has an error for the sub-dir though; so maybe the piped command is simpler/cleaner.

@livlobo Yes! that error is because it's complaining about the subdirectory during the first pass `database-backup/*`. There might be a better way to script that than including both of the paths to access everything. 

But for that reason, the command does feel cleaner.

livlobo (approved):
The example looks good to me! Thanks @kathancox

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+successfully deleted database-backup/BACKUP_MANIFEST
+successfully deleted database-backup/BACKUP_MANIFEST-CHECKSUM
+successfully deleted database-backup/data/679645557047099393.sst
+. . .
+successfully deleted database-backup/data/679645558154264579.sst
+~~~
+
+Currently, `cockroach userfile upload` will not recursively upload files from a directory. See this [known limitation](../{{site.versions["stable"]}}/cockroach-userfile-upload.html#known-limitation). It is possible to programmatically upload your files from the command line.
+
+For example, the following command finds all files under `database-backup` and pipes them through to [`xargs`](https://linux.die.net/man/1/xargs), which will execute `cockroach userfile upload` for each file. Every occurrence of `{}` in the command, is replaced with the filename.
```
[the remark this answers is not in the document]
rmloveland: (ok I just looked this up and it's the latter: https://unix.stackexchange.com/questions/616899/xargs-replstr-not-working-as-expected-not-accepted-as-a-second-argument)

in that case I recommend removing this sentence and letting users read the `xargs` man page / etc.  on the reasoning that by saying just a little about it is possibly more confusing than just telling the user "use this exact command" and leaving it at that.  by mentioning it, I think there is something there I need to understand (but in this case AFAICT I don't)

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
 successfully uploaded to userfile://testdb.public.uploads/test-data.csv
 ~~~
 
+### Upload a backup directory recursively
+
+Currently in v21.1 and prior, it's [not possible to use `cockroach userfile upload` recursively](#known-limitation) when uploading a directory of files. When you need to upload a directory (such as a backup) to `userfile` storage, it is possible to programmatically upload your files.
+
+The following example uses `userfile` storage and its features to backup a database and then restore it to a cluster. When quick testing is required, or, to protect against accidental loss of data, this workflow provides a way to recover in those scenarios when using `userfile` as your storage option.
```
rmloveland, line 132: Several versions of comment, your choice which makes more sense

-----

Minimum comment is:

> or, to protect

unnecessary comma

-----

Medium comment is:

tl;dr: this clause is not really necessary:

> When quick testing is required, or, to protect against accidental loss of data, ...

and the sentence could be: "This workflow provides ..."

The way this is phrased with the mention of accidental data loss up front is a little scary.  Also I'm not sure I fully understand what is meant.  Below, it doesn't seem like I am going to have any data loss using these instructions (at least based on my imperfect understanding).  It looks like I am doing 

CRDB userfile -> my file system -> CRDB userfile -> CRDB SQL RESTORE

which seems pretty straightforward with no data loss things happening (modulo errors etc of course, as always)

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+
+{{site.data.alerts.callout_info}}
+Only database and table-level backups are possible when using `userfile` as storage. Restoring cluster-level backups will not work because `userfile` data is stored in the `defaultdb` database, and you cannot restore a cluster with existing table data.
```
rmloveland, line 135: Could link to "cluster-level backups" so I can review that option as well before deciding

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+downloaded database-backup/BACKUP-STATISTICS to database-backup/BACKUP-STATISTICS (53 KiB)
+downloaded database-backup/BACKUP_MANIFEST to database-backup/BACKUP_MANIFEST (2.6 KiB)
+downloaded database-backup/BACKUP_MANIFEST-CHECKSUM to database-backup/BACKUP_MANIFEST-CHECKSUM (4 B)
+downloaded database-backup/data/679645557047099393.sst to database-backup/data/679645557047099393.sst (5.8 KiB)
+. . .
+downloaded database-backup/data/679645558404448257.sst to database-backup/data/679645558404448257.sst (5.8 KiB)
+downloaded database-backup/data/679645558408249345.sst to database-backup/data/679645558408249345.sst (41 KiB)
+~~~
+
+At this point, you have backed up the database to `userfile` storage and you have a copy of that backup on your local machine.
```
rmloveland, line 175: this would be clearer to me as something like

> At this point you have two copies of the data:
> - One in the database's `userfile` storage
> - One on your local machine

(Maybe the interwebs broke my brain so I don't read good anymore unless in listicle format tho)

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+├── BACKUP_MANIFEST-CHECKSUM
+└── data
+    ├── 679645557047099393.sst
+    ├── 679645557048475649.sst
+    ├── 679645557048868865.sst
+    ├── 679645558151741442.sst
+. . .
+~~~
+
+If you should need to recover that data, or, apply it for testing purposes, you can now upload these files to a cluster with `cockroach userfile upload` and [`RESTORE`](restore.html).
```
rmloveland, line 194: This comment may be unnecessary / a bit flow-breaking since we are doing this below

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+
+For the purposes of this example, [`DROP`](drop-database.html) the database:
```
rmloveland, line 196: If you delete the line above, you can also remove "For the purposes of this example"

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+
+~~~sql
+DROP DATABASE movr CASCADE;
+~~~
+
+And then, use [`cockroach userfile delete`](cockroach-userfile-delete.html) to remove this from `userfile`:
```
rmloveland, line 202: Suggest something like "Next, use ..."

v21.1/cockroach-userfile-upload.md
```diff
@@ -125,6 +125,126 @@ cockroach userfile upload /Users/maxroach/Desktop/test-data.csv userfile://testd
...
+successfully deleted database-backup/BACKUP-CHECKPOINT-CHECKSUM
+successfully deleted database-backup/BACKUP-STATISTICS
+successfully deleted database-backup/BACKUP_MANIFEST
+successfully deleted database-backup/BACKUP_MANIFEST-CHECKSUM
+successfully deleted database-backup/data/679645557047099393.sst
+. . .
+successfully deleted database-backup/data/679645558154264579.sst
+~~~
+
+Currently, `cockroach userfile upload` will not recursively upload files from a directory. See this [known limitation](../{{site.versions["stable"]}}/cockroach-userfile-upload.html#known-limitation). It is possible to programmatically upload your files from the command line.
```
rmloveland, line 219: "Currently" not needed. (TOTAL NIT, sorry!!)

v21.1/cockroach-userfile-upload.md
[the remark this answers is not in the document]
rmloveland: further comments:

1. suggest editing to even higher level e.g. "the following command finds all files under ... and runs `cockroach userfile upload` on each file ..."  and folks can read the script / read about `xargs` elsewhere if they care.  

2. Alternatively if we are mentioning `xargs` and linking to it we should do the same for `find`.  However I would maybe recommend the approach 1 above since there are many different `find` s out there, and probably many `xargs` as well

rmloveland (approved):
LGTM with a few comments.  I think the ones I believe the most are re: the `find | xargs` part and re: removing the mention of "data loss" for psychological priming reasons

PS sweet tutorial, I learned some `xargs` stuff today, thanks!

_includes/v21.1/userfile-examples/backup-userfile.md
```diff
@@ -4,8 +4,6 @@ We recommend starting backups from a time at least 10 seconds in the past using
 Only database and table-level backups are possible when using `userfile` as storage. Restoring cluster-level backups will not work because `userfile` data is stored in the `defaultdb` database, and you cannot restore a cluster with existing table data.
 {{site.data.alerts.end}}
 
-#### Database and table
-
```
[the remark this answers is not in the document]
kathancox: Haha, yes it was!

PR #10985, opened 2021-07-29
Technical advisory for TRUNCATE TABLE during index build

postamar:


postamar:
I'm guessing that CI is failing because I'm basing this PR on my forked repo. I'll try without the fork now.

PR #10988, opened 2021-07-29
Update MR docs to recommend 250ms max-offset

### LINKED ISSUE #10920: kv: lower clock max-offset recommendation 250ms for most clusters
After further discussion among the multi-region team and great analysis for the CC SRE team, we'd like to propose lowering the[ max clock offset ](https://www.cockroachlabs.com/docs/v21.1/operational-faqs.html#what-happens-when-node-clocks-are-not-properly-synchronized) for all new self-hosted clusters from 500ms to 250ms. This will provide a number of improvements including lowering the write latency for Global tables. 

We can't yet recommend converting existing clusters to 250ms because it requires taking the entire cluster offline (although, if the customer can tolerate taking the cluster down it should be safe to adjust). The caveat is that last paragraph in the Clock synchronization section as we do not want vMotion customers to go below 500 ms unless special care is taken.

rmloveland:
Fixes #10920


awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10988)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nvanbenschoten and @rmloveland)

___
*[_includes/v21.1/misc/multiregion-max-offset.md, line 1 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjYab-FnyBbktpM_U:-MgCjYab-FnyBbktpM_V:b-4atzt0) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/_includes/v21.1/misc/multiregion-max-offset.md#L1)):*
> ```Markdown
> For new clusters using the [multi-region SQL abstractions](multiregion-overview.html), we recommend lowering the [`--max-offset`](cockroach-start.html#flags-max-offset) setting to `250ms`.  This is especially helpful for lowering the write latency of [global tables](multiregion-overview.html#global-tables).
> ```

Maybe add a note here that says "adjusting this setting requires taking the cluster offline and should be done with care" or something similiar 
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjjol6BmeSHks4Ecr:-MgCjjol6BmeSHks4Ecs:b-fe4jxe) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/v21.1/when-to-use-regional-vs-global-tables.md#L22)):*
> ```Markdown
> 
> - Your application has a "read-mostly" table of reference data that is rarely updated, and that needs to be available to all regions.
> - You can accept that writes to the table will incur higher latencies from any given region, since writes use a novel [non-blocking transaction protocol](architecture/transaction-layer.html#non-blocking-transactions) that uses a timestamp "in the future".
> ```

Maybe mention here that the observed write latency is dependent on the max offset value which in turn needs to be carefully adjusted based on the clocks on your machines

rmloveland (commented):
TFTR Andy!

Nathan, are you OK with these changes?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10988#-:-MgpnxXS9fBLxCVfeuOI:b-ri142w)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187 and @nvanbenschoten)

___
*[_includes/v21.1/misc/multiregion-max-offset.md, line 1 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjYab-FnyBbktpM_U:-MgpkEV-ExsScZSwDkTX:b-xgdbz1) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/_includes/v21.1/misc/multiregion-max-offset.md#L1)):*

Maybe add a note here that says "adjusting this setting requires taking the cluster offline and should be done with care" or something similiar 
</blockquote></details>

Updated to note that restart is required!
___
*[v21.1/when-to-use-regional-vs-global-tables.md, line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjjol6BmeSHks4Ecr:-MgpmFWQ1r53UE9Q2Gfq:b-jj4bg0) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/v21.1/when-to-use-regional-vs-global-tables.md#L22)):*

Maybe mention here that the observed write latency is dependent on the max offset value which in turn needs to be carefully adjusted based on the clocks on your machines
</blockquote></details>

Added a note that this depends on max-offset!

nvanbenschoten (approved):
:lgtm: thanks Rich! I just have one comment on our wording around restarting the cluster to change the max-offset.

> Reviewed 5 of 7 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10988#-:-Mgq_eu29FLThwTTcj7l:b-4l0fiz)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187)

___
*[_includes/v21.1/misc/multiregion-max-offset.md, line 1 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjYab-FnyBbktpM_U:-Mgq_NxiClGF3X-SWu0R:b-ak3udt) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/_includes/v21.1/misc/multiregion-max-offset.md#L1)):*

Updated to note that restart is required!
</blockquote></details>

I think we'll need to be even more explicit here. This doesn't just require restarting the nodes in the cluster, it requires restarting them all at the same time. In other words, this can't be done in a rolling fashion.

rmloveland (commented):
Thanks Nathan!  Updated!

PS Sorry for the notifications noise from the huge rebase on this one.  We just created the new "v21.2" folder and I had to rebase on that to get these changes into the new docs as well

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/10988#-:-Mgvesa84P4kXxl9sce4:b-7lhcrh)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187)

___
*[_includes/v21.1/misc/multiregion-max-offset.md, line 1 at r1](https://reviewable.io/reviews/cockroachdb/docs/10988#-MgCjYab-FnyBbktpM_U:-MgvecX27R8hKmJq1lXS:bbyyc0c) ([raw file](https://github.com/cockroachdb/docs/blob/f5115a2e8094d3fcbee502a6cbb90cb45da0ba23/_includes/v21.1/misc/multiregion-max-offset.md#L1)):*

I think we'll need to be even more explicit here. This doesn't just require restarting the nodes in the cluster, it requires restarting them all at the same time. In other words, this can't be done in a rolling fashion.
</blockquote></details>

OK - updated to say "Note that this will require restarting all of the nodes in your cluster at the same time; it cannot be done with a rolling restart."

nvanbenschoten (approved):

PR #10994, opened 2021-08-02
Clarify that REGIONAL BY ROW column cannot be nullable

stevendanna:


jseldess:
Thanks, @stevendanna!

@rmloveland, PTAL. If you agree with the change, please follow-up with a PR to port this to 21.2 docs as well.

rmloveland (approved):
LGTM thanks!!! Merging ...

PR #11002, opened 2021-08-03
Adjustments to resolved option language

### LINKED ISSUE #10892: Clarify resolved setting
It is not obvious right now that resolved timestamp setting is the minimum time you will get a resolved message. We only send a resolved when we advance the change frontier. if the frontier hasn't advanced, you will not get a message.

kathancox:
Closes #10892 

Edited the description of the `resolved` option with changefeeds. "Periodic" may have been misleading readers to think that the resolved timestamp events would be emitted no matter what on a particular schedule. Also, added a sentence to detail the idea that resolved timestamps won't emit unless all ranges in the changefeed have advanced.

Made adjustments to 20.2 & 21.1, as well as enterprise and core. Although this is available in core changefeeds, adjusted language since sinks aren't a thing here. Might be more adjustments needed there.


kathancox:
also, @amruss I feel like how these resolved timestamp events emit depending on the sink, is a good candidate for the type of nuance re sink-specific docs.

kathancox:
Note to self @kathancox circle back on slack convo re resolved on 8/13 in cdc channel.

amruss (approved):
LGTM

rmloveland (approved):
:lgtm: with various optional comments

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11002#-:-MhOhVTJ-zSuJ1CQQhn9:b-jsvo4s)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @kathancox)

___
*[v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md, line 81 at r1](https://reviewable.io/reviews/cockroachdb/docs/11002#-MhOgDNm1gGIlEDJFqC4:-MhOgDNm1gGIlEDJFqC5:b-1fql4j) ([raw file](https://github.com/cockroachdb/docs/blob/be73db4ce2fe4f3ad368be33b48a9a7adef1522a/v20.2/stream-data-out-of-cockroachdb-using-changefeeds.md#L81)):*
> ```Markdown
> - Rows are sharded between Kafka partitions by the row’s [primary key](primary-key.html).
> 
> - <a name="resolved-def"></a>The `UPDATED` option adds an "updated" timestamp to each emitted row. You can also use the [`RESOLVED` option](create-changefeed.html#options) to emit "resolved" timestamp messages to each Kafka partition. A "resolved" timestamp is a guarantee that no (previously unseen) rows with a lower update timestamp will be emitted on that partition.
> ```

strictly optional, but this would be even cooler if it pointed to an anchor at the exact table element with the `resolved` option

would require adding something like `<a name="resolved-option"></a>` to the target
___
*[v21.1/changefeed-for.md, line 52 at r1](https://reviewable.io/reviews/cockroachdb/docs/11002#-MhOgcbU60fn11e6t8W1:-MhOgcbU60fn11e6t8W2:b8th5ei) ([raw file](https://github.com/cockroachdb/docs/blob/be73db4ce2fe4f3ad368be33b48a9a7adef1522a/v21.1/changefeed-for.md#L52)):*
> ```Markdown
> -------|-------|------------
> `updated` | N/A | Include updated timestamps with each row.
> `resolved` | [`INTERVAL`](interval.html) | Emits [resolved timestamp](stream-data-out-of-cockroachdb-using-changefeeds.html#resolved-def) events for the changefeed. Resolved timestamp events do not emit until all ranges in the changefeed have progressed to a specific point in time. <br><br>Set an optional minimal duration between emitting resolved timestamps. Example: `resolved='10s'`. This option will **only** emit a resolved timestamp event if the timestamp has advanced and at least the optional duration has elapsed. If unspecified, all resolved timestamps are emitted as the high-water mark advances.
> ```

Judgment call: this looks like maybe the 3rd use of this text.  Might be worth considering an include (but maybe they don't work inside tables? IDK - also IDK if the text varies ever so slightly for technical reasons, but it looks about the same across the pages to me as a lay reader of this)
___
*[v21.1/create-changefeed.md, line 100 at r1](https://reviewable.io/reviews/cockroachdb/docs/11002#-MhOh6GkEFVBAao418AJ:-MhOh6GkEFVBAao418AK:b-i6pi4a) ([raw file](https://github.com/cockroachdb/docs/blob/be73db4ce2fe4f3ad368be33b48a9a7adef1522a/v21.1/create-changefeed.md#L100)):*
> ```Markdown
> -------|-------|------------
> `updated` | N/A | Include updated timestamps with each row.<br><br>If a `cursor` is provided, the "updated" timestamps will match the [MVCC](architecture/storage-layer.html#mvcc) timestamps of the emitted rows, and there is no initial scan. If a `cursor` is not provided, the changefeed will perform an initial scan (as of the time the changefeed was created), and the "updated" timestamp for each change record emitted in the initial scan will be the timestamp of the initial scan. Similarly, when a [backfill is performed for a schema change](stream-data-out-of-cockroachdb-using-changefeeds.html#schema-changes-with-column-backfill), the "updated" timestamp is set to the first timestamp for when the new schema is valid.
> `resolved` | [`INTERVAL`](interval.html) | Emits [resolved timestamp](stream-data-out-of-cockroachdb-using-changefeeds.html#resolved-def) events per changefeed in a format dependent on the connected sink. Resolved timestamp events do not emit until all ranges in the changefeed have progressed to a specific point in time. <br><br>Set an optional minimal duration between emitting resolved timestamps. Example: `resolved='10s'`. This option will **only** emit a resolved timestamp event if the timestamp has advanced and at least the optional duration has elapsed. If unspecified, all resolved timestamps are emitted as the high-water mark advances.
> ```

re: previous comment about using includes, I think this is the same text as above? same caveats etc apply
___
*[v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md, line 80 at r1](https://reviewable.io/reviews/cockroachdb/docs/11002#-MhOhN4jCHDNihtPw1Bq:-MhOhN4jCHDNihtPw1Br:bryuttr) ([raw file](https://github.com/cockroachdb/docs/blob/be73db4ce2fe4f3ad368be33b48a9a7adef1522a/v21.1/stream-data-out-of-cockroachdb-using-changefeeds.md#L80)):*
> ```Markdown
> - Rows are sharded between Kafka partitions by the row’s [primary key](primary-key.html).
> 
> - <a name="resolved-def"></a>The `UPDATED` option adds an "updated" timestamp to each emitted row. You can also use the [`RESOLVED` option](create-changefeed.html#options) to emit "resolved" timestamp messages to each Kafka partition. A "resolved" timestamp is a guarantee that no (previously unseen) rows with a lower update timestamp will be emitted on that partition.
> ```

same as above re: maybe a more targeted anchor would be even cooler!  this is fine too

PR #11032, opened 2021-08-10
Product name liquid variables

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-4765

In preparation for serverless:
CockroachCloud Free (beta) -> CockroachDB Serverless (beta)
CockroachCloud -> CockroachDB Managed
CockroachDB Core -> CockroachDB Core
CockroachDB Enterprise -> CockroachDB Self-Hosted


gemma-shay:
cc @ianjevans

ianjevans:
We should also update the `netlify/vale/CockroachDB/Terminology.yml` config file in this PR to warn when using `CockroachCloud` etc.

ianjevans (approved):
LGTM, just need to resolve the conflicts.

mikeCRL:
@gemma Thanks again for your work here. Something I noticed that might have already been considered but I'll share in case it's helpful: With `products.db` intended to later become `CockroachDB`, I noticed that some instances of that variable will work fine as "CockroachDB" (typically cases we're we simply mean 'the system' is doing something), and others won't (typically cases where we're trying to differentiate a CC product's capabilities or instructions from CockroachDB Core or Self-Hosted).

As an example of the latter, in `/cockroachcloud/backups-page.md` the change is

old: `you can [back up and restore data manually](run-bulk-operations.html) or [back up from a self-hosted CockroachDB cluster and restore into a CockroachCloud cluster](#back-up-a-self-hosted-cockroachdb-cluster-and-restore-into-a-cockroachcloud-cluster)`  
new: `you can [back up and restore data manually](run-bulk-operations.html) or [back up from a self-hosted CockroachDB cluster and restore into a {{ site.data.products.db }} cluster](#back-up-a-self-hosted-cockroachdb-cluster-and-restore-into-a-cockroachcloud-cluster)`

This would become "back up from a self-hosted CockroachDB cluster and restore into a CockroachDB cluster"

I think we can move forward with this PR, because you've left all _existing_ instances of `CockroachDB` intact, so we should have no problem going back after this is merged and more deeply auditing the `products.db` instances to determine what to do for each one. Does that make sense? We could include this in a follow up issue.

Also, we seem to have missed instances of lowercase-E enterprise that should've been uppercase, e.g. this diff. Am I right that we should review `enterprise` as well and typically turn it into a var? 

```
-{{site.data.alerts.callout_info}}Partioning is an enterprise feature. To request and enable a trial or full enterprise license, see <a href="enterprise-licensing.html">Enterprise Licensing</a>.{{site.data.alerts.end}}
+{{site.data.alerts.callout_info}}Partioning is an enterprise feature. To request and enable a trial or full enterprise license, see <a href="enterprise-licensing.html">{{ site.data.products.enterprise }} Licensing</a>.{{site.data.alerts.end}}
```

gemma-shay:
Totally agree re: the CockroachDB confusion. I think we might be able to say "CockroachDB Cloud" or something similar, but was going to wait til marketing was finalized then follow up. Good catch on enterprise, also, will update this aft!

PR #11053, opened 2021-08-15
Add doc comment markup to generate issues from release notes for v21.1.4

mikeCRL:
Also reviewed v21.1.5 but no bullets required these for conversion to release notes.


jseldess:
@mikeCRL, at this point, I don't think we need to merge this, since all of the issues have long been generated.

PR #11057, opened 2021-08-16
Add docs for migrating to v21.1+ multiregion SQL

### LINKED ISSUE #11017: Add docs for migrating to v21.1+ multiregion SQL abstractions
Extremely high level summary of procedure:

1. Remove old bad zone configs
2. Do ... various things ... maybe add nodes ... think positive thoughts ... etc.
3. Add back new good zone configs automagically via typing in SQL statements using the new [multiregion abstractions](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html)
4. ???
5. PROFIT

We also need to add a mapping table from "old new thing" to "new new thing", e.g., duplicate indexes to `GLOBAL` tables.

Note: Also need to figure out timing / perf issues involved in the switchover.  May want to recommend just doing it all at once to get to the new end state ASAP. May result in a performance hit while the cluster applies the new zone configs under the hood, shuffles data around, etc. May need to recommend that this procedure should probably be done off-hours to reduce/mitigate perf impact for app users (TBD).

This doc goes in:

- v21.1
- (eventually) port to v21.2 ( insert :dijkstra-i-would-not-like-it: )
- Still exist in v22.1 docs but with fewer links (maybe? seems to me this kind of manual zone config procedure will be rather outre by then ...)

rmloveland:
Fixes #11017

Summary of changes:

- Add a new page, 'Migrate to Multi-Region SQL', which has:

  - Mappings from each of the legacy replication-zone-based patterns and
    the newer multi-region SQL abstractions that are designed to replace
    them.

  - Instructions for migrating from replication-zone-based workflows to
    multi-region SQL abstractions.

  - A review of the zone configuration changes that result from running
    the multi-region SQL statements.


awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @nvanbenschoten, @otan, and @rmloveland)

___
*[v21.1/migrate-to-multiregion-sql.md, line 11 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7UV5E-Ojru8l0uT_s:-MiJggfpDID01nLEEQ2N:b-ri770p) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L11)):*

Do we also want to call out the old topology patterns explicitly?  That may resonate with people more than the raw zone configs.
</blockquote></details>

Yes...I'd keep what's written but add in the previous pattern names.
___
*[v21.1/migrate-to-multiregion-sql.md, line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7V0Sq6Q36u0sMEZPm:-MiJgqnbEWWDvu0LO8pw:b2ngthx) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L22)):*

Nit: Does it make sense to move this right up to the top of the document?  Might help short-circuit things faster.
</blockquote></details>

+1 
___
*[v21.1/migrate-to-multiregion-sql.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7WScA1nO7sdgz3YK2:-MiJgugj0R35nh88r4D6:b-xwofou) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L31)):*

Does it also make sense to mention here that "Follow the workload" has no counterpart in a multi-region database?
</blockquote></details>

yes--we list that elsewhere but worth making it explicit here
___
*[v21.1/migrate-to-multiregion-sql.md, line 43 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7WiKN2hNJpjuQlBtA:-MiJgzEu7wAtnfc-g5vX:b1w51ct) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L43)):*

Do we need to say anything here about domiciling issues?  The data may move out of a geo temporarily as part of this migration.
</blockquote></details>

probably but i'd prefer to minimally mention it and link the DD doc
___
*[v21.1/migrate-to-multiregion-sql.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XLKa9nSdOjPyldjr:-MiJh3iQBgaM6cSTXIqW:b2ngthx) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L47)):*

We might also want to mention here that this work _must_ be done before you transition the database to have associated regions (and that it has to be done for all tables).
</blockquote></details>

+1 
___
*[v21.1/migrate-to-multiregion-sql.md, line 97 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XbE82iisIVsVX5sI:-MiJh9xX5iSPMtPaVT95:bt5ki01) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L97)):*

Do we worry that these instructions may be only part of the story?  In some cases will users need to modify their schemas to take into account the `crdb_region` column?  Or create the table using the `AS` clause?
</blockquote></details>

Its probably good to mention that if you have a column already you can use and link to that example
___
*[v21.1/migrate-to-multiregion-sql.md, line 174 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi34WLd0Hm8nOaHdtmd:-MiJhKo-20ZQArZCvttS:b-m288hw) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L174)):*

We might want to consider dropping this part entirely.  I worry that this may cause customers to do some digging into places that we don't want them (long term).  For customers who know that things work this way, they'll know how to get this info.
</blockquote></details>

yeah im conovinced that we should just drop based on this argument

rmloveland (commented):
Thanks!  And thanks for the great review Andy, Nathan, and Adam!

I think all feedback has been pretty well addressed *except* that I kinda disagree about removing the last section with the "here are the new zone configs" explanation (see rationale in comment below)

also I went with the `CONFIGURE ZONE DISCARD` option in that one discussion below since that seems way easier to explain

PTAL and let me know if you would like any additional edits!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057#-:-MjkE_XxAjc7QphdRzwy:b-9qwe8r)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @nvanbenschoten, and @otan)

___
*[_includes/v21.1/zone-configs/variables.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi76cD34qE22VZmQPlX:-MjkDgaE1SoIvmxAxg4z:bc4thh0) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/_includes/v21.1/zone-configs/variables.md#L9)):*

I wonder whether we should say a little bit more here about what this means for reads and writes. We say "transactions ... should be non-blocking", but that doesn't really say much about how this slows down writes but allows reads from any replica in the range. Maybe we link to the GLOBAL tables docs that spell out this trade-off in more detail?
</blockquote></details>

Updated as follows: 

- to add what you said about "slows down writes but allows reads from any replica"
- "non-blocking transactions" is now a link to more details in [the Transaction Layer docs](http://www.cockroachlabs.com/docs/v21.1/architecture/transaction-layer.html#non-blocking-transactions) (it didn't yet exist when we added this doc)
- the "GLOBAL tables" link now points to the main [GLOBAL tables doc](http://www.cockroachlabs.com/docs/v21.1/global-tables.html) as you suggested
___
*[v21.1/migrate-to-multiregion-sql.md, line 9 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi38fNEB_tDqqE7B1py:-Mjj07Lv38xTDjBaShLc:b-cm52it) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L9)):*

Nit: Good performance and low latency are sometimes the same thing.  Maybe "good performance and tuneable latency"?
</blockquote></details>

went with it! thanks!
___
*[v21.1/migrate-to-multiregion-sql.md, line 11 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7UV5E-Ojru8l0uT_s:-Mjj1Dmb1dHh5QW_HQyr:bd99vvy) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L11)):*

Yes...I'd keep what's written but add in the previous pattern names.
</blockquote></details>

Updated to say something like "using low-level mechanisms called replication zones in specific patterns we called Duplicate Indexes, ... etc"
___
*[v21.1/migrate-to-multiregion-sql.md, line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7V0Sq6Q36u0sMEZPm:-Mjj1Z0M8aLQE3FcGvXd:b-896fja) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L22)):*

+1 
</blockquote></details>

Done!
___
*[v21.1/migrate-to-multiregion-sql.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi72VRR1R4M8MgYerDL:-Mjj1xm54u1fniiXbRc8:bil5w3i) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L30)):*

+1
</blockquote></details>

Changed to say RBR and link to that section of the referenced page

Also updated the "Geo-partitioned replicas" section below to say RBR and link to that section
___
*[v21.1/migrate-to-multiregion-sql.md, line 31 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7WScA1nO7sdgz3YK2:-Mjj4TFc325tGlyuRhA4:b-mixa3m) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L31)):*

yes--we list that elsewhere but worth making it explicit here
</blockquote></details>

Updated using mostly similar language to what is on the FtW page - lemme know what you think!
___
*[v21.1/migrate-to-multiregion-sql.md, line 43 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7WiKN2hNJpjuQlBtA:-Mjj5UUP3lsJTWHZU8xw:b-gi27vu) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L43)):*

probably but i'd prefer to minimally mention it and link the DD doc
</blockquote></details>

Added a little info block at the bottom of this section that's like "hey FYI this data may move out of your specified geos temporarily, check out our domiciling page"
___
*[v21.1/migrate-to-multiregion-sql.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XLKa9nSdOjPyldjr:-Mjj83nw4SeF4RCB77ho:b-ijbpot) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L47)):*

+1 
</blockquote></details>

I'm not sure I fully understand, do you mean that basically the data movement from removing the old has to be complete before you can add the new?  at least that's how I'm parsing it.  I added a section to that effect - PTAL and see if that's close
___
*[v21.1/migrate-to-multiregion-sql.md, line 69 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7X0Sh6MgLy0YjITKO:-MjjANQx0NasgJR7wSZA:biebiji) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L69)):*

I was thinking along these lines as well. What we have here will work to set constraints and lease_preferences to empty values, but that's not quite the same as unsetting these fields. In other words, constraints from the parent database will not be inherited even after this is run. `CONFIGURE ZONE DISCARD` will clear these fields and allow inheritance to work properly, but it will throw away all other fields, which may not be desirable. I couldn't find a way to discard specific fields in a zone config. Maybe @otan knows how to do so.
</blockquote></details>

Based on this discussion I have updated this to use `CONFIGURE ZONE DISCARD` and added a disclaimer to the effect that "hey if you have other custom zone configs, you will have to reapply them" which is hopefully clear enough.

Sounds like there is nuance here but this seems like the clearest instruction to provide - shout if you object
___
*[v21.1/migrate-to-multiregion-sql.md, line 92 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi75paBAss1hctbeaiH:-MjjAMAd-bTSgTQ1-oyO:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L92)):*

s/index/indexes/

Same below.
</blockquote></details>

Updated!
___
*[v21.1/migrate-to-multiregion-sql.md, line 97 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XbE82iisIVsVX5sI:-MjkBo9W2MYihiJDYaQA:biji6uj) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L97)):*

Its probably good to mention that if you have a column already you can use and link to that example
</blockquote></details>

Updated this section with a note mentioning this and linking to the instructions on the SET LOCALITY page

Also updated this slightly based on a comment from Nathan above to mention REGIONAL BY ROW (not just "regional") and link to RBR info
___
*[v21.1/migrate-to-multiregion-sql.md, line 135 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi34K340LwlfVkg1Se2:-Mjj-qSxDxHC081tWqH3:bfy6gkp) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L135)):*

Or at least mention that ZONE is the default.
</blockquote></details>

Marked as optional in the header and revised the intro sentence and first bullet to make clear that ZONE is the default
___
*[v21.1/migrate-to-multiregion-sql.md, line 139 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7Zg55-3YkwVV9Gjae:-Mjj00fc-yXQJLgSMLg4:b-clzznb) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L139)):*

"More than one zone going down _concurrently_ ...?"
</blockquote></details>

Added the concurrently!
___
*[v21.1/migrate-to-multiregion-sql.md, line 140 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7ZlfXAwCzfLp5iqhg:-MjjM2RI3PLDMBiWQBRV:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L140)):*

"More than one region going down _concurrently_ ...?"
</blockquote></details>

Updated!
___
*[v21.1/migrate-to-multiregion-sql.md, line 174 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi34WLd0Hm8nOaHdtmd:-MjjM5csESBHWuw4qeyi:bz2a5wt) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L174)):*

yeah im conovinced that we should just drop based on this argument 
</blockquote></details>

Hmmm not sure I agree.   Previously, I have argued along the same lines - basically, "If you know enough to be touching this knife I shouldn't need to tell you not to cut yourself"

But apparently that wasn't true, right?  In some respects this whole page existing is a response to folks who were using zone configs feeling like there was not enough transitional material to allow them to move across to the new concepts.  It's not a page for new users, it's for existing zone config users who needed more educational material so they'd feel comfortable transitioning.

The explicit goal for this section is to say "look we have moved your cheese, but it's ok, it's still the same cheese.  Do not be afraid.  You can still look at the configs and see what is happening - but yay, now you don't have to."

This page's whole existence is for the short term so we can help folks transition for the longer term

And in the long term we can delete this horrible, horrible page :-) 

If you guys really feel strongly about it I'll remove it.  But I think our disclaimer at the top of this page about "if you are using MR SQL, you can ignore this" should already make clear this is not for new users.

awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @nvanbenschoten, @otan, and @rmloveland)

___
*[v21.1/migrate-to-multiregion-sql.md, line 174 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi34WLd0Hm8nOaHdtmd:-MjkUuFJ6oJHPTdEqVx2:bd20id9) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L174)):*

Hmmm not sure I agree.   Previously, I have argued along the same lines - basically, "If you know enough to be touching this knife I shouldn't need to tell you not to cut yourself"

But apparently that wasn't true, right?  In some respects this whole page existing is a response to folks who were using zone configs feeling like there was not enough transitional material to allow them to move across to the new concepts.  It's not a page for new users, it's for existing zone config users who needed more educational material so they'd feel comfortable transitioning.

The explicit goal for this section is to say "look we have moved your cheese, but it's ok, it's still the same cheese.  Do not be afraid.  You can still look at the configs and see what is happening - but yay, now you don't have to."

This page's whole existence is for the short term so we can help folks transition for the longer term

And in the long term we can delete this horrible, horrible page :-) 

If you guys really feel strongly about it I'll remove it.  But I think our disclaimer at the top of this page about "if you are using MR SQL, you can ignore this" should already make clear this is not for new users.
</blockquote></details>

I can follow that logic rich. I'll defer to your expertise here.

nvanbenschoten (approved):
:lgtm:

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057#-:-Mjkc5XH-FvfxW4_Bv18:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajstorm, @otan, and @rmloveland)

ajstorm (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057)* status: :shipit: complete! 1 of 0 LGTMs obtained

___
*[v21.1/migrate-to-multiregion-sql.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XLKa9nSdOjPyldjr:-MjlUFMJ2ao7eoCcCbgz:b17pdfy) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L47)):*

I'm not sure I fully understand, do you mean that basically the data movement from removing the old has to be complete before you can add the new?  at least that's how I'm parsing it.  I added a section to that effect - PTAL and see if that's close
</blockquote></details>

Sorry, I wasn't clear.  The data movement doesn't need to be complete, but the dropping of the zone configs must be complete.  I know that this has already merged, but probably worth another minor update to correct this part.
___
*[v21.1/migrate-to-multiregion-sql.md, line 174 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi34WLd0Hm8nOaHdtmd:-MjlUwduDEVkGrzPmch0:bmcqhro) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L174)):*

I can follow that logic rich. I'll defer to your expertise here.
</blockquote></details>

I buy your argument Rich.  LGTM.

rmloveland:
Hey @taroface would appreciate your review of this - we had to merge so it was available for a preso to some internal folks yesterday, but it would be improved by your input

taroface (commented):
Just some suggestions and a couple formatting corrections!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057#-:-Mk86Z9L0sb0MJRnTCwu:b-ohmcwz)* status: :shipit: complete! 1 of 0 LGTMs obtained

___
*[v21.1/migrate-to-multiregion-sql.md, line 13 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7pJ_m-lO0bwvRXi74:-Mk7pJ_m-lO0bwvRXi75:b9nwv52) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L13)):*
> ```
> In CockroachDB v21.1, we added
> ```

How do you feel about the more passive "CockroachDB v21.1 added"? Not sure if we refer to ourselves as "we" very often in the docs.
___
*[v21.1/migrate-to-multiregion-sql.md, line 15 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7ph5834HUnHZFuvQe:-Mk7ph5834HUnHZFuvQf:b-j6lcxw) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L15)):*
> ```
> in specific patterns we called
> ```

Ditto re: "we". I feel it adds a slight barrier in this case since it means the reader starts to wonder about the choices we made/are making.
___
*[v21.1/migrate-to-multiregion-sql.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7qgGl6edLih01HfWx:-Mk7qgGl6edLih01HfWy:bb4u1lo) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L17)):*
> ```
> worse performance
> ```

Is it possible to be more precise on what "worse" means here, or link to another doc? I'm wondering if someone already using replication zones and considering this migration would want more to go on.
___
*[v21.1/migrate-to-multiregion-sql.md, line 19 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7uw0_3XYTS8Ogfln1:-Mk7uw0_3XYTS8Ogfln2:b-eylymf) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L19)):*
> ```
> This page has instructions for transitioning from controlling multi-region clusters using replication-zone-based workflows to using multi-region SQL abstractions.
> ```

Maybe "This page describes how to migrate a multi-region cluster from using replication zones to using multi-region SQL abstractions." 
___
*[v21.1/migrate-to-multiregion-sql.md, line 21 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7vw8z8ZkQqfOPVAR-:-Mk7vw8z8ZkQqfOPVAR0:bhwdpuu) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L21)):*
> ```
> Mappings from
> ```

"Mappings between"?
___
*[v21.1/migrate-to-multiregion-sql.md, line 36 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7wXnqBmAMaVG80xX3:-Mk7wXnqBmAMaVG80xX4:bcrcf98) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L36)):*
> ```
> there is no analogous behavior provided by the multi-region SQL statements that matches Follow the Workload.
> ```

"the multi-region SQL statements do not provide a behavior that is analogous to follow-the-workload."
___
*[v21.1/migrate-to-multiregion-sql.md, line 36 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7wkU34xGLOpEN1OJo:-Mk7wkU34xGLOpEN1OJp:b-lf4it7) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L36)):*
> ```
> Follow the Workload
> ```

Looks like we style this `follow-the-workload`.
___
*[v21.1/migrate-to-multiregion-sql.md, line 51 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk80mwN5BXgxk0rk4fO:-Mk80mwO2-qbV7BjE07c:b6fzsww) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L51)):*
> ```
> next set of constraints are added
> ```

are -> is
___
*[v21.1/migrate-to-multiregion-sql.md, line 53 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk80xzJA02Xw8pqgJlo:-Mk80xzJA02Xw8pqgJlp:buyq5e6) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L53)):*
> ```
> Therefore, we recommend running this procedure at a time when you would normally perform scheduled maintenance.
> ```

Is it worth bumping this sentence to the top of the section?
___
*[v21.1/migrate-to-multiregion-sql.md, line 75 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk833gH0tUj_APSrNnj:-Mk833gH0tUj_APSrNnk:b-8g45qs) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L75)):*
> ```
> For example, 
> ```

I'd suggest cutting this
___
*[v21.1/migrate-to-multiregion-sql.md, line 100 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83QiY2LhhYVXM5dee:-Mk83QiY2LhhYVXM5def:b3yaosy) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L100)):*
> ```Markdown
> 1. Remove the manually created table partition. This will also automatically remove the replication zone configurations that were applied to the partition as part of the instructions.
> ```

Looks like an extra line break here
___
*[v21.1/migrate-to-multiregion-sql.md, line 102 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83jnD698plgBIIP31:-Mk83jnD698plgBIIP32:b-p2kmps) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L102)):*
> ```Markdown
> 
>     {% include_cached copy-clipboard.html %}
>     `ALTER TABLE users PARTITION BY NOTHING;`
> ```

This needs to be bracketed with `~~~ sql` and `~~~`. Also, remove code ticks.
___
*[v21.1/migrate-to-multiregion-sql.md, line 107 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83poz3m_YjTr_ungN:-Mk83poz3m_YjTr_ungO:b-89a4l7) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L107)):*
> ```Markdown
> 
>     {% include_cached copy-clipboard.html %}
>     `ALTER INDEX users_last_name_index PARTITION BY NOTHING;`
> ```

Ditto
___
*[v21.1/migrate-to-multiregion-sql.md, line 122 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83qol2RU4c92B04Y7:-Mk83qol2RU4c92B04Y8:b-89a4l7) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L122)):*
> ```Markdown
> 
>     {% include_cached copy-clipboard.html %}
>     `ALTER TABLE users PARTITION BY NOTHING;`
> ```

Ditto
___
*[v21.1/migrate-to-multiregion-sql.md, line 127 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83rPN1kAC9v6Xpoo9:-Mk83rPN1kAC9v6XpooA:b-89a4l7) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L127)):*
> ```Markdown
> 
>     {% include_cached copy-clipboard.html %}
>     `ALTER INDEX users_last_name_index PARTITION BY NOTHING;`
> ```

Ditto
___
*[v21.1/migrate-to-multiregion-sql.md, line 133 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84NA3CNWBiz89JVGV:-Mk84NA3CNWBiz89JVGW:b-58eptc) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L133)):*
> ```
> already
> ```

Nit: Maybe cut this word
___
*[v21.1/migrate-to-multiregion-sql.md, line 142 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84EHr4s6FLPwBz8yT:-Mk84EHr4s6FLPwBz8yU:b-oqtjw4) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L142)):*
> ```Markdown
> ~~~
> 
> ### Step 4. Add more regions to the database
> ```

This should be Step 3, unless there's a missing step before this.
___
*[v21.1/migrate-to-multiregion-sql.md, line 151 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84XukEGSqjYbrtcAV:-Mk84XukEGSqjYbrtcAW:b-xhb5r) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L151)):*
> ```Markdown
> ~~~
> 
> ### Step 5. (Optional) Configure your database survival goal
> ```

This should be Step 4.
___
*[v21.1/migrate-to-multiregion-sql.md, line 168 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84pLB4iDKOdD2cIen:-Mk84pLCBO9mQfasimuA:b-xhb4w) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L168)):*
> ```Markdown
> 
> ### Step 6. Configure table localities
> ```

This should be Step 5.
___
*[v21.1/migrate-to-multiregion-sql.md, line 188 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84wW27iWox6fjhuS5:-Mk84wW27iWox6fjhuS6:b-xhb41) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L188)):*
> ```Markdown
> For more information about when to use `GLOBAL` vs. `REGIONAL` tables, see [When to use `REGIONAL` vs `GLOBAL` tables](when-to-use-regional-vs-global-tables.html).
> 
> ### Step 7. (Optional) View the updated zone configurations
> ```

This should be Step 6.
___
*[v21.1/migrate-to-multiregion-sql.md, line 190 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk85Kr73NvIH_jWcn60:-Mk85Kr73NvIH_jWcn61:b-n4hq0j) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L190)):*
> ```
> The multi-region SQL statements operate by controlling the same replication zone configurations that you already have access to via
> ```

Maybe "The multi-region SQL statements operate on the same replication zone configurations that you can access via" (if this is still accurate)
___
*[v21.1/migrate-to-multiregion-sql.md, line 192 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk866iRC5pEte09VG0j:-Mk866iRC5pEte09VG0k:bwfx1ss) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L192)):*
> ```
> Low-latency reads and writes in a multi-region cluster
> ```

This should be title case to match with the earlier mention of this article. Also, no hyphen after "Low" according to the article title.

rmloveland (commented):
Thanks for the additional comments Ryan and Adam!  I have put the new stuff in a draft PR cockroachdb/docs#11767

Adam, I had more questions for you in our discussion below re: "how to know zone configs are done dropping" and will then update the above PR with that info

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057#-:-MkIdREo6B5dMsqxLmm0:b-q4ctiq)* status: :shipit: complete! 1 of 0 LGTMs obtained

___
*[v21.1/migrate-to-multiregion-sql.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XLKa9nSdOjPyldjr:-MkIJNyVE4qdL4tpgrRx:b7gliq4) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L47)):*

Sorry, I wasn't clear.  The data movement doesn't need to be complete, but the dropping of the zone configs must be complete.  I know that this has already merged, but probably worth another minor update to correct this part.
</blockquote></details>

Ah ok.  In that case: as a user, how do I know "dropping the old zone configs" is actually complete?  I think all I have is `SHOW ZONE CONFIGURATIONS` right?

In other words, in a process like the below, where I have e.g. 10 configs on a table foo:

1. SHOW ZONE CONFIGURATIONS FROM TABLE foo;
2. ALTER TABLE foo CONFIGURE ZONE DISCARD
3. SHOW ZONE CONFIGURATIONS FROM TABLE foo;
4. Repeat step 3 until done?

does the SHOW ZONE in step 3 continue to return a shorter and shorter list of zone configs as each config is dropped, such that I can "poll" by running that command every second and watch it go 10 .. 7 ... 3 ... etc down to 0?

or does SHOW ZONE just tell me "all good, everything is dropped" and then that process actually happens async?  in which case is there some other user-visible proxy for "zone configs are done being dropped"?
___
*[v21.1/migrate-to-multiregion-sql.md, line 13 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7pJ_m-lO0bwvRXi74:-MkILLmNDMuu2JCBWMn1:b-1xud1i) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L13)):*

> ```
> In CockroachDB v21.1, we added
> ```

How do you feel about the more passive "CockroachDB v21.1 added"? Not sure if we refer to ourselves as "we" very often in the docs.
</blockquote></details>

Works for me - updated
___
*[v21.1/migrate-to-multiregion-sql.md, line 15 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7ph5834HUnHZFuvQe:-MkIQvK6-ECBLc33w_VN:b-y8yjx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L15)):*

> ```
> in specific patterns we called
> ```

Ditto re: "we". I feel it adds a slight barrier in this case since it means the reader starts to wonder about the choices we made/are making.
</blockquote></details>

Updated to remove "we"
___
*[v21.1/migrate-to-multiregion-sql.md, line 17 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7qgGl6edLih01HfWx:-MkILPAr3VTBWpPD7Goc:btzxxru) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L17)):*

> ```
> worse performance
> ```

Is it possible to be more precise on what "worse" means here, or link to another doc? I'm wondering if someone already using replication zones and considering this migration would want more to go on.
</blockquote></details>

Unfortunately we don't have an obvious thing to link to re: the complexity of navigating the "state space" of rolling your own zone configs and tuning/debugging that vs. using something our engineers have designed for you.  Maybe we should add more on this?  Although a lot of discussion internally AIUI has been about how we may want to continue hiding zone configs from non-expert users and building higher-level features for those users instead

The huge complexity is also why I don't think we should write more here because it's way too much to explain in this space without making this doc horribly long and complex IMO.  I think the goal of this doc is to help a user with the transitional workflow, so I'm hoping the explanation of that workflow and how it ultimately maps to zone configs underneath will be enough to convince them - this may be wishful thinking tho :-) 
___
*[v21.1/migrate-to-multiregion-sql.md, line 19 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7uw0_3XYTS8Ogfln1:-MkIR55YAYlpSmT1m_lb:b-d09iku) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L19)):*

> ```
> This page has instructions for transitioning from controlling multi-region clusters using replication-zone-based workflows to using multi-region SQL abstractions.
> ```

Maybe "This page describes how to migrate a multi-region cluster from using replication zones to using multi-region SQL abstractions." 
</blockquote></details>

Nice! Much simpler 
___
*[v21.1/migrate-to-multiregion-sql.md, line 21 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7vw8z8ZkQqfOPVAR-:-MkIRCACBG7xmoz6MLBg:b-vft6o4) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L21)):*

> ```
> Mappings from
> ```

"Mappings between"?
</blockquote></details>

went with "mappings from .. to" - I think the between made total sense with the "and" that was there: "between .. and"
___
*[v21.1/migrate-to-multiregion-sql.md, line 36 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7wXnqBmAMaVG80xX3:-MkIRXs3D_WJvuOkfglV:bptm87z) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L36)):*

> ```
> there is no analogous behavior provided by the multi-region SQL statements that matches Follow the Workload.
> ```

"the multi-region SQL statements do not provide a behavior that is analogous to follow-the-workload."
</blockquote></details>

Nice! Simpler
___
*[v21.1/migrate-to-multiregion-sql.md, line 36 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk7wkU34xGLOpEN1OJo:-MkIRhLY9vrMEaD9yUpe:be6z4sw) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L36)):*

> ```
> Follow the Workload
> ```

Looks like we style this `follow-the-workload`.
</blockquote></details>

Ah! Updated
___
*[v21.1/migrate-to-multiregion-sql.md, line 51 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk80mwN5BXgxk0rk4fO:-MkIRwpL8KdEmbjhSE0s:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L51)):*

> ```
> next set of constraints are added
> ```

are -> is
</blockquote></details>

Updated!
___
*[v21.1/migrate-to-multiregion-sql.md, line 53 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk80xzJA02Xw8pqgJlo:-MkILF9J1uGezvCzAeEA:b-5422mz) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L53)):*

> ```
> Therefore, we recommend running this procedure at a time when you would normally perform scheduled maintenance.
> ```

Is it worth bumping this sentence to the top of the section?
</blockquote></details>

Great idea - moved it!
___
*[v21.1/migrate-to-multiregion-sql.md, line 75 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk833gH0tUj_APSrNnj:-MkISEMk7RiOzJ-8bUKV:b2nhc71) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L75)):*

> ```
> For example, 
> ```

I'd suggest cutting this
</blockquote></details>

CUT
___
*[v21.1/migrate-to-multiregion-sql.md, line 100 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83QiY2LhhYVXM5dee:-MkISHzJ7XPsA9uW3QyE:b-hrlaqi) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L100)):*

Looks like an extra line break here
</blockquote></details>

WHAT THE

removed
___
*[v21.1/migrate-to-multiregion-sql.md, line 102 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83jnD698plgBIIP31:-MkISXvC-hqfycbiF83u:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L102)):*

This needs to be bracketed with `~~~ sql` and `~~~`. Also, remove code ticks.
</blockquote></details>

Fixed!
___
*[v21.1/migrate-to-multiregion-sql.md, line 107 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83poz3m_YjTr_ungN:-MkISZ5e9J_poJOiV2zJ:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L107)):*

Ditto
</blockquote></details>

Fixed!
___
*[v21.1/migrate-to-multiregion-sql.md, line 122 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83qol2RU4c92B04Y7:-MkIWztm2Jvks4Q-mUr0:bt2vr0x) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L122)):*

Ditto
</blockquote></details>

Fixed 
___
*[v21.1/migrate-to-multiregion-sql.md, line 127 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk83rPN1kAC9v6Xpoo9:-MkIXJUL8O47UoQVKGop:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L127)):*

Ditto
</blockquote></details>

Fixed
___
*[v21.1/migrate-to-multiregion-sql.md, line 142 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84EHr4s6FLPwBz8yT:-MkIXSPO5t5aSqprCuLP:bped6wx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L142)):*

This should be Step 3, unless there's a missing step before this.
</blockquote></details>

Nice catch

ACTUALLY THIS WAS A TEST AND YOU PASSED RYAN

jk I just messed it up
___
*[v21.1/migrate-to-multiregion-sql.md, line 151 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84XukEGSqjYbrtcAV:-MkIX_0V8tDvDLlS5BRZ:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L151)):*

This should be Step 4.
</blockquote></details>

Fixed
___
*[v21.1/migrate-to-multiregion-sql.md, line 168 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84pLB4iDKOdD2cIen:-MkIXa4d7D5vqXsu-fTl:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L168)):*

This should be Step 5.
</blockquote></details>

Fixed
___
*[v21.1/migrate-to-multiregion-sql.md, line 188 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk84wW27iWox6fjhuS5:-MkIXcKC3_RYD1-c8PTF:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L188)):*

This should be Step 6.
</blockquote></details>

Fixed
___
*[v21.1/migrate-to-multiregion-sql.md, line 190 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk85Kr73NvIH_jWcn60:-MkIXpKY35qg8V91aOuk:bhe4ogw) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L190)):*

> ```
> The multi-region SQL statements operate by controlling the same replication zone configurations that you already have access to via
> ```

Maybe "The multi-region SQL statements operate on the same replication zone configurations that you can access via" (if this is still accurate)
</blockquote></details>

Updated to use mostly this - removed extra words
___
*[v21.1/migrate-to-multiregion-sql.md, line 192 at r2](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mk866iRC5pEte09VG0j:-MkIY-yM-RoQuO4YsSEZ:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/fb76b1ac396cbeb9979a3bd3fc4e0caf20c46b58/v21.1/migrate-to-multiregion-sql.md#L192)):*

> ```
> Low-latency reads and writes in a multi-region cluster
> ```

This should be title case to match with the earlier mention of this article. Also, no hyphen after "Low" according to the article title.
</blockquote></details>

Updated!

ajstorm (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11057)* status: :shipit: complete! 1 of 0 LGTMs obtained

___
*[v21.1/migrate-to-multiregion-sql.md, line 47 at r1](https://reviewable.io/reviews/cockroachdb/docs/11057#-Mi7XLKa9nSdOjPyldjr:-MkKORYrE-tn0hdHg6P0:b-o5s7jm) ([raw file](https://github.com/cockroachdb/docs/blob/75f4e266bb95544728244bd70d08421ff9f22520/v21.1/migrate-to-multiregion-sql.md#L47)):*

Ah ok.  In that case: as a user, how do I know "dropping the old zone configs" is actually complete?  I think all I have is `SHOW ZONE CONFIGURATIONS` right?

In other words, in a process like the below, where I have e.g. 10 configs on a table foo:

1. SHOW ZONE CONFIGURATIONS FROM TABLE foo;
2. ALTER TABLE foo CONFIGURE ZONE DISCARD
3. SHOW ZONE CONFIGURATIONS FROM TABLE foo;
4. Repeat step 3 until done?

does the SHOW ZONE in step 3 continue to return a shorter and shorter list of zone configs as each config is dropped, such that I can "poll" by running that command every second and watch it go 10 .. 7 ... 3 ... etc down to 0?

or does SHOW ZONE just tell me "all good, everything is dropped" and then that process actually happens async?  in which case is there some other user-visible proxy for "zone configs are done being dropped"?
</blockquote></details>

If you have 10 fields of a zone configuration set for table foo, `CONFIGURE ZONE DISCARD` will drop all of them with a single call.  You shouldn't need steps 3 or 4 above.  The user will know when it's complete once the `CONFIGURE ZONE DISCARD` command completes successfully.

PR #11066, opened 2021-08-17
Updates to Avro type section & remove experimental

### LINKED ISSUE #10881: Update Avro Limitations and types
The avro limitations section of the changefeed docs is out of date: https://www.cockroachlabs.com/docs/stable/create-changefeed.html#avro-limitations is out of date, especially since we're introducing new avro format support in 21.2 (everything is backported to 21.1 besides changes in decimals, so we should include these doc updates for those as well)

- Decimals must have precision specified. (still true - we may later update this) 
- Decimals with NaN or infinite values cannot be written in Avro (no longer applies - they are now written in string)
- TIME, DATE, INTERVAL, UUID, INET, ARRAY, JSONB, BIT, and collated STRING are not supported in Avro yet- we've added support for all of these so we should take this out ( For TIME, DATE, UUID, INET, JSONB - these are available at least in 20.2 so we should take them out of the list in earlier versions)
- Bit limitations - see https://github.com/cockroachdb/cockroach/pull/65869 

Updates to the Avro Types section:
- CockroachDB Type: Decimals, Avro Type: String, Bytes, Avro Logical Type: Decimal 
- CockroachDB Type: ENUMs (encoded as strings https://github.com/cockroachdb/cockroach/pull/65474)
- CockroachDB Type: Interval , Avro Type:  String
- CockroachDB Type: Array , Avro Type:  Array
- CockroachDB Type: Bit , Avro Type:  Array of LONGS
- CockroachDB Type: VarBit , Avro Type:  Array of LONGS
- CockroachDB Type: Collated String, Avro Type:  String

Add callout that Decimal type is a union between string and avro decimal type 

For Bit and Varbit encoding, we should call out some nuances and give an example of the format (example in this PR https://github.com/cockroachdb/cockroach/pull/65869): We encode BIT and VARBIT as arrays of 64-bit integers for efficiency. Since our encoding is big-endian, the last value may have many trailing zeroes. For this reason, the first value of each array gives the number of bits that are used in the last value. For example, if the bitfield is 65 bits long, this will be encoded into an array that looks like [1, <first 64 bits>, either 0 or the 64-bit integer with only the first bit set]. So you would first take every element in the array besides the first and base 2 encode it. Then you would look at the first number, which gives you the number of bits to take from the last base 2 encoded number. Then you would append all the base 2 encoded numbers together (with the caveat that the last item is truncated).

Example:
{sqlType: `VARBIT`, sql: `B'010'`, avro: `{"array":[3,4611686018427387904]}`}, // Take the 3 most significant bits of 1<<62

So first we base 2 encode '4611686018427387904', which encodes to: '0100000000000000000000000000000000000000000000000000000000000000' in 64 bits

We then take the first three bits: '010', which gives us our bit value

-------
Additionally, we are removing the experimental tag from avro, so we should change experimental-avro to just avro in the docs

kathancox:
Closes #10881 

PR includes the following updates: 

- Adjusted the Avro Limitations section in 21.2 to remove the decimal bullets that no longer apply and detail on `BIT` + `VARBIT` type encoding, and added an example on how the array encoding works.
- Adjusted same section in v21.1 & 20.2 per the details provided in the issue. 
- Edited the Avro type table for 21.2 & 21.1 (besides `DECIMAL`)... Note: Left the `DECIMAL` type in the table unchanged in 21.1.
-  Removed `experimental` from `experimental-avro` across CDC docs.


v20.2/create-changefeed.md
```diff
@@ -109,15 +109,15 @@ Option | Value | Description
...
+ Using the `format=avro`, `envelope=key_only`, and `updated` options together is rejected. `envelope=key_only` prevents any rows with updated fields from being emitted, which makes the `updated` option meaningless.
 {{site.data.alerts.end}}
 
 #### Avro limitations
 
-Currently, support for Avro is limited and experimental. Below is a list of unsupported SQL types and values for Avro changefeeds:
+You can only emit an Avro record if you are using a changefeed [connected to Kafka](#create-a-changefeed-connected-to-kafka-using-avro). Below is a list of unsupported SQL types and values for Avro changefeeds:
 
 - [Decimals](decimal.html) must have precision specified.
 - [Decimals](decimal.html) with `NaN` or infinite values cannot be written in Avro.
```
amruss, line 123: This no longer applies, we should take it out as well

v21.1/create-changefeed.md
```diff
@@ -129,7 +127,17 @@ Below is a list of unsupported SQL types and values for Avro changefeeds:
...
 
-- [`TIME`, `DATE`, `INTERVAL`](https://github.com/cockroachdb/cockroach/issues/32472), [`UUID`, `INET`](https://github.com/cockroachdb/cockroach/issues/34417), [`ARRAY`](https://github.com/cockroachdb/cockroach/issues/34420), [`JSONB`](https://github.com/cockroachdb/cockroach/issues/34421), `BIT`, and collated `STRING` are not supported in Avro yet.
+- [`BIT`](bit.html) and [`VARBIT`](bit.html) types are encoded as arrays of 64-bit integers.
+
+  For efficiency, CockroachDB encodes `BIT` and `VARBIT` bitfield types as arrays of 64-bit integers. Encoding in CockroachDB is [big-endian](https://en.wikipedia.org/wiki/Endianness), therefore the last value may have many trailing zeroes. For this reason, the first value of each array is the number of bits that are used in the last value of the array.
+
+  For instance, if the bitfield is 129 bits long, there will be 4 integers in the array. The first integer will be `1`; representing the number of bits in the last value, the second integer will be the first 64 bits, the third integer will be bits 65–128, and the last integer will either be `0` or `9223372036854775808` (i.e. the integer with only the first bit set).
```
amruss, line 134: It might make it more clear to show what `9223372036854775808` encodes to

amruss: and that we're translating base 2 encoding -> base 10 encoding, and the user needs to do the opposite downstream: take the base 10 encoded number and base 2 encode it

amruss (changes requested):
Couple of small things

kathancox:
Hey @amruss, True! I have added more detail into the `BIT` + `VARBIT` explanation. I added another example as part of showing the conversion on the user's end — I felt like two different sets of numbers/arrays would demonstrate how this works out a little better, but open to editing that!

v21.1/create-changefeed.md
```diff
@@ -129,7 +127,33 @@ Below is a list of unsupported SQL types and values for Avro changefeeds:
...
+
+    This example is base-10 encoded into an array as follows:
+
+    ~~~
+    {"array": [1, <first 64 bits>, <second 64 bits>, 0 or 9223372036854775808]}
+    ~~~
+
+    For downstream processing, it is necessary to base 2 encode every element in the array (except for the first element). The first number in the array gives you the number of bits to take from the last base 2 number — that is, the most significant bits. So, in the example above this would be `1`. Finally, all the base 2 numbers can be appended together, which will result in the original number of bits, 129.
+
+    To clarify this process, in another example in which you had a bitfield of 135 bits, the array would be as follow when base 10:
```
HonoreDB, line 144: If you're going to have a second example, I think it's a little clearer if the last used digit is a 0, to make it clear that you can't just eliminate all trailing zeroes from the last value. Like if you changed this from 135 bits to 136 bits, so that we're taking 8 bits from the last value and getting `10111110`.

HonoreDB (commented):
Thanks again, especially for the worked examples of BIT encoding.

v21.1/create-changefeed.md
kathancox: Yes, that's a really good point

_includes/v21.1/cdc/create-core-changefeed-avro.md
```diff
@@ -57,7 +57,7 @@ In this example, you'll set up a core changefeed for a single-node cluster that
 
     {% include copy-clipboard.html %}
     ~~~ sql
-    > EXPERIMENTAL CHANGEFEED FOR bar WITH format = experimental_avro, confluent_schema_registry = 'http://localhost:8081';
+    > EXPERIMENTAL CHANGEFEED FOR bar WITH format = avro, confluent_schema_registry = 'http://localhost:8081';
```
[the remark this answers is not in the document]
kathancox: Ack, OK. @amruss so I just want to remove `experimental` from avro for v21.2?

_includes/v20.2/cdc/create-core-changefeed-avro.md
```diff
@@ -57,7 +57,7 @@ In this example, you'll set up a core changefeed for a single-node cluster that
 
     {% include copy-clipboard.html %}
     ~~~ sql
-    > EXPERIMENTAL CHANGEFEED FOR bar WITH format = experimental_avro, confluent_schema_registry = 'http://localhost:8081';
+    > EXPERIMENTAL CHANGEFEED FOR bar WITH format = avro, confluent_schema_registry = 'http://localhost:8081';
```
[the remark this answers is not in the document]
amruss: I think you're right, I think just keeping it non-experimental in 21.2 is fine

_includes/v21.1/cdc/create-core-changefeed-avro.md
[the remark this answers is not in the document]
amruss: I think lets just keep it in 21.2

amruss (approved):
LGTM, assuming the core piece is also updated

kathancox:
> LGTM, assuming the core piece is also updated

Thanks @amruss I have incorporated the updated avro limitation section into the core changefeed docs. Would welcome a check of that, but I'll assign the docs reviewer too.

kathancox:
cc @mikeCRL

HonoreDB (approved):
> Reviewed 2 of 9 files at r1, 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11066)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @HonoreDB, @kathancox, and @rmloveland)

v21.1/changefeed-for.md
```diff
@@ -57,16 +57,38 @@ Option | Value | Description
...
+
+    For instance, if the bitfield is 129 bits long, there will be 4 integers in the array. The first integer will be `1`; representing the number of bits in the last value, the second integer will be the first 64 bits, the third integer will be bits 65–128, and the last integer will either be `0` or `9223372036854775808` (i.e. the integer with only the first bit set, or `1000000000000000000000000000000000000000000000000000000000000000` when base 2).
+
+    This example is base-10 encoded into an array as follows:
+
+    ~~~
+    {"array": [1, <first 64 bits>, <second 64 bits>, 0 or 9223372036854775808]}
+    ~~~
+
+    For downstream processing, it is necessary to base 2 encode every element in the array (except for the first element). The first number in the array gives you the number of bits to take from the last base-2 number — that is, the most significant bits. So, in the example above this would be `1`. Finally, all the base-2 numbers can be appended together, which will result in the original number of bits, 129.
```
rmloveland, line 77: we should say "base-2" to match our base-* language elsewhere

v21.1/changefeed-for.md
```diff
@@ -57,16 +57,38 @@ Option | Value | Description
...
+    {"array": [8, 18293058736425533439, 18446744073709551615, 13690942867206307840]}
+    ~~~
+
+    To then work with this data, you would convert each of the elements in the array to base-2 numbers, besides the first element. For the above array, this would convert to:
+
+    ~~~
+    [8, 1111110111011011111111111111111111111111111111111111111111111111, 1111111111111111111111111111111111111111111111111111111111111111, 1011111000000000000000000000000000000000000000000000000000000000]
+    ~~~
+
+    Next, you use the first number to take the number of bits from the last base-2 element, `10111110`. Finally, you append each of the base-2 numbers (the last number from the array being truncated), which in this case results in 136 bits, the original number of bits.
```
rmloveland, line 91: > Finally, you append each of the base-2 numbers (the last number from the array being truncated)

I think I get it, but the parenthetical makes the ordering feel reversed.  After reading this a few times I understand this to mean that I need to snarf the # of bits in array element 1 from the 4th element of the array to give some number _n_, and then throw away (drop) that 4th element.  Then, append array elements 2 and 3 to _n_, making one big 136-bit binary number?

may be worth revising slightly so the order is more explicit. then again I think I was able to figure it out (?) from what you wrote, so this is probably an optional comment.  And also this is not my area, so the fact I was a bit confused is not a showstopper

v21.2/changefeed-for.md
```diff
@@ -52,21 +52,41 @@ Option | Value | Description
...
 
-- [`time`, `date`, `interval`](https://github.com/cockroachdb/cockroach/issues/32472), [`uuid`, `inet`](https://github.com/cockroachdb/cockroach/issues/34417), [`array`](https://github.com/cockroachdb/cockroach/issues/34420), and [`jsonb`](https://github.com/cockroachdb/cockroach/issues/34421) are not supported in Avro yet.
+    For instance, if the bitfield is 129 bits long, there will be 4 integers in the array. The first integer will be `1`; representing the number of bits in the last value, the second integer will be the first 64 bits, the third integer will be bits 65–128, and the last integer will either be `0` or `9223372036854775808` (i.e. the integer with only the first bit set, or `1000000000000000000000000000000000000000000000000000000000000000` when base 2).
```
rmloveland, line 67: This section is repeated a bit, could maybe be an include?

v21.2/create-changefeed.md
```diff
@@ -100,36 +100,54 @@ Option | Value | Description
...
 
-Currently, support for Avro is limited and experimental. You can only emit an Avro record if you are using a changefeed [connected to Kafka](#create-a-changefeed-connected-to-kafka-using-avro).
-
-Below is a list of unsupported SQL types and values for Avro changefeeds:
+Below are clarifications for particular SQL types and values for Avro changefeeds:
 
 - [Decimals](decimal.html) must have precision specified.
-- [Decimals](decimal.html) with `NaN` or infinite values cannot be written in Avro.
+- [`BIT`](bit.html) and [`VARBIT`](bit.html) types are encoded as arrays of 64-bit integers.
+
```
rmloveland, line 125: Definitely feeling an include++

v21.2/create-changefeed.md
```diff
@@ -145,10 +163,20 @@ CockroachDB Type | Avro Type | Avro Logical Type
...
 [`TIMESTAMPTZ`](timestamp.html) | [`LONG`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) | [`TIME-MICROS`](https://avro.apache.org/docs/1.8.1/spec.html#Time+%28microsecond+precision%29)
-[`DECIMAL`](decimal.html) | [`BYTES`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) | [`DECIMAL`](https://avro.apache.org/docs/1.8.1/spec.html#Decimal)
+[`DECIMAL`](decimal.html) | [`STRING`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive), [`BYTES`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) | [`DECIMAL`](https://avro.apache.org/docs/1.8.1/spec.html#Decimal)
 [`UUID`](uuid.html) | [`STRING`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
 [`INET`](inet.html) | [`STRING`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
 [`JSONB`](jsonb.html) | [`STRING`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
+[`ENUMS`](enum.html) | [`STRING`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
```
rmloveland, line 170: Is this list of Avro supported types the same as on the other page?  If so, recommend an include

rmloveland (approved):
LGTM with a few comments which are mostly take-or-leavable

v21.2/create-changefeed.md
kathancox: At least two of the versions have slightly different availability

PR #11093, opened 2021-08-20
Update cost-based-optimizer.md

mgoddard:
The addition of "hash" here is helpful since in practice you only see merge  joins when you insert a single row, but you would see hash joins with many inserted rows.

### DIFF
--- a/v21.1/cost-based-optimizer.md
+++ b/v21.1/cost-based-optimizer.md
@@ -173,7 +173,7 @@ If it is not possible to use the algorithm specified in the hint, an error is si
 
 
 {{site.data.alerts.callout_info}}
-To make the optimizer prefer lookup joins to merge or hash joins when performing foreign key checks, set the `prefer_lookup_joins_for_fks` [session variable](set-vars.html) to `on`.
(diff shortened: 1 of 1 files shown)


ianjevans (approved):

ianjevans:
Thanks for the PR! I approved, but it'd be best if you could make the same change to the v21.2 and v20.2 topic as well.

dbist:
@RaduBerinde how many more rows need to be added before hash join turns to a lookup join? @mgoddard I'd add more color here that merge join is for a single row whereas hash join is with many rows.

RaduBerinde:
___
*[v21.1/cost-based-optimizer.md, line 176 at r2](https://reviewable.io/reviews/cockroachdb/docs/11093#-Mho0ESH6dnF13xDUY3a:-Mho0ESH6dnF13xDUY3b:b4n0g24) ([raw file](https://github.com/cockroachdb/docs/blob/930eebeb091772895af9d3cd7a7f85a5142db106/v21.1/cost-based-optimizer.md#L176)):*
> ```Markdown
> 
> To make the optimizer prefer lookup joins over merge or hash joins when performing foreign key checks, set the `prefer_lookup_joins_for_fks` [session variable](set-vars.html) to `on` (merge joins are the default for single row inserts, and hash joins are the default for multi row inserts).
> ```

I'd remove the second part, I'm not sure it's always accurate.

PR #11094, opened 2021-08-20
Added topic_naming parameter to CDC docs

### LINKED ISSUE #10935: Update Kafka topic naming
21.2 we added the ability to do arbitrary topic naming for kafka topics

We should document the query parameter:

Parameter: topic_name
Type: String
Sink: Kafka
Default value is table name
The allowed character set is [a-zA-Z0-9\\._\\-]. Kafka allows both . and _ but recommends against using both due to the possibility of bugs. We do the same. Special characters in the table name are replaced by their unicode code points when generating the Kafka topic name
You can combine with topic prefix
Note: All records will go into the topic you specify, but Schemas will still be registered separately.

For more info, Aaron wrote up a great internal doc: https://cockroachlabs.atlassian.net/wiki/spaces/CDC/pages/1636958818/Sink+Names+A+u0022+Simple+u0022+Guide

kathancox:
Closes #10935 

Added the `topic_name` query parameter to the table with the detail, example, and default. Included this option in the Topic Naming section. 

**Note**: Linked from the table description for the parameter to the already specified topic naming limitations/support.


v21.2/create-changefeed.md
```diff
@@ -152,7 +153,7 @@ CockroachDB Type | Avro Type | Avro Logical Type
 
 #### Topic Naming
 
-By default, a Kafka topic has the same name as the table that a changefeed was created on. If a changefeed was created on multiple tables, the changefeed will write to multiple topics corresponding to those table names. You can specify a [topic prefix](#query-parameters) or use the [`full_table_name` option](#options) to modify this.
+By default, a Kafka topic has the same name as the table that a changefeed was created on. If a changefeed was created on multiple tables, the changefeed will write to multiple topics corresponding to those table names. You can specify a [topic prefix](#topic-prefix-param), [an arbitrary topic name](#topic-name-param), or use the [`full_table_name` option](#full-table-option) to modify this.
```
amruss, line 156: feel like we should also mention that you can specify an arbitrary topic name and feed all the tables into that if you want.

amruss (approved):
small nit but LGTM

rmloveland (approved):
LGTM!

PR #11117, opened 2021-08-24
Updated GCR tutorial to use CLI commands instead of the web UI

ericharmeling:
Follow-up on internal email thread.


ianjevans (approved):

awoods187:
I got here late but this is great

PR #11150, opened 2021-08-27
Correct foreign key restriction for computed columns

electrum:
Stored computed columns can be a foreign key reference. Virtual columns cannot. Neither can reference a foreign key.


ianjevans (approved):
Good catch. Thanks for the PR. If you could also apply the same change to `v21.2/computed-columns.md` that would be a big help.

jseldess:
I'll just merge this an open a follow-up to port the changes to v21.2.

PR #11153, opened 2021-08-27
Add example of canceling multiple queries.

### LINKED ISSUE #10640: Add `CANCEL QUERIES` example to Manage Long-Running Queries
We document how to cancel a particular query by both query ID and selecting the ID using a subquery, but don't show how to cancel multiple queries with a single statement. We need an example of using `CANCEL QUERIES` in both the SQL reference and in Manage Long-Running Queries.

From Slack.
> is there a fast way to run CANCEL QUERY <x> on a lot of entries from `SELECT * FROM [SHOW CLUSTER QUERIES] ... `
> there is, CANCEL QUERIES. that should really be in https://www.cockroachlabs.com/docs/v20.1/manage-long-running-queries.html#cancel-long-running-queries

```
CANCEL QUERIES (SELECT query_id FROM [SHOW CLUSTER QUERIES]
      WHERE application_name = 'test_app');
```

ianjevans:
Fixes #10640.


ianjevans:
TFTR, Kathryn.

PR #11231, opened 2021-09-07
Fix CC referral ID on GCR tutorials

jseldess:
The include used in step 1.1 expects a page-level variable for the referral ID. This variable was missing.


ericharmeling (approved):
LGTM - thanks Jesse!

PR #11234, opened 2021-09-07
Update cluster scaling callouts

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-4888
https://cockroachlabs.atlassian.net/browse/CC-4831

Update that users can scale clusters up and down but cannot go from multi-node to single-node.
Add a callout that contract customers need to contact support to scale.

### DIFF
--- a/cockroachcloud/cluster-management.md
+++ b/cockroachcloud/cluster-management.md
@@ -76,7 +76,7 @@ From the **Overview** page, you can connect to your cluster. For more informatio
 You can add or remove nodes from your cluster through the Console. See [Planning your cluster](#planning-your-cluster) for cluster requirements and recommendations before proceeding.
 
 {{site.data.alerts.callout_info}}
-At this time, you cannot use the Console to scale up a single-node cluster or scale down to a single-node cluster. For these changes, contact [Support](https://support.cockroachlabs.com).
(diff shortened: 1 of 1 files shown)


emily-horing (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11234#-:-Mj5Lc-g2UsCgtPr4hEO:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @daniel-crlabs)

daniel-crlabs (commented):
Looks good, one question, can they scale from a single node to multi-node? Maybe we should add an explanation for that as well?

gemma-shay:
@daniel-crlabs yes, non-invoice customers can scale up/down in the console as long as they're not going from multi-node to single-node. we have instructions for adding/removing nodes and some recommendations at the top the [cluster management page](cockroachcloud/cluster-management.md). did you have something else in mind?

daniel-crlabs:
Ok, sounds good, as long as our docs and the UI reflect this in a way to grab the customer's attention, that's all good.
Thanks again for this.

ianjevans (approved):

PR #11238, opened 2021-09-08
Changing the event type to maybe enable changed files from forks.

ianjevans:
This is a test to see if changing the event type to `pull_request_target` will trigger the changed files workflows from forked repositories.

See the note under "Action inputs" here: https://github.com/peter-evans/create-or-update-comment


ericharmeling (approved):
LGTM AFAICT

PR #11627, opened 2021-09-10
Fix step number

KoditkarVedant:
In `Quickstart with CockroachCloud Free` tab `Download the CA certificate` and `Use the built-in SQL client` both the steps has same step number. This pull request changes the step number for `Use the built-in SQL client` to 5.


jseldess:
The GH action is failing because the PR is on a fork. Merging regardless.

jseldess:
cc @ianjevans, @gemma-shay

PR #11629, opened 2021-09-10
Adding a condition to not run list-changed-file workflow for PRs from forks.

ianjevans:
Although we'd like to list the changed files in PRs from forked repos, GitHub can't use security tokens across forks when posting the comment with the changed file output. Until we can find a solution to this, we should just turn off this workflow in PRs from forked repos.

### DIFF
--- a/.github/workflows/changed_files.yml
+++ b/.github/workflows/changed_files.yml
@@ -5,6 +5,7 @@ on:
 
 jobs:
   list-files-changed:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


ericharmeling (commented):
I don't have enough knowledge or context around these workflows to offer any proper feedback.

Deferring to you/Jesse.

jseldess (approved):
Thanks, @ianjevans. This change LGTM, although I have no way to verify it. If it fails after merge, we can revisit.

ianjevans:
TFTR, Jesse. Yeah, a big pain with these integrations comes from the inability to test before merging.

PR #11633, opened 2021-09-10
Documented `kafka_sink_config` option

### LINKED ISSUE #11001: Document kafka config option
In 21.2, we added a new WITH option for changefeeds: kafka_sink_config. We should include it amongst our WITH options and also add an "Advanced operator section" to explain the nuances of the different options, this can be a section where we add other advanced configurations.
(also backported to 21.1 so we should update those docs as well) 

Syntax:  kafka_sink_config='{"Flush": {"MaxMessages": 1000, "Frequency": "1s"}, "version": "<version number>", "RequiredAcks": "ALL" }`'

Required Acks: values "NONE", "ONE", or "ALL"
https://github.com/cockroachdb/cockroach/pull/67504 
This value specifies what we consider a successful write to kafka to be. Remember, we guarantee at least once delivery of messages. This value defines what "delivery" means. 
ONE: This is our default value. Meaning we consider a write to Kafka successful once the leader node has committed and
acknowledged the write. This carries a risk of dropped messages if the leader node acks
before replicating to a quorum of other Kafka nodes, then fails.
NONE: No kafka brokers are required to acknowledge that they have committed the message. This will decrease latency/increase throughput, but will come at the cost of lower consistency. 
ALL: This is the highest consistency level. A quorum must  be reached (most kafka brokers have committed the message) before the leader can acknowledge. 

Version: In 21.1, we upgraded to the default sarama library we used from version 0.8.2.0 to 1.0.0. You can still specify the old version using this option. 
https://github.com/cockroachdb/cockroach/pull/65389

MaxMessages and Frequency are batching paramaters that allow you to tradeoff between latency and throughput.
MaxMessages: The maximum number of messages the producer will send in a single broker request. This defaults to 1. Increasing this value allows you to send all values in a batch.
Frequency: Specifies the maximum time we will batch messages. 
In the example above: Max messages is set to 1000 and flush frequency is set to 1 second. Meaning it will flush to kafka EITHER after 1 second has passed or we batch 1000 messages, whichever comes first.  If we aren’t getting a ton of messages we will will add 1 second latency. But when there is a huge influx of messages we’re getting through them much more quickly

These are advanced operator settings in that messing with them could have a negative impact on the changefeed, so you really need to understand what you're doing when setting them.
Some warnings: 
- You could configure batching parameters to be very, very high - which would negatively impact changefeed latency. It would take a long time to see the messages come through to the sink.
- Kafka server configurations: The kafka sink could be configured to reject batches that are too large

kathancox:
Closes #11001 

Added detail on the `kafka_sink_config` option. Incorporated into v21.1 & v21.2 option table. Added a section on the nuances of this option to a new Advanced Configuration section (linked from the table). This section will also house details on the webhook config sink option too (in another PR). 

Removed the limitation on the earlier Kafka versions from the docs as a result of this option — I haven't specifically called out the `version` configuration anywhere except in the Advanced Config section, since from what I've gathered this is likely to not be used often. (Let me know if you feel this should be called out further.)

Provided a warning on the negative impacts to changefeeds that these configurations could potentially have.

### DIFF
--- a/v21.2/create-changefeed.md
+++ b/v21.2/create-changefeed.md
@@ -111,6 +111,7 @@ Option | Value | Description
 `initial_scan` / `no_initial_scan` | N/A |  Control whether or not an initial scan will occur at the start time of a changefeed. `initial_scan` and `no_initial_scan` cannot be used simultaneously. If neither `initial_scan` nor `no_initial_scan` is specified, an initial scan will occur if there is no `cursor`, and will not occur if there is one. This preserves the behavior from previous releases.<br><br>Default: `initial_scan` <br>If used in conjunction with `cursor`, an initial scan will be performed at the cursor timestamp. If no `cursor` is specified, the initial scan is performed at `now()`.
 `full_table_name` | N/A | Use fully-qualified table name in topics, subjects, schemas, and record output instead of the default table name. This can prevent unintended behavior when the same table name is present in multiple databases. <br><br>Example: `CREATE CHANGEFEED FOR foo... WITH full_table_name` will create the topic name `defaultdb.public.foo` instead of `foo`.
 `avro_schema_prefix` | Schema prefix name               | Use fully-qualified schema name for a table instead of the default table name. This allows multiple databases or clusters to share the same schema registry when the same table name is present in multiple databases.<br><br>Example: `CREATE CHANGEFEED FOR foo WITH format=experimental_avro, confluent_schema_registry='registry_url', avro_schema_prefix='super'` will register subjects as `superfoo-key` and `superfoo-value` with the namespace `super`.
+`kafka_sink_config` | [`STRING`](string.html) | <span class="version-tag">New in v21.2:</span> Set fields to configure delivery, version, and batching parameters for Kafka sinks. See [Advanced Configuration](#advanced-configuration) for more detail on configuring this option. <br><br>Default: `CREATE CHANGEFEED FOR table INTO 'kafka://localhost:9092' WITH kafka_sink_config='{"Flush": {"MaxMessages": 1, "Frequency": "1s"}, "version": "", "RequiredAcks": "ONE" }'`
 
 {{site.data.alerts.callout_info}}
  Using the `format=experimental_avro`, `envelope=key_only`, and `updated` options together is rejected. `envelope=key_only` prevents any rows with updated fields from being emitted, which makes the `updated` option meaningless.
@@ -166,6 +167,19 @@ Kafka has the following topic limitations:
 - The maximum character length of a topic name is 249.
 - Topics with a period (`.`) and underscore (`_`) can collide on internal Kafka data structures, so you should use either but not both.
 - Characters not accepted by Kafka will be automatically encoded as unicode characters by CockroachDB.
+
+## Advanced Configuration
+
+### Kafka Sink Configuration
+
+<span class="version-tag">New in v21.2:</span>The `kafka_sink_config` option allows configuration of a changefeed's message delivery, Kafka server version, and batching parameters. The configurable fields include:
+
+* `Flush.MaxMessages` and `Flush.Frequency`: Batching parameters to set up depending on latency and throughput needs. For example, if `MaxMessages` is set to 1000 and `Frequency` to 1 second, it will flush to Kafka either after 1 second or after 1000 messages are batched, whichever comes first. It's important to consider that if there are not many messages then a `"1s"` frequency will add 1 second latency. However, if there is a larger influx of messages these will be flushed quicker.
+  * `MaxMessages`: Maximum number of messages the producer will send in a single broker request. Increasing this value allows you to send all values in a batch. Default: 1. Type: `INT`.
+  * `Frequency`: Maximum time that messages will be batched. Default: `"0s"`. Type: `INTERVAL`.
+
+  
+
 
 ## Responses
 


stevendanna (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11633)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @HonoreDB, and @kathancox)

*[a discussion](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTN-0oCkLgKX7owMab:-MjTN-0oCkLgKX7owMac:bffhapk) (no related file):*
The bits related to the version field look good.  I've left some inline comments and questions for your consideration.

---
___
*[v21.1/create-changefeed.md, line 114 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTLR2J4MOjQMUT9d62:-MjTLR2J4MOjQMUT9d63:b-vf5pfb) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L114)):*
> ```Markdown
> `full_table_name` | N/A | <span class="version-tag"> New in v21.1: </span> Use fully-qualified table name in topics, subjects, schemas, and record output instead of the default table name. This can prevent unintended behavior when the same table name is present in multiple databases. <br><br>Example: `CREATE CHANGEFEED FOR foo... WITH full_table_name` will create the topic name `defaultdb.public.foo` instead of `foo`.
> `avro_schema_prefix` | Schema prefix name               | <span class="version-tag"> New in v21.1: </span> Use fully-qualified schema name for a table instead of the default table name. This allows multiple databases or clusters to share the same schema registry when the same table name is present in multiple databases.<br><br>Example: `CREATE CHANGEFEED FOR foo WITH format=experimental_avro, confluent_schema_registry='registry_url', avro_schema_prefix='super'` will register subjects as `superfoo-key` and `superfoo-value` with the namespace `super`.
> `kafka_sink_config` | [`STRING`](string.html) | Set fields to configure delivery, version, and batching parameters for Kafka sinks. See [Advanced Configuration](#advanced-configuration) for more detail on configuring this option. <br><br>Example: `CREATE CHANGEFEED FOR table INTO 'kafka://localhost:9092' WITH kafka_sink_config='{"Flush": {"MaxMessages": 1, "Frequency": "1s"}, "version": "", "RequiredAcks": "ONE" }'`
> ```

`"version": ""` works, but is equivalent to not including it at all.  Given that this should be a rarely used option, might it be clearer to drop it from this example?
___
*[v21.1/create-changefeed.md, line 114 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTMnaA5jyYl-kUAHSS:-MjTMnaA5jyYl-kUAHST:bvgy80q) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L114)):*
> ```
> Set fields to configure delivery, version, and batching parameters
> ```

Perhaps "server version" to avoid possible confusion?
___
*[v21.1/create-changefeed.md, line 191 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTLKZa4tMFuEUjxuB_:-MjTLKZa4tMFuEUjxuBa:b1wjy8i) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L191)):*
> ```Markdown
>   * `"Frequency"`: sets the maximum time that messages will be batched. Default: `"0s"`. Type: `INTERVAL`.
> 
> * `"version"`: sets the appropriate Kafka cluster version, which can be used to connect to [Kafka versions < v1.0](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) (`kafka_sink_config='{"version": "0.8.2.0"}'`). Default: `""` Type: `STRING`.
> ```

If we want consistent capitalisation, `"Version"` will also work.  Let me know, as if you think it is important, I'll add a test so that we don't regress on supporting the capitalised version of "version."

The `Default` is interesting here. `""` is indeed the default; but, when we provide empty string, we are deferring to the underlying library's default version, which I believe is 1.0.0.0. Listing 1.0.0.0 as the default might be more informative to some users.
___
*[v21.1/create-changefeed.md, line 193 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTMMbx-dhziHbNAsjI:-MjTMMbx-dhziHbNAsjJ:buqgorl) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L193)):*
> ```Markdown
> * `"version"`: sets the appropriate Kafka cluster version, which can be used to connect to [Kafka versions < v1.0](https://docs.confluent.io/platform/current/installation/versions-interoperability.html) (`kafka_sink_config='{"version": "0.8.2.0"}'`). Default: `""` Type: `STRING`.
> 
> * `"RequiredAcks"`: specifies what a successful write to Kafka is. CockroachDB [guarantees at least once delivery of messages](stream-data-out-of-cockroachdb-using-changefeeds.html#ordering-guarantees) — this value defines the _delivery_. Type: `STRING`. The possible values are:
> ```

Rather than "this value defines the _delivery_", something like "this value defines the required level of message acknowledgement from the Kafka server" might be clearer. I defer to you and @honoredb here.

HonoreDB (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11633)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @stevendanna)

___
*[v21.1/create-changefeed.md, line 191 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTLKZa4tMFuEUjxuB_:-MjVGaL_8Grxvx8dKUv_:biwh6bh) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L191)):*
<details><summary><i>Previously, stevendanna (Steven Danna) wrote…</i></summary><blockquote>

If we want consistent capitalisation, `"Version"` will also work.  Let me know, as if you think it is important, I'll add a test so that we don't regress on supporting the capitalised version of "version."

The `Default` is interesting here. `""` is indeed the default; but, when we provide empty string, we are deferring to the underlying library's default version, which I believe is 1.0.0.0. Listing 1.0.0.0 as the default might be more informative to some users.
</blockquote></details>

+1 to both.
___
*[v21.1/create-changefeed.md, line 193 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTMMbx-dhziHbNAsjI:-MjVFZ3m3XGtdJ5gQvXT:b-420kse) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L193)):*
<details><summary><i>Previously, stevendanna (Steven Danna) wrote…</i></summary><blockquote>

Rather than "this value defines the _delivery_", something like "this value defines the required level of message acknowledgement from the Kafka server" might be clearer. I defer to you and @honoredb here.
</blockquote></details>

I like the idea that we're defining the "delivery" in the delivery guarantee and I think the explanation of possible values gives enough detail to make it clear. Your (Steven's) version would be better in places where we have less real estate for a full explanation though.
___
*[v21.2/create-changefeed.md, line 178 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjVGwWr6CPyo2OAQe2R:-MjVGwWr6CPyo2OAQe2S:b-2dgwek) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.2/create-changefeed.md#L178)):*
> ```Markdown
> 
> {{site.data.alerts.callout_danger}}
> It's important to note that each of the following settings have significant impact on a changefeed's behavior, such as latency. For example, it is possible to configure batching parameters to be very high, which would negatively impact changefeed latency. As a result it would take a long time to see messages coming through to the sink. Also, the Kafka sink could be configured to reject batches that are too large.
> ```

I'd suggest "Kafka server" rather than "Kafka sink" as the configuration option you're talking about here isn't possible to set in `kafka_sink_config`. Or maybe a rephrase to something like "Also, large batches may be rejected by the Kafka server unless it's separately configured to accept a high `max.message.bytes`.'

kathancox:
> "version": "" works, but is equivalent to not including it at all. Given that this should be a rarely used option, might it be clearer to drop it from this example?

Yes, good point. From the Options table, I have removed the version field, but included it in the examples in the advanced config section. I also changed the language in the table to mention "all available" options... to try and make it clear that  users need to check out that section. What do you think @amruss 

Also, on the version field I have capped that in all instances, OK @stevendanna ?

kathancox:
> The Default is interesting here. "" is indeed the default; but, when we provide empty string, we are deferring to the underlying library's default version, which I believe is 1.0.0.0. Listing 1.0.0.0 as the default might be more informative to some users. 

Agree, I have changed the default version as Steven mentioned. I updated the example in the Advanced section (within the code block) with 0.8.2.0 to show this incorporated ... but I'm in two minds whether that is a good idea? It is _already_ listed separately within the version bullet description. Lmk your take on that one @amruss

kathancox:
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11633)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @stevendanna)
> 
> _[v21.1/create-changefeed.md, line 191 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTLKZa4tMFuEUjxuB_:-MjVGaL_8Grxvx8dKUv_:biwh6bh) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L191)):_
> 
> Previously, stevendanna (Steven Danna) wrote…
> +1 to both.
> 
> _[v21.1/create-changefeed.md, line 193 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjTMMbx-dhziHbNAsjI:-MjVFZ3m3XGtdJ5gQvXT:b-420kse) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.1/create-changefeed.md#L193)):_
> 
> Previously, stevendanna (Steven Danna) wrote…
> I like the idea that we're defining the "delivery" in the delivery guarantee and I think the explanation of possible values gives enough detail to make it clear. Your (Steven's) version would be better in places where we have less real estate for a full explanation though.
> 
> _[v21.2/create-changefeed.md, line 178 at r2](https://reviewable.io/reviews/cockroachdb/docs/11633#-MjVGwWr6CPyo2OAQe2R:-MjVGwWr6CPyo2OAQe2S:b-2dgwek) ([raw file](https://github.com/cockroachdb/docs/blob/676ba37d48ab58576fa4f1bb8e61e3c869f4c1b4/v21.2/create-changefeed.md#L178)):_
> 
> > ```gfm
> > {{site.data.alerts.callout_danger}}
> > It's important to note that each of the following settings have significant impact on a changefeed's behavior, such as latency. For example, it is possible to configure batching parameters to be very high, which would negatively impact changefeed latency. As a result it would take a long time to see messages coming through to the sink. Also, the Kafka sink could be configured to reject batches that are too large.
> > ```
> 
> I'd suggest "Kafka server" rather than "Kafka sink" as the configuration option you're talking about here isn't possible to set in `kafka_sink_config`. Or maybe a rephrase to something like "Also, large batches may be rejected by the Kafka server unless it's separately configured to accept a high `max.message.bytes`.'

On the description for `Delivery` I adapted the description within the table with @stevendanna where there is less space, and then I retained what was in the advanced section. OK? 

Yes, thanks Aaron for the better warning update, I have added that, with a link out to the  `max.messages.bytes` config on the Kafka side.

PR #11640, opened 2021-09-10
Removing notice about index use for OR.

### LINKED ISSUE #9803: Remove notice around index use for OR
This notice should be removed as of 20.2:
```
Given a query like SELECT * FROM foo WHERE a > 1 OR b > 2, even if there are appropriate indexes to satisfy both a > 1 and b > 2, the query planner performs a full table or index scan because it cannot use both conditions at once.
```

ianjevans:
Fixes #9803.


mgartner (approved):
LGTM

PR #11671, opened 2021-09-14
ADHOC: Add link to user file for bulk operation

sfong-roblox:
Migrate from CSV links to Cloud and Local File Server but not to User File


sfong-roblox:
Wait that netfliy preview is pretty cool!

jseldess:
Thank you, @sfong-roblox! 

@kathancox, can you PTAL? If you approve this change, please follow up with the same for v21.2 docs.

kathancox (approved):

PR #11759, opened 2021-09-22
New navigation proposal based on user personas and journeys

ianjevans:
This PR reorganizes the sidebar navigation into user journeys based on these user personas:

- New users, with topics under "New to CockroachDB"
- Developers, with topics under "Develop with CockroachDB"
- Operators, with topics under "Deploy CockroachDB"

There is also a Reference top-level section that will apply to all the above personas, and make it easy to find release notes. I've significantly reorganized the Reference topics.

The sidebar data JSON has been split into these files:

- `sidebar-data-personas.json`: the main container with "New to CockroachDB" topics and the includes of the other persona JSON files.
- `sidebar-data-developers.json`: for "Develop with CockroachDB"
- `sidebar-data-operators.json`: for "Deploy CockroachDB"
- `sidebar-data-reference.json`: for the Reference section

Still TBD:

- [x] Add the Architecture section and topics
- [ ] Add new topics in added in v21.2 since I branched for this project back in June.

The new navigation is only in the dev/v21.2 section of the docs.


stbof:
Is there a reason why you are using the gerund forms Developing with CockroachDB and Deploying CockroachDB for the container titles instead of the imperative Develop with CockroachDB and Deploy CockroachDB?  The latter would be consistent with the verb form in the child topics.

ianjevans:
TBH, I'm still shaking off the house style from DataStax a year on....
Changing to imperative would give more white space in the nav area.

PR #11722, opened 2021-09-20
Release notes v21.2.0 beta.1 and v21.2.0-beta.2

kathancox:
Closes #10715, #10716

Release notes for v21.2.0-beta.1 and beta.2

For review: Note that this includes **already edited** notes from the alpha that were never released. Markdown comments show the items that do not need review

PR #11761, opened 2021-09-23
Hide 21.2 docs

jseldess:
Our 21.2 docs are live before the first testing version has been released. The download instructions in the 21.2 docs therefore fail. 

When the 21.2 beta release is blessed and https://github.com/cockroachdb/docs/pull/11722/files is merged, this won't be an issue. In the meantime, if it takes much longer to get that release out, this PR is a temporary way to hide the 21.2 docs without breaking any existing incoming links.

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -21,6 +21,7 @@ exclude:
 - archived
 - package.json
 - package-lock.json
+- v21.2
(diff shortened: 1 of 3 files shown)


jseldess:
No longer needed.

PR #11763, opened 2021-09-23
Ensure we reference a tagged release of the operator

pseudomuto:
We're going to be making some changes on the master branch of the operator including some of the manifest files referenced here in the docs. Before we do, I wanted to make sure that it wouldn't break anything that we're recommending people do to use the operator.

I've added `operator_version` to the main site config and referenced that in all of the _start-cockroachdb-operator-secure.md_ files. This changes the links from `https://.../master/...` to `https://.../<version>/...`. Now when we change something on master, we don't need to worry about it breaking our documentation.

The one downside here is that we'll need to change this value after we cut a new release. I think this is worth the tradeoff and can easily be added to our deployment steps.

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -91,3 +91,4 @@ release_info:
     version: v21.2.0-alpha.1
 site_title: CockroachDB Docs
 url: https://www.cockroachlabs.com
+operator_version: v2.1.0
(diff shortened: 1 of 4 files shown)


taroface:
LGTM! Thanks for doing this. I believe this should also fix https://github.com/cockroachdb/cockroach-operator/issues/450

taroface (approved):

davidwding (approved):

Issue #10689, opened 2021-06-02
sql: modify stored procedures status from planned to X


awoods187:
In https://www.cockroachlabs.com/docs/v21.1/sql-feature-support.html we list stored procedure support as "planned." We should downgrade this to X as we don't plan to add support anytime soon.

PR #11781, opened 2021-09-24
Propagate lack of stored proc support to 21.2.

stbof:
Fixes https://github.com/cockroachdb/docs/issues/10689 in 21.2.

### DIFF
--- a/_includes/v21.2/misc/interleave-deprecation-note.md
+++ b/_includes/v21.2/misc/interleave-deprecation-note.md
@@ -1,1 +1,0 @@
-{{site.data.alerts.callout_danger}}Interleaving was deprecated in CockroachDB v20.2, and is disabled by default in CockroachDB v21.1 and later. For details, see [`INTERLEAVE IN PARENT` Deprecation](interleave-in-parent.html#deprecation).{{site.data.alerts.end}}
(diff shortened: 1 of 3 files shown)


stbof:
The last (Misc) table on the web site is broken. 
Screen Shot 2021-09-24 at 1 40 58 PM

I played around with the included warning, but the only way to fix the table was to explictly specify the warning instead of including it. Not sure why.

ianjevans (approved):
Approved. But I think the reason the include didn't work is because there's a trailing line break in `interleave-deprecation-notice.md`. Can you try removing the line break and see if the table still breaks?

stbof:
> Approved. But I think the reason the include didn't work is because there's a trailing line break in `interleave-deprecation-notice.md`. Can you try removing the line break and see if the table still breaks?

There is no trailing line break that I can see either in Atom, vi, or GH.

stbof:
@ianjevans I could not get the include file to work with the latest version of the referring file. I realized that the include exists because multiple files were including it. I'm worried that the include file still referenced in alter-primary-key and the explicit warning in sql-feature-support will get out of sync. Should I still merge this?

ianjevans:
> @ianjevans I could not get the include file to work with the latest version of the referring file. I realized that the include exists because multiple files were including it. I'm worried that the include file still referenced in alter-primary-key and the explicit warning in sql-feature-support will get out of sync. Should I still merge this?

I tried this locally, and it worked.

```
Interleaved tables | ✓ | CockroachDB Extension | [Interleaved Tables documentation](interleave-in-parent.html)<br>{% include {{ page.version.version }}/misc/interleave-deprecation-note.md -%}
```

See the [Liquid docs on whitespace control](https://shopify.github.io/liquid/basics/whitespace/).

But I also see that my local build doesn't break the table even without the `-%}`. So there's a discrepancy between local Jekyll builds and the Netlify builds, perhaps. Regardless, try pushing a commit with the `-%}` at the end of the include and see if it breaks the table.

Issue #10963, opened 2021-07-22
feature request: view all release notes for vXY.Z.* on one big page


rmloveland:
Richard Loveland (rmloveland) commented:

We have had folks ask internally.  Would likely be useful for external users as well.

Current workaround is to clone the docs repo and do the following inelegant things:

    $ cd ~/path/to/docs/releases
    $ cat v20.2.*.md > v20.2.all.md # or some version
    $ markdown v20.2.all.md > v20.2.all.html

Of course this does not really help most users.

See also:

- cockroachdb/docs#5368

Jira Issue: DOC-1496

rmloveland:
cc @mikeCRL since I'm not sure how exactly to label/categorize this

rafiss:
We've had external customers ask for this as well. The biggest pain point is that the release notes from alpha/beta versions are on a separate page, and some customers often just skip right over the alpha release notes since they wouldn't expect to find anything important there. But sometimes we do document breaking changes on alpha release notes but nowhere else.

shermanCRL:
Is there any automation that generates the current patch release pages?

rmloveland:
thanks for the additional context @rafiss !

> Is there any automation that generates the current patch release pages?

yes @shermanCRL , it's a script in the CockroachDB repo: https://github.com/cockroachdb/cockroach/blob/master/scripts/release-notes.py 

It takes an arbitrary start and end commit SHA (or tag, I think?), so I think it could be used to generate something like "show me all changes from v20.2.0 through v20.2.13" or similar on one page

vy-ton:
@jseldess do you see this being prioritized for 21.2?

PR #11783, opened 2021-09-24
Single page for all 21.1 releases

jseldess:
This is an initial experiment to understand the UX pros/cons
of putting all release notes from a major release on
a single page.

One obvious challenge: Our anchor links don't work as a user
would expect when there are multiple identical headings on
a page. The anchor link always links to the first instance.
We would need to fix this.

We'd also need to make many changes to the way release notes
are referenced (e.g., from the release note list, install pages,
sidenav, etc.) and put in place many redirects. Time-consuming
but doable and worth-while if there's consensus that this format,
or something related, is better for users.

Addresses #10963


stbof:
All these instances `{% include marketo.html %}` need to have a unique ID or they don't display.  I fixed a bug where the ID was duplicated here: https://github.com/cockroachdb/docs/pull/11775

jseldess:
Good catch, @stbof. Thank you. I'll fix that if people agree that we want to proceed with this general approach.

releases/v21.1.md
```diff
@@ -0,0 +1,3293 @@
...
+~~~
+
+### CockroachCloud
+
+- <a href="https://cockroachlabs.cloud/signup?referralId=docs_crdb_release_notes" rel="noopener" target="_blank">Get a free v21.1 cluster on CockroachCloud</a>.
+- Learn about recent updates to CockroachCloud in the [CockroachCloud Release Notes](index-cockroachcloud.html).
+
+### Feature summary
+
+This section summarizes the most significant user-facing changes in v21.1.0. For a complete list of features and changes, including bug fixes and performance improvements, see the [release notes](index.html#testing-releases) for previous testing releases. You can also search for [what's new in v21.1 in our docs](../search.html?query=%22New%20in%20v21.1%22&page=2).
```
bdarnell, line 1066: This section (the hand-summarized version of the 21.1.0 notes) shouldn't be mixed in to the middle of this giant page - I'd probably keep it on its own page. Maybe three pages per branch:
- "What's new in 21.1" - the hand-summarized overall notes. (The "what's new" name is inspired by [python](https://docs.python.org/3/whatsnew/3.9.html))
- "21.1 pre-releases" - all the alpha and beta release notes
- "21.1 patch releases"  - all the 21.1.x release notes starting with 21.1.1. Alternately, call it "21.1 releases" and start it with the mostly-auto-generated form of the 21.1.0 release notes (or a merger of the alpha/beta notes). That's probably not worth the trouble, though, since there's still a lot of editing that goes into the output of the release notes script.

jseldess:
@bdarnell, PTAL. As you suggested, I kept the stand-alone 21.1.0 page and split the rest into `v21.1 Pre-Releases` and `v21.1 Patch Releases`. With this approach, I think we'd need to rethink the Releases sidenav somewhat and make sure we're cross-referencing between these helpfully. Let me know if you have thoughts on that.

rmloveland:
random thoughts: i wonder if this is something we could do for releases from some chosen version forward (say v21.1 and on) rather than backporting, so we don’t need to do a lot of time-consuming admin stuff around redirects, etc for e.g. versions <= v20.2

Thesis: value of docs for older releases decays rapidly (even more so in cloud/serverless world)

also, i think splitting pre-releases/patches may negate some of the value of the original issue, cf this comment from @rafiss (https://github.com/cockroachdb/docs/issues/10963#issuecomment-885397742)

> The biggest pain point is that the release notes from alpha/beta versions are on a separate page, and some customers often just skip right over the alpha release notes since they wouldn’t expect to find anything important there. But sometimes we do document breaking changes on alpha release notes but nowhere else.

another thought that comes to mind is: what if we break something in a pre-release and fix it in a later patch release? then you still have to look across two different pages

finally: could we perhaps avoid redirects/admin stuff by adding the big page as another option rather than replacing what we do / have now?  e.g. we do the various smaller release notes as we do now, except that we have a big “all-in-one” page that we also update by just pasting the latest small release notes onto the bottom of as we go?

bdarnell:
> i wonder if this is something we could do for releases from some chosen version forward (say v21.1 and on) rather than backporting, so we don’t need to do a lot of time-consuming admin stuff around redirects

I definitely agree with this. 

> The biggest pain point is that the release notes from alpha/beta versions are on a separate page, and some customers often just skip right over the alpha release notes since they wouldn’t expect to find anything important there

These customers are right - the only people who should look at alpha/beta release notes are ones who are running one of those versions. Anything important from alpha/beta release notes *must* make it into the release notes for the applicable final release. Maybe this argues for a "detailed changelog" component to the x.y.0 release notes (a merged and organized version of the alpha/beta notes) in addition to the higher-level summary we have now.

> could we perhaps avoid redirects/admin stuff by adding the big page as another option rather than replacing what we do / have now?

My guess is that providing both options is going to end up being more work in the long run and I'd rather just do whatever ends up being simplest (leave the old ones as is and make whatever changes we want going forward)

rmloveland:
tl;dr: I think "everything on one page" is better than 3, for the following reasons: it is what the users seem to be asking for (easy to find things in one place), and is as simple as possible to maintain - for a given release `x.y.z`, we copy the content onto the top of the `x.y.*` page above the previous content for `x.y.z-1`

Longer comments below:

> Anything important from alpha/beta release notes must make it into the release notes for the applicable final release. Maybe this argues for a "detailed changelog" component to the x.y.0 release notes (a merged and organized version of the alpha/beta notes) in addition to the higher-level summary we have now.

In other words, this detailed changelog added to the `x.y.0` release page would include all of the items from the `x.y.0-{alpha,beta}*` notes concatenated - Is that what you mean?

This would leave us with the following 3 types of pages per major version, if I understand this PR and the above convo correctly:

1. `x.y.0-{alpha,beta}` pre-releases page (which I guess folks can ignore once x.y.0 is out?)
1. `x.y.0` "official dot oh" release page which has the high-level stuff and a detailed changelog which duplicates the content of  all the stuff from the pre-releases.
1. `x.y.1` and up on the point releases page, which has the detailed changelogs for all the point releases in one place

Thus yielding three pages in total for a given release version (IIUC)

That seems like a reasonable compromise, but still has the following limitations IMO:

- It still isn't actually "all the things for `x.y.*` in one place", so users would still have to look in two places rather than one: on the `x.y.0` page and the point-releases page
- There is also a risk of confusion since there would now be duplicated content between the pre-releases page and the "detailed changelog" of the `x.y.0` page: now users can see the same info in two places (and we have to maintain it across two places)

Based on the convo above it sounds like we are leaning away from "all the releases for `x.y.*` on one page" but I think that is the actual user desire being expressed in the linked issue.  I also think it is much simpler to maintain: for a given release `x.y.z`, just copy the content onto the top of the `x.y.*` page above the previous release notes `x.y.z-1`.

> My guess is that providing both options is going to end up being more work in the long run and I'd rather just do whatever ends up being simplest (leave the old ones as is and make whatever changes we want going forward)

I think in practice it's much simpler to make "one giant `x.y.*` page that we add stuff to" than to do 3 pages per release, two of which can have some duplicate content (again, unless I am misunderstanding)

bdarnell:
I don't feel strongly about whether it's one, two, or three pages per release. If it's going to be one big page it's important that it be organized differently than the earlier draft (which was just chronological so that the most human-readable content was stuck in the middle). I think one giant page is going to be unwieldy or intimidating for many users, but I defer to y'all and would personally prefer to err on the side of too much on a page instead of too many pages. 

One other consideration for one giant page is that we'll have a page for `/docs/releases/22.1.html` well before 22.1 is actually released. We'll need to make sure that the page clearly indicates its pre-release status since the URL won't. And there are implications for SEO, but I'm not sure which direction (is it better to let the page exist and accumulate a little bit of pagerank before the release, or will it siphon traffic away from other places and leave stale content in google caches?).

rafiss:
> AFAICT it's mostly internal folks supporting on-prem users who have
asked for this, so we should consider whether it's very important to
spend even more time supporting those use cases / needs going forward,
esp. if going forward we are focused more on cloud usage.

I wouldn't say that it's mostly on-prem users -- I think any dedicated cluster whether self-hosted or CC-hosted would care about this.

Also, I hypothesize that the only reason we hear about it from dedicated users right now is because that's how most of our large customers operate CockroachDB right now. I don't anticipate that large customers would stop caring about this problem ever.

Why do we think Cloud users won't care about release notes? It will still be very important for them to be able to identify backwards-incompatible changes in one place.

rmloveland:
Rafi Shamim ***@***.***> writes:

>     AFAICT it's mostly internal folks supporting on-prem users who
>     have
>     asked for this, so we should consider whether it's very important
>     to
>     spend even more time supporting those use cases / needs going
>     forward,
>     esp. if going forward we are focused more on cloud usage.
>
> I wouldn't say that it's mostly on-prem users -- I think any dedicated
> cluster whether self-hosted or CC-hosted would care about this.

OK! I trust your experience supporting customers more than my own
opinion/misunderstanding on this

> Also, I hypothesize that the only reason we hear about it from
> dedicated users right now is because that's how most of our large
> customers operate CockroachDB right now. I don't anticipate that large
> customers would stop caring about this problem ever.

Makes sense

> Why do we think Cloud users won't care about release notes? It will
> still be very important for them to be able to identify
> backwards-incompatible changes in one place.

I guess that my thinking based on reading release notes for a while is
that many of the items that go into the release notes could be
considered "minutiae" to most users of a cloud service that someone else
is running.

However it sounds like that is not really true based on your experience
(and others)

So what do you think of the 3 types of pages approach we've been
discussing?

1. Pre-releases
2. Main x.y.0 release
3. Point releases x.y.1 and up

Is this an approach that will meet the need from your POV?

rafiss:
I agree with your point about many of the notes being minutiae, and I also agree that Cloud users might care less about that.

But I think the breaking or backwards-incompatible changes are not minutiae. So maybe instead of trying to change all the release notes formats we should narrow the scope of this a bit: there needs to be an easier place to see breaking changes. The biggest gripe I have heard from support situations is when we document breaking changes only on alpha/beta release notes and not on the vX.Y.0 release notes. And we'd also want to address the problem of a customer going to vX.Y.6 who didn't realize they had to check the vX.Y.0 release notes for any breaking changes.

jseldess:
Hmm, we do aggregate all previously announced backward-incompatible changes on the vX.Y.0 release page, for example: https://www.cockroachlabs.com/docs/releases/v21.1.0#backward-incompatible-changes

But your point about vX.Y.N customers having to go back there to find them is a good point.

gemma-shay:
Working on a PR to put all the cloud notes on one page and running into a similar issue with anchors: https://github.com/cockroachdb/docs/pull/12334

PR #11802, opened 2021-09-27
update Systemd guidance for file descriptors limit

### LINKED ISSUE #11748: Add note about setting no files limit to SystemD service def to infinity
In our Prod Recommendations doc, we talk about setting file descriptor limits: https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#file-descriptors-limit

Edit the service definition to configure the maximum number of open files -- like this:
```
[Service]
...
LimitNOFILE=35000
```

We also say that we recommend setting this to unlimited:
```
It is recommended to set the file descriptors limit to unlimited; otherwise, the recommended limit is at least 15000 (10000 per store plus 5000 for networking). This higher limit ensures performance and accommodates cluster growth.
```

However, you will get an error in Linux if you do this:
```
LimitNOFILE=unlimited
```

Instead, you need to do:
```
LimitNOFILE=infinity
```

I think it would be helpful to add a note to the SystemD section that says:
NOTE: To set the file descriptor limit to "unlimited" in the SystemD service definition file, you need to use the value "infinity"

taroface:
Fixes #11748.

### DIFF
--- a/v20.2/recommended-production-settings.md
+++ b/v20.2/recommended-production-settings.md
@@ -306,8 +306,8 @@ CockroachDB can use a large number of open file descriptors, often more than is 
 
 For each CockroachDB node:
 
-- At a **minimum**, the file descriptors limit must be 1956 (1700 per store plus 256 for networking). If the limit is below this threshold, the node will not start.
-- It is **recommended** to set the file descriptors limit to unlimited; otherwise, the recommended limit is at least 15000 (10000 per store plus 5000 for networking). This higher limit ensures performance and accommodates cluster growth.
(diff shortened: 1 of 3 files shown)


jhatcher9999 (commented):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11802#-:-Mkd607iAei506wRv_wx:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained

kathancox (approved):
LGTM @taroface !

PR #11805, opened 2021-09-27
Update contradictory recommendation re: localities

### LINKED ISSUE #11786: Low Latency Reads and Writes in a Multi-Region Cluster Doc Update
Re: [Low Latency Reads and Writes in a Multi-Region Cluster](https://cockroachlabs.com/docs/v21.1/demo-low-latency-multi-region-deployment.html)

## Issue Description

Contradicting table locality patterns are recommended at different places in the same web page.

At one place it says (my emphasis):

> All of the tables except promo_codes have a composite primary key of city and id, in that order. This means that the rows in these tables are ordered by their geography. These tables are read from and written to very frequently. To keep read and write latency low, **you'll use the Regional table locality pattern for these tables.**

Later it says:

> All of the tables except promo_codes are geographically specific, and updated very frequently. For these tables, **the right table locality for optimizing access to their data is REGIONAL BY ROW.**

## Suggested Resolution 

Change the first reference to match the second one, which is correct.  So change the first one to:

> All of the tables except promo_codes have a composite primary key of city and id, in that order. This means that the rows in these tables are ordered by their geography. These tables are read from and written to very frequently. To keep read and write latency low, **you'll use the REGIONAL BY ROW table locality pattern for these tables.**

rmloveland:
Fixes #11786

Summary of changes:

- Instead of referring to "regional tables" generally, we update to say
  `REGIONAL BY ROW` (which is what we meant, and what the doc already
  links to).

### DIFF
--- a/v21.1/demo-low-latency-multi-region-deployment.md
+++ b/v21.1/demo-low-latency-multi-region-deployment.md
@@ -45,7 +45,7 @@ The workload you'll run against the cluster is our open-source, fictional, peer-
 
 {% include {{ page.version.version }}/misc/movr-schema.md %}
 
-All of the tables except `promo_codes` have a composite primary key of `city` and `id`, in that order. This means that the rows in these tables are ordered by their geography. These tables are read from and written to very frequently. To keep read and write latency low, you'll use the [Regional table locality pattern](multiregion-overview.html#regional-by-row-tables) for these tables.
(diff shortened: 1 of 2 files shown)


taroface (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11805#-:-MkcZZpa5Nc_hozUI5i4:b-uwkowd)* status: :shipit: complete! 1 of 0 LGTMs obtained

PR #11828, opened 2021-09-29
Revert HTTP(s) known limitation

rafiss:
After some discussion, we decided that this limitation was too broad to
be helpful for developers who may be looking for different things.


otan (approved):

ericharmeling (approved):
LGTM! Thanks for doing this @rafiss.


I went ahead and made a docs issue, in case we want to call out the no full REST API support elsewhere: https://github.com/cockroachdb/docs/issues/11829

PR #11838, opened 2021-09-29
Made `IMPORT INTO` takes table offline more prominent

### LINKED ISSUE #11123: import: make it more prominent that `import into` takes the table offline
Michael Wang (mwang1026) commented:

https://www.cockroachlabs.com/docs/v21.1/import-into.html

We make note of this, but customers constantly run into this as a big surprise

We should also update \h help text in the SQL shell

kathancox:
Closes #11123 

Adjusted the positioning of the note letting users know that `IMPORT INTO` will take the table offline while the data is imported into the table. 

- Added this to the considerations section 20.2, 21.1, 21.2
- Added a note on this right above the examples
- Adjusted some of the surrounding text into body text rather note boxes to emphasize this notice relative to importance of the other information


mwang1026:
cc @livlobo for review

livlobo (approved):
LGTM w.r.t. making it prominent.

Do we use the terminology The table is public again to mean that it is back online?

_includes/v21.1/import-into-offline.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_info}}
```
rmloveland, line 1: tbh I feel like this creeps closer to a warning to me than just "info".  meta: wish we had the middle ground "yellow caution note" for these situations

v21.1/import-into.md
```diff
@@ -10,6 +10,7 @@ The `IMPORT INTO` [statement](sql-statements.html) imports CSV, Avro, or delimit
 
 - `IMPORT INTO` works for existing tables. To import data into new tables, read the following [Import into a new table from a CSV file](#import-into-a-new-table-from-a-csv-file) example.
 - `IMPORT INTO` cannot be used within a [transaction](transactions.html) or during a [rolling upgrade](upgrade-cockroach-version.html).
+- `IMPORT INTO` takes the table **offline** before importing the data. The table will be online again once the job has completed successfully.
```
rmloveland, line 13: It's good that we're adding this note, but I wonder If 5 times on this page feels a bit on the nose stylistically? for me as a reader it's a bit much.  your call obv but my opinion is that if it's at the top under "Considerations" and at the bottom under "Known Limitations" we have definitely let the reader know.  if anything there is quite a bit of overlap between those two sections on this page and I'm wondering if we can consolidate.  At the end of the day it's more words to read that are saying the same things

also if it is technically a "known limitation" is there a GH issue we can link the user to so they know when/if this is lifted?  if not, and this is Just How It Works Now (tm) it's a shame we are deprecating `IMPORT TABLE` in favor of this which apparently does not have this limitation?  or am I wrong about that?  (Sorry to annoy but I'm reading this with my "skeptical user" hat on and feeling like I'm losing something)

v21.1/import-into.md
```diff
@@ -164,6 +163,8 @@ Pausing and then resuming an `IMPORT INTO` job will cause it to restart from the
 
 The following provide connection examples to cloud storage providers. For more information on connecting to different storage options, read [Use Cloud Storage for Bulk Operations](use-cloud-storage-for-bulk-operations.html).
 
+{% include {{ page.version.version }}/import-into-offline.md %}
```
rmloveland, line 166: from style POV this is probably the one I like the least tbh but if we do have to keep it is there a way to make every instance on this page use the include?  tons o repetition

rmloveland (approved):
LGTM but had comments - I would strongly prefer consolidation and fewer repetitions but your call obv.  Nothing blocking merge 

Separate issue and not something that can be addressed in this very topical PR but as a user I would like some clarity around how this interacts with the deprecation of `IMPORT TABLE` which, if it doesn't have these limitations (which it apparently does not according the `IMPORT` docs), makes me feel as a user like I'm losing something but not getting anything, and it isn't clear why - there is no explanation on the `IMPORT` page other than "this is happening".  If the response is "actually that's not accurate" or "please file a separate docs issue" that is fine, please let me know :-)

sorry, I am one of those annoying "why are you moving my cheese" users

PR #11872, opened 2021-10-04
Update logging-overview.md

knz:


taroface:
@knz LGTM! Could you add this line to v21.2 as well?

knz:
> Could you add this line to v21.2 as well?

Done

PR #11877, opened 2021-10-04
Fix typo in rename-index.md files

postamar:
IF EXISTS mistakenly referred to columns instead of indexes.


jseldess (approved):
LGTM. Thanks, @postamar!

PR #11885, opened 2021-10-05
Fix sslmode typo

CNeenan:


CNeenan:
Had a cert issue and saw here it should be 'sslmode' not 'sslMode' which also fixed my issue:

https://stackoverflow.com/questions/51291765/root-crt-not-found-postgresql

ianjevans (approved):

ianjevans:
Thanks for the PR!

PR #11896, opened 2021-10-06
Add docs on bounded staleness (follower) reads

### LINKED ISSUE #11016: Document Bounded Staleness Reads
Bounded staleness reads are a form of read-only queries that use a dynamic, system-determined timestamp, subject to a user-provided staleness bound, to read nearby replicas while minimizing data staleness.  They provide a new way to perform [follower reads](https://www.cockroachlabs.com/docs/v21.1/follower-reads.html) off local replicas to minimize query latency in [multiregion clusters](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview).

This issue tracks all the docs we need to write for this feature.

See also:

- [Bounded staleness reads RFC](https://github.com/cockroachdb/cockroach/blob/master/docs/RFCS/20210519_bounded_staleness_reads.md)
- cockroachdb/docs#6793
- cockroachdb/docs#9870

rmloveland:
Fixes #11016

Summary of changes:

- Update the Follower Reads page to mention that follower reads can be
  either exact staleness or bounded staleness, along with
  descriptions/examples of each

- Update AS OF SYSTEM TIME docs based on the above

- Update 'Follower Reads' deployment pattern page to provide a bounded
  staleness read example, and link to the updated FR docs

- Update links from licensing page


nvanbenschoten (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11896)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rmloveland)

___
*[v21.2/follower-reads.md, line 21 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLvDZ940_ih2HG9bpk:-MlLvDZ940_ih2HG9bpl:b-he96e4) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L21)):*
> ```Markdown
> ## How follower reads work
> 
> Each CockroachDB range tracks a property called its [_closed timestamp_](architecture/transaction-layer.html#closed-timestamps), which means that no new writes can ever be introduced below that timestamp. The closed timestamp advances forward by some target interval behind the current time. If the range receives a write at a timestamp less than its closed timestamp, the write is forced to change its timestamp, which might result in a [transaction retry error](transaction-retry-error-reference.html).
> ```

`s/below/at or below/`
`s/less than/less than or equal to/`

If you think these boundary cases are worth calling out. If they just confuse, feel free to ignore.
___
*[v21.2/follower-reads.md, line 21 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLvWYm5SbZxeRz-eVy:-MlLvWYn3NCJEQRjVONU:b-b9puze) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L21)):*
> ```Markdown
> ## How follower reads work
> 
> Each CockroachDB range tracks a property called its [_closed timestamp_](architecture/transaction-layer.html#closed-timestamps), which means that no new writes can ever be introduced below that timestamp. The closed timestamp advances forward by some target interval behind the current time. If the range receives a write at a timestamp less than its closed timestamp, the write is forced to change its timestamp, which might result in a [transaction retry error](transaction-retry-error-reference.html).
> ```

I think it would be useful to mention that the closed timestamp is advanced on the leaseholder and then a notification about this is sent to each follower. This gives a bit of intuition about how the system is coordinating to make this all safe.
___
*[v21.2/follower-reads.md, line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLwZahC51kpsxHDotY:-MlLwZahC51kpsxHDotZ:b-9xrtg3) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L23)):*
> ```Markdown
> Each CockroachDB range tracks a property called its [_closed timestamp_](architecture/transaction-layer.html#closed-timestamps), which means that no new writes can ever be introduced below that timestamp. The closed timestamp advances forward by some target interval behind the current time. If the range receives a write at a timestamp less than its closed timestamp, the write is forced to change its timestamp, which might result in a [transaction retry error](transaction-retry-error-reference.html).
> 
> With follower reads, any replica on a node can serve a read for a key as long as the time at which the operation is performed (i.e., the [`AS OF SYSTEM TIME`](as-of-system-time.html) value) is less than or equal to the node's closed timestamp.
> ```

s/on a node/in a range/

s/the node's/the range's/
___
*[v21.2/follower-reads.md, line 30 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLwybE4LTGoWTxjned:-MlLwybE4LTGoWTxjnee:bdob6z1) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L30)):*
> ```
> bounded staleness reads provide 
> ```

Looks like this sentence got cut off.
___
*[v21.2/follower-reads.md, line 38 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLxIqAC_q_R04IuBr0:-MlLxIqAC_q_R04IuBr1:bihw2yn) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L38)):*
> ```Markdown
> ### When to use exact staleness reads
> 
> As long as your [`SELECT` operations](select-clause.html) can tolerate reading stale data, follower reads can reduce read latencies and increase throughput.
> ```

These two paragraphs look out of place in a "When to use exact staleness reads" section.
___
*[v21.2/follower-reads.md, line 44 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLxRGf6WqVpGaYxDPP:-MlLxRGf6WqVpGaYxDPQ:b-e0y1y0) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L44)):*
> ```Markdown
> - Use [exact staleness](#exact-staleness-reads) follower reads when:
>   - You need multi-statement reads inside transactions
>   - You can tolerate reading older data (at least 4.8 seconds in the past), to reduce the chance that the historical query timestamp is not quite old enough to prevent blocking on a conflicting write and thus being able to be served by a local replica
> ```

I think it's worth mentioning somewhere that exact staleness reads are slightly cheaper than bounded staleness reads because they don't need to dynamically determine the query timestamp.
___
*[v21.2/follower-reads.md, line 48 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLxbam1XCmNGwrPHhI:-MlLxbam1XCmNGwrPHhJ:b-t008q9) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L48)):*
> ```Markdown
> ## Bounded Staleness Reads
> 
> An _bounded staleness read_ is a historical read that uses a dynamic, system-determined timestamp to minimize staleness while being more tolerant to replication lag than exact staleness reads. They also help increase system availability, since they provide the ability to serve reads from local replicas even in the presence of network partitions or other failures that prevent the SQL gateway from communicating with the leaseholder. For more information about when to use bounded staleness reads, see [When to use bounded staleness reads](#when-to-use-bounded-staleness-reads).
> ```

s/An/A/
___
*[v21.2/follower-reads.md, line 63 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLyN_J6ALPtL7FrVn_:-MlLyN_J6ALPtL7FrVna:b-ww5oc4) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L63)):*
> ```Markdown
> Use [bounded staleness](#bounded-staleness-reads) follower reads when you need:
> 
> - Absolute minimum staleness possible from single statement reads from follower replicas. This is possible because the historical timestamp is chosen dynamically and the least stale timestamp that can be served locally without blocking is used.
> ```

I would remove the phrase "Absolute". In the future, we may introduce optimizations that make this not "as minimally stale as possible", so this phrasing sets the wrong expectations and paints us into a bit of a corner.
___
*[v21.2/follower-reads.md, line 63 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLyh6N1YBkArvFg5dt:-MlLyh6N1YBkArvFg5du:butmsp0) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L63)):*
> ```Markdown
> Use [bounded staleness](#bounded-staleness-reads) follower reads when you need:
> 
> - Absolute minimum staleness possible from single statement reads from follower replicas. This is possible because the historical timestamp is chosen dynamically and the least stale timestamp that can be served locally without blocking is used.
> ```

"from follower replicas and without blocking on conflicting transactions".
___
*[v21.2/follower-reads.md, line 73 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLzHH09I0ax73FyYJu:-MlLzHH09I0ax73FyYJv:bu9ludi) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L73)):*
> ```Markdown
> 
> This video uses [exact staleness]](#exact-staleness-reads) follower reads; it was recorded before [bounded staleness reads](#bounded-staleness-reads) were possible with CockroachDB.
> ```

This link is broken, probably because of the double closing square bracket.
___
*[v21.2/follower-reads.md, line 80 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLzlh646MiYHiaPRlb:-MlLzlh646MiYHiaPRlc:b-p11pz4) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L80)):*
> ```Markdown
> ## Using follower reads
> 
> ### Enable/disable follower reads
> ```

Now that the setting defaults to true, can we remove this section?
___
*[v21.2/follower-reads.md, line 101 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLzPlC5Hhvj-U2wKSR:-MlLzPlC5Hhvj-U2wKSS:b-fp73re) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L101)):*
> ```Markdown
> 2. Go to the [Custom Chart Debug Page in the DB Console](ui-custom-chart-debug-page.html) and add the metric `follower_read.success_count` to the time series graph you see there. The number of follower reads performed by your cluster will be shown.
> 
> XXX: DOES THIS METRIC ALSO COUNT BOUNDED STALENESS READS?
> ```

Yes, it does.
___
*[v21.2/follower-reads.md, line 120 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLzyzw7KVCs1JZazyV:-MlLzyzw7KVCs1JZazyW:beql25t) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L120)):*
> ```Markdown
> 
> - They must be used in a [single-statement (aka implicit) transaction](transactions.html#individual-statements).
> - They must read from a single [range](architecture/overview.html#architecture-range).
> ```

s/range/row/

Even if two rows are on the same range, we still disallow bounded staleness reads over them because allowing this could lead to an app breaking when a range gets split, which would be a bad experience. Mind updating throughout this page?
___
*[v21.2/follower-reads.md, line 136 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlLxop86AGUTrSXaZZO:-MlLxop86AGUTrSXaZZP:b-ppc6yb) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L136)):*
> ```Markdown
> ~~~ sql
> SELECT code FROM promo_codes AS OF SYSTEM TIME with_min_timestamp(now() - '1μs') where code = '0_explain_theory_something';
> ```

I don't think it makes sense to use `now() - '1us'` in the examples with `with_min_timestamp`. First off, 1us is far too short of an interval for this query to actually hit a follower. Secondly, if we want an offset from the present time, we should use `with_max_staleness`. So I'd replace these both with `with_max_staleness('10s')`.
___
*[v21.2/follower-reads.md, line 155 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlM-9Vq6EXbKu0ubRBM:-MlM-9Vq6EXbKu0ubRBN:b-gmtv08) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L155)):*
> ```Markdown
> ~~~
> 
> We can verify using [`EXPLAIN`](explain.html) that the reason this query was able to perform a bounded staleness read is that it performed a point lookup from a single range and did not use any indexes:
> ```

s/range/row/
___
*[v21.2/follower-reads.md, line 155 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlM-AleEOi-jE-GgDKT:-MlM-AleEOi-jE-GgDKU:b-f2znhd) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/follower-reads.md#L155)):*
> ```Markdown
> ~~~
> 
> We can verify using [`EXPLAIN`](explain.html) that the reason this query was able to perform a bounded staleness read is that it performed a point lookup from a single range and did not use any indexes:
> ```

It actually can use indexes, they just must cover the entire query by STORING all columns. We may want to call that suggestion out somewhere.
___
*[v21.2/topology-follower-reads.md, line 97 at r1](https://reviewable.io/reviews/cockroachdb/docs/11896#-MlM-t4h6ZpSHWgin7pl:-MlM-t4h6ZpSHWgin7pm:b9nubg0) ([raw file](https://github.com/cockroachdb/docs/blob/32972f7c548cccc80c9aa016b491c9ee9d1692d9/v21.2/topology-follower-reads.md#L97)):*
> ```Markdown
>     {% include_cached copy-clipboard.html %}
>     ~~~ sql
>     SELECT code FROM postal_codes AS OF SYSTEM TIME with_min_timestamp(now() - '1μs') where id = 5;
> ```

Same point here about using `with_max_staleness`.

PR #11900, opened 2021-10-06
Add optimizer docs on locality optimized search

### LINKED ISSUE #11613: sql: plan multiple table readers on the gateway if plan is local
PR: https://github.com/cockroachdb/cockroach/pull/68679

From release notes:
> If the query is executed locally and needs to perform reads from multiple remote leaseholders, those remote reads might now be done faster. This is especially likely for the case of locality optimized search when there is a local region miss. [#68679][#68679]

rmloveland:
Fixes #11613

Summary of changes:

- Add new section on locality optimized search to the optimizer docs

- Update our previously existing include blurb on locality optimized search to link to the abovementioned optimizer docs


yuzefovich (commented):
> In particular I say in the PR that the async parallel search of other regions doesn't happen if you disable the
> vectorized engine

Yeah, that's about right. (To be precise, the remote lookups are being sent in parallel in both engines, but we have to wait until all of the lookups are fully performed - i.e. there is no short-circuiting in the row-by-row engine, which is present only in the vectorized engine.)

There is also a bit more nuance there - even if the vectorized engine is used, but the plan still contains pieces of the row-by-row infrastructure, we cannot short-circuit either. Most common operations that require the usage of row-by-row components are lookup joins. The way to see this is by running `EXPLAIN (VEC)` and observing whether there are entries with `rowexec` prefix (like `rowexec.joinReader`).

However, this is probably going too much into the weeds, so I'd probably leave things as they currently are.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11900#-:-MlM8NBA2tLg_KCE76z7:b-phwa1n)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rmloveland and @rytaft)

___
*[v21.2/cost-based-optimizer.md, line 103 at r1](https://reviewable.io/reviews/cockroachdb/docs/11900#-MlM6fQd3I8ICRxy-zcZ:-MlM6fQd3I8ICRxy-zc_:b-qo9cwq) ([raw file](https://github.com/cockroachdb/docs/blob/3deddf43c093ea7e432be9366a5707a8614daf03/v21.2/cost-based-optimizer.md#L103)):*
> ```Markdown
> ## Locality optimized search in multi-region clusters
> 
> In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), will avoid sending requests to nodes in other regions when it can instead read a value from a unique column that is stored locally.  This capability is known as _locality optimized search_.
> ```

super nit: not sure if this is intentional, but there are two spaces before `This`.
___
*[v21.2/cost-based-optimizer.md, line 105 at r1](https://reviewable.io/reviews/cockroachdb/docs/11900#-MlM6xoM-GzRWNqgJHAB:-MlM6xoM-GzRWNqgJHAC:brmxaz) ([raw file](https://github.com/cockroachdb/docs/blob/3deddf43c093ea7e432be9366a5707a8614daf03/v21.2/cost-based-optimizer.md#L105)):*
> ```Markdown
> In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), will avoid sending requests to nodes in other regions when it can instead read a value from a unique column that is stored locally.  This capability is known as _locality optimized search_.
> 
> Even if the value cannot be read locally, CockroachDB can still take advantage of the fact that some of the other regions are much closer than others (with lower latency). In this case, it performs all lookups against the remote regions in parallel and returns the result once a row is retrieved, without having wait for each lookup to come back.
> ```

nit: I think in 21.2 we support locality optimized search when it returns at most 10000 rows, so maybe do: "once a row is retrieved" -> "once it is retrieved". I'm not sure what the best to phrase it, but I'd like to not make the reader infer that only a single row locality optimized lookup is supported.
___
*[v21.2/cost-based-optimizer.md, line 107 at r1](https://reviewable.io/reviews/cockroachdb/docs/11900#-MlM7PNf4qoVEQ0Hk1It:-MlM7PNf4qoVEQ0Hk1Iu:b-jno0pm) ([raw file](https://github.com/cockroachdb/docs/blob/3deddf43c093ea7e432be9366a5707a8614daf03/v21.2/cost-based-optimizer.md#L107)):*
> ```Markdown
> Even if the value cannot be read locally, CockroachDB can still take advantage of the fact that some of the other regions are much closer than others (with lower latency). In this case, it performs all lookups against the remote regions in parallel and returns the result once a row is retrieved, without having wait for each lookup to come back.
> 
> This can result in increased performance in multi-region deployments, since it means that results can be returned from wherever they're found first without waiting for all results to return.
> ```

nit: maybe "all results" -> "all lookups"?

For the context, the way locality optimized search works is as follows:
- say we need to scan `N` rows
- suppose `L` rows are found locally (this could be 0). If `L >= N`, then no remote lookups are performed.
- if `L < N`, then we issue all remote lookups in parallel. Once all of the lookups produce `N-L` rows combined, we cancel the current outstanding lookups and finish the query execution.

Given this context, "all results" is a bit ambiguous to me.

Issue #11370, opened 2021-09-08
sql: change information_schema routines data types


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/67641

From release notes:
> Changed `information_schema.routines` data types at columns `interval_precision`, `result_cast_char_octet_length` and `result_cast_datetime_precision` to `INT`. [#67641][#67641] {% comment %}doc{% endcomment %}

Issue #11367, opened 2021-09-08
sql: added missing tables from mysql information_schema


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/66795

From release notes:
> Added tables to `information_schema` that are present on MySQL. The tables are not populated and are entirely empty. `column_statistics`, `columns_extensions`, `engines`, `events`, `files`, `keywords`, `optimizer_trace`, `partitions`, `plugins`, `processlist`, `profiling`, `resource_groups`, `schemata_extensions`, `st_geometry_columns`, `st_spatial_reference_systems`, `st_units_of_measure`, `table_constraints_extensions`, `tables_extensions`, `tablespaces`, `tablespaces_extensions`, `user_attributes`. [#66795][#66795] {% comment %}doc{% endcomment %}

Issue #11353, opened 2021-09-08
sql: adding missing tables on information_schema


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/65854

From release notes:
> Adding empty missing tables on `information_schema` for compatibility: `attributes`, `check_constraint_routine_usage`, `column_column_usage`, `column_domain_usage`, `column_options`, `constraint_table_usage`, `data_type_privileges`, `domain_constraints`, `domain_udt_usage`, `domains`, `element_types`, `foreign_data_wrapper_options`, `foreign_data_wrappers`, `foreign_server_options`, `foreign_servers`, `foreign_table_options`, `foreign_tables`, `information_schema_catalog_name`, `role_column_grants`, `role_routine_grants`, `role_udt_grants`, `role_usage_grants`, `routine_privileges`, `sql_features`, `sql_implementation_info`, `sql_parts`, `sql_sizing`, `transforms`, `triggered_update_columns`, `triggers`, `udt_privileges`, `usage_privileges`, `user_defined_types`, `user_mapping_options`, `user_mappings`, `view_column_usage`, `view_routine_usage`, `view_table_usage`. [#65854][#65854] {% comment %}doc{% endcomment %}

PR #11911, opened 2021-10-07
information_schema tables added for pg,mysql compatibility

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/11353.
Fixes https://github.com/cockroachdb/docs/issues/11367.
Fixes https://github.com/cockroachdb/docs/issues/11370.

In this PR, I've added a list of all of the empty `information_schema` tables introduced in 21.2 for compatibility with postgresql and mysql tooling. 

Note that I've removed the sections dedicated to the two empty tables included in previous versions, as these tables' contents were empty and identical to their postgresql analogs.

### DIFF
--- a/v21.2/information-schema.md
+++ b/v21.2/information-schema.md
@@ -229,7 +229,7 @@ Column | Description
 
 ### referential_constraints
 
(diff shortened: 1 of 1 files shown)


ericharmeling:
TFTRs!

PR #11912, opened 2021-10-07
Address TA 71002.

ghost:


stbof:
> :lgtm:
> 
> > Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/11912#-:-MlRZWafA8FWwHFjj1fF:bnfp4nl)_ status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ianjevans and @stbof)
> 
> _[advisories/a71002.md, line 13 at r3](https://reviewable.io/reviews/cockroachdb/docs/11912#-MlRZ3jg0bbIM7cvMv9a:-MlRZ3jg0bbIM7cvMv9b:b-ru0q58) ([raw file](https://github.com/cockroachdb/docs/blob/0204284e964681b826a97f20cbcec77676670063/advisories/a71002.md#L13)):_
> 
> > ```gfm
> > A bug was introduced in CockroachDB v21.1.9 that causes the database to drop `WHERE` predicates from prepared statements in specific circumstances.
> > 
> > The bug can only be triggered when a prepared statement references a boolean column in top-level AND clauses in `WHERE` clauses or `JOIN … ON` clauses. In such statements, it is possible for the part of the filter referencing the boolean column, or any other child of an adjacent `AND` expression to be omitted during execution.
> > ```
> 
> nit: there's another AND that could use some backticks

Thanks for the pointer @rytaft . Fixed.

PR #11939, opened 2021-10-12
Adding docs for securing operator webhooks

pseudomuto:
We commission both CA and server certificates for securing the webhooks in the operator. By default this is a no-touch process which generates a self-signed CA certificate and uses that to generate a server certificate at startup.

If people want to use their own CA for this, it is possible, but we need to tell them how. This PR addresses that.

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -91,4 +91,4 @@ release_info:
     version: v21.2.0-rc.2
 site_title: CockroachDB Docs
 url: https://www.cockroachlabs.com
-operator_version: v2.2.0
+operator_version: v2.3.0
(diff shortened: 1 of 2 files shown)


pseudomuto:
@taroface you mentioned that you were going to take this over. Just wanted to make sure that's still the case.

taroface:
@pseudomuto Yes, aiming to do it this week!

_config_base.yml
```diff
@@ -91,4 +91,4 @@ release_info:
     version: v21.2.0-rc.2
 site_title: CockroachDB Docs
 url: https://www.cockroachlabs.com
-operator_version: v2.2.0
+operator_version: v2.3.0
```
pseudomuto, line 94: This should now be `2.4.0` since #12198 has been merged

PR #11941, opened 2021-10-12
add example in v20.2

luharuka:


v20.2/alter-column.md
```diff
@@ -59,6 +59,8 @@ For an example of `ALTER COLUMN TYPE`, see [Increase a column type's precision](
 
 ## Examples
 
+>ALTER TABLE student ALTER clg_fee type DECIMAL (12,2);
```
ericharmeling, line 62: All example code snippets must be properly formatted.

The code snippet you provided should look like this:

```
~~~ sql
> ALTER TABLE student ALTER clg_fee type DECIMAL (12,2);
~~~
```

For reference, see the first code snippet in the example with the title `### Set or change a DEFAULT value`:

```
~~~ sql
> ALTER TABLE subscriptions ALTER COLUMN newsletter SET DEFAULT true;
~~~
```

v20.2/alter-column.md
```diff
@@ -59,6 +59,8 @@ For an example of `ALTER COLUMN TYPE`, see [Increase a column type's precision](
 
 ## Examples
```
ericharmeling, line 61: On this page, all examples must be within a subsection of `## Examples`. To make a subsection, create a title with a level-3 header (e.g., `### Update the data type of two columns`). 

For reference, see the example directly following `## Examples`. It has the level-3 header `### Set or change a DEFAULT value`.

v20.2/alter-column.md
ericharmeling, line 62: To be consistent with the other examples on this page, your example should have some text explaining exactly what the example is doing.

For reference, see the example with the title `### Set or change a DEFAULT value`.

v20.2/alter-column.md
ericharmeling, line 61: I think this example should be appended to the end of the examples provided, as changing the data type of two columns at once is not the most common use of `ALTER COLUMN`

ericharmeling (changes requested):
@luharuka Thanks for the contribution! Please see my comments.

PR #11970, opened 2021-10-14
Add docs on multi-tier load-based rebalancing

### LINKED ISSUE #11240: kvserver: actuate load-based replica rebalancing under heterogeneous localities
https://github.com/cockroachdb/cockroach/pull/65379 --- Release note (performance improvement): QPS-based replica rebalancing is now aware of different constraints placed on different replication zones. This means that heterogeneously loaded replication zones (for instance, regions) will achieve a more even distribution of QPS within the stores inside each such zone.  /cc @cockroachdb/kv

rmloveland:
Fixes #11240

### DIFF
--- a/v21.2/architecture/replication-layer.md
+++ b/v21.2/architecture/replication-layer.md
@@ -112,7 +112,7 @@ Finally, note that the process described above is lazily initiated: it only occu
 
 #### Leaseholder rebalancing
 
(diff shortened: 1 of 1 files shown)


aayushshah15:
These updates mostly look good to me, but the thing to note is that we had to temporarily revert https://github.com/cockroachdb/cockroach/pull/65379 because it introduced a new regression over the previous implementation. 

We're going to fix the regression and unrevert that patch soon, but this won't make it in until 21.2.2. With that in mind, I assume we want to hold off on merging this change?

rmloveland:
> this won't make it in until 21.2.2. With that in mind, I assume we want to hold off on merging this change?

Ah ok, didn't realize.  Yes, sounds good - will pause on this until v21.2.2

PR #12032, opened 2021-10-20
Cloud API reference docs for bug bash.

ianjevans:
For internal testing during the bug bash. We won't make the API public in the navigation until it's ready for the public.


ianjevans:
> In the example for PUT ([https://cockroachlabs.cloud/api/v1/orgs/{organizationId}/clusters/{clusterId}/nodes](https://cockroachlabs.cloud/api/v1/orgs/%7BorganizationId%7D/clusters/%7BclusterId%7D/nodes)) to set the number of nodes I see the following payload, which doesn't look like a real example. The description says: Region keys should match the cloud provider's zone code. For example, for Oregon, set region_name to "us-west2" for GCP and "us-west-2" for AWS. Values represent the node count.
> ...
> Would it make sense to show real regions and node counts?
> 
> The response example also includes this property which doesn't apply to that API: "serverless": { "regions": [ "string" ], "spendLimit": 0 },

The reference is generated from the [proto file in the `managed-service` project](https://github.com/cockroachlabs/managed-service/blob/master/console/apipb/api.proto). The output HTML is extremely difficult to hand-edit, unfortunately. I'll talk with @erademacher about how to make these examples more useful, but for now we'd like to just get the reference up for the bug bash tomorrow, even if it's rough around the edges.

ianjevans:
Thanks, @stbof.

Issue #11351, opened 2021-09-08
sql: "char" is supposed to truncate long values


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/66422

From release notes:
> The `"char"` column type will now truncate long values, in line with PostgreSQL. [#66422][#66422] {% comment %}doc{% endcomment %}

PR #12053, opened 2021-10-25
Documented "char" type

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/11351.

### DIFF
--- a/v21.2/string.md
+++ b/v21.2/string.md
@@ -25,6 +25,8 @@ For PostgreSQL compatibility, CockroachDB supports the following `STRING`-relate
 
 These types are functionality identical to `STRING`.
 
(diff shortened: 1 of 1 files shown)


v21.2/string.md
```diff
@@ -25,6 +25,8 @@ For PostgreSQL compatibility, CockroachDB supports the following `STRING`-relate
 
 These types are functionality identical to `STRING`.
 
+CockroachDB also supports the single-byte `"char"` special character type. As in PostgreSQL, this special type is intended for internal use in [system catalogs](system-catalogs.html), and has a storage size of 1 byte. CockroachDB truncates all values of type `"char"` to a single character.
```
stbof, line 28:
```suggestion
CockroachDB supports the single-byte `"char"` special character type. As in PostgreSQL, this special type is intended for internal use in [system catalogs](system-catalogs.html), and has a storage size of 1 byte. CockroachDB truncates all values of type `"char"` to a single character.
```

stbof (approved):
See suggestion. YMMV.

ericharmeling:
TFTR!

Issue #11070, opened 2021-08-18
Document High Throughput Changefeed settings


amruss:
if you're designing a system that needs to push through a lot of changefeed messages, whether it be steady traffic or a burst in traffic, you should ideally set up your changefeed in a certain way. We should maybe include this in the "advanced operator" section mentioned: #11001.

Resolved: Don't set resolved messages - or set a high resolved message
Batch messages
high memory budget 
avro format (JSON encoding can be a bottle neck)
parallelism in scan requests- for backfills (increase) 

heuristic is 3x the number of nodes, but we internally cap at 100 - if you set the setting you can change however you want.

Negative: higher CPU usage - for 100 node cluster, each node will issue requests to the same node, 300 requests

Turn on time based iterator cluster setting

kathancox:
Closed by #12100

Issue #11069, opened 2021-08-18
Document High Consistency Changefeed Tunings


amruss:
if you're designing a system that relies on not missing any messages coming through a changefeed, you should ideally set up your changefeed in a certain way. We should maybe include this in the "advanced operator" section mentioned: https://github.com/cockroachdb/docs/issues/11001. 

Gc.ttl cluster setting: increasing this setting gives you a higher recoverability window if the changefeed fails for some reason
high gcttl or protect_on_pause + on failure pause
Monitoring: 
kafka_sink_config: {"RequiredAcks": "ALL" }
memory budget
maybe fail on schema change if you’re worried about data corruption downstream
what sink you use: kafka great, cloud storage good, http not good- dont know if they’re lying

kathancox:
Closed by #12100

PR #12100, opened 2021-10-27
Memory budget setting added to CDC docs and advanced config section expanded

### LINKED ISSUE #10939: Document memory budget cluster setting
Also backported to 21.1

changefeed.memory.per_changefeed_limit: This cluster setting gives our internal memory buffers more room to work with during times of heavy traffic or when the sink is unreachable. Default value is 1GB. Increasing may lead to more changefeed memory usage overall

kathancox:
Closes #10939 #11069 #11070

Added documentation on the new memory budget cluster setting to the changefeed docs. Took the opportunity to incorporate this into the newer _draft_ advanced configuration setting. Adding recommendations for high throughput & consistency.

Couple of questions / specific feedback here:

* For high throughput is the setting `kv.rangefeed.catchup_scan_iterator_optimization.enabled` impactful here? (Not currently mentioned in this PR)
* Wondering if we should include the bullet on Avro vs JSON format since cloud storage and webhooks can only emit JSON? Or would we rather suggest that Kafka's performance is going to be better?


kathancox:
Thanks Abbey, I have added that setting in to 21.2. — it's not a public setting, so let me know if that's something that will/should change. I have put the tracking issue for the `diff` option beside it. I haven't created a limitation as yet since the setting is public, lmk if you think that would be best.

Re the Kafka/avro/cloud storage point. I have left as is for your/eng full review. My thought is that if I take out cloud storage from these sections then it would seem like each of these bullets are only applicable to Kafka users, but I'm assuming enabling just some of these settings could promote consistency/throughput for cloud storage users too?

miretskiy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12100)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @stevendanna)

___
*[v21.1/create-changefeed.md, line 192 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaIFO08ToFhMBSgXv2:-MnaIFO08ToFhMBSgXv3:bjeug52) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.1/create-changefeed.md#L192)):*
> ```Markdown
> ### Tuning for high consistency
> 
> When designing a system that relies on high consistency — that is, not missing any messages coming through a changefeed — consider the following settings and configuration:
> ```

I think it's important to emphasize that (to the best of our knowledge), changefeeds do not miss messages.
However, we rely on downstream sink acknowledgements to determine if the message was delivered.
So, if the acknowledgement itself is not "high consistency", then messages can be lost -- but again, *not* through the fault
of changefeed...

Perhaps we can use "durability" when describing this?  As in "tuning for high durability delivery"?  @stevendanna any additional ideas?
___
*[v21.1/create-changefeed.md, line 205 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaKXZ5EyqoIulr3ZaQ:-MnaKXZ61S7SJmfnwDd6:b-l30787) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.1/create-changefeed.md#L205)):*
> ```Markdown
> When designing a system that needs to emit a lot of changefeed messages, whether it be steady traffic or a burst in traffic, consider the following settings and configuration:
> 
> * Avoid using the [`resolved`](#resolved-option) option or set this to a higher duration. This will help to reduce emitted messages.
> ```

I think for 21.1 we should always set resolved to a high value... 10m+.
21.2 is better in how it handles this option, but in 21.1 it's always better to set this to a higher value.
___
*[v21.1/create-changefeed.md, line 209 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaJFLr3yHTPDEhn9iG:-MnaJFLr3yHTPDEhn9iH:b82ntae) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.1/create-changefeed.md#L209)):*
> ```Markdown
> * Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering for a changefeed. This is useful in situations of heavy traffic to avoid an error due to the memory buffer overfilling.
> * Use `avro` as the emitted message [format](#format) option with Kafka sinks; JSON encoding can potentially create a slowdown.
> * Increase the `changefeed.backfill.concurrent_scan_requests` setting, which controls the number of concurrent scan requests per node issued during a [backfill event]. The default behavior, when this setting is at `0`, is that each node will issue 3 scan requests per node in a cluster (to a maximum of 100). While increasing this number will allow for higher throughput, it **will increase CPU usage overall**.
> ```

I would perhaps indicate that the cluster load, including CPU and IO will increase. Ditto below.
___
*[v21.2/create-changefeed.md, line 240 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaJ0Mr0yLZxJtxK-11:-MnaJ0Ms2Z_XZYBWwe9r:b-4xfu1) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.2/create-changefeed.md#L240)):*
> ```Markdown
> * Avoid using the [`resolved`](#resolved-option) option or set this to a higher duration. This will help to reduce emitted messages.
> * Batch messages to your sink. See the [`Flush`](#kafka-flush) parameter for the `kafka_sink_config` option. When using cloud storage sinks, use the [`file_size`](#file-size) parameter to flush a file when it exceeds the specified size.
> * Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering for a changefeed. This is useful in situations of heavy traffic to avoid an error due to the memory buffer overfilling.
> ```

Technically, 21.2 -- this should not happen -- we have full pushback mechanism.  I would de-emphasize this setting in this release.
___
*[v21.2/create-changefeed.md, line 243 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaKAaM5IiaDoUxLHBT:-MnaKAaM5IiaDoUxLHBU:b-cbq04r) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.2/create-changefeed.md#L243)):*
> ```Markdown
> * Use `avro` as the emitted message [format](#format) option with Kafka sinks; JSON encoding can potentially create a slowdown.
> * Increase the `changefeed.backfill.concurrent_scan_requests` setting, which controls the number of concurrent scan requests per node issued during a backfill event. The default behavior, when this setting is at `0`, is that each node will issue 3 scan requests per node in a cluster (to a maximum of 100). While increasing this number will allow for higher throughput, it **will increase CPU usage overall**.
> * <span class="version-tag">New in v21.2:</span> Enable the `kv.rangefeed.catchup_scan_iterator_optimization.enabled` setting to have [rangefeeds](stream-data-out-of-cockroachdb-using-changefeeds.html#enable-rangefeeds) use time-bound iterators for catch-up scans when possible. Catch-up scans are run for each rangefeed request. This setting improves the overall performance of changefeeds during some [range-split operations](architecture/distribution-layer.html#range-splits). Note that the [`WITH diff` option](#diff-opt) cannot be used with this setting. [Tracking GitHub Issue](https://github.com/cockroachdb/cockroach/pull/69191)
> ```

`WITH diff` can be used -- it's just this setting is a no-op.

kathancox:
Hey @miretskiy , Thank you for the feedback! Let me know if this next commit covers your comments — I see what you mean re "high consistency".

stevendanna (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12100)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @miretskiy)

___
*[v21.2/create-changefeed.md, line 243 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaKAaM5IiaDoUxLHBT:-Mne_pk9CFxidtfdY844:bi9mfv3) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.2/create-changefeed.md#L243)):*
The update here isn't quite right. The current wording might make people believe that `WITH diff` is a no-op.  Perhaps something like:

> This setting improves the performance of changefeeds during some range-split operations.  Changefeeds using the `WITH DIFF` option do not currently receive any benefit from this setting.

On the "very advanced" side, we might want to recommend lowering `kv.rangefeed.concurrent_catchup_iterators`

kathancox:
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12100)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @miretskiy)
> 
> _[v21.2/create-changefeed.md, line 243 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaKAaM5IiaDoUxLHBT:-Mne_pk9CFxidtfdY844:bi9mfv3) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.2/create-changefeed.md#L243)):_ The update here isn't quite right. The current wording might make people believe that `WITH diff` is a no-op. Perhaps something like:
> 
> > This setting improves the performance of changefeeds during some range-split operations.  Changefeeds using the `WITH DIFF` option do not currently receive any benefit from this setting.
> 
> On the "very advanced" side, we might want to recommend lowering `kv.rangefeed.concurrent_catchup_iterators`

Ahhh good point @stevendanna, my first draft got that the wrong way round... thank you for the rework! 

On the `concurrent_catchup_iterators` setting, I'm happy to add in if that makes sense here? @amruss? If we want to include, are there any additional warnings that are needed with this one? 

My only other question is: besides the memory limit setting, the other settings here are currently not public. By doc'ing here, we're kind of making public, and in theory, maybe should be set to public on the eng side? However, I see an argument that these are _advanced_ settings that we don't necessarily want on show without fair warning...

miretskiy:
Thanks for addressing my comments, @kathancox . It's :lgtm: to me.

stevendanna (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12100)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @amruss, @kathancox, and @miretskiy)

___
*[v21.2/create-changefeed.md, line 243 at r3](https://reviewable.io/reviews/cockroachdb/docs/12100#-MnaKAaM5IiaDoUxLHBT:-Mo3mm4q6GWiij8pBURp:b-3xn6ds) ([raw file](https://github.com/cockroachdb/docs/blob/42f8d5bc8ebd32b1ae202ee37e475b7fead97cc4/v21.2/create-changefeed.md#L243)):*
Your update looks good.

> On the concurrent_catchup_iterators setting, I'm happy to add in if that makes sense here? @amruss? If we want to include, are there any additional warnings that are needed with this one?

Honestly, perhaps we should leave it out for now.  While it is important in very large clusters, it likely isn't the kind of thing we want people turning on "just because." Also, it actually doesn't fit here perfectly because we typically lower it to limit foreground impact of catch-up scans rather than to improve throughput necessarily.

> My only other question is: besides the memory limit setting, the other settings here are currently not public. By doc'ing here, we're kind of making public, and in theory, maybe should be set to public on the eng side? However, I see an argument that these are advanced settings that we don't necessarily want on show without fair warning...

I think having them in this advanced section for now is probably good. This is an "advanced" section after all.
___
*[v21.2/create-changefeed.md, line 223 at r5](https://reviewable.io/reviews/cockroachdb/docs/12100#-Mo3nz4l2x6mDI9MNjWl:-Mo3nz4l2x6mDI9MNjWm:b-g8npe2) ([raw file](https://github.com/cockroachdb/docs/blob/3d86f8d987910b8ea6c795de18cf0f1eccdce9c5/v21.2/create-changefeed.md#L223)):*
> ```Markdown
> 
> * Increase the number of seconds before [garbage collection](architecture/storage-layer.html#garbage-collection) with the [`gc.ttlseconds`](configure-replication-zones.html#gc-ttlseconds) setting to provide a higher recoverability window for data if a changefeed fails. For example, if a sink is unavailable, changes are queued until the sink is available again. While the sink is unavailable, changes will be retried until the garbage collection window is reached and then the data is removed.
>   * You can also use the [`protect_data_from_gc_on_pause`](#protect-pause) option in combination with [`on_error=pause`](#on-error) to explicitly pause a changefeed on error and to then protect the changes from garbage collection.
> ```

Note for the CDC team: We might want to make sure we have a test for this combination if we are going to document it.

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,40 @@ Kafka has the following topic limitations:
...
+
+* Increase the number of seconds before [garbage collection](architecture/storage-layer.html#garbage-collection) with the [`gc.ttlseconds`](configure-replication-zones.html#gc-ttlseconds) setting to provide a higher recoverability window for data if a changefeed fails. For example, if a sink is unavailable, changes are queued until the sink is available again. While the sink is unavailable, changes will be retried until the garbage collection window is reached and then the data is removed.
+  * You can also use the [`protect_data_from_gc_on_pause`](#protect-pause) option to protect the changes from garbage collection.
+* Determine what a successful write to Kafka is with the [`kafka_sink_config: {"RequiredAcks": "ALL"}`](#kafka-required-acks) option, which provides the highest consistency level.
+* Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering per single changefeed. Increasing this limit is useful to avoid errors in situations where the sink is unreachable and the memory buffer fills up.
```
amruss, line 192: Since we have full pushback in 21.2, this setting probably is not needed. We would strongly recommend in 21.1 though

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,40 @@ Kafka has the following topic limitations:
...
+* Use [Kafka](#kafka) or [cloud storage](#cloud-storage-sink) sinks when tuning for high durability in changefeeds.
```
amruss, line 193: I would maybe reword this and just add the reason why we have the recommendation, in that kafka and cloud storage have built-in highly advanced protocols, the webhook sink is flexible and thus it is on the user to understand how messages are acked/comitted (ie their system could just drop the message as soon as they got it we have no way of knowing, but the user does have a way of knowing)

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,40 @@ Kafka has the following topic limitations:
...
+* Ensure data accuracy downstream by using the [`schema_change_events`](#schema-events) and [`schema_schange_policy`](#schema-policy) options. For example, setting `schema_change_events=column_changes` and `schema_change_policy=stop` will trigger an error to the `cockroach.log` file on a [schema change](stream-data-out-of-cockroachdb-using-changefeeds.html#schema-changes-with-column-backfill).
```
amruss, line 194: Hmm it's not really that data is not accurate when a schema change happens, its just that downstream systems may not be set up to ingest the new format post schema change

amruss: I would also say the changefeed will fail on `schema_change_policy=stop`, that isn't obvious from this paragraph

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,40 @@ Kafka has the following topic limitations:
...
+
+### Tuning for high throughput
+
+When designing a system that needs to emit a lot of changefeed messages, whether it be steady traffic or a burst in traffic, consider the following settings and configuration:
+
+* Set the [`resolved`](#resolved-option) option to a higher duration (e.g. 10 minutes or more) to reduce emitted messages.
+* Batch messages to your sink. See the [`Flush`](#kafka-flush) parameter for the `kafka_sink_config` option. When using cloud storage sinks, use the [`file_size`](#file-size) parameter to flush a file when it exceeds the specified size.
+* Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering for a changefeed. This is useful in situations of heavy traffic to avoid an error due to the memory buffer overfilling.
+* Use `avro` as the emitted message [format](#format) option with Kafka sinks; JSON encoding can potentially create a slowdown.
```
amruss, line 203: Would add to use compression in cloud storage sinks with avro (gzip)

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,40 @@ Kafka has the following topic limitations:
...
+* Increase the `changefeed.backfill.concurrent_scan_requests` setting, which controls the number of concurrent scan requests per node issued during a backfill event. The default behavior, when this setting is at `0`, is that each node will issue 3 scan requests per node in a cluster (to a maximum of 100). While increasing this number will allow for higher throughput, it **will increase the cluster load overall**, including CPU and IO usage.
```
amruss, line 204: I'm not sure that setting this to 0 would give you the default behavior, but need to double check

v21.1/create-changefeed.md
amruss: Oh realized this is for 21.1, would apply to the 21.2 version of this page

v21.1/create-changefeed.md
amruss: Ah I realize this is 21.1 docs, so webhook wouldn't apply - applies for 21.2

v21.2/create-changefeed.md
```diff
@@ -202,9 +202,42 @@ Kafka has the following topic limitations:
...
+
+### Tuning for high durability delivery
+
+When designing a system that relies on high durability of message delivery — that is, not missing any message acknowledgement at the downstream sink — consider the following settings and configuration. Before tuning these settings we recommend reading details on our [changefeed at-least-once-delivery guarantee](stream-data-out-of-cockroachdb-using-changefeeds.html#ordering-guarantees).
+
+* Increase the number of seconds before [garbage collection](architecture/storage-layer.html#garbage-collection) with the [`gc.ttlseconds`](configure-replication-zones.html#gc-ttlseconds) setting to provide a higher recoverability window for data if a changefeed fails. For example, if a sink is unavailable, changes are queued until the sink is available again. While the sink is unavailable, changes will be retried until the garbage collection window is reached and then the data is removed.
+  * You can also use the [`protect_data_from_gc_on_pause`](#protect-pause) option in combination with [`on_error=pause`](#on-error) to explicitly pause a changefeed on error and to then protect the changes from garbage collection.
```
amruss, line 223: I would say instead of going into a failure state, because if we have a retryable error we will still retry and not pause

amruss (commented):
Left a few comments, nothing major

v21.1/create-changefeed.md
kathancox: OK, will remove this setting from the 21.2 section here.

kathancox:
@amruss I have added Rich for the docs review, but I know that we might want to edit that last bullet on the concurrent_scan_requests.

amruss:
changefeed.backfill.concurrent_scan_requests - setting to 0 will give you the default behavior, one clarification is that the default behavior is that the number of scan requests will be 3* the number of nodes, up to 100;

v21.1/create-changefeed.md
```diff
@@ -172,9 +170,41 @@ Kafka has the following topic limitations:
...
+* Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering for a changefeed. This is useful in situations of heavy traffic to avoid an error due to the memory buffer overfilling.
+* Use `avro` as the emitted message [format](#format) option with Kafka sinks; JSON encoding can potentially create a slowdown.
+* Use the [`compression` option](#compression-opt) in cloud storage sinks with JSON to compress the changefeed data files.
+* Increase the `changefeed.backfill.concurrent_scan_requests` setting, which controls the number of concurrent scan requests per node issued during a backfill event. The default behavior, when this setting is at `0`, is that each node will issue 3 scan requests per node in a cluster (to a maximum of 100). While increasing this number will allow for higher throughput, it **will increase the cluster load overall**, including CPU and IO usage.
```
rmloveland: if this is a cluster setting, would be good to make it link to the cluster settings page

v21.2/create-changefeed.md
```diff
@@ -202,9 +202,42 @@ Kafka has the following topic limitations:
...
+* Batch messages to your sink. See the [`Flush`](#kafka-flush) parameter for the `kafka_sink_config` option. When using cloud storage sinks, use the [`file_size`](#file-size) parameter to flush a file when it exceeds the specified size.
+* Set the [`changefeed.memory.per_changefeed_limit`](cluster-settings.html) cluster setting to a higher limit to give more memory for buffering for a changefeed. This is useful in situations of heavy traffic.
+* Use `avro` as the emitted message [format](#format) option with Kafka sinks; JSON encoding can potentially create a slowdown.
+* Use the [`compression` option](#compression-opt) in cloud storage sinks with JSON to compress the changefeed data files.
+* Increase the `changefeed.backfill.concurrent_scan_requests` setting, which controls the number of concurrent scan requests per node issued during a backfill event. The default behavior, when this setting is at `0`, is that each node will issue 3 scan requests per node in a cluster (to a maximum of 100). While increasing this number will allow for higher throughput, it **will increase the cluster load overall**, including CPU and IO usage.
```
rmloveland: this could also link to cluster settings

v21.2/create-changefeed.md
```diff
@@ -202,9 +202,42 @@ Kafka has the following topic limitations:
...
+* <span class="version-tag">New in v21.2:</span> Enable the `kv.rangefeed.catchup_scan_iterator_optimization.enabled` setting to have [rangefeeds](stream-data-out-of-cockroachdb-using-changefeeds.html#enable-rangefeeds) use time-bound iterators for catch-up scans when possible. Catch-up scans are run for each rangefeed request. This setting improves the performance of changefeeds during some [range-split operations](architecture/distribution-layer.html#range-splits). Changefeeds using the [`WITH diff` option](#diff-opt) do not currently receive any benefit from this setting.
```
rmloveland, line 238: "setting" could also link to cluster settings

rmloveland (approved):
LGTM, just added a couple RFLs (request for links)

Issue #11506, opened 2021-09-08
cli: make --global flag for demo public


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/62435

From release notes:
> There will now be a message upon start-up on [`cockroach demo --global`](../v21.2/cockroach-demo.html) indicating that latencies between nodes will simulate real-world latencies. [#62435][#62435]

Issue #11505, opened 2021-09-08
cli: make --global flag for demo public


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/62435

From release notes:
> The `--global` flag for [`cockroach demo`](../v21.2/cockroach-demo.html) is now advertised. This flag simulates latencies in multi-node demo clusters when the nodes are set in different regions to simulate real-life global latencies. [#62435][#62435]

Issue #11504, opened 2021-09-08
cli: make --global flag for demo public


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/62435

From release notes:
> Previously, `\demo shutdown <node_idx>` would error if `--global` was set. This will now error gracefully as an unsupported behavior. [#62435][#62435]

PR #12134, opened 2021-11-01
cockroach demo --global flag no longer experimental

gemma-shay:
Closes:
https://github.com/cockroachdb/docs/issues/11506
https://github.com/cockroachdb/docs/issues/11505
https://github.com/cockroachdb/docs/issues/11504


v21.2/cockroach-demo.md
```diff
@@ -11,6 +11,7 @@ The `cockroach demo` [command](cockroach-commands.html) starts a temporary, in-m
 - By default, `cockroach demo` starts in secure mode using TLS certificates to encrypt network communication. It also serves a local [DB Console](#connection-parameters) that does not use TLS encryption.
 - Each instance of `cockroach demo` loads a temporary [Enterprise license](https://www.cockroachlabs.com/get-cockroachdb) that expires after an hour. To prevent the loading of a temporary license, set the `--disable-demo-license` flag.
 -  `cockroach demo` opens the SQL shell with a new [SQL user](authorization.html#sql-users) named `demo`. The `demo` user is assigned a random password and granted the [`admin` role](authorization.html#admin-role).
+-  <span class="version-tag">New in v21.2:</span> The <a href="#global-flag">`--global`</a> flag is no longer experimental. This flag simulates realistic global latencies in a multi-region cluster by setting the [`--locality` flag on node startup](cockroach-start.html#locality) to three different regions.
```
ericharmeling, line 14: IMO, we don't need this bullet.

If you disagree and want to keep it, we should probably make it read more like this:
```
The <a href="#global-flag">`--global`</a> flag simulates realistic global latencies in a multi-region cluster by setting the [`--locality` flag on node startup](cockroach-start.html#locality) to three different regions.
```

gemma-shay: Removing sgtm, wasn't sure about it. Should we have a `new in version` tag in the table of flags?

ericharmeling: > Should we have a new in version tag in the table of flags?

Since the flag already existed in 21.1, probably not. Just removing the experimental designation is probably enough.

Issue #11490, opened 2021-09-08
sql, import: add metrics for max_row_size guardrails


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/69457

From release notes:
> Added four new metrics, `sql.guardrails.max_row_size_{log|err}.count{.internal}`, which are incremented whenever a large row violates the corresponding `sql.guardrails.max_row_size_{log|err}` limit. [#69457][#69457]

PR #12190, opened 2021-11-04
Add guardrail metrics. Remove obsolete metrics file.

stbof:
Flxes #11490. 

The issue mentions 4 metrics. Should I be documenting the `internal` metrics?


yuzefovich (approved):
> Should I be documenting the `internal` metrics?

I'm not sure what our policy is about this, but it looks like we didn't document similar `internal` metrics for other things (for example, there exist both `sql.distsql.select.count` and `sql.distsql.select.count.internal`, but only the former is mentioned in `metric-names.md`), so I think we should not document the new `internal` metrics either.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12190#-:-MnhAWkY4tFeYpio3nWw:b-4vxob5)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

ianjevans (approved):

Issue #11470, opened 2021-09-08
sql: cache hashed password and auth info during authentication


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/66919

From release notes:
> Added a `server.authentication_cache.enabled` [cluster setting](../v21.2/cluster-settings.html) that defaults to `true`. When enabled, this cache stores authentication-related data and will improve the latency of authentication attempts. Keeping the cache up to date adds additional overhead when using the [`CREATE`](../v21.2/create-role.html), [`ALTER`](../v21.2/alter-role.html), and [`DROP ROLE`](../v21.2/drop-role.html) commands. To minimize the overhead, any bulk `ROLE` operations should be run inside of a transaction. To make the cache more effective, any regularly-scheduled `ROLE` updates should be done all together, rather than occurring throughout the day at all times. [#66919][#66919]

ericharmeling:
The `server.authentication_cache.enabled` cluster setting is already included in external documentation.

The updates required here are just performance tips.

PR #12200, opened 2021-11-08
Added performance best practices for ROLE operations

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/11470.

This PR documents some performance best practices for `ROLE` operations, following the addition of hashed password caching for authentication.

(IMO, we don't need to document the specifics of hashed password caching, but LMK if you disagree.)

### DIFF
--- a/v21.2/authentication.md
+++ b/v21.2/authentication.md
@@ -69,7 +69,7 @@ CockroachDB offers the following methods for client authentication:
    Note that the client still needs the CA certificate to validate the nodes' certificates.
 
    {{site.data.alerts.callout_success}}
-   <span class="version-tag">New in v21.2</span>: For improved performance, CockroachDB securely caches password authentication information for users. To limit the authentication latency of users logging into a new session, we recommend that you run bulk `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)) inside of a transaction., and run any regularly-scheduled `ROLE` operations together, rather than at different times throughout the day.
+   <span class="version-tag">New in v21.2</span>: For improved performance, CockroachDB securely caches password authentication information for users. To limit the authentication latency of users logging into a new session, we recommend that you run bulk `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)) inside a transaction, and run any regularly-scheduled `ROLE` operations together, rather than at different times throughout the day.
    {{site.data.alerts.end}}
 
 - **Password authentication without TLS**
--- a/v21.2/authentication.md
+++ b/v21.2/authentication.md
@@ -68,6 +68,10 @@ CockroachDB offers the following methods for client authentication:
 
    Note that the client still needs the CA certificate to validate the nodes' certificates.
 
+   {{site.data.alerts.callout_success}}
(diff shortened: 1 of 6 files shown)


v21.2/authorization.md
```diff
@@ -145,6 +145,11 @@ We recommend the following best practices to set up access control for your clus
 - Use the `root` user only for database administration tasks such as creating and managing other [users](#sql-users), creating and managing [roles](#roles), and creating and managing databases. Do not use the `root` user for applications; instead, create users or roles with specific [privileges](#assign-privileges) based on your application’s access requirements.
 - Use the ["least privilege model"](https://en.wikipedia.org/wiki/Principle_of_least_privilege) to grant privileges to users and roles.
 
+We recommend the following best practices for optimal performance in `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)):
```
rafiss, line 148: I do think a bit of the "why" would be helpful. it's not really a performance concern with the ROLE operations themselves -- so we should say that doing it this way will have less of an effect on the latency of logging into a new session.

otherwise people may see this advice, think to themselves, "i don't have any performance issues with this command," and then ignore the advice.

rafiss (changes requested):

v21.2/authorization.md
ericharmeling: Yeah, you're totally right. I updated the PR with some more details, without going into too much detail. PTAL.

rafiss (approved):
this looks great!

ericharmeling:
> this looks great!

TFTR!

v21.2/authentication.md
```diff
@@ -68,6 +68,10 @@ CockroachDB offers the following methods for client authentication:
 
    Note that the client still needs the CA certificate to validate the nodes' certificates.
 
+   {{site.data.alerts.callout_success}}
+   <span class="version-tag">New in v21.2</span>: For improved performance, CockroachDB securely caches password authentication information for users. To limit the authentication latency of users logging into a new session, we recommend that you run bulk `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)) inside of a transaction., and run any regularly-scheduled `ROLE` operations together, rather than at different times throughout the day.
```
stbof, line 72:
```suggestion
   <span class="version-tag">New in v21.2</span>: For improved performance, CockroachDB securely caches password authentication information for users. To limit the authentication latency of users logging into a new session, we recommend that you run bulk `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)) inside a transaction, and run any regularly-scheduled `ROLE` operations together, rather than at different times throughout the day.
```

v21.2/authorization.md
```diff
@@ -145,6 +145,11 @@ We recommend the following best practices to set up access control for your clus
 - Use the `root` user only for database administration tasks such as creating and managing other [users](#sql-users), creating and managing [roles](#roles), and creating and managing databases. Do not use the `root` user for applications; instead, create users or roles with specific [privileges](#assign-privileges) based on your application’s access requirements.
 - Use the ["least privilege model"](https://en.wikipedia.org/wiki/Principle_of_least_privilege) to grant privileges to users and roles.
 
+ <span class="version-tag">New in v21.2</span>: For improved performance, CockroachDB securely caches [authentication information for users](authentication.html#client-authentication). To limit the authentication latency of users logging into a new session, we recommend the following best practices for `ROLE` operations ([`CREATE ROLE`](create-role.html), [`ALTER ROLE`](alter-role.html), [`DROP ROLE`](drop-role.html)):
+
+- Run bulk `ROLE` operations inside of a transaction.
```
stbof, line 150:
```suggestion
- Run bulk `ROLE` operations inside a transaction.
```

stbof (approved):
See suggestions.

v21.2/authentication.md
ericharmeling: Done.

v21.2/authorization.md
ericharmeling: Done.

ericharmeling (commented):
TFTRs!

PR #12203, opened 2021-11-08
v20.2.18 is the latest v20.2 version

taroface:
Made a tweak that ensures v20.2.18 loads as the "latest" v20.2 version in the side nav.


taroface:
Closing; need to update the branch first.

PR #12208, opened 2021-11-09
Clarify UPSERT vs INSERT ON CONFLICT performance

mgartner:
* Move performance note about UPSERT vs INSERT ON CONFLICT from
  "known limitations" to "insert" and "performance best practices".

* Add a sentence to clarify that performance of UPSERT and INSERT ON
  CONFLICT is the same when a table has secondary indexes.


stbof (approved):

mgartner:
Thanks for the review!

PR #12215, opened 2021-11-09
Docs for SET rollback in txn limitation

### LINKED ISSUE #12169: sql: SET does not properly rollback in a transaction
Known limitation per https://github.com/cockroachdb/cockroach/issues/69396

See also:

- #11958

kathancox:
Closes #12169 

Docs added for known limitation page 21.2 & `SET` page on `SET` not properly rolling back during a transaction.


otan (approved):
FWIW, this applies to all versions of CockroachDB past and present

rmloveland (approved):
LGTM!

do you mind opening a followup issue to add this to versions of docs <= 21.1 so we can do post v21.2 release?

something like this: https://github.com/cockroachdb/docs/issues/12232

no need to assign to yourself but please add `G-sql` label so someone in that team sees it

PR #12217, opened 2021-11-09
Add alert to Cloud API quickstart about it being experimental

ianjevans:
Do we need a link for a location to file bugs?

Do we need a link to request an invite?


marksoper:
It would be great to have a way to request an invite, yes. A link to file bugs would be nice but lower priority.
Ideally I'd want them to just email me in either case, but I don't know if there's a good way to do this.

marksoper (approved):
The text looks good. We can add the "email Mark" mechanism as a follow up if that makes sense.

PR #12229, opened 2021-11-10
Update 21.2 limitation about special characters in passwords

jseldess:


knz (approved):
thank you!

jseldess:
TFTR, @knz. I've also update the connection parameters page. Should these changes be applied to the 21.1 as well?

Issue #12172, opened 2021-11-03
sql: support expressions as ON CONFLICT targets


rmloveland:
Known limitation per https://github.com/cockroachdb/cockroach/issues/67893

See also:

- #11958

PR #12234, opened 2021-11-10
Cannot use expressions as ON CONFLICT targets

rmloveland:
Fixes cockroachdb/docs#12172


_includes/v21.2/sql/expressions-as-on-conflict-targets.md
```diff
@@ -0,0 +1,49 @@
+CockroachDB does not support expressions as `ON CONFLICT` targets. This means that unique {% if page.name == "expression-indexes.md" %} expression indexes {% else %} [expression indexes](expression-indexes.html) {% endif %} cannot be selected as arbiters for [`INSERT .. ON CONFLICT`](insert.html#on-conflict-clause) statements. For example:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+SET experimental_enable_expression_indexes=true;
+~~~
+
+~~~
+SET
```
mgartner, line 9: You can remove the SET command and SET output - as of 21.2 expression indexes are enabled by default.

mgartner (approved):
LGTM

_includes/v21.2/sql/expressions-as-on-conflict-targets.md
rmloveland: NOICE - removed

PR #12236, opened 2021-11-10
Add alert about completing upgrade before enabling admission control.

### LINKED ISSUE #12235: Recommend upgrade to 21.2, then turn on admission control
In the [admission control architecture docs](https://www.cockroachlabs.com/docs/v21.2/architecture/admission-control.html),

> We recommend enabling admission control on all layers if you decide to use admission control.

Can we also mention that we recommend upgrading to 21.2 first without enabling admission control and then enabling the feature after? This is to avoid any conflation of upgrade issues with admission control issues.

ianjevans:
Fixes #12235.

### DIFF
--- a/v21.2/architecture/admission-control.md
+++ b/v21.2/architecture/admission-control.md
@@ -14,6 +14,10 @@ Admission control is disabled by default. To enable admission control:
 
 We recommend enabling admission control on all layers if you decide to use admission control.
 
+{{site.data.alerts.important}}
+If you are upgrading to v21.2, first complete the upgrade with admission control disabled, then enable admission control after verifying the upgrade was successful.
+{{site.data.alerts.end}}
+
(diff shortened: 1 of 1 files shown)


v21.2/architecture/admission-control.md
```diff
@@ -14,6 +14,10 @@ Admission control is disabled by default. To enable admission control:
 
 We recommend enabling admission control on all layers if you decide to use admission control.
 
+{{site.data.alerts.important}}
+If you are upgrading to v21.2, first complete the upgrade with admission control disabled, then enable admission control after verifying the upgrade was successful.
```
vy-ton, line 18: `was successful and meets your expectations`

vy-ton (approved):

stbof (approved):

Issue #12170, opened 2021-11-03
opt: support locality optimized search on unique partitioned indexes on virtual columns


rmloveland:
Known limitation per https://github.com/cockroachdb/cockroach/issues/68129

See also:

- #11958

PR #10998, opened 2021-08-02
Add known issues to 21.1 for virtual computed columns.

ianjevans:
Fixes #10993 
Fixes #10997 

Adding known issues for virtual computed columns and multi-region tables in v21.1.

PR #12240, opened 2021-11-10
No locality optimized search on unique partitioned indexes on virtual computed columns

rmloveland:
Fixes cockroachdb/docs#12170


_includes/v21.2/sql/locality-optimized-search-virtual-computed-columns.md
```diff
@@ -0,0 +1,5 @@
+CockroachDB's optimizer cannot plan {% if page.name == "cost-based-optimizer.md" %} locality-optimized searches {% else %} [locality-optimized searches](cost-based-optimizer.html#locality-optimized-search-in-multi-region-clusters) {% endif %} for queries that use [partitioned unique indexes](partitioning.html#partition-using-secondary-index) on [virtual computed columns](computed-columns.html#virtual-computed-columns).
```
mgartner, line 1: We should also mention here (or somewhere else) that as of 21.2 (which adds expression indexes) "locality optimized searches will not be planned on partitioned unique expression indexes".

mgartner (approved):
FYI https://github.com/cockroachdb/docs/pull/10998 already added this known limitation for 21.1. The only difference in 21.2 is that expression indexes have been added, which have the same limitation (see my inline comment).

_includes/v21.2/sql/locality-optimized-search-virtual-computed-columns.md
rmloveland: > FYI #10998 already added this known limitation for 21.1. The only difference in 21.2 is that expression indexes have been added, which have the same limitation

Ah ok - thanks for the heads up, did not know

> We should also mention here (or somewhere else) that as of 21.2 (which adds expression indexes) "locality optimized searches will not be planned on partitioned unique expression indexes".

Sounds good - added below the main paragraph: "Locality-optimized searches will also not be planned for queries that use partitioned unique [expression indexes](expression-indexes.html)."

PR #12241, opened 2021-11-10
Fix database value in Sequelize Serverless

kernfeld-cockroach:
The first part of the value for "database" actually needs to be the cluster's tenant name (which is the cluster name plus a short ID) instead of just the cluster name. The user can easily get all of the relevant values from the "Connection parameters" tab.

### DIFF
--- a/v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
+++ b/v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
@@ -55,8 +55,13 @@ Open `app.js`, and edit the connection configuration parameters:
 
 <section class="filter-content" markdown="1" data-scope="cockroachcloud">
 
+- At the top of the file, uncomment the `const fs = require('fs');` line.
+
+    This line imports the `fs` Node module, which enables you to read in the CA cert that you downloaded from the {{ site.data.products.db }} Console.
 - Replace the values for `username`, `host`, `port`, and `database` with values from the **Connection parameters** tab of **Connection info**.
 - Replace the value for `password` with the password you created for your user.
+- Remove the `rejectUnauthorized` key-value pair.
+- Uncomment the `ca` key-value pair, and edit the `fs.readFileSync('certs/ca.crt').toString()` call to use the path to the `cc-ca.crt` file that you downloaded from the {{ site.data.products.db }} Console.
 
 </section>
 
--- a/v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
+++ b/v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
@@ -58,11 +58,8 @@ Open `app.js`, and edit the connection configuration parameters:
 - At the top of the file, uncomment the `const fs = require('fs');` line.
 
     This line imports the `fs` Node module, which enables you to read in the CA cert that you downloaded from the {{ site.data.products.db }} Console.
-- Replace the value for `username` with the user you created earlier.
+- Replace the values for `username`, `host`, `port`, and `database` with values from the "Connection parameters" tab of "Connection info."
 - Replace the value for `password` with the password you created for your user.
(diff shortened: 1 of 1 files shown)


v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
```diff
@@ -55,16 +55,8 @@ Open `app.js`, and edit the connection configuration parameters:
 
 <section class="filter-content" markdown="1" data-scope="cockroachcloud">
 
-- At the top of the file, uncomment the `const fs = require('fs');` line.
```
ericharmeling, line 58: Don't they still need to specify the root cert to authenticate?

v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
```diff
@@ -55,16 +55,8 @@ Open `app.js`, and edit the connection configuration parameters:
...
-
-    This line imports the `fs` Node module, which enables you to read in the CA cert that you downloaded from the {{ site.data.products.db }} Console.
-- Replace the value for `username` with the user you created earlier.
+- Replace the values for `username`, `host`, `port`, and `database` with values from the "Connection parameters" tab of "Connection info."
```
ericharmeling, line 58: For consistency with the rest of our Cloud docs:

s/`"Connection parameters" tab of "Connection info."`/`**Connection parameters** tab of the **Connection info** dialog.`

ericharmeling (commented):
Nice. Thanks!

I do have a couple comments/questions.

v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
kernfeld-cockroach: No, I don't think so

v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
kernfeld-cockroach: Thanks!

v21.1/build-a-nodejs-app-with-cockroachdb-sequelize.md
ericharmeling: Okay - it will run, but the connection will be insecure as long as we have `rejectUnauthorized: false`, right? 

With `rejectUnauthorized: false` (the default), you'll get the following error: 
```
SequelizeConnectionError: server requires encryption
error: server requires encryption
```

I think one of the main pain points in the cloud connection story is accessing the root cert from a local directory. If we want to update our recommendations for simple "quick-start" connections to just not use root certs at all, then we will probably need to update a lot more docs pages.

@ianjevans @gemma-shay

ianjevans: This config works:

~~~
var sequelize = new Sequelize({
  dialect: "postgres",
  username: "ian",
  password: "xxxxxxx",
  host: "free-tier.gcp-us-central1.cockroachlabs.cloud",
  port: 26257,
  database: "usable-weasel-4036.bank",
  dialectOptions: {
    ssl: true,
    native: true
  },
  logging: false,
});
~~~

Tried it after reading one of the answers to [this SO question](https://stackoverflow.com/questions/61153275/sequelizeconnectionerror-the-server-does-not-support-ssl-connections).

I tried a similar config using the regular connection string and the above `dialectOptions`, but I got an `SequelizeConnectionError: codeParamsRoutingFailed: missing cluster name in connection string` error, so Sequelize wasn't respecting the `options` query parameter.

kernfeld-cockroach: All right, I am reverting all cert-related stuff from this PR. Even were it a good idea, it should be a separate PR.

kernfeld-cockroach:
This is ready for another look.

awoods187:
Pretty cool to make a PR here! Nice work.

kernfeld-cockroach:
Oops, I didn't see that this was already approved! Will merge.

PR #12242, opened 2021-11-10
hardcode link to Operate CockroachDB on K8s in v20.2

taroface:
Replaced the liquid tag with a hardcoded version path to the Operate CockroachDB on Kubernetes doc.

cc @lnhsingh - This is unfortunately pointing to v21.1 from v20.2, for reasons that have to do with the Kubernetes Operator updates leading to changes in docs content + structure that became version-specific. I'm going to investigate the possibility of decoupling the K8s docs from CRDB versioning. In the meantime, this should be the only instance of the docs versions mismatching. Let me know if you think I should handle this differently.


kathancox (approved):

lnhsingh:
This LGTM. Thanks Ryan!

Issue #12171, opened 2021-11-03
sql: allow expression indexes to reference computed columns


rmloveland:
Known limitation per https://github.com/cockroachdb/cockroach/issues/67900

See also:

- #11958

PR #12243, opened 2021-11-10
Expression indexes can't reference computed cols

rmloveland:
Fixes cockroachdb/docs#12171


rmloveland:
(Actually this was already added to the expression indexes docs by @stbof  in f078f2f9f - we're just porting it to an include so it can live on both the 'Expression Indexes' and 'Known Limitations' pages)

mgartner (approved):
LGTM

PR #12245, opened 2021-11-10
Clarify cluster name, tenant id, and cluster id

gemma-shay:
Connection docs were not previously using cluster name, tenant IDs, and cluster IDs correctly. This PR updates the language temporarily,[ until tenant IDs get added to cluster names and simplify this process.](https://cockroachlabs.atlassian.net/browse/CC-5285)

Examples to illustrate correct meanings:
`cluster-name`=`funny-skunk`
`cluster-id`=`12a3bcde-4fa5-6789-1234-56bc7890d123`
(Serverless) `tenant-id`=`123`
(Dedicated) `short-id`=`6mc`

To download the CA cert, you need the cluster ID. To use the SQL connection string, you need the cluster name and tenant or short ID. Dedicated certs are saved using cluster names and Serverless certs are just saved as `root.crt`.

Closes https://cockroachlabs.atlassian.net/browse/CC-5363

### DIFF
--- a/_includes/cockroachcloud/sql-connection-string.md
+++ b/_includes/cockroachcloud/sql-connection-string.md
@@ -1,21 +1,21 @@
     <section class="filter-content" markdown="1" data-scope="mac">
     {% include_cached copy-clipboard.html %}
     ~~~ shell
-    cockroach sql --url 'postgresql://<user>@<cluster-name>-<short-id>.<region>.<host>:26257/<database>?sslmode=verify-full&sslrootcert='$HOME'/Library/CockroachCloud/certs/<cluster-name>-ca.crt'
+    cockroach sql --url 'postgresql://<user>@<cluster-name>-<short-id>.<region>.cockroachlabs.cloud:26257/<database>?sslmode=verify-full&sslrootcert='$HOME'/Library/CockroachCloud/certs/<cluster-name>-ca.crt'
     ~~~
     </section>
 
     <section class="filter-content" markdown="1" data-scope="linux">
     {% include_cached copy-clipboard.html %}
     ~~~ shell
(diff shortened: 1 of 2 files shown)


jaylim-crl (approved):
Thanks! This looks good to me. Note that we may want to change the URL that serverless clusters download the certs from in the near future (i.e. a cluster ID isn’t needed since the contents are the same). Right now, it’s causing confusion where users with multiple serverless clusters would think that downloading the cert on a second time will override their old root.crt file, which isn’t the case.

gemma-shay:
Thanks @jaylim-crl! Would that be part of [this update](https://cockroachlabs.atlassian.net/browse/CC-5285) or is there a separate ticket I can track?

jaylim-crl:
> Thanks @jaylim-crl! Would that be part of [this update](https://cockroachlabs.atlassian.net/browse/CC-5285) or is there a separate ticket I can track?

I don't think there's an exact ticket, but the closest would be CC-4287. I've tagged you on it.

_includes/cockroachcloud/sql-connection-string.md
```diff
@@ -17,4 +17,11 @@
     ~~~ shell
     cockroach sql --url "postgresql://<user>@<cluster-name>-<short-id>.<region>.<host>:26257/<database>?sslmode=verify-full&sslrootcert=$env:appdata\CockroachCloud\certs\$<cluster-name>-ca.crt"
     ~~~
-    </section>
\ No newline at end of file
+    </section>
+    
+    Where:
```
ianjevans, line 22: Shouldn't this define the host, region, and database variables here as well?

gemma-shay: looks like host is always `cockroachlabs.cloud`, so I'll update that and add the other two variables

ianjevans (approved):

Issue #11380, opened 2021-09-08
sql: copy hidden columns in CREATE TABLE LIKE


jseldess:
PR: https://github.com/cockroachdb/cockroach/pull/67799

From release notes:
> [`CREATE TABLE LIKE`](../v21.2/create-table.html) now copies hidden columns over. [#67799][#67799] {% comment %}doc{% endcomment %}

PR #12255, opened 2021-11-11
CREATE TABLE LIKE copies hidden columns

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/11380.

### DIFF
--- a/v21.2/create-table.md
+++ b/v21.2/create-table.md
@@ -114,7 +114,7 @@ Parameter | Description
 
 ## Create a table like an existing table
 
- CockroachDB supports the `CREATE TABLE LIKE` syntax for creating a new table based on the schema of an existing table.
+CockroachDB supports the `CREATE TABLE LIKE` syntax for creating a new table based on the schema of an existing table.
 
 The following options are supported:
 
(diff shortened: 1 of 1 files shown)


v21.2/create-table.md
```diff
@@ -127,9 +127,9 @@ The following options are supported:
 To exclude specifiers, use the `EXCLUDING` keyword. Excluding specifiers can be useful if you want to use `INCLUDING ALL`, and exclude just one or two specifiers. The last `INCLUDING`/`EXCLUDING` keyword for a given specifier takes priority.
 
 {{site.data.alerts.callout_info}}
-Column families, partitioning, interleavings, and foreign key constraints
-cannot be preserved from the old table and will have to be recreated
-manually in the new table if the user wishes.
+`CREATE TABLE LIKE` statements cannot copy [column families](column-families.html), [partitions](partitioning.html), and [foreign key constraints](foreign-key.html) from old tables. If you want these column qualifications in a new table, you must recreate them manually.
```
stbof, line 130:
```suggestion
`CREATE TABLE LIKE` statements cannot copy [column families](column-families.html), [partitions](partitioning.html), and [foreign key constraints](foreign-key.html) from other tables. If you want these column qualifications in a new table, you must recreate them manually.
```

v21.2/create-table.md
```diff
@@ -127,9 +127,9 @@ The following options are supported:
...
+
+`CREATE TABLE LIKE` copies all hidden columns (e.g., the hidden [`crdb_region`](set-locality.html#crdb_region) column in multi-region databases) from old tables to new tables.
```
stbof, line 132:
```suggestion
`CREATE TABLE LIKE` copies all hidden columns (e.g., the hidden [`crdb_region`](set-locality.html#crdb_region) column in multi-region databases) from the existing table to the new table.
```

stbof (approved):
I find the word `old` to be kind of strange. My suggested replacement is `existing`.

v21.2/create-table.md
ericharmeling: Done

v21.2/create-table.md
ericharmeling: Done

ericharmeling (commented):
TFTR!

ericharmeling:
> I find the word `old` to be kind of strange. My suggested replacement is `existing`.

I agree. I updated it here. Unfortunately, I think we use this kind of terminology in a lot of docs where we could use "existing" instead. I'll try to update it where I see it.

In general, using "old" could be perceived as non-inclusive (in addition to not being the most precise term), so I also added it to the ["Write about features and users in inclusive ways" section of the style guide](https://github.com/cockroachdb/docs/wiki/Style-Guide#write-about-features-and-users-in-inclusive-ways).

PR #12320, opened 2021-11-16
Use better links to Pebble docs

rmloveland:
We have a section on Pebble in the Architecture > Storage Layer docs which includes this info and more, so let's use that.


taroface:
LGTM

taroface (approved):

PR #12325, opened 2021-11-16
Example on `as_json` added to `SHOW BACKUP` docs & 21.1

### LINKED ISSUE #12278: Add `as_json` option to 21.1 SHOW BACKUP docs
Kathryn Hancox (kathancox) commented:

The `as_json` option has been added to the 21.2 docs, this has also been backported and can be added into the 21.1 docs.

kathancox:
Closes #12278 

Expanded documentation on the `as_json` option, also added this to v21.1 docs on the SHOW BACKUP page

### DIFF
--- a/v21.1/show-backup.md
+++ b/v21.1/show-backup.md
@@ -332,7 +332,7 @@ SELECT jsonb_pretty(manifest->'entryCounts') AS f FROM [SHOW BACKUP '/2021/11/15
   }
 ~~~
 
-To query for particular data, use the `jsonb_array_elements()` function to expand the desired elements from the JSON response. The following query returns the paths to each of the data files within the backup:
+To query for particular data, use the [`jsonb_array_elements()` function](functions-and-operators.html#jsonb-functions) to expand the desired elements from the JSON response. The following query returns the paths to each of the data files within the backup:
 
 {% include copy-clipboard.html %}
 ~~~ sql
--- a/v21.1/show-backup.md
+++ b/v21.1/show-backup.md
@@ -33,7 +33,8 @@ Parameter | Description
 Option       | Value | Description
 -------------+-------+-----------------------------------------------------
 `privileges` | N/A   |  List which users and roles had which privileges on each table in the backup. Displays original ownership of the backup.
-`encryption_passphrase`<a name="with-encryption-passphrase"></a> | [`STRING`](string.html) |  The passphrase used to [encrypt the files](take-and-restore-encrypted-backups.html) (`BACKUP` manifest and data files) that the `BACKUP` statement generates.
+`encryption_passphrase`<a name="with-encryption-passphrase"></a> | [`STRING`](string.html) |  The passphrase used to [encrypt the files](take-and-restore-encrypted-backups.html) that the `BACKUP` statement generates (the data files and its manifest, containing the backup's metadata).
+`as_json`   |  N/A  | [Display the backup's internal metadata](#show-a-backups-internal-metadata) as JSON in the response.
 
 ## Response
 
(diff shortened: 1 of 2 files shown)


v21.1/show-backup.md
```diff
@@ -34,6 +34,7 @@ Option       | Value | Description
 -------------+-------+-----------------------------------------------------
 `privileges` | N/A   |  List which users and roles had which privileges on each table in the backup. Displays original ownership of the backup.
 `encryption_passphrase`<a name="with-encryption-passphrase"></a> | [`STRING`](string.html) |  The passphrase used to [encrypt the files](take-and-restore-encrypted-backups.html) (`BACKUP` manifest and data files) that the `BACKUP` statement generates.
+`as_json`   |  N/A  | [Display the backup's manifest](#show-a-backups-manifest) as JSON in the response.
```
dt, line 37: I wonder if "manifest" is meaningful to the user or just us internally? should we say "...the backup's raw internal metadata" instead?

v21.1/show-backup.md
```diff
@@ -301,6 +302,87 @@ Or, use the `kms` option and the same KMS URI that was used to create the backup
 (20 rows)
 ~~~
 
+### Show a backup's manifest
+
+Use the `WITH as_json` option to output a backup's manifest as a JSON value:
```
dt, line 307: ditto above

No matter what we call it, we should explicitly state that this is for inspecting the _internal_ metadata, for example for debugging, and is not a stable public API, and its contents may change in the future, and that users should prefer to use a different `SHOW BACKUP` form if they can, or if they can't, file an issue describing the use-case that is not supported by the other (stable) SHOW BACKUP APIs

kathancox: Great points @dt, I've revised and drafted with "manifest" de-emphasized. I have retained it in places — though always clarifying what it contains, so it does have meaning to the user. My reasoning is that since the `as_json` option displays within a `manifest` column, referring to the metadata as within the manifest provides some context for the user on that front at least. Let me know what you think of these changes. 

I've also added a draft note to describe the usage as debugging. I'm wondering if this needs to be stronger, maybe a warning? @livlobo I assume we don't necessarily need to go as strong as only suggesting usage of this with the Support team.

dt: It is certainly _safe_ to use this without any sort of support guidance / supervision, since it is just reading, but it just is not something we're making any promises about what it'll report, and is not subject to the same external API stability guarantees as the rest of cockroach's SQL surface area including the other flavors of `SHOW BACKUP`. Its content *will* change from version to version, and should not be used for anything other than ad-hoc manual inspection.

kathancox: Thanks for the confirmation. OK, with that in mind, it seems me the note in place would suffice to provide enough recommendation to only use this for debugging or inspection. Other than that I have directed toward using other `SHOW BACKUP` options for more stable responses. Are you OK with the other changes @dt?

dt (approved):

v21.1/show-backup.md
```diff
@@ -301,6 +302,91 @@ Or, use the `kms` option and the same KMS URI that was used to create the backup
...
+For example, to return the JSON response as a [`string`](string.html) indented and with newlines use the [`jsonb_pretty()`](functions-and-operators.html#jsonb-functions) function:
+
+{% include copy-clipboard.html %}
+~~~ sql
+SELECT jsonb_pretty(f.manifest) FROM [SHOW BACKUP '/2021/11/15-150703.21' IN 's3://test/backup-test?AWS_ACCESS_KEY_ID=[placeholder]&AWS_SECRET_ACCESS_KEY=[placeholder]' WITH as_json] AS f;
+~~~
+
+~~~
+. . .
+"S3Config": {
```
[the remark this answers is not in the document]
kathancox: I get it and agree. My natural inclination is to lean toward showing "more" educationally in docs. _But_, in this case, if we're already providing a note saying this is subject to change, then this could potentially be misleading if not updated quickly enough after changes to the output (which might even be difficult to track). I'm going to go ahead and reduce the output here. Thank you for the feedback @dt

kathancox:
@livlobo I am pushing this through to docs review, but do let me know if you have further comments adjustments.

v21.1/show-backup.md
```diff
@@ -301,6 +302,57 @@ Or, use the `kms` option and the same KMS URI that was used to create the backup
...
+
+~~~ json
+  {
+      "dataSize": "458371",
+      "indexEntries": "1015",
+      "rows": "2565"
+  }
+~~~
+
+To query for particular data, use the `jsonb_array_elements()` function to expand the desired elements from the JSON response. The following query returns the paths to each of the data files within the backup:
```
rmloveland, line 335: it would be cool if this JSONB function also linked to functions-and-operators.html#jsonb-functions

rmloveland (approved):
LGTM!

Issue #12079, opened 2021-10-26
AWS Lambda Tutorial


ericharmeling:
Eric Harmeling (ericharmeling) commented:

Add a tutorial for AWS Lambda, similar to the [one for Google Cloud Run](https://www.cockroachlabs.com/docs/stable/deploy-app-gcr.html), but with AWS services.

PR #12330, opened 2021-11-17
AWS Lambda tutorial (psycopg2)

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/12079.

See https://github.com/cockroachlabs/examples-aws-lambda.


v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Step 2. Set up your cluster connection
+
+1. Once your cluster is created, the **Connection info** dialog displays. Follow the instructions to do the following:
+
+    1. Download the CockroackDB client.
+
+    1. Download the CA certificate.
+
+    1. Connect to the database with the [built-in SQL shell](cockroach-sql.html).
```
kernfeld-cockroach, line 35: This would require users to download CockroachDB, which I would ideally prefer to avoid.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+
+1. In the SQL shell, create the `bank` database that your Lambda function will use:
```
kernfeld-cockroach, line 37: I think it would be simpler to create this database from application code. Or just use `defaultdb` 🤷 .

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+1. In the `examples-aws-lambda/python` directory, create a new folder called `certs`.
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ cd examples-aws-lambda/python
+    ~~~
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ mkdir certs
```
kernfeld-cockroach, line 69: Can we just bundle the cert into the example app? @abhinavg6 I believe this is okay because the cert doesn't contain any private info.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+    docker run \
+    -e DATABASE_URL="postgresql://<username>:<password>@<serverless-host>:26257/bank?sslmode=verify-full&sslrootcert=certs/root.crt&options=--cluster%3D<cluster-name>" \
+    -p 9000:8080 init-crdb
+    ~~~
+
+    Note that this command sets the `DATABASE_URL` environment variable to the cluster connection string, where:
+    - `<username>` is the SQL user. By default, this is your {{ site.data.products.db }} account username.
+    - `<password>` is the password for the SQL user. The password will be shown only once in the **Connection info** dialog after creating the cluster.
+    - `<serverless-host>` is the hostname of the Serverless cluster.
+    - `<cluster-name>` is the short name of your cluster plus the tenant ID. For example, `funny-skunk-3`. The `<cluster-name>` is used to identify your tenant cluster on a multitenant host.
```
kernfeld-cockroach, line 113: I think we shouldn't mention the "tenant ID" in our docs. It's an implementation detail that's subject to change, and it introduces an extra concept that users shouldn't need to understand. Instead, users should be able to grab this info right from the connection parameters dialog.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+    ~~~
+
+    This command builds a Docker image based on the [AWS Lambda base image for Python](https://gallery.ecr.aws/lambda/python).
+
+1. (Optional) Run the Docker image locally:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    docker run \
+    -e DATABASE_URL="postgresql://<username>:<password>@<serverless-host>:26257/bank?sslmode=verify-full&sslrootcert=certs/root.crt&options=--cluster%3D<cluster-name>" \
```
kernfeld-cockroach, line 105: I'm worried that there's too much room for error in having users construct this DB URL manually.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,260 @@
...
+1. Configure the AWS CLI to authenticate with your AWS account:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ aws configure
+    ~~~
+
+    Follow the prompts to authenticate.
+
+    After you configure the CLI, you can use `aws configure` to list out the configuration variables. For example:
```
kernfeld-cockroach, line 150: I don't understand the benefit of having users run `aws configure list` and `aws sts get-caller-identity`.

kernfeld-cockroach:
Two big comments:

1. Unless there's a strong reason to use Docker, I think it would be better to go with the [Lambda Python runtime](https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html) instead because Docker/ECR introduces a lot of extra complexity.
2. I think it would be best to have a reviewer go through the entire tutorial before releasing it, to verify that we didn't miss anything. If it'll be hard to find someone within CockroachDB to do this, I bet we could get a manual testing service to do it. Let me know if I can be helpful in figuring that out!

kernfeld-cockroach (changes requested):

ericharmeling (commented):
Thanks for the thorough review, @kernfeld-cockroach.

See my response to your specific comments inline.

In response to your general comments:

> Unless there's a strong reason to use Docker, I think it would be better to go with the Lambda Python runtime instead because Docker/ECR introduces a lot of extra complexity.

I really think we should use the `aws` CLI (and not the AWS console) whenever possible. The CLI is more stable, less likely to decay, and much easier for users to copy and paste. 

According to the [AWS docs on `aws lambda create-function`](https://docs.aws.amazon.com/cli/latest/reference/lambda/create-function.html#description), "to create a function, you need a deployment package and an execution role". The deployment package can be a ZIP file or a Docker image. (I'm guessing that the runtime via the console either bypasses this requirement or creates a deployment package under-the-covers. Not sure.)

I tried both deployment package formats, and here are a couple reasons why I chose to use Docker over ZIP:

1. I think Docker makes things a little cleaner/more portable than just a ZIP file. The ZIP packaging/dependency management is not intuitive. The Docker workflow is pretty standard and was more recently implemented.
3. ["psycopg2 needs to be built and compiled with statically linked libraries for Linux."](https://stackoverflow.com/a/36608956) On macOS, you can just `pip install pscyopg2-binary` and it works. This just doesn't work with the ZIP deployment package. The most popular workaround is to use a third-party dependency (https://github.com/jkehler/awslambda-psycopg2), but you can also just use Docker and `psycopg2-binary` alone works. My guess is that the Dockerfile builds and compiles the necessary libraries *in the container*, whereas the ZIP file method requires you to build/compile the packages *from your machine's OS* and then copy the packages to the ZIP deployment package.

> I think it would be best to have a reviewer go through the entire tutorial before releasing it, to verify that we didn't miss anything. If it'll be hard to find someone within CockroachDB to do this, I bet we could get a manual testing service to do it. Let me know if I can be helpful in figuring that out!

I would usually ask the PM or tech lead of the product area to run through the tutorial. In the case of Google Cloud Run, we had the luxury of getting their PM to actually run through the tutorial. In the absence of DX engineers, I would ask you to quickly run through it. If you don't have the bandwidth, I can ask another writer.

Unfortunately, I ran into a strange permissions bug with AWS and had to use my personal account to test this sample. https://cockroachlabs.atlassian.net/browse/DEVINF-268

v21.2/deploy-lambda-function.md
ericharmeling: See also https://cockroachlabs.atlassian.net/browse/DOC-1856

awoods187:
LGTM with addressing Paul's comments!

ericharmeling:
@kernfeld-cockroach 

Following our discussion offline, I updated the sample code (see https://github.com/cockroachlabs/examples-aws-lambda/pull/1) and the tutorial to use a simplified deployment procedure. PTAL and let me know if you have any questions/comments.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ python3 -m pip install --only-binary :all: --platform manylinux1_x86_64  --target ./my-package -r requirements.txt
+    ~~~
+
+    {{site.data.alerts.callout_info}}    
+    To run on Amazon Linux distributions, `pscyopg2` dependencies must be compiled for Linux.
+    {{site.data.alerts.end}}
+
+1. Compress the project files to a ZIP file for deployment:
```
ianjevans, line 73: A `deployment-package.zip` file is checked into the code repo in cockroachlabs/examples-aws-lambda#1. This step creates a different ZIP file. Are both necessary?

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+The function's code is available under the `examples-aws-lambda/python` directory:
+
+~~~ shell
+.
+├── README.md
+├── deployment-package.zip  ## Lambda deployment package
+├── init_db.py              ## Lambda function source code
+├── package                 ## Psycopg dependencies
+├── requirements.txt        ## List of Python requirements
+└── root.crt                ## CA cert
```
ianjevans, line 53: The goal is eventually to remove the hard-dependency of `root.crt` in connecting to Serverless clusters, fwiw.

ericharmeling: That day cannot come soon enough.

v21.2/deploy-lambda-function.md
ericharmeling: For the purposes of completing the tutorial, no. Two deployment packages are not necessary. Because of that, I've marked this step as optional in two places. 

Something to note is that it's likely that users will end up creating their own functions (or modifying the sample function we provide). If they want to deploy their own code, they'll need to know how to package the code with the right dependencies. So I think it's important to include instructions on how to manually package the ZIP file.

ericharmeling (commented):
TFTR @ianjevans !

See my comments inline.

v21.2/deploy-lambda-function.md
gemma-shay: We added it in because a lot of users saw `<cluster-name>` and thought that just meant the short name. There was a long discussion [here](https://cockroachlabs.atlassian.net/browse/CC-5285) about it. I'm fine with removing tenant ID from docs, just noting that this distinction is definitely tripping people up.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+    zip -g my-deployment-package.zip init_db.py root.crt
+    ~~~
+
+## Step 4. Configure AWS
+
+1. Configure the AWS CLI to authenticate with your AWS account:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ aws configure
```
kernfeld-cockroach, line 101: Nit: the AWS user you use for this will need particular permissions. Beats me which ones.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+## Step 5. Deploy the function to AWS Lambda
+
+1. In the deployment package directory, use the AWS CLI to create a Lambda function:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ aws lambda create-function \
+        --function-name init-crdb \
+        --region us-east-1  \
+        --zip-file fileb://my-deployment-package.zip \
```
kernfeld-cockroach, line 129: Default to `deployment-package.zip`?

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+
+    The Lambda function needs this role to run.
+
+## Step 5. Deploy the function to AWS Lambda
+
+1. In the deployment package directory, use the AWS CLI to create a Lambda function:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ aws lambda create-function \
```
kernfeld-cockroach, line 126: Is there a sane way to make this happen without leaking the DB URL into the shell history? I can't really think of one.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+        --environment Variables={DATABASE_URL='<connection-string>'}
+    ~~~
+
+    Where `<account-id>` is your AWS account ID, and `<connection-string>` is the [connection string to the CockroachDB cluster](#connection-string).
+
+1. Invoke the function:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ aws lambda invoke --function-name init-crdb out --log-type Tail \
```
kernfeld-cockroach, line 142: This will only work if a user has their region set to `us-east-1`. I'd recommend including `--region us-east-1` in this command.

kernfeld-cockroach (approved):
Overall I'm quite jazzed! Going through the whole thing only took like 40 minutes. It might be nice to get another person to try the tutorial, because I did hit a couple bumps that required improvisation to overcome.

ericharmeling (commented):
Thanks for iterating on this, @kernfeld-cockroach 

I think I got to all of your questions, PTAL. 

I vote that we just merge this so we can have a version out there. Then we can add updates if needed later on. LMK if you disagree.

v21.2/deploy-lambda-function.md
```diff
@@ -0,0 +1,164 @@
...
+.
+├── README.md
+├── deployment-package.zip  ## Lambda deployment package
+├── init_db.py              ## Lambda function source code
+├── package                 ## Psycopg dependencies
+├── requirements.txt        ## List of Python requirements
+└── root.crt                ## CA cert
+~~~
+
+## Step 3. (Optional) Create the deployment package
```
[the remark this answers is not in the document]
kernfeld-cockroach: Good reasoning, I didn't think about create vs. update.

v21.2/deploy-lambda-function.md
kernfeld-cockroach: If AWS recommends an admin account then I think we should parrot them since it's basically guaranteed to avoid permissions problems.

kernfeld-cockroach:
LGTM, nice work!

v21.2/deploy-lambda-function.md
ericharmeling: Done.

PR #12331, opened 2021-11-17
[DOC-779] Clarify that show grants defaults to the current database

jseldess:


jseldess:
Received LGTM from @pxlogan in another channel.

pxlogan:
LGTM Thank you

PR #12350, opened 2021-11-18
Refactor Cloud API docs with warnings.

ianjevans:
Fixes CC-5426.

This PR:

- Breaks out the service account and API key creation docs into the main managing accounts doc.
- Adds the Cloud API reference and quickstart to the sidebar navigation.
- Adds a general purpose warning that can be used across the docs for experimental features.

### DIFF
--- a/_includes/common/experimental-warning.md
+++ b/_includes/common/experimental-warning.md
@@ -1,3 +1,3 @@
 {{site.data.alerts.callout_danger}}
-This feature is considered experimental, and it may not be available to all users. The functionality is subject to change, and there may be bugs.
+**This is an experimental feature**. The interface and output are subject to change.
 {{site.data.alerts.end}}
--- a/_includes/common/experimental-warning.md
+++ b/_includes/common/experimental-warning.md
@@ -1,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
+This feature is considered experimental, and it may not be available to all users. The functionality is subject to change, and there may be bugs.
+{{site.data.alerts.end}}
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -6,94 +6,15 @@ toc: true
 
 The Cloud API is a [REST interface](https://en.wikipedia.org/wiki/Representational_state_transfer) that allows users programmatic access to manage the lifecycle of clusters within their organization.
 
-{{site.data.alerts.callout_danger}}
-This page describes an early access, experimental version of the Cloud API. The interface and output are subject to change, and there may be bugs.
-{{site.data.alerts.end}}
+{% include_cached common/experimental-warning.md %}
 
 ## API reference
 
 Refer to the [full API reference documentation](../api/cloud/v1.html) for detailed descriptions of the API endpoints and options.
 
-{%comment %} START of temporary content while API is still behind a feature-flag{% endcomment %}
-## Service accounts
-
-Service accounts are used by applications accessing the API to manage {{ site.data.products.db }} clusters within the organization. Service accounts are not for human users.
-
-To create a service account:
-
-1. Click **Create Service Account**.
-1. In the **Create service account** dialog:
-    1. Enter the **Account name**.
-    1. (Optional) Enter a **Description** of the service account.
-    1. Set the **Permissions** of the service account.
-
-        Granting `ADMIN` permissions allows the service account full authorization for the organization, where the service account can create, modify, and delete clusters.
-
-        The `CREATE` permission allows the service account to create new clusters within the organization.
-
-        The `DELETE` permission allows the service account to delete clusters within the organization.
-
-        The `EDIT` permission allows the service account to modify clusters within the organization.
-
-        The `READ` permission allows the service account to get details about clusters within the organization.
-
-    1. Click **Create**.
-
-1. [Create an API key](#create-api-keys) for the newly created service account, or click **Skip** to go back to the Service Accounts table.
-
-### Modify a service account
-
-To modify the name, description, or permissions of a service account:
-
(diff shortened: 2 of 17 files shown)


_includes/common/experimental-warning.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
```
jseldess, line 1: We already have the concept of "experimental feature" and have an include that we use in some places, although it's versioned and shouldn't be: https://github.com/cockroachdb/docs/blob/master/_includes/v21.2/misc/experimental-warning.md. We also have a page that lists experimental features, though it's not exposed in the sidenav for some reason: https://www.cockroachlabs.com/docs/v21.2/experimental-features.html.

All that said, if experimental is the right categorization (and it does sound right, based on the similarities between your include and the old one), how do we combine experimental with "restricted", which is a cloud-only phenomenon, I think? cc @marksoper

marksoper: Yes, good point @jseldess - in this case it is experimental but has the additional property of being restricted.

cockroachcloud/cloud-api.md
```diff
@@ -6,94 +6,15 @@ toc: true
 
 The Cloud API is a [REST interface](https://en.wikipedia.org/wiki/Representational_state_transfer) that allows users programmatic access to manage the lifecycle of clusters within their organization.
 
-{{site.data.alerts.callout_danger}}
-This page describes an early access, experimental version of the Cloud API. The interface and output is subject to change, and there may be bugs.
```
marksoper, line 10: **Suggested wording:** 

This page describes the private Early Access version of the Cloud API. It's for **experimental use only**. The interface and output are subject to breaking changes. We would be grateful for your feedback - please email Mark Soper [contact info](https://github.com/marksoper) to request access.

_includes/common/experimental-warning.md
ianjevans: So it sounds like we should have a separate, Cloud-specific warning for private/restricted experimental features, as there aren't any cases where CRDB would have a private feature.

zaneteh-crl: +1 on having a separate message for this - I think we need to differentiate the two. 

The other "experimental features" message sounds more like the actual A/B experiments that run and our users could be in the experimentation group to get them. 

For the API feature, I like the terminology of private/restricted as these are opt-in and selected orgs that we chose for them.

cockroachcloud/cloud-api.md
zaneteh-crl: this looks good. @marksoper I would suggest maybe having an external facing email alias created for this with more team members on it. This way we can give more visibility into folks reaching out and you can still be the owner of the email group.

jseldess: Like the email alias idea. Also, I think this include needs to be crafted so it's not about an entire page. It could get included for a section of a page, for example.

_includes/common/experimental-warning.md
ianjevans: Ok, I created a separate Cloud-specific warning.

cockroachcloud/cloud-api.md
ianjevans: Should this PR wait until we have the alias ready? I will update the new Cloud-specific private/experimental warning with the contact info.

marksoper: I agree, an email group alias makes more sense.

ianjevans:
Just resurfacing this to see if I should wait for an email alias or get this merged. @marksoper

cockroachcloud/console-access-management.md
```diff
@@ -45,6 +45,84 @@ A Console Admin is an all-access role. A Console Admin can perform the following
...
+{% include_cached cockroachcloud/experimental-warning.md %}
+
+Each service account can have one or more API keys. API keys are used to authenticate and authorize service accounts when using the API. All API keys created by the account are listed under **API Access**.
+
+We recommend creating service accounts with the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege), and giving each application that accesses the API its own API key. This allows fine-grained permission and API key access to the cluster.
+
+#### Create API keys
+
+To create an API key:
+
```
gemma-shay, line 96: Add step to get to the Service Account Details page first here, same for Delete/Edit an API key

cockroachcloud/cloud-api.md
jseldess: @marksoper, can we quickly set up an email alias so we can get this published?

In the meantime, @ianjevans and @marksoper, can you work out how to extend the warning to guide readers on how to use that email alias?

gemma-shay (commented):
Couldn't figure out how to comment on the top of the **Console Access Management** page, but the second mention of "Access page" in the first paragraph needs to be bolded. Also, maybe overkill but I think the instructions under **Adminstrative tasks** could specify that you need to be on the **People** tab now that there are two tabs.

ianjevans:
> Also, maybe overkill but I think the instructions under Adminstrative tasks could specify that you need to be on the People tab now that there are two tabs.

I'll leave this for now, as it would affect all users, but definitely should be fixed once these tabs aren't under feature flags.

cockroachcloud/cloud-api.md
ianjevans: I had helpdesk create a new email group, earlyaccess@cockroachlabs.com. @marksoper is the current member, and he can add others in groups.google.com.

gemma-shay (approved):
LGTM.

Unrelated to the refactor, but because `{clusterID}` has been a source of confusion, I would suggest clarifying its format in the API doc so it doesn't get mixed up with cluster name + tenant ID.

jseldess (approved):
LGTM

PR #12352, opened 2021-11-18
Tech advisory notes: Backups fail during upgrade process a72839

kathancox:
Technical advisory notes relating to [#72839](https://github.com/cockroachdb/cockroach/issues/72839).

Backups fail during upgrade process v21.2 / v21.1

### DIFF
--- a/_data/advisories.yml
+++ b/_data/advisories.yml
@@ -1,3 +1,7 @@
+- advisory: 72839
+  summary: Backups fail during upgrade process
+  versions: 21.1.x to 21.2.0
+  date: November 18, 2021
 - advisory: 71553
   summary: SQL statements that used secondary unique indexes that were created as a result of an <code>ALTER PRIMARY KEY</code> statement can return incorrect results.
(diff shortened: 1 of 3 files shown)


rmloveland (commented):
This LGTM but I guess it's expected that this is just a floating page with no mention from the Table of Contents etc?  Never looked at one before so maybe this is expected

shermanCRL:
> This LGTM but I guess it's expected that this is just a floating page with no mention from the Table of Contents etc? Never looked at one before so maybe this is expected

Looks like it’s correct here: https://deploy-preview-12352--cockroachdb-docs.netlify.app/docs/advisories/index.html

shermanCRL:
@kathancox let’s consider calling out this technical advisory on the v21.2.0 release notes:

https://www.cockroachlabs.com/docs/releases/v21.2.0

nickelnickel (approved):
LGTM

ericharmeling (approved):
LGTM! Great work!

PR #12360, opened 2021-11-19
Update to version on table to make clearer

kathancox:
Quick fix to the versions in the advisories table to make it clearer that v21.2.0 is the affected version


shermanCRL:
LGTM

PR #12443, opened 2021-11-23
DOC-1859: Highlight sidenav for newer TAs

jseldess:
"Technical Advisories" in the sidenav isn’t highlighted when viewing some more recent TAs. This PR fixes that.


stbof:
@jseldess  Should something be added to https://cockroachlabs.atlassian.net/wiki/spaces/TS/pages/1874134401/Technical+Advisory+Process to prevent this bug? I worked on 2 advisories recently and wasn't aware that I had to update this file.

jseldess:
Good callout, @stbof. I added a step there.

PR #12459, opened 2021-11-24
Wrong column name in node_statement_statistics

wolftsao:
Column "FullTableScan" in crdb_internal.node_statement_statistics should be "full_scan".
And the data type is BOOL, so remove quotation and change value to true


stbof:
Thank you for this fix. I'm closing this because I incorporated your fix in https://github.com/cockroachdb/docs/pull/12462  in which the referenced page is being merged into another page.

PR #12460, opened 2021-11-24
DOC-1875: Corrected typo on restore cc page

kathancox:
Fixes DOC-1875

The CockroachCloud Restore data page currently notes that the **Restore to** field will create a new table, but it's actually a database, which makes sense from the text above the note.


livlobo (approved):
Thanks! 

I still have to re-read it to understand what they are asking me to drop/rename.

Could we add something like -- if you enter the name of an existing database, the table will be restored into that existing DB. If not, we will create a new database and restore it in the new databse?

rmloveland (approved):
LGTM

PR #12462, opened 2021-11-24
DOC-1874 Simplify performance recipes navigation

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-1874


taroface (changes requested):
I have a few suggestions and a rambling question about the reference to "slow queries" at the top.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12462#-:-MpIdAXiCHpT7tOirUVm:bnd3k8c)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans and @stbof)

___
*[v21.2/performance-recipes.md, line 11 at r3](https://reviewable.io/reviews/cockroachdb/docs/12462#-MpIZhE-C7_wE5rTgGwp:-MpIZhE003OG3cULGowa:b-cutui8) ([raw file](https://github.com/cockroachdb/docs/blob/5242810fcdae1d064ddadcf0e26fb6a8747c5d56/v21.2/performance-recipes.md#L11)):*
> ```Markdown
> 
> If you aren't sure whether SQL query performance needs to be improved, see [Identify slow queries](query-behavior-troubleshooting.html#identify-slow-statements).
> ```

I am somewhat confused by this statement, since at least the "full table scans" and "slow writes" recipes deal with slow/suboptimal queries. Is there more precise language we can use here, along the lines of "To determine whether slow queries are affecting cluster performance, see..."?

That said, I'm not clear whether any slow queries found could then be correlated with any of the recipes here. Are "slow queries" categorically distinct from all the below?
___
*[v21.2/performance-recipes.md, line 31 at r3](https://reviewable.io/reviews/cockroachdb/docs/12462#-MpIRmAc7IqvFQVMB5UR:-MpIRmAc7IqvFQVMB5US:b84ogs2) ([raw file](https://github.com/cockroachdb/docs/blob/5242810fcdae1d064ddadcf0e26fb6a8747c5d56/v21.2/performance-recipes.md#L31)):*
> ```Markdown
>     </tr>
>     <td><ul>
>       <li>Your application is experiencing degraded performance with the following serialization errors:
> ```

Terminology nit: We also list these errors as "transaction retry errors" in the Transaction Retry Error Reference: https://www.cockroachlabs.com/docs/v21.2/transaction-retry-error-reference.html

Unless "serialization errors" in a term in general usage, perhaps we should call these "transaction retry errors" for consistency.
___
*[v21.2/performance-recipes.md, line 82 at r3](https://reviewable.io/reviews/cockroachdb/docs/12462#-MpIUkDOB6lBVH52ssgQ:-MpIUkDOB6lBVH52ssgR:b-de9627) ([raw file](https://github.com/cockroachdb/docs/blob/5242810fcdae1d064ddadcf0e26fb6a8747c5d56/v21.2/performance-recipes.md#L82)):*
> ```Markdown
> #### Indicators that your application is experiencing contention
> 
> * Your application is experiencing degraded performance with serialization errors like `SQLSTATE: 40001`, `RETRY_WRITE_TOO_OLD`, and `RETRY_SERIALIZABLE`.
> ```

Same comment as above re: "transaction retry errors".

Also, the Transaction Retry Error Reference includes more detailed resolution steps for the two `RETRY` errors. Perhaps we should add a link to this reference under "Fix contention problems" below.
___
*[v21.2/performance-recipes.md, line 115 at r3](https://reviewable.io/reviews/cockroachdb/docs/12462#-MpIVZl51csj-Allw5TW:-MpIVZl51csj-Allw5TX:baz29he) ([raw file](https://github.com/cockroachdb/docs/blob/5242810fcdae1d064ddadcf0e26fb6a8747c5d56/v21.2/performance-recipes.md#L115)):*
> ```Markdown
> * The [Full Table/Index Scans graph](ui-sql-dashboard.html#full-table-index-scans) in the DB Console is showing spikes over time.
> 
> #### Fix full table scans in statements
> ```

The SQL Performance Best Practices page has a section on table scans. Probably worth adding a link and summary here.

https://www.cockroachlabs.com/docs/v21.2/performance-best-practices-overview.html#table-scans-best-practices

stbof:
@ianjevans do you know the history behind the opening intro about "slow queries"?

ianjevans:
> @ianjevans do you know the history behind the opening intro about "slow queries"?

It was part of the intro to an outdated performance tuning tutorial that was removed in favor of the recipes. 

>I am somewhat confused by this statement, since at least the "full table scans" and "slow writes" recipes deal with slow/suboptimal queries. Is there more precise language we can use here, along the lines of "To determine whether slow queries are affecting cluster performance, see..."?
>
>That said, I'm not clear whether any slow queries found could then be correlated with any of the recipes here. Are "slow queries" categorically distinct from all the below?

It's a fair point, but the current recipes aren't intended to be exhaustive. The idea behind the recipes was to guide users who aren't sure what is going wrong to the more detailed docs that would fix their problems, and provide a structure for adding additional recipes over time.

E.g. we wanted to capture all the different ways that a user might see the signals of workload contention and then point them to the cause and potential fixes. We wanted more docs to help users who were seeing some behavior or indicator that looked bad but they didn't know what was causing it.

taroface:
> It's a fair point, but the current recipes aren't intended to be exhaustive. The idea behind the recipes was to guide users who aren't sure what is going wrong to the more detailed docs that would fix their problems, and provide a structure for adding additional recipes over time.
>
> E.g. we wanted to capture all the different ways that a user might see the signals of workload contention and then point them to the cause and potential fixes. We wanted more docs to help users who were seeing some behavior or indicator that looked bad but they didn't know what was causing it.

Understood. Perhaps I'm reading into this too deeply, but this callout at the top stuck out to me because it implies that one might either read the performance tuning recipes or investigate the slow queries instead. I'm just wondering if the callout can be phrased in a way that better contextualizes the point being mentioned at the top, or if it would be better placed together with one of the recipes or in a "See also" at the bottom.

taroface (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12462#-:-MphfeZ-CScuXU3xjI3n:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ianjevans and @stbof)

ianjevans (approved):

PR #12494, opened 2021-12-01
Disambiguate different types of contention.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-1782


v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
 
 However, because `AS OF SYSTEM TIME` returns historical data, your reads might be stale.
 
-## Understanding and avoiding transaction contention
+## Contention
 
-Transaction contention occurs when the following three conditions are met:
+There are two types of contention:
```
kevin-v-ngo, line 328: See above comment.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
-- There are multiple concurrent transactions or statements (sent by multiple clients connected simultaneously to a single CockroachDB cluster).
-- They operate on the same data, specifically over table rows with the same index key values (either on [primary keys](primary-key.html) or secondary [indexes](indexes.html)) or using index key values that are close to each other, and thus place the indexed data on the same [data ranges](architecture/overview.html).
-- At least one of the transactions modify the data.
+- Transactions that operate on the same range but _different index key values_ will be limited by the overall hardware capacity of a single node (the range lease holder). This is referred to as _resource contention_.
```
kevin-v-ngo, line 330: This would be under 'hot spots'.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
-A set of transactions that all contend on the same keys will be limited in performance to the maximum processing speed of a single node (limited horizontal scalability). Non-contended transactions are not affected in this way.
+- Transactions that operate on the _same index key values_ (specifically, that operate on the same [column family](column-families.html) for a given index key) will be more strictly serialized to obey transaction isolation semantics. This is referred to as _transaction contention_ or _lock contention_.
 
-There are two levels of contention:
+    Transaction contention can also increase the rate of transaction restarts, and thus make the proper implementation of [client-side transaction retries](transactions.html#client-side-intervention) more critical.
 
-- Transactions that operate on the same range but different index key values will be limited by the overall hardware capacity of a single node (the range lease holder).
+### Resource contention
```
kevin-v-ngo, line 336: See above comment for this section.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
-{% include {{ page.version.version }}/performance/statement-contention.md %}
+### Transaction contention
 
-### Avoid contention
+Transaction contention occurs when the following three conditions are met:
 
-To avoid contention, you can apply multiple strategies:
+- There are multiple concurrent transactions or statements (sent by multiple clients connected simultaneously to a single CockroachDB cluster).
+- They operate on the same data, specifically over table rows with the _same index key values_ (either on [primary keys](primary-key.html) or secondary [indexes](indexes.html)) or using index key values that are close to each other, and thus place the indexed data on the same [data ranges](architecture/overview.html).
```
kevin-v-ngo, line 349: > or using index key values that are close to each other, and thus place the indexed data on the same [data ranges]

This seems like it's describing hot spots. @ajwerner can you confirm?

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
+- At least one of the transactions modify the data.
 
-- Use index key values with a more random distribution of values, so that transactions over different rows are more likely to operate on separate data ranges. See the [SQL FAQs](sql-faqs.html) on row IDs for suggestions.
+A set of transactions that all contend on the same keys will be limited in performance to the maximum processing speed of a single node (limited horizontal scalability). Non-contended transactions are not affected in this way.
```
kevin-v-ngo, line 352: This seems to be describing hot spots. Contending on the same keys would mainly cause transactions to wait (or restart) - it's less of a resource issue. @ajwerner can you confirm?

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
 - Increase [normalization](https://en.wikipedia.org/wiki/Database_normalization) of the data to place parts of the same records that are modified by different transactions in different tables. Fully normalized data allows separate transactions to modify related underlying data without causing contention. Increasing normalization has drawbacks as well as benefits, however. Denormalizing data can increase performance for certain queries by creating multiple copies of often-referenced data in separate ranges, so read-heavy workloads may have higher performance. However, denormalization adds complexity the data model, increases the chance of data inconsistency, increases data redundancy, and can have poor performance when running write-heavy workloads.
 
 - If the application strictly requires operating on very few different index key values, consider using [`ALTER ... SPLIT AT`](split-at.html) so that each index key value can be served by a separate group of nodes in the cluster.
```
kevin-v-ngo, line 375: This resolves hot spots - not transaction contention.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
 - When replacing values in a row, use [`UPSERT`](upsert.html) and specify values for all columns in the inserted rows. This will usually have the best performance under contention, compared to combinations of [`SELECT`](select-clause.html), [`INSERT`](insert.html), and [`UPDATE`](update.html).
 
 - Increase [normalization](https://en.wikipedia.org/wiki/Database_normalization) of the data to place parts of the same records that are modified by different transactions in different tables. Fully normalized data allows separate transactions to modify related underlying data without causing contention. Increasing normalization has drawbacks as well as benefits, however. Denormalizing data can increase performance for certain queries by creating multiple copies of often-referenced data in separate ranges, so read-heavy workloads may have higher performance. However, denormalization adds complexity the data model, increases the chance of data inconsistency, increases data redundancy, and can have poor performance when running write-heavy workloads.
```
kevin-v-ngo, line 373: This reduces hot spots since normalizing the data model will effectively spread data across more ranges. This is not necessarily two transactions conflicting with one another on the same key/row. @ajwerner can you confirm?

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
 - If the application strictly requires operating on very few different index key values, consider using [`ALTER ... SPLIT AT`](split-at.html) so that each index key value can be served by a separate group of nodes in the cluster.
 
+### Maximize performance in the presence of contention
+
 It is always best to avoid contention as much as possible via the design of the schema and application. However, sometimes contention is unavoidable. To maximize performance in the presence of contention, you'll need to maximize the performance of a single range. To achieve this, you can apply multiple strategies:
 
 - Minimize the network distance between the replicas of a range, possibly using zone configs and partitioning.
```
kevin-v-ngo, line 381: >- If the contending transactions operate on different keys within the same range, add more CPU power (more cores) per node. However, if the transactions all operate on the same key, this may not provide an improvement .

This resolves hot spots - not transaction contention.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,49 +323,59 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
 - If the application strictly requires operating on very few different index key values, consider using [`ALTER ... SPLIT AT`](split-at.html) so that each index key value can be served by a separate group of nodes in the cluster.
 
+### Maximize performance in the presence of contention
+
 It is always best to avoid contention as much as possible via the design of the schema and application. However, sometimes contention is unavoidable. To maximize performance in the presence of contention, you'll need to maximize the performance of a single range. To achieve this, you can apply multiple strategies:
```
kevin-v-ngo, line 379: The first two bullets here are generic resolutions to both hot spots and transaction contention. 

For contention, by minimizing network latency and faster storage devices, transactions can complete quicker reducing the wait time. For hot spots, transactions complete quicker reducing the time the node(s) was 'hot'.

Not sure how to structure this since 'hot spots' and 'contention' would be two different sections. Perhaps this belongs in a new section 'Improve transaction performance by sizing and configuring the cluster' ?

kevin-v-ngo (changes requested):
Thanks for the quick turnaround on this! My high-level feedback is to create a new section call 'Hot spots' and move the content we currently have as 'Resource contention' under that section. 

This would make it clear that 'contention' always refers to 'transaction contention'. Here's a good writeup from the field that highlights contention and hot spots as separate problems: https://github.com/cockroachlabs/cockroachdb-runbook-template/tree/main/most-common-problems#5-hot-spots.

It would be good to ask an engineer to review this as well. @ajwerner, you mind taking a look at this section?

ajwerner (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12494)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajwerner, @kevin-v-ngo, and @stbof)

___
*[v21.2/performance-best-practices-overview.md, line 349 at r2](https://reviewable.io/reviews/cockroachdb/docs/12494#-MpsFV6O5IrF4PGIFFCN-r2-349:-MqGakOC1u8-KRo0l9Sp:b-wjhtvt) ([raw file](https://github.com/cockroachdb/docs/blob/736a2af27dc671b4af2f95b3bb8abdd7845fd26a/v21.2/performance-best-practices-overview.md#L349)):*
<details><summary><i>Previously, kevin-v-ngo wrote…</i></summary><blockquote>

> or using index key values that are close to each other, and thus place the indexed data on the same [data ranges]

This seems like it's describing hot spots. @ajwerner can you confirm?

</blockquote></details>

Yeah, close to each other is explicitly not transaction contention. 
___
*[v21.2/performance-best-practices-overview.md, line 330 at r3](https://reviewable.io/reviews/cockroachdb/docs/12494#-MqGa5Qu4_7-kKI5Hpoj:-MqGa5Qu4_7-kKI5Hpok:b-3ith1) ([raw file](https://github.com/cockroachdb/docs/blob/714402406c535b3d46dac78e33d401f27ef7f101/v21.2/performance-best-practices-overview.md#L330)):*
> ```
> Hot spots occur when 
> ```
I'd say `can occur`. In particular, that's a reasonable source of write hot spots. They can occur also for other weird write patterns. Sequential is definitely the common one. However, worth putting `can occur` so you don't need to explain like:

```
.0
.1
.01
.001
.0001
.0011
.0111
.00101
...
```

Read hot spots can occur for other reasons. Consider lots of scans of an empty table or of a single key. Overlapping reads do not have contention. Read hot spots can be tough to deal with. The best solution is to use global tables and drive up the number of replicas.

stbof:
@kevin-v-ngo Do we want to call contention transaction contention or database contention. In our recent blog post we are calling it the latter. https://www.cockroachlabs.com/blog/what-is-database-contention/

kevin-v-ngo (approved):
Hi Stephanie, thanks for making the changes. This looks good!

kevin-v-ngo:
> @kevin-v-ngo Do we want to call contention _transaction contention_ or _database contention_. In our recent blog post we are calling it the latter. https://www.cockroachlabs.com/blog/what-is-database-contention/

I'd still use transaction contention to be more specific versus the general term of database contention. The key update we're making with this change is that whenever we refer to any form of contention (database contention, transaction contention, lock contention, or just contention), we only mean when the workload is forced to wait due to the serializable transaction isolation level of CRDB.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,53 +323,69 @@ If you have long-running queries (such as analytics queries that perform full ta
...
+
+- Increase [normalization](https://en.wikipedia.org/wiki/Database_normalization) of the data to place parts of the same records that are modified by different transactions in different tables. Fully normalized data allows separate transactions to modify related underlying data without causing contention. Increasing normalization has drawbacks as well as benefits, however. Denormalizing data can increase performance for certain queries by creating multiple copies of often-referenced data in separate ranges, so read-heavy workloads may have higher performance. However, denormalization adds complexity the data model, increases the chance of data inconsistency, increases data redundancy, and can have poor performance when running write-heavy workloads.
+
+- If the application strictly requires operating on very few different index key values, consider using [`ALTER ... SPLIT AT`](split-at.html) so that each index key value can be served by a separate group of nodes in the cluster.
```
rmloveland, line 342: Similarly maybe we should add another bullet here with the advice from the [schema design page on tables](https://www.cockroachlabs.com/docs/v21.2/schema-design-table.html#primary-key-best-practices) re: primary keys which says:

> If you are working with a table that must be indexed on sequential keys, use hash-sharded indexes. For details about the mechanics and performance improvements of hash-sharded indexes in CockroachDB, see our Hash Sharded Indexes Unlock Linear Scaling for Sequential Workloads blog post.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,53 +323,69 @@ If you have long-running queries (such as analytics queries that perform full ta
...
+
+- To avoid read hot spots, use global tables and increase the number of replicas.
```
rmloveland, line 344: I'm not sure the advice in this bullet is correct.

Using [`GLOBAL` tables](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html#global-tables) and [increasing the number of replicas](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#num_replicas) are different things.

Re: global tables:

- one can only use global tables in a [multi-region cluster](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html)
- global tables are only good for read-mostly workloads on data that must be accessible 
- ... and one should generally not adjust the number of replicas on a `GLOBAL` table, the whole point of the multi-region abstractions is that users should not need to do low-level replication changes
- in general I don't think we should recommend that the median user with a read hot spot should turn on multiregion + global tables without knowing more about their workload / deployment

Re: increasing replication factor:

if you are not running a "new style" multi-region cluster (link above) then I think increasing the number of replicas could make sense, but it's not straightforward and involves tradeoffs.  I'm not the expert on those tradeoffs, but AIUI: more replication costs (CPU for replication queue etc), more storage (for ranges on different nodes), etc.

my understanding is that it's not something we should recommend to users without caveats.  Not sure if we should suggest it to the median user since it could result in outcomes on their cluster they are not prepared for - I would describe it as more intermediate/advanced usage that most users should not do unless they really know what they're doing

Re: read hot spots in general

- another option for avoiding read hot spots in a multi-region deployment is to use [follower reads](https://www.cockroachlabs.com/docs/v21.1/follower-reads.html)
- we also have a [hot ranges API endpoint](https://www.cockroachlabs.com/docs/v21.2/cluster-api#resources) which is used to "Get information on ranges receiving a high number of reads or writes." which we should probably mention here since it will allow folks to zero in on these hot ranges - this section should probably also add a link from that page back to here

@ajwerner since you're already on this review do you mind fact checking me on the above? also, do you have a concrete 1-3 recommendations to avoid read hot spots?

v21.2/performance-best-practices-overview.md
```diff
@@ -323,53 +323,69 @@ If you have long-running queries (such as analytics queries that perform full ta
 
 However, because `AS OF SYSTEM TIME` returns historical data, your reads might be stale.
 
-## Understanding and avoiding transaction contention
+## Hot spots
 
-Transaction contention occurs when the following three conditions are met:
+Transactions that operate on the same range but _different index key values_ will be limited by the overall hardware capacity of a single node (the range lease holder). These are referred to as _hot spots_.
```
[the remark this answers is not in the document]
rmloveland: As you did with 'Indicators of Transaction Contention' below, I think it would be good to add a similar subsection here and point users to the [hot ranges API endpoint](https://www.cockroachlabs.com/docs/v21.2/cluster-api#resources) (see `/ranges/hot`)

rmloveland (changes requested):
Most of it looks pretty good but I think we still need revisions on the read hotspot mitigations (comment below) and agree with Kevin's comments re: moving some of the mentions of things that seem to be "hot spots" out of "contention" since we have the new hot spots section now

v21.2/performance-best-practices-overview.md
rmloveland: NB according to the link above the hash sharded indexes help with sequential write perf but have a small read cost

v21.2/performance-best-practices-overview.md
```diff
@@ -323,53 +323,69 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
-- There are multiple concurrent transactions or statements (sent by multiple clients connected simultaneously to a single CockroachDB cluster).
-- They operate on the same data, specifically over table rows with the same index key values (either on [primary keys](primary-key.html) or secondary [indexes](indexes.html)) or using index key values that are close to each other, and thus place the indexed data on the same [data ranges](architecture/overview.html).
-- At least one of the transactions modify the data.
+Hot spots can occur when a range is indexed on a column of data that is sequential in nature such that all incoming writes to the range will be the last (or first) item in the index and appended to the end of the range. As a result, the system cannot find a point in the range that evenly divides the traffic, and the range cannot benefit from load-based splitting, creating a hotspot on the single range.
```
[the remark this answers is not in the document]
stbof: Fixed.

v21.2/performance-best-practices-overview.md
```diff
@@ -323,53 +323,69 @@ If you have long-running queries (such as analytics queries that perform full ta
...
 
-A set of transactions that all contend on the same keys will be limited in performance to the maximum processing speed of a single node (limited horizontal scalability). Non-contended transactions are not affected in this way.
+Read hot spots can occur if you perform lots of scans of an empty table or of a single key.
+
+### Fix hot spots
+
+To reduce hot spots:
+
+- Use index key values with a random distribution of values, so that transactions over different rows are more likely to operate on separate data ranges. See the [SQL FAQs](sql-faqs.html) on row IDs for suggestions.
```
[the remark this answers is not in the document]
stbof: Fixed.

v21.2/performance-best-practices-overview.md
stbof: Fixed.

v21.2/performance-best-practices-overview.md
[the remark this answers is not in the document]
stbof: Fixed.

ajwerner (approved):
Sorry for the extreme review delay. 

I could see another pass to synthesize the approaches for dealing with hot spots, but, on the whole, :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12494#-:-Mt6B_as6sfSpCGd9bvM:b4opqxz)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @kevin-v-ngo, @rmloveland, and @stbof)

___
*[v21.2/performance-best-practices-overview.md, line 352 at r2](https://reviewable.io/reviews/cockroachdb/docs/12494#-MpsFV6O5IrF4PGIFFCN-r2-352:-Mt6AxEC2j84vxFueCRx:bb6un1j) ([raw file](https://github.com/cockroachdb/docs/blob/736a2af27dc671b4af2f95b3bb8abdd7845fd26a/v21.2/performance-best-practices-overview.md#L352)):*
<details><summary><i>Previously, kevin-v-ngo wrote…</i></summary><blockquote>

This seems to be describing hot spots. Contending on the same keys would mainly cause transactions to wait (or restart) - it's less of a resource issue. @ajwerner can you confirm?
</blockquote></details>

Done
___
*[v21.2/performance-best-practices-overview.md, line 344 at r5](https://reviewable.io/reviews/cockroachdb/docs/12494#-MpsFV6O5IrF4PGIFFCN-r5-344:-Mt68qoS8wRZ0GgmdFqh:brkdgkx) ([raw file](https://github.com/cockroachdb/docs/blob/edeadf33da67a5d263da278c6821680791cdc98f/v21.2/performance-best-practices-overview.md#L344)):*

I'm not sure the advice in this bullet is correct.

Using [`GLOBAL` tables](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html#global-tables) and [increasing the number of replicas](https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#num_replicas) are different things.

Re: global tables:

- one can only use global tables in a [multi-region cluster](https://www.cockroachlabs.com/docs/v21.1/multiregion-overview.html)
- global tables are only good for read-mostly workloads on data that must be accessible 
- ... and one should generally not adjust the number of replicas on a `GLOBAL` table, the whole point of the multi-region abstractions is that users should not need to do low-level replication changes
- in general I don't think we should recommend that the median user with a read hot spot should turn on multiregion + global tables without knowing more about their workload / deployment

Re: increasing replication factor:

if you are not running a "new style" multi-region cluster (link above) then I think increasing the number of replicas could make sense, but it's not straightforward and involves tradeoffs.  I'm not the expert on those tradeoffs, but AIUI: more replication costs (CPU for replication queue etc), more storage (for ranges on different nodes), etc.

my understanding is that it's not something we should recommend to users without caveats.  Not sure if we should suggest it to the median user since it could result in outcomes on their cluster they are not prepared for - I would describe it as more intermediate/advanced usage that most users should not do unless they really know what they're doing

Re: read hot spots in general

- another option for avoiding read hot spots in a multi-region deployment is to use [follower reads](https://www.cockroachlabs.com/docs/v21.1/follower-reads.html)
- we also have a [hot ranges API endpoint](https://www.cockroachlabs.com/docs/v21.2/cluster-api#resources) which is used to "Get information on ranges receiving a high number of reads or writes." which we should probably mention here since it will allow folks to zero in on these hot ranges - this section should probably also add a link from that page back to here

@ajwerner since you're already on this review do you mind fact checking me on the above? also, do you have a concrete 1-3 recommendations to avoid read hot spots?
</blockquote></details>

Increasing the replication won't help you with the hot spot if your table is not GLOBAL, unless you use follower reads (as you noted). I think the principles we can teach here are that there are two main approaches to coping with hot spots: 

1) Increase data distribution which will allow for more ranges. The hot spot existed because the data being accessed was all co-located in one range. Most of other points address this.
2) Increase load balancing across more nodes in the same range. Most transactional reads must go to the leaseholder in cockroach, that means that the opportunities for load balancing over replicas is ~zero. However, there are some features which do permit load-balancing over replicas. In these cases, more replicas will help, up to the number of nodes in the cluster. They all only help with reads, and they all come with their own tradeoffs.
    * Global tables
    * Follower reads (both the bounded staleness and the exact staleness kinds)
___
*[v21.2/performance-best-practices-overview.md, line 320 at r7](https://reviewable.io/reviews/cockroachdb/docs/12494#-Mt66t3u7kDYTf3NcIj1:-Mt66t3u7kDYTf3NcIj2:bsxafv1) ([raw file](https://github.com/cockroachdb/docs/blob/e1704674bbc96a282f9d0a85187ea4dd3a93425f/v21.2/performance-best-practices-overview.md#L320)):*
> ```Markdown
> ## Hot spots
> 
> Transactions that operate on the same range but _different index key values_ are limited by the overall hardware capacity of [the range lease holder](architecture/overview.html#terms) node. These are referred to as _hot spots_.
> ```

nit: consider `_different index keys_` or `_different rows_`  or `_different index entries_` over `_different index key values_` here and below. My brain had a hard time parsing the latter phrase. 
___
*[v21.2/performance-best-practices-overview.md, line 320 at r7](https://reviewable.io/reviews/cockroachdb/docs/12494#-Mt67APDC5d4yrvKIliH:-Mt67APDC5d4yrvKIliI:bzat28e) ([raw file](https://github.com/cockroachdb/docs/blob/e1704674bbc96a282f9d0a85187ea4dd3a93425f/v21.2/performance-best-practices-overview.md#L320)):*
> ```Markdown
> ## Hot spots
> 
> Transactions that operate on the same range but _different index key values_ are limited by the overall hardware capacity of [the range lease holder](architecture/overview.html#terms) node. These are referred to as _hot spots_.
> ```

Okay, don't hate me for using the word contention as I point out a piece of minutiae that likely isn't worth documenting. It's a minor technical nuance but it's also the fact that operating on a single range, even on one machine, can cause decreased throughput. The reason is that within the context of that range, each operation may need to access some in-memory state exclusively, to, say, choose a raft log index. This overhead we'd say is "mutex contention". It's not really very important for this piece because if anybody is worried about performance, they are using more than one node. 
___
*[v21.2/performance-best-practices-overview.md, line 324 at r7](https://reviewable.io/reviews/cockroachdb/docs/12494#-Mt67lhI1m6b346QavG2:-Mt67lhI1m6b346QavG3:b-zh828p) ([raw file](https://github.com/cockroachdb/docs/blob/e1704674bbc96a282f9d0a85187ea4dd3a93425f/v21.2/performance-best-practices-overview.md#L324)):*
> ```Markdown
> Hot spots can occur when a range is indexed on a column of data that is sequential in nature such that all incoming writes to the range will be the last (or first) item in the index and appended to the end of the range. As a result, the system cannot find a point in the range that evenly divides the traffic, and the range cannot benefit from [load-based splitting](load-based-splitting.html), creating a hot spot on the single range.
> 
> Read hot spots can occur if you perform lots of scans of an empty table or of a single key.
> ```

maybe `empty portion of a table index or a single key`?  We had a customer which had a sizable table but kept spinning querying something like:

```
CREATE TABLE tasks (
   id.        INT
   status  STRING
   INDEX (status)
);

SELECT * FROM tasks WHERE status = 'ready'; -- had a hot spot because no tasks were generally ready
```

There are a bunch of other problems in that example, so let's not dig too deeply into it. 
___
*[v21.2/performance-best-practices-overview.md, line 328 at r7](https://reviewable.io/reviews/cockroachdb/docs/12494#-Mt68RvhBXq9Km7HhOos:-Mt68RvhBXq9Km7HhOot:b-r07fn2) ([raw file](https://github.com/cockroachdb/docs/blob/e1704674bbc96a282f9d0a85187ea4dd3a93425f/v21.2/performance-best-practices-overview.md#L328)):*
> ```Markdown
> ### Find hot spots
> 
> To track down the nodes experiencing hot spots, use the [hot spot ranges API endpoint](cluster-api.html#resources).
> ```

for whatever reason, the rest of the APIs and docs call this the `hot range API endpoint` (no `spot`).
___
*[v21.2/performance-best-practices-overview.md, line 359 at r7](https://reviewable.io/reviews/cockroachdb/docs/12494#-Mt6A5hn1jrTUWY04tYn:-Mt6A5hn1jrTUWY04tYo:b-585hrd) ([raw file](https://github.com/cockroachdb/docs/blob/e1704674bbc96a282f9d0a85187ea4dd3a93425f/v21.2/performance-best-practices-overview.md#L359)):*
> To maintain this isolation, "locks" ensure that multiple transactions don't change the same row of data (for example) at the same time.

nit, what do you think about the below edit. It's intentionally more ambiguous. Contention can occur when even a high percentage of the transactions are reading the row. At least one involved transaction needs to write the row, but it could be very skewed. The worst cases we tend to see are when there are many reads and one, very large writing transaction. You do cover that below so feel free to totally ignore this one. 

`To maintain this isolation, writing transactions lock rows to prevent hazardous interactions with concurrent transactions.`  



_Code quote:_
```Markdown
To maintain this isolation, "locks" ensure that multiple transactions don't change the same row of data (for example) at the same time.
```

rmloveland (approved):
LGTM modulo adding the updated recommendations from @ajwerner - thanks!

v21.2/performance-best-practices-overview.md
stbof: Fixed.

PR #12509, opened 2021-12-06
Updated style sheet to allow horizontal scrolling on large div blocks

nickvigilante:
Description:

- Updated CSS to add horizontal scrolling on all DIV blocks.


nickvigilante:
@jseldess It looks like customstyles.css was formatted, hence the large diff. I can confirm that I only manually updated customstyles.**scss**.

jseldess:
Nice, @nickvigilante! Sorry for the delay. 

This does fix the SQL diagram issue. However, I see some unintended styling consequences that we don't want. For example, just at a glance, the search field placement in the sidebar has changed:

Screen Shot 2021-12-09 at 10 18 18 AM

And the copy icon in code blocks is not outside of the code block:

Screen Shot 2021-12-09 at 10 18 46 AM

There may be other unintended consquences. Please take a look. 

Our CSS is pretty unwieldy, so it doesn't surprise me to see these side-effects. It might be good for you to loop in one of the web devs on Marketing for guidance, @anagio or @JosephNickdow.

nickvigilante:
@jseldess Review once more? I was able to specifically target all `div` blocks that contain an `svg` block, so the two side effects you mentioned are no longer a problem.

jseldess:
Hmm, @nickvigilante, now I'm seeing the old behavior again. Can you double-check? https://deploy-preview-12509--cockroachdb-docs.netlify.app/docs/stable/insert.html#on-conflict-clause

nickvigilante:
@jseldess That's definitely odd, because I built it locally, and it's fixed. I will attach a video to the JIRA ticket. Do you have some time tomorrow to discuss?

nickvigilante:
Closing as the solution must be applied at the DIV level and not the SVG level.

PR #12512, opened 2021-12-06
Note added to Serverless FAQs re the error on cloud storage/free tier

kathancox:
Fixes DOC-2092 

Added a note to the Serverless FAQ page that mentions the error that users will receive if they attempt to run a bulk operation to cloud storage without having entered billing info first. This had come up as a question multiple times — hope is that this will add a little more discoverability while changing the actual error message currently has a few more complexities around it.


livlobo (commented):
LGTM. Would it be worth clarifying that they can still set their spend limit to zero? We just need them to input billing info; but we won't charge them. It's still free.

kathancox:
@livlobo Do you think in the answer just above this note box, that the final sentence would suffice on that front?

rmloveland (approved):
LGTM

PR #12523, opened 2021-12-07
Remove SPLIT AT note re: resplitting vs TRUNCATE

### LINKED ISSUE #11779: SPLIT AT Docs for 21.1 do not cover changed behavior for TRUNCATE
Nick Lamb (nickelnickel) commented:

As per https://github.com/cockroachdb/cockroach/pull/65940 this is no longer the case in 20.2+

> Note that when a table is truncated, it is essentially re-created in a single new empty range, and the old ranges that used to constitute the table are garbage collected. Any pre-splitting you have performed on the old version of the table will not carry over to the new version. The new table will need to be pre-split again.

This should be reworded to cover the changed behavior.

rmloveland:
Fixes #11779

### DIFF
--- a/v21.1/split-at.md
+++ b/v21.1/split-at.md
@@ -54,8 +54,6 @@ the ranges that store tables or indexes:
   workload performance that results when automatic splits are unable to keep up
   with write traffic.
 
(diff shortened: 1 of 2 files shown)


rmloveland:
Hey @jordanlewis any objections to this change?  Happy to ask for a different reviewer if that makes sense also, I know you are busy! (Would love to merge this week before PTO)

jordanlewis (approved):
LGTM. Thanks, Rich! Sorry I missed this.

rmloveland:
no worries Jordan!  thanks!

PR #12537, opened 2021-12-09
Fix typo found in --exclude-nodes example in debug-zip docs

Marjan154:
Noticed a minor typo in https://www.cockroachlabs.com/docs/stable/cockroach-debug-zip.html#general, under the exclude-nodes  flag example. Made a fix.


nickvigilante:
@Marjan154 Thank you for the contribution! We restructured our content and changed default branch names, so I applied your suggestions to our `main` branch via https://github.com/cockroachdb/docs/pull/17219.

Issue #11022, opened 2021-08-09
Add per-driver connection information


ericharmeling:
Eric Harmeling (ericharmeling) commented:

For each supported driver and ORM framework, we currently have [download and installation information](https://www.cockroachlabs.com/docs/stable/install-client-drivers.html) and [example applications](https://www.cockroachlabs.com/docs/stable/example-apps.html).

We do not have a dedicated page for connection information for each driver/ORM. I think users would find this information super helpful to have in one place (alla [Install a Driver or ORM Framework](https://www.cockroachlabs.com/docs/stable/install-client-drivers.html)), especially if we somehow linked to this page from the Connect page of the CC console (and possibly in the welcome page of the relevant `cockroach` commands).

All of the relevant connection information should be available on either the tool's example app tutorial page, or on the tool's third-party website.

Issue #12353, opened 2021-11-18
Sequelize 6.11 connection string update


ericharmeling:
_Originally posted by @kernfeld-cockroach in https://github.com/cockroachdb/examples-orms/pull/144#pullrequestreview-810382884_

PR #12538, opened 2021-12-09
Client connection reference page

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/11022.
Fixes https://cockroachlabs.atlassian.net/browse/DOC-1856.
Related to https://github.com/cockroachdb/docs/issues/12353.

This PR adds a client connection reference page with connection configuration information for all of our fully-supported client drivers and ORM frameworks. 

I collected the information presented in this reference page from various internal documents (Google Docs, Jira tickets, other GH issues) and from third-party reference documentation. As of 1/6/22, all of these connection configurations have been tested, with the latest version of each client driver/ORM. 

*Note that the connection reference provided here is specific to cloud cluster connections.*


v21.2/connection-reference.md
```diff
@@ -0,0 +1,732 @@
...
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+)
+
+func main() {
+	conn, err := pgx.Connect(context.Background(), "{connection-string}")
+  if err != nil {
```
kernfeld-cockroach, line 301: Indentation

v21.2/connection-reference.md
```diff
@@ -0,0 +1,732 @@
...
+
+<section class="filter-content" markdown="1" data-scope="python">
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="psycopg">Psycopg</button>
+  <button class="filter-button page-level" data-scope="sqlalchemy">SQLAlchemy</button>
+  <button class="filter-button page-level" data-scope="django">Django</button>
+</div>
+
+{{site.data.alerts.callout_danger}}
```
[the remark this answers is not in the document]
ericharmeling: Done.

v21.2/connection-reference.md
```diff
@@ -0,0 +1,732 @@
...
+
+{% include copy-clipboard.html %}
+~~~
+postgresql://<username>:<password>@<host>:<port>/<cluster_name>.<database>?sslmode=verify-full
+~~~
+
+You can also specify the connection configuration as individual properties:
+
+{% include copy-clipboard.html %}
+~~~ js
```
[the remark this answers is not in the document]
ericharmeling: Done. (went with `Pool`)

ericharmeling (commented):
TFTR @kernfeld-cockroach !

See my responses inline.

bdarnell (commented):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connect-to-the-database.md, line 37 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz-KlS9_B9Z1iEwI36:-Mqz-KlS9_B9Z1iEwI37:btr8rne) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connect-to-the-database.md#L37)):*
> ```Markdown
> 
> <div class="filter-content" markdown="1" data-scope="local">
> Use the connection information provided by the [`cockroach` client](cockroach-commands.html) to configure your client to connect to CockroachDB.
> ```

This is referring to the output of `cockroach start` or `start-single-node`, right? In that case the `cockroach` command is acting as the server, not a client. 

This page is pretty small now - do we still need it? 
___
*[v21.2/connection-reference.md, line 54 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2weN0-poNupsvnIk:-Mqz2weN0-poNupsvnIl:b7ctpf4) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L54)):*
> ```Markdown
> ~~~
> postgresql://<username>:<password>@<host>:<port>/<cluster_name>.<database>?sslmode=verify-full
> ```

Don't we want to prefer the cluster options parameter instead of putting the cluster name in the database name when it's supported? 
___
*[v21.2/connection-reference.md, line 139 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2GE38HigKq9FeoDd:-Mqz2GE38HigKq9FeoDe:b-atjzca) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L139)):*
> ```Markdown
> 
> To connect to a {{ site.data.products.serverless }} cluster from a Python application, you must have a valid CA certificate located at <code>~/.postgresql/root.crt</code>.<br>For instructions on downloading a CA certificate from the {{ site.data.products.db }} Console, see <a href="https://www.cockroachlabs.com/docs/v21.2/cockroachcloud/connect-to-a-serverless-cluster.html">Connect to a {{ site.data.products.serverless }} Cluster</a>.
> ```

Non-serverless clusters must also have a root certificate downloaded (in all languages, not just python and ruby). 

We should probably have some explanation of what to do if you have to work with multiple clusters (or both cockroach and postgres). (You may be able to concatenate the files, or you may want to use multiple paths and specify them explicitly on the connection string depending on the security practices of the various CAs in question). 
___
*[v21.2/connection-reference.md, line 414 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz4RMz1tqtBgXLbokk:-Mqz4RMz1tqtBgXLbokl:b-nilxvn) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L414)):*
> ```Markdown
> 
> JDBC connection URLs do not accept the `username` and `password` parameters. These parameters must be set outside of the connection string.
> ```

I think JDBC does support username and password in the connection string, it's just in a different place than the others. Append `&user={user}&password={password}`. (I think I've tested this in the past, but not today so please confirm).

andy-kimball (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 246 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqzXqoxEcsMMCDF07rC:-MqzXqoy5xGWQbggbfLD:b-9x6vml) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L246)):*
> ```Markdown
> `{host}`  | The host on which the CockroachDB node is running.
> `{port}`  | The port at which the CockroachDB node is listening.
> `{cluster_name}`  | The name of the CockroachDB cluster.
> ```

This is misleading, because it's not the name of the cluster - it's the name of the cluster _plus_ a unique number called the tenant_id. The problem is that two different customers could name their clusters the same thing, and we have to know which one to connect to. Therefore, we add a unique number that gives us that information.

I've recently written an [in-depth description](https://github.com/cockroachlabs/managed-service/pull/7411/files?short_path=30ccbd2#diff-30ccbd2ae4dd3be33ef626351c84dbe23f4e33e3c0e6ee38c9619cb020756583) of all the problems with connection routing. I'd suggest reading through that in order to get all the context of why the problem is so difficult to solve in a simple way.

ericharmeling (commented):
TFTR @bdarnell !

See my responses inline.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538#-:-MqzkS-n0GktTcQonfAn:bwpaio)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @ericharmeling, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connect-to-the-database.md, line 37 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz-KlS9_B9Z1iEwI36:-MqzG5Qs-Jdn3IyCp5RX:bmuaoyx) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connect-to-the-database.md#L37)):*
> In that case the cockroach command is acting as the server, not a client.

Error on my part. Fixed.

> This page is pretty small now - do we still need it?

I don't think so, actually... But removing it might be out of the scope of this PR, as this page is referenced throughout our docs and those references will require some redirection/rewriting. I've opened [a docs issue](https://cockroachlabs.atlassian.net/browse/DOC-2179) to handle a rework of this whole "Connect" section in a separate PR. 
___
*[v21.2/connection-reference.md, line 54 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2weN0-poNupsvnIk:-MqzHgrI3p477Mqvebvy:beq4ge7) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L54)):*

Don't we want to prefer the cluster options parameter instead of putting the cluster name in the database name when it's supported? 
</blockquote></details>

I did not know that was the preference. My fault. Updated with the cluster parameter in the connection string, and retested each connection string to verify that that format indeed was supported. 
___
*[v21.2/connection-reference.md, line 139 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2GE38HigKq9FeoDd:-MqzK1Yr9uYGLhK3nSms:b5oa8rd) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L139)):*

Non-serverless clusters must also have a root certificate downloaded (in all languages, not just python and ruby). 

We should probably have some explanation of what to do if you have to work with multiple clusters (or both cockroach and postgres). (You may be able to concatenate the files, or you may want to use multiple paths and specify them explicitly on the connection string depending on the security practices of the various CAs in question). 
</blockquote></details>

Good to note. 

I envision this page having distinct serverless, dedicated, and self-hosted connection information for each tool, separated by functioning tabs. At this point, it just has info for serverless. After the structure of this page and of each tool's section is approved for serverless, I will add dedicated and core, which will include a more general note that dedicated and self-hosted clusters require a local root certificate.

I'll keep the commits separate so that reviewers can look at the changes individually after this first iteration is approved.  I'll also have to test all of those snippets too, so it might take some time before I get those commits up.

Apologies for the churn. FWIW, I think getting the serverless snippets up and approved (at least in this PR) ASAP is going to help get design/product on the right track for the [console updates](https://cockroachlabs.atlassian.net/browse/CC-3454) more quickly.
___
*[v21.2/connection-reference.md, line 414 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz4RMz1tqtBgXLbokk:-MqzI5Fa9X9ll3r9zran:b2vli06) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L414)):*

I think JDBC does support username and password in the connection string, it's just in a different place than the others. Append `&user={user}&password={password}`. (I think I've tested this in the past, but not today so please confirm). 
</blockquote></details>

Confirmed. Fixed.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 246 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqzXqoxEcsMMCDF07rC:-MqznIeZ16gkOYj23-43:bkc9bkn) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L246)):*
> This is misleading, because it's not the name of the cluster - it's the name of the cluster plus a unique number called the tenant_id. The problem is that two different customers could name their clusters the same thing, and we have to know which one to connect to. Therefore, we add a unique number that gives us that information.

Noted. I can make that change here, but I want to make sure that DX is on the same page. @kernfeld-cockroach I know that you had some misgivings about mentioning tenant ID in the docs (https://github.com/cockroachdb/docs/pull/12330#discussion_r753459401 and https://cockroachlabs.atlassian.net/browse/DOC-1856). 

I don't think we've been very consistent in our docs about how we refer to the `clustername-tenantid` value in general, but we should make a decision one way or another and aim for consistency. Because this page is a technical reference doc, we especially want to be accurate and clear on this page.

> I've recently written an in-depth description of all the problems with connection routing. I'd suggest reading through that in order to get all the context of why the problem is so difficult to solve in a simple way.

Thanks for sharing! I did a quick read-through and found the motivation and implementation compelling.

andy-kimball:
___
*[v21.2/connection-reference.md, line 246 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqzXqoxEcsMMCDF07rC:-Mr-6EtM75UZPBpuloSl:bt1ai1q) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L246)):*

> This is misleading, because it's not the name of the cluster - it's the name of the cluster plus a unique number called the tenant_id. The problem is that two different customers could name their clusters the same thing, and we have to know which one to connect to. Therefore, we add a unique number that gives us that information.

Noted. I can make that change here, but I want to make sure that DX is on the same page. @kernfeld-cockroach I know that you had some misgivings about mentioning tenant ID in the docs (https://github.com/cockroachdb/docs/pull/12330#discussion_r753459401 and https://cockroachlabs.atlassian.net/browse/DOC-1856). 

I don't think we've been very consistent in our docs about how we refer to the `clustername-tenantid` value in general, but we should make a decision one way or another and aim for consistency. Because this page is a technical reference doc, we especially want to be accurate and clear on this page.

> I've recently written an in-depth description of all the problems with connection routing. I'd suggest reading through that in order to get all the context of why the problem is so difficult to solve in a simple way.

Thanks for sharing! I did a quick read-through and found the motivation and implementation compelling.
</blockquote></details>

I agree that we should not break out these different parts as in `<cluster-name>-<tenant-id>` for users. Instead, I think we should have a single term like `routing-id` and then tell customers to go look up their `routing-id` in the Connection Parameters section of their dashboard (similar to what @kernfeld-cockroach suggests). Today the routing ID will be `<cluster-name>-<tenant-id>`. If we adopt the proposal I linked to, then that will change to `serverless-<tenant-id>.<host-short-id>`. But in either case, _we should not describe the component parts to the user_. We should always treat it as an opaque `routing-id`, indivisible and not to be interpreted.

rafiss (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @ericharmeling, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 219 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0fHxj7FJd6ExbJhur:-Mr0fHxs8AoIHlgIhG7U:bn4pej5) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L219)):*
> ```Markdown
> 
> JDBC connection URLs do not accept the `username` and `password` parameters. These parameters must be set outside of the connection string.
> ```

i believe a URL like this might work
```
jdbc:postgresql://{host}:{port}/{database}?sslmode=verify-full&options=--cluster%3D{cluster_name}&user={username}&password={password}&sslfactory=org.postgresql.ssl.DefaultJavaSSLFactory
```
___
*[v21.2/connection-reference.md, line 621 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0h_mJ1Kb3fwrCkNH_:-Mr0h_mK47AremnyH4Ix:bop493g) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L621)):*
> ```Markdown
>   host: "<host>",
>   port: <port>,
>   database: "<cluster_name>.<database>"
> ```

instead of using `cluster.database` here, i have tested sequelize and seen that it supports options like follows:
```
var sequelize = new Sequelize({
  dialect: "postgres",
  username: "<username>",
  password: "<password>",
  host: "<host>",
  port: <port>,
  database: "<database>",
  dialectOptions: {
    options: '--cluster=<cluster_name>'
  }
});
```

but this is only for sequelize v6.10.0 and greater. actually, the URL version also only is with v6.10.0 or greater. maybe that deserves a mention? otherwise you have to do some tricky stuff, like the "before" part of this commit: https://github.com/cockroachdb/examples-orms/commit/7398e856f9e40cdbf7322ea2afb7d84d7a368e5b
___
*[v21.2/connection-reference.md, line 668 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0jDV-34KhGXLZgt_K:-Mr0jDV0AUeDjVrvRo6u:b4zly53) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L668)):*
> ```Markdown
>     host: "<host>",
>     port: <port>,
>     database: "<cluster_name>.<database>"
> ```

i haven't tested typeorm myself, but based on my reading of the docs, this should work:
```
const connection = await createConnection({
    type: "postgresql",
    username: "<username>",
    password: "<password>",
    host: "<host>",
    port: <port>,
    database: "<database>",
    options: "--cluster=<cluster_name>"
});
```
___
*[v21.2/connection-reference.md, line 684 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0jTcuDdvXJPjgd4mW:-Mr0jTcuDdvXJPjgd4mX:bce63bk) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L684)):*
> ```Markdown
>     "host": "<host>",
>     "port": <port>,
>     "database": "<cluster_name>.<database>"
> ```

ditto above comment about `options`

rafiss (changes requested):
oops sorry my comments look out of place now since i left them on revision 1. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538#-:-Mr0kvSv3rWstL7HyLto:b-bvmbpx)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @ericharmeling, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 219 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0fHxj7FJd6ExbJhur:-Mr0krxe6QoYjnBoW2Ih:b-rwspam) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L219)):*

i believe a URL like this might work
```
jdbc:postgresql://{host}:{port}/{database}?sslmode=verify-full&options=--cluster%3D{cluster_name}&user={username}&password={password}&sslfactory=org.postgresql.ssl.DefaultJavaSSLFactory
```
</blockquote></details>

ah i see ben commented this on a later revision.

ericharmeling (commented):
TFTR @rafiss !

> oops sorry my comments look out of place now since i left them on revision 1.

No worries. I think I got to all of your comments that are still relevant to the latest revision, but some of the comments are now a little out-of-place in reviewable. LMK if I missed something.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538#-:-Mr2WAHo3g2rTkLUVRBG:b88pf8o)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @emily-horing, @ianjevans, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 621 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0h_mJ1Kb3fwrCkNH_:-Mr2oFdJ2qZ6_kLIFgpk:b-4d7ofc) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L621)):*
I updated this snippet with the standard `postgresql://<username>:<password>@<host>:<port>/<database>?sslmode=verify-full&options=--cluster%3D<routing-id>` URL format - which I tested against v6.10 and v6.11. 

> actually, the URL version also only is with v6.10.0 or greater. maybe that deserves a mention?

I was actually able to get a URL to work for earlier versions, but with a slightly different format. I've added that format to the page in a new connection string snippet.

The page now lists a format for v6.10.0+ and a format for earlier versions. The earliest version I could get that second format to work with is 6.1.0 (which is like a year old at this point).

Note that the format for earlier versions appears to be backwards-compatible... so perhaps we just want to provide that one? It uses `routing-id.database` for the database name rather than the preferred `cluster=routing-id` individual parameter though, so IDK. I'm fine with providing both until we think most users will have 6.10+.
___
*[v21.2/connection-reference.md, line 668 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0jDV-34KhGXLZgt_K:-Mr2otY_7or1Dza2djdB:bv1h0cq) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L668)):*

i haven't tested typeorm myself, but based on my reading of the docs, this should work:
```
const connection = await createConnection({
    type: "postgresql",
    username: "<username>",
    password: "<password>",
    host: "<host>",
    port: <port>,
    database: "<database>",
    options: "--cluster=<cluster_name>"
});
```
</blockquote></details>

This config doesn't work for a few reasons:

- Using the `postgres` type (`postgresql` isn't a valid type) returns some interesting errors that I don't really understand. Using `cockroachdb` as the type instead resolves those errors.
- You need to specify a value for the `ssl` parameter (e.g., `ssl: true`). Not doing so results in the following error: `error: server requires encryption`.
- `options` doesn't appear to be a valid config parameter for TypeORM. I think you have to use `extras`.

I updated this config to use a URL instead of individual parameters, but if we want to use the `--cluster=<routing-id>` option, we need to specify that parameter individually because TypeORM can't parse it from the URL (or at least I can't seem to format the URL in such a way that `--cluster=<routing-id>` gets parsed properly).

Also, I actually think it's much more common for users to use an `ormconfig` file for TypeORM (but I could be wrong on this?), and then just calling `createConnection` without any arguments (see https://typeorm.io/#/using-ormconfig/creating-a-new-connection-from-the-configuration-file). But the parameters should be the same.

I've updated the page with a working configuration - LMK if you think it needs any changes.
___
*[v21.2/connection-reference.md, line 684 at r1](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mr0jTcuDdvXJPjgd4mW:-Mr3sfk40pQ-EnQakdY8:bifo0ds) ([raw file](https://github.com/cockroachdb/docs/blob/c48186fb49d689a0a55944d46f8dda56fb24704f/v21.2/connection-reference.md#L684)):*

ditto above comment about `options`
</blockquote></details>

For ActiveRecord, I was able to get a pretty standard URL to work?

But you might be referring to a different section since the PR has been revised? I am guessing it was Django - which I updated accordingly to your suggestion. Thanks!
___
*[v21.2/connection-reference.md, line 579 at r2](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqVnNud89nvvu1Br5bN-r2-579:-Mr3GihvBLXFB5g74wY4:bvbkeuu) ([raw file](https://github.com/cockroachdb/docs/blob/0d3e8266f0065bf6ccdd74935deccc51ca7404af/v21.2/connection-reference.md#L579)):*

Done. (went with `Pool`)
</blockquote></details>

Actually, changing to use `Client` instead (seems like the most basic option).
___
*[v21.2/connection-reference.md, line 246 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqzXqoxEcsMMCDF07rC:-Mr2mx2KAvxYW31Tezpb:b-1vsvdg) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L246)):*
> I agree that we should not break out these different parts as in <cluster-name>-<tenant-id> for users. Instead, I think we should have a single term like routing-id and then tell customers to go look up their routing-id in the Connection Parameters section

Okay, great. Sounds like everyone is on the same page then. Using a single term sounds like the more user-friendly, future-proof approach. 

@emily-horing I'm going to use `routing-id` for now, but please LMK if there is a specific term we want to use. @ianjevans I imagine whatever term we use will also need to be reflected in docs changes for https://cockroachlabs.atlassian.net/browse/DOC-2040.

emily-horing (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @ianjevans, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 246 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqzXqoxEcsMMCDF07rC:-Mr8BVrB7TfHb4v8NIYF:b-hijyk6) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L246)):*

> I agree that we should not break out these different parts as in <cluster-name>-<tenant-id> for users. Instead, I think we should have a single term like routing-id and then tell customers to go look up their routing-id in the Connection Parameters section

Okay, great. Sounds like everyone is on the same page then. Using a single term sounds like the more user-friendly, future-proof approach. 

@emily-horing I'm going to use `routing-id` for now, but please LMK if there is a specific term we want to use. @ianjevans I imagine whatever term we use will also need to be reflected in docs changes for https://cockroachlabs.atlassian.net/browse/DOC-2040.
</blockquote></details>

`routing-id` is ok with me for now. I think we'll want to reconsider the right convention once the changes Andy suggested are more finalized. (https://github.com/cockroachlabs/managed-service/pull/7411)

v21.2/connection-reference.md
```diff
@@ -0,0 +1,570 @@
...
+
+{% include copy-clipboard.html %}
+~~~
+postgresql://<username>:<password>@<host>:<port>/<database>?sslmode=verify-full&options=--cluster%3D<routing-id>
+~~~
+
+For more information about connecting with node-postgres, see the [official node-postgres documentation](https://node-postgres.com/features/connecting).
+
+</section>
+
```
kernfeld-cockroach, line 63: Do we want a `## Connect with Sequelize` here?

v21.2/connection-reference.md
```diff
@@ -0,0 +1,732 @@
...
+For example:
+
+{% include copy-clipboard.html %}
+~~~ python
+import psycopg
+
+conn = psycopg.connect('{connection-string}')
+~~~
+
+Psycopg accepts the following format for CockroachDB connection strings:
```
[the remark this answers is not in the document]
kernfeld-cockroach: What I was quite unsuccessfully attempting to articulate here is that, in an ideal world, we wouldn't even need to document the format of CockroachDB connection strings for Serverless because the user would simply copy the appropriate connection string directly from the CockroachDB Cloud web UI. However, I am now realizing that this is not immediately possible because the web UI doesn't generate the correct connection strings. So, I guess the first step would be for the web UI to generate the correct connection string for a particular tool. Once that's ready to go, then we could un-document this.

andy-kimball:
I see that you switched the docs to use `<routing-id>`, but I don't see where that's defined or instructions on how to get it. Did I miss something or are you doing that in a different PR?

v21.2/connection-reference.md
```diff
@@ -0,0 +1,570 @@
...
+
+To connect to CockroachDB with [node-postgres](https://node-postgres.com), create a new [`Client`](https://node-postgres.com/api/client) object with a connection string.
+
+For example:
+
+{% include copy-clipboard.html %}
+~~~ js
+const { Client } = require('pg')
+
+const connectionString = '<connection-string>'
```
kernfeld-cockroach, line 45: I'm concerned that this might encourage developers who don't know any better to paste their connection strings into their source code, which is sketch because it contains the password. What do you think of modifying these examples to instead read from an environment variable named `DATABASE_URL`?

kernfeld-cockroach (approved):
Thanks for this! I learned a lot about connecting to CockroachDB!

kernfeld-cockroach:
I think this information will be most useful for users who want to get started with an app. Can this go under "Get Started" instead?

ericharmeling (commented):
> I see that you switched the docs to use `<routing-id>`, but I don't see where that's defined or instructions on how to get it. Did I miss something or are you doing that in a different PR?

AFAIK, this is the first place where we use `<routing-id>` in the docs. As such, the only description of this term can be found in the "### Connection parameters" table(s) on this page. ATM, this description is very concise ("The routing ID of the CockroachDB cluster (e.g., `funny-skunk-123`)."). For users just copying and pasting from the console, IDK if they really need any more than this. 

For users specifying parameters individually, I do think that providing this very simple example routing ID (`funny-skunk-123`) is helpful, but we ideally also want to send users to a page with more detailed instructions on how to get their cluster's routing ID (e.g., "For instructions on how to obtain your cluster's routing ID, see [Connect to a CockroachDB Serverless (beta) Cluster](https://www.cockroachlabs.com/docs/cockroachcloud/connect-to-a-serverless-cluster.html?filters=connection-parameters)"). 

@ianjevans any objections to me updating `cockroachcloud/connect-to-a-serverless-cluster.html` (and possibly other pages) to use `<routing-id>` instead of `<cluster-name>-<tenant-id>` in this PR? Or would you prefer to wait until after the Console changes have been made?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538#-:-Ms_wg840M14PzZCGxma:bpjcfwr)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @emily-horing, @ianjevans, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 53 at r2](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqVnNud89nvvu1Br5bN-r2-53:-Ms_kofW13Zvpx2a0Cri:be6hd3q) ([raw file](https://github.com/cockroachdb/docs/blob/0d3e8266f0065bf6ccdd74935deccc51ca7404af/v21.2/connection-reference.md#L53)):*
<details><summary><i>Previously, kernfeld-cockroach (Paul Kernfeld) wrote…</i></summary><blockquote>

What I was quite unsuccessfully attempting to articulate here is that, in an ideal world, we wouldn't even need to document the format of CockroachDB connection strings for Serverless because the user would simply copy the appropriate connection string directly from the CockroachDB Cloud web UI. However, I am now realizing that this is not immediately possible because the web UI doesn't generate the correct connection strings. So, I guess the first step would be for the web UI to generate the correct connection string for a particular tool. Once that's ready to go, then we could un-document this.
</blockquote></details>

For serverless, that does sound like an ideal scenario.

In the future, this page should serve as a reference for Self-hosted and Dedicated as well. As a result, we will want to keep the explicit connection string format docs for users that can't simply cut-and-paste from the console.
___
*[v21.2/connection-reference.md, line 54 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2weN0-poNupsvnIk:-MsaTTFu1_BH_Lz75xJH:b-yeu8eq) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L54)):*

I did not know that was the preference. My fault. Updated with the cluster parameter in the connection string, and retested each connection string to verify that that format indeed was supported. 
</blockquote></details>

@bdarnell Circling back on this ... Is there a specific reason why we prefer the cluster options parameter instead of putting the cluster name in the database name? (I assume that it is just more accurate/expected to have the routing ID in a separate parameter, especially for users specifying the individual parameters).

@emily-horing 
___
*[v21.2/connection-reference.md, line 45 at r5](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqVnNud89nvvu1Br5bN-r5-45:-Ms_J_odBds_ZpQ8926E:b-jud4ko) ([raw file](https://github.com/cockroachdb/docs/blob/dee162c975bd7d73ad2517abfb68234c090c1b57/v21.2/connection-reference.md#L45)):*
>I'm concerned that this might encourage developers who don't know any better to paste their connection strings into their source code, which is sketch because it contains the password. What do you think of modifying these examples to instead read from an environment variable named DATABASE_URL?

That's a great idea. Done (and tested) for all except Django, which uses an individual configuration parameters.
___
*[v21.2/connection-reference.md, line 63 at r5](https://reviewable.io/reviews/cockroachdb/docs/12538#-MqVnNud89nvvu1Br5bN-r5-63:-Ms_lSKP3wqdttDw8ye7:b-54i8mn) ([raw file](https://github.com/cockroachdb/docs/blob/dee162c975bd7d73ad2517abfb68234c090c1b57/v21.2/connection-reference.md#L63)):*
<details><summary><i>Previously, kernfeld-cockroach (Paul Kernfeld) wrote…</i></summary><blockquote>

Do we want a `## Connect with Sequelize` here?
</blockquote></details>

Yes! Good catch. Also fixed this for TypeORM

ericharmeling (commented):
> Thanks for this! I learned a lot about connecting to CockroachDB!

Happy to hear that! TFTR!

> I think this information will be most useful for users who want to get started with an app. Can this go under "Get Started" instead?

I discussed this today with @emily-horing, who suggested the same idea. I will discuss offline with @ianjevans to iron out details.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538#-:-MsamJ-z5Ob1fK2ot4Ok:b-552gei)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @emily-horing, @ianjevans, @jordanlewis, @kernfeld-cockroach, and @rafiss)

bdarnell (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12538)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @bdarnell, @emily-horing, @ianjevans, @jordanlewis, @kernfeld-cockroach, and @rafiss)

___
*[v21.2/connection-reference.md, line 54 at r3](https://reviewable.io/reviews/cockroachdb/docs/12538#-Mqz2weN0-poNupsvnIk:-MsanVU89TWrPNH2lnx5:b-6vq5wb) ([raw file](https://github.com/cockroachdb/docs/blob/1845480bf4c6fd2375c8ec6b309bd972bbe3b84b/v21.2/connection-reference.md#L54)):*

@bdarnell Circling back on this ... Is there a specific reason why we prefer the cluster options parameter instead of putting the cluster name in the database name? (I assume that it is just more accurate/expected to have the routing ID in a separate parameter, especially for users specifying the individual parameters).

@emily-horing 
</blockquote></details>

Some tools get confused when the cluster name is added to the database name. I don't remember the specifics but for example I think one of them didn't like the fact that the database name in the connection string did not appear in SHOW DATABASES.

ianjevans:
> Previously, ericharmeling (Eric Harmeling) wrote…
> Some tools get confused when the cluster name is added to the database name. I don't remember the specifics but for example I think one of them didn't like the fact that the database name in the connection string did not appear in SHOW DATABASES.

Beekeeper Studio only partially works with Serverless clusters when connecting using the routing ID prefix. You can connect to a single database, and the other databases are listed in their UI, but you can't change databases. You have to disconnect, change the database name parameter, and reconnect. You can't use a connection string, as the `options` query parameter is ignored even if you modify it to use `cockroachdb://` instead of `postgresql://`.

Navikat Premium doesn't work at all even though it appears to connect successfully to the cluster using the routing ID prefix (as reported in the community Slack channel, and I tested it out). No databases or schema objects show up in the UI.

ericharmeling:
I think that I've addressed everyone's initial feedback. 

Here are some of the biggest changes that I've made in the last several commits:

- Filled out the Dedicated connection tab with tested connection string formats. Dedicated connection strings seem to be a lot more straightforward. I think I'm going to leave self-hosted out of this PR for now, so we can get this out ASAP.
- Moved this page from "Reference" to "Develop"/"Connect to CockroachDB", following the recent [docs nav restructure](https://github.com/cockroachdb/docs/pull/12567).
- Updated docs to use `routing-id` instead of `cluster-name.tenant-id`, and added more context on "routing ID" terminology.

Please let me know if there are any outstanding comments/concerns. If not, I'll move this to writer review early next week (Monday evening or Tuesday morning).

Note: Since returning from PTO 1/4/22, I was unable to get Sequelize connection strings to work at all, I believe due to https://github.com/cockroachdb/sequelize-cockroachdb/issues/141 / https://github.com/cockroachdb/cockroach/issues/74057.

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
+
+For a list of community-supported third-party tools, see [Third-Party Tools Supported by the Community](community-tooling.html). CockroachDB supports both native drivers and the PostgreSQL wire protocol. Most client drivers and ORM frameworks connect to CockroachDB like they connect to PostgreSQL.
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="serverless">{{ site.data.products.serverless-plan }}</button>
+  <button class="filter-button page-level" data-scope="dedicated">{{ site.data.products.dedicated }}</button>
+</div>
+
+<div class="filters clearfix">
+<button class="filter-button page-level" data-scope="js-ts">JavaScript/TypeScript</button>
```
ianjevans, line 19: Indentation is off on the first `<button>` tag.

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
 
-For a reference that lists all of the supported cluster connection parameters, see [Connection Parameters][connection_params].
+Parameter | Description
+----------|------------
+`<username>`  | The [SQL user](authorization.html#sql-users) connecting to the cluster.
+`<password>`  | The password for the SQL user connecting to the cluster.
+`<host>`  | The host on which the CockroachDB node is running.
+`<port>`  | The port at which the CockroachDB node is listening.
+`<database>`  | The name of the (existing) database.
+`<routing-id>`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 228:
```suggestion
`<routing-id>`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

_includes/cockroachcloud/sql-connection-string-free.md
```diff
@@ -1,29 +1,29 @@
...
+    cockroach sql --url "postgresql://<username>:<password>@<serverless-host>:26257/defaultdb?sslmode=verify-full&sslrootcert=$env:appdata/.postgresql/root.crt&options=--cluster=<routing-id>"
     ~~~
     </section>
 
     Where:
     - `<username>` is the SQL user. By default, this is your {{ site.data.products.db }} account username.
     - `<password>` is the password for the SQL user. The password will be shown only once in the **Connection info** dialog after creating the cluster.
     - `<serverless-host>` is the hostname of the {{ site.data.products.serverless }} cluster.
-    - `<cluster-name>-<tenant-id>` is the short name of your cluster plus the tenant ID. For example, `funny-skunk-123`. The `<tenant-id>` is used to identify your tenant cluster on a multitenant host.
+    - `<routing-id>` identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture). For example, `funny-skunk-123`.
```
ianjevans, line 26:
```suggestion
    - `<routing-id>` identifies your tenant cluster on a [multi-tenant host](architecture.html#architecture). For example, `funny-skunk-123`.
```

cockroachcloud/connect-to-a-serverless-cluster.md
```diff
@@ -109,28 +109,37 @@ To connect to your cluster with your application, use the connection string prov
...
-    "postgresql://<user>@<free-tier-host>.<region>.cockroachlabs.cloud:26257/defaultdb?sslmode=verify-full&sslrootcert=$env:appdata/.postgresql/root.crt&options=--cluster=<cluster-name>-<tenant-id>"
+    "postgresql://<username>:<password>@<serverless-host>:26257/defaultdb?sslmode=verify-full&sslrootcert=$env:appdata/.postgresql/root.crt&options=--cluster=<routing-id>"
     ~~~
     </section>
 
+    Where:
+    - `<username>` is the SQL user. By default, this is your {{ site.data.products.db }} account username.
+    - `<password>` is the password for the SQL user. The password will be shown only once in the **Connection info** dialog after creating the cluster.
+    - `<serverless-host>` is the hostname of the {{ site.data.products.serverless }} cluster.
+    - `<routing-id>` identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture). For example, `funny-skunk-123`.
```
ianjevans, line 134:
```suggestion
    - `<routing-id>` identifies your tenant cluster on a [multi-tenant host](architecture.html#architecture). For example, `funny-skunk-123`.
```

v21.2/build-a-csharp-app-with-cockroachdb.md
```diff
@@ -120,7 +120,7 @@ Where:
 
 - `{username}` and `{password}` specify the SQL username and password that you created earlier.
 - `{host-name}` is the name of the {{ site.data.products.db }} free tier host (e.g., `free-tier.gcp-us-central1.cockroachlabs.cloud`).
-- `{cluster_name}` is the name of your cluster.
+- `{routing-id}` is your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 123:
```suggestion
- `{routing-id}` is your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/build-a-python-app-with-cockroachdb.md
```diff
@@ -91,14 +91,14 @@ Before running the command, update the connection string as follows:
...
-"postgres://<username>:<password>@<globalhost>:26257/<cluster-name>.bank?sslmode=verify-full&sslrootcert=<certs_directory>/cc-ca.crt"
+"postgres://<username>:<password>@<globalhost>:26257/<routing-id>.bank?sslmode=verify-full&sslrootcert=<certs_directory>/cc-ca.crt"
 ~~~
 
 Before running the command, update the connection string that you copied [earlier](#set-up-your-cluster-connection) from the **Connection info** dialog as follows:
 
 - Replace `<username>` and `<password>` with the SQL username and password that you created earlier.
 - Replace `<globalhost>` with the name of the {{ site.data.products.serverless }} host (e.g., `free-tier.gcp-us-central1.cockroachlabs.cloud`).
-- Replace `<cluster-name>` with the name of your cluster.
+- Replace `<routing-id>` with your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 101:
```suggestion
- Replace `<routing-id>` with your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
+<div class="filter-content" markdown="1" data-scope="serverless">
+
+Parameter | Description
+----------|------------
+`{username}`  | The [SQL user](authorization.html#sql-users) connecting to the cluster.
+`{password}`  | The password for the SQL user connecting to the cluster.
+`{host}`  | The host on which the CockroachDB node is running.
+`{port}`  | The port at which the CockroachDB node is listening.
+`{database}`  | The name of the (existing) database.
+`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 434:
```suggestion
`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
```
ianjevans, line 621:
```suggestion
`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
```
ianjevans, line 747:
```suggestion
`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
 
-</section>
+Parameter | Description
+----------|------------
+`{username}`  | The [SQL user](authorization.html#sql-users) connecting to the cluster.
+`{password}`  | The password for the SQL user connecting to the cluster.
+`{host}`  | The host on which the CockroachDB node is running.
+`{port}`  | The port at which the CockroachDB node is listening.
+`{database}`  | The name of the (existing) database.
+`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 875:
```suggestion
`{routing-id}`  | Your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

v21.2/connection-parameters.md
```diff
@@ -96,7 +96,7 @@ CockroachDB supports the following `options` parameters. After the first `option
 
 Parameter | Description
 ----------|-------------
-`--cluster=<cluster name>` | Specifies the cluster name when connecting to [{{ site.data.products.serverless }}](connect-to-the-database-cockroachcloud.html#connect).
+`--cluster=<routing-id>` | Identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture). For example, `funny-skunk-123`.
```
ianjevans, line 99:
```suggestion
`--cluster=<routing-id>` | Identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture). For example, `funny-skunk-123`.
```

v21.2/hello-world-python-sqlalchemy.md
```diff
@@ -80,14 +80,14 @@ For example:
...
-cockroachdb://<username>:<password>@<globalhost>:26257/<cluster-name>.bank?sslmode=verify-full&sslrootcert=<certs_directory>/cc-ca.crt
+cockroachdb://<username>:<password>@<globalhost>:26257/<routing-id>.bank?sslmode=verify-full&sslrootcert=<certs_directory>/cc-ca.crt
 ~~~
 
 Where you update the connection string as follows:
 
 - Replace `<username>` and `<password>` with a SQL username and password.
 - Replace `<globalhost>` with the name of the {{ site.data.products.serverless }} host (e.g., `free-tier.gcp-us-central1.cockroachlabs.cloud`).
-- Replace `<cluster-name>` with the name of your cluster.
+- Replace `<routing-id>` with your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](https://www.cockroachlabs.com/docs/cockroachcloud/architecture.html#architecture).
```
ianjevans, line 90:
```suggestion
- Replace `<routing-id>` with your cluster's routing ID (e.g., `funky-skunk-123`). The routing ID identifies your tenant cluster on a [multi-tenant host](../cockroachcloud/architecture.html#architecture).
```

ianjevans (commented):
The links to the Cloud architecture doc were showing up as external. I added suggestions you can batch to fix the links.

emily-horing:
One thing that came up when I was reviewing our [latest mocks](https://www.figma.com/file/z8UDfW1DLrgoh88838kJJF/Onboarding-%3E-Connection?node-id=574%3A107165) with Rachel I wanted to bring up here: is it confusing that we're now calling it "routingID" but the actual text says --options **cluster**=<clustername+tenantid>? It seems pretty clear that this is the cluster, and there's no reference to routing here. 

Made me wonder how trivial it might be to add a second --options path that used **routing**=<clustername+tenantid>, which we could update to show in the UI without impacting any existing connections. cc @andy-kimball for thoughts on that.

@ericharmeling I'd be curious (assuming we keep everything the same) what you would suggest we call this in the parameters dialog in the UI. This would likely be used mostly to connect to tools like DataGrip (see [this issue](https://cockroachlabs.atlassian.net/browse/CC-5759) for an example Ian helped solve for). Feel free to leave comments in the Figma!

andy-kimball:
> Made me wonder how trivial it might be to add a second --options path that used routing=<clustername+tenantid>

That would be very easy to do. Can you file a JIRA issue, and we can find someone to do that work?

ericharmeling:
> Made me wonder how trivial it might be to add a second --options path that used **routing**=<clustername+tenantid>, which we could update to show in the UI without impacting any existing connections.

Great idea. We might want to document those changes (once implemented) as part of a follow-up docs PR though.

>I'd be curious (assuming we keep everything the same) what you would suggest we call this in the parameters dialog in the UI.

We could change the terminology in the docs to be "cluster routing ID" (with placeholder `cluster-routing-id`). Thoughts?

ericharmeling:
> The links to the Cloud architecture doc were showing up as external. I added suggestions you can batch to fix the links.

TFTR!

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
-  <button class="filter-button current">Local</button>
-  <a href="connect-to-the-database-cockroachcloud.html"><button class="filter-button">{{ site.data.products.db }}</button></a>
+This page documents the required connection configuration for [fully-supported third-party tools](third-party-database-tools.html).
+
+For a list of all supported cluster connection parameters, see the [`cockroach` Connection Parameters](connection-parameters.html).
+
+For a list of community-supported third-party tools, see [Third-Party Tools Supported by the Community](community-tooling.html). CockroachDB supports both native drivers and the PostgreSQL wire protocol. Most client drivers and ORM frameworks connect to CockroachDB like they connect to PostgreSQL.
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="serverless">{{ site.data.products.serverless-plan }}</button>
```
nickvigilante, line 14: Any particular reason we're using `{{ site.data.products.serverless-plan }}` and not `{{ site.data.products.serverless }}`?

nickvigilante (changes requested):
@ericharmeling Took a look at it and this was an impressive feat to accomplish. Just a few nitpicky things here and there and some broken links, but otherwise, LGTM!

v21.2/connect-to-the-database.md
ericharmeling: `{{ site.data.products.serverless-plan }}` maps to "CockroachDB Serverless" and `{{ site.data.products.serverless }}` maps to "CockroachDB Serverless (beta)"

While more accurate, "CockroachDB Serverless (beta)" is a little unnecessarily bulky to be throwing around every time we mention serverless (especially in things like tab titles). But that's just my opinion.

@ianjevans Is there a set convention for when to use `{{ site.data.products.serverless-plan }}` vs `{{ site.data.products.serverless}}`?

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
+  <button class="filter-button page-level" data-scope="psycopg">Psycopg2</button>
+  <button class="filter-button page-level" data-scope="sqlalchemy">SQLAlchemy</button>
+  <button class="filter-button page-level" data-scope="django">Django</button>
+</div>
+
+
+<div class="filter-content" markdown="1" data-scope="serverless">
+
+{{site.data.alerts.callout_info}}
+To connect to a {{ site.data.products.serverless-plan }} cluster from a Python application, you must have a valid CA certificate located at <code>~/.postgresql/root.crt</code>.<br>For instructions on downloading a CA certificate from the {{ site.data.products.db }} Console, see <a href="https://www.cockroachlabs.com/docs/v21.2/cockroachcloud/connect-to-a-serverless-cluster.html">Connect to a {{ site.data.products.serverless-plan }} Cluster</a>.
```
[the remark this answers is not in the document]
ericharmeling: Done.

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
+
+{% include copy-clipboard.html %}
+~~~
+jdbc:postgresql://{host}:{port}/{database}?user={username}&password={password}&sslmode=verify-full&sslrootcert={root-cert}
+~~~
+
+</div>
+
+{{site.data.alerts.callout_info}}
+To connect to CockroachDB with Hibernate, you must specify the [CockroachDB Hibernate dialect](https://www.cockroachlabs.com/docs/v21.2/install-client-drivers?filters=java#hibernate) in your `hibernate.cfg.xml` configuration file.
```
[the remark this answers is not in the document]
ericharmeling: Done.

v21.2/connect-to-the-database.md
```diff
@@ -1,147 +1,899 @@
...
+
+<div class="filter-content" markdown="1" data-scope="ruby">
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="ruby-pg">pg</button>
+  <button class="filter-button page-level" data-scope="activerecord">ActiveRecord</button>
+</div>
+
+{{site.data.alerts.callout_info}}
+To connect to a {{ site.data.products.serverless-plan }} cluster from a Ruby application, you must have a valid CA certificate located at <code>~/.postgresql/root.crt</code>.<br>For instructions on downloading a CA certificate from the {{ site.data.products.db }} Console, see <a href="https://www.cockroachlabs.com/docs/v21.2/cockroachcloud/connect-to-a-serverless-cluster.html">Connect to a {{ site.data.products.serverless-plan }} Cluster</a>.
```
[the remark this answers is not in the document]
ericharmeling: Done.

v21.2/developer-guide-overview.md
```diff
@@ -51,7 +52,7 @@ A growing number of popular third-party database tools offer full support for Co
 
 ## What's next?
 
-- [Start a Local Cluster](secure-a-cluster.html)
+- [Create a CockroachDB Serverless Cluster](../cockroachcloud/create-a-serverless-cluster.html)
```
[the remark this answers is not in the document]
ericharmeling: Done.

v21.2/multiregion-scale-application.md
```diff
@@ -67,7 +67,7 @@ Scaling application deployments in multiple regions can greatly improve latency
 
 For guidance on connecting to CockroachDB from an application deployment, see one of the following pages:
 
-- For connecting to managed, {{ site.data.products.db }} deployments, see [Connect to Your {{ site.data.products.dedicated }} Cluster](../cockroachcloud/connect-to-your-cluster.html) and [Connect to the Database ({{ site.data.products.dedicated }})](connect-to-the-database-cockroachcloud.html).
+- For connecting to managed, {{ site.data.products.db }} deployments, see [Connect to Your {{ site.data.products.dedicated }} Cluster](../cockroachcloud/connect-to-your-cluster.html) and [Connect to the Database ({{ site.data.products.dedicated }})](connect-to-the-database.html).
```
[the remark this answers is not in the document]
ericharmeling: Done.

ericharmeling (commented):
@nickvigilante TFTR! Much appreciated. I think I got to all your comments.

v21.2/connect-to-the-database.md
ianjevans: The intent was to separate the name of the product from what type of cluster the user would see in the Cloud UI. But in practice it has turned into a short name for the product.

The problem with that is eventually we're going to drop the "(beta)" from the product name and the two variables are going to be identical.

Now I'm wondering if we really want "(beta)" there at all.

nickvigilante: @ianjevans I think we want the beta there for now while the product is still in beta, but the nice thing is once it's out of beta, we just remove 7 characters from 1 file and it gets updated across the board.

Like I wrote in Slack, I think we should stick with just `serverless` and repurpose `serverless-plan` for what it was originally intended to be, which I still don't fully understand what that is 😄 

As bulky as it is, we will eventually move out of beta, and if for some reason the product name changes again, at least we'll be able to update it rather effortlessly. Regardless, we're going to have to do a universal find and replace of `serverless-plan` if we've been using them interchangeably, so probably keeping it for now will not make much of a difference in the end.

nickvigilante (approved):
@ericharmeling LGTM!

PR #12544, opened 2021-12-10
Product area category per page via front-matter

jseldess:
This PR is an experiment to see how we might use a
front-matter field to more easily bucket pages
for analytics.

Specifically, we set an href on the page based
on the category front-matter field and then use
JS to auto-click the link so it will get
tracked in Google Analytics as a default event.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -36,6 +36,15 @@ layout: default
       <div id="toc"></div>
    {% endif %}
 
+   {% if page.category %}
+      <a href="#{{ page.category }}" id=#doc-category" data-proofer-ignore></a>
+      <script>
+        $(document).ready(function(){
+          $('#doc-category').trigger('click');
+        });
+      </script>
+   {% endif %}
+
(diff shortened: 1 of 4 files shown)


jseldess:
Example: The `area:cloud` front-matter field on the Cloud quickstart gets generated as this on-page:
Screen Shot 2021-12-10 at 2 49 25 PM

mikeCRL:
@jseldess I like the use of front matter but I'm not clear on how the JS would affect GA, likely due to my limited knowledge of its custom JS features or our specific setup (e.g. if we're doing something similar with other data and know this works in another context). Please let me know if you can shed further light or recommend a doc page about the mechanism here. After doing a little reading, I thought we might use [Custom Dimensions in GA](https://support.google.com/analytics/answer/2709828) and it appears there are options for setting the value via JS as it's done on that page, or via a meta tag as described [here](https://marcusrauhut.com/passing-clean-meta-tag-data-to-google-analytics/) (albeit with less need for data transformation).

jseldess:
Closing in favor of https://github.com/cockroachdb/docs/pull/12878.

Issue #12237, opened 2021-11-10
Document Knex support


ericharmeling:
Eric Harmeling (ericharmeling) commented:

null

Issue #3137, opened 2018-05-14
Add KnexJs tutorial to Build an App 


knz:
Raphael 'kena' Poss (knz) commented:

> Latest version of KnexJS query builder is compatible with CockroachDB. A KnexJS tutorial would be a great addition to documentation.

From https://github.com/cockroachdb/cockroach/issues/15528

jseldess:
We have closed this issue because it is more than 3 years
old. If this issue is still relevant, please add a comment and re-open
the issue. Thank you for your contribution to CockroachDB docs!

PR #12563, opened 2021-12-13
Knex.js full support + sample app tutorial

ericharmeling:
Fixes https://github.com/cockroachdb/docs/issues/12237, https://github.com/cockroachdb/docs/issues/3137.


ericharmeling:
Gunna wait on https://github.com/cockroachdb/cockroach/pull/73773 to merge. 

Will remote_include the version number at https://github.com/cockroachdb/cockroach/pull/73773/files#diff-8785fb38f8b764db0a010bee31819ea9c28810714cf987e945064017f89db8cdR22

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Create a free cluster
+
+{% include cockroachcloud/quickstart/create-a-free-cluster.md %}
+
+### Set up your cluster connection
+
+1. Navigate to the cluster's **SQL Users** page, and create a new user, with a new password.
+
+1. Navigate to the **Cluster Overview page**, select **Connect**.
```
nickvigilante, line 34: @ericharmeling Does the name of the page include the word "page"? If not, probably unbold it.

Edit: looks like the tab is just called **Overview**.

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
```diff
@@ -0,0 +1,115 @@
...
+
+<section class="filter-content" markdown="1" data-scope="cockroachcloud">
+
+### Create a free cluster
+
+{% include cockroachcloud/quickstart/create-a-free-cluster.md %}
+
+### Set up your cluster connection
+
+1. Navigate to the cluster's **SQL Users** page, and create a new user, with a new password.
```
nickvigilante, line 32: I just tested this and it looks like a user is already created by default when spinning up a new serverless cluster. Do we need a second user on top of this?

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. Navigate to the **Cluster Overview page**, select **Connect**.
+
+1. Take note of the connection information. You'll use it to connect to the database later in this tutorial.
```
nickvigilante, line 36: I actually think we should move this into create-a-free-cluster.md and emphasize the bit about the password not being available, but that's outside the scope of this ticket. Is it fair to assume that users should already know to copy their connection information down prior to this point?

nickvigilante (commented):
@ericharmeling Just did a review and I think it looks great! I tested all the links and the nav bar, and everything is functional.

Two out of scope comments:

1. I think we should update the Quickstart steps for serverless to remind people to copy their username and password. Can I just create a new JIRA ticket to do that?
2. I love the tabs for the NodeJS/KnexJS at the top, but it seems clunky to copy the exact same `div` block across all 4 tabs. Should we think about modularizing it into a single include, in the event we have to add a new tab? Or would that be too complicated?

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
ericharmeling: (*I replaced the `### Set up your cluster connection` section with a new include. See my general comment about that new include file.*)

I don't think that we do, tbh. Removed this step in `_includes/v21.2/app/sample-setup.md`, `_includes/v21.1/app/sample-setup.md`, and `_includes/v20.2/app/sample-setup.md`.

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
ericharmeling: *(I replaced the ### Set up your cluster connection section with a new include. See my general comment about that new include file.)*

Great catch. Fixed in `_includes/v21.2/app/sample-setup.md`, `_includes/v21.1/app/sample-setup.md`, and `_includes/v20.2/app/sample-setup.md`.

v21.2/build-a-nodejs-app-with-cockroachdb-knexjs.md
ericharmeling: I think the connection information gathering is separate from `create-a-free-cluster.md` for a reason. Namely, not all connection flows are the same.

For example, `_includes/cockroachcloud/quickstart/set-up-your-cluster-connection.md` instructs the user to download a CA cert, but a local CA cert is actually not necessary for this particular driver. I created a new include file for that alternate workflow (`cockroachcloud/quickstart/set-up-your-cluster-connection-no-cert.md`) and used it here. That include also mentions the password bit.

ericharmeling (commented):
TFTR @nickvigilante !

See my inline response to your inline comments.

re: your other suggestions:

> I think we should update the Quickstart steps for serverless to remind people to copy their username and password. Can I just create a new JIRA ticket to do that?

That's a good idea - I think `_includes/cockroachcloud/quickstart/set-up-your-cluster-connection.md` already does this though. `_includes/cockroachcloud/quickstart/create-a-free-cluster.md` and `_includes/cockroachcloud/quickstart/set-up-your-cluster-connection.md` are separate includes because not all connection workflows require you to download the root CA. I created a new include file for that alternate workflow (`cockroachcloud/quickstart/set-up-your-cluster-connection-no-cert.md`). That include also mentions the password bit.

>I love the tabs for the NodeJS/KnexJS at the top, but it seems clunky to copy the exact same div block across all 4 tabs. Should we think about modularizing it into a single include, in the event we have to add a new tab? Or would that be too complicated?

That's a good idea, but I think that is out of the scope of this particular tutorial. Whatever modular solution we come up with will need to update the button dynamically to show the current page's tab. Could you open an issue?

nickvigilante (approved):
LGTM!

PR #12564, opened 2021-12-13
Clarify definition of join hint and provide example.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-502


rmloveland (approved):
LGTM

I think this section is the same in the v21.1 docs.  Do you mind adding these patches there too?  (since it's still a [supported version](https://www.cockroachlabs.com/docs/releases/release-support-policy.html))

PR #12605, opened 2021-12-16
Exclude old versions from link checking

jseldess:
- Tell google to ignore 19.2, which is unsupported
- Tell htmltest to ignore docs for all unsupported versions


ianjevans (approved):

ianjevans:
Shoot, this is hitting the ignored directory bug in htmltest. Trying to switch to a different htmltest repo that has the fix.

PR #12628, opened 2021-12-17
Explain apply joins.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-419


rytaft (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12628)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)

___
*[v21.2/joins.md, line 170 at r1](https://reviewable.io/reviews/cockroachdb/docs/12628#-Mt9EpmiDiSCI_2VvRMO:-Mt9EpmiDiSCI_2VvRMP:bgbcbir) ([raw file](https://github.com/cockroachdb/docs/blob/d92f118fbd580da965fef0f18100fd7a4ea87bf1/v21.2/joins.md#L170)):*
> ```Markdown
> Apply join is the operator that executes a lateral join if the optimizer is not able to de-correlate it (i.e., rewrite the query to use a regular join). Most of the time, the optimizer can de-correlate most queries. However, there are some cases where the optimizer cannot perform this rewrite, and `apply-join` would show up in the [`EXPLAIN`](explain.html#join-queries) output for the query. The optimizer also replaces correlated subqueries with apply joins, and therefore `apply-join` may appear in the `EXPLAIN` output even if `LATERAL` was not used.
> 
> Apply joins are inefficient because they must be executed one row at a time. The left side row must be used to construct the right side row, and only then can the execution engine determine if the two rows should be output by the join. This corresponds to an `O(n*m)` time complexity. Hash joins are much more efficient. A hash table is constructed using rows from the smaller side of the join, and then the larger side of the join is used to probe into the hash table using the ON conditions of the join. This corresponds to an `O(n+m)` time complexity.
> ```

Nice. Instead of only talking about hash joins, though, you might say:
"... This corresponds to an `O(n*m)` time complexity. Other types of joins supported by CockroachDB (e.g., hash join, merge join, and lookup join) are generally much more efficient. For example, with a hash join, a hash table is constructed ..."

Then you could consider including links to the sections that describe each of those join types.

rytaft (approved):
:lgtm: just one broken link

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12628#-:-MtAO8v9BRlICAA3vXud:b-go880q)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

___
*[v21.2/joins.md, line 172 at r2](https://reviewable.io/reviews/cockroachdb/docs/12628#-MtANckz1OAHnTxr_Cha:-MtANckz1OAHnTxr_Chb:b6ut1em) ([raw file](https://github.com/cockroachdb/docs/blob/e84b81ee17629d79e82d3b1af5d298d4e21474f1/v21.2/joins.md#L172)):*
> ```Markdown
> Apply joins are inefficient because they must be executed one row at a time. The left side row must be used to construct the right side row, and only then can the execution engine determine if the two rows should be output by the join. This corresponds to an `O(n*m)` time complexity.
> 
> Other types of joins supported by CockroachDB (e.g., [hash join](#hashs-join), [merge join](#merge-joins), and [lookup join](#lookup-joins)) are generally much more efficient. For example, with a hash join, a hash table is constructed using rows from the smaller side of the join, and then the larger side of the join is used to probe into the hash table using the `ON` conditions of the join. This corresponds to an `O(n+m)` time complexity.
> ```

`#hashs-join` -> `#hash-joins`

rmloveland (approved):
LGTM @stbof !  Sorry for delay.

Do you mind porting these changes to v21.1 docs as well?  According to our [release support policy](https://www.cockroachlabs.com/docs/releases/release-support-policy.html) we will be supporting v21.1 until November of this year, and my understanding from looking at the issue is that this applies to that version as well (I think v19.x is mentioned)

stbof:
> LGTM @stbof ! Sorry for delay.
> 
> Do you mind porting these changes to v21.1 docs as well? According to our [release support policy](https://www.cockroachlabs.com/docs/releases/release-support-policy.html) we will be supporting v21.1 until November of this year, and my understanding from looking at the issue is that this applies to that version as well (I think v19.x is mentioned)

Done.

PR #12685, opened 2022-01-07
DOC-1380: Added note about bounded staleness reads retry with nearst_only

nickvigilante:


v21.2/as-of-system-time.md
```diff
@@ -33,8 +33,8 @@ Format | Notes
...
+`with_min_timestamp(TIMESTAMPTZ, [nearest_only])` | <span class="version-tag">New in v21.2:</span> Defines a minimum [timestamp](timestamp.html) at which to perform the [bounded staleness read](follower-reads.html#bounded-staleness-reads). The actual timestamp of the read may be equal to or later than the provided timestamp, but cannot be before the provided timestamp. This is useful to request a read from nearby followers, if possible, while enforcing causality between an operation at some point in time and any dependent reads. This function accepts an optional `nearest_only` argument that will error if the reads cannot be serviced from a nearby replica. If a follower read can't be served when `nearest_only` is set to `true` and a schema change is made, CockroachDB will retry the read from before the schema change was made so long as the schema change occurred after the specified `TIMESTAMPTZ`.
```
otan, line 36: i'm trying to think of this makes the most sense to read from an external user.

Maybe `The query may fetch a result from an older schema if a read from the current cannot be satisified with a follower read.`?

otan (commented):
cc @nvanbenschoten in case you have opinions on wording

v21.2/as-of-system-time.md
nickvigilante: I'm happy to omit the detail or use your suggestion. Your call.

otan: Yeah let's omit it. The commit in actuality fixes a previous limitation related to schema changes. Now it's an internal detail that doesn't matter too much.

nvanbenschoten: +1 on omiting the detail. I think it will only serve to confuse readers.

nickvigilante: @otan @nvanbenschoten I omitted it from follower-reads and as-of-system-time, since it's the same message across all of it. Is the bit in transactions still good?

v21.2/transactions.md
```diff
@@ -125,6 +125,10 @@ transaction to retry, the client needs to be involved in retrying the whole
 transaction and so you should write your transactions to use
 [client-side intervention](#client-side-intervention).
 
+#### Bounded staleness reads
+
+In the event [bounded staleness reads](follower-reads.md#bounded-staleness-reads) are used along with either the `with_min_timestamp` or `with_max_staleness` functions and the `nearest_only` parameter set to `true`, the query will throw an error if it can't be served by a nearby replica. The exception to this is if a schema change is detected within the bounds of the time interval set in the function. In this scenario, CockroachDB will attempt the read again from before the schema change.
```
otan, line 130: I would be in favour of deleting the `The exception to this is if a schema change is detected within the bounds of the time interval set in the function. In this scenario, CockroachDB will attempt the read again from before the schema change.` bit. Otherwise, LGTM!

nickvigilante: Done! TFTR!

otan (approved):

nickvigilante:
@lnhsingh @gemma-shay Can you review this for me, please?

v21.2/transactions.md
```diff
@@ -125,6 +125,10 @@ transaction to retry, the client needs to be involved in retrying the whole
 transaction and so you should write your transactions to use
 [client-side intervention](#client-side-intervention).
 
+#### Bounded staleness reads
+
+In the event [bounded staleness reads](follower-reads.html#bounded-staleness-reads) are used along with either the `with_min_timestamp` or `with_max_staleness` functions and the `nearest_only` parameter set to `true`, the query will throw an error if it can't be served by a nearby replica.
```
gemma-shay, line 130: Only suggestion is linking the functions to this page: https://www.cockroachlabs.com/docs/v21.2/functions-and-operators#date-and-time-functions. LGTM otherwise!

gemma-shay (approved):

v21.2/transactions.md
nickvigilante: Added. Thank you!

lnhsingh:
I can't seem to use Reviewable since this PR is merged, but here's my feedback:

- nit: "functions" should be "function" since it's either the `with_min_timestamp` _function_ or the `with_max_staleness` _function_
- other nit: I think it should be "parameter _is_ set to `true`", for better flow (but feel free to disregard since this is mostly a style opinion)

PR #12709, opened 2022-01-10
retriable -> retryable

taroface:
Updated "retriable errors" to "retryable errors" for style consistency in the docs. The `SHOW TRACE` docs still have "retriable", but since this appears to be excerpted from the CLI output, I didn't change it.

(This update only touched on release notes, but a few were recent.)

Context: https://cockroachlabs.slack.com/archives/C01NFCXULTX/p1641830038000700


kathancox (approved):
Awesome, thanks for changing these Ryan! LGTM

PR #12719, opened 2022-01-12
DOC-2291: Searching now occurs within the site itself

nickvigilante:


nickvigilante:
This is a current feature of Algolia to limit the search cost.

PR #12728, opened 2022-01-13
Remove scale up from one node limitation

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-2305


emily-horing (changes requested):
i think we still want to include a limitation but it should say that scale down from multi node to 1 is not available (right now it says that scale up from 1 is not possible)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12728#-:-MtKKHO_A0ow3u8gxBYQ:b-s6lewp)* status: :shipit: complete! 0 of 0 LGTMs obtained

ianjevans (approved):

PR #12741, opened 2022-01-14
Update the crdb version on the "latest" cloud release notes

jseldess:
The notes are dated based on the date on which email
communication goes out. But since cloud isn't versioned,
the date isn't meaningful beyond that. Since the
"Latest Release Notes" page is the only place where
users can find out what version of crdb the cloud offerings
are running, it makes sense to keep the versions up to date.

I know there's broader work happening around the cloud release
not process that might resolve this issue. As part of that,
we should think of a way to communicate the current crdb
version for cloud that's release note date-agnostic.


gemma-shay (approved):
lgtm!

PR #12753, opened 2022-01-18
Remove incorrect example text from on_error option cdc

kathancox:
Quick fix removal of the example for `on_error` option. The example here actually is a retryable error, so `on_error` would not actually help in that scenario.


taroface (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12753#-:-Mtn-Kt8E2Xbtzka-CFm:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained

PR #12769, opened 2022-01-19
Don't reference source archives for building v21.2

rickystewart:
Building from source archives has been broken since v21.2.0, and from
v22.1.0 onward with the Bazel migration there won't be source archives
at all. Cloning the upstream repo and building from the clone is the
most correct and future-proof way to build from source.

### DIFF
--- a/v21.2/install-cockroachdb-linux.md
+++ b/v21.2/install-cockroachdb-linux.md
@@ -159,65 +159,7 @@ true
 
 <div id="build-from-source-linux" class="install-option">
   <h3>Build from Source</h3>
-  <ol>
-    <li>
-      <p>Install the following prerequisites, as necessary:</p>
-
-      <table>
-        <tr>
-          <td>C++ compiler</td>
-          <td>Must support C++ 11. GCC prior to 6.0 does not work due to <a href="https://gcc.gnu.org/bugzilla/show_bug.cgi?id=48891">this issue</a>. On macOS, Xcode should suffice.</td>
-        </tr>
-        <tr>
-          <td>Go</td>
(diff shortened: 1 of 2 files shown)


v21.2/install-cockroachdb-linux.md
```diff
@@ -158,6 +158,7 @@ true
 
 <div id="build-from-source-linux" class="install-option">
   <h3>Build from Source</h3>
+  <p>Consult the <a href="https://wiki.crdb.io/wiki/spaces/CRDB/pages/181338446/Getting+and+building+CockroachDB+from+source">Wiki</a> for specific, up-to-date information.</p>
```
jseldess, line 161: Given this change, rather than duplicate instructions, I think we should keep this sentence only and remove the rest. I'd suggest revising this sentence slightly to:

```
<p>See the <a href="https://wiki.crdb.io/wiki/spaces/CRDB/pages/181338446/Getting+and+building+CockroachDB+from+source">public wiki</a> for guidance.</p>
```

v21.2/install-cockroachdb-mac.md
```diff
@@ -173,6 +173,7 @@ true
 
 <div id="build-from-source" class="install-option">
   <h3>Build from source</h3>
+  <p>Consult the <a href="https://wiki.crdb.io/wiki/spaces/CRDB/pages/181338446/Getting+and+building+CockroachDB+from+source">Wiki</a> for specific, up-to-date information.</p>
```
jseldess, line 176: Same as above.

jseldess (changes requested):
Sorry for the delay, @rickystewart. LGTM overall. Just have a suggestion to take it a bit further.

jseldess (approved):
LGTM. Thanks, @rickystewart!

PR #12780, opened 2022-01-20
New home for style guide.

stbof:


mikeCRL:
For posterity: This is better, because we can't collaborate via git/PRs with the GitHub wiki. It will also be more visible to contributors.

PR #12784, opened 2022-01-20
DOC-987: Added more information on unique_rowid() to SERIAL

nickvigilante:
Fixes DOC-987

### DIFF
--- a/v21.2/serial.md
+++ b/v21.2/serial.md
@@ -4,13 +4,7 @@ summary: The SERIAL pseudo-type produces integer values automatically.
 toc: true
 ---
 
-The `SERIAL` pseudo [data type](data-types.html) is a keyword that can
-be used *in lieu* of a real data type when defining table columns. It
-is approximately equivalent to using an [integer type](int.html) with
-a [`DEFAULT` expression](default-value.html) that generates different
-values every time it is evaluated. This default expression in turn
-ensures that inserts that do not specify this column will receive an
-automatically generated value instead of `NULL`.
+The `SERIAL` pseudo [data type](data-types.html) is a keyword that can be used *in lieu* of a real data type when defining table columns. It is approximately equivalent to using an [integer type](int.html) with a [`DEFAULT` expression](default-value.html) that generates different values every time it is evaluated. This default expression in turn ensures that inserts that do not specify this column will receive an automatically generated value instead of `NULL`.
 
 {{site.data.alerts.callout_info}}
 `SERIAL` is provided only for compatibility with PostgreSQL. New applications should use real data types and a suitable `DEFAULT` expression.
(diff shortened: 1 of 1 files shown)


nickvigilante (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12784)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lin-crl and @lnhsingh)

___
*[v21.2/serial.md, line 239 at r1](https://reviewable.io/reviews/cockroachdb/docs/12784#-MvZNIBXAJjIV0aE-tyk:-MvZYJIRB7tsor3W0aRV:b-yxji42) ([raw file](https://github.com/cockroachdb/docs/blob/8c278458ba70e67f3545c4f6bdac38c5ba8ab1f3/v21.2/serial.md#L239)):*

Adding a drive-by review: The information hierarchy here seems off for a couple of reasons - 

1. Usually, the Examples section(s) are the last one the page before See Also.
2. I don't think `unique_rowid()` should be a H2, when other H2s are "Modes of operation", "Examples", etc. I think it makes more sense to have it nested under "Modes of operation" and not make the user jump all the way down to the bottom of the page for this info.

That being said, the bottom of the existing page ("Examples", "Auto-incrementing is not always sequential", "Additional examples") can be cleaned up to make more sense as well (I know this is legacy content and you did not write it). Let's chat offline about it.
</blockquote></details>

I fixed 1 and 2, so that Examples is at the end of the page.

The "Additional examples" section was less of an example on how to use `SERIAL` and more of an example of how auto-incrementing is not always sequential, so I renamed that header to "Additional scenarios".

lnhsingh (commented):
One more nit and then LGTM. Feel free to publish

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12784#-:-Mx-joOmDrdLc9xczcip:bh8b18x)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lin-crl, @lnhsingh, and @nickvigilante)

___
*[v21.2/serial.md, line 179 at r2](https://reviewable.io/reviews/cockroachdb/docs/12784#-Mx-jj8O8FshXYcFV2V-:-Mx-jj8O8FshXYcFV2V0:b-35xsj4) ([raw file](https://github.com/cockroachdb/docs/blob/7a4d6bd53d31b2eafbe95408d3a5c89746e2ba35/v21.2/serial.md#L179)):*
> ```Markdown
> If this happens, CockroachDB cannot guarantee whether `x < y` or `x > y`. Even though the transactions "clearly" occurred one "after" the other, perhaps there was a clock skew between the two nodes and the system time of the second node is set earlier than the first node.
> 
> ## Examples
> ```

Extreme nit: The heading here should probably be "Example" since there's only one right now

nickvigilante (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12784)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lin-crl and @lnhsingh)

___
*[v21.2/serial.md, line 179 at r2](https://reviewable.io/reviews/cockroachdb/docs/12784#-Mx-jj8O8FshXYcFV2V-:-Mx-smP36D8vcxnjgEd3:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/7a4d6bd53d31b2eafbe95408d3a5c89746e2ba35/v21.2/serial.md#L179)):*

Extreme nit: The heading here should probably be "Example" since there's only one right now
</blockquote></details>

Done.

PR #12787, opened 2022-01-20
Update multiregion migration instructions

rmloveland:
Fixes DOC-2357

Geo-partitioned leaseholders were incorrectly mapped to Regional By
Table tables with Region survival goals in our 'Migrate to Multi-region
SQL' docs.

They are now mapped to Regional By Row tables w/ Region survival goals.

Applied to both v21.1 and v21.2 docs.


awoods187 (approved):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12787)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm)

rmloveland:
Thanks for the review @awoods187 !  

@ajstorm since you already commented in the chat to the effect that this should be RBR and this makes that update, I'm gonna merge this now to avoid any further user confusion

ajstorm (commented):
Coming late to this.  LGTM too.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12787#-:-MtxQ9q0BkWFNOYS3H_A:bdj8ou3)* status: :shipit: complete! 0 of 0 LGTMs obtained

PR #12803, opened 2022-01-21
Fixed more broken links during nightly build

nickvigilante:


.htmltest.yml
```diff
@@ -13,6 +13,11 @@ IgnoreURLs:
   - "https://crates.io/*"
   - "https://docs.pipenv.org/*"
   - "https://arxiv.org/*"
+  - "https://www.microsoft.com/*"
+  - "https://v1-18.docs.kubernetes.io/*"
```
ianjevans, line 17: This will get rid of the errors, but the links are still broken.

nickvigilante: You're right, but @taroface said he's going to do work into decoupling the K8s docs with our docs.

Ryan, is there a JIRA ticket for that? If not, do you want me to create one?

PR #12703, opened 2022-01-10
New package versions.

stbof:

PR #12805, opened 2022-01-21
DOC-2375 Remove Gemfile.lock from the repo, add it to .gitignore.

ianjevans:
This PR fixes DOC-2375, removing explicit `Gemfile.lock` dependencies from the repo.

Writers will have to generate `Gemfile.lock` again after this PR merges. You can do this with the following commands:

```
make clean
make vendor
make cockroachdb
```


stbof:
Would it make sense to also add `.bundle/config`? See https://github.com/cockroachdb/docs/pull/12703/files.

stbof:
> Would it make sense to also remove `.bundle/config` from repo and add to .gitignore? See https://github.com/cockroachdb/docs/pull/12703/files.

Nevermind. I think the value in config is a mistake.

stbof (approved):

PR #12817, opened 2022-01-24
DOC-1535: Remove limitations of computed columns in FKs

nickvigilante:
Fixes DOC-1535

### DIFF
--- a/v21.1/computed-columns.md
+++ b/v21.1/computed-columns.md
@@ -19,7 +19,6 @@ Computed columns are especially useful when used with [`JSONB`](jsonb.html) colu
 Computed columns:
 
 - Cannot be used to generate other computed columns.
-- Cannot reference a [foreign key](foreign-key.html).
 - Behave like any other column, with the exception that they cannot be written to directly.
 - Are mutually exclusive with [`DEFAULT`](default-value.html).
 
--- a/v21.1/foreign-key.md
+++ b/v21.1/foreign-key.md
@@ -26,7 +26,7 @@ To read more about how foreign keys work, see our [What is a Foreign Key? (With 
 **Foreign Key Columns**
 
 - Foreign key columns must use their referenced column's [type](data-types.html).
-- A foreign key column cannot be a [computed column](computed-columns.html).
+- A foreign key column cannot be a virtual [computed column](computed-columns.html), but it can be a stored computed column.
 - A single column can have multiple foreign key constraints. For an example, see [Add multiple foreign key constraints to a single column](#add-multiple-foreign-key-constraints-to-a-single-column).
 
 **Referenced Columns**
--- a/v21.2/computed-columns.md
+++ b/v21.2/computed-columns.md
@@ -27,7 +27,6 @@ Computed columns are especially useful when used with [`JSONB`](jsonb.html) colu
 Computed columns:
 
 - Cannot be used to generate other computed columns.
-- Cannot reference a [foreign key](foreign-key.html).
 - Behave like any other column, with the exception that they cannot be written to directly.
 - Are mutually exclusive with [`DEFAULT`](default-value.html) and [`ON UPDATE`](create-table.html#on-update-expressions) expressions.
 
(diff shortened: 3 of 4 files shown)


v21.1/computed-columns.md
```diff
@@ -28,7 +27,6 @@ Virtual computed columns:
 - Are not stored in the table's primary index.
 - Are recomputed as the column data in the expression changes.
 - Cannot be used as part of a primary key, `FAMILY` definition, in `CHECK` constraints, or `FOREIGN KEY` constraints.
-- Cannot be a [foreign key](foreign-key.html) reference.
```
mgartner, line 31: This restriction still remains. It's quite confusing and I had to double check myself. Here are examples of the differences:

A computed column expression can now reference a foreign key column, so this is allowed:

```sql
defaultdb> CREATE TABLE parent (p INT PRIMARY KEY);
CREATE TABLE

defaultdb> CREATE TABLE child (c INT PRIMARY KEY, c_p INT REFERENCES parent (p), v INT AS (c_p + 10) VIRTUAL);
CREATE TABLE
```

But this is still NOT allowed:

```sql
defaultdb> CREATE TABLE parent (p INT PRIMARY KEY, v INT AS (p + 10) VIRTUAL, UNIQUE INDEX (v))
CREATE TABLE

defaultdb> CREATE TABLE child (c INT PRIMARY KEY, v INT REFERENCES parent (v))
ERROR: unimplemented: virtual column "v" cannot be referenced by a foreign key
SQLSTATE: 0A000
HINT: You have attempted to use a feature that is not yet implemented.
See: https://go.crdb.dev/issue-v/59671/v22.1
```

v21.1/foreign-key.md
```diff
@@ -26,7 +26,6 @@ To read more about how foreign keys work, see our [What is a Foreign Key? (With
 **Foreign Key Columns**
 
 - Foreign key columns must use their referenced column's [type](data-types.html).
-- A foreign key column cannot be a [computed column](computed-columns.html).
```
mgartner, line 29: This should read: "A foreign key column cannot be a [virtual computed column](...)" (see my second example above). You might want to clarify that a foreign key column **can** be a stored computed column, though.

v21.2/computed-columns.md
```diff
@@ -36,7 +35,6 @@ Virtual computed columns:
 - Are not stored in the table's primary index.
 - Are recomputed as the column data in the expression changes.
 - Cannot be used as part of a primary key, `FAMILY` definition, in `CHECK` constraints, or `FOREIGN KEY` constraints.
-- Cannot be a [foreign key](foreign-key.html) reference.
```
mgartner, line 39: ditto: leave as-is

v21.2/foreign-key.md
```diff
@@ -26,7 +26,6 @@ To read more about how foreign keys work, see our [What is a Foreign Key? (With
 **Foreign Key Columns**
 
 - Foreign key columns must use their referenced column's [type](data-types.html).
-- A foreign key column cannot be a [computed column](computed-columns.html).
```
mgartner, line 29: ditto: see my previous comment for the same line in 21.1 docs

mgartner (changes requested):

v21.1/computed-columns.md
nickvigilante: Oof, that's tricky. Took me a while to wrap my head around it, but I think I understand. Basically, you can create a computed column that references a foreign key, but you can't create a foreign key that references a computed column.

Done.

v21.2/computed-columns.md
nickvigilante: Done.

v21.2/foreign-key.md
nickvigilante: Done.

mgartner (approved):

lnhsingh (commented):
LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12817#-:-MuDguyi18qRuqsMxjex:bb6ywmx)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @gemma-shay)

gemma-shay (approved):

PR #12849, opened 2022-01-26
Use single quotes everywhere we UPDATE region name

rmloveland:
Follow-up to cockroachdb/docs#10926


ericharmeling (approved):
LGTM! (though IDK where else we might be doing this...)

rmloveland:
Eric Harmeling ***@***.***> writes:

> @ericharmeling approved this pull request.
>
> LGTM! (though IDK where else we might be doing this...)

Best just not to think about it until we have

https://github.com/cockroachdb/docs/issues/7067

:-)

PR #12850, opened 2022-01-26
DOC-2209 Fast follow for nav restructure

ianjevans:
This PR resolves some of the issues we noticed after new doc navigation structure was merged. Specifically:

- Releases has been moved from Reference to the top level.
- New prominent link to Core binary/release note download from docs landing page.
- Consistent naming of overview topics.
- Expanded names of Hello World and Simple CRUD topics in the nav.


stbof (approved):
This looks great! Huge improvement.

jseldess (approved):
LGTM with a preference.

ianjevans:
> LGTM with a preference.

I don't see a review comment. What is your preference?

_includes/v21.2/sidebar-data/get-started.json
```diff
@@ -22,26 +22,26 @@
           "title": "Hello World",
           "items": [
             {
-              "title": "Go",
+              "title": "Build a Hello World App Using Go",
```
jseldess, line 25: Comment got lost somehow. Or I probably forgot to add it. 

I'm not sure why we're spelling this out in each case. It makes the sidenav very cluttered:

Screen Shot 2022-01-31 at 12 03 04 PM

Perhaps we're trying to ensure that the sidenav link matches the page title, but that's also not true in these cases.

stbof:
> Sorry, @ianjevans. Just added the missing comment.

@ianjevans was responding to my comment that simply saying `Using Ruby`, `Using Python` and requiring the user to figure out the context wasn't friendly either. I don't know the solution. The current nav titles are a bit long.

jseldess:
Yeah, I'd prefer to just go back to the language names for now. We might want to explore another approach where we just have one "page" with another level of language toggles at the top. In reality, it would be many pages, but there would just be one "Build a Hello World App" link and one "Build a Simple CRUD App" link.

PR #12859, opened 2022-01-27
Add NOVIEWACTIVITYREDACTED and NOSQLLOGIN roles

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2315
https://cockroachlabs.atlassian.net/browse/DOC-2359

Also alphabetized role options.


v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
+`VIEWACTIVITYREDACTED`/`NOVIEWACTIVITYREDACTED` | Allow or disallow a role to see other roles' queries and sessions using `SHOW STATEMENTS`, `SHOW SESSIONS`, and the Statements and Transactions pages in the DB Console, and the usage of [statements diagnostics bundle](ui-statements-page.html#diagnostics), which can contain PII information. Without this role option, the `SHOW` commands only show the role's own data and the DB Console pages are unavailable. It is possible to have both `VIEWACTIVITY` and `VIEWACTIVITYREDACTED`, and `VIEWACTIVITYREDACTED` takes precedence on restrictions. <br><br>By default, the role option is set to `NOVIEWACTIVITYREDACTED` for all non-admin roles.
```
kevin-v-ngo, line 60: This is great! Can we clarify that the prevention of using the statements diagnostics bundle is only in the console? The user can still run EXPLAIN ANALYZE debug via SQL CLI. 

"and the usage of statements diagnostics bundle in the DB Console"

kevin-v-ngo (commented):
Small clarification from me. @thtruo can you take a look as well? NOSQLLOGIN looked fine to me.

kevin-v-ngo:
Adding @maryliag to help take a look as well

v21.2/alter-role.md
abhinavg6: Another thing is that we are using both SQLNOLOGIN and NOSQLLOGIN in the first point. Should use the available option uniformly. 

Also, maybe worth adding a line or two on how that works with LOGIN / NOLOGIN. @thtruo - comments on that?

_includes/common/sql/statement-bundle-warning.md
```diff
@@ -1,2 +1,4 @@
-{{site.data.alerts.callout_danger}} Statement bundles can contain unredacted user data including histograms and placeholders that contain real data samples, statements that have real data, and database schema which could be sensitive. Be careful when generating and sharing statement bundles.
+{{site.data.alerts.callout_danger}}Statement bundles can contain unredacted user data including histograms and placeholders that contain real data samples, statements that have real data which can contain PII information, and database schema that could be sensitive. Be careful when generating and sharing statement bundles.
+
+<br><br>To allow or disallow a role from seeing other roles' <a href="{{ link_prefix }}ui-statements-page.html#diagnostics">statements diagnostics bundles</a>, set the <code>NOVIEWACTIVITYREDACTED</code> <a href="{{ link_prefix }}create-role.html#role-options">role option</a>.
```
maryliag, line 3: This role is not about seeing other roles bundles or not, is about seeing bundles in general, independent if it was created by another role. 
Also, the NO___ option is usually more to remove the role option, so a user should grant or revoke the VIEWACTIVITYREDACTED role.

The sentence should be "To allow or disallow a role from seeing statements diagnostics bundle, set the VIEWACIVITYREDACTED role option"

v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
-`VALID UNTIL timestamp`/`VALID UNTIL NULL` | The date and time (in the [`timestamp`](timestamp.html) format) after which the password is not valid.
+`SQLLOGIN`/`SQLNOLOGIN` | Allow or disallow a role to log in using the SQL CLI with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `SQLNOLOGIN` prevents the role from logging in using the SQL CLI with any authentication method. Logins to the DB Console are permitted with `NOSQLLOGIN` set.
+`VALID UNTIL` |  The date and time (in the [`timestamp`](timestamp.html) format) after which the [password](#parameters) is not valid.
+`VIEWACTIVITY`/`NOVIEWACTIVITY` | Allow or disallow a role to see other roles' [queries](show-statements.html) and [sessions](show-sessions.html) using `SHOW STATEMENTS`, `SHOW SESSIONS`, and the [Statements](ui-statements-page.html) and [Transactions](ui-transactions-page.html)  pages in the DB Console. Without this role option, the `SHOW` commands only show the role's own data and the DB Console pages are unavailable. <br><br>By default, the role option is set to `NOVIEWACTIVITY` for all non-admin roles.
```
maryliag, line 59: One thing worth mentioning is that "Without this role... DB Console pages are unavailable", but if the user has the new VIEWACTIVITYREDACTED role, those pages will be available.

v21.2/alter-role.md
maryliag, line 60: Similar to my other comment, if the user doesn't have VIEWACTIVITYREDACTED but have VIEWACTIVITY they will be able to see DB Console pages.

thtruo (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12859)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @abhinavg6, @kevin-v-ngo, @maryliag, @stbof, and @thtruo)

___
*[v21.2/alter-role.md, line 57 at r3](https://reviewable.io/reviews/cockroachdb/docs/12859#-MuST95wCNNorADv0q9k:-MuST95wCNNorADv0q9l:bjyjiuv) ([raw file](https://github.com/cockroachdb/docs/blob/f52ae3fd3cf1d52e93256cb31913e76e5e1e0466/v21.2/alter-role.md#L57)):*
> ```Markdown
> `LOGIN`/`NOLOGIN` | Allow or disallow a role to log in with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `NOLOGIN` prevents the role from logging in using any authentication method.
> `MODIFYCLUSTERSETTING`/`NOMODIFYCLUSTERSETTING` | Allow or disallow the role to modify the [cluster settings](cluster-settings.html) with the `sql.defaults` prefix. <br><br>By default, the role option is set to `NOMODIFYCLUSTERSETTING` for all non-admin roles.
> `SQLLOGIN`/`SQLNOLOGIN` | Allow or disallow a role to log in using the SQL CLI with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `SQLNOLOGIN` prevents the role from logging in using the SQL CLI with any authentication method. Logins to the DB Console are permitted with `NOSQLLOGIN` set.
> ```

Suggestion: could we rephrase along the lines of "Setting the role option to `NOSQLLOGIN` prevents the role from logging into the SQL CLI while retaining DB Console login."



_Code quote:_
```Markdown
Setting the role option to `SQLNOLOGIN` prevents the role from logging in using the SQL CLI with any authentication method. Logins to the DB Console are permitted with `NOSQLLOGIN` set.
```
___
*[v21.2/alter-role.md, line 57 at r3](https://reviewable.io/reviews/cockroachdb/docs/12859#-MuSTbpYA-Q-MHBNSCDq:-MuSTbpYA-Q-MHBNSCDr:b-fv1cci) ([raw file](https://github.com/cockroachdb/docs/blob/f52ae3fd3cf1d52e93256cb31913e76e5e1e0466/v21.2/alter-role.md#L57)):*
> ```Markdown
> `LOGIN`/`NOLOGIN` | Allow or disallow a role to log in with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `NOLOGIN` prevents the role from logging in using any authentication method.
> `MODIFYCLUSTERSETTING`/`NOMODIFYCLUSTERSETTING` | Allow or disallow the role to modify the [cluster settings](cluster-settings.html) with the `sql.defaults` prefix. <br><br>By default, the role option is set to `NOMODIFYCLUSTERSETTING` for all non-admin roles.
> `SQLLOGIN`/`SQLNOLOGIN` | Allow or disallow a role to log in using the SQL CLI with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `SQLNOLOGIN` prevents the role from logging in using the SQL CLI with any authentication method. Logins to the DB Console are permitted with `NOSQLLOGIN` set.
> ```

Good callout - the role option should be relabeled as`NOSQLLOGIN`. 

I'd also recommend we add an additional sentence of two along the lines of how: "It is possible to have both `NOSQLLOGIN` and `LOGIN` set for a role and `NOSQLLOGIN` takes precedence on restrictions." 



_Code quote:_
```Markdown
SQLNOLOGIN
```

thtruo:
cc-ing @dhartunian to take a look as well

v21.2/cockroach-sql.md
```diff
@@ -13,6 +13,10 @@ To exit the interactive shell, use `\q`, `quit`, `exit`, or `ctrl-d`.
 If you want to experiment with CockroachDB SQL but do not have a cluster already running, you can use the [`cockroach demo`](cockroach-demo.html) command to open a shell to a temporary, in-memory cluster.
 {{site.data.alerts.end}}
 
+## Prerequisites
+
+The [role option of the user](create-role.html#role-options) logging in must be `SQLLOGIN`. If the user's role option has been set to `SQLNOLOGIN`, the user cannot log in using the SQL CLI with any authentication method.
```
dhartunian, line 18: `SQLNOLOGIN` should be `NOSQLLOGIN` (the negative is always prepended)

I'm a bit worried about the wording here implying that the user has to *take action* before being able to log into SQL. Is it obvious that they have this role by default? (under the hood, adding the `NOSQLLOGIN` role introduces blocking but the "absence" of this block is taken to affirm that they have `SQLLOGIN`, does that make sense? `LOGIN` and `NOLOGIN` work the same way, the user doesn't have to do anything to have `LOGIN` by default.)

stbof: Thanks for the catch. I missed that one.

So does the prereq have to be 


The [role option of the user](create-role.html#role-options) logging in must be `SQLLOGIN`, which is granted by default. If the user's role option has been set to `SQLNOLOGIN`, the user cannot log in using the SQL CLI with any authentication method.

dhartunian: Correct. It appears we did not have this section previously despite the fact that `LOGIN` and `NOLOGIN` have the same effect on SQL logins so I'm also wondering if this section is necessary. If it is, it should also reference the `LOGIN` role option, I think.

stbof: I think we should have a section describing required permissions on any function that won't work if the user doesn't have it. IMO, the fact that we didn't have it before is a shortcoming.

dhartunian: 👍🏻  LGTM thx!

v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
+`VIEWACTIVITYREDACTED`/`NOVIEWACTIVITYREDACTED` | Allow or disallow a role to see other roles' queries and sessions using `SHOW STATEMENTS`, `SHOW SESSIONS`, and the Statements and Transactions pages in the DB Console, and the usage of [statements diagnostics bundle](ui-statements-page.html#diagnostics), which can contain PII information, in the DB Console. Without this role option, the `SHOW` commands only show the role's own data and the DB Console pages are unavailable. It is possible to have both `VIEWACTIVITY` and `VIEWACTIVITYREDACTED`, and `VIEWACTIVITYREDACTED` takes precedence on restrictions. If the user has `VIEWACTIVITY` but doesn't have `VIEWACTIVITYREDACTED`, they will be able to see DB Console pages. <br><br>By default, the role option is set to `NOVIEWACTIVITYREDACTED` for all non-admin roles.
```
maryliag, line 60: I think we need to make a little more clear that with the VIEWACTIVITYREDACTED role they can see the pages on the Console, but they won't see the diagnostic. Maybe something on the lines of:
"Allow or disallow ... Transactions pages in the DB Console. With this role, the user won't have access to the usage of statements diagnostics bundle, which can contain PII information, in the DB Console."

stbof:
@maryliag Please have another look. I'm finding these roles incredibly confusing, especially since the doc before would say things like "With this role ..." and I had no idea what role it was referring to: `VIEWACTIVITY` or `NOVIEWACTIVITY`.

v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
+`VIEWACTIVITYREDACTED`/`NOVIEWACTIVITYREDACTED` | Allow or disallow a role to see other roles' queries and sessions using `SHOW STATEMENTS`, `SHOW SESSIONS`, and the Statements and Transactions pages and the [statements diagnostics bundle](ui-statements-page.html#diagnostics), which can contain PII information, in the DB Console. With `NOVIEWACTIVITYREDACTED`, the user won't have access to the usage of statements diagnostics bundle in the DB Console. It is possible to have both `NOVIEWACTIVITY` and `NOVIEWACTIVITYREDACTED`, and `NOVIEWACTIVITYREDACTED` takes precedence on restrictions. If the user has `VIEWACTIVITY` but doesn't have `VIEWACTIVITYREDACTED`, they will be able to see DB Console pages. <br><br>By default, the role option is set to `NOVIEWACTIVITYREDACTED` for all non-admin roles.
```
maryliag, line 60: You added the NO there assuming by mistake.
- With `VIEWACTIVITYREDACTED`, the user won't have access to the...
- It is possible to have both `VIEWACTIVITY` and `VIEWACTIVITYREDACTED`, and `VIEWACTIVITYREDACTED` takes precedence on restrictions

stbof: It was deliberate. To me the roles should be named in a parallel structure. If you can't do something the role should start with NO. I find it super confusing how these roles work.

maryliag: `VIEWACTIVITYREDACTED`/`NOVIEWACTIVITYREDACTED` | Allow or disallow a role to see other roles' queries and sessions using `SHOW STATEMENTS`, `SHOW SESSIONS`, and the Statements and Transactions pages in the DB Console. With `VIEWACTIVITYREDACTED`, the user won't have access to the usage of statements diagnostics bundle, which can contain PII information, in the DB Console. It is possible to have both `VIEWACTIVITY` and `VIEWACTIVITYREDACTED`, and `VIEWACTIVITYREDACTED` takes precedence on restrictions. If the user has `VIEWACTIVITY` but doesn't have `VIEWACTIVITYREDACTED`, they will be able to see DB Console pages and have access to the usage of statements diagnostics bundle. <br><br>By default, the role option is set to `NOVIEWACTIVITYREDACTED` for all non-admin roles.

maryliag (approved):
Great! Looks good to me!

v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
-`CANCELQUERY`/`NOCANCELQUERY` | Allow or disallow the role to cancel [queries](cancel-query.html) and [sessions](cancel-session.html) of other roles. Without this role option, roles can only cancel their own queries and sessions. Even with this role option, non-admins cannot cancel admin queries or sessions. This option should usually be combined with `VIEWACTIVITY` so that the role can view other roles' query and session information. <br><br>By default, the role option is set to `NOCANCELQUERY` for all non-admin roles.
+`CREATEROLE`/`NOCREATEROLE` |  Allow or disallow the new role to create, [alter](alter-role.html), and [drop](drop-role.html) other non-admin roles. <br><br>By default, the role option is set to `NOCREATEROLE` for all non-admin users.
```
rmloveland, line 54: in future I think we should move alphabetization to a separate style PR instead of mixed with new content.  kinda hard for me to understand what's changing here

v21.2/alter-role.md
```diff
@@ -35,7 +35,7 @@ table td:first-child {
 
 Parameter | Description
 ----------|-------------
-`role_name` | Specify the name of the role that you want to alter.
+`role_name` | The name of the role to alter.
```
rmloveland, line 38: Would also be nice to hold to style PR in future.  Different brain required for this than reading new content.  Harder to switch between them

v21.2/alter-role.md
```diff
@@ -45,18 +45,19 @@ Parameter | Description
...
+`LOGIN`/`NOLOGIN` | Allow or disallow a role to log in with one of the [client authentication methods](authentication.html#client-authentication). Setting the role option to `NOLOGIN` prevents the role from logging in using any authentication method.
 `MODIFYCLUSTERSETTING`/`NOMODIFYCLUSTERSETTING` | Allow or disallow the role to modify the [cluster settings](cluster-settings.html) with the `sql.defaults` prefix. <br><br>By default, the role option is set to `NOMODIFYCLUSTERSETTING` for all non-admin roles.
-`PASSWORD password`/`PASSWORD NULL` | Let the role [authenticate their access to a secure cluster](authentication.html#client-authentication) using this new password. Passwords should be entered as a [string literal](sql-constants.html#string-literals). For compatibility with PostgreSQL, a password can also be entered as an identifier. <br><br>To prevent a role from using [password authentication](authentication.html#client-authentication) and to mandate [certificate-based client authentication](authentication.html#client-authentication), [set the password as `NULL`](#prevent-a-role-from-using-password-authentication).
```
rmloveland, line 58: so is the PASSWORD role gone now?  Not seeing it here.  Its removal doesn't seem to be mentioned in the PR description or in either of the linked docs tickets (at a glance anyway)

v21.2/create-role.md
```diff
@@ -7,9 +7,9 @@ toc: true
...
 
 {{site.data.alerts.callout_info}}
- <code>CREATE ROLE</code> is no longer an Enterprise feature and is now freely available in the core version of CockroachDB. Also, since the keywords `ROLE` and `USER` can now be used interchangeably in SQL statements for enhanced Postgres compatibility.
+The keywords `ROLE` and `USER` can be used interchangeably in SQL statements for enhanced Postgres compatibility.
 
- `CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the [`NOLOGIN`](#parameters) option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the [`LOGIN`](#parameters) option to achieve the same result as `CREATE USER`.
+`CREATE USER` is equivalent to the statement `CREATE ROLE`, with one exception: `CREATE ROLE` sets the [`NOLOGIN`](#parameters) option by default, preventing the new role from being used to log in to the database. You can use `CREATE ROLE` and specify the [`LOGIN`](#parameters) option to achieve the same result as `CREATE USER`.
```
rmloveland, line 12: Same as above re: easier to review without these style changes in a new content PR

v21.2/create-role.md
```diff
@@ -45,20 +45,34 @@ Unless a role is a member of the admin role, additional [privileges](#parameters
 
 ## Parameters
 
-| Parameter | Description |
+<style>
```
rmloveland, line 48: yay markdown is great :-D

rmloveland: do we have a docs site issue filed for this?  looks like a one-off fix due to length of the phrase in the Parameter column right?

(I'm not asking for any change in this PR but I think we should have this solved in the site design b/c this is gonna be fragile)

v21.2/create-user.md
```diff
@@ -41,18 +43,29 @@ table td:first-child {
...
+`WITH role_option` | Apply a [role option](#role-options) to the role.
+
+## Role options
+
+Role option | Description
+------------|--------------
+`CANCELQUERY`/`NOCANCELQUERY` | Allow or disallow the role to cancel [queries](cancel-query.html) and [sessions](cancel-session.html) of other roles. Without this role option, roles can only cancel their own queries and sessions. Even with this role option, non-admins cannot cancel admin queries or sessions. This option should usually be combined with `VIEWACTIVITY` so that the role can view other roles' query and session information. <br><br>By default, the role option is set to `NOCANCELQUERY` for all non-admin roles.
```
rmloveland, line 54: (nothing to do, but: same re: adding new content and the sorting - looks all new and is pretty hard to see what changed)

rmloveland (approved):
LGTM - don't claim to understand all the changed content but trusting in the eng/PM review.

v21.2/alter-role.md
stbof: PASSWORD isn't a role. It's a parameter. I think it was mistaken to have it as a role option.

v21.2/create-role.md
stbof: I asked @mikeCRL about alphabetizing and he approved.

PR #12860, opened 2022-01-27
Update aggregation interval and stats flushing.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2424

### DIFF
--- a/_includes/common/ui/statistics.md
+++ b/_includes/common/ui/statistics.md
@@ -6,9 +6,11 @@
   {% assign page_prefix = "ui-" %}
 {% endif %}
 
-Statistics are aggregated once an hour and organized by [Aggregation Interval](#statement-fingerprint-properties). Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
+Statistics aggregation is controlled by the [aggregation interval](#statement-fingerprint-properties) property. Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
 
-Statistics are persisted in statistics tables in the [`crdb_internal`]({{ link_prefix }}crdb-internal.html) system catalog. The default retention period of the statistics tables is based on the number of rows up to 10 million records. When this threshold is reached, the oldest records are deleted.
+The `sql.stats.aggregation.interval` [cluster setting](cluster-settings.html) controls the size of the aggregation interval, with a default value of 1 hour.
+
+Aggregated statistics are flushed from memory to statistics tables in the [`crdb_internal`]({{ link_prefix }}crdb-internal.html) system catalog every 10 minutes. The default retention period of the statistics tables is based on the number of rows up to 10 million records. When this threshold is reached, the oldest records are deleted.
 
 To reset SQL statistics in the DB Console UI and `crdb_internal` system catalog, click **clear SQL stats**.
 


_includes/common/ui/statistics.md
```diff
@@ -6,9 +6,11 @@
   {% assign page_prefix = "ui-" %}
 {% endif %}
 
-Statistics are aggregated once an hour and organized by [Aggregation Interval](#statement-fingerprint-properties). Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
+Statistics aggregation is controlled by the [aggregation interval](#statement-fingerprint-properties) property. Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
 
-Statistics are persisted in statistics tables in the [`crdb_internal`]({{ link_prefix }}crdb-internal.html) system catalog. The default retention period of the statistics tables is based on the number of rows up to 10 million records. When this threshold is reached, the oldest records are deleted.
+The `sql.stats.aggregation.interval` [cluster setting]({{ link_prefix }}cluster-settings.html) controls the size of the aggregation interval, with a default value of 1 hour.
```
kevin-v-ngo, line 11: Can we remove the 'sql.stats.aggregation.interval' mention? We haven't set any guardrails (or even best practices) on this setting and don't want users configuring this. Thanks!

kevin-v-ngo: We should even consider adding a note on the cluster-settings page that it's not recommended to alter the flush or aggregation interval settings.

stbof: > Can we remove the 'sql.stats.aggregation.interval' mention? We haven't set any guardrails (or even best practices) on this setting and don't want users configuring this. Thanks!

I can, but it's also in the release notes PR. Shall I remove it there too?

 The default SQL stats flush interval is now 10 minutes. A new cluster setting `sql.stats.aggregatinon.interval` controls the aggregation interval of [SQL stats](ui-statements-page.html#statement-statistics), with a default value of 1 hour. [#75524][#75524]

stbof: > We should even consider adding a note on the cluster-settings page that it's not recommended to alter the flush or aggregation interval settings.

What flush setting are you referring to? AFAIK we only have a new aggregation interval setting.

kevin-v-ngo: My mistake, I thought we introduced a new flush setting as well but it seems like we kept the old setting and decoupled the aggregation into another setting.

 https://www.cockroachlabs.com/docs/stable/cluster-settings.html. 

But yes, if we can remove it from the release notes, that would be great. Mainly we don't want to advertise these settings as users shouldn't really be changing them. They're exposed in case for potential cluster wide issues where we may need to increase they value.

stbof: > We should even consider adding a note on the cluster-settings page that it's not recommended to alter the flush or aggregation interval settings.

I can't edit the cluster-settings page because it's generated. If you want to hide it you have to ask engineering to set it to private I think. @maryliag?

maryliag: I have no idea how to do this. Do you have any example where this was done or who I can ask about it?

stbof: Investigating.

stbof: According to @rmloveland, 

Raphael on KV definitely does
Pretty sure Becca, Radu, Marcus on SQL Queries would know / have done it
Oliver, Adam, Rafi on SQL Experience almost certainly know

kevin-v-ngo: So the entire page is not editable unless there are code changes? For now, we can call this out on the statistics.md page to not update these two settings.

stbof: There is only one setting: `sql.stats.aggregation.interval`. The cluster settings page is generated from code so the change must come from engineering. Let's see what @maryliag can do first.

maryliag: To make the change to remove this it would mean pushing code, which would delay the release and I don't want to do that, so we have 2 options here:
1 - Leave as is, at least for this version, and remove on the following one. The major concern was that this value can't be more than `sql.stats.aggregation.interval`, which is already documented here, so it's not that bad to have this documented.
2 - I can generate the new file locally I send to you and maybe you can use that?

One thing I noticed is that the file has changed already, where you able to do it? Can you use the file with your changes?

stbof: 1) I have removed all documentation of `sql.stats.aggregation.interval` per @kevin-v-ngo's request. The only doc remaining is the generated one, which he as also requested be removed.

2) I did not change any generated files. Which one are you referring to.

maryliag: I was referring to the only one I can see on this review.
Which one is used here? https://www.cockroachlabs.com/docs/stable/cluster-settings.html

If the generated one is used for that case, can I generate a new one and send to you and you use that instead? If not, then we will keep as is for this release and remove on the next.

stbof: > I was referring to the only one I can see on this review. Which one is used here? https://www.cockroachlabs.com/docs/stable/cluster-settings.html
> 
> If the generated one is used for that case, can I generate a new one and send to you and you use that instead? If not, then we will keep as is for this release and remove on the next.

Those are the docs for the  21.2.4 release.

stbof: I don't touch the generated file `https://raw.githubusercontent.com/cockroachdb/cockroach/release-21.2/docs/generated/settings/settings.html` or the file `https://github.com/cockroachdb/docs/blob/master/v21.2/cluster-settings.md` that references the generated file.

maryliag: Yes, but that page will be updated once we publish 21.2.5, correct? I think Kevin's concern is that we don't want to show the new setting on that page, so the question is: 
1 - which file is used to generate that page?
2 - Can you change that file or use something that I can generate locally and send to you?

maryliag: Okay, if you can't change, we should keep as is. 
cc @kevin-v-ngo  I don't want to delay the release, I will remove for the following one

maryliag (approved):

_includes/common/ui/statistics.md
stbof: I think https://raw.githubusercontent.com/cockroachdb/cockroach/release-21.2/docs/generated/settings/settings.html is already updated. It's generated from the latest version of the code, which has the new property.

No I can't use a different file. The whole point is that we import the generated file.

The only workaround is for me to put a note in the documentation to not change the value of `sql.stats.aggregation.interval`.

stbof: @kevin-v-ngo 

Since I no longer discuss `sql.stats.aggregation.interval` in the docs at all, and the _only_ place it appears is in the cluster-settings page, please tell me how you want to spin this in the regular docs.

kevin-v-ngo: Yeah, that's fine. Main thing is to do our best to not highlight this cluster setting since we haven't validated with values lower than the default. And the aggregation interval must be greater than the flush interval (we don't enforce this) or there would be odd stats values. 

I think there are other cluster settings as well on this page which we don't encourage users to configure so we can leave it there for now. 

There's a warning on the page already

> Many cluster settings are intended for tuning CockroachDB internals. Before changing these settings, we strongly encourage you to discuss your goals with Cockroach Labs; otherwise, you use them at your own risk.

maryliag: Update: Ian mention that I can backport the change to remove this and the Docs will pickup without the need of a new version being creating, so I created the PR with the change :)

kevin-v-ngo (approved):
Looks great!

PR #12864, opened 2022-01-27
Fix syntax error for node-postgres connection docs

kernfeld-cockroach:
Our current connection docs contain a "set literal," which is not
supported in JavaScript and causes a syntax error. A solution is
to pass the connection string directly into the client without
wrapping it in a JavaScript object.


ericharmeling (approved):
LGTM - thanks!

PR #12802, opened 2022-01-21
Update mapping from zcfgs to multiregion SQL

rmloveland:
Fixes DOC-2357 (all the way this time)

Summary of changes:

- Update to use the correct mapping from zone configs to multiregion SQL
  statements (again)

- Except this time, move the mapping into an include file, and use the
  include file in both places the mapping is referenced

PR #12866, opened 2022-01-27
Add (last?) correct mapping of geo* pattern to RBR

rmloveland:
Fixes DOC-2357

Summary of changes:

- Fixed the (hopefully) last incorrect mapping from the old skool
  geo-partitioned leaseholder pattern to the new REGIONAL BY ROW table
  locality + ZONE survival goal on the 'Migrate to Multi-region SQL'
  page.

- Previously, we were incorrectly referring to a mapping from
  geo-partitioned leaseholders to REGIONAL BY TABLE locality + ZONE
  survival.

### DIFF
--- a/v21.1/migrate-to-multiregion-sql.md
+++ b/v21.1/migrate-to-multiregion-sql.md
@@ -131,7 +131,7 @@ If you applied the [geo-partitioned leaseholders][geo_leaseholders] pattern, the
     ALTER INDEX users_last_name_index PARTITION BY NOTHING;
     ~~~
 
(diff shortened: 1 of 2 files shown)


rmloveland:
Eric this is the last reference (I hope) that needs updating on this page.

Skipping technical review since we already confirmed this info via reviews in #12787 and #12802

Really hope this is the last one (I actually read the page this time) 😓

ericharmeling (approved):
LGTM!

rmloveland:
woohoo! thanks!

PR #12869, opened 2022-01-27
Conceptual section how changefeed works

kathancox:
Fixes [DOC-1509](https://cockroachlabs.atlassian.net/browse/DOC-1509)

Draft for docs on how a changefeed works — conceptual.


kathancox:
For review: see https://deploy-preview-12869--cockroachdb-docs.netlify.app/docs/v21.2/change-data-capture-overview

v21.2/change-data-capture-overview.md
```diff
@@ -23,6 +23,26 @@ The main feature of CDC is the changefeed, which targets an allowlist of tables,
 
 See [Ordering Guarantees](use-changefeeds.html#ordering-guarantees) for detail on CockroachDB's at-least-once-delivery-guarantee as well as explanation on how rows are emitted.
 
+## How does a changefeed work?
+
+When a changefeed is started on a node, that node becomes the aggregator node for the changefeed job (CRDB Node 2 in the diagram). The aggregator node acts as an administrator: keeping track of all other nodes during job execution and the changefeed work as it completes. In the unlikely event that the changefeed’s aggregator node were to fail during the job, the aggregator role will move to a different node and the changefeed will restart from the last checkpoint.
```
miretskiy, line 28: > When a changefeed is started on a node, that node becomes the aggregator node for the changefeed job (CRDB Node 2 in the diagram). The aggregator node acts as an administrator

Internally, we have "change aggregator" processors; and a single change frontier processor. Aggregators run on all nodes, while frontier is a single node.  Perhaps we can "rename" frontier processor to be "changefeed coordinator"?  And use "change aggregator" to mean "the thing that runs on each node"?

Also, note re picture: THe coordinator node runs "both" processors (so it functions as both an aggregator and a coordinator).

miretskiy: > The aggregator node acts as an administrator: keeping track of all other nodes during job execution and the changefeed work as it completes. In the unlikely event that the changefeed’s aggregator node were to fail during the job, the aggregator role will move to a different node

I think it is important to emphasize that this functionality is applicable only to enterprise changefeeds.  Core style changefeeds do not even have a job.  The caller is expected to handle any hickups -- so, if the node where coordinator is running is restarted, the SQL connection used in core style changefeeds will return an error -- the caller is responsible for restarts.

Another thing to note is that the picture shows enterprise style changefeed -- (kafka).  So, perhaps this text makes sense, provided that we make it clear that this whole thing is for "enterprise".

It might also make sense to have another picture -- very similar to the one you have, but specifically for core style changefeed where each aggregator sends data back to the coordinator, and the coordinator sends the data back to the caller through SQL conneciton.

v21.2/change-data-capture-overview.md
```diff
@@ -23,6 +23,26 @@ The main feature of CDC is the changefeed, which targets an allowlist of tables,
...
+
+The changefeed job will run across nodes in the cluster to access changed data in the watched table. Typically, the [leaseholder](architecture/replication-layer.html#leases) for a particular range (or the range’s replica) determines which node emits the changefeed data.
+
+As each node is working, it sends back checkpoint progress to the aggregator node that gathers this information to update the high-water mark timestamp. The high-water mark acts as a checkpoint for the changefeed’s job progress, and guarantees that all changes before (or at) the timestamp have been emitted. If restarted, the changefeed will send duplicate messages from the high-water mark timestamp. For more detail on when messages are emitted, see [Ordering Guarantees](use-changefeeds.html#ordering-guarantees).
```
miretskiy, line 32: We should make it clear that this checkpointing and restarting is enterprise only feature.

v21.2/change-data-capture-overview.md
```diff
@@ -23,6 +23,26 @@ The main feature of CDC is the changefeed, which targets an allowlist of tables,
...
+
+<img src="{{ 'images/v21.2/changefeed-structure.png' | relative_url }}" alt="Changefeed process in a 3-node cluster" style="max-width:100%" />
+
+With the [`resolved` option](create-changefeed.html#resolved-option) specified when a changefeed is started, the aggregator node will send the resolved timestamp (i.e., the high-water mark) to each endpoint in the sink. For example, when using [Kafka](changefeed-sinks.html#kafka) this will be sent as a message to each partition; for [cloud storage](changefeed-sinks.html#cloud-storage-sink), this will be emitted as a resolved timestamp file.
```
miretskiy, line 36: The resolved option, btw, is the only mechanism how core style changefeeds would know how far along they are if an error occurred.

kathancox:
Thank @miretskiy and apologies for the delay in getting to your comments. I had to put this down for release work. 

I have adapted this section to be Enterprise changefeeds only and I've opened a separate issue to cover Core changefeeds. I have included your feedback on the coordinator/aggregator naming and simplified the diagram. 

Would love another review from you!

amruss:
I really like the diagrams

v21.2/change-data-capture-overview.md
```diff
@@ -25,6 +25,26 @@ The main feature of CDC is the changefeed, which targets an allowlist of tables,
...
+
+The changefeed job will run across nodes in the cluster to access changed data in the watched table. Typically, the [leaseholder](architecture/replication-layer.html#leases) for a particular range (or the range’s replica) determines which node emits the changefeed data.
+
+Each node uses its aggregator processors to send back checkpoint progress to the coordinator that gathers this information to update the high-water mark timestamp. The high-water mark acts as a checkpoint for the changefeed’s job progress, and guarantees that all changes before (or at) the timestamp have been emitted. If restarted, the changefeed will send duplicate messages from the high-water mark timestamp. For more detail on when messages are emitted, see [Ordering Guarantees](use-changefeeds.html#ordering-guarantees).
```
amruss, line 34: "If restarted, the changefeed will send duplicate messages from the high-water mark timestamp."

i think this sentence could be worded a bit more clearly, maybe "If restarted, the changefeed will send duplicate messages starting at the high watermark time to the current time"

amruss (approved):
One minor nit but besides that LGTM

andf-crl (approved):
Great overview of a complicated topic! I added some comments regarding specific terms and a few language nits, but otherwise LGTM! Nothing blocking.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12869#-:-N4bst2J4j5eDvg0uRJL:bcbn9uf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @kathancox and @miretskiy)

___
*[`v22.1/change-data-capture-overview.md` line 30 at r4](https://reviewable.io/reviews/cockroachdb/docs/12869#-N4bjT0p9hPLHHfDTDRi:-N4bjT0p9hPLHHfDTDRj:b-1489y5) ([raw file](https://github.com/cockroachdb/docs/blob/3108f19a7bacd398a03615a5527ade6e42a31445/v22.1/change-data-capture-overview.md#L30)):*
> ```Markdown
> ## How does an Enterprise changefeed work?
> 
> When an {{ site.data.products.enterprise }} changefeed is started on a node, that node becomes the _coordinator_ for the changefeed job (**Node 2** in the diagram). The coordinator node acts as an administrator: keeping track of all other nodes during job execution and the changefeed work as it completes. In the unlikely event that the changefeed’s coordinating node were to fail during the job, that role will move to a different node and the changefeed will restart from the last checkpoint.
> ```

Clarification: `checkpoint` is introduced here, and used three times in this section. Is there somewhere we could link to define this? Is this tied to the `resolved` timestamp i.e. high-water mark? If so, consider calling it the high-water mark instead and linking to (I think) [here](https://www.cockroachlabs.com/docs/stable/create-changefeed.html#start-a-new-changefeed-where-another-ended). We don't seem to use `checkpoint` much elsewhere (mostly just [here](https://www.cockroachlabs.com/docs/stable/use-changefeeds.html#messages), which is how I mentally connected it to `resolved`).

Note that the above link contains its own link to high-water mark, which takes us [here](https://www.cockroachlabs.com/docs/v22.1/monitor-and-debug-changefeeds#monitor-a-changefeed), which does not appear to contain this term.

nit: Abbey calls this _coordinator_ node an _aggregator_ in her [blog on the subject](https://cockroachlabs.com/blog/change-data-capture-for-performance-durability/#how-does-a-changefeed-work). Ideally we use the same term. I prefer _coordinator_ because it appears to be what our tooling calls it, as shown [here](https://www.cockroachlabs.com/docs/stable/create-changefeed.html#start-a-new-changefeed-where-another-ended) (i.e. `coordinator_id`).

EDIT: reading further, and following your later [link](https://deploy-preview-12869--cockroachdb-docs.netlify.app/docs/v22.1/create-changefeed#resolved-option), I see that `resolved` and `high-water mark` are related -- but discrete -- concepts. Defining checkpoint here, or describing its relationship to the high-water mark, would likely have been enough to get me thinking in the right way in paragraph 1.
___
*[`v22.1/change-data-capture-overview.md` line 32 at r4](https://reviewable.io/reviews/cockroachdb/docs/12869#-N4bpRfg-vGOvamxjWBW:-N4bpRfg-vGOvamxjWBX:b-sdvjtl) ([raw file](https://github.com/cockroachdb/docs/blob/3108f19a7bacd398a03615a5527ade6e42a31445/v22.1/change-data-capture-overview.md#L32)):*
> ```Markdown
> When an {{ site.data.products.enterprise }} changefeed is started on a node, that node becomes the _coordinator_ for the changefeed job (**Node 2** in the diagram). The coordinator node acts as an administrator: keeping track of all other nodes during job execution and the changefeed work as it completes. In the unlikely event that the changefeed’s coordinating node were to fail during the job, that role will move to a different node and the changefeed will restart from the last checkpoint.
> 
> The changefeed job will run across nodes in the cluster to access changed data in the watched table. Typically, the [leaseholder](architecture/replication-layer.html#leases) for a particular range (or the range’s replica) determines which node emits the changefeed data.
> ```

nit: `across nodes` --> `across all nodes`
___
*[`v22.1/change-data-capture-overview.md` line 34 at r4](https://reviewable.io/reviews/cockroachdb/docs/12869#-N4bmWDM9rn1ESp_QFL5:-N4blbNnAzJ-aZ6_PFbR:bvtswjj) ([raw file](https://github.com/cockroachdb/docs/blob/3108f19a7bacd398a03615a5527ade6e42a31445/v22.1/change-data-capture-overview.md#L34)):*
> ```Markdown
> The changefeed job will run across nodes in the cluster to access changed data in the watched table. Typically, the [leaseholder](architecture/replication-layer.html#leases) for a particular range (or the range’s replica) determines which node emits the changefeed data.
> 
> Each node uses its aggregator processors to send back checkpoint progress to the coordinator that gathers this information to update the high-water mark timestamp. The high-water mark acts as a checkpoint for the changefeed’s job progress, and guarantees that all changes before (or at) the timestamp have been emitted. If restarted, the changefeed will send duplicate messages starting at the high-water mark time to the current time. For more detail on when messages are emitted, see [Ordering Guarantees](use-changefeeds.html#ordering-guarantees).
> ```

nit:  `coordinator that gathers` --> `coordinator, which gathers`

Suggestion: What do you think of replacing `For more detail` sentence at the end of this paragraph with your final sentence in the `What is change data capture` section? I think that line would work perfectly here! (*at-least-once delivery guarantee* really helped me to understand the discussed "duplicate emitance" & high-water mark through current timestamp approach described in this paragraph).
___
*[`v22.1/change-data-capture-overview.md` line 38 at r4](https://reviewable.io/reviews/cockroachdb/docs/12869#-N4brI8m47bXOZFb4Dn8:-N4bqnMz6jEEnsNThxxr:bcg53xc) ([raw file](https://github.com/cockroachdb/docs/blob/3108f19a7bacd398a03615a5527ade6e42a31445/v22.1/change-data-capture-overview.md#L38)):*
Question: if `resolved` is unspecified, is the resolved timestamp sent at all? If not, perhaps this tiny change (`the` suggests must be present always from some source, `a` for not necessarily. Yes, this is a nit).

> the coordinator will send a resolved timestamp
___
*[`v22.1/change-data-capture-overview.md` line 40 at r4](https://reviewable.io/reviews/cockroachdb/docs/12869#-N4bqzQUBKXUCcUOt5fd:-N4bqzQUBKXUCcUOt5fe:b-gtybqq) ([raw file](https://github.com/cockroachdb/docs/blob/3108f19a7bacd398a03615a5527ade6e42a31445/v22.1/change-data-capture-overview.md#L40)):*
> ```Markdown
> With the [`resolved` option](create-changefeed.html#resolved-option) specified when a changefeed is started, the coordinator will send the resolved timestamp (i.e., the high-water mark) to each endpoint in the sink. For example, when using [Kafka](changefeed-sinks.html#kafka) this will be sent as a message to each partition; for [cloud storage](changefeed-sinks.html#cloud-storage-sink), this will be emitted as a resolved timestamp file.
> 
> As rows are updated, added, and deleted in the targeted table(s), the node sends the row changes through the [rangefeed mechanism](use-changefeeds.html#enable-rangefeeds) to the changefeed encoder, which encodes these changes into the [final message format](use-changefeeds.html#responses). The message is emitted from the encoder to the sink—it can emit to any endpoint in the sink. In the diagram example, this means that the messages can emit to any Kafka Broker.
> ```

These two paragraphs are very helpful and greatly clarify -- nice!

kathancox:
TFTRs all. @andf-crl I adjusted as per your feedback. And your point on the position of checkpoint and high-water mark led me to a better flow for the first couple of paragraphs. So thank you!

PR #12872, opened 2022-01-28
Added redirect from removed changefeed file

kathancox:
Added a redirect from the now removed stream data out of cockroachdb using changefeeds page for v21.2.


ianjevans (approved):

kathancox:
TFTR @ianjevans

PR #12873, opened 2022-01-28
Correction to BACKUP CHECKPOINT note on BACKUP

kathancox:
Fixes [DOC-2386](https://cockroachlabs.atlassian.net/browse/DOC-2386)

This is a small clarification to the current note on the backup page that specifies that the presence of the `BACKUP CHECKPOINT` file in the backup destination means that the job was not successful. That is not actually the case, as this file is not necessarily cleared up. 

Adjusted this note to the `BACKUP MANIFEST` file's presence indicating that the job was successful.


livlobo (approved):
Thanks for making all these small fixes, Kathryn. They all add up. LGTM!

taroface (approved):
LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12873#-:-Muq0tf42wKKuAi15fLm:b-85i648)* status: :shipit: complete! 0 of 0 LGTMs obtained

PR #12904, opened 2022-02-01
DOC-463: Added note about dropping unique indexes

nickvigilante:
Fixes DOC-463

- Added a note to clarify unique indexes created with `CREATE TABLE` cannot be dropped without the `CASCADE` flag

### DIFF
--- a/v20.2/drop-index.md
+++ b/v20.2/drop-index.md
@@ -37,6 +37,10 @@ The user must have the `CREATE` [privilege](authorization.html#assign-privileges
 
 ### Remove an index with no dependencies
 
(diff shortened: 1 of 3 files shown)


ajwerner (approved):
I hate this bug. One day we'll get around to fixing it. I don't know the policy for mentioning bugs. The bug here is that we treat such an index as a unique constraint internally. 

Otherwise, :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12904#-:-N42fg-t5RqTlimunR4n:bn7emox)* status: :shipit: complete! 1 of 0 LGTMs obtained

PR #12924, opened 2022-02-02
DOC-500: Add more details to `ALTER PARTITION`

nickvigilante:
Fixes DOC-500

### DIFF
--- a/_includes/v20.2/zone-configs/create-a-replication-zone-for-a-table-partition.md
+++ b/_includes/v20.2/zone-configs/create-a-replication-zone-for-a-table-partition.md
@@ -1,4 +1,6 @@
+{% unless include.hide-enterprise-warning == "true" %}
 {% include enterprise-feature.md %}
+{% endunless %}
 
 Once [partitions have been defined for a table or a secondary index](partition-by.html), to control replication for a partition, use `ALTER PARTITION <partition> OF INDEX <table@index> CONFIGURE ZONE`:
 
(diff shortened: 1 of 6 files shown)


lnhsingh (commented):
Style suggestions for tightening up the language of the definitions that apply to all versions. Feel free to merge once addressed

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12924#-:-MvZIGD65_Tf8nf3MDG4:bkuqe4m)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @lnhsingh and @nickvigilante)

___
*[v20.2/alter-partition.md, line 25 at r1](https://reviewable.io/reviews/cockroachdb/docs/12924#-MvZGQjZ2ZAokF1Uynwz:-MvZGQjZ2ZAokF1Uynx-:bd0q2fl) ([raw file](https://github.com/cockroachdb/docs/blob/19e75489666b4fd86ce2ed0b13a91b3425425e54/v20.2/alter-partition.md#L25)):*
> ```Markdown
>  Parameter | Description
> -----------+-------------
> `table_name` | The name of the [table](create-table.html) whose [replication zone configurations](configure-replication-zones.html) you want to change.
> ```

Style suggestion on the definitions in this table:

Instead of saying "The name of the [object] whose replication zone configs you want to change" I would suggest changing "whose" to "with the". By using "whose", you anthropomorphize the db object.

You could also make the end more concise and change "you want to change" to "to change" or "to modify".

So my suggestion would be: "The name of the table with the replication zone configurations to modify."

PR #12878, opened 2022-01-28
Categories for analytics

jseldess:

PR #12929, opened 2022-02-03
DOC-2500: Set analytics front-matter for Migrate docs

jseldess:
Follow-up to https://github.com/cockroachdb/docs/pull/12878.


jseldess:
@kathancox, I see an "Export Spatial Data", but not a general instructional page on exporting other forms of data (e.g., complimentary to the `EXPORT` reference page). Do we need that?

kathancox (approved):

kathancox:
Yes, we don't **_currently_** have a page for general export or import other than their sql reference pages (and for import the more specific migrate pages). I do need to break down the "Run Bulk Ops in Cloud" page, which currently is under Manage --> Backup/Restore, because this page contains import/export cloud examples. 

So, with all that being said, we need a space under Migrate for that content, which could also be a space for general export/import content.

PR #12959, opened 2022-02-07
DOC-547: Added note about name resolution by search path

nickvigilante:
Fixes DOC-547

- Named resolution by search path will try to resolve `pg_extension` prior to `public` schema.


otan (approved):

gemma-shay (approved):
lgtm!

PR #12714, opened 2022-01-11
Move hello worlds under Get Started

jseldess:
- Move hello worlds under Get Started and remove the sidenav reference to the outdated
hello-world-example-apps.html page.
- Update links to point to example-apps instead.

PR #12981, opened 2022-02-09
Removed Hello World Example Apps overview page

ericharmeling:
Follow-up on https://cockroachdb.slack.com/archives/CP4D9LD5F/p1644353610688869.


ericharmeling:
Follow-up on https://github.com/cockroachdb/docs/pull/12714

jseldess:
Thanks for catching this, @ericharmeling. Since we're removing a page, shouldn't we add it to our server-side redirects, just in case? https://github.com/cockroachdb/docs/blob/master/_redirects#L339

ericharmeling:
> Since we're removing a page, shouldn't we add it to our server-side redirects, just in case? https://github.com/cockroachdb/docs/blob/master/_redirects#L339

Yes. Good catch. Opened https://github.com/cockroachdb/docs/pull/13038, which also resolves https://cockroachlabs.atlassian.net/browse/DOC-2310.

PR #12991, opened 2022-02-09
Serverless function best practices

ericharmeling:
Fixes https://cockroachlabs.atlassian.net/browse/DOC-2283.

Related to https://github.com/cockroachlabs/examples-aws-lambda/pull/3.

### DIFF
--- a/v21.2/serverless-function-best-practices.md
+++ b/v21.2/serverless-function-best-practices.md
@@ -1,0 +1,116 @@
+---
+title: Serverless Function Best Practices
+summary: Best practices for optimizing the performance of serverless functions (e.g., AWS Lambda functions or Google Cloud Functions) that connect to CockroachDB.
+toc: true
+referral_id: docs_serverless_functions
+docs_area: get_started
+---
+
+This page provides best practices for optimizing the performance of serverless functions (e.g., [AWS Lambda functions](https://aws.amazon.com/lambda) and [Google Cloud Functions](https://cloud.google.com/functions)) that connect to CockroachDB.
+
+## Use connection pools
+
+Use connection pools to manage the lifecycle of database connections established by serverless functions. Connection pools health-check connections and re-establish broken connections in the event of a communication error.
+
+When creating connection pools in serverless functions:
+
+- Set the maximum connection pool size to 1, unless your function is multi-threaded and establishes multiple concurrent requests to your database within a single function instance.
+
+- Do not set a minimum idle connection count. The connection pool should be free to open connections as needed.
+
+- Set the maximum lifetime on the connection pool to 30 minutes.
+
+## Persist connection pools across function invocations
+
+If you plan to invoke a serverless function at a high frequency, you should configure the function to persist connection pools across function invocations, to limit the number of new connection attempts to the database. This is typically done by initializing the connection pool variable outside the scope of the serverless function definition.
+
+For example, suppose you are writing an AWS Lambda function that [`INSERT`](insert.html)s data into a table on a CockroachDB cluster, and you plan to run this query every few seconds. To persist a connection pool across invocations, initialize the connection pool variable outside of the [handler function](https://docs.aws.amazon.com/lambda/latest/dg/foundation-progmodel.html) definition and then define the connection pool in the handler only if the pool does not already exist.
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="node">node.JS</button>
+  <button class="filter-button page-level" data-scope="python">Python</button>
+</div>
+
+<section class="filter-content" markdown="1" data-scope="node">
+
+A node.JS function that implements this pattern could look similar to the following:
+
+{% include_cached copy-clipboard.html %}
+~~~ js
+const { Pool } = require('pg')
+
+let pool
+
+const insertRows = async (p) => {
+  const client = await p.connect()
+  try {
+    await client.query('INSERT INTO table (col1, col2) VALUES (val1, val2)')
+  } catch (err) {
+    console.log(err.stack)
+  } finally {
+    client.release()
+  }
+}
+
+exports.handler = async (context) => {
+  if (!pool) {
+    const connectionString = process.env.DATABASE_URL
+    pool = new Pool({
+      connectionString,
+      max: 1,
+      maxLifetimeSeconds = 1800
+    })
+  }
+
+  await insertRows(pool)
+}
+~~~
+
+</section>
+
+<section class="filter-content" markdown="1" data-scope="python">
+
+A Python function that implements this pattern could look similar to the following:
+
+{% include_cached copy-clipboard.html %}
+~~~ python
+from psycopg2.pool import SimpleConnectionPool
+
+pool = None
+
+
+def query(p):
+  conn = p.getconn()
+  with conn.cursor() as cur:
+      cur.execute("INSERT INTO table (col1, col2) VALUES (val1, val2)")
+      conn.commit()
+
+def lambda_handler(context):
+    global pool
+
+    if not pool:
+        pool = SimpleConnectionPool(0, 1, dsn=os.environ['DATABASE_URL'])
+
+    query(pool)
+
+    return
+~~~
+
+</section>
+
+## Use {{ site.data.products.serverless }}
+
+As a database-as-a-service, {{ site.data.products.serverless }} abstracts away the complexity of deploying, scaling, and load-balancing your database. Additionally, idle database connections to CockroachDB use very little memory (~20-30 KiB) when compared to PostgreSQL (~2-10 MiB).
+
+To create a free {{ site.data.products.serverless }} cluster:
+
+{% include cockroachcloud/quickstart/create-a-free-cluster.md %}
+
+## Deploy serverless functions in the same region as your database
+
+To minimize network latency, you should deploy your serverless functions in the same region as your database deployment. If your serverless function provider does not offer deployments in the same region as your database deployment, choose the region closest to the region where your database is deployed.
+
+## See also
+
+- [Create and Deploy an AWS Lambda Function Built on CockroachDB](deploy-lambda-function.html)
+- [Connect to your Cluster](connect-to-the-database.html)
--- a/_includes/v21.2/sidebar-data/develop.json
+++ b/_includes/v21.2/sidebar-data/develop.json
@@ -421,23 +421,6 @@
               ]
             }
           ]
-        },
-        {
-          "title": "Application Deployment Tools",
-          "items": [
-            {
-              "title": "Google Cloud Run",
-              "urls": [
-                "/${VERSION}/deploy-app-gcr.html"
-              ]
-            },
-            {
-              "title": "AWS Lambda",
-              "urls": [
-                "/${VERSION}/deploy-lambda-function.html"
-              ]
-            }
-          ]
         }
       ]
     },
--- a/_includes/v21.2/sidebar-data/get-started.json
+++ b/_includes/v21.2/sidebar-data/get-started.json
@@ -125,6 +125,29 @@
               ]
             }
           ]
+        },
+        {
+          "title": "Serverless",
+          "items": [
+            {
+              "title": "AWS Lambda",
+              "urls": [
+                "/${VERSION}/deploy-lambda-function.html"
+              ]
+            },
+            {
+              "title": "Google Cloud Run",
+              "urls": [
+                "/${VERSION}/deploy-app-gcr.html"
+              ]
+            },
+            {
+              "title": "Serverless Function Best Practices",
+              "urls": [
+                "/${VERSION}/serverless-function-best-practices.html"
+              ]
+            }
+          ]
         }
       ]
     },
--- a/v21.2/deploy-app-gcr.md
+++ b/v21.2/deploy-app-gcr.md
@@ -4,7 +4,7 @@ summary: Learn how to use Google Cloud Run (GCR) and CockroachDB Serverless to d
 toc: true
 twitter: false
 referral_id: docs_gcr_django
-docs_area: 
+docs_area: get_started
 ---
 
 This tutorial shows you how to use Google Cloud Run to deploy a containerized Django application that communicates with a {{ site.data.products.serverless }} cluster.
--- a/v21.2/deploy-lambda-function.md
+++ b/v21.2/deploy-lambda-function.md
@@ -4,7 +4,7 @@ summary: Learn how to use AWS Lambda and CockroachDB Serverless.
 toc: true
 twitter: false
 referral_id: docs_lambda_python
-docs_area: 
+docs_area: get_started
 ---
 
 This tutorial shows you how to create an [AWS Lambda](https://aws.amazon.com/lambda) function that communicates with a {{ site.data.products.serverless }} cluster.


andy-kimball (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 14 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVo5Yv2UQ4UzQeMICC:-MvVo5Yw5nqbg-s2ffKT:b-lfvcfq) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L14)):*
> ```Markdown
> We recommend using connection pools to manage the lifecycle of database connections established from serverless functions. Connection pools health-check connections and re-establish broken connections in the event of a communication error.
> 
> The maximum size of the connection pool should be determined by the concurrency support of the function-as-a-service (FaaS) runtime. Do not exceed the maximum number of concurrent requests allowed by your FaaS provider, and do not set a minimum idle connection count. The connection pool should be free to open connections as needed.
> ```

How is the size of the connection pool connected to the maximum concurrent requests allowed by the FaaS provider? It sounds like the "maximum concurrent requests" is the number of instances of my Serverless functions I can have running concurrently. But the size of the connection pool is about concurrency _within_ my Serverless function, which I don't think the FaaS runtimes care about. AFAIK, there are no restrictions on concurrent requests within a Serverless function instance. For example, I could spin up 100 goroutines in my Go Lambda function and open up 100 connections to my database.

What I expected here is that the size of the connection pool should not exceed the maximum number of concurrent requests that you expect to make to the database _within your Serverless function_. Probably in 99% of cases, that's 1, since most apps will have a single thread of execution. In that case, you don't want more than 1 connection in your pool.
___
*[v21.2/serverless-function-best-practices.md, line 20 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVpzjc5-JP-Q-Sk_VL:-MvVpzjc5-JP-Q-Sk_VM:bte1fm3) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L20)):*
> ```Markdown
> ## Persist database connections across function invocations
> 
> 
> ```

I think it's more like, "calling Serverless functions can result in a very high rate of new connections being established to your database". The number of concurrent connections would actually be _lower_ if you close each connection after your function runs. But, that'd be tremendously taxing on your database, since you'd have to then re-establish that connection the next time the function is invoked, and establishing new connections is very expensive (due to TLS, user authentication, etc).

ericharmeling (commented):
TFTR @andy-kimball. I made some updates in response to your feedback. PTAL.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991#-:-Mvdq5Mt6Uvc13BrnsXH:b-iumkv0)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 14 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVo5Yv2UQ4UzQeMICC:-MvZtGPo5ZxLYe492vt7:berl8vs) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L14)):*
> It sounds like the "maximum concurrent requests" is the number of instances of my Serverless functions I can have running concurrently. But the size of the connection pool is about concurrency within my Serverless function, which I don't think the FaaS runtimes care about. AFAIK, there are no restrictions on concurrent requests within a Serverless function instance. 

[The official Google Cloud Function docs](https://cloud.google.com/functions/docs/concepts/exec#auto-scaling_and_concurrency) make it sound like the other way around (i.e., that they restrict concurrent requests to a single function instance to 1, but there is a much higher limit to the number of requests across instances of the same function):

> Each instance of a function handles only one concurrent request at a time. This means that while your code is processing one request, there is no possibility of a second request being routed to the same instance. 

> In cases where inbound request volume exceeds the number of existing instances, Cloud Functions may start multiple new instances to handle requests. This automatic scaling behavior allows Cloud Functions to handle many requests in parallel, each using a different instance of your function.

[The Lambda docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) are a little fuzzier, but suggest the same to be true for Lambda functions:

> When your function is invoked, Lambda allocates an instance of it to process the event. When the function code finishes running, it can handle another request. If the function is invoked again while a request is still being processed, another instance is allocated, which increases the function's concurrency. 

The wording is a little tricky though, so perhaps I am misunderstanding what they are saying?

> What I expected here is that the size of the connection pool should not exceed the maximum number of concurrent requests that you expect to make to the database within your Serverless function. 

You're right - my wording here was misleading. Updated.
___
*[v21.2/serverless-function-best-practices.md, line 20 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVpzjc5-JP-Q-Sk_VL:-Mv_QB7n2_qiVU41opMu:bq7go1u) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L20)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

I think it's more like, "calling Serverless functions can result in a very high rate of new connections being established to your database". The number of concurrent connections would actually be _lower_ if you close each connection after your function runs. But, that'd be tremendously taxing on your database, since you'd have to then re-establish that connection the next time the function is invoked, and establishing new connections is very expensive (due to TLS, user authentication, etc).
</blockquote></details>

I just removed this entire intro section.

andy-kimball (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 14 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVo5Yv2UQ4UzQeMICC:-Mve0-FEEwqCvocZe7Fg:bbvc7ks) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L14)):*

> It sounds like the "maximum concurrent requests" is the number of instances of my Serverless functions I can have running concurrently. But the size of the connection pool is about concurrency within my Serverless function, which I don't think the FaaS runtimes care about. AFAIK, there are no restrictions on concurrent requests within a Serverless function instance. 

[The official Google Cloud Function docs](https://cloud.google.com/functions/docs/concepts/exec#auto-scaling_and_concurrency) make it sound like the other way around (i.e., that they restrict concurrent requests to a single function instance to 1, but there is a much higher limit to the number of requests across instances of the same function):

> Each instance of a function handles only one concurrent request at a time. This means that while your code is processing one request, there is no possibility of a second request being routed to the same instance. 

> In cases where inbound request volume exceeds the number of existing instances, Cloud Functions may start multiple new instances to handle requests. This automatic scaling behavior allows Cloud Functions to handle many requests in parallel, each using a different instance of your function.

[The Lambda docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) are a little fuzzier, but suggest the same to be true for Lambda functions:

> When your function is invoked, Lambda allocates an instance of it to process the event. When the function code finishes running, it can handle another request. If the function is invoked again while a request is still being processed, another instance is allocated, which increases the function's concurrency. 

The wording is a little tricky though, so perhaps I am misunderstanding what they are saying?

> What I expected here is that the size of the connection pool should not exceed the maximum number of concurrent requests that you expect to make to the database within your Serverless function. 

You're right - my wording here was misleading. Updated.
</blockquote></details>

OK, I see the confusion. There are 3 different possible meanings for "concurrency":

1. The number of concurrent instances of a particular Serverless Function running.
2. The number of concurrent requests sent to any one instance.
3. The number of concurrent threads/goroutines/requests/etc running _within_ any one instance.

#1 is not too interesting for client-side connection pooling decisions, since every instance is completely independent. It only matters for server-side connection pooling, but that's our problem, not the customer's problem. I think we should take some time in the docs to try and explain the difference, b/c it's a constant source of confusion.

#2 is always a max of 1 for all cloud providers, so I don't think you should be calling it out as something they need to worry about.

#3 is a concern, but it's also something that should be very rare for customers to do.

IMO, the message should be: use a max connection pool size of 1, unless you're doing fancy stuff like spinning up goroutines/threads/async requests, such that you're establishing multiple concurrent requests to your DB _in the same function_.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 14 at r1](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvVo5Yv2UQ4UzQeMICC:-MvskG6M8-Vee9u4O4NU:byf0d01) ([raw file](https://github.com/cockroachdb/docs/blob/74d1119c56757f26c977b04004e9e87c4598af42/v21.2/serverless-function-best-practices.md#L14)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

OK, I see the confusion. There are 3 different possible meanings for "concurrency":

1. The number of concurrent instances of a particular Serverless Function running.
2. The number of concurrent requests sent to any one instance.
3. The number of concurrent threads/goroutines/requests/etc running _within_ any one instance.

#1 is not too interesting for client-side connection pooling decisions, since every instance is completely independent. It only matters for server-side connection pooling, but that's our problem, not the customer's problem. I think we should take some time in the docs to try and explain the difference, b/c it's a constant source of confusion.

#2 is always a max of 1 for all cloud providers, so I don't think you should be calling it out as something they need to worry about.

#3 is a concern, but it's also something that should be very rare for customers to do.

IMO, the message should be: use a max connection pool size of 1, unless you're doing fancy stuff like spinning up goroutines/threads/async requests, such that you're establishing multiple concurrent requests to your DB _in the same function_.
</blockquote></details>

Makes sense. Thanks for clarifying. I pared this down to just touch on 3, as you suggested.

JeffSwenson (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @ericharmeling, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 16 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt2Ysk9x5ItreHD4DB:-Mvt2Ysk9x5ItreHD4DC:bvk487q) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L16)):*
> ```Markdown
> When creating connection pools in serverless functions:
> 
> - Set the maximum connection pool size to 1, unless your function is multi-threaded and establishes multiple concurrent requests to your database within a single function instance.
> ```

cc: @andy-kimball 

This is accurate advice for lambda, but it is not accurate for Google Cloud Run. Google Cloud Run will serve 80 concurrent requests from a single container by default https://cloud.google.com/run/docs/about-concurrency.

This is the main reason I brought up the idea of setting the max connection pool size to the default concurrency of the runtime. If a Google Cloud Run user set the max connection pool size to 1 and each db transaction is 50ms, they would see a per request latency of 80 * 50ms = 4 seconds before Google Cloud run would add more workers. 
___
*[v21.2/serverless-function-best-practices.md, line 57 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt5HE_9vw5_WCwq2rs:-Mvt5HEa1hhYzrmPG4--:b-es69on) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L57)):*
> ```Markdown
>     pool = new Pool({
>       connectionString,
>       max: 1,
> ```

I would like to add a max lifetime recommendation of 30 minutes. That is the default max lifetime in hikari and it allows us to update sql proxies without breaking connections. I checked and node-postgres does not supply a default max lifetime.

https://github.com/brettwooldridge/HikariCP#frequently-used

```
maxLifetimeSeconds = 1800
```
___
*[v21.2/serverless-function-best-practices.md, line 58 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt4qUV1iVccAcKLVLA:-Mvt4qUV1iVccAcKLVLB:bd1nt7m) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L58)):*
> ```Markdown
>       connectionString,
>       max: 1,
>       min: 0
> ```

Does pg.pool support the min parameter? I read through node-postgres and I do not see this in the pg-pool package. Judging by the code, node-postgres always behaves as if min = 0.

https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js
___
*[v21.2/serverless-function-best-practices.md, line 74 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt6fc16l1H0oxJmk4l:-Mvt6fc16l1H0oxJmk4m:b-xobsgf) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L74)):*
> ```Markdown
> ~~~ python
> from psycopg2.extras import register_uuid
> ```

nit: should we recommend psycopg3? It was released 2021-10-03.

https://www.psycopg.org/articles/2021/10/13/psycopg-30-released/

I did not find pscopg2 support for max lifetime. This is the config I recommend for psycopg3:

```
psycopg_pool.ConnectionPool(..., min_size=0, max_size = 1, max_lifetime=1800)
```

JeffSwenson (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @ericharmeling, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 16 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt2Ysk9x5ItreHD4DB:-Mvu9Po0CO_Adc95eXqX:b-bdww5g) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L16)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

This documentation is specifically about "Running in a Serverless Function", and I think it should not attempt to cover "Running in a Serverless Container". I think trying to generalize the advice by asking customers to know the default concurrency of their particular runtime is too confusing and hard to explain. Better to just have different sections in the documentation for Serverless Functions vs. Serverless Containers.
</blockquote></details>

That is fair. This documentation is accurate when applied to AWS Lambda and Google Cloud Run specifically. I like the idea of having a separate doc for runtimes that support >1 concurrency and we can explicitly list runtimes that fall into that category. I'm not sure there is a clear separation between container runtimes and function runtimes. For example: AWS lambda has a container runtime option https://docs.aws.amazon.com/lambda/latest/dg/images-create.html.

ericharmeling (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 10 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-MvtY-ez7LbP1TEv6v4-:-Mvt_xZ6EggZE2Nf72mi:bt2vr1b) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L10)):*
<details><summary><i>Previously, andy-kimball (Andy Kimball) wrote…</i></summary><blockquote>

This line still requires the user to dig around for esoteric details about their serverless runtime, which I don't think is necessary, unless we find a Serverless Function runtime that allows concurrency > 1.
</blockquote></details>

Fixed.
___
*[v21.2/serverless-function-best-practices.md, line 16 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt2Ysk9x5ItreHD4DB:-MvtKNan7vITsewoUgJF:b-wj921r) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L16)):*
>  Google Cloud Run will serve 80 concurrent requests from a single container by default https://cloud.google.com/run/docs/about-concurrency.

From the [Google Cloud Functions docs on concurrency](https://cloud.google.com/functions/docs/concepts/exec#auto-scaling_and_concurrency), it doesn't seem like this applies to Google Cloud Functions:

> Each instance of a function handles only one concurrent request at a time. This means that while your code is processing one request, there is no possibility of a second request being routed to the same instance.

As you note, Cloud Functions aren't really distinct from GCR containers. It looks like Cloud Functions are run in GCR containers with a max request concurrency of 1, as shown in the graphic on that page:



>  Better to just have different sections in the documentation for Serverless Functions vs. Serverless Containers.

> I like the idea of having a separate doc for runtimes that support >1 concurrency and we can explicitly list runtimes that fall into that category. 


+1 to both of your suggestions. If you're both good with it, I'll leave the recommendation here as-is, and we can save the container recommendations for a different doc (or a later iteration of this doc).
___
*[v21.2/serverless-function-best-practices.md, line 57 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt5HE_9vw5_WCwq2rs:-MvtK49eDXCQt7mBui8K:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L57)):*
<details><summary><i>Previously, JeffSwenson (Jeff Swenson) wrote…</i></summary><blockquote>

I would like to add a max lifetime recommendation of 30 minutes. That is the default max lifetime in hikari and it allows us to update sql proxies without breaking connections. I checked and node-postgres does not supply a default max lifetime.

https://github.com/brettwooldridge/HikariCP#frequently-used

```
maxLifetimeSeconds = 1800
```
</blockquote></details>

Done.
___
*[v21.2/serverless-function-best-practices.md, line 58 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt4qUV1iVccAcKLVLA:-MvtGIyfEW8gliiKf0VT:bshacmd) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L58)):*
Good catch.

> Does pg.pool support the min parameter? 

It does not appear to... though I was [able to deploy a node app that specifies this parameter](https://github.com/cockroachlabs/examples-aws-lambda/pull/3/files#diff-ddf78f02f4a8e0bccaa29f6a7b1c18871b90fcf4f273fba9163c0510e5cc212cR46). Odd.

I removed the parameter (which should be enough, given the recommendation is to *not* set a minimum).
___
*[v21.2/serverless-function-best-practices.md, line 74 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt6fc16l1H0oxJmk4l:-MvtHBHU38HPh6U7ok21:b-lf0n00) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L74)):*
<details><summary><i>Previously, JeffSwenson (Jeff Swenson) wrote…</i></summary><blockquote>

nit: should we recommend psycopg3? It was released 2021-10-03.

https://www.psycopg.org/articles/2021/10/13/psycopg-30-released/

I did not find pscopg2 support for max lifetime. This is the config I recommend for psycopg3:

```
psycopg_pool.ConnectionPool(..., min_size=0, max_size = 1, max_lifetime=1800)
```
</blockquote></details>

I'll leave this up to @kernfeld-cockroach ... [In the past, we've opted against documenting pscyopg3 connections](https://github.com/cockroachdb/docs/pull/12538#discussion_r769049822), as we don't claim to have official support for psycopg3, and AFAIK, we don't have any dedicated nightly tests for psycopg3. 

FWIW, we do have a couple issues for pscyopg3 support ([mainter support](https://cockroachlabs.atlassian.net/browse/DXTPT-2) and [docs](https://cockroachlabs.atlassian.net/browse/DOC-2080)), but those are currently on the backlog.

kernfeld-cockroach (commented):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andy-kimball, @ericharmeling, @JeffSwenson, and @kernfeld-cockroach)

___
*[v21.2/serverless-function-best-practices.md, line 74 at r3](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mvt6fc16l1H0oxJmk4l:-Mvz_Bgg3q5GonqImCH7:b-tu4v7x) ([raw file](https://github.com/cockroachdb/docs/blob/433e688d1fc7e8ae06c3fb56f2ec7acea392702c/v21.2/serverless-function-best-practices.md#L74)):*

I'll leave this up to @kernfeld-cockroach ... [In the past, we've opted against documenting pscyopg3 connections](https://github.com/cockroachdb/docs/pull/12538#discussion_r769049822), as we don't claim to have official support for psycopg3, and AFAIK, we don't have any dedicated nightly tests for psycopg3. 

FWIW, we do have a couple issues for pscyopg3 support ([mainter support](https://cockroachlabs.atlassian.net/browse/DXTPT-2) and [docs](https://cockroachlabs.atlassian.net/browse/DOC-2080)), but those are currently on the backlog.
</blockquote></details>

Right, we shouldn't mention psycopg3 until we support it.

andy-kimball (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991#-:-Mw-732Q3RpkZJaRcgZi:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @andy-kimball, @ericharmeling, @JeffSwenson, and @kernfeld-cockroach)

v21.2/serverless-function-best-practices.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Serverless Function Best Practices
+summary: Best practices for optimizing the performance of serverless functions (e.g., AWS Lambda functions or Google Cloud Functions) that connect to CockroachDB.
+toc: true
+referral_id: docs_serverless_functions
+---
```
jseldess, line 6: Don't forget to add `docs_area`: https://github.com/cockroachdb/docs/blob/master/CONTRIBUTING.md#pages

JeffSwenson (approved):

v21.2/serverless-function-best-practices.md
ericharmeling: Done.

nickvigilante (changes requested):
Just a few nits, but otherwise LGTM!

> Reviewed 1 of 3 files at r1, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991#-:-Mw2B_l67eLTW-ccO3Xx:b-4lkk0y)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @andy-kimball, @ericharmeling, @JeffSwenson, and @jseldess)

___
*[v21.2/serverless-function-best-practices.md, line 13 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw29MZ57e96ygS2LB_t:-Mw29MZ57e96ygS2LB_u:bmhikb) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L13)):*
> ```Markdown
> ## Use connection pools
> 
> We recommend using connection pools to manage the lifecycle of database connections established by serverless functions. Connection pools health-check connections and re-establish broken connections in the event of a communication error.
> ```

Do we say "We recommend" or should we say "Cockroach Labs recommends"? Or something completely different to pull the focus away from the writer as Stephanie suggests?

https://cockroachlabs.slack.com/archives/C01NFCXULTX/p1641833812002200
___
*[v21.2/serverless-function-best-practices.md, line 101 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2AvzN0_QwIT1FkUQa:-Mw2AvzN0_QwIT1FkUQb:b2tgt0s) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L101)):*
> ```Markdown
> </section>
> 
> ## Use CockroachDB {{ site.data.products.serverless }}
> ```

Omit the extra CockroachDB.



_Code quote:_
```Markdown
## Use CockroachDB {{ site.data.products.serverless }}
```
___
*[v21.2/serverless-function-best-practices.md, line 103 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2BPtp6onAS9sv0NyP:-Mw2BPtp6onAS9sv0NyQ:buodkiy) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L103)):*
> ```Markdown
> ## Use CockroachDB {{ site.data.products.serverless }}
> 
> As a database-as-a-service, CockroachDB {{ site.data.products.serverless }} abstracts away the complexity of deploying, scaling, and load-balancing your database. Additionally, idle database connections to CockroachDB use very little memory (~20-30 KiB) when compared to PostgreSQL (~2-10 MiB).
> ```

Same nit.
___
*[v21.2/serverless-function-best-practices.md, line 105 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2BVt20iNq7xKGpGE2:-Mw2BVt20iNq7xKGpGE3:buodkiy) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L105)):*
> ```Markdown
> As a database-as-a-service, CockroachDB {{ site.data.products.serverless }} abstracts away the complexity of deploying, scaling, and load-balancing your database. Additionally, idle database connections to CockroachDB use very little memory (~20-30 KiB) when compared to PostgreSQL (~2-10 MiB).
> 
> To create a free CockroachDB {{ site.data.products.serverless }} cluster:
> ```

Same nit.

ericharmeling (commented):
TFTRs! 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/12991#-:-Mw2G-MI61LO-b3KaCiR:bodi8jh)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @andy-kimball, @JeffSwenson, @jseldess, and @nickvigilante)

___
*[v21.2/serverless-function-best-practices.md, line 13 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw29MZ57e96ygS2LB_t:-Mw2Dmb8A6Y6qKRiIU2Z:b-w2hu74) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L13)):*
<details><summary><i>Previously, nickvigilante (Nick Vigilante) wrote…</i></summary><blockquote>

Do we say "We recommend" or should we say "Cockroach Labs recommends"? Or something completely different to pull the focus away from the writer as Stephanie suggests?

https://cockroachlabs.slack.com/archives/C01NFCXULTX/p1641833812002200
</blockquote></details>

I'm just going to make this a direct imperative.

FWIW, I personally like "we recommend" vs "Cockroach Labs recommends" because it follows the [style guide's](https://github.com/cockroachdb/docs/blob/master/StyleGuide.md) general "be helpful, humble, positive, and friendly" principle. I find the third-person to be (literally) impersonal and little too formal for a modern docset.
___
*[v21.2/serverless-function-best-practices.md, line 101 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2AvzN0_QwIT1FkUQa:-Mw2Fx5x8UB9J_yUU2JQ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L101)):*
<details><summary><i>Previously, nickvigilante (Nick Vigilante) wrote…</i></summary><blockquote>

Omit the extra CockroachDB.
</blockquote></details>

Done.
___
*[v21.2/serverless-function-best-practices.md, line 103 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2BPtp6onAS9sv0NyP:-Mw2FxSp9LE6BXqdAgUN:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L103)):*
<details><summary><i>Previously, nickvigilante (Nick Vigilante) wrote…</i></summary><blockquote>

Same nit.
</blockquote></details>

Done.
___
*[v21.2/serverless-function-best-practices.md, line 105 at r5](https://reviewable.io/reviews/cockroachdb/docs/12991#-Mw2BVt20iNq7xKGpGE2:-Mw2FxqLBEhhAsoTxReZ:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/867619fe1938d479d41000416c0a64faac9ffbf2/v21.2/serverless-function-best-practices.md#L105)):*
<details><summary><i>Previously, nickvigilante (Nick Vigilante) wrote…</i></summary><blockquote>

Same nit.
</blockquote></details>

Done.

nickvigilante:
@ericharmeling Ah, understood. Friendly is good. LGTM!

PR #13003, opened 2022-02-10
Document NO_FULL_SCAN.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-1711
https://cockroachlabs.atlassian.net/browse/DOC-1666

### DIFF
--- a/_includes/v21.2/misc/force-index-selection.md
+++ b/_includes/v21.2/misc/force-index-selection.md
@@ -4,6 +4,8 @@ By using the explicit index annotation, you can override [CockroachDB's index se
 Index selection can impact [performance](performance-best-practices-overview.html), but does not change the result of a query.
 {{site.data.alerts.end}}
 
+##### Force index scan
+
 The syntax to force a scan of a specific index is:
 
 {% include copy-clipboard.html %}
@@ -18,7 +20,9 @@ This is equivalent to the longer expression:
 > SELECT * FROM table@{FORCE_INDEX=my_idx};
 ~~~
 
-The syntax to force a **reverse scan** of a specific index is:
+##### Force reverse scan
+
+The syntax to force a reverse scan of a specific index is:
 
 {% include copy-clipboard.html %}
 ~~~ sql
@@ -58,6 +62,8 @@ you can check the scan direction with:
 (2 rows)
 ~~~
 
+##### Force partial index scan
+
 To force a [partial index scan](partial-indexes.html), your statement must have a `WHERE` clause that implies the partial index filter.
 
 {% include copy-clipboard.html %}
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13003)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rytaft and @stbof)

___
*[_includes/v21.2/misc/force-index-selection.md, line 144 at r1](https://reviewable.io/reviews/cockroachdb/docs/13003#-MvbBjDpBoXn8wec-8-q:-MvbBjDpBoXn8wec-8-r:bqicjc4) ([raw file](https://github.com/cockroachdb/docs/blob/ed73ebfa7a400fbdc88678251d49ee58fc699d2f/_includes/v21.2/misc/force-index-selection.md#L144)):*
> ```Markdown
> ~~~
> 
> To prevent a full scan of a partial index, specify `NO_FULL_SCAN` in combination with a specific partial index using `FORCE_INDEX=index_name`.
> ```

I think we need to explain that a scan of an entire partial index is not considered a full scan.
I don't quite understand what the combination of `NO_FULL_SCAN` and `FORCE_INDEX=..` does exactly.. Does it force a plan where the index is used but the scan is also constrained? @rytaft should know

rytaft (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13003)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @RaduBerinde)

___
*[_includes/v21.2/misc/force-index-selection.md, line 144 at r1](https://reviewable.io/reviews/cockroachdb/docs/13003#-MvbBjDpBoXn8wec-8-q:-MvsIw4nCzhm_GQEvruZ:b5v2etm) ([raw file](https://github.com/cockroachdb/docs/blob/ed73ebfa7a400fbdc88678251d49ee58fc699d2f/_includes/v21.2/misc/force-index-selection.md#L144)):*
<details><summary><i>Previously, RaduBerinde wrote…</i></summary><blockquote>

I think we need to explain that a scan of an entire partial index is not considered a full scan.
I don't quite understand what the combination of `NO_FULL_SCAN` and `FORCE_INDEX=..` does exactly.. Does it force a plan where the index is used but the scan is also constrained? @rytaft should know
</blockquote></details>

That's correct -- if `NO_FULL_SCAN` is used without `FORCE_INDEX`, a full scan of a partial index may be planned, since it's technically not a full table scan.

When `NO_FULL_SCAN` is used with `FORCE_INDEX`, we will try to build a query plan where the specific index is constrained. So if used with a partial index, this will prevent a full scan of the partial index.

rmloveland (approved):
LGTM

PR #13012, opened 2022-02-11
Sketch of simple api policy

jseldess:


marksoper:
@jseldess we should add a top-level section for the ccloud CLI alongside the Cloud API

jseldess:
Thanks, @marksoper. Added `ccloud`.

PR #13059, opened 2022-02-17
DOC-2679: Update Netlify build to fail if Jekyll build fails

nickvigilante:
Fixes DOC-2679

### DIFF
--- a/netlify/build
+++ b/netlify/build
@@ -2,6 +2,7 @@
 
 function build {
 	bundle exec jekyll build --trace --config _config_base.yml,$1
+	exit $?
 }
 
 # Transform API spec for API docs generation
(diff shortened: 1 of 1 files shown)


lnhsingh (approved):

netlify/build
```diff
@@ -14,6 +14,7 @@ popd
 gem install bundler
 bundle install
 build _config_cockroachdb.yml
+exit $?
```
jseldess, line 17: I don't know bash very well, but why are we handling this differently than for htmltest?

```
if [[ $? != 0 ]]; then
  exit 1
fi;
```

jseldess:
@nickvigilante, can you create a distinct Jira issue for the work you're doing here? Essentially, something like: `Fail Netlify build and block merge when Jekyll build fails`.

jseldess (approved):
LGTM

ianjevans (approved):

PR #13063, opened 2022-02-17
Revert "DOC-2679: Update Netlify build to fail if Jekyll build fails"

nickvigilante:
Reverts cockroachdb/docs#13059


nickvigilante:
@ianjevans @jseldess @lnhsingh I had to revert this because it broke "stable".

PR #13093, opened 2022-02-23
When to use GLOBAL tables vs. REGIONAL plus Follower Reads

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-1001


nvanbenschoten (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13093#-:-Mx04CPZAGpos9chWUkw:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

___
*[v21.2/topology-patterns.md, line 48 at r1](https://reviewable.io/reviews/cockroachdb/docs/13093#-Mx03t6H-VB8dpu-5-vB:-Mx03t6H-VB8dpu-5-vC:bf0vldj) ([raw file](https://github.com/cockroachdb/docs/blob/59a62a9482448bd02c7f0cfa58e66f7e73d2a113/v21.2/topology-patterns.md#L48)):*
> ```Markdown
> If you want low-latency read-only access to your data in multiple regions, Cockroach Labs recommends that you default to using stale follower reads on `REGIONAL` tables (the default locality). However, there are two reasons you might want to upgrade a table to `GLOBAL`:
> 
> - You want low-latency consistent (non-stale) read access to the table in multiple regions from read-write transactions. One case where this is important is if the table is referenced by a [foreign key](foreign-key.html) from a [`REGIONAL BY ROW`](regional-tables.html#regional-by-row-tables) table. In these cases, the foreign key check that accompanies a write cannot use a stale read because it must be transactionally consistent with the write. To keep this foreign key check fast, you can slow down writes to the reference table by making it `GLOBAL`.
> ```

Instead of "you can slow down writes to the reference table by making it", consider saying something like:

"To keep this foreign key check fast, you can make the reference table GLOBAL, at the expense of slower writes to that table."

v21.2/topology-patterns.md
```diff
@@ -45,6 +43,13 @@ The multi-region patterns described below are almost always table-specific. For
 In [multi-region databases](multiregion-overview.html), the resiliency of each database depends on its [survival goal settings](multiregion-overview.html#survival-goals).
 {{site.data.alerts.end}}
 
+If you want low-latency read-only access to your data in multiple regions, Cockroach Labs recommends that you default to using stale follower reads on `REGIONAL` tables (the default locality). However, there are two reasons you might want to upgrade a table to `GLOBAL`:
+
+- You want low-latency consistent (non-stale) read access to the table in multiple regions from read-write transactions. One case where this is important is if the table is referenced by a [foreign key](foreign-key.html) from a [`REGIONAL BY ROW`](regional-tables.html#regional-by-row-tables) table. In these cases, the foreign key check that accompanies a write cannot use a stale read because it must be transactionally consistent with the write. To keep this foreign key check fast, you can slow down writes to the reference table by making it `GLOBAL`.
```
stbof, line 48:
```suggestion
- You want low-latency consistent (non-stale) read access to the table in multiple regions from read-write transactions. One case where this is important is if the table is referenced by a [foreign key](foreign-key.html) from a [`REGIONAL BY ROW`](regional-tables.html#regional-by-row-tables) table. In these cases, the foreign key check that accompanies a write cannot use a stale read because it must be transactionally consistent with the write. To keep this foreign key check fast, you can make the reference table `GLOBAL`, at the expense of slower writes to that table.
```

nvanbenschoten (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13093)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @rmloveland and @stbof)

rmloveland (approved):
LGTM

PR #13026, opened 2022-02-14
Reorganize performance articles.

stbof:
See description in https://cockroachlabs.atlassian.net/browse/DOC-2576.

- Move rule application tutorial to new page.
- Adds missing links to troubleshooting articles.
- Consistent titles in TOC and pages.

PR #13102, opened 2022-02-24
Add Enterprise features to TOC and fix link.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-1167

Benchmarking orphan fixed in https://github.com/cockroachdb/docs/pull/13026.


jseldess (approved):
LGTM, but the page is very self-hosted centric, as discussed in Slack. We should eventually update it to be clear which enterprise features Cloud customers get on Serverless vs. Dedicated.

PR #13109, opened 2022-02-24
CBO style fixes

ghost:


rytaft (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13109)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)

___
*[v21.2/cost-based-optimizer.md, line 117 at r1](https://reviewable.io/reviews/cockroachdb/docs/13109#-MwhuvKt5AsZnF3qjIQf:-Mwhum_i8BLaC4vnKwD0:b-q8e29n) ([raw file](https://github.com/cockroachdb/docs/blob/ffec8bd9399f33eec6c9c7d135f1dcaf585bbdec/v21.2/cost-based-optimizer.md#L117)):*
> ```Markdown
> ### Requirements
> 
> To realize locality optimized search you must specify a [`CHECK` constraint](check.html) when you create the table. For example, if you typically perform a query that has a locality constraint:
> ```

The need for `CHECK` constraints isn't really related to locality optimized search. I guess it will become relevant in 22.1 when we add support for using locality optimized search with `LIMIT`, but it's not relevant for LOS in 21.2. 

Users of partitioned tables who are *not* using the new multi-region abstractions should consider explicitly adding a`CHECK` constraint or using an `ENUM` type column on the partitioning column(s) for general improved performance. Note that as far as the optimizer is concerned, an `ENUM` type is equivalent to a `CHECK` constraint since it limits the possible column values to a finite set of known values. Using an `ENUM` type / `CHECK` constraint for the partitioning column(s) is especially helpful for tables that use `PARTITION BY LIST` if users want to ensure that the optimizer can effectively avoid scanning data that's not necessary for the query. The `CHECK` constraint or `ENUM` column should restrict the possible values of the partitioning column to those that map to partitions according to the `PARTITION BY LIST` clause. `DEFAULT` partitions and `PARTITION BY RANGE` should be avoided for best performance.

Users of the new multi-region abstractions need not concern themselves with this. The `crdb_region` column is the default partitioning column for `REGIONAL BY ROW` tables, and it already has an `ENUM` type. This is another reason why the new multi-region abstractions are easier to use -- there's no need to think about the need for a `CHECK` constraint / `ENUM` -- it's already taken care of.

`CHECK` constraints and `ENUM` types are also helpful on non-partitioning columns in general, since it gives the optimizer more information about the allowed values of a column, and enables various different types of optimizations. This applies to non-partitioned and non-multi-region tables as well.

Let me know if this is confusing or you want more info...

rytaft:
___
*[v21.2/cost-based-optimizer.md, line 117 at r1](https://reviewable.io/reviews/cockroachdb/docs/13109#-MwhuvKt5AsZnF3qjIQf:-MwisTWxDfvdwp_NwbV1:b-f3kxsu) ([raw file](https://github.com/cockroachdb/docs/blob/ffec8bd9399f33eec6c9c7d135f1dcaf585bbdec/v21.2/cost-based-optimizer.md#L117)):*

The need for `CHECK` constraints isn't really related to locality optimized search. I guess it will become relevant in 22.1 when we add support for using locality optimized search with `LIMIT`, but it's not relevant for LOS in 21.2. 

Users of partitioned tables who are *not* using the new multi-region abstractions should consider explicitly adding a`CHECK` constraint or using an `ENUM` type column on the partitioning column(s) for general improved performance. Note that as far as the optimizer is concerned, an `ENUM` type is equivalent to a `CHECK` constraint since it limits the possible column values to a finite set of known values. Using an `ENUM` type / `CHECK` constraint for the partitioning column(s) is especially helpful for tables that use `PARTITION BY LIST` if users want to ensure that the optimizer can effectively avoid scanning data that's not necessary for the query. The `CHECK` constraint or `ENUM` column should restrict the possible values of the partitioning column to those that map to partitions according to the `PARTITION BY LIST` clause. `DEFAULT` partitions and `PARTITION BY RANGE` should be avoided for best performance.

Users of the new multi-region abstractions need not concern themselves with this. The `crdb_region` column is the default partitioning column for `REGIONAL BY ROW` tables, and it already has an `ENUM` type. This is another reason why the new multi-region abstractions are easier to use -- there's no need to think about the need for a `CHECK` constraint / `ENUM` -- it's already taken care of.

`CHECK` constraints and `ENUM` types are also helpful on non-partitioning columns in general, since it gives the optimizer more information about the allowed values of a column, and enables various different types of optimizations. This applies to non-partitioned and non-multi-region tables as well.

Let me know if this is confusing or you want more info...
</blockquote></details>

Another important note is that in order for the optimizer to take advantage of `CHECK` constraints and `ENUM` type columns, the expression cannot evaluate to `NULL`. As a result, these optimizations only apply to columns that are also `NOT NULL`.

stbof:
> > Reviewed 1 of 1 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13109)_ status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)
> 
> _[v21.2/cost-based-optimizer.md, line 117 at r1](https://reviewable.io/reviews/cockroachdb/docs/13109#-MwhuvKt5AsZnF3qjIQf:-Mwhum_i8BLaC4vnKwD0:b-q8e29n) ([raw file](https://github.com/cockroachdb/docs/blob/ffec8bd9399f33eec6c9c7d135f1dcaf585bbdec/v21.2/cost-based-optimizer.md#L117)):_
> 
> > ```gfm
> > ### Requirements
> > 
> > To realize locality optimized search you must specify a [`CHECK` constraint](check.html) when you create the table. For example, if you typically perform a query that has a locality constraint:
> > ```
> 
> The need for `CHECK` constraints isn't really related to locality optimized search. I guess it will become relevant in 22.1 when we add support for using locality optimized search with `LIMIT`, but it's not relevant for LOS in 21.2.
> 
> Users of partitioned tables who are _not_ using the new multi-region abstractions should consider explicitly adding a`CHECK` constraint or using an `ENUM` type column on the partitioning column(s) for general improved performance. Note that as far as the optimizer is concerned, an `ENUM` type is equivalent to a `CHECK` constraint since it limits the possible column values to a finite set of known values. Using an `ENUM` type / `CHECK` constraint for the partitioning column(s) is especially helpful for tables that use `PARTITION BY LIST` if users want to ensure that the optimizer can effectively avoid scanning data that's not necessary for the query. The `CHECK` constraint or `ENUM` column should restrict the possible values of the partitioning column to those that map to partitions according to the `PARTITION BY LIST` clause. `DEFAULT` partitions and `PARTITION BY RANGE` should be avoided for best performance.
> 
> Users of the new multi-region abstractions need not concern themselves with this. The `crdb_region` column is the default partitioning column for `REGIONAL BY ROW` tables, and it already has an `ENUM` type. This is another reason why the new multi-region abstractions are easier to use -- there's no need to think about the need for a `CHECK` constraint / `ENUM` -- it's already taken care of.
> 
> `CHECK` constraints and `ENUM` types are also helpful on non-partitioning columns in general, since it gives the optimizer more information about the allowed values of a column, and enables various different types of optimizations. This applies to non-partitioned and non-multi-region tables as well.
> 
> Let me know if this is confusing or you want more info...

So does the recommendation in https://cockroachlabs.atlassian.net/browse/DOC-720 apply to the old way of handling multi-region data? If so, where would it be appropriate to document this?  See https://cockroachlabs.atlassian.net/browse/DOC-2718 for a discussion of  issues with the old docs.

stbof:
> _[v21.2/cost-based-optimizer.md, line 117 at r1](https://reviewable.io/reviews/cockroachdb/docs/13109#-MwhuvKt5AsZnF3qjIQf:-MwlsLVN0zk2m1PVqVdl:bamgh6) ([raw file](https://github.com/cockroachdb/docs/blob/ffec8bd9399f33eec6c9c7d135f1dcaf585bbdec/v21.2/cost-based-optimizer.md#L117)):_
> 
> > So does the recommendation in https://cockroachlabs.atlassian.net/browse/DOC-720 apply to the old way of handling multi-region data?
> 
> That's correct -- the Zendesk ticket referenced in that issue was all about how to avoid high latencies with the old multi-region primitives.
> 
> > If so, where would it be appropriate to document this? See https://cockroachlabs.atlassian.net/browse/DOC-2718 for a discussion of issues with the old docs.
> 
> I think it's probably important to keep documentation around somewhere for the old-multi-region primitives, since the new multi-region abstractions cannot yet cover all use cases, and migration from old to new also seems to be difficult. Ideally there would be a disclaimer at the top saying that the new abstractions are the preferred approach to multi-region, but power users or those with special circumstances / legacy setups could still use the old approach.

There is a problem with the old documentation as described in the cited issue. The v21 (and soon v22) docs point to links in the v20 docs because those docs weren't moved forward. The links are confusing because v20 is scheduled to go out of support in May 2022 and there is a warning to that effect. In May the v20 docs will say that the version is _not supported_, which is not helpful even if the old features are still supported in the product. IOW, if there are no docs for a feature in the current version of the docs, effectively the feature is not supported.  https://cockroachlabs.atlassian.net/browse/DOC-2718 is saying that either we need to move those docs forward to the current version, or stop supporting those features. Please comment in that Jira if you feel strongly one way or the other.

rytaft:
___
*[v21.2/cost-based-optimizer.md, line 117 at r1](https://reviewable.io/reviews/cockroachdb/docs/13109#-MwhuvKt5AsZnF3qjIQf:-Mwlwkj824f4C9RWzSKL:b-v39ewq) ([raw file](https://github.com/cockroachdb/docs/blob/ffec8bd9399f33eec6c9c7d135f1dcaf585bbdec/v21.2/cost-based-optimizer.md#L117)):*
<details><summary><i>Previously, stbof (Stephanie Bodoff) wrote…</i></summary><blockquote>

There is a problem with the old documentation as described in the cited issue. The v21 (and soon v22) docs point to links in the v20 docs because those docs weren't moved forward. The links are confusing because v20 is scheduled to go out of support in May 2022 and there is a warning to that effect. In May the v20 docs will say that the version is _not supported_, which is not helpful even if the old features are still supported in the product. IOW, if there are no docs for a feature in the current version of the docs, effectively the feature is not supported.  https://cockroachlabs.atlassian.net/browse/DOC-2718 is saying that either we need to move those docs forward to the current version, or stop supporting those features. Please comment in that Jira if you feel strongly one way or the other.

In any case, I will back out this change in the CBO doc until I have guidance on where to put this information.

</blockquote></details>

I just commented on the jira ticket.

stbof:
This is now simply style fixes.

rytaft (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13109#-:-Mx0GUR75n4eAlqHP0jk:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

PR #13110, opened 2022-02-24
Accessibility, update to 2021 cloud report, capitalization.

stbof:


v21.2/performance.md
```diff
@@ -8,17 +8,17 @@ docs_area: manage
...
+This document is about CockroachDB performance on benchmarks. For guidance on tuning real workloads, see [SQL Best Practices](performance-best-practices-overview.html), and for guidance on data location techniques to minimize network latency, see [Topology Patterns](topology-patterns.html).
 {{site.data.alerts.end}}
 
 ## Scale
 
-TPC-C provides the most realistic and objective measure for OLTP performance at various scale factors, and CockroachDB can process **1.68M tpmC with 140,000 warehouses, resulting in an efficiency score of 95%.** As shown in the chart below, this is a 40% improvement over the results from CockroachDB 19.2.
+TPC-C provides the most realistic and objective measure for OLTP performance at various scale factors. CockroachDB 21.1 can process **1.68M tpmC with 140,000 warehouses, resulting in an efficiency score of 95%.** As shown in the following chart, this is a 40% improvement over the results from CockroachDB 19.2.
```
rmloveland, line 19: I don't think adding this version number in this PR is a good idea.  These pages go through a lot of vetting every time they are updated.  Style edits are fine but I'm against any content additions.  Especially in a big "style changes" PR like this.  Too easy to miss mistakes. If you want to make non-style content edits I'd rather send you to engineering reviewers

(This may sound extreme but I've spent literally weeks having back-and-forths on some of these pages in the past, it's very easy to introduce small errors)

rmloveland (changes requested):
LGTM except the addition of the non-style edit (version number) on the performance page.  Other than that all looks great!

v21.2/performance.md
stbof, line 19:
```suggestion
TPC-C provides the most realistic and objective measure for OLTP performance at various scale factors. CockroachDB can process **1.68M tpmC with 140,000 warehouses, resulting in an efficiency score of 95%.** As shown in the following chart, this is a 40% improvement over the results from CockroachDB 19.2.
```

stbof: Thanks for the great catch. Totally agree. Not sure why I put that in. Fixed.

stbof: @rmloveland Now I remember why I put the version in. There's a comparison with CockroachDB 19.2 at the end. It's a weird comparison if one side has a version and the other doesn't. Thoughts?

rmloveland: IMO the best thing is to leave this alone until the next content (non-style) edit

rmloveland (approved):
LGTM once that one content addition is removed

stbof:
> LGTM once that one content addition is removed

Fixed in https://github.com/cockroachdb/docs/pull/13110/commits/dc25df2068325b9a043fcf724bff1a58842beefa

PR #13137, opened 2022-03-01
(retrying) minor tweaks and corrections to node shutdown doc

taroface:
- Minor corrections and wording tweaks to Node Shutdown doc.
- Adjusted the server-side redirect to use stable and not v21.2.


kathancox (approved):
LGTM!

PR #13191, opened 2022-03-08
Fix for page formatting with SSO include.

ianjevans:
https://cockroachlabs.atlassian.net/browse/DOC-2846


stbof (approved):
Looks good. Could you add the fix to the Style Guide?

nickvigilante (approved):
I checked to make sure the cockroachcloud/authentication.md file that also references this include works fine, and it looks like it does. LGTM!

PR #13194, opened 2022-03-08
DOC-1554: Added IF [NOT] EXISTS syntax to ADD/DROP REGION

nickvigilante:
Fixes DOC-1554, DOC-1556


gemma-shay (approved):

nickvigilante:
@gemma-shay TFTR!

PR #13205, opened 2022-03-09
Clarify wording for Open SQL Transactions chart

rafiss:
This came up as a point of confusion on this community slack thread: https://cockroachdb.slack.com/archives/CP4D9LD5F/p1646720306266029

### DIFF
--- a/v21.1/ui-sql-dashboard.md
+++ b/v21.1/ui-sql-dashboard.md
@@ -14,7 +14,7 @@ The **SQL** dashboard displays the following time series graphs:
 
 ## Open SQL Sessions
 
-- In the node view, the graph shows the number of connections currently open between the client and the selected node.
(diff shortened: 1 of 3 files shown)


stbof:
currently open SQL transactions on the node.

It seems redundant to say "currently open". Doesn't open mean the transaction is current?

stbof (changes requested):
See my comment about removing "currently"  from the other bullets instead of adding it to the Open SQL Transactions bullet.

maryliag:
@rafiss just checking if this PR is still valid, since I didn't see any chances based on Stephanie's comments

rafiss:
I forgot about this. I can incorporate the comments; thanks Stephanie.

stbof (approved):

PR #13263, opened 2022-03-16
Updated version tags in style guide

ericharmeling:
Fixes https://cockroachlabs.atlassian.net/browse/DOC-2415.

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -631,19 +631,21 @@ Examples help show the feature in action. Examples follow a basic format:
 
 ### Version tags
 
-Use HTML version tags to denote new or updated features for the version. Version tags can be put at the end of a heading, if the whole section describes something that is new or updated. Otherwise, version tags can introduce a paragraph, sentence, or description in a table cell.
+Version tags inform users of new and updated features in CockroachDB, and could motivate users to upgrade to the latest major or minor version of CockroachDB. Version tags also help us identify new and updated features that we can call out in [our GA release notes](https://cockroachlabs.atlassian.net/wiki/spaces/ED/pages/402718726/GA+Release+Checklist).
 
-To insert a version tag, use the following code:
+To add a version tag, use the following HTML `span`:
 
 ~~~
 <span class="version-tag">New in vX.X:</span>
 ~~~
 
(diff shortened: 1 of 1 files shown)


StyleGuide.md
```diff
@@ -631,19 +631,21 @@ Examples help show the feature in action. Examples follow a basic format:
...
-To insert a version tag, use the following code:
+To add a version tag, use the following HTML `span`:
 
 ~~~
 <span class="version-tag">New in vX.X:</span>
 ~~~
 
+Version tags can be put at the end of a heading, if the whole section describes something that is new or updated. Otherwise, version tags can introduce a paragraph, sentence, or description in a table cell.
+
 Note: Version tags cannot be used in bulleted lists items. To denote a new feature in a bulleted list, start the bulleted item with "**New in vX.X:**".
```
stbof, line 644: A drive by comment. The  Style and tone section of this guide says: Use active voice instead of passive.  Could we observe that in the Style Guide as well?

Version tags can be put at the end of a heading -> You can put version tags at the end of a heading.

Note: Version tags cannot be used in bulleted lists items.  -> Note: Do not use version tags  in bulleted list items.

StyleGuide.md
```diff
@@ -631,19 +631,21 @@ Examples help show the feature in action. Examples follow a basic format:
...
 
-#### Examples
+If a feature is new in a GA release, use the major release number for the release version tag (e.g., `New in v22.1`).
 
-- [`CREATE TABLE`](https://www.cockroachlabs.com/docs/stable/create-table.html)
+If a feature has been backported to a previous version in a patch release, use the minor release number for the release version tag (e.g., `New in v21.1.9`).
```
kathancox, line 648: So, if we have a new feature in v22.1, for example, and it has been backported to v21.1.16, would we include `New in v21.1.16` in the v22.1 docs as well? (Just making sure I understand correctly!)

StyleGuide.md
ericharmeling: Great point! Done.

StyleGuide.md
ericharmeling: Great question. No, I am suggesting that we put a `New in v22.1` in the v22.1 docs and a `New in v21.1.16` in the v21.1 docs (and there would likely also be a `New in v21.2.x` in the v21.2 docs). I didn't really make that explicit though - just updated the PR.

rmloveland (approved):
LGTM

PR #13280, opened 2022-03-17
Add FAQ entry on using certificates when connecting to Serverless.

ianjevans:


cockroachcloud/serverless-faqs.md
```diff
@@ -105,6 +105,14 @@ You can submit feedback or log any bugs you find through [this survey](https://f
...
+
+Some drivers and ORMs don't use the system root certificates, however. In those cases, you need to download a root certificate file and configure your client to use that certificate when connecting to your cluster. You can [download the certificate](connect-to-a-serverless-cluster.html#step-2-connect-to-your-cluster) by following the instructions in the {{ site.data.products.db }} Console. Configure your client to use this certificate (for example, by setting `sslrootcert=<path to the root certificate>` in your connection string) and to use SSL (for example, by setting `sslmode=verify-full` in your connection string) to connect to your cluster.
```
stbof, line 112:
```suggestion
However, some drivers and ORMs don't use the system root certificates. In those cases, you need to download a root certificate file and configure your client to use that certificate when connecting to your cluster. You can [download the certificate](connect-to-a-serverless-cluster.html#step-2-connect-to-your-cluster) by following the instructions in the {{ site.data.products.db }} Console. Configure your client to use this certificate (for example, by setting `sslrootcert=<path to the root certificate>` in your connection string) and to use SSL (for example, by setting `sslmode=verify-full` in your connection string) to connect to your cluster.
```

stbof (approved):
Minor wording suggestion.

cockroachcloud/serverless-faqs.md
stbof: Are there other ways to configure the client?

ianjevans: There's no standard way, so it depends on the driver/ORM. Many use the libpq connection string format that we favor, but others (JDBC and C# prominently) use different connection strings or have individual parameters.

cockroachcloud/connect-to-a-serverless-cluster.md
```diff
@@ -62,7 +62,7 @@ To connect to your cluster with the [built-in SQL client](../{{site.versions["st
 
     {% include cockroachcloud/download-the-binary.md %}
 
-1. In your terminal, run the second command from the dialog to create a new `certs` directory on your local machine and download the CA certificate to that directory:
+1. In your terminal, run the second command from the dialog to download the CA certificate to the default PostgreSQL certificate directory:
```
gemma-shay, line 65: Looking at the current **Connect** dialog, it seems like the second command was removed?

PR #12729, opened 2022-01-13
DOC-1045, DOC-1046: Added information about NOT VISIBLE flag for columns

nickvigilante:
Fixes DOC-1045, DOC-1046

Missing SQL grammar diagram.

PR #13340, opened 2022-03-24
Add docs for Row-Level TTL

rmloveland:
Fixes DOC-2252

Summary of changes:

- Add a new page, 'Batch Delete Expired Data with Row-Level TTL', which
  explains how to use the new TTL feature in v22.1, along with how
  it (roughly) works via examples, some limitations, etc.

- Add 'Row-Level TTL' to the list of supported features as a "common
  extension"

- Add links to the Row-Level TTL docs from various other places that
  mention "bulk-deletish" things


rmloveland:
@otan wanted to mention I did look through [this list of your PRs](https://github.com/cockroachdb/cockroach/pulls?q=is:pr+ttl+author:otan+is:closed) and there were some things I wasn't sure about.  They may need to be added, but I'll need more info.  Here are some questions:

- I think there was one about non-descending PKs not working yet? Does TTL still work OK with UUID PKs?  b/c that's what I used in the examples

- also not sure if there's a more proper way to list out the TTL cluster settings than what I showed using the example there

- wasn't sure how best to describe the validation logic around which storage params need to be set/unset so other ones can be set/unset.  I tried to capture the straightforward errors but happy to update if there's something missing there

- finally just anything glaring you see wrong or missing would be great to know about

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+| `ttl_range_concurrency`                                  | How many concurrent batches of expired rows are being worked on at a time. Default: `cpu_core_count`. Minimum: 1.                                                            |
+| `ttl_row_stats_poll_interval`                            | If set, counts rows and expired rows on the table to report as Prometheus metrics while the TTL job is running. Unset by default, meaning no stats are fetched and reported. |
+| `ttl_pause` <a name="param-ttl-pause"></a>               | If set, stops the TTL job from executing.                                                                                                                                    |
+| `ttl_job_cron` <a name="param-ttl-job-cron"></a>         | Frequency at which the TTL job runs, specified using [CRON syntax](https://cron.help).                                                                                       |
+| `ttl_automatic_column`                                   | If set, use the value of the `crdb_internal_expiration` hidden column. This is unset when `ttl_expiration_expression` is used.                                               |
```
otan, line 93: Unfortunately `ttl_expiration_expression` is not available yet, so it just appears and is mostly unconfigurable.

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+| `ttl` <a name="param-ttl"></a>                           | Signifies if a TTL is active. Automatically set.                                                                                                                             |
+| `ttl_select_batch_size`                                  | How many rows to [select](select-clause.html) at one time from the set of expired rows. Default: 500. Minimum: 1.                                                            |
+| `ttl_delete_batch_size`                                  | How many rows to [delete](delete.html) at a time. Default: 100. Minimum: 1.                                                                                                  |
+| `ttl_delete_rate_limit`                                  | Maximum number of rows to be deleted per second (rate limit). Default: 0 (no limit).                                                                                         |
+| `ttl_range_concurrency`                                  | How many concurrent batches of expired rows are being worked on at a time. Default: `cpu_core_count`. Minimum: 1.                                                            |
```
otan, line 89: The default is actually 1 now. If it helps to document, we split scans by ranges, we only process `ttl_range_concurrency` ranges at a time.

_includes/v22.1/known-limitations/row-level-ttl-limitations.md
```diff
@@ -0,0 +1,2 @@
+- You cannot use [foreign keys](foreign-key.html) to create references to or from a table that uses Row-Level TTL.
+- [`SELECT` queries](selection-queries.html) against tables with Row-Level TTL enabled do not filter out expired rows from the result set. This feature may be added in a future release. For now, follow the instructions in [Filter out expired rows from a selection query](row-level-ttl.html#filter-out-expired-rows-from-a-selection-query).
```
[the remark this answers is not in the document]
otan: i would also probably add the following (if we care about performance):

* We don't execute the deletes to be close to the range, e.g. something from europe-west may delete from us-east.
* Deletes do not make use of any available index on crdb_internal_expiration which would speed things up.
* There is no way to customize a delete expiration yet

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+             |     CONSTRAINT ttl_test_pkey PRIMARY KEY (id ASC)
+             | ) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:03:00':::INTERVAL)
+(1 row)
+~~~
+
+## TTL storage parameters
+
+The settings that control the behavior of Row-Level TTL are provided using [storage parameters](sql-grammar.html#opt_with_storage_parameter_list). These parameters can be set during table creation using [`CREATE TABLE`](#create-a-table-with-row-level-ttl), added to an existing table using the [`ALTER TABLE`](#add-row-level-ttl-to-an-existing-table) statement, or [reset to default values](#reset-a-storage-parameter-to-its-default-value).
+
+| Description                                              | Option                                                                                                                                                                       |
```
otan, line 82: some of these settings (as you say) have a default of the cluster setting, instead of the defaults mentioned here. maybe worth adding a column to signify that.

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+title: Batch Delete Expired Data with Row-Level TTL
+summary: Automatically delete rows of expired data older than a specified interval.
+toc: true
+keywords: ttl,delete,deletion,bulk deletion,bulk delete,expired data,expire data,time to live,row-level ttl,row level ttl
+docs_area: develop
+---
+
+{% include {{page.version.version}}/sql/row-level-ttl.md %}
+
+By using Row-Level TTL, you can avoid the complexity of writing and managing scheduled jobs from the application layer to mark rows as expired and perform the necessary deletions. Doing it yourself can become complicated due to the need to balance the timeliness of the deletions vs. the potentially negative performance impact of those deletions on foreground traffic from your application.
```
otan, line 11: I would add this is a feature in preview mode - it may not satisfy production requirements due to some of the performance issues that may occur in large volume data sets (subject to feedback from @vy-ton)

otan (commented):
looking good, thanks rich!

> I think there was one about non-descending PKs not working yet? Does TTL still work OK with UUID PKs? b/c that's what I used in the examples

yes, it does not work with a PRIMARY KEY that looks like this: `PRIMARY KEY (id DESC)`, `PRIMARY KEY (id ASC, id2 DESC)`

> also not sure if there's a more proper way to list out the TTL cluster settings than what I showed using the example there

i've added a suggestion, but yeah that's more or less it.

> wasn't sure how best to describe the validation logic around which storage params need to be set/unset so other ones can be set/unset. I tried to capture the straightforward errors but happy to update if there's something missing there

basically, `ttl` and `ttl_automatic_column` and `ttl_expire_after` is always set. this will change in future.

> finally just anything glaring you see wrong or missing would be great to know about

gotcha!

_includes/v22.1/known-limitations/row-level-ttl-limitations.md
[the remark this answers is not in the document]
rmloveland: thanks Oliver, updated this a bit based on that info

the only one I left out is "customize a delete expiration yet" - do you mind explaining more about what that means?  do you mean customizing the delete expiration per row (somehow) instead of setting it for the whole table using the `ttl_expire_after`?

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+
+- Outbox pattern: When events are written to an outbox table and published to an external system like [Kafka](https://en.wikipedia.org/wiki/Apache_Kafka), those events must be deleted to prevent unbounded growth in the size of the outbox table.
+
+## How it works
+
+At a high level, Row-Level TTL works by:
+
+1. Issuing a [selection query](selection-queries.html) at a [historical timestamp](as-of-system-time.html), yielding a set of rows that are eligible for deletion.
+1. Given the set of rows eligible for deletion, split what would logically be a single [`DELETE`](delete.html) statement to remove those rows into multiple SQL queries that will eventually have the equivalent effect by operating on a subset of the eligible rows.
+1. As part of the above process, deciding how many rows to [`SELECT`](select-clause.html) and [`DELETE`](delete.html) at once in each of the above queries.
```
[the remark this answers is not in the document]
rmloveland: sounds good - updated!

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+1. Running the SQL queries described above in parallel as [background jobs](show-jobs.html).
+1. All the while applying a rate limit to control the rate of those background deletion queries to minimize the performance impact on foreground application queries.
```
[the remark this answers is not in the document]
rmloveland: updated to note that - assume you mean this admission control? https://www.cockroachlabs.com/docs/stable/architecture/admission-control.html

anyway that's what I wrote, please let me know if it's wrong

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+  description TEXT,
+  inserted_at TIMESTAMP default current_timestamp()
+) WITH (ttl_expire_after = '3 minutes');
+~~~
+
+has the following effects:
+
+<a name="crdb-internal-expiration"></a>
+
+1. Creates a repeating [scheduled job](#view-ttl-jobs) for the `events` table.
```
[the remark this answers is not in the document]
rmloveland: Updated! thanks for the catch. Started with `events` here, then switched.

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,323 @@
...
+  table_name |                                                                                           create_statement
+-------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+  ttl_test   | CREATE TABLE public.ttl_test (
+             |     id UUID NOT NULL DEFAULT gen_random_uuid(),
+             |     description STRING NULL,
+             |     inserted_at TIMESTAMP NULL DEFAULT current_timestamp():::TIMESTAMP,
+             |     crdb_internal_expiration TIMESTAMPTZ NOT VISIBLE NOT NULL DEFAULT current_timestamp():::TIMESTAMPTZ + '00:03:00':::INTERVAL ON UPDATE current_timestamp():::TIMESTAMPTZ + '00:03:00':::INTERVAL,
+             |     CONSTRAINT ttl_test_pkey PRIMARY KEY (id ASC)
+             | ) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:03:00':::INTERVAL)
+(1 row)
```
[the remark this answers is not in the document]
rmloveland: Added!

v22.1/row-level-ttl.md
rmloveland: Just deleted that second sentence for now.  No sense getting into it too much until later release, I think.  let me know if you disagree

v22.1/row-level-ttl.md
rmloveland: It helps!  I rewrote this sentence to use what I understood from your comment.  PTAL

v22.1/row-level-ttl.md
rmloveland: I added that it is in *beta*, along with your performance comment, and linked to the (updated) limitations section which now mentions some performance things

rmloveland:
@otan wanted to mention I am on PTO next week so I will be able to make updates when I return from that

@vy-ton adding you as a reviewer now, given the above you have a week or so to take a peek if you want!

_includes/v22.1/known-limitations/row-level-ttl-limitations.md
```diff
@@ -0,0 +1,5 @@
+- You cannot use [foreign keys](foreign-key.html) to create references to or from a table that uses Row-Level TTL.
+- Any queries you run against tables with Row-Level TTL enabled do not filter out expired rows from the result set (this includes [`UPDATE`s](update.html) and [`DELETE`s](delete.html)). This feature may be added in a future release. For now, follow the instructions in [Filter out expired rows from a selection query](row-level-ttl.html#filter-out-expired-rows-from-a-selection-query).
+- The queries executed by Row-Level TTL are not yet optimized for performance:
+  - They do not use any indexes that may be available on the [`crdb_internal_expiration` column](row-level-ttl.html#crdb-internal-expiration).
+  - They do not take into account [node localities](cockroach-start.html#locality).
```
otan, line 5: ah one more, "all deletes are run on a single node, instead of being distributed" if that makes sense

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+| `ttl_row_stats_poll_interval`                            | If set, counts rows and expired rows on the table to report as Prometheus metrics while the TTL job is running. Unset by default, meaning no stats are fetched and reported. | N/A                                 |
+| `ttl_pause` <a name="param-ttl-pause"></a>               | If set, stops the TTL job from executing.                                                                                                                                    | N/A                                 |
+| `ttl_job_cron` <a name="param-ttl-job-cron"></a>         | Frequency at which the TTL job runs, specified using [CRON syntax](https://cron.help).                                                                                       | N/A                                 |
+| `ttl_automatic_column`                                   | If set, use the value of the `crdb_internal_expiration` hidden column.                                                                                                       | N/A                                 |
```
otan, line 110: maybe as a note "currently always set"

_includes/v22.1/known-limitations/row-level-ttl-limitations.md
otan: ah, another limitation: "unable to customise the TTL based on the values of other columns on that row"

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+
+~~~
+          id         |        label         | schedule_status |        next_run        |  state  | recurrence | jobsrunning | owner |            created            |     command
+---------------------+----------------------+-----------------+------------------------+---------+------------+-------------+-------+-------------------------------+-------------------
+  747608117920104449 | sql-stats-compaction | ACTIVE          | 2022-03-25 16:00:00+00 | pending | @hourly    |           0 | node  | 2022-03-25 15:31:31.444067+00 | {}
+  747609229470433281 | row-level-ttl-112    | ACTIVE          | 2022-03-25 16:00:00+00 | NULL    | @hourly    |           0 | root  | 2022-03-25 15:37:10.613056+00 | {"tableId": 112}
+(2 rows)
+~~~
+
+## TTL storage parameters
```
otan, line 95: is it worth also mentioning the metrics one can use to monitor the effectiveness of the TTL? https://github.com/cockroachdb/cockroach/blob/b9a675266303528d71b388ac35dfb87f145cc7a8/pkg/sql/ttl/ttljob/ttljob.go#L167-L244 contains a list of all metrics.

Note the metrics by default are aggregated. If the user wishes to have metrics broken down by table, if `set cluster setting server.child_metrics.enabled = true;` is set AND the table has the storage parameter (ttl_label_metrics=True) (this is necessary to reduce cardinality of the metrics).

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+
+To specify a TTL when creating a table, use the SQL syntax shown below. For example, to create a new table with rows that expire after 15 minutes, execute a statement like the following:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+CREATE TABLE events (
+  id UUID PRIMARY KEY default gen_random_uuid(),
+  description TEXT,
+  inserted_at TIMESTAMP default current_timestamp()
+) WITH (ttl_expire_after = '3 minutes');
```
otan, line 126: nit: 15, not 3, due to

> that expire after 15 minutes

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+Insert some data; it should work as expected:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+INSERT INTO events (description) VALUES ('a thing'), ('another thing'), ('yet another thing');
+~~~
+
+~~~
+INSERT 3
+~~~
```
otan, line 142: maybe add a `SELECT *, crdb_internal_expiration FROM events` to show the rows and their expirations

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+~~~ sql
+SELECT * FROM [SHOW SCHEDULES] WHERE label LIKE 'row-level-ttl-%';
+~~~
+
+~~~
+          id         |       label       | schedule_status |        next_run        | state | recurrence | jobsrunning | owner |            created            |     command
+---------------------+-------------------+-----------------+------------------------+-------+------------+-------------+-------+-------------------------------+-------------------
+  745396332264194049 | row-level-ttl-114 | ACTIVE          | 2022-03-17 21:00:00+00 | NULL  | @hourly    |           0 | root  | 2022-03-17 20:01:48.033267+00 | {"tableId": 114}
+(1 row)
+~~~
```
otan, line 169: maybe add a `SHOW JOBS` when the TTL job kicks off to show running jobs as well?

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+
+~~~
+ALTER TABLE
+~~~
+
+### View TTL storage parameters on a table
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+SELECT relname, reloptions FROM pg_class WHERE relname = 'events';
```
otan, line 188: `SHOW CREATE TABLE` is also another valid way

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,342 @@
...
+  inserted_at TIMESTAMP default current_timestamp()
+) WITH (ttl_expire_after = '3 minutes');
+~~~
+
+has the following effects:
+
+<a name="crdb-internal-expiration"></a>
+
+1. Creates a repeating [scheduled job](#view-ttl-jobs) for the `ttl_test` table.
+2. Adds a hidden column `crdb_internal_expiration` of type [`TIMESTAMPTZ`](timestamp.html) to represent the TTL.
```
otan, line 57: do we have a hyperlink to what a hidden column does (i.e. doesn't show up on `*`). if so, let's link to that!

otan (commented):
Thanks for the updates rich. Caught some extra stuff on the second pass (sorry about that)!

_includes/v22.1/known-limitations/row-level-ttl-limitations.md
rmloveland: Added both of these!

v22.1/row-level-ttl.md
rmloveland: Added!

v22.1/row-level-ttl.md
rmloveland: Yes I think it's def worth mentioning! Added a table with the metrics and those notes, PTAL

Would these metrics be visible in the UI using the [advanced debug page](https://www.cockroachlabs.com/docs/v21.2/ui-debug-pages)?  

Or is this the type of thing that would be piped out to Prometheus a la https://www.cockroachlabs.com/docs/v21.2/monitor-cockroachdb-with-prometheus.html  so users should expect to get at them that way?

(Sorry, I don't know much about CRDB metrics things)

v22.1/row-level-ttl.md
rmloveland: Fixed by updating to say "other than the default" and linking to that param's docs, which list the default.  thanks for catching this!

v22.1/row-level-ttl.md
rmloveland: Did it!

v22.1/row-level-ttl.md
rmloveland: Sorry, I don't understand.  I typed in `SHOW JOBS` but only saw the schema change job about changing the TTL storage param

`SHOW JOBS`

```
        job_id       |   job_type    |                                                        description                                                         | statement | user_name |  status   | running_status |          created           |          started           |          finished          |          modified          | fraction_completed | error | coordinator_id |      trace_id       |          last_run          |          next_run          | num_runs | execution_errors
---------------------+---------------+----------------------------------------------------------------------------------------------------------------------------+-----------+-----------+-----------+----------------+----------------------------+----------------------------+----------------------------+----------------------------+--------------------+-------+----------------+---------------------+----------------------------+----------------------------+----------+-------------------
  750771143879491585 | SCHEMA CHANGE | ALTER TABLE test.public.events SET (ttl_expire_after = '90 days')                                                          |           | root      | succeeded | NULL           | 2022-04-05 19:39:30.317168 | 2022-04-05 19:39:30.448445 | 2022-04-05 19:39:30.458783 | 2022-04-05 19:39:30.457604 |                  1 |       |              1 | 4535995618078903066 | 2022-04-05 19:39:30.448446 | 2022-04-05 19:40:00.448446 |        1 | {}
```

v22.1/row-level-ttl.md
rmloveland: Updated to lead with an example with `SHOW CREATE TABLE`, left the other as an alternative

v22.1/row-level-ttl.md
rmloveland: I looked and we actually don't appear to!  Filed https://cockroachlabs.atlassian.net/browse/DOC-3200 to address as a followup task (not sure when, but it's filed at least :-} )

otan: following some PR advice, i think we should follow the DDL here and use `NOT VISIBLE` column externally here

v22.1/row-level-ttl.md
otan: they are both! the advanced debug page shows all these stats, and prometheus scrapes these stats from the same data source (from the `/_status/vars` endpoint on the db console)

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,400 @@
...
+| `jobs.row_level_ttl.select_duration`      | Duration for select requests during row level TTL.               | `nanoseconds`        | Histogram |
+| `jobs.row_level_ttl.delete_duration`      | Duration for delete requests during row level TTL.               | `nanoseconds`        | Histogram |
+| `jobs.row_level_ttl.rows_selected`        | Number of rows selected for deletion by the row level TTL job.   | `num_rows`           | Counter   |
+| `jobs.row_level_ttl.rows_deleted`         | Number of rows deleted by the row level TTL job.                 | `num_rows`           | Counter   |
+| `jobs.row_level_ttl.num_active_ranges`    | Number of active workers attempting to delete for row level TTL. | `num_active_workers` | Count     |
+| `jobs.row_level_ttl.total_rows`           | Approximate number of rows on the TTL table.                     | `total_rows`         | Count     |
+| `jobs.row_level_ttl.total_expired_rows`   | Approximate number of expired rows on the TTL table.             | `total_expired_rows` | Count     |
+
+By default, these metrics are aggregated. If you want to have metrics broken out by table, you must take the following steps:
```
otan, line 129: Maybe "If you want to have metrics labelled by the table name at the risk of added cardinality`

v22.1/row-level-ttl.md
```diff
@@ -0,0 +1,400 @@
...
+
+- Set the `server.child_metrics.enabled` [cluster setting](cluster-settings.html) to `true`.
+- Set the `ttl_label_metrics` storage parameter to `true`. (This step is necessary to reduce the cardinality of the metrics.)
```
otan, line 132: Just doing the above means all non-ttl_label_metrics'd tables will look like `jobs.row_level_ttl.total_expired_rows{relation_name=default}`. Setting this flag makes it broken out by it's extra table and *adds* cardinality.

(please feel free to give it a shot! it's confusing! you can look at the [`/status/varz`](http://localhost:8081/_status/vars)` endpoint to see how these metric names get reported over these settings.

v22.1/row-level-ttl.md
otan: `SHOW JOBS` will show the job when the TTL starts running. you may need to set `ttl_job_cron` to `* * * * *` and then wait a minute or two to show up.

otan (commented):
thanks rich! looking close, just some more work on the metrics bit i'd say!

rmloveland (commented):
thanks Oliver! made changes based on your comments - PTAL

(also added a question about whether my query to filter out expired rows had the predicate backwards? see below)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340#-:-N-8WGcn6u4z1PXcIADU:b-sc2nud)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan and @vy-ton)

___
*[v22.1/row-level-ttl.md, line 57 at r2](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r2-57:-N-8N82N0byNHJEhXK57:b-egnmfm) ([raw file](https://github.com/cockroachdb/docs/blob/6a1edb87772490f20a09d06f30860f8391b6e5b3/v22.1/row-level-ttl.md#L57)):*

following some PR advice, i think we should follow the DDL here and use `NOT VISIBLE` column externally here
</blockquote></details>

do you mean we should say "Adds a `NOT VISIBLE` column called `crdb_internal_expiration` ..."

I went ahead and called it that, let me know if that was wrong

Also I discovered we have a docs PR in progress on `NOT VISIBLE` columns, which we can link to ASAP once it is in: https://github.com/cockroachdb/docs/pull/12729
___
*[v22.1/row-level-ttl.md, line 95 at r2](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r2-95:-N-8OzkE8lIw3SjtunmU:b2toxu2) ([raw file](https://github.com/cockroachdb/docs/blob/6a1edb87772490f20a09d06f30860f8391b6e5b3/v22.1/row-level-ttl.md#L95)):*

they are both! the advanced debug page shows all these stats, and prometheus scrapes these stats from the same data source (from the `/_status/vars` endpoint on the db console)
</blockquote></details>

NOICE - ok I added a sentence about them being available on the debug page or in Prometheus if one uses that
___
*[v22.1/row-level-ttl.md, line 169 at r2](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r2-169:-N-8URcy5OZIBq1eZvuL:bhls1l0) ([raw file](https://github.com/cockroachdb/docs/blob/6a1edb87772490f20a09d06f30860f8391b6e5b3/v22.1/row-level-ttl.md#L169)):*

`SHOW JOBS` will show the job when the TTL starts running. you may need to set `ttl_job_cron` to `* * * * *` and then wait a minute or two to show up.
</blockquote></details>

Oh nice! yeah I see it now

As a result of this I revamped things a bit and made these into two different examples: 'show scheduled TTL jobs' and 'show running TTL jobs' - please let me know what you think!
___
*[v22.1/row-level-ttl.md, line 129 at r3](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r3-129:-N-8PEWF0jjQem9oCU03:b-xrbfyu) ([raw file](https://github.com/cockroachdb/docs/blob/caf025494e60276ef8a4d99823057b6fcca5321c/v22.1/row-level-ttl.md#L129)):*

Maybe "If you want to have metrics labelled by the table name at the risk of added cardinality`
</blockquote></details>

Updated to say that!
___
*[v22.1/row-level-ttl.md, line 132 at r3](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r3-132:-N-8SfQK4b5eL2iRuLFM:bfioain) ([raw file](https://github.com/cockroachdb/docs/blob/caf025494e60276ef8a4d99823057b6fcca5321c/v22.1/row-level-ttl.md#L132)):*

Just doing the above means all non-ttl_label_metrics'd tables will look like `jobs.row_level_ttl.total_expired_rows{relation_name=default}`. Setting this flag makes it broken out by it's extra table and *adds* cardinality.

(please feel free to give it a shot! it's confusing! you can look at the [`/status/varz`](http://localhost:8081/_status/vars)` endpoint to see how these metric names get reported over these settings.
</blockquote></details>

Oh yeah I see it now.  thanks for the prompt.  Removed the parenthetical here which, combined with the edit above about it adding more cardinality, hopefully is more true :-) 
___
*[v22.1/row-level-ttl.md, line 316 at r3](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-8QRN6ExDzfZQg7Bmu:-N-8QRN752-Z455LrT2j:b-5jpjkj) ([raw file](https://github.com/cockroachdb/docs/blob/caf025494e60276ef8a4d99823057b6fcca5321c/v22.1/row-level-ttl.md#L316)):*
> ```Markdown
> ~~~ sql
> SELECT * FROM events WHERE crdb_internal_expiration < now();
> ```

Hey Oliver can you check me on something? I just ran through these instructions again and I think this should be `> now()` ( I think I had it flipped) since in this case we only want to see rows whose expiration is in the future (and thus still alive) - right?

(my quick local test seems to confirm this so I updated the code to use `>`  b/c I'm pretty sure this was wrong but would appreciate your confirmation - can revert if needed)

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @otan and @vy-ton)

___
*[v22.1/row-level-ttl.md, line 82 at r1](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r1-82:-N-UDGmz0Z_EHfvpc13p:b-849pej) ([raw file](https://github.com/cockroachdb/docs/blob/ac94881a4cd91f886295e40b3cfec7a480ef8c62/v22.1/row-level-ttl.md#L82)):*

looks good
</blockquote></details>

NOICE
___
*[v22.1/row-level-ttl.md, line 84 at r1](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r1-84:-N-UDiGeDXekIy83EOIx:b-fiqbtg) ([raw file](https://github.com/cockroachdb/docs/blob/ac94881a4cd91f886295e40b3cfec7a480ef8c62/v22.1/row-level-ttl.md#L84)):*

there's also no "default" as it is always specified
</blockquote></details>

Ah!  Copied that from the old RFC I think.  Removed mention of "default" and instead said it is a required parameter to enable TTL
___
*[v22.1/row-level-ttl.md, line 126 at r2](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r2-126:-N-UCzZLC3cscO0Zwfub:bq703db) ([raw file](https://github.com/cockroachdb/docs/blob/6a1edb87772490f20a09d06f30860f8391b6e5b3/v22.1/row-level-ttl.md#L126)):*

there is no default though!
</blockquote></details>

weird, not sure where I got that idea from!  Updated to say "after a 3 minute interval"
___
*[v22.1/row-level-ttl.md, line 129 at r3](https://reviewable.io/reviews/cockroachdb/docs/13340#-MyxzVmN2OdKGI4VQcG1-r3-129:-N-UCrCg6eF01N04CoPT:bqmjwcm) ([raw file](https://github.com/cockroachdb/docs/blob/caf025494e60276ef8a4d99823057b6fcca5321c/v22.1/row-level-ttl.md#L129)):*

maybe clarify with "metrics are aggregated, meaning that all TTL tables will report the metrics under the same label".
</blockquote></details>

Added!
___
*[v22.1/row-level-ttl.md, line 316 at r3](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-8QRN6ExDzfZQg7Bmu:-N-UDF2u2eyGz6FJR_tZ:bgdws1l) ([raw file](https://github.com/cockroachdb/docs/blob/caf025494e60276ef8a4d99823057b6fcca5321c/v22.1/row-level-ttl.md#L316)):*

ah yes, my bad i missed that. use `>` for non-expired rows.
</blockquote></details>

ok thanks!

otan (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @vy-ton)

vy-ton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling and @rmloveland)

___
*[_includes/v22.1/known-limitations/row-level-ttl-limitations.md, line 1 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xEi22BJy2rnJMHoHt:-N-xEi22BJy2rnJMHoHu:b-7g4p8v) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/_includes/v22.1/known-limitations/row-level-ttl-limitations.md#L1)):*
> ```Markdown
> - You cannot use [foreign keys](foreign-key.html) to create references to or from a table that uses Row-Level TTL.
> ```

Let's link Github issues for some of these limitations

* `cannot use foreign keyrs` - https://github.com/cockroachdb/cockroach/issues/76407
* `TTL cannot be customized` - https://github.com/cockroachdb/cockroach/issues/76916
* `not yet optimized for performance` - https://github.com/cockroachdb/cockroach/issues/76914
___
*[v22.1/row-level-ttl.md, line 11 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xC_dd2fXWOiVEc5p3:-N-xC_dd2fXWOiVEc5p4:b-g7ofcl) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L11)):*
> ```Markdown
> {% include {{page.version.version}}/sql/row-level-ttl.md %}
> 
> Row-Level TTL is in **beta** because it may not satisfy production requirements due to performance issues that may occur in large volume data sets. For more information see [Limitations](row-level-ttl.html#limitations).
> ```

`is in preview because it may not satisfy all functional and performance requirements of your workload`

Product Marketing is going to standardize on `preview` instead of `beta`. I'd also like to avoid the prod vs non-prod framing
___
*[v22.1/row-level-ttl.md, line 28 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xGrv84HjQDo5s7s5S:-N-xGrv84HjQDo5s7s5T:b-w7fbct) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L28)):*
> ```Markdown
> 
> - Issuing a [selection query](selection-queries.html) at a [historical timestamp](as-of-system-time.html), yielding a set of rows that are eligible for deletion.
> - Given the set of rows eligible for deletion, split what would logically be a single [`DELETE`](delete.html) statement to remove those rows into multiple SQL queries that will eventually have the equivalent effect by operating on a subset of the eligible rows.
> ```

Found this hard to follow  - we issue batched DELETES for expired row
___
*[v22.1/row-level-ttl.md, line 87 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xIGDI2JFYkkeBrq_S:-N-xIDDvCSIBx-zc2ZvZ:b-85e7sy) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L87)):*
> ```Markdown
> |----------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------------|
> | `ttl_expire_after` <a name="param-ttl-expire-after"></a> | The [interval](interval.html) when a TTL will expire. This parameter is required to enable TTL. Minimum value: '1 microsecond'.                                              | N/A                                 |
> | `ttl` <a name="param-ttl"></a>                           | Signifies if a TTL is active. Automatically set.                                                                                                                             | `sql.ttl.job.enabled`               |
> ```

`sql.ttl.job.enabled` is the cluster equivalent of `ttl_pause` - it does not enable/disable the feature but rather enable/disable the TTL deletion job
___
*[v22.1/row-level-ttl.md, line 88 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xJ1K7AAhmEZUpYWo2:-N-xJ1K7AAhmEZUpYWo3:bo2e8lt) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L88)):*
> ```Markdown
> | `ttl_expire_after` <a name="param-ttl-expire-after"></a> | The [interval](interval.html) when a TTL will expire. This parameter is required to enable TTL. Minimum value: '1 microsecond'.                                              | N/A                                 |
> | `ttl` <a name="param-ttl"></a>                           | Signifies if a TTL is active. Automatically set.                                                                                                                             | `sql.ttl.job.enabled`               |
> | `ttl_select_batch_size`                                  | How many rows to [select](select-clause.html) at one time from the set of expired rows. Default: 500. Minimum: 1.                                                            | `sql.ttl.default_select_batch_size` |
> ```

@otan Is this from the `set of expired rows`? I thought it was the batch size for the selection query mentioned in `How it works`.
___
*[v22.1/row-level-ttl.md, line 94 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xLHqU3hFpg4M0dDvt:-N-xLHqU3hFpg4M0dDvu:b-z191c4) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L94)):*
> ```Markdown
> | `ttl_row_stats_poll_interval`                            | If set, counts rows and expired rows on the table to report as Prometheus metrics while the TTL job is running. Unset by default, meaning no stats are fetched and reported. | N/A                                 |
> | `ttl_pause` <a name="param-ttl-pause"></a>               | If set, stops the TTL job from executing.                                                                                                                                    | N/A                                 |
> | `ttl_job_cron` <a name="param-ttl-job-cron"></a>         | Frequency at which the TTL job runs, specified using [CRON syntax](https://cron.help).                                                                                       | N/A                                 |
> ```

We should mention the default
___
*[v22.1/row-level-ttl.md, line 123 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xK8Vr1Z6oxyWvzjQc:-N-xK8Vr1Z6oxyWvzjQd:bmgjqxm) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L123)):*
> ```Markdown
> ### Create a table with row-level TTL
> 
> To specify a TTL when creating a table, use the SQL syntax shown below. For example, to create a new table with rows that expire after a 3 minute [interval](interval.html), execute a statement like the following:
> ```

Can we make our examples with a TTL of weeks or months?
___
*[v22.1/row-level-ttl.md, line 197 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xL0Im5_UPH6Y7PPjw:-N-xL0Im5_UPH6Y7PPjx:b7m0ahx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L197)):*
> ```Markdown
> ### View running TTL jobs
> 
> You can use [`SHOW JOBS`](show-jobs.html) to see any running TTL jobs by executing the following query:
> ```

Mention DBConsole Jobs page as well

In addition, I believe @otan is improving the `description` field for TTL jobs so we will need to update the screenshot when that happens

vy-ton:
@rmloveland Awesome docs!

Let's merge this PR but then I'd like to discuss with you whether features like TTL should be split across multiple pages. There's a lot covered in a single page including:
* What is TTL and why might it be useful?
* High-level architecture and how it works
* Reference and exmaples

vy-ton (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling and @rmloveland)

___
*[v22.1/row-level-ttl.md, line 180 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xvv__Eddckgx-DzCJ:-N-xvv__Eddckgx-DzCK:b-ewff71) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L180)):*
> ```Markdown
> ### View scheduled TTL jobs
> 
> You can use [`SHOW SCHEDULES`](show-schedules.html) to view all TTL-related scheduled jobs by executing the following query:
> ```

Need to edit https://www.cockroachlabs.com/docs/v21.2/show-schedules.html since it no longer describes just backup schedules

otan (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @rmloveland, and @vy-ton)

___
*[v22.1/row-level-ttl.md, line 88 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xJ1K7AAhmEZUpYWo2:-N-yMAlFEdT39W5ZzS6I:bdhgk3x) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L88)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

@otan Is this from the `set of expired rows`? I thought it was the batch size for the selection query mentioned in `How it works`.
</blockquote></details>

oh yes, this is from all rows.

ericharmeling (approved):
Very well done, Rich! Exciting new feature in CRDB, this. And these docs are pretty incredible.

I just had a few very minor nits. But other than those, LGTM.

> Reviewed 6 of 10 files at r1, 2 of 4 files at r2, 1 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340#-:-N-yOnI3873AwFRIe4Ul:b-o1i9xe)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @rmloveland and @vy-ton)

___
*[_includes/v22.1/sql/row-level-ttl.md, line 1 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yP6d97HuFzBBnA9ts:-N-yP6d97HuFzBBnA9tt:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/_includes/v22.1/sql/row-level-ttl.md#L1)):*
> ```Markdown
> <span class="version-tag">New in v22.1</span>: CockroachDB has **beta** support for Time to Live ("TTL") expiration on table rows, also known as _Row-Level TTL_. Row-level TTL is a mechanism whereby rows from a table are considered "expired" and can be automatically deleted once those rows have been stored longer than a specified expiration time.
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/known-limitations.md, line 200 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yQ_e10i8r4sLqkUDg:-N-yQ_e10i8r4sLqkUDh:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/known-limitations.md#L200)):*
> ```Markdown
> {% include {{page.version.version}}/sql/expressions-as-on-conflict-targets.md %}
> 
> ### Row-level TTL limitations
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/row-level-ttl.md, line 15 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPUkQ9IrBpBxHHY5x:-N-yPUkQ9IrBpBxHHY5y:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L15)):*
> ```Markdown
> By using Row-Level TTL, you can avoid the complexity of writing and managing scheduled jobs from the application layer to mark rows as expired and perform the necessary deletions. Doing it yourself can become complicated due to the need to balance the timeliness of the deletions vs. the potentially negative performance impact of those deletions on foreground traffic from your application.
> 
> Use cases for row-level TTL include:
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/row-level-ttl.md, line 28 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yQqKlCmWIhU-5EnsG:-N-yQqKlCmWIhU-5EnsH:bkrj939) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L28)):*
> ```Markdown
> 
> - Issuing a [selection query](selection-queries.html) at a [historical timestamp](as-of-system-time.html), yielding a set of rows that are eligible for deletion.
> - Given the set of rows eligible for deletion, split what would logically be a single [`DELETE`](delete.html) statement to remove those rows into multiple SQL queries that will eventually have the equivalent effect by operating on a subset of the eligible rows.
> ```

nit: grammar consistency



_Suggestion:_
```Markdown
splitting
```
___
*[v22.1/row-level-ttl.md, line 37 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yRBhN-DZpJAuOpKIk:-N-yRBhN-DZpJAuOpKIl:b8a9ykt) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L37)):*
> ```Markdown
> ## When are rows deleted?
> 
> Once rows are expired (that is, are older than the specified [TTL interval](#param-ttl-expire-after)), they are eligible to be deleted. However, eligible rows may not be deleted right away - instead, they are scheduled for deletion using a [background job](#view-scheduled-ttl-jobs) that is run at the interval defined by the `ttl_job_cron` [storage parameter](#ttl-storage-parameters).
> ```

nit: grammar



_Suggestion:_
```Markdown
away. Instead
```
___
*[v22.1/row-level-ttl.md, line 86 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yRjPu4NtfyyDYyGhw:-N-yRjPu4NtfyyDYyGhx:bzdqul4) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L86)):*
> ```Markdown
> | Description                                              | Option                                                                                                                                                                       | Associated cluster setting          |
> |----------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------------|
> | `ttl_expire_after` <a name="param-ttl-expire-after"></a> | The [interval](interval.html) when a TTL will expire. This parameter is required to enable TTL. Minimum value: '1 microsecond'.                                              | N/A                                 |
> ```

nit: code style



_Suggestion:_
```Markdown
`'1 microsecond'`
```
___
*[v22.1/row-level-ttl.md, line 121 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPeIR8mG9Fxd0fGtN:-N-yPeIS17EZ9-jPBkgu:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L121)):*
> ```Markdown
> ## Examples
> 
> ### Create a table with row-level TTL
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/row-level-ttl.md, line 165 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPiNDB-SkuunpfZPM:-N-yPiNDB-SkuunpfZPN:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L165)):*
> ```Markdown
> ~~~
> 
> ### Add row-level TTL to an existing table
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/row-level-ttl.md, line 333 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPorY2JTL1K9trba8:-N-yPorY2JTL1K9trba9:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L333)):*
> ```Markdown
> ~~~
> 
> ### Remove row-level TTL from a table
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```
___
*[v22.1/row-level-ttl.md, line 377 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPrhM5_UOMLvBFLR7:-N-yPrhM5_UOMLvBFLR8:b7pmccf) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L377)):*
> ```Markdown
> ~~~
> 
> If you try to reset a [TTL storage parameter](#ttl-storage-parameters) but resetting that paraemeter would result in an invalid state of the TTL subsystem, CockroachDB will signal an error. For example, there is only one way to [remove row-level TTL from a table](#remove-row-level-ttl-from-a-table). If you try to remove the TTL from a table by resetting the `ttl_expire_after` storage parameter you set earlier, you will get the following error:
> ```

nit: casing consistency



_Suggestion:_
```Markdown
Row-Level TTL
```

vy-ton:
I cross-referenced with [the spec](https://docs.google.com/document/d/1NBXzzXMWSEopll27h_ZF0jyImXX3m4GEfpR36ixJc4c/edit#heading=h.qvlozyo3kma3) and wanted to call out a few more items:

- [ ] Adding TTL to an existing table will result in a schema change that sets `crdb_internal_expiration` for all rows. Depending on table size, this could be an expensive operation.
- [ ] Add example for user story `I want to set or override the row-level TTL on individual row(s) so that I can change the TTL expiration for compliance or troubleshooting reasons.`
- [ ] On `UPDATE`, a row with a manual row-level TTL override will lose the override value and expiration will be set to now() + TTL.

rmloveland (commented):
Thanks for the review Vy!

Re: this list from the spec, I've added this info: the first to the 'add to an existing table' example, and the latter two to a new example about overriding the TTL for a single row - PTAL

I have addressed your other inline comments below, I think - PTAL as well

Re: splitting this doc out, I think we could definitely do that going forward.  Maybe we could do a big split-up by the next release when it moves out of preview, once we (presumably) get more user feedback and questions etc.  It definitely kind of grew a bit larger than I might have planned!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340#-:-N02q4KI3qrBTa5XCmvt:b-fuwzwg)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @otan, and @vy-ton)

___
*[_includes/v22.1/known-limitations/row-level-ttl-limitations.md, line 1 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xEi22BJy2rnJMHoHt:-N02Dujo-0bVIjF4CXd3:b4a2h7q) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/_includes/v22.1/known-limitations/row-level-ttl-limitations.md#L1)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Let's link Github issues for some of these limitations

* `cannot use foreign keyrs` - https://github.com/cockroachdb/cockroach/issues/76407
* `TTL cannot be customized` - https://github.com/cockroachdb/cockroach/issues/76916
* `not yet optimized for performance` - https://github.com/cockroachdb/cockroach/issues/76914

</blockquote></details>

Thanks!  Added links
___
*[_includes/v22.1/sql/row-level-ttl.md, line 1 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yP6d97HuFzBBnA9ts:-N02-bvf4nX692N9qJW7:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/_includes/v22.1/sql/row-level-ttl.md#L1)):*

nit: casing consistency
</blockquote></details>

Fixed
___
*[v22.1/known-limitations.md, line 200 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yQ_e10i8r4sLqkUDg:-N02-g2WCkMd2JJivjQN:bt2vr0y) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/known-limitations.md#L200)):*

nit: casing consistency
</blockquote></details>

Fixed!
___
*[v22.1/row-level-ttl.md, line 11 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xC_dd2fXWOiVEc5p3:-N02-tF6-fc8IDgx-zoO:b-zctzit) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L11)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

`is in preview because it may not satisfy all functional and performance requirements of your workload`

Product Marketing is going to standardize on `preview` instead of `beta`. I'd also like to avoid the prod vs non-prod framing


</blockquote></details>

Makes sense - updated!
___
*[v22.1/row-level-ttl.md, line 15 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPUkQ9IrBpBxHHY5x:-N0200X_5KNjFSKFocwj:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L15)):*

nit: casing consistency
</blockquote></details>

Fixed
___
*[v22.1/row-level-ttl.md, line 28 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xGrv84HjQDo5s7s5S:-N020Qdy9Z_zvICoGedQ:bptgyas) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L28)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Found this hard to follow  - we issue batched DELETES for expired row
</blockquote></details>

changed this line to "- Issuing batched [`DELETE`](delete.html) statements for the expired rows."
___
*[v22.1/row-level-ttl.md, line 28 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yQqKlCmWIhU-5EnsG:-N020WXrAz6aSvtQ3DIh:bn44i5t) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L28)):*

nit: grammar consistency
</blockquote></details>

Fixed by above removal
___
*[v22.1/row-level-ttl.md, line 37 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yRBhN-DZpJAuOpKIk:-N020c4B9Xr9AWZKUHMe:b33qp0t) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L37)):*

nit: grammar
</blockquote></details>

HOW DARE YOU COMMENT ON MY RANDOM INSERTIONS OF DASHES ERICCCCCCC

OK fixed :-) 
___
*[v22.1/row-level-ttl.md, line 86 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yRjPu4NtfyyDYyGhw:-N020rx6D36Cb-K8yWxA:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L86)):*

nit: code style
</blockquote></details>

Fixed
___
*[v22.1/row-level-ttl.md, line 87 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xIGDI2JFYkkeBrq_S:-N022uw2B9Hu9FrXMWES:bous7iz) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L87)):*

oh missed this - yes there's no default for this, and yes it refers to a cluster wide span of the job.
ttl_pause does *not* override sql.ttl.job.enabled if the latter is set to true, so it's not fair for it as a default. it is meant to be more a "feature flag" which is a quite disable on the whole cluster, so may be worth mentioning separately.
</blockquote></details>

Ah!  Changed this table field to N/A

Added a new example section "Disable TTL jobs for the whole cluster" to show using the cluster setting as a feature flag
___
*[v22.1/row-level-ttl.md, line 88 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xJ1K7AAhmEZUpYWo2:-N021Wtk-nC_WadBNr91:b-bw8e1h) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L88)):*

oh yes, this is from all rows.
</blockquote></details>

Updated!
___
*[v22.1/row-level-ttl.md, line 94 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xLHqU3hFpg4M0dDvt:-N02HIxl4cys0L6hTR1v:b-batt8g) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L94)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

We should mention the default
</blockquote></details>

Do you mean the default of `ttl_job_cron`?  I can't seem to view the default value of `ttl_job_cron`.  Am I holding it wrong @otan ?

```
CREATE TABLE events (
  Id UUID PRIMARY KEY default gen_random_uuid(),
  description TEXT,
  inserted_at TIMESTAMP default current_timestamp()
) WITH (ttl_expire_after = '3 months');

CREATE TABLE

SELECT relname, reloptions FROM pg_class WHERE relname = 'events';

  relname |                                reloptions
----------+----------------------------------------------------------------------------
  events  | NULL
  events  | {ttl='on',ttl_automatic_column='on',ttl_expire_after='3 mons':::INTERVAL}
(2 rows)
```

Not seeing any value for `til_job_cron`

Also did a search in the code and did not find a default value (or else am I missing it?)

```
$ cd ~/go/src/github.com/cockroachdb/cockroach/
$ ack 'ttl_job_cron'
docs/RFCS/20220120_row_level_ttl.md
178:`ttl_job_cron` | Frequency the job runs, specified using the CRON syntax.

pkg/sql/catalog/tabledesc/structured.go
2547:			appendStorageParam(`ttl_job_cron`, fmt.Sprintf(`'%s'`, cron))

pkg/sql/catalog/tabledesc/ttl.go
44:		if err := ValidateTTLCronExpr("ttl_job_cron", ttl.DeletionCron); err != nil {

pkg/sql/logictest/testdata/logic_test/row_level_ttl
284:statement error invalid cron expression for "ttl_job_cron"
285:CREATE TABLE tbl () WITH (ttl_expire_after = '10 seconds', ttl_job_cron = 'bad expr')
292:) WITH (ttl_expire_after = '10 minutes'::interval, ttl_job_cron = '@daily')
303:) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:10:00':::INTERVAL, ttl_job_cron = '@daily')
315:ALTER TABLE tbl SET (ttl_job_cron = '@weekly')
326:) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:10:00':::INTERVAL, ttl_job_cron = '@weekly')
336:ALTER TABLE tbl RESET (ttl_job_cron)
367:ALTER TABLE no_ttl_table SET (ttl_job_cron = '@weekly')

pkg/sql/paramparse/paramobserver.go
375:	`ttl_job_cron`: {
```
___
*[v22.1/row-level-ttl.md, line 121 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPeIR8mG9Fxd0fGtN:-N023DfnAfg9z5qxHIa3:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L121)):*

nit: casing consistency
</blockquote></details>

Fixed
___
*[v22.1/row-level-ttl.md, line 123 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xK8Vr1Z6oxyWvzjQc:-N025FmM7j0_QUxRKYkU:bo22id6) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L123)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Can we make our examples with a TTL of weeks or months?
</blockquote></details>

changed to 3 month interval 
___
*[v22.1/row-level-ttl.md, line 165 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPiNDB-SkuunpfZPM:-N02C5-p7QOjbmkGZOmP:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L165)):*

nit: casing consistency
</blockquote></details>

Fixed
___
*[v22.1/row-level-ttl.md, line 180 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xvv__Eddckgx-DzCJ:-N02GVzO5uHpBpZZPq00:b-hkgcpt) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L180)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Need to edit https://www.cockroachlabs.com/docs/v21.2/show-schedules.html since it no longer describes just backup schedules
</blockquote></details>

Thanks for the catch!  Updated that page to make it less backup-specific, and add some references to Row-Level TTL as well
___
*[v22.1/row-level-ttl.md, line 197 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xL0Im5_UPH6Y7PPjw:-N02Ch3s4cTmUUqwIb9Q:b1n88no) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L197)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

Mention DBConsole Jobs page as well

In addition, I believe @otan is improving the `description` field for TTL jobs so we will need to update the screenshot when that happens
</blockquote></details>

Ah yes!  Updated to mention the Jobs page and link to those docs
___
*[v22.1/row-level-ttl.md, line 333 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPorY2JTL1K9trba8:-N02CohI9oJSOkUJSrRX:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L333)):*

nit: casing consistency
</blockquote></details>

Fixed
___
*[v22.1/row-level-ttl.md, line 377 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-yPrhM5_UOMLvBFLR7:-N02Cur-EbnBD5Fyw09x:b-886gsx) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L377)):*

nit: casing consistency
</blockquote></details>

Fixed

rmloveland (commented):
Thanks for the review Eric!  I think I fixed the things you pointed out below, please let me know if I missed anything

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340#-:-N02s6PK6dasJP-bv5qr:b-cwzc8i)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @otan, and @vy-ton)

vy-ton (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @otan, and @vy-ton)

___
*[v22.1/row-level-ttl.md, line 94 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xLHqU3hFpg4M0dDvt:-N03cJtcAP6y848IMfTv:b-fzabzb) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L94)):*

Do you mean the default of `ttl_job_cron`?  I can't seem to view the default value of `ttl_job_cron`.  Am I holding it wrong @otan ?

```
CREATE TABLE events (
  Id UUID PRIMARY KEY default gen_random_uuid(),
  description TEXT,
  inserted_at TIMESTAMP default current_timestamp()
) WITH (ttl_expire_after = '3 months');

CREATE TABLE

SELECT relname, reloptions FROM pg_class WHERE relname = 'events';

  relname |                                reloptions
----------+----------------------------------------------------------------------------
  events  | NULL
  events  | {ttl='on',ttl_automatic_column='on',ttl_expire_after='3 mons':::INTERVAL}
(2 rows)
```

Not seeing any value for `til_job_cron`

Also did a search in the code and did not find a default value (or else am I missing it?)

```
$ cd ~/go/src/github.com/cockroachdb/cockroach/
$ ack 'ttl_job_cron'
docs/RFCS/20220120_row_level_ttl.md
178:`ttl_job_cron` | Frequency the job runs, specified using the CRON syntax.

pkg/sql/catalog/tabledesc/structured.go
2547:			appendStorageParam(`ttl_job_cron`, fmt.Sprintf(`'%s'`, cron))

pkg/sql/catalog/tabledesc/ttl.go
44:		if err := ValidateTTLCronExpr("ttl_job_cron", ttl.DeletionCron); err != nil {

pkg/sql/logictest/testdata/logic_test/row_level_ttl
284:statement error invalid cron expression for "ttl_job_cron"
285:CREATE TABLE tbl () WITH (ttl_expire_after = '10 seconds', ttl_job_cron = 'bad expr')
292:) WITH (ttl_expire_after = '10 minutes'::interval, ttl_job_cron = '@daily')
303:) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:10:00':::INTERVAL, ttl_job_cron = '@daily')
315:ALTER TABLE tbl SET (ttl_job_cron = '@weekly')
326:) WITH (ttl = 'on', ttl_automatic_column = 'on', ttl_expire_after = '00:10:00':::INTERVAL, ttl_job_cron = '@weekly')
336:ALTER TABLE tbl RESET (ttl_job_cron)
367:ALTER TABLE no_ttl_table SET (ttl_job_cron = '@weekly')

pkg/sql/paramparse/paramobserver.go
375:	`ttl_job_cron`: {
```
</blockquote></details>

The default is `@hourly` which I can only find with`show schedule`. 

I filed https://github.com/cockroachdb/cockroach/issues/80215

rmloveland (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13340)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ericharmeling, @otan, and @vy-ton)

___
*[v22.1/row-level-ttl.md, line 94 at r5](https://reviewable.io/reviews/cockroachdb/docs/13340#-N-xLHqU3hFpg4M0dDvt:-N06K0yTEBGzVl8EbPhM:bc9op8z) ([raw file](https://github.com/cockroachdb/docs/blob/b53e62adc9d6e6afdb702e78d166cbc36b4be6ff/v22.1/row-level-ttl.md#L94)):*
<details><summary><i>Previously, vy-ton (Vy Ton) wrote…</i></summary><blockquote>

The default is `@hourly` which I can only find with`show schedule`. 

I filed https://github.com/cockroachdb/cockroach/issues/80215
</blockquote></details>

Oh cool, thanks Vy! Updated the doc to note the default

PR #13359, opened 2022-03-26
Doc/elixir

devalexandre:
Add elixir code example using Postgrex driver


devalexandre:
@dbist I didn't understand, why broke the actions

ianjevans:
@devalexandre There's issues with using security tokens in GitHub Action in forked repositories, and that's why the actions are failing for your PR. It doesn't reflect actual errors.

dangdennis:
@ianjevans Can you confirm if the example still works?

I'm consistently getting tls handshake errors `{bad_cert,hostname_check_failed}`. If you don't set `verify: :verify_peer` in ssl_opts, we only get a warning (but valid)
```
13:11:59.562 [warning] Description: 'Authenticity is not established by certificate path validation'
     Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'
```
But with :verify_peer there, tls handshake always fails. 
```
13:03:31.831 [error] Postgrex.Protocol (#PID<0.234.0>) failed to connect: ** (DBConnection.ConnectionError) ssl connect: TLS client: In state wait_cert_cr at ssl_handshake.erl:2075 generated CLIENT ALERT: Fatal - Handshake Failure
 {bad_cert,hostname_check_failed} - {:tls_alert, {:handshake_failure, 'TLS client: In state wait_cert_cr at ssl_handshake.erl:2075 generated CLIENT ALERT: Fatal - Handshake Failure\n {bad_cert,hostname_check_failed}'}}
 ```

PR #13150, opened 2022-03-02
v22.1.0-alpha.2 release notes.

ianjevans:

PR #13375, opened 2022-03-28
Update v22.1.0-alpha.1.md

stbof:
This feature is not user focused and is only needed in the context of the gist surfaced in crdb_internal.statement_statistics and is documented in https://github.com/cockroachdb/docs/pull/13150.


kathancox (approved):
lgtm

PR #13382, opened 2022-03-28
Document plan gist surfaced in crdb_internal.statement_statistics

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2857

New section: https://deploy-preview-13382--cockroachdb-docs.netlify.app/docs/v22.1/crdb-internal#detect-suboptimal-and-regressed-plans


maryliag (approved):
:lgtm: 

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13382#-:-N-3jnbY033dlKKRqwRa:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @kevin-v-ngo)

v22.1/crdb-internal.md
```diff
@@ -578,6 +579,102 @@ WHERE metadata @> '{"db":"movr"}' AND (metadata @> '{"stmtTyp":"TypeDDL"}' OR me
...
+                    table: rides@rides_pkey
+                    spans: FULL SCAN
+
+~~~
+
+After you add an [index on the `start_time` column in the `rides` table](apply-statement-performance-rules.html#rule-2-use-the-right-index), you can see the change in the plans for the same query.
+
+{% include copy-clipboard.html %}
+~~~ sql
+SELECT crdb_internal.decode_plan_gist(statistics->'statistics'->'planGists'->>0) AS plan FROM movr.crdb_internal.statement_statistics WHERE substring(metadata ->> 'query',1,35)='SELECT name, count(rides.id) AS sum';
```
kevin-v-ngo, line 639: Related to my previous question, I wonder if we can structure the output a bit better here. Going through this example, we just show two plans as the result with no other context:

image

It would be nice to group a plan as a single row in the result with additional context like the statement text and aggregation interval (when it was captured) as well.

kevin-v-ngo: @stbof, i wrote a query using string_agg function that should clean up the output a bit. Let me know what you think. 

```
SELECT 
  statistics->'statistics'->'planGists'->>0 as plan_id,
  string_agg(  crdb_internal.decode_plan_gist(statistics->'statistics'->'planGists'->>0), '
  ')
AS plan,
substring(metadata ->> 'query',1,60) AS statement_text
FROM movr.crdb_internal.statement_statistics
WHERE substring(metadata ->> 'query',1,35)='SELECT name, count(rides.id) AS sum'
group by metadata ->> 'query', statistics->'statistics'->'planGists'->>0;
```

stbof: OK. I see that that gives the side-by-side view you were looking for. I clearly need to beef up my SQL skills to be able to get output like this. I have a few questions:

- Do we need the plan ID in the output
- Do you want the statement text on the left or the right? 
- Do you want the aggregation interval as you mentioned in your first comment?
- Do we want to reverse order, so that we can see the plan _after_ the index is added?
- Do we want to show other statistics, so that the plan behavior is clear? If so, which one?

kevin-v-ngo: I am no expert either - had to looke up how to merge multiple rows into a single row and column. There could be a better way to do this. Perhaps @cucaroach may know :)

- It's not needed but i couldn't get the query to work since we can't use the gist function in the group by. And since the plan_id represents the plan itself, it was the best candidate to ensure we had one row per plan.
- Yes, first column would be good!
- Yes, let's add the aggregation interval (probably take the max for the most recent time interval)
- Yes, that would be good. Perhaps order by latency desc so that the FULL SCAN is the first row
- Yes, that's a good idea. rowsRead and runLat are good stats to show the change after the index is added.


```
SELECT 
substring(metadata ->> 'query',1,60) AS statement_text,
  statistics->'statistics'->'planGists'->>0 as plan_id,
  string_agg(  crdb_internal.decode_plan_gist(statistics->'statistics'->'planGists'->>0), '
  ')
AS plan,
max(aggregated_ts) as timestamp_interval,
max(statistics -> 'statistics' -> 'rowsRead' -> 'mean') AS num_rows_read_mean,
max(statistics -> 'statistics' -> 'runLat' -> 'mean') AS runtime_latency_mean
FROM movr.crdb_internal.statement_statistics
WHERE substring(metadata ->> 'query',1,35)='SELECT name, count(rides.id) AS sum'
group by metadata ->> 'query', statistics->'statistics'->'planGists'->>0;
```

cucaroach: No sorry, nothing comes to mind.   The side by side view is slick though!

maryliag (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13382)* status: :shipit: complete! 0 of 0 LGTMs obtained (and 1 stale) (waiting on @cucaroach, @kevin-v-ngo, and @maryliag)

___
*[v22.1/crdb-internal.md, line 604 at r3](https://reviewable.io/reviews/cockroachdb/docs/13382#-N-3gM1QwNIW0bNSx6TM-r3-604:-N-Y44uz8esxCiYbrorU:b-lefrih) ([raw file](https://github.com/cockroachdb/docs/blob/d423e8a2a4e04012428109733ea09f9c07088a57/v22.1/crdb-internal.md#L604)):*
<details><summary><i>Previously, kevin-v-ngo wrote…</i></summary><blockquote>

_"you should select only the planGists column and apply the function.."_

It's unfortunate that we can't list all the statement text and their plans side by side. For instance, a great query to help users is "Give me the slowest queries with full scans and their plans".

@maryliag, how do we do this today in the console? Do we only select a single column as well?
</blockquote></details>

On the console we get this data with 2 steps: first I get the plan gists list, and then for each one I call the decode function separately, and then for each call I combine all the rows into a single string.

If you run the decode function you will see the plan itself and `(X rows)`
If the decode function was able to return everything as a single string, meaning one row, instead of one row per line, then we could combine with other columns
cc @cucaroach maybe this could be an improvement

kevin-v-ngo (approved):
Looks good Stephanie! Thanks for making the updates.

v22.1/crdb-internal.md
```diff
@@ -578,6 +579,85 @@ WHERE metadata @> '{"db":"movr"}' AND (metadata @> '{"stmtTyp":"TypeDDL"}' OR me
...
+  rides.start_time BETWEEN '2018-12-31 00:00:00' AND '2020-01-01 00:00:00'
+GROUP BY
+  name
+ORDER BY
+  sum DESC
+LIMIT
+  10;
+~~~
+
+To decode plan gists stored in `statistics->'statistics'->'planGists'`, you use the `crdb_internal.decode_plan_gist` function. The following example shows a query before (2nd row) and after (1st row) you add an [index on the `start_time` column in the `rides` table](apply-statement-performance-rules.html#rule-2-use-the-right-index). You can see the change in the plans for the same query and the reduced number of rows read and latency when you use the index.
```
rmloveland, line 604: This paragraph is a little hard to parse.  The order of the before and after is reversed from the order a reader would normally expect.  I was able to puzzle it out but perhaps this could be simplified to use shorter sentences, something like

"To decode plan gists ..., use the .. function as shown in the following query.

The query output shows the performance impact of adding an [index on the  ...]().  The second row of the output shows the performance before the index was added.  The first row of the output shows the improved performance after the index was added."

rmloveland (approved):
LGTM, had a comment about the para explaining the query but it could go as is

rmloveland (approved):
LGTM, thanks!

PR #13399, opened 2022-03-29
Rate-limiting prevents editing AWS disk

gemma-shay:
Closes DOC-2845


gemma-shay:
@jramos-crl is the only limit here on AWS disks?

jramos-crl (approved):
@gemma-shay yes, the only rate-limiting is on AWS disks. Change looks good to me!

PR #13403, opened 2022-03-29
Removed http mentions from backup/restore

kathancox:
Fixes [DOC-3059](https://cockroachlabs.atlassian.net/browse/DOC-3059)

Removed leftover mentions of http storage for backup and restore that were in includes files. Moved restore's source privilege include to its own file since it differs from show backup etc.


kathancox:
cool thanks @msbutler, those changes are just italics --> bold as per some style guidelines!

andf-crl (commented):
Looks good, but I think SHOW BACKUP might also be within scope for this ticket:

https://deploy-preview-13403--cockroachdb-docs.netlify.app/docs/stable/show-backup.html#required-privileges

Perhaps also IMPORT and IMPORT INTO?

https://deploy-preview-13403--cockroachdb-docs.netlify.app/docs/stable/import.html#source-privileges
https://deploy-preview-13403--cockroachdb-docs.netlify.app/docs/stable/import-into.html#source-privileges

Otherwise LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13403#-:-MzMZ5oaBwYK6m2Mjukn:b8u4cjl)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @taroface)

msbutler:
+1 to removing references in SHOW BACKUP

I don't tthink we should remove references to http for import IMPORT / IMPORT INTO. The user can still import multiple files in a single IMPORT statement using HTTP if each file has its own URL. This use case may merit its own example in the docs. @livlobo @dt do you agree?

taroface (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13403)* status: :shipit: complete! 0 of 0 LGTMs obtained

kathancox:
> +1 to removing references in SHOW BACKUP
> 
> I don't tthink we should remove references to http for import IMPORT / IMPORT INTO. The user can still import multiple files in a single IMPORT statement using HTTP if each file has its own URL. This use case may merit its own example in the docs. @livlobo @dt do you agree?

I checked in with Liv, and IMPORT + http can remain.

PR #13493, opened 2022-04-04
add connection_wait to node shutdown docs

taroface:
Fixes DOC-2874.

- Documented new `connection_wait` setting for v22.1.
- Minor tweaks to Node Shutdown docs (copy changes that I forgot to port to v22.1).
- Also clarified that draining nodes will have `SUSPECT` status in DB Console.


taroface:
@ZhouXing19, I believe I incorporated your guidance from DOC-2874. However, I left it implicit that the `connection_wait` setting should be used in conjunction with a connection pool. 

If it's important to explicitly recommend/require the use of a connection pool with `connection_wait`, then can you help me understand the reasoning for this? (Is it because max connection lifetime of all clients can only be configured when using a pool? Just guessing...)

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13493)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ZhouXing19)

___
*[v22.1/node-shutdown.md, line 74 at r1](https://reviewable.io/reviews/cockroachdb/docs/13493#-MzuNBW23ZOmk2G5HZT-:-MzvFFUd60DAXggLXNvv:bg23qku) ([raw file](https://github.com/cockroachdb/docs/blob/cea9e8b46e5b58c66dce459d993c006b9b2db2c0/v22.1/node-shutdown.md#L74)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

I wonder if we can rename `Connection phase` to something similar to `Waiting-connections-to-close phase`, and `Local query phase` to `Draining-SQL-connections phase`, so that the name shows the server's behavior at this stage.

Note that during `connection_wait`, **only SQL connections are restricted**. Some other type of connections, such as HTTP connections, are still allowed throughout the draining process. I'm not sure if we have specified that the `client connections` here refers to **SQL connections**.
</blockquote></details>

Updated to reflect the point about SQL connections - thank you for clarifying this for me!

Also renamed three phases - in the interest of simplicity, would "SQL wait" and "SQL drain" (and "DistSQL drain") be sufficient? I also tried to make the "SQL drain" (`server.shutdown.query_wait`) phase description more accurate, but let me know if I made any mistakes here.
___
*[v22.1/node-shutdown.md, line 162 at r1](https://reviewable.io/reviews/cockroachdb/docs/13493#-MzuPMa00CMCHaKkJxje:-Mzv1Jm6-lesWLbLo5nk:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/cea9e8b46e5b58c66dce459d993c006b9b2db2c0/v22.1/node-shutdown.md#L162)):*
<details><summary><i>Previously, ZhouXing19 (Jane Xing) wrote…</i></summary><blockquote>

> If it's important to explicitly recommend/require the use of a connection pool with connection_wait, then can you help me understand the reasoning for this? (Is it because max connection lifetime of all clients can only be configured when using a pool? Just guessing...)

I think it's fine that we don't explicitly mention that users should configure connection_wait only when using a connection pool. But maybe we can mention something like
```
When needed, we suggest setting `server.shutdown.connection_wait` in accordance with the maximum life time of a SQL connection, which is usually configurable via a connection pool.
```
</blockquote></details>

Done.

ZhouXing19 (changes requested):
:lgtm: modulo a nit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13493#-:-MzvO0EZ8Kkalb4zIKJn:b-9byf81)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @taroface and @ZhouXing19)

___
*[v22.1/node-shutdown.md, line 74 at r1](https://reviewable.io/reviews/cockroachdb/docs/13493#-MzuNBW23ZOmk2G5HZT-:-MzvMm-FEBFFNufJYdVQ:b-uxu6ie) ([raw file](https://github.com/cockroachdb/docs/blob/cea9e8b46e5b58c66dce459d993c006b9b2db2c0/v22.1/node-shutdown.md#L74)):*

Updated to reflect the point about SQL connections - thank you for clarifying this for me!

Also renamed three phases - in the interest of simplicity, would "SQL wait" and "SQL drain" (and "DistSQL drain") be sufficient? I also tried to make the "SQL drain" (`server.shutdown.query_wait`) phase description more accurate, but let me know if I made any mistakes here.
</blockquote></details>

Yes, I think `SQL wait` and `SQL drain` are good. Thanks for making the changes!
___
*[v22.1/node-shutdown.md, line 162 at r1](https://reviewable.io/reviews/cockroachdb/docs/13493#-MzuPMa00CMCHaKkJxje:-MzvNBp84vWdcgjepWiS:b-c4wmxb) ([raw file](https://github.com/cockroachdb/docs/blob/cea9e8b46e5b58c66dce459d993c006b9b2db2c0/v22.1/node-shutdown.md#L162)):*

Done.
</blockquote></details>

Looks good! Just a nit: "in accordance with the maximum lifetime ... This will cause more frequent reconnections. Set..." are italic now. Is this intended?

taroface:
> Looks good! Just a nit: "in accordance with the maximum lifetime ... This will cause more frequent reconnections. Set..." are italic now. Is this intended?

TFTR! Hmm I don't see italics in preview or in the Markdown, so I think it's ok. But let me know if I misunderstood.

ZhouXing19 (approved):

kathancox (approved):
LGTM Ryan!

taroface:
@MichaelTrestman Ping on review, please!

PR #13494, opened 2022-04-04
Fix css callout

gemma-shay:
https://cockroachlabs.atlassian.net/browse/CC-6572


nickvigilante (approved):

gemma-shay:
thanks @nickvigilante!

Issue #13562, opened 2022-04-09
Add  download link in release pages for the new `cockroach-sql` binary + simplify some Cloud tutorials


knz:
Raphael 'kena' Poss (knz) commented:

tldr: the docs should  be modified to:
- add a new download link
- simplify some Cloud tutorials/docs to recommend use the small binary instead of the large one

Context:

In https://github.com/cockroachdb/cockroach/issues/75800, we've added a new `cockroach-sql` to the published artifacts along `cockroach` and the geospatial library.

This is a smaller executable (<20MB compared to `cockroach`'s 150MB+ size) which users can download when they just want to access an existing cluster remotely.

This is part of a wider initiative originally requested by @bdarnell [here](https://github.com/cockroachdb/cockroach/issues/56058), which found recently a new purpose via @kernfeld-cockroach  with our focus to Cloud: since CC end-users are not expected to run/administrate their own cluster any more, we don't need to orient them to download the full `cockroach` binary any more.

In the longer term, we wish to provide users with a CC-specific `ccloud` CLI which *embeds* the SQL shell, obviating the need for a separate download entiely. However, we're not there yet. The new `cockroach-sql` binary provides an intermediate improvement upon the current situation.

Epic: CC-5993

Jira Issue: DOC-3283

knz:
@ianjevans do you have recommendations about who in the docs team is best to own this?

bdarnell:
We probably need installation instructions similar to `install-cockroachdb-*.html`. The process is simpler (no geo libraries!) but we should still give people explicit instructions. (we should also make sure that all the various installation methods work, including `brew`)

On windows, this could be a big win since `cockroach-sql` wouldn't need to recommend docker (could we remove the "experimental" banner completely or could there still be readline-related experimentalness?)

knz:
> could there still be readline-related experimentalness

IIRC libedit is simply not linked in on windows, and we don't get line editing at all. The UX is definitely sub-par.

knz:
personally i'd prefer a future for windows where we depend on WSL and simply ship the linux binary and provide instructions on how to run in in a WSL shell. That would be so much easier all around.

mari-crl:
:information_source: Hello! I am a human and not at all a robot! Look at my very human username! :robot: :notes:
:thinking: Although I tried very hard to figure out what to do with this issue, more powerful human brains will need to help me. (specifically: Both Github and Jira issues were modified after exalate problems) :confounded:
[...]

PR #13574, opened 2022-04-11
Add download link for cockroach-sql

knz:
Informs #13562.
Epic: CC-5993

cc @jseldess


nickvigilante:
@knz This looks good, but can you also apply the change you made in https://github.com/cockroachdb/cockroach/pull/79788 to [this file](https://github.com/cockroachdb/docs/blob/master/_includes/releases/release-downloads-docker-image.md)?

knz:
> can you also apply the change you made in https://github.com/cockroachdb/cockroach/pull/79788 to [this file](https://github.com/cockroachdb/docs/blob/master/_includes/releases/release-downloads-docker-image.md?rgh-link-date=2022-04-11T18%3A43%3A01Z)?

Done.

Can you also point me to who/how the file `releases.csv` is generated? We need to change the generation to add a new column.

nickvigilante (approved):

nickvigilante:
@knz As a stop gap until I figured out a good way to maintain the existing functionality of the Python script while satisfying the docs team's purposes, I put a copy of it here:

https://github.com/cockroachdb/ed-tools/blob/main/docs/release-notes/release-notes.py

knz:
ok, are you expecting me to update that copy to add the new column?

nickvigilante:
@knz I can update it if you want me to; I just need to know the logic used to calculate it (under what situations is it true or false?).

knz:
It's true for v22.1-beta.1 and all later versions after that
it's false for all versions before v22.1 (so also false for later patch releases of v21.2 etc)

PR #13579, opened 2022-04-11
update K8s Operator cert rotation to avoid downtime

taroface:
Fixes DOC-2467.

- Updated the K8s Operator cert rotation steps to avoid downtime.
- Updated the cert generation snippet to use the Operator's default namespace.


taroface:
@pseudomuto @chrisseto Part of the old instructions included annotating the cluster to trigger a full restart (resulting in downtime). Can you confirm this is no longer necessary with the reordered steps?

davidwding (approved):

PR #13585, opened 2022-04-12
Bump operator version to v2.5.3

pseudomuto:
We recently released a new version of the public operator containing a
bugfix that has been coming up in support tickets and Slack (issues with
`cockroachDBVersion` is set and `image` is not).


pseudomuto:
TFTR!

PR #13590, opened 2022-04-12
Document non-experimental DateStyle and IntervalStyle session parameters

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-2593

`datestyle_enabled` and `intervalstyle_enabled` now have no effect because they are always true.
(I just deleted references to those variables, but let me know if there should be a more direct note to say that they are no longer necessary)
There is also no longer any issue with those variables and casting between INTERVALs and STRINGs.


e-mbrown (approved):
LGTM

ianjevans (approved):

PR #13593, opened 2022-04-12
document topology spread constraints

taroface:
Fixes DOC-2993.

- Added some docs on the topology spread constraints feature. 

I'm letting the K8s docs do the heavy lifting here. It was difficult to think of an example where these rules would be effectively used, since our default behavior and existing guidance already emphasizes running deployments with an even distribution of pods across failure zones. But if anyone can point out an example that would make sense, I'm happy to add it.

### DIFF
--- a/v21.2/schedule-cockroachdb-kubernetes.md
+++ b/v21.2/schedule-cockroachdb-kubernetes.md
@@ -13,6 +13,7 @@ This page describes how to configure the following, using the [Operator](https:/
 - [Node affinities](#add-a-node-affinity)
 - [Pod affinities and anti-affinities](#add-a-pod-affinity-or-anti-affinity)
 - [Taints and tolerations](#taints-and-tolerations)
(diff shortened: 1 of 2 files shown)


davidwding (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13593#-:-N02i1WHB2EY-3q7OZNI:bg7igkv)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @chrisseto and @pseudomuto)

nickvigilante (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13593#-:-N0af2Ap35Kw8P_fO9Ij:b-uwkowd)* status: :shipit: complete! 2 of 0 LGTMs obtained (waiting on @chrisseto and @pseudomuto)

PR #13609, opened 2022-04-14
Add docs on super regions

rmloveland:
Fixes DOC-2581
Fixes DOC-2899
Fixes DOC-3099
Fixes DOC-3110

Summary of changes:

- Add section on super regions to 'Multiregion Overview' page

- Add SQL reference pages with examples:

  - 'ADD SUPER REGION'

  - 'DROP SUPER REGION'

  - 'ALTER SUPER REGION'

  - 'SHOW SUPER REGIONS'

- Add links to the above statements as needed from the 'ALTER DATABASE'
  and 'SQL Statements' pages

### DIFF
--- a/v22.1/add-region.md
+++ b/v22.1/add-region.md
@@ -132,5 +132,8 @@ SHOW REGIONS FROM DATABASE foo;
 - [`SET PRIMARY REGION`](set-primary-region.html)
 - [`DROP REGION`](drop-region.html)
 - [`SHOW REGIONS`](show-regions.html)
+- [`ADD SUPER REGION`](add-super-region.html)
+- [`DROP SUPER REGION`](drop-super-region.html)
+- [`SHOW SUPER REGIONS`](show-super-regions.html)
 - [`ALTER TABLE`](alter-table.html)
 - [SQL Statements](sql-statements.html)
--- a/v22.1/add-super-region.md
+++ b/v22.1/add-super-region.md
@@ -1,0 +1,94 @@
+---
+title: ADD SUPER REGION
+summary: The ADD SUPER REGION statement creates a set of regions where data from regional tables with home rows in the super region is stored in the super region.
+toc: true
+docs_area: reference.sql
+---
+
+ The `ALTER DATABASE .. ADD SUPER REGION` [statement](sql-statements.html) adds a [super region](multiregion-overview.html#super-regions) to a [multi-region database](multiregion-overview.html).
+
+{% include enterprise-feature.md %}
+
+{{site.data.alerts.callout_info}}
+`ADD SUPER REGION` is a subcommand of [`ALTER DATABASE`](alter-database.html).
+{{site.data.alerts.end}}
+
+{% include common/experimental-warning.md %}
+
+## Synopsis
+
+<div>
+{% remote_include https://raw.githubusercontent.com/cockroachdb/generated-diagrams/release-22.1/grammar_svg/alter_database_add_super_region.html %}
+</div>
+
+## Parameters
+
+| Parameter       | Description                                                                                              |
+|-----------------+----------------------------------------------------------------------------------------------------------|
+| `database_name` | The database to which you are adding a [super region](multiregion-overview.html#super-regions).          |
+| `name`          | The name of the [super region](multiregion-overview.html#super-regions) being added to this database.    |
+| `name_list`     | The super region consists of this set of [database regions](multiregion-overview.html#database-regions). |
+
+## Required privileges
+
+To add a super region to a database, the user must have one of the following:
+
+- Membership to the [`admin`](security-reference/authorization.html#admin-role) role for the cluster.
+- Either [ownership](security-reference/authorization.html#object-ownership) or the [`CREATE` privilege](security-reference/authorization.html#supported-privileges) for the database.
+
+## Considerations
+
+{% include {{page.version.version}}/sql/super-region-considerations.md %}
+
+## Examples
+
+The examples in this section use the following setup.
+
+{% include {{page.version.version}}/sql/multiregion-example-setup.md %}
+
+#### Set up movr database regions
+
+{% include {{page.version.version}}/sql/multiregion-movr-add-regions.md %}
+
+#### Set up movr global tables
+
+{% include {{page.version.version}}/sql/multiregion-movr-global.md %}
+
+#### Set up movr regional tables
+
+{% include {{page.version.version}}/sql/multiregion-movr-regional-by-row.md %}
+
+### Enable super regions
+
+{% include {{page.version.version}}/sql/enable-super-regions.md %}
+
+### Add a super region to a database
+
+To add a super region to a multi-region database, use the `ALTER DATABASE ... ADD SUPER REGION` statement:
+
+{% include copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE movr ADD SUPER REGION "usa" VALUES "us-east1", "us-west1";
+~~~
+
+~~~
+ALTER DATABASE ADD SUPER REGION
+~~~
+
+### Allow user to modify a primary region that is part of a super region
+
+{% include {{page.version.version}}/sql/enable-super-region-primary-region-changes.md %}
+
+## See also
+
+- [Multi-Region Capabilities Overview](multiregion-overview.html)
+- [Super regions](multiregion-overview.html#super-regions)
+- [`SET PRIMARY REGION`](set-primary-region.html)
+- [`DROP SUPER REGION`](drop-super-region.html)
+- [`ALTER SUPER REGION`](alter-super-region.html)
+- [`SHOW SUPER REGIONS`](show-super-regions.html)
+- [`DROP REGION`](drop-region.html)
+- [`SHOW REGIONS`](show-regions.html)
+- [`ALTER TABLE`](alter-table.html)
+- [`ALTER DATABASE`](alter-database.html)
+- [SQL Statements](sql-statements.html)
--- a/v22.1/drop-super-region.md
+++ b/v22.1/drop-super-region.md
@@ -1,0 +1,104 @@
+---
+title: DROP SUPER REGION
+summary: The DROP SUPER REGION statement drops a super region from a multi-region database.
+toc: true
+docs_area: reference.sql
+---
+
+ The `ALTER DATABASE .. DROP SUPER REGION` [statement](sql-statements.html) drops a [super region](multiregion-overview.html#super-regions) from a [multi-region database](multiregion-overview.html).
+ 
+{% include enterprise-feature.md %}
+
+{{site.data.alerts.callout_info}}
+`DROP SUPER REGION` is a subcommand of [`ALTER DATABASE`](alter-database.html).
+{{site.data.alerts.end}}
+
+{% include common/experimental-warning.md %}
+
+## Synopsis
+
+<div>
+{% remote_include https://raw.githubusercontent.com/cockroachdb/generated-diagrams/release-22.1/grammar_svg/alter_database_drop_super_region.html %}
+</div>
+
+## Parameters
+
+| Parameter       | Description                                                                                               |
+|-----------------+-----------------------------------------------------------------------------------------------------------|
+| `database_name` | The database from which you are dropping a [super region](multiregion-overview.html#super-regions).       |
+| `name`          | The name of the [super region](multiregion-overview.html#super-regions) being dropped from this database. |
+
+## Required privileges
+
+To drop a super region from a database, the user must have one of the following:
+
+- Membership to the [`admin`](security-reference/authorization.html#roles) role for the cluster.
+- Either [ownership](security-reference/authorization.html#object-ownership) or the [`CREATE` privilege](security-reference/authorization.html#supported-privileges) for the database.
+
+## Considerations
+
+{% include {{page.version.version}}/sql/super-region-considerations.md %}
+
+## Examples
+
+The examples in this section use the following setup.
+
+{% include {{page.version.version}}/sql/multiregion-example-setup.md %}
+
+#### Set up movr database regions
+
+{% include {{page.version.version}}/sql/multiregion-movr-add-regions.md %}
+
+#### Set up movr global tables
+
+{% include {{page.version.version}}/sql/multiregion-movr-global.md %}
+
+#### Set up movr regional tables
+
+{% include {{page.version.version}}/sql/multiregion-movr-regional-by-row.md %}
+
+### Enable super regions
+
+{% include {{page.version.version}}/sql/enable-super-regions.md %}
+
+### Drop a super region from a database
+
+To drop a super region from a multi-region database, use a [`DROP SUPER REGION`](drop-super-region.html) statement:
+
+{% include copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE movr DROP SUPER REGION "usa";
+~~~
+
+~~~
+ALTER DATABASE DROP SUPER REGION
+~~~
+
+Note that you cannot [drop a region](drop-region.html) that is part of a super region until you either [alter the super region](alter-super-region.html) to remove it, or [drop the super region](drop-super-region.html) altogether.
+
+For example, using the super region that was added in [`ADD SUPER REGION`](add-super-region.html#add-a-super-region-to-a-database):
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE movr DROP REGION "us-west1";
+~~~
+
+~~~
+ERROR: region us-west1 is part of super region usa
+SQLSTATE: 2BP01
+HINT: you must first drop super region usa before you can drop the region us-west1
+~~~
+
+## See also
+
+- [Multi-Region Capabilities Overview](multiregion-overview.html)
+- [Super regions](multiregion-overview.html#super-regions)
+- [`SET PRIMARY REGION`](set-primary-region.html)
+- [`ADD SUPER REGION`](add-super-region.html)
+- [`ALTER SUPER REGION`](alter-super-region.html)
+- [`SHOW SUPER REGIONS`](show-super-regions.html)
+- [`ADD REGION`](add-region.html)
+- [`SHOW REGIONS`](show-regions.html)
+- [`ALTER TABLE`](alter-table.html)
+- [`ALTER DATABASE`](alter-database.html)
+- [SQL Statements](sql-statements.html)
--- a/v22.1/multiregion-overview.md
+++ b/v22.1/multiregion-overview.md
@@ -66,6 +66,36 @@ To show all of a database's regions, execute the [`SHOW REGIONS FROM DATABASE` s
 If the default survival goals and table localities meet your needs, there is nothing else you need to do once you have set a database's primary region.
 {{site.data.alerts.end}}
 
+## Super regions
+
+{% include common/experimental-warning.md %}
+
+<span class="version-tag">New in v22.1:</span> Super regions allow you to define a set of [database regions](#database-regions) such that the following [schema objects](schema-design-overview.html#database-schema-objects) will have all of their replicas stored _only_ in regions that are members of the super region:
+
+- [Regional tables](#regional-tables) whose home region is a member of the super region.
+- Any row of a [regional by row table](#regional-by-row-tables) whose [home region](set-locality.html#crdb_region) is a member of the super region.
+
+The primary use case for super regions is data domiciling. As mentioned above, data from [regional](#regional-tables) and [regional by row](#regional-by-row-tables) tables will be stored only in regions that are members of the super region. Further, if the super region contains 3 or more regions and if you use [`REGION` survival goals](#survive-region-failures), the data domiciled in the super region will remain available if you lose a region.
+
+{% include {{page.version.version}}/sql/super-region-considerations.md %}
+
+<a name="enable-super-regions"></a>
+
+For more information about how to enable and use super regions, see:
+
+- [`ADD SUPER REGION`](add-super-region.html)
+- [`DROP SUPER REGION`](drop-super-region.html)
+- [`ALTER SUPER REGION`](alter-super-region.html)
+- [`SHOW SUPER REGIONS`](show-super-regions.html)
+
+Note that super regions take a different approach to data domiciling than [`ALTER DATABASE ... PLACEMENT RESTRICTED`](placement-restricted.html). Specifically, super regions make it so that all [replicas](architecture/overview.html#architecture-replica) (both voting and [non-voting](architecture/replication-layer.html#non-voting-replicas)) are placed within the super region, whereas `PLACEMENT RESTRICTED` makes it so that there are no non-voting replicas.
+
+For more information about data domiciling using `PLACEMENT RESTRICTED`, see [Data Domiciling with CockroachDB](data-domiciling.html).
+
+{{site.data.alerts.callout_info}}
+Super regions rely on the underlying [replication zone system](configure-replication-zones.html), which was historically built for performance, not for domiciling. The replication system's top priority is to prevent the loss of data and it may override the zone configurations if necessary to ensure data durability. For more information, see [Configure Replication Zones](https://www.cockroachlabs.com/docs/v21.2/configure-replication-zones#types-of-constraints).
+{{site.data.alerts.end}}
+
 ## Survival goals
 
 A _survival goal_ dictates how many simultaneous failure(s) a database can survive. All tables within the same database operate with the **same survival goal**. Each database can have its own survival goal setting.
--- a/_includes/v22.1/sidebar-data/reference.json
+++ b/_includes/v22.1/sidebar-data/reference.json
@@ -163,6 +163,12 @@
                 ]
               },
               {
+                "title": "<code>ADD SUPER REGION</code> (Enterprise)",
+                "urls": [
+                  "/${VERSION}/add-super-region.html"
+                ]
+              },
+              {
                 "title": "<code>ALTER CHANGEFEED</code> (Enterprise)",
                 "urls": [
                   "/${VERSION}/alter-changefeed.html"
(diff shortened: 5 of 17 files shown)


RichardJCai:
Sweet, thanks for putting this out.

> 1. Can you please confirm the permissions needed to use DROP SUPER REGION? (see note marked 'XXX' on drop-super-region.md)
> 2. Can you please confirm the permissions needed to use ADD SUPER REGION? (see note marked 'XXX' on add-super-region.md)

Good question, this one follows the general requirement for multiregion database ops which is `CREATE` privilege or being the owner of the database.

> 3. Similar to my question on [cockroachdb/cockroach#79912](https://github.com/cockroachdb/cockroach/issues/79912#issuecomment-1098384620), do you mind explaining what the meaning of the `sql.defaults.override_alter_primary_region_in_super_region.enabled` setting is? I found [this comment](https://github.com/cockroachdb/cockroach/blob/master/pkg/sql/sessiondatapb/local_only_session_data.proto#L236) which seems to explain it, but I was wondering: what do we mean by "modify" a primary region in that comment? as in, change the primary region A of a database to be some other region B while region A is part of a super region? (see other note marked 'XXX' on add-super-region.md)

Your understanding is correct, we have this to prevent users shooting themselves in the foot. Altering the primary region could cause say a regional table to move out of a super region it was part of, we want them to explicitly acknowledge that this can be possible through setting the cluster setting.

> 4. Also, is there a corresponding session setting to the `sql.defaults.override_alter_primary_region_in_super_region.enabled` cluster setting above?

Yes it is `override_multi_region_zone_config`

> 5. Can you please confirm the permissions needed to use SHOW SUPER REGIONS? (see note marked 'XXX' on show-super-regions.md)

The rows will only be populated if they have any privilege on the corresponding database, ie `CREATE`

> 6. What is the correct SQL syntax diagram to link to for SHOW SUPER REGIONS? I've tried a few variations of show_super_region*.html and have been getting 404s (see other note marked 'XXX' on show-super-regions.md)

It's part of the `show_regions.html` I believe which might not make that much sense

rmloveland:
Thanks Richard!  I updated things and I this is RFAL

> Sweet, thanks for putting this out.

Thanks for writing the feature!

> Good question, this one follows the general requirement for multiregion database ops which is `CREATE` privilege or being the owner of the database.

OK thanks.  I think the stuff that's there is basically right, then

> Your understanding is correct, we have this to prevent users shooting themselves in the foot. Altering the primary region could cause say a regional table to move out of a super region it was part of, we want them to explicitly acknowledge that this can be possible through setting the cluster setting.

OK - added some words around "this is off by default for your safety - do this only if you know what you are doing"

> Yes it is `override_multi_region_zone_config`

Added

> The rows will only be populated if they have any privilege on the corresponding database, ie `CREATE`

Updated the SHOW SUPER REGIONS to list this as a perm

> It's part of the `show_regions.html` I believe which might not make that much sense

Oh I see.  Added it (I think it's probably fine to be one diagram, it's all to do with regions n stuff)

rmloveland:
Hey @RichardJCai thanks for answering my questions earlier - this should now be ready for your official review

In particular please let me know if:

- something in here is incorrect
- something is missing that users need to understand to use super regions
- you see anything else in general that could be improved

Thanks!

v22.1/add-super-region.md
```diff
@@ -0,0 +1,115 @@
+---
+title: ADD SUPER REGION
+summary: The ADD SUPER REGION statement adds a super region made up of a set of regions such that data from regional tables will be stored in those regions.
```
RichardJCai, line 3: I'd explicitly call out regional by row tables as well. 
Also I'd say such that if the regional table's region or regional by row partition region is part of the super region, then all data from that table / partition will be stored in that super region.

v22.1/add-region.md
```diff
@@ -132,5 +132,8 @@ SHOW REGIONS FROM DATABASE foo;
 - [`SET PRIMARY REGION`](set-primary-region.html)
 - [`DROP REGION`](drop-region.html)
 - [`SHOW REGIONS`](show-regions.html)
+- [`ADD SUPER REGION`](add-super-region.html)
+- [`DROP SUPER REGION`](drop-super-region.html)
+- [`SHOW SUPER REGIONS`](show-super-regions.html)
```
RichardJCai, line 137: There was also ALTER super region support added - just wanted to make sure that this is also on your radar but it can be added afterwards.
https://github.com/cockroachdb/cockroach/pull/78367

v22.1/add-super-region.md
```diff
@@ -0,0 +1,115 @@
...
+{% remote_include https://raw.githubusercontent.com/cockroachdb/generated-diagrams/release-22.1/grammar_svg/alter_database_add_super_region.html %}
+</div>
+
+## Parameters
+
+| Parameter       | Description                                                                                                       |
+|-----------------+-------------------------------------------------------------------------------------------------------------------|
+| `database_name` | The database to which you are adding a [super region](multiregion-overview.html#super-regions).                   |
+| `name`          | The name of the [super region](multiregion-overview.html#super-regions) being added to this database.             |
+| `name_list`     | The names of the [database regions](multiregion-overview.html#database-regions) being added to this super region. |
```
RichardJCai, line 30: s/being added to this super region 

I'd say something like the super region consists of this set of regions

v22.1/add-super-region.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Required privileges
+
+To add a super region to a database, the user must have one of the following:
+
+- Membership to the [`admin`](security-reference/authorization.html#admin-role) role for the cluster.
+- Either [ownership](security-reference/authorization.html#object-ownership) or the [`CREATE` privilege](security-reference/authorization.html#supported-privileges) for the database and all [`REGIONAL BY ROW`](multiregion-overview.html#regional-by-row-tables) tables in the database.
```
RichardJCai, line 37: I believe it's we only check for CREATE on the database and nothing about the table here.

v22.1/drop-super-region.md
```diff
@@ -0,0 +1,101 @@
...
+|-----------------+-----------------------------------------------------------------------------------------------------------|
+| `database_name` | The database from which you are dropping a [super region](multiregion-overview.html#super-regions).       |
+| `name`          | The name of the [super region](multiregion-overview.html#super-regions) being dropped from this database. |
+
+## Required privileges
+
+To drop a super region from a database, the user must have one of the following:
+
+- Membership to the [`admin`](security-reference/authorization.html#roles) role for the cluster.
+- Membership to the [owner](security-reference/authorization.html#object-ownership) role, or the [`CREATE` privilege](security-reference/authorization.html#supported-privileges), for the database and all [`REGIONAL BY ROW`](multiregion-overview.html#regional-by-row-tables) tables in the database.
```
RichardJCai, line 36: Ditto with the privilege comment in add-super-region

v22.1/drop-super-region.md
```diff
@@ -0,0 +1,101 @@
...
+{% include copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE movr DROP SUPER REGION "usa";
+~~~
+
+~~~
+ALTER DATABASE DROP SUPER REGION
+~~~
+
+Note that you cannot [drop a region](drop-region.html) that is part of a super region until you first [drop the super region](drop-super-region.html). For example, using the "usa" super region that was added in [`ADD SUPER REGION`](add-super-region.html#add-a-super-region-to-a-database):
```
RichardJCai, line 77: Once alter super region docs are added, we can also callout that you can use alter to remove the region from the super region

v22.1/multiregion-overview.md
```diff
@@ -66,6 +66,31 @@ To show all of a database's regions, execute the [`SHOW REGIONS FROM DATABASE` s
...
 {{site.data.alerts.end}}
 
+## Super regions
+
+{% include common/experimental-warning.md %}
+
+<span class="version-tag">New in v22.1:</span> Super regions allow you to define a set of [database regions](#database-regions) such that the following [schema objects](schema-design-overview.html#database-schema-objects) will have all of their replicas stored _only_ in regions that are members of the super region:
+
+- [Regional tables](#regional-tables) whose home region is a member of the super region.
+- Any row of a [regional by row table](#regional-by-row-tables) whose [home region](set-locality.html#crdb_region) is a member of the super region.
```
RichardJCai, line 76: any partition

RichardJCai (commented):
Nice, this looks great overall

v22.1/add-super-region.md
rmloveland: Added mention of RBR

I modified your suggested text slightly to say regional by row's 'home region' since even tho they use partitions the normie RBR user is supposed to not have to care about partitions too much (which is what we tell them at least on https://www.cockroachlabs.com/docs/dev/partitioning.html)

v22.1/add-super-region.md
rmloveland: Makes sense - updated to "The super region consists of this set of [database regions](multiregion-overview.html#database-regions)"

v22.1/add-super-region.md
rmloveland: OK, removed the mention of RBR tables - PTAL

rmloveland:
Thanks for the review @awoods187 !  Working on these updates now (esp. adding the `ALTER SUPER REGION` stuff)

rmloveland (commented):
Andy I made a bunch of updates based on your feedback - PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13609#-:-N1ABx8o6DCCbWbI97Xy:b-37r601)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @RichardJCai)

___
*[`_includes/v22.1/sql/super-region-considerations.md` line 7 at r2](https://reviewable.io/reviews/cockroachdb/docs/13609#-N0qhMKZBQK-K5memvJx:-N1Eaxv60hFE5IzqWnTt:bpep3u3) ([raw file](https://github.com/cockroachdb/docs/blob/00120684f97e4ce39823eba87854ce9f87321ce4/_includes/v22.1/sql/super-region-considerations.md#L7)):*

or alter the super region no?
</blockquote></details>

Updated with a mention of altering the super region that links to ALTER SUPER REGION docs
___
*[`v22.1/add-region.md` line 137 at r1](https://reviewable.io/reviews/cockroachdb/docs/13609#-N-dcfPDxaMeZQKW3qnP-r1-137:-N1Eb3B-DohkmFbD5S_s:b-n13ivh) ([raw file](https://github.com/cockroachdb/docs/blob/fd30b6d5957828fd3e7607119553cbbeefa9525c/v22.1/add-region.md#L137)):*

I'd add it into these docs because otherwise there are a large number of usability gaps
</blockquote></details>

Makes sense - Just added to latest round of changes - PTAL
___
*[`v22.1/drop-super-region.md` line 77 at r1](https://reviewable.io/reviews/cockroachdb/docs/13609#-N-cUJlcWUWvc4coUZKn-r1-77:-N1EbPFW6TEPSpQSPHmD:b-wykpw0) ([raw file](https://github.com/cockroachdb/docs/blob/fd30b6d5957828fd3e7607119553cbbeefa9525c/v22.1/drop-super-region.md#L77)):*

Thanks - added a comment to https://cockroachlabs.atlassian.net/browse/DOC-3008 with a link to this comment so I don't forget
</blockquote></details>

Updated to link to the ALTER SUPER REGION docs added to this PR
___
*[`v22.1/multiregion-overview.md` line 76 at r1](https://reviewable.io/reviews/cockroachdb/docs/13609#-N-cUJlcWUWvc4coUZKh-r1-76:-N1EbShm-zoZPfznFWPV:brv5xyg) ([raw file](https://github.com/cockroachdb/docs/blob/fd30b6d5957828fd3e7607119553cbbeefa9525c/v22.1/multiregion-overview.md#L76)):*

I'm in favor of not using partition. I think any row is fine and the unit (SQL) that we want people to think in
</blockquote></details>

OK thanks
___
*[`v22.1/multiregion-overview.md` line 78 at r2](https://reviewable.io/reviews/cockroachdb/docs/13609#-N0qi6UV1equx7fZJ3_x:-N1EcrKl88-0oyynxP7e:b-ne0vw7) ([raw file](https://github.com/cockroachdb/docs/blob/00120684f97e4ce39823eba87854ce9f87321ce4/v22.1/multiregion-overview.md#L78)):*

I think we need to explain that data will stay in the super region and that you can lose a region if region n=3+ 
</blockquote></details>

Let me know if I understand what you meant.  Here's what I added:

"The primary use case for super regions is data domiciling. As mentioned above, data from [regional](#regional-tables) and [regional by row](#regional-by-row-tables) tables will be stored only in regions that are members of the super region. Further, if the super region contains 3 or more regions and if you use [`REGION` survival goals](#survive-region-failures), the data domiciled in the super region will remain available if you lose a region."

.. but I'm less sure about the latter part about needing REGION survival goal. But it seems like it would still be necessary, right?  otherwise I'm still only at 3 replicas in 3 regions and losing 1 region puts me down to 2 replicas and I can't make progress on voting with an even number of replicas.

but maybe I misunderstood - please let me know

awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13609)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @RichardJCai, and @rmloveland)

___
*[`v22.1/alter-super-region.md` line 3 at r3](https://reviewable.io/reviews/cockroachdb/docs/13609#-N1jUOEICCH05J9X7TWD:-N1jUOEICCH05J9X7TWE:b-dyzuhz) ([raw file](https://github.com/cockroachdb/docs/blob/2ca1fa6739fbe6f73bb33f6d6f660f93da9849eb/v22.1/alter-super-region.md#L3)):*
> ```Markdown
> ---
> title: ALTER SUPER REGION
> summary: The ALTER SUPER REGION statement adds a super region made up of a set of regions such that all data from regional tables or regional by row tables with home rows in the super region is stored in the super region
> ```

shouldnt this say alters a super region?
___
*[`v22.1/multiregion-overview.md` line 78 at r2](https://reviewable.io/reviews/cockroachdb/docs/13609#-N0qi6UV1equx7fZJ3_x:-N1jUgnv8C7B3mKN57vl:bs9jjpt) ([raw file](https://github.com/cockroachdb/docs/blob/00120684f97e4ce39823eba87854ce9f87321ce4/v22.1/multiregion-overview.md#L78)):*

Let me know if I understand what you meant.  Here's what I added:

"The primary use case for super regions is data domiciling. As mentioned above, data from [regional](#regional-tables) and [regional by row](#regional-by-row-tables) tables will be stored only in regions that are members of the super region. Further, if the super region contains 3 or more regions and if you use [`REGION` survival goals](#survive-region-failures), the data domiciled in the super region will remain available if you lose a region."

.. but I'm less sure about the latter part about needing REGION survival goal. But it seems like it would still be necessary, right?  otherwise I'm still only at 3 replicas in 3 regions and losing 1 region puts me down to 2 replicas and I can't make progress on voting with an even number of replicas.

but maybe I misunderstood - please let me know
</blockquote></details>

technically you can make progress on 2 replicas if a quorum is 3, because the two that are up can agree. You are right that you will need region survival so that we spread the replicas out (but not so that we have 5 although that will help with latency). All in all I think its fine as written.

rmloveland (commented):
Thanks for the review Andy!  Thanks for pushing to just get ALTER SUPER REGION in now, should have started from there. Will pass this on to Eric for docs review now

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13609#-:-N1nf0il3L5U_hhVDhPG:brfaa5e)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, and @RichardJCai)

___
*[`v22.1/alter-super-region.md` line 3 at r3](https://reviewable.io/reviews/cockroachdb/docs/13609#-N1jUOEICCH05J9X7TWD:-N1nennR48FefKMapza5:b-tmrbqr) ([raw file](https://github.com/cockroachdb/docs/blob/2ca1fa6739fbe6f73bb33f6d6f660f93da9849eb/v22.1/alter-super-region.md#L3)):*

shouldnt this say alters a super region?
</blockquote></details>

Yes!  thanks for the catch - fixed it
___
*[`v22.1/multiregion-overview.md` line 78 at r2](https://reviewable.io/reviews/cockroachdb/docs/13609#-N0qi6UV1equx7fZJ3_x:-N1net7HD0cScD1Rq8Lz:b9q73kn) ([raw file](https://github.com/cockroachdb/docs/blob/00120684f97e4ce39823eba87854ce9f87321ce4/v22.1/multiregion-overview.md#L78)):*

technically you can make progress on 2 replicas if a quorum is 3, because the two that are up can agree. You are right that you will need region survival so that we spread the replicas out (but not so that we have 5 although that will help with latency). All in all I think its fine as written. 
</blockquote></details>

ok thanks for explaining this Andy! glad to hear the doc text is closer to correct than my comment :-)

v22.1/multiregion-overview.md
RichardJCai: I see, that makes sense to me we can stick to row.

v22.1/multiregion-overview.md
```diff
@@ -66,6 +66,32 @@ To show all of a database's regions, execute the [`SHOW REGIONS FROM DATABASE` s
...
+
+The primary use case for super regions is data domiciling. As mentioned above, data from [regional](#regional-tables) and [regional by row](#regional-by-row-tables) tables will be stored only in regions that are members of the super region. Further, if the super region contains 3 or more regions and if you use [`REGION` survival goals](#survive-region-failures), the data domiciled in the super region will remain available if you lose a region.
```
RichardJCai, line 78: s/Further/Furthermore (?)

RichardJCai (approved):
Nice work!

v22.1/multiregion-overview.md
rmloveland: I just checked the macOS thesaurus and it appears they are synonyms

Screen Shot 2022-05-17 at 10 13 59 AM

ericharmeling (changes requested):
Looks great, Rich (as always)! I had a few comments.

> Reviewed 9 of 15 files at r1, 7 of 8 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13609#-:-N2Hk2Xn7Zt44HsLbChn:b-hzfbrs)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @RichardJCai, and @rmloveland)

___
*[`_includes/v22.1/sql/super-region-considerations.md` line 6 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2Hf9vnAj80rk8kF1Tn:-N2Hf9vnAj80rk8kF1To:b45h) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/_includes/v22.1/sql/super-region-considerations.md#L6)):*
> ```Markdown
> - Super regions [must be enabled](#enable-super-regions).
> - Super regions can only contain [database regions](multiregion-overview.html#database-regions) that have already been added with [`ADD REGION`](add-region.html).
> - Each database region can only belong to one super region. In other words, given two super regions _A_ and _B_, the set of database regions in _A_ must be disjoint from the set of database regions in _B_.
> ```





_Suggestion:_
```Markdown
disjointed
```
___
*[`v22.1/alter-super-region.md` line 2 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2HiFJYDbEjQQxj8g6V:-N2HiFJZBpoVOxZpeucX:b-c6a1f6) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/v22.1/alter-super-region.md#L2)):*
> ```Markdown
> ---
> title: ALTER SUPER REGION
> ```

I'm not seeing this page in the ToC (and therefore, the preview). You should probably also reference it in `v22.1/alter-database.md`, `v22.1/add-super-region.md`,  `v22.1/multiregion-overview.md`, and `v22.1/sql-statements.md` (at least)?



_Code quote:_
```Markdown
ALTER SUPER REGION
```
___
*[`v22.1/show-super-regions.md` line 2 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2HjhfB9aGJuxLGbLQ-:-N2HjhfB9aGJuxLGbLQ0:b-3a7mtd) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/v22.1/show-super-regions.md#L2)):*
> ```Markdown
> ---
> title: SHOW SUPER REGIONS
> ```

It's up to you, but I would just consolidate this page with `SHOW REGIONS`, since they are kind of the same statement and use the same diagram and have similar examples?



_Code quote:_
```Markdown
SHOW SUPER REGIONS
```

RichardJCai:
One last thing actually, I think we should also call out zone config limitations here like we do in 
https://www.cockroachlabs.com/docs/stable/data-domiciling.html#limitations

Notably this point since SUPER REGIONS relies on zone configs. 
```
[Zone configs](https://www.cockroachlabs.com/docs/v21.2/configure-replication-zones) can be used for data placement but these features were historically built for performance, not for domiciling. The replication system's top priority is to prevent the loss of data and it may override the zone configurations if necessary to ensure data durability. For more information, see [Configure Replication Zones](https://www.cockroachlabs.com/docs/v21.2/configure-replication-zones#types-of-constraints).
```

I think we may want to document this more generally as well (the limitations) and that if constraints cannot be satisfied we prioritize durability (ie we will up replicate to a node outside of the constrained regions).

On the engineering side I think we should explore adding a zone config setting to not upreplicate in these cases and allow the data to become unavailable instead (?) if there are hard requirements for data domiciling.

rmloveland (commented):
Thanks for the review Eric!  Appreciate the feedback (and catching some errors that should have been obvious!) I think all your comments were addressed, PTAL

Richard, thanks for your comment.  It's a good point since this feature is also primarily for domiciling.  I added a slightly edited version of the text you link from the domiciling page to the 'Super regions' section of the 'Multiregion Capabilities Overview' page. 

For now I'm hoping that's enough to let us get something out there for the v22.1.0 release next week, but I also agree that we should probably add more info about this generally to the docs (and I'm sure there are users who would like to have the proposed feature you describe).

Let me know if you think this change is enough for now

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13609#-:-N2IZdhw205MOX2YMDAz:b9ct3yf)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ajstorm, @awoods187, @ericharmeling, and @RichardJCai)

___
*[`_includes/v22.1/sql/super-region-considerations.md` line 6 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2Hf9vnAj80rk8kF1Tn:-N2IXFgR7ol2Fo5IXVgs:b-t3f2wo) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/_includes/v22.1/sql/super-region-considerations.md#L6)):*
> ```Markdown
> - Super regions [must be enabled](#enable-super-regions).
> - Super regions can only contain [database regions](multiregion-overview.html#database-regions) that have already been added with [`ADD REGION`](add-region.html).
> - Each database region can only belong to one super region. In other words, given two super regions _A_ and _B_, the set of database regions in _A_ must be disjoint from the set of database regions in _B_.
> ```

this usage is def awkward but it's pretty common in math writing on sets (for good or ill :-} ) - I added a link to the wikipedia page on disjoint sets which also has that usage, so hopefully it will be less awkward with a bit more context
___
*[`v22.1/alter-super-region.md` line 2 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2HiFJYDbEjQQxj8g6V:-N2IT3pD5-0ctPfgv9gR:b5a8rvp) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/v22.1/alter-super-region.md#L2)):*

I'm not seeing this page in the ToC (and therefore, the preview). You should probably also reference it in `v22.1/alter-database.md`, `v22.1/add-super-region.md`,  `v22.1/multiregion-overview.md`, and `v22.1/sql-statements.md` (at least)?
</blockquote></details>

Added to 'ALTER DATABASE' page

Added to 'SQL Statements' page

Thanks for catching these omissions!

Looked and there is a reference in v22.1/multiregion-overview.md in the 'Super regions' section

There is a reference from v22.1/add-super-region under 'Considerations' where we talk about altering a super region, and we also link to it from the 'See also'

I wonder if there's something wonky with the revisions view also?  I just squashed everything and re-pushed , please let me know if it's still missing from your POV

Again thanks for catching this stuff!
___
*[`v22.1/show-super-regions.md` line 2 at r4](https://reviewable.io/reviews/cockroachdb/docs/13609#-N2HjhfB9aGJuxLGbLQ-:-N2IU5uw4hx-5a3nCjy-:b-ma6skf) ([raw file](https://github.com/cockroachdb/docs/blob/62003ea26c4451f434345bcd4c0a6e360475ba95/v22.1/show-super-regions.md#L2)):*

It's up to you, but I would just consolidate this page with `SHOW REGIONS`, since they are kind of the same statement and use the same diagram and have similar examples?
</blockquote></details>

good point - Def thought about it, but I think right now that could be confusing since the super regions stuff is behind a setting, so not everyone will have it enabled.  Also think having a separate page might be a bit better for findability in general

created an issue to look at merging once this super regions stuff is fully GA 
https://cockroachlabs.atlassian.net/browse/DOC-3438

ericharmeling (approved):
Thanks for getting to my comments. Looks great. :)

rmloveland:
@RichardJCai I'd like to merge this PR sometime tomorrow unless you have any additional feedback on my update to the _Super regions_ section of the **Multiregion Capabilities Overview** page re: super regions/zcfgs being historically built for availability

PR #13623, opened 2022-04-15
Update Statements and Statement Details pages.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2231
https://cockroachlabs.atlassian.net/browse/DOC-3287
https://cockroachlabs.atlassian.net/browse/DOC-2890
https://cockroachlabs.atlassian.net/browse/DOC-2944
https://cockroachlabs.atlassian.net/browse/DOC-2858
https://cockroachlabs.atlassian.net/browse/DOC-2515
https://cockroachlabs.atlassian.net/browse/DOC-707
https://cockroachlabs.atlassian.net/browse/DOC-2625

The single pages for statements and transactions shared between Cloud Console and DB Console had to be split because 21.2 and 22.1 have different features.


_includes/v22.1/ui/statement-details.md
```diff
@@ -0,0 +1,150 @@
...
+
+<img src="{{ 'images/v22.1/ui_activate_diagnostics.png' | relative_url }}" alt="Statements diagnostics" style="border:1px solid #eee;max-width:80%" />
+
+To activate diagnostics collection:
+
+1. Click the **Activate diagnostics** button. <span class="version-tag">New in v22.1</span> The **Activate statement diagnostics** dialog displays.
+
+    <img src="{{ 'images/v22.1/ui_activate_diagnostics_dialog.png' | relative_url }}" alt="Statements diagnostics" style="border:1px solid #eee;max-width:80%" />
+
+1. Choose whether to activate collection on the next statement execution (default) or if execution latency exceeds a certain time. If you choose the latter, accept the default latency of 100 milliseconds, or specify a different time.
```
kevin-v-ngo, line 96: Can we call out the performance overhead for enabling diagnostics? All executions of the statement fingerprint will run slower with diagnostics enabled until diagnostics are collected.

v22.1/ui-statements-page.md
```diff
@@ -7,12 +7,31 @@ docs_area: reference.db_console
...
+## Search and filter
+
+By default, the Statements page shows SQL statements from all applications and databases running on the cluster.
+
+### Search statements and statement plans
+
+To search using the search field, type a string over `Search Statements` and press `Enter`. The search will look for matches in:
+
+- Statement fingerprints.
+- <span class="version-tag">New in v22.1:</span> [Statement plans](#explain-plans). For example, suppose you want to identify statements using a specific index. You can search for a specific index and if the index appears in the statement plan text the search will return that statement fingerprint.
```
kevin-v-ngo, line 31: @maryliag, do you know if Gerardo's changes for searching by index is supported against all plans? Since we capture all plans now in 22.1, did this break/impact Gerardos feature to search by index at all?

kevin-v-ngo (commented):
Few comments. Thanks Stephanie!

_includes/v21.2/ui/statement-details.md
```diff
@@ -0,0 +1,145 @@
...
+  - **Total executions**: the total number of executions of statements with this fingerprint. It is calculated as the sum of first attempts and retries.
+  - **Retries**: the cumulative number of [retries]({{ link_prefix }}transactions.html#transaction-retries) of statements with this fingerprint within the aggregation interval.
+  - **Max Retries**: the highest number of retries of a single statement with this fingerprint within the aggregation interval. For example, if three statements with the same fingerprint had to be retried 0, 1, and 5 times, then the Max Retries value for the fingerprint is 5.
+
+### Diagnostics
+
+The **Diagnostics** section allows you to activate and download diagnostics for the SQL statement fingerprint.
+
+{{site.data.alerts.callout_info}}
+The **Diagnostics** tab is not enabled for {{ site.data.products.serverless }} clusters.
```
maryliag, line 53: should we also mention this tab is also not enabled if user has `VIEWACTIVITYREDACTED` role?

v21.2/ui-statements-page.md
```diff
@@ -7,12 +7,28 @@ docs_area: reference.db_console
 
 {% include {{ page.version.version }}/ui/admin-access.md %}
 
-The **Statements** page helps you:
+The **Statements** page  of the DB Console helps you:
```
maryliag, line 10: this page exists on both DB and CC Console, so I don't think we should be saying only DB Console here (do we even use this term publicly?)

v22.1/ui-statements-page.md
```diff
@@ -7,12 +7,31 @@ docs_area: reference.db_console
 
 {% include {{ page.version.version }}/ui/admin-access.md %}
 
-The **Statements** page helps you:
+The **Statements** page  of the DB Console helps you:
```
maryliag, line 10: same as above

v22.1/ui-statements-page.md
maryliag: is just looking at the sampled plan. It would be to expensive to decode every single plan to do the search on top of it

v21.2/ui-statements-page.md
stbof: Yes, we do use the term DB Console publicly: https://deploy-preview-13623--cockroachdb-docs.netlify.app/docs/dev/ui-overview.html

This particular page appears only in the CockroachDB docs: https://deploy-preview-13623--cockroachdb-docs.netlify.app/docs/dev/ui-statements-page.html

There's a different page for CC Console: https://deploy-preview-13623--cockroachdb-docs.netlify.app/docs/cockroachcloud/statements-page.html

The pages are constructed from shared components that are included into the parent page.

v22.1/ui-statements-page.md
stbof: See my response above.

_includes/v21.2/ui/statement-details.md
stbof: We do have the warning at the bottom of this section: https://deploy-preview-13623--cockroachdb-docs.netlify.app/docs/v22.1/ui-statements-page.html#diagnostics that says you can't download bundles.  Somehow I missed the tab behavior. I'll fix that. Thanks for the catch.

stbof:
@maryliag I have questions about your comment about sampled plan: https://github.com/cockroachdb/docs/pull/13623#discussion_r866839695 and also about how to describe fields in the Statements table now that aggregation interval no longer is displayed.

v22.1/ui-statements-page.md
maryliag: Every time we execute a statement, it uses a plan, we sample a few of them to get the most recent stats from it.

@kevin-v-ngo I don't know how we can describe this, because is not straight forward to tell the user which one was sample, also we might stop doing the sampling part, since we no longer the sample plan anywhere. Do you think is worth removing this part of the docs?

_includes/v21.2/ui/statement-details.md
```diff
@@ -0,0 +1,147 @@
+## Statement Details page
+
+Click a SQL statement fingerprint to open **Statement Details**. For each statement fingerprint, the details include:
+
+- [Overview](#overview)
+- [Diagnostics](#diagnostics)
+- [Explain plan](#explain-plan)
+- [Execution stats](#execution-stats)
+
+The Statement Details page supports the search param `aggregated_ts`. If set, the page displays the statement statistics aggregated at that interval. If unset, the page displays the statement statistics aggregated over the date range specified on the Statements page.
```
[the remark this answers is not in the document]
maryliag: Looks like this part was removed, but let me know if you need any extra information

maryliag:
> Execution Count | Cumulative number of executions of statements with this fingerprint within the aggregation interval.
Should aggregation interval be replaced with date range?

Yes

> Tooltips in many of the fields are referring to a time interval and link to a location (https://www.cockroachlabs.com/docs/stable/ui-statements-page.html#time-interval) that no longer exists.
Should that link be to https://www.cockroachlabs.com/docs/stable/ui-statements-page.html#date-range instead? If that's the right location, I could add a time-interval anchor right above date-range.

Yes, we should make the change

Would you mind opening an issue to update the tooltips to use date range instead of aggregation interval? Thank you!




> Statistics aggregation is controlled by the aggregation interval property, set to 1 hour. Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
where aggregation interval used to be in the table. However, that property no longer exists in the table. What should it say instead?

It's used on the database table, not the ui table, that is what this tooltip is explaining and that value still exists. I see this tooltip being used only on the clear stats, and there is not change that should be done for this tooltip.

stbof:
> > Execution Count | Cumulative number of executions of statements with this fingerprint within the aggregation interval.
> > Should aggregation interval be replaced with date range?
> 
> Yes
> 
> > Tooltips in many of the fields are referring to a time interval and link to a location (https://www.cockroachlabs.com/docs/stable/ui-statements-page.html#time-interval) that no longer exists.
> > Should that link be to https://www.cockroachlabs.com/docs/stable/ui-statements-page.html#date-range instead? If that's the right location, I could add a time-interval anchor right above date-range.
> 
> Yes, we should make the change
> 
> Would you mind opening an issue to update the tooltips to use date range instead of aggregation interval? Thank you!
> 
> > Statistics aggregation is controlled by the aggregation interval property, set to 1 hour. Statistics between two hourly intervals belong to the nearest hour rounded down. For example, a statement execution ending at 1:50 would have its statistics aggregated in the 1:00-2:00 interval.
> > where aggregation interval used to be in the table. However, that property no longer exists in the table. What should it say instead?
> 
> It's used on the database table, not the ui table, that is what this tooltip is explaining and that value still exists. I see this tooltip being used only on the clear stats, and there is not change that should be done for this tooltip.

That paragraph isn't in a tooltip. It appears in the docs here:

https://deploy-preview-13623--cockroachdb-docs.netlify.app/docs/stable/ui-statements-page.html#statement-statistics

v22.1/ui-statements-page.md
kevin-v-ngo: Yes, we should remove it since the sampled plan is no longer being used anywhere and we are tracking to remove it in the future since it doesn't represent the actual plan and could be a false signal.

I do wonder if there is a way we can perform the search so that it's not expensive. This was a great feature. Perhaps we can file a tracking improvement here and revisit this especially with future improvements with the obs service.

stbof: @kevin-v-ngo I'm not sure what you are referring to by "remove it". Do you mean remove the entire bullet that says you can search plans?

kevin-v-ngo: FYI - https://github.com/cockroachdb/cockroach/issues/81485

rmloveland (approved):
LGTM

PR #13633, opened 2022-04-18
Removed guidance to place version tags in headers from the style guide

ericharmeling:
Follow-up on https://cockroachlabs.slack.com/archives/C01NFCXULTX/p1650321116723369.


stbof (approved):

stbof:
Thanks for the quick fix!

PR #13665, opened 2022-04-21
Quick fix to diagram URL

kathancox:
The sql diagram on the 22.1 session vars page is not showing for the `SET SESSION` sql: https://www.cockroachlabs.com/docs/dev/set-vars.html#set-session 

Corrected the URL.


ericharmeling (approved):
LGTM!

PR #13666, opened 2022-04-21
Link read and write amp to our own descriptions

rmloveland:
Fixes DOC-3361
Fixes DOC-420

Summary of changes:

- Update all the old 'write amplification' links to Wikipedia so they
  now point to our own description of write amp in the Pebble
  architecture docs instead.

- Do the same thing for mentions of 'read amplification'


ericharmeling (approved):
Nice! LGTM

PR #13669, opened 2022-04-21
Cloud architecture reorganization and clarification

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-3347
Edit architecture page layout
Add deployment option recommendation


ianjevans (commented):
Looks good, but the Serverless architecture diagram is way too big, which makes it seem low-res. Maybe play around with the `style="max-width:100%"` setting.

ianjevans (approved):

PR #13671, opened 2022-04-21
Clarify ALTER ROLE ALL behavior

rafiss:
The `root` user cannot be configured in this way.


rmloveland (approved):
LGTM, thanks @rafiss !

PR #13685, opened 2022-04-22
TA 79066

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-3364

### DIFF
--- a/_data/advisories.yml
+++ b/_data/advisories.yml
@@ -1,6 +1,6 @@
 - advisory: a79066
   summary: Data key rotation is inadvertently disabled if the store key hasn't changed since the last node start.
-  versions: All clusters with encryption-at-rest enabled running versions of CockroachDB v20.2.x, v21.1.0-v21.1.18, v21.2.0-v21.2.9, and v22.1.0.
+  versions: All clusters with encryption-at-rest enabled running versions of CockroachDB v20.2.x, v21.1.0-v21.1.18, and v21.2.0-v21.2.9.
   date: May 9, 2022
 - advisory: a79384
   summary: The optimizer has been found to create logically incorrect query plans in some cases.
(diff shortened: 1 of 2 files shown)


itsbilal (approved):
:lgtm: thanks for putting this up!

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13685#-:-N0bAu5G0Tdpu0KdQyL3:bp55wli)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @nickelnickel and @stbof)

___
*[`advisories/a79066.md` line 12 at r4](https://reviewable.io/reviews/cockroachdb/docs/13685#-N0bAexeAqCWyfovzbvf:-N0bAexf7KkwB7-ddZcV:b-dr22x0) ([raw file](https://github.com/cockroachdb/docs/blob/f88d228f8c45198cc0210ffd3ab0092b988b68b8/advisories/a79066.md#L12)):*
> ```Markdown
> ## Description
> 
> CockroachDB [Encryption-At-Rest (EAR)](../{{site.versions["stable"]}}/security-reference/encryption.html#encryption-at-rest-enterprise) provides automatic data key rotation at a configurable interval, defaulting to every week. While the store key is set by the operator when they start CockroachDB, data keys are managed by CockroachDB's EAR layer. Data key rotation improves data security by reducing the effect of any single exposed data key, as newer data files would be encrypted with newer keys.
> ```

"By reducing the effect of exposure of any single data key" is probably better now that we no longer say blast radius.

nickelnickel (approved):
I thought this was being published May 2nd along with 21.2.10. Otherwise LGTM.

stbof:
> I thought this was being published May 2nd along with 21.2.10. Otherwise LGTM.

It's waiting for 21.1.19 on May 9.

PR #13691, opened 2022-04-22
Override to keep text white for the Start Instantly button

ianjevans:
Fixes a CSS issue where hovering over the Start Instantly button in the header of our docs pages turns the text to the background color.


nickvigilante (approved):
Thanks for fixing this.

PR #13709, opened 2022-04-25
Add additional troubleshooting entry for certificate signed by unknown authority errors

ianjevans:
Users may encounter errors when connecting to Serverless clusters when they have existing certificates in `~/.cockroach-certs`.

Fixes DOC-3377.

### DIFF
--- a/cockroachcloud/troubleshooting-page.md
+++ b/cockroachcloud/troubleshooting-page.md
@@ -38,11 +38,11 @@ Error: x509: certificate signed by unknown authority
 Failed running "sql"
 ~~~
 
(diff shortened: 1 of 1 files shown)


stbof (approved):
The length of the string in the `summary:`  metadata looks too short. Did we change the length test?

ianjevans:
> The length of the string in the `summary:` metadata looks too short. Did we change the length test?

Nope, but Jest passed the test according to the build log. Maybe because that line is 49 characters long plus CR? Can't really tell from the [code](https://github.com/cockroachdb/docs/blob/master/__tests__/frontmatter.js#L15).

PR #13749, opened 2022-04-29
DOC-270: Run Vale against changed files only

nickvigilante:
Fixes DOC-270

### DIFF
--- a/netlify/build
+++ b/netlify/build
@@ -14,19 +14,21 @@ npx snippet-enricher-cli --targets="shell_curl" --input=spec_30.json > spec_30_e
 rm spec_30.json
 popd
 
-gem install bundler
-bundle install
+gem install bundler --silent
+bundle install --quiet
 build _config_cockroachdb.yml
 
 cp _site/docs/_redirects _site/_redirects
 cp _site/docs/404.html _site/404.html
 
 # Set up htmltest
-go get -u github.com/cockroachdb/htmltest
-pushd $GOPATH/src/github.com/cockroachdb/htmltest && ./build.sh && popd
+
+go get -u github.com/cockroachdb/htmltest > /dev/null
+pushd $GOPATH/src/github.com/cockroachdb/htmltest && ./build.sh>/dev/null 2>&1 && popd
 
 # Run htmltest to check external links on scheduled nightly runs
 # (see .github/workflows/nightly.yml)
+
 if [[ "$INCOMING_HOOK_TITLE" = "nightly" ]]; then
 	$GOPATH/src/github.com/cockroachdb/htmltest/bin/htmltest
 	if [[ $? != 0 ]]; then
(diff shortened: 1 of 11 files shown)


netlify/build
```diff
@@ -41,10 +43,19 @@ if [[ "$CONTEXT" = "production" ]]; then
 fi;
 
 # Set up Vale, a text linter
-wget -O- https://github.com/errata-ai/vale/releases/download/v1.2.7/vale_1.2.7_Linux_64-bit.tar.gz | tar -xz vale
+wget --quiet -O- https://github.com/errata-ai/vale/releases/download/v1.2.7/vale_1.2.7_Linux_64-bit.tar.gz | tar -xz vale >/dev/null 2>/dev/null
```
ianjevans, line 46: The latest Vale is v2.16. Should we try to upgrade?

v22.1/add-constraint.md
```diff
@@ -19,6 +19,8 @@ The [`DEFAULT`](default-value.html) and [`NOT NULL`](not-null.html) constraints
 
 {% include {{ page.version.version }}/sql/combine-alter-table-commands.md %}
 
+We don't like this file bababooie asdlfkmflk
```
ianjevans, line 22: Misspellings currently break the build. I get that Warnings are easy to ignore, but I'm worried that such a strict policy will cause more headaches than it solves for new terms we haven't yet added to the exception list.

ianjevans (commented):
Looks good. I left a couple review comments.

I also think that Vale errors need to be surfaced in the PR if possible. If we can eliminate the need to go to the Netlify build log easily, we should do that. But that may not be possible.

v22.1/add-constraint.md
nickvigilante: I'll send out a poll to the team to see what people would want to see in terms of functionality.

nickvigilante:
> Looks good. I left a couple review comments.
> 
> I also think that Vale errors need to be surfaced in the PR if possible. If we can eliminate the need to go to the Netlify build log easily, we should do that. But that may not be possible.

I'm thinking about adding a separate check entirely that just displays the output of Vale and splitting up the output.

netlify/build
nickvigilante: Done.

v22.1/add-constraint.md
nickvigilante: I personally would prefer that avenue, and I can add documentation in the wiki as to how to edit the list of terms.

stbof:
@nickvigilante Could you add a broken link to this PR?

stbof:
> @stbof Broken links is an htmltest thing, but there's also another GitHub action for that 😄 

Actually I forgot, I really want to find out broken links locally, not in the build. It takes too long.

ianjevans (approved):
Looks great. Having the Vale messages in the changed files is so much better.

PR #13765, opened 2022-05-02
Added forward-looking guidance to Style Guide

ericharmeling:
Fixes https://cockroachlabs.atlassian.net/browse/DOC-2072.

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -81,6 +81,11 @@ Other general guidance about language and tone:
 
     **Example:** If you leave versioned binaries on your servers, you do not need to do anything.
 
+- Avoid using forward-looking language when writing about supported syntax and behavior:
+    - Do not suggest that a feature may or may not be added in a future release.
+    - Do not use the words "yet" and "currently" when writing about a feature that we do or do not support.
+    - Do not reference the internal product roadmap.
+
 ## Inclusive language
 
 We want our education materials to be inclusive and written with diversity in mind. This section provides general guidelines, best practices, and examples when writing docs or training materials. This is a work in progress and we welcome any feedback and additions.


StyleGuide.md
```diff
@@ -81,6 +81,11 @@ Other general guidance about language and tone:
 
     **Example:** If you leave versioned binaries on your servers, you do not need to do anything.
 
+- To help users set reasonable expectations for the future of CockroachDB's feature set, avoid using forward-looking language when writing about supported syntax and behavior:
```
rmloveland, line 84: Suggest making 2nd clause first so it's more obvious

eg something like "Avoid using forward-looking ..."

IMO the justification clause can be dropped, if someone wants to litigate decision history they can read the Jira issues or #docs-style.  But we shouldn't impose that cost on every reader of this guide, since most of the time I just want an answer and don't care that much about why

----

meta comment that is not really about this PR but prompted: I don't think explanation for _any_ guidance in the style guide is needed.  When I'm looking at this I almost always want to find an answer ASAP: how do I ...?   If I need justification I can always ask someone on the team for that context/history

put another way: I want the fastest answer to my style question at least cost.  if I want a justification for a style guide decision, I can pay the higher cost of digging into decision making.  But folks who just want the info should not have to pay that higher cost of scanning more words

like I said, not really about your PR but prompted by that one clause

rmloveland (approved):
LGTM, had a suggestion and a weird meta-comment (of course :-) )

StyleGuide.md
kathancox: Yes, I think I agree with Rich... though I did have to justify this today 😁 

Also, I think the style guide is a mixture of instructions and instructions + justification right now, so either way would work imo. 

(This is a very unhelpful review.... I like BOTH ways!)

kathancox (approved):
LGTM!

StyleGuide.md
ericharmeling: > IMO the justification clause can be dropped

Done!

ericharmeling: Re: the meta comment, maybe this could be added as a discussion item for the next style team meeting?

ericharmeling (commented):
TFTRs!

StyleGuide.md
rmloveland: Added it to next meeting agenda, thanks!

PR #13795, opened 2022-05-04
CC release notes edit

gemma-shay:
Remove metrics --only avail in 22.1 which is not out yet


kevin-v-ngo (approved):
Thanks Gemma! Looks good.

PR #13807, opened 2022-05-05
Fix New in markup.

stbof:
Add colon.

### DIFF
--- a/v22.1/show-schedules.md
+++ b/v22.1/show-schedules.md
@@ -130,6 +130,6 @@ To view a specific schedule, use the schedule's `id`:
 - [`RESUME JOB`](pause-job.html)
 - [`CANCEL JOB`](cancel-job.html)
 - [Take Full and Incremental Backups](take-full-and-incremental-backups.html)
-- [Row-Level TTL](row-level-ttl.html) <span class="version-tag">New in v22.1</span>
+- [Row-Level TTL](row-level-ttl.html) <span class="version-tag">New in v22.1:</span>
 - [Use the Built-in SQL Client](cockroach-sql.html)
 - [`cockroach` Commands Overview](cockroach-commands.html)
--- a/v22.1/show-schedules.md
+++ b/v22.1/show-schedules.md
@@ -6,7 +6,7 @@ docs_area: reference.sql
 ---
 
 The `SHOW SCHEDULES` [statement](sql-statements.html) lists all of the currently active [backup schedules](create-schedule-for-backup.html).
- 
+
 <span class="version-tag">New in v22.1:</span> It also lists all of the currently active scheduled jobs for [Row-Level TTL](row-level-ttl.html).
 
 ## Required privileges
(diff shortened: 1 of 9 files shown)


v22.1/show-schedules.md
```diff
@@ -130,6 +130,6 @@ To view a specific schedule, use the schedule's `id`:
 - [`RESUME JOB`](pause-job.html)
 - [`CANCEL JOB`](cancel-job.html)
 - [Take Full and Incremental Backups](take-full-and-incremental-backups.html)
-- [Row-Level TTL](row-level-ttl.html) <span class="version-tag">New in v22.1</span>
+- **New in v22.1:** [Row-Level TTL](row-level-ttl.html)
```
rmloveland, line 133: I don't think we should change from using our standard "new in XX" version tag just so there can be a colon.  For the following reasons:

1. it's different visually from the colored tag we normally use, and thus doesn't send the same signal to the reader
2. less important but still a thing: the existing tag is used by our ["bump docs version" script](https://github.com/cockroachdb/ed-tools/blob/main/docs/bump-major-version/bump_major_version.py#L181)

I'm fine with adding a colon to all uses of the existing tag though

v22.1/ui-databases-page.md
```diff
@@ -68,15 +68,15 @@ The table details include:
 - **Size**: the approximate disk size of all replicas of this table on the cluster.
 - **Replicas**: the number of [replicas](architecture/replication-layer.html) of this table on the cluster.
 - **Ranges**: the number of [ranges](architecture/glossary.html#architecture-range) in this table.
-- **Table Stats Last Updated**: <span class="version-tag">New in v22.1</span> the last time table statistics were created or updated.
-- **Auto Stats Collection**: <span class="version-tag">New in v22.1</span> whether [automatic statistics collection](cost-based-optimizer.html#control-automatic-statistics) is enabled.
+- **New in v22.1:** **Table Stats Last Updated**: the last time table statistics were created or updated.
```
rmloveland, line 71: As above I don't think we should change from our version tag scheme

v22.1/ui-databases-page.md
```diff
@@ -68,15 +68,15 @@ The table details include:
...
+- **New in v22.1:** **Auto Stats Collection**: whether [automatic statistics collection](cost-based-optimizer.html#control-automatic-statistics) is enabled.
```
rmloveland, line 72: Same as above re: we should use our existing version tag scheme

rmloveland (changes requested):
LGTM except where the version tag is changed to bold text (reasons explained inline below)

v22.1/show-schedules.md
stbof: I'm just implementing what's in the style guide: https://github.com/cockroachdb/docs/blob/master/StyleGuide.md#version-tags 

Note: Do not use version tags in bulleted lists items. To denote a new feature in a bulleted list, start the bulleted item with "**New in vX.X:**".

PR #13818, opened 2022-05-05
Explain UPSERT behavior with a subset of columns

mgartner:
The behavior of `UPSERT` when a subset of column values is provided is confusing, so I think it deserves some explanation in our docs (see https://github.com/cockroachdb/cockroach/issues/73984). This is my quick attempt to explain it, but I'm hoping that the docs team can take over from here to refine the language and make it as clear as possible. I've include some examples which should help understand the behavior.

I think this should be documented for all past versions that support UPSERT, though I've only included it in 22.1 in this PR.

### DIFF
--- a/v22.1/upsert.md
+++ b/v22.1/upsert.md
@@ -7,6 +7,8 @@ docs_area: reference.sql
 
 The `UPSERT` [statement](sql-statements.html) is semantically equivalent to [`INSERT ON CONFLICT`](insert.html#on-conflict-clause), but the two may have slightly different [performance characteristics](#considerations). It inserts rows in cases where specified values do not violate uniqueness constraints, and it updates rows in cases where values do violate uniqueness constraints.
 
+An `UPSERT` statement that affects a proper subset of columns in the target table behaves differently depending on whether or not the target columns are explicit in the statement. If no target columns are explicitly specified, `UPSERT` will always write the default values for columns without explicit values in the `UPSERT` statement. If target columns are explicitly specified, `UPSERT` will not override the values of existing columns that do not have new values in the `UPSERT` statement. For more details, see [`UPSERT` TODO](#upsert-a-proper-subset-of-columns) below.
+
 To learn more about how to perform and when to use an upsert in CockroachDB, PostgreSQL, and MySQL, see [Upsert in SQL: What is an Upsert, and When Should You Use One?](https://www.cockroachlabs.com/blog/sql-upsert/) .
 
 ## Considerations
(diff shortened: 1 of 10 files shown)


v22.1/upsert.md
```diff
@@ -237,6 +239,77 @@ In such a case, you would need to use the [`INSERT ON CONFLICT`](insert.html) st
...
++----+------+----------+
+| id | name | balance  |
++----+------+----------+
+|  1 |   a1 | 10000.50 |
+|  2 |   b1 | 20000.75 |
+|  3 |   c1 |  6325.20 |
++----+------+----------+
+~~~
+
+Upserting a proper subset of columns without specifying the column names will write the default values of the unspecified columns when there is a conflict on the primary key. Notice how the account with `id` of `1` has a balance of `0` (the column's default value) after the `UPSERT`:
```
stbof, line 275: @mgartner What do you mean "when there is a conflict on the primary key"?  I thought the point is simply that the target columns are missing.

mgartner: I was trying to convey that this slight behavioral difference has no effect when the `UPSERT` inserts a row, which happens when there is no conflict on the primary key. The default column value will always be used. Perhaps this is obvious — there's no other value that could be used.

Example:

```
defaultdb> CREATE TABLE accounts (
    id INT PRIMARY KEY,
    name STRING,
    balance DECIMAL(10, 2) DEFAULT 0
);
CREATE TABLE

defaultdb> SELECT * FROM accounts;
  id | name | balance
-----+------+----------
(0 rows)

defaultdb> UPSERT INTO accounts VALUES (1, 'a1');
INSERT 1

defaultdb> UPSERT INTO accounts (id, name) VALUES (2, 'b1');
INSERT 1

defaultdb> SELECT * FROM accounts;
  id | name | balance
-----+------+----------
   1 | a1   |    0.00
   2 | b1   |    0.00
(2 rows)
```

stbof:
@mgartner I pushed an update with some style and organization changes. If you approve I will propagate to other versions.

v22.1/upsert.md
```diff
@@ -5,23 +5,47 @@ toc: true
...
+
+To learn more about how to perform and when to use an upsert in CockroachDB, PostgreSQL, and MySQL, see [Upsert in SQL: What is an Upsert, and When Should You Use One?](https://www.cockroachlabs.com/blog/sql-upsert/).
 
 ## Considerations
 
-- `UPSERT` considers uniqueness only for [Primary Key](primary-key.html) columns. `INSERT ON CONFLICT` is more flexible and can be used to consider uniqueness for other columns. For more details, see [How `UPSERT` transforms into `INSERT ON CONFLICT`](#how-upsert-transforms-into-insert-on-conflict) below.
+- An `UPSERT` statement affecting a proper subset of columns behaves differently depending on whether or not you specify the target columns in the statement.
+
+    - If you specify target columns (e.g., `UPSERT INTO accounts (id, name) VALUES (2, 'b2');`), the values of columns that do not have new values in the `UPSERT` statement will not be updated.
+    - If you do not specify the target columns (e.g., `UPSERT INTO accounts VALUES (2, 'b2');`), the value of columns that do not have new values in the `UPSERT` statement will be updated to the default values.
```
mgartner, line 38: nit: "to **their** default values"

mgartner (commented):
This looks great! Thank you!!

PR #13824, opened 2022-05-06
Added docs for S3 storage classes backups, exports, changefeeds

kathancox:
Fixes [DOC-2438](https://cockroachlabs.atlassian.net/browse/DOC-2438)

Added documentation on the new S3_storage_class parameter to 22.1 & 21.2 documentation: 

- Longer section to Use Cloud Storage For Bulk Ops
- Short-form examples to `CREATE CHANGEFEED`, `EXPORT`, and `BACKUP`
- Some optimizations of the URL section in backup


_includes/v21.2/misc/storage-classes.md
```diff
@@ -0,0 +1,13 @@
+Use the parameter to set one of the following [storage classes](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#AmazonS3-PutObject-request-header-StorageClass):    
+
+- `STANDARD`
```
dt, line 3: Are we worried that this could become stale as amazon changes their allowed options? would it make sense to just link to their docs instead of duplicating it?

dt (approved):

andf-crl (approved):
LGTM with small nits -- and agreed with David's comment regarding just external linking to upstream. Nice one!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13824#-:-N2TBenYBOb-v4PvjNUT:b-ovtx4s)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl, @dt, and @kathancox)

___
*[`_includes/v21.2/misc/storage-classes.md` line 3 at r2](https://reviewable.io/reviews/cockroachdb/docs/13824#-N1Oz9ig7F5Aq36Tj7_k-r2-3:-N2T43gxBlmvlFL-3JDw:b-t3d625) ([raw file](https://github.com/cockroachdb/docs/blob/51d7ad69cb7a25c876e95732ee904e44b9e6101d/_includes/v21.2/misc/storage-classes.md#L3)):*

Yes, and no (kind of). It's an annoying balance, because I would rather list the options here so users can just have one stop. But, am I realistically going to keep on top of the changes — no. Also, there's a high probability that users will need to go to the Amazon docs anyway. @andf-crl as you do your docs review, can you let me know your thoughts? I feel like David's question has kind of convinced me to remove and link out.
</blockquote></details>

Agreed with @dt. For example, AWS just added `GLACIER_IR` [Dec last year](https://techcrunch.com/2021/12/01/aws-launches-four-new-storage-services-aimed-at-improving-data-management/). Best indeed to reference their canonical resource via an external link. However, to keep the one-stop-shop approach we can just use a common or unlikely-to-change upstream storage class in our examples. I like your choice of `INTELLIGENT_TIERING` because it is indeed a somewhat common "set-and-forget" choice for your S3 storage, even as you might introduce or retire underlying S3 storage tiers: thus likely to be relevant and recognizable to the reader. Nice one.

Likely this means retiring this include and just updating the inline copy to generalize a bit. I've included a suggestion in line alongside such an instance later in this review.
___
*[`v21.2/backup.md` line 95 at r2](https://reviewable.io/reviews/cockroachdb/docs/13824#-N2T7giw3rQf04tzaR6K:-N2T7giw3rQf04tzaR6L:b-leaf3w) ([raw file](https://github.com/cockroachdb/docs/blob/51d7ad69cb7a25c876e95732ee904e44b9e6101d/v21.2/backup.md#L95)):*
> ```Markdown
> 
> {{site.data.alerts.callout_success}}
> Backups support [Amazon S3 storage classes](#backup-with-an-s3-storage-class). For more detail, see [Additional cloud storage feature support](use-cloud-storage-for-bulk-operations.html#additional-cloud-storage-feature-support).
> ```

Nit: since this is a UL of "for more info" links, I think it's safe to add this fourth "see more info" directly in line as UL #4.

Nit: should we indicate "New in v21.2.6" (and 22.1 for the other instance) here (and on the other statements)? Certainly optional since we do so centrally on `use-cloud-storage-for-bulk-operations.md`, but this (and other statements) is likely where the user will first encounter mention of this new feature (outside of release notes).
___
*[`v21.2/use-cloud-storage-for-bulk-operations.md` line 235 at r2](https://reviewable.io/reviews/cockroachdb/docs/13824#-N2T8di_0nsAxJeSWfEY:-N2T8di_0nsAxJeSWfEZ:bd69g9c) ([raw file](https://github.com/cockroachdb/docs/blob/51d7ad69cb7a25c876e95732ee904e44b9e6101d/v21.2/use-cloud-storage-for-bulk-operations.md#L235)):*
nit: I think we need a tiny intro to this code to indicate it is an example (i.e. that `INTELLIGENT_TIERING` is user-provided) with something like:

> ... to configure a storage class type. For example, the following S3 connection URI specifies the `INTELLIGENT_TIERING` storage class:

Same nit for the three statement examples, with maybe a slight tweak:

>  use the `S3_STORAGE_CLASS` parameter with the desired class, in this case `INTELLIGENT_TIERING`:

kathancox:
Hey @andf-crl, Thanks for the review, as always super helpful! Yep, adapted without the callout list — I've kept the links out in an include since it is repeated/exactly the same across pages. Love the other suggestions and have resolved. In the backup file URL section, have kept the tip/note box for now since it the features are somewhat different to the basic necessary parameter-forming help. I'm also working on this section in another PR this coming week.

Issue #13696, opened 2022-04-22
SHOW RANGES Doc Update : describe purpose of WHERE clause


hand-crdb:
Re: [SHOW RANGES](https://cockroachlabs.com/docs/v21.2/show-ranges.html)

## Issue Description

The examples do not explain the purpose of this part of the queries: `WHERE "start_key" NOT LIKE '%Prefix%'`

## Suggested Resolution 

Explain the purpose of the above `WHERE` clause.

PR #13833, opened 2022-05-09
Explained purpose of where clause

chaneth8:
Related issue:https://github.com/cockroachdb/docs/issues/13696 

This is my first PR to an open source project, so I'm open to any feedback :)

### DIFF
--- a/v19.2/show-ranges.md
+++ b/v19.2/show-ranges.md
@@ -50,6 +50,8 @@ Field | Description
 
 ### Show ranges for a table (primary index)
 
+The following query only selects rows whose start_key does not have prefix as a substring. 
+
 {% include copy-clipboard.html %}
 ~~~ sql
 > SELECT * FROM [SHOW RANGES FROM TABLE vehicles] WHERE "start_key" NOT LIKE '%Prefix%';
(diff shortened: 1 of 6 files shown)


rmloveland:
Hi @chaneth8 !  thanks for your interest in contributing to the CockroachDB docs!

Unfortunately there is more nuance to #13696 than there appears to be at first glance.  We need to explain not only what the SQL is doing (which this PR does), but why the user needs to add the `WHERE` clause, which will likely involve some more explanation of why the output of the [`SHOW RANGES`](https://www.cockroachlabs.com/docs/stable/show-ranges.html) statement is what it is, and also some explanation of the way CockroachDB maps from SQL data to key-value storage.

(Good places to learn more about how CockroachDB maps SQL data to key-value storage are [in this older blog post](https://www.cockroachlabs.com/blog/sql-in-cockroachdb-mapping-table-data-to-key-value-storage/) and [this encoding tech note](https://github.com/cockroachdb/cockroach/blob/master/docs/tech-notes/encoding.md) and probably some other things I don't know about right now).

I think this PR as is will not meet the requirements of #13696, and the topic has enough low-level detail that I don't think we will be able to easily evolve this PR into what is needed, so I'm gonna close it out

That said - since you said you are just getting started contributing to open source documentation, we have a list of [good first issues](https://github.com/cockroachdb/docs/labels/good-first-issue) you may like to try and work on instead.  Please feel free to ask questions on any of those issues, and/or add me as a reviewer if you have a PR you think will fix one of them.  Thanks!

Issue #13826, opened 2022-05-06
inconsistency in docs for `cockroach node drain` CLI cmd


msbutler:
Michael Butler (msbutler) commented:

this doc page is not aligned with the CLI tool: https://www.cockroachlabs.com/docs/v21.2/cockroach-node#synopsis
cockroach node drain doesn’t take any arguments. But in the docs, it says that it takes node ID as an argument.
Docs:
`$ cockroach node drain <node ID> <flags>`
CLI:
```
$ cockroach node drain 1
ERROR: unknown command "1" for "cockroach node drain"
Failed running "node drain"
```

Jira Issue: DOC-3623

PR #13834, opened 2022-05-09
Removed arguments

chaneth8:
Related issue: https://github.com/cockroachdb/docs/issues/13826

### DIFF
--- a/v21.2/cockroach-node.md
+++ b/v21.2/cockroach-node.md
@@ -79,7 +79,7 @@ $ cockroach node recommission <node IDs> <flags>
 Drain nodes:
 
 ~~~ shell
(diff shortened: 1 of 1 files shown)


jseldess:
@taroface, would you please take a look at this change? If this command does, in fact, not accept node ID as an argument, we can probably move forward with this (and port the change to 22.1 docs), but we'll need to revisit other areas of our docs where `drain` is depicted with node ID as an argument, e.g., https://www.cockroachlabs.com/docs/v21.2/node-shutdown?filters=decommission#drain-a-node-manually.

taroface:
Thanks for the submission, @chaneth8. I've opened PR #14033 as a more complete fix.

PR #13835, opened 2022-05-09
Add docs on gateway node connection limits

rmloveland:
Fixes DOC-2582

Summary of changes:

- Add an include file that describes the new
  `server.max_connections_per_gateway` cluster setting, and the error
  code that results if the max connections are exceeded

- Reference the include from various related places in the documentation that
  mention connection pools, admission control, etc.  Specifically, the
  following pages:

  - Admission Control

  - Connection Pooling

  - CREATE DATABASE (specifically the CONNECTION LIMIT syntax)

  - Known Limitations (where memory monitoring is mentioned
    w.r.t. simultaneous client connections)

  - Recommended Production Settings (where connection pools are
    referenced)

  - UI SQL Dashboard (where the number of connections is mentioned)

- Bonus: Added an anchor link to a good definition of 'gateway node' in
  the SQL Layer page of the architecture docs, and link to it from the
  include where we mention the gateway node.


_includes/v22.1/sql/server-side-connection-limit.md
```diff
@@ -0,0 +1 @@
+<span class="version-tag">New in v22.1</span>: To control the maximum number of non-superuser ([`root`](security-reference/authorization.html#root-user) user or other [`admin` role](security-reference/authorization.html#admin-role)) connections a [gateway node](architecture/sql-layer.html#gateway-node) can have open at one time, use the `server.max_connections_per_gateway` [cluster setting](cluster-settings.html).  If a new non-superuser connection would exceed this limit, the error message `"sorry, too many connections"` is returned, along with error code `53300`.
```
ecwall, line 1: `"sorry, too many connections"` -> `"sorry, too many clients already"`

ecwall (approved):
Just 1 small change.

_includes/v22.1/sql/server-side-connection-limit.md
rmloveland: Fixed - thanks @ecwall !

ericharmeling (approved):
Looks good!

PR #13838, opened 2022-05-10
Fix number of PRs cited as fixed in the release.

stbof:


stbof:
I guess it's never 1-1 (link to PR). My bad.

PR #13844, opened 2022-05-12
Improve index best practices.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2801

Best practices are updated here: https://deploy-preview-13844--cockroachdb-docs.netlify.app/docs/dev/schema-design-indexes.html#best-practices

Other fixes are to correct statement output or to conform to style guidelines.


rytaft (approved):
Just one issue below. Otherwise, :lgtm:

> Reviewed 1 of 3 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13844#-:-N2K-oaf91ffR5s4yDrd:bnb1wgx)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

___
*[`v22.1/schema-design-indexes.md` line 89 at r2](https://reviewable.io/reviews/cockroachdb/docs/13844#-N2Jyvf1EOcPhDKjUFoU:-N2Jyvf1EOcPhDKjUFoV:bxmps60) ([raw file](https://github.com/cockroachdb/docs/blob/40dcba58b68bc0607a8294f9044c27fcfa3e71e8/v22.1/schema-design-indexes.md#L89)):*
> ```Markdown
> - Use a [`STORING` clause](create-index.html#parameters) to store columns of data that you want returned by common queries, but that you do not plan to use in query filters.
> 
>     The `STORING` clause specifies columns that are not part of the index key but should be stored in the index. If a column is specified in a query, and the column is neither indexed nor stored in an index, CockroachDB will perform a full scan of the table, which can result in poor performance. For an example, see [Example](#example).
> ```

This isn't quite right. We may perform a full scan in this case, but we may perform an index join instead if a suitable secondary index exists. However, if the optimizer determines that the index join is too expensive, then we will perform a full table scan.
___
*[`v22.1/schema-design-indexes.md` line 91 at r2](https://reviewable.io/reviews/cockroachdb/docs/13844#-N2JzOsB17OU-iPhx2LG:-N2JzOsB17OU-iPhx2LH:b3mek5t) ([raw file](https://github.com/cockroachdb/docs/blob/40dcba58b68bc0607a8294f9044c27fcfa3e71e8/v22.1/schema-design-indexes.md#L91)):*
> ```Markdown
>     The `STORING` clause specifies columns that are not part of the index key but should be stored in the index. If a column is specified in a query, and the column is neither indexed nor stored in an index, CockroachDB will perform a full scan of the table, which can result in poor performance. For an example, see [Example](#example).
> 
> - Review the [specialized indexes](schema-design-overview.html#specialized-indexes) and decide if you need to create a specialized index instead of a standard index.
> ```

nit: can you give a hint here about what specialized indexes are?

rytaft (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13844#-:-N2NeGHTDS-4FcxK8o-1:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @rmloveland and @stbof)

___
*[`v22.1/schema-design-indexes.md` line 89 at r3](https://reviewable.io/reviews/cockroachdb/docs/13844#-N2MgM-FA8fQcXCl4YgQ:-N2MeInO6S5_KHXFjH_2:bveym2k) ([raw file](https://github.com/cockroachdb/docs/blob/a099b4b3c64001f59fa53cd1ffae33295d11644e/v22.1/schema-design-indexes.md#L89)):*
> ```Markdown
> - Use a [`STORING` clause](create-index.html#parameters) to store columns of data that you want returned by common queries, but that you do not plan to use in query filters.
> 
>     The `STORING` clause specifies columns that are not part of the index key but should be stored in the index. If a column is specified in a query, and the column is neither indexed nor stored in an index, CockroachDB may perform a full scan in this case, but we may perform an index join instead if a suitable secondary index exists. However, if the optimizer determines that the index join is too expensive, then CockroachDB will perform a full table scan, which can result in poor performance. For an example, see [Example](#example).
> ```

nit: suggest rewriting as:
If a column is specified in a query, and the column is neither indexed nor stored in an index, CockroachDB may either perform a full scan or perform an index join if a suitable secondary index exists. However, if the optimizer determines that the index join is too expensive, then CockroachDB will perform a full table scan.

rmloveland (approved):
LGTM

PR #13845, opened 2022-05-12
Update developer pages to suggest Serverless instead of Dedicated

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-3288

### DIFF
--- a/v21.2/bulk-delete-data.md
+++ b/v21.2/bulk-delete-data.md
@@ -22,7 +22,7 @@ Exercise caution when batch deleting rows from tables with foreign key constrain
 Before reading this page, do the following:
 
 - [Install CockroachDB](install-cockroachdb.html).
-- [Start a local cluster](secure-a-cluster.html), or [create a {{ site.data.products.dedicated }} cluster](../cockroachcloud/create-your-cluster.html).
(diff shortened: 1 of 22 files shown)


v21.2/bulk-delete-data.md
```diff
@@ -22,7 +22,7 @@ Exercise caution when batch deleting rows from tables with foreign key constrain
 Before reading this page, do the following:
 
 - [Install CockroachDB](install-cockroachdb.html).
-- [Start a local cluster](secure-a-cluster.html), or [create a {{ site.data.products.dedicated }} cluster](../cockroachcloud/create-your-cluster.html).
+- [Start a local cluster](secure-a-cluster.html) or [create a {{ site.data.products.serverless }} cluster](../cockroachcloud/create-a-serverless-cluster.html).
```
jseldess, line 25: Instead of using these, I'd suggest we use the 2 quickstart variations.

Start a local cluster: https://www.cockroachlabs.com/docs/cockroachcloud/quickstart.html?filters=local
Start a cloud cluster: https://www.cockroachlabs.com/docs/cockroachcloud/quickstart.html

I'd probably switch the order and start with cloud. Also, I think we can remove the "Install CockroachDB" link. If folks choose the local cluster option, they'll get there.

jseldess (changes requested):

PR #13849, opened 2022-05-13
[DOC-3429] Fix ordered list numbering in filtered content

mdlinville:
DOC-3429
See the duplicated first bullet at https://www.cockroachlabs.com/docs/v21.2/orchestrate-cockroachdb-with-kubernetes-multi-cluster.html#step-3-use-the-built-in-sql-client.


mdlinville:
@ianjevans I'm not sure I'm fixing this correctly. I don't understand how the sections work. I'm puzzled that in the preview with these changes applied, there is only one list item showing, rather than the 5 I'd expect. Can you help me spot what I'm missing?

ianjevans (approved):

Issue #13788, opened 2022-05-03
Recommending setting a leasing preference for the system database when setting up multi-region


fqazi:
For multi-region databases, we should in the documentation recommend setting a leasing preference for the system database since it can lead to more consistent performance for schema changes. Additionally, we should recommend running schema changes from whatever region that has the leasing preference set since it can lead to better performance. This change in general is beneficial for multi-region users.

For example:
`alter database system configure zone using lease_preferences = '[[+region=us-east]]';`

PR #13851, opened 2022-05-14
Added Recommendation

chaneth8:
Related issue: https://github.com/cockroachdb/docs/issues/13788


jseldess:
@chaneth8, thank you for working on this issue. Unfortunately, this is a rather complex part of our product, and so the content we add needs to go beyond the simple recommendation you've added. I believe we need to provide an example of how to set a lease preference for the system database. But beyond that, we need to explain why this is important to do, and find the best place for those instructions to be provided. 

All that considered, I think this doc work is better suited for a writer at Cockroach Labs, so I'm going to close this PR and make sure it's on the radar of the appropriate writer. Thank you, nonetheless, for taking time and effort to improve our docs!

PR #13869, opened 2022-05-17
Reverting long enum names in Cloud API.

ianjevans:


erademacher (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13869)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @ianjevans and @marksoper)

___
*[`cockroachcloud/cloud-api.md` line 229 at r1](https://reviewable.io/reviews/cockroachdb/docs/13869#-N2Hrpw_6E2aYH7IyFG5:-N2Hrpw_6E2aYH7IyFG6:b-8fl3t) ([raw file](https://github.com/cockroachdb/docs/blob/8f347b699afbe442b2906f0108b829ed0e53273d/cockroachcloud/cloud-api.md#L229)):*
> ```Markdown
>   {{site.data.alerts.end}}
>   - `{cluster_name}` is the name of the cluster you specified when creating the cluster.
>   - `{cloud_provider}` is the name of the cloud infrastructure provider on which you want your cluster to run. Possible values are: `GCP` and `AWS`. The default value is `UNSPECIFIED`.
> ```

Do we have to give a default value here?  This is a required field and `UNSPECIFIED` is not actually a valid option. I'm actually working on a change that will remove all the 'UNSPECIFIED' zero values from the OpenAPI output (effectively hiding them).

ianjevans:
> Do we have to give a default value here? This is a required field and `UNSPECIFIED` is not actually a valid option. I'm actually working on a change that will remove all the 'UNSPECIFIED' zero values from the OpenAPI output (effectively hiding them).

We don't, no. I'll remove it.

erademacher (approved):
:lgtm: Thanks Ian!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13869#-:-N2Huc0r1ECqoZCiSe2H:b-5qyq5)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @marksoper)

PR #13873, opened 2022-05-17
Added more detail to GC warning on incremental backups

kathancox:
Fixes [DOC-2410](https://cockroachlabs.atlassian.net/browse/DOC-2410)

The error mentioned in the issue is already publicly documented — to provide more visibility have added a link out to the errors page from the incremental backup section v21.1+


livlobo (approved):
LGTM!

kathancox:
@andf-crl No that's a good point, have added! Thanks for the review!

PR #13904, opened 2022-05-20
DOC-351: Add height restriction and scrolling on `pre` blocks

nickvigilante:
Fixes DOC-351

### DIFF
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -7554,10 +7554,10 @@ section.footer {
     .bs-callout--info h4, .bs-callout--info .bs-callout__label {
       color: #190F33; }
   .bs-callout--version {
-    background-color: rgba(105, 51, 255, 0.2);
-    border-left-color: #6933FF }
+    background-color: #FF7595;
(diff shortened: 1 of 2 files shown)


jseldess (changes requested):
Thanks for taking this on, @nickvigilante. The UX still needs some work.

When it's a block without a copy button, the scrollbar is position well, but it doesn't show up unless you hover over it, which is a bit tricky.

When it's a block with a copy button, the scrollbar position is somewhat odd. See screenshot:
Screen Shot 2022-05-25 at 2 20 37 PM

nickvigilante:
@jseldess I was a bit overzealous here, thinking one line would fix this 😄 

I'm looking at the code of Confluent's website, and I see that they have their `<pre>` block inside of another `<div>` block, and that `<div>` block also contains some `<span>` blocks containing the copy/expand/collapse buttons. I'm gonna have to look into this more. For now, I'm closing this PR to do some more local testing, but I'll keep the JIRA open and continue the conversation there.

PR #13907, opened 2022-05-20
DOC-3375: Improve usability of TAs

nickvigilante:
Fixes DOC-3375


jseldess (changes requested):
This looks great overall, but on the index, the advisories no longer seems sorted in reverse chronologically order by date, like on the live page. I think it's best to retain that.

nickvigilante:
@jseldess Good catch. I've fixed that now.

jseldess:
There's a conflict to resolve, though, before merging.

PR #13622, opened 2022-04-15
PKI/TLS improvements

MichaelTrestman:
This PR creates:
A tutorial on managing certs with Vault: https://deploy-preview-13622--cockroachdb-docs.netlify.app/docs/v22.1/manage-certs-vault.html
An overview page on TLS/PKI security in CRDB: https://deploy-preview-13622--cockroachdb-docs.netlify.app/docs/v22.1/security-reference/transport-layer-security.html
It also reorganizes and refines existing documentation on certs throughout the docs.

https://cockroachlabs.atlassian.net/browse/DOC-428
https://cockroachlabs.atlassian.net/browse/DOC-3194
https://cockroachlabs.atlassian.net/browse/DOC-3193
https://cockroachlabs.atlassian.net/browse/DOC-2717

PR #13910, opened 2022-05-23
temp remove --insecure deprecation notice

taroface:
Fixes DOC-2807.

The docs have been running an outdated statement that `--insecure` "will be deprecated in v21.1" and replaced with other flags. Temporarily removing this statement until any alternatives can be documented.


MichaelTrestman (approved):
Unfortunate state of affairs, but LGTM given lack of cogent alternative.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13910#-:-N3K_IeK7Ab7z-qq_DBY:brp0qrz)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @taroface)

taroface:
> I have a blurb clarifying the difference between the SSL modes in an upcoming PR. If you have bandwidth and the inclination, you could add the beefed up description somewhere.
> 
> https://deploy-preview-13622--cockroachdb-docs.netlify.app/docs/v22.1/security-reference/transport-layer-security.html#cockroachdbs-tls-support-and-operating-modes
> 
> #13622

@MichaelTrestman TFTR! This actually made me realize that `--accept-sql-without-tls` is already documented in the `cockroach-start` and `cockroach-start-single-node` docs. Is that flag intended to be the alternative to `--insecure`? If so, I could try to clarify the intended usage in this PR. But I could use more context, in case you know of any other issues that reference how these 2 flags relate to each other.

PR #13952, opened 2022-05-26
Updates for versioned Cloud API.

ianjevans:
Closes DOC-2194.

This PR uses Rapidoc for our API reference docs with some light theming to match the current Cockroach Lab style. It adds documentation on setting a version header.


stbof:
It looks a lot better. Is there a reason why the line widths of these buttons are
Screen Shot 2022-07-18 at 9 45 06 AM
 different.

taroface:
Agree that this is looking much more readable! I have a few style comments:

Perhaps the example request/response blocks, code samples, and parameter fields could all have the same formatting (i.e., grey background without border) for visual simplicity:

image

One thing that's been confusing me about the default template is the vertical spacing between elements. Something like this would be easier to scan IMO:

image

This example is adding:
- `line-height: normal` to `h2`
- `margin-top: 15px` to `.req-res-title`

And removing:
- `margin-top: 24px` from `.top-gap`
- `margin-bottom: 8px` from `.req-res-title`

ianjevans:
> Perhaps the example request/response blocks, code samples, and parameter fields could all have the same formatting (i.e., grey background without border) for visual simplicity:

I think the reason the Request body block has a white background is to visually distinguish it as an editable pane, for customizing the request.

I've added the rest of the overrides.

ianjevans:
> I think the response buttons would look better with a narrower line.

Done.

api/cloud/v1.md
```diff
@@ -0,0 +1,116 @@
+---
+title: CockroachDB Cloud API
+summary: Learn about the CockroachDB Cloud API
+layout: api
+docs_area: reference.api
+---
+
+{% assign releases = site.data.cloud_api_releases %}
+
+<style>
```
natelong, line 10: Nit: Could we pull this CSS out into a separate file?

api/cloud/v1.md
```diff
@@ -0,0 +1,116 @@
...
+
+    if (typeof selectedVersion !== 'undefined') {
+      doc.loadSpec('https://cockroachlabs.cloud/assets/docs/ccapi_spec_' + selectedVersion + '.json');
+    } else {
+      doc.loadSpec('https://cockroachlabs.cloud/assets/docs/ccapi_spec_latest.json');
+    }
+  }
+</script>
+<div class="apidocs">
+  {% comment %}Replace with remote URL once it is complete{% endcomment %}
```
natelong, line 88: Nit: Is this comment out of date?

api/cloud/v1.md
```diff
@@ -0,0 +1,116 @@
...
+            mono-font = "SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace"
+            show-method-in-nav-bar = "as-colored-text"
+            allow-server-selection = "false"
+            use-path-in-nav-bar = "false"
+            schema-description-expanded = "true"
+            allow-spec-file-download = "true">
+    <div slot="nav-logo" style="display: flex; align-items: center; justify-content: center;"> 
+      <img src = "https://d33wubrfki0l68.cloudfront.net/1c17b3053b29646cdddc53965186a02179b59842/69991/docs/images/cockroachlabs-logo-170.png" style="width:170px; margin-right: 20px" alt="Cockroach Labs logo"> <span style="color:#fff"> <b>nav-logo</b> slot </span>
+    </div>
+    <div slot="overview">
```
natelong, line 106: I love the version switcher, but I feel like it should be higher precedence on the page. I'm imagining having it in the sidebar so that it's always visible, because it's so important that people don't get the wrong version.

cockroachcloud/cloud-api.md
```diff
@@ -40,6 +40,37 @@ Authorization: Bearer {secret_key}
 
 Where `{secret_key}` is the [secret key string you stored when you created the API key in the Console](console-access-management.html#create-api-keys).
 
+## Set the API version
+
+The `Cc-Version` HTTP header specifies the version of the Cloud API to use. If you omit the `Cc-Version` header, the Cloud API will default to version `2022-03-31` (the initial release of the Cloud API). The Cloud API uses date-based versions of the form `YYYY-MM-DD`, in [ISO 8601 format](https://www.w3.org/TR/NOTE-datetime).
```
natelong, line 45: @marksoper Should we mention here that this behavior is temporary? I think we might want this message to be from the perspective that you should always be using the `cc-version` header, but for backwards compatibility reasons, we will support omitting it for a while (I think we agreed on 3 months?).

Maybe something like this:
> The `Cc-Version` HTTP header specifies the version of the Cloud API to use. The Cloud API uses date-based versions of the form `YYYY-MM-DD`, in [ISO 8601 format](https://www.w3.org/TR/NOTE-datetime).
>
> Note: Until January 1, 2023, omitting the `Cc-Version` header will default to version `2022-03-31` (the initial release of the Cloud API). After January 1, 2023, omitting the `Cc-Version` header will return an HTTP 400 response with the message "Missing Cc-Version".

natelong (approved):
Added some comments from my perspective, but I wouldn't block this PR on them--these are good changes. Thank you!

api/cloud/v1.md
ianjevans: Unfortunately no. I'm not clear why, but adding these CSS elements to our main CSS files ended up getting overriden by RapiDoc.

cockroachcloud/cloud-api.md
ianjevans: In general, we avoid documenting potential future changes in the docs. The docs describe how the product works at the present. If we decide to change how we implement version pinning, we've already set expectations among our customers, and caused unnecessary confusion.

ianjevans: If we implemented a deprecation warning when users omit the version header, we would add a corresponding warning in the docs.

natelong: If that's a hard line, then I would just leave out the part about what happens when you omit the version header. At the time that these docs are going up, the official stance is that you need to include the version header. We're going to have a default temporarily for backwards compatibility reasons, but IMO that's something that only existing users should know about.

gemma-shay:
just a nit but can we change CockroachCloud to CockroachDB Cloud?

ianjevans:
> just a nit but can we change CockroachCloud to CockroachDB Cloud?

Unfortunately, we can't. It's derived from the [service name in the code](https://github.com/cockroachlabs/managed-service/blob/master/console/apipb/api.proto#L1650). I think Jesse had asked @erademacher if this could be changed previously, but I don't remember what came out of this.

ianjevans:
> Confused by the "Fill Example" button behavior. Under fields with no example, it just does nothing. Can we remove or gray out in these cases? (e.g. https://deploy-preview-13952--cockroachdb-docs.netlify.app/docs/api/cloud/v1.html#get-/api/v1/clusters)

That's controlled by RapiDoc, and I don't see an option to disable it for those cases.

gemma-shay:
> > Confused by the "Fill Example" button behavior. Under fields with no example, it just does nothing. Can we remove or gray out in these cases? (e.g. https://deploy-preview-13952--cockroachdb-docs.netlify.app/docs/api/cloud/v1.html#get-/api/v1/clusters)
> 
> That's controlled by RapiDoc, and I don't see an option to disable it for those cases.

Ah, well. LGTM otherwise!

PR #13958, opened 2022-05-26
Add link to index recommendations from best practices.

stbof:
Needs the update in https://github.com/cockroachdb/docs/pull/13999 that adds index recommendations to the plan properties for the link to make sense.


mgartner (approved):
LGTM

PR #13961, opened 2022-05-26
Add special character escaping

andf-crl:
- Add _Escaping special characters_ section covering Jekyll and Unicode situations.
- Linked from _Code block_ and _Placeholders_


taroface:
LGTM! Thank you!

PR #13965, opened 2022-05-26
Fix typo.

stbof:


StyleGuide.md
```diff
@@ -736,7 +736,7 @@ For some examples, see [Connect to a CockroachDB Cluster](https://www.cockroachl
 
 #### Escaping special characters
```
stbof, line 737:
```suggestion
#### Escape special characters
```

stbof: Use imperative instead of gerund.

andf-crl (commented):
Thank you for `achieve`.

However:

Your heading change breaks two links and changes a consideration into an always-true guidance, which it must not be. Escaping special characters in, say, a monospace link would render the Unicode as-is ([example](https://github.com/cockroachdb/docs/blob/9e1582ebf1f5bd662ef966db8dcb81bdaa621e69/v22.1/alter-role.md?plain=1#L41-L42)). I recommend we keep the header as is: the phrasing of the two now-broken links suggests the use cases where this heading is appropriate, as does the use of the word `may` just underneath.

andf-crl:
"How to" sound good, thanks!

stbof:
> "How to" sound good, thanks!

Do you want me to update the heading and fix the links?

andf-crl:
Either: as-proposed or as-is is sufficient for me! Happy to approve either choice you make.

stbof:
> Either: as-proposed or as-is is sufficient for me! Happy to approve either choice you make.

Done.

andf-crl (approved):
LGTM, thank you!

PR #13972, opened 2022-05-27
Latest patch rollback clarification

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-3562

### DIFF
--- a/cockroachcloud/upgrade-policy.md
+++ b/cockroachcloud/upgrade-policy.md
@@ -31,7 +31,7 @@ Major version upgrades are automatic for {{ site.data.products.serverless }} clu
 
 ### Rollback support
 
-When upgrading a {{ site.data.products.dedicated }} cluster to a new major version, once all nodes are running the new version, you have approximately 72 hours before the upgrade is automatically finalized. During this window, if you see unexpected behavior, you can [trigger a rollback to the previous major version](upgrade-to-v21.2.html#roll-back-the-upgrade) from the {{ site.data.products.db }} Console.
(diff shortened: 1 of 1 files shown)


cockroachcloud/upgrade-policy.md
```diff
@@ -31,7 +31,7 @@ Major version upgrades are automatic for {{ site.data.products.serverless }} clu
 
 ### Rollback support
 
-When upgrading a {{ site.data.products.dedicated }} cluster to a new major version, once all nodes are running the new version, you have approximately 72 hours before the upgrade is automatically finalized. During this window, if you see unexpected behavior, you can [trigger a rollback to the previous major version](upgrade-to-v21.2.html#roll-back-the-upgrade) from the {{ site.data.products.db }} Console.
+When upgrading a {{ site.data.products.dedicated }} cluster to a new major version, once all nodes are running the new version, you have approximately 72 hours before the upgrade is automatically finalized. During this window, if you see unexpected behavior, you can [trigger a rollback to the previous major version](upgrade-to-v21.2.html#roll-back-the-upgrade) from the {{ site.data.products.db }} Console. Note that you will roll back to the latest patch version of the previous major version, not necessarily the same patch version you were running before you initiated the upgrade.
```
rmloveland, line 34: I would remove "necessarily" so this just says "..., not the same patch version" since the "necessarily" introduces some ambiguity

May be worth considering making this its own separate paragraph to make sure folks see it, since I could imagine some users might find this behavior surprising

rmloveland (approved):
LGTM

PR #13973, opened 2022-05-27
Fix issue with new-in includes in MD tables

nickvigilante:
Continuation of fix for DOC-3423


stbof (approved):
Spot checked 3 tables. LGTM. Sorry I missed it before.

nickvigilante:
@stbof No need to apologize. I think I'm equally responsible to make sure it's working as well 😄

PR #13989, opened 2022-05-31
Clarified CRDB does not support incremental restores in docs

kathancox:
Fixes [DOC-3404](https://cockroachlabs.atlassian.net/browse/DOC-3404)

Short sentence and note accompanying incremental restore examples & restore type table.


shermanCRL (approved):

dt (approved):

_includes/v22.1/backups/no-incremental-restore.md
```diff
@@ -0,0 +1 @@
+When you restore from an incremental backup, you're restoring from the latest full backup and latest (or [specific](restore.html#restore-a-specific-backup)) incremental backup. You cannot restore an incremental backup without a full backup. See [Restore types](restore.html#restore-types) for detail on the types of backups you can restore.
```
stevendanna, line 1: Looking at the original discussion, I am not sure this would have cleared up the confusion.  I wonder if something like

> When you restore from an incremental backup, the _entire_ table, database, or cluster is restored. Both the latest incremental backup and the full backup are used during this process. Further, you cannot restore over a table, database, or cluster with existing data.

It's still a bit indirect in that someone who expects an "incremental restore" to exist will need to infer that it doesn't from those facts.

kathancox: I see what you mean on perhaps not 100% covering the original conversation. We do cover that you can't restore over existing data in another section, but I have adapted to include that here as well and to link out. Lmk what you think of the changes @stevendanna thanks for the feedback!

stevendanna (approved):

_includes/v22.1/backups/no-incremental-restore.md
stevendanna: Looks good to me!

andf-crl (approved):
LGTM! Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13989#-:-N3QJvwR9x7j_sC6VKUQ:b3is8u8)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @kathancox and @stevendanna)

PR #13993, opened 2022-05-31
DOC-3448: Add tip on improving schema change latency for MR

nickvigilante:
Fixes DOC-3448

### DIFF
--- a/_includes/v22.1/performance/lease-preference-system-database.md
+++ b/_includes/v22.1/performance/lease-preference-system-database.md
@@ -1,0 +1,8 @@
+To reduce latency while making {% if page.name == "online-schema-changes.md" %}online schema changes{% else %}[online schema changes](online-schema-changes.html){% endif %}, we recommend specifying a `lease_preference` [zone configuration](configure-replication-zones.html) on the `system` database to a single region and running all subsequent schema changes from a node within that region. For example, if the majority of online schema changes come from machines that are geographically close to `us-east1`, run the following:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE system CONFIGURE ZONE USING constraints = '{"+region=us-east1": 1}', lease_preferences = '[[+region=us-east1]]';
+~~~
+
+Run all subsequent schema changes from a node in the specified region.
--- a/netlify/vale/vocab.txt
+++ b/netlify/vale/vocab.txt
@@ -1,3 +1,4 @@
+abovementioned
 absentia
 accipitridae
 addons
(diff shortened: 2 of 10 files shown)


ajwerner (approved):
I'm cool with this, thanks for adding something. Is it worth adding any more explanations? 

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/13993#-:-N3yMYiF8n_XH4mEA_yo:b-9bgv58)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @nickvigilante)

nickvigilante:
@ajwerner I didn't think so at the time, unless there was a page I missed in my search. I checked with @rmloveland on this, who also agreed on those same pages.

Rich, can I get your review on this while Stephanie is out?

_includes/v22.1/performance/lease-preference-system-database.md
```diff
@@ -0,0 +1,8 @@
+To reduce latency while making online schema changes, we recommend specifying a `lease_preference` on the `system` database to a single region and running all subsequent schema changes from a node within that region. For example, if the majority of online schema changes come from machines that are geographically close to `us-east1`, run the following:
```
rmloveland, line 1: This looks really good! Two small things:

First, suggest making first mention of 'online schema changes' a link to https://www.cockroachlabs.com/docs/stable/online-schema-changes.html (appropriate to the docs version of course)

Since this is also going on the schema changes page, you can use the Liquid syntax like the below to make it not a link on that page  (ex is taken from https://github.com/cockroachdb/docs/commit/f4cc14a2013bdd8baf3a08feed883fe516a7f7c7)

```
{% if page.name == "multiregion-overview.md" %} using the higher-level multiregion abstractions, you must use Super Regions {% else %} using the higher-level [multiregion abstractions](multiregion-overview.html), you must use [Super Regions](multiregion-overview.html#super-regions) {% endif %}
```

Second, suggest adding a link to https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html somewhere in your mention of `lease_preference`, e.g. "we recommend specifying the `lease_preference` [zone configuration](configure-replication-zones.html) ..."

_includes/v22.1/performance/lease-preference-system-database.md
```diff
@@ -0,0 +1,8 @@
...
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+ALTER DATABASE system CONFIGURE ZONE USING constraints = '{"+region=us-east1": 1}', lease_preferences = '[[+region=us-east1]]';
+~~~
+
+After running this successfully, run all subsequent schema changes from a node existing in the abovementioned region.
```
rmloveland, line 8: nit: could be simplified a bit as something like "run all subsequent ... from a node in the specified region"

rmloveland (approved):
LGTM, had a couple little comments - thanks Nick, this is awesome!

netlify/vale/vocab.txt
```diff
@@ -1,3 +1,4 @@
+abovementioned
```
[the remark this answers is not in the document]
kathancox: Thankkk you, I didn't know that. I will do that now and commit!

PR #14009, opened 2022-06-02
Added `end_time` option to changefeed docs

kathancox:
Fixes [DOC-2984](https://cockroachlabs.atlassian.net/browse/DOC-2984)

Added the `end_time` option to `CREATE CHANGEFEED` and `EXPERIMENTAL CHANGEFEED FOR` documentation.


miretskiy (approved):

v22.1/changefeed-for.md
```diff
@@ -54,6 +54,7 @@ Option | Value | Description
...
 `envelope` | `key_only` / `row` | Use `key_only` to emit only the key and no value, which is faster if you only want to know when the key changes.<br><br>Default: `envelope=row`
 `cursor` | [Timestamp](as-of-system-time.html#parameters)  | Emits any changes after the given timestamp, but does not output the current state of the table first. If `cursor` is not specified, the changefeed starts by doing a consistent scan of all the watched rows and emits the current value, then moves to emitting any changes that happen after the scan.<br><br>`cursor` can be used to start a new changefeed where a previous changefeed ended.<br><br>Example: `CURSOR=1536242855577149065.0000000000`
+`end_time` | [Timestamp](as-of-system-time.html#parameters) | <span class="version-tag">New in v22.1:</span> Indicate the time when a changefeed should stop. Provide a future timestamp to `end_time` in number of nanoseconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time). For example, `end_time="1654117200000000"`. At the specified time, the changefeed will stop running and return a `successful` status.
```
samiskin, line 57: It could be better to word this as something more like
> Indicate the time for which the changefeed will emit all events up to the given timestamp and then complete with a `successful` status.

Since the changefeed job itself could potentially still be running after the `end_time` passes if the timestamp of the actual events its emitting are still lagging behind.

samiskin: Would "Indicate the timestamp" rather than "a timestamp" be fine?  Very minor but personally "the" gives me more confidence that this is the event timestamp, since it must be referring to the only timestamp we've ever talked about so far, while "a" kind of reminds me that this could have multiple meanings and I feel a little less certain.

v22.1/changefeed-for.md
```diff
@@ -54,6 +54,7 @@ Option | Value | Description
...
 `envelope` | `key_only` / `row` | Use `key_only` to emit only the key and no value, which is faster if you only want to know when the key changes.<br><br>Default: `envelope=row`
 `cursor` | [Timestamp](as-of-system-time.html#parameters)  | Emits any changes after the given timestamp, but does not output the current state of the table first. If `cursor` is not specified, the changefeed starts by doing a consistent scan of all the watched rows and emits the current value, then moves to emitting any changes that happen after the scan.<br><br>`cursor` can be used to start a new changefeed where a previous changefeed ended.<br><br>Example: `CURSOR=1536242855577149065.0000000000`
+`end_time` | [Timestamp](as-of-system-time.html#parameters) | <span class="version-tag">New in v22.1:</span> Indicate a timestamp up to which the changefeed will emit all events and then complete with a `successful` status. Provide a future timestamp to `end_time` in number of nanoseconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time). For example, `end_time="1654117200000000"`.
```
samiskin, line 57: Would "Indicate the timestamp" rather than "a timestamp" be fine? Very minor but personally "the" gives me more confidence that this is the event timestamp, since it must be referring to the only timestamp we've ever talked about so far, while "a" kind of reminds me that this could have multiple meanings and I feel a little less certain.

kathancox: Sure Shiranka, I understand. Added!

andf-crl (approved):
LGTM with one question for engineering, and one small nit on the epoch notation.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14009#-:-N4hOyPT0zVrGoU861XC:bst5daj)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl, @kathancox, and @samiskin)

___
*[`v22.1/changefeed-for.md` line 57 at r2](https://reviewable.io/reviews/cockroachdb/docs/14009#-N3ZgQTJo1WCjMkPhP9G-r2-57:-N4hLkZb7ebkZEhLGHSU:bq636rj) ([raw file](https://github.com/cockroachdb/docs/blob/caf626bf1fbeceab278fe22529880958e8bd6e3f/v22.1/changefeed-for.md#L57)):*

Sure Shiranka, I understand. Added!
</blockquote></details>

nit: I believe the nanosecond version of your unix epoch timestamp would be `1654117200000000000`. `1654117200000000` I believe is notation in milliseconds (for `Wed Jun 01 2022 21:00:00 GMT+0000`).
___
*[`v22.1/create-changefeed.md` line 127 at r3](https://reviewable.io/reviews/cockroachdb/docs/14009#-N4hQ5Tw6MFQDUNfRvh3:-N4hQ5Tw6MFQDUNfRvh4:bh8zieb) ([raw file](https://github.com/cockroachdb/docs/blob/02abe8b9beedb6bfe0a37b05609395c7bd5b23f6/v22.1/create-changefeed.md#L127)):*
> ```Markdown
> `envelope` | `key_only` / `wrapped` | Use `key_only` to emit only the key and no value, which is faster if you only want to know when the key changes.<br><br>Default: `envelope=wrapped`
> <a name="cursor-option"></a>`cursor` | [Timestamp](as-of-system-time.html#parameters)  | Emit any changes after the given timestamp, but does not output the current state of the table first. If `cursor` is not specified, the changefeed starts by doing an initial scan of all the watched rows and emits the current value, then moves to emitting any changes that happen after the scan.<br><br>When starting a changefeed at a specific `cursor`, the `cursor` cannot be before the configured garbage collection window (see [`gc.ttlseconds`](configure-replication-zones.html#replication-zone-variables)) for the table you're trying to follow; otherwise, the changefeed will error. With default garbage collection settings, this means you cannot create a changefeed that starts more than 25 hours in the past.<br><br>`cursor` can be used to [start a new changefeed where a previous changefeed ended.](#start-a-new-changefeed-where-another-ended)<br><br>Example: `CURSOR='1536242855577149065.0000000000'`
> `end_time` | [Timestamp](as-of-system-time.html#parameters) | <span class="version-tag">New in v22.1:</span> Indicate the timestamp up to which the changefeed will emit all events and then complete with a `successful` status. Provide a future timestamp to `end_time` in number of nanoseconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time). For example, `end_time="1654117200000000"`.
> ```

Question: upstream code suggests that `end_time` cannot be used along with `no_initial_scan` (this appears to be [tested for explicitly](https://github.com/sherman-grewal/cockroach/blob/d0fd09c396b9e9dd7361e412fb3720d313b5ddc1/pkg/ccl/changefeedccl/changefeed_test.go#L3766-L3774)). Should this be included in one (or both) of these descriptions?

kathancox:
@andf-crl Do you mean `initial_scan_only`? If so, yes, I missed that. I have another PR for the unification of the `initial_scan` syntax that can't move forward until a bug is fixed, so in my head I had lumped that into that PR forgetting this would come first. I will add that in

andf-crl:
Yes: `initial_scan_only` in the code and `no_initial_scan` in the [docs](https://deploy-preview-14009--cockroachdb-docs.netlify.app/docs/v22.1/create-changefeed.html#options). Sounds good!

kathancox:
> Yes: `initial_scan_only` in the code and `no_initial_scan` in the [docs](https://deploy-preview-14009--cockroachdb-docs.netlify.app/docs/v22.1/create-changefeed.html#options). Sounds good!

Ahhh yes, `initial_scan = only` I was holding on that bug. `initial_scan_only`/`no_initial_scan` these are two different things. (The former had not been added to the docs before the unification of the `initial_scan` values had been worked on. Hence `initial_scan_only` not being present in public docs yet.) I have added the following:

- `initial_scan_only` (soon to be replaced with `initial_scan = yes|no|only`
- updated the timestamp 
- Added missing options from the core page and the general updates there

PR #14026, opened 2022-06-06
Fix redirects for old RN URLs without .html suffix

nickvigilante:
Formerly, anyone trying to follow a link like /docs/releases/v21.1.5 would be given a 404. This should resolve that.


nickvigilante:
@jseldess I think this change did the trick:

https://deploy-preview-14026--cockroachdb-docs.netlify.app/docs/releases/v21.1.15
https://deploy-preview-14026--cockroachdb-docs.netlify.app/docs/releases/v21.1.15.html

Both of these redirect to the correct spot in the docs.

jseldess (approved):
Just tested. LGTM!

PR #14027, opened 2022-06-06
Added missing kafka config fields & rearranged section format

kathancox:
Fixes [DOC-2687](https://cockroachlabs.atlassian.net/browse/DOC-2687) and [DOC-3625](https://cockroachlabs.atlassian.net/browse/DOC-3625)

Added missing configurations options to `kafka_sink_config` (`Bytes` and `Messages`). Also rearranged the format of the section to match the same in the webhook configuration.

### DIFF
--- a/v21.2/changefeed-sinks.md
+++ b/v21.2/changefeed-sinks.md
@@ -74,7 +74,7 @@ The configurable fields are as follows:
 Field              | Type                | Description      | Default
 -------------------+---------------------+------------------+-------------------
 `Flush.MaxMessages` | [`INT`](int.html)  | Sets the maximum number of messages the producer can send in a single broker request. Any messages beyond the configured limit will be blocked. Increasing this value allows all messages to be sent in a batch. | `1`
-`Flush.Messages`   | [`INT`](int.html)   | Determines when to send a batch of messages based on the configured the number of messages. | `0`
+`Flush.Messages`   | [`INT`](int.html)   | Configure the number of messages the changefeed should batch before flushing. | `0`
 `Flush.Bytes`      | [`INT`](int.html)   | When the total byte size of all the messages in the batch reaches this amount, it should be flushed. | `0`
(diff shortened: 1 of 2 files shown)


kathancox:
Preview to page: https://deploy-preview-14027--cockroachdb-docs.netlify.app/docs/stable/changefeed-sinks.html#kafka-sink-configuration

kathancox:
> LGTM, although I think the distinction between MaxMessages and Messages isn't clear in the table right now

Yes, I agree — the distinction needs to be stronger somehow. I added some small tweaks. Appreciate any feedback/ideas on distinguishing these two fields!

miretskiy (approved):

andf-crl (approved):
Just one question, otherwise LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14027#-:-N3ygQ3b8G8JwCC1VZxe:bi4q63i)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl and @kathancox)

___
*[`v21.2/changefeed-sinks.md` line 77 at r3](https://reviewable.io/reviews/cockroachdb/docs/14027#-N3yekD97Y8evVU4biI4:-N3yekD97Y8evVU4biI5:b-br551e) ([raw file](https://github.com/cockroachdb/docs/blob/d4bbb41da2320a03afec1bbac58430c6e1c6c43d/v21.2/changefeed-sinks.md#L77)):*
Question: If I leave `Flush.Messsages` set to its default of `0`, does that mean "unlimited" or "disable this method of calculating when to flush"? If  not, doesn't a value of `0` negate any functionality from `Flush.MaxMessages`? Similarly with `Flush.Bytes` and `Flush.Frequency`: their default values likely disable them as methods of determining flush behavior. If so, it would be helpful to indicate with something like:

> A value of `0` causes this parameter to be ignored in determining when to batch requests.

PR #14061, opened 2022-06-07
troubleshoot excessive snapshot rebalance/recovery rates

taroface:
Fixes DOC-2409.

- Document KB issue https://cockroachlabs.atlassian.net/wiki/spaces/CKB/pages/2273673403/Rebalance+and+Recovery+Rates+-+Scaling+nodes+up+or+down
- A few opportunistic corrections and tweaks to troubleshooting and other docs.

@kevinkokomani @mattcrdb Note that I've kept this vague in terms of what it means for settings to be "too high". I'd love to include more detailed guidance than this, if it's at all possible, but I realize it's a more complex issue.


taroface:
Hi @irfansharif, would you be able to provide any input on @kevinkokomani's question above (recommended max values of `kv.snapshot_rebalance.max_rate` and `kv.snapshot_recovery.max_rate`)?

irfansharif:
It's difficult to give a blanket number for those values, the defaults today are somewhat conservative but as knobs, we don't like them very much and would like to remove them altogether in favor of automatic controls with CRDB: https://github.com/cockroachdb/cockroach/issues/80607. Perhaps we can say that these rates can be destabilizing or not at all depending on your environment, and our current recommendations are that you stay within 2x of the default setting (pulled entirely out of thin air).

taroface:
Thanks for this context, @irfansharif! I'll use this wording:

```
However, if the settings are too high when nodes are added to the cluster, this can cause degraded performance and node crashes. We recommend **not** increasing these values by more than 2x their [default values](cluster-settings.html) without explicit approval from Cockroach Labs.
```

MichaelTrestman (approved):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14061)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mattcrdb)

PR #14270, opened 2022-06-13
Use include_cached to improve performance

nickvigilante:
I was able to shave another 10 seconds off build time by converting existing includes to use `include_cached`.


stbof (approved):
Nice!

PR #14281, opened 2022-06-14
DOC-4530 tighten grep filename match

andf-crl:
Addresses: DOC-4530

- Tightens `grep` match on `$filename` on our `changed_files.yaml` GH action to capture cases where `filename.md` is actually `more-to-filename.md`; current `grep` syntax matches both.
- `cdc.md` file in here to test that this works, will pull for final merge ofc.


nickvigilante (approved):
Good catch! I didn't see this in testing.

PR #14295, opened 2022-06-15
update recs per 2022 Cloud Report

taroface:
Fixes DOC-3742.
Fixes DOC-3391.

Updated the following per 2022 Cloud Report:

- Recommended instance types for GCP, AWS, Azure
- Added some additional workload-specific recommendations

Also moved the instance recs into includes so that these are updated across all docs.

### DIFF
--- a/_includes/v21.2/prod-deployment/cloud-report.md
+++ b/_includes/v21.2/prod-deployment/cloud-report.md
@@ -1,0 +1,1 @@
+Cockroach Labs creates a yearly cloud report focused on evaluating hardware performance. For more information, see the [2022 Cloud Report](https://www.cockroachlabs.com/guides/2022-cloud-report/).
--- a/_includes/v21.2/prod-deployment/recommended-instances-aws.md
+++ b/_includes/v21.2/prod-deployment/recommended-instances-aws.md
@@ -1,0 +1,9 @@
+- Use general-purpose [`m6i` or `m6a`](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/general-purpose-instances.html) VMs with SSD-backed [EBS volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html). For example, Cockroach Labs has used `m6i.2xlarge` for performance benchmarking. If your workload requires high throughput, use network-optimized `m5n` instances. To simulate bare-metal deployments, use `m5d` with [SSD Instance Store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ssd-instance-store.html).
+
+    - `m5` and `m5a` instances, and [compute-optimized `c5`, `c5a`, and `c5n`](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/compute-optimized-instances.html) instances, are also acceptable.
+
+    {{site.data.alerts.callout_danger}}
+    CockroachDB does not support Arm-based `m6g` instances.
+
+    **Do not** use [burstable performance instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances.html), which limit the load on a single core.
+    {{site.data.alerts.end}}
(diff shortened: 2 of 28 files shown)


taroface:
@keith-mcclellan Let me know if you'd like to tag another reviewer here.

Also, I had a couple questions:

- Is it still OK to recommend the compute-optimized instances? We already had one listed for Azure, and I went ahead and mentioned the AWS `c` types as well.
- Do the recommendations under `### Cloud-specific recommendations` conflict with anything under `### Sizing` at all (or do we think having guidance on smaller vs. larger nodes in 2 places might be confusing)?

keith-mcclellan:
In general this looks good, one caveat is we need to call out provisioning IOPS and Throughput for gp3 volumes (aws)

With gp3 you get 3000 IOPS and 125MiB/Sec throughput by default, if you need more than either of those they need to be provisioned  separately

For example, an 8 vcpu box needs 4000 IOPS for example, and a 16 vCPU box we would want a min of 8000 IOPS and 250 MiB/sec Throughput

taroface:
> For example, an 8 vcpu box needs 4000 IOPS for example, and a 16 vCPU box we would want a min of 8000 IOPS and 250 MiB/sec Throughput

@keith-mcclellan The [recommendation in our docs](https://www.cockroachlabs.com/docs/stable/recommended-production-settings.html#disk-i-o) is 30 MB/s per vCPU, which comes out to 480 MB/s for a 16 vCPU machine. Can you help me square this away? Or is this a different value?

keith-mcclellan:
Honestly, this is the first I've seen the 30 MB/s recommendation per vCPU - TIL.

I'm not sure we can push it that much in most cases, but it doesn't hurt to overprovision things a bit,. So let's say 240 MB/s for the 8 vCPU instances and 480 MB/s for the 16 vCPU instances then. That work?

keith-mcclellan (approved):
LGTM

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14295)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @kathancox and @mwang1026)

___
*[`v21.2/recommended-production-settings.md` line 163 at r2](https://reviewable.io/reviews/cockroachdb/docs/14295#-N4d2DCdzYMsgGcLK5tM-r2-163:-N7RvP_09bTzyiFluVzc:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/fe24d4a32a12b1d87db624a4e91cfb50bb3379c5/v21.2/recommended-production-settings.md#L163)):*

Total nit: "more complicated workloads" perhaps I'm missing a point, but with workloads I think of "complexity", so maybe "complex workloads". Having said that I think either gets the point across!
</blockquote></details>

Done.
___
*[`v21.2/recommended-production-settings.md` line 165 at r2](https://reviewable.io/reviews/cockroachdb/docs/14295#-N4d2DCdzYMsgGcLK5tM-r2-165:-N7Rvb534JdX83fMbndC:b-dcbopc) ([raw file](https://github.com/cockroachdb/docs/blob/fe24d4a32a12b1d87db624a4e91cfb50bb3379c5/v21.2/recommended-production-settings.md#L165)):*

Nit: Option to de-passive this sentence, not sure "individual" makes sense or helps here:

Because storage cost influences the cost of running a workload more than the cost of (individual?) instances, larger nodes offer a better....


</blockquote></details>

Much better now, thanks!
___
*[`v22.1/recommended-production-settings.md` line 163 at r2](https://reviewable.io/reviews/cockroachdb/docs/14295#-N4d2DCdzYMsgGcLK5tW-r2-163:-N7Rvcqg6skq-d-8EtCc:b-896fix) ([raw file](https://github.com/cockroachdb/docs/blob/fe24d4a32a12b1d87db624a4e91cfb50bb3379c5/v22.1/recommended-production-settings.md#L163)):*

Same comments in this section as per v21.2
</blockquote></details>

Done.

PR #14308, opened 2022-06-16
DOC-2413: Add note about unsupported syntax

nickvigilante:
Fixes DOC-2413


nickvigilante:
Oops, wrong branch name.

PR #13167, opened 2022-03-03
Restore zone config pattern docs.

stbof:
https://cockroachlabs.atlassian.net/browse/DOC-2718

PR #14310, opened 2022-06-16
Zone config restoration followup.

stbof:
Followup to https://github.com/cockroachdb/docs/pull/13167.

Complete restoration by:

- Updating docs to point to current version instead of v20.2.
- Updating restored pages to point to correct image path.


nickvigilante (approved):
LGTM!

PR #14319, opened 2022-06-21
Add local linkcheck and updated exclude list from htmltest.

ianjevans:
This PR adds a `make linkcheck` target to check local links before pushing PRs. It updates the build files to use the mainline `wjdp/htmltest` branch instead of the `cockroachdb/htmltest` fork, as the upstream main branch has the necessary fixes to allow us to use it in our docs.

The `make vendor` or `make bootstrap` targets will now install `htmltest` locally.

I also updated the external links so nightly builds should pass link checks.

To get everything working, you may need to clean and re-bootstrap your local environment.

~~~
make clean
make bootstrap
make htmltest
~~~

The process of checking links before publishing a PR is as follows:

1. Stage and commit the PR changes in your local PR branch.
2. Run `make linkcheck`. Fix any links that are broken.
3. Publish or push your local PR branch.


stbof:
@ianjevans I ran the steps, but got an error on the last one:

➜  docs git:(local-linkcheck) make htmltest
make: *** No rule to make target 'htmltest'.  Stop.

.htmltest.yml
```diff
@@ -3,16 +3,28 @@ ExternalTimeout: 30
...
   - ^docs/dev
+  - ^docs/v1.0
+  - ^docs/v1.1
   - ^docs/v1.2
+  - ^docs/v2.0
+  - ^docs/v2.1
+  - ^docs/v19.1
+  - ^docs/v19.2
+  - ^docs/v20.1
+  - ^docs/releases
```
nickvigilante, line 14: I'm gonna take a wild guess and say the build errors are coming from this line here. Can you try omitting to see if that fixes it?

nickvigilante:
@stbof I think it's a typo. Try `make linkcheck`.

stbof (approved):
Hooray!

.htmltest.yml
ianjevans: Trying this right now.

nickvigilante: Looks like that fixed it. I think the problem is if we add a directory to that list then any file not being ignored that points to a file that is ignored throws an error, I think? Seems buggy.

PR #14333, opened 2022-06-22
Fix typos.

stbof:


stbof:
@nickvigilante I thought all the "included" version tags were reverted in tables.

nickvigilante (approved):
@stbof Ah, you're right. I forgot it's because in MD tables it doesn't like includes. LGTM!

stbof:
> @stbof Ah, you're right. I forgot it's because in MD tables it doesn't like includes. LGTM!

Yes, and I thought you backed the include out from all the tables when the issue was raised. I wonder if there are others still lurking in MD tables.

PR #14336, opened 2022-06-22
Removed include_cached new-in.html from MD tables

nickvigilante:


stbof:
In `v21.2/create-changefeed.md` there are a lot of instances of `<span class="version-tag">New in v21.2:</span>`. I thought we wanted to use only `**New in v21.2:**.

nickvigilante:
@stbof Check now. There were several `<span>` tags that either I missed the first time or others added after I did my initial pass.

stbof (approved):
I didn't check every single replacement, but a scan (no pun intended) says this is good.

nickvigilante:
@stbof Nothing funky is happening with the version switcher, so I'm merging.

PR #14338, opened 2022-06-22
Added Storage Permissions section to Cloud Storage bulk page

kathancox:
Fixes [DOC-581](https://cockroachlabs.atlassian.net/browse/DOC-581) [DOC-1525](https://cockroachlabs.atlassian.net/browse/DOC-1525)

Added a Storage Permissions section to the Cloud Storage bulk page. 

This introduces the minimum permission required for backup, restore, import, export, and changefeeds. For both versions v21.2 and v22.1, where backup has permission changes from 21.2 to 22.1. 

Also some _general_ detail to S3 and GCS storage permissions with prompts to check out the respective docs for setup.


kathancox:
V21.2 preview: https://deploy-preview-14338--cockroachdb-docs.netlify.app/docs/v21.2/use-cloud-storage-for-bulk-operations

v22.1 preview: https://deploy-preview-14338--cockroachdb-docs.netlify.app/docs/stable/use-cloud-storage-for-bulk-operations.html

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+[Restore](restore.html)     | List<br><br>Get                    | List: During a restore job, it is necessary to pass the backup's directory. This contains a manifest file that describes the backup's metadata and data. Restores require access to read these files from the storage bucket to process the backup correctly. <br> Get: Restores need access to retrieve files from the backup.
+[Import](import.html)       | Get                          | Imports pull the requested file(s) from the storage bucket.
```
stevendanna, line 240: s/pull/read/ to be consistent with the other descriptions

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+[Backup](backup.html)       | Write<br>Get<br>List<br>Delete | Write/Get: Backups write the backup data to the storage bucket. During a backup job, a `BACKUP CHECKPOINT` file will be written that tracks the progress of the backup. Should the backup resume after a [pause](pause-job.html), it will read the checkpoint file to find a suitable resume point. This can include overwriting backup work that had happened before the pause.<br> List: Backups need list access to the files already in the bucket. If the backup is paused, or there is a retryable error, the backup will read the `BACKUP CHECKPOINT` file.<br> Delete: Backups may delete files and progress during the job after a pause in order to resume the job from a certain point.
+[Restore](restore.html)     | List<br><br>Get                    | List: During a restore job, it is necessary to pass the backup's directory. This contains a manifest file that describes the backup's metadata and data. Restores require access to read these files from the storage bucket to process the backup correctly. <br> Get: Restores need access to retrieve files from the backup.
```
stevendanna, line 239: For List, I'm not sure this describes why we need the permission since it only mentions that the directory contains a manifest and that we need read on that manifest. Do we want these to be comprehensive lists of when we use the operations?  The number of uses of List grew between 21.2 and 22.1, but we can easily generate the list for both. if that is the goal.

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+[Import](import.html)       | Get                          | Imports pull the requested file(s) from the storage bucket.                   
+[Export](export.html)       | Write                        | Exports need write access to the storage bucket to create individual export file(s) from the exported data.
+[Enterprise changefeeds](create-changefeed.html)  | Write  | Changefeeds will write messages files to the storage bucket as it emits row changes.
```
stevendanna, line 242: Do we use the term "messages files" elsewhere?  If not, perhaps something like "Changefeeds will write files to the storage bucket that contain row changes and resolved timestamps. "

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+</div>
+
+<section class="filter-content" markdown="1" data-scope="s3">
+
+These [actions](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_Simple_Storage_Service.html) are the minimum access permissions to be set in an Amazon S3 bucket policy:
+
+Operation    | S3 permission                                                                          
+-------------+----------------------------------------------------------------------------------
+Backup       | `s3:PutObject`, `s3:ListBucket`, `s3:DeleteObject`, `s3:GetObject`
+Restore      | `s3:GetObject`, `s3:ListBucket`
```
stevendanna, line 256: [nit] the ordering of the options isn't consistent between these two rows

v22.1/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,85 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+
+This section describes the minimum permissions required to run CockroachDB bulk operations. While we provide the required permissions for Amazon S3 and Google Cloud Storage, the provider's documentation provides detail on the setup process and different options regarding access management.
+
+Depending on the actions a bulk operation performs, it will require different access permissions to a cloud storage bucket. 
+
+This table outlines the actions that each operation performs against the storage bucket:
+
+Operation                   | Permission                   | Description                                                                     
+----------------------------+------------------------------+---------------------------------------------------
+[Backup](backup.html)       | Write<br>List                  | Write: Backups write the backup data to the bucket/container. During a backup job, a `BACKUP CHECKPOINT` file will be written that tracks the progress of the backup.<br> Read: Backups need list access to the files already in the bucket. If the backup is [paused](pause-job.html), or there is a retryable error, the backup will read the `BACKUP CHECKPOINT` file.
```
stevendanna, line 238: It is correct that Delete is no longer needed, but I wonder if we want to call out somewhere that if the user doesn't provide delete, then we can't cleanup checkpoints on failure. So it is a small tradeoff.

The description here describes why we need to be able to read, but Get isn't included in the permission list or the example policies below.

stevendanna (changes requested):
Thanks for working on this! It will be nice to be able to link customers to this.

v22.1/use-cloud-storage-for-bulk-operations.md
kathancox: Yes I can add a note on that tradeoff, I think that's a great idea. If you had delete permission, would the checkpoint files only be deleted on failure? There wouldn't be clean up of the checkpoint files if the backup is successful... and that's because we only write multiple if there is a pause/retryable error and we don't overwrite?

On the second point, that was my bad, I had "Read" here but that should have been "List" in the description column. On testing this, `Get` wasn't required for backups in v22.1.

v21.2/use-cloud-storage-for-bulk-operations.md
kathancox: We do use "message" elsewhere, but in the case of cloud storage just using "files" is better. And I prefer yours!

v22.1/use-cloud-storage-for-bulk-operations.md
kathancox: Update, I added a draft note to v22.1. Looking at it though, I think I need to define that this is on failure, if you can confirm?

v22.1/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,85 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+
+Depending on the actions a bulk operation performs, it will require different access permissions to a cloud storage bucket. 
+
+This table outlines the actions that each operation performs against the storage bucket:
+
+Operation                   | Permission                   | Description                                                                     
+----------------------------+------------------------------+---------------------------------------------------
+[Backup](backup.html)       | Write<br><br>List                | Write: Backups write the backup data to the bucket/container. During a backup job, a `BACKUP CHECKPOINT` file will be written that tracks the progress of the backup.<br> List: Backups need list access to the files already in the bucket. If the backup is [paused](pause-job.html), or there is a retryable error, the backup will read the `BACKUP CHECKPOINT` file. <br>**Note:** To clean up `BACKUP CHECKPOINT` files that the backup job has written, you would need to also include a delete permission in your bucket policy (e.g., `s3:DeleteObject`). However, delete is **not** necessary for backups  to complete successfully in v22.1 and later.
```
stevendanna, line 238: extra space: "backups  to"

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+
+This table outlines the actions that each operation performs against the storage bucket:
+
+Operation                   | Permission                   | Description                                                                     
+----------------------------+------------------------------+---------------------------------------------------
+[Backup](backup.html)       | Write<br>Get<br><br>List<br><br>Delete | Write/Get: Backups write the backup data to the storage bucket. During a backup job, a `BACKUP CHECKPOINT` file will be written that tracks the progress of the backup. Should the backup resume after a [pause](pause-job.html), it will read the checkpoint file to find a suitable resume point. This can include overwriting backup work that had happened before the pause.<br> List: Backups need list access to the files already in the bucket. If the backup is paused, or there is a retryable error, the backup will read the `BACKUP CHECKPOINT` file.<br> Delete: Backups may delete files and progress during the job after a pause in order to resume the job from a certain point.
```
stevendanna, line 238: > List: Backups need list access to the files already in the bucket. If the backup is paused, or there is a retryable error, the backup will read the `BACKUP CHECKPOINT` file.

In v21.2 we use List for finding previous backup in an incremental change. So perhaps something like

> List: Backups need list access to the files already in the bucket. For example, BACKUP uses List to find previously taken backups when executing an incremental backup.

v22.1/use-cloud-storage-for-bulk-operations.md
stevendanna: > On testing this, Get wasn't required for backups in v22.1.

I think Get is required in v22.1.  For example, if you pause and resume a backup, you need Get in order to read the backup checkpoint.

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+[Restore](restore.html)     | List<br><br><br>Get                    | List: During a restore job, it is necessary to pass the backup's directory. This contains a manifest file that describes the backup as well as other versioned subdirectories and files. Restores require access to read these files from the storage bucket to process the backup correctly. <br> Get: Restores need access to retrieve files from the backup.
```
stevendanna, line 239: Similar to Backup, v21.2 uses List to find the incremental backups.

v22.1/use-cloud-storage-for-bulk-operations.md
stevendanna: My comments on the 21.2 page apply here as well. In 22.1, List is additionally used during Backup the fine the _latest_ checkpoint. List is additionally used during Restore to find the latest LATEST file.

v21.2/use-cloud-storage-for-bulk-operations.md
```diff
@@ -225,6 +225,87 @@ A custom root CA can be appended to the system's default CAs by setting the `clo
...
+[Backup](backup.html)       | Write<br><br>Get<br><br>List<br><br>Delete | Write: Backups write the backup data to the bucket/container. During a backup job, a `BACKUP CHECKPOINT` file will be written that tracks the progress of the backup.<br>Get: Backups need get access after a [pause](pause-job.html) to read the checkpoint files on [resume](resume-job.html).<br>List: Backups need list access to the files already in the bucket. For example, `BACKUP` uses list to find previously taken backups when executing an incremental backup.<br> Delete: Backups may delete files and progress during the job after a pause in order to resume the job from a certain point. This can include overwriting backup work that had happened before the pause.
+[Restore](restore.html)     | Get<br>List          | Get: Restores need access to retrieve files from the backup.<br>List: During a restore job, it is necessary to pass the backup's directory. This contains metadata files that describe the backup as well as other versioned subdirectories and files. Restore also requires access to the `LATEST` file in order to read the latest available backup.
```
stevendanna, line 239: non-blocking: But I think that the `List` description here sounds more like why we need Get.  If you wanted, the reasons we need list here are similar to why we need list in Backup, namely to find other backups in the backup collection.

PR #14346, opened 2022-06-23
Added release note for backup/protected timestamp bug fix

kathancox:
This is a quick fix to add a release note for the recent bug fix to protected timestamps & requests.


andf-crl (approved):
LGTM as is, or suggested a slight rephrase to highlight the previous behavior.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14346#-:-N5G_ium4zK2D2jv9Adz:bx1bep4)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl and @kathancox)

___
*[`_includes/releases/v22.1/v22.1.2.md` line 36 at r1](https://reviewable.io/reviews/cockroachdb/docs/14346#-N5GWfeR-mE-imIr5DcD:-N5GWfeR-mE-imIr5DcE:b-l1bgep) ([raw file](https://github.com/cockroachdb/docs/blob/550596c1884281f64b87d37c52a9472bac5414d7/_includes/releases/v22.1/v22.1.2.md#L36)):*
nit: This is a complicated root cause; I think it might benefit from a mention of the behavior we would have seen previously: backups failing due to being GC'd midway. What do you think of:

> Fixed a bug where some KV requests, particularly export requests, issued during a [backup](../v22.1/backup.html) could be incorrectly rejected, causing the backup to fail with a `batch timestamp <ts> must be after replica GC threshold` error. This was because [protected timestamps](../v22.1/architecture/storage-layer.html#protected-timestamps) were not considered when computing the [garbage collection threshold](../v22.1/architecture/storage-layer.html#garbage-collection) of the key span being backed up, causing some longer-running backups to fail due to being garbage collected before they could complete. [#82757][#82757]

PR #14356, opened 2022-06-23
Remove unused includes from old versions of the docs

nickvigilante:


stbof:
Are these files no longer included by their respective CockroachDB versions, or are they referenced by those files but they are no longer used because we are no longer building those versions?  I'd worry about removing them unless we are also removing the referencing files.

nickvigilante:
@stbof The former. They aren't referenced in any document in their respective versions (or in any version for that matter, seeing as both Jekyll and htmltest built successfully).

stbof (approved):

PR #14386, opened 2022-06-28
Language nit for unrestricted dedicated storage options

gemma-shay:
Change to "we recommend" 150 GiB limitation


jramos-crl (approved):
Looks good to me!

PR #14436, opened 2022-07-01
DOC-886: Make feedback widget more visible

jseldess:
This PR moves the per-page docs feedback widget from
the bottom of each page to below the toc and adjusts
styling accordingly. Bottom of page placement is typical
on doc sites, but I'd like to see whether the new
position elicits more responses. On pages with long tocs,
the widget is less accessible. However, on the majory
of pages, the feedback prompt is now much more visible
throughout a user's time on  page, not just when the
bottom is reached.

This PR also:

- Changes the "No" response to prompt for
more details via a link to a templated GH issue in
our docs repo.
- Removes the "Contribute" drop-down options. The options
are almost never leveraged by external users. The source
files are still in place, so we can restore this if
we hear people miss it.


jseldess:
Screen Shot 2022-07-01 at 3 44 12 PM

Screen Shot 2022-07-01 at 3 45 03 PM

Screen Shot 2022-07-01 at 3 45 36 PM

mikeCRL (approved):
LGTM. One caveat is that a page with no TOC won't get the widget, e.g. `docs/stable/cluster-settings.html`. This is probably a worthwhile tradeoff for surfacing better feedback on the vast majority of pages.

jseldess:
Good point, @mikeCRL. I just pushed a commit to address this. When there's no page toc, the feedback widget now shows at the bottom of the page:

Screen Shot 2022-07-02 at 12 23 56 PM

The UX is still not great for pages with long tocs, however. In the future, I will try to find a way to show the widget at both places on all pages. Right now, when I do that, the No feedback message/GH link doesn't work properly in both places.

PR #14454, opened 2022-07-06
Define volatility and xref.

stbof:
DOC-4588

Depends on https://github.com/cockroachdb/cockroach/pull/84038.


v22.1/computed-columns.md
```diff
@@ -60,7 +60,7 @@ Parameter | Description
 ----------|------------
 `column_name` | The [name](keywords-and-identifiers.html#identifiers) of the computed column.
 `<type>` | The [data type](data-types.html) of the computed column.
-`<expr>` | The immutable [scalar expression](scalar-expressions.html) used to compute column values. You cannot use any functions, such as `now()` or `nextval()`, that are not immutable.
+`<expr>` | The [immutable](functions-and-operators.html#function-volatility) [scalar expression](scalar-expressions.html) used to compute column values. You cannot use functions such as `now()` or `nextval()`, that are not immutable.
```
nickvigilante, line 63:
```suggestion
`<expr>` | The [immutable](functions-and-operators.html#function-volatility) [scalar expression](scalar-expressions.html) used to compute column values. You cannot use functions such as `now()` or `nextval()` that are not immutable.
```

nickvigilante (approved):

PR #14481, opened 2022-07-07
Update row-level TTL docs with add'l details

rmloveland:
Fixes DOC-3545

Summary of changes:

- Note that TTL update runs a schema change; also add some specifics
  around per-row overrides vs. updating the whole table

- High metrics cardinality can lead to bad perf (see CDC doc
  description)

- List required privileges for TTL

- Show an example of updating `ttl_expire_after` once already set

### DIFF
--- a/v22.1/row-level-ttl.md
+++ b/v22.1/row-level-ttl.md
@@ -79,7 +79,7 @@ SHOW CREATE TABLE ttl_test;
 
 ## TTL storage parameters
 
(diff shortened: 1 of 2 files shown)


otan (approved):

ianjevans (approved):

v22.1/row-level-ttl.md
```diff
@@ -455,6 +461,14 @@ Row-level TTL interacts with [backup and restore](backup-and-restore-overview.ht
 
 - When you [`RESTORE`](restore.html) from a backup, all row-level TTL information associated with the tables being restored (including TTL expiration times) is also restored. Any expired rows in the restored tables are eligible to be [deleted](#when-are-rows-deleted) by the [TTL job](#how-it-works).
 
+## Required Privileges
+
+To add or update Row-Level TTL settings on a table, you must have one of the following:
+
+- Membership to the [`admin`](security-reference/authorization.html#roles) role for the cluster.
```
vy-ton, line 468: @ecwall @rafiss Do we have an issue for taking a second look at privileges for 22.2? I don't think we want to add more to the `admin` role.

PR #14490, opened 2022-07-08
Show feedback widget in 2 places

jseldess:
Follow-up to https://github.com/cockroachdb/docs/pull/14436

Previously, the feedback widget would show either under the
page toc or, when there is no toc, at the bottom of the page.
Now, the widget shows up in both places. When the user
interacts with one, the other is hidden so as to avoid
confusion and/or duplicate submissions.

The footer overlaps the toc, which causes a strange UX
in some cases. That can be handled separately.


andf-crl (approved):
This is great! Some tiny design feedback you can feel free to disregard if you disagree!

- Class `feedback-question` blends in with right TOC entries a little. Italics, bold, or a different font size could help separate it from the TOC. Looks fine at bottom (indent helps it stand out), but I'd probably update both to be consistent.
- Classes `yes-button` and `no-button` look very similar and are small and close together: they might benefit from one or more of:
  - an increase in size
  - a little more padding between them
  - consider aligning the "wrist" of each hand with the other, such that thumb direction stands out even more. Since we have text directly above us, this in practice would just be a lowering of the x position of thumbs down.
   Otherwise it's rather easy to click the wrong tiny button by accident.
- I see the same footer UI overlap you do with certain scroll positions. Minor indeed, for another PR.

All minor, optional feedback. Nice work!

jseldess:
TFTR, @andf-crl! Please take another look at the thumbs/up down images.

andf-crl:
Nice, I think that really helps differentiate! If you're looking to keep tweaking, I feel like the icons are just a little too far from `Was this helpful?`. What does it look like if we lose, say, `4px` between text and icon row? Perhaps also a slight indent, so that thumbs-up wrist is under mid-`W` of `Was`, to further indicate "not second discrete concept, still related to first".

But also: fully shippable as is!

PR #14507, opened 2022-07-11
Added Dedicated multi-region restore workflow

kathancox:
Fixes [DOC-4905](https://cockroachlabs.atlassian.net/browse/DOC-4905)

Added an additional workflow to the Dedicated section of the CockroachCloud Restore page. This includes the matching of regions between databases.

### DIFF
--- a/cockroachcloud/backups-page.md
+++ b/cockroachcloud/backups-page.md
@@ -113,7 +113,7 @@ To restore a database:
 
     The **Restore database** module displays with backup details.
 
-1. In the **Restore to** field, enter the name of the destination database.
+1. In the **Restore to** field, enter the name of the destination database. For multi-region databases, see [Restore a multi-region database](#restore-a-multi-region-database).
 
     {{site.data.alerts.callout_info}}
     [Resolve any naming conflicts](#resolve-a-database-naming-conflict) by using [`DROP`](../{{site.versions["stable"]}}/drop-database.html) or [`RENAME`](../{{site.versions["stable"]}}/rename-database.html) on the existing database. If you enter a unique name in the **Restore to** field, a new database will be created.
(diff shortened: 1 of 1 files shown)


livlobo (approved):
Looks good! Thanks for the quick turnaround @kathancox!

andf-crl (commented):
Thanks, nice work! I was a bit confused by the MR and non-MR steps, and choosing between them. I think presenting the choice a little earlier would go a long way in getting the right procedural in front of the user. Feel free to merge if you like your changes, or run it by me again if you want another look. Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14507#-:-N6s_8lB9bk2T0wxtoIZ:bojnim7)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl and @kathancox)

___
*[`cockroachcloud/backups-page.md` line 116 at r1](https://reviewable.io/reviews/cockroachdb/docs/14507#-N6sOk3E96aFaWsrI78k:-N6sOk3E96aFaWsrI78l:b9odwgv) ([raw file](https://github.com/cockroachdb/docs/blob/a63d9a9382cd2364f8202c87ea5e375927c10f35/cockroachcloud/backups-page.md#L116)):*
> ```Markdown
>     The **Restore database** module displays with backup details.
> 
> 1. In the **Restore to** field, enter the name of the destination database. For multi-region databases, see [Restore a multi-region database](#restore-a-multi-region-database).
> ```

Step 3 of the non-MR (do we call it Single Region?) steps introduces the MR alternative. Since MR actually requires you to run MR-specific steps first ("the workaround") and then complete the task with the regular non-MR steps (as indicated in its step 4) , the MR user finds themselves doing:

1. Read "### Restore a database" through to step 3.
2. Find that step 3 of the above sends us to "#### Restore a multi-region database", so go there instead
3. Find that step 4 of the above sends us back to "### Restore a database", where we encounter its Step 3 again (but I think we are meant to start with step 4 of non-MR)

Perhaps, to avoid the jumping, we can move the choice of which procedural to follow to somewhere before both of them. I could see this either as:

a. Under the "## Ways to restore data" section, "Restore a database" could be broken into two:

Restore a database
Restore a multi-region database
Restore a table

(then no need to later nest MR under "### Restore a database", and it can be introduced separately)

-- OR --

b. Introduce the choice briefly as first sentence under "### Restore a database", helping the user decide which sub-procedural to follow.
___
*[`cockroachcloud/backups-page.md` line 138 at r1](https://reviewable.io/reviews/cockroachdb/docs/14507#-N6sSfLh8JNkaLjpjIAq:-N6sSfLh8JNkaLjpjIAr:b-3avy5c) ([raw file](https://github.com/cockroachdb/docs/blob/a63d9a9382cd2364f8202c87ea5e375927c10f35/cockroachcloud/backups-page.md#L138)):*
nit: the ticket called it a workaround, but to the user this is just the correct procedural for their use case. Perhaps something like:

> If you need to restore a MR DB you have backed up to a different database name, you must use the following procedure to do so.
___
*[`cockroachcloud/backups-page.md` line 168 at r1](https://reviewable.io/reviews/cockroachdb/docs/14507#-N6sTbfo95a4NcdkHDC0:-N6sTbfo95a4NcdkHDC1:b-tuc414) ([raw file](https://github.com/cockroachdb/docs/blob/a63d9a9382cd2364f8202c87ea5e375927c10f35/cockroachcloud/backups-page.md#L168)):*
> ```Markdown
>     ~~~
> 
> 1. From the Console, go to the **Backups** page and choose the backup you want to restore. Click on the number of tables in the database. You will find a list of all the tables contained in the database's backup. Click **Restore** for each table you want to restore into the new database. Provide the new database's name (e.g., `new_movr`) in **Restore to** for the **Destination database** name.
> ```

nit: `all the tables` --> `all tables`

nit: Most sentences here are direct instructions. We could make these sub items of #3, in a fashion similar to:

3. From the Console, go to the Backups page and perform the following:
     a. Choose the backup you want to restore.
     b. Click on the number of tables in the database. You will find a list of all the tables contained in the database's backup. 
     c. Click Restore for each table you want to restore into the new database. Provide the new database's name (e.g., new_movr) in Restore to for the Destination database name.
___
*[`cockroachcloud/backups-page.md` line 170 at r1](https://reviewable.io/reviews/cockroachdb/docs/14507#-N6sVcDA2-Sp0rRGE7t5:-N6sVcDA2-Sp0rRGE7t6:bb51ghv) ([raw file](https://github.com/cockroachdb/docs/blob/a63d9a9382cd2364f8202c87ea5e375927c10f35/cockroachcloud/backups-page.md#L170)):*
> ```Markdown
> 1. From the Console, go to the **Backups** page and choose the backup you want to restore. Click on the number of tables in the database. You will find a list of all the tables contained in the database's backup. Click **Restore** for each table you want to restore into the new database. Provide the new database's name (e.g., `new_movr`) in **Restore to** for the **Destination database** name.
> 
> 1. Continue through the [Restore a database](#restore-a-database) workflow.
> ```

nit: while I understand the desire to reuse steps, this leaves things a little vague: "Return to the non-MR steps, but really only for its steps 4-6." 

Consider adding non-MR steps 4-6 directly here so no bouncing around is required.

kathancox:
Hey @andf-crl, Thanks for the review! I have adapted this to your comments. Totally makes sense! thanks so much

PR #14510, opened 2022-07-11
Added note to performance section on pausing jobs after max retries

kathancox:
Fixes [DOC-3506](https://cockroachlabs.atlassian.net/browse/DOC-3506)

Added a note to the performance section of `RESTORE`, `IMPORT`, & `BACKUP` on pausing if reaching maximum retries. 

Also, opportunistic edit adding new version feature to a prior bullet that I missed in the original PR. 

Furthermore, added `IMPORT` performance bullets to `IMPORT INTO` as I believe these all belong there too and had been missing.

### DIFF
--- a/v22.1/backup.md
+++ b/v22.1/backup.md
@@ -134,7 +134,7 @@ This improves performance by decreasing the likelihood that the `BACKUP` will be
 
 `BACKUP` will initially ask individual ranges to backup but to skip if they encounter an intent. Any range that is skipped is placed at the end of the queue. When `BACKUP` has completed its initial pass and is revisiting ranges, it will ask any range that did not resolve within the given time limit (default 1 minute) to attempt to resolve any intents that it encounters and to **not** skip. Additionally, the backup's transaction priority is then set to `high`, which causes other transactions to abort until the intents are resolved and the backup is finished.
 
-{% include_cached new-in.html version="v22.1" %} A backup job will [pause](pause-job.html) instead of entering a `failed` state if it continues to encounter transient errors once it has retried a maximum number of times. Once the backup has paused, you can [resume](resume-job.html) or [cancel](cancel-job.html).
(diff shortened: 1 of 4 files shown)


kathancox:
Hey @adityamaru, while I was adding this note to restore, backup, and import. I noticed that import into didn't have the performance points that the import page has. Reading through them, it seemed as if they would apply to import into too, could you also check that's correct?

adityamaru (approved):

andf-crl (approved):
LGTM with tiny optional nit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14510#-:-N7wBY2-ChAmMNNE8Mpw:b-jmywvs)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl and @kathancox)

___
*[`v22.1/backup.md` line 137 at r1](https://reviewable.io/reviews/cockroachdb/docs/14510#-N7wAqL-BWXsAEAk0EdZ:-N7wAqL-BWXsAEAk0Ed_:bc1lzia) ([raw file](https://github.com/cockroachdb/docs/blob/3e5e54be91898a902161ce76e8050a4955b5a002/v22.1/backup.md#L137)):*
> ```Markdown
> `BACKUP` will initially ask individual ranges to backup but to skip if they encounter an intent. Any range that is skipped is placed at the end of the queue. When `BACKUP` has completed its initial pass and is revisiting ranges, it will ask any range that did not resolve within the given time limit (default 1 minute) to attempt to resolve any intents that it encounters and to **not** skip. Additionally, the backup's transaction priority is then set to `high`, which causes other transactions to abort until the intents are resolved and the backup is finished.
> 
> {% include_cached new-in.html version="v22.1" %} A backup job will [pause](pause-job.html) instead of entering a `failed` state if it continues to encounter transient errors once it has retried a maximum number of times. Once the backup has paused, you can [resume](resume-job.html) or [cancel](cancel-job.html).
> ```

nit: you can either resume or cancel it.

x4 in this review.

PR #14530, opened 2022-07-13
Add more Row-level TTL details

rmloveland:
Fixes the last bits of DOC-3545

Followup to cockroachdb/docs#14481

Summary of changes:

- Update the required privileges to remove incorrect reference to
  `admin` role

- Add note that the schema change backfill of
  `crdb_internal_expiration` (and its performance impact) is necessary
  in part due to the limitation that the TTL cannot be customized based
  on the values of other columns in the row

  - Also note that adding TTL to new tables is not affected by this
    limitation


rmloveland:
@otan !!! couple of changes in this PR:

- I updated the required perms stuff based on your comment https://github.com/cockroachdb/docs/pull/14481#discussion_r919436508
- tryna make it clearer that new tables created with TTL are happy but big old tables with TTL added may be sad (and how that is linked to the inability to customize TTL based on row values)

PTAL! and thanks!

@vy-ton FYI this PR exists and is aiming to address your latest comments on DOC-3545 so we can call that one done

otan (approved):

ianjevans (approved):

PR #14533, opened 2022-07-13
DOC-5001: Make Vale ignore Liquid blocks

nickvigilante:
Fixes DOC-5001


ianjevans (approved):
Looks good. The only other Vale weirdness I've seen recently is Value throwing spell check errors on URLs in links (e.g. [link text](https://example.com/word-that-causes-vale-error.html)).

PR #14540, opened 2022-07-14
DOC-5012: Update Algolia search index to include TAs and API docs

nickvigilante:
Fixes DOC-5012


jseldess (approved):
LGTM! Thanks, @nickvigilante!

jseldess:
cc @ianjevans, we'll have to do this for the `ccloud` docs, if they're in a new subdirectory.

PR #14545, opened 2022-07-14
TA-82079

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-5009

### DIFF
--- a/advisories/a82079.md
+++ b/advisories/a82079.md
@@ -1,0 +1,180 @@
+---
+title: Technical Advisory 82079
+advisory: A-82079
+summary: If a <code>CREATE MATERIALIZED VIEW</code> statement fails, all objects referenced in its <code>SELECT</code> query wil be unusable.
+toc: true
+affected_versions: v21.2.0-v21.2.12, v22.1.0
+advisory_date: 2022-07-18
+docs_area: releases
+---
+'
+Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
+
+## Description
+
+The problem can occur on any cluster running versions v21.2.0 to v21.2.12 or v22.1.0. If a `CREATE MATERIALIZED VIEW` statement fails, all the tables, views, sequences and types referenced in its `SELECT` query will become unusable.
+
+This problem can be diagnosed by running the following query:
+
+~~~ sql
+WITH
+	tables
+		AS (
+			SELECT
+				id, tab
+			FROM
+				(
+					SELECT
+						id,
+						crdb_internal.pb_to_json(
+							'cockroach.sql.sqlbase.Descriptor',
+							descriptor
+						)->'table'
+							AS tab
+					FROM
+						system.descriptor
+				)
+			WHERE
+				tab IS NOT NULL
+		),
+	old_depended_on_by_entries
+		AS (
+			SELECT
+				id, json_array_elements(tab->'dependedOnBy') AS entry
+			FROM
+				tables
+		),
+	ids_of_descriptors_that_need_upsert
+		AS (
+			SELECT
+				DISTINCT id
+			FROM
+				old_depended_on_by_entries
+			WHERE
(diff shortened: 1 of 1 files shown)


advisories/a82079.md
```diff
@@ -0,0 +1,180 @@
...
+							new_depended_on_by_arr
+						),
+						ARRAY['version'],
+						((tab->>'version')::INT8 + 1)::STRING::JSONB
+					),
+					ARRAY['modificationTime']
+				)
+			)
+		),
+		false
```
postamar, line 167: This `false` actually needs to be replaced with `true`.

advisories/a82079.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Statement
+
+This is resolved in CockroachDB by PR [#82087](https://github.com/cockroachdb/cockroach/pull/82087) which fixes the missing removal of back-references to the materialized view.
+
+The fix has been applied to maintenance releases of CockroachDB v21.2.13 and v22.1.1. This public issue is tracked by [#82079](https://github.com/cockroachdb/cockroach/pull/82079).
+
+## Mitigation
+
+Users of CockroachDB versions v21.2.0 to v21.2.12 or v22.1.0 are encouraged to upgrade to [v21.2.13](../releases/v21.2.html#v21-2-13) and [v22.1.1](../releases/v22.1.html#v22-1-1) or a later version.
```
postamar, line 72: Add a sentence to the effect of: `Upgrading to a later maintenance release only prevents the problem from appearing, it will not fix an existing problem.`

advisories/a82079.md
```diff
@@ -0,0 +1,180 @@
...
+affected_versions: v21.2.0-v21.2.12, v22.1.0
+advisory_date: 2022-07-18
+docs_area: releases
+---
+
+Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
+
+## Description
+
+The problem can occur on any cluster running versions v21.2.0 to v21.2.12 or v22.1.0. If a `CREATE MATERIALIZED VIEW` statement fails, all the tables, views, sequences and types referenced in its `SELECT` query will become unusable.
```
postamar, line 15: Please also add:
```

Any statement or query referencing them will trigger an error matching the following regular expression:

    invalid depended-on-by relation back reference: referenced table ID [0-9]*: referenced descriptor not found

```

postamar (commented):
Recent support events have brought to light a few issues with the TA such as it was worded. Sorry about this!

Everything else looks good otherwise.

postamar (approved):

advisories/a82079.md
```diff
@@ -0,0 +1,183 @@
...
+			)
+		),
+		true
+	)
+FROM
+	tables
+	JOIN new_depended_on_by_arrs ON tables.id = new_depended_on_by_arrs.id;
+~~~
+
+We encourage users to [back up](../{{site.versions["stable"]}}/backup-and-restore-overview.html) their clusters before undertaking such a repair.
```
rmloveland, line 177: How do I know if my repair attempt succeeded?

Do we have a clearer recommendation? e.g., "if the repair attempt succeeds, X will happen.  If X does not happen, restore from a backup"

basically as a user I would no longer trust my cluster in this state without some more guarantees tbh

rmloveland: @postamar this ^^ is really a question for you I think, not really a docs copy edit type comment

rmloveland (approved):
LGTM as is if we're pressed for time, but I had a couple comments (and a question about how to know if the repair succeeded / what to do )

rmloveland:
Marius Posta ***@***.***> writes:

> The repair succeeds if the query succeeds. Re-running the diagnostic
> query can confirm this. For the paranoid, there's the option of doing
> just that (repair + re-diagnosis) in an explicit transaction.

OK thanks for explaining this Marius!  And Gemma, I see you updated the
doc with this info - thanks!

PR #14548, opened 2022-07-14
TA-82576

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-5010


advisories/a82576.md
```diff
@@ -0,0 +1,206 @@
+---
+title: Technical Advisory 82576
+advisory: A-82576
+summary: Adding a column to a table which references a sequence, or creating a table with columns referencing sequences, adds an incomplete back-reference to the column to the sequence metadata.
```
postamar, line 4: Minor comment: I'm concerned that this repetition of "to" in the last clause might be confusing. To be clear, the back-reference exists within the sequence metadata, and refers to both the column and the table the column is in.

postamar (approved):

advisories/a82576.md
gemma-shay: would it make more sense to just say "...adds an incomplete back-reference to the sequence metadata."

postamar: Yes, absolutely.

advisories/a82576.md
```diff
@@ -0,0 +1,207 @@
...
+                    ),
+                    ARRAY['modificationTime']
+                )
+            )
+        ),
+        true
+       )
+  FROM depended_on_by_arrs JOIN tables ON id = seq_id;
+~~~
+
```
rmloveland, line 202: Similar to the other TA, IMO it would be good to add a line like "if this query succeeds, the problem is repaired" (or whatever the better sentence you wrote in the other TA was)

rmloveland (approved):
LGTM with one suggestion

PR #14550, opened 2022-07-15
Add back contribute options to docs pages

jseldess:
These options were accidentally removed as part of
experiments in https://github.com/cockroachdb/docs/pull/14436.

Adding these back with a few changes:
- The drop-down now offers 2 options: Edit This Page and Report Doc Issue.
  The previous third option, Suggest New Content, rarely got engagement
  and is probably covered by Report Doc Issue.
- The Report Doc Issue now uses the same GH issue template as the
  Was this page helpful? No vote.


stbof:
LGTM

PR #14590, opened 2022-07-19
Document session variables related to full scans.

stbof:
DOC-1712
DOC-1667
DOC-1668

### DIFF
--- a/v21.2/cost-based-optimizer.md
+++ b/v21.2/cost-based-optimizer.md
@@ -104,13 +104,13 @@ When `sql.stats.histogram_collection.enabled` is set to `false`, histograms are 
 
 ## Control whether the optimizer creates a plan with a full scan
 
-Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may decide that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and determine that a statement plan isn't using the secondary index, it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
+Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may determine that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and find that a statement plan is not using the secondary index, then it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
 
 You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html).
 
 If you disable full scans, you can set the `large_full_scan_rows` session variable to specify the maximum table size allowed for a full scan. If no alternative plan is possible, the optimizer will return an error.
 
-If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you don't provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
+If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you do not provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
 
 ## Locality optimized search in multi-region clusters
 
--- a/v21.2/cost-based-optimizer.md
+++ b/v21.2/cost-based-optimizer.md
@@ -102,6 +102,16 @@ SET CLUSTER SETTING sql.stats.histogram_collection.enabled = false;
 
 When `sql.stats.histogram_collection.enabled` is set to `false`, histograms are never collected, either as part of automatic statistics collection or by manually invoking [`CREATE STATISTICS`](create-statistics.html).
 
+## Control whether the optimizer creates a plan with a full scan
+
+Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may decide that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and determine that a statement plan isn't using the secondary index, it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
+
+You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html).
+
+If you disable full scans, you can set the `large_full_scan_rows` session variable to specify the maximum table size allowed for a full scan. If no alternative plan is possible, the optimizer will return an error.
+
+If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you don't provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
+
 ## Locality optimized search in multi-region clusters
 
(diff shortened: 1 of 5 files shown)


rytaft (commented):
Sorry for the delay!

> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14590#-:-N7alVetE7PfiwXJQ0IF:b2c5wy4)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)

___
*[`v22.1/cost-based-optimizer.md` line 188 at r3](https://reviewable.io/reviews/cockroachdb/docs/14590#-N7aiFiu8pETBfEGjpOn:-N7aiFiu8pETBfEGjpOo:b-ejj0to) ([raw file](https://github.com/cockroachdb/docs/blob/9662875f1734349145b1f1dee98904ac3aa1428a/v22.1/cost-based-optimizer.md#L188)):*
> ```Markdown
> Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may decide that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and determine that a statement plan isn't using the secondary index, it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
> 
> You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html). If `disallow_full_table_scans` is `true` and you set the `large_full_scan_rows` [session variable](set-vars.html). If no alternative plan is possible, an error will be returned.
> ```

This sentence doesn't quite make sense: "If `disallow_full_table_scans` is `true` and you set the `large_full_scan_rows` [session variable](set-vars.html)." I think you're missing some words to explain that `large_full_scan_rows` determines the maximum table size allowed for a full scan when `disallow_full_table_scans` is set to true.
___
*[`v22.1/cost-based-optimizer.md` line 190 at r3](https://reviewable.io/reviews/cockroachdb/docs/14590#-N7aiswq-gGFtcjLM3tF:-N7aiswq-gGFtcjLM3tG:b-jc770) ([raw file](https://github.com/cockroachdb/docs/blob/9662875f1734349145b1f1dee98904ac3aa1428a/v22.1/cost-based-optimizer.md#L190)):*
> ```Markdown
> You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html). If `disallow_full_table_scans` is `true` and you set the `large_full_scan_rows` [session variable](set-vars.html). If no alternative plan is possible, an error will be returned.
> 
> If `disallow_full_table_scans` is `true` and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you don't provide an index hint, the full scan will be logged and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
> ```

nit: This is slightly confusing the way it's worded: "If you don't provide an index hint, the full scan will be logged and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated."
I think you should make it clear that an error will still be returned in this case, but the full scan will *also* be logged and the metric will be updated. As written, it sounds like we return an error in the force index case, but not otherwise.

rytaft (approved):
:lgtm:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14590#-:-N7bXhqJ0fKHOMs9ioOj:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

v21.2/cost-based-optimizer.md
```diff
@@ -102,6 +102,16 @@ SET CLUSTER SETTING sql.stats.histogram_collection.enabled = false;
 
 When `sql.stats.histogram_collection.enabled` is set to `false`, histograms are never collected, either as part of automatic statistics collection or by manually invoking [`CREATE STATISTICS`](create-statistics.html).
 
+## Control whether the optimizer creates a plan with a full scan
+
+Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may decide that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and determine that a statement plan isn't using the secondary index, it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
```
taroface, line 107:
```suggestion
Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may determine that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and find that a statement plan is not using the secondary index, then it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
```

taroface: I was slightly slow to parse the second sentence, so am making these suggestions for readability. Feel free to disregard. However, I feel that "determine" is better than "decide" because the latter can imply (at least to me) that the optimizer's choice might be arbitrary and/or incorrect.

v21.2/cost-based-optimizer.md
```diff
@@ -102,6 +102,16 @@ SET CLUSTER SETTING sql.stats.histogram_collection.enabled = false;
...
+
+You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html).
+
+If you disable full scans, you can set the `large_full_scan_rows` session variable to specify the maximum table size allowed for a full scan. If no alternative plan is possible, the optimizer will return an error.
+
+If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you don't provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
```
taroface, line 113:
```suggestion
If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you do not provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
```

taroface (changes requested):
Left a few suggestions.

v22.1/cost-based-optimizer.md
```diff
@@ -181,6 +181,16 @@ The optimizer uses `avg_size` to cost scans and relevant joins. Costing scans pe
 
 Cockroach Labs recommends that you allow the optimizer to consider column size when costing plans. If you are an advanced user and need to disable using `avg_size` for troubleshooting or performance tuning reasons, you can disable it by setting the `cost_scans_with_default_col_size` [session variable](set-vars.html) to true with `SET cost_scans_with_default_col_size=true`.
 
+## Control whether the optimizer creates a plan with a full scan
+
+Even if you have [secondary indexes](schema-design-indexes.html), the optimizer may decide that a full table scan will be faster. For example, if you add a secondary index to a table with a large number of rows and determine that a statement plan isn't using the secondary index, it is likely that performing a full table scan using the primary key is faster than doing a secondary index scan plus an [index join](indexes.html#example).
```
taroface, line 186: Sorry I somehow missed this - the suggestions from v21.2 should be applied here as well.

v22.1/cost-based-optimizer.md
```diff
@@ -181,6 +181,16 @@ The optimizer uses `avg_size` to cost scans and relevant joins. Costing scans pe
...
+
+You can disable statement plans that perform full table scans with the `disallow_full_table_scans` [session variable](set-vars.html).
+
+If you disable full scans, you can set the `large_full_scan_rows` session variable to specify the maximum table size allowed for a full scan. If no alternative plan is possible, the optimizer will return an error.
+
+If you disable full scans, and you provide an [index hint](table-expressions.html#force-index-selection), the optimizer will try to avoid a full scan while also respecting the index hint. If this is not possible, the optimizer will return an error. If you don't provide an index hint, the optimizer will return an error, the full scan will be logged, and the `sql.guardrails.full_scan_rejected.count` [metric](ui-overview-dashboard.html) will be updated.
```
taroface, line 192: Ditto

PR #14603, opened 2022-07-21
Add join example.

stbof:
DOC-3191


yuzefovich (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14603#-:-N7Wd1TD2m9snVWGLUky:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

v22.1/explain-analyze.md
```diff
@@ -212,6 +212,51 @@ For example, the following `EXPLAIN ANALYZE` statement executes a simple query a
 (30 rows)
 ~~~
 
+If you perform a join, the estimated max memory allocation is also reported for the join. For example:
```
taroface, line 215: Perhaps call out the `maximum memory usage` field and value in the following example? Or maybe just the value, if we think the field names might be subject to change.

taroface (approved):
One suggestion for clarity; otherwise LGTM.

v22.1/explain-analyze.md
stbof: The example is demonstrating that `estimated max memory allocated` was added to the `lookup join` component of the `EXPLAIN ANALYZE` output.

PR #14614, opened 2022-07-22
Rework known limitation to part of product docs

kathancox:
Fixes [DOC-4996](https://cockroachlabs.atlassian.net/browse/DOC-4996)

Removed restore historical data information from known limitation sections. This is understood as part of the product rather than a limitation. That is, it's just a fact of using restore on objects, therefore shouldn't be classified as a limitation.


v22.1/restore.md
```diff
@@ -92,6 +92,8 @@ You can restore:
 - [Databases](#databases)
 - [Tables](#tables)
 
+An object's historical data is not preserved upon `RESTORE`. This means that if you issue an `AS OF SYSTEM TIME` query on a restored object, the query will fail or the response will be incorrect because there is no historical data to query. For example, if you restore a table at `2022-07-13 10:38:00`, it is not then possible to read or [back up](backup.html) that table at `2022-07-13 10:37:00` or earlier. This is also the case for backups with [`revision_history`](backup.html#with-revision-history) that might try to initiate a revision start time earlier than `2022-07-13 10:38:00`.
+
```
msbutler, line 96: nit: I suspect a casual reader doesn't know what the term "historical data" means. Could it be worth saying something along the lines of: 

RESTORE will only restore the latest value of each key, or  the latest value AS OF SYSTEM TIME, if specified. In other words, historical data will not be restored even if the backup was specified with revision history.  

One consequence of this is that you cannot issue an AS OF SYSTEM TIME query on a restored object, the query will fail or the response will be incorrect because there is no historical data to query. For example, if you restore a table at `2022-07-13 10:38:00`, it is not then possible to read or [back up](backup.html) that table at `2022-07-13 10:37:00` or earlier.

kathancox: That's definitely something to consider and I'll adjust to make that a little more clear. We do use "historical data" quite often through our docs as part of backup, restore, and AS OF SYSTEM TIME docs. 

I think I'll need to reframe "key" in your example here to something more generic like data (or something more specific like a table in an example). We haven't typically referred to keys in this way, unless in arch docs. 

I'll make some adjustments and push another commit shortly 👍

v21.1/restore.md
```diff
@@ -102,7 +102,7 @@ You can restore:
...
+`RESTORE` will only restore the latest value of an object (table, database, cluster), or the latest value as per an [`AS OF SYSTEM TIME` restore](#as-of-system-time). That is, a restore will not include historical data even if you ran your backup with [`revision_history`](backup.html#with-revision-history). This means that if you issue an `AS OF SYSTEM TIME` query on a restored object, the query will fail or the response will be incorrect because there is no historical data to query. For example, if you restore a table at `2022-07-13 10:38:00`, it is not then possible to read or [back up](backup.html) that table at `2022-07-13 10:37:00` or earlier. This is also the case for backups with `revision_history` that might try to initiate a revision start time earlier than `2022-07-13 10:38:00`.
```
msbutler, line 106: nit: I'd change "latest value on" to "latest data in". Other than that. looks good to me!

kathancox: Cool, thanks for re-reviewing so quickly and your suggestions Michael! I'll make that change now.

msbutler (approved):

andf-crl (approved):
:lgtm:  LGTM!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14614#-:-N7wAaVw8WaxgsMzs1RC:b-di8w4)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @andf-crl, @kathancox, and @msbutler)

___
*[`v22.1/restore.md` line 96 at r1](https://reviewable.io/reviews/cockroachdb/docs/14614#-N7aUqR8fvxizjHCk88_-r1-96:-N7wAKeVB2OFj2QRU7Ra:b1ll4e8) ([raw file](https://github.com/cockroachdb/docs/blob/e6c92b5604f7c1ccbf445274e3e156a6884e255a/v22.1/restore.md#L96)):*

That's definitely something to consider and I'll adjust to make that a little more clear. We do use "historical data" quite often through our docs as part of backup, restore, and AS OF SYSTEM TIME docs. 

I think I'll need to reframe "key" in your example here to something more generic like data (or something more specific like a table in an example). We haven't typically referred to keys in this way, unless in arch docs. 

I'll make some adjustments and push another commit shortly 👍 
</blockquote></details>

Nice work, I think this paragraph reads quite clearly!

PR #14638, opened 2022-07-26
add upgrade paths and release terminology

taroface:
Fixes DOC-524.

```
Realizing that this draft does not yet work if:

- The docs version corresponds to a testing release (i.e., the docs are being viewed on DEV before the major release is GA)
- The docs version is the initial major release (i.e., with the `0` patch number). 

I'll account for these in a new commit.
```

EDIT: Above should be handled correctly now.

The logic in this PR is summarized here: https://cockroachlabs.atlassian.net/wiki/spaces/~5d815e4401e2cb0c301faf7e/blog/2022/08/03/2665447546/Dynamic+upgrade+docs


taroface:
@nickvigilante Would you mind checking over the liquid syntax I used here? I added a few since we last spoke.

nickvigilante (approved):
Checked all the Liquid in the v21.1 version and it looked good to me!

nickvigilante:
> EDIT: I just noticed that the syntax to pull the given upgrade version is `page.release_info.version`. I could use your help in determining whether this will always produce the same value as `latest_production.version`, which I added to use. If so, I should probably just stick with one of these.

Both `page.release_info.version` and `latest_production.version` should produce the same value. The former gets fed from `_config_base.yml` and uses the version plug-in in the `_plugins` folder. the latter is calculated based on the data in the `_data` folder. Either can be used, although I eventually want to retire the need for the version plug-in. However, I opt for less code where possible, so I suggest using `page.release_info.version`.

taroface:
@BramGruneir I've written this blog to explain what I'm doing with the variables in this PR: https://cockroachlabs.atlassian.net/wiki/spaces/~5d815e4401e2cb0c301faf7e/blog/2022/08/03/2665447546/Dynamic+upgrade+docs

Let me know whether this fulfills what you had in mind for an upgrade path!

taroface:
@BramGruneir friendly ping for review!

kathancox (approved):
Phew! That's a lot 22.1 22.2 21.2, but it all makes sense and LGTM!

PR #14641, opened 2022-07-27
DOC-4573 v22.1.5 release notes

andf-crl:
Addresses: DOC-4573

- Release notes for v22.1.5

Guess: the following selected changelog entries might not be intended for end users:

```
- Fixed the following builtins so that users can only run them if they have [`SELECT`](../v22.1/selection-queries.html) privileges on the relevant tables: `crdb_internal.revalidate_unique_constraints_in_all_tables`, `crdb_internal.revalidate_unique_constraints_in_table`, and `crdb_internal.revalidate_unique_constraint`. [#84269][#84269]
- `crdb_internal.decode_plan_gist` will no longer produce an internal error when it is used to decode a plan gist for which no schema information is available. [#84726][#84726]
- The [sampled query telemetry log] ... (x2) (as Telemetry channel is described as "Telemetry events for internal usage")
- Introduced a new `troubleshooting_mode_enabled` ... (telemetry data also)
- In case client-provided session parameters fail to parse ... (engineer changelog clearly not intended for non-engineer consumption)
```

Should I remove some or all of these?

[v22.1.5.md](https://deploy-preview-14641--cockroachdb-docs.netlify.app/docs/releases/v22.1.html#v22-1-5)

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -91,10 +91,10 @@ release_info:
     start_time: 2022-06-29 11:07:02.191084 +0000 UTC
     version: v21.2.13
   v22.1:
-    build_time: 2022-07-19 00:00:00 (go1.17)
+    build_time: 2022-07-28 00:00:00 (go1.17)
     docker_image: cockroachdb/cockroach
-    name: v22.1.4
-    start_time: 2022-07-18 12:08:36.898533 +0000 UTC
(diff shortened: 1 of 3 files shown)


MichaelTrestman (approved):
LGTM!

PR #14642, opened 2022-07-27
Clarify that outbox use case for TTL requires CDC

rmloveland:
Fixes DOC-4973


otan (approved):

ianjevans (approved):
Should these changes be brought forward to v22.2 as well?

PR #14643, opened 2022-07-27
Update for IN SCHEMA.

stbof:
DOC-5110

### DIFF
--- a/v22.1/alter-default-privileges.md
+++ b/v22.1/alter-default-privileges.md
@@ -32,8 +32,7 @@ Parameter | Description
 `FOR ALL ROLES` | Alter the default privileges on objects created by all users/roles.
 `GRANT ...` | Grant a default privilege or list of privileges on all objects of the specified type to a role/user, or a list of roles/users.
 `REVOKE ...` | Revoke a default privilege or list of privileges on all objects of the specified type from a role/user, or a list of roles/users.
-`IN SCHEMA qualifiable_schema_name` | **New in v22.1:** If specified, the default privileges are altered for objects created in that schema. If an object has default privileges specified at
-the database and at the schema level, the union of the default privileges is taken.
+`IN SCHEMA qualifiable_schema_name` | **New in v22.1:** If specified, the default privileges are altered for objects created in that schema. If an object has default privileges specified at the database and at the schema level, the union of the default privileges is taken.
 
 {{site.data.alerts.callout_info}}
 If you do not specify a `FOR ...` clause, CockroachDB alters the default privileges on objects created by the current user.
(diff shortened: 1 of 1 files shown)


RichardJCai (approved):

v22.1/alter-default-privileges.md
```diff
@@ -32,7 +32,8 @@ Parameter | Description
 `FOR ALL ROLES` | Alter the default privileges on objects created by all users/roles.
 `GRANT ...` | Grant a default privilege or list of privileges on all objects of the specified type to a role/user, or a list of roles/users.
 `REVOKE ...` | Revoke a default privilege or list of privileges on all objects of the specified type from a role/user, or a list of roles/users.
-`IN SCHEMA qualifiable_schema_name` | This clause is a no-op. The syntax is supported for compatibility with PostgreSQL.
+`IN SCHEMA qualifiable_schema_name` | **New in v22.1:** If specified, the default privileges are altered for objects created in that schema. If an object has default privileges specified at
```
taroface, line 35:
```suggestion
`IN SCHEMA qualifiable_schema_name` | **{% include_cached new-in.html version="v22.1" %}** If specified, the default privileges are altered for objects created in that schema. If an object has default privileges specified at
```

v22.1/alter-default-privileges.md
```diff
@@ -32,7 +32,8 @@ Parameter | Description
...
+the database and at the schema level, the union of the default privileges are taken.
```
taroface, line 36: See suggestion below; also, it looks like there is an extra line break above this block.

```suggestion
the database and at the schema level, the union of the default privileges is taken.
```

taroface (changes requested):
Couple nits and suggesting use of a liquid tag, which I just discovered while editing the style guide: https://github.com/cockroachdb/docs/blob/master/StyleGuide.md#version-tags

v22.1/alter-default-privileges.md
stbof: The Liquid tag doesn't work in tables.

taroface: Ah I overlooked this. But it seems `<span class="version-tag">New in vXX.Y:</span>` is the guidance for tables.

taroface:
@stbof Please see my reply about using the `span` tag for the "New in {version}" text. Also, the unwanted line break still appears to be here:

image

v22.1/alter-default-privileges.md
stbof: At one point we were using the format for version tags that's in this PR and it's also used in a number of other files. Since we have to do a global S&R for all those instances, I'd like to let this one go and fix them all at the same time. I'd also like to revisit the style for tables and plan to bring it up in the Style team. I think the `<span ...` is overkill.

taroface: Works for me!

stbof:
> @stbof Please see my reply about using the `span` tag for the "New in {version}" text. Also, the unwanted line break still appears to be here:
> 
> image

Fixed.

taroface (approved):

PR #14654, opened 2022-07-28
Added bullet to avro limitations on byte changes

kathancox:
Fixes [DOC-3651](https://cockroachlabs.atlassian.net/browse/DOC-3651)

Added a bullet to the Avro limitations outlining how byte types can be changed by Avro clients downstream.


_includes/v22.1/cdc/avro-limitations.md
```diff
@@ -1,3 +1,7 @@
+- [Decimals](decimal.html) must have precision specified.
+- [`BYTE`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) (or its alias `BYTEA`) is often used to store machine-readable data. When you stream these types through a changefeed with [`format=avro`](create-changefeed.html#format), CockroachDB does not encode or change the data. However, Avro clients can often include escape sequences so the data is in a printable format. A potential solution, is to hex-encode `BYTEA` values when initially inserting them into CockroachDB. This will ensure that you can consistently decode the hexadecimal. Note that this process will increase record size.
```
HonoreDB, line 2: The CockroachDB column type affected is BYTES, as well as its aliases BYTEA and BLOB. These all get encoded as the Avro type BYTES. Note that unlike other types in SQL or Avro, BYTES is pluralized--there's no singular BYTE type.

Otherwise, technical details LGTM. Suggest removing the comma after "a potential solution." Also, I feel like maybe it'd read better with some kind of description of why the escape sequences and printable format can be a problem requiring a solution, like "...in a printable format, which can interfere with deserialization."

kathancox: OK, I've edited the types + link here — do you think the small adjustment I have made reflects your point here? Yes, thanks for catching the typos on bytes and the errant comma there. Incorporated your other suggestions. I also added that mapping to the table as it wasn't there. Alphabetized while I was in the area.

_includes/v21.2/cdc/avro-limitations.md
```diff
@@ -1,3 +1,7 @@
+- [Decimals](decimal.html) must have precision specified.
+- [`BYTES`](bytes.html) (or its aliases `BYTEA` and `BLOB`) is often used to store machine-readable data. When you stream these types through a changefeed with [`format=avro`](create-changefeed.html#format), CockroachDB does not encode or change the data. However, Avro clients can often include escape sequences so the data is in a printable format, which can interfere with deserialization. A potential solution is to hex-encode `BYTEA` values when initially inserting them into CockroachDB. This will ensure that you can consistently decode the hexadecimal. Note that this process will increase record size.
```
andf-crl, line 2: > ... is often used to store machine-readable data. When you stream these types ...

nit: `is often` and `these types` should match as either singular or plural.

> ... escape sequences so the data is  ...

nit: `so the data is` --> `to present the data`

> ... hex-encode `BYTEA` values ...

`BYTEA` --> `BYTES` or `BYTES (or its aliases BYTEA and BLOB)` -- to match the main variant we present initially in this UL entry.

> This will ensure that you can consistently decode the hexadecimal.

nit: `you` --> `Avro clients`

> Note that this process will increase record size


optional nit: "this process" could potentially mean using `BYTES` in general. Tiny clarification could sort: `this process` --> `hex-encoding values at insertion`, or something similar. Very nit, very optional.

v21.2/use-changefeeds.md
```diff
@@ -197,35 +197,33 @@ The following sections provide information on Avro usage with CockroachDB change
...
 
 Below is a mapping of CockroachDB types to Avro types:
 
 CockroachDB Type | Avro Type | Avro Logical Type
 -----------------+-----------+---------------------
-[`INT`](int.html) | [`LONG`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
+[`ARRAY`](array.html) | [`ARRAY`](https://avro.apache.org/docs/1.8.1/spec.html#schema_complex) |
+[`BIT`](bit.html) | Array of [`LONG`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
 [`BOOL`](bool.html) | [`BOOLEAN`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
+[`BYTES`](bytes.html) | [`BYTES`](https://avro.apache.org/docs/1.8.1/spec.html#schema_primitive) |
```
andf-crl, line 211: Question: would it be useful to include `BYTEA` and `BLOB` here in some fashion, somewhat similar to `TIMESTAMP` and `TIMESTAMPZ`? Maybe, as aliases rather than a variant, same line unlinked is more appropriate? Unsure.

```
[`BYTES`](bytes.html), `BYTEA`, `BLOB` | [`BYTES`](avrolink)
```

Love the alpha reordering, thank you!

andf-crl (approved):
LGTM with small language suggestions. Thanks!

v21.2/use-changefeeds.md
kathancox: Cool nice changes! I think it makes sense to just additional lines for the aliases like the time related types. TFTR!

PR #14662, opened 2022-07-28
Add OID type.

stbof:
DOC-3383

### DIFF
--- a/v22.1/data-types.md
+++ b/v22.1/data-types.md
@@ -24,6 +24,7 @@ Type | Description | Example
 [`INT`](int.html) | A signed integer, up to 64 bits. | `12345`
 [`INTERVAL`](interval.html) | A span of time.  | `INTERVAL '2h30m30s'`
 [`JSONB`](jsonb.html) | JSON (JavaScript Object Notation) data.  | `'{"first_name": "Lola", "last_name": "Dog", "location": "NYC", "online" : true, "friends" : 547}'`
+[`OID`](oid.html) | An unsigned 32 bit integer. | `4294967295`
 [`SERIAL`](serial.html) | A pseudo-type that combines an [integer type](int.html) with a [`DEFAULT` expression](default-value.html).  | `148591304110702593`
 [`STRING`](string.html) | A string of Unicode characters. | `'a1b2c3'`
 [`TIME`<br>`TIMETZ`](time.html) | `TIME` stores a time of day in UTC.<br> `TIMETZ` converts `TIME` values with a specified time zone offset from UTC. | `TIME '01:23:45.123456'`<br> `TIMETZ '01:23:45.123456-5:00'`
--- a/v22.1/data-types.md
+++ b/v22.1/data-types.md
@@ -24,7 +24,7 @@ Type | Description | Example
 [`INT`](int.html) | A signed integer, up to 64 bits. | `12345`
 [`INTERVAL`](interval.html) | A span of time.  | `INTERVAL '2h30m30s'`
 [`JSONB`](jsonb.html) | JSON (JavaScript Object Notation) data.  | `'{"first_name": "Lola", "last_name": "Dog", "location": "NYC", "online" : true, "friends" : 547}'`
-[`OID`](oid.html) | An unsigned 32 bit integer. | `4294967295`
+[`OID`](oid.html) | An unsigned 32-bit integer. | `4294967295`
 [`SERIAL`](serial.html) | A pseudo-type that combines an [integer type](int.html) with a [`DEFAULT` expression](default-value.html).  | `148591304110702593`
 [`STRING`](string.html) | A string of Unicode characters. | `'a1b2c3'`
 [`TIME`<br>`TIMETZ`](time.html) | `TIME` stores a time of day in UTC.<br> `TIMETZ` converts `TIME` values with a specified time zone offset from UTC. | `TIME '01:23:45.123456'`<br> `TIMETZ '01:23:45.123456-5:00'`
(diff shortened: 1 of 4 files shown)


v22.1/oid.md
```diff
@@ -0,0 +1,19 @@
+---
+title: OID
+summary: The OID data type stores an unsigned 32 bit integer value.
+toc: true
+docs_area: reference.sql
+---
+
+The `OID` (Object Identifier) [data type](data-types.html) stores an unsigned 32 bit integer value. OIDs are used internally as primary keys for tables in [system schema](system-catalogs.html). OIDs are not used in user-created tables.
```
ajwerner, line 8: sadly I think `OIDs` can be used in user-created tables.
```
create table baz (k oid primary key);
CREATE TABLE
```

Our OIDs are very similar to postgres's. There are "subtypes" of oids which correspond to different types. It's rather frustratingly complex. https://www.postgresql.org/docs/14/datatype-oid.html

One note is that pg_catalog is synthesized in cockroach whereas it's a real table in postgres. The structure of our OIDs is bespoke for each of the subtypes. We can omit that for now. I'm fine keeping this doc vague, but it is an area which could use further expansion.

To clarify what I mean, a table has a `regclass`. One can cast a string name to a regclass: 

```
CREATE TABLE foo (i INT);
SELECT 'foo'::regclass;
SELECT 'foo'::regclass::oid;
```

```
CREATE TABLE
  regclass
------------
  foo
(1 row)
  oid
-------
  116
(1 row)
```

ajwerner (commented):
I'm afraid that my comment is only going to muddy the waters.

v22.1/oid.md
stbof: @ajwerner How do you want to move this forward?  Shall I just remove the sentence `OIDs are not used in user-created tables.`? What use cases should I document?

postamar: I suggest something along the lines of `OIDs should not be used in user-created tables. Values of this type are not guaranteed to be stable across major releases`.

rafiss: i think we should also mention that users should not rely on OIDs to be globally unique. each OID subtype is only unique within certain namespace, and it probably isn't worth getting into all the details of that.

stbof:
@ajwerner  Please review use cases and limitations.

ajwerner (approved):
:lgtm:

I like it! Provides useful information and avoids going into the weeds. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14662#-:-N8VP9oqC4sashwN8JJs:bn2xozx)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @ajwerner, @postamar, and @rafiss)

v22.1/data-types.md
```diff
@@ -24,6 +24,7 @@ Type | Description | Example
 [`INT`](int.html) | A signed integer, up to 64 bits. | `12345`
 [`INTERVAL`](interval.html) | A span of time.  | `INTERVAL '2h30m30s'`
 [`JSONB`](jsonb.html) | JSON (JavaScript Object Notation) data.  | `'{"first_name": "Lola", "last_name": "Dog", "location": "NYC", "online" : true, "friends" : 547}'`
+[`OID`](oid.html) | An unsigned 32 bit integer. | `4294967295`
```
taroface, line 27:
```suggestion
[`OID`](oid.html) | An unsigned 32-bit integer. | `4294967295`
```

v22.1/oid.md
```diff
@@ -0,0 +1,29 @@
+---
+title: OID
+summary: The OID data type stores an unsigned 32 bit integer value.
```
taroface, line 3:
```suggestion
summary: The OID data type stores an unsigned 32-bit integer value.
```

v22.1/oid.md
```diff
@@ -0,0 +1,29 @@
...
+toc: true
+docs_area: reference.sql
+---
+
+The `OID` (Object Identifier) [data type](data-types.html) stores an unsigned 32 bit integer value.
```
taroface, line 8:
```suggestion
The `OID` (Object Identifier) [data type](data-types.html) stores an unsigned 32-bit integer value.
```

taroface: Does "Object Identifier" need to be capitalized? Postgres docs imply no: https://www.postgresql.org/docs/current/datatype-oid.html

v22.1/oid.md
```diff
@@ -0,0 +1,29 @@
...
+
+OIDs are used internally as primary keys for tables in [system catalogs](system-catalogs.html), notably `information_schema` and `pg_catalog`.
+
+CockroachDB supports many [functions](functions-and-operators.html#built-in-functions) that accept OIDs as argument types and return OIDs, and [operators](functions-and-operators.html#operators) that operate on OIDs and other data types. These functions are used by are used by drivers and ORMs and you can use these functions to introspect schema.
```
taroface, line 12:
```suggestion
CockroachDB supports many [functions](functions-and-operators.html#built-in-functions) that accept OIDs as argument types and return OIDs, and [operators](functions-and-operators.html#operators) that operate on OIDs and other data types. These functions are used by drivers and ORMs and you can use these functions to introspect your schema.
```

taroface (approved):
Looks good; a few nits.

v22.1/oid.md
```diff
@@ -0,0 +1,29 @@
+---
+title: OID
+summary: The OID data type stores an unsigned 32-bit integer value.
+toc: true
+docs_area: reference.sql
+---
+
+The object identifier (OID) [data type](data-types.html) stores an unsigned 32-bit integer value.
```
stbof, line 8:
```suggestion
The object identifier (`OID`) [data type](data-types.html) stores an unsigned 32-bit integer value.
```

PR #14664, opened 2022-07-29
Update alter-column.md

BazenGabrakirstos:
added comma between double arrows


nickvigilante:
@BazenGabrakirstos Unfortunately, this syntax causes the Liquid parser to fail, so I've closed this PR. Thank you anyway!

PR #14668, opened 2022-07-29
Update unsupported-postgres-features.md

stbof:
Move reason and workaround closer to limitation so it's not missed.

### DIFF
--- a/_includes/v22.1/sql/unsupported-postgres-features.md
+++ b/_includes/v22.1/sql/unsupported-postgres-features.md
@@ -1,10 +1,14 @@
 - Stored procedures and functions.
 - Triggers.
 - Events.
-- User-defined functions (UDF).
-- FULLTEXT functions and indexes.
-- Drop primary key. Each table must have a primary key associated with it. You can [drop and add a primary key constraint within a single transaction](drop-constraint.html#drop-and-add-a-primary-key-constraint).
-- XML Functions.
+- User-defined functions (UDFs).
+- `FULLTEXT` functions and indexes.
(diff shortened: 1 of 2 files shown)


_includes/v22.1/sql/unsupported-postgres-features.md
```diff
@@ -1,12 +1,10 @@
...
-- XA syntax
-- Creating a database from a template
-
-<a name="drop-primary-key"></a> * Each table must have a primary key associated with it. You can, however, [drop and add a primary key constraint within a single transaction](drop-constraint.html#drop-and-add-a-primary-key-constraint).
+- Stored procedures and functions.
+- Triggers.
+- Events.
+- User-defined functions (UDF).
+- FULLTEXT functions and indexes.
+- Drop primary key. Each table must have a primary key associated with it. You can [drop and add a primary key constraint within a single transaction](drop-constraint.html#drop-and-add-a-primary-key-constraint).
```
taroface, line 6: I would personally use this approach, assuming this renders correctly. I think it's clearest if this bulleted list is simply a list of features, and further explanation is kept separate.

```suggestion
- Drop primary key. 
  {{site.data.alerts.callout_info}}
  Each table must have a primary key associated with it. You can [drop and add a primary key constraint within a single transaction](drop-constraint.html#drop-and-add-a-primary-key-constraint).
  {{site.data.alerts.end}}
```

taroface (commented):
Just 1 suggestion, but your call

taroface (approved):
LGTM!

PR #14700, opened 2022-08-02
Added note to client key description on encrypted key error

kathancox:
Fixes [DOC-3761](https://cockroachlabs.atlassian.net/browse/DOC-3761)

Added a note to warn users that client keys can potentially be encrypted. If the key isn't decrypted before passing, they will receive an error. Inclued in changefeed docs v21.1 – v22.1 for the `client_key` param.


kathancox:
Here is a preview for the 22.1 docs: https://deploy-preview-14700--cockroachdb-docs.netlify.app/docs/stable/changefeed-sinks.html#kafka

amruss (approved):
LGTM

andf-crl (approved):
LGTM: with one suggestion:

- the decrpyt command changes the filename of the key we use in our changefeed statement from `key.pem` to `key.decrypt.pem`. You definitely don't want to directly clobber `key.pem` with the same name (i.e. overwrite), so probably best here to just note something at the end, maybe:

> Once decrypted, be sure to update your changefeed statement to use the new `key.decrpyt.pem` file instead.

PR #14710, opened 2022-08-03
Expand on why you can cancel alter type that drops enum values.

stbof:
DOC-4586

### DIFF
--- a/v22.1/alter-type.md
+++ b/v22.1/alter-type.md
@@ -10,7 +10,8 @@ The `ALTER TYPE` [statement](sql-statements.html) modifies a user-defined, [enum
 {% include {{ page.version.version }}/misc/schema-change-stmt-note.md %}
 
 {{site.data.alerts.callout_info}}
-You can only [cancel](cancel-job.html) `ALTER TYPE` [schema change jobs](online-schema-changes.html) that drop values. All other `ALTER TYPE` schema change jobs are non-cancellable.
+- When you drop a value, CockroachDB searches through all of the rows that could contain the enum value, which could take a long time. Therefore, you can [cancel](cancel-job.html) `ALTER TYPE` [schema change jobs](online-schema-changes.html) that drop values.
+- All other `ALTER TYPE` schema change jobs are non-cancellable.
 {{site.data.alerts.end}}
 
 ## Synopsis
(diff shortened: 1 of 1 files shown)


postamar (approved):
LGTM

v22.1/alter-type.md
```diff
@@ -10,7 +10,8 @@ The `ALTER TYPE` [statement](sql-statements.html) modifies a user-defined, [enum
 {% include {{ page.version.version }}/misc/schema-change-stmt-note.md %}
 
 {{site.data.alerts.callout_info}}
-You can only [cancel](cancel-job.html) `ALTER TYPE` [schema change jobs](online-schema-changes.html) that drop values. All other `ALTER TYPE` schema change jobs are non-cancellable.
+- When you drop a value, CockroachDB searches through all of the rows that could contain the enum value, which could take a long time. Therefore, you can [cancel](cancel-job.html) `ALTER TYPE` [schema change jobs](online-schema-changes.html) that drop values.
```
taroface, line 13:
```suggestion
You can only [cancel](cancel-job.html) `ALTER TYPE` [schema change jobs](online-schema-changes.html) that drop values. This is because when you drop a value, CockroachDB searches through every row that could contain the `ENUM` value, which could take a long time.

All other `ALTER TYPE` schema change jobs are non-cancellable.
```

taroface: Making two suggestions:

- Leading with the key point, as the original version did, that only `ALTER TYPE` jobs that drop values are cancellable. It is less intuitive IMO to lead with the explanation about CRDB searching all rows.
- Removing the bullet points - I usually avoid these in callouts, especially without leading text that is not in a list, because it looks visually off (and because the combo of callout box + bullet points somehow sends mixed signals). This can be up for discussion in style committee.

v22.1/alter-type.md
```diff
@@ -10,7 +10,8 @@ The `ALTER TYPE` [statement](sql-statements.html) modifies a user-defined, [enum
...
+- All other `ALTER TYPE` schema change jobs are non-cancellable.
```
taroface, line 14:
```suggestion
```

taroface (changes requested):
Suggesting an edit

taroface (approved):

PR #14711, opened 2022-08-04
Correct alpha order of SQL statements

kathancox:
This is a quick fix to the alphabetical order of the SQL statements. Seems like there was some sort of clash between PRs as the `ADD SUPER REGION` docs were added that changed the ordering on `ALTER BACKUP`.


andf-crl (approved):
LGTM, thanks!

PR #14716, opened 2022-08-04
use liquid syntax in snippets

taroface:
Fixes [DOC-2815](https://cockroachlabs.atlassian.net/browse/DOC-2815).

### DIFF
--- a/v21.2/node-shutdown.md
+++ b/v21.2/node-shutdown.md
@@ -395,11 +395,11 @@ cockroach node status --decommission --certs-dir=certs --host={address of any li
 ~~~
 
 ~~~
-  id |     address     |   sql_address   |              build              |         started_at         |         updated_at         | locality | is_available | is_live | gossiped_replicas | is_decommissioning |   membership   | is_draining
------+-----------------+-----------------+---------------------------------+----------------------------+----------------------------+----------+--------------+---------+-------------------+--------------------+----------------+--------------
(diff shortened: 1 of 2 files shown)


taroface:
@kathancox I'm actually not 100% sure about this solution, as it required manually reformatting the table column width in the output snippets. Until we have a more advanced solution, I sort of like knowing that a snippet reflects the actual output of the version it was tested on. However, I understand wanting not to show a testing version, or a version that mismatches the actual page being viewed. Wondering if you have any thoughts on this.

kathancox:
> @kathancox I'm actually not 100% sure about this solution, as it required manually reformatting the table column width in the output snippets. Until we have a more advanced solution, I sort of like knowing that a snippet reflects the actual output of the version it was tested on. However, I understand wanting not to show a testing version, or a version that mismatches the actual page being viewed. Wondering if you have any thoughts on this.

@taroface Interesting! It feels "neater" with the liquid, but what is the motivation behind changing that [discussed]? I kind of assumed sort of like the switchover to the new version (v22.2 coming up would have `v22.1.blah` in the output). But, as you say, what happens if something changes in a minor version and the command works differently/there is a bug... the output will imply that it has been tested and works fine on that version. 

Keeping these static might be more transparent for the user — the version in the output and such, is the version that has been tested by one of the docs team.

There are probably some edge case benefits and cons that I'm missing here. 

Update after talking to @taroface, this could be a good team/style team discussion point.

kathancox (approved):
LGTM!

PR #14717, opened 2022-08-04
Update the splash screen for new versions of docs

nickvigilante:
When a new version of the docs is cut, this page will be prepared
to handle users who may try to manually access this URL in between
when the docs are cut and the first alpha becomes available.

### DIFF
--- a/releases/v22.1.md
+++ b/releases/v22.1.md
@@ -30,5 +30,5 @@ To upgrade to {{ page.major_version }}, see [Upgrade to CockroachDB {{ page.majo
 {% endfor %}
 
 {% else %}
-Currently, no releases are available for this version.
+No releases are available for this version. See the [Releases](https://www.cockroachlabs.com/docs/releases/index.html) page for all available releases.
 {% endif %}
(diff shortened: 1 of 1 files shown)


releases/v22.1.md
```diff
@@ -27,3 +28,7 @@ To upgrade to {{ page.major_version }}, see [Upgrade to CockroachDB {{ page.majo
 {% for r in rel %}
 {% include releases/{{ page.major_version }}/{{ r.version }}.md release=r.version %}
 {% endfor %}
+
+{% else %}
+Currently, no releases are available for this version.
```
mikeCRL, line 33:
```suggestion
No releases are available for this version. See the [Releases](https://www.cockroachlabs.com/docs/releases/index.html) page for all available releases.
```

May as well link out, I suppose, though this should be sufficient either way.

mikeCRL (approved):
LGTM with one idea as a suggestion.

releases/v22.1.md
nickvigilante: Agreed. Done.

PR #14726, opened 2022-08-05
Add Explain Alter example.

stbof:
DOC-2191

### DIFF
--- a/v22.1/explain.md
+++ b/v22.1/explain.md
@@ -333,6 +333,28 @@ Because the `INSERT` includes an `ON CONFLICT` clause, the query requires more t
 
 ### Alter queries
 
+If you alter a table to split a range as described in [Split a table](split-at.html#split-a-table), the `EXPLAIN` command returns the target table and index names and a `NULL` expiry timestamp:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+EXPLAIN ALTER TABLE users SPLIT AT VALUES ('chicago'), ('new york'), ('seattle');
+~~~
+
+~~~
+               info
+----------------------------------
+  distribution: local
+  vectorized: true
+
+  • split
+  │ index: users@users_pkey
+  │ expiry: CAST(NULL AS STRING)
+  │
+  └── • values
+        size: 1 column, 3 rows
+(9 rows)
+~~~
+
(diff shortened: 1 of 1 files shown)


ajwerner (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14726#-:-N925W5T6Svr6iDZvhWT:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @stbof)

v22.1/explain.md
```diff
@@ -331,6 +331,52 @@ Time: 3ms total (execution 3ms / network 0ms)
 
 Because the `INSERT` includes an `ON CONFLICT` clause, the query requires more than a simple `insert` operation. CockroachDB must check the provided values against the values in the database, to ensure that the `UNIQUE` constraint on `name`, `city`, and `id` is not violated. The output also lists the indexes available to detect conflicts (the `arbiter indexes`), including the `users_city_id_name_key` index.
 
+### Alter queries
+
+If you alter a table to split a range as described in [Split a table](split-at.html#split-a-table), the `EXPLAIN` command returns the target table and index names and a `NULL` expiry timestamp:
```
taroface, line 336: This is a simpler example, but it doesn't address the concern I raised that this section "reads as though all of the information is only returned when a split is performed". To me, it would make sense to instead use an `ALTER TABLE` example that does not perform any split, and therefore doesn't return an `expiry` timestamp.

stbof: The Jira being addressed is the new output for `RELOCATE` and `SPLIT AT`: The output of `EXPLAIN ALTER INDEX/TABLE ... RELOCATE/SPLIT` now includes the target table/index name and, for the SPLIT AT variants, the expiry timestamp.

Since `RELOCATE` isn't being documented, this PR addresses the JIra.

stbof: I tried a few more `EXPLAIN ALTER` commands and they typically don't have much output. For example:

```
explain ALTER TABLE users ADD COLUMN id UUID NOT NULL DEFAULT gen_random_uuid();
         info
-----------------------
  distribution: local
  vectorized: true

  • alter table
```

taroface: Oh, I was missing the context that `RELOCATE` is deprecated. In that case, sorry to hold this up and LGTM!

taroface (approved):

PR #14727, opened 2022-08-05
DOC-5249: Improve performance of filter tabs

nickvigilante:
Fixes DOC-5249

I reduced build time for the filter tabs by 3.5 seconds.

### DIFF
--- a/_includes/cockroachcloud/filter-tabs/cloud-faqs.md
+++ b/_includes/cockroachcloud/filter-tabs/cloud-faqs.md
@@ -1,0 +1,4 @@
+{% assign tab_names_html = "CockroachDB Serverless;CockroachDB Dedicated" %}
+{% assign html_page_names = "serverless-faqs.html;frequently-asked-questions.html" %}
+
+{% include filter-tabs.md tab_names=tab_names_html page_names=html_page_names page_folder="cockroachcloud" %}
--- a/_includes/cockroachcloud/filter-tabs/cluster-management.md
+++ b/_includes/cockroachcloud/filter-tabs/cluster-management.md
@@ -1,0 +1,4 @@
+{% assign tab_names_html = "CockroachDB Serverless;CockroachDB Dedicated" %}
+{% assign html_page_names = "serverless-cluster-management.html;cluster-management.html" %}
+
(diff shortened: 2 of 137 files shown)


nickvigilante:
@mikeCRL I know this is a doozy of a PR, but if it helps you to test, only focus on clicking one of the links under each include modified in [this comment](https://github.com/cockroachdb/docs/pull/14727#issuecomment-1206809907). I just ran through a battery of tests myself and I did a review of every file manually, so I'm confident this will be fine, but let me know if you need help understanding this change.

mikeCRL (commented):
I know the goal was build efficiency, however I think this is also a more straightforward way to create/specify—and maintain visibility of the use of—tabs. Thank you, Nick.

I tested every page per your suggestion and found no issues.

Not a blocker for merging, but I'm wondering if it would be possible to avoid the need to hard code the following line in every filter tab include:
`{% include filter-tabs.md tab_names=tab_names_html page_names=html_page_names page_folder=page.version.version %}`

@lnhsingh given the significant change in how writers will specify filter tabs, can I ask for your sign-off on that aspect, and on implementing this without advance team review? I am OK with this.

lnhsingh:
@nickvigilante / @mikeCRL - this change LGTM, but I would feel more comfortable if one of the writers also signed off (@ianjevans can you take a quick look since you've used a lot of tabs in your docs?)

nickvigilante:
@mikeCRL I can't think of a good way to remove that line from each of the includes. The intermediate includes are meant to be uniform across each of the pages, and without copying/pasting the code from `filter-tabs.md` in each intermediate include, this was the best solution I found. However, I'm open to other ideas.

ianjevans:
I think there's a real advantage to using attributes to set the tab group details over the page front matter. It keeps the configuration in the filter include.

I definitely think we'll need a demo of this at the next docs weekly (and record it for people to check out later) so everyone knows how to use the new syntax.

ianjevans (approved):

nickvigilante:
@ianjevans More than happy to do so.

@lnhsingh @mikeCRL I won't be able to demo this before I cut the 22.2 docs but I will send a notification out today to the other writers and work on updating the Style Guide this afternoon.

mikeCRL (approved):
Thanks for the additional detail. Sounds good to me. Thanks, Nick.

PR #14751, opened 2022-08-09
Add note that CC admins receive alerts for minor upgrades

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-4978


rmloveland (approved):
LGTM

PR #14756, opened 2022-08-09
Document psycopg3 support.

ianjevans:
This PR documents how to use the psycopg3 driver with CockroachDB.

Closes DOC-4977.


kernfeld-cockroach:
These docs look good to me. A couple thoughts on the app code itself:

1. There is some commented-out code that should be removed.
2. It's best practice to include an exact version number of each dependency in `requirements.txt`.
3. It's not necessary to create an `ids` variable in `create_accounts`. You could just do `return [id1, id2]`.
4. I suspect the error handling can be simplified. Why do we need to swallow `ValueError`? Why is it beneficial to catch, log, and re-throw `Exception`? Why are we catching `psycopg.OperationalError` in only one place?
5. The sleep variable is named `sleep_ms` but I think it's actually measured in seconds.

ianjevans:
> These docs look good to me. A couple thoughts on the app code itself:
> 
> 1. There is some commented-out code that should be removed.
> 2. It's best practice to include an exact version number of each dependency in `requirements.txt`.
> 3. It's not necessary to create an `ids` variable in `create_accounts`. You could just do `return [id1, id2]`.
> 4. I suspect the error handling can be simplified. Why do we need to swallow `ValueError`? Why is it beneficial to catch, log, and re-throw `Exception`? Why are we catching `psycopg.OperationalError` in only one place?
> 5. The sleep variable is named `sleep_ms` but I think it's actually measured in seconds.

I think I addressed all of these in the latest commit in the code repo.

kernfeld-cockroach (approved):

_includes/v22.1/misc/tooling.md
```diff
@@ -23,6 +23,7 @@ For a list of tools supported by the CockroachDB community, see [Third-Party Too
 | Java | [JDBC](https://jdbc.postgresql.org/download.html#others) | {% remote_include https://raw.githubusercontent.com/cockroachdb/cockroach/master/pkg/cmd/roachtest/tests/pgjdbc.go ||var supportedPGJDBCTag = "||"\n\n %} | Full | N/A | [Build a Java App with CockroachDB (JDBC)](build-a-java-app-with-cockroachdb.html) |
 | JavaScript | [pg](https://www.npmjs.com/package/pg) | 8.2.1 | Full | N/A | [Build a Node.js App with CockroachDB (pg)](build-a-nodejs-app-with-cockroachdb.html) |
 | Python | [psycopg2](https://www.psycopg.org/docs/install.html) | 2.8.6 | Full | N/A | [Build a Python App with CockroachDB (psycopg2)](build-a-python-app-with-cockroachdb.html) |
+| Python | [psycopg3](https://www.psycopg.org/psycopg3/docs/) | 3.0.16 | Full | N/A | [Build a Python App with CockroachDB (psycopg3)](build-a-python-app-with-cockroachdb-psycopg3.html) |
```
gemma-shay, line 26: I think the two python cells should be combined like the Go example if possible

ianjevans:
@gemma-shay Can I get your review on this?

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Step 3. Install the Psycopg 3 driver
+
+`psycopg[binary]` is the sample app's only third-party module dependency.
+
+To install `psycopg[binary]`, run the following command:
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+$ pip install "psycopg[binary]"
```
gemma-shay, line 49: personally had to use `pip3` for this, not sure if that would be the case for others

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
```diff
@@ -0,0 +1,107 @@
...
+1. Run the code:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ cd example-app-python-psycopg3
+    ~~~
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    $ python example.py
```
gemma-shay, line 89: also had to use `python3` here

gemma-shay: second part of this comment didn't get posted for some reason, but I got this error when running the example: `CRITICAL:root:database connection failed
CRITICAL:root:invalid cast: uuid -> int`

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
kernfeld-cockroach: Good point. This would indeed be `pip3` on many systems; `pip` often still refers to the pip for Python 2.

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
ianjevans: You have an existing `accounts` table with the old schema. `DROP TABLE accounts` to clear it out and re-run.

ianjevans: > also had to use `python3` here

Done.

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
ianjevans: Done.

_includes/v22.1/misc/tooling.md
ianjevans: Done.

v22.1/build-a-python-app-with-cockroachdb-psycopg3.md
gemma-shay: ah good catch. lgtm!

gemma-shay (approved):

PR #14758, opened 2022-08-09
Clarify Support Policy for Cloud vs Self-Hosted

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-3007

Part of the ticket request was to put the self-hosted and cloud support policies on the same page, which I don't think would make sense. Confirmed instead that each support policy page specifies which product it's for and links to the other one.

### DIFF
--- a/cockroachcloud/upgrade-policy.md
+++ b/cockroachcloud/upgrade-policy.md
@@ -5,7 +5,7 @@ toc: true
 docs_area: manage
 ---
 
-This page describes the upgrade policy for {{ site.data.products.db }}.
+This page describes the upgrade policy for {{ site.data.products.db }}. For self-hosted clusters, see the CockroachDB [Release Support Policy](../releases/release-support-policy.html).
 
 {{ site.data.products.db }} supports the latest major version of CockroachDB and the version immediately preceding it. Support for these versions includes patch version updates and security patches.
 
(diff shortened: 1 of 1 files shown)


cockroachcloud/upgrade-policy.md
```diff
@@ -39,7 +39,7 @@ If you see problems after the upgrade has been finalized, it will not be possibl
 
 ### End of Support for older CockroachDB versions
 
-As CockroachDB releases new major versions, older versions reach their End of Support (EOS) on {{ site.data.products.db }}. A CockroachDB version reaches EOS when it is 2 major versions behind the latest version. For example, now that CockroachDB v21.2 has been released, CockroachDB v20.2 has reached EOS.
+As CockroachDB releases new major versions, older versions reach their End of Support (EOS) on {{ site.data.products.db }}. A CockroachDB version reaches EOS when it is two major versions behind the latest version. For example, now that CockroachDB v21.2 has been released, CockroachDB v20.2 has reached EOS.
```
rmloveland, line 42: Suggest also editing the last sentence in the paragraph to be past tense since those are pretty old versions

something like

"for example when v21.2 was released, v20.2 reached EOS"

rmloveland (approved):
LGTM but I suggest at least tagging Joel for a quick look as well since they filed the issue

DuskEagle (approved):
LGTM

PR #14786, opened 2022-08-11
Document s2_level_mod parameter.

stbof:
We have examples in https://www.cockroachlabs.com/docs/v22.1/spatial-indexes.html#create-a-spatial-index-that-uses-all-of-the-tuning-parameters that use this parameter, but no definition.


ajwerner (commented):
The code says `s2_level_mod` must be between 1 and 3. https://github.com/cockroachdb/cockroach/blob/965f7c68b70ee46383af1946622935b0c579adfe/pkg/sql/storageparam/indexstorageparam/index_storage_param.go#L118

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14786#-:-N9Xk5tk-8J2t_0wcaUj:b4t90ko)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)

ajwerner (approved):
Otherwise, the commentary LGTM and the existing condition is also enforced here: https://github.com/cockroachdb/cockroach/blob/965f7c68b70ee46383af1946622935b0c579adfe/pkg/sql/storageparam/indexstorageparam/index_storage_param.go#L147

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/14786#-:-N9XkNLB5c_cVWrAkIbd:b-wiqy7l)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @stbof)

PR #14796, opened 2022-08-11
Updated link text to Dedicated multi-region DB restore to be more descriptive

kathancox:
This is a quick fix to update the link text for the alternate workflow users must take to restore a multi-region database on a v21.2 Dedicated cluster.

### DIFF
--- a/cockroachcloud/backups-page.md
+++ b/cockroachcloud/backups-page.md
@@ -113,7 +113,7 @@ To restore a database:
 
     The **Restore database** module displays with backup details.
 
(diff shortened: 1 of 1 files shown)


andf-crl (commented):
Question: If I want to restore a multi-region database on v22.1 or later, which do I follow:
- `#restore-a-database`
- `#restore-a-multi-region-database-to-a-new-database-in-v21-2`

(users must choose based on whether they are v21.2 and also whether they are MR, which makes picking between the two presented options a bit tricky)

Also, it'd be preferable to preface both procedurals with a bit of intro to help the user choose which applies. Currently, the user must progress through method 1 steps 1-3 before seeing that they need to switch to method 2, where steps 1 and 2 are identical anyways. Perhaps a new tabset under `Restore a multi-region database` that includes parallel instructions for v21.2 and v22.1+? Not sure.

PR #14805, opened 2022-08-12
Developer vs sql admin clarification

gemma-shay:
DOC-2884

Developers have sql admin privs on a cluster. Calling this out under both the SQL User and Developer sections as it can be a security issue.

### DIFF
--- a/cockroachcloud/console-access-management.md
+++ b/cockroachcloud/console-access-management.md
@@ -36,7 +36,7 @@ Every {{ site.data.products.db }} user is either a Developer or a Console Admin 
 
 ### Developer
 
-A Developer is a limited-access role. A Developer cannot invite Team Members to the Console or create new SQL users. Note that Developers can still create [SQL Users](#sql-users) with the [`admin` role](../{{site.versions["stable"]}}/security-reference/authorization.html#admin-role) on a cluster.
+A Developer is a limited-access role. A Developer cannot invite Team Members to the Console or create new SQL users. Note that Developers can still create [SQL Users](#sql-users) with the [`admin` role](../{{site.versions["stable"]}}/security-reference/authorization.html#admin-role) on a cluster and therefore privileges for all resources across the cluster.
 
 To access a cluster, you need to ask a Console Admin for the username and password of a SQL user. To find out who your Console Admin is, check the **Access** page.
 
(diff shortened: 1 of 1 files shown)


kevin-v-ngo:
This looks fine to me Gemma! 

But is there another issue tracking a docs update to call out that Developer and Console Admins have access to all information surfaced on the SQL Activity and Database pages because they belong to the SQL admin role?

kevin-v-ngo:
Ah so this issue is meant to address this JIRA: https://cockroachlabs.atlassian.net/browse/DOC-2884

We should clarify and have something along the lines of,

Console Admins and Developers have access to all the information on the SQL Activity and Databases page in the CockroachCloud Console because they belong to the SQL admin role.

gemma-shay:
@kevin-v-ngo Does that last commit sound good?

cockroachcloud/console-access-management.md
```diff
@@ -32,11 +32,15 @@ Learn more about [managing SQL users' privileges](../{{site.versions["stable"]}}
 
 ## Roles
 
-Every {{ site.data.products.db }} user is either a Developer or a Console Admin for the organization.
+Every {{ site.data.products.db }} user is either a Developer or a Console Admin for the organization. 
+
+{{site.data.alerts.callout_danger}}
+Both Console Admins and Developers have access to all the information on the **SQL Activity** and **Databases** pages in the {{ site.data.products.db }} Console because they can create SQL users with the [`admin` role](../{{site.versions["stable"]}}/security-reference/authorization.html#admin-role).
```
kevin-v-ngo, line 38: It's not because they can create SQL users with the admin role that they have access to the pages. It's because they are accessing the pages **using the admin role**. And the admin role has the highest privileges such as viewing all pages in the console _and_ creating SQL users. 

We can remove everything after Console...

Or if we want to be specific, change it to, "because they are part of the SQL admin role."

kevin-v-ngo (commented):
Small suggestion

cockroachcloud/console-access-management.md
gemma-shay: gotcha, edited

kevin-v-ngo (approved):
Thanks for the changes Gemma!

rmloveland (approved):
LGTM!

PR #14836, opened 2022-08-16
feat: speed and performance part 1

angel-carvajal:
Feat: improve seo performance
- Defer all JS possible
- Move JS to inline script tag to avoid extra loading content
- change font-display to swap
- add whenAvailable method for defer activity


nickvigilante (changes requested):
In the latest build preview, the entire sidebar tree is expanded. We should only expand the sidebar at the particular file that is opened.

nickvigilante:
The sidebar loads between .5 seconds and 1 second after the rest of the page content. Both Mike and Lauren found this to be less than ideal. Is it possible to either load the sidebar at the very beginning of the page load or to have the sidebar and page content to load all at the same time? Anything to minimize the gap that the sidebar isn't rendered.

nickvigilante:
This looks MUCH better. Great work! I'm just gonna verify with Mike and Lauren and then we should be good to merge.

nickvigilante:
Also, can you add a sentence or two in the PR description outlining the changes you made?

PR #14837, opened 2022-08-16
DOC-5369: Cutting the v22.2 docs

nickvigilante:
Fixes DOC-5369


_config_base.yml
```diff
@@ -98,5 +95,14 @@ release_info:
     name: v22.1.5
     start_time: 2022-07-27 13:34:27.897839 +0000 UTC
     version: v22.1.5
+  v22.2:
+    build_time: 2022-08-16 10:15:48.318507 (go1.17)
+    docker_image: cockroachdb/cockroach-unstable
+    name: v22.2-alpha.1
+    start_time: 2022-08-16 10:15:48.318607 +0000 UTC
+    version: v22.2-alpha.1
```
mikeCRL: What's the impact of including v22.2 generally, and specifically `v22.2-alpha.1` prior to its release?

Also, not sure if related, but I am seeing an empty v22.2 section here
https://deploy-preview-14837--cockroachdb-docs.netlify.app/docs/releases/index.html

Shouldn't we disable that until the alpha release? Should we be moving some steps to the first alpha release notes PR (or a separate PR after this one but before that one)?

andf-crl (approved):
LGTM after both a diff spot check and a staged HTML spot check. Especially looked for:
- remote_includes
- upgrade guides
- references into and out of out-of-band cockroachcloud dir
- releases stub page and all releases listing (thanks for clarifying that v22.2 is expected to be missing from `releases > CockroachDB` sidebar until first alpha)

Everything looks great -- fantastic work! Let 's get another set of eyes on this though before merge.

Also see DOC-5369 linked tickets for follow-up tasks post-merge (I'll do 'em).

v22.2/security-reference/scram-authentication.md
```diff
@@ -67,10 +67,6 @@ This section details how to use SCRAM authentication rather than plaintext passw
 1. [Enable SCRAM-SHA-256 password encryption](#enabling-scram-sha-256-authentication-for-new-users-roles)
 2. [Migrate pre-existing users](#migrating-existing-users-roles-to-scram-sha-256-authentication)
 
-{{site.data.alerts.callout_info}}
-SCRAM authentication can only be used after the upgrade to v22.1 has been finalized.
-{{site.data.alerts.end}}
-
```
mikeCRL: @MichaelTrestman Do we need a similar but revised note for the v22.2 docs? If a customer were on an upgrade path from v21.2, would't this still be true, but need slight rewording, e.g.

> If you are running a version of CockroachDB prior to v22.1, SCRAM authentication can only be used after an upgrade to v22.1 has been finalized.

mikeCRL: Thanks, Andrew. BTW, just fixed my blockquote - I meant for the last version number there to be "v21.2".

_config_base.yml
nickvigilante: That's a good note. I will update the releases index to suppress any versions that do not contain a release.

If we do not include this in the config, then the main problem we'll face is we won't see v22.2 in the version switcher. Other links may also not resolve correctly and there may be blank bits of text that use `page.version.version` anywhere in the page. I think it's necessary to keep it.

I can instead use `v22.2.0 (unreleased)` or something similar to indicate that it's currently not released. What do you think?

_data/versions.csv
```diff
@@ -10,3 +10,4 @@ v20.2,2020-11-10,2021-11-10,2022-05-10,v20.1
 v21.1,2021-05-18,2022-05-18,2022-11-18,v20.2
 v21.2,2021-11-16,2022-11-16,2023-05-16,v21.1
 v22.1,2022-05-24,2023-05-24,2023-11-24,v21.2
+v22.2,N/A,N/A,N/A,v22.1
```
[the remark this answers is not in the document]
nickvigilante: Adding this helped suppress some errors within `upgrade-cockroach-version.md` and `upgrade-cockroachdb-kubernetes.md `. I was able to correct that temporarily by adding the `{% if rd %}` syntax within those files, but it's not necessary if we just include the row for `v22.2`.

v22.2/upgrade-cockroach-version.md
```diff
@@ -66,7 +66,7 @@ In case a decommissioning process is hung, [recommission](node-shutdown.html?fil
...
+Review the [backward-incompatible changes in {{ page.version.version }}](../releases/{{ page.version.version }}.html{% if rd %}{% unless rd.release_date == "N/A" or rd.release_date > today %}#{{ page.version.version | replace: ".", "-" }}-0-backward-incompatible-changes{% endunless %}{% endif %}) and [deprecated features](../releases/{{ page.version.version }}.html#{% if rd %}{% unless rd.release_date == "N/A" or rd.release_date > today %}{{ page.version.version | replace: ".", "-" }}-0-deprecations{% endunless %}{% endif %}). If any affect your deployment, make the necessary changes before starting the rolling upgrade to {{ page.version.version }}.
```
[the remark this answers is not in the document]
nickvigilante: Backward-incompatible changes and deprecated features are not populated until the GA release notes are pushed. The logic here to hide the anchor is necessary until that time to suppress any errors fro htmltest.

v22.2/security-reference/scram-authentication.md
nickvigilante: Yep, since upgrading to v22.1 is a required step to upgrade to v22.2 and above, we figured that note was no longer necessary.

_data/versions.csv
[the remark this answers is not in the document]
nickvigilante: Also, we don't update the release support policy table until we have tangible release dates, which doesn't happen until GA. I wouldn't expect that page to update just yet.

v22.2/security-reference/scram-authentication.md
MichaelTrestman: Yup, I concur that the note is no longer needed!

nickvigilante:
@mikeCRL I fixed some issues with the releases index and I undid the change to `upgrade-cockroach-version.md` and `upgrade-cockroachdb-kubernetes.md`. There are 2 outstanding comments left. Hoping to get your final OK on those before I merge.

_config_base.yml
mikeCRL: > I can instead use v22.2.0 (unreleased) or something similar to indicate that it's currently not released. What do you think?

Do you mean that you'd change the value of `name` and `version` to something that's easier for us to understand, but it won't actually be shown on site? If so, that's fine by me. Maybe just `v22.2-alpha.1 (unreleased)` so the next RN PR can just drop 'unreleased'?

nickvigilante: I don't fully understand why we split name and version and keep them as the same thing, but whatever we set as `version` is the text that is shown directly in the version switcher drop down.

I'll take a separate follow-up action to understand the version plug-in and why it's built the way it's built.

What should we display publicly for end users?

mikeCRL: To recap, before alpha 1 is released, we need reasonable text for the dropdown re. what version you're seeing. Maybe

`v22.2 (dev / pre-alpha)` ?

mikeCRL: Edit: After discussion re. limitations, `v22.2 (dev)` is fine. Thanks!

mikeCRL (approved):
Approving, pending update of version drop-down text to `v22.2 (dev)`

Issue #14814, opened 2022-08-13
download/use instructions for the cockroach-sql binary


knz:
Previously discussed with @ianjevans and @nickvigilante here: https://github.com/cockroachdb/cockroach/issues/81246

we were causing UX issues by shipping a "naked" binary for `cockroach-sql`.

This was fixed in https://github.com/cockroachdb/cockroach/pull/86051 - we're now going to have archives containing the one file instead.

We need to decide what to do:
- do we keep the standalone binary for the previous 22.1 releases? If so, we'd need separate instructions and hyperlinks in docs.
- otherwise, we need to coordinate with @rail to ensure we use the same format for all the releases.

In any case, the hyperlink generation on the docs site needs to change before the next 22.1 release, otherwise we'll get broken links.

nickvigilante:
My vote: update all previous sql binaries to be in an archive, so the instructions are uniform.

knz:
thanks nick. @rail ^^

knz:
@nickvigilante , Rail has advised we can't transform the past archives easily. This means that we should either make-do with different download types (naked before, now properly packaged); or remove the links to the previous downloads.

What do you think?

nickvigilante:
My initial thoughts are people may ask us what happened to the old SQL binaries if we remove them. I think it's less effort overall if we just provide instructions on how to use the two different binaries.

PR #14845, opened 2022-08-16
Change download links for SQL shells to archives/tarballs

nickvigilante:
Informs #14814


rail:
The tarballs are up and we can proceed with this PR

stbof:
The table is really wide and there are some inconsistencies in the labels. Also, do we really need Precompiled if it's already labeled as a binary? How about:

64-bit Binary (Full)     Full Binary SHA256     64-bit Binary (SQL shell)    SQL Shell Binary SHA256

nickvigilante:
@stbof There may be subsequent CSS changes that need to be made on tables that we can address in a separate JIRA. For now, let me know how this looks.

stbof (approved):
So much better!

PR #14861, opened 2022-08-17
DOC-5373 new Vale exclusions

andf-crl:
Addresses: DOC-5373

- Adds a number of new Vale spelling / caps exclusions to `vocab.txt`, drawn from going through the v22.2 docs cut and seeing what Vale errors presented themselves on this rare pass through the entire corpus.
	- Took opportunity to `sort -f` the list, but returned `_include` to its initial location.
	- Besides alpha sort, only extant entry modified by this PR is `sstables` in favor of new `SSTables`. This might not be the right thing, as there are many instances of `sstables` (lc) in the corpus in code sample, etc. Too noisy? Human-typed text should always use `SSTables`, but maybe not worth the hassle of too many false-positives in code samples. Leaning towards reverting this back to how I found it for `sstables`. Let me know what you think.


netlify/vale/vocab.txt
```diff
@@ -470,14 +485,15 @@ schemaless
 schimmel
 scrollable
 scrollbars
-seif
 seeked
+seif
+SELinux
```
nickvigilante, line 490: Can we also add the all-lowercase form of this for commands?

nickvigilante (approved):
Approved with one comment.

andf-crl:
Thanks @nickvigilante , added lower `selinux`, good call. Also added back in lower `sstables` and `sstable`. Just know it'd result in noise complaints from the neighbors otherwise. TYFR!

PR #14866, opened 2022-08-17
Remove unnecessary table styles.

stbof:


mdlinville (approved):
LGTM, checked several different files from very wide to very narrower browser widths with no surprises.

PR #14882, opened 2022-08-18
DOC-3670 update prometheus operator

andf-crl:
Addresses: DOC-3670

- Updates hardlink to CoreOS Prometheus Operator to `v0.58.0` from `v0.47.1`
- Doubles up on notice to replace example hardcoded version with latest version
- Adds `--server-side` to `apply` step
- Updates command output to match use of `--server-side`

Question:

- Your ZD ticket has the end user split the yaml payload into two files, and only suggests using `--server-side` with one of them. Is that because your client already had an extant config, and new users -- i.e. those following this docs page -- would not need to split as you have?

Alternatives:

- If approach: "look up latest, then replace hardcoded with your findings" is too cumbersome, we can instead just direct users to download latest `bundle.yaml` from [here](https://github.com/prometheus-operator/prometheus-operator/releases/) directly, then run `kubectl apply -f /local/path/to/bundle.yaml --server-side` instead. What do you think?

[v22.2/monitor-cockroachdb-kubernetes.md](https://deploy-preview-14882--cockroachdb-docs.netlify.app/docs/dev/monitor-cockroachdb-kubernetes.html#configure-prometheus) | [v22.1/monitor-cockroachdb-kubernetes.md](https://deploy-preview-14882--cockroachdb-docs.netlify.app/docs/stable/monitor-cockroachdb-kubernetes.html#configure-prometheus) | [v21.2/monitor-cockroachdb-kubernetes.md](https://deploy-preview-14882--cockroachdb-docs.netlify.app/docs/v21.2/monitor-cockroachdb-kubernetes.html#configure-prometheus) | [v21.1/monitor-cockroachdb-kubernetes.md](https://deploy-preview-14882--cockroachdb-docs.netlify.app/docs/v21.1/monitor-cockroachdb-kubernetes.html#configure-prometheus)

### DIFF
--- a/v21.1/monitor-cockroachdb-kubernetes.md
+++ b/v21.1/monitor-cockroachdb-kubernetes.md
@@ -68,31 +68,32 @@ If you're on Hosted GKE, before starting, make sure the email address associated
     This ensures that there is a Prometheus job and monitoring data only for the `my-release-cockroachdb` service, not for the `my-release-cockroach-public` service.
     </section>
 
-1. Install [CoreOS's Prometheus Operator](https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/release-0.43/bundle.yaml):
+1. Determine the latest version of [CoreOS's Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator/releases/) and run the following to download and apply the latest `bundle.yaml` definition file:
 
     {{site.data.alerts.callout_info}}
-    We recommend checking for the latest Prometheus Operator [release version](https://github.com/prometheus-operator/prometheus-operator/blob/master/RELEASE.md) and specifying this version in the following command.
+    Be sure to specify the latest [CoreOS Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator/releases/) version in the following command, in place of this example's use of version `v0.58.0`.
     {{site.data.alerts.end}}
 
     {% include_cached copy-clipboard.html %}
     ~~~ shell
     $ kubectl apply \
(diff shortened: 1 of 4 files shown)


daniel-crlabs:
>Your ZD ticket has the end user split the yaml payload into two files, and only suggests using --server-side with one of them. Is that because your client already had an extant config, and new users -- i.e. those following this docs page -- would not need to split as you have? 

The reason I split was the following, as I told the customer below, it is not a requirement, it is just more organized and easier to follow.

>Open the bundle and split it into 2 different files (this will make it easier to manage and be more organized) for each of the separate definitions.

andf-crl:
> > Your ZD ticket has the end user split the yaml payload into two files, and only suggests using --server-side with one of them. Is that because your client already had an extant config, and new users -- i.e. those following this docs page -- would not need to split as you have?
> 
> The reason I split was the following, as I told the customer below, it is not a requirement, it is just more organized and easier to follow.
> 
> > Open the bundle and split it into 2 different files (this will make it easier to manage and be more organized) for each of the separate definitions.

Thanks for this explanation Daniel, I'll leave as is then per your comment. Do approve this PR for publish otherwise? Thanks!

daniel-crlabs:
I'm not sure I'm the one to approve things :-) I just created the request since I ran into this and thought it would be great to have our documentation updated.

andf-crl:
Who then, if not the progenitor of the idea, the detailer of the steps to reproduce the issue, and the suggester of the textual fix, now made through the efforts of this PR? I propose that you are in fact the ideal person to review this PR!

But for next time. I'll move this one along out of band.

MichaelTrestman (approved):
LGTM

PR #14886, opened 2022-08-18
[DOC-417] Provide workaround to run the cockroach binary on macOS

mdlinville:
[DOC-417]

- [x] Verified with 22.1.5 binary that the workaround is still required
- [x] Local build
- [x] Local linkcheck


stbof:
Don't you have to click Open again so that it's saved as an exception. See https://support.apple.com/en-us/HT202491?

mdlinville:
> Don't you have to click Open again so that it's saved as an exception. See https://support.apple.com/en-us/HT202491?

You're right, I missed that because I didn't follow through and run the command the second time. Fixed now.

stbof (approved):

PR #14895, opened 2022-08-19
[DOC-593] Add info about case-sensitivity of locality zones

mdlinville:
[DOC-593]
- [x] Local build
- [x] Local linkcheck


v22.2/configure-replication-zones.md
```diff
@@ -99,7 +99,7 @@ When starting a node with the [`cockroach start`](cockroach-start.html) command,
...
...ster based on locality, with the order of locality tiers determining the priority. Locality can also be used to influence the location of data replicas in various ways using replication zones.<br><br>When there is high latency between nodes, CockroachDB uses locality to move range leases closer to the current workload, reducing network round trips and improving read performance. See [Follow-the-workload](topology-follow-the-workload.html) for more details.<div class="bs-callout bs-callout--danger"><div class="bs-callout__label">Warning:</div>Repeating an exact locality value has no effect, but locality values are case-sensitive. For example, from the point of view of CockroachDB, the following values result in two separate localities:<div class="highlight"><pre><code style="border: 0" class="language-shell" data-lang="shell"><span class="nt">--locality</span><span class="o">=</span><span class="nv">datacenter</span><span class="o">=</span>us-east-1 and <span class="nt">--locality</span><span class="o">=</span><span class="nv">datacenter</span><span class="o">=</span>US-EAST-1</code></pre></div>This type of configuration error can lead to issues that are difficult to diagnose.</div>
```
mdlinville, line 102: I filed https://cockroachlabs.atlassian.net/browse/DOC-5428 to address the need to use `style="border: 0"` in this context.

stbof: Although it's not documented anywhere, I believe we don't use rendered [callouts](https://github.com/cockroachdb/docs/blob/master/StyleGuide.md#tips-notes-and-warnings) in tables for precisely the problem you are working around here.  Instead,  I've seen manual callouts using **Note:**. For an example, see https://www.cockroachlabs.com/docs/v22.1/cockroach-start-single-node.html#general

Perhaps a topic for the doc team meeting.

mdlinville: The admonition within the table is a bit of a red herring (ha). Our CSS doesn't have a broad enough selector to avoid the default border on `<pre>` tags in this context. I'd wager that our intention is for code blocks to behave the same regardless of which block level element they occur within. So I'll keep the JIRA as-is, but I've modified these to use your suggested type of note.

nickvigilante (approved):

PR #14913, opened 2022-08-23
fixed typo

BWalti:
it's cockroach you know :)


nickvigilante:
@BWalti Super embarrassing we don't know our own product name, but thank you for the contribution! We went through a restructuring of our docs as well as a default branch rename, so I went and applied this change in #17217.

PR #14918, opened 2022-08-23
Fix to filter issue on cloud storage bulk ops pages

kathancox:
To fix the filter issue on the Use Cloud Storage for Bulk Ops page, adjusted the name ID of the second filter. Although this means the filters won't adjust together between the two relevant sections, it seems a worthwhile trade-off to keep the filters in both sections that are distinct from one another. Also, this makes the most sense in the case that further _different_ storage options are added, or where storage options aren't applicable to the two sections in question. https://www.cockroachlabs.com/docs/v21.2/use-cloud-storage-for-bulk-operations

### DIFF
--- a/v21.2/use-cloud-storage-for-bulk-operations.md
+++ b/v21.2/use-cloud-storage-for-bulk-operations.md
@@ -288,11 +288,11 @@ This table outlines the actions that each operation performs against the storage
 
 
 <div class="filters clearfix">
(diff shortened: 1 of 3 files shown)


andf-crl (approved):
LGTM! I see what you mean about tabs not switching together on-page for AWS and GCP, but agree with you that this is probably the best path forward given our filtering/linking constraints. Nice fix, thank you!

kathancox:
Thanks Andrew!

PR #14938, opened 2022-08-25
add EAR store key generation with Vault

MichaelTrestman:
Add doc for generating encryption at rest (EAR) store key with Hashicorp Vault Transit Secret Engine

https://cockroachlabs.atlassian.net/browse/DOC-3309


v22.2/encryption.md
```diff
@@ -17,20 +17,96 @@ Cockroach determines which encryption algorithm to use based on the size of the
...
+
+~~~txt
+Key     Value
+---     -----
+keys    map[1:6kkUAMq4EgIJKEmFxl97bQaBn0n9FL6FbRoUeCBtVhE=]
+name    crdb-key
+type    aes256-gcm96
+~~~
+
+But stop, since the exportable key is “just” 256 bits and cockroachDB has a special mechanic here. Let me quote our docs “The key file must contain random data making up the key ID (32 bytes) and the actual key (16, 24, or 32 bytes depending on the encryption algorithm).”
```
mdlinville, line 78: This seems a little informal. It's not ideal to lead the reader into a mistake that they need to then recover from. I wonder if it's worth explaining this limitation / workaround earlier in the process.

v22.2/encryption.md
```diff
@@ -17,20 +17,96 @@ Cockroach determines which encryption algorithm to use based on the size of the
...
+
+So, the 256 bits from the exportable key would be required as an identifier and there would be no bits left for the actual key. So we have to tweak it a bit. We are going to add some random data at the beginning to use as key identifier and the actual Vault key as the real encryption key. 
+
+copy the key into a file e.g. key_from_valut.key
```
mdlinville, line 82:
```suggestion
1. Copy the key into a file named `key_from_value.key`.
```
It's nice to avoid Latin. If they use a different file name, they need to also change the following examples, so maybe it's better to just give them static instructions.

mdlinville (commented):
Some early comments, apologies if you didn't actually want review yet.

v22.2/encryption.md
pschulz1: AFAIK there is no other way/order of doing things.

pschulz1 (commented):
Looks good to me.

PR #14992, opened 2022-08-31
clarify use of bold text with UI elements

taroface:
Clarifying use of bold text when documenting UI elements, and explicitly recommending when to avoid this.

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -443,7 +443,7 @@ Enter a line break between a heading and its content.
 
 Use bold text to emphasize an important word or phrase, or to create visual separation and callouts (e.g., **Example:**). Do not combine bold with italic.
 
-Bold text should be used whenever referring to the name of a UI section or field. The name should only be in bold if it appears verbatim in the UI. Some UI sections, such as tables, are not labeled in the UI and should not be bolded when named in the documentation.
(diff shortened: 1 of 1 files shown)


StyleGuide.md
```diff
@@ -441,8 +441,10 @@ Enter a line break between a heading and its content.
 
 #### Bold
 
-Use bold text to emphasize an important word or phrase, when referring to the name of a UI section or field, or to create visual separation and callouts (e.g., **Example:**). Do not combine bold with italic.
+Use bold text to emphasize an important word or phrase, or to create visual separation and callouts (e.g., **Example:**). Do not combine bold with italic.
 
+Bold text should be used whenever referring to the name of a UI section or field. The name should only be in bold if it appears verbatim in the UI. Some UI sections, such as tables, are not labeled in the UI and should not be bolded when named in the documentation.
```
stbof, line 446:
```suggestion
Use bold text you refer to the name of a UI section or field. The name should be in bold only if it appears verbatim in the UI. If a UI element, such as a table, is not labeled in the UI, do not bold when you reference the element.
```

stbof (changes requested):

StyleGuide.md
taroface: I'd like to further edit and use your suggestion as follows. Look ok?

```suggestion
Use bold text when you refer to the name of a UI section or field. The name should be in bold only if it appears verbatim in the UI. If a UI element, such as a table, is not labeled in the UI, do not bold when you reference the element in the documentation.
```

stbof: LGTM

PR #14995, opened 2022-09-01
Tell git to make nicer diffs for Markdown files

rmloveland:
... by setting up the .gitattributes file to recognize Markdown
files (*.md).

This will result in diffs that are easier to read since the hunk context
will prefer to display nearby Markdown headings rather than random
nearby text.

Hopefully a small quality of life improvement.


nickvigilante (approved):
LGTM! I can't wait to see how it works.

PR #15044, opened 2022-09-08
Improve sitemap rendering and meta tagging for SEO

nickvigilante:
Enhancements:

1. Updated head.html to inject robots noindex tag for all pages that don't belong to the sitemap.
2. The sitemap now includes all pages in dev/ that don't exist in stable/.

Fixes DOC-5567


andf-crl (approved):
LGTM logic-wise, but am certainly no Liquid expert! Thanks for this!

JosephNickdow (commented):
LGTM @nickvigilante

PR #15047, opened 2022-09-08
Remove noindex tag from main docs index

nickvigilante:
We don't want the docs site itself not to be indexed. This fixes that bug.


JosephNickdow (commented):
LGTM @nickvigilante

PR #15035, opened 2022-09-08
[DOC-4748] Add ARM64 download links for Linux

mdlinville:
## HTML Previews
[Releases index](https://deploy-preview-15035--cockroachdb-docs.netlify.app/docs/releases/index.html) (the diff is difficult to review)
[Releases 22.2 minor version page with ARM downloads](https://deploy-preview-15035--cockroachdb-docs.netlify.app/docs/releases/v22.2.html)
[Releases 22.1 minor version page without ARM downloads](https://deploy-preview-15035--cockroachdb-docs.netlify.app/docs/releases/v22.1.html)
[Installation](https://deploy-preview-15035--cockroachdb-docs.netlify.app/docs/v22.2/install-cockroachdb-linux.html)

## Changes
- Update Linux installation instructions to provide info about ARM
- Add linux_arm field to _data/releases.csv
- Add a Javascript confirm on click of each Experimental download, both on the Releases index and on minor-version pages (all Windows, and Linux ARM)
- Add logic to releases/index.md and _includes/releases/release-downloads-docker-image.md (which actually builds out all of the minor-release pages, despite its name) to display the new links
- Define Testing and Experimental at the top level of the Releases Index page
- Completely refactor releases/index.md in other ways:
    - Change the looping logic to generate separate tables per version per platform
    - Always show SHA256 links
    - Improve link text for downloadable versions, Github links, and source code
    - For Docker, add a link to the cockroachdb image tags page
    - Remove CSS classes that were adding extra padding
 
## Tests
- [x] Local build
- [x] Link check
- [x] Download links download the correct things

PR #15064, opened 2022-09-09
Update htmltest to use curl instead of go get

nickvigilante:
`go get` is deprecated and it was causing funky things with the build. This will take care of that.


mdlinville (approved):
https://github.com/cockroachdb/docs/pull/15035 builds with this patch.

PR #15065, opened 2022-09-09
Remove debug tags from netlify build

nickvigilante:


netlify/build
```diff
@@ -23,10 +23,8 @@ cp _site/docs/404.html _site/404.html
 
 # Set up htmltest
 
-curl https://htmltest.wjdp.uk | bash
-echo "debug 0"
+curl https://htmltest.wjdp.uk | bash > /dev/null 2>&1
```
mdlinville, line 26: This will fail if that site is down but you won't find out because you're redirecting STDERR ... that OK?

mdlinville (approved):

netlify/build
nickvigilante: I fixed it in the event that we can't download htmltest.

netlify/build
```diff
@@ -23,10 +23,12 @@ cp _site/docs/404.html _site/404.html
 
 # Set up htmltest
 
-curl https://htmltest.wjdp.uk | bash
-echo "debug 0"
+curl -s https://htmltest.wjdp.uk | bash
+if [[ $? != 0 ]]; then
+	echo "Failed to install htmltest"
+	exit 1
+fi;
```
mdlinville, line 30:
```suggestion
fi
```
It will work but it's non-standard syntax.

mdlinville (approved):

netlify/build
nickvigilante: There are other parts of this file that have it, so I'll log a new PR for it on Monday.

PR #15073, opened 2022-09-12
DOC-5593: Add lastmod tag to sitemap

nickvigilante:
Fixes DOC-5593

This helps pave the way for Google to effectively identify when content changes, with the exception of modifying any includes referenced in the file. After this is merged, we can then trigger a sitemap update during the Netlify build and have Google identify which pages need to change.


andf-crl (approved):
LGTM! Will we all need to `gem install jekyll-last-modified-at` at some point also?

nickvigilante:
@andf-crl Yes, but a better solution would be to merge changes from Master into each of our local repos.

PR #15092, opened 2022-09-14
DOC-5659: Excluded the sidebar from search engine indexing

nickvigilante:
Fixes DOC-5659 and the following:

- Tidies up the default and page layouts
- Moves certain elements to include_cached for better performance

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,4 +1,4 @@
-<div class="footer py-5">
+<div class="footer py-5" data-nosnippet>
   <div class="container">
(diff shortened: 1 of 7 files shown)


mikeCRL (approved):
LGTM. As we discussed, I suspect that adding `data-nosnippet` may only affect the snippets that appear for search results and _not_ exclude the nav from the index itself, but this change can only be of help to users; let's see the result, and we can later decide whether further changes are needed to more fully exclude the nav.

PR #15097, opened 2022-09-14
Add Jobs page to Cloud docs.

stbof:
DOC-4718
DOC-3648

There is no 22.1 version of these docs, and to get this to build I had to use the `dev` site version in these definitions:

```
{% if page.cloud == true %}
  {% capture link_prefix %}../{{site.versions["dev"]}}/{% endcapture %}
  {% assign page_prefix = "" %}
{% else %}
  {% assign link_prefix = "" %}
  {% assign page_prefix = "ui-" %}
{% endif %}
```

`{% capture version_prefix %}{{site.versions["dev"]}}/{% endcapture %}`


ianjevans (approved):

stbof:
@ianjevans Thanks for the review. Should I have to change `stable` to `dev`? How do I know when to I change it back?

ianjevans:
> @ianjevans Thanks for the review. Should I have to change `stable` to `dev`? How do I know when to change it back?

This should only be temporary, until v22.2 is released. I think this is the kind of thing we should be able to catch with a script when a release goes GA. What do you think @nickvigilante?

nickvigilante:
@stbof @ianjevans We can programmatically determine if a page exists in either `dev` or `stable` by using `site.pages`. It should look roughly something like this:

```
{% assign dev_link_prefix = site.versions["dev"] | append: "/" %}
{% assign stable_link_prefix = site.versions["stable"] | append: "/" %}

{% assign dev_pages = site.pages | where_exp: "dev_pages", "dev_pages.path contains dev_link_prefix" | map: path %}
{% assign stable_pages = site.pages | where_exp: "stable_pages", "stable_pages.path contains stable_link_prefix" | map: path %}

{% if page.cloud == true %}
  {% if stable_pages contains include.page_name %}
    {% capture link_prefix %}../{{ stable_link_prefix }}{% endcapture %}
  {% elsif dev_pages contains include.page_name %}
    {% capture link_prefix %}../{{ dev_link_prefix }}{% endcapture %}
  {% endif %}
  {% assign page_prefix = "" %}
{% else %}
  {% assign link_prefix = "" %}
  {% assign page_prefix = "ui-" %}
{% endif %}
```

However, every reference to this include must also pass in a page name variable such as `add-column.md`. For instance:

{% include common/ui/jobs-page.md page_name="add-column.md" %}

And what that will do is check to see if the current page you're on is a cloud page, and if it is, check to see if the desired destination file is in `stable`. If it is, use `stable` as the link prefix. Otherwise, check if it's in `dev`, and use the `dev` link prefix.

nickvigilante:
My comment above requires there to be an actual file within the docs/v22.2 folder to work. It doesn't work on includes.

Side note: I'd love for there to be some kind of site variable that lists the includes.

PR #15106, opened 2022-09-15
Fix title and content spacing

nickvigilante:


JosephNickdow (commented):
Looks good @nickvigilante

PR #15119, opened 2022-09-18
Update build-a-go-app-with-cockroachdb-gorm.md

quanuw:
Fix sample navigation in step 3.


nickvigilante:
@quanuw Thank you for the contribution! I incorporated this change as part of https://github.com/cockroachdb/docs/pull/17211.

PR #15123, opened 2022-09-19
Add docs for secondary regions

rmloveland:
Fixes DOC-4753
Fixes DOC-5507
Fixes DOC-5573
Fixes DOC-5665

Summary of changes:

- Add 'Secondary regions' section to the Multiregion Overview page, with links to the new SQL pages:

  - SET SECONDARY REGION (including note that this implicitly does an ADD REGION)

  - DROP SECONDARY REGION

- Add info re: interaction of Super Regions and Secondary Regions

- Update CREATE DATABASE page to show how to add a SECONDARY REGION at table creation time

- Update SHOW REGIONS page so output shows secondary regions

- Add links to Secondary Regions things from other pages

- Fix a small error where the `cockroach` command was referred to as `cockroach211`, which would ofc not work


rmloveland:
@RichardJCai added you to review due to https://github.com/cockroachdb/cockroach/pull/76620
@e-mbrown added you to review due to https://github.com/cockroachdb/cockroach/pull/84999 

Please let me know what you think - thanks!

rmloveland:
> We changed [secondary region syntax](https://github.com/cockroachdb/docs/issues/15010) a bit and [Added secondary regions to SHOW REGION](https://github.com/cockroachdb/docs/issues/14970) if you want to include them in this PR.

Thanks Ebony!  I've updated the PR as follows:

- the `SHOW REGIONS` page lists the `secondary_region` column
- the `SET SECONDARY REGION` page has a note about the implicit region add if you make a region secondary

> If not this looks good to me!

Thank you for the review!

rmloveland:
@RichardJCai do you see any additional updates/etc needed on this one?

RichardJCai:
> @RichardJCai do you see any additional updates/etc needed on this one?

Nope, LGTM!

v22.2/multiregion-overview.md
```diff
@@ -100,6 +100,27 @@ For more information about data domiciling using `PLACEMENT RESTRICTED`, see [Da
 Super regions rely on the underlying [replication zone system](configure-replication-zones.html), which was historically built for performance, not for domiciling. The replication system's top priority is to prevent the loss of data and it may override the zone configurations if necessary to ensure data durability. For more information, see [Configure Replication Zones](https://www.cockroachlabs.com/docs/v21.2/configure-replication-zones#types-of-constraints).
 {{site.data.alerts.end}}
 
+{{site.data.alerts.callout_info}}
+If you are using super regions with the [secondary region](#secondary-regions) feature, the secondary region must be inside or outside of the same super region as the [primary region](set-primary-region.html).
```
ianjevans, line 104: I think this needs to be clarified. "inside or outside of the same super region" is confusing, to me. Are you saying that if the primary region is within a super region, the secondary region also must be in a super region and that super region must be different than the primary region?

By definition, any region is either inside or outside of the primary region's super region.

v22.2/multiregion-overview.md
```diff
@@ -100,6 +100,27 @@ For more information about data domiciling using `PLACEMENT RESTRICTED`, see [Da
...
+
+This behavior is an improvement over versions of CockroachDB prior to v22.2.  In those versions, when the primary region failed, the [leaseholders](architecture/replication-layer.html#leases) would be transferred to another [database region](#database-regions) at random, which could have negative effects on performance.
+
+For more information about how to use secondary regions, see:
+
+- [`SET SECONDARY REGION`](set-secondary-region.html)
+- [`DROP SECONDARY REGION`](drop-secondary-region.html)
+
+{{site.data.alerts.callout_info}}
+If you are using a secondary region and a [super region](#super-regions), the secondary region must be inside or outside of the same super region as the [primary region](set-primary-region.html).
```
ianjevans, line 121: Same issue with this callout as the previous comment.

v22.2/multiregion-overview.md
dikshant: I was looking into this a bit more because I was also curious how super regions and secondary regions would interact. I used the `movr` database on a 3 node multi region `{us-east4, us-west2, us-central1}` cluster.

Added a super region `us-east-west`
```
root@localhost:26257/defaultdb> ALTER DATABASE movr ADD SUPER REGION "us-east-west" VALUES "us-east4", "us-west2";
ALTER DATABASE ADD SUPER REGION


Time: 1.461s total (execution 1.461s / network 0.000s)
````

Then I tried to add a secondary region that is OUTSIDE the super region and I got this error:
```
root@localhost:26257/defaultdb> ALTER DATABASE movr SET SECONDARY REGION "us-central1";
ERROR: the secondary region must be in the same super region as the current primary region
SQLSTATE: 42P12
```

Similarly if you try to move the primary region out of a super region:
```
root@localhost:26257/defaultdb> ALTER DATABASE movr SET PRIMARY REGION "us-central1";
ERROR: moving the primary region into or out of a super region could have undesirable data placement implications as all regional tables in the primary region will now moved.

the old primary region us-east4 is a member of super region us-east-west
HINT: You can enable this operation by running `SET alter_primary_region_super_region_override = 'on'`.
```
As expected, secondary cannot be the same as the primary even though it is in the super region which makes sense.
```
root@localhost:26257/defaultdb> ALTER DATABASE movr SET SECONDARY REGION "us-east4";
ERROR: the secondary region cannot be the same as the current primary region
SQLSTATE: 42P12
```

Finally doing:
```
root@localhost:26257/defaultdb> ALTER DATABASE movr SET SECONDARY REGION "us-west2";
ALTER DATABASE SET SECONDARY REGION


Time: 73ms total (execution 73ms / network 0ms)
```
Works as expected.

So maybe we want it to say:

> If you are using super regions with the [secondary region](#secondary-regions) feature, the secondary region must be one of the regions inside of the super region but not the primary region.

Thoughts?

e-mbrown: Thats the gist! 
In the case where there are multiple super regions assigned to a database, lets say SR-A, and SR-B.

If the primary region is in SR-A, the secondary region must be a region within SR-A. It can not be apart of SR-B.

If the primary region is not a region within SR-A or SR-B(outside of the super region) the secondary region can not be a region within SR-A or SR-B

ianjevans: Suggested text:
> If you are using super regions in your cluster, there are additional constraints when using secondary regions. If the primary region is in a super region, the secondary region must be a region within the primary's super region. If the primary region is not in a super region, the secondary region must not be within a super region.

rmloveland: Thanks for these comments!  I updated it to use the suggested text from Ian - PTAL

v22.2/multiregion-overview.md
rmloveland: Updated using the same new callout (with slight change to link out to super region section)

ianjevans (approved):

PR #15171, opened 2022-09-22
Update monitor-cockroachdb-with-prometheus.md

GerDner:
Link points to non existing file.


nickvigilante:
@GerDner Thank you for the contribution! We went through a restructuring of our content as well as a default branch rename, so I applied your change as part of #17218.

PR #15033, opened 2022-09-07
Schema Conversion Tool docs; overhaul Migrations Overview and sidebar

taroface:
Fixes DOC-4921.
Fixes DOC-5243.
Fixes DOC-5463.

- Add reference doc for Migrations Page (Schema Conversion Tool)
- Overhaul Migration Overview to reflect the actual migration steps (schema, data, application)
- Reorganize Migrations navigation in sidebar
- Recommend AWS DMS more prominently in PostgreSQL, MySQL, and Oracle migration docs

Note:

- These changes apply to v21.2 - v22.2 docs
- A tutorial for using the SCT is planned for a separate PR. It will be added to the current `migrate-from-postgres.md` doc. In a discussion with @vy-ton I realized I should prioritize updating the Migration Overview doc above the SCT tutorial, since the UI links directly to the Migration Overview. This is **v1** of a new Migration Overview and mostly reframes & reorganizes existing information in our docs.
  - I made an educated guess on how to address "Unimplemented features" for now, and am open to corrections and improvements.

PR #15176, opened 2022-09-22
temporarily remove migrations docs; implement redirects

taroface:
In #15033 I hid the links to some currently incorrect migrations docs, but did not remove the docs from the repo, and they were still accessible through search and other links.

- Removed the docs from v21.2-v22.2
- Implemented server-side redirects for those versions

The docs will be replaced and redirects will be removed once the instructions are updated.


kathancox:
Looks like there are some link removals/changes that need to happen here... I'll wait till you've done that and review! :)

taroface:
Thanks @kathancox! Closing in favor of #15179.

PR #15187, opened 2022-09-23
Fix certain redirect items

nickvigilante:
- Asterisks are not supported inside destination URLs. It is treated as a plaintext asterisk. I replaced them with `:version` placeholders per the [Netlify documentation](https://docs.netlify.com/routing/redirects/redirect-options/#placeholders).
- For pages that have only temporarily moved, update the status code from 301 to 302.


andf-crl (approved):
LGTM! Nice work.

PR #15192, opened 2022-09-23
infra: add Segment snippet + page impression to all pages

sjbarag:
Using Segment for page tracking enables consistent user IDs with the CockroachDB Cloud console, and allows us to configure where events are sent without redeploying this site. Add the standard Segment snippet, which fires a page impression on-load by default.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -51,6 +51,14 @@ var pageConfig = {
 
 <script src="{{ 'js/jquery.min.js' | relative_url }}"></script>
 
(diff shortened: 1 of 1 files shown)


sjbarag:
Just an FYI that I have the Segment source disabled server-side for now.  I'll enable it next week, but if anyone reviewing this needs it turned on sooner just let me know!

nickvigilante (approved):
LGTM! Do you have merge access to docs?

sjbarag:
> LGTM! Do you have merge access to docs?

I do!  Thanks for the review!

PR #15074, opened 2022-09-12
Insights

stbof:
Addresses roadmap and product change Jiras:

- DOC-4730
- DOC-4741
- DOC-5328
- DOC-5588
- DOC-5557
- DOC-5556
- DOC-5545
- DOC-5512
- DOC-5495
- DOC-5484
- DOC-5480
- DOC-5470
- DOC-5433
- DOC-5432
- DOC-5418
- DOC-5414
- DOC-5410
- DOC-5325
- DOC-5304
- DOC-5165
- DOC-5469
- DOC-5004
- DOC-3243
- DOC-3046
- DOC-5650
- DOC-5664
- DOC-5527
- DOC-5496
- DOC-5667
- DOC-5382
- DOC-5299
- DOC-5277
- DOC-5124
- DOC-5884

PR #15193, opened 2022-09-23
Provide a placeholder for the link to the Insights page in the Insights UI.

stbof:
The 22.2 Insights UI pages point to `/docs/stable/ui-insights-page.html` but that page won't exist until GA.  In fact, `/docs/dev/ui-insights-page.html` probably won't be published either when the beta is released because [the PR](https://github.com/cockroachdb/docs/pull/15074) is still under development.  This is a placeholder page in `/docs/v22.1/ui-insights-page.html` which maps to  `/docs/stable/ui-insights-page.html` that says the feature is under development so that the link from the UI doesn't return 404.

### DIFF
--- a/v22.1/ui-insights-page.md
+++ b/v22.1/ui-insights-page.md
@@ -1,6 +1,6 @@
 ---
 title: Insights Page
-summary: The Insights page exposes problems that CockroachDB has detected in your workloads and schemas and offers recommendations to improve the performance of your workloads.
+summary: The Insights page exposes problems that CockroachDB has detected in your workloads and schemas.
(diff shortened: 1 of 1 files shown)


maryliag (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15193#-:-NCgBmls4uLbEpVVi0RC:bnfp4nl)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @taroface)

taroface (approved):
LGTM

PR #15213, opened 2022-09-26
[DOC-5792] Fix copy buttons in code blocks

mdlinville:
## Changes
In v22.2, v22.1, v21.2, v21.1, and cockroachcloud:
- Add blank line before copy-widget includes
- Add blank line before `</section>`s and `</div>`s near the above changes
- Add blank line before section headings within `<section>`s
- In a few places, add code highlighting hint and copy-widget include to code blocks the user should be able to copy

I'm sure there are places I've missed because regex is hard, but this should fix most of the existing problems with the copy widget in code blocks.

## Previews
See [this comment](https://github.com/cockroachdb/docs/pull/15213#issuecomment-1258654075) for linked previews.

## Tests
- [x] Local build
- [x] Link check


stbof:
When I click the copy button I don't get a purple checkmark. Which means it's not working in the sense that it's not giving the right feedback.

mdlinville:
@nickvigilante PTAL, the copy buttons still appear broken even with the added blank lines.

mdlinville:
> @mdlinville The problem exists within the following includes:
> 
> _includes/cockroachcloud/download-the-binary.md _includes/cockroachcloud/download-the-cert.md _includes/cockroachcloud/sql-connection-string.md
> 
> Fix those in the exact same way and this problem should be fixed.

I fixed it in all of these and found a couple more files there the problem exists. It makes me wonder how many more there are across the docset.

nickvigilante (approved):

stbof (approved):

PR #15219, opened 2022-09-27
Use $1 as placeholder.

stbof:
DOC-5749
DOC-5747

### DIFF
--- a/_includes/common/ui/statements-page.md
+++ b/_includes/common/ui/statements-page.md
@@ -31,8 +31,8 @@ The following statements cannot be represented by the same fingerprint:
 - `INSERT INTO orders(product_id, customer_id, transaction_id) VALUES (380, 11, 11098)`
 - `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES (380, 11, 11098)`
 - `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, 11, 11098)`
-- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $2, 11098)`
-- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $2, $3)`
+- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $1, 11098)`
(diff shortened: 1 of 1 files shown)


ericharmeling (approved):
LGTM

_includes/common/ui/statements-page.md
```diff
@@ -31,8 +31,8 @@ The following statements cannot be represented by the same fingerprint:
 - `INSERT INTO orders(product_id, customer_id, transaction_id) VALUES (380, 11, 11098)`
 - `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES (380, 11, 11098)`
 - `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, 11, 11098)`
-- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $2, 11098)`
-- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $2, $3)`
+- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $1, 11098)`
+- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $1, $1)`
```
maryliag, line 35: This change is not correct. The change I made is when creating the fingerprint the value is replaced by `$1`, the value you're showing here is the statement being executed, so you should keep the values as is here, because those are the original values.

You can update line 19 to say:
`For multiple SQL statements to be represented by a fingerprint, they must be identical aside from their literal values and placeholders:`
then you keep the example with only literal as is and add below line 27 something like:


- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $2, 11098)`
- `INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($3, $4, 300)`

The preceding SQL statements have the fingerprint:
`INSERT INTO new_order(product_id, customer_id, transaction_id) VALUES ($1, $1, _)`

maryliag (changes requested):

_includes/common/ui/statements-page.md
stbof: Working on it.

PR #15220, opened 2022-09-27
Fix path to diagram.

stbof:
DOC-5518

### DIFF
--- a/v22.2/show-backup.md
+++ b/v22.2/show-backup.md
@@ -24,7 +24,7 @@ For guidance on the syntax for `SHOW BACKUP FROM`, see the [examples](#examples)
 ## Synopsis
 
 <div>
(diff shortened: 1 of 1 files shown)


nickvigilante:
I don't think we want this just yet. @kathancox was going to be making a change to convert just this diagram to a manual include until Engineering fixes the sql.y file for this command.

stbof:
> I don't think we want this just yet. @kathancox was going to be making a change to convert just this diagram to a manual include until Engineering fixes the sql.y file for this command.

I assumed I could go ahead because DOC-5517 was marked as fixed. Then I saw it wasn't backported to 22.2 and I asked @chengxiong-ruan  to do that: https://github.com/cockroachdb/cockroach/pull/88838.  

There's a bit of disconnect here.

nickvigilante:
Once that PR gets merged, I'm happy to have this merged.

PR #15229, opened 2022-09-28
Removing cockroach-sql for v22.1.7 and v22.1.6.

ianjevans:
Related to https://github.com/cockroachdb/cockroach/issues/88833. The `cockroach-sql` binary was incorrectly bundled as the full binary for v22.1.7 and v22.1.6. This PR removes the download from the docs.


rail (approved):
Stamp.

PR #15253, opened 2022-09-29
update minimum recommended versions for AWS DMS

otan:
After https://github.com/cockroachdb/cockroach/issues/88229, we want to advertise a higher minimum version.


taroface (approved):
LGTM!

PR #15276, opened 2022-10-03
(do not merge) test nesting SQL subcommands in nav

taroface:
Per DOC-5850, we might try nesting SQL subcommands underneath their parent commands in the Reference nav. This PR shows an example of doing so underneath `ALTER TABLE` (v22.2 doc). 

Preview link (see nav): https://deploy-preview-15276--cockroachdb-docs.netlify.app/docs/v22.2/alter-table.html

This is definitely not ready to go, as-is. A few observations:

- Depending on browser window width, side nav doesn't expand enough to show most/all of the nested statement name.
- Because `ALTER TABLE` itself has to open the `alter-table.md` doc, clicking the arrow does not hide the nested subcommands after the list is shown, but only opens `alter-table.md` again.
- When you open a SQL statement doc, the sidebar doesn't actually scroll to show that statement in the nav, which diminishes the utility of using the nav to display the subcommands in context. I may have imagined that this behavior happened at all, and am only realizing it because the SQL statements list is so long (@nickvigilante can you confirm that the nav doesn't behave this way?).

### DIFF
--- a/_includes/v22.2/sidebar-data/reference.json
+++ b/_includes/v22.2/sidebar-data/reference.json
@@ -139,18 +139,6 @@
                 ]
               },
               {
-                "title": "<code>ADD COLUMN</code>",
-                "urls": [
-                  "/${VERSION}/add-column.html"
-                ]
-              },
-              {
-                "title": "<code>ADD CONSTRAINT</code>",
-                "urls": [
-                  "/${VERSION}/add-constraint.html"
-                ]
-              },
-              {
                 "title": "<code>ADD REGION</code> (Enterprise)",
                 "urls": [
                   "/${VERSION}/add-region.html"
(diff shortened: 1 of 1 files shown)


nickvigilante:
@taroface Confirmed as far as I know that navgoco (the software we use for our sidebar) doesn't automatically scroll to the position of the open item. Can you log a ticket for that?

_includes/v22.2/sidebar-data/reference.json
```diff
@@ -250,6 +226,122 @@
                 "title": "<code>ALTER TABLE</code>",
                 "urls": [
                   "/${VERSION}/alter-table.html"
```
nickvigilante, line 228: I suggest omitting the `urls` block here and moving it down as a separate item within the `items` block to prevent all of those subcommands from being selected while the page is open.

ianjevans:
I finally got a good chance to look at the output. I think I'm leaning much more toward including all subcommand docs in the main command document, using `_includes`. It has the added benefit of making the actual command navigation much simpler.

taroface:
> I finally got a good chance to look at the output. I think I'm leaning much more toward including all subcommand docs in the main command document, using `_includes`. It has the added benefit of making the actual command navigation much simpler.

Thanks for looking! Initial feedback from the Queries team was also that all subcommands on one statement page is preferable to what we have now.

taroface:
Closing; this will be addressed soon with https://cockroachlabs.atlassian.net/browse/DOC-5850

PR #15279, opened 2022-10-03
rowid is dropped when adding a new primary key

stbof:
DOC-5381


ajwerner (commented):
Here's the nuance, it's only dropped if you run the statement outside of an explicit transaction, and if there are no secondary indexes, and a laundry-list of other limitations. 

One of the tasks I've been putting off, and can't really put off much longer, is going through and documenting exactly the cases where the declarative schema changer will be used, and the differences it'll buy you. This is one of those differences. I think we'll probably need a page or something to explain the situation. I'll get something up in the next day or so. 

I recognize that it would have been better to have done that sooner. Sorry.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15279#-:-NDUdu1l7G2VRhUT0PX9:boh9w08)* status: :shipit: complete! 0 of 0 LGTMs obtained

PR #15294, opened 2022-10-04
analytics: persist Segment user and group ID & traits with non-standard names

sjbarag:
The Segment JS client library locally persists the user’s ID and traits [1] and the user’s group ID and group properties [2] by default, using predefined names to store those values. Unfortunately, this means that any other Segment integration on the same domain that *doesn't* limit its persistence to a specific subdomain [3] (e.g. Open edX instance hosted on university.cockroachlabs.com) will overwrite any values persisted by any other Segment integration on that domain. In other words, the most recent analytics.identify() call sets the user ID for all other subdomains.

For example, loading cockroachlabs.cloud would correctly set the CC user ID for www.cockroachlabs.com, but loading university.cockroachlabs.com would overwrite the CC user ID until the user reloads cockroachlabs.cloud. All user interactions that happen in between will be attributed to the Open edX user ID, not the CC user ID. Set custom persistence names for user and group cookies and localStorage, so that anything persisted by the uncustomizable university.cockroachlabs.com integration doesn't overwrite this integration's user and group info.

[1] https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/persistence/#user-settings [2] https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/persistence/#group-settings [3] https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/persistence/#cookie-settings


nickvigilante (approved):
Makes sense and looks good to me!

JosephNickdow (approved):

sjbarag:
Thanks for reviewing, friends! 🙂

PR #15296, opened 2022-10-05
Removed pages with redirects from the sitemap

nickvigilante:
- Some pages like `install-cockroachdb.md` and the Python Peewee wrapper serve as in-page redirects and are hurting our SEO. I'm excluding those from the sitemap.
- Other pages that are redirected inside of the `_redirects` file should also be excluded because they are also hurting our SEO.
- We also have issues not being able to index the releases and advisories indexes properly, so Google can't actually process those pages. I've removed the redirects that are causing this.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -6,8 +6,6 @@ layout: null
 / /docs/ 301!
 /docs/v2.2/* /docs/v19.1/:splat 301!
 /docs/managed/* /docs/cockroachcloud/:splat 301!
-/docs/releases/ /docs/releases/index.html 301!
-/docs/advisories/ /docs/advisories/index.html 301!
 /docs/advisories/advisories.html /docs/advisories/index.html 301
 
 # Pages undergoing maintenance
@@ -544,4 +542,3 @@ layout: null
 {% endfor -%}
 
 /docs/frequently-asked-questions.html /docs/stable/frequently-asked-questions.html  301
-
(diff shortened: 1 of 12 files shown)


sitemap.xml
```diff
@@ -3,7 +3,7 @@ layout: null
 ---
 <?xml version="1.0" encoding="UTF-8"?>
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
-  {%- assign stable_pages_names = site.pages | where_exp: "stable_pages", "stable_pages.url contains site.versions['stable']" | map: "name" -%}
+  {%- assign stable_pages_names = site.pages | where_exp: "stable_pages", "stable_pages.url contains site.versions['stable']" | where_exp: "stable_pages", "stable_pages.name != 'index.md'" | map: "name" -%}
```
nickvigilante, line 6: This is here because we want to include /docs/dev/index.md in the sitemap so that any new dev pages can be reached by Google and other search engines. The logic I built here basically says exclude all stable page names that are in the list, but we're excluding something from an exclude list, which means it's being included instead.

_redirects
```diff
@@ -6,8 +6,6 @@ layout: null
 / /docs/ 301!
 /docs/v2.2/* /docs/v19.1/:splat 301!
 /docs/managed/* /docs/cockroachcloud/:splat 301!
-/docs/releases/ /docs/releases/index.html 301!
-/docs/advisories/ /docs/advisories/index.html 301!
```
andf-crl, line 10: Just leaving a comment per our discussion: these redirects may be used by downstream consumers, TSE org, etc. As long as we are OK with potential downstream impact here (we can't know who uses redirects, for example!), then this LGTM. Understand the SEO gains from this now, thanks for the helpful explanations!

andf-crl (approved):
LGTM, thanks!

PR #15094, opened 2022-09-14
Add DROP OWNED BY

stbof:
DOC-4770
DOC-5465  Depends on system privileges.
Blocks https://github.com/cockroachdb/docs/pull/15298

PR #15298, opened 2022-10-05
Add statements supported by declarative schema changer.

stbof:
DOC-4777

Depends on https://github.com/cockroachdb/docs/pull/15094

### DIFF
--- a/v22.2/online-schema-changes.md
+++ b/v22.2/online-schema-changes.md
@@ -64,17 +64,23 @@ If a schema change fails, the schema change job will be cleaned up automatically
 
 The declarative schema changer is the next iteration of how schema changes will be performed in CockroachDB. By planning schema change operations in a more principled manner, the declarative schema changer will ultimately make transactional schema changes more stable. You can identify jobs that are using the declarative schema changer by running [`SHOW JOBS`](show-jobs.html) and finding jobs with a `job_type` of `NEW SCHEMA CHANGE`.
 
-The following statements use the declarative schema changer by default:
+The following statements can use the declarative schema changer, and are enabled to do so by default:
 
+- [`ADD COLUMN`](add-column.html)
+- [`ALTER PRIMARY KEY`](alter-primary-key.html)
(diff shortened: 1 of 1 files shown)


v22.2/online-schema-changes.md
```diff
@@ -64,17 +64,23 @@ If a schema change fails, the schema change job will be cleaned up automatically
 
 The declarative schema changer is the next iteration of how schema changes will be performed in CockroachDB. By planning schema change operations in a more principled manner, the declarative schema changer will ultimately make transactional schema changes more stable. You can identify jobs that are using the declarative schema changer by running [`SHOW JOBS`](show-jobs.html) and finding jobs with a `job_type` of `NEW SCHEMA CHANGE`.
 
-The following statements use the declarative schema changer by default:
+The following statements can use the declarative schema changer, and are enabled to do so by default:
```
ajwerner, line 67: I didn't know about this section. This is good. I think we need to say more about when it is used. In particular, the declarative schema changer is not used in an explicit transaction or a multi-statement implicit transaction; it is only used by default for single statement, implicit transactions. 

There's also a list of special cases for some of these statements where it is not used. We should document them. Let me make a list now.

v22.2/online-schema-changes.md
```diff
@@ -64,17 +64,23 @@ If a schema change fails, the schema change job will be cleaned up automatically
...
 
+- [`ADD COLUMN`](add-column.html)
+- [`ALTER PRIMARY KEY`](alter-primary-key.html)
+- [`COMMENT ON`](comment-on.html)
+- [`CREATE INDEX`](create-index.html)
```
ajwerner, line 72: We didn't implement this one by default.

ajwerner (commented):
Thanks for putting this up. I had a google doc I had started to type but this is easier.

ajwerner:
I guess this depends on https://github.com/cockroachdb/docs/pull/15094. I can take this from here. Thanks for getting this going!

PR #15301, opened 2022-10-05
[DOC-5767] Add info about graceful termination on Kubernetes and CockroachDB Dedicated

mdlinville:
[DOC-5767]

This PR adds info about the Kubernetes `terminationGracePeriodSeconds` setting for both Kubernetes deployments and CockroachDB Dedicated clusters. It also gives some advice about what you should and shouldn't do on Kubernetes deployments, based on conversation with @chrisseto.

TODO: Backport the changes to 22.1 and 21.2 after the text is finalized.

## Out of scope

This content is tricky underneath. This change aims to be surgical and small. For that reason, these items are out of scope:

-  Fixing this page to be less manual-deploy-only.
- Making sense of the page's structure from the source.

## Previews
[v22.2/node-shutdown.md](https://deploy-preview-15301--cockroachdb-docs.netlify.app/docs/v22.2/node-shutdown.html)

## Tests
- [x] Local build
- [x] Linkcheck


chrisseto (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15301)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @mdlinville and @pseudomuto)

___
*[`v22.2/node-shutdown.md` line 32 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiNMJq6-PRqT189yCV:-NDiNMJq6-PRqT189yCW:bimvfhj) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L32)):*
> ```Markdown
> Most of the guidance in this page is most relevant to manual deployments. In a Kubernetes deployment, you don't manually decommission or drain individual nodes. Instead, each cluster node is managed within Kubernetes.
> 
> - Do not attempt to decommission a specific node that is part of a Kubernetes deployment using the advice in this article. This can lead to data loss or loss of quorum in the CockroachDB cluster, and there is generally no need to do so. If you think that you need to decommission a specific node, [contact support](https://support.cockroachlabs.com/hc/en-us) for assistance.
> ```

`This can lead to data loss or loss of quorum in the CockroachDB cluster`

This isn't quite true. It's safe to do so but will lead to a weird state within Kubernetes because of how StatefulSets work. How would you feel about rephrasing this to something along the lines of:

"Due to the nature of StatefulSets, it's only safe to decommission a the Cockroach node with the highest StatefulSet ordinal in preparation for scaling down the StatefulSet. If you think you need to decommission an other node, contact support"
___
*[`v22.2/node-shutdown.md` line 34 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiO1Wf7Xt3zd4Uqrxx:-NDiO1Wf7Xt3zd4Uqrxy:b-ohxnbx) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L34)):*
> ```Markdown
> - Do not attempt to decommission a specific node that is part of a Kubernetes deployment using the advice in this article. This can lead to data loss or loss of quorum in the CockroachDB cluster, and there is generally no need to do so. If you think that you need to decommission a specific node, [contact support](https://support.cockroachlabs.com/hc/en-us) for assistance.
> - While it's possible to drain a CockroachDB node that is part of a Kubernetes deployment, there is generally no need to do so. When you upgrade, downgrade, or change the configuration of a CockroachDB deployment on Kubernetes, you apply the changes using a [rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/), which applies the change to one node at a time. On a given node, Kubernetes sends a `SIGTERM` signal to the `cockroach` process. When the `cockroach` process receives this signal, it drains the node before terminating itself. Kubernetes then applies the updated deployment to the cluster node, restarts the `cockroach` process, and re-joins the cluster. Refer to [Cluster Upgrades](upgrade-cockroachdb-kubernetes.html).
> - When you scale a Kubernetes deployment of CockroachDB down, there is no need to manually decommission or drain nodes. Instead, update the value of `nodes` in the deployment specification and apply the change using a rolling update. Refer to [Cluster Scaling](scale-cockroachdb-kubernetes.html).
> ```

Hmmm, maybe we need to have different call outs for operator managed clusters and manually/helm managed clusters. For clusters provisioned using the operator, this is true. helm/manual clusters will require the user to run the decommission command themselves.
___
*[`v22.2/node-shutdown.md` line 41 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiOL6T2BmSkSyY3Ja1:-NDiOL6T2BmSkSyY3Ja2:b-u384lj) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L41)):*
> ```Markdown
> ### Termination grace period on Kubernetes
> 
> After Kubernetes issues a termination request to the `cockroach` process on a cluster node, it waits for a maximum of the deployment's `terminationGracePeriodSeconds` before forcibly terminating the process. If `terminationGracePeriodSeconds` is too short, the `cockroach` process could be killed before it can shut down cleanly, client applications could be disrupted, and data loss can result.
> ```

Consider linking out to [Kubernetes' Pod lifecycle documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) here?

Based on a recent thread in slack, I'm not sure if there can actually be data loss here? (They mention that CRDB can be kill -9'd) Given that this is a warning, I'm not sure if we care about being 100% accurate.
___
*[`v22.2/node-shutdown.md` line 71 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiPBAS5-xFuA0lw9Mt:-NDiPBAS5-xFuA0lw9Mu:b9arsxc) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L71)):*
> ```Markdown
> Client applications that connect to CockroachDB must handle disconnections gracefully, because from the point of view of the client, cluster maintenance is always a possibility.
> 
> - A client application's connection pool should have a maximum lifetime that is shorter than the Kubernetes deployment's `terminationGracePeriodSeconds` setting, which is always 300 seconds (5 minutes) on {{ site.data.products.dedicated }}.
> ```

Do we have an official stance on how nitty-gritty we get with how CC is deployed? It feels a bit strange to tell users about this aspect of dedicate instead of just providing guidance about what settings to configure in their connection pool.

mdlinville (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15301)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @chrisseto and @pseudomuto)

___
*[`v22.2/node-shutdown.md` line 32 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiNMJq6-PRqT189yCV:-NDjbs_r3pzvS0y2iy8_:b6smf3h) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L32)):*
<details><summary><i>Previously, chrisseto (Chris Seto) wrote…</i></summary><blockquote>

`This can lead to data loss or loss of quorum in the CockroachDB cluster`

This isn't quite true. It's safe to do so but will lead to a weird state within Kubernetes because of how StatefulSets work. How would you feel about rephrasing this to something along the lines of:

"Due to the nature of StatefulSets, it's only safe to decommission a the Cockroach node with the highest StatefulSet ordinal in preparation for scaling down the StatefulSet. If you think you need to decommission an other node, contact support"
</blockquote></details>

I reworked this bit and integrated your suggestion.
___
*[`v22.2/node-shutdown.md` line 34 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiO1Wf7Xt3zd4Uqrxx:-NDjbvFS2z2IyBwoTfiJ:b-b0ah59) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L34)):*
<details><summary><i>Previously, chrisseto (Chris Seto) wrote…</i></summary><blockquote>

Hmmm, maybe we need to have different call outs for operator managed clusters and manually/helm managed clusters. For clusters provisioned using the operator, this is true. helm/manual clusters will require the user to run the decommission command themselves.
</blockquote></details>

I reworked this to be more clear.
___
*[`v22.2/node-shutdown.md` line 41 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiOL6T2BmSkSyY3Ja1:-NDjcY-88RDRDpVr-BPB:b-ti0onm) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L41)):*
> Kubernetes' Pod lifecycle documentation
Added the link and reworked the text.
___
*[`v22.2/node-shutdown.md` line 71 at r1](https://reviewable.io/reviews/cockroachdb/docs/15301#-NDiPBAS5-xFuA0lw9Mt:-NDjceon5LmFhsgyJvdv:b-t8sa3m) ([raw file](https://github.com/cockroachdb/docs/blob/a86298f528127f53c12b3b2c59e3c0b5c58c4643/v22.2/node-shutdown.md#L71)):*
<details><summary><i>Previously, chrisseto (Chris Seto) wrote…</i></summary><blockquote>

Do we have an official stance on how nitty-gritty we get with how CC is deployed? It feels a bit strange to tell users about this aspect of dedicate instead of just providing guidance about what settings to configure in their connection pool.
</blockquote></details>

I reworked this not to talk about the underlying Kubernetes config for Dedicated.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- How to [perform node shutdown](#perform-node-shutdown) by manually draining or decommissioning a node.
 
 <div class="filters filters-big clearfix">
     <button class="filter-button" data-scope="drain">Drain</button>
     <button class="filter-button" data-scope="decommission">Decommission</button>
 </div>
 
+## Decommissioning and draining on Kubernetes
+
+Most of the guidance in this page is most relevant to manual deployments that don't use Kubernetes. If you use Kubernetes to deploy CockroachDB, draining and decommissioning work the same way, but Kubernetes handles them on your behalf. In a Kubernetes deployment, you don't usually decommission or drain individual nodes manually. Instead, each cluster node is managed within Kubernetes.
```
mdlinville, line 30: The Oxford comma warning is a false positive and won't block merging this PR.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+- Although the `kubectl drain` command is used for routine maintenance of Kubernetes clusters, it has direct relevance to the concept of draining a CockroachDB node. For details on this command, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
+
+Refer to the sections about [Termination grace period on Kubernetes](#termination-grace-period-on-kubernetes) and [Client connection timeout](#client-connection-timeout), as well as the pages about [Cluster upgrades](upgrade-cockroachdb-kubernetes.html) and [Cluster scaling](scale-cockroachdb-kubernetes.html) on Kubernetes.
```
taroface, line 44: Can you be more specific about what kinds of information people are going to find on those pages? Generally we try to write along the lines of "For more information about X topic, refer to...".

```suggestion
Refer to the sections about [Termination grace period on Kubernetes](#termination-grace-period-on-kubernetes) and [Client connection timeout](#client-connection-timeout), as well as to [Cluster Upgrades](upgrade-cockroachdb-kubernetes.html) and [Cluster Scaling](scale-cockroachdb-kubernetes.html) on Kubernetes.
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+### Termination grace period on Kubernetes
```
taroface, line 46: This section needs to be reconciled with the existing https://www.cockroachlabs.com/docs/stable/node-shutdown.html#termination-grace-period, which already mentions K8s and includes separate notes on this setting. Perhaps remove the info there (and reuse it here if still appropriate) and link to this section.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+After Kubernetes issues a termination request to the `cockroach` process on a cluster node, it waits for a maximum of the deployment's [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) before forcibly terminating the process. If `terminationGracePeriodSeconds` is too short, the `cockroach` process could be killed before it can shut down cleanly, client applications could be disrupted, and data loss can result.
```
taroface, line 48:
```suggestion
After Kubernetes issues a termination request to the `cockroach` process on a cluster node, it waits for a maximum of the deployment's [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) before forcibly terminating the process. If `terminationGracePeriodSeconds` is too short, the `cockroach` process could be terminated before it can shut down cleanly, client applications could be disrupted, and data loss can result.
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+If undefined, Kubernetes sets `terminationGracePeriodSeconds` to 30 seconds. This is too short for the `cockroach` process to terminate gracefully before Kubernetes terminates it forcibly. Do not set `terminationGracePeriodSeconds` to `0`, which prevents Kubernetes from detecting and terminating a stuck pod.
```
taroface, line 50: I think we use "terminate" as a synonym for "kill", which is not graceful.

```suggestion
If undefined, Kubernetes sets `terminationGracePeriodSeconds` to 30 seconds. This is too short for the `cockroach` process to stop gracefully before Kubernetes terminates it forcibly. Do **not** set `terminationGracePeriodSeconds` to `0`, which prevents Kubernetes from detecting and terminating a stuck pod.
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+For clusters deployed using the CockroachDB public operator, `terminationGracePeriodSeconds` defaults to 300 seconds (5 minutes).
```
taroface, line 52:
```suggestion
For clusters deployed using the CockroachDB Kubernetes Operator, `terminationGracePeriodSeconds` defaults to 300 seconds (5 minutes).
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+For clusters deployed using the CockroachDB Helm chart or a manual StatefulSet, the default depends upon the values file or manifest you used when you created the cluster.
+
+Cockroach Labs recommends that you:
+
+- Set `terminationGracePeriodSeconds` to no shorter than 300 seconds (5 minutes). This recommendation has been validated over time for many production workloads. In most cases, a value higher than 600 seconds (10 minutes) is not required. If CockroachDB takes longer than 10 minutes to gracefully terminate, this may indicate an underlying configuration problem. Test the value you select against representative workloads before rolling out the change to production clusters.
```
taroface, line 57:
```suggestion
- Set `terminationGracePeriodSeconds` to no shorter than 300 seconds (5 minutes). This recommendation has been validated over time for many production workloads. In most cases, a value higher than 600 seconds (10 minutes) is not required. If CockroachDB takes longer than 10 minutes to gracefully shut down, this may indicate an underlying configuration problem. Test the value you select against representative workloads before rolling out the change to production clusters.
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- Set `terminationGracePeriodSeconds` to be at least 5 seconds longer than the configured [drain timeout](#server-shutdown-drain_wait), to allow the node to complete draining before Kubernetes removes the Kubernetes pod for the CockroachDB node.
+- Ensure that the **sum** of all `server.shutdown.*` settings for the CockroachDB cluster do not exceed the deployment's `terminationGracePeriodSeconds`, to reduce the likelihood that a node must be terminated forcibly.
+
+  - [server.shutdown.drain_wait](#server-shutdown-drain_wait)
+  - [server.shutdown.connection_wait](#server-shutdown-connection_wait)
+  - [server.shutdown.query_wait](#server-shutdown-query_wait)
```
taroface, line 63: This probably needs to be `server.shutdown.query_wait` times two. Refer to the callout about drain timeouts elsewhere in this doc.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+  - [server.shutdown.lease_transfer_wait](#server-shutdown-lease_transfer_wait)
+
```
taroface, line 65: Add code backticks to the linked settings.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+  For more information about these settings, refer to [Cluster settings](#cluster-settings). Refer also to the [Kubernetes documentation about pod termination](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination).
+
+### Client connection timeout
+
+Client applications that connect to CockroachDB must handle disconnections gracefully, because from the point of view of the client, cluster maintenance is always a possibility.
+
+- A client application's connection pool should have a maximum lifetime that is shorter than the Kubernetes deployment's `terminationGracePeriodSeconds` setting. Cockroach Labs recommends aligning the value for [server.shutdown.connection_wait](#server-shutdown-connection_wait) with the connection pool's maximum lifetime.
+- If a client application is connected to a cluster node that is draining, it will be disconnected, and must include the ability to try to connect again. Refer to [Connection Retry Logic](#connection-retry-loop).
```
taroface, line 73:
```suggestion
- If a client application is connected to a cluster node that is draining, it will be disconnected, and must include the ability to try to connect again. Refer to [Connection retry loop](#connection-retry-loop).
```

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- Ensure that the **sum** of all `server.shutdown.*` settings for the CockroachDB cluster do not exceed the deployment's `terminationGracePeriodSeconds`, to reduce the likelihood that a node must be terminated forcibly.
+
+  - [server.shutdown.drain_wait](#server-shutdown-drain_wait)
+  - [server.shutdown.connection_wait](#server-shutdown-connection_wait)
+  - [server.shutdown.query_wait](#server-shutdown-query_wait)
+  - [server.shutdown.lease_transfer_wait](#server-shutdown-lease_transfer_wait)
+
+  For more information about these settings, refer to [Cluster settings](#cluster-settings). Refer also to the [Kubernetes documentation about pod termination](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination).
+
+### Client connection timeout
```
taroface, line 68: I'm generally not clear on why client connections need a dedicated section under the Kubernetes header. For example, you could include `- A client application's connection pool should have a maximum lifetime that is shorter than the Kubernetes deployment's `terminationGracePeriodSeconds` setting.` in the list of considerations in the preceding section.

v22.2/node-shutdown.md
```diff
@@ -264,7 +317,7 @@ If the `cockroach` process has not terminated at the end of the grace period, a
 
 - When using [`systemd`](https://www.freedesktop.org/wiki/Software/systemd/) to run CockroachDB as a service, set the termination grace period with [`TimeoutStopSec`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec=) setting in the service file.
 
-- When using [Kubernetes](kubernetes-overview.html) to orchestrate CockroachDB, set the termination grace period with `terminationGracePeriodSeconds` in the [StatefulSet manifest](deploy-cockroachdb-with-kubernetes.html?filters=manual#configure-the-cluster).
+- When using [Kubernetes](kubernetes-overview.html) to orchestrate CockroachDB, set the termination grace period with `terminationGracePeriodSeconds` in the [StatefulSet manifest](deploy-cockroachdb-with-kubernetes.html?filters=manual#configure-the-cluster). Refer to [Decommissioning and draining on Kubernetes](#decommissioning-and-draining-on-kubernetes).
```
taroface, line 320: As mentioned above, maybe just remove this to avoid duplication with the other section you added?

taroface (changes requested):
I'm in general agreement with your approach here. My comments are primarily about ways to make this update more surgical and small.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+    - If you deployed a cluster using the CockroachDB Kubernetes operator, there is no advantage to this approach. Instead, update the pod spec for the deployment to reduce the value of `nodes:` and apply the change using a [rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/). Kubernetes will notice that there are now too many nodes and will reduce them and clean up their storage automatically.
+
+    - If you deployed the cluster using Helm or a manual StatefulSet, the best way to scale down a cluster is to manually decommission and drain the highest-order node. After that node is decommissioned, drained, and terminated, you can repeat the process to further reduce the cluster's size.
```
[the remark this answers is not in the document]
mdlinville: The Kubernetes Operator team says that manually decommissioning and draining a node on Kubernetes doesn't usually make sense and that the right way to do it is to just scale down the cluster. But if you deployed using Helm or a manual StatefulSet, and you need to scale down the cluster, then you need to do it manually, one node at a time, starting from the highest-order node. Trying to decommission a node that isn't the highest-order node won't fix the PVC that is associated with the node.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+    Refer to [Cluster Scaling](scale-cockroachdb-kubernetes.html).
+
+- There is generally no need to manually drain a node that is not being decommissioned, regardless of how you deployed the cluster in Kubernetes. When you upgrade, downgrade, or change the configuration of a CockroachDB deployment on Kubernetes, you apply the changes using a [rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/), which applies the change to one node at a time. On a given node, Kubernetes sends a `SIGTERM` signal to the `cockroach` process. When the `cockroach` process receives this signal, it starts draining itself. After draining is complete or the [termination grace period](#termination-grace-period-on-kubernetes) expires (whichever happens first), Kubernetes terminates the `cockroach` process and then removes the node from the Kubernetes cluster. Kubernetes then applies the updated deployment to the cluster node, restarts the `cockroach` process, and re-joins the cluster. Refer to [Cluster Upgrades](upgrade-cockroachdb-kubernetes.html).
```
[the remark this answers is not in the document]
mdlinville: In this case, "manually drain" would mean connecting to the cluster node directly, sending a SIGTERM to the cockroach process, and waiting for that to finish before manually evicting the pod. I updated the text a bit.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+- Although the `kubectl drain` command is used for routine maintenance of Kubernetes clusters, it has direct relevance to the concept of draining a CockroachDB node. For details on this command, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```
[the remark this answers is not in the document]
mdlinville: Sure, `kubectl drain` effectively puts a node into "maintenance mode" by terminating all of its pods and then waiting for human intervention. In a production Kubernetes cluster, most maintenance isn't done manually by humans, but is done by adjusting the deployment and applying the adjustment using a rolling update.

v22.2/node-shutdown.md
mdlinville: I think much of the content in this section is reworked from that previous content. I found a few duplicated bits to remove.

v22.2/node-shutdown.md
mdlinville: Hmm, I'm not sure about that. The setting is called Termination Grace Period. But I'm ok with saying "stop" here.

v22.2/node-shutdown.md
mdlinville: I took away the heading and turned the section into a bullet.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
-
-Also see our documentation for [cluster upgrades](upgrade-cockroachdb-kubernetes.html) and [cluster scaling](scale-cockroachdb-kubernetes.html) on Kubernetes.
-{{site.data.alerts.end}}
-
 This page describes:
 
-- The details of the [node shutdown sequence](#node-shutdown-sequence).
-- How to [prepare for graceful shutdown](#prepare-for-graceful-shutdown) by implementing connection retry logic and coordinating load balancer, process manager, and cluster settings.
-- How to [perform node shutdown](#perform-node-shutdown) by draining or decommissioning the node.
+- How node shutdown works when CockroachDB is deployed using [Kubernetes](#decommissioning-and-draining-on-kubernetes) or in a [{{ site.data.products.dedicated }} cluster](#decommissioning-and-draining-on-cockroachdb-dedicated).
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- The details of the [node shutdown sequence](#node-shutdown-sequence) from the point of view of the `cockroach` process on a CockroachDB cluster node.
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- How to [prepare for graceful shutdown](#prepare-for-graceful-shutdown) by implementing connection retry logic and coordinating load balancer, client application server, process manager, and cluster settings.
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+- How to [perform node shutdown](#perform-node-shutdown) by manually draining or decommissioning a node.
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
 <div class="filters filters-big clearfix">
     <button class="filter-button" data-scope="drain">Drain</button>
     <button class="filter-button" data-scope="decommission">Decommission</button>
 </div>
 
+## Decommissioning and draining on Kubernetes
+
+Most of the guidance in this page is most relevant to manual deployments that don't use Kubernetes. If you use Kubernetes to deploy CockroachDB, draining and decommissioning work the same way, but Kubernetes handles them on your behalf. In a Kubernetes deployment, you don't usually decommission or drain individual nodes manually. Instead, each cluster node is managed within Kubernetes.
+
+- Whether you deployed a cluster using the CockroachDB Operator, Helm, or a manual StatefulSet, the resulting deployment is a StatefulSet. Due to the nature of StatefulSets, it's safe to decommission **only** the Cockroach node with the highest StatefulSet ordinal in preparation for scaling down the StatefulSet. If you think you need to decommission any other node, consider the following advice and [contact support](https://support.cockroachlabs.com/hc/en-us) for assistance.
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
...
+
+    - If you deployed a cluster using the CockroachDB Kubernetes operator, there is no advantage to this approach. Instead, update the pod spec for the deployment to reduce the value of `nodes:` and apply the change using a [rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/). Kubernetes will notice that there are now too many nodes and will reduce them and clean up their storage automatically.
```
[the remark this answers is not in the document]
mdlinville: Done.

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
mdlinville: Done.

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
mdlinville: Done

v22.2/node-shutdown.md
```diff
@@ -13,23 +13,76 @@ There are two ways to handle node shutdown:
 
 - To permanently remove the node from the cluster, **decommission** the node and then terminate the `cockroach` process. This is done when scaling down a cluster or reacting to hardware failures. With a decommission, the data is moved out of the node. Replica rebalancing creates network traffic throughout the cluster, which makes a decommission heavyweight.
 
-{{site.data.alerts.callout_success}}
-This guidance applies to manual deployments. If you have a Kubernetes deployment, terminating the `cockroach` process is handled through the Kubernetes pods. The `kubectl drain` command is used for routine cluster maintenance. For details on this command, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```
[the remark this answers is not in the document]
mdlinville: DOne.

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+- Set `terminationGracePeriodSeconds` to be at least 5 seconds longer than the configured [drain timeout](#server-shutdown-drain_wait), to allow the node to complete draining before Kubernetes removes the Kubernetes pod for the CockroachDB node.
+- Ensure that the **sum** of the following `server.shutdown.*` settings for the CockroachDB cluster do not exceed the deployment's `terminationGracePeriodSeconds`, to reduce the likelihood that a node must be terminated forcibly.
+
+  - [`server.shutdown.drain_wait`](#server-shutdown-drain_wait)
+  - [`server.shutdown.connection_wait`](#server-shutdown-connection_wait)
+  - [`server.shutdown.query_wait`](#server-shutdown-query_wait) times two
+  - [`server.shutdown.lease_transfer_wait`](#server-shutdown-lease_transfer_wait)
+
+  For more information about these settings, refer to [Cluster settings](#cluster-settings). Refer also to the [Kubernetes documentation about pod termination](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination).
```
taroface, line 837: Nit: This should be indented to the same level as the preceding bullets.

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+- Whether you deployed a cluster using the CockroachDB Operator, Helm, or a manual StatefulSet, the resulting deployment is a StatefulSet. Due to the nature of StatefulSets, it's safe to decommission **only** the Cockroach node with the highest StatefulSet ordinal in preparation for scaling down the StatefulSet. If you think you need to decommission any other node, consider the following recommendations and [contact Support](https://support.cockroachlabs.com/hc/en-us) for assistance.
+
+    - If you deployed a cluster using the [CockroachDB Kubernetes Operator](deploy-cockroachdb-with-kubernetes.html), there is no advantage to manually decommissioning the node with the highest StatefulSet ordinal. Instead, update the specification for the Kubernetes deployment to reduce the value of `nodes:` and apply the change using a [rolling update](https://kubernetes.io/docs/tutorials/kubernetes-basics/update/update-intro/). Kubernetes will notice that there are now too many nodes and will reduce them and clean up their storage automatically.
```
taroface, line 805: I'm still confused by the use of "manually" to describe decommissioning in K8s, since the intro paragraph here mentions that "you don't usually decommission or drain a node interactively ... Instead, each cluster node is managed within Kubernetes."

Can you clarify what "manually" means in this context?

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+    - If you deployed the cluster using [Helm](deploy-cockroachdb-with-kubernetes.html?filters=helm) or a [manual StatefulSet](deploy-cockroachdb-with-kubernetes.html?filters=manual), the best way to scale down a cluster is to manually decommission and drain the highest-order node. After that node is decommissioned, drained, and terminated, you can repeat the process to further reduce the cluster's size.
```
taroface, line 807: Ditto

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
 On the **Cluster Overview** page of the DB Console, the [node status](ui-cluster-overview-page.html#node-status) of the node should be `LIVE`. After a few minutes, you should see replicas rebalanced to the nodes.
 </section>
 
+## Decommissioning and draining on Kubernetes
+
+Most of the guidance in this page is most relevant to manual deployments that don't use Kubernetes. If you use Kubernetes to deploy CockroachDB, draining and decommissioning work the same way for the `cockroach` process, but Kubernetes handles them on your behalf. In a Kubernetes deployment, you don't usually decommission or drain a node interactively. In fact, if you attempt to do so, Kubernetes will automatically undo your change so that the cluster remains in its desired state as defined by the Kubernetes deployment. Instead, each cluster node is managed within Kubernetes.
```
taroface, line 801: Avoid contractions where a clear instruction is given - see style guide on this.

```suggestion
Most of the guidance in this page is most relevant to manual deployments that don't use Kubernetes. If you use Kubernetes to deploy CockroachDB, draining and decommissioning work the same way for the `cockroach` process, but Kubernetes handles them on your behalf. In a Kubernetes deployment, you do not usually decommission or drain a node interactively. In fact, if you attempt to do so, Kubernetes will automatically undo your change so that the cluster remains in its desired state as defined by the Kubernetes deployment. Instead, each cluster node is managed within Kubernetes.
```

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+- Although the `kubectl drain` command is used for manual maintenance of Kubernetes clusters, it has little direct relevance to the concept of draining a CockroachDB node. The `kubectl drain` command gracefully terminates each pod running on a node so that the node can be shut down (in the case of physical hardware) or deleted (in the case of a virtual machine). For details on this command, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```
taroface, line 813: The terminology around "node" gets confusing. Is this correct (i.e., "node" refers to a K8s instance)?

```suggestion
- Although the `kubectl drain` command is used for manual maintenance of Kubernetes clusters, it has little direct relevance to the concept of draining a CockroachDB node. The `kubectl drain` command gracefully terminates each pod running on a Kubernetes node so that it can be shut down (in the case of physical hardware) or deleted (in the case of a virtual machine). For details on this command, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/).
```

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+Refer to the sections about [Termination grace period on Kubernetes](#termination-grace-period-on-kubernetes). For more details about managing CockroachDB on Kubernetes, refer to [Cluster upgrades](upgrade-cockroachdb-kubernetes.html) and [Cluster scaling](scale-cockroachdb-kubernetes.html).
```
taroface, line 815:
```suggestion
Refer to [Termination grace period on Kubernetes](#termination-grace-period-on-kubernetes). For more details about managing CockroachDB on Kubernetes, refer to [Cluster upgrades](upgrade-cockroachdb-kubernetes.html) and [Cluster scaling](scale-cockroachdb-kubernetes.html).
```

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+### Termination grace period on Kubernetes
+
+After Kubernetes issues a termination request to the `cockroach` process on a cluster node, it waits for a maximum of the deployment's [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) before forcibly terminating the process. If `terminationGracePeriodSeconds` is too short, the `cockroach` process could be terminated before it can shut down cleanly and client applications could be disrupted.
```
taroface, line 819:
```suggestion
After Kubernetes issues a termination request to the `cockroach` process on a cluster node, it waits for a maximum of the deployment's [`terminationGracePeriodSeconds`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) before forcibly terminating the process. If `terminationGracePeriodSeconds` is too short, the `cockroach` process may be terminated before it can shut down cleanly and client applications may be disrupted.
```

v22.2/node-shutdown.md
```diff
@@ -800,6 +796,62 @@ If the decommissioning node has already reached the [draining stage](#draining),
...
+
+<a id="decommissioning-and-draining-on-cockroachdb-dedicated"></a>
+
+## Decommissioning and draining on {{ site.data.products.dedicated }}
+
+Most of the guidance in this page is most relevant to manual deployments, although decommissioning and draining work the same way behind the scenes in a {{ site.data.products.dedicated }} cluster. {{ site.data.products.dedicated }} clusters have a termination grace period of 300 seconds (5 minutes). This setting is not configurable.
+
+Client applications that connect to CockroachDB must handle disconnections gracefully, because from the point of view of the client, cluster maintenance is always a possibility.
+
+- Client application that connect to {{ site.data.products.dedicated }} should use connection pools that have a maximum lifetime that is shorter than 300 seconds (5 minutes).
```
taroface, line 852:
```suggestion
- Client applications that connect to {{ site.data.products.dedicated }} should use connection pools that have a maximum lifetime that is shorter than 300 seconds (5 minutes).
```

taroface (approved):
A couple nits, and re-requesting a clarification, but otherwise LGTM assuming eng approved the tech details here.

v22.2/node-shutdown.md
mdlinville: Yes, you would not SSH into the Kubernetes node and issue a SIGTERM to the `cockroach` process. Instead. Kubernetes notices that the cluster now has too many nodes and selects one of them to send SIGTERM to. The cockroach process receives the SIGTERM and does the same thing it would have done if a sysadmin issued it manually.

PR #15309, opened 2022-10-06
remove finalize modal step

taroface:
Reflects UX change in https://github.com/cockroachlabs/managed-service/pull/10148


Jeremyyang920 (approved):
Thanks for updating!

PR #15326, opened 2022-10-10
Clarify how Cloud interacts with cloud providers

ianjevans:
Fixes DOC-4584.


mdlinville (approved):
LGTM. I wonder if it makes sense to add this info to other Cloud topics too.

ianjevans:
> LGTM. I wonder if it makes sense to add this info to other Cloud topics too.

I searched through the Cloud topics for "cloud provider" and these seemed to be the most appropriate places.

PR #15348, opened 2022-10-12
Force any /dev/ canonical pages to redirect to /dev/

nickvigilante:
Currently, we have some new pages, such as `ALTER FUNCTION` to support UDFs. We want these pages to be part of our search index, but Google can't index them because the canonical of these pages is set to v22.2/ instead of dev/. This PR fixes that.


andf-crl (approved):
Interesting! And clever fix! LGTM.

PR #15349, opened 2022-10-12
[DOC-4826] Document Cloud organization audit log API

mdlinville:
[DOC-4826]

Reviewers:
- Please review directly in Github rather than ReviewBoard.

## Previews

[cockroachcloud/cloud-api.md](https://deploy-preview-15349--cockroachdb-docs.netlify.app/docs/cockroachcloud/cloud-api.html)
[cockroachcloud/cloud-org-audit-logs.md](https://deploy-preview-15349--cockroachdb-docs.netlify.app/docs/cockroachcloud/cloud-org-audit-logs.html)

## Tests
- [x] Local build
- [x] Linkcheck

### DIFF
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -413,7 +413,7 @@ Where:
 
 A request that includes no parameters exports roughly 200 entries in ascending order, starting from when your {{ site.data.products.db }} organization was created.
 
-If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned. If the results are returned in ascending order, `next_starting_from` is always returned.
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned when it reaches the time when the  {{ site.data.products.db }} organization was created. If the results are returned in ascending order, `next_starting_from` is always returned.
 
 {% include_cached copy-clipboard.html %}
 ~~~ json
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -401,19 +401,23 @@ To export audit logs for activities and events related to your Cloud organizatio
 {% include_cached copy-clipboard.html %}
 ~~~ shell
 curl --request GET \
-  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom={timestamp},SortOrder={sort_order},limit={limit}' \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp},sortOrder={sort_order},limit={limit}' \
   --header 'Authorization: Bearer {secret_key}'
 ~~~
 
 Where:
 
-  - `{timestamp}` is an [RFC3339 timestamp](https://www.ietf.org/rfc/rfc3339.txt) that indicates the first log entry to fetch. If unspecified, defaults to the time when the Cloud organization was created if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
-  - `{sort_order}` is either `ascending` (the default) or `descending`.
+  - `{timestamp}` is an [RFC3339 timestamp](https://www.ietf.org/rfc/rfc3339.txt) that indicates the first log entry to fetch. If unspecified, defaults to the time when the Cloud organization was created if `{sort_order}` is `ASC`, or the current time if `{sort_order}` is `DESC`.
+  - `{sort_order}` is either `ASC` (the default) or `DESC`.
   - `{limit}` indicates roughly how many entries to return. If any entries would be returned for a timestamp, all entries for that timestamp are always returned. Defaults to `200`.
 
 A request that includes no parameters exports roughly 200 entries in ascending order, starting from when your {{ site.data.products.db }} organization was created.
 
-If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned when it reaches the time when the  {{ site.data.products.db }} organization was created. If the results are returned in ascending order, `next_starting_from` is always returned.
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and, depending on the circumstances, a `next_starting_from` field.
+
+- If `{sort_order}` is `ASC`, next_starting_from` is always returned.
+- If `{sort_order}` is `DESC`, then `next_starting_from` is returned as long as earlier audit logs are available. It is not returned when the earliest log entry is reached (when the {{ site.data.products.db }} organization was created).
+
 
 {% include_cached copy-clipboard.html %}
 ~~~ json
@@ -421,7 +425,7 @@ If the request was successful, the client will receive a JSON array consisting o
   "entries": [
     "{entry_array}"
   ],
-  "NextStartingFrom": "{timestamp}"
+  "next_starting_from": "{timestamp}"
 }
 ~~~
 
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -390,6 +390,46 @@ Where:
 
 If the `DELETE` request was successful the client will not receive a response payload.
 
+<a id="cloud-audit-logs"></a>
+
+## Export {{ site.data.products.db }} audit logs
+
+{% include feature-phases/preview-opt-in.md %}
+
+To export audit logs for activities and events related to your {{ site.data.products.db }} organization, send a `GET` request to the `/v1/auditlogevents` endpoint. The service account associated with the secret key must have `ADMIN` [permission](console-access-management.html#service-accounts).
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom={timestamp},SortOrder={sort_order},limit={limit}' \
+  --header 'Authorization: Bearer {secret_key}'
+~~~
+
+Where:
+
+  - `{timestamp}` indicates the first log entry to fetch. If unspecified, `{timestamp}` is set to the beginning of time if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
+  - `{sort_order}` is either `ascending` (the default) or `descending`.
+  - `{limit}` indicates roughly how many entries to return. It is expected that slightly more results will be returned, because if any entries exist for a timestamp, all entries for that timestamp are always fetched. It defaults to `200`.
+
+A request that includes no parameters exports roughly 200 entries are returned in ascending order, starting from the first entry for your {{ site.data.products.db }} organization.
+
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field.
+
+{% include_cached copy-clipboard.html %}
+~~~ json
+{
+  "entries": [
+    "{entry_array}"
+  ],
+  "NextStartingFrom": "{timestamp}"
+}
+~~~
+
+Where:
+
+  - `{entry_array}` is a structured JSON array of audit log entries.
+  - `{timestamp}` indicates the timestamp to send to export the next batch of results.
+
 ## List all clusters in an organization
 
 To list all active clusters within an organization, send a `GET` request to the `/v1/clusters` endpoint. The service account associated with the secret key must have `ADMIN` or `READ` [permission](console-access-management.html#service-accounts) to list an organization's clusters.
@@ -514,7 +554,7 @@ Where:
   {{site.data.alerts.callout_info}}
   The cluster ID used in the Cloud API is different than the routing ID used when [connecting to clusters](connect-to-a-serverless-cluster.html).
   {{site.data.alerts.end}}
-  
+
 If the request was successful, the client will receive a list of SQL users.
 
 ~~~ json
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -392,7 +392,7 @@ If the `DELETE` request was successful the client will not receive a response pa
 
 <a id="cloud-audit-logs"></a>
 
-## Export {{ site.data.products.db }} Organization audit logs
+## Export Cloud Organization audit logs
 
 {% include feature-phases/preview-opt-in.md %}
 
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -392,11 +392,11 @@ If the `DELETE` request was successful the client will not receive a response pa
 
 <a id="cloud-audit-logs"></a>
 
-## Export {{ site.data.products.db }} audit logs
+## Export {{ site.data.products.db }} Organization audit logs
 
 {% include feature-phases/preview-opt-in.md %}
 
-To export audit logs for activities and events related to your {{ site.data.products.db }} organization, send a `GET` request to the `/v1/auditlogevents` endpoint. The service account associated with the secret key must have `ADMIN` [permission](console-access-management.html#service-accounts).
+To export audit logs for activities and events related to your Cloud organization, send a `GET` request to the `/v1/auditlogevents` endpoint. The service account associated with the secret key must have `ADMIN` [permission](console-access-management.html#service-accounts).
 
 {% include_cached copy-clipboard.html %}
 ~~~ shell
@@ -407,13 +407,13 @@ curl --request GET \
 
 Where:
 
-  - `{timestamp}` indicates the first log entry to fetch. If unspecified, `{timestamp}` is set to the beginning of time if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
+  - `{timestamp}` is an [RFC3339 timestamp](https://www.ietf.org/rfc/rfc3339.txt) that indicates the first log entry to fetch. If unspecified, defaults to the time when the Cloud organization was created if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
   - `{sort_order}` is either `ascending` (the default) or `descending`.
-  - `{limit}` indicates roughly how many entries to return. It is expected that slightly more results will be returned, because if any entries exist for a timestamp, all entries for that timestamp are always fetched. It defaults to `200`.
+  - `{limit}` indicates roughly how many entries to return. If any entries would be returned for a timestamp, all entries for that timestamp are always returned. Defaults to `200`.
 
-A request that includes no parameters exports roughly 200 entries are returned in ascending order, starting from the first entry for your {{ site.data.products.db }} organization.
+A request that includes no parameters exports roughly 200 entries in ascending order, starting from when your {{ site.data.products.db }} organization was created.
 
-If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field.
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned. If the results are returned in ascending order, `next_starting_from` is always returned.
 
 {% include_cached copy-clipboard.html %}
 ~~~ json
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -401,7 +401,7 @@ To export audit logs for activities and events related to your Cloud organizatio
 {% include_cached copy-clipboard.html %}
 ~~~ shell
 curl --request GET \
-  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp},sortOrder={sort_order},limit={limit}' \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp}&sortOrder={sort_order}&limit={limit}' \
   --header 'Authorization: Bearer {secret_key}'
 ~~~
 
--- a/cockroachcloud/cloud-api.md
+++ b/cockroachcloud/cloud-api.md
@@ -402,7 +402,8 @@ To export audit logs for activities and events related to your Cloud organizatio
 ~~~ shell
 curl --request GET \
   --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp}&sortOrder={sort_order}&limit={limit}' \
-  --header 'Authorization: Bearer {secret_key}'
+  --header 'Authorization: Bearer {secret_key}' \
+  --header 'Cc-Version: {api_version}'
 ~~~
 
 Where:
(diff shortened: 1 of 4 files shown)


cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,46 @@ Where:
...
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom={timestamp},SortOrder={sort_order},limit={limit}' \
+  --header 'Authorization: Bearer {secret_key}'
+~~~
+
+Where:
+
+  - `{timestamp}` indicates the first log entry to fetch. If unspecified, `{timestamp}` is set to the beginning of time if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
```
abhinavg6, line 410: Again, should contextualize "beginning of time" in terms of the lifetime of an organization and our retention policy (if any), based on @rgcase 's response to similar comments in other file.

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,46 @@ Where:
...
+  - `{sort_order}` is either `ascending` (the default) or `descending`.
+  - `{limit}` indicates roughly how many entries to return. It is expected that slightly more results will be returned, because if any entries exist for a timestamp, all entries for that timestamp are always fetched. It defaults to `200`.
```
abhinavg6, line 412: This statement is not very clear - "It is expected that slightly more results will be returned, because if any entries exist for a timestamp, all entries for that timestamp are always fetched". Are we saying that we could return more than 200 results in case of default limit? Basically what is "more results" relative to?

cc @rgcase

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,46 @@ Where:
...
+
+A request that includes no parameters exports roughly 200 entries are returned in ascending order, starting from the first entry for your {{ site.data.products.db }} organization.
```
abhinavg6, line 414:
```suggestion
A request that includes no parameters exports roughly 200 entries in ascending order, starting from the first entry for your {{ site.data.products.db }} organization.
```

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,46 @@ Where:
...
+
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field.
```
abhinavg6, line 416: @rgcase Would we return "next_starting_from" even when there no more results to return?

cockroachcloud/cloud-api.md
abhinavg6, line 410: Should we also refer to the expected format of "timestamp" here, explaining the different parts of it - year, month, day, hour, minute, seconds, microseconds?

abhinavg6 (changes requested):
Done one full pass. Mainly one meta question that affects potential changes at a few places, and then certain minor suggestions on referring to "cloud organization" rather than the overall cloud or dedicated product in this case.

cockroachcloud/cloud-api.md
mdlinville: If I understand it correctly: There is a bucket for each timestamp, and if your results include even a single item in a bucket, you'll instead get the whole bucket.

rgcase: Yes that's the idea. It's not expected that slightly more *will* be returned, it's just possible but honestly probably so unlikely that it'll never happen.

The reason for this is to make pagination work. If the user passed in a timestamp `t`, it'll return events with timestamp greater (or less) than `t`. But if we get unlucky and there are multiple events with the same timestamp, and the boundary lands between them, then the next attempt to fetch events using the latest timestamp from the results will end up skipping over the later of the events with the same timestamp.

We're talking about millisecond precision timestamps here, so in practice we'll probably never log two events with exactly the same timestamp, and even less likely, two events with the same timestamp in the same organization. Even if that were to happen, it would have to land exactly at the 200th (or whatever limit they pass) event in the response for this to even happen.

It's possible I drew too much attention to this scenario for how unlikely it is to happen, so feel free to soften it a bit or just mention it in passing that there's a remote possibility that the API will return more results than the `limit`.

cockroachcloud/cloud-api.md
rgcase: There's no log expiry at the moment, but if we want to choose a retention policy where they expire we can. At the moment the default will be the beginning of time, or the create time of the organization, whichever is later :).

cockroachcloud/cloud-api.md
rgcase: If the results are returned in descending order (latest to earliest) then we won't return a `next_starting_from` if there are no more results, because it's impossible to ever have more results show up in the past. If we're paginating in the other direction we will return a `next_starting_from` because more events may come in the future.

cockroachcloud/cloud-api.md
rgcase: This is the same time format as every other timestamp in the CC API, formatted as RFC3339 format https://www.ietf.org/rfc/rfc3339.txt. There's a detailed explanation here https://github.com/protocolbuffers/protobuf/blob/7d1362c7ba96ef13de713869513ca32d79a02931/src/google/protobuf/timestamp.proto#L108.

abhinavg6: Got it, thanks Ryan. Let's discuss retention later then.

@mdlinville - In that case, I think we should document that "beginning of time" refers to the creation of a cloud organization.

abhinavg6: Got it, thanks. I think we should explain this because I think this is the first time an API would expose date/time as an input parameter.

cockroachcloud/cloud-api.md
abhinavg6: Got it, this is helpful. Yeah would be good to change this a bit to indicate that we may return more results than the limit if the timestamp for some of the events is exactly same, or something like that.

cockroachcloud/cloud-audit-logs.md
```diff
@@ -0,0 +1,119 @@
...
+
+{% include feature-phases/preview-opt-in.md %}
+
+After your organization is enrolled in the preview, you can begin exporting {{ site.data.products.db }} audit logs.
+
+This page provides some examples of exporting {{ site.data.products.db }} audit logs. For details about each parameter and its defaults, refer to the API specification for the [`auditlogevents` endpoint](cloud-api.html#cloud-audit-logs).
+
+## Export audit logs in ascending order
+
+This example requests audit logs without defining the starting timestamp, sort order, or limit. By default, the first 200 audit logs for your {{ site.data.products.db }} organization are returned in ascending order.
```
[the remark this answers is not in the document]
abhinavg6: As per Ryan's comment above, I think we should indicate that "first 200 logs in ascending order" refers to time of creation of the cloud organization.

cockroachcloud/cloud-audit-logs.md
```diff
@@ -0,0 +1,119 @@
...
+This example requests the 300 most recent audit logs, starting from the current timestamp.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?SortOrder=descending,limit=300' \
+  --header 'Authorization: Bearer {secret_key}'
+~~~
+
+To request the next batch of entries in the same direction, send a second request with the same values for `SortOrder` and `limit` and set `StartingFrom` to the value of `next_starting_from`.
```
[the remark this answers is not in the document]
abhinavg6: Based on the above explanation from Ryan, we should indicate that one wouldn't get next_starting_from if they reach the time of creation of the organization.

abhinavg6 (changes requested):

cockroachcloud/cloud-audit-logs.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Export CockroachDB Cloud audit logs
+summary: Learn about exporting Cockroach Cloud audit logs.
+toc: true
+docs_area: manage
+---
+
+{{ site.data.products.db }} captures audit logs when many types of events occur, such as when a cluster is created or when a user logs in to {{ site.data.products.db }}. Any user who is an admin of the organization can export these audit logs using the [`auditlogevents` endpoint](cloud-api.html#cloud-audit-logs) of the [Cloud API](/docs/cockroachcloud/cloud-api.html).
```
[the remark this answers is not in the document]
mdlinville: FYI, `{{ site.data.products.db }}` expands to exactly "CockroachDB Cloud".

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
+---
+title: Export CockroachDB Cloud organization audit logs
+summary: Learn about exporting Cockroach Cloud organization audit logs.
+toc: true
+docs_area: manage
+---
+
+{{ site.data.products.db }} captures audit logs when many types of events occur, such as when a cluster is created or when a user is added to or removed from an organization. Any user who is an admin of the organization can export these audit logs using the [`auditlogevents` endpoint](cloud-api.html#cloud-audit-logs) of the [Cloud API](/docs/cockroachcloud/cloud-api.html).
```
abhinavg6, line 8:
```suggestion
{{ site.data.products.db }} captures audit logs when many types of events occur, such as when a cluster is created or when a user is added to or removed from an organization. Any user in an organization with an admin-level service account can export these audit logs using the [`auditlogevents` endpoint](cloud-api.html#cloud-audit-logs) of the [Cloud API](/docs/cockroachcloud/cloud-api.html).
```

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,46 @@ Where:
...
+
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned. If the results are returned in ascending order, `next_starting_from` is always returned.
+
+{% include_cached copy-clipboard.html %}
+~~~ json
+{
+  "entries": [
+    "{entry_array}"
+  ],
+  "NextStartingFrom": "{timestamp}"
```
abhinavg6, line 424: Shouldn't the field name be `next_starting_from`?

abhinavg6 (changes requested):

cockroachcloud/cloud-api.md
mdlinville: I copied the result structure from an example @rgcase provided. Maybe he can verify.

rgcase: Abhinav is right, it's `next_starting_from`. Not sure what example response would have said `NextStartingFrom` but I just verified it to be sure.

rgcase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15349)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @abhinavg6 and @mdlinville)

___
*[`cockroachcloud/cloud-api.md` line 406 at r2](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElV1lZ9tPxF0edKLQA:-NElV1lZ9tPxF0edKLQB:b62bw8e) ([raw file](https://github.com/cockroachdb/docs/blob/1ba74e664ddd2e995ce1efd9f0cc8203ad78c10d/cockroachcloud/cloud-api.md#L406)):*
> ```Markdown
>   --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom={timestamp},SortOrder={sort_order},limit={limit}' \
>   --header 'Authorization: Bearer {secret_key}'
> ~~~
> ```

Query parameters should be in lowerCamelCase form and separated by ampersands, so this should be:
```
'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp}&sortOrder={sort_order}&limit={limit}'
```




_Code quote:_
```Markdown
StartingFrom={timestamp},SortOrder={sort_order},limit={limit
```
___
*[`cockroachcloud/cloud-api.md` line 411 at r2](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElVSil7o1E0L8W0fkE:-NElVSil7o1E0L8W0fkF:b-fu5m2c) ([raw file](https://github.com/cockroachdb/docs/blob/1ba74e664ddd2e995ce1efd9f0cc8203ad78c10d/cockroachcloud/cloud-api.md#L411)):*
> ```Markdown
> 
>   - `{timestamp}` is an [RFC3339 timestamp](https://www.ietf.org/rfc/rfc3339.txt) that indicates the first log entry to fetch. If unspecified, defaults to the time when the Cloud organization was created if `{sort_order}` is `ascending`, or the current time if `{sort_order}` is `descending`.
>   - `{sort_order}` is either `ascending` (the default) or `descending`.
> ```

Not sure how specific you want to be here, but in our API we use the strings `"ASC"` and `"DESC"` when they're passed to/from the API.



_Code quote:_
```Markdown
is either `ascending` (the default) or `descending`.
```
___
*[`cockroachcloud/cloud-api.md` line 416 at r2](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElVkHy2C45n2fmuIhZ:-NElVkHz8El0PPOaE2Ij:bskteuk) ([raw file](https://github.com/cockroachdb/docs/blob/1ba74e664ddd2e995ce1efd9f0cc8203ad78c10d/cockroachcloud/cloud-api.md#L416)):*
> ```Markdown
> A request that includes no parameters exports roughly 200 entries in ascending order, starting from when your {{ site.data.products.db }} organization was created.
> 
> If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and a `next_starting_from` field. If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned. If the results are returned in ascending order, `next_starting_from` is always returned.
> ```

This is only true if there are no more results in the descending direction. If there are then we'll include a `next_starting_from`.



_Code quote:_
```Markdown
If the results are returned in descending order (latest to earliest), then `next_starting_from` is not returned.
```
___
*[`cockroachcloud/cloud-org-audit-logs.md` line 80 at r3](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElW5nl20eAmL3u-ENf:-NElW5nl20eAmL3u-ENg:b-gfa7a4) ([raw file](https://github.com/cockroachdb/docs/blob/0b75df3d7ef38f77c3efcd2e80b3e8f44a395199/cockroachcloud/cloud-org-audit-logs.md#L80)):*
> ```Markdown
> ~~~ shell
> curl --request GET \
>   --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom=2022-10-09T02:40:35.054818Z' \
> ```

`StartingFrom -> `startingFrom`.



_Code quote:_
```Markdown
StartingFrom
```
___
*[`cockroachcloud/cloud-org-audit-logs.md` line 91 at r3](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElWC3ICSzCTKb-5g6j:-NElWC3ICSzCTKb-5g6k:by4x2t8) ([raw file](https://github.com/cockroachdb/docs/blob/0b75df3d7ef38f77c3efcd2e80b3e8f44a395199/cockroachcloud/cloud-org-audit-logs.md#L91)):*
> ```Markdown
> ~~~ shell
> curl --request GET \
>   --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?SortOrder=descending,limit=300' \
> ```

`sortOrder=DESC&limit=300`



_Code quote:_
```Markdown
SortOrder=descending,limit=300
```
___
*[`cockroachcloud/cloud-org-audit-logs.md` line 95 at r3](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElWInJ776sQIB-RRez:-NElWInJ776sQIB-RRf-:b-tlgui4) ([raw file](https://github.com/cockroachdb/docs/blob/0b75df3d7ef38f77c3efcd2e80b3e8f44a395199/cockroachcloud/cloud-org-audit-logs.md#L95)):*
> ```Markdown
> ~~~
> 
> To request the next batch of entries in the same direction, send a second request with the same values for `SortOrder` and `limit` and set `StartingFrom` to the value of `next_starting_from`. When there are no more results to fetch (because you have reached when your {{ site.data.products.db }} organization was created), no `next_starting_from` field is returned.
> ```

`sortOrder` and `startingFrom`



_Code quote:_
```Markdown
SortOrder` and `limit` and set `StartingFrom`
```
___
*[`cockroachcloud/cloud-org-audit-logs.md` line 105 at r3](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElWRsL6otBnzVRBpeW:-NElWRsL6otBnzVRBpeX:br3i711) ([raw file](https://github.com/cockroachdb/docs/blob/0b75df3d7ef38f77c3efcd2e80b3e8f44a395199/cockroachcloud/cloud-org-audit-logs.md#L105)):*
> ```Markdown
> ~~~ shell
> curl --request GET \
> ```

Here as well, `sortOrder={ASC, DESC}` and `startingFrom`.



_Code quote:_
```Markdown
curl --request GET \
  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom=2022-10-09T02:40:00.262143Z&SortOrder=ascending' \
  --header 'Authorization: Bearer {secret_key}'
~~~

Next, retrieve roughly 200 less recent entries for the specified timestamp and earlier.

~~~ shell
curl --request GET \
  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?StartingFrom=2022-10-09T02:40:00.262143Z&SortOrder=descending' \
  --header 'Authorization: Bearer {secret_key}'
~~~
```

abhinavg6 (approved):

rgcase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15349)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @abhinavg6 and @mdlinville)

___
*[`cockroachcloud/cloud-api.md` line 406 at r2](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElV1lZ9tPxF0edKLQA:-NElvEmW5qZH8qh9qqIM:b-escup7) ([raw file](https://github.com/cockroachdb/docs/blob/1ba74e664ddd2e995ce1efd9f0cc8203ad78c10d/cockroachcloud/cloud-api.md#L406)):*
<details><summary><i>Previously, rgcase (Ryan Case) wrote…</i></summary><blockquote>

Query parameters should be in lowerCamelCase form and separated by ampersands, so this should be:
```
'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp}&sortOrder={sort_order}&limit={limit}'
```

</blockquote></details>

Still seeing `,` being used as a separator here instead of `&`.
___
*[`cockroachcloud/cloud-org-audit-logs.md` line 91 at r3](https://reviewable.io/reviews/cockroachdb/docs/15349#-NElWC3ICSzCTKb-5g6j:-NElv0z7-7TY0P07azuO:b-ag8v4k) ([raw file](https://github.com/cockroachdb/docs/blob/0b75df3d7ef38f77c3efcd2e80b3e8f44a395199/cockroachcloud/cloud-org-audit-logs.md#L91)):*
<details><summary><i>Previously, rgcase (Ryan Case) wrote…</i></summary><blockquote>

`sortOrder=DESC&limit=300`
</blockquote></details>

Same thing here, `,` -> `&`.

rgcase (approved):
:shipit:

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,50 @@ Where:
...
+
+{% include feature-phases/preview-opt-in.md %}
+
+To export audit logs for activities and events related to your Cloud organization, send a `GET` request to the `/v1/auditlogevents` endpoint. The service account associated with the secret key must have `ADMIN` [permission](console-access-management.html#service-accounts).
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom={timestamp}&sortOrder={sort_order}&limit={limit}' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 405:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,50 @@ Where:
...
+~~~
+
+Where:
+
+  - `{timestamp}` is an [RFC3339 timestamp](https://www.ietf.org/rfc/rfc3339.txt) that indicates the first log entry to fetch. If unspecified, defaults to the time when the Cloud organization was created if `{sort_order}` is `ASC`, or the current time if `{sort_order}` is `DESC`.
+  - `{sort_order}` is either `ASC` (the default) or `DESC`.
+  - `{limit}` indicates roughly how many entries to return. If any entries would be returned for a timestamp, all entries for that timestamp are always returned. Defaults to `200`.
```
ianjevans, line 412:
```suggestion
  - `{limit}` indicates roughly how many entries to return. If any entries would be returned for a timestamp, all entries for that timestamp are always returned. Defaults to `200`.
  - `{api_version}` is the [Cloud API version](#set-the-api-version).
```

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Export audit logs in ascending order
+
+This example requests audit logs without defining the starting timestamp, sort order, or limit. By default, the earliest 200 audit logs for your {{ site.data.products.db }} organization are returned in ascending order, starting from when the organization was created.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 24:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
...
+If you get an error, verify that the feature is enabled for your {{ site.data.products.db }} organization.
+{{site.data.alerts.end}}
+
+To export the next batch of entries, send a second request and set `startingFrom` to the value of `next_starting_from`, `2022-10-09T02:40:35.054818Z`.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom=2022-10-09T02:40:35.054818Z' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 81:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Export audit logs in descending order
+
+This example requests the 300 most recent audit logs, starting from the current timestamp.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?sortOrder=DESC&limit=300' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 92:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
...
+
+This example shows how to retrieve the 200 events on each side of a given timestamp by invoking the API twice, with the same timestamp and a different sort order for each request. The sort order determines whether the specified timestamp is at the beginning or end of the list. These examples use the default value for `limit`.
+
+First, retrieve roughly 200 entries for the specified timestamp and later.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom=2022-10-09T02:40:00.262143Z&sortOrder=ASC' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 107:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

cockroachcloud/cloud-org-audit-logs.md
```diff
@@ -0,0 +1,123 @@
...
+~~~
+
+Next, retrieve roughly 200 less recent entries for the specified timestamp and earlier.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+curl --request GET \
+  --url 'https://cockroachlabs.cloud/api/v1/auditlogevents?startingFrom=2022-10-09T02:40:00.262143Z&sortOrder=DESC' \
+  --header 'Authorization: Bearer {secret_key}'
```
ianjevans, line 116:
```suggestion
  --header 'Authorization: Bearer {secret_key}'
  --header 'Cc-Version: {api_version}'
```

ianjevans (commented):
Looks good, but the examples need to set the `Cc-Version` header as version that is used when you omit the header is the old version.

cockroachcloud/cloud-api.md
```diff
@@ -390,6 +390,50 @@ Where:
...
+
+A request that includes no parameters exports roughly 200 entries in ascending order, starting from when your {{ site.data.products.db }} organization was created.
+
+If the request was successful, the client will receive a JSON array consisting of a list of log `entries` and, depending on the circumstances, a `next_starting_from` field.
+
+- If `{sort_order}` is `ASC`, next_starting_from` is always returned.
```
mdlinville, line 418:
```suggestion
- If `{sort_order}` is `ASC`, `next_starting_from` is always returned.
```

PR #15363, opened 2022-10-13
Serverless diagram replacement

gemma-shay:


andf-crl (approved):
This LGTM with one question for Product surrounding k8s storage.

- k8s docs refer to storage as "persistent volume claims", and don't use language similar to "storage pods" anywhere that I could see. This morning's TES talk on our k8s infr by Towfiqa also used "persistent volume claim". I'm not certain if that language is appropriate here, but let's ask Product how we should refer to the logical storage that our k8s infra uses. I would like the TES slides from this morning, but don't appear to have access! You'd be looking for "CRDB Orchestration on K8s" on 10-13-22 in Highchart. There's an infra chart there that looks really similar to yours, but with some slightly different names.

- Also, I see that you've ordered tenant numbers across AZs. I agree with this call, since we don't use the fact that they're otherwise out of order for anything (i.e. nothing explicit about tenant number in the copy [here](https://deploy-preview-15363--cockroachdb-docs.netlify.app/docs/cockroachcloud/architecture.html#architecture)), so this is OK by me!

JeffSwenson:
> * k8s docs refer to storage as "persistent volume claims", and don't use language similar to "storage pods" anywhere that I could see. This morning's TES talk on our k8s infr by Towfiqa also used "persistent volume claim". I'm not certain if that language is appropriate here, but let's ask Product how we should refer to the logical storage that our k8s infra uses. I would link the TES slides from this morning, but don't appear to have access! You'd be looking for "CRDB Orchestration on K8s" on 10-13-22 in Highchart. There's an infra chart there that looks really similar to yours, but with some slightly different names.

In this case the "storage pod" is a component in CRDB. It is frequently referred to as a "kv pod" but that is a legacy term. The terminology comes from: https://cockroachlabs.atlassian.net/wiki/spaces/CRDB/pages/2431942778/Multi-tenant+CockroachDB#Architectural-terms. 

In the diagram the persistent volume claim is labeled as "block storage" which I think is reasonable.

JeffSwenson (approved):
LGTM

PR #15366, opened 2022-10-13
Bring Vale into the future

nickvigilante:
- Move Vale directory out from Netlify into its own directory
- Add a new Make target to run Vale on changed .md files prior to `git commit`
- Use new vocab structure in docs here: https://vale.sh/docs/topics/vocab/
- Update Vale action to use Reviewdog
 - Vale no longer fails when it finds an error in the text.


v22.2/add-column.md
```diff
@@ -11,6 +11,12 @@ docs_area: reference.sql
 
 {% include {{ page.version.version }}/sql/combine-alter-table-commands.md %}
 
+This is a new sentence.
+
+Bababa testtest
+test test test test test test this should fail repetition asdflkmakl;m4w3;lraksfmdl;xcvkn;4uitnu3warksfljdfndlasjkfnlakjwrnalfkjsndkljn
```
mdlinville, line 17: This seems non-ideal to have a separate failure for each repetition.

vale/Vocab/CockroachDB/accept.txt
```diff
@@ -23,17 +17,10 @@ autovacuum
 avro
 AZ
 AZs
```
mdlinville, line 19: Could this and the preceding be combined into a regex?

mdlinville (commented):
Does it make sense to add some reject rules?

v22.2/add-column.md
nickvigilante: I can possibly tweak the RegEx to catch one or more repetitions.

nickvigilante: Found a better RegEx.

vale/Vocab/CockroachDB/accept.txt
nickvigilante: I was going to work on that in more detail in a separate PR, but I've fixed that line specifically for this PR.

nickvigilante:
> Does it make sense to add some reject rules?

@mdlinville We can, but 1) I don't know any reject rules we need right now. Can you think of any? Also, would this block successful merging of this PR?

mdlinville:
> > Does it make sense to add some reject rules?
> 
> @mdlinville We can, but 1) I don't know any reject rules we need right now. Can you think of any? Also, would this block successful merging of this PR?

No, I can't think of any that aren't already covered by another test. It makes sense to use an iterative approach.

mdlinville (approved):

PR #15393, opened 2022-10-18
Added docs for the granular CHANGEFEED privilege model

kathancox:
Fixes [DOC-4731](https://cockroachlabs.atlassian.net/browse/DOC-4731), [DOC-5658](https://cockroachlabs.atlassian.net/browse/DOC-5658)

Added docs for the newer changefeed privilege model. This includes both Enterprise and Core changefeeds.

### DIFF
--- a/_includes/v22.2/cdc/privilege-model.md
+++ b/_includes/v22.2/cdc/privilege-model.md
@@ -22,4 +22,6 @@ You can add `CHANGEFEED` to the user or role's [default privileges](security-ref
 ALTER DEFAULT PRIVILEGES GRANT CHANGEFEED ON TABLES TO user;
 ~~~
 
+{{site.data.alerts.callout_info}}
 Users with the `CONTROLCHANGEFEED` role option must have `SELECT` on each table, even if they are also granted the `CHANGEFEED` privilege.
+{{site.data.alerts.end}}
--- a/_includes/v22.2/cdc/privilege-model.md
+++ b/_includes/v22.2/cdc/privilege-model.md
@@ -1,0 +1,25 @@
+{{site.data.alerts.callout_info}}
+Starting in v22.2, CockroachDB introduces a new changefeed privilege model that provides finer control over a user's privilege to run changefeeds. 
+
+There is continued support for the [legacy privilege model](#legacy-privilege-model) in v22.2, however it **will be removed** in a future release. We recommend implementing the new privilege model that follows in this section for all changefeeds.
+{{site.data.alerts.end}}
+
+{% include_cached new-in.html version="v22.2" %} You can [grant](grant.html#grant-privileges-on-specific-tables-in-a-database) a user the `CHANGEFEED` privilege to allow them to create changefeeds on a specific table:
+
+{% include_cached copy-clipboard.html %}
+~~~sql
+GRANT CHANGEFEED ON TABLE example_table TO user;
+~~~
+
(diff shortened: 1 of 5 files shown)


HonoreDB (approved):

amruss:
Where is information about the privilege model located? It's not obvious to me that it fits into any of these categories - maybe it should have its own high level page?
Screen Shot 2022-10-24 at 1 52 09 PM

amruss:
Actually I see its in the CREATE CHANGEFEED section - that is probably the right place for it, feel free to disregard my earlier comment

amruss (approved):
This LGTM, my one suggestion is that we have a blurb about external connections and how those can be used along with this privilege model, but I think that can come with the external connections docs

kathancox:
> This LGTM, my one suggestion is that we have a blurb about external connections and how those can be used along with this privilege model, but I think that can come with the external connections docs

@amruss OK cool, I have a PR out for the external connection docs right now. It feels like it would be maybe be best placed in the changefeed docs... that is the combination of the changefeed privilege model and external connection privileges. However, do you think it's best to wait on highlighting this until 1. We have more sink support for changefeeds and external connections. and 2. External connections is also still in preview?

_includes/v22.2/cdc/privilege-model.md
```diff
@@ -0,0 +1,25 @@
...
+Since you can grant the `CHANGEFEED` privilege to a user or role **without** them needing the `SELECT` privilege on a table, these users will be able to create changefeeds, but they will not be able to run a `SELECT` query on that data directly. {% if page.name == "create-changefeed.md" %} However, these users could still read this data indirectly if they have read access to the [sink](changefeed-sinks.html), or create a "sinkless" changefeed that emits messages to the SQL session.{% else %} However, these users will be able to read this data indirectly as `EXPERIMENTAL CHANGEFEED FOR` emits changefeed messages to the SQL session. {% endif %}
+
+You can add `CHANGEFEED` to the user or role's [default privileges](security-reference/authorization.html#default-privileges) with [`ALTER DEFAULT PRIVILEGES`](alter-default-privileges.html#grant-default-privileges-to-a-specific-role):
+
+{% include_cached copy-clipboard.html %}
+~~~sql
+ALTER DEFAULT PRIVILEGES GRANT CHANGEFEED ON TABLES TO user;
+~~~
+
+Users with the `CONTROLCHANGEFEED` role option must have `SELECT` on each table, even if they are also granted the `CHANGEFEED` privilege.
```
andf-crl, line 25: nit: this sentence is possibly unneeded, given that we mentioned this already on line 14. It also seems a bit abrupt following a demonstration of new syntax to discuss a limitation of old.

andf-crl (approved):
LGTM with one tiny nit. Thanks!

_includes/v22.2/cdc/privilege-model.md
kathancox: Yes and no. I think line 14 is the opposite really, and if someone approaches privileges from the role option pov they will kind of want to know that if they then try and grant the new priv that's not going allow the user to run a cf without having first granted `SELECT` too. Not my favorite solution, but I have added this into a note box in the same position — moving it to the legacy section was not preferred especially as we're referencing the new privilege.

But it's a good point, that did feel a pretty abrupt way to end that section. Hopefully, this note box will bridge the worlds between legacy and shiny new.

PR #15407, opened 2022-10-19
Add audit logs Private Preview spec, query param version

ianjevans:
Fixes DOC-5986.

This PR adds the audit log Private Preview spec to the version selector. It also adds the ability to set the version as a query parameter by setting `version=<version>`. The version is the date for GA specs, and the subdirectory of `https://cockroachlabs.cloud/assets/docs/private-preview/` for Private Preview specs.

E.g. for audit logs, the URL is `/docs/api/cloud/v1.html?version=audit-logs`.

### DIFF
--- a/_data/cloud_api_releases.csv
+++ b/_data/cloud_api_releases.csv
@@ -1,3 +1,4 @@
-version,release_date,withdrawn,latest
-2022_09_20,2022-09-20,false,true
-2022_03_31,2022-03-31,false,false
+spec,version,release_date,withdrawn,private-preview,latest
+ccapi_spec_2022_09_20,2022-09-20,2022-09-20,false,false,true
+private-preview/audit_logs/openapi,audit-logs,2022-10-18,false,true,false
+ccapi_spec_2022_03_31,2022-03-31,2022-03-31,false,false,false
--- a/api/cloud/v1.md
+++ b/api/cloud/v1.md
@@ -74,21 +74,24 @@ docs_area: reference.api
   function changeVersion() {
     const doc = document.getElementById('rd');
     selectedVersion = document.getElementById("versionSelector").value;
-    doc.loadSpec('https://cockroachlabs.cloud/assets/docs/' + selectedVersion + '.json');
+    let specPath = "";
+    if (selectedVersion === "latest") {
+      specPath = "ccapi_spec_latest";
+    } else {
+      specPath = selectedVersion;
+    }
+    doc.loadSpec('https://cockroachlabs.cloud/assets/docs/' + specPath + '.json');
   }
 
   window.addEventListener('load', (event) => {
-    const params = new Proxy(new URLSearchParams(window.location.search), {
-      get: (searchParams, prop) => searchParams.get(prop),
-    });
-    // Get the value of the version query parameter
-    let version = params.version;
-    if (version != null) {
+    const urlParams = new URLSearchParams(window.location.search);
+    const version = urlParams.get('version');
+    if (version !== null) {
       let select = document.getElementById("versionSelector");
       // find the version by the option text (not the label)
       for (var i = 0; i < select.length; i++){
-        var option = select.options[i];
-        if (version == option.text) {
+        const option = select.options[i];
+        if (version === option.text) {
           // refresh to the new spec
           select.value = option.value;
           select.dispatchEvent(new Event('change'));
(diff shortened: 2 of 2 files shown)


api/cloud/v1.md
```diff
@@ -74,9 +74,29 @@ docs_area: reference.api
   function changeVersion() {
     const doc = document.getElementById('rd');
     selectedVersion = document.getElementById("versionSelector").value;
-    doc.loadSpec('https://cockroachlabs.cloud/assets/docs/ccapi_spec_' + selectedVersion + '.json');
+    doc.loadSpec('https://cockroachlabs.cloud/assets/docs/' + selectedVersion + '.json');
   }
 
+  window.addEventListener('load', (event) => {
+    const params = new Proxy(new URLSearchParams(window.location.search), {
```
fantapop, line 81: I had never seen this usage of Proxy before. Its kind of cool.  In this case however, I'd recommend just using the 3rd  option seen here: https://fedingo.com/how-to-get-query-string-values-in-javascript/. Since we're only getting a single parameter here we don't save much by changing the api.  It could just be:

```js
const urlParams = new URLSearchParams(window.location.search);
const version = urlParams.get('version');
```

api/cloud/v1.md
```diff
@@ -74,9 +74,29 @@ docs_area: reference.api
...
+      get: (searchParams, prop) => searchParams.get(prop),
+    });
+    // Get the value of the version query parameter
+    let version = params.version;
+    if (version != null) {
+      let select = document.getElementById("versionSelector");
+      // find the version by the option text (not the label)
+      for (var i = 0; i < select.length; i++){
+        var option = select.options[i];
```
fantapop, line 90: In modern javascript, there's pretty much never a reason to use `var` any more.  These can be:

```js
  for (let i = 0; i < select.length; i++){
        const option = select.options[i];
```

api/cloud/v1.md
```diff
@@ -74,9 +74,29 @@ docs_area: reference.api
...
+        if (version == option.text) {
```
fantapop, line 91: the 2 character comparison operators are generally not used anymore either.  Recommend replacing with === and !== (above).  These 3 character operators also check for type equality.  Do double check it still works as expected because the != null is probably a bit different than !== null.

mdlinville:
I looked at the [preview](https://deploy-preview-15407--cockroachdb-docs.netlify.app/docs/api/cloud/v1.html) and set `version` to `audit-logs` and I may be wrong, but I think the parameters like `starting_from` and `sort_order` should actually be `startingFrom` and `sortOrder`. See https://github.com/cockroachdb/docs/pull/15349.

mdlinville (changes requested):
Questions about the `auditevents` API.

rgcase:
> I looked at the [preview](https://deploy-preview-15407--cockroachdb-docs.netlify.app/docs/api/cloud/v1.html) and set `version` to `audit-logs` and I may be wrong, but I think the parameters like `starting_from` and `sort_order` should actually be `startingFrom` and `sortOrder`. See #15349.

I wonder if this is actually something I did wrong? Seems like other APIs seem to use snake case for query params, but I don't think I did anything special to make it camel case?

_data/cloud_api_releases.csv
```diff
@@ -1,3 +1,4 @@
-version,release_date,withdrawn,latest
-2022_09_20,2022-09-20,false,true
-2022_03_31,2022-03-31,false,false
\ No newline at end of file
+spec,version,release_date,withdrawn,private-preview,latest
+ccapi_spec_2022_09_20,2022-09-20,2022-09-20,false,false,true
+private-preview/audit_logs/openapi,audit-logs,2022-10-18,false,true,false
```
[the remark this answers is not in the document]
ianjevans: Chris and I discussed this in Slack. Best course long-term is to keep the case of the paths and version in sync. Proposal is to change the asset paths to:

```
/assets/docs/api/<version>/openapi.json
/assets/docs/api/latest/openapi.json
/assets/docs/api/private-preview/<version>/openapi.json
```

Where `<version>` matches the GA version (e.g. `2022-09-20`) or the private preview name (e.g. `audit-logs`).

rgcase:
Wow, actually it seems like you can use either camelcase or snakecase and they'll both be accepted. I'm sorry @mdlinville for directing you down the wrong path. I just tried it with camelcase myself and it worked, so I assumed that was the correct way. It's pretty surprising that it works for both.

mdlinville (approved):

_data/cloud_api_releases.csv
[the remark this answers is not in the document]
fantapop: I'll update my private preview pipeline PR to reflect these

fantapop (approved):

PR #15409, opened 2022-10-19
Steer users away from sql.defaults.* settings

rmloveland:
Fixes DOC-4754

Summary of changes:

- Add a `sql.defaults.*` notice as an include

- Add that include to all the places in the docs where `sql.defaults.*`
  is mentioned

- NB. We did *not* change all uses of `SET CLUSTER SETTING
  sql.defaults.*` to use `ALTER ROLE ALL` for the following reasons:

  - Making a docs change of that scope and ensuring we don't break the
    pages in question would require a large amount of manual QA

  - The use of `ALTER ROLE` is merely encouraged (the old settings are
    not yet officially _deprecated_), and the SQL client prints a
    message to the user. So that amount of work is not strictly required
    yet.

### DIFF
--- a/_includes/v22.2/sql/sql-defaults-cluster-settings-deprecation-notice.md
+++ b/_includes/v22.2/sql/sql-defaults-cluster-settings-deprecation-notice.md
@@ -1,0 +1,3 @@
+{{site.data.alerts.callout_info}}
+<span class="version-tag">New in v22.2:</span> Use [`ALTER ROLE ALL SET {sessionvar} = {val}`](alter-role.html#set-default-session-variable-values-for-all-users) instead of the `sql.defaults.*` [cluster settings](cluster-settings.html). This allows you to set a default value for all users for any [session variable](set-vars.html) that applies during login, making the `sql.defaults.*` cluster settings redundant.
+{{site.data.alerts.end}}
--- a/v22.2/alter-role.md
+++ b/v22.2/alter-role.md
@@ -234,6 +234,21 @@ root@:26257/movr> SHOW timezone;
 (1 row)
 ~~~
 
+### Set default session variable values for all users
+
+To set a default value for all users for any [session variable](set-vars.html) that applies during login, issue a statment like the following:
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+ALTER ROLE ALL SET sql.spatial.experimental_box2d_comparison_operators.enabled = "on";
+~~~
+
+~~~
+ALTER ROLE
+~~~
+
+{% include {{page.version.version}}/sql/sql-defaults-cluster-settings-deprecation-notice.md %}
+
 ## See also
 
 - [`DROP ROLE`](drop-role.html)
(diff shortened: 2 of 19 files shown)


e-mbrown (approved):
LGTM

rafiss (commented):
this might be a nitpick, but i remember we (Vy/Ebony/Adam/Me) had some convos in the release where we decided that this would be more of a message like "nudge to encourage ALTER ROLE" rather than a "deprecating the sql.defaults cluster settings"

the only practical difference is that we don't actually have plans to ever remove the existing sql.defaults cluster settings. we just think the ALTER ROLE syntax is more useful, so we were wanting to point people towards that

in the context of this PR, maybe the only difference would be to use a word other than "deprecate" in the `includes` notice

rmloveland:
> this might be a nitpick, but i remember we (Vy/Ebony/Adam/Me) had some convos in the release where we decided that this would be more of a message like "nudge to encourage ALTER ROLE" rather than a "deprecating the sql.defaults cluster settings"

ah ok - appreciate the context

> the only practical difference is that we don't actually have plans to ever remove the existing sql.defaults cluster settings. we just think the ALTER ROLE syntax is more useful, so we were wanting to point people towards that

aha!  gotcha

> in the context of this PR, maybe the only difference would be to use a word other than "deprecate" in the `includes` notice

yes can def do that - will update

meta comment: "deprecation" does not have to mean "we may/will eventually remove this", there is also a sense in which it means only that it is discouraged in favor of something else, but will not necessarily be removed.  According to [the wikipedias](https://en.wikipedia.org/wiki/Deprecation):

> In several fields, especially [computing](https://en.wikipedia.org/wiki/Computing), deprecation is the discouragement of use of some terminology, feature, design, or practice, typically because it has been superseded or is no longer considered efficient or safe, without completely removing it or prohibiting its use. Typically, deprecated materials are not completely removed to ensure legacy compatibility or back up practice in case new methods are not functional in an odd scenario.
> It can also imply that a feature, design, or practice will be removed or discontinued entirely in the future.[[1]](https://en.wikipedia.org/wiki/Deprecation#cite_note-1)

This matches my understanding from other contexts/roles that deprecation does not necessarily mean "being removed"

However, the common usage at CRL seems to have evolved towards "deprecation means we may/will eventually change/remove this"

is it your sense that deprecated means "we will change/remove this" among the Eng org ? that's kinda what I've seen here anecdotally

rmloveland:
@rafiss I updated the PR to remove the phrase "which are deprecated", and also updated the PR description and commit message to make clear this is not a deprecation but merely a nudge or whatever. Otherwise it's exactly the same

what do you think?

ianjevans (approved):
One of the Vale linter messages is valid, but otherwise lgtm.

rmloveland:
Ian Evans ***@***.***> writes:

> @ianjevans approved this pull request.
>
> One of the Vale linter messages is valid, but otherwise lgtm.

Thanks Ian!

I looked at the Vale output in the diff [1] but it doesn't look like
those messages are referring to anything that was added in this change?

The one I saw that seems valid ("older/newer") is on the optimizer page
- is that the one you meant? I agree it is valid but don't wanna do in
this PR, would rather leave to a style PR.

Sorry if I'm totally misunderstanding you tho - entirely possible!

[1] https://github.com/cockroachdb/docs/pull/15409/files

PR #15431, opened 2022-10-20
DOC-5237 add 4 mvcc range metrics

andf-crl:
Addresses: DOC-5237

- Added 4 new metrics to `metric-names.md`: `rangekeybytes`, `rangekeycount`, `rangevalbytes`, and `rangevalcount`.

Question:

- Assuming these are not for Serverless also, or I'd need to also update `metric-names-serverless.md`.

Staging:

[metric-names.md](https://deploy-preview-15431--cockroachdb-docs.netlify.app/docs/v22.2/ui-custom-chart-debug-page#available-metrics)

### DIFF
--- a/_includes/v22.2/metric-names.md
+++ b/_includes/v22.2/metric-names.md
@@ -139,9 +139,13 @@ Name | Help
 `range.snapshots.rcvd-bytes` | Number of snapshot bytes received
 `range.snapshots.sent-bytes` | Number of snapshot bytes sent
 `range.splits` | Number of range splits
+`rangekeybytes` | Number of bytes taken up by range keys (e.g. MVCC range tombstones)
+`rangekeycount` | Count of all range keys (e.g. MVCC range tombstones)
 `ranges.unavailable` | Number of ranges with fewer live replicas than needed for quorum
 `ranges.underreplicated` | Number of ranges with fewer live replicas than the replication target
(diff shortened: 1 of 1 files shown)


erikgrinaker:
>Assuming these are not for Serverless also, or I'd need to also update metric-names-serverless.md.

These should be exposed in serverless too. They are analogous to the point key metrics, e.g. `keybytes`, which are already exposed in serverless.

andf-crl:
@erikgrinaker Sorry I misspoke: To be avail in Serverless, these metrics would need to also be present in this file, IIRC: https://github.com/cockroachlabs/managed-service/blob/master/pkg/otel/assets/metrics.yaml. I believe this would require a ticket to the KV Obs team. The `metric-names-serverless.md` file I referenced is populated directly from this `yaml` file.

erikgrinaker:
I see, thanks @andf-crl. I wrote up https://cockroachlabs.atlassian.net/browse/CC-8391 for this, hope I put it in the right place.

Sorry for the late reply btw, been sick for a while.

erikgrinaker (approved):

andf-crl:
No worries, and TYFR! Hope you're feeling better!

MichaelTrestman (approved):
LGTM

andf-crl:
TYFR both!

PR #15442, opened 2022-10-21
DOC-5726: Force absolute URLs in sidebar

nickvigilante:
This will help remove some of the burden Google has to navigate multiple URLs while crawling our site. The hope is that this will increase our crawl budget. This should not have any end-user impact.

Additionally, this now allows search to search the local instance running Jekyll, whether that be the Netlify build or the local `make standard` target.

### DIFF
--- a/_config_base.yml
+++ b/_config_base.yml
@@ -138,4 +138,3 @@ sass:
   sass_dir: css
   style: compressed
 site_title: CockroachDB Docs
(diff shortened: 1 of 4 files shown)


lnhsingh:
Very cool!

ianjevans (approved):

lnhsingh (approved):

PR #15447, opened 2022-10-21
Docs for restoring backups across versions

kathancox:
Fixes [DOC-1221](https://cockroachlabs.atlassian.net/browse/DOC-1221)

Initial draft policy for restoring backups across major versions.

### Preview

https://deploy-preview-15447--cockroachdb-docs.netlify.app/docs/dev/restoring-backups-across-versions.html

### DIFF
--- a/v21.2/restoring-backups-across-versions.md
+++ b/v21.2/restoring-backups-across-versions.md
@@ -10,6 +10,10 @@ This page describes the support Cockroach Labs provides for restoring backups ac
 - [Support for restoring backups into a higher version](#support-for-restoring-backups-into-a-higher-version)
 - [Support for long-term backup archival](#support-for-long-term-backup-archival)
 
+{{site.data.alerts.callout_info}}
+We refer to versions on this page in the "v21.2.x" format, because the minor versions of CockroachDB are not considered in backup eligibility. For example, CockroachDB considers both v21.2.8 and v21.2.14 as v21.2 clusters. 
+{{site.data.alerts.end}}
+
 ## Support for restoring backups into a higher version
 
 Cockroach Labs supports restoring a backup taken on a cluster on a specific major version into a cluster that is on the same version or the next major version. Therefore, when you're upgrading your cluster from version N to N+1, if you took a backup on major version N, you can restore it to your cluster on either major version N or N+1.
--- a/v21.2/restoring-backups-across-versions.md
+++ b/v21.2/restoring-backups-across-versions.md
@@ -7,48 +7,48 @@ docs_area: manage
 
 This page describes the support Cockroach Labs provides for restoring backups across major versions of CockroachDB. The sections outline the support and examples for the most common use cases when restoring backups across versions:
 
-- [Support for restoring backups into a newer version](#support-for-restoring-backups-into-a-newer-version)
-- [Restore into a newer version](#restore-into-a-newer-version)
+- [Support for restoring backups into a higher version](#support-for-restoring-backups-into-a-higher-version)
+- [Restore into a higher version](#restore-into-a-higher-version)
 - [Support for long-term backup archival](#support-for-long-term-backup-archival)
 
-## Support for restoring backups into a newer version
+## Support for restoring backups into a higher version
 
-Cockroach Labs supports restoring a backup on one major version following the major version on which you took the backup. Therefore, if you took a backup on major version N, you can restore it on major version N+1.
+Cockroach Labs supports restoring a backup taken on a cluster on a specific major version into a cluster that is on the next major version. Therefore, if you took a backup on major version N, you can restore it on major version N+1.
 
-For example, backups taken on v21.1 can be restored into v21.2. Backups taken on v21.2 can be restored into v22.1. The following table outlines this support:
+For example, backups taken on v21.1.x can be restored into v21.2.x. Backups taken on v21.2.x can be restored into v22.1.x. The following table outlines this support:
 
-Backup taken on version | Restorable into version
-------------------------+--------------------------
-20.1                    | 20.2
-20.2                    | 21.1
-21.1                    | 21.2
-21.2                    | 22.1
-22.1                    | 22.2
+Backup taken on version   | Restorable into version
+--------------------------+--------------------------
+20.1.x                    | 20.2.x
+20.2.x                    | 21.1.x
+21.1.x                    | 21.2.x
+21.2.x                    | 22.1.x
+22.1.x                    | 22.2.x
 
-See the following section, [Restore into a newer version](#restore-into-a-newer-version), for an example of of restoring from multiple major versions previous to a newer version.
+See the following section, [Restore into a higher version](#restore-into-a-higher-version), for an example of of restoring from multiple major versions previous to a higher version.
 
 {{site.data.alerts.callout_info}}
-Cockroach Labs does **not** support restoring backups from a newer version into a previous version. 
+Cockroach Labs does **not** support restoring backups from a higher version into a lower version. 
 {{site.data.alerts.end}}
 
-## Restore into a newer version
+## Restore into a higher version
 
 When you upgrade from version N to N+1, you can restore version N backup into version N+1. Then, you can take a new backup on the version you just upgraded to. If you need to restore a backup that was taken with on an earlier version, you would repeat this process until you have fully upgraded to the desired version.
 
-For example, you have taken a backup on v20.2 and you want to upgrade to v22.1. You would upgrade and back up through the following process:
+For example, you have taken a backup on v20.2.x and you want to upgrade to v22.1.x. You would upgrade and back up through the following process:
 
-1. Upgrade from v20.2 to v21.1:
+1. Upgrade from v20.2.x to v21.1.x:
     1. Finalize your upgrade. See [Upgrade to CockroachDB v21.1](../v21.1/upgrade-cockroach-version.html) for the necessary steps to upgrade.
-    1. [Restore your backup](restore.html#examples) from v20.2 into the v21.1 upgraded cluster.
-    1. [Take a new backup](backup.html#examples) on the v21.1 cluster.
-1. Upgrade from v21.1 to v21.2:
+    1. [Restore your backup](restore.html#examples) from v20.2.x into the v21.1.x upgraded cluster.
+    1. [Take a new backup](backup.html#examples) on the v21.1.x cluster.
+1. Upgrade from v21.1.x to v21.2.x:
     1. Finalize your upgrade. See [Upgrade to CockroachDB v21.2](../v21.2/upgrade-cockroach-version.html) for the necessary steps to upgrade.
-    1. Restore your backup from v21.1 into the v21.2 upgraded cluster.
-    1. Take a new backup on the v21.2 cluster.
-1. Upgrade from v21.2 to v22.1:
+    1. Restore your backup from v21.1.x into the v21.2.x upgraded cluster.
+    1. Take a new backup on the v21.2.x cluster.
+1. Upgrade from v21.2.x to v22.1.x:
     1. Finalize your upgrade. See [Upgrade to CockroachDB v22.1](../v22.1/upgrade-cockroach-version.html) for the necessary steps to upgrade.
-    1. Restore your backup from v21.2 into the v22.1 upgraded cluster.
-    1. Take a new backup on the v22.1 cluster.
+    1. Restore your backup from v21.2.x into the v22.1.x upgraded cluster.
+    1. Take a new backup on the v22.1.x cluster.
 
 See the [Upgrade Policy](../cockroachcloud/upgrade-policy.html) page for details on version upgrades to {{ site.data.products.db }} clusters.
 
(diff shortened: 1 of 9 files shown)


v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+This page describes the support Cockroach Labs provides for restoring backups across major versions of CockroachDB. The sections outline the support and examples for the most common use cases when restoring backups across versions:
+
+- [Support for restoring backups into a newer version](#support-for-restoring-backups-into-a-newer-version)
+- [Restore into a newer version](#restore-into-a-newer-version)
+- [Support for long-term backup archival](#support-for-long-term-backup-archival)
+
+## Support for restoring backups into a newer version
+
+Cockroach Labs supports restoring a backup on one major version following the major version on which you took the backup. Therefore, if you took a backup on major version N, you can restore it on major version N+1.
```
livlobo, line 16: This is hard to express, the current text sounds a bit confusing to me. How does this sound?

You can restore a backup taken on a cluster on a specific major version (e.g. version N), into a cluster that is on the next major version (e.g. version N+1).

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+For example, backups taken on v21.1 can be restored into v21.2. Backups taken on v21.2 can be restored into v22.1. The following table outlines this support:
```
livlobo, line 18: we can write these as v21.1.x, v21.2.x, to express that the minor version does not matter.

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+20.1                    | 20.2
+20.2                    | 21.1
+21.1                    | 21.2
+21.2                    | 22.1
+22.1                    | 22.2
+
+See the following section, [Restore into a newer version](#restore-into-a-newer-version), for an example of of restoring from multiple major versions previous to a newer version.
+
+{{site.data.alerts.callout_info}}
+Cockroach Labs does **not** support restoring backups from a newer version into a previous version.
```
livlobo, line 31: from a higher version into a lower version? (newer is relative to date e.g. 22.1.100 may be "newer" than 22.2.0?)

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+1. Upgrade from v21.2 to v22.1:
+    1. Finalize your upgrade. See [Upgrade to CockroachDB v22.1](../v22.1/upgrade-cockroach-version.html) for the necessary steps to upgrade.
+    1. Restore your backup from v21.2 into the v22.1 upgraded cluster.
+    1. Take a new backup on the v22.1 cluster.
+
+See the [Upgrade Policy](../cockroachcloud/upgrade-policy.html) page for details on version upgrades to {{ site.data.products.db }} clusters.
+
+## Support for long-term backup archival
+
+When you need to archive a backup for the long term, we recommend that you also archive the CockroachDB binary of the version that the backup was taken on. However, for this method, it is necessary that the binary can still be run **and** is able to upgrade to a version that is supported.
```
livlobo, line 57: Hmm, Im not sure this second part is necessary?

livlobo (approved):
Looks good overall, Kathryn. Thanks for documenting this! 

Would you mind adding David to review?

v21.2/restoring-backups-across-versions.md
kathancox: I've altered this a little to match up here.

v21.2/restoring-backups-across-versions.md
kathancox: Good point... I have changed the language around this

v21.2/restoring-backups-across-versions.md
kathancox: Yes, that's a good idea. I have changed all version numbers to this pattern as I think it makes it clearer.

v21.2/restoring-backups-across-versions.md
kathancox: I am assuming you mean from "However,..." onward? If so, this was something that was part of the original doc and it seemed important to mention as a consideration?

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+For a true archival copy that is not dependent on CockroachDB at all, running an [`EXPORT INTO CSV`](export.html) and archiving the files would be a better approach instead of taking a backup.
```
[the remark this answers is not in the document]
kathancox: I have added in a ref to changefeed as export for now.

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+This page describes the support Cockroach Labs provides for restoring backups across major versions of CockroachDB. The sections outline the support and examples for the most common use cases when restoring backups across versions:
+
+- [Support for restoring backups into a higher version](#support-for-restoring-backups-into-a-higher-version)
+- [Restore into a higher version](#restore-into-a-higher-version)
+- [Support for long-term backup archival](#support-for-long-term-backup-archival)
+
+## Support for restoring backups into a higher version
+
+Cockroach Labs supports restoring a backup taken on a cluster on a specific major version into a cluster that is on the next major version. Therefore, if you took a backup on major version N, you can restore it on major version N+1.
```
dt, line 16: maybe this is too obvious to need to state, but should we say "on the same version _or_  the next major..."?

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+    1. Restore your backup from v21.2.x into the v22.1.x upgraded cluster.
+    1. Take a new backup on the v22.1.x cluster.
+
+See the [Upgrade Policy](../cockroachcloud/upgrade-policy.html) page for details on version upgrades to {{ site.data.products.db }} clusters.
+
+## Support for long-term backup archival
+
+When you need to archive a backup for the long term, we recommend that you also archive the CockroachDB binary of the version that the backup was taken on. However, for this method, it is necessary that the binary can still be run **and** is able to upgrade to a version that is supported.
+
+For a true archival copy that is not dependent on CockroachDB at all, running an [`EXPORT INTO CSV`](export.html) and archiving the files would be a better approach instead of taking a backup. You can also use changefeeds to export data in this way, see [Export Data with Changefeeds](export-data-with-changefeeds.html) for more detail.
```
dt, line 59: should we suggest EXPORT, given we know it does not work on larger data sizes, or should we suggest CDC and CSV format?

v21.2/restoring-backups-across-versions.md
dt: I'm also not really sure what the However... here is saying. Like, you can use the archived binary and restore regardless of whether or not you then upgrade, so that seems unrelated?

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+See the following section, [Restore into a higher version](#restore-into-a-higher-version), for an example of of restoring from multiple major versions previous to a higher version.
+
+{{site.data.alerts.callout_info}}
+Cockroach Labs does **not** support restoring backups from a higher version into a lower version. 
+{{site.data.alerts.end}}
+
+## Restore into a higher version
+
+When you upgrade from version N to N+1, you can restore version N backup into version N+1. Then, you can take a new backup on the version you just upgraded to. If you need to restore a backup that was taken with on an earlier version, you would repeat this process until you have fully upgraded to the desired version.
```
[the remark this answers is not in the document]
kathancox: @livlobo Do you feel there is need to outline an example restore/upgrade process here? I'm not sure of if we've had users misunderstanding this, and so, an example process is useful... let me know your thoughts and I can alter as best suits.

v21.2/restoring-backups-across-versions.md
livlobo: Hmm, I think we should stick with CDC and CSV and remove EXPORT.

v21.2/restoring-backups-across-versions.md
livlobo: Good catch, worth stating the obvious.

v21.2/restoring-backups-across-versions.md
livlobo: Yes, the "And is unable to upgrade" is not required.

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,64 @@
...
+
+For example, you have taken a backup on v20.2.x and you want to upgrade to v22.1.x. You would upgrade and back up through the following process:
+
+1. Upgrade from v20.2.x to v21.1.x:
+    1. Finalize your upgrade. See [Upgrade to CockroachDB v21.1](../v21.1/upgrade-cockroach-version.html) for the necessary steps to upgrade.
+    1. [Restore your backup](restore.html#examples) from v20.2.x into the v21.1.x upgraded cluster.
+    1. [Take a new backup](backup.html#examples) on the v21.1.x cluster.
```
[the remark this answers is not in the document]
livlobo: I think the suggestion came from TSE that best practice is to backup before an upgrade. Either way, definitely don't need the restore steps.
 
 I'm good with removing the intermediate steps completely and lean on upgrade docs to spell out the upgrade process.

v21.2/restoring-backups-across-versions.md
kathancox: @dt @livlobo That makes sense for v22.1 and v22.2. However, for v21.2 in changefeeds, we don't have `format = csv`, `initial_scan = 'only'`, or `end_time` options... so I have left export in this version. Lmk if you want me to alter the language around export in this version at all, given the limitations you mention.

v21.2/restoring-backups-across-versions.md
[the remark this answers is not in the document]
kathancox: I've totally removed this section and merged into the previous, since this really was a reiteration.

dt (approved):

andf-crl (approved):
LGTM, just a question about how we talk about minor versions. Thanks!

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,42 @@
...
+---
+
+This page describes the support Cockroach Labs provides for restoring backups across major versions of CockroachDB. The sections outline the support for the most common use cases when restoring backups across versions:
+
+- [Support for restoring backups into a higher version](#support-for-restoring-backups-into-a-higher-version)
+- [Support for long-term backup archival](#support-for-long-term-backup-archival)
+
+## Support for restoring backups into a higher version
+
+Cockroach Labs supports restoring a backup taken on a cluster on a specific major version into a cluster that is on the same version or the next major version. Therefore, when you're upgrading your cluster from version N to N+1, if you took a backup on major version N, you can restore it to your cluster on either major version N or N+1.
```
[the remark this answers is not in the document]
dt: if you want to get pedantic, the _software_ actually supports more than is documented here -- some versions might, in fact, allow you to restore a backup that is older than from previous major version -- but you're on your own when doing so and cockroach _labs_ is covering our bases by saying it isn't supported (we never said it was, and now we're explicitly saying it isn't). If you technically just go by what the software lets you do, or arguably "supports", then you can restore a backup from cockroach v1.0 using v22.1, but CR*L* would not support that if you hit any snags.

But I'm being extra, extra pedantic with this nit-of-a-nit, so feel free to ignore me.

v21.2/restoring-backups-across-versions.md
```diff
@@ -0,0 +1,42 @@
...
+20.1.x                    | 20.2.x
+20.2.x                    | 21.1.x
+21.1.x                    | 21.2.x
+21.2.x                    | 22.1.x
+22.1.x                    | 22.2.x
+
+See the [Upgrade documentation](../{{site.versions["stable"]}}/upgrade-cockroach-version.html) for the necessary steps to finalize your upgrade. For {{ site.data.products.db }} clusters, see the [Upgrade Policy](../cockroachcloud/upgrade-policy.html) page.
+
+{{site.data.alerts.callout_info}}
+Cockroach Labs does **not** support restoring backups from a higher version into a lower version.
```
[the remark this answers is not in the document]
dt: no, for the purposes of backups, we're talking cluster version, not binary version, so there's no minor revisions in play -- v 22.1.8 and v22.1.14 are both v22.1 in the eyes of the cluster version.

v21.2/restoring-backups-across-versions.md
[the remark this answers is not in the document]
andf-crl: Got it. Then it is policy. If we have such defined policy, then it's appropriate to use the actor supporting it: CRL. Thanks for the clarification!

v21.2/restoring-backups-across-versions.md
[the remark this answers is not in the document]
andf-crl: Great, thanks! Let's be sure to capture a version of this in this doc: that minor version is not considered in backup eligibility, only major. Thank you!

PR #15460, opened 2022-10-24
DOC-6020 fix broken Grafana replication url

andf-crl:
Addresses: DOC-6020

- Engineering changed the name of an upstream resource from `replicas.json` to `replication.json`. This docs PR to update the docs to use the new name.
	- Upstream rename was introduced for CRDB v21.2, so sending this change back that far.

Staging:
[monitor-cockroachdb-with-prometheus.md](https://deploy-preview-15460--cockroachdb-docs.netlify.app/docs/stable/monitor-cockroachdb-with-prometheus.html#step-5-visualize-metrics-in-grafana)


MichaelTrestman (commented):
LGTM

PR #15476, opened 2022-10-25
Added section to scheduled backups on protected timestamps

kathancox:
Fixes [DOC-4784](https://cockroachlabs.atlassian.net/browse/DOC-4784)

Added a section to the scheduled backup page to describe how backup schedules work with protected timestamps. Also, included some cross-referencing from the incremental backup and garbage collection section.


v22.2/create-schedule-for-backup.md
```diff
@@ -90,10 +90,17 @@ If you encounter a bug, please [file an issue](file-an-issue.html).
...
 - The `AS OF SYSTEM TIME` clause cannot be set on scheduled backups. Scheduled backups are started shortly after the scheduled time has passed by an internal polling mechanism and are automatically run with `AS OF SYSTEM TIME` set to the time at which the backup was scheduled to run.
 - If you want to schedule a backup using temporary credentials, we recommend that you use `implicit` authentication; otherwise, you'll need to drop and then recreate schedules each time you need to update the credentials.
 
+### Protected timestamps and scheduled backups 
+
+{% include_cached new-in.html version="v22.2" %} When using scheduled backups, [protected timestamps](architecture/storage-layer.html#protected-timestamps) will ensure that data is protected from garbage collection even if the [GC TTL](configure-replication-zones.html#gc-ttlseconds) is lower than the frequency of scheduled backups. For example, if a backup fails, protected timestamps will continue to protect all keys that have not yet been backed up.
```
adityamaru, line 98: Scheduled backups ensure that the data that needs to be backed up is protected from garbage collection until it has been successfully backed up to a certain time. This active management of protected timestamps means that scheduled backups can run at a cadence independent of the datas' GC TTL. This is unlike non-scheduled/out of band backups that are tightly coupled to the GC TTL (link to our existing section on this dependency maybe?)

The data being backed up will not be eligible for garbage collection until a successful backup completes. At this point, the schedule will release the old protected timestamp record and write a new one to protect data for the next backup that is scheduled to run. As a result, it is necessary to consider that when a scheduled backup fails there will be an accumulation of data until the next successful backup. Resolving the backup failure or dropping the backup scheduled will make the data eligible for garbage collection once again.

We recommend monitoring for your backup schedule to alert for failed backups. See [Set up monitoring for the backup schedule](manage-a-backup-schedule.html#set-up-monitoring-for-the-backup-schedule) for more detail.

adityamaru: ^ suggested tweaks, feel free to use it in a way you think is appropriate!

adityamaru: I don't know if its also worth linking to `exclude_data_from_backup` from here as a way to prevent a protection from holding up GC on a table even during a scheduled backup.

adityamaru (changes requested):

v22.2/create-schedule-for-backup.md
livlobo: +1 to link/include -->  exclude_data_from_backup here

kathancox: Thank you both, I have incorporated the feedback with very slight tweaks and also added a reference to the `exclude_data_from_backup` option.

_includes/v22.2/backups/scheduled-backups-tip.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_success}}
+ You can create [schedules for periodic backups](manage-a-backup-schedule.html) in CockroachDB. We recommend using scheduled backups to automate daily backups of your cluster.
```
andf-crl, line 2: nit: extra leading space

v22.2/create-schedule-for-backup.md
```diff
@@ -90,10 +90,19 @@ If you encounter a bug, please [file an issue](file-an-issue.html).
...
 - If you want to schedule a backup using temporary credentials, we recommend that you use `implicit` authentication; otherwise, you'll need to drop and then recreate schedules each time you need to update the credentials.
 
+### Protected timestamps and scheduled backups 
+
+{% include_cached new-in.html version="v22.2" %} Scheduled backups ensure that the data to be backed up is protected from garbage collection until it has been successfully backed up to a certain time. This active management of [protected timestamps](architecture/storage-layer.html#protected-timestamps) means that you can run scheduled backups at a cadence independent from the [GC TTL](configure-replication-zones.html#gc-ttlseconds) of the data. This is unlike non-scheduled backups that are tightly coupled to the GC TTL. See [Garbage collection and backups](take-full-and-incremental-backups.html#garbage-collection-and-backups) for more detail.
```
andf-crl, line 98: Question: what does "to a certain time" mean here? The sentence looks perfectly-functional without this qualifier. Initially I assumed "up to the configured `gc.ttlseconds` value", i.e. 25 hours (by default) or 10m (by recommendation)), but: clarifying that this is _not_ the case is in fact the goal of this PR! So I'm not sure what we mean here. Line 100 goes on to state this wonderfully, maybe that's the leading sentence?

andf-crl (approved):
LGTM with small little ideas. Nice work!

v22.2/create-schedule-for-backup.md
kathancox: "to a certain time" here would represent the last backup. That is, the scheduled backup would be backing up everything (depending on whether it is full or incremental) since the last backup and that would be based on the time here. I think, as you allude to, this sentence works just fine without "to a certain time" as scheduled backups almost explicitly apply this chain of backups, which is then referred to in the next sentence.

PR #15498, opened 2022-10-28
DOC-5978: Remove unnecessary/internal redirects from the docs

nickvigilante:
Fixes DOC-5978 (fix internal redirects to reduce the number of 301s)
Fixes DOC-6059 (add the ability to override a page's canonical)


mikeCRL (approved):
Looks good. Spot checked many URLs just to be sure the new ones point to the right place and all were fine. Thanks, Nick.

PR #15520, opened 2022-10-31
DOC-6056: Move sidebar JS out of the main page

nickvigilante:
Fixes DOC-6056

This will help cut down on some of the JS bloat we have in each of our web pages. This reduced each file by about 114 KB, and at around 4500 files to generate, that will save about 0.5 GB of space/post processing.


nickvigilante:
@JosephNickdow Does this JS need to be deferred? I didn't defer it since the original script wasn't deferred, but if you think there would be a benefit to it, I can add it.

JosephNickdow (commented):
@nickvigilante I did some local testing and saw a small increase in performance with that script deferred, would you be able to take a look and confirm?

nickvigilante:
I added in the `defer`.

PR #15529, opened 2022-11-01
DOC-6102: Move fragile redirect to a separate include file

nickvigilante:
Fixes DOC-6102

This PR moves a fragile redirect that allows the /dev and /stable aliases within the docs to function properly so that people don't trip over it accidentally.


mikeCRL (approved):
LGTM

PR #15544, opened 2022-11-03
add "invisible" and "not visible" as synonyms

taroface:
This should enable our docs on not visible indexes to surface when someone searches "invisible indexes".


nickvigilante (changes requested):
Can you merge master into your branch? The Netlify build will fail if you don't.

nickvigilante (approved):
Huh. I seriously thought it would fail. Apologies. LGTM!

taroface:
> Huh. I seriously thought it would fail. Apologies. LGTM!

NP! Sorry that I forgot to do that, though.

PR #15547, opened 2022-11-03
DOC-6101: Add CODEOWNERS to docs repo

nickvigilante:
This will automatically add docs-infra-prs to specific changed files.

Fixes DOC-6101


.github/CODEOWNERS
```diff
@@ -0,0 +1,26 @@
+/.bundle/ @cockroachdb/docs-infra-prs
+/.github/ @cockroachdb/docs-infra-prs
+/__tests__/ @cockroachdb/docs-infra-prs
+/_layouts/  @cockroachdb/docs-infra-prs
+/_plugins/  @cockroachdb/docs-infra-prs
+/archived/  @cockroachdb/docs-infra-prs
```
mikeCRL, line 6: I don't think this should be on `docs-infra-prs`

.github/CODEOWNERS
```diff
@@ -0,0 +1,26 @@
...
+/js/  @cockroachdb/docs-infra-prs
+/netlify/ @cockroachdb/docs-infra-prs
+/vale/  @cockroachdb/docs-infra-prs
+_redirects  @cockroachdb/docs-infra-prs
+.gitattributes  @cockroachdb/docs-infra-prs
+.gitignore  @cockroachdb/docs-infra-prs
+.htmltest.yaml  @cockroachdb/docs-infra-prs
+.vale.ini @cockroachdb/docs-infra-prs
+Gemfile @cockroachdb/docs-infra-prs
+LICENSE @cockroachdb/docs-infra-prs
```
mikeCRL, line 20: I don't think this should be on `docs-infra-prs`. Maybe a different group for docs management—but I'd leave this out for now if you are OK with that. Let me know if you disagree.

.github/CODEOWNERS
```diff
@@ -0,0 +1,26 @@
...
+Makefile  @cockroachdb/docs-infra-prs
+navgoco_license @cockroachdb/docs-infra-prs
```
mikeCRL, line 22: I don't know if this should be on `docs-infra-prs`. What are your thoughts? Also, would you agree that it's a problem that it's hard for people to tell what the license applies to? Do you know if there is any context we can add at the top of the content or file name without 'adjusting' the license?

mikeCRL (changes requested):

.github/CODEOWNERS
nickvigilante: Fair. Will remove.

.github/CODEOWNERS
nickvigilante: That's fair. I'll remove.

.github/CODEOWNERS
nickvigilante: Upon first glance, I can imagine people will have no idea what `navgoco_license` means without having to look it up. I know it's the software we use to render the sidebar because I've had to research it.

I'm not sure if the license is even necessary for us to include in the repo, because the project has been defunct for 7 years. We can discuss that next Thursday.

mikeCRL (approved):

PR #15557, opened 2022-11-04
Added warning certain S3 storage classes not supported with incremental backups

kathancox:
Fixes [DOC-6133](https://cockroachlabs.atlassian.net/browse/DOC-6133)

Added a warning to the S3 storage class examples for backups that incremental backups are incompatible with particular S3 Glacier storage classes.


kathancox:
https://deploy-preview-15557--cockroachdb-docs.netlify.app/docs/dev/use-cloud-storage-for-bulk-operations.html#amazon-s3-storage-classes

shermanCRL (approved):
Thanks!

_includes/v21.2/misc/storage-class-glacier-incremental.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_danger}}
+[Incremental backups](take-full-and-incremental-backups.html#incremental-backups) are **not** compatible with the S3 Glacier Flexible Retrieval or Glacier Deep Archive storage classes. Incremental backups require ad-hoc reading of previous backups. The Glacier Flexible Retrieval or Glacier Deep Archive storage classes do not allow immediate access to S3 objects without first restoring the objects. See Amazon's documentation on [Restoring an archived object](https://docs.aws.amazon.com/AmazonS3/latest/userguide/restoring-objects.html) for more detail.
```
andf-crl, line 2: nit: we could combine sentences 2 and 3 since 2 directly explains 3. What do you think of something like:

> ... storage classes. Incremental backups require ad-hoc reading of previous backups, which is not possible with  the Glacier Flexible Retrieval or Glacier Deep Archive storage classes as they do not allow immediate access to S3 objects without first restoring the objects. See ...

andf-crl (approved):
LGTM with one tiny optional nit, and one thought:

We don't explicitly say that the two S3 Glacier tiers are "supported for full and cluster backups, just not for incremental backups." It might be useful to stick that briefly somewhere: that CRDB supports these S3 storage tiers for non-incremental. Unless we want to lean on "we only test against S3 Standard" as "we don't recommend any non-Standard tier, actually", in which case: nevermind! Just a thought, feel free to disregard!

PR #15562, opened 2022-11-04
Add netlify.toml

nickvigilante:
This is to test out the Lighthouse plugin and get recommendations for improvement of our docs site.


andf-crl (approved):
LGTM -- cool idea, let's check it out!

PR #15584, opened 2022-11-09
loose index scan with recursive CTE

taroface:
DOC-4747

Add an example of a loose index scan using a recursive CTE.

### DIFF
--- a/v22.2/common-table-expressions.md
+++ b/v22.2/common-table-expressions.md
@@ -301,6 +301,8 @@ While adding a limit to prevent infinite recursion works for testing and debuggi
 
 You can use a recursive CTE to perform a loose index scan, which speeds up certain queries that would otherwise require a full scan. A loose index scan reads noncontiguous ranges of an index by performing multiple shorter scans.
 
+In this example, compare the latencies when scanning an index with and without a recursive CTE:
+
 Create a table:
 
 {% include_cached copy-clipboard.html %}
@@ -323,7 +325,7 @@ Create an index:
 CREATE INDEX ON test (n);
 ~~~
 
-Issue a statement to count the number of distinct values:
+Issue a statement to count the number of distinct values, without using a recursive CTE:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
@@ -341,7 +343,7 @@ SELECT COUNT(DISTINCT n) FROM test;
 Time: 273ms total (execution 273ms / network 0ms)
 ~~~
 
-This statement has a high latency because it reads every row in the index.
+This statement has a high latency because it reads every row in the index. You can see this using [`EXPLAIN`](explain.html):
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
@@ -392,7 +394,7 @@ The initial subquery uses the [`LIMIT`](limit-offset.html) and [`ORDER BY`](orde
 Time: 13ms total (execution 13ms / network 0ms)
 ~~~
 
-The recursive CTE has a low latency because it performs 10 limited scans of the index, each reading only one row and skipping the rest.
+The recursive CTE has a low latency because it performs 10 limited scans of the index, each reading only one row and skipping the rest. You can see this using [`EXPLAIN`](explain.html):
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
--- a/v22.2/common-table-expressions.md
+++ b/v22.2/common-table-expressions.md
@@ -295,7 +295,7 @@ SELECT * FROM cte LIMIT 10;
 (10 rows)
 ~~~
 
-While this practice works for testing and debugging, Cockroach Labs does not recommend it in production.
+While adding a limit to prevent infinite recursion works for testing and debugging, Cockroach Labs does not recommend it in production. It is best practice to ensure that recursive subqueries have an explicit end condition.
 
 ### Loose index scan using a recursive CTE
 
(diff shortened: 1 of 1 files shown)


DrewKimball (commented):
Looks pretty good to me. Might be worth mentioning that this pattern incurs the overhead of a new scan for each iteration, and so it will be slower *per row* than a full scan. That makes it faster than a full scan for cases where a lot of rows are skipped, but slower if not.

Also, should we add the blog post this query was pulled from? (https://malisper.me/the-missing-postgres-scan-the-loose-index-scan/)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15584#-:-NGSZynF0nyuvw6adlgQ:b-rswt3e)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @awoods187, @dikshant, and @taroface)

___
*[`v22.2/common-table-expressions.md` line 298 at r1](https://reviewable.io/reviews/cockroachdb/docs/15584#-NGSb244AIQqyc2Ix6AH:-NGSb244AIQqyc2Ix6AI:b-xrxn0v) ([raw file](https://github.com/cockroachdb/docs/blob/8261dffbe0876447a8390989d9ccac9ba7580ded/v22.2/common-table-expressions.md#L298)):*
> ```Markdown
> ~~~
> 
> While this practice works for testing and debugging, Cockroach Labs does not recommend it in production.
> ```

Since we're at it, I seem to remember there being some confusion about whether this line refers to adding a limit to potentially infinite-looping recursive ctes, or recursive ctes in general. Maybe we could clarify it now? Could be something like `While adding a limit to prevent infinite recursion works for testing and debugging, Cockroach Labs does not recommend it in production. It is best practice to ensure that recursive subqueries have an explicit end condition.`
___
*[`v22.2/common-table-expressions.md` line 330 at r1](https://reviewable.io/reviews/cockroachdb/docs/15584#-NGS_6E6CMRsqc34M_jT:-NGS_6E6CMRsqc34M_jU:bnrc97a) ([raw file](https://github.com/cockroachdb/docs/blob/8261dffbe0876447a8390989d9ccac9ba7580ded/v22.2/common-table-expressions.md#L330)):*
> ```Markdown
> ~~~ sql
> EXPLAIN ANALYZE SELECT COUNT(DISTINCT n) FROM test;
> ```

Is there a reason why we omit the timing here and not below? Might be nice to see the concrete comparison.

awoods187 (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15584)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dikshant and @taroface)

taroface (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15584)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @dikshant and @DrewKimball)

___
*[`v22.2/common-table-expressions.md` line 298 at r1](https://reviewable.io/reviews/cockroachdb/docs/15584#-NGSb244AIQqyc2Ix6AH:-NGStT0l0HPjA1xsJyb_:b-cmt49t) ([raw file](https://github.com/cockroachdb/docs/blob/8261dffbe0876447a8390989d9ccac9ba7580ded/v22.2/common-table-expressions.md#L298)):*
<details><summary><i>Previously, DrewKimball (Drew Kimball) wrote…</i></summary><blockquote>

Since we're at it, I seem to remember there being some confusion about whether this line refers to adding a limit to potentially infinite-looping recursive ctes, or recursive ctes in general. Maybe we could clarify it now? Could be something like `While adding a limit to prevent infinite recursion works for testing and debugging, Cockroach Labs does not recommend it in production. It is best practice to ensure that recursive subqueries have an explicit end condition.`
</blockquote></details>

Done! Thanks.
___
*[`v22.2/common-table-expressions.md` line 330 at r1](https://reviewable.io/reviews/cockroachdb/docs/15584#-NGS_6E6CMRsqc34M_jT:-NGSvTyDEna5g2tQXGrw:b-y7lrjj) ([raw file](https://github.com/cockroachdb/docs/blob/8261dffbe0876447a8390989d9ccac9ba7580ded/v22.2/common-table-expressions.md#L330)):*
<details><summary><i>Previously, DrewKimball (Drew Kimball) wrote…</i></summary><blockquote>

Is there a reason why we omit the timing here and not below? Might be nice to see the concrete comparison.
</blockquote></details>

I took it out to save some space. I was on the fence, though. I'll put it back.

DrewKimball (approved):
:lgtm: Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15584#-:-NGSy2rw7VRI6pDLHwH_:b-g2a5fb)* status: :shipit: complete! 1 of 0 LGTMs obtained (waiting on @awoods187 and @dikshant)

taroface:
> Might be worth mentioning that this pattern incurs the overhead of a new scan for each iteration, and so it will be slower _per row_ than a full scan. That makes it faster than a full scan for cases where a lot of rows are skipped, but slower if not.

I just noticed this part of your comment - have added this info to the doc.

v22.2/common-table-expressions.md
```diff
@@ -295,7 +295,155 @@ SELECT * FROM cte LIMIT 10;
 (10 rows)
 ~~~
 
-While this practice works for testing and debugging, Cockroach Labs does not recommend it in production.
+While adding a limit to prevent infinite recursion works for testing and debugging, Cockroach Labs does not recommend it in production. It is best practice to ensure that recursive subqueries have an explicit end condition.
+
+### Loose index scan using a recursive CTE
+
```
kathancox, line 301: General nit: (similar to the PR I just reviewed on trigram indexes) some of the lead-ins to the SQL are a little abrupt. The intro here to this example is mainly telling me what it is, rather than how the example will flow/accomplish. It could be beneficial to add a sentence underneath this to only mention that the example will first scan an index without recursive CTE and then use recursive CTE.

v22.2/common-table-expressions.md
```diff
@@ -295,7 +295,155 @@ SELECT * FROM cte LIMIT 10;
...
+     10
+(1 row)
+
+
+Time: 273ms total (execution 273ms / network 0ms)
+~~~
+
+This statement has a high latency because it reads every row in the index.
+
+{% include_cached copy-clipboard.html %}
```
kathancox, line 346: The two `EXPLAIN` statements aren't necessarily introduced. I like the sentences that precede them as explanation for the statement, but an additional "the following `EXPLAIN` statement analyzes x...:" This might help lead the user through a little more. This could be a me thing, so totally optional.

kathancox (approved):
LGTM, with optional nits!

PR #15585, opened 2022-11-09
Promote multi-region trial clusters on CockroachDB dedicated

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-3278

Just some minor additions to prompt trial dedicated users to create multi-region clusters

### DIFF
--- a/cockroachcloud/frequently-asked-questions.md
+++ b/cockroachcloud/frequently-asked-questions.md
@@ -39,7 +39,7 @@ If you want to create a cluster in a disabled region, please [contact Support](h
 - Create the first cluster in your organization
 - Select 9 or fewer nodes (we recommend starting with 3 so you can try scaling)
 - Select up to 4 vCPUs of compute and 150 GiB of storage (the trial code will not apply to larger clusters)
-- Select 3 or fewer regions
+- Select a single region or 3 regions
 - Don't remove the pre-applied trial code at check out
 
 Once the 30-day period is over, your cluster can be scaled beyond the trial period hardware limitations. You can create other paid clusters at any time. If Cockroach Labs has provided you with additional codes, you can use those on applicable clusters. For extended trial options, [contact us](https://www.cockroachlabs.com/contact-sales/).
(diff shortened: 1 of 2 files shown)


cockroachcloud/frequently-asked-questions.md
```diff
@@ -39,6 +39,7 @@ If you want to create a cluster in a disabled region, please [contact Support](h
 - Create the first cluster in your organization
 - Select 9 or fewer nodes (we recommend starting with 3 so you can try scaling)
 - Select up to 4 vCPUs of compute and 150 GiB of storage (the trial code will not apply to larger clusters)
+- Select 3 or fewer regions
```
rmloveland, line 42: Is our UI going to let people select 2? Seems risky if so (unless I am misunderstanding something?)

My understanding is a 2 region cluster means you can't meaningfully do [multi-region survival goal stuff](http://localhost:4000/docs/dev/multiregion-overview.html#survive-region-failures): If one region goes down the cluster will get stuck and not be able to make progress, since consensus will require nodes in 2/3 regions.  Maybe that's fine tho?  As long as users understand the tradeoffs.  Seems like a footgun to me tho 

Also our docs [call out 2 region multi-region deployments as an anti-pattern](https://www.cockroachlabs.com/docs/stable/topology-patterns.html#anti-patterns)

Suggest getting this reviewed by the PM/Eng who worked on this (altho I guess if its in the UI the decision has been made - but I'd still rather not be the sole reviewer on this)

rmloveland (changes requested):

cockroachcloud/frequently-asked-questions.md
gemma-shay: No, you're right that they still can't pick 2, I just wasn't sure what the best way to phrase it was. Could say "Select 1 or 3 regions" but I thought that sounded a little awkward.

rmloveland: Ah ok makes sense!  Thanks for explaining

rmloveland (approved):
LGTM

PR #15587, opened 2022-11-09
Clarify that we don't support an http api for serverless

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-5887


rmloveland (approved):
LGTM!

PR #15594, opened 2022-11-10
Added replacement `show backup` diagram as local

kathancox:
Fixes DOC-3010

Replaced the remote include for the show backup SQL diagrams for 22.1 & 22.2 since the diagram pulling from the `cockroach` repo exposes some parts of the show backup SQL that are not intended for users. Added a local include instead with the intention of figuring out how to resolve the misalignment between what's in the `cockroach` SQL.y file and what should be user facing.


nickvigilante (approved):
🚀

PR #15632, opened 2022-11-16
accessiBe trial

nickvigilante:
accessiBe is a software that assesses 508 compliance and ensures our website is accessible to all.

https://accessibe.com


nickvigilante:
After discussing this with Mike, we decided not to pursue accessiBe at this time.

PR #15638, opened 2022-11-17
Enable multiple filters per section

ianjevans:
This PR fixes DOC-6209.

With this PR, you can set multiple content filters in a section by adding the filter to the `data-scope` attribute value, separated by spaces.

~~~ html
<section class="filter-content" markdown="1" data-scope="mac linux">

This content will appear if the user selects the Mac or Linux filter.

</section>
~~~


nickvigilante (approved):
This code makes complete sense. You're basically creating an array separated by spaces and then checking which out of the array is the one currently selected and rendering it. Very slick.

cockroachcloud/quickstart.md
```diff
@@ -92,27 +92,20 @@ postgresql://maxroach:ThisIsNotAGoodPassword@free-tier4.aws-us-west-2.cockroachl
...
 
 <section class="filter-content" markdown="1" data-scope="python">
 
 The **Connect to cluster** dialog shows information about how to connect to your cluster.
 
 1. Select **General connection string** from the **Select option** dropdown.
-1. Open a new terminal on your local machine, and run the **CA Cert download command** provided in the **Download CA Cert** section. This certificate is required by most Python clients connecting to {{ site.data.products.db }}.
-1. Open the **General connection string** section, then copy the connection string provided and save it in a secure location.
+2. Open a new terminal on your local machine, and run the **CA Cert download command** provided in the **Download CA Cert** section. This certificate is required by most Python clients connecting to {{ site.data.products.db }}.
+3. Open the **General connection string** section, then copy the connection string provided and save it in a secure location.
```
taroface, line 108: Curious about the reason for renumbering these. We typically use `1.` for all ordered lists, per style guide.

taroface (approved):
LGTM! Thanks for this excellent change. I had an unrelated question about list numbering.

cockroachcloud/quickstart.md
ianjevans: This was unintentional. I think maybe VS Code automatically modified these?

nickvigilante: I will try to add a Vale test to catch this. @nickvigilante

PR #15646, opened 2022-11-17
DOC-6215: Add Vale check for any ordered lists that don't start with 1.

nickvigilante:
Fixes DOC-6215


ianjevans:
Just did a search for what this would catch using:

~~~ shell
ag '^\s*[23456789][0-9]*\.'
~~~

And I found this false positive:

~~~
v19.1/sql-constants.md
140:3.1415
142:6.626e-34
~~~

Probably not a big deal, as it's non-blocking.

nickvigilante:
I updated it to `^\s*[23456789][0-9]*\.\D`, so that it would skip over those. I also went ahead and just updated everything current to use 1.

mikeCRL (approved):
LGTM to merge after @nickvigilante explains change to the team via Slack (and perhaps after sharing vs code solution to default to 'continue' list numbering with `1.`).

_includes/cockroachcloud/cdc/create-core-changefeed.md
```diff
@@ -11,28 +11,29 @@ In this example, you'll set up a core changefeed on your {{ site.data.products.s
 
     {% include cockroachcloud/cdc/core-csv.md %}
 
-3. Enable the `kv.rangefeed.enabled` [cluster setting](../{{site.versions["cloud"]}}/cluster-settings.html):
+
```
nickvigilante, line 14:
```suggestion
```

PR #15651, opened 2022-11-18
Update Rapidoc to 9.3.3

ianjevans:
Rapidoc 9.3.3 includes some bug fixes that better show the schema of request and response objects. This PR updates the library file from our fork with 9.3.3.


fantapop (approved):
I wouldn't normally have a merge commit in a PR but it seems like there are other ones in the docs repo and I'm not that familiar with the process used on the docs team so 🤷 .  You should be able to get rid of it by rebase your branch off of master. But either way looks good to me.

ianjevans:
> I wouldn't normally have a merge commit in a PR but it seems like there are other ones in the docs repo and I'm not that familiar with the process used on the docs team so 🤷 . You should be able to get rid of it by rebase your branch off of master. But either way looks good to me.

We tend to use merge commits so reviewers can view changes since their last review. Rebasing would require force pushing, which resets the commit history.

fantapop:
Sorry, I just meant that there was a merge commit in your PR.  It looks like it was squashed anyways and didn't end up in the final repo.

PR #15653, opened 2022-11-18
Use SNI connection strings for serverless docs

ianjevans:
This PR replaces all references to the old connection string using either `--options=cluster%3d<routing-id>` or `<routing-id>.<database name>` with the new SNI format URLs.

Fixes CC-6286.


_includes/cockroachcloud/sql-connection-string-free.md
```diff
@@ -2,7 +2,7 @@
 
     {% include_cached copy-clipboard.html %}
     ~~~ shell
-    cockroach sql --url 'postgresql://<username>:<password>@<serverless-host>:26257/defaultdb?sslmode=verify-full&sslrootcert='$HOME'/.postgresql/root.crt&options=--cluster=<routing-id>'
+    cockroach sql --url 'postgresql://<username>:<password>@<serverless-host>:26257/defaultdb?sslmode=verify-full&sslrootcert='$HOME'/.postgresql/root.crt'
```
emily-horing, line 5: Nit: we use <> in the javascript examples but all the other examples seem to be using {} for placeholder text.

cockroachcloud/migrate-from-serverless-to-dedicated.md
```diff
@@ -27,7 +27,7 @@ In {{ site.data.products.serverless }} clusters, all external service integratio
 {% include copy-clipboard.html %}
 ~~~ shell
 $ cockroach sql \
---url 'postgres://<username>:<password>@free-tier.<region>.cockroachlabs.cloud:26257?sslmode=verify-full&sslrootcert=<path/to/certs_dir>/cc-ca.crt&options=--cluster=<cluster_name>' \
+--url 'postgres://<username>:<password>@<cluster-name>.<region code>.cockroachlabs.cloud:26257?sslmode=verify-full&sslrootcert=<path/to/certs_dir>/cc-ca.crt' \
```
emily-horing, line 30: is this right? Wouldn't it be the {host} not {cluster-name}?

cockroachcloud/troubleshooting-page.md
```diff
@@ -38,12 +38,6 @@ Error: x509: certificate signed by unknown authority
 Failed running "sql"
 ~~~
 
-There are two possible causes of this error: incorrect [routing ID](../{{site.versions["stable"]}}/connect-to-the-database.html#connection-parameters) in the connection string, and CA certificate conflicts in the `cockroach` certificate search path.
-
-<h4>Solution: incorrect routing ID in the connection string</h4>
-
-Check if you are using the right routing ID in the [connection method](connect-to-your-cluster.html#step-3-connect-to-your-cluster). You can find your routing ID in the {{ site.data.products.db }} Console by navigating to **Cluster Overview** > **Connect** > **Select option/language** and select **General connection string**, and then locating the parameter `cluster={routing-id}` in your connection string.
```
emily-horing, line 45: I'm not sure we want to remove this fully. Seems like here is a good place to show the new SNI and say "switch to SNI login for this tool" as a solution.

cockroachcloud/troubleshooting-page.md
```diff
@@ -65,7 +59,7 @@ FATAL: CodeParamsRoutingFailed: rejected by BackendConfigFromParams: Invalid clu
 
 <h4>Solution</h4>
 
-Check that you are using the correct cluster and database names. You can find these parameters in the {{ site.data.products.db }} Console by navigating to **Cluster Overview** > **Connect** > **Step 2. Connect** > **Connection parameters**. For most tools, the full name of your database should be in the format `<routing-id>.<database>` for {{ site.data.products.serverless }} clusters.
+Check that you are using the correct cluster and database names. You can find these parameters in the {{ site.data.products.db }} Console by navigating to **Cluster Overview** > **Connect** > **Step 2. Connect** > **Connection parameters**.
```
emily-horing, line 62: Is there a spot where we do want to show/explain that there are multiple options to connect? What if the SNI version doesn't work for a GUI tool or something? Will there be somewhere we point users to?

emily-horing (approved):
Some nits but overall lgtm. This makes things appear a lot simpler!

cockroachcloud/connect-to-a-serverless-cluster.md
```diff
@@ -71,17 +71,16 @@ For connection examples and code snippets in your language, see the following:
...
 
   </section>
   <section class="filter-content" markdown="1" data-scope="cockroachdb-client">
 
-1. In the **Download CA Cert** section of the dialog, select your operating system, and use the command provided to download the CA certificate to the default PostgreSQL certificate directory on your machine.
-1. In the **Download the latest CockroachDB Client** section of the dialog, select your operating system, and use the command provided to install CockroachDB.
-1. Copy the [`cockroach sql`](../stable/cockroach-sql.html) command and connection string provided in the **Connect** modal, which will be used in the next step (and to connect to your cluster in the future).
-1. In your terminal, enter the copied `cockroach sql` command and connection string to start the [built-in SQL client](../{{site.versions["stable"]}}/cockroach-sql.html).
+2. In the **Download CA Cert** section of the dialog, select your operating system, and use the command provided to download the CA certificate to the default PostgreSQL certificate directory on your machine.
```
gemma-shay, line 78: Is the numbered list instead of all 1s correct?

cockroachcloud/troubleshooting-page.md
gemma-shay: My understanding is also that the routing id still exists, we're just not showing it in the connection string. Maybe we should have some guidance for users who still need it...?

gemma-shay (approved):

cockroachcloud/migrate-from-serverless-to-dedicated.md
ianjevans: Yeah, good point. I'll change it to <host>.

cockroachcloud/troubleshooting-page.md
ianjevans: There isn't any tool where the user can get the old-style connection string. Recreating the old-style connection string from the new parameters would be really confusing to document.

AFAIK in the bug-bash we didn't find any tools that were incompatible with the SNI version.

IMO, if this is important, then we should add an option to the Cloud Console and `ccloud` to get the old-style connection string.

cockroachcloud/connect-to-a-serverless-cluster.md
ianjevans: No, it isn't. I got bit by VS Code reformatting the ordered lists. Fixing.

_includes/cockroachcloud/sql-connection-string-free.md
ianjevans: Yeah, that's annoying. Since this is included in multiple topics that use `<>` in the non-included text, I'll have to punt on this for now.

ianjevans:
> My understanding is also that the routing id still exists, we're just not showing it in the connection string. Maybe we should have some guidance for users who still need it...?

I added a note to the connection info about how the routing ID is now included in host name.

PR #15666, opened 2022-11-22
[DOC-5483] Document new limitation when upgrading to an alpha version or manual build

mdlinville:
[DOC-5483]
Relates to https://github.com/cockroachdb/cockroach/pull/86345

## Previews
[releases/index.md](https://deploy-preview-15666--cockroachdb-docs.netlify.app/docs/releases/index.html)
[v22.2/upgrade-cockroach-version.md](https://deploy-preview-15666--cockroachdb-docs.netlify.app/docs/v22.2/upgrade-cockroach-version.html)

## Tests
- [x] Local build
- [x] Linkcheck


dt (approved):
you also can no longer actually do such a one-way upgrade without passing an opt-in environment variable but I'm not sure we really want to go into too much detail about that so LGTM as is

PR #15667, opened 2022-11-22
Added docs on changes 22.2 to control schedule privileges

kathancox:
Fixes DOC-5616

Adjusted the privileges section for pause, drop, resume, show schedule statements so that it lists owners of scheduled backups can use these statements to control schedules.


andf-crl (approved):
LGTM!

PR #15675, opened 2022-11-22
Included reference to change in CDC CSV format 22.2

kathancox:
Fixes DOC-5870

Included a couple of lines for the change to the CSV format. This has been added to the Changefeed Messages, `CREATE CHANGEFEED` options table, and Export Data with Changefeeds pages.

Also, added a few small formatting changes.


_includes/v22.2/cdc/csv-changefeed-format.md
```diff
@@ -0,0 +1 @@
+{% include_cached new-in.html version="v22.2" %} Changefeeds emit the same CSV format as [`EXPORT`](export.html). In v22.1, changefeeds emitted CSV data that wrapped some values in single quotes, which were not wrapped when exporting data with the `EXPORT` statement.
```
andf-crl, line 1: nit: we can just say `Previously, changefeeds emitted ...` to avoid worrying about whether previous behavior also existed in v21.2 or earlier. Pinning to a specific previous release suggests that it only applies to that singular previous release.

v22.2/export-data-with-changefeeds.md
```diff
@@ -5,7 +5,7 @@ toc: true
 docs_area: stream_data
 ---
 
-{% include_cached new-in.html version="v22.1" %} When you create an {{ site.data.products.enterprise }} changefeed, you can include the [`initial_scan = 'only'`](create-changefeed.html#initial-scan) option to specify that the changefeed should only complete a table scan. The changefeed emits messages for the table scan and then the job completes with a `succeeded` status. As a result, you can create a changefeed with `initial_scan = 'only'` to [export](export.html) data out of your database.
```
andf-crl, line 8: Oooh, I thought our GA release process was supposed to catch and delete these previous-version new-in's when we cut the next one. Surprised to see this here.

andf-crl (approved):
LGTM! Just some optional thoughts. Thanks!

v22.2/export-data-with-changefeeds.md
kathancox: Yeah... me too, not sure how that could have happened.

_includes/v22.2/cdc/csv-changefeed-format.md
kathancox: The reason I pinned this was because I knew that the `format='csv'` option only became available in v22.1. You feel it is still necessary to change in that case?

andf-crl: Gotcha, not at all then! Thanks for clarification.

PR #15682, opened 2022-11-23
Removed cloud filters from bulk pages due to bug

kathancox:
Fixes DOC-6110

Removed the cloud filters from the backup, import, import into, export, restore from v21.1 to v22.2. The filters were breaking the sidebar TOC since the section headings were the same. On weighing up whether to change the headings or remove the filters, it seemed that removing the filters had far more benefits, notably:

- No need for awkward, long example headings.
- Removes a lot of repetitive content from the docs repo without losing much in terms of benefit.
- Enables us to add content with more ease. 
- Where there are cloud-specific examples in the future, we can add specific headings (the S3 storage class feature is a good example of this).

cc @taroface for migration-related sql


kathancox:
Hey @andf-crl just dropping the review request for now as I need to make some additional changes!

_includes/v21.1/backups/bulk-auth-options.md
```diff
@@ -1 +1 @@
-The examples below provide connection strings to Amazon S3, Google Cloud Storage, and Azure Storage. For guidance on connecting to other storage options or using other authentication parameters, read [Use Cloud Storage for Bulk Operations](use-cloud-storage-for-bulk-operations.html#example-file-urls).
+The following examples provide connection strings to Amazon S3. For guidance on connecting to other storage options or using other authentication parameters, read [Use Cloud Storage for Bulk Operations](use-cloud-storage-for-bulk-operations.html#example-file-urls). Furthermore, the examples use the **default** `AUTH=specified` parameter. For more detail on how to use `implicit` authentication with Amazon S3 buckets, read [Use Cloud Storage for Bulk Operations — Authentication](use-cloud-storage-for-bulk-operations.html#authentication).
```
andf-crl, line 1: suggestion: since two caveats to following examples, what do you think of bullet points for each? Maybe something like:

> The following examples make use of:
> - Amazon S3 connection strings. For guidance on connecting to other storage options or using other authentication parameters instead, read [Use Cloud Storage for Bulk Operations](use-cloud-storage-for-bulk-operations.html#example-file-urls).
> -  The **default** `AUTH=specified` parameter. For guidance on using `AUTH=implicit` authentication with Amazon S3 buckets instead, read [Use Cloud Storage for Bulk Operations — Authentication](use-cloud-storage-for-bulk-operations.html#authentication).

supernit: added `instead` to help drive home that these links take us away from current page approach to alternate options.

Could see this on the other like includes as well.

andf-crl (approved):
LGTM, just one optional suggestion. Thanks!

PR #15708, opened 2022-11-29
display filtered content inline

taroface:
Filtered content does not filter or display correctly when a section block is part of a list (I think I'm the only one who has tried this). In order to filter and display properly, the entire list can be formatted in HTML rather than Markdown. Example:

```
<ul>
<li>blah</li>
<li>blah</li>
<section class="filter-content" markdown="1" data-scope="filter">
<li>filtered blah</li>
</section>
<li>blah</li>
</ul>
```

To avoid inconsistent line spacing, the filtered section needs to display as `inline` rather than `block`. After testing on some existing docs, this CSS change doesn't appear to affect how other filtered content is displayed. I believe it only improves the situation described above. Let me know if I should add any explicit tests to this PR.

### DIFF
--- a/css/components/_filtertabs.scss
+++ b/css/components/_filtertabs.scss
@@ -262,7 +262,7 @@
 	display: none;
 
 	&.current {
-		display: block;
+		display: inline;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


taroface:
Added a fix for the Node Shutdown doc. Previously the last bullet in the reformatted list was incorrectly displayed with the `drain` as well as the `decommission` tab (i.e., wasn't actually being filtered).

nickvigilante (approved):

taroface:
TFTR!

PR #15709, opened 2022-11-29
Update performance benchmarking docs for v22.2

rmloveland:
Fixes DOC-5815
Partially addresses DOC-5511

Summary of changes:

- Update performance overview page to:

  - Clarify that the versions under discussion are v21.1 and v19.2

  - Remove the misleading "p95 latency" metric from the chart on that page, which was confusing because it showed a higher value for the "better" result.  More docs are needed around this to explain the nuances (based on the convo in DOC-5511), but at least removing the misleading stuff for now

- Reviewed other pages in the performance benchmarking section and did not find anything else that mentions specific versions of CockroachDB

### DIFF
--- a/v21.2/performance.md
+++ b/v21.2/performance.md
@@ -16,7 +16,7 @@ This document is about CockroachDB performance on benchmarks. For guidance on tu
 
 ## Scale
 
-TPC-C provides the most realistic and objective measure for OLTP performance at various scale factors. CockroachDB can process **1.68M tpmC with 140,000 warehouses, resulting in an efficiency score of 95%.** As shown in the following chart, this is a 40% improvement over the results from CockroachDB 19.2.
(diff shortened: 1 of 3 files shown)


rmloveland:
@sean- 

since you commented on DOC-5511: this docs PR removes the misleading p95 from the TPC-C comparison chart (for all supported versions v21.2+, per https://www.cockroachlabs.com/docs/releases/release-support-policy)

however, it does _not_ address [your comment](https://cockroachlabs.atlassian.net/browse/DOC-5511?focusedCommentId=122395) re: the docs should provide the add'l nuance on why the latency numbers are higher on the "better" result

therefore we're leaving DOC-5511 open for future work to add that nuance at some point (maybe these benchmarks will get re-run Soon (tm) and we'll update that page altogether, hopefully)

This specific PR is just to get us into a (slightly?) better state for the v22.2.0 marketing release date next Tuesday (6 December 2022)

ianjevans (approved):

PR #15652, opened 2022-11-18
DOC-5188 DOC-5296 RBAC new sys privs

andf-crl:
Addresses: DOC-5188, DOC-5296

- Added new system privilege versions of existing roles:
	- VIEWACTIVITY
	- VIEWACTIVITYREDACTED
	- VIEWCLUSTERSETTING
	- CANCELQUERY
	- NOSQLLOGIN
- Adds new observability system privileges:
	- VIEWDEBUG
	- VIEWCLUSTERMETADATA
- Updated mention of deprecated roles with new system privileges, framing roles as "legacy" (to match live CDC and BulkIO language for this effort)
- Added "or user in the `admin` role" to several statements, as your code confirms you check for both.

NOTES:

- Ignore double "Role Options" on create-user, etc pages. Have D O C - 6 2 2 8 to fix sitewide after this effort (present in live dev docs, not related to this PR).

QUESTIONS:

- Updated `hot ranges` page to indicate that `VIEWCLUSTERMETADATA` priv is now required. Explicitly indicated that legacy `VIEWACTIVITY` and `VIEWACTIVITYREDACTED` roles still work, but that new `VIEWACTIVITY` and `VIEWACTIVITYREDACTED` privs do not. Just confirming that this correct. (i.e. you permit legacy roles to work in v22.2 for this one release only, and users should be using the new `VIEWCLUSTERMETADATA` priv for this)?
- Any uses for `VIEWDEBUG` outside of (undocumented) debug endpoints and (seemingly-undocumented) trace-taking? Otherwise, it now appears only on the privs list table, and nowhere else. Or:
- Is [this](https://www.cockroachlabs.com/docs/stable/show-trace.html#required-privileges) still accurate, or maybe does [this](https://www.cockroachlabs.com/docs/stable/cockroach-debug-job-trace.html) now require `VIEWDEBUG`?
- Another clarification: `NOSQLLOGIN` is deprecated ... but `SQLLOGIN` is not? Same with all inverse-refutation-named privs basically: am I missing something in my notation that the one listed (in your code) is deprecated, while the one not (present in your code) is still alive? Both versions still work for me in v22.2.0-rc.3, but then even the deprecated ones are guaranteed to last one more GA release, per design. Just checking, thank you.

SELECTED STAGING:

[...]

PR #15641, opened 2022-11-17
Add docs on global MODIFYCLUSTERSETTING privilege

rmloveland:
Fixes DOC-4749
Fixes DOC-4597

Summary of changes:

- Add new Global privilege to list of privileges

- Add 'Grant global privileges on the entire cluster' example to `GRANT` docs

- Add 'Revoke global privileges on the entire cluster' example to `REVOKE` docs

- Update the required privileges in the `SET CLUSTER SETTING` docs

- Drive-by add `SHOW ROLES` link to the `SHOW DEFAULT PRIVILEGES` doc

PR #15718, opened 2022-12-01
DOC-6264 central place for RBAC content

andf-crl:
Addresses: DOC-6264

- Added a central place for all new system privilege / RBAC content on the existing `v22.2/security-reference/authorization.md` page.
	- Added support chart, incl which privs replace which roles, and which are net-new
	- Did a first pass at introducing the concept of RBAC & inheritance
	- Tried to de-obfuscate nearly-identically-named concepts with a generous peppering of `{{site.data.alerts.callout_info}}`
	- Tried to describe things to users that haven't really been described to me. Did I guess right?

Research

- Some engineers call it [global](https://github.com/cockroachdb/docs/pull/15641#pullrequestreview-1190712479), some engineers call it [system](https://github.com/cockroachdb/docs/pull/15652#issuecomment-1330919151), some engineers use [both](https://github.com/cockroachdb/cockroach/pull/80580/files). What use does a database company have for consistency? :joy:
- The commands themselves suggest that we go with `system`:
	```
	GRANT SYSTEM
	SHOW SYSTEM GRANTS
	```
- System privileges are introduced alongside role option. As such:
	- They do not replace current, regular-old privileges like `ALL` or `WRITE`. They are pretty much unrelated, even insofar as requiring different management commands `GRANT SYSTEM` and `SHOW SYSTEM GRANTS`.
	- Not all role options are deprecated, only the ones that have new system privilege variants introduced alongside v22.2. See the support chart I've included in this PR for which ones.
	- If a system privilege and role option are both defined, the system privilege takes precedence. Guess: this only applies to the 5 new sys privs which have old role option counterparts.

Next steps:

- Instances in the docs that discuss `global privilege` should be updated to use `system privileges`. Rich, think this is just your PR. I'll make that update before GA.
- Instances in the docs that link to system privileges that currently use `[system privilege](security-reference/authorization.html#supported-privileges)` should be updated to use `[system privileges](security-reference/authorization.html#system-privileges)` instead. Kathryn this is yours and my PRs. I'll make this update also before GA.
- Someone has to have a better way of saying: privileges and system privileges are not the same, but are related, in some peculiarly nuanced way, that I don't understand. I put something base level in there, but probably need an engineer.

Resources:

- Original RFC for concept: RBAC @ CRDB: https://github.com/cockroachdb/cockroach/pull/80580/files
- Rich's massively-helpful assembled chart of support + comment thread with engineering: https://github.com/cockroachdb/docs/pull/15641 Thank you Rich!!
- Tommy's comment to D O C - 4 6 4 8.
- I drew language for this PR from other writer's excellent work on this topic:
	- Kathryn: https://github.com/cockroachdb/docs/pull/15393
	- Rich: https://github.com/cockroachdb/docs/pull/15641

Naming

Just for fun, it appears that I can do this and remain correct:

```
<span class="version-tag">New in v22.2:</span> When assigning system privileges (but not regular privileges) to roles (which are different than role options), use the `NOSQLLOGIN` system privilege instead of the deprecated `SLQLOGIN` role option, and the `VIEWACTIVITY` system privilege instead of the deprecated `NOVIEWACTIVITY` role option.
```

Staging

[v22.2/security-reference/authorization.md](https://deploy-preview-15718--cockroachdb-docs.netlify.app/docs/dev/security-reference/authorization.html)

### DIFF
--- a/v22.2/security-reference/authorization.md
+++ b/v22.2/security-reference/authorization.md
@@ -7,19 +7,32 @@ docs_area: reference.security
 
 Authorization, generally, is the control over **who** (users/roles) can perform **which actions** (e.g read, write, update, delete, grant, etc.) to **which resources or targets** (databases, functions, tables, clusters, schemas, rows, users, jobs, etc.).
 
-{{site.data.alerts.callout_info}}
 CockroachDB has a unified authorization model, meaning that a given user's permissions are governed by the same policies in different contexts such as accessing the SQL shell or viewing data from the DB Console.
+
+## Authorization models
+
+{{site.data.alerts.callout_info}}
+Starting in v22.2, CockroachDB introduces a new granular [system-level privilege model](#system-level-privileges) that provides finer control over a user's ability to work with the database. This new system-level privilege model is intended to replace the existing [role options model](#role-options) in a future release of CockroachDB. As such, any legacy role options that now have corresponding system-level privilege versions are **deprecated in CockroachDB v22.2**, though both are supported alongside each other in v22.2. We recommend familiarizing yourself with the new system-level privilege model, and implementing it where possible.
 {{site.data.alerts.end}}
+
+<span class="version-tag">New in v22.2:</span> CockroachDB offers two authorization models:
+
+Authorization Model         | Features
+----------------------------|---------------------------------------
+[System-level Privileges](#system-level-privileges)  | <ul><li> Introduced in CockroachDB v22.2.</li><li> Supported in CockroachDB v22.2, alongside the existing role options.</li><li>Apply cluster-wide. A system-level privilege is granted at the cluster level, and is inherited via role membership, similar to how [object-level privileges](#privileges) are inherited.</li><li>Are granted with the [`GRANT`](../grant.html) statement using the `SYSTEM` parameter, and viewed with the [`SHOW SYSTEM GRANTS`](../show-system-grants.html) statement.</li><li>May be assigned with the [`GRANT OPTION`](../grant.html), which allows the assigned user or role to further grant that same system-level privilege to other users or roles.</li></ul>
+[Role Options](#role-options) |  <ul><li> [Specific role options](#system-level-privileges) which have had corresponding system-level privileges introduced in CockroachDB v22.2 are now **deprecated**.</li><li>Supported in CockroachDB v22.2, alongside the new system-level privileges.</li><li>Apply only to specific users, and are not inheritable via role membership.</li><li>Are granted with the [`GRANT`](../grant.html) statement, and viewed with the [`SHOW GRANTS`](../show-grants.html) statement.</li><li>May be assigned with the [`GRANT OPTION`](../grant.html), which allows the assigned user or role to further grant that same role option to other users or roles.</li></ul>
+
+If a system-level privilege exists with the same name as a role option, the system-level privilege should be used.
 
 ## Users and roles
 
-There is no technical distinction between a role or user in CockroachDB. A role/user can:
+Both [authorization models](#authorization-models) make use of the concept of user and roles. There is no technical distinction between a role or user in CockroachDB. A role/user can:
 
 - Be permitted to log in to the [SQL shell](../cockroach-sql.html).
 - Be granted [privileges](#privileges) to specific actions and database objects.
 - Be a member of other users/roles, inheriting their privileges.
 - Have other users/roles as members that inherit its privileges.
-- Be configured with other [role options](#role-options)
+- Be configured with other [role options](#role-options).
 
 We refer to these as "roles" when they are created for managing the privileges of their member "users" and not for logging in directly, which is typically reserved for "users".
 
@@ -48,6 +61,10 @@ The `root` user is created by default for each cluster. The `root` user is assig
 For secure clusters, in addition to [generating the client certificate](../authentication.html#client-authentication) for the `root` user, you can assign or change the password for the `root` user using the [`ALTER USER`](../alter-user.html) statement.
 
 ## Roles
+
+{{site.data.alerts.callout_info}}
+This section describes roles. For role options like `CREATEROLE`, see [role options](#role-options).
+{{site.data.alerts.end}}
 
 A role is a group of users and/or other roles for which you can grant or revoke privileges as a whole. To simplify access management, create a role and grant privileges to the role, then create SQL users and grant them membership to the role.
 
@@ -119,15 +136,43 @@ Roles and users can be granted the following privileges:
 
 {% include {{ page.version.version }}/sql/privileges.md %}
 
+### System-level privileges 
+
+<span class="version-tag">New in v22.2:</span> System-level privileges offer more granular control over a user's actions when working with CockroachDB, compared to the [role options authorization model](#role-options).
+
+System-level privileges are a special kind of privilege that apply cluster-wide, meaning that the privilege is not tied to any specific object in the database.
+
+You can work with system-level privileges using the [`GRANT `](../grant.html) statement with the `SYSTEM` parameter, and the [`SHOW SYSTEM GRANTS`](../show-system-grants.html) statement.
+
+The following table lists the new system-level privileges introduced with CockroachDB v22.2, and indicates which system-level privileges replace role options:
+
+New System-level Privilege  | Replaces Legacy Role Option
+----------------------------|---------------------------------------
+`MODIFYCLUSTERSETTING`      | Yes: the `MODIFYCLUSTERSETTING` and `NOMODIFYCLUSTERSETTING` role option
+`EXTERNALCONNECTION`        | No, new in v22.2
+`VIEWACTIVITY`              | Yes: the `VIEWACTIVITY` and `NOVIEWACTIVITY` role options
+`VIEWACTIVITYREDACTED`      | Yes: the `VIEWACTIVITYREDACTED` and `NOVIEWACTIVITYREDACTED` role options
+`VIEWCLUSTERSETTING`        | Yes: the `VIEWCLUSTERSETTING` and `NOVIEWCLUSTERSETTING` role options
+`CANCELQUERY`               | Yes: the `CANCELQUERY` and `NOCANCELQUERY` role options
+`NOSQLLOGIN`                | Yes: the `SQLLOGIN` and `NOSQLLOGIN` role options
+`VIEWCLUSTERMETADATA`       | No, new in v22.2
+`VIEWDEBUG`                 | No, new in v22.2
+`BACKUP`                    | No, new in v22.2
+`RESTORE`                   | No, new in v22.2
+`EXTERNALIOIMPLICITACCESS`  | No, new in v22.2
+`CHANGEFEED`                | No, new in v22.2
+
+If a system-level privilege exists with the same name as a role option, the system-level privilege should be used. Some role options do not have a corresponding system-level privilege, since they configure per-user attributes. For those system-level privileges that replace legacy role options (such as `VIEWACTIVITY`), if both the system-level privilege and its legacy role option are specified for a user/role, the system-level privilege will take precedence and the legacy role option will be ignored.
+
 ### Managing privileges
 
 Use the [`GRANT`](../grant.html) and [`REVOKE`](../revoke.html) statements to manage privileges for users and roles.
 
 Take the following points into consideration while granting privileges to roles and users:
 
-- When a role or user is granted privileges for a database, that role or user is not automatically granted access to any new or existing objects within that database. To change access to those objects, see [Default privileges](#default-privileges).
+- When a role or user is granted privileges for a database, that role or user is not automatically granted access to any new or existing objects within that database. To change access to those objects, see [Default privileges](#default-privileges). This does not apply to system-level privileges, which apply cluster-wide.
 - When a role or user is granted privileges for a table, the privileges are limited to the table.
-- In CockroachDB, privileges are granted to users and roles at the database and table levels. They are not yet supported for other granularities such as columns or rows.
+- In CockroachDB, privileges are granted to users and roles at the database and table levels, or cluster-wide at the system level. They are not yet supported for other granularities such as columns or rows.
 - The `root` user automatically belongs to the `admin` role and has the `ALL` privilege for new databases.
 - For privileges required by specific statements, see the documentation for the respective [SQL statement](../sql-statements.html).
 
@@ -165,3 +210,4 @@ We recommend the following best practices to set up access control for your clus
 
 - Run bulk `ROLE` operations inside a transaction.
 - Run regularly-scheduled `ROLE` operations together, rather than at different times throughout the day.
+- Generally, if a [system-level privilege](#system-level-privileges) exists with the same name as a [role option](#role-options), the system-level privilege should be used. 


andf-crl:
@rafiss Hi friend, would you mind helping out with one more RBAC PR? This to try to invent a sort of central resource for the RBAC system priv changes introduced in v22.2 since it appears missing as of nearly-now-GA. Thanks! Feel free to pass along to another engineer if preferred. Appreciate your help!

rafiss (changes requested):
regarding "privileges and system privileges are not the same, but are related, in some peculiarly nuanced way, that I don't understand. I put something base level in there, but probably need an engineer."

here's my best summary: "normal privileges" are defined on objects (such as tables, databases, types). "system privileges" are defined "globally" (i.e., across the whole cluster), hence the name "global privileges"

from a user's perspective, they are the same except for needing slightly different syntaxes since "system privileges" cannot have the "ON" clause in the GRANT statement. the most important similarity is that they have the same inheritance semantics 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15718#-:-NIEThBf-c_Zw0LNEXWQ:bkhjbt2)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl, @kathancox, and @rmloveland)

___
*[`v22.2/security-reference/authorization.md` line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEPvjF1E6k3tF604vY:-NIEPvjF1E6k3tF604vZ:b-ixq02y) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L15)):*
> ```Markdown
> 
> Starting in v22.2, CockroachDB introduces a new granular [system privilege model](#system-privileges) that provides finer control over a user's ability to work with the database. This new system privilege model is intended to replace the existing [role options model](#role-options) in a future release of CockroachDB. As such, any legacy role options that now have corresponding system privilege versions are **deprecated in CockroachDB v22.2** and **will be removed in a future release of CockroachDB**, though both are supported alongside each other for v22.2. We recommend familiarizing yourself with the new system privilege model, and implementing it wherever possible.
> ```

i don't think we need to mention that they are getting removed, since we don't actually have plans to do that.
___
*[`v22.2/security-reference/authorization.md` line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEPT364-F3uvZzzXHe:-NIEPT364-F3uvZzzXHf:b-jgo2et) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L22)):*
> ```Markdown
> Authorization Model         | Features
> ----------------------------|---------------------------------------
> [System Privileges](#system-privileges)  | <ul><li> Introduced in CockroachDB v22.2.</li><li> Supported in CockroachDB v22.2, alongside the existing role options.</li><li>Inheritable with the [`GRANT OPTION`](#inheriting-system-privileges).</li></ul>
> ```

"Inheritiable with the GRANT OPTION" doesn't seem right to me. System privileges are always inherited no matter what. The GRANT OPTION just says if a user has a system privilege, that they can in turn grant that privilege to other users. the GRANT OPTION is a flag that's defined for each privilege that a user has. Also, the GRANT OPTION exists for normal privileges as well as system privileges.

the `#inheriting-system-privileges` section needs updating too
___
*[`v22.2/security-reference/authorization.md` line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEQ8Av58BdnLbnp9H5:-NIEQ8Av58BdnLbnp9H6:b-4z3xgk) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L23)):*
> ```Markdown
> ----------------------------|---------------------------------------
> [System Privileges](#system-privileges)  | <ul><li> Introduced in CockroachDB v22.2.</li><li> Supported in CockroachDB v22.2, alongside the existing role options.</li><li>Inheritable with the [`GRANT OPTION`](#inheriting-system-privileges).</li></ul>
> [Role Options](#role-options) |  <ul><li> [Specific role options](#system-privileges) which have had corresponding system privilege versions introduced in CockroachDB v22.2 are now **deprecated** and **will be removed in a future release of CockroachDB**.</li><li>Supported in CockroachDB v22.2, alongside the new system privileges.</li><li>Not inheritable. Role options are explicit per user/role.</li></ul>
> ```

"explicit per user/role" doesn't feel like the most clear way to phrase this. what about "... are not inherited from parent roles."
___
*[`v22.2/security-reference/authorization.md` line 25 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEQO610oZ1v9BDNa2Q:-NIEQO610oZ1v9BDNa2R:byiiv79) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L25)):*
> ```Markdown
> [Role Options](#role-options) |  <ul><li> [Specific role options](#system-privileges) which have had corresponding system privilege versions introduced in CockroachDB v22.2 are now **deprecated** and **will be removed in a future release of CockroachDB**.</li><li>Supported in CockroachDB v22.2, alongside the new system privileges.</li><li>Not inheritable. Role options are explicit per user/role.</li></ul>
> 
> The intention is that role options should be reserved for user administration permissions, while system privileges should be used to define database actions such as reads and writes. Generally, if a system privilege exists that covers an authorization need in your workflow, you should use the system privilege instead of a role option.
> ```

I don't quite follow this sentence: "The intention is that role options should be reserved for user administration permissions, while system privileges should be used to define database actions such as reads and writes."

minimal correction: instead of "define" it should say "control" or "restrict"

but I think we can just change this to say something like "If a system privilege exists with the same name as a role option, the system privilege should be used. Some role options do not have a corresponding system privilege, since they configure per-user attributes"
___
*[`v22.2/security-reference/authorization.md` line 179 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIERspD9rr_6LCD0STy:-NIERspD9rr_6LCD0STz:bq890tn) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L179)):*
> ```Markdown
> New System Privilege        | Replaces Legacy Role Option
> ----------------------------|---------------------------------------
> `MODIFYCLUSTERSETTING`      | No, new in v22.2
> ```

correction: `MODIFYCLUSTERSETTING` does replace the old role option
___
*[`v22.2/security-reference/authorization.md` line 193 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIESKVfCk1Lp34_E_UD:-NIESKVg71quQsINkRgr:b-4bt8dj) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L193)):*
> ```Markdown
> `CHANGEFEED`                | No, new in v22.2
> 
> Generally, if a system privilege exists that covers an authorization need in your workflow, you should prefer the system privilege over the role option. For those system privileges that replace legacy role options, as listed above (such as `VIEWACTIVITY`), if both the system privilege and its legacy role option are specified for a user/role, the system privilege will take precedence and the legacy role option will be ignored.
> ```

"the system privilege will take precedence and the legacy role option will be ignored."

correction: for the ones where both the system privilege and the legacy role option exist, _either_ the system privilege or the role option are sufficient to receive that capability.
___
*[`v22.2/security-reference/authorization.md` line 197 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIESXVZ7bcsU4s8Ip3r:-NIESXVZ7bcsU4s8Ip3s:bkww6s5) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L197)):*
> ```Markdown
> ### Inheriting system privileges
> 
> System privileges can be assigned with the `GRANT OPTION`, allowing the user assigned to further grant the same privilege to other users. This inheritance of privileges allows for, say, a lead developer or chief database architect to be given a specific set of permissions which they can then assign to their team as needed, without needing to involve the originating privilege granter. In practice, this often means that a company's security and compliance team can standardize on specific role-based access privileges by team and issue that set of privileges only once, to a designed team lead or manager, who can then freely issue those same permissions across that team without involving central security.
> ```

this paragraph is conflating two different concepts.

- "inheriting" is when user A is a member of role B. if role B has a system privilege, then user A also has that privilege.
- the GRANT OPTION is just about which privileges a user can grant to a different user. Users granted a privilege with WITH GRANT OPTION can in turn grant that privilege to others. this is NOT the same as "inheriting"

side note: the GRANT OPTION itself also does get inherited. so in the example I wrote a few sentences ago, if role B has `VIEWCLUSTERSETTING WITH GRANT OPTION`, then user A also has `VIEWCLUSTERSETTING WITH GRANT OPTION`
___
*[`v22.2/security-reference/authorization.md` line 199 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIETVXo6fJQlakhg5Or:-NIETVXo6fJQlakhg5Os:bgrz3jr) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L199)):*
> ```Markdown
> System privileges can be assigned with the `GRANT OPTION`, allowing the user assigned to further grant the same privilege to other users. This inheritance of privileges allows for, say, a lead developer or chief database architect to be given a specific set of permissions which they can then assign to their team as needed, without needing to involve the originating privilege granter. In practice, this often means that a company's security and compliance team can standardize on specific role-based access privileges by team and issue that set of privileges only once, to a designed team lead or manager, who can then freely issue those same permissions across that team without involving central security.
> 
> To issue inheritable system privileges, use the `GRANT OPTION` to the [`GRANT`](../grant.html) statement, like so:
> ```

i would just remove this sentence and example. grant options are already explained on https://www.cockroachlabs.com/docs/stable/grant.html so I don't think we need to re-explain here

andf-crl (commented):
Thank you so much, you filled in many missing pieces. Primarily: my conflation of the two inheritances, indeed! I've updated language everywhere based on your comments -- please let me know if I missed anything.

Docs is currently struggling with build issues given the SSO migration, so I don't have an updated stage version for you. I will try to troubleshoot tomorrow as well. If you don't mind a docs-as-code review, though, the docs-as-code is ready for another review pass!

Thank you again for your patience and help!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15718#-:-NIEnLpL-aZIEFpLz-cD:bkux39g)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @kathancox, @rafiss, and @rmloveland)

___
*[`v22.2/security-reference/authorization.md` line 15 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEPvjF1E6k3tF604vY:-NIEZW3u04kabZZtUtiY:b-5g1d70) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L15)):*

i don't think we need to mention that they are getting removed, since we don't actually have plans to do that.
</blockquote></details>

Thank you, I've removed this, but:

- Other engineers gave us different directions: DOC-5505 // https://github.com/cockroachdb/cockroach/pull/86918 and DOC-5558 // https://github.com/cockroachdb/cockroach/pull/87188 both indicate near-future support removal, the former explicitly naming v23.1 as the removal target.
- Other docs work on this topic was already approved by PM and eng alike with "removed in a future version" language (example: https://github.com/cockroachdb/docs/pull/15393/files). You're the first to say otherwise in fact.
___
*[`v22.2/security-reference/authorization.md` line 22 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEPT364-F3uvZzzXHe:-NIEaW0CDwX_j85ExS6C:b-eg4e91) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L22)):*

"Inheritiable with the GRANT OPTION" doesn't seem right to me. System privileges are always inherited no matter what. The GRANT OPTION just says if a user has a system privilege, that they can in turn grant that privilege to other users. the GRANT OPTION is a flag that's defined for each privilege that a user has. Also, the GRANT OPTION exists for normal privileges as well as system privileges.

the `#inheriting-system-privileges` section needs updating too
</blockquote></details>

Thank you: conflating is exactly the right word here: I have untangled the two concepts; let me know how close I got. I dare say the `#inheriting-system-privileges` section needs _removing_ instead!
___
*[`v22.2/security-reference/authorization.md` line 23 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEQ8Av58BdnLbnp9H5:-NIEbGL84NDX0yQJfpQU:b4nphl4) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L23)):*

"explicit per user/role" doesn't feel like the most clear way to phrase this. what about "... are not inherited from parent roles."
</blockquote></details>

Removed this in favor of "cluster-wide" per above comment, thanks!
___
*[`v22.2/security-reference/authorization.md` line 25 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIEQO610oZ1v9BDNa2Q:-NIEbi_i8gHVZxzb9Egq:b-h439n7) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L25)):*

I don't quite follow this sentence: "The intention is that role options should be reserved for user administration permissions, while system privileges should be used to define database actions such as reads and writes."

minimal correction: instead of "define" it should say "control" or "restrict"

but I think we can just change this to say something like "If a system privilege exists with the same name as a role option, the system privilege should be used. Some role options do not have a corresponding system privilege, since they configure per-user attributes"
</blockquote></details>

This is mostly verbatim from Richard: https://github.com/RichardJCai/cockroach/blob/1b39be615caaddc9ec48e84fdbb49fd59d8b181a/docs/RFCS/20220426_rbac_extensions_system_privileges.md?plain=1#L29

I've updated to use your language in this -- and the other two -- locations it applies. Thank you!
___
*[`v22.2/security-reference/authorization.md` line 179 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIERspD9rr_6LCD0STy:-NIEcONr03C3geUEYtFu:bqbjnu0) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L179)):*

correction: `MODIFYCLUSTERSETTING` does replace the old role option
</blockquote></details>

Thank you -- you are correct!
___
*[`v22.2/security-reference/authorization.md` line 193 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIESKVfCk1Lp34_E_UD:-NIEhUC2-mn2479nO3U-:bfs3mkc) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L193)):*

"the system privilege will take precedence and the legacy role option will be ignored."

correction: for the ones where both the system privilege and the legacy role option exist, _either_ the system privilege or the role option are sufficient to receive that capability.
</blockquote></details>

I believe the intent from the PM I heard this from is: if the end user had previously been assigned `NOVIEWACTIVITY` the role option and is now assigned `VIEWACTIVITY` the system privilege. It was passed to me that the system priv would override in this case: the user previously barred from `VIEWING` the `ACTIVITY` suddenly can. Is this not true? 

Either way, you're right that I can reword this for clarity: I will, based on your response.
___
*[`v22.2/security-reference/authorization.md` line 197 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIESXVZ7bcsU4s8Ip3r:-NIEd9OFDlBYt4oOHwlM:bybhski) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L197)):*

this paragraph is conflating two different concepts.

- "inheriting" is when user A is a member of role B. if role B has a system privilege, then user A also has that privilege.
- the GRANT OPTION is just about which privileges a user can grant to a different user. Users granted a privilege with WITH GRANT OPTION can in turn grant that privilege to others. this is NOT the same as "inheriting"

side note: the GRANT OPTION itself also does get inherited. so in the example I wrote a few sentences ago, if role B has `VIEWCLUSTERSETTING WITH GRANT OPTION`, then user A also has `VIEWCLUSTERSETTING WITH GRANT OPTION`
</blockquote></details>

Yes indeed. Conflating is exactly the right word. I've untangled the two, let me know how close I got.
___
*[`v22.2/security-reference/authorization.md` line 199 at r1](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIETVXo6fJQlakhg5Or:-NIEcfIY1kty36F6Z-n_:b-s0ix7h) ([raw file](https://github.com/cockroachdb/docs/blob/d5a83b478b645665d84de8d7047889d295f85d22/v22.2/security-reference/authorization.md#L199)):*

i would just remove this sentence and example. grant options are already explained on https://www.cockroachlabs.com/docs/stable/grant.html so I don't think we need to re-explain here
</blockquote></details>

Actually, I've removed this whole section: under impression "inheritance is differentiating feature, so doc it explicitly" but was chasing the wrong interpretation of "inheritance". Pointing to `GRANT` is definitely the right call for this. Thank you!

andf-crl:
Thanks to @nickvigilante, I've now also fixed the build issues: staging is now up-to-date with the requested changes. Thanks Nick!

andf-crl:
One more update: think I've found a better way to more closely tie sys privs with privs in general, now that I understand this all better. Let me know what you think of the changes (sys privs --> system-level privs throughout, and moved all system-level priv content to within privs header). Thanks again!

andf-crl:
@rafiss mind taking another look? I think I've addressed all your comments -- thank you!

rafiss (changes requested):
this looks pretty good now! my only bit of feedback is about the notion of "inheriting to objects" which i think should be corrected.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/docs/15718#-:-NIf5DfN0VwGB9fZGiQc:b-3p56ch)* status: :shipit: complete! 0 of 0 LGTMs obtained (waiting on @andf-crl, @kathancox, and @rmloveland)

___
*[`v22.2/security-reference/authorization.md` line 22 at r3](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIf3oVR8wEIJvGylHcj:-NIf3oVR8wEIJvGylHck:b-jn165j) ([raw file](https://github.com/cockroachdb/docs/blob/4d1c0653d037fdf91b19ef8ed07de3ae7e8d765d/v22.2/security-reference/authorization.md#L22)):*
> A system-level privilege is granted at the cluster level, and is inherited by all objects within the cluster (such as databases, tables, or types).

I think we could reword this a bit: the system-level privileges don't really apply to objects in the cluster. they instead control actions that the user can do on the cluster (e.g. view cluster settings, or create changefeeds)

Instead this should probably say that system-level privileges are inherited via role membership, similar to how object-level privileges are inherited
___
*[`v22.2/security-reference/authorization.md` line 23 at r3](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIf4JVS16QEDpJZnUH9:-NIf4JVS16QEDpJZnUHA:b-ajbrst) ([raw file](https://github.com/cockroachdb/docs/blob/4d1c0653d037fdf91b19ef8ed07de3ae7e8d765d/v22.2/security-reference/authorization.md#L23)):*
> Apply only to a specific object (such as database, table, or type). Role options are not inheritable by other objects.

instead of this, i would say "apply only to a specific users" and maybe also "are not inheritable via role membership"
___
*[`v22.2/security-reference/authorization.md` line 143 at r3](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIf4YTe5pICUzLMyBim:-NIf4YTe5pICUzLMyBin:b-d3jz4z) ([raw file](https://github.com/cockroachdb/docs/blob/4d1c0653d037fdf91b19ef8ed07de3ae7e8d765d/v22.2/security-reference/authorization.md#L143)):*
> ```Markdown
> <span class="version-tag">New in v22.2:</span> System-level privileges offer more granular control over a user's actions when working with CockroachDB, over the legacy [role options authorization model](#role-options).
> 
> System-level privileges are a special kind of privilege that apply cluster-wide, meaning that the privilege is inherited to all objects within a cluster, such as databases, tables, and types.
> ```

instead of "the privilege is inherited to all objects within a cluster, such as databases, tables, and types" i would say "the privilege is not tied to any specific object in the database"
___
*[`v22.2/security-reference/authorization.md` line 173 at r3](https://reviewable.io/reviews/cockroachdb/docs/15718#-NIf52NhCvxF6K7cG35_:-NIf52NhCvxF6K7cG35a:bjavbms) ([raw file](https://github.com/cockroachdb/docs/blob/4d1c0653d037fdf91b19ef8ed07de3ae7e8d765d/v22.2/security-reference/authorization.md#L173)):*
> ```Markdown
> Take the following points into consideration while granting privileges to roles and users:
> 
> - When a role or user is granted privileges for a database, that role or user is not automatically granted access to any new or existing objects within that database. To change access to those objects, see [Default privileges](#default-privileges). This does not apply to system-level privileges, which apply cluster-wide, and are therefore automatically inherited by all objects within the cluster, whether existing or newly-created.
> ```

i would remove ", and are therefore automatically inherited by all objects within the cluster, whether existing or newly-created."

andf-crl:
@rafiss Thanks! PTAL for the latest changes!

andf-crl:
@rafiss and TY for all the help fixing it up!

@rmloveland & @kathancox please take a quick look to see if everything makes sense to you (either of you). I'll update your PRs to match language used in this PR (per PR desc) once LGTMd. Thanks so much team!

v22.2/security-reference/authorization.md
```diff
@@ -119,15 +136,43 @@ Roles and users can be granted the following privileges:
 
 {% include {{ page.version.version }}/sql/privileges.md %}
 
+### System-level privileges 
+
+<span class="version-tag">New in v22.2:</span> System-level privileges offer more granular control over a user's actions when working with CockroachDB, over the legacy [role options authorization model](#role-options).
```
kathancox, line 141: The two clauses in this sentence read a little disconnected — possibly because of the doubling up of "over" with two slightly different meanings. Maybe, slightly rephrase to something like:

System-level privileges offer more granular control over a user's actions when working with CockroachDB, compared to the legacy role options authorization model.

v22.2/security-reference/authorization.md
```diff
@@ -119,15 +136,43 @@ Roles and users can be granted the following privileges:
...
+`VIEWACTIVITYREDACTED`      | Yes: the `VIEWACTIVITYREDACTED` and `NOVIEWACTIVITYREDACTED` role options
+`VIEWCLUSTERSETTING`        | Yes: the `VIEWCLUSTERSETTING` and `NOVIEWCLUSTERSETTING` role options
+`CANCELQUERY`               | Yes: the `CANCELQUERY` and `NOCANCELQUERY` role options
+`NOSQLLOGIN`                | Yes: the `SQLLOGIN` and `NOSQLLOGIN` role options
+`VIEWCLUSTERMETADATA`       | No, new in v22.2
+`VIEWDEBUG`                 | No, new in v22.2
+`BACKUP`                    | No, new in v22.2
+`RESTORE`                   | No, new in v22.2
+`EXTERNALIOIMPLICITACCESS`  | No, new in v22.2
+`CHANGEFEED`                | No, new in v22.2
```
kathancox, line 163: Yes, the `CHANGEFEED` system-level priv is new in v22.2, but there is a `CONTROLCHANGEFEED` role option. (Different name, so maybe even bringing this up is a moot point.) When I worked on the PR for the changefeed privileges, we didn't talk about it as "replacing" the role option, but perhaps it is, like the other role options mentioned here. Let me know if you need me to check on that, or whether we can just define that here.

kathancox (approved):
One nit and one question, but otherwise LGTM!

v22.2/security-reference/authorization.md
andf-crl: Thanks! Good notice! My guess is that `CHANGEFEED` is indeed intended to replace `CONTROLCHANGEFEED`: please do check in with your PM to be sure! I'll merge this as is, but happy to add in what you find after the fact. Thanks!

andf-crl:
@kathancox TYFR!

PR #15752, opened 2022-12-08
Removed ref to diff option not applying to TBI cluster setting

kathancox:
Fixes DOC-4527

This sentence is no longer applicable and removed from the relevant versions.


jayshrivastava (approved):

andf-crl (approved):
LGTM!

PR #15788, opened 2022-12-09
Properly document the keyboard shortcuts for the SQL shell.

### LINKED ISSUE #15787: Explain to mac users how to enable alt-based shortcuts
See discussion here: https://github.com/cockroachdb/cockroach/issues/93353

On macOs terminal applications, the handling of the "Alt" key is not enabled by default.
However, doing so is rather simple. We should document it.

knz:
Fixes #15787.

NB: this is applicable to all versions as far as v2.x. Once the review converges we can copy the change to the previous versions.

NB2: v23.1 will need an update, not doing this here because the v23.1 directory has not yet been created in the doc repo.

cc @rmloveland @nickvigilante


nickvigilante (approved):
Looks awesome! Thanks for the contribution! I'd wait for @rmloveland to review though.

rmloveland (approved):
LGTM!

@knz this is glorious! thank you

knz:
so you mean you're also OK with me backporting this change to the v22.1 and v21.2 doc trees?

rmloveland:
> so you mean you're also OK with me backporting this change to the v22.1 and v21.2 doc trees?

yep!  assuming they're getting the new SQL shell experience (which you would know better than me :-} )

knz:
This PR is about the shortcuts that were in effect ever since crdb v2.0.
For the new stuff, we need to work on #15822 instead.

PR #15799, opened 2022-12-12
Added docs for the Azure cluster setting for additional buffers

kathancox:
Fixes [DOC-6005](https://cockroachlabs.atlassian.net/browse/DOC-6005), [DOC-6015](https://cockroachlabs.atlassian.net/browse/DOC-6015), [DOC-6026](https://cockroachlabs.atlassian.net/browse/DOC-6026), [DOC-6003](https://cockroachlabs.atlassian.net/browse/DOC-6003)

Added docs in the Performance section of the `BACKUP` docs for the Azure buffer cluster setting.

Introduced a new subsection in Performance related to the two cluster settings we already had called out. 

(Future work to consider Performance its own page and build out the backup-related settings here with further descriptions/tuning information over what the Cluster Settings page holds. Audit settings for backups on cluster page, consider something along the lines of: https://www.cockroachlabs.com/docs/stable/advanced-changefeed-configuration.html for backups)

### DIFF
--- a/v22.1/backup.md
+++ b/v22.1/backup.md
@@ -140,7 +140,16 @@ This improves performance by decreasing the likelihood that the `BACKUP` will be
 
 {% include_cached new-in.html version="v22.1" %} A backup job will [pause](pause-job.html) instead of entering a `failed` state if it continues to encounter transient errors once it has retried a maximum number of times. Once the backup has paused, you can either [resume](resume-job.html) or [cancel](cancel-job.html) it.
 
-{% include {{ page.version.version }}/backups/file-size-setting.md %}
+### Backup performance configuration
+
+Cluster settings provide a means to tune a CockroachDB cluster. The following cluster settings are helpful for configuring backup files and performance:
+
+Setting    | Description
+-----------+-----------------
+`bulkio.backup.file_size` | Set a target for the amount of backup data written to each backup file.<br><br>**Default:** `128 MiB`
+`cloudstorage.azure.concurrent_upload_buffers` | Improve the speed of backups to Azure Storage by increasing `cloudstorage.azure.concurrent_upload_buffers` to `3`. This setting configures the number of concurrent buffers that are used during file uploads to Azure Storage. It is important to note that the higher this setting the more data that is held in memory, which can increase the risk of OOMs if there is not sufficient memory on each node.<br><br>**Default:** `1`
+
+For a complete list, including all cluster settings related to backups, see the [Cluster Settings](cluster-settings.html) page.
 
 ## Viewing and controlling backups jobs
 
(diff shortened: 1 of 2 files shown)


kathancox:
Hey @adityamaru, friendly ping on this docs review!

v22.1/backup.md
```diff
@@ -140,7 +140,16 @@ This improves performance by decreasing the likelihood that the `BACKUP` will be
...
 {% include_cached new-in.html version="v22.1" %} A backup job will [pause](pause-job.html) instead of entering a `failed` state if it continues to encounter transient errors once it has retried a maximum number of times. Once the backup has paused, you can either [resume](resume-job.html) or [cancel](cancel-job.html) it.
 
-{% include {{ page.version.version }}/backups/file-size-setting.md %}
+### Backup performance configuration
+
+Cluster settings provide a means to tune a CockroachDB cluster. The following cluster settings are helpful for configuring backup files and performance:
+
+Setting    | Description
+-----------+-----------------
+`bulkio.backup.file_size` | Set a target for the amount of backup data written to each backup file.<br><br>**Default:** `128 MiB`
```
adityamaru, line 149: Should we warn users against setting this value too low? If they set it too low then it will result in many small SSTs being created as part of the backup, which in turn could result in degraded restore performance because of the need to keep track of so many small files.

adityamaru: Maybe we should also highlight that this is the maximum target size that we strive to hit, but it is possible that we create files of a smaller size than this for various reasons.

adityamaru (changes requested):

v22.1/backup.md
kathancox: Yes, definitely worthwhile. Thanks for providing the information here for that!

kathancox: @adityamaru I'd appreciate another quick review! I also changed up the formatting (sorry if that distracts!)

adityamaru (approved):
Thanks and sorry for the delay!

v22.1/backup.md
```diff
@@ -140,7 +140,25 @@ This improves performance by decreasing the likelihood that the `BACKUP` will be
...
-{% include {{ page.version.version }}/backups/file-size-setting.md %}
+### Backup performance configuration
+
+Cluster settings provide a means to tune a CockroachDB cluster. The following cluster settings are helpful for configuring backup files and performance:
+
+#### `bulkio.backup.file_size` 
+
+Set a target for the amount of backup data written to each backup file. This is the maximum target size the backup will reach, but it is possible files of a smaller size are created during the backup job.
+
+If `bulkio.backup.file_size` is set too low, it will cause the backup job to create many small SST files, which could impact a restore job's performance because it will need to keep track of so many small files.
```
andf-crl, line 151: Question: do we have an idea of what constitutes "too low"? Guess: "too low for your deployment's average SST size", but that's difficult to both advise of and track on the end user's side. If there's no guidance from engineering on the value of this tunable, ignore me and leave this as is.

andf-crl (approved):
LGTM, just one optional question. Thanks!

v22.1/backup.md
kathancox: I have adapted this after discussing with Aditya. We want to refer to the lowering from the default here, which I have changed. TFTR!

PR #15802, opened 2022-12-12
disable forecasted stats

taroface:
DOC-5501
DOC-5493


michae2 (approved):
LGTM! Thank you!

v22.2/cost-based-optimizer.md
```diff
@@ -159,6 +157,47 @@ sql_stats_automatic_collection_min_stale_rows = 2000);
...
+    id INT PRIMARY KEY,
+    balance DECIMAL);
+
+ALTER TABLE accounts
+SET (sql_stats_forecasts_enabled = false);
+~~~
+
+The current table settings are shown in the `WITH` clause output of `SHOW CREATE TABLE`:
+
+~~~ sql
```
kathancox, line 186: Nit: You can remove the sql syntax highlighting from the output block here. And, the columns are slightly off, small adjustment needed.

kathancox: (I also think there is a output block like this earlier on the page that you could take the opportunity to edit!) 😺

kathancox (approved):
One tiny nit, but otherwise LGTM!

PR #15807, opened 2022-12-13
Update recommended TimeoutStopSec to 300 seconds

mikeCRL:
Fixes DOC-6376


mdlinville (approved):
LGTM. I pushed a commit into this branch to add a note about the updated recommendation in the 22.2 and 22.1 docs. My thinking is that admins doing new deployments might decide to update the `systemd` configuration on their existing clusters if they read the note. Of course feel free to remove that commit if you don't think it's helpful.

shannonbradshaw:
Any reason we can't just merge this? I know Mike's out, but it doesn't make a lot of sense to wait IMO.

shannonbradshaw:
Unless we're still waiting on field input.

mdlinville:
I'll go ahead and merge. I was just waiting on Mike since I added some content on top of his.

PR #15812, opened 2022-12-14
document EXPLAIN parameters

taroface:
DOC-5235

- Document `FULL SCAN (SOFT LIMIT)` in the `EXPLAIN` output.
- Added output parameter documentation to to the `EXPLAIN` doc (it was previously only in the `EXPLAIN ANALYZE` doc). I am not sure if I have included all of the possible `EXPLAIN` output parameters. @michae2 can you verify?
- Also added some output parameters that weren't previously documented. I made a best guess on some of the definitions, which may be wrong.


michae2 (approved):
This isn't every possible property but the ones listed here look correct.

v22.2/explain.md
```diff
@@ -52,11 +52,29 @@ A successful `EXPLAIN` statement returns a table with the following details in t
 
 Detail | Description
 -----------|-------------
-Global properties | Properties that apply to the entire query plan. Global properties include `distribution` and `vectorized`.
-Statement plan tree properties | A tree representation of the hierarchy of the statement plan.
-index recommendations: N | Number of index recommendations followed by a list of index actions and SQL statements to perform the actions.
+[Global properties](#global-properties) | The properties and statistics that apply to the entire statement plan. Global properties include `distribution` and `vectorized`.
+[Statement plan tree properties](#statement-plan-tree-properties) | A tree representation of the hierarchy of the statement plan.
+Node details | The properties, columns, and ordering details for the current statement plan node in the tree.
+index recommendations | Number of index recommendations followed by a list of index actions and SQL statements to perform the actions.
```
kathancox, line 58: Is there a reason why this one is lowercased on "index"?

kathancox (approved):
LGTM! One tiny, possible nit

v22.2/explain.md
taroface: TFTR! It's because the field is lowercase in the output. I've added code formatting to the parameters to make this clearer.

PR #15820, opened 2022-12-14
Update SELECT FOR UPDATE re: serialization errors

rmloveland:
Fixes DOC-6257

Summary of changes:

- Add a paragraph to the "SELECT FOR UPDATE" overview that notes that the locking provided by SFU does not guarantee ordering in *all* cases and can result in serializable errors (40001)

  - We also make sure the magic "retry error" strings appear for searchability: specifically, `40001` and `restart transaction`.

- h/t @kai-niemi for the language here, which is adapted from his work in an internal design document (hence no link, otherwise we would add it)

### DIFF
--- a/_includes/v22.2/sql/select-for-update-overview.md
+++ b/_includes/v22.2/sql/select-for-update-overview.md
@@ -6,6 +6,8 @@ Because this queueing happens during the read operation, the [thrashing](https:/
 
 As a result, using `SELECT FOR UPDATE` leads to increased throughput and decreased tail latency for contended operations.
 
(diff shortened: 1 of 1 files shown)


nvanbenschoten (approved):

rmloveland:
TFTR @nvanbenschoten !

FYI @ianjevans I'm gonna merge this since it's so tiny and it's Breather Week (tm) and also the docs issue is a bit urgent due to a customer ask vs. I'm out on vacation for 2 weeks after today.  I don't think anything here is objectionable but feel free to yell at me if you disagree

rmloveland:
@nvanbenschoten :

Is this docs change valid for the `SELECT FOR UPDATE` docs for v21.2 and v22.1 as well?

If so,  I will create another PR to add this change to those versions

nvanbenschoten:
@rmloveland yes, these docs changes are valid for both of those versions.

PR #15825, opened 2022-12-14
JSON subscripting operators and examples

taroface:
DOC-4652

@otan I have updated the example output in `JSONB.md` to reflect the latest version. However, is it an issue that both column names are either `?column?` or `user_profile` (per the example)?


otan (approved):
looks correct to me (it matches postgres)

kathancox (approved):
LGTM Ryan!

PR #15828, opened 2022-12-14
DROP INDEX behavior in hash-sharded indexes

taroface:
DOC-3491

- Document `DROP INDEX ... CASCADE` for dropping `STORED` columns created by hash-sharded indexes
- Update v21.2-v22.2 docs (added some basic info on hash-sharded indexes that appeared to be missing in v21.2; not sure if fully correct)

It looks like the `CASCADE` requirement is only in v22.2, which means the guidance will differ in both v22.1 and v21.2. Let me know if this isn't correct.


taroface:
@Xiang-Gu @ajwerner friendly ping on review!

Xiang-Gu (approved):
LGTM! Thanks for doing this

kathancox (commented):
This LGTM Ryan, I was initially confused that v22.1 & v22.2 are described differently having read the release note, but I see you pointed that out in the PR description.

PR #15849, opened 2022-12-16
Exclude release note contributors from Vale

nickvigilante:
This is a better way of configuring Vale to exclude any div block with the release-note-contributors class added.

I believe the current Vale errors are a result of the Vale action using the config that's in master and not in this PR, so they can be ignored. I tested this locally and it did what I expected it to do.

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -4,8 +4,10 @@ MinAlertLevel = suggestion
 
 Vocab = CockroachDB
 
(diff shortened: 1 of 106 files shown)


andf-crl (approved):
Nice, LGTM! Explicitly defining a div class and then `IgnoredClasses`ing it seems like the way! Did not know this existed! Great solution.

PR #15868, opened 2022-12-20
CRDB-18740: Add CSS to handle new table anchors for cluster settings

nickvigilante:
Part of CRDB-18740

This CSS will allow us to use the `.anchored` class and add page anchors with the hoverable link button.


andf-crl (approved):
LGTM, thanks so much for this effort! Can't wait to roll this out docs-side!

PR #15869, opened 2022-12-20
disable revision history before running DMS migration

taroface:
Instruct users to disable backup revision history before running DMS (and re-enable after migration completes).


otan (approved):

v21.2/aws-dms.md
```diff
@@ -24,6 +24,9 @@ Complete the following items before starting this tutorial:
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
 - Ensure you have a secure, publicly available CockroachDB cluster running v21.2.16 or later.
+- If you intend to migrate to a {{ site.data.products.db }} cluster, you must first **disable revision history** for cluster backups.
```
kathancox, line 27: Optional: Might be helpful to include a link to this page take-backups-with-revision-history-and-restore-from-a-point-in-time.html

kathancox (approved):
LGTM, one optional thing!

v21.2/aws-dms.md
awoods187, line 27: DMS lets you both one-time migrate all data or replicate ongoing changes. This guidance seems fine for the one-time migration but I wonder about the ongoing replication as that could last weeks depending upon your migration rules. We should check with @otan but I believe that we need both the initial migration and ongoing replication to be complete. If so, we may want to specify. On the other hand, maybe the language as it is currently worded is sufficient since it says before and after you migrate.

awoods187 (approved):
Overall LGTM

v21.2/aws-dms.md
otan: Hmm actually, now that you mention it I would say this tip only applies if you are using replication as a part of the migration strategy.

> we need both the initial migration and ongoing replication to be complete

Yeah it'd need to be ongoing.

taroface: Do you mean that the tip to disable revision history only applies when someone is using replication on an ongoing basis, and not when they are using a replication task for a one-time migration?

otan: not quite - on AWS DMS, if a users select "migrate existing data and ongoing replication" or "ongoing replication" only then they need this. if they have only "migrate existing data" then they don't need this

taroface: OK, I've attempted to clarify this. PTAL!

otan (approved):
you bloody legend

taroface:
Thanks to the build breaking, I just noticed that `ALTER BACKUP SCHEDULE` is v22.2 only, and CC isn't on v22.2 until first week of Jan. Unless there is an alternate syntax we can readily document, I think it's best to wait until then to merge this. Or if anyone disagrees, I can amend the recommendation here to "contact Support" and then update it once CC is on v22.2.

otan (approved):

PR #15872, opened 2022-12-21
Add markdown="1" to all release note contributor divs

nickvigilante:
It seems that adding the div block stopped allowing markdown text inside the div from rendering. This should fix that. An accompanying PR to fix the `release-notes.py` script will follow.


andf-crl (approved):
LGTM, thanks for quick fix!

PR #15883, opened 2022-12-22
chore: trigger netlify deployment

angel-carvajal:
Chore:
- Upgrade Jquery to latest version using CDN

Fix:
- Deprecated methods
- Deprecated syntax


_includes/head.html
```diff
@@ -63,7 +63,7 @@
   }
 </script>
 
-<script src="{{ 'js/jquery.min.js' | relative_url }}"></script>
```
nickvigilante, line 66: Can we delete the `js/jquery.min.js` file?

nickvigilante (approved):
One comment, but LGTM otherwise.

_includes/head.html
angel-carvajal: yup, let me delete it

PR #15884, opened 2022-12-22
update serverless -> dedicated migration steps

taroface:
DOC-2622

- Update serverless -> dedicated migration steps to use `EXPORT` rather than local storage
- Fix some misformatted links elsewhere

### DIFF
--- a/cockroachcloud/migrate-from-serverless-to-dedicated.md
+++ b/cockroachcloud/migrate-from-serverless-to-dedicated.md
@@ -77,7 +77,7 @@ First, upload your {{ site.data.products.serverless }} data to a cloud storage l
     ~~~
 
     {{site.data.alerts.callout_success}}
-    For more information about using cloud storage with CockroachDB, see [Use Cloud Storage for Bulk Operations](use-cloud-storage-for-bulk-operations.html).
+    For more information about using cloud storage with CockroachDB, see [Use Cloud Storage for Bulk Operations](../{{site.current_cloud_version}}/use-cloud-storage-for-bulk-operations.html).
     {{site.data.alerts.end}}
 
 ## Step 2. Import the CSV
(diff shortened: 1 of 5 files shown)


otan (commented):
hmm, does BACKUP and RESTORE not work for serverless -> dedicated?

awoods187 (approved):
LGTM 

I'd check with @andy-kimball and  @amruss  to make sure serverless and export are positioned correctly

andy-kimball:
Seems pretty tough to do a database migration this way - table by table? Is that really our best answer to this problem? Like @otan, I'd expect BACKUP/RESTORE to work and be much faster/easier.

taroface:
I conferred with @awoods187 and have created a [ticket](https://cockroachlabs.atlassian.net/browse/DOC-6461) to track refactoring this doc with BACKUP/RESTORE (as well as bring in the latest migration recommendations).

This PR is only meant to correct the outdated info that `EXPORT` couldn't be used with serverless clusters.

otan:
> This PR is only meant to correct the outdated info that EXPORT couldn't be used with serverless clusters.

but you're using export here? do you mean export into local files doesn't work with serverless clusters?

taroface:
> > This PR is only meant to correct the outdated info that EXPORT couldn't be used with serverless clusters.
> 
> 
> 
> but you're using export here? do you mean export into local files doesn't work with serverless clusters?

Sorry, yeah. Previously this doc recommended export to local files only because it was presumably written when EXPORT from serverless directly to cloud storage was still not possible. The update simplifies the existing steps so that they are using EXPORT direct to S3 instead of exporting to local files and then uploading to S3.

otan (approved):
this is fine for now i guess, but yeah BACKUP/RESTORE seems much better

kathancox (approved):
LGTM!

PR #15913, opened 2023-01-03
Update SELECT FOR UPDATE re: serialization errors

rmloveland:
Followup work for DOC-6257

This commit ports the changes to the SELECT FOR UPDATE docs that were added in cockroachdb/docs#15820 to the v21.2 and v22.1 docs as well, since those versions are still supported.


rmloveland:
FYI @ianjevans these changes have already had technical review in #15820 but happy to make any content changes as needed (to this PR as well as the already-merged v22.2 text, since they're the same)

ianjevans (approved):

PR #15914, opened 2023-01-03
Potential change to SQL option display

kathancox:
Fixes DOC-6435

This is a exploratory PR to figure out if there is a better way to display SQL option and statement query parameters rather than in a table. Potentially, this is only applicable to the `CREATE CHANGEFEED` SQL statement that has many options and query parameters, which often have dense, complex descriptions. (One of the motivations behind this is that several of these options really require even more detailed explanation.)

Pros:
- More "breathing space" for descriptions and better examples without moving that content to another page.
- Each option has a heading that we (and a user) can easily grab a link to, compare to the anchor links we create.
- Avoid the awkward different font when using bullets in a table.
- Some would argue that this kind of information isn't what a table should be used for.

Cons: 
- The long body text is a lot to scroll through.

In the PR, the Query parameter section has the addition of a bulleted list of the parameters. This is to test whether such a list would help users locate the parameters with a quick click. Left this bulleted list off of the Options section to see which is preferable.

### Current status
Screenshot 2023-01-03 at 3 08 22 PM

### Potential status
Screenshot 2023-01-03 at 2 51 53 PM


kathancox:
Hey @mikeCRL, Here is a style-related PR that it would be good to get your review/thoughts on, along with the style team. Thanks!

kathancox:
Closing this PR, although I think this is something we still need to consider, this PR is so far out of date that it would be quicker to redo rather than try to resolve all of the merge conflicts from v23.1 work

PR #15944, opened 2023-01-08
Fix a typo in Multi-region Capabilities Overview page 

MantasMiksys:
The docs say that database's primary region is defined in Step 1 while actually it is Step 2.


nickvigilante:
@MantasMiksys Thank you for your contribution! We recently moved our default branch from `master` to `main` as well as restructured our docs content. I went ahead and reapplied the changes in this PR to #17210.

PR #15982, opened 2023-01-11
DOC-5851: Remove old client-side redirects

nickvigilante:
Fixes DOC-5851


mikeCRL (approved):
[Checked all 65 changed pages](https://docs.google.com/spreadsheets/d/1qBmuQIjSsFQr-Nt-8GopEe_DXmbpDkEdZ98icSIPQvQ/edit#gid=0) and reached the intended destination in all cases. Thanks for the help preparing that list and a test plan. LGTM!

PR #15994, opened 2023-01-12
Fix deprecated SASS syntax to help reduce Netlify build log size

nickvigilante:


andf-crl (approved):
LGTM (with one nit on comment placement), but I really don't know SASS ([SAS](https://www.sas.com/en_us/home.html) on the other hand 😂  ) . Logically, makes sense to me as far as I understand the logic though! Hope this helps!

PR #16001, opened 2023-01-13
Added incremental location specifics to locality-aware backups page

kathancox:
Fixes DOC-3265

Added a short example and sentence to note that the `incremental_location` option requires the same number of URIs as the full backup location.

Also, added a sentence that touches on [#79135](https://github.com/cockroachdb/cockroach/pull/79135) because it didn't seem like that was covered here.

Other miscellaneous cleanup.

## Preview

https://deploy-preview-16001--cockroachdb-docs.netlify.app/docs/stable/take-and-restore-locality-aware-backups.html#create-an-incremental-locality-aware-backup


v22.2/take-and-restore-locality-aware-backups.md
```diff
@@ -117,29 +117,43 @@ To restore from a specific backup, use [`RESTORE FROM {subdirectory} IN ...`](re
...
+	~~~
+
+	To view the available subdirectories, use [`SHOW BACKUPS`](restore.html#view-the-backup-subdirectories).
+
+- If you are creating an incremental backup using the [`incremental_location`](backup.html#options) option to send your incremental backups to a different location, you must include the same number of locality-aware URIs for the full backup destination and the `incremental_location` option:
+
+	{% include_cached copy-clipboard.html %}
+	~~~ sql
+	BACKUP INTO LATEST IN
+		('s3://us-east-bucket?COCKROACH_LOCALITY=default', 's3://us-west-bucket?COCKROACH_LOCALITY=region%3Dus-west') WITH incremental_location = ('s3://us-east-bucket-2?COCKROACH_LOCALITY=default', 's3://us-west-bucket-2?COCKROACH_LOCALITY=region%3Dus-west');
```
msbutler, line 153: it might be worth mentioning this current limitation: https://github.com/cockroachdb/cockroach/issues/82912

msbutler (approved):
looks good to me!

v22.2/take-and-restore-locality-aware-backups.md
kathancox: Ah, yes, that's a good idea. I will add that in. Thank you!

v22.1/take-and-restore-locality-aware-backups.md
```diff
@@ -117,29 +117,43 @@ To restore from a specific backup, use [`RESTORE FROM {subdirectory} IN ...`](re
...
-And if you want to explicitly control where your incremental backups go, specify the subdirectory in the `BACKUP` statement:
+	{{site.data.alerts.callout_info}}
+	When [restoring from an incremental locality-aware backup](#restore-from-an-incremental-locality-aware-backup), you need to include **every** locality ever used, even if it was only used once.
 
-{% include_cached copy-clipboard.html %}
-~~~ sql
-> BACKUP INTO {subdirectory} IN
-		('s3://us-east-bucket?COCKROACH_LOCALITY=default', 's3://us-west-bucket?COCKROACH_LOCALITY=region%3Dus-west');
-~~~
+	It is necessary that the same localities be included for every incremental backup in the series of backups; however, only the `default` locality is required.
```
andf-crl, line 135: nit: think we can do without this sentence in this reshuffle: seems to be adequately conveyed just above in line 133, and might be confusing with "only the default is required" seeming to counter "the same localities be included". In reality, for a loc-aware incremental restore: both are required for the thing to succeed. "required" for `default` only means that the code does some input/param validation for this value. The restore will still fail, even if `default` is included to pass input validation, because the 101st incremental I ran in this series used `us-east` once and I forgot to include it here. Just a thought!

v22.1/take-and-restore-locality-aware-backups.md
```diff
@@ -117,29 +117,43 @@ To restore from a specific backup, use [`RESTORE FROM {subdirectory} IN ...`](re
...
+	{{site.data.alerts.end}}
+
+- If you want to explicitly control the subdirectory for your incremental backups:
```
andf-crl, line 138: nit: Ideally all three bullet points introduced as "there is different syntax depending:" are introduced in the same way. I'm a big fan of the direct approach you used in the first one:

> - To append your incremental backup to the full backup in the [`incrementals` directory]:
> - To explicitly control the subdirectory for your incremental backup:
> - To append your incremental backup to the full backup using the [`incremental_location`] option to send your ...

andf-crl (approved):
Nice, LGTM with two optional language nits. Thanks!

v22.1/take-and-restore-locality-aware-backups.md
kathancox: I actually think (and this was probably past me who wrote this) the first sentence in this note block is referring to the restore: every locality ever used must be included in the restore statement. Then, the sentence referring to `default` is for actually taking the backup... that is the `default` is required. If that's the case, then it's poor note construction. I'm going to double check this and rearrange as appropriate. But, totally see your point.

kathancox: Sorted, thanks!

v22.1/take-and-restore-locality-aware-backups.md
kathancox: Done.

v22.1/take-and-restore-locality-aware-backups.md
andf-crl: thanks, looks good, nice work sorting!!

PR #16015, opened 2023-01-17
DOC-5851: Undo noindex in the docs

nickvigilante:
Fixes DOC-5851

I've added comments in the event we want to put it back.


gemma-shay (approved):
LGTM with the caveat that I'm not super familiar with how this page works, but seems like a low-risk change

PR #16033, opened 2023-01-19
Release notes v22.2.3

kathancox:
Fixes DOC-4612


_includes/releases/v22.2/v22.2.3.md
```diff
@@ -0,0 +1,53 @@
...
+
+<h3 id="v22-2-3-bug-fixes">Bug fixes</h3>
+
+- In the PostgreSQL extended protocol mode, it was possible for auto-commits to not execute logic for DDL statements, when certain DML (insert/update/delete) and DDL statements were combined in an implicit transaction. Auto-commits are now disabled inside the planner if any DDL statements were executed earlier. [#93717][#93717]
+- Fixed a panic that occurred when using a SQL cursor to access tables in the [`crdb_internal`](../v22.2/crdb-internal.html) schema. [#94443][#94443]
+- Fixed a bug that caused incorrect selectivity estimation for queries with ORed predicates all referencing a common single table. [#94664][#94664]
+- Fixed a bug in join queries that involves tables with unique constraints using `LIMIT`, `GROUP BY`, and `ORDER BY` clauses to ensure the [optimizer](../v22.2/cost-based-optimizer.html) considers streaming group-by with no TopK operation when possible. This is often the most efficient query plan. [#94603][#94603]
+- Fixed a bug where certain [`GRANT`](../v22.2/grant.html) or [`REVOKE`](../v22.2/revoke.html) commands on a user that does not exist would error with the correct PG code. [#94735][#94735]
```
andf-crl, line 23: @mikeCRL @nickvigilante This release notes dupes one under SQL Language Changes, both from [PR #94735](https://github.com/cockroachdb/cockroach/pull/94735). Here's the relevant PR desc:

> Release note (bug fix): Previously, certain GRANT or REVOKE commands on a user which does not exist would error with the correct PG code. This is now fixed.
> Release note (sql change): Error messages for missing users sometimes had different forms. This is now unified as role/user "user" does not exist.

We can either have the automation discard dupes like this, or teach engineering not to do this.

But for the here and now it looks like another manual adjustment: deleting one of them.

andf-crl (approved):
LGTM, just one dupe note. Thanks!

_includes/releases/v22.2/v22.2.3.md
nickvigilante: I think this would require follow-up with Oliver on what his rationale was for splitting the RN. I'm not fully sure I understand what each does, but I think they're different. The first may be an issue with running GRANT and REVOKE commands on users that don't exist. The second may be that various "missing user" error messages not strictly related to "GRANT" or "REVOKE" are now unified in a single format.

andf-crl: Right: but would you want to do that each time a dupe message appears: follow up with eng to ask why? This happens often enough (think I've now seen at least 5 of these) that automation may wish to consider how to handle dupe RNs in terms of the ongoing process.

Just noting it here such that it is considered for the future.

nickvigilante: If by dupe you mean multiple release notes within a single commit, then I have been told that is a valid commit message submission on behalf of Engineering. I had to build logic into the product change issue script to handle those situations and create two separate GH issues for those commits.

kathancox: fyi, I'm checking in with Oliver to clarify this case.

kathancox: In this case from Oliver, it seems we should maybe include both as the error message format _and_ the error code.

PR #16036, opened 2023-01-19
feat: improve sidebar blocking time

angel-carvajal:
Feats
- Create new json files to simplify the creation of the html of the main level (done)
- Loading sub sections on demand (done)
- Integrate navgoco with solution (still fixing it, currently only expand and collapse works properly on main level, sublevels only expand works)
- Lazy load all the logic behind the sidebar so we reduce the blocking time (done)

Fix
- Beautify json files

Chore
- Add couple of files to improve js styling


_includes/sidebar.html
```diff
@@ -5,8 +5,22 @@
...
+      {% for option in page.sidebar_data_titles %}
+        <li class="tier-1">
+          {% if option.isExternal %}
+            <a href="{{option.url}}">{{option.title}}</a>
+          {% elsif option.url %}
+            {% assign currentVersion = page.dir | replace: "/", "" %}
+            {% if currentVersion == site.versions.stable %}
+              {% assign currentVersion = "stable" %}
+            {% endif %}
+            <a href="{{site.url}}/docs{{option.url | replace: "VERSION", currentVersion }}">{{option.title}}</a>
```
nickvigilante, line 18:
```suggestion
            <a href="{{site.url}}{{site.baseurl}}{{option.url | replace: "VERSION", currentVersion }}">{{option.title}}</a>
```

nickvigilante (changes requested):
One change, and this should be ready for merge. Please let me know when you're ready for me to merge.

_includes/sidebar.html
angel-carvajal: fixed, still one pending issue

PR #16039, opened 2023-01-19
Include note that key in value opt is required for changefeeds on cloud storage external connections

kathancox:
Fixes DOC-6633

Added note to relevant pages that `key_in_value` is necessary when you run a changefeed to an external connection that represents a cloud storage sink.

## Preview

https://deploy-preview-16039--cockroachdb-docs.netlify.app/docs/stable/create-external-connection.html#changefeed-sinks-as-external-connections


HonoreDB (approved):

andf-crl (approved):
LGTM!

PR #16016, opened 2023-01-17
DOC-6601 TA C20230118

andf-crl:
Addresses: DOC-6601

- Adds TA C20230118
- Moves entire TA section to new left-nav loc: **Releases > Technical Advisories** to somewhat suggest applicability to both other same-level entries in this loc: CC and CRDB.
- Adjusts TA index page (listing) to now also reference new CC-applicability for TAs.
- Minor language adjustments

Staging

- [advisories/index.md](https://deploy-preview-16016--cockroachdb-docs.netlify.app/docs/advisories/index.html)

PR #16051, opened 2023-01-20
DOC-6643 fix links missed by builder

andf-crl:
Addresses: DOC-6643

- Fixes links to non-CC content that 404d. These links were not caught as broken by the builder, as https://github.com/cockroachdb/docs/pull/16016 passed all checks for build & merge - spawning separate investigation in D O C - 6 6 4 4 to troubleshoot.

Staging

WIP


andf-crl:
All green! But do I trust it .... ?

PR #16054, opened 2023-01-20
Add headings before the filter selectors

ianjevans:
This PR is part of the OKR to improve the connection docs. A quick fix is to make the content filters on the connection topic more usable by adding headings.

It also includes a necessary fix to our Javascript file for handling embedded filter groups. Up until now, our filters assumed that each filter group was stand-alone. E.g. you have an OS filter group for Mac, Linux, and Windows, and a deployment filter group for Serverless, Dedicated, and Self-Hosted, and each filter group should exist independent of the other.

The problem in this topic is that the driver/ORM filter groups are nested below the language filter group. E.g. if the user selects Sequelize at first (under the Javascript/Typescript filter), and then selects Java, the URL parameters kept Sequelize as an active filter even though the parent language changed.

With the changes in this PR, we check if filter groups are defined within the scope of another filter, and if so, we store the parent scope of the filter group. If the user changes a parent filter scope, we also remove the child scope from the existing URL parameters.


kathancox:
Cool! I don't know if this would be better without visually seeing it, but what about a sentence before the filters and numbering the selectors (though practically you can select in any order, I suppose):

> To connect to your CockroachDB cluster, select the following:
> 
> 1. Your deployment
> [filters]
> 2. Your language
> [filters]
> 3. Your ORM or driver

That might be too fussy! But, it came to mind. I like your change though!

nickvigilante (approved):

_includes/v22.2/connect/connection-url.md
```diff
@@ -0,0 +1,19 @@
+<div class="filter-content" markdown="1" data-scope="mac linux">
+Set a `DATABASE_URL` environment variable to your connection string.
+
+{% include_cached copy-clipboard.html %}
+~~~ shell
+export DATABASE_URL="{connection string}"
+~~~
```
shannonbradshaw, line 7: Setting the env variable at the top of the connect section is a nice improvement!

Did you implement it by using a placeholder and an include to simplify what edits would be required for this revision (as opposed to using the connection string template as the value of DATABASE_URL in each data-scope)?

ianjevans: Yes. The actual connection string will differ based on the deployment, so I added info about how to find the connection string earlier, when the user selects their deployment.

The vast majority of the examples can use a connection string (exceptions are Java, which uses a JDBC connection string, and Django, which requires parameters). So we have includes for general connection strings and JDBC connection strings, and don't need anything else for Django.

gemma-shay (approved):
lgtm! thanks for doing this

emily-horing (approved):
LGTM!

_includes/v22.2/connect/connection-url.md
shannonbradshaw: Thanks for wrapping this up @ianjevans! LGTM

shannonbradshaw: @ianjevans apologies for the delayed follow up. I didn't see your response because I had notifications going to the wrong email address. Fixed now, but sorry to keep you waiting. I see this was LGTMed and merged. Great!

PR #16055, opened 2023-01-20
Clarify that you must modify the general connection string with Sequelize.

ianjevans:
This PR clarifies the format of connection strings when using SQLAlchemy.

Closes DOC-6587.


css/bs/bootstrap-grid.css
```diff
@@ -0,0 +1,57 @@
+/*
```
nickvigilante, line 1: Can you give these files an extension of `.scss` instead and delete the error comments?

nickvigilante (changes requested):
If you're in the office this week, let me know and we can work on this together.

gemma-shay:
I can see in the convo history that you were working on this, but something weird is still going on with the code blocks. 
Screenshot 2023-02-02 at 3 22 56 PM

ianjevans:
> I can see in the convo history that you were working on this, but something weird is still going on with the code blocks. Screenshot 2023-02-02 at 3 22 56 PM

Yes, the CSS change is now in #16174, which ideally we would merge before this PR so I can pull it in.

gemma-shay (approved):
gotcha, this lgtm!

PR #16057, opened 2023-01-20
Add detail to remove region for dedicated

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-6123

Add requested details to remove region functionality for dedicated clusters

### DIFF
--- a/cockroachcloud/cluster-management.md
+++ b/cockroachcloud/cluster-management.md
@@ -137,6 +137,8 @@ You can add or remove up to nine regions at a time through the Console. Note tha
 
 ### Remove a region from your cluster
 
+When you remove a region from your cluster, the node in that region with the highest ordinal will be decommissioned first. Any ranges on that node will be up-replicated to other nodes, and once decommission is complete that node will be shut down. This process is then repeated for every other node in the region. To remove a region from your cluster:
+
 1. Navigate to the cluster's **Overview** page.
 1. Select **Actions > Edit cluster**.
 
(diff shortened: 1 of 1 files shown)


gemma-shay:
@jramos-crl this info seems a little excessive to me for cloud, but it was requested. lmk what you think

jramos-crl (approved):
@gemma-shay agreed but seems like there have been some questions about it so hopefully, the extra context helps. This looks good to me!

cockroachcloud/cluster-management.md
```diff
@@ -137,6 +137,8 @@ You can add or remove up to nine regions at a time through the Console. Note tha
 
 ### Remove a region from your cluster
 
+When you remove a region from your cluster, the node in that region with the highest ordinal will be decommissioned first. Any ranges on that node will be up-replicated to other nodes, and once decommission is complete that node will be shut down. This process is then repeated for every other node in the region. To remove a region from your cluster:
```
rmloveland, line 140: suggest making 'decommissioned' a link to https://www.cockroachlabs.com/docs/v22.2/node-shutdown?filters=decommission#decommission-the-node to offer more detail, if wanted

'up-replicated' could be a link to these Replication Dashboard docs that show up-replication after decommission: https://www.cockroachlabs.com/docs/stable/ui-replication-dashboard.html#snapshot-data-received 

finally the fact that we can remove a region implies more than one region, which implies it is a multi-region cluster.  Suggest a slight modification to the text to say something like "when you remove a region from a [multi-region cluster](https://www.cockroachlabs.com/docs/cockroachcloud/plan-your-cluster.html#multi-region-clusters), ..." since due to links and search, a user can land directly on this sub-heading of the page without seeing the context at beginning of parent section

rmloveland (approved):
LGTM

PR #16067, opened 2023-01-23
Updated the envelope option in changefeed docs for missing row & detail

kathancox:
Fixes DOC-6656

Added `row` to the `CREATE CHANGEFEED` enterprise docs as this was missing, but a viable option. It seems that this is only supported by Kafka and sinkless changefeeds.

Furthermore, test `row` with `EXPERIMENTAL CHANGEFEED FOR` and `wrapped` is also the default for core changefeeds. The docs say that `row` is the default... this PR updates that.

## Preview

Enterprise: https://deploy-preview-16067--cockroachdb-docs.netlify.app/docs/v22.2/create-changefeed#options

Core: https://deploy-preview-16067--cockroachdb-docs.netlify.app/docs/v22.2/changefeed-for#options


HonoreDB (approved):
Technical details LGTM. I think "metadata" might be clearer than "top-level" but other reviewers might disagree.

PR #16076, opened 2023-01-24
[DOC-6275] rename bulk operations cloud storage page

ashleykwtam:
Addresses DOC-6275:
- removes "bulk ops" from page name and fixes affected changes on other pages
- change for v22.2, v23.1

Preview:
- [Use Cloud Storage 23.1](https://deploy-preview-16076--cockroachdb-docs.netlify.app/docs/dev/use-cloud-storage.html)
- [Use Cloud Storage 22.2](https://deploy-preview-16076--cockroachdb-docs.netlify.app/docs/stable/use-cloud-storage.html)


_redirects
```diff
@@ -377,6 +377,7 @@ layout: null
 /docs/cockroachcloud/planning-your-serverless-cluster /docs/cockroachcloud/learn-about-pricing.html 301
 /docs/cockroachcloud/migrate-from-free-to-dedicated /docs/cockroachcloud/migrate-from-serverless-to-dedicated.html 301
 /docs/cockroachcloud/free-faqs /docs/cockroachcloud/serverless-faqs.html 301
+/docs/v22.2/use-cloud-storage-for-bulk-operations.html /docs/v22.2/use-cloud-storage.html 301
```
nickvigilante, line 380:
```suggestion
/docs/v22.2/use-cloud-storage-for-bulk-operations.html /docs/v22.2/use-cloud-storage.html 301
/docs/dev/use-cloud-storage-for-bulk-operations.html /docs/dev/use-cloud-storage.html 301
/docs/stable/use-cloud-storage-for-bulk-operations.html /docs/stable/use-cloud-storage.html 301
```

nickvigilante: You need to make sure `dev` and `stable` aliases are accounted for in setting redirects.

nickvigilante (changes requested):

_redirects
ashleykwtam: added, thanks Nick!

nickvigilante (approved):

kathancox:
@nickvigilante We want to update the name of this page... currently it is "Use Cloud Storage for Bulk Operations". Ashley is only updating v22.2, but ideally we still want the version dropdown to direct users to the previous versions without the "page does not exist" callout. Is there a way to do that if you change the page name for one version (but all the content is otherwise the same, or representative of the feature for said version)?

kathancox:
@ashleykwtam We'll also want to remove "Bulk Operations" from the nave "File Storage for Bulk Operations"... this can just be "File Storage" as it contains all the file storage options.

livlobo (approved):
Left nav still says "bulk", but othrerwise lgtm. thx!

ashleykwtam:
@kathancox added some changes and implemented it for 23.1 too!

kathancox:
Hey @ashleykwtam, Thanks for adding to v23.1!

To remediate the version dropdown issue where the title name change means that you can't select another version of the page, could you try adding `key` to the frontmatter at the top of the page for the new version pages where the title has changed?

So, in v22.2 and v23.1 add:
`key: use-cloud-storage-for-bulk-operations.html`

See the following image as an example:
image

I had to test this out a bit when I added, but I *think* what I mention above should work. Let me know if not, or you have questions!

kathancox:
> Hey @ashleykwtam, Thanks for adding to v23.1!
> 
> To remediate the version dropdown issue where the title name change means that you can't select another version of the page, could you try adding `key` to the frontmatter at the top of the page for the new version pages where the title has changed?
> 
> So, in v22.2 and v23.1 add: `key: use-cloud-storage-for-bulk-operations.html`
> 
> See the following image as an example: image
> 
> I had to test this out a bit when I added, but I _think_ what I mention above should work. Let me know if not, or you have questions!

Oh, looks like you did this for v22.2, but not v23.1... is that correct? It still doesn't seem to work on the preview. Could you add to v23.1 and then see if you can move between versions in your edited local build?

kathancox (approved):

PR #16088, opened 2023-01-25
remove outdated delete statistics guidance

taroface:
[DOC-5347](https://cockroachlabs.atlassian.net/browse/DOC-5347)

### DIFF
--- a/_includes/v22.2/misc/delete-statistics.md
+++ b/_includes/v22.2/misc/delete-statistics.md
@@ -2,16 +2,14 @@ To delete statistics for all tables in all databases:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> DELETE FROM system.table_statistics WHERE true;
+DELETE FROM system.table_statistics WHERE true;
 ~~~
 
 To delete a named set of statistics (e.g, one named "users_stats"), run a query like the following:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> DELETE FROM system.table_statistics WHERE name = 'users_stats';
+DELETE FROM system.table_statistics WHERE name = 'users_stats';
 ~~~
-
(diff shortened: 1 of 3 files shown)


_includes/v22.2/misc/delete-statistics.md
```diff
@@ -2,16 +2,14 @@ To delete statistics for all tables in all databases:
...
 
 To delete a named set of statistics (e.g, one named "users_stats"), run a query like the following:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> DELETE FROM system.table_statistics WHERE name = 'users_stats';
+DELETE FROM system.table_statistics WHERE name = 'users_stats';
 ~~~
 
-After deleting statistics, restart the nodes in your cluster to clear the statistics caches.
```
mgartner, line 15: cc @rytaft this advice is out-of-date, correct?

mgartner (approved):
LGTM

_includes/v22.2/misc/delete-statistics.md
rytaft: Not that I know of -- do you know what change fixed this? Is it due to the fact that we're using rangefeeds instead of Gossip?

mgartner: Correct. With https://github.com/cockroachdb/cockroach/pull/62563 a deletion of a stats row will trigger the range feed handle in `stats_cache.go`, and the timestamp of the delete rangefeed event will differ from the cache's `lastTS`:

https://github.com/cockroachdb/cockroach/blob/3bc0992225673c2e8c80288a5c619d78f8b928cd/pkg/sql/stats/stats_cache.go#L155

So a stats refresh should be triggered.

rytaft: Makes sense - thanks!

kathancox (approved):
LGTM!

PR #16092, opened 2023-01-25
Update vale.yml

nickvigilante:
Change filter mode to added to limit the number of results


andf-crl (approved):
Nice!  Let's check it out!

PR #16095, opened 2023-01-25
Add Linux flag to base64 command PubSub

kathancox:
Fixes DOC-5788

### DIFF
--- a/v22.1/changefeed-examples.md
+++ b/v22.1/changefeed-examples.md
@@ -392,7 +392,9 @@ You'll need access to a [Google Cloud Project](https://cloud.google.com/resource
     gcloud iam service-accounts keys create key.json --iam-account=cdc-demo@cockroach-project.iam.gserviceaccount.com
     ~~~
 
-    Next, base64 encode your credentials key:
+    Next, base64 encode your credentials key using the command specific to your platform.
+
+    If you're working on macOS:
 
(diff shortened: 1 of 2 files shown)


v22.1/changefeed-examples.md
```diff
@@ -399,6 +399,13 @@ You'll need access to a [Google Cloud Project](https://cloud.google.com/resource
     cat key.json | base64
     ~~~
 
+    If you're working on Linux, run the following to ensure that lines are not wrapped in the output:
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ shell
+    cat key.json | base64 -w 0
+    ~~~
+
```
andf-crl, line 408: nit: I see this more of an either/or. The Linux command doesn't work on Mac, and the Mac command results in output that breaks the `CREATE CHANGEFEED...` command on Linux. I think we can safely avoid by making this something like:

> Next, base64 encode your credentials key using the following command specific to your platform:
>
> If you're working on Linux, run:
>    ~~~ shell
>   cat key.json | base64 -w 0
>    ~~~
>
> If you're working on macOS, run:
>    ~~~ shell
>   cat key.json | base64
>    ~~~

Otherwise LGTM!

andf-crl (approved):
LGTM just a nit regarding either/or approach here. Thanks! Quick work!

PR #16097, opened 2023-01-25
Update PG compat page to mention UDFs

rmloveland:
Fixes DOC-6631


ianjevans (approved):

rmloveland:
@mgoddard FYI since you filed the issue about the UDF support being listed on the PG compat page.  Should be live on the site in a few minutes, thanks for filing that issue!

PR #16098, opened 2023-01-25
Statement and Transactions pages correction

gemma-shay:
Statement and Transactions pages incorrectly list 10mil rows as default limit for persisted stats, 1 mil is correct.
https://cockroachlabs.atlassian.net/browse/DOC-5893


xinhaoz (approved):
Thanks for the correction!

xinhaoz (approved):

PR #16112, opened 2023-01-25
Add style guide rule about using "we"

lnhsingh:
Closes DOC-2708

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -67,6 +67,11 @@ Other general guidance about language and tone:
 
     **Example:** Now that you have a database, user, and a table, run the following code to insert rows into the table.
 
+- Do not use the personal pronoun “we.” Instead, use:
+
+    - “CockroachDB” for when you are talking about something the _product_ does or supports.
+    - “Cockroach Labs” for when you are talking about something “we” (as the group of people developing CockroachDB) support or recommend.
+
 - For [tutorials and examples](#tutorials-and-examples), we recommend you use the second-person point of view (e.g., you). These docs should be more casual and conversational, as if they are teaching the user, but still straightforward and clear.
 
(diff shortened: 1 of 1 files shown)


StyleGuide.md
```diff
@@ -67,6 +67,11 @@ Other general guidance about language and tone:
 
     **Example:** Now that you have a database, user, and a table, run the following code to insert rows into the table.
 
+- Do not use the personal pronoun “we.” Instead, use:
+
+    - “CockroachDB” for when you are talking about something the _product_ does or supports.
+    - “Cockroach Labs” for when you are talking about something “we” (as the group of people developing CockroachDB) support or recommend.
```
MichaelTrestman, line 73: Hmm, i think when it would be tempting to use "we" is when you'd have to repeat CRDB or CRL 6 times in a couple sentences, like "Cockroach Labs keeps careful track of x, y and z. Cockroach Labs also monitors 1, 2 and 3. Cockroach Labs' approach to yada yada..."
Can we not say "we" when it's obvious who "we" would be, as per normal English usage?
I'm not questioning that we (docs) shouldn't just use the royal we willy nilly, I agree that is obnoxious and potentially unclear if there isn't a clear pronoun anchor and the reader is just supposed to assume that 'we' are the company/product/whoever.

MichaelTrestman: i'm just not clear what the motivation would be for avoiding the clearly anchored usage.

lnhsingh: If we want to convey that we've actually tried things and are recommending as a company and you'll run into a problem.
Or if it's "strongly recommend" then it's CRL
Try to use "we recommend" more sparingly. Only use if it's a recommendation from eng or the field.

lnhsingh: Ok, updated the recommendation for using "we" based on the team discussion yesterday. @MichaelTrestman, can PTAL?

PR #16113, opened 2023-01-25
Guidance on shortened product names

MichaelTrestman:
Completes **Docs Style Fixit 1/25/23** issue https://cockroachlabs.atlassian.net/browse/DOC-2075


lnhsingh (commented):
This LGTM. Can you add a follow up Jira issue for thinking through / updating the variables to use "CockroachDB {{..}}" for the first instance on a page and just the variable for everything following?

PR #16116, opened 2023-01-25
DOC-6510 correct disk ops include values

andf-crl:
Addresses: DOC-6510

- Corrected include that referred to disk ops counter in milliseconds, as it is a raw numeric increment.
	- The modified include is included in two places: verified that this change is correct to make for both.

Thanks Mark for the great feedback!!

Staging

[ui-hardware-dashboard.html](https://deploy-preview-16116--cockroachdb-docs.netlify.app/docs/stable/ui-hardware-dashboard.html#disk-ops-in-progress)


world2mark (approved):

world2mark (approved):

andf-crl:
Thanks so much @world2mark !

PR #16117, opened 2023-01-25
Style fix

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-6425

"CockroachDB Dedicated-managed service account" is technically incorrect because service accounts are created for your CockroachDB Cloud organization, not Serverless or Dedicated clusters specifically. Removed hyphen and changed to "CockroachDB Cloud service account."

### DIFF
--- a/cockroachcloud/export-logs.md
+++ b/cockroachcloud/export-logs.md
@@ -281,7 +281,7 @@ Perform the following steps to enable log export from your {{ site.data.products
 
 	1. In the GCP console, visit the [IAM admin page](https://console.cloud.google.com/iam-admin) for your project.
  1. Click the **+ Grant Access** button.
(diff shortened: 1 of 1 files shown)


andf-crl (approved):
Definitely approve of this change, but this language came from here: https://www.cockroachlabs.com/docs/cockroachcloud/cmek-ops-gcp.html#step-1-collect-the-required-information.

I was told to mirror.

Since you're changing it here, it would make sense to also add where it came from to this PR.

But happy to pre-approve with an LGTM, thanks for making this change!

gemma-shay:
I think the GCP project is actually unique to each Dedicated cluster, so that one makes more sense. Are you ok with leaving that hyphen in? It's a little awkward but does make sense.

andf-crl:
Sure then! You absolutely know this better than I, approved!

PR #16119, opened 2023-01-25
add warning about orphaned org migration abort scenario

MichaelTrestman:


cockroachcloud/configure-cloud-org-sso.md
```diff
@@ -48,6 +48,24 @@ Before you enable Cloud Organization SSO, notify your members about what to expe
 
 During enablement of the feature, a list of affected members is shown, and those members are also notified individually.
 
+### Ensure at least one of your org admins is an original org member
+
+{{site.data.alerts.callout_success}}
+Troubleshooting tip: If your migration fails obscurely, confirm that at least one org admin belongs to no other CockroachDB Cloud Organization.
+{{site.data.alerts.end}}
+
+<!--  @jason-crl I know this isn't quite right, plz help me phrase this stuff more accurately -->
```
MichaelTrestman, line 57: @jason-crl 
I know that the 'identity being split' is not quite right... there aren't two identities, just one accessed through the two domains. but I can't remember the way you described it...

jason-crl: #### ------
> Under the hood, when an Organization migrates to Cloud Organization SSO from basic SSO, the identity of each member must be migrated to the new domain. If an org members *also belongs to other CockroachDB Cloud Organizations*, their identity must be split, to create an identity on the new org at the new SSO domain, while maintaining their identity on their other orgs at the basic SSO shared domain. This is all done automatically by the platform, and requires no action by customer org administrators.

Overall this is correct.

#### ------
> However, this aspect of the migration can cause problems in a situation where *all members with the admin role joined the org through invitation*. Theoretically, a migration that succeeded *except for the migration of the admin identities* could lead to the organization being unrecoverably orphaned without admin access.

`all members with the admin role joined the org through invitation` isn't quite right.  It's not about admins joining through invitation, but rather admins being members of multiple orgs.  Unclear if we need to explain the negative case here, since our Enablement flow won't ever let this happen (ie that's why the check exists).

#### ------
> To prevent this unlikely but serious failure scenario, the migration will instead abort. However, there is currently no error message that will indicate the cause of the failure.

We could indicate which general error message appears: `Cloud Organization SSO cannot be enabled`

#### ------
> If no human administrator has an available SSO account that is not already tied to another org, you must create a new admin user without SSO (authenticating with username/password) to perform the migration. This powerful but weakly secured admin user should be deleted after the migration, in keeping with general access-management hygiene.

Trying out different wording to improve clarity:
_If every single admin member belongs to multiple orgs, the Enablement cannot proceed.  In this case, you must create a temporary admin user (authenticating with username/password would be easiest) to perform the migration.  This temporary admin user should then be deleted after Cloud Org SSO is enabled._

#### ------
Overall though, I worry about diving into too much detail around the internals of identity splitting and auth domains (`shared domain`).  Trying to consider how much the customer actually needs to know.  Personally I would opt for a more general treatment, something like this for the intro.  Then adding any necessary bits from above.  cc'ing @abhinavg6 for his perspective.

_When Cloud Organization SSO is enabled for an Organization, certain members will need to be re-added to the Organization.  This occurs for members that belong to multiple CockroachDB Cloud Organizations, as their membership in other organizations interferes with their transition from Basic SSO to Cloud Org SSO.  Affected members simply need to be re-added to the Org, once Cloud Org SSO is enabled._

jason-crl (changes requested):

cockroachcloud/configure-cloud-org-sso.md
```diff
@@ -48,6 +48,24 @@ Before you enable Cloud Organization SSO, notify your members about what to expe
 
 During enablement of the feature, a list of affected members is shown, and those members are also notified individually.
 
+### Ensure at least one of your org admins is an original org member
```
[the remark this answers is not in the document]
MichaelTrestman: whoops, forgot to change the title after we talked.

cockroachcloud/configure-cloud-org-sso.md
```diff
@@ -48,6 +48,24 @@ Before you enable Cloud Organization SSO, notify your members about what to expe
...
+
+{{site.data.alerts.callout_success}}
+Troubleshooting tip: If your migration fails obscurely, confirm that at least one org admin belongs to no other CockroachDB Cloud Organization.
```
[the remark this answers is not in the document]
MichaelTrestman: thanks!

cockroachcloud/configure-cloud-org-sso.md
MichaelTrestman: Except this isn't true anymore. They dont' need to be re-added, right?

MichaelTrestman: They are auto-re-invited now?

jason-crl: That's right, the re-invite is automatic now.  Good call, "re-added" isn't the best wording here.  I was trying to focus on the end result that's visible, which is that the user gets kicked out and needs to re-join.

So maybe: "certain members will need to re-join the Organization (via automatically sent re-invites)" or something

cockroachcloud/configure-cloud-org-sso.md
```diff
@@ -48,6 +48,22 @@ Before you enable Cloud Organization SSO, notify your members about what to expe
...
 
+### Ensure that at least one organization admin belongs to no other CockroachDB Cloud Organization
+
+{{site.data.alerts.callout_success}}
+Troubleshooting tip: If your migration fails with the error: `Cloud Organization SSO cannot be enabled`, confirm that at least one organization admin belongs to no other CockroachDB Cloud Organization.
+{{site.data.alerts.end}}
+
+<!--  @jason-crl I know this isn't quite right, plz help me phrase this stuff more accurately -->
+
+Under the hood, when an Organization migrates to Cloud Organization SSO from basic SSO, the identity of each member must be migrated to the new domain. If an organization members *also belongs to other CockroachDB Cloud Organizations*, their identity must be split, to create an identity on the new organization at the new SSO domain, while maintaining their identity on their other orgs at the basic SSO shared domain. This is all done automatically by the platform, and requires no action by customer organization administrators.
```
ianjevans, line 59:
```suggestion
Under the hood, when an Organization migrates to Cloud Organization SSO from basic SSO, the identity of each member must be migrated to the new domain. If an organization member *also belongs to other CockroachDB Cloud Organizations*, their identity must be split, to create an identity on the new organization at the new SSO domain, while maintaining their identity on their other orgs at the basic SSO shared domain. This is all done automatically by the platform, and requires no action by customer organization administrators.
```

ianjevans: FWIW, I agree with @jason-crl that these are implementation details that aren't actionable by users, so they should be omitted.

cockroachcloud/configure-cloud-org-sso.md
```diff
@@ -48,6 +48,22 @@ Before you enable Cloud Organization SSO, notify your members about what to expe
...
+
+However, this aspect of the migration can cause problems in a situation where *all members with the admin role belong to other orgs*. Theoretically, a migration that succeeded *except for the migration of the admin identities* could lead to the organization being unrecoverably orphaned without admin access. To prevent this unlikely but serious failure scenario, the migration will instead abort. However, there is currently no specific error message that will indicate the cause of the failure, and you will see the generic error `Cloud Organization SSO cannot be enabled`.
```
ianjevans, line 61: As above, this paragraph can be omitted in favor of the one below.

MichaelTrestman: done.

MichaelTrestman:
@jason-crl @ianjevans how does the updated copy look as far as balancing TMI with actionable?

ianjevans (approved):

jason-crl:
May need to be updated to reflect the new enablement/auto-reinvite mechanism:

In the **Communicate to your members** section:
>   - Members who are also members of other organizations must be re-added to your organization. If they sign in using an authentication method with [autoprovisioning](#autoprovisioning) enabled, they are automatically added upon successful sign-in. _Otherwise, they must be re-invited._ If they previously had the [console admin](console-access-management.html#console-admin) role, it must be granted to them again.

PR #16121, opened 2023-01-25
remove "please" from instructions

taroface:
[DOC-2070](https://cockroachlabs.atlassian.net/browse/DOC-2070)

- Removed "please" specifically for technical instructions.
- Updated style guide. I'm not sure if this wording is clear enough.


kathancox (approved):
I think that sounds good to me Ryan... as in "please" for something beyond the normal scope of work we lay out in docs (is what I gather from this).

PR #16128, opened 2023-01-26
document 22.2 session vars workaround

taroface:
[DOC-6689](https://cockroachlabs.atlassian.net/browse/DOC-6689)

This can't be merged until `copy_from_retries_enabled` is in a point release.

### DIFF
--- a/v22.1/aws-dms.md
+++ b/v22.1/aws-dms.md
@@ -24,6 +24,15 @@ Complete the following items before starting this tutorial:
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
 - Ensure you have a secure, publicly available CockroachDB cluster running v22.1.8 or later.
+    - If your CockroachDB cluster is running v22.1.14 or later, set the following session variable using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
+
+        {% include_cached copy-clipboard.html %}
+        ~~~ sql
+        ALTER ROLE {username} SET copy_from_retries_enabled = true;
+        ~~~
+
+        This prevents a potential issue when migrating especially large tables.
+
 - Manually create all schema objects in the target CockroachDB cluster. This is required in order for AWS DMS to populate data successfully.
     - If you are migrating from a PostgreSQL database, [use the **Schema Conversion Tool**](../cockroachcloud/migrations-page.html) to convert and export your schema. Ensure that any schema changes are also reflected on your PostgreSQL tables, or add [transformation rules](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.CustomizingTasks.TableMapping.SelectionTransformation.Transformations.html). If you make substantial schema changes, the AWS DMS migration may fail.
 
@@ -177,6 +186,15 @@ The `BatchApplyEnabled` setting can improve replication performance and is recom
     ~~~
 
     Try selecting **Full LOB mode** in your [task settings](#step-2-2-task-settings). If this does not resolve the error, select **Limited LOB mode** and gradually increase the **Maximum LOB size** until the error goes away. For more information about LOB (large binary object) modes, see the [AWS documentation](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.LOBSupport.html).
+
+- If you encounter an `ABORT_REASON_CLIENT_REJECT` error when migrating an especially large table, and are running v22.1.14 or later, set the following session variable using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ sql
+    ALTER ROLE {username} SET copy_from_retries_enabled = true;
+    ~~~
+
+    Then retry the migration.
 
 - Run the following query from within the target CockroachDB cluster to identify common problems with any tables that were migrated. If problems are found, explanatory messages will be returned in the `cockroach sql` shell.
 
--- a/v22.1/aws-dms.md
+++ b/v22.1/aws-dms.md
@@ -23,7 +23,7 @@ Complete the following items before starting this tutorial:
 
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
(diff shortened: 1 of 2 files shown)


v22.2/aws-dms.md
```diff
@@ -22,6 +22,18 @@ Complete the following items before starting this tutorial:
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
 - Ensure you have a secure, publicly available CockroachDB cluster running v22.1.0 or later.
+    - If your CockroachDB cluster is running v22.2.3 or later, set the following session variables using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
```
otan, line 25: * I think you'd want to make note here that you'd need to create a user for DMS that you'll use for your replication endpoint. This session variable is only going got be available for v22.2.4 or higher.
* It may be worth noting that without doing this, you may end up with a DMS failure for larger tables (in some ways it is optional). Maybe worth adding as a troubleshooting step for later as well.
* This will probably apply to v22.1 as well.

taroface: > This will probably apply to v22.1 as well at the next point release for it (looking like v22.1.14+)

Will `copy_from_atomic_enabled` be backported? I wasn't able to find info about this.

otan: ah my bad, `copy_from_atomic_enabled` already exists in v22.2.0. turning that off improves things but doesn't fix it unless retries are enabled.

`copy_from_atomic_enabled` doesn't exist in v22.1, so only the `copy_from_retries_enabled` setting is required.

thanks for dealing with the complexity around all this!

taroface: @otan Updated, PTAL. I also updated the minimum recommended version on the v22.2 doc to `v22.2.0`; I think this may have been incorrect previously?

v22.2/aws-dms.md
```diff
@@ -21,7 +21,21 @@ Complete the following items before starting this tutorial:
 
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
-- Ensure you have a secure, publicly available CockroachDB cluster running v22.1.0 or later.
+- Ensure you have a secure, publicly available CockroachDB cluster running v22.2.0 or later, and have created a [SQL user](security-reference/authorization.html#sql-users) that you can use for your AWS DMS [target endpoint](#step-1-create-a-target-endpoint-pointing-to-cockroachdb).
```
otan, line 24: Wonder if we should just set the min to .4. Do we already have a prereq of creating a unique migration user?

v22.2/aws-dms.md
```diff
@@ -176,6 +190,20 @@ The `BatchApplyEnabled` setting can improve replication performance and is recom
 
     Try selecting **Full LOB mode** in your [task settings](#step-2-2-task-settings). If this does not resolve the error, select **Limited LOB mode** and gradually increase the **Maximum LOB size** until the error goes away. For more information about LOB (large binary object) modes, see the [AWS documentation](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.LOBSupport.html).
 
+- If you encounter an `ABORT_REASON_CLIENT_REJECT` error when migrating an especially large table, and are running v22.2.4 or later, set the following session variables using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
```
otan, line 193: Replace this with TransactionProtoWithRefreshError

v22.1/aws-dms.md
```diff
@@ -24,6 +24,15 @@ Complete the following items before starting this tutorial:
 - Configure a [replication instance](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_ReplicationInstance.Creating.html) in AWS.
 - Configure a [source endpoint](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.html) in AWS pointing to your source database.
 - Ensure you have a secure, publicly available CockroachDB cluster running v22.1.8 or later.
+    - If your CockroachDB cluster is running v22.1.14 or later, set the following session variable using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
```
ajstorm, line 27: Maybe this goes without saying, but I'm wondering if we should say here that if you're migrating large tables (and define what large is), that customers should upgrade to at least 22.1.14

v22.1/aws-dms.md
```diff
@@ -178,6 +187,15 @@ The `BatchApplyEnabled` setting can improve replication performance and is recom
 
     Try selecting **Full LOB mode** in your [task settings](#step-2-2-task-settings). If this does not resolve the error, select **Limited LOB mode** and gradually increase the **Maximum LOB size** until the error goes away. For more information about LOB (large binary object) modes, see the [AWS documentation](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.LOBSupport.html).
 
+- If you encounter an `ABORT_REASON_CLIENT_REJECT` error when migrating an especially large table, and are running v22.1.14 or later, set the following session variable using [`ALTER ROLE ... SET {session variable}`](alter-role.html#set-default-session-variable-values-for-a-role):
```
ajstorm, line 190: We might want to define what we mean by large here.  Also, can we provide more information as to where we'd expect customers to see that error.  Sometimes it shows up in the detailed cloud flare logging.  Sometimes you need to dig into the CRDB logs to find it.

v22.2/aws-dms.md
ajstorm: +1 on saying that .4 is the minimum

ajstorm (commented):
Thanks for making this change Ryan.  Mostly minor comments.

v22.1/aws-dms.md
taroface: @otan Can you help me define "large" here? I guessed it was on the order of millions of rows.

awoods187:
Looks like this is coming on Feb 6 for 22.1.x and Feb 22.2.x on Feb 13--so almost time to merge

v22.1/aws-dms.md
otan: Yeah millions is a good approximation

v22.2/aws-dms.md
taroface: @otan I take it you mean `TransactionRetryWithProtoRefreshError` and used this, but let me know if I'm wrong.

v22.1/aws-dms.md
taroface: @ajstorm I've added this, but since we recommend v22.2.4 by default on 22.2 clusters, would it be simpler to recommend v22.1.14 by default on 22.1 (or is that going too far)?

otan: 22.1.14 will have the requisite changes

taroface:
Closing this in favor of #16204 and #16205, which I'll merge separately as the patch releases are available.

PR #16157, opened 2023-01-31
Cloud February Release Notes

gemma-shay:
Release notes February 2023
https://cockroachlabs.atlassian.net/browse/DOC-6838
New process of PM submission in effect instead of script


releases/cloud.md
```diff
@@ -15,6 +15,42 @@ Get future release notes emailed to you:
...
+## February 6, 2023
+
+<h3> General changes </h3>
+
+- The following features are now available for {{ site.data.products.serverless }} clusters running CockroachDB [v22.2.0](v22.2.html) or later:
+  - [Query cancellation](../{{site.current_cloud_version}}/cancel-query.html#considerations) using the PostgreSQL wire protocol (pgwire).
+  - [`EXPLAIN ANALYZE`](../{{site.versions["stable"]}}/explain-analyze.html) now gives an estimate of the [Request Units (RUs)](../cockroachcloud/learn-about-request-units.html) consumed by a query.
+  - All {{ site.data.products.serverless }} users can now use cloud storage for [`IMPORT`](../{{site.versions["stable"]}}/import.html), [`BACKUP`](../{{site.versions["stable"]}}/backup.html), and [change data capture (CDC)](../{{site.versions["stable"]}}/change-data-capture-overview.html) without entering billing information.
+  - [`SHOW RANGES`](../{{site.versions["stable"]}}/show-ranges.html) is now supported on {{ site.data.products.serverless }}.
+  - The [GC TTL](../{{site.versions["stable"]}}/row-level-ttl.html) for deleted values is lowered from 24 hours to 1.25 hours.
```
rmloveland, line 27: GC TTL refers to the zone config `gc.ttl_seconds` and the docs are at https://www.cockroachlabs.com/docs/v21.2/configure-replication-zones.html#gc-ttlseconds

Row-level TTL is something else (it deletes data for you in the background that you are no longer interested in).  I don't think we want Row-level TTL docs here, since that is something users have to configure themselves via SQL

unfortunately we call them both TTL so it's confusing!

rmloveland (approved):
LGTM except that I think you need a different link for the GC TTL (explanation in comment)

PR #16164, opened 2023-02-02
Lock jekyll-sass-converter to an older version

nickvigilante:
This is to help work around the build failures in Cloudflare.


nickvigilante:
We have newer CSS that needs to be supported. Do not merge.

PR #16192, opened 2023-02-06
Follow up on cloud org sso docs

MichaelTrestman:


jason-crl:
Thinking we should include at least a small mention about how some members may need to accept the auto re-invites, just so admins are aware some of their members might go through this.  Something like this?

_Members who are also members of other organizations may need to be re-added to your organization after Cloud Org SSO is enabled.  Affected members will be automatically sent an invite to re-join your organization._

PR #16199, opened 2023-02-07
Add Backup and Restore monitoring page to DR docs

kathancox:
Fixes DOC-6118, DOC-6646

Added a new backup and restore monitoring page to the DR docs. Some of these metrics may be surfaced elsewhere in the docs (or will be), but the point is to consolidate DR observability on this one page. This first iteration of the page includes the following:

- Available metrics (However, after a conversation with David in Slack, it seems that not _all_ available metrics in Prometheus are actually useful. As of opening the PR, I have removed the idle metric as this seems to not be applicable to backup/restore jobs.)
- Reference to relevant `SHOW` SQL statements for further investigation.
- Cloud/Self-hosted availability
- Prometheus/Datadog metric availability differences and links out to their respective setup instructions.

## Preview

https://deploy-preview-16199--cockroachdb-docs.netlify.app/docs/stable/backup-and-restore-monitoring.html


v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+See the [Monitor CockroachDB with Prometheus](monitor-cockroachdb-with-prometheus.html) tutorial for guidance on installing and setting up Prometheus and Alertmanager to track metrics.
+
+### Available metrics
+
+Metric | Description 
+-------+-------------
+`schedules_backup_succeeded` | The number of scheduled backup jobs that have succeeded.
+`schedules_backup_started` | The number of scheduled backup jobs that have started.
+`schedules_backup_last_completed_time` | The Unix timestamp of the most recently completed backup by a schedule specified as maintaining this metric.
```
dt, line 37: should we mention the parameter on the schedule that is used to specify this?

v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+docs_area: manage
+---
+
+CockroachDB includes metrics to monitor [backup](backup.html), [restore](restore.html), and [scheduled backup](create-schedule-for-backup.html) jobs. You can use monitoring integrations to alert when there are anomalies, such as backups that have failed or restore jobs encountering a retryable error.
+
+Depending on whether you are using a {{ site.data.products.dedicated }} or {{ site.data.products.core }} cluster, you can use the following to monitor backup and restore metrics for your CockroachDB cluster:
+
+- [Prometheus endpoint](#prometheus-endpoint): {{ site.data.products.dedicated }}, {{ site.data.products.core }}
+- [Datadog integration](#datadog-integration): {{ site.data.products.dedicated }}
+
```
dt, line 14: I wonder if we should specifically call out the schedules_backup_last_completed_time metric as the recommended metric to monitor, and/or mention that whichever schedule -- if you have more than one -- maintains the backups you'd use in a disaster should be configured to update this metric.

dt (approved):

v22.2/backup-and-restore-monitoring.md
kathancox: Ah, @dt this parameter: `updates_cluster_last_backup_time_metric`... we don't currently document this. Was it because if CC customers use it on their scheduled backups it could mess with the CRL-run scheduled backups in some way? If that's the case, I still think we can document it here, might just have to come with that warning. That is, if the CC issue is a thing.

livlobo (approved):
Thanks for documenting this Kathryn.

Re: backup monitoring, I am wondering if we can provide some additional guidance / tips for monitoring.

@dt  had suggested a while back that we should tell customers to monitor for "lack of successful backup" rather than "backup failure". Something along those lines would be helpful!

v22.2/backup-and-restore-monitoring.md
dt: I'll defer to @livlobo but I think we should doc it; I don't love the idea that we just hope nobody finds it as a strategy for CC , and it is generally useful to non-CC users for the same reason CC wanted it: this is _the_ thing you'd want to be monitoring -- short of having a tool go look in the buckets themselves for successful backups but that is annoying to do -- to know that you're backing up as expected.

kathancox: Gotcha I made some adjustments to see what you both think. For now, I have added in the option (see the screenshot below), can remove if @livlobo doesn't want to document. I may have gone a bit overboard on the warnings here, but it's a starting point at least, if we're going to document:
image

kathancox:
I have specifically called out the last completed metrics and monitoring for a lack of successful backup as recommendations.

v22.2/backup-and-restore-monitoring.md
livlobo: We really need to find a way to separate the managed service backup monitoring from the customer owned backup monitoring. For the problem of customers accidentally overwriting the manages service schedules, we said, we'd rely on monitoring. Now, I feel like we're exposing the monitoring to potential clobbering?

kathancox: @livlobo as part of another PR: https://github.com/cockroachdb/docs/pull/16316 on re-organizing the nav, the proposal is sections on managed service & customer-owned backups, how about this monitoring page (as it stands) lives under the customer-owned backup section? Then, this screenshot above, should actually say "For **customer-owned backups only**" ?

dt: I think we're worrying about the potential for clobbering way too much without evidence that it will happen and/or is a problem if it does happen. I don't even know if we need the warnings here. I'd just say let's document the facts of the flag: if set, the schedule will update this metric, which is a good way to monitor that backups are succeeding as expected. I don't think we need to talk one way for cloud and another elsewhere or warn people about anything.

v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,80 @@
...
+### Available metrics
+
+We recommend the following guidelines:
+
+- Use the `schedules_backup_last_completed_time` metric. Configure this metric to monitor the backup schedule you would use in a disaster.
+- Monitor for a lack of successful backups rather than failed backups. A stuck scheduled job will not increment the `schedules_backup_failed` metric. Instead, alert on the `schedules_backup_last_completed_time` metric if the timestamp has not moved forward as expected.
+
+Metric | Description 
+-------+-------------
```
andf-crl, line 39: This metrics section has multiple problems, few of which can be fixed all the way downstream here on the Docs team:

Metrics use dot notation to demarcate along metrics families, as seen [here](https://www.cockroachlabs.com/docs/stable/metrics.html) and in http://localhost:8080/_admin/v1/chartcatalog.

For example:
- `schedules_round_reschedule_skip` should be `schedules.round.reschedule-skip`
- `schedules_backup_last_completed_time` should be `schedules.BACKUP.last-completed-time`
- etc.

There is no rubric by which underscore characters may be intuited to their intended character replacement, so you'll have to go through each by hand via guess and check. I recommend http://localhost:8080/_admin/v1/chartcatalog and trying single words at a time, since a `_` could be any of `.` `-` or `_`.

Did engineering give you a resource like this one: https://github.com/cockroachlabs/managed-service/blob/master/pkg/otel/assets/metrics.yaml

If so, that uses some strange internal representation that does not match what is shown to end users in the third-party integrations named on this page.

Further, the descriptions here, while helpful, are manual and do not match the metrics descriptions shown to users in the third-party integrations named on this page, in the DB Console custom chart: http://localhost:8080/#/debug/chart (click `Select a metric ...`) and in our own: http://localhost:8080/_admin/v1/chartcatalog

I don't know what to do about this, I haven't solved this on my end either in Obs space.

@dt can you confirm the above please.

Big picture: This begins a slippery slope, where we list metrics in multiple places, where they can now get out of date at twice the volume. I don't know what to do about this either.

v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,80 @@
...
+
+To use the Datadog integration with your **{{ site.data.products.dedicated }}** cluster, you can:
+
+- Export the following schedule backup metrics to Datadog using the [Cloud API](../cockroachcloud/cloud-api.html). To set this up, see [Export Metrics From a CockroachDB Dedicated Cluster](../cockroachcloud/export-metrics.html?filters=datadog-metrics-export).
+- Access the Cloud Console **Monitoring** page to enable the integration. To set this up, see [Monitor CockroachDB Dedicated with Datadog](../cockroachcloud/monitoring-page.html#monitor-cockroachdb-dedicated-with-datadog).
+
+### Available metrics in Datadog
+
+Metric | Description 
+-------+-------------
```
andf-crl, line 69: This metrics list will also need the same adjustment.

v22.2/backup-and-restore-overview.md
```diff
@@ -71,6 +71,16 @@ For detail on additional cloud storage features CockroachDB supports:
 
 {% include {{ page.version.version }}/misc/note-egress-perimeter-cdc-backup.md %}
 
+## Backup and restore observability
+
+You can verify that your stored backups are restorable with backup validation. A successful restore completely validates a backup, however, the validation tools offer a faster alternative and return an error message if a backup is not valid. There are three "levels" of verifying backups that give increasing validation coverage depending on the amount of runtime you want to invest in validating backups.
```
andf-crl, line 76: nit: `While a successful ... a backup, the validation tool offers a faster ...`

v22.2/backup-and-restore-overview.md
```diff
@@ -71,6 +71,16 @@ For detail on additional cloud storage features CockroachDB supports:
...
+
+See the [Backup Validation](backup-validation.html) page for detail and examples.
+
+You can track backup jobs with metrics that cover scheduled backups, status of running jobs, and details on completed or failed jobs. You can alert on these metrics via the Prometheus endpoint or the Datadog integration.
```
andf-crl, line 80: nit: `with` --> `using`.

`with` suggests (backup jobs with metrics), while `using` suggests (backup jobs) using (metrics).

andf-crl (changes requested):
I left a question here for engineering. I believe that the metrics listing uses the wrong notation, and have raised a few related points regarding, not all of which are solvable by Docs.

v22.2/backup-and-restore-monitoring.md
kathancox: @andf-crl I picked them up "literally" from the status/vars endpoint hence the underscores.

- Re the descriptions, I think we wanted to be able to add detail to the metrics here (as a docs page) where we felt necessary for guidance. This feels akin to most other areas of the docs site? Is the metrics page you link to automatically pulled then? My sense is we want to move forward with this page and add nuanced guidance where needed. Let me know if I haven't captured your meaning here.

v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,80 @@
...
+toc: true
+docs_area: manage
+---
+
+CockroachDB includes metrics to monitor [backup](backup.html), [restore](restore.html), and [scheduled backup](create-schedule-for-backup.html) jobs. You can use monitoring integrations to alert when there are anomalies, such as backups that have failed or restore jobs encountering a retryable error.
+
+Depending on whether you are using a {{ site.data.products.dedicated }} or {{ site.data.products.core }} cluster, you can use the following to monitor backup and restore metrics for your CockroachDB cluster:
+
+- [Prometheus endpoint](#prometheus-endpoint): {{ site.data.products.dedicated }}, {{ site.data.products.core }}
+- [Datadog integration](#datadog-integration): {{ site.data.products.dedicated }}
```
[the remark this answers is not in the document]
kathancox: Backup metrics have not been added to Datadog on SH hence the need to outline on this page. 

🤔 Hmmm, I must have had a moment when I added dedicated to Prometheus here.

v22.2/backup-and-restore-monitoring.md
andf-crl: Ah gotcha, thank you. My understanding is that `_status/vars` is more of a consumable API rather than a notation with end users in mind. However, it is indeed what Prometheus consumes to populate its output, so it is indeed the notation that will be shown to Prometheus users in Prometheus, you're right. I withdraw that comment, then, given that this doc is essentially all about metrics surfaced through Prometheus, and the different notation that appears in the other three locations is essentially out of scope to this particular doc.

Re: descriptions: the intent is indeed to automatically populate our currently-out-of-date metrics docs reference, but I have not been able to solve the fundamental problems raised above, so am unlikely to be able to enact this in the short term. Since the short term is all any of us have control over: I withdraw my comment on descriptions as well.

Apologies: I won't get any further in the way of this PR.

andf-crl (approved):
Changing my review to: LGTM with tiny nits. Thank you Kathryn.

v22.2/backup-and-restore-monitoring.md
kathancox: @andf-crl No, no! Definitely not getting in the way, I appreciate your expertise here! To your comment, I will keep as is for now. On the second point, it would be good to stay in the loop with your intentions on the overarching metrics docs, I would generally rather that we can stay consistent across the docs. If there is a way to also add some nuance on the page, while still automating I would be into that. This is most definitely a new page, and therefore, it's very earliest iteration! As anything comes up around the metrics work you're doing, please feel free to open issues for me to update this page, or directly alter the page!

v22.2/backup-and-restore-overview.md
kathancox: Yes, much smoother! I also totally forgot to add this section to v23.1 ooooops!

v22.2/backup-and-restore-overview.md
kathancox: Done!

v22.2/backup-and-restore-monitoring.md
kathancox: As per earlier comment, leaving as-is for now! Thanks!

v22.2/backup-and-restore-monitoring.md
```diff
@@ -0,0 +1,80 @@
...
+You can access the [Prometheus endpoint](monitoring-and-alerting.html#prometheus-endpoint) (`http://<host>:<http-port>/_status/vars`) for backup and restore metrics with **{{ site.data.products.dedicated }} or {{ site.data.products.core }}** clusters.
+
+See the [Monitor CockroachDB with Prometheus](monitor-cockroachdb-with-prometheus.html) tutorial for guidance on installing and setting up Prometheus and Alertmanager to track metrics.
+
+### Available metrics
+
+We recommend the following guidelines:
+
+- Use the `schedules_backup_last_completed_time` metric. Configure this metric to monitor the backup schedule you would use in a disaster.
+- Monitor for a lack of successful backups rather than failed backups. A stuck scheduled job will not increment the `schedules_backup_failed` metric. Instead, alert on the `schedules_backup_last_completed_time` metric if the timestamp has not moved forward as expected.
```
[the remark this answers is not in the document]
kathancox: Yes, good idea! Done

PR #16212, opened 2023-02-08
Fix netlify.toml based on advice from support

nickvigilante:


mikeCRL (approved):
LGTM. Matches their [docs](https://docs.netlify.com/integrations/build-plugins/create-plugins/#inputs).

PR #16213, opened 2023-02-08
Performance fix on the sidebar

nickvigilante:
We added an SEO tweak to ensure that our sidebar produced full URLs
in our sidebar to limit the scope to which Google could crawl URLs on
our docs site. The problem is we're now using absolute URLs instead of
relative URLs in each of our Netlify builds, which are adding a hamper
on the speed of our builds. This PR should fix that.


mikeCRL (approved):
LGTM in the preview

PR #16220, opened 2023-02-08
Remove unsupported pgwire cancellation note

gemma-shay:
I think this was accidentally left over after https://cockroachlabs.atlassian.net/browse/DOC-4761 was completed


rmloveland (approved):
LGTM, thanks!

PR #16223, opened 2023-02-08
Install Lighthouse plugin

nickvigilante:


package-lock.json
```diff
@@ -0,0 +1,10688 @@
+{
```
ianjevans, line 1: I think we recently removed `package-lock.json` from the repo so it would get regenerated every time someone ran the bundler.

nickvigilante: Ack. I removed it to see how that would behave.

nickvigilante:
@ianjevans How does it look now?

ianjevans (approved):

PR #16237, opened 2023-02-10
DOC-6919 clarify ccloud service account

andf-crl:
Addresses: DOC-6919

- Clarified that the service account referenced in GCP steps is the CCloud service account - not a GCP service account.
	- Tagged most references to `service account` under the GCP tab with `CockroachDB Cloud` to clarify.
	- Added new FAQ item explicitly calling this out.

Staging

[export-logs.md](https://deploy-preview-16237--cockroachdb-docs.netlify.app/docs/cockroachcloud/export-logs.html)

### DIFF
--- a/cockroachcloud/export-logs.md
+++ b/cockroachcloud/export-logs.md
@@ -32,7 +32,7 @@ Method | Required permissions | Description
 `POST` | `ADMIN` or `EDIT` | Enables log export, or updates an existing log export configuration.
 `DELETE` | `ADMIN` | Disables log export, halting all log export to AWS CloudWatch or GCP Cloud Logging.
 
-See [Service accounts](console-access-management.html#service-accounts) for instructions on configuring a service account with these required permissions.
+See [Service accounts](console-access-management.html#service-accounts) for instructions on configuring a {{ site.data.products.db }} service account with these required permissions.
 
 ## Log name format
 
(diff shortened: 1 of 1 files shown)


cockroachcloud/export-logs.md
```diff
@@ -249,7 +249,7 @@ Perform the following steps to enable log export from your {{ site.data.products
 	1. Click on the name of your cluster.
 	1. Find your cluster ID in the URL of the single cluster overview page: `https://cockroachlabs.cloud/cluster/{your_cluster_id}/overview`.
 
-1. Determine the service account to use for exporting logs from your {{ site.data.products.dedicated }} cluster. This log export service account requires specific naming syntax to be recognized by the log export feature. This command uses the third-party JSON parsing tool [`jq`](https://stedolan.github.io/jq/download/) to isolate just the needed `id` (GCP cluster ID) and `account_id` (GCP account ID) fields, and combines them for you in the required form:
+1. Determine the {{ site.data.products.db }} service account to use for exporting logs from your {{ site.data.products.dedicated }} cluster. This log export service account requires specific naming syntax to be recognized by the log export feature. This command uses the third-party JSON parsing tool [`jq`](https://stedolan.github.io/jq/download/) to isolate just the needed `id` (GCP cluster ID) and `account_id` (GCP account ID) fields, and combines them for you in the required form:
```
dhartunian, line 252: This is not correct. The service account here is the **GCP** service account not the CockroachCloud one. Sorry for causing more confusion 😓 .

Here's what I think we should do: let's not use the phrase "service account" at all here and say "principal" instead. So something like "Determine the GCP Principal to grant permissions to from your account". The user needs the name of the thing to allow to write to their GCP logs. That thing happens to be a service account also but we can say Principal here because in step 5c we say to grant access to this Principal.

Furthermore I think under "The resulting log export service account should resemble the following example:" we should have some sort of warning or note that says: "This principal refers to a resource that is owned by Cockroach Labs and is created automatically along with your cluster" to clarify that they should **not** create that service account when they see that string.

Let me know what you think. Thanks for iterating on this.

andf-crl: Thanks @dhartunian, your suggested copy is very helpful! I might have succeeded in re-confusing myself, so please let me know if my changes get it right, or if I need to make more tweaks -- apologies if so.

Thanks so much for the help!

cockroachcloud/export-logs.md
```diff
@@ -262,13 +262,15 @@ Perform the following steps to enable log export from your {{ site.data.products
...
+    The resulting GCP principal should resemble the following example:
 
     {% include_cached copy-clipboard.html %}
     ~~~
     crl-logging-user-a1c42be2e53b@crl-prod-abc.iam.gserviceaccount.com
     ~~~
 
+    This GCP principal refers to a resource that is owned by Cockroach Labs and is created automatically along with your cluster. You **do not** need to create this account in GCP; it is already present for use by your cluster.
+
 1. Create a new role with the required permissions in your GCP project:
```
dhartunian, line 274: Do you think it's worth splitting the directions more severely between "stuff you do with CRL API" and "stuff you do with GCP". These steps mix both and I wonder if that can be clarified

dhartunian (approved):
👍🏻 thx for making these changes. I added a comment re: step groups but that can be addressed separately.

PR #16244, opened 2023-02-13
DOC-6315: Add loop in Cloud Releases docs

nickvigilante:
This will help us implement a similar system to Cloud releases as we do for Self-Hosted releases.

Fixes DOC-6315


gemma-shay:
Is this just creating individual files for every date? But they'll still all be on the same page?

nickvigilante:
@gemma-shay Correct. Essentially, you'll just throw a new release note MD file in `_includes/releases/cloud` and add a new row to `_data/cloud_releases.csv`, both of which I will automate in the managed-service `release-notes.py` script, and the main `releases/cloud.md` file will iterate through the list and render the releases in reverse chronological order.

gemma-shay:
Ohh is this so the last SHA can get picked up by the script?

nickvigilante:
Yep, exactly. The data file stores the previous SHA, so all you'll have to do is pass in a new SHA and the script will know the bounds by which it's supposed to run.

gemma-shay:
Awesome, that would be great. My only concern is that we often throw in an ad-hoc release note for anything urgent that comes up between official release notes. These are usually part the docs PR for the feature and are often feature-flagged things that the script wouldn't pick up anyway. I guess we would need to also document the process for doing this manually or just have an easy way to generate a blank file. Or maybe there's another solution... hm

gemma-shay:
for example, https://www.cockroachlabs.com/docs/releases/cloud.html#february-2-2023

nickvigilante:
I suppose for those cases, what I can do is just iterate through the list and find the last valid SHA, and if you have a new off-cycle RN you want to publish, it can just add a row to `cloud_releases.csv` with a null SHA. And then for the next full hotfix, it will search for the latest valid SHA and then use that for comparison. Would that satisfy that use case? I can also create a script in ed-tools that can handle that.

gemma-shay:
I feel like we should change this at the same time as the script so I don't get stuck during the next release notes. Can also wait til next month if you want. What do you think?

nickvigilante:
Sure, can do. I'll put this PR on hold until I can get a PR for the script, and then I'll bring in the newer RNs into this PR.

gemma-shay (approved):

PR #16245, opened 2023-02-13
DOC-6403 v23.1.0-alpha.2 release notes

andf-crl:
Addresses: DOC-6403

- Release notes for v23.1.0-alpha.2

Process Feedback

- See [diff](https://github.com/cockroachdb/docs/pull/16245/commits/b52ec840ea2b54dc92201aba341b5cd63d4df664) for an example of the manual effort required each time we do these. This diff represents the changes required by hand to get the automated output from `release_notes.py` into human-readable form, even before any copy review. Specifically:
	- Red/Green are entire paragraphs, not sporadic punctuation or added links. Scope: out of control.
	- Basic typos abound. Missing punctuation abounds. Every engineer phrases things differently, so manual edits for basic conformity to style guide -- and even just for actual correctness -- are extensive.
		- Some basic typos invert the intended meaning. Likely the following is very wrong, but I can't tell in which direction:
			- 	`Previously, certain GRANT or REVOKE commands on a user which does not exist would error with the correct PG code. This is now fixed. [#94677][#94677]`
	- No standardization of phrasing, feature names, and product names:
		- CC Console vs DB Console vs Admin Console
		- DB Console page names are all over the place, some refer to page names, some to graph names, some to tabs.
	- Extensive deleting = wasted time reading and judging an item to be N/A. Determination of whether I delete a RN should in fact be an engineering design decision, agreed upon during feature development:
		- ✅ `No docs needed`, and/or:
		- ✅ `No release note required`
	- Specialized knowledge requirement . For example, I happened to know that when engineers say "global privileges" they mean "system-level privileges". Former means nothing to every single CRDB user (outside of CRL), latter is documented and linkable. I'm sure there's CDC or KV stuff in here that I left incorrect because I don't happen to know that the engineer is wrong.
		- Which `crdb_internal` namespace (tables? schemas? built-in functions?) is public and which is CRDB-internal? I left all, can never tell intent with these.
	- New ticket linking policy (`Epic: None` and `CRDB-12345` link) is not expected by `release_notes.py`, and as such is not parsed / stripped / anything, but instead dumped raw. More manual deletions.
		- New "instructions to docs" section (i.e. `Our CTE docs should be updated with the following changes:`) is similarly unexpected, and also dumped raw. More manual deletions.
		- Also don't need any intra-eng banter, like: ` Informs: https://github.com/cockroachdb/cockroach/issues/XXXXX`
	- Many many duplicate release notes. I left every single one as-is because the last time I asked "it was both a bug fix and a feature so it should be in both". It's in both now too, but there are 12 such boths. Probably not actually intended, but I don't have the information to determine. Example: #96029.
	- `Changes without release note annotation` Policy: if no release note annotation, no release note. Deleted all.

Technical Issue:

- Sorting of releases on v23.1.md appears to be inverted from all previous:
	- [v23.1.md](https://deploy-preview-16245--cockroachdb-docs.netlify.app/docs/releases/v23.1.html#v23-1-0-alpha-2) (latest at bottom)
	- [v22.2.md](https://www.cockroachlabs.com/docs/releases/v22.2.html) (latest at top)

Staging

[v23.1.md](https://deploy-preview-16245--cockroachdb-docs.netlify.app/docs/releases/v23.1.html#v23-1-0-alpha-2)


MichaelTrestman (approved):
LGTM. Huge RN, wow!

PR #16259, opened 2023-02-14
Fix spec download, upgrade to RapiDoc v9.3.4

ianjevans:
This PR fixes the issue where the download and view links never updated when changing versions.

It also updates RapiDoc to v9.3.4.

Fixes CC-9141.


nickvigilante (approved):
I can't see the diff of `js/rapidoc-min.js`, but the deploy preview LGTM.

PR #16279, opened 2023-02-16
DOC-6964: Fix an issue where advisories wouldn't appear in the sidenav

nickvigilante:
Fixes DOC-6964

### DIFF
--- a/_includes/sidebar-releases.json
+++ b/_includes/sidebar-releases.json
@@ -58,13 +58,13 @@
     "/releases/kubernetes-operator.html"
   ]
 },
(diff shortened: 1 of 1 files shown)


mdlinville:
Something seems wrong in the preview. When I tried to view the Docs homepage, I am shown the `/` even though it looks like I am in `/docs/`. Then when I expand and click on Releases / Advisories, I am immediately taken to https://deploy-preview-16279--cockroachdb-docs.netlify.app/docs/advisories/a30821.html rather than the index. PTAL.

mdlinville (approved):

PR #16280, opened 2023-02-16
DOC-6661 update txn retry reference page

andf-crl:
Addresses: DOC-6661

- Removed actionable guidance from `transactions.md` and moved directly to `transaction-retry-error-reference.md`, since the latter is directly linked to by code and cloud.
	- Highlighted much more prominently the specific actions to take across all transaction retry errors, split into two categories:
		- Client-side retry handling (what to do when you get them), and
		- Minimizing transaction retry errors (what to do to prevent them or lessen the number of them)
	- Retained error-message-specific guidance as is, excepting that each applicable step 1 now links to the all-errors guidance for client-side retry handling, rather than out-of-document.
- Updated links throughout the corpus to direct users seeking actionable guidance to the new location that provides it
	- Did not update links that discussed automatic retries, or both auto and client-side intervention-type retries together, as `transactions.md` still retains a stub section that introduces these concepts, and in fact addresses auto-type in its entirely there.

Questions

- This PR is only for v23.1. Should this be sent back to v22.2? Further back?
- The [Testing transaction retry logic](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/transaction-retry-error-reference#testing-transaction-retry-logic) section is large and seems like it might only be useful to a few readers of this page. Should I break it out into a separate page and just link to it instead? It is linked to from two other pages, so not sure. EDIT: We've agreed that it should: Moved it and the example code section out to a dedicated page, linked to from its previous home.
- DOC-6661 asks for clarity on the closed timestamp interval, but that information already appears to be present and thorough in the [existing docs](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/transaction-retry-error-reference#retry_serializable), including links to the specific sections called out in the `State of Contention` doc. I checked for accuracy as well: seems OK to me! Please provide specific guidance on what remains to be done, if any. Thank you! EDIT: We've agreed to hold on this until further along in the process (after other writers have made their contribution).

Staging

[v23.1/transaction-retry-error-reference.md](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html) | [v23.1/transactions.md](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/transactions#transaction-retries)


andf-crl:
Hi @kevin-v-ngo , here is the first step of the improved transaction contention reference documentation work. Please let me know if you'd like to see any changes!

Feel free to pass this along to your favorite engineer or engineers if you'd prefer. Thanks so much!

andf-crl:
Hi @kevin-v-ngo ! In our meeting we discussed:

- We will hold off on any timestamp push language changes until Kathryn has done her portion of this epic: as her efforts will rely in part on the changes in this PR, we'll revisit if improvements can be made to this transaction retry reference page only after she's completed her work.
- I will add further context to the action items mentioned under [Minimize transaction retry errors](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html#minimize-transaction-retry-errors) as drawn from your _State of Contention_ gdoc.
- I will move [Testing transaction retry logic](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html#testing-transaction-retry-logic) to a difference page to streamline this one.

Anything else you'd like to see done? Please make a comment -- I am otherwise unaware! Thanks again.

andf-crl:
@kevin-v-ngo I've made the changes you hinted at in our 1:1 meeting:

- Updated [Minimizing transaction retry errors](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html#minimize-transaction-retry-errors) section to more closely match the language and context you use in your "State of Contention" doc.
- Moved the retry logic example pseudocode -- and its testing subsection -- out to its own dedicated doc. Not linking in left-nav, at least for now.

Ready for a look and your comment please! If you don't mind I'm going to reach out to engineering directly as well -- other writers are waiting on this effort to begin their work -- I must get this moving through the review process. Thank you!

kevin-v-ngo (changes requested):
Here's my initial review!

I think we're missing same mitigations we should call out for users. It would be good to provide _some_ context as to why we suggest these mitigations. And then users can choose to go to the existing concepts doc to learn more via links. 

We need to balance simplicity in the mitigation steps for users who are actively troubleshooting an issue and more advance users who care and have the time to learn about the details and link out to the transaction layer reference doc.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,185 @@
...
+
+Calling `run_transaction` without an `op` input sets `inject_retry_errors_enabled` as `true` until the final retry attempt, before which the `inject_retry_errors_enabled` is set back to `false`. For all attempts except the last one, CockroachDB will inject a retryable serialization error for the client to handle. If the client cannot handle the error properly, the retry logic isn't working properly.
+
+### Minimize transaction retry errors
+
+While the steps described in [Client-side retry handling](#client-side-retry-handling) detail how to configure your application to restart a failed transaction, there are also a number of changes you can make to your application logic to prevent transaction retry errors from occurring in the first place:
+
+1. Design your schema and queries to reduce contention. For more information about how contention occurs and how to avoid it, see [Transaction Contention](performance-best-practices-overview.html#transaction-contention). In particular, if you are able to [send all of the statements in your transaction in a single batch](transactions.html#batched-statements), CockroachDB can usually automatically retry the entire transaction for you.
```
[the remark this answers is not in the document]
kevin-v-ngo: >Some of these concepts are pretty complex to describe all here to we'd probably need to link out to the existing transaction layer doc?

I think the behavior of timestamps being pushed and then revalidated in phase 2 (commit) is important to link to.

kevin-v-ngo:
> 

I just your comment here Andrew! Apologies, I forgot to "Finish my Review" and you might've addressed some of them. Let me review again.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,94 @@
...
+
+1. Break down larger transactions into smaller ones (e.g., [bulk deletes](bulk-delete-data.html)) wherever possible, and limit the number of affected rows by following [performance-tuning best practices](apply-statement-performance-rules.html) (e.g., query performance tuning, index design and maintenance, etc.). Not only will transactions run faster and hold locks for a shorter duration but also read-write transactions that affect many rows increases the likelihood of retry errors. For example, as the size of reads (number of rows) increases, the chances of read invalidation when the transaction’s timestamp is pushed due to a conflicting write increases. As the size of writes (number of rows written) increases, the chances of the transaction’s timestamp getting bumped by concurrent reads increases.
```
kevin-v-ngo, line 77: I was thinking we can separate this out into two bullets. Query tuning and Bulk Deletes. 

Although related, I do think the former is more common than the later and the context is a bit different.


Limit the number of affected rows by following [performance-tuning best practices](https://www.cockroachlabs.com/docs/stable/apply-statement-performance-rules.html) (e.g., query performance tuning, index design and maintenance, etc.). Reading less rows will decrease the likelihood of concurrent writes having their timestamp pushed (we should link out here again: https://www.cockroachlabs.com/docs/stable/architecture/transaction-layer.html#read-refreshing) which can lead to retry errors.

Break down larger transactions into smaller ones (e.g., [bulk deletes](https://www.cockroachlabs.com/docs/stable/bulk-delete-data.html)) to have transactions hold locks for a shorter duration. This will also decrease the likelihood of pushed timestamps and retry errors. For instance, as the size of writes (number of rows written) decreases, the chances of the (bulk delete) transaction’s timestamp getting bumped by concurrent reads decreases.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,94 @@
...
+
+1. Break down larger transactions into smaller ones (e.g., [bulk deletes](bulk-delete-data.html)) wherever possible, and limit the number of affected rows by following [performance-tuning best practices](apply-statement-performance-rules.html) (e.g., query performance tuning, index design and maintenance, etc.). Not only will transactions run faster and hold locks for a shorter duration but also read-write transactions that affect many rows increases the likelihood of retry errors. For example, as the size of reads (number of rows) increases, the chances of read invalidation when the transaction’s timestamp is pushed due to a conflicting write increases. As the size of writes (number of rows written) increases, the chances of the transaction’s timestamp getting bumped by concurrent reads increases. 
+
+1. Use [`SELECT FOR UPDATE`](select-for-update.html) to aggressively lock rows that will later be updated in the transaction. Locking (blocking) earlier in the transaction will not allow other write transactions to conflict and cause read invalidation when the transaction’s timestamp is pushed.
```
kevin-v-ngo, line 79: My rationale for SELECT for UPDATE in the state of contention doc was a bitt difficult to understand. I've updated it to the following:

SELECT FOR UPDATE locks rows that will later be updated in the transaction. Locking (blocking) earlier in the transaction will not allow other concurrent write transactions to conflict which leads to a situation where we would return out-of-date information subsequently returning a retry error (RETRY_WRITE_TOO_OLD).

Thoughts? Feel free to edit.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,185 @@
...
+
+1. Design your schema and queries to reduce contention. For more information about how contention occurs and how to avoid it, see [Transaction Contention](performance-best-practices-overview.html#transaction-contention). In particular, if you are able to [send all of the statements in your transaction in a single batch](transactions.html#batched-statements), CockroachDB can usually automatically retry the entire transaction for you.
+
+1. Use historical reads with [`SELECT ... AS OF SYSTEM TIME`](as-of-system-time.html).
+
+1. In some cases, [`SELECT FOR UPDATE`](select-for-update.html) can also be used to reduce the occurrence of serialization errors.
+
+1. [High priority transactions](transactions.html#transaction-priorities) are less likely to experience serialization errors than low priority transactions. Adjusting transaction priorities usually does not change how many serialization errors occur, but it can help control which transactions experience them.
+
```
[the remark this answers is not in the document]
kevin-v-ngo: Same comment before your latest edit. Just swapped order to prioritize column family mitigation over closed timestamp interval

v23.1/transaction-retry-error-reference.md
```diff
@@ -63,19 +103,23 @@ This "conversational" design means that there is no way for the server to always
...
 _Description_:
 
 The `RETRY_WRITE_TOO_OLD` error occurs when a transaction _A_ tries to write to a row _R_, but another transaction _B_ that was supposed to be serialized after _A_ (i.e., had been assigned a higher timestamp), has already written to that row _R_, and has already committed. This is a common error when you have too much contention in your workload.
 
 _Action_:
 
-1. Retry transaction _A_ as described in [client-side retry handling](transactions.html#client-side-intervention).
+1. Retry transaction _A_ as described in [client-side retry handling](#client-side-retry-handling).
 1. Design your schema and queries to reduce contention. For more information about how contention occurs and how to avoid it, see [Transaction Contention](performance-best-practices-overview.html#transaction-contention). In particular, if you are able to [send all of the statements in your transaction in a single batch](transactions.html#batched-statements), CockroachDB can usually automatically retry the entire transaction for you.
```
kevin-v-ngo, line 123: Can't we just link to this section? https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html#minimize-transaction-retry-errors

Users might think "single batch" is the only mitigation.

kevin-v-ngo: Also we should call out/emphasize the SELECT FOR UPDATE mitigation

v23.1/transaction-retry-error-reference.md
```diff
@@ -97,13 +143,13 @@ The `RETRY_SERIALIZABLE` error occurs in the following cases:
...
    1. Use historical reads with [`SELECT ... AS OF SYSTEM TIME`](as-of-system-time.html).
 
 1. If you encounter case 3 above, the solution is to:
    1. Increase the `kv.closed_timestamp.target_duration` setting to a higher value. As described above, this will impact the freshness of data available via [Follower Reads](follower-reads.html) and [CDC changefeeds](change-data-capture-overview.html).
-   1. Retry transaction _A_ as described in [client-side retry handling](transactions.html#client-side-intervention).
+   1. Retry transaction _A_ as described in [client-side retry handling](#client-side-retry-handling).
    1. Design your schema and queries to reduce contention. For more information about how contention occurs and how to avoid it, see [Transaction Contention](performance-best-practices-overview.html#transaction-contention). In particular, if you are able to [send all of the statements in your transaction in a single batch](transactions.html#batched-statements), CockroachDB can usually automatically retry the entire transaction for you.
```
kevin-v-ngo, line 153: Can't we just link to this entire section? https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/dev/transaction-retry-error-reference.html#minimize-transaction-retry-errors

v23.1/transaction-retry-error-reference.md
```diff
@@ -97,13 +143,13 @@ The `RETRY_SERIALIZABLE` error occurs in the following cases:
 _Action_:
```
kevin-v-ngo, line 143: >When a transaction A is forced to refresh (i.e., change its timestamp) due to hitting the maximum [closed timestamp](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/architecture/transaction-layer#closed-timestamps) interval (closed timestamps enable [Follower Reads](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/follower-reads#how-stale-follower-reads-work) and [Change Data Capture (CDC)](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/change-data-capture-overview)). This can happen when transaction A is a long-running transaction, and there is a write by another transaction to data that A has already read. If this is the cause of the error, the solution is to increase the kv.closed_timestamp.target_duration setting to a higher value. Unfortunately, there is no indication from this error code that a too-low closed timestamp setting is the issue. Therefore, you may need to rule out cases 1 and 2 (or experiment with increasing the closed timestamp interval, if that is possible for your application - see the note below).


This is actually good context to add to the shared section (Minimize transaction retry errors) above. If no other mitigations work, change the setting it this is a long-runing transaction

kevin-v-ngo (changes requested):
Second review (after your edits). Thanks!

andf-crl (commented):
Hi @kevin-v-ngo -- thanks so much for the feedback!

I've commented inline to comments that survived, and will respond here generally to comments from previous commits that have disappeared:

- I've went through all specific errors and added, where applicable:
	- Clearer language directing to the "minimize" section.
	- Brief indication of especially-important remediations, such as SFU for `RETRY_WRITE_TOO_OLD`.
	-  A end-of-section "for more info" also pointing back to "minimize" section.
- Ensured that "minimize" section matches your contention gdoc and github comments more closely.  Specifically:
	- Broke longer recommendations into smaller, discrete recommendations.
	- Linked out to concepts as noted.
- Clarified that Serialization errors can also be remediated via refactor and tune.

Ready for another look! Thanks so much!

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,96 @@
...
+Suppose that the client is [a Java application using JDBC](build-a-java-app-with-cockroachdb.html), or an analyst typing [`BEGIN`](begin-transaction.html) directly to [a SQL shell](cockroach-sql.html). In either case, the client is free to issue a `BEGIN`, wait an arbitrary amount of time, and issue additional statements. Meanwhile, other transactions are being processed by the system, potentially writing to the same data.
 
-1. In some cases, [`SELECT FOR UPDATE`](select-for-update.html) can also be used to reduce the occurrence of serialization errors.
+This "conversational" design means that there is no way for the server to always retry the arbitrary statements sent so far inside an open transaction. If there are different results for any given statement than there were at an earlier point in the currently open transaction's lifetime (likely due to the operations of other, concurrently-executing transactions), CockroachDB must defer to the client to decide how to handle that situation. This is why we recommend [keeping transactions as small as possible](performance-best-practices-overview.html#transaction-contention).
```
kevin-v-ngo, line 29: Re-reading this section, does this seem out of place? It's an action but also links out to an overview doc. Maybe just remove it?

>This is why we recommend [keeping transactions as small as possible](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/performance-best-practices-overview#transaction-contention).

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,96 @@
...
+
+1. Break down larger transactions into smaller ones (e.g., [bulk deletes](bulk-delete-data.html)) to have transactions hold locks for a shorter duration. This will also decrease the likelihood of [pushed timestamps](architecture/transaction-layer.html#timestamp-cache) and retry errors. For instance, as the size of writes (number of rows written) decreases, the chances of the (bulk delete) transaction’s timestamp getting bumped by concurrent reads decreases.
+
+1. Design your applications to reduce network round trips by [sending statements in transactions as a single batch](transactions.html#batched-statements) (e.g., using [common table expressions](common-table-expressions.html)).
```
kevin-v-ngo, line 79: Can we elaborate why?

Batching allows CRDB to automaticall retry when previous reads are invalidated at a pushed timestamp.

https://www.cockroachlabs.com/docs/stable/transactions.html#automatic-retries

kevin-v-ngo: Same reason as 4 below. The conditions for retry is less than 16kb and has not yet streamed results back. The later is why users should batch.

kevin-v-ngo (changes requested):
Looking good! I'll finish reviewing the Transaction retry error reference section this week.

v23.1/transaction-retry-error-reference.md
andf-crl: This is under the `Overview` section: it is indeed intended as an overview. If you want this page 100% "do as we tell you" I can indeed remove the entire `## Overview` section. Isn't it helpful to be aware of serializable ordering always, auto-rety txns whenever possible, etc. I think there is some merit to providing this background before then proceeding to use these same concepts in actionable guidance throughout the rest of the page.

I've deleted just the "Suppose that a client is ... This is why we recommend keeping transactions as small as possible." section. What do you think? (I think it's still valuable to keep a brief overview of the concepts we use throughout the rest of the page).

v23.1/transaction-retry-error-reference.md
andf-crl: Now #3 and #4 have nearly identical descriptions. To clarify then, can you elaborate on what you meant in your most recent comment, so I can add? 

> The conditions for retry is less than 16kb and has not yet streamed results back. The later is why users should batch.

Maybe (total guess):

> Batching allows CockroachDB to [automatically retry](transactions.html#automatic-retries) a transaction when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache), as long as the total transaction side (data written) is less than `16kb` and has not yet streamed any results back to the client.

andf-crl:
Ready for another look @kevin-v-ngo , but also need to get this to copy reivew, and out to the other writers who are waiting on it. Moving to copy review, but feel free to leave any lingering comments / answers to my questions as you find time. Thanks!

v23.1/transaction-retry-error-example.md
```diff
@@ -0,0 +1,118 @@
+---
+title: Client side retry handling example
+summary: Client side transaction retry error retry handling example and testing
+toc: true
+docs_area: reference.transaction_retry_error_example
+---
+
+When a [transaction](transactions.html) is unable to complete due to another concurrent or recent transaction attempting to write to the same data, CockroachDB will [automatically attempt to retry the failed transaction](transactions.html#automatic-retries) without involving the client (i.e., silently). If the automatic retry is not possible or fails, a [transaction retry error](transaction-retry-error-reference.html)is emitted to the client. Multiple transactions trying to write to the same data is also known as contention.
```
MichaelTrestman, line 8: 'attempt to retry' is redundant, yeah?

```suggestion
When a [transaction](transactions.html) is unable to complete due to another concurrent or recent transaction attempting to write to the same data, CockroachDB will [automatically retry the failed transaction](transactions.html#automatic-retries) without involving the client (i.e., silently). If the automatic retry is not possible or fails, a [transaction retry error](transaction-retry-error-reference.html)is emitted to the client. Multiple transactions trying to write to the same data is also known as contention.
```

v23.1/transaction-retry-error-example.md
```diff
@@ -0,0 +1,118 @@
...
+                time.sleep(sleep_ms)
+
+            except psycopg2.Error as e:
+                logging.debug("got error: %s", e)
+                raise e
+
+        raise ValueError(f"Transaction did not succeed after {max_retries} retries")
+~~~
+
+`run_transaction` takes a SQL-executing function `op(conn)` and attempts to run the function, retrying on serialization failures (exposed in pscyopg2 as the [`SerializationFailure` exception class](https://www.psycopg.org/docs/errors.html#sqlstate-exception-classes)) with exponential backoff, until reaching a maximum number of tries.
```
MichaelTrestman, line 59: This explanation of the function makes more sense to me above the code block? Maybe like "suppose you've written a wrapper function `run_transaction`, which takes a ..."

v23.1/transaction-retry-error-example.md
```diff
@@ -0,0 +1,118 @@
...
+                time.sleep(sleep_ms)
+
+            except psycopg2.Error as e:
+                logging.debug("got error: %s", e)
+                raise e
+
+        raise ValueError(f"Transaction did not succeed after {max_retries} retries")
+~~~
+
+Calling `run_transaction` without an `op` input sets `inject_retry_errors_enabled` as `true` until the final retry attempt, before which the `inject_retry_errors_enabled` is set back to `false`. For all attempts except the last one, CockroachDB will inject a retryable serialization error for the client to handle. If the client cannot handle the error properly, the retry logic isn't working properly.
```
MichaelTrestman, line 109: The error handling bit, after "For all..." seems separate and important enough to have its own header, nah?

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,92 @@
 ---
 title: Transaction Retry Error Reference
-summary: A list of the transaction retry (serialization) errors emitted by CockroachDB, including likely causes and user actions for mitigation.
+summary: A list of the transaction retry errors emitted by CockroachDB, including likely causes and user actions for mitigation.
 toc: true
 docs_area: reference.transaction_retry_error_reference
 ---
 
-This page has a list of the transaction retry error codes emitted by CockroachDB.
+When a [transaction](transactions.html) is unable to complete due to another concurrent or recent transaction attempting to write to the same data, CockroachDB will [automatically attempt to retry the failed transaction](transactions.html#automatic-retries) without involving the client (i.e., silently). If the automatic retry is not possible or fails, a _transaction retry error_ is emitted to the client. Multiple transactions trying to write to the same data is also known as contention.
```
MichaelTrestman, line 8: Same point as I raised on the other file about maybe referring to a glossary definition of contention in order to consolidate, and also provide a bit more info with the canonical definition.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,92 @@
...
+Transaction retry errors fall into two categories:
 
-The failure to place a transaction into a serializable ordering usually happens due to a [transaction conflict](architecture/transaction-layer.html#transaction-conflicts) with another concurrent or recent transaction that is trying to write to the same data. When multiple transactions are trying to write to the same data, this state is also known as contention. When serialization errors occur due to contention, the transaction needs to be retried by the client as described in [client-side retry handling](transactions.html#client-side-intervention).
+- **Serialization Errors** indicate that a transaction failed because it could not be placed into a [serializable ordering](demo-serializable.html) among all of the currently-executing transactions. These errors are generally addressed with client-side intervention, where the client [initiates a restart of the transaction](#client-side-retry-handling), and [adjusts application logic and tunes queries](#minimize-transaction-retry-errors) for greater performance.
```
MichaelTrestman, line 12: Not 100% sure as this isn't my area, but this wording sounds incorrect. "Could not be placed into a serializable ordering" seems garbled/redundant, because the 'series' and 'order' are the same concept. So you could say that the transaction 'could not be serialized' or 'could not be placed into an (valid? determinate?) ordering', but combining into 'serializable ordering' seems problematic. Although maybe I just don't get it, totes possible

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,92 @@
...
+
+1. Design your applications to reduce network round trips by [sending statements in transactions as a single batch](transactions.html#batched-statements) (e.g., using [common table expressions](common-table-expressions.html)). Batching allows CockroachDB to [automatically retry](transactions.html#automatic-retries) a transaction when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache).
+
+1. Limit the size of the result sets of your transactions so that CockroachDB is more likely to [automatically retry](transactions.html#automatic-retries) when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache).
```
MichaelTrestman, line 77: Maybe I just don't understand database stuff? I never work with SQL logic over in security land. But I don't at all follow this "...so that..."

Also, isn't this recommendation (limit the size of result sets) at odds with the previous one (batch transactions together), insofar as batched transactions will often have larger result sets?

v23.1/transaction-retry-error-reference.md
```diff
@@ -63,27 +101,38 @@ This "conversational" design means that there is no way for the server to always
...
 
+Each transaction retry error listed includes an example error as it would appear from the context of the client, a description of the circumstances that cause that error, and specific guidance for addressing the error.
+
 ### RETRY_WRITE_TOO_OLD
 
 ```
 TransactionRetryWithProtoRefreshError: ... RETRY_WRITE_TOO_OLD ...
 ```
 
+_Error type_: Serialization error
```
MichaelTrestman, line 112: I'm definitely not in love with the strange formatting with italics for these subsection headers?

MichaelTrestman (approved):
A few suggestions, mostly LGTM!

v23.1/transaction-retry-error-reference.md
andf-crl: You're right that these two sound about the same, but I was asked to make them explicit separate guidance. I'll ask about this specifically to be sure. Thanks!

v23.1/transaction-retry-error-reference.md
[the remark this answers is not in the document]
kathancox: Hey Andrew, just peeking on these docs as I start my work... should we mention the actual name of the setting here? It links to the doc I'm working on, but the setting name is mentioned there either. In docs, it was actually hard to find `kv.closed_timestamp.target_duration` as it is not public?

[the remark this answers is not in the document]
kathancox: I suppose this setting can have adverse effects on a bunch of things, so that is probably why it is not public. Given we mentioning it in docs though, do we need to think about it's status?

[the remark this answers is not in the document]
andf-crl: Agreed, I think we should definitely name it here, if it's part of official guidance (the intent of this page). How else would users be able to implement it if we don't name it!! I'll make that update, thanks much, great call!

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,92 @@
...
+- **Advanced users, such as library authors**: See [Advanced Client-Side Transaction Retries](advanced-client-side-transaction-retries.html).
 
-This "conversational" design means that there is no way for the server to always retry the arbitrary statements sent so far inside an open transaction. If there are different results for any given statement than there were at an earlier point in the currently open transaction's lifetime (likely due to the operations of other, concurrently-executing transactions), CockroachDB must defer to the client to decide how to handle that situation. This is why we recommend [keeping transactions as small as possible](performance-best-practices-overview.html#transaction-contention).
+#### Client-side retry handling example
 
-## Error reference
+For a conceptual example of application-defined retry logic, and testing that logic against your application's needs, see the [client-side retry handling example](transaction-retry-error-example.html).
+
+### Minimize transaction retry errors
```
kevin-v-ngo, line 67: @michae2 and @DrewKimball - Do y'all mind giving this section a review? It would be good to get your set of eyes on this from the SQL side of things. I know you two have been engaged in several escalations related to retry errors.

v23.1/transaction-retry-error-reference.md
kevin-v-ngo: Oh yeah! Agree to keep the Overview section. I was just commenting on "This is why we recommend keeping transactions as small as possible." It felt random to have an action in this part of the doc.

v23.1/transaction-retry-error-reference.md
kevin-v-ngo: That's right! **Both** conditions (resultset size and less than 16KB) must be satisfied. 

For 3), If the user does not batch and they create a multi-statement transaction that does >1 roundtrip, CRDB will not automatically retry (even if the resultset is <16KB).

For 4), if the user sends their transaction as a single batch, but resultset hits 16KB where CRDB starts streaming results back, CRDB will not automatically retry.

v23.1/transaction-retry-error-reference.md
kevin-v-ngo: Overall goal is to allow CRDB to automatically retry. If user does not batch, no auto-retry. If user batches, but resultset hits 16KB, no auto-retry. If batch, and resultset is <16KB, 👍

But yeah, they are related since batching will increase the resultset.

hand-crdb (commented):
Here is some feedback:

**(1)** Under "Client-side retry handling" the example code contradicts what the text says.

The text says

> Your application should include client-side retry handling when the statements are sent individually, such as:

But the example shown is when multiple statements are sent together, not sent individually.  This is reinforced by:

* There is just one prompt
* The copy widget for the example copies all statements together in a single piece of text.

---

**(2)** Item 6 under "Minimize transaction retry errors" is confusing because it conflates transaction **staleness** with transaction **priority**.  It says:

> Use historical reads ([SELECT ... AS OF SYSTEM TIME](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/as-of-system-time)), preferably [bounded staleness reads](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/follower-reads#when-to-use-bounded-staleness-reads) or [exact staleness with follower reads](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/follower-reads#run-queries-that-use-exact-staleness-follower-reads) when possible to reduce conflicts with other writes. This reduces the likelihood of conflicts as fewer writes will happen at the historical timestamp. More specifically, writes’ timestamps are less likely to be pushed by historical reads as they would [when the read has a higher priority level](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/architecture/transaction-layer#transaction-conflicts).

Transaction staleness and transaction priority are different things.

---

**(3)** Item 3 under "Minimize transaction retry errors" is confusing because it conflates sending multiple statements in a single batch with combining multiple operations in a single statement using common table expressions.

It says:

> Design your applications to reduce network round trips by [sending statements in transactions as a single batch](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/transactions#batched-statements) (e.g., using [common table expressions](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/common-table-expressions)).

A common table expression is not an example of sending multiple statements in a transaction as a single batch.  It is an example of combining multiple operations (that otherwise would be performed via separate statements) into a single statement.

So, while common table expressions can be part of a strategy for minimizing round trips to the database, conflating these concepts is ultimately confusing.


THANK YOU

andf-crl:
@hand-crdb Thank you. I'll investigate overriding our code block to more clearly show 4 separate, unbatched statements.

The rest of your guidance directly contradicts Kevin's, and the [State of Contention report](https://docs.google.com/document/d/1_VpO6DTc9krVoIBCYhiRy5LujcRer5pxiso27Cn9jKI/edit#heading=h.px0q8ikemnol) on which these changes were based. Please search for `common table expressions` and `staleness reads` in the above-linked gdoc for verbatim guidance. Should you have perhaps been consulted in the initial creation of this gdoc?

@kevin-v-ngo can you comment on the above comments? Which conflicting guidance is correct here?

DrewKimball (commented):
I have some comments around organization + a few nits, but overall looks good.

kevin-v-ngo:
Thanks for the review Steven!!

> **(2)** Item 6 under "Minimize transaction retry errors" is confusing because it conflates transaction **staleness** with transaction **priority**. It says:
> 
> > Use historical reads ([SELECT ... AS OF SYSTEM TIME](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/as-of-system-time)), preferably [bounded staleness reads](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/follower-reads#when-to-use-bounded-staleness-reads) or [exact staleness with follower reads](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/follower-reads#run-queries-that-use-exact-staleness-follower-reads) when possible to reduce conflicts with other writes. This reduces the likelihood of conflicts as fewer writes will happen at the historical timestamp. More specifically, writes’ timestamps are less likely to be pushed by historical reads as they would [when the read has a higher priority level](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/architecture/transaction-layer#transaction-conflicts).
> 
> Transaction staleness and transaction priority are different things.


Perhaps we should be more clear here. If a Read Transaction with a higher priority encounters a lower priority Write Transaction (its intent), the Write Transaction will have its timestamp pushed. This may cause the Write Transaction to retry if it fails its refresh. If the Read Transaction uses AOST (staleness reads), it will less likely encounter the Write Transaction (write intents) which will reduce retries. This is why AOST reads is preferable if the application can staleness. Correct @hand-crdb?




> 
> **(3)** Item 3 under "Minimize transaction retry errors" is confusing because it conflates sending multiple statements in a single batch with combining multiple operations in a single statement using common table expressions.
> 
> It says:
> 
> > Design your applications to reduce network round trips by [sending statements in transactions as a single batch](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/transactions#batched-statements) (e.g., using [common table expressions](https://deploy-preview-16280--cockroachdb-docs.netlify.app/docs/v23.1/common-table-expressions)).
> 
> A common table expression is not an example of sending multiple statements in a transaction as a single batch. It is an example of combining multiple operations (that otherwise would be performed via separate statements) into a single statement.
> 
> So, while common table expressions can be part of a strategy for minimizing round trips to the database, conflating these concepts is ultimately confusing.

Steven is technical correct and we should clarify! There can be other ways to send multiple statements in a transaction as a single batch - CTEs is one way. Perhaps we can be more explicit instead of calling out CTE in the "(e.g., )".  Clarify that one possible approach is to refactor your application to take advantage of CTEs to send as a single batch.

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,91 @@
...
+
+1. Use historical reads ([`SELECT ... AS OF SYSTEM TIME`](as-of-system-time.html)), preferably [bounded staleness reads](follower-reads.html#when-to-use-bounded-staleness-reads) or [exact staleness with follower reads](follower-reads.html#run-queries-that-use-exact-staleness-follower-reads) when possible to reduce conflicts with other writes. This reduces the likelihood of conflicts as fewer writes will happen at the historical timestamp. More specifically, writes’ timestamps are less likely to be pushed by historical reads as they would [when the read has a higher priority level](architecture/transaction-layer.html#transaction-conflicts).
```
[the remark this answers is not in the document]
kevin-v-ngo: Yeah I commented above as well from Steven. Wording was a bit poorly on my part here. This is only the scenario when the read has the higher priority and encounters a low priority write which causes the write's timestamp to be pushed. If staleness read, it would reduce the likelihood of running into the intent in the first place correct?

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,91 @@
...
+
+1. Design your applications to reduce network round trips by [sending statements in transactions as a single batch](transactions.html#batched-statements) (e.g., using [common table expressions](common-table-expressions.html)). Batching allows CockroachDB to [automatically retry](transactions.html#automatic-retries) a transaction when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache). When a multi-statement transaction is not batched, and takes more than a single round trip, CockroachDB cannot automatically retry the transaction.
+
+1. Limit the size of the result sets of your transactions to under 16KB, so that CockroachDB is more likely to [automatically retry](transactions.html#automatic-retries) when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache). When a transaction returns a result set over 16KB, even if that transaction has been sent as a single batch, CockroachDB cannot automatically retry the transaction.
```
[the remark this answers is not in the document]
kevin-v-ngo: Great point! This can actually be configured - 16KB is default

v23.1/transaction-retry-error-reference.md
```diff
@@ -1,54 +1,91 @@
...
+
+1. Break down larger transactions into smaller ones (e.g., [bulk deletes](bulk-delete-data.html)) to have transactions hold locks for a shorter duration. This will also decrease the likelihood of [pushed timestamps](architecture/transaction-layer.html#timestamp-cache) and retry errors. For instance, as the size of writes (number of rows written) decreases, the chances of the (bulk delete) transaction’s timestamp getting bumped by concurrent reads decreases.
+
+1. Design your applications to reduce network round trips by [sending statements in transactions as a single batch](transactions.html#batched-statements) (e.g., using [common table expressions](common-table-expressions.html)). Batching allows CockroachDB to [automatically retry](transactions.html#automatic-retries) a transaction when [previous reads are invalidated](architecture/transaction-layer.html#read-refreshing) at a [pushed timestamp](architecture/transaction-layer.html#timestamp-cache). When a multi-statement transaction is not batched, and takes more than a single round trip, CockroachDB cannot automatically retry the transaction.
```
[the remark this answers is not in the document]
kevin-v-ngo: > some of these recommendations are specific to a particular retry error type. I'm not sure how helpful it is to have them all under a general recommendations section at least without some more context given about what sort of errors they each prevent.

Yeah, good callout Drew. Although these are generally good best practices for performance, only some will directly minimize specific retry errors. @andf-crl, perhaps we can label each action with some type of key and reference each key in the retry error section? Or we can just duplicate the action/content in each retry error section. Thoughts?

v23.1/transaction-retry-error-reference.md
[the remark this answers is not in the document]
DrewKimball: >If staleness read, it would reduce the likelihood of running into the intent in the first place correct?

Yeah, sounds right to me.

andf-crl:
Hi @kevin-v-ngo , any update on your end? I think you were waiting on consensus from engineering? Perhaps:

> This is why AOST reads is preferable if the application can staleness. Correct @hand-crdb?

And

> Perhaps we can be more explicit instead of calling out CTE in the "(e.g., )". Clarify that one possible approach is to refactor your application to take advantage of CTEs to send as a single batch.

But Steven said explicitly that CTEs are not "one possible approach to" but instead "not an example of":

> A common table expression is not an example of sending multiple statements in a transaction as a single batch.

Which should the docs reflect? Let me know, engineering, what engineering agrees on. Thanks tem.

kevin-v-ngo:
> But Steven said explicitly that CTEs are not "one possible approach to" but instead "not an example of":
> 
> > A common table expression is not an example of sending multiple statements in a transaction as a single batch.

You can just say that you can use CTEs to combine multiple statements into a single batch. It's what we guide customers to do so that it can increase the amount of internal retries.

shannonbradshaw:
I suggest we capture the suggested improvements by @DrewKimball in a DOC issue for later improvements and merge this in its current form. This PR has been in review for a month. It would be good to get these improvements published.

andf-crl:
Thanks folks. This PR is being merged as is, due to the following direct instructions:

- [From Shannon B.](https://github.com/cockroachdb/docs/pull/16280/#issuecomment-1477748085):
> I suggest we capture the suggested improvements by @DrewKimball in a DOC issue for later improvements and merge this in its current form. This PR has been in review for a month. It would be good to get these improvements published.

- [From Kevin N.](https://github.com/cockroachdb/docs/pull/16280/#issuecomment-1476839748):
> Overall, I approve of this PR. Nothing blocking on my end.

Per Shannon's request, I have created D O C - 7 2 6 8 to capture remaining items raised in the above comments. 

Thanks all for the reviews.

PR #16284, opened 2023-02-16
crdb_internal.cluster_contention_events examples++

rmloveland:
Addresses some of the spirit of DOC-3395

Helps out with DOC-6663 and friends as well

### DIFF
--- a/v23.1/crdb-internal.md
+++ b/v23.1/crdb-internal.md
@@ -268,6 +268,32 @@ SELECT * FROM crdb_internal.cluster_contention_events;
 (9 rows)
 ~~~
 
+#### View the tables/indexes with the most time under contention
+
+To view the [tables](create-table.html) and [indexes](indexes.html) with the most cumulative time under [contention](performance-best-practices-overview.html#transaction-contention) since the last server restart, run the query below.
+
+{{site.data.alerts.callout_info}}
+The default tracing behavior captures a small percent of transactions so not all contention events will be recorded. When investigating transaction contention, you can set the `sql.trace.txn.enable_threshold` [cluster setting](cluster-settings.html#setting-sql-trace-txn-enable-threshold) to always capture contention events.
+{{site.data.alerts.end}}
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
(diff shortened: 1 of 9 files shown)


rmloveland:
@yuzefovich added you since it looks like you added the `crdb_internal.cluster_contention_events` table

@jonstjohn added you to review since this query is snarfed from a wiki page you wrote (and I figured it should be in docs!)

FYI this is happening because of a big initiative in Docs/Education around contention via DOC-6652

yuzefovich:
I did add `crdb_internal.cluster_contention_events` table, but it's been a while, and I think @j82w is now most-up-to-date on the contention stuff, so I'll defer to him on the review.

j82w (approved):

v23.1/crdb-internal.md
```diff
@@ -268,6 +268,28 @@ SELECT * FROM crdb_internal.cluster_contention_events;
 (9 rows)
 ~~~
 
+#### View the tables/indexes with the most time under contention
+
+To view the [tables](create-table.html) and [indexes](indexes.html) with the most cumulative time under [contention](performance-best-practices-overview.html#transaction-contention) since the last server restart, run the following query:
```
j82w, line 273: The contention events are based on tracing. Tracing by default is only enabled for a very small percentage of requests by default. It probably should mention this to avoid confusion. Related issue: https://github.com/cockroachdb/cockroach/issues/96453

rmloveland: thanks @j82w !

I've added a note to that effect (which also appears at https://www.cockroachlabs.com/docs/dev/ui-insights-page#transaction-executions-view) that says 

> The default tracing behavior captures a small percent of transactions so not all contention events will be recorded. When investigating transaction contention, you can set the sql.trace.txn.enable_threshold [cluster setting](https://www.cockroachlabs.com/docs/v23.1/cluster-settings) to always capture contention events.

I've also filed https://cockroachlabs.atlassian.net/browse/DOC-6967 to add that info to all of the `crdb_internal` things that look at contention, since presumably they all use that setting

ianjevans (approved):

PR #16297, opened 2023-02-19
DOC-6275 added key for v23

ashleykwtam:
Addresses DOC-6275
- continuation of changing the nav and renaming 'Use Cloud Storage' page
- fixes dropdown error for v23.1

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -382,7 +382,6 @@ layout: null
 /docs/cockroachcloud/migrate-from-free-to-dedicated /docs/cockroachcloud/migrate-from-serverless-to-dedicated.html 301
 /docs/cockroachcloud/free-faqs /docs/cockroachcloud/serverless-faqs.html 301
 /docs/v22.2/use-cloud-storage-for-bulk-operations.html /docs/v22.2/use-cloud-storage.html 301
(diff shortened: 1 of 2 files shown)


_redirects
```diff
@@ -382,7 +382,6 @@ layout: null
 /docs/cockroachcloud/migrate-from-free-to-dedicated /docs/cockroachcloud/migrate-from-serverless-to-dedicated.html 301
 /docs/cockroachcloud/free-faqs /docs/cockroachcloud/serverless-faqs.html 301
 /docs/v22.2/use-cloud-storage-for-bulk-operations.html /docs/v22.2/use-cloud-storage.html 301
-/docs/v23.1/use-cloud-storage-for-bulk-operations.html /docs/v23.1/use-cloud-storage.html 301
```
nickvigilante, line 385: Any reason you're moving the redirect here? I think it's still necessary to include.

nickvigilante (commented):
One question on a change.

_redirects
ashleykwtam: added back above

ashleykwtam:
thanks Nick!

PR #16316, opened 2023-02-22
Corrected the navigation for backup/restore

kathancox:
Fixes DOC-6675

This is a slight rearrange of the backup and restore nav.

Currently, the navigation is arranged into cloud/self-hosted. This is actually really confusing because the self-hosted dropdown contains nearly all the features that are also available in cloud. In fact, we should group by customer-owned / managed-service backups as this determines more of what users can do with a backup.

## Preview

https://deploy-preview-16316--cockroachdb-docs.netlify.app/docs/stable/backup-and-restore-overview.html

### DIFF
--- a/_includes/v22.2/backups/feature-availability-note.md
+++ b/_includes/v22.2/backups/feature-availability-note.md
@@ -1,0 +1,1 @@
+This feature is available in **{{ site.data.products.dedicated }}**, **{{ site.data.products.serverless }}**, and **{{ site.data.products.core }}**.
--- a/_includes/v22.2/backups/feature-availability-note.md
+++ b/_includes/v22.2/backups/feature-availability-note.md
@@ -1,1 +1,0 @@
-This feature is available in **{{ site.data.products.dedicated }}**, **{{ site.data.products.serverless }}**, and **{{ site.data.products.core }}**.
--- a/_includes/v22.2/backups/support-products.md
+++ b/_includes/v22.2/backups/support-products.md
@@ -1,0 +1,3 @@
+## Supported products
+
+The feature described on this page is available in **{{ site.data.products.dedicated }}**, **{{ site.data.products.serverless }}**, and **{{ site.data.products.core }}**.
--- a/_includes/v22.2/sidebar-data/manage.json
+++ b/_includes/v22.2/sidebar-data/manage.json
@@ -97,7 +97,7 @@
       ]
     },
     {
-      "title": "Back Up and Restore Data",
+      "title": "Backups and Restores",
       "items": [
         {
           "title": "Backup and Restore Overview",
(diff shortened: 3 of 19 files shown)


kathancox:
@livlobo This is the renav that we discussed. The only thing that is different from what I shared with you in Slack, is that in the v23.1 pages (like full + incr, locality-aware, etc.) I have added a sentence at the top that the feature is available in self-hosted, dedicated, and serverless. I'm not sure this is necessary. But I kept it in the one version for you to check out.

kathancox:
image

@livlobo Re the monitoring PR #16199, I would rebase that on top of this renav and include the monitoring page in customer-owned backups for now.... I think?

kathancox:
Hey @andf-crl, pushing this to docs review as it's a slight change to the backup/restore nav. Can you read the comments above? You'll see that I have added the product list to one version at the top of each page... this was a test and I'm kind of against the idea now, but I would love your thoughts on that too!

_includes/v22.2/backups/feature-availability-note.md
```diff
@@ -0,0 +1 @@
+This feature is available in **{{ site.data.products.dedicated }}**, **{{ site.data.products.serverless }}**, and **{{ site.data.products.core }}**.
```
andf-crl, line 1: Hmm, I'm actually torn here. It is a little hard to know which this page applies to, until the more extensive left-nav CC vs SH changes are implemented (that Mike has been talking about), so I don't mind a little bit of explicitness. That being said, I would definitely prefer for applicability to be entirely determinable by left-nav when we get to that stage.

What do you think about retaining this explicit mention for now, until Mike's re-org proposal is completed?

Or: If you don't like it right up top as the first text on-page, I think a `## Supported Products` section or similar would also be fine, perhaps as a section of its own just before `## Use Key Management Service` on [this page](https://deploy-preview-16316--cockroachdb-docs.netlify.app/docs/dev/take-and-restore-encrypted-backups.html).

I really could go either way on this for the interim. Up to you.

andf-crl (approved):
I like the changes!

I'm torn on the explicit SH vs CC callout -- could really go either way! Left a comment describing my torn-ness.

I do have one tiny comment beyond the changes you've presented: I'd drop the `Use` from the like `[Use Managed-Service Backups]` link on [this page](https://deploy-preview-16316--cockroachdb-docs.netlify.app/docs/cockroachcloud/take-and-restore-customer-owned-backups.html) to match your new page name. Or perhaps just link the bolded header, like you do for `[Customer-owned backups]`: on [this page](https://deploy-preview-16316--cockroachdb-docs.netlify.app/docs/cockroachcloud/use-managed-service-backups.html). Very up to you.

Otherwise, LGTM!

_includes/v22.2/backups/feature-availability-note.md
kathancox: I really like the Supported products idea and so I have added this in. After discussing with Liv yesterday, it was clear that customers are confused by which product can do what. And so until our IA in general directs users this way, I think this heading provides the clarity needed.

PR #16321, opened 2023-02-22
Add example of using dates in ttl expressions

ianjevans:
Adds an example of using `ttl_expiration_expression` with a `DATE` column and explicitly setting the timezone.

Closes DOC-6918.

### DIFF
--- a/v23.1/row-level-ttl.md
+++ b/v23.1/row-level-ttl.md
@@ -212,7 +212,7 @@ CREATE TABLE events (
 );
 ~~~
 
-A `ttl_expiration_expression` that uses an existing `TIMESTAMPTZ` column.
+A `ttl_expiration_expression` that uses an existing `TIMESTAMPTZ` column:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
@@ -226,7 +226,7 @@ CREATE TABLE events (
 );
 ~~~
 
-When using a `ttl_expiration_expression` on a `DATE` or `TIMESTAMPTTZ` column use `AT TIME ZONE` to explicitly set the timezone for the expression. By setting the timezone to UTC in the expression, you set an exact time when the delete should be performed, regardless of the local timezone of the node.
+When using a `ttl_expiration_expression` on a `DATE` or `TIMESTAMPTTZ` column, use `AT TIME ZONE` to explicitly set the time zone for the expression. By setting the time zone to UTC in the expression, you set an exact time when the delete should be performed, regardless of the local time zone of the node.
 
 ### Create a table with `ttl_expire_after`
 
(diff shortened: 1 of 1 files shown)


fabiog1901:
Hi Ian, I don't see the example on TIMESTAMPTZ as mentioned in the ticket I raised. it's good we have now an example for DATE, but I think it's far more important to have the example for TIMESTAMPTZ.

ianjevans:
> Hi Ian, I don't see the example on TIMESTAMPTZ as mentioned in the ticket I raised. it's good we have now an example for DATE, but I think it's far more important to have the example for TIMESTAMPTZ.

Ok, I added back the `TIMESTAMPTZ` example to the same section, since they accomplish the same things. Rafi had said that ideally the column would be a `DATE`.

rafiss:
i agree with having the TIMESTAMPTZ example too

v23.1/row-level-ttl.md
```diff
@@ -194,6 +194,40 @@ The `ttl_expiration_expression` parameter has the following requirements:
...
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  description TEXT,
+  start_date DATE DEFAULT now() NOT NULL,
+  end_date DATE NOT NULL
+) WITH (
+  ttl_expiration_expression = '((end_date::TIMESTAMP) + INTERVAL ''30 days'') AT TIME ZONE ''UTC'''
+);
+~~~
+
+A `ttl_expiration_expression` that uses an existing `TIMESTAMPTZ` column.
```
gemma-shay, line 215:
```suggestion
A `ttl_expiration_expression` that uses an existing `TIMESTAMPTZ` column:
```

v23.1/row-level-ttl.md
```diff
@@ -194,6 +194,40 @@ The `ttl_expiration_expression` parameter has the following requirements:
...
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  description TEXT,
+  start_date TIMESTAMPTZ DEFAULT now() NOT NULL,
+  end_date TIMESTAMPTZ NOT NULL
+) WITH (
+  ttl_expiration_expression = '((end_date AT TIME ZONE ''UTC'') + INTERVAL ''30 days'') AT TIME ZONE ''UTC'''
+);
+~~~
+
+When using a `ttl_expiration_expression` on a `DATE` or `TIMESTAMPTTZ` column use `AT TIME ZONE` to explicitly set the timezone for the expression. By setting the timezone to UTC in the expression, you set an exact time when the delete should be performed, regardless of the local timezone of the node.
```
gemma-shay, line 229:
```suggestion
When using a `ttl_expiration_expression` on a `DATE` or `TIMESTAMPTTZ` column, use `AT TIME ZONE` to explicitly set the time zone for the expression. By setting the time zone to UTC in the expression, you set an exact time when the delete should be performed, regardless of the local time zone of the node.
```

gemma-shay (approved):
Just nits, lgtm

PR #16379, opened 2023-02-28
Make ~all mentions of "contention" link to docs

rmloveland:
Preliminary work addressing DOC-6663

Summary of changes:

- Wide-ranging "housekeeping" PR that makes ~all places where 'contention' is found as a bare word into *links* to the contention info on the 'SQL Performance Best Practices' page

- Adds more specific links to cluster settings useful for contention debugging / monitoring

- Makes many mentions of 'retry errors' into sentences that use the searchable strings '`40001` error code' and `restart transaction`, which in turn link to the 'Common Errors' page (which in turn links to the 'Transaction Retry Error Reference')

- Adds links to various contention-related tables in `crdb_internal`

- Adds links to `crdb_internal.cluster_locks` in some appropriate places

- Adds more links to the 'Hot Spots' section of the 'SQL Performance Best Practices' page, e.g. from the `ALTER TABLE ... SPLIT AT` docs

- Makes the 'Indicators your application is experiencing transaction contention' section from 'Performance Recipes' also available on the 'SQL Performance Best Practices' page (via an include)

- Applies to the following versions:

  - v23.1 (including things in _includes/ folder; cockroachcloud/ was checked and did not have any instances of the term 'contention')


v23.1/admission-control.md
```diff
@@ -5,15 +5,15 @@ toc: true
...
+CockroachDB supports an admission control system to maintain cluster performance and availability when some nodes experience high load. When admission control is enabled, CockroachDB sorts request and response operations into work queues by priority, giving preference to higher priority operations. Internal operations critical to node health, like [node liveness heartbeats](cluster-setup-troubleshooting.html#node-liveness-issues), are high priority. The admission control system also prioritizes transactions that hold [locks](architecture/transaction-layer.html#writing), to reduce [contention](performance-best-practices-overview.html#transaction-contention) by releasing [locks](crdb-internal.html#cluster_locks) in a timely manner.
```
AlexTalks, line 8: Are the three separate links here at all confusing?

v23.1/admission-control.md
```diff
@@ -44,7 +44,7 @@ Admission control is enabled  by default for all layers.
...
+Higher priority operations are processed first. The criteria for determining higher and lower priority operations is different at each processing layer, and is determined by the CPU and storage I/O of the operation. Write operations in the [KV storage layer](architecture/storage-layer.html) in particular are often the cause of performance bottlenecks, and admission control prevents [the Pebble storage engine](architecture/storage-layer.html#pebble) from experiencing high [read amplification](architecture/storage-layer.html#read-amplification). Critical cluster operations like node heartbeats are processed as high priority, as are transactions that hold locks in order to avoid [contention](performance-recipes.html#transaction-contention) by releasing [locks](architecture/transaction-layer.html#writing).
```
AlexTalks, line 47: s/"by releasing locks"/"and release locks earlier".

AlexTalks (approved):
As mentioned when we discussed, I think these changes are a strict improvement on what we currently have.  I do think it would be useful to review these docs going forward and attempt to clarify at each point what we mean by "contention", as it's a broad term.  **Lock contention**, which is specifically two transactions vying for (i.e. conflicting on) the same lock, is a bit different to **serializability conflicts**, which have to do with two transactions conflicting in their orders of operations performed on overlapping read-write sets.  Both are types of **transaction conflicts**, and may fit under the broad umbrella of the term "contention", however it will be important to distinguish these types of transaction conflict and their behavior going forward.

v23.1/admission-control.md
rmloveland: Yes probably.  I've updated this to make only the first mention of "locks" a link to https://www.cockroachlabs.com/docs/v22.1/crdb-internal.html#cluster_locks.  My thinking for linking to that page is that it may be more practical for most users, and it itself links to the architecture docs on locks at https://www.cockroachlabs.com/docs/v22.1/architecture/transaction-layer#concurrency-control

v23.1/admission-control.md
rmloveland: Updated (in both places) to say "and release locks earlier".  Also the 2nd mention of locks is no longer a link, as in my previous comment.

rmloveland:
> As mentioned when we discussed, I think these changes are a strict improvement on what we currently have. I do think it would be useful to review these docs going forward and attempt to clarify at each point what we mean by "contention", as it's a broad term. **Lock contention**, which is specifically two transactions vying for (i.e. conflicting on) the same lock, is a bit different to **serializability conflicts**, which have to do with two transactions conflicting in their orders of operations performed on overlapping read-write sets. Both are types of **transaction conflicts**, and may fit under the broad umbrella of the term "contention", however it will be important to distinguish these types of transaction conflict and their behavior going forward.

Thanks for summarizing and explaining this Alex.  I will be sure to draw this distinction more clearly in [DOC-7148: Update "SQL Performance Best Practices - Transaction Contention" to distinguish lock contention vs. Serialization conflicts/retries](https://cockroachlabs.atlassian.net/browse/DOC-7148)

rmloveland:
@ianjevans FYI this is mostly adding a bunch of links.  If you are happy with the v23.1 changes I will backport to earlier versions

ianjevans (approved):

PR #16319, opened 2023-02-22
DOC-6988 metrics export GTM preview

andf-crl:
Addresses: DOC-6988

- Switches GTM include from `preview-opt-in.md` to `preview.md`, removing "enroll" language.

Staging

[export-metrics.md](https://deploy-preview-16319--cockroachdb-docs.netlify.app/docs/cockroachcloud/export-metrics.html)

PR #16396, opened 2023-03-01
DOC-7102 fix metrics export step 6

andf-crl:
Addresses: DOC-7102

- Corrects AWS step 6 to use standard permissions template from log export.

Staging

[export-metrics.md](https://deploy-preview-16396--cockroachdb-docs.netlify.app/docs/cockroachcloud/export-metrics.html)

### DIFF
--- a/cockroachcloud/export-metrics.md
+++ b/cockroachcloud/export-metrics.md
@@ -11,7 +11,7 @@ cloud: true
 Exporting metrics to AWS CloudWatch is only available on {{ site.data.products.dedicated }} clusters which are hosted on AWS, and were created after August 11, 2022. Metrics export to Datadog is supported on all {{ site.data.products.dedicated }} clusters regardless of creation date.
 
 {{site.data.alerts.callout_info}}
-{% include_cached feature-phases/limited-access.md %}
+{% include_cached feature-phases/preview.md %}
 {{site.data.alerts.end}}
 
 ## The `metricexport` endpoint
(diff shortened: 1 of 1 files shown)


thtruo (commented):
Change in step 6 LGTM!

Before approving, I just noticed the docs in prod is referring to metric export as a "limited access" feature. I thought we had already updated the disclaimer to say "preview" from https://github.com/cockroachdb/docs/pull/16319 a few days ago.

Was that intentional? Let me know if you need me to file another issue or if we can update this within this PR

andf-crl:
@thtruo You are correct, looks like Lauren changed the include I was using behind my back! Staging for the change you reference is still up [here](https://deploy-preview-16319--cockroachdb-docs.netlify.app/docs/cockroachcloud/export-metrics.html). Does that look correct to you as-is? If so I'll restore that deleted include file and get that copy live as well as part of this PR. Good catch!!

thtruo:
@andf-crl ack!

> Staging for the initial state of this include is still up [here](https://deploy-preview-16319--cockroachdb-docs.netlify.app/docs/cockroachcloud/export-metrics.html). Does that look correct to you as-is?

Yes, this looks right!

And to reiterate, step 6 config matches the one mentioned in log export, so this still looks good to me

abarganier:
LGTM, going to run through the steps myself real quick to triple check, then I'll approve 🙂

cockroachcloud/export-metrics.md
```diff
@@ -94,25 +94,26 @@ Perform the following steps to enable metrics export from your {{ site.data.prod
...
-                "Action": "sts:AssumeRole"
-            },
-            {
-                "Effect": "Allow",
-                "Principal": {
-                    "AWS": "arn:aws:iam::{your_aws_acct_id}:root"
-                },
-                "Action": "sts:AssumeRole"
+                "Resource": [
+                    "arn:aws:logs:*:{your_aws_acct_id}:log-group:{log_group_name}:*"
```
abarganier, line 107: I noticed that I received errors with this resource name.

We were able to fix by using this instead:
```
"Resource": [
    "arn:aws:logs:*:{your_aws_acct_id}:log-group:{log_group_name}",
    "arn:aws:logs:*:{your_aws_acct_id}:log-group:{log_group_name}:log-stream:*"
]
```

Once we swapped to this, it worked 😎

abarganier (approved):
Thanks for the fix! As noted in Slack I think we should take some time during stability to overhaul these docs to use the AWS CLI instead.

For now though, let's get the Policy JSON fixed up.

andf-crl:
@thtruo , @abarganier here are the three changes you two requested in [one easy link](https://github.com/cockroachdb/docs/compare/2aaec10f7b8fec997d82198e1f9bdd00403108ed..07c8388771b25c92c891c651dd25aa6945d5b4a8). LMK if you'd like to see changes, thanks!!

abarganier (approved):
LGTM, thanks for the quick turnaround!

andf-crl:
TYFR all!

PR #16413, opened 2023-03-03
You cannot drop a column ref'd in a partial index

rmloveland:
Fixes DOC-7050

Relates to DOC-7029


rmloveland:
Hi @mgartner !  Mind taking a look at this docs PR? it attempts to update the docs based on the limitation described in [TA-96924](https://www.cockroachlabs.com/docs/advisories/a96924), which has been published

Entirely possible there is nuance here that this docs update does not capture, happy to update the PR as needed!

ianjevans (approved):

PR #16421, opened 2023-03-06
Fix structured content in page header

nickvigilante:
We were getting an error because the summary in one page had unescaped double quotes. This should fix that.

image


mikeCRL (approved):
Checked the rendered JSON and it looks good and passes JSON lint. Thanks!

PR #16478, opened 2023-03-10
[DOC-202] Document troubleshooting info about what to do if a node's timezone changes

mdlinville:
[DOC-202] Document troubleshooting info about what to do if a node's timezone changes

## Previews
[v22.2/cluster-setup-troubleshooting.md](https://deploy-preview-16478--cockroachdb-docs.netlify.app/docs/v22.2/cluster-setup-troubleshooting.html)
[v23.1/cluster-setup-troubleshooting.md](https://deploy-preview-16478--cockroachdb-docs.netlify.app/docs/v23.1/cluster-setup-troubleshooting.html)

## Tests
- [x] Local build
- [x] Linkcheck

### DIFF
--- a/src/current/v22.2/cluster-setup-troubleshooting.md
+++ b/src/current/v22.2/cluster-setup-troubleshooting.md
@@ -304,6 +304,34 @@ The solution is to add connection retry logic to your application.
 
 ## Clock sync issues
 
+#### A node's timezone data has changed
+
+CockroachDB relies on each node's operating system timezone database to resolve named timezones like `America/New_York`, and for applying daylight saving rules. This can lead to inconsistent results across a cluster when coercing date/time strings into temporal types.
+
+Ensure that each node's operating system uses the same timezone database. In general, Cockroach Labs recommends that each node has an identical hardware and software configuration.
+
+Timezone data is read from the following sources in order of preference:
+
+1. The `ZONEINFO` operating system environment variable, if it is set to point to a `zoneinfo.zip` file such as [the one included in Go](https://github.com/golang/go/tree/master/lib/time) or to the directory that results from unzipping a `zoneinfo.zip`.
+1. The operating system's timezone database, sometimes called `tzdata` or `zoneinfo`. The following common locations are searched:
+
+      - `/usr/share/zoneinfo/`
+      - `/usr/share/lib/zoneinfo/`
+      - `/usr/lib/locale/TZ/`
+
+    {{site.data.alerts.callout_info}}
+    The Windows operating system's timezone database is not used by CockroachDB on Windows.
+    {{site.data.alerts.end}}
+
+1. The timezone database that is embedded in Go.
+
+[Restart](node-shutdown.html) each node if:
+
+- The operating system's timezone database is updated.
+- Any node's timezone changes.
+- The definition of a timezone changes due to government or geopolitical decisions.
+- The `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
+
 #### Node clocks are not properly synchronized
 
 See the following FAQs:
--- a/src/current/v23.1/cluster-setup-troubleshooting.md
+++ b/src/current/v23.1/cluster-setup-troubleshooting.md
@@ -304,6 +304,34 @@ The solution is to add connection retry logic to your application.
 
 ## Clock sync issues
 
+#### A node's timezone data has changed
+
+CockroachDB relies on each node's operating system timezone database to resolve named timezones like `America/New_York`, and for applying daylight saving rules. This can lead to inconsistent results across a cluster when coercing date/time strings into temporal types.
+
+Ensure that each node's operating system uses the same timezone database. In general, Cockroach Labs recommends that each node has an identical hardware and software configuration.
+
+Timezone data is read from the following sources in order of preference:
+
+1. The `ZONEINFO` operating system environment variable, if it is set to point to a `zoneinfo.zip` file such as [the one included in Go](https://github.com/golang/go/tree/master/lib/time) or to the directory that results from unzipping a `zoneinfo.zip`.
+1. The operating system's timezone database, sometimes called `tzdata` or `zoneinfo`. The following common locations are searched:
+
+      - `/usr/share/zoneinfo/`
+      - `/usr/share/lib/zoneinfo/`
+      - `/usr/lib/locale/TZ/`
+
+    {{site.data.alerts.callout_info}}
+    The Windows operating system's timezone database is not used by CockroachDB on Windows.
+    {{site.data.alerts.end}}
+
+1. The timezone database that is embedded in Go.
+
+[Restart](node-shutdown.html) each node if:
+
+- The operating system's timezone database is updated.
+- Any node's timezone changes.
+- The definition of a timezone changes due to government or geopolitical decisions.
+- The `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
+
 #### Node clocks are not properly synchronized
 
 See the following FAQs:


knz (approved):
Maybe add a note that this is needed even when the operator uses the ZONEINFO env var to point to a zip file. CockroachDB caches the content of the zip file internally and will only refresh it upon restart.

mdlinville:
Added a bit more context about the order of operations and the TZDATA OS environment variable, after discussion with @knz

knz (approved):

mdlinville:
@knz PTAL :)

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
 
 ## Clock sync issues
 
+#### A node's timezone data has changed
```
ianjevans, line 307: Just noticing that these entries are h4, while the parent section is an h2. This may be intended, but it caught my eye.

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
...
+    {{site.data.alerts.end}}
+
+1. The timezone database that is embedded in Go.
+
+In the following situations, [restart](node-shutdown.html) each node:
+
+- If the operating system's timezone database is updated.
+- If any node's timezone changes.
+- If the definition of a timezone changes due to government or geopolitical decisions.
+- If the `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
```
ianjevans, line 329:
```suggestion
[Restart](node-shutdown.html) each node if:

- The operating system's timezone database is updated.
- Any node's timezone changes.
- The definition of a timezone changes due to government or geopolitical decisions.
- The `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
```

v23.1/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
...
+    {{site.data.alerts.end}}
+
+1. The timezone database that is embedded in Go.
+
+In the following situations, [restart](node-shutdown.html) each node:
+
+- If the operating system's timezone database is updated.
+- If any node's timezone changes.
+- If the definition of a timezone changes due to government or geopolitical decisions.
+- If the `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
```
ianjevans, line 329:
```suggestion
[Restart](node-shutdown.html) each node if:

- The operating system's timezone database is updated.
- Any node's timezone changes.
- The definition of a timezone changes due to government or geopolitical decisions.
- The `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
```

ianjevans (approved):
Just some suggestions. Looks good.

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
...
 
+#### A node's timezone data has changed
+
+CockroachDB relies on each node's operating system timezone database to resolve named timezones like `America/New_York`, and for applying daylight saving rules. This can lead to inconsistent results across a cluster when coercing date/time strings into temporal types.
+
+Ensure that each node's operating system uses the same timezone database. In general, Cockroach Labs recommends that each node has an identical hardware and software configuration.
+
+Timezone data is read from the following sources in order of preference:
+
+1. The `ZONEINFO` operating system environment variable, if it is set to point to a `zoneinfo.zip` file such as the one that is included in Go.
```
knz, line 315: ZONEINFO can point either to a zip filr or a directory containing the tzdata. This can thus also be used e.g. on OS variants what have their tzdata in non-standard locations.

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
...
+1. The operating system's timezone database, sometimes called `tzdata` or `zoneinfo`.
```
knz, line 316: To clarify: "The OS's tz database, ..., as found in one of the following standard locations

        "/usr/share/zoneinfo/",
        "/usr/share/lib/zoneinfo/",
        "/usr/lib/locale/TZ/",

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,30 @@ The solution is to add connection retry logic to your application.
...
+
+    {{site.data.alerts.callout_info}}
+    CockroachDB on Windows does not use the operating system's timezone database.
```
knz, line 319: This callout is not exactly correct. on windows there is a tz database, but it has a different format and is relatively incomplete wrt crdb's needs.

v22.2/cluster-setup-troubleshooting.md
mdlinville: Yeah, I didn't want to deal with that in this PR. I suspect it might have been to suppress them in the in-page TOC.

v22.2/cluster-setup-troubleshooting.md
mdlinville: So where do we get our ZONEINFO in CRDB for Windows? Do we require the user to download a `zoneinfo.zip` or do we include it in the Windows installer or something else?

knz: We raise our hands up in the air and let them figure it out, because we don't really support windows anyways...

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,34 @@ The solution is to add connection retry logic to your application.
...
 
+#### A node's timezone data has changed
+
+CockroachDB relies on each node's operating system timezone database to resolve named timezones like `America/New_York`, and for applying daylight saving rules. This can lead to inconsistent results across a cluster when coercing date/time strings into temporal types.
+
+Ensure that each node's operating system uses the same timezone database. In general, Cockroach Labs recommends that each node has an identical hardware and software configuration.
+
+Timezone data is read from the following sources in order of preference:
+
+1. The `ZONEINFO` operating system environment variable, if it is set to point to a `zoneinfo.zip` file such as [the one included in Go](https://github.com/golang/go/tree/master/lib/time) or to the directory that results from unzipping a `zoneinfo.zip`.
```
knz, line 315: This phrasing makes the reader think that it's Go who decides the format. That's not the right sentiment.

The format is standardized - it's called "tzdata" and it's a file hierarchy where all systems (except windows) do the same. The Go runtime supports loading this data *either from a directory* which is the most common way or from a zip file -- with the zip simply containing a copy of the tzdata in the standard format.

Here are the most common scenarios, **from the end user's perspective** and in decreasing order of commonality:

1. the tzdata is in one of the standard locations (/usr/share/zoneinfo etc). In this case they have nothing to do.
2. the tzdata is provided by the OS or the container base image, in a non-standard directory. In this case they set ZONEINFO to point to that directory and call it a day.
3. they are building a custom container (via their own Dockerfile or whatnot) and *they didn't know they needed a copy of tzdata*. In this case, we need to remind them that they should include "tzdata installation" as one of the build steps for their container image, usually via the base image package manager. Otherwise things won't work.
4. the user is working on a size-restricted environment (e.g. small containers or whatnot) and they want a "smaller tzdata database" that's just sufficient for the needs of their apps. In this case, they can take a copy of the tzdata on a standard system; trim it down to just what they need; put the result in a  zip file, then make the crdb deployment use that via the ZONEINFO env var.

Of course use case (1) and (2) are far more common than (3). And case (4) is just exotic.

Finally, there's the windows case. We just needs to tell them that:
- on windows, go tries to emulate tzdata on a "best effort" basis using the tz information in the windows registry. There may be discrepancies with unix systems. 
- if they really want the same tzdata as on unix, the user can take a copy of the tzdata directory on a unix system, put that copy on windows, and point to it via the ZONEINFO env var. They can either copy the full directory with all the files, or just a subset, or perhaps bundle all this in a zip file. We don't have an opinion.

v22.2/cluster-setup-troubleshooting.md
```diff
@@ -304,6 +304,34 @@ The solution is to add connection retry logic to your application.
...
+    {{site.data.alerts.end}}
+
+1. The timezone database that is embedded in Go.
+
+[Restart](node-shutdown.html) each node if:
+
+- The operating system's timezone database is updated.
+- Any node's timezone changes.
+- The definition of a timezone changes due to government or geopolitical decisions.
+- The `ZONEINFO` operating system environment variable is updated or if the file it points to is updated.
```
knz, line 333: ... if the file(s) it points to is/are updated ...

(it can point to a directory)

PR #16481, opened 2023-03-13
IMPORT, schema changes don't use protected ts

rmloveland:
Fixes DOC-5193

Relates to https://github.com/cockroachdb/cockroach/pull/68043


adityamaru (approved):
Thanks!

stevendanna:
I believe some parts of schema changes in 23.1 do technically use protected timestamps.

adityamaru:
🤦 I found all usages of `Protect` when reviewing this but I forgot that they switched over to using a different jobs PTS manager. Apologies @rmloveland I believe backfill schema changes still use PTS records.

rmloveland:
Thanks @stevendanna and @adityamaru for pointing this out!

I have filed DOC-7465 to update this to add back schema changes to the list of things that use protected ts

@stevendanna do you think simply adding the link back to the list (that was removed in this PR) would suffice?

PR #16488, opened 2023-03-14
Fix SSO description

msmygit:
Fix SSO description


nickvigilante:
@msmygit Thank you for your contribution! Looks like we already incorporated this as part of https://github.com/cockroachdb/docs/pull/16838.

PR #16504, opened 2023-03-16
Clarify locality optimized search

mgartner:
I found the _locality optimized search_ docs a bit too brief and slightly misleading, so here are some possible edits to make. Feedback is appreciated. I can apply the changes to the other versions once it looks good. Also, if it's easier for someone else to take this over, that's fine be me too.

### DIFF
--- a/v23.1/cost-based-optimizer.md
+++ b/v23.1/cost-based-optimizer.md
@@ -250,7 +250,7 @@ You can instruct the optimizer to use indexes marked as `NOT VISIBLE` with the [
 
 ## Locality optimized search in multi-region clusters
 
-In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), will avoid sending requests to nodes in other regions when it can instead read a value from a unique column that is stored locally. This capability is known as _locality optimized search_.
+In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), may perform a _locality optimized search_ to attempt to avoid high latency, cross-region communication between nodes. If query results could be produced using only region-local rows, the database will first search for rows in the gateway node's region. The search only continues in remote regions if rows in the local region did not satisfy the query.  Examples of queries that can use _localtiy optimized search_ include unique key lookups and queries with `LIMIT` clauses.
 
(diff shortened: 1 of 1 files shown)


v23.1/cost-based-optimizer.md
```diff
@@ -250,7 +250,7 @@ You can instruct the optimizer to use indexes marked as `NOT VISIBLE` with the [
 
 ## Locality optimized search in multi-region clusters
 
-In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), will avoid sending requests to nodes in other regions when it can instead read a value from a unique column that is stored locally. This capability is known as _locality optimized search_.
+In [multi-region deployments](multiregion-overview.html), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), may perform a _locality optimized search_ to attempt to avoid high latency, cross-region communication between nodes. If query results could be produced using only region-local rows, the database will first search for rows in the gateway node's region. The search only continues in remote regions if rows in the local region did not satisfy the query.  Examples of queries that can use _localtiy optimized search_ include unique key lookups and queries with `LIMIT` clauses.
```
taroface, line 253:
```suggestion
In multi-region deployments with [regional by row tables](multiregion-overview.html#regional-by-row-tables), the optimizer, in concert with the [SQL engine](architecture/sql-layer.html), may perform a *locality optimized search* to attempt to avoid high-latency, cross-region communication between nodes. If a query can be satisfied by reading rows in the local region, the database will first search for rows in the gateway node's region. The search only continues in remote regions if rows in the local region did not satisfy the query.  Examples of queries that can use locality optimized search include unique key lookups and queries with [`LIMIT`](limit-offset.html) clauses.
```

taroface (changes requested):
Thank you for opening this PR! I made some light suggestions based on my current understanding, but it's possible I made things wrong again by rephrasing.

After we finalize this copy, can you also reproduce the change on the v22.2 doc, since that is the currently visible set?

v23.1/cost-based-optimizer.md
taroface: OK, I thought that this might be your intent with that phrase (and I wanted to be optimistic that the situation wasn't as nuanced). In this case I think it's fine to use more words, and what you wrote here works well for that line:

```
If there is a possibility that the results of a query all live in local rows, the database will first search for rows in the gateway node's region. 
```

taroface: @mgartner Just following up on this. Does the above phrasing sound OK to you? If so, I'm happy to take over the PR if you prefer.

taroface:
Closing in favor of #16713.

PR #16506, opened 2023-03-16
Update all GC TTL and range size mentions to DRY

rmloveland:
... where DRY = Don't Repeat Yourself

Addresses:

DOC-6539
DOC-6540
DOC-6820
DOC-6834

Summary of changes:

- Update all mentions in v23.1 docs of the default GC TTL (`gc.ttlseconds`) and default/max range size (`range_max_bytes`) to point to their respective entries on the 'Configure Replication Zones' page.

- Update the description of the `gc.ttlseconds` zcfg variable to link to more salient info re: MVCC, range splits, etc. to provide the reader with add'l context

- Make a few small text updates to weave in the above, as well as a few opportunistic link additions (to range splitting docs, for example)


rmloveland:
@irfansharif added you since my understanding is you are the primary engineer on the change to the GC TTL default for v23.1 based on https://github.com/cockroachdb/cockroach/issues/89233, https://github.com/cockroachdb/cockroach/pull/93836, and your filing of https://github.com/cockroachdb/cockroach/issues/98105

@kathancox adding you as well per our discussion earlier today.  I suspect some of these changes will be obsoleted by your work on DOC-1012, DOC-6654, and friends, but hopefully this PR is not too disruptive b/c the changes are small and will ensure consistent pointing to one place for the GC TTL default values (even if for a short time because your subsequent work will be rewriting some of this stuff Very Soon (tm))

rmloveland:
also @irfansharif there will almost certainly be more doc updates to come re: the new GC TTL

this is a first pass at just pointing users to one place for the updated defaults and making sure when they get to that one place we have more cross-linking to concepts like MVCC and range splits they need to understand if they are going to be changing this setting

v23.1/delete-data.md
```diff
@@ -220,7 +220,7 @@ with conn.cursor() as cur:
...
+Because of the way CockroachDB works under the hood, deleting data from the database does not immediately reduce disk usage.  Instead, records are marked as "deleted" and processed asynchronously by a background garbage collection process.  This process runs [at a specified interval](configure-replication-zones.html#gc-ttlseconds) to allow sufficient time for running [backups](take-full-and-incremental-backups.html) and running [time travel queries using `AS OF SYSTEM TIME`](as-of-system-time.html).  The garbage collection interval is controlled by the [`gc.ttlseconds`](configure-replication-zones.html#gc-ttlseconds) setting.
```
irfansharif, line 223: I think this was+is incorrect. The garbage collection process is running continuously, not every 25hrs or some controllable interval. It's more that data outside the configured window is _eligible_ to get garbage collected, but we don't go look for it every 25hrs.

v23.1/backup.md
```diff
@@ -81,7 +81,7 @@ CockroachDB stores full backups in a backup collection. Each full backup in a co
 `LATEST` | Append an incremental backup to the latest completed full backup's subdirectory.
 <a name="collectionURI-param"></a> `collectionURI` | The URI where you want to store the backup. (Or, the default locality for a locality-aware backup.)<br/><br/>For information about this URL structure, see [Backup File URLs](#backup-file-urls).
 `localityURI`   | The URI containing the `COCKROACH_LOCALITY` parameter for a non-default locality that is part of a single locality-aware backup.
-`timestamp` | Back up data as it existed as of [`timestamp`](as-of-system-time.html). The `timestamp` must be more recent than your cluster's last garbage collection (which defaults to occur every 25 hours, but is [configurable per table](configure-replication-zones.html#replication-zone-variables)).
+`timestamp` | Back up data as it existed as of [`timestamp`](as-of-system-time.html). The `timestamp` must be more recent than your cluster's last garbage collection (which occurs at a [configurable interval](configure-replication-zones.html#gc-ttlseconds)).
```
irfansharif, line 84: The same comment as below. The GC TTL does not control the interval during which we're collecting garbage, it controls how old an over-written data has to be _at least_ before it's considered for garbage collection. The garbage collection itself is actually happening continuously.

_includes/v23.1/zone-configs/variables.md
```diff
@@ -1,8 +1,8 @@
...
...EM TIME`](as-of-system-time.html) queries, also known as [Time Travel Queries](select-clause.html#select-historical-data-time-travel).<br><br>It is not recommended to set this below `600` (10 minutes); doing so will cause problems for [long-running queries](manage-long-running-queries.html). Also, since all versions of a row are stored in a single range that never splits, it is not recommended to set this so high that all the changes to a row in that time period could add up to more than [the default range size](#range-max-bytes); such oversized ranges could contribute to the server [running out of memory](cluster-setup-troubleshooting.html#out-of-memory-oom-crash) or other problems. {{site.data.alerts.callout_info}} Ensure that you set `gc.ttlseconds` long enough to accommodate your [backup schedule](create-schedule-for-backup.html), otherwise your incremental backups will fail with [the error message `protected ts verification error`](common-errors.html#protected-ts-verification-error). For example, if you set up your backup schedule to recur at an interval greater than `gc.ttlseconds`, all your incremental backups will fail.{{site.data.alerts.end}} **Default:** `14400` (4 hours)
```
irfansharif, line 5: Re: "Ensure that you set `gc.ttlseconds` long enough to accommodate your..." this is now less applicable. GC TTL now has nothing to do with _scheduled_ backups since scheduled backups with revision history will automatically prevent garbage collection of relevant data regardless of gc.ttlseconds. If the scheduled backups don't use revision history, whether we GC overwritten data or not is irrelevant, since it does not appear in the backup.

The whole "keep your GC TTL large enough to accommodate your backups" only applies if you're not using scheduled backups, and are effectively issuing new backup statements yourself every N hours. If you're doing that, and also want revision history, then you do need to keep your GC TTL large enough. But our guidance is to prefer using scheduled backups.

Do we have other references to this "coupling" between backups and the GC TTL?

irfansharif (approved):
This is great! Thanks for setting the bed for the forthcoming GC TTL specific documentation. It LGTM, but I noticed some out-of-date or incorrect information that I've commented on below. Those corrections are unrelated to what this PR itself is changing, but appeared adjacent to text this PR changed. So it caught my eye.

_includes/v23.1/zone-configs/variables.md
kathancox: This link might be helpful here, re coupling of non-scheduled backups to GC TTL https://www.cockroachlabs.com/docs/v22.2/create-schedule-for-backup#protected-timestamps-and-scheduled-backups

kathancox (approved):
From a docs pov, LGTM! I added a potentially helpful link re Irfan's comment.

v23.1/backup.md
rmloveland: how about

> The `timestamp` must be more recent than your cluster's garbage collection TTL (which is controlled by the [`gc.ttlseconds` replication zone variable](configure-replication-zones.html#gc-ttlseconds)).

I think that does not quite capture the nuance where something could potentially be older than `timestamp` but still work kinda by accident because something hasn't been GC'd yet but is eligible to.  But I think it might be better to just say the timestamp must be more recent.  What do you think?

rmloveland: PS will port this language change to all versions v21.2+ as mentioned in another comment

_includes/v23.1/zone-configs/variables.md
rmloveland: > Re: "Ensure that you set `gc.ttlseconds` long enough to accommodate your..." this is now less applicable. GC TTL now has nothing to do with _scheduled_ backups since scheduled backups with revision history will automatically prevent garbage collection of relevant data regardless of gc.ttlseconds. If the scheduled backups don't use revision history, whether we GC overwritten data or not is irrelevant, since it does not appear in the backup.

> The whole "keep your GC TTL large enough to accommodate your backups" only applies if you're not using scheduled backups, and are effectively issuing new backup statements yourself every N hours. If you're doing that, and also want revision history, then you do need to keep your GC TTL large enough. But our guidance is to prefer using scheduled backups.

> Do we have other references to this "coupling" between backups and the GC TTL?



> This link might be helpful here, re coupling of non-scheduled backups to GC TTL https://www.cockroachlabs.com/docs/v22.2/create-schedule-for-backup#protected-timestamps-and-scheduled-backups

Awesome, thanks Kathryn - I've linked to this in the revised text, PTAL

rmloveland:
Thanks for the reviews @irfansharif and @kathancox !  I have made updates based on your comments.  Please let me know if I got them approximately right, or if further changes are needed

v23.1/backup.md
irfansharif: That works better, thanks. If we're being pedantic, I would change "cluster's garbage collection TTL" to "data's garbage collection TTL" since gc.ttlseconds applies over data, like specific schemas, databases, or "range default", and not the cluster.

_includes/v23.1/zone-configs/variables.md
irfansharif: This part is not correct:

```
For example, if you set up your backup schedule to recur at an interval greater than `gc.ttlseconds`, all your incremental backups will fail.
```

What instead happens now is that if you're using scheduled backups with revision history + an interval greater than the GC TTL, we'll effectively only GC data once it's been backed up, independent of the GC TTL. These backups "chain together" because we know the full schedule, and prevent GC of revisiones you wanted to have backed up. So backups will not fail. They used in in some releases older than 22.2, the chaining was enabled by default in 22.2: https://github.com/cockroachdb/cockroach/pull/79976.

v23.1/backup.md
rmloveland: Yes let's be pedantic!  I have updated all versions v21.2+ to say "cluster's garbage collection TTL". Thank you!

_includes/v23.1/zone-configs/variables.md
rmloveland: I have removed that sentence altogether in the latest update (to both v22.2 and v23.1 docs).  That section now reads:

> If you are not yet using [scheduled backups](create-schedule-for-backup.html), and instead issuing [`BACKUP`](backup.html) statements manually, you must ensure that you set `gc.ttlseconds` long enough to accommodate your manual backup schedule. Otherwise, your incremental backups will fail with [the error message `protected ts verification error`](common-errors.html#protected-ts-verification-error). We recommend using [scheduled backups](create-schedule-for-backup.html) instead, which automatically [use protected timestamps](create-schedule-for-backup.html#protected-timestamps-and-scheduled-backups) to ensure they succeed.

what do you think?

irfansharif: That looks great!

ianjevans (approved):
LGTM!

PR #16556, opened 2023-03-22
adding miss "," in log configuration

lin-crl:
The configuration is missing a "," before `min_level`
```
             log_name: "devops",
             channels: ["OPS", "HEALTH", "STORAGE"]
             min_level: "WARNING"
```


andf-crl:
Thanks @lin-crl , I caught this the same time you did: the fix will be in https://github.com/cockroachdb/docs/pull/16569. Closing this PR as that one has this same fix. Good notice though!!

PR #16571, opened 2023-03-22
Add external-io flag error for implicit access to AWS

kathancox:
Fixes DOC-7236

This adds the note on the external-io-implicit flag to the AWS S3 section of the Cloud Storage Auth page.

**Note**: We do not yet have Azure implicit and so this is not yet relevant to that section.


taroface (approved):
LGTM! Thank you.

kathancox:
Thanks for catching this @taroface!

PR #16632, opened 2023-03-29
TA 98779

kathancox:
Restore TA 

## Preview

https://deploy-preview-16632--cockroachdb-docs.netlify.app/docs/advisories/a98779

### DIFF
--- a/advisories/a98779.md
+++ b/advisories/a98779.md
@@ -1,0 +1,47 @@
+---
+title: Technical Advisory 98779
+advisory: A-98779
+summary: A restore job can potentially skip some data files upon resumption of an in-progress <code>RESTORE</code>, which could lead to missing rows after the job succeeds.
+toc: true
+affected_versions: v22.2.6
+advisory_date: 2023-03-29
+docs_area: releases
+---
+
+Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
+
+## Description
+
+A bug was introduced in CockroachDB v22.2.6 that can cause a restore job to skip some data files upon resumption of an in-progress [`RESTORE`](../{{site.versions["stable"]}}/restore.html). This could potentially lead to some missing rows after the restore job succeeds.
+
+A resumption could occur due to manual user action, such as [pausing](../{{site.versions["stable"]}}/pause-job.html)  and [resuming](../{{site.versions["stable"]}}/resume-job.html) the job, or, if a job is automatically retried internally, such as after a node restart. Therefore, this failure might not be noticed by an operator.
+
+## Statement
+
+This is resolved in CockroachDB by [#99066](https://github.com/cockroachdb/cockroach/pull/99066), which removes the incorrect optimization, and reverts back to the behavior of v22.2.5.
+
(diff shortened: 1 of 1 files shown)


shermanCRL (approved):

advisories/a98779.md
```diff
@@ -0,0 +1,47 @@
...
+
+A bug was introduced in CockroachDB v22.2.6 that can cause a restore job to skip some data files upon resumption of an in-progress [`RESTORE`](../{{site.versions["stable"]}}/restore.html). This could potentially lead to some missing rows after the restore job succeeds.
+
+A resumption could occur due to manual user action, such as [pausing](../{{site.versions["stable"]}}/pause-job.html)  and [resuming](../{{site.versions["stable"]}}/resume-job.html) the job, or, if a job is automatically retried internally, such as after a node restart. Therefore, this failure might not be noticed by an operator.
+
+## Statement
+
+This is resolved in CockroachDB by [#99066](https://github.com/cockroachdb/cockroach/pull/99066), which removes the incorrect optimization, and reverts back to the behavior of v22.2.5.
+
+The fix has been applied to maintenance releases of CockroachDB [v22.2.7](../releases/v22.2.html#v22-2-7) and later. This problem can also be mitigated immediately by updating the `SET CLUSTER SETTING bulkio.restore.use_simple_import_spans` cluster setting. See the following [Mitigation](#mitigation) section for details.
```
andf-crl, line 23: Ah, since we give the full `SET CLUSTER SETTING` example below, I think we should simplify this just to:

```
by updating the `bulkio.restore.use_simple_import_spans` cluster setting.
```

(I would link CS but ... it's [not there](https://www.cockroachlabs.com/docs/dev/cluster-settings.html)! This is maintained by eng: `cockroachdb/cockroach/{{ page.release_info.crdb_branch_name }}/docs/generated/settings/settings.html` )

It's technically not correct as is without the value to set the CS to: `true`, so let's save the full command for the `## Mitigation` section then.

advisories/a98779.md
```diff
@@ -0,0 +1,47 @@
...
+SET CLUSTER SETTING bulkio.restore.use_simple_import_spans = true;
+```
+
+This setting has already been automatically applied to all affected {{ site.data.products.db }} clusters.
+
+## Impact
+
+For users of CockroachDB v22.2.6, a restore that is resumed for any reason may give incorrect results. The resume could be due to manual job control actions, automatic retries if nodes restart, or after some types of internal rebalancing events.
+
+Enabling the cluster setting described in the [Mitigation](#mitigation) section, as well as upgrading to a later or earlier patch version, will prevent this.
```
andf-crl, line 44: nit: `as well as` --> `or`

(two options, only one of which need be pursued)

nit: `or earlier` seems out of place. Do we support downgrades? Changing the cluster setting seems way less intrusive than a cluster downgrade to get around this issue. Recommend (unless engineering really pushed for this):

```
... upgrading to v22.2.6 or later, will prevent this issue.
```

andf-crl (approved):
LGTM, just two small language nits. Thanks!

advisories/a98779.md
kathancox: Good point, I was only meant to copy over the cluster setting, so I have removed the errant `SET CLUSTER SETTING` from the top section.

advisories/a98779.md
kathancox: Yes, checked in with eng, we're remove downgrade. Thanks for pointing this out @andf-crl !

PR #16634, opened 2023-03-29
Update AWS recommended production guidelines to include AWS Placement Groups

sean-:


srosenberg:
I am a bit worried about bit overly prescriptive. Placement Groups don't make any hard guarantees, and they come with some limitations. Maybe we could just reference AWS’ best practices [1] instead?

Also, it's not something we could test exhaustively, given our limited resources. Thus, deferring to best practices published by a trusted third-party seems like a reasonable thing to do.

[1] https://docs.aws.amazon.com/parallelcluster/latest/ug/best-practices.html

sean-:
@srosenberg What guarantees are you worried Placement Groups don't offer when `strategy=partition`?  Empirically, customers using Placement Groups have validated that this primitive works as advertised due to their lack of correlated failures within the same cluster (e.g., two clusters share similar hardware, such as an EBS storage rack, but only suffer a localized outage within their respective clusters).  Or, TL;DR: I don't think we need to be in the business of validating the functionality that AWS advertises on their tin.  ?

srosenberg:
The part that seems worrisome is the added operational risk which may result in (temporary) unavailability due to a capacity issue that's outside of the customer's control.  From [1],

> If you try to add more instances to the placement group later, or if you try to launch more than one instance type in the placement group, you increase your chances of getting an insufficient capacity error.

> If you stop an instance in a placement group and then start it again, it still runs in the placement group. However, the start fails if there isn't enough capacity for the instance.

> If you receive a capacity error when launching an instance in a placement group that already has running instances, stop and start all of the instances in the placement group, and try the launch again. Starting the instances may migrate them to hardware that has capacity for all of the requested instances.

If you think that's not our concern, then I suppose this guidance is ok.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-groups.html#placement-groups-strategies

srosenberg (approved):

sean-:
In talking with AWS, potential capacity concerns are something they can address via support, and our responsibility is to point customers in the right direction to maximize cluster resiliency and availability.  Thank you for the review and discussion!

PR #16652, opened 2023-03-31
Clarify that EXPORT doesn't require AOST

ianjevans:
Fixes DOC-7320.

### DIFF
--- a/v22.1/export.md
+++ b/v22.1/export.md
@@ -113,6 +113,10 @@ CockroachDB types map to [Parquet types](https://github.com/apache/parquet-forma
 | [`TIMESTAMPTZ`](timestamp.html) | byte array | `STRING` <br>Note: exporting to microsecond precision. |
 | [`ARRAY`](array.html) | Encoded as a repeated field; <br>each array value is encoded as per the preceding types. | `nil` |
 
+## Exports and `AS OF SYSTEM TIME`
+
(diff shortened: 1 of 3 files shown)


fabiog1901 (approved):

v22.1/export.md
```diff
@@ -113,6 +113,10 @@ CockroachDB types map to [Parquet types](https://github.com/apache/parquet-forma
 | [`TIMESTAMPTZ`](timestamp.html) | byte array | `STRING` <br>Note: exporting to microsecond precision. |
 | [`ARRAY`](array.html) | Encoded as a repeated field; <br>each array value is encoded as per the preceding types. | `nil` |
 
+## Exports and `AS OF SYSTEM TIME`
+
+It is not necessary to use an [`AS OF SYSTEM TIME`](as-of-system-time.html) clause in `EXPORT` statements, even though they are long-running queries. When the `EXPORT` statement is executed there is an implied `AS OF SYSTEM TIME`: the start of the statement's execution. The risk of [contention](performance-best-practices-overview.html#transaction-contention) is low because potentially conflicting transactions would need an effective `AS OF SYSTEM TIME` of the exact same time as the `EXPORT` statement's start time.
```
mdlinville, line 118:
```suggestion
The [`AS OF SYSTEM TIME`](as-of-system-time.html) clause is not required in `EXPORT` statements, even though they are long-running queries. When the `EXPORT` statement is executed. If it is omitted, `AS OF SYSTEM TIME` is implicitly set to the start of the statement's execution. The risk of [contention](performance-best-practices-overview.html#transaction-contention) is low because transactions would need to have exactly the same `AS OF SYSTEM TIME` value as the `EXPORT` statement's start time.
```

mdlinville (approved):
Take a look at my suggested change. I think it might be a bit easier to parse.

v22.1/export.md
ianjevans: Thanks, Matt. I'm going to incorporate your suggestion, but there's a stray fragment so it'll be a separate commit.

mdlinville (approved):

PR #16658, opened 2023-04-03
DOC-7358 Add tabset guidance

andf-crl:
Addresses: DOC-7358

- Added guidance on using tabsets, linking into them from without of them, synchronizing state across tabsets, etc.

Original PR: https://github.com/cockroachdb/docs/pull/14940

Staging

[StyleGuide.md](https://deploy-preview-16658--cockroachdb-docs.netlify.app/docs/StyleGuide.md)

PR #16659, opened 2023-04-03
Update accept.txt

andf-crl:
Added recommended vocab entries from https://github.com/cockroachdb/docs/pull/16658. Thanks Nick!

### DIFF
--- a/vale/Vocab/CockroachDB/accept.txt
+++ b/vale/Vocab/CockroachDB/accept.txt
@@ -389,7 +389,7 @@ Sysbench
 Sysdig
 Syslog
 tablespaces?
-tabset[s]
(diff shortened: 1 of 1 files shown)


vale/Vocab/CockroachDB/accept.txt
```diff
@@ -277,6 +277,7 @@ preloaded
 preparable
 prepend(ed|ing|s)
 Prisma
+procedural[s]
```
nickvigilante, line 280: My brain has a case of the Mondays. I definitely made a bad suggestion here. Sorry! Use this instead:

```suggestion
procedurals?
```

nickvigilante (approved):

PR #16660, opened 2023-04-03
DOC-7356: Add more strict checking of nondescript link titles

nickvigilante:
Fixes DOC-7356


andf-crl (approved):
Yesss, approve of this effort, great call!

PR #16678, opened 2023-04-04
Add video to Online Schema Changes

lnhsingh:
Closes DOC-7361

### DIFF
--- a/_includes/youtube.html
+++ b/_includes/youtube.html
@@ -1,0 +1,13 @@
+{% comment %}
+This include is meant to embed YouTube videos within pages in the docs.
+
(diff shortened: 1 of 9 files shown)


mikeCRL (approved):
The video isn't loading for me in the preview. On Chrome, I get a gray box with an icon in it; when I hover, it adds: "www.youtube.com refused to connect." On Safari, I just get a blank area where the video should be. (Same on your end?)

I wonder if this is only the case in the preview. I'd merge this to find out, and then inform Nick, but we could do it the other way around (or just try a different embed format first, if you want?)

nickvigilante (approved):

PR #16689, opened 2023-04-05
fix wrong format within table

eiDear:
OCSP should be inclued in Authentication row, but is is not actually. So modify attribute rowspan from 7 to 8 in table row ```<td rowspan="8"><a href="authentication.html">Authentication</a></td>```.


nickvigilante:
@eiDear Thank you for your contribution! We incorporated this change as part of #17209.

PR #16700, opened 2023-04-06
Quickstart with Cloud Console

gemma-shay:
Change the pared down quickstart from ccloud-only to console-only due to multiple bugs coming up with ccloud in the past weeks

Related bug: https://cockroachlabs.atlassian.net/browse/CC-9799

### DIFF
--- a/cockroachcloud/quickstart.md
+++ b/cockroachcloud/quickstart.md
@@ -7,117 +7,58 @@ referral_id: docs_quickstart_free
 docs_area: get_started
 ---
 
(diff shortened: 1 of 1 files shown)


gemma-shay:
@lnhsingh I feel like this actually might better than ccloud long-term... I guess because the UI is more primary/robust than a command line tool by nature.

lnhsingh:
This LGTM. Agree- UI is probably best for now, until ccloud reaches parity (at least). Plus then the user can get acquainted with the UI, especially if this is the first time they're experiencing CRDB.

PR #16528, opened 2023-03-18
DOC-6759 added incomplete backups section

ashleykwtam:
Addresses DOC-6759:
- On Cloud Console, Dedicated users can now view complete backups separately from incomplete backups. 
- Incomplete Backups section has been added to the page along with an image

Preview:
- [Use Managed Service Backups](https://deploy-preview-16528--cockroachdb-docs.netlify.app/docs/cockroachcloud/use-managed-service-backups.html?filters=dedicated#incomplete-backups)

PR #16711, opened 2023-04-10
Fix serverless content on use managed backups page

ashleykwtam:
- Fixed Serverless section due to missing content

Preview:
[Use Managed Service Backups](https://deploy-preview-16711--cockroachdb-docs.netlify.app/docs/cockroachcloud/use-managed-service-backups.html?filters=dedicated)

### DIFF
--- a/cockroachcloud/use-managed-service-backups.md
+++ b/cockroachcloud/use-managed-service-backups.md
@@ -41,7 +41,6 @@ Your cluster's **Backups** page displays a list of your full and incremental clu
 
 For each backup, the following details display:
 
-<div class="filter-content" markdown="1" data-scope="dedicated">
 - **Data From**: The date and time the backup was taken. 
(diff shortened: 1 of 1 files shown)


ashleykwtam:
@kathancox when I worked on the epic [DOC-6759](https://github.com/cockroachdb/docs/pull/16528) it already addressed removing the 'Status' field. I noticed that the Serverless section was displaying incorrectly so this PR just updates that error.

kathancox:
This LGTM to @ashleykwtam, but can you update the PR description and title of the PR as it doesn't reflect the error you're correcting here. It might be better to mark the JIRA in a way that we know that the status field was removed with the prior PR and that it doesn't need work.

kathancox (approved):
LGTM, let's merge this ASAP since content is missing from the page.

ashleykwtam:
Done. Will update the JIRA accordingly, thanks Kathryn!

PR #16737, opened 2023-04-11
DOC-7423: Add overflow scrolling to all table elements

nickvigilante:
Fixes DOC-7423


JosephNickdow (approved):
I think is fine @nickvigilante

PR #16740, opened 2023-04-11
Homebrew instructions for cockroach-sql

ianjevans:
Fixes DOC-7443.


gemma-shay (commented):
LGTM but the OS tabs in `cockroach-sql-binary` are defaulting to linux instead of mac. this might have been pre-existing but can we switch the order to default to mac?

gemma-shay (approved):

PR #16753, opened 2023-04-12
Update docs on multiple active portal support

rmloveland:
Fixes DOC-4751, DOC-7418, DOC-7431


ZhouXing19 (approved):
LGTM! @rafiss could you give this a look as well?

rafiss (approved):
these docs lgtm!

one other thought i had, which i'm not sure we need to prioritize: i noticed that our docs don't talk about "prepared statements" at all. those are pretty fundamental for understanding what portals are. but since it's a PG wire protocol feature, maybe that means we don't document it on our site? i'm not sure what our approach is.

rmloveland:
> these docs lgtm!

awesome thanks!

> one other thought i had, which i'm not sure we need to prioritize: i noticed that our docs don't talk about "prepared statements" at all. those are pretty fundamental for understanding what portals are. but since it's a PG wire protocol feature, maybe that means we don't document it on our site? i'm not sure what our approach is.

ya we should have that.  we have a couple of open issues about it

DOC-2091
DOC-7025

7025 in particular I just found and it looks like there is important info we need to share with users in that one

I reassigned both of those issues to @taroface b/c they appear to be in the SQL Queries product area but he'll let me know if that's wrong, and if so I'll find out where they should go to get done

ianjevans (approved):
Not sure how the bad links got polluted into this branch, but you can rebase `master` to get a working build, looks like.

rmloveland:
> Not sure how the bad links got polluted into this branch, but you can rebase `master` to get a working build, looks like.

Thanks for the heads up!

PR #16769, opened 2023-04-14
Updates to transaction layer doc

kathancox:
Fixes DOC-6662

- Adjustments to language for clarification 
- Closed timestamp description added with reference to minimizing transaction retry error reference


v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,14 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
 
 Closed timestamps provide the guarantees that are used to provide support for low-latency historical (stale) reads, also known as [Follower Reads](../follower-reads.html). Follower reads can be particularly useful in [multi-region deployments](../multiregion-overview.html).
 
+When [contention](../performance-best-practices-overview.html#understanding-and-avoiding-transaction-contention) happens between transactions, the closed timestamp can provide some mitigation. Increasing the closed timestamp interval can reduce the likelihood of long-running transactions having their [timestamps pushed](#timestamp-cache), and therefore, causing the transaction to retry. However, note that increasing the closed timestamp interval can have an impact on the following:
```
kevin-v-ngo, line 120: > and therefore, causing the transaction to retry.

Not always so perhaps loosening the language?

and therefore, reduces the likelihood of the transaction retrying.

kevin-v-ngo (commented):
Looks good so far! Small nit

kevin-v-ngo (approved):
This PR looks good to me! Thanks.

kathancox:
Hey @AlexTalks! Just a friendly fresh ping here.

v23.1/architecture/transaction-layer.md
```diff
@@ -50,11 +50,11 @@ CockroachDB provides the following types of reads:
 
 ### Commits (phase 2)
 
-CockroachDB checks the running transaction's record to see if it's been `ABORTED`; if it has, it restarts the transaction.
+CockroachDB checks the running transaction's record to see if it's been `ABORTED`; if it has, it [throws an error](../transaction-retry-error-reference.html) to the client.
```
AlexTalks, line 53: nit: Do we want to specify that it is a retriable error here?

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,14 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
 
 Closed timestamps provide the guarantees that are used to provide support for low-latency historical (stale) reads, also known as [Follower Reads](../follower-reads.html). Follower reads can be particularly useful in [multi-region deployments](../multiregion-overview.html).
 
+When [contention](../performance-best-practices-overview.html#understanding-and-avoiding-transaction-contention) happens between transactions, the closed timestamp can provide some mitigation. Increasing the closed timestamp interval can reduce the likelihood of long-running transactions having their [timestamps pushed](#timestamp-cache), and therefore, reduces the likelihood of the transaction retrying. However, note that increasing the closed timestamp interval can have an impact on the following:
```
AlexTalks, line 120: This is correct, but a double-edged sword.  Considering a scenario in which a long-running transaction named txn1 holds write locks on some keys at time t=1, and txn2 is waiting to read those same keys at t=1:

1. If txn1 has it's timestamp pushed forward by the closed timestamp, moving its writes to t=2, txn2 may be able to proceed and read the keys at t=1.
2. If the closed timestamp interval is raised, that now means that long-running txn1 is causing txn2 to be long-waiting, as it can't read the keys at t=1 until txn1 is done or pushed into the future.

This is in addition to the other callouts listed below, and I think when talking about contention, it's worth mentioning that while increasing the closed timestamp interval may decrease retry errors, in may also increase lock wait latencies.

AlexTalks (changes requested):
These look like some good improvements!  I think some clarification on the impacts of the closed timestamp on contention are needed but beyond that this looks good to me.

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+
+- [Follower reads](../follower-reads.html): Reads will be forced to read from the leaseholder, which may affect latency/throughput.
+- [Change data capture](../change-data-capture-overview.html): There could be a lag in changefeed messages.
+- [Statistics collection](../cost-based-optimizer.html#table-statistics): Collection can place more load on the leaseholder.
+
+While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. For example, a long-running transaction (txn 1) holds [write locks](#writing) on keys at time t=1, and txn 2 is waiting to [read](#reading) those same keys at t=1. The following scenarios may occur depending on whether the closed timestamp has been increased or not:
+
+- If txn 1 has its timestamp pushed forward by the closed timestamp, moving its writes to time t=2, txn 2 may be able to proceed and read the keys at t=1. The chances of retryable errors have increased.
```
AlexTalks, line 128: Small nit: it may almost be preferable to write this as "When txn1 has its timestamp pushed", rather than if.

AlexTalks (approved):
Left a small comment.

More explanation:
For long-running transactions, and in general for almost any transaction, timestamp bumping can be almost inevitable.  The transaction may encounter a key written at a higher timestamp, may find some keys are in a range with a new lease (in which case the timestamp has to be moved to after the lease start), and so on.  The closed timestamp setting _may_ help in some cases, but it may not.

While the general rules stated here do apply, attempting to avoid transaction timestamps getting bumped may be a futile exercise, which is why these caveats are important.

In any case I don't want to block on this further because I think the updates here are beneficial, but if you see room for clarification please go ahead!

v23.1/architecture/transaction-layer.md
```diff
@@ -50,11 +50,11 @@ CockroachDB provides the following types of reads:
 
 ### Commits (phase 2)
 
-CockroachDB checks the running transaction's record to see if it's been `ABORTED`; if it has, it restarts the transaction.
+CockroachDB checks the running transaction's record to see if it's been `ABORTED`; if it has, it [throws a retryable error](../transaction-retry-error-reference.html) to the client.
 
 In the common case, it sets the transaction record's state to `STAGING`, and checks the transaction's pending write intents to see if they have succeeded (i.e., been replicated across the cluster).
```
mdlinville, line 55:
```suggestion
In the common case, it sets the transaction record's state to `STAGING` and checks the transaction's pending write intents to see if they have been successfully replicated across the cluster.
```
I'm confused about what "In the common case" refers to. I'm also confused about how line 53 relates to line 55 and 57. Who would have set the transaction record to ABORTED? If it's ABORTED, we throw the retryable error and move on. Otherwise, WE set the transaction record's state to STAGING (what else might it have been before?) and check for successful write attempts. When those succeed, then what is the status of the transaction record's state?

v23.1/architecture/transaction-layer.md
```diff
@@ -101,7 +101,7 @@ As part of providing serializability, whenever an operation reads a value, we st
 
 The timestamp cache is a data structure used to store information about the reads performed by [leaseholders](replication-layer.html#leases). This is used to ensure that once some transaction *t1* reads a row, another transaction *t2* that comes along and tries to write to that row will be ordered after *t1*, thus ensuring a serial order of transactions, aka serializability.
 
-Whenever a write occurs, its timestamp is checked against the timestamp cache. If the timestamp is less than the timestamp cache's latest value, we attempt to push the timestamp for its transaction forward to a later time. Pushing the timestamp might cause the transaction to restart in the second phase of the transaction (see [read refreshing](#read-refreshing)).
+Whenever a write occurs, its timestamp is checked against the timestamp cache. If the timestamp is less than the timestamp cache's latest value, we attempt to push the timestamp for its transaction forward to a later time. Pushing the timestamp might cause the transaction to restart [during the commit time](#commits-phase-2) of the transaction (see [read refreshing](#read-refreshing)).
```
mdlinville, line 104:
```suggestion
Whenever a write occurs, its timestamp is checked against the timestamp cache. If the timestamp is earlier than the timestamp cache's latest value, we attempt to push the timestamp for its transaction forward to a later time. Pushing the timestamp might cause the transaction to restart [during the commit time](#commits-phase-2) of the transaction (see [read refreshing](#read-refreshing)).
```
I don't love the use of "we" in this topic. Who is it? CockroachDB? The client? The end user? It might be better to avoid "we" to be more clear about what the actor is.

"during the commit time of the transaction" reads really strange to me. Do we mean until the transaction is committed?

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
 
 Closed timestamps provide the guarantees that are used to provide support for low-latency historical (stale) reads, also known as [Follower Reads](../follower-reads.html). Follower reads can be particularly useful in [multi-region deployments](../multiregion-overview.html).
 
+For long-running transactions, and generally any transaction, it can be inevitable that their timestamp is pushed, such as when they encounter a key written at a higher timestamp. When this kind of [contention](../performance-best-practices-overview.html#understanding-and-avoiding-transaction-contention) happens between transactions, the closed timestamp **may** provide some mitigation. Increasing the closed timestamp interval may reduce the likelihood of long-running transactions having their [timestamps pushed](#timestamp-cache), and therefore, reduces the likelihood of the transaction retrying. However, note that increasing the closed timestamp interval should be **carefully adjusted**, because it can have an impact on the following:
```
mdlinville, line 120:
```suggestion
The timestamp for any transaction, especially a long-running transaction, could be [pushed](#timestamp-cache). An example is when a transaction encounters a key written at a higher timestamp. When this kind of [contention](../performance-best-practices-overview.html#understanding-and-avoiding-transaction-contention) happens between transactions, the closed timestamp **may** provide some mitigation. Increasing the closed timestamp interval may reduce the likelihood that a long-running transaction's timestamp is pushed and must be retried. Thoroughly test any adjustment to the closed timestamp interval before deploying the change in production, because such an adjustment can have an impact on:
```
Forgive me if this reworking hurts more than it helps. It's weird to me to link "timestamps pushed" so long after you've introduced the concept. I've also removed some words to try to make the message a bit more clear and less wiggly. WDYT?

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+
+- [Follower reads](../follower-reads.html): Reads will be forced to read from the leaseholder, which may affect latency/throughput.
```
mdlinville, line 122:
```suggestion
- [Follower reads](../follower-reads.html): Latency or throughput may be increased, because reads are served only by the leaseholder.
```

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- [Change data capture](../change-data-capture-overview.html): There could be a lag in changefeed messages.
```
mdlinville, line 123:
```suggestion
- [Change data capture](../change-data-capture-overview.html): Latency of changefeed messages may be increased.
```
Latency of ... reading? Writing? Maybe you meant a different kind of "lag".

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- [Statistics collection](../cost-based-optimizer.html#table-statistics): Collection can place more load on the leaseholder.
```
mdlinville, line 124:
```suggestion
- [Statistics collection](../cost-based-optimizer.html#table-statistics): Collection can place more load on the leaseholder.
```
More load when compared with what? than on other nodes? Than before adjusting the setting?

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+
+While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. For example, a long-running transaction (txn 1) holds [write locks](#writing) on keys at time t=1, and txn 2 is waiting to [read](#reading) those same keys at t=1. The following scenarios may occur depending on whether the closed timestamp has been increased or not:
```
mdlinville, line 126:
```suggestion
While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. Consider an example where the long-running transaction `txn 1` holds [write locks](#writing) on keys at time `t=1`, and the long-running transaction `txn 2` is waiting to [read](#reading) those same keys at `t=1`. The following scenarios may occur, depending on whether or not the closed timestamp has been increased:
```

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+
+- When txn 1 has its timestamp pushed forward by the closed timestamp, moving its writes to time t=2, txn 2 may be able to proceed and read the keys at t=1. The chances of retryable errors have increased.
```
mdlinville, line 128:
```suggestion
- When the timestamp for `txn 1` is pushed forward by the closed timestamp, to time `t=2`, `txn 2` may then be able to read the keys at `t=1`. The likelihood of retryable errors has increased.
```
There is some subject-predicate disagreement and what seems to me like a dangling modifier. Let me know if I've captured your intention with the edit.

"The chances of...." is really informal, "the chance of" is slightly less informal, but "the likelihood of" is really what you intend here, I think.

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,19 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- If the closed timestamp interval is raised, long-running txn 1 is making txn 2 [wait](../performance-recipes.html#waiting-transaction), because txn 2 cannot read the keys at t=1 until txn 1 is done or pushed into the future. The chances of lock contention have increased.
```
mdlinville, line 129:
```suggestion
- If the closed timestamp interval is increased, `txn1` may cause `txn2`  to [wait](../performance-recipes.html#waiting-transaction), because `txn 2` cannot read the keys at `t=1` until `txn 1` is complete or is pushed into the future. The likelihood of lock contention has increased.
```

v23.1/architecture/transaction-layer.md
```diff
@@ -129,7 +142,7 @@ However, `client.Txn` is actually just a wrapper around `TxnCoordSender`, which
 
 - Dealing with transactions' state. After a transaction is started, `TxnCoordSender` starts asynchronously sending heartbeat messages to that transaction's transaction record, which signals that it should be kept alive. If the `TxnCoordSender`'s heartbeating stops, the transaction record is moved to the `ABORTED` status.
 - Tracking each written key or key range over the course of the transaction.
-- Clearing the accumulated write intent for the transaction when it's committed or aborted. All requests being performed as part of a transaction have to go through the same `TxnCoordSender` to account for all of its write intents, which optimizes the cleanup process.
+- Clearing the accumulated write intent for the transaction when it's committed or aborted. All operations being performed as part of a transaction have to go through the same `TxnCoordSender` to account for all of its write intents, which optimizes the cleanup process.
```
mdlinville, line 145:
```suggestion
- Clearing the accumulated write intent for the transaction when it's committed or aborted. All of a transaction's operations go through the same `TxnCoordSender` to account for all of its write intents and optimize the cleanup process.
```

v23.1/architecture/transaction-layer.md
kathancox: This part I did not work on specifically other than add a link, however, I do think you're right that the second paragraph is weirdly phrased. I am adjusting it to clear up some of those odd areas

v23.1/architecture/transaction-layer.md
kathancox: Done!

v23.1/architecture/transaction-layer.md
kathancox: Done!

kathancox: Yes, it is latency

v23.1/architecture/transaction-layer.md
kathancox: It will cause an increase in load during collection

v23.1/architecture/transaction-layer.md
kathancox: Changed!

v23.1/architecture/transaction-layer.md
kathancox: Changed, yep, agree!

v23.1/architecture/transaction-layer.md
kathancox: Done!

v23.1/architecture/transaction-layer.md
```diff
@@ -117,16 +117,16 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- [Follower reads](../follower-reads.html): Latency or throughput may be increased, because reads are served only by the leaseholder.
+- [Change data capture](../change-data-capture-overview.html): Latency of changefeed messages may be increased.
+- [Statistics collection](../cost-based-optimizer.html#table-statistics): Load placed on the leaseholder may increase during collection.
 
-While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. For example, a long-running transaction (txn 1) holds [write locks](#writing) on keys at time t=1, and txn 2 is waiting to [read](#reading) those same keys at t=1. The following scenarios may occur depending on whether the closed timestamp has been increased or not:
+While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. Consider an example where the long-running transaction `txn 1` holds [write locks](#writing) on keys at time `t=1`, and the long-running transaction `txn 2` is waiting to [read](#reading) those same keys at `t=1`. The following scenarios may occur depending on whether or not the closed timestamp has been increased:
```
mdlinville, line 126:
```suggestion
While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. Consider an example where:

1. The long-running transaction `txn 1` holds [write locks](#writing) on keys at time `t=1`.
1. The long-running transaction `txn 2` is waiting to [read](#reading) those same keys at `t=1`.

The following scenarios may occur depending on whether or not the closed timestamp has been increased:
```
This is complicated and it might reduce the cognitive load to break it up just a little. WDYT?

v23.1/architecture/transaction-layer.md
```diff
@@ -117,16 +117,16 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
 
-- When txn 1 has its timestamp pushed forward by the closed timestamp, moving its writes to time t=2, txn 2 may be able to proceed and read the keys at t=1. The chances of retryable errors have increased.
-- If the closed timestamp interval is raised, long-running txn 1 is making txn 2 [wait](../performance-recipes.html#waiting-transaction), because txn 2 cannot read the keys at t=1 until txn 1 is done or pushed into the future. The chances of lock contention have increased.
+- When the timestamp for `txn 1` is pushed forward by the closed timestamp, moving its writes to time `t=2`, `txn 2` may then be able to proceed and read the keys at `t=1`. The likelihood of retryable errors has increased.
```
mdlinville, line 128:
```suggestion
- If the timestamp for `txn 1` is pushed forward by the closed timestamp, its writes are moved to  time `t=2`, and it may then be able to read the keys at `t=1` before writing its changes at `t=2`. The likelihood of retryable errors has increased.
```
Is this correct? I changed it to If to match the next bullet.

v23.1/architecture/transaction-layer.md
```diff
@@ -117,16 +117,16 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- If the closed timestamp interval is increased, `txn1` may cause `txn2`  to [wait](../performance-recipes.html#waiting-transaction), because `txn 2` cannot read the keys at `t=1` until `txn 1` is complete or is pushed into the future. The likelihood of lock contention has increased.
```
mdlinville, line 129:
```suggestion
- If the closed timestamp interval is increased, `txn 2` may need to [wait](../performance-recipes.html#waiting-transaction) for `txn 1` to complete or to be pushed into the future before `txn 2` can proceed. The likelihood of lock contention has increased.
```

v23.1/architecture/transaction-layer.md
```diff
@@ -117,6 +117,24 @@ Note that closed timestamps are valid even if the leaseholder changes, since the
...
+- [Statistics collection](../cost-based-optimizer.html#table-statistics): Load placed on the leaseholder may increase during collection.
+
+While increasing the closed timestamp may decrease retryable errors, it may also increase lock latencies. Consider an example where:
+
+1. The long-running transaction `txn 1` holds [write locks](#writing) on keys at time `t=1`.
+1. The long-running transaction `txn 2` is waiting to [read](#reading) those same keys at `t=1`.
+
+The following scenarios may occur depending on whether or not the closed timestamp has been increased:
+
+- When the timestamp for `txn 1` is pushed forward by the closed timestamp, its writes are moved to  time `t=2`, and it can try again to read the keys at `t=1` before writing its changes at `t=2`. The likelihood of retryable errors has increased.
```
mdlinville, line 133: nit: double space in case you have a followup planned.

PR #16771, opened 2023-04-14
Recommend changefeeds for exporting data over `EXPORT`

kathancox:
Fixes DOC-7323

This PR adds the following to encourage users to create changefeeds for data export over `EXPORT`:

- Note to the top of the `EXPORT` page 
- Additional benefits of changefeeds to Export Data with Changefeeds page

### DIFF
--- a/v23.1/export-data-with-changefeeds.md
+++ b/v23.1/export-data-with-changefeeds.md
@@ -9,7 +9,7 @@ When you create an {{ site.data.products.enterprise }} changefeed, you can inclu
 
 {% include_cached new-in.html version="v23.1" %} You can also [schedule a changefeed](#create-a-scheduled-changefeed-to-export-filtered-data) to use a changefeed initial scan for exporting data on a regular cadence. 
 
-The benefits of using changefeeds for this use case compared to an [export](export.html), include:
+The benefits of using changefeeds for this use case instead of [export](export.html), include:
 
 - Changefeeds are jobs, which can be [paused](pause-job.html), [resumed](resume-job.html), [cancelled](cancel-job.html), [scheduled](create-schedule-for-changefeed.html), and [altered](alter-changefeed.html).
 - There is observability into a changefeed job using [`SHOW CHANGEFEED JOBS`](show-jobs.html#show-changefeed-jobs) and the [Changefeeds Dashboard](ui-cdc-dashboard.html) in the DB Console.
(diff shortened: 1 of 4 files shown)


amruss (approved):
LGTM

v23.1/export-data-with-changefeeds.md
```diff
@@ -9,11 +9,12 @@ When you create an {{ site.data.products.enterprise }} changefeed, you can inclu
 
 {% include_cached new-in.html version="v23.1" %} You can also [schedule a changefeed](#create-a-scheduled-changefeed-to-export-filtered-data) to use a changefeed initial scan for exporting data on a regular cadence. 
 
-The benefits of using changefeeds for this use case compared to an export, include:
+The benefits of using changefeeds for this use case compared to an [export](export.html), include:
```
rmloveland, line 12: I think this would be more emphatic advice with "for this ... instead of ..." (here and in other use)

v23.1/export.md
```diff
@@ -5,6 +5,10 @@ toc: true
 docs_area: reference.sql
 ---
 
+{{site.data.alerts.callout_info}}
```
rmloveland, line 8: I'd guess this is to head off questions but a part of me is like "why say this" if it's _not_ needed.  However I'm sure people will ask / are asking

v23.1/export.md
```diff
@@ -15,7 +19,7 @@ Using the [CockroachDB distributed execution engine](architecture/sql-layer.html
 If you do not need distributed exports, you can [export tabular data in CSV format](#non-distributed-export-using-the-sql-client).
 
 {{site.data.alerts.callout_info}}
-`EXPORT` no longer requires an {{ site.data.products.enterprise }} license.
```
rmloveland, line 18: I'd almost reverse the order of these callouts since if we recommend something else we could probably say that first thing

also I think this could be "green" (tip) or even "Red" if you wanna go hardcore ("DEFINITELY don't use this")

rmloveland (approved):
LGTM just had random thoughts

v23.1/export.md
kathancox: I agree, I think this maybe was added when it was no longer required and now remains.... I don't feel super confident about removing as part of this PR

v23.1/export-data-with-changefeeds.md
kathancox: I like that, changed!

v23.1/export.md
kathancox: Agree, switched these round. I have kept it as a note for now.... I felt very in between whether this should be a tip or warning, and so kept it boring with a note. I'll keep track of how the messaging around export vs changefeeds involves and maybe upgrade to a warning if it moves that way.

PR #16772, opened 2023-04-14
Landing page cards can be in any docs section

ianjevans:
This is a follow up PR to the new cards layout for landing pages. Instead of using the `docs_area` as the key for finding cards, use a new `cards_section` front matter variable.


nickvigilante (approved):
One nit, but it LGTM otherwise!

nickvigilante (approved):

PR #16776, opened 2023-04-17
DOC-7325: Rename variables to something more logical

nickvigilante:
Part of DOC-7325


lnhsingh (approved):

nickvigilante:
@lnhsingh I diffed the entire docs from before I made the variable name change to after I made the variable name change, and I was able to see the directories be identical, which means that the rename was complete and successful. I'm going to merge now and help Rich out with his PR, since his PR uses the previous variable names.

PR #16789, opened 2023-04-18
Update Raft election, lease intervals, etc.

rmloveland:
Fixes DOC-6375, DOC-7125

Summary of changes:

- Update 'Distribution Layer' page with updated network timeout and mention the `COCKROACH_NETWORK_TIMEOUT` env var that controls it

- Update 'Replication Layer' page with:

  - A table listing all of the relevant intervals that can be controlled with the `COCKROACH_RAFT_ELECTION_TIMEOUT_TICKS` variable

  - More links to the table wherever the values in that table are referred to

- Finally, all of the intervals are stored in variables so their values can be referenced and updated in one place going forward


rmloveland:
Hi @erikgrinaker these docs changes are based on your work in cockroachdb/cockroach#92542 and cockroachdb/cockroach#91947

Please let me know if I got some details wrong, I read through your comments in the code PRs but probably didn't understand everything properly

also open to any general feedback/comments on how we're presenting this info, or anything else that comes to mind

thanks!

v23.1/architecture/distribution-layer.md
```diff
@@ -87,7 +87,9 @@ In relationship to other layers in CockroachDB, the distribution layer:
 
 gRPC is the software nodes use to communicate with one another. Because the distribution layer is the first layer to communicate with other nodes, CockroachDB implements gRPC here.
 
-gRPC requires inputs and outputs to be formatted as protocol buffers (protobufs). To leverage gRPC, CockroachDB implements a protocol-buffer-based API defined in `api.proto`.
+gRPC requires inputs and outputs to be formatted as protocol buffers (protobufs). To leverage gRPC, CockroachDB implements a protocol-buffer-based API.
+
+By default, the network timeout for RPC connections between nodes is {{site.data.cockroach_network_timeout}}, with a connection timeout of {{site.data.cockroach_network_connection_timeout}}, in order to reduce unavailability and tail latencies during infrastructure outages. This can be changed via the environment variable `COCKROACH_NETWORK_TIMEOUT` which defaults to {{site.data.cockroach_network_timeout}}.
```
erikgrinaker, line 92: This is only kind of true. gRPC network timeouts only apply from the server's point of view, not the client's (in this case the client is another CRDB node), and the client is the one that usually matters. We additionally send an RPC ping every second to check that RPC connections are alive, but these have a fairly high timeout of 6 seconds (3x network timeout) because they're subject to contention with RPC traffic (so e.g. during a large import they can get delayed by several seconds).

This additional complexity is unfortunate, but currently necessary -- ideally we would only rely on the gRPC network timeout, but gRPC doesn't let us do that from the client side.

v23.1/architecture/replication-layer.md
```diff
@@ -207,6 +209,20 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+
+Constant | Default value | Notes
+---------|---------------|------
+[Raft](#raft) election timeout | {{site.data.constants.raft_election_timeout}} | The actual election timeout is multiplied by a random factor to avoid ties.
+[Raft](#raft) proposal timeout | {{site.data.constants.raft_proposal_timeout}} |
+[Node heartbeat interval](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.node_heartbeat_interval}} |
+[Lease interval](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.lease_interval}} |
+[Lease acquisition timeout](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.lease_acquisition_timeout}} |
+
+The values of the intervals above can be tuned using the environment variable `COCKROACH_RAFT_ELECTION_TIMEOUT_TICKS`, which specifies the minimum number of Raft ticks must pass before holding an election. It defaults to {{site.data.cockroach_raft_election_timeout_ticks}}. If changed, it will scale these bintervals proportionally.
```
erikgrinaker, line 224: This is no longer entirely true. In cockroachdb/cockroach#98616 we split out the lease interval to a separate env var `COCKROACH_RANGE_LEASE_DURATION` with a default of 6 seconds, which is independent of the Raft election timeout -- the node heartbeat interval is half of this, so 3 seconds. Similarly, cockroachdb/cockroach#98559 split out the proposal timeout to `COCKROACH_RAFT_REPROPOSAL_TIMEOUT_TICKS` defaulting to 6 ticks (3 seconds given the current 500 ms tick interval).

_data/constants.yml
```diff
@@ -0,0 +1,8 @@
+raft_election_timeout: 2s
+raft_proposal_timeout: 2-4s
+node_heartbeat_interval: 3s
+lease_interval: 6s
+lease_acquisition_timeout: 4s
+cockroach_network_timeout: 2s
+cockroach_network_connection_timeout: 4s
+cockroach_raft_election_timeout_ticks: 10
```
erikgrinaker, line 8: In cockroachdb/cockroach#98584, the tick interval `COCKROACH_RAFT_TICK_INTERVAL` was changed from 200 ms to 500 ms, so the Raft election timeout was reduced from 10 ticks to 4 ticks (both are equivalent to 2 seconds in wall time).

_data/constants.yml
```diff
@@ -0,0 +1,8 @@
+raft_election_timeout: 2s
+raft_proposal_timeout: 2-4s
```
erikgrinaker, line 2: The election timeout is 2-4 seconds, the proposal timeout is 3 seconds (in practice 3-6 seconds).

erikgrinaker (commented):
Thanks for updating this!

v23.1/architecture/distribution-layer.md
rmloveland: I added another paragraph with this info, slightly revised. PTAL since I may have misunderstood a bit

v23.1/architecture/replication-layer.md
rmloveland: I've revised this table to explain that these new environment variables can be used, as well as updated the table so that the various values are shown as what they are calculated from, e.g. "n_ticks * tick_value" (where appropriate)

also updated the table to use the updated tick info (and env var) from your other comment

_data/constants.yml
rmloveland: This is now updated so that the variables in the file represent the various foo_ticks variables you mentioned in other comments, and the table in replication-layer.md now shows a value something like "multiply the ticks times the interval" to describe the value

in practice it means that eg the table on `replication-layer.md` shows a value for the raft election timeout like 'N ticks * M ms' rather than a range of say "2-4s" as shown here, followed by something in the description like

> Controlled by `COCKROACH_RAFT_REPROPOSAL_TIMEOUT_TICKS`, which is then multiplied by the default tick interval to determine the value. This value is then multiplied by a random factor to avoid election ties.

rmloveland:
@erikgrinaker this is updated based on your previous comments and should be RFAL

erikgrinaker:
Sorry for the hold-up, I just got out of a non-stop L2 rotation and am still working my way through the backlog. Will get back to you shortly.

v23.1/architecture/distribution-layer.md
erikgrinaker: Yep, that's accurate, thanks!

v23.1/architecture/replication-layer.md
```diff
@@ -207,6 +209,23 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
 
 In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically based on the relative load across the nodes within a cluster. For more information, see the `kv.allocator.load_based_rebalancing` and `kv.allocator.qps_rebalance_threshold` [cluster settings](../cluster-settings.html). Note that depending on the needs of your deployment, you can exercise additional control over the location of leases and replicas by [configuring replication zones](../configure-replication-zones.html).
 
+### Important values and timeouts
+
+The following table lists some important values used by CockroachDB's replication layer:
+
+Constant | Default value | Notes
+---------|---------------|------
+[Raft](#raft) election timeout | {{site.data.constants.cockroach_raft_election_timeout_ticks}} * {{site.data.constants.cockroach_tick_interval}} | Controlled by `COCKROACH_RAFT_ELECTION_TIMEOUT_TICKS`, which is then multiplied by the default tick interval to determine the timeout value. This value is then multiplied by a random factor to avoid election ties.
```
erikgrinaker, line 218:
```suggestion
[Raft](#raft) election timeout | {{site.data.constants.cockroach_raft_election_timeout_ticks}} * {{site.data.constants.cockroach_tick_interval}} | Controlled by `COCKROACH_RAFT_ELECTION_TIMEOUT_TICKS`, which is then multiplied by the default tick interval to determine the timeout value. This value is then multiplied by a random factor of 1-2 to avoid election ties.
```

v23.1/architecture/replication-layer.md
```diff
@@ -207,6 +209,23 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+[Raft](#raft) proposal timeout | {{site.data.constants.cockroach_raft_reproposal_timeout_ticks}} * {{site.data.constants.cockroach_tick_interval}} | Controlled by `COCKROACH_RAFT_REPROPOSAL_TIMEOUT_TICKS`, which is then multiplied by the default tick interval to determine the value. This value is then multiplied by a random factor to avoid election ties.
```
erikgrinaker, line 219: I'd remove the last sentence. This doesn't really have anything to do with elections, it's how often we repropose/retry a Raft command if we haven't gotten a result yet. It can technically happen within a factor of 1-2 of this, because we only check it once per reproposal interval, but I don't think we need to get that technical.

v23.1/architecture/replication-layer.md
```diff
@@ -207,6 +209,23 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+[Lease interval](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.cockroach_range_lease_duration}} | Controlled by `COCKROACH_RANGE_LEASE_DURATION`.
+[Lease acquisition timeout](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.cockroach_range_lease_acquisition_timeout}} |
```
erikgrinaker, line 221: This is 2x the Raft election timeout (without randomization). Not sure if we need to specify that though, your call.

v23.1/architecture/replication-layer.md
```diff
@@ -207,6 +209,23 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+[Node heartbeat interval](#how-leases-are-transferred-from-a-dead-node) | {{site.data.constants.cockroach_range_lease_duration}} / 2 | Used to determine if you're having [node liveness issues](../cluster-setup-troubleshooting.html#node-liveness-issues).  This is calculated as one half of the lease interval.
+Raft tick interval | {{site.data.constants.cockroach_raft_tick_interval}} | Controlled by `COCKROACH_RAFT_TICK_INTERVAL`. Used to calculate various replication-related timeouts.
+
+The values of the intervals above can be tuned as follows:
+
+using the environment variable `COCKROACH_RAFT_ELECTION_TIMEOUT_TICKS`, which specifies the minimum number of Raft ticks must pass before holding an election. It defaults to {{site.data.cockroach_raft_election_timeout_ticks}}. If changed, it will scale these bintervals proportionally.
```
erikgrinaker, line 227: This isn't really true anymore. Only the lease acquisition timeout is now tied to the Raft election timeout, the other values are separate (arguably we should make that a separate value too, but it's too late now to do that for 23.1). I'd just nix this paragraph.

erikgrinaker (approved):
LGTM once comments are addressed. Thanks again for updating this!

I will say that tuning these values isn't trivial, since there are several interactions and aspects to consider. We can write something up around this if you'd like, or we can just consider this to be for advanced users who have some idea of the implications.

v23.1/architecture/replication-layer.md
rmloveland: Will leave it out for now in service of expediency and see if someone cares.  Can always add later

v23.1/architecture/replication-layer.md
rmloveland: Ah thanks, I think I meant to delete after latest update and missed it.  Removed entirely.

ianjevans:
> attractive nuisance

My absolute favorite legal term.

ianjevans (approved):

PR #16805, opened 2023-04-19
DOC-7485 self service cluster restores available for Dedicated

ashleykwtam:
Addresses DOC-7485:
- Customers for Dedicated can now restore clusters themselves in the CC console

Preview:
- [Use Managed Service Backups - Restore a cluster](https://deploy-preview-16805--cockroachdb-docs.netlify.app/docs/cockroachcloud/use-managed-service-backups.html?filters=dedicated#restore-a-cluster)
- [Cloud FAQs](https://deploy-preview-16805--cockroachdb-docs.netlify.app/docs/cockroachcloud/frequently-asked-questions.html#who-is-responsible-for-backup)


cockroachcloud/use-managed-service-backups.md
```diff
@@ -125,6 +124,20 @@ Additional ways to restore data:
 - [Back up a self-hosted CockroachDB cluster and restore into a {{ site.data.products.db }} cluster](#back-up-a-self-hosted-cockroachdb-cluster-and-restore-into-a-cockroachdb-cloud-cluster)
 - [Back up and restore data manually](take-and-restore-customer-owned-backups.html)
 
+### Restore a cluster
+
+To restore a cluster:
+
+1. Find the cluster backup on the **Backups** page.
```
benbardin, line 131: I'm not a markdown expert, would you confirm these items should all say 1? :)

cockroachcloud/use-managed-service-backups.md
```diff
@@ -125,6 +124,20 @@ Additional ways to restore data:
...
+1. Click **Restore** for the cluster you want to restore.
+
+    The **Restore cluster** module displays with backup details.
+
+1. Click **Continue**
+1. Once you have reviewed the restore details, click **Restore**.
+
+   The **Restore Jobs** tab will show you the status of your restore and update when the restore job has been created successfully.
```
benbardin, line 139: Add a warning? This will wipe all data currently in the database to replace it with the restore. This is obvious, and we say so in the GUI, but someone will still miss this and the more documentation when that happens the better. Thanks!

cockroachcloud/use-managed-service-backups.md
ashleykwtam: I'm not sure why they are all '1.' but that's how our markdown renders numbered lists, you can view in the Preview! :)

cockroachcloud/use-managed-service-backups.md
ashleykwtam: Good call! Better to be explicit about it!

benbardin (approved):

cockroachcloud/use-managed-service-backups.md
```diff
@@ -125,6 +124,23 @@ Additional ways to restore data:
...
+### Restore a cluster
+
+To restore a cluster:
+
+1. Find the cluster backup on the **Backups** page.
+1. Click **Restore** for the cluster you want to restore.
+
+    The **Restore cluster** module displays with backup details.
+
+1. Click **Continue**
```
kathancox, line 136: Nit: period at the end of **Continue**

cockroachcloud/use-managed-service-backups.md
```diff
@@ -125,6 +124,23 @@ Additional ways to restore data:
...
+    
+    The restore will completely erase all data in the cluster. All cluster data will be replaced with the data from the backup.
```
kathancox, line 138: Do we want to add this to a warning note?

Also, where this is placed, under 3, it seems like the data is erased at this point — is that true? 

I think adding this to a warning box will help with this second point ^

cockroachcloud/use-managed-service-backups.md
```diff
@@ -125,6 +124,23 @@ Additional ways to restore data:
 - [Back up a self-hosted CockroachDB cluster and restore into a {{ site.data.products.db }} cluster](#back-up-a-self-hosted-cockroachdb-cluster-and-restore-into-a-cockroachdb-cloud-cluster)
 - [Back up and restore data manually](take-and-restore-customer-owned-backups.html)
 
+### Restore a cluster
+
+To restore a cluster:
+
+1. Find the cluster backup on the **Backups** page.
```
kathancox, line 131: You'll know better than me, but the screenshot at the top of the Dedicated section... is that still correct? That is the Backups Page too. I know this is a different section, but I wonder if that needs updating? It might not, I have no idea. Is there a restore button that is actually on the side of that?

kathancox (commented):
Just a couple comments here. You may also want to add a short sentence on the CockroachDB Cloud FAQs within the backup question on the page that references this restore section for CockroachDB Ded managed-service backups.

Also, before merging can you double check that the cc release happened yesterday? I think it did. Perhaps check with Ben or Gemma will probably know!

cockroachcloud/use-managed-service-backups.md
ashleykwtam: Good catch. The Backups page screenshot is still correct but now has the 'Restore' button on the side. Updating the image to include this in the feature!

cockroachcloud/use-managed-service-backups.md
kathancox: Maybe where you have it makes sense then, if a reader is just about to, or has just clicked the continue step!

Thanks for the screen grab! Yeah, adding that into a warning callout in the docs will be nice symmetry here. (As Ben noted.)

kathancox (approved):
Before merging, can you check on whether this is live or not? (I think so) Worth double-checking! 

Cool, LGTM!

ashleykwtam:
Ben confirmed this is live! Good to merge for docs!

PR #16840, opened 2023-04-25
DOC-6980, DOC-7044 fix restore table by adding skip_missing_udfs option

ashleykwtam:
Addresses DOC-6980, DOC-7044:
- Previously UDFs could not be referenced from other objects, so backup and restore only needed to read and write UDF descriptors without worrying about missing UDFs when doing `RESTORE TABLE`
- Currently, there is a limitation that if you backup the whole database, but only restore an individual table which uses a UDF, the restore won’t find the UDF. Thus, the UDF is considered missing. 
- A new `RSTORE` command option `skip_missing_udfs` is added, so that when the option is given, checking for UDF dependencies will be skipped if UDF descriptors are missing (ie if UDFs are used in the table). 
- Any table that has a referenced UDF and is being restored will fail and thus **needs** to use `skip_missing_udfs`
- If UDF dependencies are missing and `skip_missing_udfs` flag is given, the DEFAULT expression would be dropped during the RESTORE

Preview
- [RESTORE](https://deploy-preview-16840--cockroachdb-docs.netlify.app/docs/dev/restore.html#options)

Needs clarification:
- I'm unsure where else I might need to add `skip_missing_udfs`, other than in the Options table. Would we want to include a paragraph/description about it in the Tables section of the page? I see a section discussing restoring user-defined types (e.g. ENUMs). Additionally, would we want to include an example of this?


kathancox:
Hey @ashleykwtam, I don't have much context on this without a closer review. 

You may want to discuss with an engineer the best description here. I think this could be described in a more user-facing way. "Use to skip checking UDF dependencies if UDF descriptors are missing." I would want to understand the effect this has on a restore job, because this isn't directly referenced. "if UDF descriptors are missing" trips me up here.

You may want to take a look at this section as well, once you've consulted with an engineer: https://deploy-preview-16840--cockroachdb-docs.netlify.app/docs/dev/restore.html#object-dependencies 

Interested in what you find out from engineering! Let me know if you have any other questions!

chengxiong-ruan (approved):
LGTM!

v23.1/restore.md
```diff
@@ -208,7 +209,7 @@ In general, two types are compatible if they are the same kind (e.g., an enum is
 
 ### Object dependencies
 
-Dependent objects must be restored at the same time as the objects they depend on.
+Dependent objects must be restored at the same time as the objects they depend on. Referenced UDFs are not restored and require the [`skip_missing_udfs`](#skip_missing_udfs) option.
```
kathancox, line 212: I suggest moving this new sentence below the table here. Since the existing sentence really helps to set context for the table following it, I think they should follow one another.

kathancox (approved):
One small suggestion, otherwise LGTM

PR #16863, opened 2023-04-28
Update SHOW DATABASES example output

ianjevans:
This PR fixes the example output of `SHOW DATABASES` to match the current output, and fixes the incorrect list of default database.

Fixes DOC-7511.
Fixes DOC-7374.


sjbarag (approved):
Thanks for the quick turnaround, @ianjevans!

PR #16876, opened 2023-05-02
Add docs on delegated Raft snapshots

rmloveland:
Fixes DOC-6815, DOC-6850

### DIFF
--- a/v23.1/architecture/replication-layer.md
+++ b/v23.1/architecture/replication-layer.md
@@ -108,6 +108,17 @@ Each replica can be "snapshotted", which copies all of its data as of a specific
 
 After loading the snapshot, the node gets up to date by replaying all actions from the Raft group's log that have occurred since the snapshot was taken.
 
+CockroachDB clusters running v23.1 and later can send _delegated snapshots_. Delegated snapshots can be sent by a Raft follower on behalf of the leader of a range. Which follower is chosen depends on the locality of the follower being nearest the replica that is the final recipient of the snapshot. If the follower is not able to send the snapshot quickly, the attempt is cancelled and the Raft leader sends the snapshot instead. If the follower is not able to send a snapshot that will be valid for the recipient, the request is rerouted to the leader.
+
+Sending data locally using delegated snapshots has the following benefits:
+
+- Snapshot transfers are faster
+- Snapshot transfers use less WAN bandwidth
+- Network costs are lower for operators of multi-region deployments
+- User traffic is less likely to be negatively impacted by snapshots
+
+Delegated snapshots are managed automatically by the cluster with no need for user involvement.
+
 ### Leases
 
 A single node in the Raft group acts as the leaseholder, which is the only node that can serve reads or propose writes to the Raft group leader (both actions are received as `BatchRequests` from [`DistSender`](distribution-layer.html#distsender)).


rmloveland:
@andrewbaptist it appears the cluster settings in the code in https://github.com/cockroachdb/cockroach/pull/83991 were different than the ones in the PR comment, so I went with the former

Open to not including the description of the settings so explicitly, since it's not clear we necessarily want folks changing them?  In any case they will appear in the autogenerated cluster settings docs for the very curious

This is mostly based on a mix of the PR commit message and the press release Gdoc.  Please feel free to suggest any corrections/improvements as you see fit of course!

v23.1/architecture/replication-layer.md
```diff
@@ -108,6 +108,20 @@ Each replica can be "snapshotted", which copies all of its data as of a specific
 
 After loading the snapshot, the node gets up to date by replaying all actions from the Raft group's log that have occurred since the snapshot was taken.
 
+CockroachDB clusters running v23.1 and later can send _delegated snapshots_.  Delegated snapshots can be sent by a Raft follower instead of the leader of a range.  Which follower is chosen depends on the locality of the follower being nearest the replica that is the final recipient of the snapshot. If the follower is not able to send the snapshotf quickly, the attempt is cancelled and the Raft leader sends the snapshot instead. If the follower is not able to send a snapshot that will be valid for the recipient, the request is rerouted to the leader.
```
andrewbaptist, line 111: Change "Delegated snapshots can be sent by a Raft follower instead of the leader of a range." to
"Delegated snapshots can be sent by a Raft follower on behalf of the leader of a range."

Typo snapshotf -> snapshot

v23.1/architecture/replication-layer.md
```diff
@@ -108,6 +108,20 @@ Each replica can be "snapshotted", which copies all of its data as of a specific
...
+Sending data locally using delegated snapshots has the following benefits:
+
+- Snapshot transfers are faster
+- Snapshot transfers use less WAN bandwidth
+- Network costs are lower for operators of multi-region deployments
+- User traffic is less likely to be negatively impacted by snapshots
+
+Delegated snapshots are enabled by default, and managed automatically by the cluster with no need for user involvement.  If you want to control how delegated snapshots work, the following cluster settings are available:
+
+- [`kv.snapshot_delegation.num_follower`](../cluster-settings.html#setting-kv-snapshot-delegation-num-follower): The number of delegates to try when sending snapshots, before falling back to sending from the leaseholder.  If set to `0`, snapshot delegation is disabled.
```
andrewbaptist, line 122: I'm not sure if you want to mention this only applies if the entire cluster is on v23.1

The settings were renamed I checked the code and they are currently. (Sorry for the confusion)

kv.snapshot_delegation.max_delegation_attempts (replaces num_followers)
kv.snapshot_delegations.queue_limit (replaces num_requests)

See https://github.com/cockroachdb/cockroach/blob/master/pkg/kv/kvserver/replica_command.go#L2978 for details (including the help text).

andrewbaptist (changes requested):
Thanks for doing this, a few small changes.

v23.1/architecture/replication-layer.md
rmloveland: Updated that line, and fixed the typo!  (note to self: spellcheckkkk)

v23.1/architecture/replication-layer.md
rmloveland: Thanks for the updated info.  I took another look and it appears these settings are not marked as public in the code so they are not appearing in the [v23.1 cluster settings docs](https://www.cockroachlabs.com/docs/v23.1/cluster-settings.html) (which are generated automatically from the cockroach repo, but only for "public" settings).

Was that intentional?  If so I would recommend we remove the cluster setting info from this PR for consistency with how we document other cluster settings.

If this was not intentional and these settings should be "public", I can add some info here from your comment.  For now I have removed mention of the cluster settings

Advanced users can always see all the non-public cluster settings and their help text using `SHOW ALL CLUSTER SETTINGS` if they want to. 

PTAL and let me know what you think

andrewbaptist (approved):
:lgtm:

ianjevans (approved):
Looks good, but see my comment about possibly linking to the setting that enables/disables delegated snapshots.

v23.1/architecture/replication-layer.md
```diff
@@ -108,6 +108,17 @@ Each replica can be "snapshotted", which copies all of its data as of a specific
...
+CockroachDB clusters running v23.1 and later can send _delegated snapshots_. Delegated snapshots can be sent by a Raft follower on behalf of the leader of a range. Which follower is chosen depends on the locality of the follower being nearest the replica that is the final recipient of the snapshot. If the follower is not able to send the snapshot quickly, the attempt is cancelled and the Raft leader sends the snapshot instead. If the follower is not able to send a snapshot that will be valid for the recipient, the request is rerouted to the leader.
+
+Sending data locally using delegated snapshots has the following benefits:
+
+- Snapshot transfers are faster
+- Snapshot transfers use less WAN bandwidth
+- Network costs are lower for operators of multi-region deployments
+- User traffic is less likely to be negatively impacted by snapshots
+
+Delegated snapshots are enabled by default, and managed automatically by the cluster with no need for user involvement.
```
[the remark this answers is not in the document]
rmloveland: Oh - that is a remnant from when we were gonna add the cluster settings, but they're private.  Amended to 

> Delegated snapshots are managed automatically by the cluster with no need for user involvement.

PR #16879, opened 2023-05-02
Add new-in tags to SQL shell topic as footnotes

ianjevans:


lnhsingh (approved):
Approving so your not blocked, but see my last comment

PR #16885, opened 2023-05-03
Correct changefeed from system priv v22.2

kathancox:
Fixes DOC-7574

Removed `CHANGEFEED` priv from system-level privilege table in v22.2 and added to supported privilege table.

### DIFF
--- a/_includes/v22.2/sql/privileges.md
+++ b/_includes/v22.2/sql/privileges.md
@@ -1,26 +1,27 @@
 Privilege | Levels
 ----------|------------
 `ALL` | Database, Schema, Table, Sequence, Type
+`BACKUP` | System, Database, Table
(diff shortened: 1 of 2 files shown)


rmloveland:
> Hey @rmloveland, I updated the v22.2 version of the privileges page we were discussing the other day — for a changefeed fix. I haven't made the same wider change as per the 23.1 PR you reviewed. That is, removing that 2nd table. I erred here on the side of a small change, and just letting that system-level / role option comparison table stand in v22.2. What do you think? I could also remove — I'm v split here.

I think doing the small change is fine in v22.2 as the notes about "x added in v22.2" are still relevant in these docs

ianjevans (approved):

kathancox:
Thank you @ianjevans !

PR #16889, opened 2023-05-04
Add Confluent Schema Registry to client cert & key parameters

kathancox:
Fixes DOC-6848

Added Confluent Schema Registry to the `client_cert` and `client_key` query params.


kathancox:
@HonoreDB Besides listing that these two parameters can be used with Confluent Schema Registries, is there anything else that should be noted that differs from what is already present?

HonoreDB (approved):
LGTM, these should all be intuitive as they behave identically to the params of the same name in Kafka and webhook URIs.

ashleykwtam (approved):
LGTM, all the links check out :)

PR #16894, opened 2023-05-05
Remove the preview label from changefeed webhook sink

kathancox:
Fixes DOC-7149

Removed the preview label from webhook sink across the changefeed docs.
Updated the "beta" label on Pub/Sub to "Preview" that was missed on one of the pages.


amruss (approved):
LGTM

PR #16895, opened 2023-05-05
Fix the v20.1 Change Data Capture page version dropdown

kathancox:
Fixes DOC-7590

Right now, users who search for "cockroachdb cdc" on Google will run into the Change Data Capture Overview page (good), but the second result will be the very old Change Data Capture page (bad). And it's particularly bad, since over the last few years the `key` element hasn't been added to frontmatters to control where v20.1 visitors on that page will end up. They cannot select the newer versions, and so end up stuck.


kathancox:
Closing this PR — need to rework this and will approach in a different way.

PR #16896, opened 2023-05-05
Deprecate controlchangefeed role option note to table

kathancox:
Fixes DOC-6512

The `CONTROLCHANGEFEED` role option is deprecated as of v23.1. This is already noted on the `CREATE CHANGEFEED` SQL page. This PR adds this to the role option table that we pull into a few different pages (e.g., `ALTER USER`, `CREATE ROLE`...)


taroface (approved):
LGTM

PR #16897, opened 2023-05-05
DOC-5611: Mark all Cockroach Labs-owned domains as internal

nickvigilante:
Fixes DOC-5611

All URLs owned directly by Cockroach Labs are now no longer listed as external links


mdlinville (approved):

js/customscripts.js
```diff
@@ -30,6 +30,18 @@ function renderTOC() {
 
 var $versionSwitcher, versionSwitcherBottom = Infinity;
 
+const cockroachDomains = [
```
mikeCRL, line 33: @nickvigilante LGTM. You could also add `cockroachlabs.atlassian.net` in case we ever link to the public wiki https://cockroachlabs.atlassian.net/wiki/spaces/CRDB/overview

mikeCRL (approved):

js/customscripts.js
nickvigilante: I'll send a follow-up PR.

PR #16924, opened 2023-05-10
updates to UDF functionality

taroface:
DOC-6753

- Add example for `SETOF` type
- Add example for `RETURNS RECORD`
- Document UDF inlining
- Remove some features from unsupported UDF syntax list (i.e., they are supported now)

### DIFF
--- a/v23.1/create-function.md
+++ b/v23.1/create-function.md
@@ -31,8 +31,8 @@ If you grant `EXECUTE` privilege as a default privilege at the database level, n
 Parameter | Description
 ----------|------------
 `func_create_name` | The name of the function.
-`func_arg` | A function argument.
-`func_arg_type` | The type returned by the function.
+`func_param` | A function argument.
+`func_param_type` | The type returned by the function. 
 `opt_routine_body` | The body of the function. For allowed contents, see [User-Defined Functions: Overview](user-defined-functions.html#overview).
 
 ## Example of a simple function
@@ -41,14 +41,14 @@ The following statement creates a function to compute the square of integers:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> CREATE OR REPLACE FUNCTION sq(a INT) RETURNS INT AS 'SELECT a*a' LANGUAGE SQL;
+CREATE OR REPLACE FUNCTION sq(a INT) RETURNS INT AS 'SELECT a*a' LANGUAGE SQL;
 ~~~
 
 The following statement invokes the `sq` function:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> SELECT sq(2);
+SELECT sq(2);
 ~~~
 
 ~~~
@@ -68,12 +68,12 @@ The following statement defines a function that returns the total number of MovR
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> CREATE OR REPLACE FUNCTION num_users() RETURNS INT AS 'SELECT count(*) from users' LANGUAGE SQL;
+CREATE OR REPLACE FUNCTION num_users() RETURNS INT AS 'SELECT COUNT(*) FROM users' LANGUAGE SQL;
 ~~~
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> SELECT num_users();
+SELECT num_users();
 ~~~
 
 ~~~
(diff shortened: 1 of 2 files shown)


v23.1/create-function.md
```diff
@@ -89,21 +89,75 @@ The following statement defines a function that returns the total revenue for ri
...
+In the function subquery, the latest `end_time` timestamp is used to determine the most recently completed ride.
+
+{% include_cached copy-clipboard.html %}
+~~~ sql
+CREATE OR REPLACE FUNCTION last_rider() RETURNS RECORD LANGUAGE SQL AS $$
+  SELECT * FROM users WHERE id = (
+    SELECT rider_id FROM rides WHERE end_time = (SELECT max(end_time) FROM rides)
+  )
+$$;
+~~~
```
taroface, line 147: @mgartner Is there a better way to define this example function? Also, does this indenting look ok?

v23.1/user-defined-functions.md
```diff
@@ -12,12 +12,15 @@ A user-defined function (UDF) is a named function defined at the database level
...
-- The return type can be a built-in type, user-defined enum, implicit record type, or `VOID`. `VOID` indicates that there is no return type and `NULL` will always be returned. If the return type of the function is not `VOID`, the last statement of a UDF must be a `SELECT`.
+- An argument has a _mode_ and a _type_. CockroachDB supports the `IN` argument mode. The type can be a built-in type, [user-defined `ENUM`](enum.html), or implicit record type. CockroachDB **does not** support default values for arguments.
+- The return type can be a built-in [type](data-types.html), user-defined [`ENUM`](enum.html), [`RECORD`](create-function.html#create-a-function-that-returns-a-record-type), implicit record type, or `VOID`.
+    - Preceding a type with `SETOF` indicates that a set, or multiple rows, should be returned. For an example, see [Create a function that returns a set of results](create-function.html#create-a-function-that-returns-a-set-of-results).
```
taroface, line 17: @mgartner Can `SETOF` precede any type?

mgartner: Yes.

mgartner: "should be returned" is potentially misleading - the in a SETOF UDF the last statement in the function body can return any number of rows: 0, 1, or more. It might be more correct to say "may be returned".

v23.1/user-defined-functions.md
```diff
@@ -105,6 +92,192 @@ If you do not specify a schema for the function `add` when you create it, the de
...
+                KV gRPC calls: 1
+                estimated max memory allocated: 20 KiB
+                sql cpu time: 11µs
+                estimated row count: 1 (100% of the table; stats collected 42 seconds ago)
+                table: b@b_pkey
+                spans: FULL SCAN
+    (57 rows)
+    ~~~
+
+    The query takes only `4ms` to execute because the function is inlined as an equality comparison `(a) = (b)` with a much lower overhead.
```
mgartner, line 279: nit: maybe improve this with something like: "because the function is inlined **and transformed to a join with an equality comparison, which has much less overhead than invoking a function for each row scanned in table `a`.**

mgartner (approved):
Looks great!

v23.1/create-function.md
```diff
@@ -31,8 +31,8 @@ If you grant `EXECUTE` privilege as a default privilege at the database level, n
 Parameter | Description
 ----------|------------
 `func_create_name` | The name of the function.
-`func_arg` | A function argument.
-`func_arg_type` | The type returned by the function.
+`func_param` | A function argument.
```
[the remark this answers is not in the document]
DrewKimball: Alternatively, maybe we leave it as `arg` for the docs, and let the more precise wording be internal only. Maybe @mgartner has further thoughts on that.

v23.1/user-defined-functions.md
```diff
@@ -12,12 +12,15 @@ A user-defined function (UDF) is a named function defined at the database level
...
-- The return type can be a built-in type, user-defined enum, implicit record type, or `VOID`. `VOID` indicates that there is no return type and `NULL` will always be returned. If the return type of the function is not `VOID`, the last statement of a UDF must be a `SELECT`.
+- An argument has a _mode_ and a _type_. CockroachDB supports the `IN` argument mode. The type can be a built-in type, [user-defined `ENUM`](enum.html), or implicit record type. CockroachDB **does not** support default values for arguments.
+- The return type can be a built-in [type](data-types.html), user-defined [`ENUM`](enum.html), [`RECORD`](create-function.html#create-a-function-that-returns-a-record-type), implicit record type, or `VOID`.
+    - Preceding a type with `SETOF` indicates that a set, or multiple rows, should be returned. For an example, see [Create a function that returns a set of results](create-function.html#create-a-function-that-returns-a-set-of-results).
+    - `VOID` indicates that there is no return type and `NULL` will always be returned. If the return type of the function is not `VOID`, the last statement of a UDF must be a `SELECT`.
```
DrewKimball, line 18: Currently only `Select` is supported in UDFs, so each statement in a UDF must be a `SELECT` in any event.

v23.1/create-function.md
```diff
@@ -68,12 +68,12 @@ The following statement defines a function that returns the total number of MovR
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-> CREATE OR REPLACE FUNCTION num_users() RETURNS INT AS 'SELECT count(*) from users' LANGUAGE SQL;
+CREATE OR REPLACE FUNCTION num_users() RETURNS INT AS 'SELECT COUNT(*) FROM users' LANGUAGE SQL;
```
DrewKimball, line 71: [nit] function names should be left lower case (count).

DrewKimball (commented):
Looks pretty good, just a few nits mostly.

v23.1/create-function.md
taroface: OK, thank you! I'll leave as-is, although that example is insightful to me.

v23.1/user-defined-functions.md
taroface: Noted - commenting the last line out until relevant.

v23.1/create-function.md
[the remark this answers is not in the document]
taroface: Yeah... "argument" seems to be all over the UDF feature doc. Do you mean that everything called a UDF "argument" is actually a "parameter"? If so, I can file a follow-up issue for this.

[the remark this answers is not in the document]
DrewKimball: `argument` is the value that is supplied to the function when it is called; `parameter` is the name+type in the function definition. I'm unsure if it really matters for our public docs, but an issue wouldn't hurt.

kathancox (approved):
LGTM! looks great!

PR #16940, opened 2023-05-11
enforce_home_region_follower_reads_enabled session var

taroface:
DOC-6784

Document the `enforce_home_region_follower_reads_enabled` session setting.


_includes/v23.1/misc/session-vars.md
```diff
@@ -22,6 +22,7 @@
...
+| <a name="enforce-home-region-follower-reads-enabled"></a> `enforce_home_region_follower_reads_enabled` | If `on`, allows the [`enforce_home_region`](cost-based-optimizer.html#control-whether-queries-are-limited-to-a-single-region) setting to perform `AS OF SYSTEM TIME` [follower reads](follower-reads.html) to determine a query's [home region](multiregion-overview.html#table-localities), if any. | `off` | Yes | Yes |
```
msirek, line 25: Maybe change this to be more similar to https://github.com/cockroachdb/cockroach/blob/3042c4fee9457b4f05318fcfa9836433203430cb/pkg/sql/sessiondatapb/local_only_session_data.proto#L346-L348

Either use "determine and report" or "detect and report", or something similar in place of "determine", and also mention that `enforce_home_region` must also be set to `on` for this setting to have any effect.

msirek (changes requested):

_includes/v23.1/misc/session-vars.md
msirek: Also, would this be the place to mention that `enforce_home_region_follower_reads_enabled` is a preview feature?

taroface: Done and done - thanks for pointing out that this is in preview! PTAL.

msirek (approved):

kathancox (approved):
LGTM!

PR #16823, opened 2023-04-21
update contention performance recipes & terminology

taroface:
DOC-6664

- Updated the Performance Tuning page with prescriptive guidance for waiting transactions, transaction retry errors, and overall cluster contention (recorded contention events).
- Streamlined the SQL Performance Best Practices section to include only preventative/proactive guidance. Previously this page also included troubleshooting/diagnostics steps, which was redundant with the Performance Tuning page.
- Parsed the existing "reduce transaction contention" guidance into 1) client-side and 2) client- & server-side recommendations.
- Unified our "transaction contention" semantics, where I encountered it, to the best of my current understanding.
- Added Hot Spots guidance to the Performance Tuning page. **Note:** Overall, our existing docs and the requested changes emphasize _transaction contention_ rather than the umbrella topic of "contention". From what I understand, hot spots indicate _resource contention_ and may increase the occurrence of _transaction contention_, but are not intrinsically an outcome of transaction contention. For this reason, I haven't grouped hot spots underneath the transaction contention guidance. Hot spots were & are still a separate heading. However, I did note that hot spots may also be observed when transaction contention is observed. Let me know if I'm mistaken anywhere or if a different approach is preferred.
- Removed the prescriptive steps under "Hanging and stuck queries" on the SQL Troubleshooting page. This now links back to the Performance Tuning recipe. Also, the old steps used complex SQL statements likely because they were written before we had the ability to diagnose waiting transactions in the UI. I don't think we lose much by replacing them with the UI steps, unless there is a strong use case for querying internal tables.
- Added & removed includes where I found that necessary.

We have some remaining gaps:

- Throughout the documentation (e.g. `performance-recipes.html#transaction-contention`, `transaction-retry-error-reference.html`, possibly others), we don't sufficiently differentiate serializable and lock contention. In the former link, only lock contention is specifically defined. I believe @rmloveland will address this in https://cockroachlabs.atlassian.net/browse/DOC-7148.
- (Less urgently) The docs for the Transactions page on Cloud (`cockroachcloud/transactions-page.html`) may need to be updated with the Active Executions table, which is documented on DB Console (`ui-transactions-page.html#active-executions-table`). Unless this doesn't exist on Cloud? cc @ianjevans @gemma-shay 

@kevin-v-ngo Is there an eng reviewer you can suggest for this PR?

PR #16945, opened 2023-05-11
follow up contention update with AlexTalks feedback

taroface:
This is a fast-follow PR that incorporates feedback from @AlexTalks on #16823, which was already merged.


_includes/v23.1/performance/reduce-contention.md
```diff
@@ -4,7 +4,7 @@
...
+- Use historical reads ([`SELECT ... AS OF SYSTEM TIME`](as-of-system-time.html)), preferably [bounded staleness reads](follower-reads.html#when-to-use-bounded-staleness-reads) or [exact staleness with follower reads](follower-reads.html#run-queries-that-use-exact-staleness-follower-reads) when possible to reduce conflicts with other writes. This reduces the likelihood of [`RETRY_SERIALIZABLE`](transaction-retry-error-reference.html#retry_serializable) errors as fewer writes will happen at the historical timestamp. More specifically, writes' timestamps are less likely to be pushed by historical reads as they would [when the read has a higher priority level](architecture/transaction-layer.html#transaction-conflicts). Note that if the `AS OF SYSTEM TIME` value is below the closed timestamp, the read canot be invalidated.
```
kathancox, line 7: Ooops, little typo here
```suggestion
- Use historical reads ([`SELECT ... AS OF SYSTEM TIME`](as-of-system-time.html)), preferably [bounded staleness reads](follower-reads.html#when-to-use-bounded-staleness-reads) or [exact staleness with follower reads](follower-reads.html#run-queries-that-use-exact-staleness-follower-reads) when possible to reduce conflicts with other writes. This reduces the likelihood of [`RETRY_SERIALIZABLE`](transaction-retry-error-reference.html#retry_serializable) errors as fewer writes will happen at the historical timestamp. More specifically, writes' timestamps are less likely to be pushed by historical reads as they would [when the read has a higher priority level](architecture/transaction-layer.html#transaction-conflicts). Note that if the `AS OF SYSTEM TIME` value is below the closed timestamp, the read cannot be invalidated.
```

kathancox (approved):

taroface:
TFTR @kathancox!! This is why we do docs review

PR #16971, opened 2023-05-15
Update recommendations re: GC TTL

rmloveland:
Fixes DOC-7622


bdarnell (commented):
> Smaller values can save disk space if values are frequently overwritten

Smaller values can also improve performance (especially for queue-like workloads); I think performance is a more common reason for decreasing `gc.ttlseconds` than disk space.

> Do not set this below `600` (10 minutes); doing so will cause problems for long-running queries

"If I never run queries that are longer than one minute, can I set this lower?" That is, is 10 minutes tied to the presumed lifetime of your longest-running query (an assumption we don't seem to make on the linked page), or is it still the minimum regardless? I'd borrow language from the next bullet: "the smallest value that is regularly tested is 600 (10 minutes)".

> If you are going to set it to a value anywhere between `90000` (25 hours) and `604800` (7 days), make sure to do extensive testing with your workload beforehand.

This feels needlessly detailed. I would just pick one recommended upper bound. (and I don't think it's terribly important which one we pick - apps with low update rates will be fine either way, and apps with very high update rates will be in trouble either way, and I suspect (with no evidence whatsoever) that it's not very common for the TTL value that gives your app trouble to fall in the 25h-7d range. 

> if set too high, all the changes to a row in that time period could add up to more than [the default range size](#range-max-bytes)

This is less important than it used to be since the default range size has increased. (But also, what matters is the `range_max_bytes` for *this* range, whether it is the default or not, so `s/default/maximum/`)

> Such oversized ranges could contribute to the server [running out of memory](cluster-setup-troubleshooting.html#out-of-memory-oom-crash)

Now that we have streaming snapshots, are large ranges still an OOM risk?

> We recommend using [scheduled backups](create-schedule-for-backup.html) instead, which automatically [use protected timestamps](create-schedule-for-backup.html#protected-timestamps-and-scheduled-backups) to ensure they succeed.

AIUI, because of the way scheduled incremental backups prevent GC, they more or less prevent you from benefiting from further reducing the GC TTL. That is, with manual backups, if you set GC TTL to less than the incremental backup interval, the GC wins and breaks backups. With scheduled backups, the backup takes priority and prevents GC. Either way, it is never a good idea to set a GC TTL to less than your incremental backup interval. Right?


There's a lot of text here and different formatting (opening paragraph, bullets, callout_info, final distant statement of the default). Do we still need the scheduled backup callout in a box or should it be a bullet along with the other "things that can go wrong if you set this poorly"? I'd like to get the default closer to the recommended min/max just to make the numbers more easily comparable.

Here's an alternative organization that I think flows a little better:

The number of seconds overwritten [MVCC values]({{link_prefix}}architecture/storage-layer.html#mvcc) will be retained before [garbage collection]({{link_prefix}}architecture/storage-layer.html#garbage-collection).

**Default:** `14400` (4 hours)

Smaller values can save disk space and improve performance if values are frequently overwritten or for queue-like workloads. The smallest value we regularly test is 600 (10 minutes); smaller values are unlikely to be beneficial because of the frequency with which GC runs. If you use incremental backups, the GC TTL must be greater than the interval between incremental backups.

Larger values increase the interval allowed for AOST queries and allow for less frequent incremental backups. The largest value we regularly test is 90000 (25 hours). Increasing the GC TTL is not meant to be a solution for long-term retention of history; for that you should handle versioning in the schema design at the application layer. Setting the TTL too high could cause problems if the retained versions of a single row approach the maximum range size.

rmloveland:
Ben Darnell ***@***.***> writes:

>>     Smaller values can save disk space if values are frequently
>>     overwritten

> Smaller values can also improve performance (especially for queue-like
> workloads); I think performance is a more common reason for decreasing
> gc.ttlseconds than disk space.

Good to know

>>     Do not set this below 600 (10 minutes); doing so will cause
>>     problems for long-running queries

> "If I never run queries that are longer than one minute, can I set
> this lower?" That is, is 10 minutes tied to the presumed lifetime of
> your longest-running query (an assumption we don't seem to make on the
> linked page), or is it still the minimum regardless? I'd borrow
> language from the next bullet: "the smallest value that is regularly
> tested is 600 (10 minutes)".

>>     If you are going to set it to a value anywhere between 90000 (25
>>     hours) and 604800 (7 days), make sure to do extensive testing
>>     with your workload beforehand.

> This feels needlessly detailed. I would just pick one recommended
> upper bound. (and I don't think it's terribly important which one we
> pick - apps with low update rates will be fine either way, and apps
> with very high update rates will be in trouble either way, and I
> suspect (with no evidence whatsoever) that it's not very common for
> the TTL value that gives your app trouble to fall in the 25h-7d range.

Makes sense to me, a simple min/max is easier to understand

>>     if set too high, all the changes to a row in that time period
>>     could add up to more than the default range size

> This is less important than it used to be since the default range size
> has increased. (But also, what matters is the range_max_bytes for this
> range, whether it is the default or not, so s/default/maximum/)

This should be fixed in the latest push - I used your suggested language
below, which says "maximum range size"

>>     Such oversized ranges could contribute to the server running out
>>     of memory

> Now that we have streaming snapshots, are large ranges still an OOM
> risk?

I don't know, perhaps @irfansharif does.  Latest update is based on your
suggestions below and removes mention of this OOM stuff FWIW.

>>     We recommend using scheduled backups instead, which automatically
>>     use protected timestamps to ensure they succeed.

> AIUI, because of the way scheduled incremental backups prevent GC,
> they more or less prevent you from benefiting from further reducing
> the GC TTL. That is, with manual backups, if you set GC TTL to less
> than the incremental backup interval, the GC wins and breaks backups.
> With scheduled backups, the backup takes priority and prevents GC.
> Either way, it is never a good idea to set a GC TTL to less than your
> incremental backup interval. Right?

Yes, that's my understanding as well from the existing note text - we
updated it recently due to the protected timestamps work (described in
https://www.cockroachlabs.com/blog/protected-timestamps-for-less-garbage/)
so the error message `protected ts verification error` would be easy to
find.

To cover this, I've pulled in the content of the note where we recommend
using a backup schedule instead so users don't have to worry as much
about GC TTL vs backups.

> There's a lot of text here and different formatting (opening
> paragraph, bullets, callout_info, final distant statement of the
> default). Do we still need the scheduled backup callout in a box or
> should it be a bullet along with the other "things that can go wrong
> if you set this poorly"? I'd like to get the default closer to the
> recommended min/max just to make the numbers more easily comparable.

+1, min/max sounds great to me

> Here's an alternative organization that I think flows a little better:

> The number of seconds overwritten MVCC values will be retained before
> garbage collection.

> Default: 14400 (4 hours)

> Smaller values can save disk space and improve performance if values
> are frequently overwritten or for queue-like workloads. The smallest
> value we regularly test is 600 (10 minutes); smaller values are
> unlikely to be beneficial because of the frequency with which GC runs.
> If you use incremental backups, the GC TTL must be greater than the
> interval between incremental backups.

> Larger values increase the interval allowed for AOST queries and allow
> for less frequent incremental backups. The largest value we regularly
> test is 90000 (25 hours). Increasing the GC TTL is not meant to be a
> solution for long-term retention of history; for that you should
> handle versioning in the schema design at the application layer.
> Setting the TTL too high could cause problems if the retained versions
> of a single row approach the maximum range size.

Thanks, this organization is def better

I updated the text to basically this, with a couple of tweaks to retain
some add'l details that I think will still be useful. Specifically:

- Added the longer message about non-scheduled incremental backups to
  end of first paragraph (which was previously in the separate info
  block)

- Added the bit about versions being stored in an unsplit range to end
  of second paragraph

PTAL

bdarnell (approved):

ianjevans (approved):

PR #16980, opened 2023-05-15
[DOC-7165] Document --max-go-memory flag

mdlinville:
[DOC-7165] Document --max-go-memory flag

## Previews
[v23.1/cockroach-start-single-node.md](https://deploy-preview-16980--cockroachdb-docs.netlify.app/docs/v23.1/cockroach-start-single-node.html)
[v23.1/cockroach-start.md](https://deploy-preview-16980--cockroachdb-docs.netlify.app/docs/v23.1/cockroach-start.html)

## Tests
- [x] Local build
- [x] Linkcheck


v23.1/cockroach-start-single-node.md
```diff
@@ -41,7 +41,9 @@ Flag | Description
...
+`--cache` | The total size for caches, shared evenly if there are multiple storage devices. This can be a percentage (notated as a decimal or with `%`) or any bytes-based unit, for example: <br><br>`--cache=.25`<br>`--cache=25%`<br>`--cache=1000000000 ----> 1000000000 bytes`<br>`--cache=1GB ----> 1000000000 bytes`<br>`--cache=1GiB ----> 1073741824 bytes` <br><br><strong>Note:</strong> If you use the `%` notation, you might need to escape the `%` sign, for instance, while configuring CockroachDB through `systemd` service files. For this reason, it's recommended to use the decimal notation instead.<br><br>**Note:** The sum of `--cache`,
```
yuzefovich, line 44: Something is off here - the last sentence for `--cache` is incomplete, and that sentence seems to continue after the new `--max-go-memory` section.

v23.1/cockroach-start-single-node.md
```diff
@@ -41,7 +41,9 @@ Flag | Description
...
+`--max-go-memory` | The maximum soft memory limit for the Go runtime, which influences the behavior of Go's garbage collection. Defaults to `--max-sql-memory` x 2.5, but cannot exceed `--cache` x 1.15, and cannot exceed 90% of the node's available RAM. To disable the soft memory limit, set `--max-go-memory` to `0` (not recommended).
```
yuzefovich, line 45: It defaults to `x 2.25`, not `x 2.5`.

Also the relationship between `--max-go-memory` and `--cache` is not direct one like "but cannot exceed `--cache` x 1.15" implies. The formula for calculation is as follows:
1. choose the initial default as `default := --max-sql-memory x 2.25`
2. if `(default + --cache x 1.15) > 90% RAM`, then reduce `default` to `90% RAM - --cache x 1.15`.

If this calculation is too complicated, then perhaps we should just mention `--max-sql-memory x 2.25` as the default value, that is then clamped at 90% of RAM.

yuzefovich (approved):
LGTM, thanks!

v23.1/cockroach-start-single-node.md
```diff
@@ -46,6 +46,7 @@ Flag | Description
...
+`--max-go-memory` | The maximum soft memory limit for the Go runtime, which influences the behavior of Go's garbage collection. Defaults to `--max-sql-memory x 2.25`, but cannot exceed 90% of the node's available RAM. To disable the soft memory limit, set `--max-go-memory` to `0` (not recommended).
```
taroface, line 49: This is less straightforward, but I feel like only the flag should be code-formatted. I don't feel especially strongly about this.

```suggestion
`--max-go-memory` | The maximum soft memory limit for the Go runtime, which influences the behavior of Go's garbage collection. Defaults to 2.25 times `--max-sql-memory`, but cannot exceed 90% of the node's available RAM. To disable the soft memory limit, set `--max-go-memory` to `0` (not recommended).
```

v23.1/cockroach-start.md
```diff
@@ -64,6 +64,7 @@ Flag | Description
...
+`--max-go-memory` | The maximum soft memory limit for the Go runtime, which influences the behavior of Go's garbage collection. Defaults to `--max-sql-memory x 2.25`, but cannot exceed 90% of the node's available RAM. To disable the soft memory limit, set `--max-go-memory` to `0` (not recommended).
```
taroface, line 67: Ditto

taroface (approved):
Just a nit/suggestion

v23.1/cockroach-start-single-node.md
mdlinville: I did it this way for readability so that the formula wouldn't get lost among the text. I think I would like to leave this as is, since you don't feel strongly.

PR #16992, opened 2023-05-16
Updated example output for DROP VIEW

ianjevans:
Fixes DOC-7609.


srosenberg (approved):
Thanks for fixing! LGTM.

PR #17006, opened 2023-05-16
fix link to COPY diagram

taroface:
DOC-7662

The path to the COPY diagram was updated.


taroface:
@kathancox I still need to verify in preview that the diagram correctly displays now. The path is based on https://github.com/cockroachdb/generated-diagrams/blob/release-23.1/grammar_svg/copy.html but it wasn't loading for me on my local build.

shannonbradshaw (approved):
LGTM

PR #17010, opened 2023-05-16
Add the webhook and pub/sub performance cluster setting to docs

kathancox:
Fixes DOC-7349, DOC-7414

Added a note for the new webhook and Pub/Sub cluster settings that allow the performance improvements.

Note: I haven't added setting these embedded in examples because the intention is to backport these settings to be default on.

### DIFF
--- a/_includes/v23.1/cdc/pubsub-performance-setting.md
+++ b/_includes/v23.1/cdc/pubsub-performance-setting.md
@@ -1,0 +1,3 @@
+{{site.data.alerts.callout_info}}
+{% include_cached new-in.html version="v23.1" %} Enable the `changefeed.new_pubsub_sink_enabled` [cluster setting](cluster-settings.html) to improve the throughput of changefeeds emitting to Pub/Sub sinks.
+{{site.data.alerts.end}}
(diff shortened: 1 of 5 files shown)


amruss (approved):
LGTM

_includes/v23.1/cdc/pubsub-performance-setting.md
```diff
@@ -0,0 +1,3 @@
+{{site.data.alerts.callout_info}}
+{% include_cached new-in.html version="v23.1" %} Enable the `changefeed.new_pubsub_sink_enabled` [cluster setting](cluster-settings.html) to improve the throughput of changefeeds emitting to Pub/Sub sinks.
```
lnhsingh, line 2: Would it make sense to link to `changefeed-sinks.html#google-cloud-pub-sub` here?

lnhsingh (approved):
LGTM - just a (non-blocking) question on whether we should cross-link or not.

_includes/v23.1/cdc/pubsub-performance-setting.md
kathancox: This is a good point! I haven't mainly because the areas where this sits is right by sentences referring to the changefeed sinks page. (Or it sits on the Changefeed Sinks page). But, I might as well add an if/else block in case a reader doesn't see the Changefeed Sinks link.

PR #17011, opened 2023-05-16
Update build-netlify.sh

nickvigilante:
Add the Algolia deployment step back to the build script. This also disables the Lighthouse plug-in until we figure out what we want to do with it and so it doesn't produce errors.

### DIFF
--- a/build/build-netlify.sh
+++ b/build/build-netlify.sh
@@ -51,10 +51,10 @@ if [[ "$INCOMING_HOOK_TITLE" = "nightly" ]]; then
 fi;
 
 # Run Algolia if building master
-# if [[ "$CONTEXT" = "production" ]]; then
-# 	echo "Building Algolia index..."
-# 	ALGOLIA_API_KEY=$PROD_ALGOLIA_API_KEY bundle exec jekyll algolia --config _config_base.yml,_config_url.yml --builds-config _config_cockroachdb.yml
-# fi;
+if [[ "$CONTEXT" = "production" ]]; then
+	echo "Building Algolia index..."
+	ALGOLIA_API_KEY=$PROD_ALGOLIA_API_KEY bundle exec jekyll algolia --config _config_base.yml,_config_url.yml --builds-config _config_cockroachdb.yml
+fi;
 
 # Run htmltest, but skip checking external links to speed things up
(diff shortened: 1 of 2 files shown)


build/build-netlify.sh
```diff
@@ -51,10 +51,10 @@ if [[ "$INCOMING_HOOK_TITLE" = "nightly" ]]; then
 fi;
 
 # Run Algolia if building master
-# if [[ "$CONTEXT" = "production" ]]; then
-# 	echo "Building Algolia index..."
-# 	ALGOLIA_API_KEY=$PROD_ALGOLIA_API_KEY bundle exec jekyll algolia --config _config_base.yml,_config_url.yml --builds-config _config_cockroachdb.yml
-# fi;
+if [[ "$CONTEXT" = "production" ]]; then
```
mdlinville, line 54:
```suggestion
if [ "$CONTEXT" == "production" ]; then
```
The non-POSIX version of `test` always looks like a syntax error to me since `==` is usually used to test equality, and there's not much call for it unless you're doing regex pattern-match testing.

build/build-netlify.sh
```diff
@@ -51,10 +51,10 @@ if [[ "$INCOMING_HOOK_TITLE" = "nightly" ]]; then
...
+	echo "Building Algolia index..."
+	ALGOLIA_API_KEY=$PROD_ALGOLIA_API_KEY bundle exec jekyll algolia --config _config_base.yml,_config_url.yml --builds-config _config_cockroachdb.yml
```
mdlinville, line 56:
```suggestion
	ALGOLIA_API_KEY=${PROD_ALGOLIA_API_KEY} bundle exec jekyll algolia --config _config_base.yml,_config_url.yml --builds-config _config_cockroachdb.yml"
```
This syntax is somewhat more thread-safe. Careful to be sure that $PROD_ALGOLIA_API_KEY is actually set unless it being unset wouldn't cause a problem. As it is, if it was empty it would just be the `bundle exec....` bit that runs and might be difficult to debug. Probably not a significant issue, but this code is not resilient against shell-scripting injection such as adding a `sudo` or `rm -rf /;` to it.

mdlinville (approved):

PR #17013, opened 2023-05-17
Add example of moving a column to a new column family

ianjevans:
Fixes DOC-7566.

- [x] add to v22.2 after content review


smcvey (approved):

src/current/v23.1/alter-table.md
```diff
@@ -911,6 +911,50 @@ $ cockroach demo bank
 (1 row)
 ~~~
 
+##### Move a column from one column family to another
+
+To move a column from one column family to another column family, create a temporary, [non-visible](create-table.html#not-visible-property) [stored computed column](computed-columns.html) in the target column family, then rename the columns. Once this succeeds, you can drop the original, now renamed column.
```
rmloveland, line 916: I think it would be good to add a line about why one might want to do this.  I didn't know why one would want to do this but DOC-7566 has it: "say I create a table with just the default primary family. Then later I realize 3 of the 10 columns get frequently updated so I want to move them to their own family."

In fact I would really prefer even more detail than that: _why_ do I care that only 3/10 of the columns get frequently updated?  Is this a substantial optimization? if so, how substantial (roughly)?

src/current/v23.1/alter-table.md
```diff
@@ -911,6 +911,50 @@ $ cockroach demo bank
...
+
+    {% include_cached copy-clipboard.html %}
+    ~~~ sql
+    SET sql_safe_updates = false;
+    ALTER TABLE bank DROP COLUMN old_name;
+    SET sql_safe_updates = true;
+    ~~~
+
+    {{site.data.alerts.callout_info}}
+    You must set the `sql_safe_updates` [session variable](set-vars.html) to `false` to drop a column in a table that has data.
```
rmloveland, line 953: This can be a direct link to the session variable e.g. https://www.cockroachlabs.com/docs/stable/set-vars.html#sql-safe-updates

src/current/v23.1/alter-table.md
rmloveland, line 916: First use of 'column family' could be a link to our overview docs on the subject at https://www.cockroachlabs.com/docs/stable/column-families.html

rmloveland: ok maybe the 2nd question about perf would be answered by the link to https://www.cockroachlabs.com/docs/stable/column-families.html#default-behavior

but I still think it would be good to add a line about the "later I realize 3/10 are frequently updated" etc

src/current/v23.1/alter-table.md
```diff
@@ -911,6 +911,50 @@ $ cockroach demo bank
...
+    {{site.data.alerts.end}}
+
+Moving a column to another column family executes writes to the underlying storage equal to two times the number of rows. For example, if the table has 10 million rows, there will be 20 million writes to the storage layer: 10 million writes when creating the temporary stored computed column, and 10 million writes when removing the original column.
```
rmloveland, line 956: 'storage layer' could be a link to https://www.cockroachlabs.com/docs/stable/architecture/storage-layer.html

if folks care enough to mess with column families they should probably also know about LSMs etc

rmloveland (approved):
LGTM just had some link requests and one ask for a bit of motivation

PR #17029, opened 2023-05-17
Fix issues with sitemap and canonicals

nickvigilante:
Now that v23.1 is GA, both the dev and stable paths point to v23.1. In the sitemap logic and in the page head, it would automatically force a page's canonical to point to the dev URL. This PR fixes that.


JosephNickdow (commented):
Looks like this will fix it @nickvigilante

PR #17039, opened 2023-05-18
DOC-7212 remove restore v21.2 multi-region database

ashleykwtam:
Addresses DOC-7212:
- We currently document restoring a multi-region database to a new database in v21.2 for managed-service backups as this procedure is different than other versions that can use the regular workflow
- This PR removes the v21.2 section as it is no longer relevant

Preview:
- [Use Managed Service Backups](https://deploy-preview-17039--cockroachdb-docs.netlify.app/docs/cockroachcloud/use-managed-service-backups.html?filters=dedicated#restore-a-database)


kathancox (approved):
Great to see this gone! LGTM

PR #17082, opened 2023-05-24
Update query parameters to caps for external connections

kathancox:
Fixes DOC-7599

The issue points to a problem in handling the `COCKROACH_LOCALITY` parameter when running with external connections. The reporter mentioned this should be appended to the external connection when running the backup. However, after testing, it turns out query parameters must be capped when running `CREATE EXTERNAL CONNECTION`. Here is the eng issue to fix that: https://github.com/cockroachdb/cockroach/issues/103779

In the meantime, this PR adds notes to the examples and query parameter sections for changefeeds and backup SQL statements. To standardize query parameters, and where users may copy from docs, this PR caps all of the lists of query parameters across backup/changefeed docs.

Note: there is some consolidation needed of the changefeed query parameter docs (they are listed across multiple places  duplicated). https://cockroachlabs.atlassian.net/browse/DOC-7737


mdlinville (approved):
Looks like your editor is now trimming whitespace. :)

PR #17083, opened 2023-05-24
Update pr-reviews.yml to run on pull request target

nickvigilante:
This will allow for secrets to be used to run the script needed to check for the necessary PR reviews across forks.

Additionally, each time a commit is pushed or auto-merge is enabled, the script will run.

### DIFF
--- a/.github/workflows/pr-reviews.yml
+++ b/.github/workflows/pr-reviews.yml
@@ -1,10 +1,12 @@
 name: PR Review
 on: 
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - reopened
       - ready_for_review
+      - synchronize
+      - auto_merge_enabled
   pull_request_review:
     types:
       - submitted


mikeCRL:
Can you further explain the secrets aspect? My current assumption is that the change will ensure this action to check for appropriate approvals will now run (via `synchronize`) on every commit and (via `auto_merge_enabled`) when auto merge is enabled. Is that right?

nickvigilante:
@mikeCRL In order to query the GitHub GraphQL API, we need to pass in a token into the Go script. This makes sure that we're authenticated and have the privileges to view docs-prs group members and PR reviewers. `pull_request_target` will run this action in each PR on the base branch (master) instead of the HEAD branch (whatever code is running in the PR). GitHub blocks secrets from running in the HEAD branch for security reasons, and my code needs secrets to function correctly.

Your assumption about `synchronize` and `auto_merge_enabled` is correct, but separate from the secrets aspect.

mikeCRL:
Ah, I missed the significance of `pull_request_target` until now, and I appreciate all the context. SGTM/LGTM! Thanks.

PR #17098, opened 2023-05-25
Update _config_cockroachdb.yml

nickvigilante:
Fix a bug where canonicals are not showing correctly


JosephNickdow (approved):
Looks good @nickvigilante - checked a few pages and the sitemap. Everything looks correct.

PR #17105, opened 2023-05-26
Include information about cold starts consuming RUs in docs

gemma-shay:
DOC-7634


gemma-shay:
@alyshanjahani-crl can you review when you get a chance?

alyshanjahani-crl (approved):

cockroachcloud/serverless-faqs.md
```diff
@@ -79,11 +79,13 @@ Yes, your free cluster has been automatically migrated to {{ site.data.products.
...
 
-When automatic statistics collection starts your cluster may consume RUs above the 100 RUs per second baseline when your workload is otherwise consuming RUs below the baseline. You can [turn off automatic statistics collection](../{{site.current_cloud_version}}/cost-based-optimizer.html#enable-and-disable-automatic-statistics-collection-for-clusters) to avoid these RU bursts, but the cost-based optimizer may choose inefficient statement plans as it doesn't have access to the latest statistics.
+When automatic statistics collection starts you may see a spike in RU usage. You can [turn off automatic statistics collection](../{{site.current_cloud_version}}/cost-based-optimizer.html#enable-and-disable-automatic-statistics-collection-for-clusters) to avoid these RU bursts, but the cost-based optimizer may choose inefficient statement plans if it doesn't have access to the latest statistics.
```
mdlinville, line 82:
```suggestion
When automatic statistics collection starts, the cluster may experience a spike in RU usage. You can [disable automatic statistics collection](../{{site.current_cloud_version}}/cost-based-optimizer.html#enable-and-disable-automatic-statistics-collection-for-clusters) to avoid these RU bursts, but this may cause the cost-based optimizer to choose inefficient statement plans based on outdated statistics.
```
Maybe not for this PR, but "RU usage" reads strange to me, since RUs are an abstract unit that shows usage. WDYT? Also, is the spike temporary? periodic? Under what circumstances would you _not_ see the increased RUs? I'm just wondering if we can firm up "may" a bit.

cockroachcloud/serverless-faqs.md
```diff
@@ -79,11 +79,13 @@ Yes, your free cluster has been automatically migrated to {{ site.data.products.
...
+
+Refreshing the Cloud Console will also consume RUs.
```
mdlinville, line 84: Since there could be a lot of clients consuming the Cloud Console, is there any way to help the customer predict how many RUs Cloud Console usage is likely to incur or to figure out if a single client or session is consuming more than expected?

cockroachcloud/serverless-faqs.md
```diff
@@ -79,11 +79,13 @@ Yes, your free cluster has been automatically migrated to {{ site.data.products.
...
 
 ### What is the cold start latency of a {{ site.data.products.serverless }} cluster?
 
-When a {{ site.data.products.serverless }} cluster is idle, it will scale down to zero and consume no RUs. When the cluster becomes active again it will begin serving requests within a fraction of a second, typically around 600 milliseconds.
+When a {{ site.data.products.serverless }} cluster is idle, it will scale down to zero and consume no RUs. When the cluster becomes active again it will begin serving requests within a fraction of a second, typically around 600 milliseconds. Note that cold starts consume RUs, so you may see a spike in usage.
```
mdlinville, line 88: I find myself wondering what it scales down to zero of. I guess it's RUs, since it doesn't really have nodes? What about "will scale down and consume no RUs"? I'm less familiar with Serverless so ignore this if I'm misunderstanding.

mdlinville (approved):
Just a few small suggestions, non-blocking. LGTM!

cockroachcloud/serverless-faqs.md
alyshanjahani-crl: typically its the amount of RUs needed to list the sql users and databases in the cluster (since we populate that info in the connect modal for the cluster).

However, for the same point in the above comment i'd prefer not including numbers.

cockroachcloud/serverless-faqs.md
alyshanjahani-crl: > wondering what it scales down to zero of. 

Good point. I dont think users are aware that scale ups and downs are in reference to sql server instances (kubernetes pods). You could also think of it as scale to 0 meaning 0 CPUs.

cockroachcloud/serverless-faqs.md
mdlinville: Makes sense. It seems more that it is rendering Cloud Console pages than specifically refreshing Cloud Console.

cockroachcloud/serverless-faqs.md
mdlinville: Thanks for clarifying, it makes sense. I guess that it's difficult to observe an idle cluster that is consuming no RUs, if viewing its pages in Cloud Console consumes RUs.

mdlinville (approved):

PR #17118, opened 2023-05-30
Correct version description in release notes

kathancox:
Quick fix to the 23.1 GA release notes. (It should not read two minor versions — it is major versions and not quite two.)

### DIFF
--- a/_includes/releases/v23.1/v23.1.0.md
+++ b/_includes/releases/v23.1/v23.1.0.md
@@ -326,7 +326,7 @@ If the node executing the backup does not have a specific range to be backed up,
   <tr>
    <td><p>Enforce supported backup versions</p>
    </td>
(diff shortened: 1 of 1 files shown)


kathancox:
oh weird, @mikeCRL I just copied from the link above this... it looks like a couple full URLs slipped in, shall I change these all up? Any reason why they're different here?

mikeCRL:
@kathancox @ashleykwtam I believe I responded to this matter off the thread here, but for the record, yes, all links should be relative. This must have been an oversight in the GA Release Notes PR—not sure if it was one of my commits or someone else's, but I believe we don't test for this, so it would have to be caught manually. (Thanks for catching and asking.)

@nickvigilante should this be a Vale consideration? I assume it would've already made sense to enforce internal links' structure as part of our future standardized/pretty URL conversion. Should I create an issue for the URL rules enforcement aspect?

PR #17139, opened 2023-05-31
Add docs on `transaction_timeout` session var

rmloveland:
Fixes DOC-5896

NB. Unlike previous session variables, there is no corresponding `sql.defaults.*` version of this one, since that namespace in the cluster settings is deprecated.


taroface (approved):
LGTM! Sorry I'm late on this; just got back to the office.

PR #17148, opened 2023-06-01
Add Java and C# bulk delete examples

ianjevans:
Working examples of bulk deletes in Java (JDBC) and C# (Npgsql).

I also rearranged the topics to emphasize TTL.

Fixes DOC-6574.


ianjevans:
Ping @dikshant and @ecwall for reviews.

dikshant:
@rafiss can you help review some of the new Java examples here if you have a moment.

src/current/v22.1/bulk-delete-data.md
```diff
@@ -76,22 +98,185 @@ while True:
...
+            } while (cont);
+        }
+        catch(Exception e) {
+            return;
+        }
+    }
+
+    public static void main(String[] args) throws SQLException {
+        Connection connection = null;
+        try {
```
ecwall, line 152: You should be able to use try-with-resources here to automatically close the connection(please double check this works):
```
    public static void main(String[] args) throws SQLException {
        // create the datasource for the JDBC driver
        PGSimpleDataSource ds = new PGSimpleDataSource();
        ds.setApplicationName("docs_bulk_delete_java");
        // get the cluster JDBC URL from an environment variable
        ds.setUrl(Optional.ofNullable(System.getenv("JDBC_DATABASE_URL")).orElseThrow(
              () -> new IllegalArgumentException("JDBC_DATABASE_URL is not set.")));
        try (Connection connection = ds.getConnection()) {
            // call the method to perform the deletes
            deleteData(connection);
        } catch (SQLException e) {
            e.printStackTrace();
        } 
    }
```

src/current/v22.1/bulk-delete-data.md
```diff
@@ -76,22 +98,185 @@ while True:
...
+      DeleteRows(connStringBuilder.ConnectionString);
+    }
+
+    static void DeleteRows(string connString)
+    {
+      // create the data source with the connection string
+      using var dataSource = NpgsqlDataSource.Create(connString);
+      {
+        using var connection = dataSource.OpenConnection();
+        bool cont = true;
```
ecwall, line 210: Use `var` everywhere for consistency (unless this is invalid here - I have not written c# in a long time).

src/current/v22.1/bulk-delete-data.md
```diff
@@ -76,22 +98,185 @@ while True:
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Optional;
+
+import org.postgresql.ds.PGSimpleDataSource;
+
+public class App {
+
+    public static void deleteData(Connection conn) {
+        boolean cont = true;
```
ecwall, line 123: Use `var` for consistency with c# example.

src/current/v22.1/bulk-delete-data.md
```diff
@@ -76,22 +98,185 @@ while True:
...
+                // the warehouse ID of the final row of this batch.
+                warehouseId = reader.GetInt32(0);
+              }
+              Console.WriteLine("Warehouse ID is now " + warehouseId); 
+            }
+            else {
+              // All the rows have been deleted, so break out of the loop
+              cont = false;
+            }
+            reader.Close();
```
ecwall, line 239: `reader` (and `connection` below) should be automatically closed https://stackoverflow.com/questions/5243398/will-a-using-block-close-a-database-connection

src/current/v22.1/bulk-delete-data.md
ianjevans: `var` is used in lieu of the type, so I think it would be cleaner to keep `bool`.

src/current/v22.1/bulk-delete-data.md
ianjevans: Good catch. Removing.

src/current/v22.1/bulk-delete-data.md
ianjevans: Ok, tested it and it works correctly.

dikshant (approved):

PR #17149, opened 2023-06-01
Fix the Files Changed bot

nickvigilante:


.github/workflows/changed_files.yml
```diff
@@ -85,9 +88,10 @@ jobs:
               major_version=${path[-2]}
               file="<li>${file}:<ul><li><a href=\"https://deploy-preview-$pr_num--cockroachdb-docs.netlify.app/docs/releases/${major_version}.html\" target=\"_blank\" rel=\"noopener\">releases/${major_version}.md</a></li></ul></li>"
```
ianjevans, line 89: Do you know the reason why `target="_blank"` isn't working here? It's a nit, but navigating out of the PR is a little annoying.

nickvigilante:
<a href="https://www.google.com" target="_blank">test target blank url</a>

.github/workflows/changed_files.yml
nickvigilante: Seems like a limitation of GitHub.

PR #17162, opened 2023-06-02
Correct the callout for cdc monitoring on main pages

kathancox:
Quick fix to the monitoring callout that should be at the top of pages


kathancox:
@mdlinville I missed these instances in https://github.com/cockroachdb/docs/pull/17084, so this PR updates those with the same language

mdlinville (approved):

PR #17201, opened 2023-06-07
Update v23.1 backcompat language re: shared ranges

rmloveland:
Fixes DOC-7819


knz (approved):

src/current/_includes/releases/v23.1/v23.1.0.md
```diff
@@ -435,9 +435,9 @@ Before [upgrading to CockroachDB v23.1](../v23.1/upgrade-cockroach-version.html)
...
+- CockroachDB now supports sharing storage ranges across multiple indexes/tables. This behavior is not enabled by default in v23.1, but will be enabled by default in a future release. When this feature is enabled, there will no longer be a guarantee that there is at most one SQL object (e.g., table/index/sequence/materialized view) per storage range.  When that happens, the columns `table_id`, `database_name`, `schema_name`, `table_name` and `index_name` in `crdb_internal.ranges` and `.ranges_no_leases` will become meaningless, since a range will not be attributed to a single table/index anymore.  For the v23.1 release, the default behavior of [`SHOW RANGES`](../v23.1/show-ranges.html) is retained, but users should consider setting the [cluster setting `sql.show_ranges_deprecated_behavior.enabled` to `false`](../v23.1/cluster-settings.html#setting-sql-show-ranges-deprecated-behavior-enabled). This will have the following effects that will become the defaults in a future release:
```
taroface, line 438: Just some optional tweaks to clarify (IMO). One thing I'm still unclear on is how the SHOW RANGES default behavior relates to the preceding info about how ranges are stored. Is it possible to be more explicit about this?

```suggestion
- CockroachDB now supports sharing storage ranges across multiple indexes/tables. This behavior is not enabled by default in v23.1, but will be enabled by default in a future release. When the behavior is enabled, there will no longer be a guarantee that there is at most one SQL object (e.g., table/index/sequence/materialized view) per storage range. As a result, the columns `table_id`, `database_name`, `schema_name`, `table_name` and `index_name` in `crdb_internal.ranges` and `.ranges_no_leases` will become meaningless, since a range will no longer be attributed to a single table/index. In v23.1, the default behavior of [`SHOW RANGES`](../v23.1/show-ranges.html) is retained, but you should consider setting the [`sql.show_ranges_deprecated_behavior.enabled`](../v23.1/cluster-settings.html#setting-sql-show-ranges-deprecated-behavior-enabled) cluster setting to `false`. This will have the following effects that will become the defaults in a future release:
```

taroface (changes requested):
Couple suggestions and one small request!

src/current/_includes/releases/v23.1/v23.1.0.md
rmloveland: Thanks, will def take this suggestion!

>  Is it possible to be more explicit about this?

I think for now I'd rather leave it to another doc.  If a user really really wants to get into the weeds of how ranges map to schema objects, I don't think a sentence in this section of the release notes is going to be able to teach them about that.  Also our docs as they currently exist don't really have a clear explanation of exactly how that will work under the new UA world (AIUI).  Having said that I think we _will_ be able to point them to such info once we work on issues like e.g. https://cockroachlabs.atlassian.net/browse/CRDB-28402 and its linked issues!

rmloveland:
hi @taroface is it ok to merge this with those changes added?

taroface (approved):
Sorry I didn't notice the new changes - LGTM!

PR #17239, opened 2023-06-12
Return CDC Query Params to lowercase

kathancox:
Fixes DOC-7928

Right now query parameters for CDC need to be lowercase in order to pass correctly. 

After some testing with creating external connections, it had seemed that all query parameters needed to be capped in order for an external connection to create. However, this is actually only backup and restore related query parameters and S3_STORAGE_CLASS. I capitalized all of the CDC query params in [Update query parameters to caps for external connections by kathancox · Pull Request #17082 · cockroachdb/docs](https://github.com/cockroachdb/docs/pull/17082) but that was actually not correct! If you use CDC query params to create an external connection or a changefeed, it will not work. 

None of this is ideal behavior, and will be worked on from eng so that every param is case insensitive: [cloud: URI parameters should be caps lock agnostic · Issue #103779 · cockroachdb/cockroach](https://github.com/cockroachdb/cockroach/issues/103779) 

But we need to update the CDC params again to lowercase, as a customer has already run into this error since I capitalized everything.

**This doesn't affect cloud storage authentication params.**


mdlinville (approved):
Verified the changes using the preview links across a few random pages and versions. One thing I noticed is that some of the references pages have this info in a heading called **Parameters** and others **Query Parameters**. I wondered if that was intentional or is tech debt to address in the future.

kathancox:
> Verified the changes using the preview links across a few random pages and versions. One thing I noticed is that some of the references pages have this info in a heading called **Parameters** and others **Query Parameters**. I wondered if that was intentional or is tech debt to address in the future.

Oh there definitely might be some tech debt/historical stuff hanging around that I could do with clearing up. Though, I think on some of the SQL pages the Parameters section is meant to represent the parameters in the SQL diagram (the parameters of the SQL statement. And then, query parameters are those that append to the URI of the sink/storage for DR/CDC. So, with that being said, I think it could be clarified. I'll open a docs-fixit for this.

PR #17244, opened 2023-06-13
Add more detail to envelope options for changefeeds

kathancox:
Fixes DOC-7635

Improved the documentation for the `envelope` option for changefeeds.

- Added the missing `bare` value
- Corrected sink compatibility
- Incorporated a more detailed section to the Changefeed Messages page so there is more coverage of what customers can expect from the different options here

This is one of several issues to improve the changefeed message documentation.


src/current/v23.1/cdc-queries.md
```diff
@@ -100,7 +100,9 @@ You can **not** use the following functions with CDC queries:
 
 CDC queries allow you to customize your changefeed for particular scenarios. This section outlines several possible use cases for CDC queries.
 
-See [`CREATE CHANGEFEED`](create-changefeed.html) for examples on using the foundational syntax to create a changefeed. For information on sinks, see the [Changefeed Sinks](changefeed-sinks.html) page.
+CDC queries use `envelope=bare` message format by default. This omits the `after` key and stores any [metadata in the `crdb` field](#filter-columns). Depending on how you are filtering or adapting the message format with a CDC query and which sink you're emitting to, message format can vary from some of the example cases in this section.
```
HonoreDB, line 103: `omits the "after" key` could maybe be misread as saying that it omits the values that would be under the "after" key. I don't know a terse way of fixing that though, something more precise would be `puts the output of the SELECT clause at the top level of the message instead of under an "after" key`.

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,103 @@ This page describes the format and behavior of changefeed messages. You will fin
...
 
-When a changefeed targets a table with multiple column families, the family name is appended to the table name as part of the topic. See [Tables with columns families in changefeeds](changefeeds-on-tables-with-column-families.html#message-format) for guidance.
-
-For webhook sinks, the response format arrives as a batch of changefeed messages with a `payload` and `length`. Batching is done with a per-key guarantee, which means that messages with the same key are considered for the same batch. Note that batches are only collected for row updates and not [resolved timestamps](create-changefeed.html#resolved-option):
+For [webhook sinks](changefeed-sinks.html#webhook-sink), the response format arrives as a batch of changefeed messages with a `payload` and `length`. Batching is done with a per-key guarantee, which means that messages with the same key are considered for the same batch. Note that batches are only collected for row updates and not [resolved timestamps](create-changefeed.html#resolved-option):
```
HonoreDB, line 50: Sorry, not part of this PR, but I'm noticing that I don't understand what "messages with the same key are considered for the same batch" means.

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,103 @@ This page describes the format and behavior of changefeed messages. You will fin
...
+
+{{site.data.alerts.callout_info}}
+Changefeeds created with [`EXPERIMENTAL CHANGEFEED FOR`](changefeed-for.html) or [`CREATE CHANGEFEED`](create-changefeed.html) with no sink specified (sinkless changefeeds) produce messages without the envelope metadata fields of changefeeds emitting to sinks.
+{{site.data.alerts.end}}
+
+The following sections provide examples of changefeed messages that are emitted when you specify each of the supported `envelope` options. Other [changefeed options](create-changefeed.html#options) can affect the message envelope and the what messages are emitted. Therefore, the examples are a guide for what you can expect when only the `envelope` option is specified.
+
+### `wrapped`
+
+`wrapped` is the default envelope structure for changefeed messages. This envelope contains an array of the primary key, a top-level field for the type of message, and the current state of the row (or `null` for [deleted rows](#delete-messages)). Kafka sinks do not contain an array of the primary key because the key is part of the message metadata.
```
HonoreDB, line 70: The array of the primary key is there when using a sink that doesn't have a message key as part of its protocol: cloud storage or webhook. Kafka and gcpubsub don't have it by default, but this is what the `key_in_value` option does when provided for those sinks.

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,103 @@ This page describes the format and behavior of changefeed messages. You will fin
...
+
+- Cloud storage sink:
+
+    ~~~
+    {"__crdb__": {"key": ["washington dc", "cd48e501-e86d-4019-9923-2fc9a964b264"]}, "city": "washington dc", "creation_time": "2019-01-02T03:04:05", "current_location": "87247 Diane Park", "ext": {"brand": "Fuji", "color": "yellow"}, "id": "cd48e501-e86d-4019-9923-2fc9a964b264", "owner_id": "a616ce61-ade4-43d2-9aab-0e3b24a9aa9a", "status": "available", "type": "bike"}
+    ~~~
+
+### `key_only`
+
+`key_only` emits only the key and no value, which is faster if you only need to know the key of the changed row. This envelope option is only supported for [Kafka sinks](changefeed-sinks.html#kafka) or sinkless changefeeds.
```
HonoreDB, line 96: Should be supported for gcpubsub too.

HonoreDB (commented):
These are great! Some comments on the wording. In general the intent of our envelope/sink/format options (there are some gaps in execution) is that the primary key is always there _somewhere_. When we're using a protocol and format that has a specific spot for a key, we use that. Otherwise, we require an envelope that has room for metadata so that we can put it there.

src/current/v23.1/changefeed-messages.md
kathancox: Hmmm, re-reading not sure what I meant by this when I introduced webhooks into the docs. I assume I was trying to get at how batches are evaluated? Do you know how that would involve the key?

src/current/v23.1/changefeed-messages.md
kathancox: I tested this and was getting an incompatible sink error with Pub/Sub?

HonoreDB: Ack, sorry. I believe you but I'm not sure why it's not supported.

src/current/v23.1/changefeed-messages.md
HonoreDB: Okay, I think I get it. Webhook sinks obey the same per-key ordering guarantee as other sinks, so they can't make multiple requests in parallel if they contain different events concerning the same primary key--then the order the server received them would be nondeterministic. This is true whether or not the messages are being batched--if you have batching turned off, you might get a message about k1 and k2 at the same time, but you won't get another message about k1 until you've acknowledged the first one. If the messages are being batched, you won't get two batches at the same time with overlapping keys, but you might get a single batch containing two messages about k1, since you can tell the order from their position within the batch.

src/current/v23.1/changefeed-messages.md
kathancox: Oh yeah, honestly I tried again, because I worried I misremembered. I was testing on a serverless cluster to gcpubsub — should I write up an issue for this?

src/current/v23.1/cdc-queries.md
kathancox: I've adjusted this a little to hopefully get more clarity in this. Let me know if you think it needs some more adjustment.

src/current/v23.1/changefeed-messages.md
kathancox: I have added more detail based on your feedback. I think I have captured a better few sentences here. I do have an issue coming up to improve the ordering guarantee + schema change/backfill parts of this message page, so I hope to make this kind of info even clearer.

src/current/v23.1/changefeed-messages.md
kathancox: OK, I have added more detail on this and also included an example of `wrapped` + `key_in_value` for Kafka.

src/current/v23.1/cdc-queries.md
kathancox: @HonoreDB Actually on this, I think I need to add more description... we only have the `crdb` field for CDC queries when there is additional information that cockroach is sending so like updated, resolved, otherwise the messages will actually be completely "bare" like 

```
{"city": "seattle", "type": "skateboard"}
{"city": "washington dc", "type": "bike"}
{"city": "washington dc", "type": "scooter"}
{"city": "washington dc", "type": "bike"}
{"city": "washington dc", "type": "skateboard"}
{"city": "washington dc", "type": "scooter"}
{"city": "washington dc", "type": "scooter"}
```

kathancox: I hadn't understood this, having just updated the example that contained the `updated` option... so will add to this now

src/current/v23.1/create-changefeed.md
```diff
@@ -161,15 +161,15 @@ Option | Value | Description
...
-<a name="key-in-value"></a>`key_in_value` | N/A | Make the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for [cloud storage sinks](changefeed-sinks.html#cloud-storage-sink), [webhook sinks](changefeed-sinks.html#webhook-sink), and [GC Pub/Sub sinks](changefeed-sinks.html#google-cloud-pub-sub). {% include {{ page.version.version }}/cdc/cloud-storage-external-connection.md %}
+<a name="key-in-value"></a>`key_in_value` | N/A | Add a primary key array to the emitted message. This makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for [cloud storage sinks](changefeed-sinks.html#cloud-storage-sink), [webhook sinks](changefeed-sinks.html#webhook-sink), and [GC Pub/Sub sinks](changefeed-sinks.html#google-cloud-pub-sub).
```
kathancox, line 172: @HonoreDB I hope this addition here makes sense? I found the previous definition didn't really capture what we had been talking about regarding the primary key.

kathancox:
@HonoreDB Could you take a look at my updates from your review? Also, let me know if you want me to file any issues around what I found with PubSub or whether we're expecting these results right now. Thanks!

src/current/v23.1/changefeed-messages.md
HonoreDB: It's very possible we made this decision for a good reason and I'm just misremembering, sorry. Don't think we need an issue for it, we'll just revisit it if we need other options for some reason.

src/current/v23.1/create-changefeed.md
HonoreDB: Makes sense to me!

HonoreDB (approved):

kathancox:
TFTR @HonoreDB !

kathancox:
cc @mikeCRL

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,137 @@ This page describes the format and behavior of changefeed messages. You will fin
 
 ## Responses
 
-By default, changefeed messages emitted to a [sink](changefeed-sinks.html) contain keys and values of the watched table entries that have changed, with messages composed of the following fields:
+By default, changefeed messages emitted to a [sink](changefeed-sinks.html) contain keys and values of the watched table rows that have changed. The message will contain the following fields depending on the type of emitted change and the [options](create-changefeed.html#options) you specified to create the changefeed:
 
-- **Key**: An array always composed of the row's `PRIMARY KEY` field(s) (e.g., `[1]` for `JSON` or `{"id":{"long":1}}` for Avro).
+- **Key**: An array composed of the row's `PRIMARY KEY` field(s) (e.g., `[1]` for `JSON` or `{"id":{"long":1}}` for Avro).
```
ashleykwtam, line 28:
```suggestion
- **Key**: An array composed of the row's `PRIMARY KEY` field(s) (e.g., `[1]` for `JSON` or `{"id":{"long":1}}` for `avro`).
```

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,137 @@ This page describes the format and behavior of changefeed messages. You will fin
...
+
+## Message envelopes
+
+The _envelope_ defines the structure of a changefeed message. You can use the [`envelope`](create-changefeed.html#envelope) option to manipulate the changefeed envelope. The values that the `envelope` option accepts are compatible with different [changefeed sinks](changefeed-sinks.html), and the structure of the message will vary depending on the sink.
+
+{{site.data.alerts.callout_info}}
+Changefeeds created with [`EXPERIMENTAL CHANGEFEED FOR`](changefeed-for.html) or [`CREATE CHANGEFEED`](create-changefeed.html) with no sink specified (sinkless changefeeds) produce messages without the envelope metadata fields of changefeeds emitting to sinks.
+{{site.data.alerts.end}}
+
+The following sections provide examples of changefeed messages that are emitted when you specify each of the supported `envelope` options. Other [changefeed options](create-changefeed.html#options) can affect the message envelope and the what messages are emitted. Therefore, the examples are a guide for what you can expect when only the `envelope` option is specified.
```
ashleykwtam, line 68:
```suggestion
The following sections provide examples of changefeed messages that are emitted when you specify each of the supported `envelope` options. Other [changefeed options](create-changefeed.html#options) can affect the message envelope and what messages are emitted. Therefore, the examples are a guide for what you can expect when only the `envelope` option is specified.
```

src/current/v23.1/changefeed-messages.md
```diff
@@ -23,35 +23,137 @@ This page describes the format and behavior of changefeed messages. You will fin
...
+
+### `wrapped`
+
+`wrapped` is the default envelope structure for changefeed messages. This envelope contains an array of the primary key (or the key as part of the message metadata), a top-level field for the type of message, and the current state of the row (or `null` for [deleted rows](#delete-messages)).
+
+The message envelope contains a primary key array when your changefeed is emitting to a sink that does not have a message key as part of its protocol, e.g., cloud storage, webhook sinks, or Google Pub/Sub. By default, messages emitted to Kafka sinks do not have the primary key array, because the key is part of the message metadata. If you would like messages emitted to Kafka sinks to contain a primary key array, you can use the [`key_in_value`](create-changefeed.html#key-in-value) option. Refer to the following message outputs for examples of this.
```
ashleykwtam, line 74: nit: can leave as is

```suggestion
The message envelope contains a primary key array when your changefeed is emitting to a sink that does not have a message key as part of its protocol, (e.g., cloud storage, webhook sinks, or Google Pub/Sub). By default, messages emitted to Kafka sinks do not have the primary key array, because the key is part of the message metadata. If you would like messages emitted to Kafka sinks to contain a primary key array, you can use the [`key_in_value`](create-changefeed.html#key-in-value) option. Refer to the following message outputs for examples on this.
```

src/current/_includes/v23.1/cdc/bare-envelope-cdc-queries.md
```diff
@@ -0,0 +1 @@
+{% if page.name == "cdc-queries.md" %}CDC queries{% else %}[CDC queries](cdc-queries.html){% endif %} use [`envelope=bare`](create-changefeed.html#envelope) message format by default. The `bare` message envelope will place the output of the `SELECT` clause at the top level of the message instead of under an `"after"` key. When there is additional information that the changefeed is sending, such as [`updated`](create-changefeed.html#updated-option) or [`resolved`](create-changefeed.html#resolved-option) timestamps, the messages will include a `crdb` field containing this information.
```
ashleykwtam, line 1:
```suggestion
{% if page.name == "cdc-queries.md" %}CDC queries{% else %}[CDC queries](cdc-queries.html){% endif %} use [`envelope=bare`](create-changefeed.html#envelope) message format by default. The `bare` message envelope places the output of the `SELECT` clause at the top level of the message instead of under an `"after"` key. When there is additional information that the changefeed is sending, such as [`updated`](create-changefeed.html#updated-option) or [`resolved`](create-changefeed.html#resolved-option) timestamps, the messages will include a `crdb` field containing this information.
```

ashleykwtam: Would it be more clear to say "instead of below"? The usage of under here confuses me

src/current/v23.1/create-changefeed.md
ashleykwtam, line 172:
```suggestion
<a name="key-in-value"></a>`key_in_value` | N/A | Adds a primary key array to the emitted message. This makes the [primary key](primary-key.html) of a deleted row recoverable in sinks where each message has a value but not a key (most have a key and value in each message). `key_in_value` is automatically used for [cloud storage sinks](changefeed-sinks.html#cloud-storage-sink), [webhook sinks](changefeed-sinks.html#webhook-sink), and [GC Pub/Sub sinks](changefeed-sinks.html#google-cloud-pub-sub).
```

ashleykwtam (changes requested):
Hi Kathryn, submitting this preliminary round of reviews/suggestions on copy basics first!

src/current/_includes/v23.1/cdc/bare-envelope-cdc-queries.md
kathancox: I think when we're talking about fields in a JSON message, under makes sense

src/current/v23.1/changefeed-messages.md
kathancox: In the Avro section lower on this page, I only use the `format` value `avro` when talking about the record type. So `JSON` should come out of the backticks here.

src/current/v23.1/create-changefeed.md
kathancox: Changed the agreement of the previous option description here, "Overrides" to "Override" because this matches the general consistency of this table. It is also more of a call to action for the user when it is in imperative like this.

ashleykwtam: Works for me!

src/current/v23.1/changefeed-messages.md
ashleykwtam: Sounds good!

kathancox:
Thank you for finding those changes @ashleykwtam ! Look forward to the next one :)

PR #17258, opened 2023-06-14
DOC-7927 update cloud cluster restore

ashleykwtam:


adityamaru (changes requested):
This looks good to me. There was some discussion in https://cockroachlabs.slack.com/archives/C04K6JU2BN1/p1686842400313369 to add a Known Limitations section is it worth doing that in this PR? The following limitations come to mind:

- Cannot restore dedicated to serverless and vice versa.
- Self-service cluster restore is disabled for CMEK clusters.
- Restores on AWS that take longer than 36 hours may run into authentication errors because of expired credentials.

@livlobo to confirm if this list looks good!

adityamaru (approved):
discussed offline, known limitations can go as a follow up

livlobo (changes requested):
Hi Ashley- the basic flow looks good.

Can you add the "Notes" from this deck?
https://docs.google.com/presentation/d/15hKbn1c3I2M5HCrpA21qrd3WWSsNi6VXhu5iQwO-Rbo/edit#slide=id.g24d52189200_0_14

And also mention that - cross-cluster restore is not available for CMEK clsuters, yet.

ashleykwtam:
Hi @livlobo I added a Known limitations section and drew from the Notes in the slidedeck you provided. @adityamaru is there a tracking issue out for these known limitations? Please let me know if there is anything missing from this list, thanks!

livlobo (approved):
LGTM

src/current/cockroachcloud/use-managed-service-backups.md
```diff
@@ -139,10 +139,18 @@ To restore a cluster:
...
+
+    {{site.data.alerts.callout_info}}
+    Only active clusters are displayed. You can perform a cross-cluster restore across clusters that belong to the same organization. Incompatible versions cannot be selected and restoring {{ site.data.products.dedicated }} to {{ site.data.products.serverless }} or vice versa does not work. 
+    {{site.data.alerts.end}}
+
 1. Click **Continue**.
 
+1. Enter the name of the destination cluster. 
+
     {{site.data.alerts.callout_danger}}
```
MichaelTrestman, line 152: From a security perspective, I'd like to see this warning higher up (maybe at the top of the "restore" section). I'd also like to see it noted that *because* of how destructive and disruptive this operation is (anyone who has it or can get the API for a service account that has can hurt your business reeeal bad), it should be performed extremely carefully, and the roles that are capable of performing it should be restricted with extreme care according to the POLP https://en.wikipedia.org/wiki/Principle_of_least_privilege

MichaelTrestman: Also, for orgs that have opted in to fine-grained access--and very soon for all orgs--it is now possible to perform these operations with the new Cluster Operator and Cluster Admin role, as well as Org Admin and the deprecated Org admin (legacy)--which corresponds to what is now Org Admin (yeah, this is confusing...).
Could you please add that? I.e., on line 118, instead of "Org Administrators..." it should say like "Users with the Org Administrator, Org Administrator (legacy), Cluster Operator or Cluster Admin roles can..."
Here is where the links to those roles are:
https://www.cockroachlabs.com/docs/cockroachcloud/authorization#cluster-operator

For context, the explanation about the access model transition is at the top of the page there: https://www.cockroachlabs.com/docs/cockroachcloud/authorization

src/current/cockroachcloud/use-managed-service-backups.md
```diff
@@ -256,6 +264,17 @@ To back up a self-hosted CockroachDB cluster into a {{ site.data.products.db }}
     RESTORE DATABASE example_database FROM '2021/03/23-213101.37' IN 'gs://{bucket name}/{path/to/backup}?AUTH=specified&CREDENTIALS={encoded key}';
     ~~~
 
+## Known limitations
+
+- For [restoring a cluster](#restore-a-cluster):
+    - Restoring a backup taken on cluster running a newer version of CockroachDB into a cluster that is on an older version does not work. See [Restoring Backups Across Versions](../{{site.current_cloud_version}}/restoring-backups-across-versions.html).
```
MichaelTrestman, line 270:
```suggestion
    - Restoring a backup taken on cluster running a newer version of CockroachDB into a cluster that is on an earlier version does not work. See [Restoring Backups Across Versions](../{{site.current_cloud_version}}/restoring-backups-across-versions.html).  
```

MichaelTrestman (changes requested):
Left a few suggestions for strengthening the security awareness around this really critical function

MichaelTrestman (approved):
looks good!

PR #17259, opened 2023-06-14
DOC-7876: Update performance-tuning recipes for unused and missing indexes recommendations

florence-crl:
Addresses: DOC-7876 

(A) Updated Performance Tuning Recipes page, (1) added references to Schema Insights tab to drop unused indexes and create missing indexes. (2) Fixed right TOC by adding hyphens to liquid delimiters. (B) Updated Insights page, (1) changed link to point to Index Recommendations section of Databases page, (2) fixed table format. (C) Updated Transactions page, fixed bullet list format.

### DIFF
--- a/src/current/v22.2/performance-recipes.md
+++ b/src/current/v22.2/performance-recipes.md
@@ -107,7 +107,7 @@ Full table scans often result in poor statement performance.
     FROM crdb_internal.node_statement_statistics
     WHERE full_scan = true;
     ~~~
-* Viewing the statement plan on the [Statement details page](ui-statements-page.html#statement-fingerprint-page) in the DB Console indicates that the plan contains full table scans.
+* Viewing the statement plan on the [**Statement Fingerprint** page](ui-statements-page.html#statement-fingerprint-page) in the DB Console indicates that the plan contains full table scans.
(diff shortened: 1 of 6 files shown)


lnhsingh (approved):

src/current/v22.2/performance-recipes.md
```diff
@@ -117,6 +117,8 @@ Not every full table scan is an indicator of poor performance. The [cost-based o
 
 [Examine the statements](sql-tuning-with-explain.html) that result in full table scans and consider adding [secondary indexes](schema-design-indexes.html#create-a-secondary-index).
 
+In the DB Console, visit the [**Schema Insights** tab](ui-insights-page.html#schema-insights-tab) on the [**Insights** page](ui-insights-page.html) and check if there are any insights to create missing indexes. A missing index may cause a statement to have a [suboptimal plan](ui-insights-page.html#statement-execution-details).
```
kevin-v-ngo, line 120: Nit: Perhaps call out that these missing index recommendations are generated **based on slow executions**

Not all full table scans are necessarily bad as this document calls out but if the execution was slow (based on the insights threshold), then it's likely the recommendation is valid and if the plan had a full table scan, it's likely that it should be removed with an index.

kevin-v-ngo (approved):
Small nit. Looks great Florence!

PR #17261, opened 2023-06-14
DOC-7744 add recommendations regarding incremental backups for v22.2

ashleykwtam:
Addresses DOC-7744:
- Documents a limit of 48 incremental backups for v22.2 as discussed in this [thread](https://cockroachlabs.slack.com/archives/C2C5FKPPB/p1685010673145739)
- Most of the language is borrowed from v23.1 work that was done. Added examples under the 48 incremental backup limit

Preview
-  [incremental backup frequency](https://deploy-preview-17261--cockroachdb-docs.netlify.app/docs/v22.2/take-full-and-incremental-backups#recommendations-for-incremental-backup-frequency)


src/current/v22.2/take-full-and-incremental-backups.md
```diff
@@ -149,12 +149,20 @@ To view the available backup subdirectories, use [`SHOW BACKUPS`](show-backup.ht
...
 
 If you are taking backups on a regular cadence, we recommend [creating a schedule](create-schedule-for-backup.html) for your backups.
 
+### Recommendations for incremental backup frequency
+
+Incremental backups form chains between full backups. Each incremental backup contains only the data that has changed since a base set of backups. This base set of backups must include one full backup and can include many incremental backups, which are smaller and faster to produce than full backups. You can take incremental backups either as of a given timestamp or with full [revision history](take-backups-with-revision-history-and-restore-from-a-point-in-time.html).
+
+Cockroach Labs recommends taking incremental backups every 10 minutes. CockroachDB supports up to 48 incremental backups between full backups. This may vary based on your specific use-case, so we recommend testing within your own environment and workloads. This can look like:
```
MichaelTrestman, line 160: None of the schedules you list below meet the recommendation of 10 minute. Also, if CRDB only supports 48 incrementals/day = 30 min/incremental, than how is it even possible to meet that recommendation without doing more than one full/day? I feel like I'm missing something or else the math doesn't add up?

MichaelTrestman (changes requested):
The recommendation seems strangely at odds with with the discussion of backup scheduling, seems like this needs another look?

src/current/v22.2/take-full-and-incremental-backups.md
MichaelTrestman: Also, that is a pretty extreme backup schedule (10 min), which would only ever be appropriate for critical production data (you certainly wouldn't want to wast the juice and storage to backup junk data every 10 min right?

kathancox:
Hey @ashleykwtam, Perhaps Liv could provide some detail here on how the recommendations around the 10-minute mark may differ between v23.1 and v22.2? 

I like the example cases here. Michael also makes a good point that in v22.2 that incremental recommendation does not match up with the 48 total.

Maybe not for this PR, but defining a full/incremental schedule for production-critical workloads vs non-critical might be worthwhile here.

src/current/v22.2/take-full-and-incremental-backups.md
ashleykwtam: Perhaps it would make sense to remove the recommendation on 10 minutes? That was the recommendation given for v23.1 but for 22.2 it'd be hard for CockroachDB to get down to 10 minute intervals since it can only do 48 incrementals between fulls.

ashleykwtam:
@livlobo wanted your thoughts on this - we wish to document incrementals for 22.2 but the recommendation of taking backups every 10 minutes was for 23.1 and doesn't make sense for 22.2 given the limitations of CRDB. Would we want to include a recommendation or just document the incremental intervals without? Thanks!

MichaelTrestman (approved):
LGTM!

PR #17230, opened 2023-06-09
Release notes v23.1.3

kathancox:

PR #17276, opened 2023-06-15
Fix to a version number in v23.1.3 release notes

kathancox:
There was an error in the version number from the original PR release note, as reported by @msirek (corrected here)

Also, looking back over the release notes, during the review process of the v23.1.3 release notes other errors in these version numbers were introduced.

https://github.com/cockroachdb/docs/pull/17230


src/current/_includes/releases/v23.1/v23.1.3.md
```diff
@@ -76,7 +76,7 @@ A [bug](https://github.com/cockroachdb/cockroach/issues/104798) was discovered i
...
 - Fixed a bug in v23.1.0 where the `node_id` field would be omitted from logs. The `node_id` value has now been restored to the logs. [#103798][#103798]
 - Fixed a bug where the collection of `KV bytes read` and `KV gRPC calls` execution statistics during [`EXPLAIN ANALYZE`](../v23.1/explain-analyze.html) could be incorrect (it would remain at zero) in some cases. The bug was introduced in the v23.1.0 release. [#104194][#104194]
-- Fixed an incorrect results bug in v22.1.11 and v23.1.3 and above when the `optimizer_use_improved_computed_column_filters_derivation` [session setting](../v23.1/set-vars.html) is `true`. Predicates on [computed columns](../v23.1/computed-columns.html) are derived when an ORed predicate on a column in the computed column expression is present. [#104276][#104276]
+- Fixed an incorrect results bug in v22.2.10 and v23.1.2 and above when the `optimizer_use_improved_computed_column_filters_derivation` [session setting](../v23.1/set-vars.html) is `true`. Predicates on [computed columns](../v23.1/computed-columns.html) are derived when an ORed predicate on a column in the computed column expression is present. [#104276][#104276]
```
mdlinville, line 79:
```suggestion
- Fixed an incorrect results bug introduced in v23.1.2 when the `optimizer_use_improved_computed_column_filters_derivation` [session setting](../v23.1/set-vars.html) is `true`. Predicates on [computed columns](../v23.1/computed-columns.html) are derived when an ORed predicate on a column in the computed column expression is present. [#104276][#104276]
```

kathancox: My feeling is that it is not that confusing to also include v22.2.10 here. It is possible that someone could be doing that upgrade and looking through these notes.

msirek: Note that the bug only occurs when the new session setting (introduced in 22.2.10 and 23.1.2) `optimizer_use_improved_computed_column_filters_derivation` is set to `true`.  It defaults to `false` in those versions, so customers would only be impacted if they manually set this to `true`.  Feel free to adjust/rewrite the release notes if they're confusing.

kathancox: OK thanks Mark. I think we can probably go ahead with just the update version numbers in the PR. However, since we're kind of talking about the setting (when `true`) and the bug being introduced at the same time. Perhaps we can lead with when the bug occurs, and mention when the setting was introduced after.

> Fixed an incorrect results bug when the `optimizer_use_improved_computed_column_filters_derivation` [session setting](../v23.1/set-vars.html) is `true`. Predicates on [computed columns](../v23.1/computed-columns.html) are derived when an ORed predicate on a column in the computed column expression is present. This setting was introduced in v22.2.10 and v23.1.2. [#104276][#104276]

I'll merge the original in the PR though later today unless anyone has objections.

PR #17290, opened 2023-06-16
Remove overly prescriptive deprecation phrasing

rmloveland:
Fixes DOC-8128

This feature of sharing ports between SQL and RPC conns still works in the v23.1 release cycle and has not been removed (yet), so this PR updates the language to avoid confusion for users who carefully read the release notes.

While there, also audited other uses in the release notes of "will be removed" and the rest look OK and/or are accurate based on some quick testing with v23.1 (AFAICT).


rmloveland:
superseded by https://github.com/cockroachdb/docs/pull/17294

PR #17292, opened 2023-06-16
DOC-8112: Move redirects to YML

nickvigilante:
Fixes DOC-8112


src/current/_data/redirects.yml
```diff
@@ -0,0 +1,841 @@
...
+- destination: create-a-serverless-cluster.html
+  sources: ['create-a-free-cluster.html']
+  versions: ['cockroachcloud']
+
+- destination: learn-about-pricing.html
+  sources: ['planning-your-serverless-cluster.html']
+  versions: ['cockroachcloud']
+
+- destination: managing-access.html
+  sources: ['console-access-management.html', 'user-authorization.html']
```
mdlinville, line 97:
```suggestion
  sources: [
      'console-access-management.html',
      'user-authorization.html'
    ]
```
WDYT for readability and maintenance?

mdlinville (approved):
Format looks great. When you add the usage instructions to the wiki, consider leaving a note at the top of this file to request to keep the destinations alphabetized within a grouping keep the sources alphabetized within a destination, and if the list of sources gets long, put each on its own line.

src/current/_data/redirects.yml
```diff
@@ -227,40 +170,80 @@
...
+  - demo-data-replication.html
   versions: ['v19.2', 'v20.1', 'v20.2', 'v21.1']
 
 - destination: example-apps.html
   sources: ['build-an-app-with-cockroachdb.html']
   versions: ['v21.2']
 
+- destination: fips.html
+  sources: ['fips-compliance.html']
+  versions: ['v23.1']
```
mdlinville, line 211: This is weird because it will also apply to all future versions. I wonder if the version should just be 'stable' here.

nickvigilante: If for some reason we ever link to `v23.1/fips-compliance.html` or `dev/fips-compliance.html`, I believe this kind of redirect is necessary.

mdlinville (approved):
Let's merge and see what shakes out!

PR #17340, opened 2023-06-26
follow up fixes for IP allowlisting for serverless

MichaelTrestman:


src/current/cockroachcloud/network-authorization.md
```diff
@@ -15,12 +15,19 @@ You can authorize network access to your cluster by:
...
     - Enhanced network security (no access through public IPs i.e. no transit over public networks).
     - Direct connection from application deployment that do not have static public IPs
-    - Reduced lower network latency.
+    - Reduced network latency.
     - This option can also be helpful if you need more than the current maximum limit of authorized networks per cluster.
 
+**Prerequisite**: {% include cockroachcloud/cluster-operator-prereq.md %}
+
+{{site.data.alerts.callout_success}}
+You should use PrivateLink or VPC peering if you need to allowlist more IP addresses than allowed by the maximum (20 for Dedicated clusters and 50 for Serverless), if your servers’ IP addresses are not static, or if you want to limit your cluster's exposure to the public internet.
```
kathancox, line 26: I don't know how this would look in this callout box... (and this in the intended meaning!), but what about a structure like this:

> You should use PrivateLink or VPC peering if:

> - You need to allowlist more IP addresses than allowed by the maximum (20 for Dedicated clusters and 50 for Serverless).
> - Your servers’ IP addresses are not static.
> - You want to limit your cluster's exposure to the public internet.

src/current/cockroachcloud/network-authorization.md
```diff
@@ -41,10 +48,20 @@ Authorized network access can be managed from the {{ site.data.products.db }} co
 
 `https://cockroachlabs.cloud/cluster/{ your cluster UUID}/networking`
 
-You can add up to 20 IP addresses to your allowlist. If your application servers’ IP addresses are not static, or you want to limit your cluster's exposure to the public network, you can use VPC Peering or AWS PrivateLink instead (Dedicated clusters only).
+Serverless and Dedicated clusters support different maximum number of IP allowlist rules/authorized sub-networks:
```
kathancox, line 51: Maybe an "a" before "different" here? "... support a different maximum number of IP..."

src/current/cockroachcloud/network-authorization.md
```diff
@@ -41,10 +48,20 @@ Authorized network access can be managed from the {{ site.data.products.db }} co
...
+Dedicated|20
+Serverless|50
 
+{{site.data.alerts.callout_info}}
+While developing and testing your application, you may add `0.0.0.0/0` to the allowlist, which allows all networks. However, before moving into production, make sure you delete the `0.0.0.0/0` network.
+{{site.data.alerts.end}}
+
+If your application servers’ IP addresses are not static, or you want to limit your cluster's exposure to the public network, you can connect to your {{ site.data.products.dedicated }} clusters using VPC Peering or AWS PrivateLink instead.
+
+Refer to:
```
kathancox, line 64: These points are not rendering correctly. I think you may need to leave a line space in the markdown after "Refer to:"

kathancox (approved):
One bullet list isn't rendering and a couple of suggestions, otherwise LGTM.

src/current/cockroachcloud/network-authorization.md
MichaelTrestman: thanks!

src/current/cockroachcloud/network-authorization.md
MichaelTrestman: thanks

PR #17363, opened 2023-06-28
DOC-8144: Document new automatic CPU profiler cluster settings

florence-crl:
Addresses: DOC-8144

Created Automatic CPU Profiler page. Updated Troubleshooting Overview page with link to CPU Profiler page, debug zip and debug tsdump.

### DIFF
--- a/src/current/v23.1/automatic-cpu-profiler.md
+++ b/src/current/v23.1/automatic-cpu-profiler.md
@@ -1,0 +1,81 @@
+---
+title: Automatic CPU Profiler
+summary: The Automatic CPU Profiler captures CPU profiles automatically to help investigate cluster issues involving CPU load.
+keywords: cpu profile, cpu profiles, cpu profiler, automatic cpu profiler
+toc: true
+docs_area: manage
+---
+
+{% include_cached new-in.html version="v23.1" %} This feature automatically captures CPU profiles, which can make it easier to investigate and troubleshoot spikes in CPU usage or erratic CPU load on certain nodes. A CPU profile shows the functions that use the most CPU time, sampled over a window of time. You can collect a CPU Profile manually on the [Advanced Debug page](ui-debug-pages.html). However, it may be difficult to manually capture a profile of a short CPU spike at the right point in time. Automatic CPU profile capture enables the investigation of CPU load in this and other cases, such as periodic high CPU.
+
+{{site.data.alerts.callout_danger}}
+We strongly recommend only using the Automatic CPU Profiler when working directly with the [Cockroach Labs support team](support-resources.html).
+{{site.data.alerts.end}}
+
+## Configuration
+
+You can configure automatic CPU profile capture with the following [cluster settings](cluster-settings.html):
+
+Cluster Setting | Description | Default Value | Recommended Value
+----------------|-------------|---------------|------------------
+`server.cpu_profile.enabled` | Indicates if the Automatic CPU Profiler is on or off. | `false` | 
+`server.cpu_profile.cpu_usage_combined_threshold` | The baseline value for when CPU profiles should be taken. Collect profiles from each node that meets threshold. | MAX integer, such as `9223372036854775807` | `80`
+`server.cpu_profile.interval` | The period of time after which the [high-water mark](#high-water-mark-threshold) resets to the baseline value. | `5m0s` (5 minutes) | `1m40s` (100 seconds)
+`server.cpu_profile.duration` | The length of time a CPU profile is taken. | `10s` (10 seconds) | `1s` or `2s`
+`server.cpu_profile.total_dump_size_limit` | Maximum combined disk size for preserving CPU profiles. | `128 MiB` (128 Mebibytes) |
+
+### Enabling automatic CPU profile capture
+
+To enable automatic CPU profile capture, you must [set](set-cluster-setting.html) both of the following cluster settings:
+
+- Set `server.cpu_profile.enabled` to `true`
+- Set `server.cpu_profile.cpu_usage_combined_threshold` to a value between `0` and `100`. Preferably, use the [recommended value](#recommended-values).
+
+{{site.data.alerts.callout_info}}
+Only setting `server.cpu_profile.enabled` to `true` will not generate a CPU profile.
+{{site.data.alerts.end}}
+
+### Recommended values
+
+- Set `server.cpu_profile.cpu_usage_combined_threshold` to `80` for 80%.
+- Set `server.cpu_profile.duration` to a lower value, for example `1s` or `2s`. This minimizes the impact of [overhead](#overhead) on your cluster compared to the current default value.
+- Set `server.cpu_profile.interval` to a lower value, for example `1m40s` (1 minute 40 seconds).
+
+### High-water mark threshold 
+
+The Automatic CPU Profiler runs asynchronously in the background. After every second, the Automatic CPU Profiler checks if the CPU usage exceeds the high-water mark threshold. If so, it captures a CPU profile. If a profile capture is already in progress, a second profile is not taken.
+
+The Automatic CPU Profiler uses the configuration options to determine the high-water mark threshold. For example, with `duration` set to `2s` , `interval` set to `1m40s`, and `cpu_usage_combined_threshold` set to `80`:
+
+- At `time0` the CPU usage polled is 82 percent which exceeds the baseline threshold of `80`, so a `2s` profile is captured and the high-water mark threshold becomes `82`.
+- After the `2s` profile capture, the Automatic CPU Profiler continues to check every second if the CPU usage now exceeds `82` percent.
+- At `time1` the CPU usage polled is 85 percent, another profile is taken for `2s` and the high-water mark threshold becomes `85`.
+- At `time2`, the `1m40s` interval after `time0`, the high-water mark threshold is reset to the baseline threshold of `80`.
+- The Automatic CPU Profiler continues its every second polling and captures profiles when CPU usage exceeds the high-water mark threshold beginning at the `80` percent baseline.
+
+## Accessing CPU profiles
+
+- The Automatic CPU Profiler saves the captured CPU profiles to disk on each node's file system in the [logging directory](configure-logs.html#logging-directory). The default path is `cockroach-data/logs/pprof_dump`.
+- The only way to get these CPU profiles is to access the node file system.
+- Enabling the Automatic CPU Profiler does **not** add CPU profiles to [debug zips](cockroach-debug-zip.html).
+
+## Overhead
+
+Enabling the automatic CPU profile capture on a cluster will add overhead to the cluster in the form of potential increases in latency and CPU usage.
+
+{{site.data.alerts.callout_info}}
+The decision to enable this feature should be done when advised by the [Cockroach Labs support team](support-resources.html).
+{{site.data.alerts.end}}
+
+- Monitor the following metrics:
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent)
+- We anticipate a sub-10% regression on these foreground latency metrics. This overhead to your cluster may be deemed acceptable in order to collect CPU profiles that are necessary to troubleshoot problems in your cluster. Please consult with Cockroach Labs Support.
+- Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a `server.cpu_profile.duration` (for example, 1 second) increase in latency to capture the CPU profile, consider enabling the Automatic CPU Profiler.
+
+{{site.data.alerts.callout_danger}}
+Do not have the Automatic CPU Profiler enabled by default.
+
+If you enabled the Automatic CPU Profiler and then notice unacceptable overhead to your cluster, we recommend you immediately disable the Automatic CPU Profiler.
+{{site.data.alerts.end}}
--- a/src/current/v23.1/troubleshooting-overview.md
+++ b/src/current/v23.1/troubleshooting-overview.md
@@ -23,3 +23,9 @@ If you experience an issue when using CockroachDB, try these steps to resolve th
 - If you cannot resolve the issue yourself, the following tools can help you move forward:
   - [Support Resources](support-resources.html) identify ways you can get help with troubleshooting.
   - [File an Issue](file-an-issue.html) provides details on how to file an issue that you're unable to resolve.
+ 
+- In a support escalation, you may be directed to use the following features by the [Cockroach Labs support team](support-resources.html):
+
+  - [`cockroach debug zip`](cockroach-debug-zip.html)
+  - [`cockroach debug tsdump`](cockroach-debug-tsdump.html)
+  - [Automatic CPU Profiler](automatic-cpu-profiler.html)


src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Configuration
+
+You can configure automatic CPU profile capture with the following [cluster settings](cluster-settings.html):
+
+Cluster Setting | Description | Default Value | Recommended Value
+----------------|-------------|---------------|------------------
+`server.cpu_profile.enabled` | The ON/OFF switch of the CPU profiler | `false` | 
+`server.cpu_profile.cpu_usage_combined_threshold` | The baseline value for when CPU profiles should be taken. Collect profiles from each node that meets threshold. | MAX integer, such as `9223372036854775807` | `80`
+`server.cpu_profile.interval` | When the high water mark resets and a new CPU profile can be taken | `5m0s` (5 minutes) | `1m40s` (100 seconds)
```
Santamaura, line 21: I think we should elaborate more here since this isn't the whole truth. While the high water mark does reset at this provided interval this may not be the only case when a profile can be taken. 

A profile is taken when the combined CPU percentage exceeds the current high water mark. When this happens, that value becomes the new high water mark. For e.g. if we start out with a default threshold is 80 and our combined CPU usage is 82 we take a profile and the high water mark is now 82. Next time around (in 1s), let's say the value is 85 then we will take a profile again and the high water mark now is 85. So we do take profiles in shorter time spans than the interval it just becomes more unlikely that a profile is taken since the value to exceed becomes higher. The high water mark will reset to 80 once the interval has passed.

florence-crl: review above comment with @Santamaura

* so the cpu profiler (as well as the other automatic profilers) run asynchronously in the background and after every second we call the function which does all the logic to see whether we should take a profile or not
* so basically we check after every second whether we should take a profile or not
* we also don't take profiles if other profiles are already in progress

Another example:
if the `default server.cpu_profile.duration` is `10s` , the `server.cpu_profile.interval`  is `5m0s`, and `server.cpu_profile.cpu_usage_combined_threshold` is `80`:
- If at time0 the CPU exceeds the default threshold of 80 with 82 CPU, a 10s profile is taken and the high watermark becomes 82.
- After the 10s profile capture, the profiler continues to check every second if the CPU exceed 82.
- If at time1 the CPU is 85, another profile is taken for 10s and the high watermark becomes 85.
- At time2 (time0 + 5m0s ), the high watermark is reset to 80.

The diagram (taken from the enablement slide desk) is not exactly correct.

florence-crl: added High-water mark section

Santamaura (approved):
This looks good to me now

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+- At `time0` the CPU usage polled is 82 percent which exceeds the baseline threshold of `80`, so a `10s` profile is captured and the high-water mark threshold becomes `82`.
+- After the `10s` profile capture, the Profiler continues to check every second if the CPU usage now exceeds `82` percent.
+- At `time1` the CPU usage polled is 85 percent, another profile is taken for `10s` and the high-water mark threshold becomes `85`.
+- At `time2`, the `5m0s` interval after `time0`, the high-water mark threshold is reset to the baseline threshold of `80`.
+- The Profiler continues its every second polling and captures profiles when CPU usage exceeds the high-water mark threshold beginning at the `80` percent baseline.
+
+## Accessing CPU profiles
+
+- The CPU profiler saves the captured CPU profiles to disk in the node filesystem. The default path is `/cockroach-data/logs/pprof_dump`.
+- The only way to get the CPU profiles is to access the node filesystem.
```
thtruo, line 55: nit: "these"

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+- Enabling the CPU profiler does **not** add CPU profiles to [debug zips](cockroach-debug-zip.html).
+
+## Overhead
```
thtruo, line 58: I recommend we add a few introductory sentences to provide context behind overhead. We should communicate explicitly that enabling the automatic CPU profile capture on a cluster will add overhead to the cluster in the form of potential increases in latency and CPU usage. The decision to enable this should be done when advised by the Cockroach Labs Support team.

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Monitor the following metrics:
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent) 
+- A less than 10% regression on these foreground latency metrics indicates acceptable overhead.
+- The overhead is a cost to pay to collect the CPU profile needed to troubleshoot. Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a 1 second increase in latency to capture the CPU profile, enable the CPU profiler.
```
thtruo, line 65: Nit: suggest ordering this first, and then proceed with monitoring recommendations and commentary on 10% regression

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+- The only way to get the CPU profiles is to access the node filesystem.
+- Enabling the CPU profiler does **not** add CPU profiles to [debug zips](cockroach-debug-zip.html).
+
+## Overhead
+
+- Monitor the following metrics:
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent) 
+- A less than 10% regression on these foreground latency metrics indicates acceptable overhead.
```
thtruo, line 64: We should rework this bullet. It's not universally "acceptable" in all use cases for all customers. It's "acceptable" from the perspective of our engineering teams as a starting point. 

Suggestion to combine this and the next bullet together and reframe to something like: "We anticipate a sub 10% regression on these foreground latency metrics. This overhead to your cluster may be deemed acceptable in order to collect CPU profiles that are necessary to troubleshoot problems in your cluster. Please consult with Cockroach Labs Support."

src/current/v23.1/cpu-profiler.md
thtruo, line 65: In a previous comment, I suggested combining the previous bullet with this bullet. 

I do like keeping this section, but have additional suggestions:

> Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a 1 second increase in latency to capture the CPU profile, enable the CPU profiler.

^ Let's make the language around "1 second" more generic, since that depends on the value of `server.cpu_profile.duration`

Perhaps along the lines of: "Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a `server.cpu_profile.duration` (e.g. 1 second) increase in latency to capture the CPU profile, `server.cpu_profile.duration` **consider** enabling the automatic CPU profiler."

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+- Set `server.cpu_profile.cpu_usage_combined_threshold` to a value between `0` and `100`.
+
+{{site.data.alerts.callout_info}}
+Only setting `server.cpu_profile.enabled` to `true` will not generate a CPU profile.
+{{site.data.alerts.end}}
+
+### Recommended values
+
+- Set `server.cpu_profile.cpu_usage_combined_threshold` to `80` for 80%.
+- Set `server.cpu_profile.duration` to a lower value, for example `1s` or `2s`. This reduces [overhead](#overhead).
```
thtruo, line 37: Suggestion to tweak the language to say something along the lines of: "This minimizes the impact of overhead on your cluster compared to the current default values"

^ Reasoning is that the current statement "This reduces overhead" makes it sound like enabling this will reduce overhead on your cluster. That's not true. What's more accurate is that these lower values will result in **less** overhead than what would have resulted from the default cluster setting values

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent) 
+- A less than 10% regression on these foreground latency metrics indicates acceptable overhead.
+- The overhead is a cost to pay to collect the CPU profile needed to troubleshoot. Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a 1 second increase in latency to capture the CPU profile, enable the CPU profiler.
+
+{{site.data.alerts.callout_danger}}
+Do not have the CPU profiler enabled by default. Only enable it when advised by Cockroach Labs support.
+
+If the overhead is unacceptable, disable the CPU profiler.
```
thtruo, line 70: Suggestion to this sentence: "If you enabled the automatic CPU profiler and then notice unacceptable overhead to your cluster, we recommend you immediately disable the automatic CPU profiler"

thtruo (dismissed):
Thanks for taking the first stab @florence-crl - overall, it's great to see more informative reference docs describing this new functionality over automatic CPU profile capture. I've left a few suggestions/nits in specific places.

High-level feedback: we should make it more explicit that these CPU profiles are intended for Cockroach Labs Support to utilize to troubleshoot customer clusters whenever a customer escalates a support issue to CRL. These CPU profiles are not friendly or very easily interpretable for end users. I suggest we incorporate similar language in places like https://www.cockroachlabs.com/docs/stable/cockroach-debug-zip.html (e.g. "The other debug subcommands are useful only to Cockroach Labs." or "We recommend using [XYZ] subcommands only when directed by the [Cockroach Labs support team](https://www.cockroachlabs.com/docs/v23.1/support-resources)."). 

I also think we're missing updates to the "Cluster Settings" page https://www.cockroachlabs.com/docs/stable/cluster-settings.html

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
+---
+title: CPU Profiler
```
thtruo, line 2: Suggestion: let's call this what it is - it's the "Automatic CPU Profiler". 

I think calling it "CPU Profiler" makes it sound more generic about CPU profiles as a whole. We should probably have separate docs describing CPU Profiles, but that's outside the scope of this page. That's why I think we should add the `Automatic` prefix in all the places we reference "CPU Profiler". WDYT?

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+summary: Verify that your cluster's data replication, data placement, and zone configurations are working as expected.
```
thtruo, line 3: Nit: this summary does not match what this page is about.

src/current/v23.1/troubleshooting-overview.md
```diff
@@ -20,6 +20,8 @@ If you experience an issue when using CockroachDB, try these steps to resolve th
 
 - [Replication Reports](query-replication-reports.html) let you query the status of your cluster's data replication, data placement, and zone constraint conformance.
 
+- The [CPU Profiler](cpu-profiler.html) lets you investigate CPU load by automatically capturing CPU profiles.
```
thtruo, line 23: Suggestion: Refer to this as the "Automatic CPU Profiler". 

As mentioned in another comment, the CPU profiles are geared for our CRL Support team to use to investigate erratic or spikey CPU load on a cluster. They're not designed with the end user in mind. So I recommend we tweak the language to not say "the automatic cpu profiler lets **you** investigate..." and more along the lines of it is another option to investigate these problems with CRL Support

thtruo: Actually on second thought, I'm not sure it makes sense to include this bullet in this page. Reading through this page, it sounds like actions that the user should take to troubleshoot issues. Where do escalations with CRL Support fit in? Is that outside the scope of this page? If so, let's remove this bullet. If not, then we should be more comprehensive here and bring up debug zips as well

src/current/v23.1/cpu-profiler.md
florence-crl: Done

src/current/v23.1/cpu-profiler.md
florence-crl: Done

src/current/v23.1/cpu-profiler.md
florence-crl: Done

src/current/v23.1/cpu-profiler.md
florence-crl: Changed all references to this feature to "Automatic CPU Profiler"

src/current/v23.1/cpu-profiler.md
florence-crl: Updated Summary.

src/current/v23.1/cpu-profiler.md
```diff
@@ -0,0 +1,71 @@
...
+## Overhead
+
+- Monitor the following metrics:
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent) 
+- A less than 10% regression on these foreground latency metrics indicates acceptable overhead.
+- The overhead is a cost to pay to collect the CPU profile needed to troubleshoot. Overhead only occurs during profile capture, not when it is idle. If the cluster can tolerate a 1 second increase in latency to capture the CPU profile, enable the CPU profiler.
+
+{{site.data.alerts.callout_danger}}
```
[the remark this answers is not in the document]
florence-crl: Added warning to the top before "Configuration" section.

src/current/v23.1/troubleshooting-overview.md
florence-crl: Made bullet point more comprehensive, added `cockroach debug zip` and `cockroach debug tsdump`.

florence-crl:
> I also think we're missing updates to the "Cluster Settings" page https://www.cockroachlabs.com/docs/stable/cluster-settings.html

I created https://github.com/cockroachdb/cockroach/issues/105646 Set public = t for server.cpu_profile cluster settings #105646. This change will add the new settings to the generated cluster settings docs list.

src/current/v23.1/cpu-profiler.md
florence-crl: Modified to say "This minimizes the impact of [overhead](#overhead) on your cluster compared to the current default value."

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+## Overhead
+
+Enabling the automatic CPU profile capture on a cluster will add overhead to the cluster in the form of potential increases in latency and CPU usage. The decision to enable this feature should be done when advised by the [Cockroach Labs support team](support-resources.html).
+
+- Monitor the following metrics:
+  - [P99 latency](ui-sql-dashboard.html#service-latency-sql-99th-percentile)
+  - P50 latency by creating a [custom chart](ui-custom-chart-debug-page.html) for the `sql.exec.latency-p50` metric
+  - [CPU usage](ui-hardware-dashboard.html#cpu-percent)
+- We anticipate a sub-10% regression on these foreground latency metrics. This overhead to your cluster may be deemed acceptable in order to collect CPU profiles that are necessary to troubleshoot problems in your cluster. Please consult with Cockroach Labs Support.
+- Overhead only occurs during profile capture, not when it is idle. The default of `server.cpu_profile.duration` is 1 second. If the cluster can tolerate a `server.cpu_profile.duration` increase in latency to capture the CPU profile, consider enabling the Automatic CPU Profiler.
```
thtruo, line 71: > The default of `server.cpu_profile.duration` is 1 second. 

Nit: The default isn't 1 second. Suggest we cut out this sentence.

thtruo (dismissed):
Nice, this iteration looks great @florence-crl I left a nit - once that's addressed this is good to go

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Sorry about that incorrect rewrite. I have changed it to: "If the cluster can tolerate a server.cpu_profile.duration (for example, 1 second) increase in latency to capture the CPU profile, consider enabling the Automatic CPU Profiler."

thtruo (approved):
LGTM ty!

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+We strongly recommend only using the Automatic CPU Profiler when working directly with the [Cockroach Labs support team](support-resources.html).
+{{site.data.alerts.end}}
+
+## Configuration
+
+You can configure automatic CPU profile capture with the following [cluster settings](cluster-settings.html):
+
+Cluster Setting | Description | Default Value | Recommended Value
+----------------|-------------|---------------|------------------
+`server.cpu_profile.enabled` | The ON/OFF switch of the Automatic CPU Profiler | `false` |
```
lnhsingh, line 23: nit: `The ON/OFF switch` sounds colloquial. I'd suggest something like "Indicates if the Automatic CPU Profiler is on or off."

lnhsingh: Also, all of the descriptions in the table should have a period at the end

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+Cluster Setting | Description | Default Value | Recommended Value
+----------------|-------------|---------------|------------------
+`server.cpu_profile.enabled` | The ON/OFF switch of the Automatic CPU Profiler | `false` | 
+`server.cpu_profile.cpu_usage_combined_threshold` | The baseline value for when CPU profiles should be taken. Collect profiles from each node that meets threshold. | MAX integer, such as `9223372036854775807` | `80`
+`server.cpu_profile.interval` | The period of time after which the high-water mark resets to the baseline value | `5m0s` (5 minutes) | `1m40s` (100 seconds)
+`server.cpu_profile.duration` | The length of time a CPU profile is taken | `10s` (10 seconds) | `1s` or `2s`
+`server.cpu_profile.total_dump_size_limit` | Maximum combined disk size for preserving CPU profiles | `128 MiB` (128 Mebibytes) |
+
+### Enabling automatic CPU profile capture
+
```
lnhsingh, line 30: I think you need an intro the explains you need to use both of the settings

```
To enable automatic CPU profile capture, you must [set](set-cluster-setting.html) both of the following cluster settings:
- `server.cpu_` to `true`
- `server.cpu_` ...

File: src/current/v23.1/automatic-cpu-profiler.md [hunk 15]
```diff
@@ -0,0 +1,77 @@
...
+- Set `server.cpu_profile.enabled` to `true`
+- Set `server.cpu_profile.cpu_usage_combined_threshold` to a value between `0` and `100`.
+
```
lnhsingh, line 33: We should also call out that we have recommended values and link to the next section

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+## Configuration
+
+You can configure automatic CPU profile capture with the following [cluster settings](cluster-settings.html):
+
+Cluster Setting | Description | Default Value | Recommended Value
+----------------|-------------|---------------|------------------
+`server.cpu_profile.enabled` | The ON/OFF switch of the Automatic CPU Profiler | `false` | 
+`server.cpu_profile.cpu_usage_combined_threshold` | The baseline value for when CPU profiles should be taken. Collect profiles from each node that meets threshold. | MAX integer, such as `9223372036854775807` | `80`
+`server.cpu_profile.interval` | The period of time after which the high-water mark resets to the baseline value | `5m0s` (5 minutes) | `1m40s` (100 seconds)
```
lnhsingh, line 25: Should "high-water mark" link to the high-water mark section below?

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+- At `time0` the CPU usage polled is 82 percent which exceeds the baseline threshold of `80`, so a `10s` profile is captured and the high-water mark threshold becomes `82`.
+- After the `10s` profile capture, the Automatic CPU Profiler continues to check every second if the CPU usage now exceeds `82` percent.
+- At `time1` the CPU usage polled is 85 percent, another profile is taken for `10s` and the high-water mark threshold becomes `85`.
+- At `time2`, the `5m0s` interval after `time0`, the high-water mark threshold is reset to the baseline threshold of `80`.
+- The Automatic CPU Profiler continues its every second polling and captures profiles when CPU usage exceeds the high-water mark threshold beginning at the `80` percent baseline.
+
+## Accessing CPU profiles
+
+- The Automatic CPU Profiler saves the captured CPU profiles to disk in the node filesystem. The default path is `/cockroach-data/logs/pprof_dump`.
+- The only way to get these CPU profiles is to access the node filesystem.
```
lnhsingh, line 59: Are there docs you can link to that outline how to access the node filesystem?

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
...
+- Enabling the Automatic CPU Profiler does **not** add CPU profiles to [debug zips](cockroach-debug-zip.html).
+
+## Overhead
+
+Enabling the automatic CPU profile capture on a cluster will add overhead to the cluster in the form of potential increases in latency and CPU usage. The decision to enable this feature should be done when advised by the [Cockroach Labs support team](support-resources.html).
```
lnhsingh, line 64: The sentence about "advised by support" should probably be in a callout.

lnhsingh (approved):
Some comments around style, links, etc.

Once addressed, feel free to merge

src/current/v23.1/automatic-cpu-profiler.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Automatic CPU Profiler
+summary: The Automatic CPU Profiler captures CPU profiles automatically to help investigate cluster issues involving CPU load.
+keywords: cpu profile, cpu profiles, cpu profiler, automatic cpu profiler
+toc: true
+docs_area: manage
+---
+
+{% include_cached new-in.html version="v23.1" %} This feature automatically captures CPU profiles. A CPU profile shows the functions that use the most CPU time, sampled over a window of time.
```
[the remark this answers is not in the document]
florence-crl: Done

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Done

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Done

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Done

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Added link to High-water mark threshold section.

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: Done

src/current/v23.1/automatic-cpu-profiler.md
florence-crl: added link to [logging directory](https://www.cockroachlabs.com/docs/stable/configure-logs.html#logging-directory)

PR #17375, opened 2023-06-29
Remove locale from sites that automatically redirect to the reader's locale

mdlinville:
v23.1 only, not intending to try to backport.

TODO: `advisories/`, `releases/`, subsidiary includes

Related to https://cockroachlabs.atlassian.net/browse/DOC-8230, which requests that we begin to test newly-added links that shouldn't include the static locale.

## Previews
See https://github.com/cockroachdb/docs/pull/17375#issuecomment-1613748697

## Tests
- [x] Linkcheck
- [x] Local build


nickvigilante (changes requested):
Even though we're not backporting, we should still do this for all v23.1 includes, as well as `advisories`, `cockroachcloud`, `releases`, and all of those includes. Otherwise, what I've seen so far looks excellent. Thanks for taking this on!

PR #17378, opened 2023-06-30
Docs Fixit: Add note to Sink docs on pub/sub emulator

kathancox:
Fixes DOC-8126

Added a note to the Changefeed Sinks page that the Pub/Sub emulator can be used with the env vars Google provides.


kathancox:
@shermanCRL Is this sufficient as a note for the Pub/Sub emulator? Also, do you think we need to mention that we do not do testing on this as well, or maybe that would be a bit overboard?

shermanCRL (approved):
Looks great, thanks.

ashleykwtam (approved):
LGTM :)

PR #17380, opened 2023-06-30
Docs fixit: Better direct users to sink page from SQL page

kathancox:
Fixes DOC-8250

Very minor edit to bring the direction to the main Changefeed Sinks page up to the top of the sink URI section on the changefeed SQL page. Users can get more information on the Sinks page and so we want them to know about this if they land on the SQL page first.


ashleykwtam (approved):
LGTM!

PR #17382, opened 2023-06-30
Update name of Raft tick interval to match YAML

rmloveland:
Fixes a bug on the 'Architecture > Replication Layer' page where the value of the Raft tick interval was failing to render in the chart of important constants because the YAML variable in the docs page didn't match the YAML variable in the `_data/constants.yml` file.


taroface (approved):
LGTM!

PR #17407, opened 2023-07-06
Backups page now called Backup and Restore

gemma-shay:
Discovered this change via the release notes and opened a ticket https://cockroachlabs.atlassian.net/browse/DOC-8275


ashleykwtam (approved):
LGTM, thanks for catching this Gemma!

PR #17418, opened 2023-07-07
external storage: clarify when `self` is usable

dt:


src/current/v23.1/use-cloud-storage.md
```diff
@@ -53,7 +53,7 @@ You can disable the use of implicit credentials when accessing external cloud st
...
+<sup>2</sup>   Using a `nodeID` is required and the data files will be in the `extern` directory of the specified node. In most cases (including single-node clusters), using `nodelocal://1/<path>` is sufficient. If every node has the [`--external-io-dir`](cockroach-start.html#flags-external-io-dir) flag pointed to a common NFS mount or other form of network-backed, shared or synchronized storage, the word `self` can be used instead of a node ID, to that indicate each node should write individual data files to its own `extern` directory however using this when the external IO directory is not shared will lead to files being scattered across arbitrary individual nodes.
```
msbutler, line 56: nit: remove "that" before "indicate"

msbutler (approved):

src/current/v23.1/use-cloud-storage.md
kathancox, line 56:
```suggestion
<sup>2</sup>   Using a `nodeID` is required and the data files will be in the `extern` directory of the specified node. In most cases (including single-node clusters), using `nodelocal://1/<path>` is sufficient. If every node has the [`--external-io-dir`](cockroach-start.html#flags-external-io-dir) flag pointed to a common NFS mount, or other form of network-backed, shared, or synchronized storage, you can use the word `self` instead of a node ID to indicate that each node should write individual data files to its own `extern` directory. However, using `self` when the external IO directory is not shared will lead to files being scattered across arbitrary individual nodes.
```

kathancox (approved):
Made a few suggestions on nits, otherwise LGTM

src/current/v23.1/use-cloud-storage.md
dt: I had actually dropped the last sentence entirely before opening this, but I just realized I pushed changed only to my fork, but opened this from the branch on the main repo by accident. Updated now.

PR #17421, opened 2023-07-07
Update canonicals and redirects to point to naked URLs

nickvigilante:
Fixes DOC-8289


mdlinville (approved):
Noting that this is still Draft

PR #17422, opened 2023-07-07
Technical Advisory 99561

kathancox:


kathancox:
@rimadeodhar Here is the draft of the TA, I have edited some language in the first section, which removes the reference to the code line directly since we have the issue and PR links throughout. Let me know if you're OK with this.

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+
+This is resolved in CockroachDB by [#105828](https://github.com/cockroachdb/cockroach/pull/105828), which makes a post-deserialization change to alter the encoding version for secondary indexes that are marked as primary index encoded and there are no mutations. This PR will also add validation to detect tables with this same problem so that they can be fixed.
+
+The fix will be applied to v22.2.12 and v23.1.6.
+
+This public issue is tracked by [#99561](https://github.com/cockroachdb/cockroach/issues/99561).
+
+## Mitigation
+
+Users of CockroachDB v22.2.x are encouraged to upgrade to [v22.2.12](../releases/v22.2.html) upon its release. Similarly, users of v23.1.0 — v23.1.5 are encouraged to update to [v23.1.6](../releases/v23.1.html) upon its release.
```
rimadeodhar, line 29: Nit:
```suggestion
Users of CockroachDB v22.2.x are encouraged to upgrade to [v22.2.12](../releases/v22.2.html) upon its release. Similarly, users of v23.1.0 — v23.1.5 are encouraged to upgrade to [v23.1.6](../releases/v23.1.html) upon its release.
```

rimadeodhar (approved):

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Technical Advisory 99561
+advisory: A-99561
+summary: When adding a new column using the new declarative schema changer, CockroachDB retrieves a unique secondary index mistakenly, assuming it to be a primary index.
+toc: true
+affected_versions: v22.2.x, v23.1.0-v23.1.5
+advisory_date: 2023-07-07
```
kathancox, line 7: Final edit: Date to be updated to pub date

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+docs_area: releases
+---
+
+Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
+
+## Description
+
+In CockroachDB v20.2 and v21.1 (the first two versions to support [`ALTER PRIMARY KEY`](../v21.1/alter-primary-key.html)), there was a bug where in some cases the index automatically created to provide uniqueness for the existing primary key would have its `EncodingType` field marked as `PrimaryIndexEncoding`. This behavior was fixed in [#66599](https://github.com/cockroachdb/cockroach/pull/66599), but left secondary indexes with primary encoding.
```
mdlinville, line 15: I think the first sentence is doing too much work. Perhaps the first thing is to explain that an index is automatically created to ensure that the primary key is unique, then summarize the bug and explain that it is fixed. Then the reader is set up to understand the rest of the TA. WDYT?

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+
+In CockroachDB v22.2 and later, the implementation for [`ALTER TABLE`](../v22.2/alter-table.html#add-column) began using the new declarative schema changer framework. When adding a new column using the new [declarative schema changer](../v22.2/online-schema-changes.html#declarative-schema-changer), CockroachDB retrieves a unique secondary index mistakenly, assuming it to be a primary index. As a result, the secondary index is modified to store the new columns. CockroachDB expects there to be only one primary index, but in the case outlined here, there are two, because a secondary index has been incorrectly labeled as a primary one. This will cause the new column data to be present only in the secondary index and could lead to irrecoverable data loss, if the secondary index is deleted by the user.
```
mdlinville, line 17: Maybe move the "new declarative schema changer" link earlier to the first mention. I'm also not sure if it's correct to link to the v22.x docs, even though that is when the new one was introduced.

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Statement
+
+This is resolved in CockroachDB by [#105828](https://github.com/cockroachdb/cockroach/pull/105828), which makes a post-deserialization change to alter the encoding version for secondary indexes that are marked as primary index encoded and there are no mutations. This PR will also add validation to detect tables with this same problem so that they can be fixed.
```
mdlinville, line 21: I get lost in the description of the fix. I think "primary index encoded" might be easier to understand as "encoded as primary indexes", and I'm confused about "and there are no mutations" -- does it mean "where there are no mutations? Secondary indexes that are encoded as primary indexes where there are no mutations? (No mutations on what?)

kathancox: @rimadeodhar Can you clarify the mutations part here? I assumed it meant mutations hadn't occurred on the secondary indexes, but I realize this is unclear.

src/current/advisories/a99561.md
kathancox: I've moved the link here, and changed the version

src/current/advisories/a99561.md
kathancox: I have added an initial sentence here, but I'm not sure if it is what you were envisioning. Feel free to adapt.

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+affected_versions: v22.2.x, v23.1.0-v23.1.5
+advisory_date: 2023-07-10
+docs_area: releases
+---
+
+Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
+
+## Description
+
+A primary [index](../v23.1/indexes.html) is automatically created for each table, which indexes the table's [primary key](../v23.1/primary-key.html), or another unique value when there is no primary key. In CockroachDB v20.2 and v21.1 (the first two versions to support [`ALTER PRIMARY KEY`](../v21.1/alter-primary-key.html)), there was a bug where in some cases the index, which is automatically created to provide uniqueness for the existing primary key, would have its `EncodingType` field marked as `PrimaryIndexEncoding`. This behavior was fixed in [#66599](https://github.com/cockroachdb/cockroach/pull/66599), but left [secondary indexes](../v23.1/schema-design-indexes.html) with primary encoding.
```
mdlinville, line 15:
```suggestion
A primary [index](../v23.1/indexes.html) is automatically created for each table, which indexes the table's [primary key](../v23.1/primary-key.html), or another unique value when there is no primary key. In CockroachDB v20.2 and v21.1 (the first two versions to support [`ALTER PRIMARY KEY`](../v21.1/alter-primary-key.html)), there was a bug where in some cases the automatically-created index's `EncodingType` field marked as `PrimaryIndexEncoding`. This behavior was fixed in [#66599](https://github.com/cockroachdb/cockroach/pull/66599), but left [secondary indexes](../v23.1/schema-design-indexes.html) with primary encoding.
```
After adding the first sentence, does this suggestion make sense? I'm confused now because it seems like the bug is not in the primary index but in secondary indexes that are claiming to be primary indexes.

src/current/advisories/a99561.md
kathancox: @mdlinville I have updated this language per Rafi for this sentence.

src/current/advisories/a99561.md
```diff
@@ -12,13 +12,13 @@ Publication date: {{ page.advisory_date | date: "%B %e, %Y" }}
...
+A primary [index](../v23.1/indexes.html) is automatically created for each table, which indexes the table's [primary key](../v23.1/primary-key.html), or another unique value when there is no primary key. In CockroachDB v20.2 and v21.1 (the first two versions to support [`ALTER PRIMARY KEY`](../v21.1/alter-primary-key.html)), there was a bug where in some cases the automatically created index would have its `EncodingType` field marked as `PrimaryIndexEncoding`. This behavior was fixed in [#66599](https://github.com/cockroachdb/cockroach/pull/66599), but left [secondary indexes](../v23.1/schema-design-indexes.html) with primary encoding.
```
mdlinville, line 15: I'm still confused. We start by talking about primary indexes, and we finish with talking about secondary indexes that look like primary indexes. I think we need to explain more whether the bug is with primary indexes or with secondary indexes.

kathancox: @mdlinville I _think_ I see the problem....

Automatically created indexes can refer to the secondary index as well. That is, columns where there is a `UNIQUE` constraint, and therefore CRDB auto creates a secondary index. I think the first sentence/second sentence should also include this information re secondary indexes. 

Perhaps @rimadeodhar when you are making further amendments you could confirm this?

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+
+To disable the declarative schema changer for all users, use the [cluster setting](../v23.1/cluster-settings.html):
+
+{% include_cached copy-clipboard.html %}
+~~~sql
+SET CLUSTER SETTINGS sql.defaults.use_declarative_schema_changer = off
+~~~
+
+Additionally, run the corruption detection query outlined in [this gist](https://gist.github.com/rimadeodhar/752bdf89d69c9d7fea3be5ed9acce3b6) to determine if there is any such secondary index on the cluster with primary key encoding. If such corruption exists, disable the declarative schema changer as outlined previously prior to running any schema change and contact our [support team](https://support.cockroachlabs.com/) to fix the cluster as soon as possible.
+
```
rimadeodhar, line 46:
```suggestion
Until such an upgrade, please run the query outlined in [this gist](https://gist.github.com/fqazi/2b71b05ce8f6b028267ed3b4498c4f6a) to check if your cluster is in a potentially corrupt state. If the query returns any data, please reach out to support for help to resolve the inconsistency. Until support has had a chance to fix your cluster, please refrain from running any schema changes on the impacted table including DROP INDEX or ALTER TABLE as it could lead to data loss.

```

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,51 @@
...
+In CockroachDB v22.2 and later, the implementation for [`ALTER TABLE`](../v22.2/alter-table.html#add-column) began using the new [declarative schema changer](../v23.1/online-schema-changes.html#declarative-schema-changer) framework. When adding a new column using the new declarative schema changer, CockroachDB retrieves a unique secondary index mistakenly, assuming it to be a primary index. As a result, the secondary index is modified to store the new columns. CockroachDB expects there to be only one primary index, but in the case outlined here, there are two, because a secondary index has been incorrectly labeled as a primary one. This will cause the new column data to be present only in the secondary index and could lead to irrecoverable data loss, if the secondary index is deleted by the user.
+
+## Statement
+
+This is resolved in CockroachDB by [#105828](https://github.com/cockroachdb/cockroach/pull/105828), which makes CockroachDB automatically repair the encoding type for secondary indexes when accessing the table. This PR will also add validation to detect tables with this same problem so that they can be fixed.
```
rimadeodhar, line 21: The last line can be removed as the validation is not being backported in this PR.

```suggestion
This is resolved in CockroachDB by [#105828](https://github.com/cockroachdb/cockroach/pull/105828), which makes CockroachDB automatically repair the encoding type for secondary indexes when accessing the table.
```

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,37 @@
...
+
+In CockroachDB v22.2 and later, the implementation for [`ALTER TABLE`](../v22.2/alter-table.html#add-column) began using the new [declarative schema changer](../v23.1/online-schema-changes.html#declarative-schema-changer) framework. When adding a new column using the new declarative schema changer, CockroachDB retrieves a unique secondary index mistakenly, assuming it to be a primary index. As a result, the secondary index is modified to store the new columns. CockroachDB expects there to be only one primary index, but in the case outlined here, there are two, because a secondary index has been incorrectly labeled as a primary one. This will cause the new column data to be present only in the secondary index and could lead to irrecoverable data loss, if the secondary index is deleted by the user.
```
rimadeodhar, line 17:
```suggestion
In CockroachDB v22.2 and later, the implementation for [`ALTER TABLE`](../v22.2/alter-table.html#add-column) began using the new [declarative schema changer](../v23.1/online-schema-changes.html#declarative-schema-changer) framework. When adding a new column using the new declarative schema changer, CockroachDB retrieves a unique secondary index mistakenly, assuming it to be a primary index. As a result, the secondary index is modified to store the new columns. CockroachDB expects there to be only one primary index, but in the case outlined here, there are two, because a secondary index has been incorrectly labeled as a primary one. This will cause the new column data to be present only in the secondary index and could lead to irrecoverable data loss, if the secondary index is deleted by the user.  Additionally, it could also lead to inconsistent query results as the primary index does not contain data corresponding to the new column.
```

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Mitigation
+
+Users of CockroachDB v22.2.x are encouraged to upgrade to [v22.2.12](../releases/v22.2.html) upon its release. Similarly, users of v23.1.0 — v23.1.5 are encouraged to update to [v23.1.6](../releases/v23.1.html) upon its release.
+
+Until such an upgrade, run the query outlined in [this gist](https://gist.github.com/fqazi/2b71b05ce8f6b028267ed3b4498c4f6a) to check if your cluster is in a potentially corrupt state. If the query returns any data, contact our [support team](https://support.cockroachlabs.com/) to resolve the inconsistency. Until our support team has had a chance to fix the cluster, refrain from running any schema changes on the impacted table including [`DROP INDEX`](../v23.1/drop-index.html) or [`ALTER TABLE`](../v23.1/alter-table.html) as it could lead to data loss.
+
+## Impact
+
+Upon upgrade to CockroachDB v22.2.0 and above, a bug from v20.2 and v21.1 could cause the secondary index to be modified upon `ALTER TABLE..ADD COLUMN` instead of the existing primary index. If the user drops the secondary index, all data related to the new column will be deleted resulting in irrecoverable data loss.
```
rimadeodhar, line 35:
```suggestion
Upon upgrade to CockroachDB v22.2.0 and above, a bug from v20.2 and v21.1 could cause the secondary index to be modified upon `ALTER TABLE..ADD COLUMN` instead of the existing primary index. If the user drops the secondary index, all data related to the new column will be deleted resulting in irrecoverable data loss. Additionally, until this cluster is repaired, there is a risk of inconsistent query results as the primary index does not contain data for the newly added columns.
```

mdlinville (approved):

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,45 @@
...
+
+In CockroachDB v22.2.0 and above, the implementation for [`ALTER TABLE`](../v22.2/alter-table.html#add-column) began using the new [declarative schema changer](../v23.1/online-schema-changes.html#declarative-schema-changer) framework. When adding a new column using the new declarative schema changer, a bug caused the incorrect retrieval of the unique secondary index rather than the primary index. As a result, the secondary index is modified to store the new columns rather than the primary index. That is, `scpb.FindPrimaryIndex` expects there to be only one `scpb.FindPrimaryIndex`, but in this case there are two, because the secondary index is incorrectly labeled as a primary one. This causes the new column data to only be present in the secondary index, which could lead to irrecoverable data loss, if the secondary index is deleted by the user. Furthermore, it could lead to inconsistent query results as the primary index does not contain data corresponding to the new column.
```
kathancox, line 17: @rimadeodhar Is this sentence necessary? Can we remove?

> That is, `scpb.FindPrimaryIndex` expects there to be only one `scpb.FindPrimaryIndex`, but in this case there are two, because the secondary index is incorrectly labeled as a primary one.

I would say we don't typically refer to lines of CRDB code in docs. It seems like the meaning is there without this. But, let me know if there is a specific reason we should keep this?

rimadeodhar: Yeah, I agree. I think we can remove it.

src/current/advisories/a99561.md
```diff
@@ -0,0 +1,45 @@
...
+
+This is resolved in CockroachDB by [#105828](https://github.com/cockroachdb/cockroach/pull/105828), which makes CockroachDB automatically repair the encoding type for secondary indexes when accessing the table. In summary, this encoding type fix will prevent this issue from occurring, and additional validation will be added to precent data loss.
+
+The fix will be applied to v22.2.12 and v23.1.6.
+
+This public issue is tracked by [#99561](https://github.com/cockroachdb/cockroach/issues/99561).
+
+## Mitigation
+
+Users of CockroachDB who have had v20.2 or v21.1 deployed on their cluster should run the query outlined in [this gist](https://gist.github.com/fqazi/2b71b05ce8f6b028267ed3b4498c4f6a) to check if your cluster is in a potentially corrupt state. If the query returns any data, contact our [support team](https://support.cockroachlabs.com/) to resolve the inconsistency. Until our support team has had a chance to fix the cluster, refrain from running any schema changes on the impacted table including [`DROP INDEX`](../v23.1/drop-index.html) or [`ALTER TABLE`](../v23.1/alter-table.html) as it could lead to data loss.
```
rimadeodhar, line 35: Can we replace this link with the actual query as a code snippet within the TA?

PR #17435, opened 2023-07-10
DOC-8230: Add a test for localized external links

nickvigilante:
Fixes DOC-8230


vale/CockroachDB/LocalizedLinks.yml
```diff
@@ -0,0 +1,8 @@
+extends: existence
+message: "Link contains a localization element (e.g., `en`, `en-us`). Try to remove it from the URL: '%s'"
+link: https://github.com/cockroachdb/docs/blob/main/StyleGuide.md#links
+scope: raw
+level: warning
+nonword: true
+tokens:
+  - 'https:\/\/.*?(\ben\b).*?(?="|\| ))'
```
mdlinville, line 8: I think this is too broad -- it would trigger on `/enticing/` for example. I think we need to make the test more specific. The main variants I have seen are `/en/`, `/en-us/`, `/en_us/`, and the same strings as `GET` parameters. WDYT?

nickvigilante: Fixed.

mdlinville (approved):

PR #17445, opened 2023-07-11
Maintenance window clarifications

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-7871
https://cockroachlabs.atlassian.net/browse/DOC-8304


iibuaka (approved):

src/current/cockroachcloud/cluster-management.md
```diff
@@ -165,9 +165,9 @@ When you remove a region from a [multi-region](plan-your-cluster.html#multi-regi
   
 ## Set an upgrade window
 
-From your cluster's [**Overview** page](cluster-overview-page.html), you can view and manage the patch upgrade window for your cluster. To help keep your clusters updated while minimizing disruption and downtime, set a window of time when your cluster is experiencing the lowest traffic.
+From your cluster's [**Overview** page](cluster-overview-page.html), you can view and manage the [patch upgrade](upgrade-policy.html#patch-version-upgrades) window for your cluster. During the window, your cluster may experience restarts, degraded performance, and downtime for single node clusters. To help keep your clusters updated while minimizing disruptions, set a window of time when your cluster is experiencing the lowest traffic. Note that upgrades may not always be completed by the end of the window, and other kinds of cluster maintenance may occur outside the window.
```
mdlinville, line 168:
```suggestion
From your cluster's [**Overview** page](cluster-overview-page.html), you can view and manage the [patch upgrade](upgrade-policy.html#patch-version-upgrades) window for your cluster. During the window, your cluster may experience restarts, degraded performance, and downtime for single-node clusters. To help keep your clusters updated while minimizing disruptions, set a window of time when your cluster is experiencing the lowest traffic. Note that upgrades may not always be completed by the end of the window, and other kinds of cluster maintenance may occur outside the window.
```
I think you'd also get downtime if you had a two-node cluster.  I wonder if it's worth just saying "reduced availability during the maintenance window"?

src/current/cockroachcloud/upgrade-policy.md
```diff
@@ -17,7 +17,7 @@ This page describes the upgrade policy for {{ site.data.products.db }}. For self
...
+For {{ site.data.products.dedicated }} clusters, [Organization Admins](authorization.html#org-administrator-legacy) can [set a weekly 6-hour upgrade window](cluster-management.html#set-an-upgrade-window) during which available patch upgrades will be applied. During the window, your cluster may experience restarts, degraded performance, and downtime for single node clusters. Note that upgrades may not always be completed by the end of the window, and other kinds of cluster maintenance can occur outside the window. Patch upgrades can also be [deferred for 60 days](cluster-management.html#set-an-upgrade-window). If no upgrade window is configured, {{ site.data.products.dedicated }} clusters will be automatically upgraded to the latest supported patch version as soon as it becomes available.
```
mdlinville, line 20: If you make. changes above, it is worth keeping this in sync.

mdlinville (approved):

src/current/cockroachcloud/cluster-management.md
gemma-shay: i think this would be true, but we don't allow 2-node clusters on dedicated

mdlinville (approved):

PR #17454, opened 2023-07-12
[CRDB-29671] Clarify the experimental nature of Auto-TLS

knz:
See https://cockroachlabs.atlassian.net/browse/CRDB-29671

We need to clarify the experimental nature of the functionality. "Alpha" suggested we're ready to put it in front of customers and it's not yet the case.


mikeCRL (changes requested):
I'd convert all `EXPERIMENTAL` instances in titles/headings to title case `Experimental` and in body text to sentence case `experimental`.

mikeCRL:
FYI @MichaelTrestman and @abhinavg6 on this change regarding a feature I documented a while back before you joined CRL.

PR #17457, opened 2023-07-12
Serverless does not support cross-cluster restores

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-8324
You can't do cross-cluster restores through the CC Console, but you can use self-managed backups.


gemma-shay:
@marksoper how does this language look?

marksoper (approved):

PR #17458, opened 2023-07-12
SCT filtering + find/replace

taroface:
DOC-8325

- Document SCT filtering.
- Add note on Replace All functionality.
- Slightly simplify layout of SCT doc.


src/current/cockroachcloud/migrations-page.md
```diff
@@ -253,22 +245,43 @@ To edit a statement, click the **Edit** button or the statement itself and enter
...
+- [Suggestions](#suggestions)
+- [Compatibility Notes](#compatibility-notes) (non-PostgreSQL schemas)
+
+{{site.data.alerts.callout_info}}
+If a filter is not available, `No Options` is displayed.
+{{site.data.alerts.end}}
+
+Click **Apply** to filter the **Statements** list using the selected filters.
+
+To remove a filter, click the `x` next to the active filter. To remove all active filters, click **Clear filters** above the **Statements** list.
```
ryanluu12345, line 264: "To remove an active filter, click the `x` next to the active filter name."

ryanluu12345:
It generally looks good to me. However I don't see the removal of the docs for "Collapse statements". We no longer support collapsing statements now that we have filtering.

Also, I'm not sure I follow, but why did we "Add note on Replace All functionality". I don't think anything changed in that regard.

taroface:
> It generally looks good to me. However I don't see the removal of the docs for "Collapse statements". We no longer support collapsing statements now that we have filtering.
> 
> Also, I'm not sure I follow, but why did we "Add note on Replace All functionality". I don't think anything changed in that regard.

TFTR! There was a mention of "Collapse statements" that was removed in this doc. And as discussed separately, we hadn't documented Replace All, so that was added here.

Amruta-Ranade (approved):
LGTM!

ryanluu12345:
LGTM now! Thanks for addressing :)

PR #17487, opened 2023-07-17
Improving performance of Jekyll build

nickvigilante:
Caching the sidenav, because the only thing it uses to load itself is which version's JSON file to load.

Fixes DOC-8363


MichaelTrestman (approved):
LGTM

PR #17510, opened 2023-07-20
Noindex docs pages based on the following criteria:

nickvigilante:
Don't index the page if both of the following criteria are true:

- The page canonical doesn't match the visited URL
- The version of the docs is expired


JosephNickdow (approved):
Looks good to me @nickvigilante I spot checked a few as well and it seems to be working as expected.

PR #17518, opened 2023-07-21
Add Serverless and MR info to connection pool doc

ianjevans:
This PR updates the connection pool docs with information on Serverless and multi-region clusters. It also includes a new section on monitoring new SQL connections to ensure the connection pool has been configured correctly.

- [x] Content review.
- [x] Backport to v22.2, v22.1.

Fixes DOC-7401.
Fixes DOC-2111.
Fixes DOC-6586.
Fixes DOC-7643.
Fixes DOC-7895.
Fixes DOC-8652.

This PR does not address how to use `pgbouncer` or size connection pools when applications are deployed and autoscaled using Kubernetes. It's difficult to give specific guidance on using either `pgbouncer` or Kubernetes as different workloads and deployments will have different requirements.


theodore-hyman:
@ianjevans Thank you for the efforts here, I saw that I was requested as reviewer so I have dropped some thoughts. I defer to docs team and engineering or whoever you fine folks work with to validate technical knowledge. So feel free to completely ignore me if you want. But providing below my 2 cents for what it's worth.  

ramblings/thoughts:

I understand the hesitancy to try to address kubernetes and autoscaling apps (which actually do seem to be the norm for our customer base from my experience so this is unfortunate) - one way to address customers who are have spiky workloads that scale up and down dramatically, would be to use an out of process (read: not in the application container itself) connection pooler, being pgbouncer. Pgbouncer is run on an external system (not in CRDB and not in the app container but on its own VM or container) which allows the sizing of the pool to remain more constant. I still think this is a gap that we don't even make a brief mention of this and link to pgbouncer resources elsewhere. Anyways.

actual feedback: 

I do have to say I see more technical details in the main Dedicated and Self Hosted options around why we suggest it and how connection pooling works and this is very welcome to see, much improved. 

```
Similar to single-region CockroachDB Serverless clusters, the purpose of connection pools in multi-region CockroachDB Serverless clusters is to reduce connection latency. Add a connection pool for each region, and size the idle pool for your application and workload.

The cluster will automatically scale up to meet demand, so there's no need to set a maximum connection pool size.
```

connection pools are not setup on CRDB, they are setup on the application side (unless its pgbouncer), and its pretty common for there to be multiple application containers running per region, so this guidance doesn't make sense to me. "add a connection pool per region" -- functionally, what I think we're trying to say is, make sure that you have connection pooling configured in your application on all regional deployments of your application, to be consistently following our best practices. IF you were talking about pgbouncer (external to app connection pooling) then the statement "add a connection pool per region" would make sense. 

"size the idle pool for your application and workload." unsure how you would size an idle pool, I guess that's basically saying "hey customer you can set your idle/minimum pool size to whatever you want" which I guess is fine, but I imagine that could be a little confusing, regarding the use of the word size here, I'm not sure how anyone would actually size the idle pool, but perhaps this is an area that I am just not that knowledgeable in. 

Thank you

JeffSwenson:
> I understand the hesitancy to try to address kubernetes and autoscaling apps (which actually do seem to be the norm for our customer base from my experience so this is unfortunate) - one way to address customers who are have spiky workloads that scale up and down dramatically, would be to use an out of process (read: not in the application container itself) connection pooler, being pgbouncer. Pgbouncer is run on an external system (not in CRDB and not in the app container but on its own VM or container) which allows the sizing of the pool to remain more constant. I still think this is a gap that we don't even make a brief mention of this and link to pgbouncer resources elsewhere. Anyways.

IMO we should rarely recommend pgbouncer with CRDB. For example: I don't think we would ever offer pgbouncer as a service in CC. There are applications for which pgbouncer provides marginal benefits, but it doesn't outweigh the operation complexity. Connections in CRDB are relatively cheap. It's true that DB measured latency increases when the DB is saturated, but connection pooling mostly pushes the latency from waiting for a response to the DB to waiting for a connection from the pool.

There is a fundamental relationship between the scaling of an application and the scaling of the DB. Each application worker should have enough connections to the DB that under normal operations, the worker can saturate it's own CPU, memory, or network resources. The number of workers they scale out should have a limit that is around the saturation point of the database, because it doesn't make sense to add more workers only to bottlneck on the DB.

theodore-hyman:
@JeffSwenson 

> There are applications for which pgbouncer provides marginal benefits, but it doesn't outweigh the operation complexity

In an environment where there the load is always constant from the app to the DB, I would agree with you.

However, it is very common to get this question in the context of autoscaling kubernetes pods on the application side, where they autoscale up to increase resources if there is a spike in the application traffic.

I don't believe that the answer of "just look at your average traffic and size your connection pools accordingly" properly addresses this. Without using an external load balancer like pgbouncer the complexity of solving for this type of requirement is much much higher.

> There are applications for which pgbouncer provides marginal benefits, but it doesn't outweigh the operation complexity

Disagreed. For spiky workloads, it is not uncommon for customers to let the avg load only be 10-20 or 30% on the DB side to leave overhead for the spikes.

> The number of workers they scale out should have a limit that is around the saturation point of the database, because it doesn't make sense to add more workers only to bottlneck on the DB.

The entire purpose of using an external load balancing is to be able to control this with more granularity - if you set the connection pool in the autoscaling containers, then the amount of workers and their respective connection pool total size can balloon out of control. When you have an autoscaling group of works that is constrained by pgbouncer to a set "maximum" pool size, this allows you to use that extra headroom that I mentioned above in a more controlled way.

> connection pooling mostly pushes the latency from waiting for a response to the DB to waiting for a connection from the pool.

To your above point here, I acknowledge that part of the issue here may be more educational in nature than technical. I acknowledge that a big part of what is motivating me to try to push for more improvements here is our lack of a clear answer to the question "how do I setup my connection pool sizing on autoscaling kubernetes container groups, and how do I follow your guidance in that context". Perhaps if we had a better way to help our customers think about this topic, or at least understand the impact of spiky workloads in the context of connection pooling, then we would all feel more comfortable using that approach. Right now from my POV I see a lot of differing opinions here and I am not convinced either way. At least the pgbouncer approach makes sense in a spiky environment regarding how to size your pools. If the guidance is simply "don't worry about it, just let your autoscaling app side workers create massive connection pools when they scale up" -- then we ideally need some kind of way to help validate that this is indeed not an issue, some technical explanation, something, because from my anecdotal experience we do have a number of customers running with pgbouncer, they appear to get value from it, and we have CS team members recommending this.

JeffSwenson:
> The entire purpose of using an external load balancing is to be able to control this with more granularity - if you set the connection pool in the autoscaling containers, then the amount of workers and their respective connection pool total size can balloon out of control. When you have an autoscaling group of works that is constrained by pgbouncer to a set "maximum" pool size, this allows you to use that extra headroom that I mentioned above in a more controlled way.

My core point here to say here is it doesn't make sense to scale from 5 workers to 10 workers if the database (or the pgbouncer connection pool) is exhausted at 5 workers.

> because from my anecdotal experience we do have a number of customers running with pgbouncer, they appear to get value from it, and we have CS team members recommending this.

We can chat about this somewhere else, but if there are any use cases where pgbouncer is valuable with CRDB, those should be converted into feature requests for the database. Unlike postgres, CRDB connections are not fundamentally expensive, having high number of connections to the database is fine.

theodore-hyman:
@JeffSwenson 

> I think the key observation for how we should give guidance is the 4 active connections per vcpu is a soft limit, not a hard limit. If a user exceeds 4 active connections, throughput drops slightly, but it doesn't immediately fall off a cliff.

agreed

> If it were my autoscaling application I would use the following formula for picking the max_workers:

This is where the challenge arises. I have had this conversation myself and observed many CS colleagues having the same conversation with customers. We all arrive at this conclusion. We come up with a formula, something like what you have written. Then the customer asks "ok, how do we actually implement that? There is no way to implement that in my kubernetes pods" then we all come to the same conclusion - the only way that we know to keep the number of total connections consistent is to use an external connection pooler - pgbouncer.

JeffSwenson:
> @JeffSwenson
> 
> > I think the key observation for how we should give guidance is the 4 active connections per vcpu is a soft limit, not a hard limit. If a user exceeds 4 active connections, throughput drops slightly, but it doesn't immediately fall off a cliff.
> 
> agreed
> 
> > If it were my autoscaling application I would use the following formula for picking the max_workers:
> 
> Then the customer asks "ok, how do we actually implement that? There is no way to implement that in my kubernetes pods" 

Every platform I'm aware of allows for autoscalers to specify the maximum number of pods or containers. In the case of k8s it is the "maxReplicas" field on the autoscaler spec. 
image

Now, I agree that there is no way to do this perfectly if they have multiple deployments sharing a cluster, since each deployment autoscales independently. But in that case I would probably just use the same formula for every deployment and accept the database can be overloaded. 

Using pbbouncer doesn't prevent them from exhausting their access to the database. If they scale out their application too wide they will still see increased latency. The latency will occur when trying to check out connections from pgbouncer instead of waiting on a response from CRDB.

theodore-hyman:
@ianjevans thank you for following up on this. I have shared my stance and thoughts and I agree it seems we are lacking consensus at the moment. 

@sean- @a-entin can one of you please weigh in on why it is important that we at least have *some* kind of reference to pgbouncer in our documentation, even if it is as simple as a link to pgbouncer and saying that it is an option for connection pooling. This is a follow up to a conversation that we had internally nearly a year ago, ref: https://cockroachlabs.slack.com/archives/CHVV403F0/p1664548004412699

dikshant (approved):
LGTM! I think we can merge this for now. And we will add guidance around pgbouncer and autoscaling apps separately.

src/current/v23.1/connection-pooling.md
```diff
@@ -66,7 +114,7 @@ Validating connections is typically handled automatically by the connection pool
 
 <section class="filter-content" markdown="1" data-scope="java">
 
-In this example, a Java application similar to the [basic JDBC example](build-a-java-app-with-cockroachdb.html) uses the [PostgreSQL JDBC driver](https://jdbc.postgresql.org/) and [HikariCP](https://github.com/brettwooldridge/HikariCP) as the connection pool layer to connect to a CockroachDB cluster. The database is being run on 10 cores across the cluster.
+In this example, a Java application similar to the [basic JDBC example](build-a-java-app-with-cockroachdb.html) uses the [PostgreSQL JDBC driver](https://jdbc.postgresql.org/) and [HikariCP](https://github.com/brettwooldridge/HikariCP) as the connection pool layer to connect to a {{ site.data.products.dedicated }} CockroachDB cluster. The database is being run on 10 cores across the cluster.
```
rmloveland, line 117: 'CockroachDB Dedicated CockroachDB cluster' ... dang. I think I saw some docs style team discussion about this fly by but didn't get it until now.  Anyway hope we revisit this one someday

rmloveland (approved):
LGTM!

src/current/v23.1/connection-pooling.md
```diff
@@ -49,14 +65,46 @@ If you have a large number of services connecting to the same cluster, make sure
...
 
 Validating connections is typically handled automatically by the connection pool. For example, in HikariCP the connection is validated whenever you request a connection from the pool, and the `keepaliveTime` property allows you to configure an interval to periodically check if the connections in the pool are valid. Whatever connection pool you use, make sure connection validation is enabled when running your application.
 
+<section class="filter-content" markdown="1" data-scope="dedicated selfhosted">
+
+## Monitor new connections
+
+The [`sql.new_conns` metric](metrics.html#available-metrics) exposes the number of new SQL connections per second. A properly configured connection pool will show this value to be in the low single digits. A misconfigured connection pool will result in much higher values. You can expose and monitor this metric in the [DB Console](ui-custom-chart-debug-page.html).
```
sean-, line 104: Can we add some treatment here to suggest connection pools jitter their connection, please?

> Periodic spikes in new connections, also known as thundering herds or connection storms, should be avoided.  Most connection pools offer the ability to stagger their connection age through the use of a connection age "jitter."  Using a connection jitter of 10% of the connection age should be sufficient to smooth out the rate of new connections to a cluster (e.g. max connection age of 300s and a connection age jitter of 30s).

src/current/v23.1/connection-pooling.md
```diff
@@ -96,7 +144,7 @@ Connection conn = ds.getConnection();
 
 <section class="filter-content" markdown="1" data-scope="go">
 
-In this example, a Go application similar to the [basic pgx example](build-a-go-app-with-cockroachdb.html) uses the [pgxpool library](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool) to create a connection pool on a CockroachDB cluster. The database is being run on 10 cores across the cluster.
+In this example, a Go application similar to the [basic pgx example](build-a-go-app-with-cockroachdb.html) uses the [pgxpool library](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool) to create a connection pool on a {{ site.data.products.dedicated }} CockroachDB cluster. The database is being run on 10 cores across the cluster.
```
sean-, line 147: We should update this to `jackc/pgx/v5` and also note that applications using `jackc/pgx` can also include the following query parameters:

* `pool_max_conn_lifetime=300s`
* `pool_max_conn_lifetime_jitter=30s`

https://github.com/jackc/pgx/blob/389931396eea77f4e4b369f659a33b2bff179ae5/pgxpool/pool.go#L270-L286

For example:

```
postgres://my_app@crdb.example.com:26257/crdb_app?sslmode=verify-ca&pool_max_conns=10&pool_max_conn_lifetime=300s&pool_max_conn_lifetime_jitter=30s
```

sean- (commented):
A few small suggestions

sean- (approved):
LGTM!

src/current/v23.1/connection-pooling.md
```diff
@@ -29,34 +29,90 @@ Connection pooling can be a enabled as a feature of the driver, a separate libra
...
 {% include {{page.version.version}}/sql/server-side-connection-limit.md %} This may be useful in addition to your connection pool settings.
 
-## Validating connections in a pool
+</section>
+
+### Size connection pools in multi-region clusters
+
+<section class="filter-content" markdown="1" data-scope="serverless">
+
+Similar to single-region {{ site.data.products.serverless }} clusters, the purpose of connection pools in multi-region {{ site.data.products.serverless }} clusters is to reduce connection latency. Add a connection pool for each region, and size the idle pool for your application and workload.
```
andy-kimball, line 84: Similar comments to above. However, I don't know why the idle pool sizing is mentioned here and not up there; there should be no difference between single and multi-region in that regard. Can't we just say, "there's no difference in the recommendation between single and multi-region"?

src/current/v23.1/connection-pooling.md
```diff
@@ -29,34 +29,90 @@ Connection pooling can be a enabled as a feature of the driver, a separate libra
...
+
+## Validate connections in a pool
 
 After a connection pool initializes connections to CockroachDB clusters, those connections can occasionally break. This could be due to changes in the cluster topography, or rolling upgrades and restarts, or network disruptions. CockroachDB {{ site.data.products.cloud }} clusters periodically are restarted for patch version updates, for example, so previously established connections would be invalid after the restart.
 
 Validating connections is typically handled automatically by the connection pool. For example, in HikariCP the connection is validated whenever you request a connection from the pool, and the `keepaliveTime` property allows you to configure an interval to periodically check if the connections in the pool are valid. Whatever connection pool you use, make sure connection validation is enabled when running your application.
 
+<section class="filter-content" markdown="1" data-scope="dedicated selfhosted">
+
+## Monitor new connections
```
andy-kimball, line 108: This section applies to Serverless just as much as Dedicated/self-hosted - it should be made a common section for all of them.

src/current/v23.1/connection-pooling.md
andy-kimball, line 84: I don't see a need to create a connection pool for each region. That's b/c we have a very strong recommendation that an app worker would run in the same region(s) as the database, and then only connect to its local region. That is, an app worker running in `us-east-1` would only connect to a DB node running in `us-east-1`. It should not be connecting to remote regions. We do not want to dispense any advice that might make customers think we recommend doing this.

src/current/v23.1/connection-pooling.md
```diff
@@ -29,34 +29,90 @@ Connection pooling can be a enabled as a feature of the driver, a separate libra
...
 
 Storage and network performance also will affect the ability of a thread to fully execute. If a thread is blocked by network or storage latency, adding connections to the pool is a good idea so other threads can execute while the original thread is being blocked.
 
 Cockroach Labs performed lab testing of various customer workloads and found no improvement in scalability beyond:
 
-**connections = (number of cores * 4)**
+**active connections = (number of cores * 4)**
```
andy-kimball, line 64: How is this actionable, though? Connection pools do not allow the # active connections to be limited, they only allow the # total connections to be limited. Furthermore, most apps have a number of worker nodes, each of which has its own connections to the database, which complicates things further.

andy-kimball:
I've written [a short doc](https://docs.google.com/document/d/1cDcSj0v8RcMM-qweClopWSlDYQ362V4Kx4B94Qm637I/edit#heading=h.q5m0v6it4z9n) describing how customers should configure their connection pool.

src/current/v23.1/connection-pooling.md
ianjevans: We don't expose `sql.new_conns` in the Cloud Console.

ianjevans: Is it correct that the SQL Connection Attempts graph is how `sql.new_conns` is exposed?

andy-kimball: Yes, that's correct.

src/current/v23.1/connection-pooling.md
```diff
@@ -29,34 +29,90 @@ Connection pooling can be a enabled as a feature of the driver, a separate libra
...
+
+For multi-region clusters, create a connection pool per region, and size the maximum connection pool for each region in your cluster using the same formula as a single-region cluster.
+
+For example, if you have 3 regions in your cluster, and each region has 12 vCPUs, create a connection pool for each region, with each connection pool having a maximum pool size of 48 (12 [processor cores] * 4).
+
+</section>
+
+## Validate connections in a pool
 
 After a connection pool initializes connections to CockroachDB clusters, those connections can occasionally break. This could be due to changes in the cluster topography, or rolling upgrades and restarts, or network disruptions. CockroachDB {{ site.data.products.cloud }} clusters periodically are restarted for patch version updates, for example, so previously established connections would be invalid after the restart.
```
andy-kimball, line 102: > CockroachDB {{ site.data.products.cloud }} clusters periodically are restarted for patch version updates, for example, so previously established connections would be invalid after the restart.

This is not the case for Serverless and hopefully will not be the case for Dedicated soon. Should we be using this as an example going forward, since it might give customers the wrong idea?

ianjevans:
Ok, I've refactored the topic in the PR to move the recommendations into their own section, and streamlined a lot of the info. LMK if this is on the right track.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+### Set the maximum lifetime of connections
+
+The maximum lifetime of a connection should be set to no less than 5 minutes and no more than 30 minutes. Applications that connect to {{ site.data.products.serverless }} clusters should set this to 30 minutes, while applications that connect to {{ site.data.products.dedicated }} and {{ site.data.products.core }} clusters should set this to 5 minutes.
+
+### Set the maximum number of open connections
+
+The maximum number of open connections should be set to a value that is greater than the maximum number of concurrent queries that your application worker expects to make. Doing so minimizes the need to establish new connections.
```
sean-, line 44: > The maximum number of open connections should be set to a value that is greater than the maximum number of concurrent queries that your application worker expects to make. Doing so minimizes the need to establish new connections.

Wait, that's not right: the connection pool acts as a governor, too, to prevent the target cluster from being overwhelmed with active queries.  How about:

> The maximum number of open connections should be scaled based on the workload and capabilities of the target cluster.  The sweet spot will be the minimum of a) the number of vCPUs and IOPS the cluster can perform, times some scaling factor (suggest starting with 4), and b) the maximum number of concurrent queries all workloads targeting this cluster expect to make.  Doing so minimizes the need to establish new connections and keeps the target cluster safe.

If we want to throw in a formula, how about something like this:

* `scaling_factor = 4`
* `num_vcpus_per_node = 16`
* `num_nodes_in_crdb_cluster = 6`
* `num_active_connections = ${num_vcpus_per_node} * ${num_nodes_in_crdb_cluster} * ${scaling_factor} # 384`

If we plug in the avg query time, we can figure out the avg qps:

* `avg_query_time = 10ms`
* `est_num_qps = ${num_active_connections} / ${avg_query_time} # est 38400`

From there, you can add headroom by scaling up the cluster for burst capacity.

* `est_num_qps = 38400`
* `est_num_requests_per_worker = 1600`
* `worker_target_utilization = 0.8`
* `est_num_client_workers = ${est_num_qps} / ${est_num_requests_per_worker} / ${worker_target_utilization} # est 30 workers`

And a max connections per worker of:

* `conn_burst_capacity = 1.5`
* `max_conns = (${num_active_connections} / ${est_num_client_workers}) * ${conn_burst_capacity}`

This gives us a recommendation of about 20 max connections per worker to prevent overload of the target cluster and ensure available connections on every worker to handle this workload.

If some enterprising individual wanted to turn this into an interactive JavaScript calculator presented as a table with various default values, ❤️ very much.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+### Set the maximum number of idle connections
+
+Set the maximum number of idle connections to the same value as the maximum number of open connections.
```
sean-, line 48: This is recommended only for crdb, and if there is a jitter to a connections age to prevent a connection storm (a.k.a., thundering herd).

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+These are Cockroach Labs recommendations for settings common to most connection pooling software. If your application has multiple workers, each worker should be configured based on its own connections and workload.
 
-Creating the appropriate size pool of connections is critical to gaining maximum performance in an application. Too few connections in the pool will result in high latency as each operation waits for a connection to open up. But adding too many connections to the pool can also result in high latency as each connection thread is being run in parallel by the system. The time it takes for many threads to complete in parallel is typically higher than the time it takes a smaller number of threads to run sequentially.
+### Set the maximum lifetime of connections
+
+The maximum lifetime of a connection should be set to no less than 5 minutes and no more than 30 minutes. Applications that connect to {{ site.data.products.serverless }} clusters should set this to 30 minutes, while applications that connect to {{ site.data.products.dedicated }} and {{ site.data.products.core }} clusters should set this to 5 minutes.
```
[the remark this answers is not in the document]
andy-kimball: Ideally, no. But the fact is that 30 minutes is the right setting for Serverless, given there are no load balancing issues or rolling restart issues. The only reason 5 minutes is being suggested for Dedicated/self-hosted is because of those products' limitations. 5 minutes means a lot of unnecessary cycling of connections, with higher risk of elevated tail latency and higher TLS CPU costs for the customer.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+### Set the maximum number of idle connections
+
+Set the maximum number of idle connections to the same value as the maximum number of open connections.
+
+### Set the maximum lifetime of idle connections
+
+Set the maximum lifetime of idle connections to the same value as the maximum lifetime of connections.
 
-Each processor core can only execute one thread at a time. When there are more threads than processor cores, the system will use context switching to [time-slice](https://wikipedia.org/wiki/Preemption_(computing)#Time_slice) the thread execution. For example, if you have a system with a single core and two threads, processing threads 1 and 2 in parallel results in the system context switching to pause execution of thread 1 and begin executing thread 2, and then pause execution of thread 2 to resume executing thread 1. Executing thread 1 completely and then executing thread 2 will be faster because the system doesn't need to context switch, even though thread 2 had to wait until thread 1 fully completed to begin executing.
+### Avoid spikes in new connections
```
andy-kimball, line 54: I think this is the "break point" on the page, where we move from specific connection pooling settings to general advice for managing the connection pool? Shouldn't we have a new sub-section to mark the boundary? The top of the page should just include the absolute bare minimum advice for customers to avoid the worst problems that happen if they don't configure connection pooling. Everything further down should be more detailed "optimization" advice that helps them squeeze out additional performance.

src/current/v23.1/connection-pooling.md
andy-kimball: None of that works for Serverless, though, since there are not fixed number of vCPUs and nodes. I think this section should have solid, basic advice that's no more than 1-2 sentences. Further down the page we should have more detailed advice that explains "advanced" configuration for optimizing app throughput. I was envisioning two sections with names like: "Configure the Connection Pool" and "Optimize the Connection Pool".

The existing docs page is just overwhelming, with so much detail that I fear customer's eyes will glaze over and they'll end up doing nothing at all.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+## Size connection pools
+
+<div class="filters clearfix">
+  <button class="filter-button page-level" data-scope="serverless"><strong>{{ site.data.products.serverless }}</strong></button>
+  <button class="filter-button page-level" data-scope="dedicated"><strong>{{ site.data.products.dedicated }}</strong></button>
+  <button class="filter-button page-level" data-scope="selfhosted"><strong>{{ site.data.products.core }}</strong></button>
+</div>
+
+If your connection pool is properly configured, the total number of connections to your cluster should be at least 100 times larger than the number of new connections per second to your cluster.
```
andy-kimball, line 82: I'd soften this advice a bit, just b/c there may be valid cases where this advice does not apply. Maybe throw in the word "typical" or "usually", etc.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+Idle connections in CockroachDB do not consume many resources compared to PostgreSQL. Unlike PostgreSQL, which has a hard limit of 5000 connections, CockroachDB can safely support tens of thousands of connections.
+
+<section class="filter-content" markdown="1" data-scope="serverless">
+
+In {{ site.data.products.serverless }} clusters, the purpose of the connection pool is to avoid establishing a new SQL connection each time a query is run. You do not need to set a maximum number of simultaneous connections when configuring your connection pool. {{ site.data.products.serverless }} clusters will automatically scale up to meet the demand from multiple simultaneous client connections.
```
andy-kimball, line 88: > You do not need to set a maximum number of simultaneous connections when configuring your connection pool.

This contradicts the advice above where we do say to configure max conns to be > max concurrent queries. It also contradicts advice for Serverless functions, where we say that max conns = 1 in most cases.

This gets into the discussion above around "Set the maximum number of open connections"; maybe we're just forced to give different advice for different products.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+The [SQL Connection Attempts graph]({% link cockroachcloud/metrics-page.md %}#identify-sql-problems) shows how many new connections are being created each second. The [SQL Open Sessions graph]({% link cockroachcloud/metrics-page.md %}#sql-open-sessions) shows the total number of SQL client connections across the cluster. To determine if your connection pool is correctly configured use the metrics from these graphs in the following formula:
+
+**SQL Connection Attempts < SQL Open Sessions/100**
+
+That is, divide the number of open SQL sessions by 100. This result should be less than the number of new SQL connections.
```
andy-kimball, line 94: "the number of new SQL connections per second."

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+<section class="filter-content" markdown="1" data-scope="dedicated selfhosted">
+The [`sql.new_conns` metric]({% link {{ page.version.version }}/metrics.md %}#available-metrics) and [SQL Connection Rate graph]({% link {{ page.version.version }}/ui-sql-dashboard.md %}#sql-connection-rate) expose the number of new SQL connections per second.
+</section>
+
+<section class="filter-content" markdown="1" data-scope="serverless">
+The [SQL Connection Attempts graph]({% link cockroachcloud/metrics-page.md %}#identify-sql-problems) shows the number of new SQL connections per second.
+</section>
+
+A misconfigured connection pool will result in a very large number of new connections per second.
```
andy-kimball, line 138: If I'm a customer, I'd immediately ask what is "a very large number". 10/second? 100/second?

This advice should be the same for Serverless/Dedicated/self-hosted. I suggested a 100:1 ratio as a basic rule of thumb. We can use something different, but I think we should be consistent across the products. IMO, @dikshant and team should own the decision about what to recommend here.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+The [SQL Open Sessions graph]({% link cockroachcloud/metrics-page.md %}#sql-open-sessions) shows the number of active connections per second on your cluster.
+
+</section>
+
+If your connection pool is properly configured, the total number of active connections to your cluster should be at least 100 times larger than the number of new connections per second to your cluster.
+
+## Serverless functions
+
+{{site.data.alerts.callout_info}}
+"Serverless" here refers to stateless, programmatic functions deployed in a cloud environment that provides an execution framework to provision resources dynamically, such as Amazon Lambda functions. It does not refer to {{ site.data.products.serverless }} clusters.
```
andy-kimball, line 167: Do you want to further clarify by saying that Serverless functions can be used with Dedicated or self-hosted clusters?

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
 
 Connection pooling can be a enabled as a feature of the driver, a separate library used in conjunction with a driver, a feature of an application server, or a proxy server that acts as a gateway to the database server.
 
+Creating the appropriate size pool of connections is critical to gaining maximum performance in an application. Too few connections in the pool will result in high latency as each operation waits for a connection to open up. But adding too many connections to the pool can also result in high latency as each connection thread is being run in parallel by the system. The time it takes for many threads to complete in parallel is typically higher than the time it takes a smaller number of threads to run sequentially.
```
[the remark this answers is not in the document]
rafiss: I agree with Sean's changes - we should clarify that the "too many connections" is only a problem if there are too many connections _actively executing a query at the same time_

src/current/v23.1/connection-pooling.md
rafiss: I vote for removing this paragraph. It seems unlikely to me that we'd have a customer suffer from setting a max pool size for a serverless cluster, even though they don't need to.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+<section class="filter-content" markdown="1" data-scope="dedicated selfhosted">
+
+The [`sql.conns` metric]({% link {{ page.version.version }}/metrics.md %}#available-metrics) and [Active SQL Statements graph]({% link {{ page.version.version }}/ui-sql-dashboard.md %}#active-sql-statements) show the number of active connections per second on your cluster or node.
+
+Using the following formula:
+
+**connections = (number of cores * 4)**
+
+if the number of active connections exceeds 4 times the number of cores in your cluster your application is likely not achieving maximum throughput, and you should reduce the maximum number of connections in your connection pool, or consider adding more cores to each node in your cluster.
```
rafiss, line 150: "or consider adding more cores to each node in your cluster." is maybe a bit too specific. Another option is to add more nodes to the cluster.

So this could be rephrased as "scale the cluster by adding more nodes or cores per node" or something.

src/current/v23.1/connection-pooling.md
rafiss: maybe we can say something like "A misconfigured connection pool will result in most database operations requiring a new connection to be established."

src/current/v23.1/connection-pooling.md
rafiss: +1 to the point about readability. I almost wonder if we should make these numerical recommendations formatted in a table instead of textual sections.

src/current/v23.1/connection-pooling.md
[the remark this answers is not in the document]
dikshant: > Do we want different recommendations between self-hosted, Dedicated, and Serverless?

Maybe in the future. For now, I say let's keep it simple and explain the tradeoffs like Rafi suggested. We can document what our server side timeouts are (soon to be changed for Dedicated) and build from there. I would suggest:

> In CockroachDB Dedicated, the nodes are configured to wait as long as 31 minutes for clients to terminate their connections before initiating a restart or shutdown. We recommend setting your client's maximum connection lifetime to 30 minutes or less, and your choice should be guided by your application's specific needs:
> 1. **Choose a Longer Lifetime for Stability and Lower Latency:** If your application benefits from stable connections and reduced frequency of connection attempts, setting a longer maximum connection lifetime can be beneficial. This can improve connection experience for CockroachDB Dedicated cluster updates by minimizing disruptions caused by node restarts.
> 2. **Opt for a Shorter Lifetime for Efficient Load Balancing:** If your application requires the ability to quickly redistribute load across newly introduced nodes in the cluster, a shorter maximum connection lifetime is advisable. Note that in CockroachDB Serverless, this is less of a concern as the system is designed to automatically balance client connections.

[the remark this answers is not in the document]
dikshant: For now, let's keep this as is. And the change I am suggesting above is better done as part of [CC-25475](https://cockroachlabs.atlassian.net/browse/CC-25475). I'll leave a comment there.

src/current/v23.1/connection-pooling.md
dikshant: I like @sean-'s suggestion of giving guidance based on the workload characteristics but I agree that it doesn't make any sense to give guidance based on the number of vCPUs for Serverless. I think it is unavoidable that this will morph into having differing guidance based on a per product basis.

I propose that we leave this current section as is which should be good enough for most cases. We can make a separate change for something like "Advanced connection pool sizing guidance" where we go over the details @sean- described. Especially the interactive calculator would be helpful for customers looking to fine tune their setup. I opened an issue to keep track of that here: https://github.com/cockroachdb/docs/issues/17820

src/current/v23.1/connection-pooling.md
dikshant: @andy-kimball is there a reason 100 is a good rule of thumb? Most drivers and ORMs that set a max number of connections default to a fairly small number of connections (2-10). There are exceptions like npgsql which go as high as 100 but the connection lifetime tends to be long enough (30 mins). Which makes it pretty difficult to hit 100 connections/sec. Plus I have seen some customers who run their applications with 0.5 vCPUs, which will definitely choke when trying to create too many connections at once (around 10).

We can make the consequences a bit clear for now:
> A misconfigured connection pool will result in most database operations requiring a new connection to be established which will increase query latency.

I'll do some investigating as part of https://github.com/cockroachdb/docs/issues/17820 and add a more concrete number there.

dikshant:
I think we should move this forward incrementally and restrict this PR to what it's title suggests. There are two topics which is difficult to cover as a one size fits all solution between Serverless, Dedicated and Self-hosted. I have spun those into separate issues for now and I will work with @ianjevans on prioritizing those after we get this in. 

As part of that work, I'll investigate the defaults our customers are currently running to answer the impact of connections and connection lifetimes on query performance as well as DB load. 

For max connection lifetimes, @ianjevans  let's do that as part of [CC-25475](https://cockroachlabs.atlassian.net/browse/CC-25475) right after this.

I also made https://github.com/cockroachdb/docs/issues/17820 to discuss adding a more comprehensive connection pooling guidance that @sean- suggested. I am also particularly interested in exploring how the [connection pooling suggestions Hikari outlines](https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing), works for CRDB vs Postgres as part of this.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,149 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+To validate that your connection pool is correctly configured, monitor the SQL connection metrics and graphs on your cluster.
+
+### Monitor new connections
+
+<section class="filter-content" markdown="1" data-scope="dedicated selfhosted">
+The [`sql.new_conns` metric]({% link {{ page.version.version }}/metrics.md %}#available-metrics) and [SQL Connection Rate graph]({% link {{ page.version.version }}/ui-sql-dashboard.md %}#sql-connection-rate) expose the number of new SQL connections per second.
+</section>
+
+<section class="filter-content" markdown="1" data-scope="serverless">
```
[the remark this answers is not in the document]
ianjevans: You should see a section heading "Monitor new connections" with content that changes based on which filter is selected: Serverless, Dedicated, or Self-Hosted.

src/current/v23.1/connection-pooling.md
```diff
@@ -25,37 +25,151 @@ Each time an application reads or writes data, it will request one of the connec
...
+
+### Validate connections in a pool
+
+After a connection pool initializes connections to CockroachDB clusters, those connections can occasionally break. This could be due to changes in the cluster topography, or rolling upgrades and restarts, or network disruptions.
+
+Validating connections is typically handled automatically by the connection pool. For example, in HikariCP the connection is validated whenever you request a connection from the pool, and the `keepaliveTime` property allows you to configure an interval to periodically check if the connections in the pool are valid. Whatever connection pool you use, make sure connection validation is enabled when running your application.
```
rmloveland, line 66: Since we're mentioning HikariCP and one of its config properties, suggest linking to that info.  IDK if this is the best place, but I found it mentioned in the project README: https://github.com/brettwooldridge/HikariCP#frequently-used

rmloveland (approved):
LGTM!

I know this PR has been A Journey (tm), so I don't think it's necessary here, but I think a good followup would be to edit the 'Connection Pools' section of https://www.cockroachlabs.com/docs/stable/recommended-production-settings to remove what appears to be some dupe instructions around a cluster setting, total # of connections, etc.  (I'm not sure without a more careful read if it matches what is here or is now superseded by what is here)

theodore-hyman:
TYSM team massive improvement here appreciate the collaboration and efforts

PR #17544, opened 2023-07-25
DOC-8310: Update MOLT Verify with new build instructions

Amruta-Ranade:
Addresses [DOC-8310](https://cockroachlabs.atlassian.net/jira/software/c/projects/DOC/issues/DOC-8310)


cotedm (approved):
LGTM

otan (approved):

gemma-shay (commented):
Just had some formatting suggestions. Also, we might need more input here because I don't think this works right out of the box. Testing on my own terminal, it seemed like I had to add the downloaded file to my path and run the full name of the file (`./downloads/molt.darwin.amd64 ...`), not just `./molt`.

Amruta-Ranade:
> Just had some formatting suggestions. Also, we might need more input here because I don't think this works right out of the box. Testing on my own terminal, it seemed like I had to add the downloaded file to my path and run the full name of the file (`./downloads/molt.darwin.amd64 ...`), not just `./molt`.

Hmm we do say "Optionally, add the binary to your `PATH` so you can execute the `molt verify` command from any shell." @otan should the user have to use the full file name? or am I missing a step?

cotedm:
I think it's a good point that I hadn't caught because I had just run with the full file name `./molt.darwin.arm64`. I think the move is to add a step to rename the downloaded binary to `molt` and then run it.

gemma-shay (approved):
thx!

PR #17548, opened 2023-07-25
IA security / CLI edits

lnhsingh:
- Removed the old manage.json file as it's no longer used
- Removed the erroneous security links from the Reference > Cockroach Commands section
- Reordered the Reference > Security section to match the proposal

### DIFF
--- a/src/current/_includes/v23.1/sidebar-data/reference.json
+++ b/src/current/_includes/v23.1/sidebar-data/reference.json
@@ -85,70 +85,6 @@
           {
             "title": "Overview",
             "urls": [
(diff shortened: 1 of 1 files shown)


taroface:
The System Catalogs menu is nested under Security but needs to be pulled out to top level below Security:

image

In Security, something vaguely feels off about "SASL/SCRAM-SHA-256 Secure Password-based Authentication" going right above "Authentication", but I don't know enough to say. I know this was in the IA and I can't recall why. Just putting it out there.

Otherwise LGTM!

lnhsingh:
TFTR @taroface! I adjusted the Authetication section as well.

PR #17553, opened 2023-07-26
DOC-8113: Use Jekyll links to link to different docs pages

nickvigilante:
Fixes DOC-8113

What this PR does:

- For all docs that have links pointing to the same folder (e.g., v23.1, cockroachcloud, releases), use `{% link path/to/page.md %}`.
- For all docs that have links pointing to a different folder, use the fully qualified URL (e.g., `https://www.cockroachlabs.com/docs/releases/v23.1`). This is because we'll eventually be breaking off each folder into its own microsite, so rather than update the link twice, we just update it once.
- For any reference to `/{{ site.versions["stable"] }}/` or `/stable/` within links in cockroachcloud only, we replace it with `/{{ site.current_cloud_version }}/`.
- Updated all advisories page to point links to docs to the latest applicable version to the tech advisory instead of stable docs.
- Updated all releases pages to point links to the docs of the same version as the release instead of the stable docs.
- Moved all HTML comments to Liquid comments, as the contents of HTML comments both get parsed by the Liquid parser and appear on the production site.


mdlinville (approved):
LGTM, skimmed the latest pagediff.txt and did a spot check of several of the updated URLs.

mikeCRL (approved):

PR #17557, opened 2023-07-26
Update language around webhook retry.max config setting

kathancox:
Fixes DOC-8305

Updates language to clarify the type of retry that `webhook_sink_config = Retry.Max` will affect.


HonoreDB (approved):

src/current/v23.1/changefeed-sinks.md
```diff
@@ -420,7 +420,7 @@ Field              | Type                | Description      | Default
 `Flush.Messages`   | [`INT`](int.html)   | The batch is flushed and its messages are sent when it contains this many messages. | `0`
 `Flush.Bytes`      | [`INT`](int.html)   | The batch is flushed when the total byte sizes of all its messages reaches this threshold. | `0`
 `Flush.Frequency`  | [`INTERVAL`](interval.html) | When this amount of time has passed since the **first** received message in the batch without it flushing, it should be flushed. | `"0s"`
-`Retry.Max`        | [`INT`](int.html) | The maximum number of attempted retries after sending a message batch in an HTTP request fails. Specify either an integer greater than zero or the string `inf` to retry indefinitely. | `3`
+`Retry.Max`        | [`INT`](int.html) | The maximum number of attempted HTTP retries after sending a message batch in an HTTP request fails. Specify either an integer greater than zero or the string `inf` to retry indefinitely. This only affects HTTP retries, not other causes of [duplicate messages](changefeed-messages.html#duplicate-messages). Note that setting this field to a low number will not prevent the changefeed from retrying indefinitely. | `3`
```
ashleykwtam, line 423: "low number" is a little ambiguous here - can a number be provided or does it vary for the retries?

ashleykwtam (changes requested):
"low number" is a little ambiguous here - can a number be provided or does it vary for the retries?

src/current/v23.1/changefeed-sinks.md
kathancox: Low number is not really the point here, I would say. Because we're saying this only affects HTTP retries, not more general cases of changefeed retries. And "low number" could be dependent on a user's specific case. 

@HonoreDB Are we still correct here, if we simply say "Note that setting this field will not precent the changefeed from retrying."?

ashleykwtam: Waiting on @HonoreDB to confirm else LGTM!

HonoreDB: Yup, setting this field will not prevent a changefeed from retrying.

PR #17546, opened 2023-07-25
New ccloud flag to skip ip allowlist checks

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-8205

Added flag info and release note for new ccloud version + flag
Version for install instructions was already updated

PR #17566, opened 2023-07-27
Update example cluster name to be positive

gemma-shay:
follow up to https://github.com/cockroachdb/docs/pull/17546
inspired by Blue's Clues


mdlinville (approved):
This makes me so happy today.

PR #17585, opened 2023-07-31
Disable edge function within the docs repo

nickvigilante:
We experienced a brief outage on Thursday that affected some writers' ability to run Netlify builds on their PRs. This seems to be related to an outage on Netlify's end. While the outage was outside of our control, we no longer need the edge function directly within the docs repo, as we have the edge function hosted in the main Marketing site to strip extensions and slashes for us.


florence-crl (approved):
LGTM

PR #17604, opened 2023-08-03
Edit SH > Backup section

lnhsingh:
Closes DOC-8415


kathancox (approved):
I think this looks good according to Liv's comment in the issue, and we do not need to differentiate the backup type in SH.

I had one general question, which does not block, more curiosity: are we moving to just naming landing pages "Overview" with thought that implementing breadcrumbs in the future will be a signal for users on location.... and breadcrumbs will be enough for users to land on this page from Google say?

lnhsingh:
> are we moving to just naming landing pages "Overview" with thought that implementing breadcrumbs in the future will be a signal for users on location.... and breadcrumbs will be enough for users to land on this page from Google say?

Yeah, I think breadcrumbs in the future will help users locate where they are. However, the current changes are just to the sidenav, so if a user were to come to a page through Google, the large page title will still say `Backup and Restore Overview` (for example). I don't think it'll be too jarring.

PR #17605, opened 2023-08-03
Product name variables

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-7036

-Removed CockroachDB from Serverless, Dedicated, Cloud, and Self-Hosted variables
-Enterprise was already used without CockroachDB
-Added the word "CockroachDB" before all instances of affected variable use in docs
-Changed the `db` variable name to `cloud` (was originally just my attempt at shorthand for CockroachDB Cloud, but it's confusing)


mdlinville (approved):
I did a spot check and the variable changes LGTM. However, I noticed on pages like https://deploy-preview-17605--cockroachdb-docs.netlify.app/docs/v20.1/build-a-nodejs-app-with-cockroachdb-sequelize.html#step-2-create-a-database and https://deploy-preview-17605--cockroachdb-docs.netlify.app/docs/v20.1/build-a-nodejs-app-with-cockroachdb.html that the "Create a database" section is blank. This is specific to 20.1 and also exists in production: https://www.cockroachlabs.com/docs/v20.1/build-a-nodejs-app-with-cockroachdb-sequelize Any idea what might be going on?

gemma-shay:
@mdlinville not sure, the missing content is in includes so it must be some issue with that, but i checked over the files and they all appear to exist & have content.

mdlinville:
> @mdlinville not sure, the missing content is in includes so it must be some issue with that, but i checked over the files and they all appear to exist & have content.

This is just a theory, but try replacing `include` with `include_cached` on lines 19 and 23 of `src/current/v20.1/build-a-nodejs-app-with-cockroachdb.md`. Where the exact same include content should be shown (there are no include-or-page-level variables to resolve), it should always use the cache to avoid nondeterministic results on following pages that have the same include.

mdlinville:
> https://deploy-preview-17605--cockroachdb-docs.netlify.app/docs/v20.1/build-a-nodejs-app-with-cockroachdb.html

It's not `include_cached`. It looks like the filter-tabs changed at 20.2 and were partially backported to 20.1, but the tutorial wasn't updated to use the "local" and "cockroachcloud" filter-tabs, instead just has a "local" one, but the content itself is not tagged into `<section>`s to match them. Maybe not worth worrying about it for such an old version.

mdlinville (approved):

PR #17606, opened 2023-08-03
Better explain how CRDB fits into the CAP theorem

ianjevans:
This PR fixes some language about how CRDB fits into the CAP theorem framework of distributed systems.

Fixes DOC-8273.


src/current/v22.2/developer-basics.md
```diff
@@ -16,15 +16,19 @@ Note that the sections that follow were written for the purposes of orienting ap
 
 ### How transactions work in CockroachDB
 
-CockroachDB is designed to make your data scalable and consistent. All stored records are replicated across a distributed deployment of database instances, and all database transactions committed in CockroachDB satisfy [ACID properties](https://en.wikipedia.org/wiki/ACID) and the [CAP theorem](https://en.wikipedia.org/wiki/CAP_theorem). This means that when you store data in CockroachDB, the data will be valid and anomaly-free, even in the event of a system error or power failure.
+CockroachDB is designed to make your data scalable and consistent. All stored records are replicated across a distributed deployment of database instances, and all database transactions committed in CockroachDB satisfy [ACID properties](https://en.wikipedia.org/wiki/ACID). This makes CockroachDB a CAP-Consistent (CP) system under the [CAP theorem](https://wikipedia.org/wiki/CAP_theorem), and means that when you store data in CockroachDB, the data will be valid and anomaly-free, even in the event of a system error or power failure.
```
bdarnell, line 19: There is unfortunately no connection between the C in ACID and the C in CAP, so saying "this makes" is not correct. I suggest removing any mention of the CAP theorem from this paragraph and putting the new CAP-related paragaphs in a new section "How replication works in CockroachDB" (or omitting them? ACID transactions belong in a "developer basics" doc but I'm not sure the CAP theorem does)

src/current/v22.2/developer-basics.md
```diff
@@ -16,15 +16,19 @@ Note that the sections that follow were written for the purposes of orienting ap
...
+
+As a CAP-Consistent (CP) system under the CAP theorem, CockroachDB prioritizes data consistency. Each piece of data has at least 3 replicas. When reading or writing data, one of the replicas is granted a [lease](architecture/replication-layer.html#leases), or temporary ownership, of the data. The lease expires after a few seconds by default. The leaseholder can complete a read without needing to communicate with the other replicas. When the leaseholder is sent a write request, a majority of the replica nodes must be able to communicate with each other to coordinate the write. This ensures that the most recent write is always available to subsequent reads.
```
bdarnell, line 23: The first sentence kind of gets the causality backwards - we didn't set out to make a CAP-consistent database and therefore we prioritize data consistency; we set out to prioritize data consistency, and therefore we are classified as consistent under the CAP theorem. I think the CAP theorem itself should be a parenthetical or footnote - if you don't know about it, you don't need to know about it (in a "developer basics" doc); we just mention it for the benefit of folks who already know it and are looking for that keyword. Instead of the CAP theorem, the best framework for talking about our consistency model is "no stale reads" (https://www.cockroachlabs.com/blog/consistency-model/)

The key concept in our approach to consistency (and durability) is not the leaseholder, it's the quorum: writes go to a quorum of replicas before being considered complete. Data in CRDB is available as long as a quorum of replicas is available. The leaseholder is an optimization layered on top of quorums that improves read performance (and I think I'd omit it from "developer basics").

src/current/v22.2/developer-basics.md
ianjevans: We don't really talk about the CAP theorem in the architecture docs. Is that a better place for this info?

bdarnell (approved):

lnhsingh (approved):

PR #17623, opened 2023-08-07
DOC-6855: Enable deprecation warnings and fix math.div issue

nickvigilante:
Fixes DOC-6855


nickvigilante:
Abandoning this for now, because I think there's a bug in Bootstrap in this version. We should look to upgrade our version of Bootstrap to the latest version and separate our desired settings from the actual BS CSS files.

PR #17629, opened 2023-08-08
DOC-7633 recommendations for backup validation plan

ashleykwtam:
Addresses DOC-7633:
- Providing recommendations on how customers should test their backups following an inquiry from a partner from Erlang Systems (supporting Hollywoodbets) 

To consider:
- @dt gives % confidence levels in each option - should we include these in the recommendations?

Preview:
https://deploy-preview-17629--cockroachdb-docs.netlify.app/docs/stable/backup-validation#recommendations


livlobo (commented):
Thanks for adding these recommendations Ashley. These are helpful!

I think the difference between periodic and regular is a bit fuzzy. 

Could we use something like: 
Run schema_only backups very frequently
Run verify table data, frequently 
Run restore, somewhat frequently

We could also add something like:
The actual frequency you should run backup validation tests depends on the criticality of your data, the rate of change in your database, and your organization's disaster recovery and business continuity requirements.

As for confidence percentages, we can mention:
schema_only provides high confidence in teh recoverability of the backup
verify_data provides slightly higher confidence than schema only, because it checks for corruptions ... but we expect these to be rare (provide reasoning for why it;s oly slightly better)
restore - provides full confidence

cc @dt

PR #17633, opened 2023-08-08
add known limitation + workaround on using DMS with RBR tables

taroface:
DOC-6208

- Document a known limitation + workaround when using DMS to migrate to a multi-region cluster with RBR tables.
- Clarify the `crdb_region` table in the RBR table description.


otan (approved):
tables with hash sharded indexes will have a similar problem, if you want to introduce that later.

taroface:
> tables with hash sharded indexes will have a similar problem, if you want to introduce that later.

For sure. Can you provide more context on this, and we can add to a separate issue?

otan:
in general, hidden columns are not well supported by DMS. examples: `rowid` is an implicit hidden column, `regional by row` adds a hidden `crdb_region` column at the front. hash sharded indexes (https://www.cockroachlabs.com/docs/stable/hash-sharded-indexes#:~:text=Hash%2Dsharded%20indexes%20distribute%20sequential,small%20cost%20to%20read%20performance.) have the same issue as it adds an implicit `crdb_internal_ts_shard_16` column at the start

Amruta-Ranade (approved):

PR #17658, opened 2023-08-14
Remove images of cc console

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-6847
Console images get outdated extremely quickly, so opted to remove these.


mdlinville (commented):
Should you remove the actual images from the repo too?

mdlinville (approved):

PR #17661, opened 2023-08-14
Add note to Style Guide on using Liquid comments over HTML comments

nickvigilante:


StyleGuide.md
```diff
@@ -1365,6 +1366,49 @@ Considerations:
...
+There may be situations that require adding an explanation below a piece of content or to temporarily suppress content within a page. In those situations, use [Liquid comments](https://shopify.github.io/liquid/tags/template/#comment).
+
+Page source:
+
+```
+This sentence is visible!
+{% comment %}
+This sentence is not visible!
+{% endcomment %}
+This sentence is also {% comment %}not {% endcomment %}visible.
```
mdlinville, line 1380: I think only the word "not" is not visible, right? Does it really hide a whole line that has a comment in the middle of it?

StyleGuide.md
```diff
@@ -1365,6 +1366,49 @@ Considerations:
...
+```
+
+Final page HTML:
+
+```
+<p>This sentence is visible!</p>
+<p>This setence is also visible.</p>
+```
+
+Do not use HTML comments (`<!-- -->`), as the content will still be made available in the page source in production. Additionally, any HTML comment content is subject to the Liquid parser, so any Liquid inside of an HTML comment is still processed and can produce errors (e.g., an include or a link is broken).
```
mdlinville, line 1390:
```suggestion
Do not use HTML comments (`<!-- -->`), because HTML comments are visible in the page's HTML source code in production. Additionally, any HTML comment content must be processed by the Liquid parser, so any Liquid within an HTML comment is still processed and can produce errors, such as a broken include or broken link.
```

StyleGuide.md
```diff
@@ -1365,6 +1366,49 @@ Considerations:
...
+{% endcomment %}
+```
+
+```
+{% comment %}
+FIXME: Update example SQL commands
+{% endcomment %}
+```
+
+Many popular code editors feature extensions that can highlight `TODO`s across the repository. One such extension is [Todo Tree](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree).
```
mdlinville, line 1410: I have reservations about the allusion to profanity in the itemName. I wonder if there is another option we could point to instead.

mdlinville (approved):
Approved with a suggestion and a reservation.

StyleGuide.md
nickvigilante:
```
{% comment %}
This sentence is not visible!
{% endcomment %}
```
The paragraph containing `This sentence is not visible!` gets suppressed. In line 1380, only the word `not` and the space following it is suppressed. It doesn't hide the whole line. It only hides what is strictly in between the `{% comment %}` and `{% endcomment %}` tags.

StyleGuide.md
nickvigilante: Found an alternative. It's unfortunate, because it's a good extension.

StyleGuide.md
mdlinville, line 1380:
```suggestion
This sentence is visible except for a single commented word: {% comment %}not{% endcomment %}
```

StyleGuide.md
```diff
@@ -1365,6 +1366,49 @@ Considerations:
...
+This sentence is not visible!
+{% endcomment %}
+This sentence is also {% comment %}not {% endcomment %}visible.
+```
+
+Final page HTML:
+
+```
+<p>This sentence is visible!</p>
+<p>This setence is also visible.</p>
```
mdlinville, line 1387: This output is no longer accurate either.

mdlinville (approved):

PR #17668, opened 2023-08-15
Recommend 4 hour CRON schedule for TTL

ianjevans:
This PR also adds some additional info about how TTL works with admission control.

It also removes the performance limitation language when using secondary indexes.

Fixes DOC-7698.
Fixes DOC-8440.


dikshant:
Maybe it would be good to explain how Scans work with TTL and how it is more efficient than doing a full table scan here. @rafiss @ecwall can you help explain this a bit more.

src/current/v22.2/row-level-ttl.md
```diff
@@ -352,7 +352,7 @@ SELECT relname, reloptions FROM pg_class WHERE relname = 'events';
 
 ### Control how often the TTL job runs
 
-Setting a TTL on a table controls when the rows therein are considered expired, but it only says that such rows _may_ be deleted at any time after the expiration. To control how often the TTL deletion job runs, use the [`ttl_job_cron` storage parameter](#param-ttl-job-cron), which supports [CRON syntax](https://cron.help).
+Setting a TTL on a table controls when the rows therein are considered expired, but it only says that such rows _may_ be deleted at any time after the expiration. To control how often the TTL deletion job runs, use the [`ttl_job_cron` storage parameter](#param-ttl-job-cron), which supports [CRON syntax](https://cron.help). Cockroach Labs recommends setting `ttl_job_cron` to match the [`gc.ttlseconds`]({% link {{ page.version.version }}/configure-replication-zones.md %}#gc-ttlseconds) setting, which is the garbage collection interval for the cluster. The default value of `gc.ttlseconds` is 14400, or 4 hours. The CRON pattern for every four hours is `'0 */4 * * *'`.
```
dikshant, line 355: > Cockroach Labs recommends setting `ttl_job_cron` to match the [`gc.ttlseconds`]
I think we should say Cockroach Labs recommends setting `ttl_job_cron` to be equal to or higher than the [`gc.ttlseconds`].

src/current/v23.2/row-level-ttl.md
```diff
@@ -26,7 +26,7 @@ At a high level, Row-Level TTL works by:
...
 - As part of the above process, deciding how many rows to [`SELECT`]({% link {{ page.version.version }}/select-clause.md %}) and [`DELETE`]({% link {{ page.version.version }}/delete.md %}) at once in each of the above queries.
 - Running the SQL queries described above in parallel as [background jobs]({% link {{ page.version.version }}/show-jobs.md %}).
-- To minimize the performance impact on foreground application queries, the background deletion queries are rate limited; they are also submitted at a lower priority level using the [admission control system]({% link {{ page.version.version }}/admission-control.md %}).
+- To minimize the performance impact on foreground application queries, the background deletion queries are rate limited; they are also submitted at a lower priority level using the [admission control system]({% link {{ page.version.version }}/admission-control.md %}). When foreground traffic increases, CockroachDB will reduce the resources allocated to TTL deletes to handle the foreground traffic. When foreground traffic decreases, CockroachDB will increase the resources allocated to TTL deletes.
```
dikshant, line 29: For 23.2 we can also add a bit about how we now use the elastic CPU limiter to limit TTL selects from taking too much of cluster resources. That work is explained in more detail on this PR: https://github.com/cockroachdb/cockroach/pull/108815

dikshant (changes requested):
Added a few comments.

dikshant (approved):
LGTM :shipit:

src/current/v23.2/row-level-ttl.md
```diff
@@ -26,7 +26,9 @@ At a high level, Row-Level TTL works by:
...
+
+- {% include_cached new-in.html version="v23.2" %} Latency of row-level TTL queries is further reduced by using the elastic CPU limiter, which dynamically controls the total CPU percentage used by row-level TTL reads. The elastic CPU limiter can be disabled for row-level TTL queries by setting the `kvadmission.ttl_read_elastic_control.enabled` and `sqladmission.low_pri_read_response_elastic_control.enabled` [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}) to `false`.
```
rmloveland, line 31: intentional that we are documenting non-public cluster settings here?  not seeing either of these show up in https://www.cockroachlabs.com/docs/v23.2/cluster-settings 

could just as easily get support issues from people turning these off and having problems, right?  my understanding is if cluster settings are not public/visible we don't support customers using them without Support team involved

meta comment (not fixable in a PR but for your consideration @dikshant) : 

do we have broad alignment with Eng on what cluster settings are public vs non-public, and what that actually means?  I feel like if we audited our docs for "non-public" cluster settings we would find A TON.  So at that point what even is the point of this public/non-public distinction? And does everyone align on what that distinction is? (end meta comment lol)

rmloveland (approved):
LGTM!

src/current/v23.2/row-level-ttl.md
rafiss, line 31: is `kvadmission.ttl_read_elastic_control.enabled`` a typo? i don't see this defined anywhere in the code and i am not familiar with it.

rafiss: i think this should be `kvadmission.low_pri_read_elastic_control.enabled`

ianjevans: It's from this PR, linked to from a previous comment from Dikshant. https://github.com/cockroachdb/cockroach/pull/108815

rafiss: ah i see, that PR was still in review at that time. an earlier draft called it `kvadmission.ttl_read_elastic_control.enabled` but it is now `kvadmission.low_pri_read_elastic_control.enabled` and was merged with that name.

PR #17669, opened 2023-08-15
Update head.html

nickvigilante:
Fix a bug where older docs pages were still being indexed


rmloveland (approved):
LGTM - I don't claim to understand this but if it works it works!

PR #17679, opened 2023-08-16
Update heading in stream data sidenav

kathancox:
Fixes DOC-8583

This is a quick fix PR to rename one of the headings in the Stream Data side nav to better describe the pages that fall underneath and to better relate to the [heading on the CDC overview page](https://www.cockroachlabs.com/docs/stable/change-data-capture-overview#optimize-a-changefeed-for-your-workload).

### DIFF
--- a/src/current/_includes/v22.2/sidebar-data/stream.json
+++ b/src/current/_includes/v22.2/sidebar-data/stream.json
@@ -38,7 +38,7 @@
       ]
     },
     {
-      "title": "Work with Changefeeds",
+      "title": "Optimize Changefeeds",
       "items": [
         {
           "title": "Change Data Capture Transformations",
--- a/src/current/_includes/v23.1/sidebar-data/stream-data.json
+++ b/src/current/_includes/v23.1/sidebar-data/stream-data.json
@@ -38,7 +38,7 @@
       ]
     },
     {
-      "title": "Work with Changefeeds",
+      "title": "Optimize Changefeeds",
       "items": [
         {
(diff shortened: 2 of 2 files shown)


mdlinville (approved):
LGTM but I suspect that link titles within pages also need to be changed, and it might be worthwhile to backport the results to 22.2 also. WDYT?

kathancox:
> LGTM but I suspect that link titles within pages also need to be changed, and it might be worthwhile to backport the results to 22.2 also. WDYT?

@mdlinville Hmmm, so this is only a dropdown title in the side nav, so not used as part of a link for a page title, or really referenced anywhere. Though maybe I'm not thinking about this correctly?

Re v22.2. The CDC overview changes were only updated for v23.1, but I'm definitely opening to just updating the sidenav title anyway as it does not affect anything other than the dropdown.

mdlinville:
Makes sense. LGTM. On Aug 16, 2023, at 11:35 AM, Kathryn Hancox ***@***.***> wrote:﻿

LGTM but I suspect that link titles within pages also need to be changed, and it might be worthwhile to backport the results to 22.2 also. WDYT?

@mdlinville Hmmm, so this is only a dropdown title in the side nav, so not used as part of a link for a page title, or really referenced anywhere. Though maybe I'm not thinking about this correctly?
Re v22.2. The CDC overview changes were only updated for v23.1, but I'm definitely opening to just updating the sidenav title anyway as it does not affect anything other than the dropdown.

PR #17688, opened 2023-08-17
self-hosted core in overview table

MichaelTrestman:
This was a direct product request:
https://cockroachlabs.slack.com/archives/C01PHNMUFLN/p1692291727879799?thread_ts=1692281978.236079&cid=C01PHNMUFLN

https://cockroachlabs.atlassian.net/browse/DOC-8593

### DIFF
--- a/src/current/v23.1/security-reference/security-overview.md
+++ b/src/current/v23.1/security-reference/security-overview.md
@@ -18,7 +18,7 @@ docs_area: reference.security
 
 Learn more: [Integrate {{ site.data.products.dedicated }} with Satori]({% link {{ page.version.version }}/satori-integration.md %})
 
-### Self-Hosted Core
+### Self-Hosted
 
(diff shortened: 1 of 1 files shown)


kathancox:
As discussed in the slack thread, it might be good to prefix Enterprise with "Self-Hosted" here to differentiate. Otherwise LGTM.

src/current/v23.1/security-reference/security-overview.md
```diff
@@ -18,7 +18,7 @@ docs_area: reference.security
 
 Learn more: [Integrate {{ site.data.products.dedicated }} with Satori]({% link {{ page.version.version }}/satori-integration.md %})
 
-### Self-Hosted
+### Self-Hosted Core
```
abhinavg6, line 21: I think this should remain as-is, as it's about the overall SH offering.

abhinavg6 (changes requested):
Requesting one change, thanks.

abhinavg6 (approved):
LGTM

PR #17700, opened 2023-08-18
Add docs on CPU-based replica rebalancing

rmloveland:
Fixes:

- DOC-6917
- DOC-6989
- DOC-6999
- DOC-7746
- DOC-8564

Summary of changes:

- Architecture > Replication Layer: Update the 'Load-based splitting' section to describe how replica count & CPU usage are used to rebalance replicas (not disk usage)
- Cluster Setup Troubleshooting: Add a new section 'Disk usage is different across nodes in the cluster' to explain that this is expected behavior, and link to load-based replication architecture docs
- Load-based splitting: Add a new section 'Monitor load-based splitting' so users can know where to look in logs and metrics to see how it is working
- Node Shutdown: Update the 'Size and replication factor' to explain how replicas are rebalanced when a node is decommissioned and immediately replaced, and include a link to docs about how replica rebalancing works

### DIFF
--- a/src/current/v23.1/architecture/replication-layer.md
+++ b/src/current/v23.1/architecture/replication-layer.md
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
 
 #### Load-based replica rebalancing
 
-In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically based on the relative load across the nodes within a cluster. For more information, see the `kv.allocator.load_based_rebalancing` and `kv.allocator.qps_rebalance_threshold` [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}). Note that depending on the needs of your deployment, you can exercise additional control over the location of leases and replicas by [configuring replication zones]({% link {{ page.version.version }}/configure-replication-zones.md %}).
+In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically across the cluster based on a combination of:
+
+1. Replica count.
+1. CPU usage (if [`kv.allocator.load_based_rebalancing.objective`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-load-based-rebalancing-objective) is set to `cpu`, which is the default in CockroachDB v23.1 and later)
+
+Note that disk utilization per node is not one of the rebalancing criteria. For more information, see [Disk utilization is different across nodes in the cluster]({% link {{ page.version.version }}/cluster-setup-troubleshooting.md %}#disk-utilization-is-different-across-nodes-in-the-cluster).
+
+Load-based replica rebalancing operates in conjunction with [load-based splitting]({% link {{ page.version.version }}/load-based-splitting.md %}) of [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range).
+
+For more information on how to control load-based rebalancing, see the following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}):
+
+- [`kv.allocator.store_cpu_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-store-cpu-rebalance-threshold)
+- [`kv.allocator.range_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-range-rebalance-threshold)
+
+Depending on the needs of your deployment, you can exercise additional control over the locations of leases and replicas using [multi-region SQL]({% link {{ page.version.version }}/multiregion-overview.md %}), or by [configuring replication zones]({% link {{ page.version.version }}/configure-replication-zones.md %}).
 
 ### Important values and timeouts
 
--- a/src/current/v23.1/cluster-setup-troubleshooting.md
+++ b/src/current/v23.1/cluster-setup-troubleshooting.md
@@ -441,6 +441,21 @@ CockroachDB's built-in disk stall detection works as follows:
 
 - During [node liveness heartbeats](#node-liveness-issues), the [storage engine]({% link {{ page.version.version }}/architecture/storage-layer.md %}) writes to disk as part of the node liveness heartbeat process.
 
+#### Disk utilization is different across nodes in the cluster
+
+This is expected behavior.
+
+CockroachDB uses [load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing) to spread [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) across the nodes in a cluster.
+
+The rebalancing criteria for load-based replica rebalancing do not include the percentage of disk utilized per node. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. The proportions of larger and smaller ranges on each node balance each other out on average, so disk utilization differences across nodes should be relatively small.
+
+However, in some cases a majority of the largest (or smallest) ranges are on one node, which will result in imbalanced utilization. Normally that shouldn't be a problem with [sufficiently provisioned storage]({% link {{ page.version.version }}/recommended-production-settings.md %}#storage). If this imbalance is causing issues, please [contact Support]({% link {{ page.version.version }}/support-resources.md %}) for guidance you on how to manually rebalance your cluster's disk usage.
+
+Finally, note that although the replica allocator does not rebalance based on disk utilization during normal operation, it does have the following mechanisms to help protect against [full disks](#disks-filling-up):
+
+- It will avoid moving replicas onto a disk that is more than 92.5% full.
+- It will start moving replicas off of a disk that is 95% full.
+
 ## CPU issues
 
 #### CPU is insufficient for the workload
--- a/src/current/v23.1/load-based-splitting.md
+++ b/src/current/v23.1/load-based-splitting.md
@@ -69,6 +69,27 @@ However, by leveraging load-based splitting, the cluster can understand load on 
 
 This benefit is further amplified by [load-based rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing), which ensures that all nodes contain replicas with a roughly equal load. By evenly distributing load throughout your cluster, it's easier to prevent bottlenecks from arising, as well as simplifying hardware forecasting.
 
+## Monitor load-based splitting
+
+The [`INFO`]({% link {{ page.version.version }}/logging.md %}#info) log message
+
+~~~
+no split key found: ...
+~~~
+
+indicates that CockroachDB wants to [split the range]({% link {{ page.version.version }}/architecture/distribution-layer.md %}#range-splits) due to load, but couldn’t find a key to split at.
+
+Usually this log message can be ignored, unless it repeatedly shows up, which can indicate there is a load imbalance problem in the cluster. If there is a load imbalance problem, it could be because a [hot range]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}) cannot be split (because it's really a [hot key]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}#range-report)).
+
+You can see how often a split key cannot be found over time by looking at the following [time-series metric]({% link {{ page.version.version }}/monitoring-and-alerting.md %}#prometheus-endpoint):
+
+- `kv.loadsplitter.nosplitkey`
+
+This metric is directly related to the log message described above.
+
+For more information about how to reduce hot spots (including hot ranges) on your cluster, see [Hot spots]({% link {{ page.version.version }}/performance-best-practices-overview.md %}#hot-spots).
+
 ## See also
 
 - [`SET CLUSTER SETTING`]({% link {{ page.version.version }}/set-cluster-setting.md %})
+- [Load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing)
--- a/src/current/v23.1/node-shutdown.md
+++ b/src/current/v23.1/node-shutdown.md
@@ -290,6 +290,10 @@ To determine an appropriate termination grace period:
 
 Before decommissioning a node, make sure other nodes are available to take over the range replicas from the node. If fewer nodes are available than the replication factor, CockroachDB will automatically reduce the replication factor (for example, from 5 to 3) to try to allow the decommission to succeed. However, the replication factor will not be reduced lower than 3. If three nodes are not available, the decommissioning process will hang indefinitely until nodes are added or you update the zone configurations to use a replication factor of 1.
 
+Note that when you decommission a node and immediately add another node, CockroachDB does **not** simply move all of the replicas from the decommissioned node to the newly added node. Instead, replicas are placed across all nodes in the cluster. This speeds up the decommissioning process by spreading the load. The new node will eventually "catch up" with the rest of the cluster.
+
+This can lead to disk utilization imbalance across nodes. **This is expected behavior**, since disk utilization per node is not one of the rebalancing criteria. For more information, see [Disk utilization is different across nodes in the cluster]({% link {{ page.version.version }}/cluster-setup-troubleshooting.md %}#disk-utilization-is-different-across-nodes-in-the-cluster).
+
 #### 3-node cluster with 3-way replication
 
 In this scenario, each range is replicated 3 times, with each replica on a different node:
--- a/src/current/v23.2/architecture/replication-layer.md
+++ b/src/current/v23.2/architecture/replication-layer.md
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
 
 #### Load-based replica rebalancing
 
-In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically based on the relative load across the nodes within a cluster. For more information, see the `kv.allocator.load_based_rebalancing` and `kv.allocator.qps_rebalance_threshold` [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}). Note that depending on the needs of your deployment, you can exercise additional control over the location of leases and replicas by [configuring replication zones]({% link {{ page.version.version }}/configure-replication-zones.md %}).
+In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically across the cluster based on a combination of:
+
+1. Replica count.
+1. CPU usage (if [`kv.allocator.load_based_rebalancing.objective`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-load-based-rebalancing-objective) is set to `cpu`, which is the default in CockroachDB v23.1 and later)
+
+Note that disk utilization per node is not one of the rebalancing criteria. For more information, see [Disk utilization is different across nodes in the cluster]({% link {{ page.version.version }}/cluster-setup-troubleshooting.md %}#disk-utilization-is-different-across-nodes-in-the-cluster).
+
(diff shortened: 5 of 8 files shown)


src/current/v23.1/cluster-setup-troubleshooting.md
```diff
@@ -441,6 +441,16 @@ CockroachDB's built-in disk stall detection works as follows:
...
 
+#### Disk usage is different across nodes in the cluster
+
+This is expected behavior.
+
+CockroachDB uses [load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing) to spread [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) across the nodes in a cluster.
+
+The rebalancing criteria for load-based replica rebalancing do not include 'disk usage per node'. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. Larger and smaller ranges balance each other out on average, so disk usage differences across nodes should be relatively small.
+
+However, in some cases a majority of the largest (or smallest) ranges are on one node. Normally that shouldn't be a problem with [sufficiently provisioned storage]({% link {{ page.version.version }}/recommended-production-settings.md %}#storage). If it is causing issues, please [contact Support]({% link {{ page.version.version }}/support-resources.md %}) so they can guide you on how to manually rebalance your cluster's disk usage.
```
kvoli, line 452: Do we mention that there are guards against overfilling a disk somewhere?

The disk usage isn't being balanced, however we will move things around if we hit the thresholds described in https://github.com/cockroachdb/cockroach/pull/97409.

It could be helpful to clarify that, so that users don't see this as a signal we have no protection against full disks.

src/current/v23.1/cluster-setup-troubleshooting.md
```diff
@@ -441,6 +441,16 @@ CockroachDB's built-in disk stall detection works as follows:
 
 - During [node liveness heartbeats](#node-liveness-issues), the [storage engine]({% link {{ page.version.version }}/architecture/storage-layer.md %}) writes to disk as part of the node liveness heartbeat process.
 
+#### Disk usage is different across nodes in the cluster
+
+This is expected behavior.
+
+CockroachDB uses [load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing) to spread [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) across the nodes in a cluster.
+
+The rebalancing criteria for load-based replica rebalancing do not include 'disk usage per node'. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. Larger and smaller ranges balance each other out on average, so disk usage differences across nodes should be relatively small.
```
kvoli, line 450: Do you think we should also make mention of utilization? We strongly don't recommend running with different disk sizes, however the remaining capacity is what users care about -- which is more tightly linked with utilization, than usage.

kvoli (commented):
Thanks Rich! Just a couple questions.

src/current/v23.1/cluster-setup-troubleshooting.md
rmloveland: > Do we mention that there are guards against overfilling a disk somewhere?

I searched the docs and it appears we do not mention any guards against overfilling. We mention [provisioning sufficient capacity](https://www.cockroachlabs.com/docs/v23.1/recommended-production-settings#storage), [monitoring capacity to avoid going over 80%](https://www.cockroachlabs.com/docs/v23.1/common-issues-to-monitor#storage-capacity) and [using emergency ballast files](https://www.cockroachlabs.com/docs/v23.1/cluster-setup-troubleshooting#automatic-ballast-files)

> The disk usage isn't being balanced, however we will move things around if we hit the thresholds described in https://github.com/cockroachdb/cockroach/pull/97409.

Thanks for that info! I have updated this section to add a bulleted list with some information about how the allocator helps protect against full disks - PTAL

Re: the specific cluster settings in https://github.com/cockroachdb/cockroach/pull/97409, they don't appear to be set as "public" and therefore don't appear in the [cluster settings docs page](https://www.cockroachlabs.com/docs/stable/cluster-settings), so I haven't mentioned them explicitly.  I can mention them here if hiding them from the auto-generated docs was not intentional.  We can also leave it like this with a prose description and let the settings be revealed to users through a support interaction if we'd prefer folks not touch them otherwise.

src/current/v23.1/cluster-setup-troubleshooting.md
rmloveland: Ah ok I think I was using them as synonyms but after reading your comment it sounds like:

- usage: "total GiB (or whatever) of storage on disk per node"
- utilization: "percentage of disk capacity used per node"

Is that right?  If so then the latter is what we are actually talking about here, since different disks having different "percent fullnesses" is what was driving the customers to file issues, yes?

I have updated the PR to change "usage" to "utilization", I hope I've used the term correctly, please take a look and let me know if not

src/current/v23.1/cluster-setup-troubleshooting.md
kvoli: The latter option sounds best. These settings are useful but don't have significant guardrails, so the prose option is good.

src/current/v23.1/cluster-setup-troubleshooting.md
kvoli: > Is that right? If so then the latter is what we are actually talking about here, since different disks having different "percent fullnesses" is what was driving the customers to file issues, yes?

Yup

> I have updated the PR to change "usage" to "utilization", I hope I've used the term correctly, please take a look and let me know if not

Will do.

kvoli (approved):
LGTM!

src/current/v23.1/load-based-splitting.md
```diff
@@ -69,6 +69,27 @@ However, by leveraging load-based splitting, the cluster can understand load on
...
+
+The [INFO]({% link {{ page.version.version }}/logging.md %}#info) log message
+
+~~~
+no split key found: ...
+~~~
+
+indicates that CockroachDB wants to [split the range]({% link {{ page.version.version }}/architecture/distribution-layer.md %}#range-splits) due to load, but couldn’t find a key to split at.
+
+Usually this log line can be ignored, unless it repeatedly shows up, which can indicate there is a load imbalance problem in the cluster. If there is a load imbalance problem, it could be because a [hot range]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}) can’t be split (because it's really a [hot key]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}#range-report)).
```
taroface, line 82: I noticed that there are some smart quotes here - they should be replaced with straight quotes afaik.

```suggestion
Usually this log message can be ignored, unless it repeatedly shows up, which can indicate there is a load imbalance problem in the cluster. If there is a load imbalance problem, it could be because a [hot range]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}) cannot be split (because it's really a [hot key]({% link {{ page.version.version }}/ui-hot-ranges-page.md %}#range-report)).
```

src/current/v23.1/load-based-splitting.md
```diff
@@ -69,6 +69,27 @@ However, by leveraging load-based splitting, the cluster can understand load on
...
 
+## Monitor load-based splitting
+
+The [INFO]({% link {{ page.version.version }}/logging.md %}#info) log message
+
+~~~
+no split key found: ...
+~~~
+
+indicates that CockroachDB wants to [split the range]({% link {{ page.version.version }}/architecture/distribution-layer.md %}#range-splits) due to load, but couldn’t find a key to split at.
```
taroface, line 80:
```suggestion
indicates that CockroachDB wants to [split the range]({% link {{ page.version.version }}/architecture/distribution-layer.md %}#range-splits) due to load, but couldn’t find a key to split into smaller ranges.
```

src/current/v23.1/node-shutdown.md
```diff
@@ -290,6 +290,10 @@ To determine an appropriate termination grace period:
 
 Before decommissioning a node, make sure other nodes are available to take over the range replicas from the node. If fewer nodes are available than the replication factor, CockroachDB will automatically reduce the replication factor (for example, from 5 to 3) to try to allow the decommission to succeed. However, the replication factor will not be reduced lower than 3. If three nodes are not available, the decommissioning process will hang indefinitely until nodes are added or you update the zone configurations to use a replication factor of 1.
 
+Note that when you decommission a node and immediately add another node, CockroachDB does *not* simply move all of the replicas from the decommissioned node to the newly added node. Instead, replicas are placed across all nodes in the cluster. This speeds up the decommissioning process by spreading the load. The new node will eventually "catch up" with the rest of the cluster.
```
taroface, line 293: Bolding per our style (ital is for new terms)

```suggestion
Note that when you decommission a node and immediately add another node, CockroachDB does **not** simply move all of the replicas from the decommissioned node to the newly added node. Instead, replicas are placed across all nodes in the cluster. This speeds up the decommissioning process by spreading the load. The new node will eventually "catch up" with the rest of the cluster.
```

taroface (commented):
Some nits plus a couple clarifying questions

src/current/v23.1/architecture/replication-layer.md
```diff
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
 
 #### Load-based replica rebalancing
 
-In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically based on the relative load across the nodes within a cluster. For more information, see the `kv.allocator.load_based_rebalancing` and `kv.allocator.qps_rebalance_threshold` [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}). Note that depending on the needs of your deployment, you can exercise additional control over the location of leases and replicas by [configuring replication zones]({% link {{ page.version.version }}/configure-replication-zones.md %}).
+In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically across the cluster based on a combination of:
+
+1. Replica count
```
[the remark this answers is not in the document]
rmloveland: Fixed

src/current/v23.1/architecture/replication-layer.md
```diff
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+
+For more information on how to control load-based rebalancing, see the following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}):
+
+- [`kv.allocator.store_cpu_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-store-cpu-rebalance-threshold)
+- [`kv.allocator.range_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-range-rebalance-threshold)
+
+Note that disk utilization per node is not one of the rebalancing criteria. For more information, see [Disk utilization is different across nodes in the cluster]({% link {{ page.version.version }}/cluster-setup-troubleshooting.md %}#disk-utilization-is-different-across-nodes-in-the-cluster).
+
+Depending on the needs of your deployment, you can exercise additional control over the location of leases and replicas using [Multi-Region SQL]({% link {{ page.version.version }}/multiregion-overview.md %}), or by [configuring replication zones]({% link {{ page.version.version }}/configure-replication-zones.md %}).
```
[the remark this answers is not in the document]
rmloveland: ✔️

src/current/v23.1/architecture/replication-layer.md
```diff
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+
+Load-based replica rebalancing operates in conjunction with [load-based splitting]({% link {{ page.version.version }}/load-based-splitting.md %}) of [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range).
+
+For more information on how to control load-based rebalancing, see the following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}):
+
+- [`kv.allocator.store_cpu_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-store-cpu-rebalance-threshold)
+- [`kv.allocator.range_rebalance_threshold`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-range-rebalance-threshold)
+
+Note that disk utilization per node is not one of the rebalancing criteria. For more information, see [Disk utilization is different across nodes in the cluster]({% link {{ page.version.version }}/cluster-setup-troubleshooting.md %}#disk-utilization-is-different-across-nodes-in-the-cluster).
```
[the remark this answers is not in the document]
rmloveland: Moved it below the list of criteria - makes sense to me, thanks

src/current/v23.1/architecture/replication-layer.md
```diff
@@ -220,7 +220,21 @@ Rebalancing is achieved by using a snapshot of a replica from the leaseholder, a
...
+In addition to the rebalancing that occurs when nodes join or leave a cluster, replicas are also rebalanced automatically across the cluster based on a combination of:
+
+1. Replica count
+2. CPU usage (if [`kv.allocator.load_based_rebalancing.objective`]({% link {{ page.version.version }}/cluster-settings.md %}#setting-kv-allocator-load-based-rebalancing-objective) is set to `cpu`, which is the default in CockroachDB v23.1 and later)
+
+Load-based replica rebalancing operates in conjunction with [load-based splitting]({% link {{ page.version.version }}/load-based-splitting.md %}) of [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range).
```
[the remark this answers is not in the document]
rmloveland: AIUI they work together:

- the load-based splitting is what cuts up frequently accessed keys into smaller ranges

- while load-based rebalancing is what shuffles the ranges around

src/current/v23.1/cluster-setup-troubleshooting.md
```diff
@@ -441,6 +441,21 @@ CockroachDB's built-in disk stall detection works as follows:
 
 - During [node liveness heartbeats](#node-liveness-issues), the [storage engine]({% link {{ page.version.version }}/architecture/storage-layer.md %}) writes to disk as part of the node liveness heartbeat process.
 
+#### Disk utilization is different across nodes in the cluster
+
+This is expected behavior.
+
+CockroachDB uses [load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing) to spread [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) across the nodes in a cluster.
+
+The rebalancing criteria for load-based replica rebalancing do not include 'percentage of disk utilized per node'. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. Larger and smaller ranges balance each other out on average, so disk utilization differences across nodes should be relatively small.
```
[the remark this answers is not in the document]
rmloveland: SGTM! revised as suggested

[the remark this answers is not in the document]
rmloveland: Ya I think so!

Revised to say

> The proportion of larger and smaller ranges on each node balance each other out on average

src/current/v23.1/cluster-setup-troubleshooting.md
```diff
@@ -441,6 +441,21 @@ CockroachDB's built-in disk stall detection works as follows:
...
+
+However, in some cases a majority of the largest (or smallest) ranges are on one node, which will result in imbalanced utilization. Normally that shouldn't be a problem with [sufficiently provisioned storage]({% link {{ page.version.version }}/recommended-production-settings.md %}#storage). If this imbalance is causing issues, please [contact Support]({% link {{ page.version.version }}/support-resources.md %}) so they can guide you on how to manually rebalance your cluster's disk usage.
```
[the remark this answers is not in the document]
rmloveland: ✔️

src/current/v23.1/load-based-splitting.md
```diff
@@ -69,6 +69,27 @@ However, by leveraging load-based splitting, the cluster can understand load on
 
 This benefit is further amplified by [load-based rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing), which ensures that all nodes contain replicas with a roughly equal load. By evenly distributing load throughout your cluster, it's easier to prevent bottlenecks from arising, as well as simplifying hardware forecasting.
 
+## Monitor load-based splitting
+
+The [INFO]({% link {{ page.version.version }}/logging.md %}#info) log message
```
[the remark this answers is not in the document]
rmloveland: ✔️

src/current/v23.1/load-based-splitting.md
rmloveland: ✔️

src/current/v23.1/load-based-splitting.md
rmloveland: y'know I'm  hesitant to take this one, since I think it's changing the meaning slightly.  I think the key "being split at" is not the same as "the key being split".  

@kvoli can you please verify if this change is ok and/or explain if this is actually a distinction here between "to split at" vs "to split into ranges"?  my sense is there is a difference but not an expert on this

src/current/v23.1/node-shutdown.md
rmloveland: ✔️

rmloveland:
TFTR @taroface ! Just pushed all your recommended changes and hopefully answered the questions (???)

save that one change that I had a question for @kvoli about the meaning of

src/current/v23.1/load-based-splitting.md
taroface: No worries -- fyi I took that wording from our existing https://www.cockroachlabs.com/docs/stable/load-based-splitting

src/current/v23.1/cluster-setup-troubleshooting.md
```diff
@@ -447,9 +447,9 @@ This is expected behavior.
 
 CockroachDB uses [load-based replica rebalancing]({% link {{ page.version.version }}/architecture/replication-layer.md %}#load-based-replica-rebalancing) to spread [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) across the nodes in a cluster.
 
-The rebalancing criteria for load-based replica rebalancing do not include 'percentage of disk utilized per node'. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. Larger and smaller ranges balance each other out on average, so disk utilization differences across nodes should be relatively small.
+The rebalancing criteria for load-based replica rebalancing do not include the percentage of disk utilized per node. Not all [ranges]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-range) are the same size. The proportion of larger and smaller ranges on each node balance each other out on average, so disk utilization differences across nodes should be relatively small.
```
taroface, line 450: Small tweak:

```
The proportions of larger and smaller ranges on each node balance each other out on average
```

taroface (approved):
Just a small tweak, but LGTM!

src/current/v23.1/architecture/replication-layer.md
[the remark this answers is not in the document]
taroface: Oh, in that case I feel that something like "also involves" instead of "operates in conjunction with" is more straightforward and clear here, but I don't feel strongly.

src/current/v23.1/load-based-splitting.md
kvoli: The different between "to split at" and "to split into smaller ranges" is that the "to split at" implies there will be 2 ranges post split (to me at least), while "to split into smaller ranges" could imply >= 2 ranges. The log line is about splitting into 2 ranges (as with all splits), and concerns a single range we are trying to find a key to split on.

"key to split at" is more appropriate here imo, we are talking about a single split, not a series of splits into smaller ranges.

rmloveland: ok thanks for clarifying this! gonna leave it then

src/current/v23.1/cluster-setup-troubleshooting.md
rmloveland: Updated!

rmloveland:
> Just a small tweak, but LGTM!

AWESOME!!!!1!

Just porting these changes to v23.2 docs folder, then merging ...

PR #17702, opened 2023-08-18
DOC-3071: pkg/server: Add last modified timestamp and coordinator ID to /jobs

florence-crl:
Addresses: DOC-3071

Updated Jobs pages for Cloud and DB Consoles, renamed Modified Timestamp to Last Modified Timestamp with updated images.


gemma-shay (approved):
lgtm! sorry for delay

PR #17755, opened 2023-08-25
Update data domiciling docs

rmloveland:
Fixes:

- DOC-7796
- DOC-7812

Summary of changes:

- Update 'Data Domiciling with CockroachDB' as follows:

  - Add a (default) option to use the `ALTER DATABASE ... ADD SUPER REGION` statement

  - Retain the (NON-default) instructions to use `ALTER DATABASE ... PLACEMENT RESTRICTED` in part because those instructions also work and are already vetted.  Also, having more ways to do a thing is usually good.

  - Update the instructions on checking for noncompliant replicas to use the HTTP-based 'critical nodes status endpoint' instead of the now-very-deprecated and non-working SQL replication reports.

  - Also, update the example cluster setup to be a secure cluster, such that using the critical nodes status endpoint shows an example of how to get an authentication cookie (of which there is currently only one in our entire doc set, unless I'm missing something).


src/current/v23.1/data-domiciling.md
```diff
@@ -80,90 +135,238 @@ ALTER DATABASE movr ADD REGION "us-west1";
...
+
+To ensure that data on EU-based users, vehicles, etc. from [`REGIONAL BY ROW` tables]({% link {{ page.version.version }}/regional-tables.md %}#regional-by-row-tables) is stored only on EU-based nodes in the cluster, you can use [Super regions]({% link {{ page.version.version }}/multiregion-overview.md %}#super-regions). This will ensure that [regional tables]({% link {{ page.version.version }}/table-localities.md %}#regional-tables) and [regional by row tables]({% link {{ page.version.version }}/table-localities.md %}#regional-by-row-tables) whose [home regions]({% link {{ page.version.version }}/table-localities.md %}) are in the super region will have all of their [replicas]({% link {{ page.version.version }}/architecture/overview.md %}#architecture-replica) stored _only_ in regions that are members of the super region.
+
+To use this statement, you must set the `enable_super_regions` [session setting]({% link {{ page.version.version }}/set-vars.md %}) or the [`ALTER ROLE ALL SET sql.defaults.multiregion_placement_policy.enabled = on` statement]({% link {{ page.version.version }}/alter-role.md %}#set-default-session-variable-values-for-all-users):
```
rafiss, line 339: nit: this should either say `ALTER ROLE ALL SET enable_multiregion_placement_policy = true` OR say `SET CLUSTER SETTING sql.defaults.multiregion_placement_policy.enabled = true`.

(sql.defaults.multiregion_placement_policy.enabled is the cluster setting which controls the default value for the enable_multiregion_placement_policy session setting)

but also, is `enable_multiregion_placement_policy` even the right setting here? that controls the PLACEMENT syntax. perhaps you meant to use `SET CLUSTER SETING sql.defaults.super_regions.enabled = true`?

rafiss (approved):
i didn't verify each command myself, but all the information and instructions lgtm! just had one nit

src/current/v23.1/data-domiciling.md
rmloveland: Ah you're right the only setting in this example should be `enable_super_regions`, I think this was a copypasta error on my part

I have updated this sentence to recommend doing `ALTER ROLE ALL SET enable_super_regions = on`, and removed mention of that other cluster setting which was incorrect for super regions

(aside: I have also updated this PR to add some simple docs on the session variables page for `enable_super_regions` since we did not have any there (probably b/c this was experimental or something when first added moons ago)

rmloveland: PS thanks for the review @rafiss !!!!

dikshant (changes requested):
Can we also leave a note about users can't store PII in columns that are part of an index if they want data to not be domiciled in incorrect regions.

src/current/v23.1/data-domiciling.md
```diff
@@ -15,10 +15,12 @@ Using CockroachDB as part of your approach to data domiciling has several limita
...
+    1. [`ALTER DATABASE ... PLACEMENT RESTRICTED`]({% link {{ page.version.version }}/alter-database.md %}#placement) statement, which constrains the voting and non-voting replicas for a partition or table to be stored in only the [home regions]({% link {{ page.version.version }}/multiregion-overview.md %}#table-localities) associated with those rows or tables. This statement has the limitation that it will not work for databases that are set up with [region survival goals]({% link {{ page.version.version }}/multiregion-survival-goals.md %}#survive-region-failures).
```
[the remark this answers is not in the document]
rmloveland: SGTM - I added a bunch of **Recommended** and **Not recommended** text to the descriptions of each throughout the document.  I also added a note to the effect that "most users should use ADD SUPER REGION due to limitations of PLACEMENT RESTRICTED.  The documentation for PLACEMENT RESTRICTED is to preserve backwards compatibility" or similar

PTAL and let me know what you think!

src/current/v23.1/data-domiciling.md
```diff
@@ -29,17 +31,70 @@ This page assumes you are already familiar with:
...
+Then, as part of implementing a data domiciling strategy, you will apply restricted replica settings in [Step 4](#step-4-apply-stricter-replica-placement-settings) using one of the following statements:
 
-{{site.data.alerts.callout_info}}
-{% include {{page.version.version}}/sql/super-regions-for-domiciling-with-region-survivability.md %}
-{{site.data.alerts.end}}
+1. [`ALTER DATABASE ... ADD SUPER REGION`]({% link {{ page.version.version }}/alter-database.md %}#add-super-region): If you need to work with databases with [region survival goals]({% link {{ page.version.version }}/multiregion-survival-goals.md %}#survive-region-failures), you must use [super regions]({% link {{ page.version.version }}/multiregion-overview.md %}#super-regions).
+1. [`ALTER DATABASE ... PLACEMENT RESTRICTED`]({% link {{ page.version.version }}/alter-database.md %}#placement): If you need to work with databases with [zone survival goals]({% link {{ page.version.version }}/multiregion-survival-goals.md %}#survive-zone-failures), you can use this statement.
+
```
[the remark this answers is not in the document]
rmloveland: good point.  updated to say that SUPER REGIONS does both, but you must use SUPER REGIONS for region survival

also rephrased PLACEMENT RESTRICTED in a similar way to say that it only works for zone

rmloveland:
> Can we also leave a note about users can't store PII in columns that are part of an index if they want data to not be domiciled in incorrect regions.

AIUI we have such a note, it's the first one in the 'Limitations' section (if I'm understanding you correctly, please let me know if not):

> - When columns are [indexed]({% link {{ page.version.version }}/indexes.md %}), a subset of data from the indexed columns may appear in [meta ranges]({% link {{ page.version.version }}/architecture/distribution-layer.md %}#meta-ranges) or other system tables. CockroachDB synchronizes these system ranges and system tables across nodes. This synchronization does not respect any multi-region settings applied via either the [multi-region SQL statements]({% link {{ page.version.version }}/multiregion-overview.md %}), or the low-level [zone configs]({% link {{ page.version.version }}/configure-replication-zones.md %}) mechanism.

src/current/v23.1/data-domiciling.md
```diff
@@ -80,90 +135,240 @@ ALTER DATABASE movr ADD REGION "us-west1";
...
+{{site.data.alerts.callout_info}}
+[`ALTER DATABASE ... ADD SUPER REGION`]({% link {{ page.version.version }}/alter-database.md %}#add-super-region) does not affect the replica placement for [global tables]({% link {{ page.version.version }}/global-tables.md %}), which are designed to provide fast, up-to-date reads from all [database regions]({% link {{ page.version.version }}/multiregion-overview.md %}#database-regions).
+{{site.data.alerts.end}}
+
+</div>
+
+<div class="filter-content" markdown="1" data-scope="placement-restricted">
+
+**This method is not recommended due to limitations. It is documented for backwards compatibility. Most users should use `ALTER DATABASE ... ADD SUPER REGION`**
```
dikshant, line 363: > **This method is not recommended due to limitations. It is documented for backwards compatibility. Most users should use `ALTER DATABASE ... ADD SUPER REGION`**

Can we make this:

> **This method is not recommended and documented for backwards compatibility. Most users should use `ALTER DATABASE ... ADD SUPER REGION` which provides region survival as well as data placement.**

dikshant (approved):
LGTM! One final comment 🎉 thanks for working on this!

src/current/v23.1/data-domiciling.md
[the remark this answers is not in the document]
dikshant: Thanks!

src/current/v23.1/data-domiciling.md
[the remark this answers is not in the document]
dikshant: Thanks! I left a comment below with a slight tweak. Otherwise LGTM!

src/current/v23.1/data-domiciling.md
rmloveland: Yes, I updated that language ever so slightly to say

> **This method is not recommended, and is documented for backwards compatibility. Most users should use `ALTER DATABASE ... ADD SUPER REGION` which allows for region survival as well as providing data placement.**

shannonbradshaw:
@arulajmani can we please get a review on this? It's been open for more than a month and I'd like to give our users the benefit of these updates.

rmloveland:
FYI the contents of this PR have now been approved by Legal.  I will now take the following steps:

1. update the PR so it will merge
1. port the changes to the v23.2 docs as well
1. merge it

PR #17775, opened 2023-08-29
DOC-8349 cloud storage cluster settings

ashleykwtam:
Addresses DOC-8349: 
- Added additional cluster settings that limit the read and write rates to cloud storage

Preview:
https://deploy-preview-17775--cockroachdb-docs.netlify.app/docs/stable/backup#cloud-storage-cluster-settings

Note:
- looks like the [Cluster Settings](https://www.cockroachlabs.com/docs/v23.1/cluster-settings) list managed by eng does not contain these options 

closed out https://github.com/cockroachdb/docs/pull/17684 and cut another branch due to the v23.2 docs repo cut


src/current/v22.2/backup.md
```diff
@@ -177,6 +177,34 @@ Improve the speed of backups to Azure Storage by increasing `cloudstorage.azure.
 
 **Default:** `1`
 
+#### Cloud storage cluster settings
+
+The following cluster settings limit the read and write rates to [cloud storage]({% link {{ page.version.version }}/use-cloud-storage.md %}). A user may choose to use these settings if their backups overwhelm the network. These settings limit throughput and as a result backups will take longer. The designated `<provider>`s include `s3`, `gs`, and `azure`.
```
dt, line 182: Backups _and changefeeds_ share this limit so perhaps mentioning here when we say it’ll make backups slower

src/current/v22.2/backup.md
```diff
@@ -177,6 +177,34 @@ Improve the speed of backups to Azure Storage by increasing `cloudstorage.azure.
...
 
+#### Cloud storage cluster settings
+
+The following cluster settings limit the read and write rates to [cloud storage]({% link {{ page.version.version }}/use-cloud-storage.md %}). A user may choose to use these settings if their backups overwhelm the network. These settings limit throughput and as a result backups will take longer. The designated `<provider>`s include `s3`, `gs`, and `azure`.     
+
+#### `cloudstorage.<provider>.write.node_rate_limit`
+
+Limit the number of bytes per second per node across operations writing to the designated cloud storage provider if non-zero.
+
+**Default:** `0 B`
```
dt, line 188: Would it be less confusing to call the default unlimited/no-limit ?

src/current/v22.2/backup.md
```diff
@@ -177,6 +177,34 @@ Improve the speed of backups to Azure Storage by increasing `cloudstorage.azure.
...
+
+#### `cloudstorage.<provider>.write.node_burst_limit`
+
+Burst limit the number of bytes per second per node across operations writing to the designated cloud storage provider if non-zero.
```
dt, line 192: Burst is a little different than rate; I think we must have another rate limiter somewhere where we document the burst limit from which we could copy?

dt (approved):

src/current/v22.2/backup.md
ashleykwtam: the data type is byte-size so I think it would make sense to include the `0 B`, but perhaps we can write "unlimited, `0 B`"

src/current/v22.2/backup.md
ashleykwtam: changed burst rate to mention that bytes are handled concurrently

MichaelTrestman (approved):
LGTM

PR #17776, opened 2023-08-29
Add parquet to changefeed format option

kathancox:
Fixes DOC-6799

Adds documentation for the parquet to the `format` option for changefeeds. 
- Added to core and enterprise changefeed SQL pages.
- Updated the exporting with changefeed page + changefeed messages page.


jayshrivastava (approved):
Looks great!

src/current/v23.1/changefeed-for.md
```diff
@@ -82,6 +82,8 @@ Option | Value | Description
 
 #### Avro limitations
 
+Creating a changefeed using Avro is available in Core and [{{ site.data.products.enterprise }} changefeeds](create-changefeed.html) with the [`confluent_schema_registry`](create-changefeed.html#confluent-registry) option.
```
ashleykwtam, line 85: Do we still make mention of Core? Or are we only calling it Self-Hosted now and referring it to `CockroachDB {{ site.data.products.core }}` ?

ashleykwtam (approved):
LGTM, other than the mentions of core!

src/current/v23.1/changefeed-for.md
kathancox: Ah, so this is not referring to Core as in Self-Hosted CockroachDB, it is actually [Core changefeeds](https://www.cockroachlabs.com/docs/v23.1/changefeed-for), which are our implementation of changefeeds that customers do not need an Enterprise license for. Thanks for checking that though!

PR #17778, opened 2023-08-29
Super regions can use *one or more* DB regions

rmloveland:
Fixes DOC-7813


rmloveland:
@dikshant this wording update is pretty small but paired with the toy example in https://github.com/cockroachdb/docs/pull/17755 it will make it pretty clear that super regions can have just one region in them, I think

Please let me know if you disagree tho!

dikshant (approved):
LGTM 🚀. Thanks!

PR #17789, opened 2023-08-30
clarify UPSERT vs. INSERT performance

taroface:
DOC-7998

@mgartner, I did not mention the `insert fast path` behavior with FKs as this seemed somewhat niche compared to the two INSERT vs. UPSERT scenarios. However, let me know if that was wrong and I'll find a place to include it.

cc @DingDLu


src/current/_includes/v23.2/sql/insert-vs-upsert.md
```diff
@@ -1,9 +1,3 @@
...
-without reading, making it faster. This may be particularly useful if
-you are using a simple SQL table of two columns to [simulate direct KV access]({% link {{ page.version.version }}/sql-faqs.md %}#can-i-use-cockroachdb-as-a-key-value-store).
-In this case, be sure to use the `UPSERT` statement.
+When inserting or updating columns on a table that does not have [secondary indexes]({% link {{ page.version.version }}/indexes.md %}), Cockroach Labs recommends using an `UPSERT` statement instead of `INSERT ON CONFLICT DO UPDATE`. Whereas `INSERT ON CONFLICT` always performs a read, the `UPSERT` statement writes without reading, making it faster. This may be useful if you are using a simple SQL table of two columns to [simulate direct KV access]({% link {{ page.version.version }}/sql-faqs.md %}#can-i-use-cockroachdb-as-a-key-value-store).
 
-For tables with secondary indexes, there is no performance difference between `UPSERT` and `INSERT ON CONFLICT`.
+If the table has a secondary index, there is no performance difference between `UPSERT` and `INSERT ON CONFLICT`. However, `INSERT` may not scan the table for existing values. This can provide a performance improvement over `UPSERT`.
```
mgartner, line 3: nit: Maybe make it super clear that the second `INSERT` lacks `ON CONFLICT`, like:

> However, `INSERT` without an `ON CONFLICT` clause may not scan...

src/current/v23.1/insert.md
```diff
@@ -55,10 +47,7 @@ rather than inferring one using a column list, which is the default behavior.
 
 ### `INSERT ON CONFLICT` vs. `UPSERT`
 
-As a short-hand alternative to the `ON CONFLICT` clause, you can use the [`UPSERT`]({% link {{ page.version.version }}/upsert.md %})
-statement. However, `UPSERT` does not let you specify the column(s) with
-the unique constraint; it always uses the column(s) from the primary
-key. Using `ON CONFLICT` is therefore more flexible.
+As a short-hand alternative to the `ON CONFLICT` clause, you can use the [`UPSERT`]({% link {{ page.version.version }}/upsert.md %}) statement. However, `UPSERT` does not let you specify the column(s) with the unique constraint; it always uses the column(s) from the primary key. Using `ON CONFLICT` is therefore more flexible.
```
mgartner, line 50: nit: It's short-hand for `INSERT .. ON CONFLICT .. DO UPDATE ..`, specifically when all the insert columns are listed, and the have the same values assigned in the `DO UPDATE` clause. For example:

```sql
INSERT INTO t(a, b, ..., n)
VALUES ('1', '2', ..., 'n')
ON CONFLICT DO UPDATE SET
  a = '1',
  b = '2',
  ...
  n = 'n';

-- is equivalent to:

UPSERT INTO t(a, b, ..., n)
VALUES ('1', '2', ..., 'n');
```

mgartner: > does not let you specify the column(s) with the unique constraint

This is a little unclear to me. I think it could be better worded like:

> does not let you specify columns to infer a unique constraint as an arbiter; it always uses the primary key as the arbiter

This might require defining the term "arbiter". An arbiter is a unique constraint used to check for conflicts during execution of `INSERT ON CONFLICT`. `UPSERT` always uses the primary index as the arbiter.

mgartner (approved):
LGTM, just a few optional nits.

src/current/v23.1/insert.md
taroface: Thank you! I used this to clarify the UPSERT and the INSERT docs.

Amruta-Ranade (approved):

mgartner (approved):

PR #17795, opened 2023-08-31
Introduce improve changefeed latency section

kathancox:
Fixes DOC-7906

This PR adds a section to the Advanced Changefeed Configuration page around cluster settings that customers can tweak in order to improve changefeed latency.

- three cluster settings to use in combination
- some detail for each and a brief recommendation on what to change
- warnings around how adjusting these settings may affect transactions

### DIFF
--- a/src/current/v23.1/advanced-changefeed-configuration.md
+++ b/src/current/v23.1/advanced-changefeed-configuration.md
@@ -6,7 +6,7 @@ docs_area: stream_data
 ---
 
 {{site.data.alerts.callout_danger}}
-The configurations and settings explained on this page will have a significant impact on a changefeed's behavior and could potentially affect a cluster's performance. Thoroughly test before deploying any changes to productions.
+The configurations and settings explained on this page will have a significant impact on a changefeed's behavior and could potentially affect a cluster's performance. Thoroughly test before deploying any changes to production.
 {{site.data.alerts.end}}
 
 The following sections describe performance, settings, configurations, and details to tune [changefeeds]({% link {{ page.version.version }}/change-data-capture-overview.md %}):
--- a/src/current/v23.1/advanced-changefeed-configuration.md
+++ b/src/current/v23.1/advanced-changefeed-configuration.md
@@ -33,38 +33,41 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
 
 For a more technical explanation of elastic CPU, refer to the [Rubbing control theory on the Go scheduler](https://www.cockroachlabs.com/blog/rubbing-control-theory/) blog post.
 
-### Improving latency
+### Latency in changefeeds
 
-Latency in changefeeds is the sum of the time it takes to:
+When you are running large workloads, changefeeds can encounter or cause latency in a cluster in the following ways:
 
-- Commit writes to the database.
-- Encode [changefeed messages]({% link {{ page.version.version }}/changefeed-messages.md %}).
-- Deliver the message to the [sink]({% link {{ page.version.version }}/changefeed-sinks.md %}).
+- Changefeeds can have an impact on SQL latency in the cluster generally.
+- Changefeeds can encounter latency in **events** emitting. This latency is the total time CockroachDB takes to:
+    - Commit writes to the database.
+    - Encode [changefeed messages]({% link {{ page.version.version }}/changefeed-messages.md %}).
+    - Deliver the message to the [sink]({% link {{ page.version.version }}/changefeed-sinks.md %}).
 
-To improve changefeed latency, you can use the following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}) to reduce bursts of rangefeed work so that updates are paced steadily over time. It is necessary to adjust these cluster settings together in order to have an effect on changefeed latency. However, note that these can have an effect on running transactions, and should be adjusted and tested with your particular workload.
+The following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}) reduce bursts of [rangefeed]({% link {{ page.version.version }}/create-and-configure-changefeeds.md %}#enable-rangefeeds) work so that updates are paced steadily over time.
+
+{{site.data.alerts.callout_danger}}
+We do **not** recommend adjusting these settings unless you are running a large workload, or are working with the Cockroach Labs [support team]({{ link_prefix }}support-resources.html).
+{{site.data.alerts.end}}
 
 #### `kv.closed_timestamp_target_duration`
 
-Each [range]({% link {{ page.version.version }}/architecture/overview.md %}#range) in CockroachDB tracks a property called the [_closed timestamp_]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps), which means that no new writes can ever be introduced at or below that timestamp. Decreasing the target duration can reduce changefeed latency, so that the changefeed job can [checkpoint]({% link {{ page.version.version }}/how-does-an-enterprise-changefeed-work.md %}) more frequently and deliver messages to the sink.
+**Default:** `3s`
 
-Default: `3s`
-Recommendation: Decreasing to `1s`–`2s`.
+This setting controls the frequency of checkpoints for each [range]({% link {{ page.version.version }}/architecture/overview.md %}#range). A changefeed aggregates these checkpoints across all ranges, and once the timestamp on all the ranges advances, the changefeed can then [checkpoint]({% link {{ page.version.version }}/how-does-an-enterprise-changefeed-work.md %}). As a result, the higher the value of this setting the longer it can take for a changefeed to checkpoint. It is important to note that a changefeed at default configuration does not checkpoint more often than once every 30 seconds.
+
+In clusters running large-scale workloads, increasing this setting can help to lower the potential impact of changefeeds on SQL latency. That is, an increase in the setting could lower the load on the cluster. This is important for workloads with tables in the TB range of data. However, for most workloads, we recommend leaving this setting at the default of `3s`.
 
 {{site.data.alerts.callout_danger}}
-Thoroughly test any decrease to the closed timestamp interval before deploying the change in production, because such an adjustment can cause [contention]({% link {{ page.version.version }}/performance-best-practices-overview.md %}#understanding-and-avoiding-transaction-contention) between transactions.
+Thoroughly test any adjustment in cluster settings before deploying the change in production.
 {{site.data.alerts.end}}
-
-#### `kv.closed_timestamp.side_transport_interval`
-
-When there is a period of no write activity, the _side-transport_ is a node-level process that fixes a timestamp to close for groups of ranges. This cluster setting adjusts how often the side-transport process will attempt to move the [closed timestamp]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps) forward. The smaller the interval, the more frequently the side-transport will attempt to move the timestamp forward.
-
-This will also affect changefeed checkpoints. That is, if the interval is decreased the changefeed will be able to read and checkpoint more frequently, rather than collecting changes into fewer, less frequent checkpoints. To improve changefeed latency, we recommend decreasing the `kv.closed_timestamp.side_transport_interval` cluster setting.
-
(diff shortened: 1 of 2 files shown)


kathancox:
Hey @amruss @miretskiy, could you take a look at this PR and check out the accuracy here on the settings descriptions? As well as anything else we might want to mention?

kathancox:
Hey @amruss @miretskiy, just a friendly reminder ping on this one, thanks!

src/current/v23.1/advanced-changefeed-configuration.md
```diff
@@ -33,6 +33,39 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
...
+- Deliver the message to the [sink]({% link {{ page.version.version }}/changefeed-sinks.md %}).
+
+To improve changefeed latency, you can use the following [cluster settings]({% link {{ page.version.version }}/cluster-settings.md %}) to reduce bursts of rangefeed work so that updates are paced steadily over time. It is necessary to adjust these cluster settings together in order to have an effect on changefeed latency. However, note that these can have an effect on running transactions, and should be adjusted and tested with your particular workload.
+
+#### `kv.closed_timestamp_target_duration`
+
+Each [range]({% link {{ page.version.version }}/architecture/overview.md %}#range) in CockroachDB tracks a property called the [_closed timestamp_]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps), which means that no new writes can ever be introduced at or below that timestamp. Decreasing the target duration can reduce changefeed latency, so that the changefeed job can [checkpoint]({% link {{ page.version.version }}/how-does-an-enterprise-changefeed-work.md %}) more frequently and deliver messages to the sink.
+
```
miretskiy, line 49: This section is mostly incorrect -- sorry, @kathancox .
It's not on you, it's just these
settings have been the source of great many problems in the past, and continue to be very problematic.

Recently, the default on master was bumped to 3s (from old value of 0 -- which then used the value of the setting below -- the 200ms `kv.closed_timestamp.side_transport_interval` setting).  We have some customers who run with even higher values.

My first recommendation is: don't touch this setting unless you know what you're doing (more on this below).

My second recommendation is to never set it low.  If this setting was ever altered by the customer, I recommend they reset this setting value to 3s or higher.

So, what does this setting do? it controls the frequency of checkpoints for each range.  Changefeed aggregates those checkpoints across *all* ranges, and once all ranges advance -- then the changefeed can checkpoint.  Therefore, the higher the value of this setting the longer it could take for changefeed to checkpoint.  It is important to understand that in its default configuration changefeed does not checkpoint more often than once every 30 seconds.    So, by setting this setting to be low, you're actually not making changefeed do anything other than adding more load on your cluster.

As mentioned above, some customers set this setting to be as high as 30s.  And the reason why you may want to do that is to attempt to lower the impact of changefeeds on SQL latency.  The reasons are rather deep and technical, but I think we can comfortably say that increasing this setting will reduce the load on the cluster and, conversely,  decreasing it will do exact opposite.
This is particularly important for very large scale customers, and I think we should call these out.  Basically, unless you're in TB range, there is no reason for you to touch this setting. If you have very large tables, consider *increasing* this setting.


I do not recommend it being set higher than say 20-30s.  In all honesty, the new default of 3 seconds is pretty good.

Next, we need to understand, and *fix* misunderstanding that somehow this setting impacts changefeed latency.  As I explained above, this setting *could* impact how often changefeed can checkpoints.  It could impact how often changefeed will emit "resolved" events.  But this does *not* impact how long it will take for changefeed to emit real event (an update).  I suspect the reason for this is the "max changefeed latency" graph we had for very long time where the time since the last checkpoint was treated as "latency" -- and let's be very clear: it is *not* event latency.

src/current/v23.1/advanced-changefeed-configuration.md
```diff
@@ -33,6 +33,39 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
...
+{{site.data.alerts.end}}
+
+#### `kv.closed_timestamp.side_transport_interval`
+
+When there is a period of no write activity, the _side-transport_ is a node-level process that fixes a timestamp to close for groups of ranges. This cluster setting adjusts how often the side-transport process will attempt to move the [closed timestamp]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps) forward. The smaller the interval, the more frequently the side-transport will attempt to move the timestamp forward.
+
+This will also affect changefeed checkpoints. That is, if the interval is decreased the changefeed will be able to read and checkpoint more frequently, rather than collecting changes into fewer, less frequent checkpoints. To improve changefeed latency, we recommend decreasing the `kv.closed_timestamp.side_transport_interval` cluster setting.
+
+For more information on the side-transport interval, refer to [An epic read on follower reads](https://www.cockroachlabs.com/blog/follower-reads-stale-data/) on the Cockroach Labs Blog.
+
```
miretskiy, line 64: I think we should change our recommendation with respect to this setting.

Do not touch this setting.
Set the `kv.closed_timestamp_target_duration` to 3 seconds if it is 0 (that's the old default, which fell back to this setting).

It was a mistake (corrected on master, and backported, I think) to conflate these 2 settings together.  One is purely about rangefeed checkpoints.  This one IS about follower reads.  You certainly don't want to make your follower read queries slower because you're trying to tune changefeeds.

src/current/v23.1/advanced-changefeed-configuration.md
```diff
@@ -33,6 +33,39 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
...
+#### `kv.rangefeed.closed_timestamp_smear_interval`
+
+This setting provides a mechanism to pace the [closed timestamp]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps) notifications to follow replicas. Decreasing the closed timestamp smear interval makes rangefeed closed timestamp delivery less spiky, which can reduce its impact on foreground SQL query latency.
```
miretskiy, line 67: I think we should mention that the default is 1ms; and that's reasonably.
It can be set to higher values, but anything higher than 20ms probably will not matter.

I also think we should expand this section to include an explanation of what this setting actually does.  Let's say you have a fairly large table, and one of the nodes in the cluster is hosting 5000 ranges from this table.  Normally, the the rangefeed system will wake up every `kv.closed_timestamp_target_duration` (say 3s) and every 3 seconds it'll publish checkpoints for all 6000 ranges.  What this setting says is: take 3s frequency and divide it into 1ms chunks -- there are 3000 of those chunks. And instead of publishing checkpoints for all 6000 ranges, publish checkpoints for 2 ranges -- but do this every 1 millisecond.

This sounds like a lot of work, but what this does is that it generates smooth and predictable load, instead of spiky, large (6000 ranges!) load.

miretskiy (changes requested):

src/current/v23.1/advanced-changefeed-configuration.md
kathancox: That's OK, thanks for the detail. There wasn't much information on this around, so it is not surprising this needs some adjustment!

I have synthesized your explanation about this setting into the section, and not recommended any change unless customers have large workloads. I also adapted the intro to the overall section to mention how we're not just talking about changefeed event latency. 

Does these changes but reflect what we want here?

src/current/v23.1/advanced-changefeed-configuration.md
kathancox: I have actually removed the `side_transport_interval` setting with this change. It seems like we don't really want to talk about this here? Correct me though, if we do actually want to say something about it here.

src/current/v23.1/advanced-changefeed-configuration.md
kathancox: I have included your example here — I think this is a good illustration. What do you think?

kathancox:
@miretskiy I have gone through your feedback and made some changes. Let me know what you think. I think it has made the section more of an "about" these settings rather than actively calling on the user to change them, which seems maybe more sensible — but, let me know your thoughts!

Thanks for the detailed feedback, I do appreciate it.

miretskiy (approved):
Very nice!

src/current/v23.1/advanced-changefeed-configuration.md
miretskiy: It's great.

src/current/v23.1/advanced-changefeed-configuration.md
miretskiy: Yes; that's great.

src/current/v23.1/advanced-changefeed-configuration.md
```diff
@@ -33,6 +33,42 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
...
+Thoroughly test any adjustment in cluster settings before deploying the change in production.
+{{site.data.alerts.end}}
+
+#### `kv.rangefeed.closed_timestamp_smear_interval`
+
+**Default:** `1ms`
+
+This setting provides a mechanism to pace the [closed timestamp]({% link {{ page.version.version }}/architecture/transaction-layer.md %}#closed-timestamps) notifications to follower replicas. Decreasing the closed timestamp smear interval makes rangefeed closed timestamp delivery less spiky, which can reduce its impact on foreground SQL query latency.
+
+For example, if you have a large table, and one of the nodes in the cluster is hosting 6000 ranges from this table. Normally, the rangefeed system will wake up every `kv.closed_timestamp_target_duration` (default `3s`) and every 3 seconds it will publish checkpoints for all 6000 ranges. In this scenario, the `kv.rangefeed.closed_timestamp_smear_interval` setting takes the `3s` frequency and divides it into `1ms` chunks. Instead of publishing checkpoints for all 6000 ranges, it will publish checkpoints for 2 ranges every `1ms`. This produces a more predictable and level load, rather than spiky, large bursts of workload.
```
ashleykwtam, line 70: nit:

```suggestion
For example, suppose your cluster has a large table, and one of the nodes in the cluster is hosting 6000 ranges from this table. Normally, the rangefeed system will wake up every `kv.closed_timestamp_target_duration` (default `3s`) and every 3 seconds it will publish checkpoints for all 6000 ranges. In this scenario, the `kv.rangefeed.closed_timestamp_smear_interval` setting takes the `3s` frequency and divides it into `1ms` chunks. Instead of publishing checkpoints for all 6000 ranges, it will publish checkpoints for 2 ranges every `1ms`. This produces a more predictable and level load, rather than spiky, large bursts of workload.
```

src/current/v23.1/advanced-changefeed-configuration.md
```diff
@@ -33,6 +33,42 @@ kvadmission.rangefeed_catchup_scan_elastic_control.enabled
...
+
+#### `kv.closed_timestamp_target_duration`
+
+**Default:** `3s`
+
+This setting controls the frequency of checkpoints for each [range]({% link {{ page.version.version }}/architecture/overview.md %}#range). A changefeed aggregates these checkpoints across all ranges, and once the timestamp on all the ranges advances, the changefeed can then [checkpoint]({% link {{ page.version.version }}/how-does-an-enterprise-changefeed-work.md %}). As a result, the higher the value of this setting the longer it can take for a changefeed to checkpoint. It is important to note that a changefeed at default configuration does not checkpoint more often than once every 30 seconds.
+
+In clusters running large-scale workloads, increasing this setting can help to lower the potential impact of changefeeds on SQL latency. That is, an increase in the setting could lower the load on the cluster. This is important for workloads with tables in the TB range of data. However, for most workloads, we recommend leaving this setting at the default of `3s`.
+
+{{site.data.alerts.callout_danger}}
```
ashleykwtam, line 60: should this warning be placed after both cluster settings?

ashleykwtam (approved):
LGTM, other than a small nit!

src/current/v23.1/advanced-changefeed-configuration.md
kathancox: This is a valid alternative! Totally a preference thing, I just find that "suppose" sounds a bit colloquial / dialog vocab like.  I'm going to keep as is for now, but certainly open to discussion on this.

src/current/v23.1/advanced-changefeed-configuration.md
kathancox: This one warning I kept to the upper setting because the lower one we really don't anticipate customers changing. I have altered part of the language in that setting's description to ensure it's a little clearer that the default is actually good here.

PR #17834, opened 2023-09-08
Include missing options to core changefeed sql page

kathancox:
Fixes DOC-8692

Adds options to the `EXPERIMENTAL CHANGEFEED FOR` options table, that work with the statement and were missing from the table.


kathancox:
While this work still needs to be done, this PR is now very out of date and lower priority. Closing this work for now

PR #17869, opened 2023-09-12
DOC-8571 change backup process to not send requests for spans

ashleykwtam:
Addresses DOC-8571, DOC-8629:
- Added a note on the backup process as backups now skip contacting the ranges for tables on which `exclude_data_from_backup is set`, and can thus succeed even if an excluded table is unavailable

Preview:
https://deploy-preview-17869--cockroachdb-docs.netlify.app/docs/stable/take-full-and-incremental-backups#exclude-a-tables-data-from-backups


src/current/v23.1/take-full-and-incremental-backups.md
```diff
@@ -283,7 +283,9 @@ For more examples on how to schedule backups that take full and incremental back
...
 
-It is important to note that the backup will still contain the table, but it will be empty. Setting this parameter prevents the cluster or database backup from delaying [GC TTL]({% link {{ page.version.version }}/configure-replication-zones.md %}#gc-ttlseconds) on the key span for this table, and it also respects the configured GC TTL. This is useful when you want to set a shorter garbage collection window for tables containing high-churn data to avoid an accumulation of unnecessary data.
+It is important to note that the backup responds with an empty table, excluding the data. The backup process does not send requests for spans from excluded tables when conducting database, table, or cluster backups. Tenant backups will dispatch export requests to every range for the tenant span, and the ranges that host excluded tables will persist in responding with no data. The backups will succeed even when a table is unavailable, if, and only if, that table is excluded.
```
dt, line 286: I don't know if "responds" is quite what I'd say, since that could be ambiguous as to whether we're talking about the _response_ to the backup statement or the backup produced.

maybe "the backup produced contains" ? 

I do wonder if "does not send requests for spans" might be in lower level terms than most users reading this will be understanding? Perhaps the backup process "does not contact the KV storage layer to retrieve the contents of excluded..." ?

Tenant backups are not currently user-facing features, so I think we can drop the penultimate sentence entirely. 

So I might change this paragraph to:
> A a table which excludes its data from backups will still appear in backups, for example of the cluster or of the database that contains it, but will appear empty. The backup process will not attempt to contact KV storage layer ranges to retrieve data for tables which are excluded from backup. Thus backups may continue to succeed even when some ranges or regions are unavailable if and only if the unavailable ranges are in tables which are excluded from backups.

dt (approved):

MichaelTrestman (approved):
LGTM

dt (approved):

PR #17871, opened 2023-09-12
Azure upgrades

gemma-shay:
https://cockroachlabs.atlassian.net/browse/DOC-7937

Guessing at Oct 1 release but can update
Is it accurate that Azure clusters cannot currently be upgraded?


mdlinville (commented):
I think you also need to remove the relevant limitations from `cockroachdb-dedicated-on-azure.md`.

gemma-shay:
@mdlinville i'm not seeing anything related to upgrades/maintenance on https://www.cockroachlabs.com/docs/cockroachcloud/cockroachdb-dedicated-on-azure

mdlinville:
> @mdlinville i'm not seeing anything related to upgrades/maintenance on https://www.cockroachlabs.com/docs/cockroachcloud/cockroachdb-dedicated-on-azure

I woonder if it is worth adding "but can be upgraded" after "can't be modified" here.

PR #17882, opened 2023-09-13
Update the cdc prev example to select columns and filter rows

kathancox:
Fixes DOC-8800

Updates the `cdc_prev` examples on the CDC queries page, so that they do not just `SELECT *`. The reason for this is that customers can use a regular changefeed with the `diff` option to achieve the same thing, if they do not need filtering.

The updated examples better fit with how we recommend customers to use either CDC queries and regular changefeeds.


src/current/v23.1/cdc-queries.md
```diff
@@ -174,18 +174,22 @@ To emit the previous state of a row, it is necessary to explicitly call `cdc_pre
...
-CREATE CHANGEFEED INTO 'external://sink' AS SELECT *, cdc_prev FROM movr.rides;
+CREATE CHANGEFEED INTO 'external://sink' AS SELECT rider_id, vehicle_id, cdc_prev FROM movr.rides;
 ~~~
 
 To emit the previous state of a column, you can specify this as a named field from the `cdc_prev` tuple with the following syntax:
 
 {% include_cached copy-clipboard.html %}
 ~~~ sql
-CREATE CHANGEFEED INTO 'external://sink' AS SELECT *, cdc_prev FROM movr.vehicles WHERE (cdc_prev).status = 'in_use';
+CREATE CHANGEFEED INTO 'external://sink' AS SELECT owner_id, current_location, cdc_prev FROM movr.vehicles WHERE (cdc_prev).status = 'in_use';
```
miretskiy, line 184: perhaps SELECT (cdc_prev).owner_id as old_owner_id, (cdc_prev).status as old_status to make it clear that you don't need to select the entire "cdc_prev" and can also limit it to the columns that are interesting.

It's okay if you think that omitting this makes it clearer.

miretskiy (approved):

ashleykwtam (approved):
LGTM!

src/current/v23.1/cdc-queries.md
kathancox: I have updated — it seems a better fit here for the example to only request the `cdc_prev` data on a column.

kathancox:
TFTRs!

